PR #1297, opened 2018-11-28
Change NotifyUsers to allow only data JSON element

pedropombeiro:
Required for status-im/status-react#6772

This PR moves from using Notifications API to Messaging API (where only key/value pairs are sent in the message, and the message must be processed by the recipient app before being displayed).
The most important part is setting the message priority to `High`, so that Firebase can invoke the background task on the recipient phone to decode and display the push notification.

I'm also patching the `go-fcm` package (in https://github.com/status-im/go-fcm/commit/024ca6a2c5444c93980f558f91c35a2defebd362) to return the actual error from FCM if one exists (the current implementation was silently ignoring errors). It could have saved me hours of debugging which only ended when I finally decided to try the same request with `curl` and saw the reason for the error. It now outputs this:

```
Status Code   : 400
Success       : 0
Fail          : 0
Canonical_ids : 0
Topic MsgId   : 0
Topic Err     : "data" key "from" is a reserved keyword
```


ghost:
### Pull Request Checklist

notifications/push/fcm/notification.go
```diff
@@ -25,30 +26,35 @@ func NewNotification(key string) NotificationConstructor {
...
 	}
 
-	fmt.Println(payload.Title, payload.Body)
-
-	n.client.NewFcmRegIdsMsg(tokens, data)
-	n.client.SetNotificationPayload(&payload)
-	_, err := n.client.Send()
+	n.client.NewFcmRegIdsMsg(tokens, dataPayload)
+	resp, err := n.client.Send()
+	if err == nil {
```
dshulyak: nil will be returned on L59. you need either to remove error instanation of the new err variable on L49 or restructure the code.

i suggest to follow common structure, if err != nil {return err} and then marshal json

```go
var err error
if err == nil {
	err := errors.New("1")
	fmt.Println(err)
}
fmt.Println(err)

1  
<nil>
```

notifications/push/fcm/notification.go
```diff
@@ -25,30 +26,35 @@ func NewNotification(key string) NotificationConstructor {
 		client := fcm.NewFcmClient(key).
 			SetDelayWhileIdle(true).
 			SetContentAvailable(true).
+			SetPriority(fcm.Priority_HIGH).
```
dshulyak: makes sense to mention it in PR description, what kind of effect it will have?

api/backend.go
```diff
@@ -446,10 +444,10 @@ func (b *StatusBackend) SelectAccount(address, password string) error {
 }
 
 // NotifyUsers sends push notifications to users.
-func (b *StatusBackend) NotifyUsers(message string, payload fcmlib.NotificationPayload, tokens ...string) error {
-	err := b.newNotification().Send(message, payload, tokens...)
+func (b *StatusBackend) NotifyUsers(dataPayloadJSON string, tokens ...string) error {
+	err := b.newNotification().Send(dataPayloadJSON, tokens...)
 	if err != nil {
-		b.log.Error("Notify failed", "error", err)
+		b.log.Error("NotifyUsers failed", "error", err, "dataPayloadJSON", dataPayloadJSON)
```
dshulyak: it is common to print error as a last value

notifications/push/fcm/notification.go
```diff
@@ -25,30 +26,35 @@ func NewNotification(key string) NotificationConstructor {
...
 }
 
-// Send send to the tokens list.
-func (n *Notification) Send(body string, payload fcm.NotificationPayload, tokens ...string) error {
-	data := map[string]string{
-		"msg": body,
-	}
+// Send sends a push notification to the tokens list.
+func (n *Notification) Send(dataPayloadJSON string, tokens ...string) error {
+	log.Debug("calling Firebase Send", "dataPayloadJSON", dataPayloadJSON)
```
dshulyak: i like the payload name more, it is obvious what it is, not sure if JSON part is that important

notifications/push/fcm/notification.go
pedropombeiro: I just realized I forgot to clean up this code that I used for debugging why the notifications weren't reaching the react side. I'll do that now, sorry about that.

notifications/push/fcm/notification.go
pedropombeiro: Done, I'll also add a comment in the code. It allows Firebase to call the new background task in the Android app which will take care of displaying the PN.

notifications/push/fcm/notification.go
pedropombeiro: The JSON in the name is to inform the callers that the string will be interpreted (unmarshalled) as a JSON object.

pedropombeiro:
Addressed PR comments and moved to status-im fork of `go-fcm`

Gopkg.toml
```diff
@@ -35,6 +35,10 @@
   name = "github.com/golang/protobuf"
   version = "1.1.0"
 
+[[constraint]]
+  name = "github.com/status-im/go-fcm"
+  revision = "024ca6a2c5444c93980f558f91c35a2defebd362"
```
dshulyak, line 40: it was possible to keep modules at the same path, like for geth, not sure if it makes sense 

```
   name = "github.com/ethereum/go-ethereum"
   version = "=v1.8.17"
   source = "github.com/status-im/go-ethereum"
```

dshulyak (approved):

Gopkg.toml
adambabik: :+1:

In this case, it does work because `go-fcm` does not contain any "submodules". So we either need to remember to change import paths appropriately in our fork or apply this suggestion.

lib/library.go
```diff
@@ -417,7 +416,7 @@ func makeJSONResponse(err error) *C.char {
 
 // NotifyUsers sends push notifications by given tokens.
 //export NotifyUsers
-func NotifyUsers(message, payloadJSON, tokensArray *C.char) (outCBytes *C.char) {
+func NotifyUsers(dataPayloadJSON, tokensArray *C.char) (outCBytes *C.char) {
```
adambabik, line 419: Breaking change. There should be a bump of minor in `VERSION` in this PR.

notifications/push/fcm/notification.go
```diff
@@ -1,14 +1,16 @@
 package fcm
 
 import (
+	"encoding/json"
 	"fmt"
 
-	"github.com/NaySoftware/go-fcm"
+	"github.com/ethereum/go-ethereum/log"
+	fcm "github.com/status-im/go-fcm"
```
adambabik: `fcm` not needed.

notifications/push/fcm/notification.go
```diff
@@ -25,30 +27,35 @@ func NewNotification(key string) NotificationConstructor {
...
+		log.Error("error calling Firebase Send", "err", err.Error())
+		return err
+	}
 
-	n.client.NewFcmRegIdsMsg(tokens, data)
-	n.client.SetNotificationPayload(&payload)
-	_, err := n.client.Send()
+	if !resp.Ok {
+		err = fmt.Errorf("failed to send notification: code=%d err=%s", resp.StatusCode, resp.Err)
+		resp.PrintResults()
```
adambabik: Is it possible to wrap it in `log.Error`?

notifications/push/fcm/notification.go
```diff
@@ -25,30 +27,35 @@ func NewNotification(key string) NotificationConstructor {
...
+	if err != nil {
+		return err
 	}
 
-	fmt.Println(payload.Title, payload.Body)
+	n.client.NewFcmRegIdsMsg(tokens, dataPayload)
+	resp, err := n.client.Send()
+	if err != nil {
+		log.Error("error calling Firebase Send", "err", err.Error())
+		return err
```
adambabik, line 47: I am not sure who call it but usually logging is in higher abstract layers. Here I would only do:
```go
return fmt.Errorf("failed to send notification: %v", err)
```

and do logging in the caller.

Gopkg.toml
pedropombeiro: I wasn't aware of the `source` functionality, updated PR with it, thanks :+1:

notifications/push/fcm/notification.go
pedropombeiro: I think this was added automatically by my editor... Removed.

notifications/push/fcm/notification.go
pedropombeiro: It's not, but with the new FCM form we don't actually need this anymore as we surface the error message in the error object, so I'll remove this line.

notifications/push/fcm/notification.go
pedropombeiro: Good point. Sorry about the state of this PR, since the big one is on the react side, I guess I didn't pay as much attention to this one as I should have.

pedropombeiro:
@adambabik are you OK with the current state of the PR?

adambabik (approved):

lib/library.go
cammellos: Breaking change on a public interface is a major bump, not a minor, let's stick to https://semver.org/ (bar the fact that we have not yet reached 1.0.0), I have yet again builds failing for this and takes some time to figure out

pedropombeiro: Good point, I'll create a PR to bump the version to 0.18.0

PR #1298, opened 2018-11-30
Notify user when the device is missing

cammellos:
A few things:

1) We notify `status-react` only when the message is not targeting our device
2) Shorten the time we refresh the bundle, faster turn over is more secure and gives us faster feedback
3) Notify of added bundles even if the message is not sent to this device


services/shhext/api.go
```diff
@@ -418,9 +418,16 @@ func (api *PublicAPI) processPFSMessage(msg *whisper.Message) error {
 
 	handler := EnvelopeSignalHandler{}
 
+	// Notify of added bundles
+	if response != nil && response.AddedBundles != nil {
```
adambabik: Any reason why this logic was moved here? Generally, checking the `err` should be done first and if it's not `nil`, processing other return values should not be continued as it's usually unclear if these values are correct.

If it's possible that `AddedBundles` has a correct value even though `err` is not `nil`, I would think about better interface.

cammellos: > If it's possible that AddedBundles has a correct value even though err is not nil, I would think about better interface.

Correct, that might be the case, is there an idiomatic way to express this?

A few things I can think of:

1) Split the `HandleMessage` in 2 methods, `HandleBundles` & `HandleMessage`, but I feel that breaks encapsulation (i.e. the caller should not care about calling two functions to handle a single message)
2) Pass a function as  a parameter, say `onSucessfulBundleProcessed`
3) change the struct to embeds the errors, but then we would not be using the second argument, which is probably weird
4) Return a slice of errors, with one being the `message` error & one the `bundle` errors

Basically the problem is essentially this one:
https://www.reddit.com/r/golang/comments/58kt6u/returning_multiple_errors_as_a_slice/

I have an sequence of operations to be performed independently, `addingBundles`, `decryptMessage`, but not to break encapsulation I want them to be performed in the same function `handleMessage` and be notified of the success/failure of both.

What do you reckon?

pedropombeiro: I would vote for 2) but would like to hear @adambabik's opinion.

services/shhext/chat/encryption.go
```diff
@@ -238,8 +239,8 @@ func (s *EncryptionService) DecryptPayload(myIdentityKey *ecdsa.PrivateKey, thei
 	}
 
 	// We should not be sending a signal if it's coming from us, as we receive our own messages
-	if msg == nil {
-		return nil, ErrSessionNotFound
+	if msg == nil && myIdentityKey.PublicKey != *theirIdentityKey {
```
pedropombeiro: I feel it's more readable if we switch the operands in the comparison:

```suggestion
	if msg == nil && *theirIdentityKey != myIdentityKey.PublicKey {
```

pedropombeiro (changes requested):

services/shhext/api.go
adambabik: One more way to look at this is whether the caller needs to care about new bundles when handling a message. Can `chat.ProtocolService` have a global handler of new bundles which is passed when it is instantiated?
```go
protocol := NewChatProtocolService(WithNewBundlesHandler(func () { ... })
```

If processing new bundles might be needed in a place where `HandleMessage` is called, I like the second option where you can pass a callback:
```go
message, err := HandleMessage(privateKey, publicKey, msg.Payload, NewBundlesHandler())
```
Here, you have two options:
1. Pass it as a next parameter which would be required and it would usually be `nil` if a global handler can be configured,
2. Use a struct which would wrap this handler so that it can be extended in the future without changing the signature of this method. Not sure if this is important at this stage. In this case, if no options are needed, an empty struct would need to be passed.

I am not convinced if returning multiple errors is a better way because, first of all, the caller needs to be aware that there are multiple errors and examine all of them in a proper order to handle all cases. I haven't seen any solution that would solve that nicely, however, something like:
```go
message, bundles, err := HandleMessage(...)
if err.Bundle != nil {
    // handle error and return here? What about err.Message? 
    // Assumes the caller knows that err.Bundle needs to be handled first.
} else {
    // handle bundles
}
if err.Message != nil {
    // handle error
}
```
would not be that bad. Of course, it involves some additional glue code like declaring an additional type, providing methods to make it `error` interface compatible etc.

Generally, I think that the fact of handling added bundles when calling `HandleMessage` breaks encapsulation, just not during the method call but during the response processing. Hence, that global handler plus a callback seems the most appropriate.

cammellos: > Can chat.ProtocolService have a global handler of new bundles which is passed when it is instantiated? 

I think that should be fine, otherwise I can go for a callback, thanks

cammellos:
@adambabik implemented, is this what you had in mind? thanks!

adambabik (approved):

adambabik:
@cammellos looks great!

cammellos:
@adambabik thanks! @PombeirP ready to be reviewed again, applied your suggestion, also in case of approval, please do not merge, I still want to build a status-react build and make sure it works as expected, thanks!

pedropombeiro (approved):

PR #1300, opened 2018-12-03
stop adding 'v' before the contents of VERSION file

jakubgs:
This is messing with uploading of releases. Some tags we use have a `v` before them and some don't. So I'm just going to depend entirely on what is in `VERSION` and try not to add anything.

### DIFF
--- a/_assets/ci/lib.groovy
+++ b/_assets/ci/lib.groovy
@@ -9,7 +9,7 @@ def timestamp() {
 
 def suffix() {
   if (params.RELEASE == true) {
-    return 'v' + readFile("${env.WORKSPACE}/${env.STATUS_PATH}/VERSION").trim()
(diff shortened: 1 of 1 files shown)


_assets/ci/lib.groovy
```diff
@@ -9,7 +9,7 @@ def timestamp() {
 
 def suffix() {
   if (params.RELEASE == true) {
-    return 'v' + readFile("${env.WORKSPACE}/${env.STATUS_PATH}/VERSION").trim()
+    return readFile("${env.WORKSPACE}/${env.STATUS_PATH}/VERSION").trim()
```
adambabik, line 12: 👍 we should not be using `v` anymore.

adambabik (approved):

pedropombeiro (approved):

PR #1301, opened 2018-12-03
echo Linux instead of android for linux task

rcullito:
keeping the printout consistent with the make task `statusgo-linux`


pedropombeiro (approved):

adambabik (approved):

rcullito:
hey @PombeirP just looking to see if there is anything manually that I should do with travis-ci in order for this final check to pass? thanks!

pedropombeiro:
@rcullito not sure if that's because you're using a fork... Maybe you can try from status-im itself.

rcullito:
ah, ok. thanks pedro. just spun up https://github.com/status-im/status-go/pull/1307

Issue #1290, opened 2018-11-24
Mail server data synchronization


adambabik:
# Problem

Due to lack of data synchronisation between mail servers, our users experience the following issues:
1. New mail servers do not contain messages sent beforehand,
2. Intermittent problems with mail servers lead to single messages being lost.

This can be frustrating because two users can see totally different results while both claiming to use Status-provided mail servers.

# Implementation

**1. Protocol level**

These problems can be fixed on protocol level, i.e. providing a way to synchronize messages between two mail servers using devp2p protocol and a direct p2p connection. As each connection is handled in a separate goroutine, this solution won't block other peers from accessing historic messages. 

**2. Storage level**

Alternative solution is redesigning storage layer and using distributed or decentralized databases. In this model, mail servers act as "front-ends" to access the database. From the user perspective nothing will change.

We have two options for the storage layer: distributed databases like Apache Cassandra and Elasticsearch or decentralized databases like IPFS and swarm.

The first option is easier to implement and by selecting a proper database, it's very easy to provide additional functionality like full-text search in public chats. However, the synchronization of data is possible only using a devp2p protocol and this needs to be implemented to solve the first problem. 

Also, it requires pluggable data layer because if someone wants to run a mail server on their hardware, they need to be able to sync the old messages and running specific database like Elasticsearch can't be a requirement.

Decentralized databases can scale better and provide more autonomy for each mail server. Drawbacks are that the implementation is more difficult and providing additional functionality like full-text search may require an additional index layer. Even delivering historic messages may be a challenge and require an additional index layer.

The biggest advantage, however, is the fact that there is no need for a protocol level solution as data is synchornized by the storage layer. The only required thing is to agree on the data paths as well as index and data files schemas.

# Acceptance Criteria

1. It's possible to run a new mail server and sync old messages from other mail servers,
2. Intermittent issues and downtimes do not lead to missing messages; in other words, it's possible for a mail server to be constantly in-sync with other mail servers.

# Notes

The implementation should be fairly simple and achievable in a short period of time. This issue is irrespective of the research on the new communication.

cc @mandrigin @dshulyak

adambabik:
I'd like to work on a POC first that validates usage of a decentralized database and how complex storage and retrieving of messages would be.

mandrigin:
Sounds good to me!

adambabik:
Update: currently I am working on a protocol level synchronization as it's needed even if we decide to use a distributed storage for our mail servers.

[...]

PR #1302, opened 2018-12-04
Support Mail Server data synchronization

adambabik:
These changes add a support for syncing data between two Mail Servers. It still lacks features that allow triggering syncing easily.

This is more like a preparation to support data syncing between Mail Servers rather than a final implementation. It does not give an easy way to start syncing. This will be solved in the next PR.

Important changes:
- [x] Processing requests by mail servers was refactored.

Todo:
- [x] merged https://github.com/status-im/whisper/pull/13 first.

Relates to #1290


mailserver/mailserver.go
```diff
@@ -288,24 +290,56 @@ func (s *WMailServer) DeliverMail(peer *whisper.Peer, request *whisper.Envelope)
...
+			if err := s.sendEnvelopes(peer, bundle, batch); err != nil {
+				errCh <- err
+				break
+			}
+		}
+		close(errCh)
+	}()
+
+	start := time.Now()
+	nextPageCursor, lastEnvelopeHash := s.processRequestInBundles(
```
adambabik, line 322: Some stress tests for this fragment would be nice. We use a buffered channel but if there are more than 5 bundles, sending them through the network will become a bottleneck. It is still better than the previous implementation which was sending bundles without batching.

mailserver/mailserver_test.go
```diff
@@ -330,31 +330,41 @@ func (s *MailserverSuite) TestRequestPaginationLimit() {
 	s.Nil(cursor)
 	s.Equal(params.limit, limit)
 
-	envelopes, _, cursor, err := s.server.processRequest(nil, lower, upper, bloom, limit, nil, false)
-	s.NoError(err)
-	for _, env := range envelopes {
-		receivedHashes = append(receivedHashes, env.Hash())
+	iter := s.server.createIterator(lower, upper, cursor)
```
adambabik: This fragment repeats in a few places. It can be extracted to a method.

gravityblast (approved):

Gopkg.lock
```diff
@@ -822,12 +822,12 @@
   revision = "fbcc46a78cd43fef95a110df664aab513116a850"
 
 [[projects]]
-  digest = "1:6cb252f27feb57ef0e8406556c259d903c0ecff2ab0d2200ca85773b3561777d"
+  branch = "sync-messages"
```
gravityblast: just as reminder, we have to remember to change this after the PR on whisper is merged

mailserver/mailserver.go
```diff
@@ -288,24 +290,56 @@ func (s *WMailServer) DeliverMail(peer *whisper.Peer, request *whisper.Envelope)
...
 		requestsBatchedCounter.Inc(1)
 	}
 
-	_, lastEnvelopeHash, nextPageCursor, err := s.processRequest(
-		peer,
-		lower, upper,
+	iter := s.createIterator(lower, upper, cursor)
+	defer iter.Release()
+
+	bundles := make(chan []*whisper.Envelope, 5)
```
dshulyak, line 308: is it just to avoid blocking or this number means something else?

dshulyak (approved):

mailserver/mailserver.go
adambabik: Just avoid blocking. Pretty arbitrary.

PR #1303, opened 2018-12-05
Persist selected mail server using separate monitor

### LINKED ISSUE #1285: Provide information about selected mail servers to status-go
# Problem

We need to have information about selected mail servers when [confirmation](https://github.com/status-im/status-go/issues/1278) is received.

Several alternatives:
- assume that we are always connected with a mail server from TrustedMailServers. This information already available in status-go configuration. But it doesn't include info about manually added mail servers. So we may get into inconsistent state, and never show a confirmation that message was delivered.
- implicitly use last mail server that was used for fetching history. it seems to provide consistent information, but may sound a bit hacky. one potential inconsistency would be if we send a message before sending a request for history
- explicitly get a selected list of mail servers from a user when application is started, or this list was changed

I would prefer the last one, but i've seen a couple of discussions how to handle mail servers connection better. This is definitely will have an impact on that so maybe we can have a discussion here.

# Implementation

Add a binding to status-go library: UseMailServers that accepts lists of enodes and returns error if one of those is not parseable.

The binding does 3 things:
- ensures that we are connected with one mail server from that list
- provides configuration for requesting history
- provides configuration to envelope tracker

History request will use any connected peer from that list, unless mailserverPeer argument is passed. if it is passed it will preserve current logic.

Envelope tracker will acknowledge an envelope if confirmation was received from any peer in the list.

# Acceptance Criteria

Each part should work correctly according to the logic described. And information about used mail servers should be consistent across different components.

cc @yenda @adambabik @mandrigin @cammellos @PombeirP

dshulyak:
This change allows to connect to the mail server that we were using before the app was restarted. Separate loop is listening for whisper events, and when we receive event that request was completed we will update time on a peer record.

Records are stored in leveldb. Body of the record is marshaled using json. At this point the only field is a timestamp when record was used. 

This loop doesn't control connections, it only tracks what mail server we ended up using. It works asynchronously to connection management loop. Which tracks events that are related to connection state and expiry of the requests.

When app starts we look into the database and select the most recently used record. This record is added to connection management loop first. So if this server is available we will stick to using it. If we weren't able to connect to the same server in configured timeout (5s) we will try to connect to any other server from list of active servers.

closes: https://github.com/status-im/status-go/issues/1285

### DIFF
--- a/services/shhext/mailservers/cache.go
+++ b/services/shhext/mailservers/cache.go
@@ -1,0 +1,142 @@
+package mailservers
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/ethereum/go-ethereum/p2p/enode"
+	"github.com/status-im/status-go/db"
+	"github.com/syndtr/goleveldb/leveldb"
+	"github.com/syndtr/goleveldb/leveldb/iterator"
+	"github.com/syndtr/goleveldb/leveldb/util"
+)
+
+// NewPeerRecord returns instance of the peer record.
+func NewPeerRecord(node *enode.Node) PeerRecord {
+	return PeerRecord{node: node}
+}
+
+// PeerRecord is set data associated with each peer that is stored on disk.
+// PeerRecord stored with a enode as a key in leveldb, and body marshalled as json.
+type PeerRecord struct {
+	node *enode.Node
+
+	// last time it was used.
+	LastUsed time.Time
+}
+
+// Encode encodes PeerRecords to bytes.
+func (r PeerRecord) Encode() ([]byte, error) {
+	return json.Marshal(r)
+}
+
+// ID returns enode identity of the node.
+func (r PeerRecord) ID() enode.ID {
+	return r.node.ID()
+}
+
+// Node returs pointer to original object.
+// enode.Node doensn't allow modification on the object.
+func (r PeerRecord) Node() *enode.Node {
+	return r.node
+}
+
+// EncodeKey returns bytes that will should be used as a key in persistent storage.
+func (r PeerRecord) EncodeKey() ([]byte, error) {
+	return r.Node().MarshalText()
+}
+
+// NewCache returns pointer to a Cache instance.
+func NewCache(db *leveldb.DB) *Cache {
+	return &Cache{db: db}
+}
+
+// Cache is wrapper for operations on disk with leveldb.
+type Cache struct {
+	db *leveldb.DB
+}
+
+// Replace deletes old and adds new records in the persistent cache.
+func (c *Cache) Replace(nodes []*enode.Node) error {
+	batch := new(leveldb.Batch)
+	iter := createPeersIterator(c.db)
+	defer iter.Release()
+	newNodes := nodesToMap(nodes)
+	for iter.Next() {
+		record, err := unmarshalKeyValue(keyWithoutPrefix(iter.Key()), iter.Value())
+		if err != nil {
+			return err
+		}
+		if _, exist := newNodes[record.ID()]; exist {
+			delete(newNodes, record.ID())
+		} else {
+			batch.Delete(iter.Key())
+		}
+	}
+	for _, n := range newNodes {
+		enodeKey, err := n.MarshalText()
+		if err != nil {
+			return err
+		}
+		// we put nil as default value doesn't have any state associated with them.
+		batch.Put(db.Key(db.MailserversCache, enodeKey), nil)
+	}
+	return c.db.Write(batch, nil)
+}
+
+// LoadAll loads all records from persistent database.
+func (c *Cache) LoadAll() (rst []PeerRecord, err error) {
+	iter := createPeersIterator(c.db)
+	for iter.Next() {
+		record, err := unmarshalKeyValue(keyWithoutPrefix(iter.Key()), iter.Value())
+		if err != nil {
+			return nil, err
+		}
+		rst = append(rst, record)
+	}
+	return rst, nil
+}
+
+// UpdateRecord updates single record.
+func (c *Cache) UpdateRecord(record PeerRecord) error {
+	enodeKey, err := record.EncodeKey()
+	if err != nil {
+		return err
+	}
+	value, err := record.Encode()
+	if err != nil {
+		return err
+	}
+	return c.db.Put(db.Key(db.MailserversCache, enodeKey), value, nil)
+}
+
+func unmarshalKeyValue(key, value []byte) (record PeerRecord, err error) {
+	enodeKey := key
+	node := new(enode.Node)
+	err = node.UnmarshalText(enodeKey)
+	if err != nil {
+		return record, err
+	}
+	record = PeerRecord{node: node}
+	if len(value) != 0 {
+		err = json.Unmarshal(value, &record)
+	}
+	return record, err
+}
+
+func nodesToMap(nodes []*enode.Node) map[enode.ID]*enode.Node {
+	rst := map[enode.ID]*enode.Node{}
+	for _, n := range nodes {
+		rst[n.ID()] = n
+	}
+	return rst
+}
+
+func createPeersIterator(level *leveldb.DB) iterator.Iterator {
+	return level.NewIterator(util.BytesPrefix([]byte{byte(db.MailserversCache)}), nil)
+}
+
+// keyWithoutPrefix removes first byte from key.
+func keyWithoutPrefix(key []byte) []byte {
+	return key[1:]
+}
--- a/services/shhext/mailservers/cache_test.go
+++ b/services/shhext/mailservers/cache_test.go
@@ -23,14 +23,14 @@ func containsNode(nodes []*enode.Node, node *enode.Node) error {
 			return nil
 		}
 	}
-	return fmt.Errorf("nodes %+s do not containt  %s", nodes, node)
+	return fmt.Errorf("nodes %+s do not contain  %s", nodes, node)
 }
 
 func TestReplaceRecords(t *testing.T) {
 	nodesNumber := 3
 	cache := newInMemCache(t)
 	nodes := make([]*enode.Node, nodesNumber)
-	// First round is a sanity check that record were written.
+	// First round is a sanity check that records were written.
 	fillWithRandomNodes(t, nodes)
 	require.NoError(t, cache.Replace(nodes))
 	records, err := cache.LoadAll()
--- a/services/shhext/mailservers/cache_test.go
+++ b/services/shhext/mailservers/cache_test.go
@@ -1,0 +1,80 @@
+package mailservers
+
+import (
+	"fmt"
+	"testing"
+	"time"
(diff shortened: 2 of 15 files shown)


mandrigin (approved):
lgtm

services/shhext/mailservers/cache.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+// NewPeerRecord returns instance of the peer record.
+func NewPeerRecord(node *enode.Node) PeerRecord {
+	return PeerRecord{node: node}
+}
+
+// PeerRecord is set data associated with each peer that is stored on disk.
+type PeerRecord struct {
+	node *enode.Node
```
adambabik, line 22: Any reason why `node` is unexported by `LastUsed` exported?

services/shhext/mailservers/cache.go
```diff
@@ -0,0 +1,119 @@
...
+
+// Cache is wrapper for operations on disk with leveldb.
+type Cache struct {
+	db *leveldb.DB
+}
+
+// Replace delets old and adds new records in the persistent cache.
+func (c *Cache) Replace(nodes []*enode.Node) error {
+	batch := new(leveldb.Batch)
+	iter := c.db.NewIterator(util.BytesPrefix([]byte{byte(db.MailserversCache)}), nil)
```
adambabik: Maybe extract creating an iterator to a new method on `Cache`?

services/shhext/mailservers/cache.go
```diff
@@ -0,0 +1,119 @@
...
+	defer iter.Release()
+	newNodes := nodesToMap(nodes)
+	for iter.Next() {
+		record, err := unmarshalKeyValue(iter.Key()[1:], iter.Value())
```
adambabik: Magic `[1:]`. It would be nice to have some functions in `db` package that could help with keys management.

services/shhext/mailservers/cache.go
```diff
@@ -0,0 +1,119 @@
...
+func NewCache(db *leveldb.DB) *Cache {
+	return &Cache{db: db}
+}
+
+// Cache is wrapper for operations on disk with leveldb.
+type Cache struct {
+	db *leveldb.DB
+}
+
+// Replace delets old and adds new records in the persistent cache.
```
adambabik:
```suggestion
// Replace deletes old and adds new records in the persistent cache.
```

services/shhext/mailservers/cache.go
```diff
@@ -0,0 +1,119 @@
...
+		} else {
+			batch.Delete(iter.Key())
+		}
+	}
+	for _, n := range newNodes {
+		enodeKey, err := n.MarshalText()
+		if err != nil {
+			return err
+		}
+		batch.Put(db.Key(db.MailserversCache, enodeKey), nil)
```
adambabik, line 82: I would explain why we put `nil` as a value. Actually, what is missing in this file, is a description how the records are stored in this key-value store, like what are the key and value here.

services/shhext/mailservers/cache.go
```diff
@@ -0,0 +1,119 @@
...
+
+func unmarshalKeyValue(key, value []byte) (record PeerRecord, err error) {
+	enodeKey := key
+	node := new(enode.Node)
+	err = node.UnmarshalText(enodeKey)
+	if err != nil {
+		return record, err
+	}
+	record = PeerRecord{node: node}
+	if len(value) != 0 {
```
adambabik, line 121:
```suggestion
	if len(value) > 0 {
```

services/shhext/mailservers/connmanager.go
```diff
@@ -148,6 +139,26 @@ func (ps *ConnectionManager) Stop() {
 	ps.quit = nil
 }
 
+func processReplacement(srv PeerAdderRemover, target int, timeout time.Duration, newNodes []*enode.Node, events <-chan *p2p.PeerEvent, connected map[enode.ID]struct{}, current map[enode.ID]*enode.Node) map[enode.ID]*enode.Node {
```
adambabik: Any reason this function is not a method on `ConnectionManager`? We could avoid that many arguments.

services/shhext/mailservers/connmonitor.go
```diff
@@ -0,0 +1,85 @@
+package mailservers
+
+import (
+	"sync"
+	"time"
+
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/p2p/enode"
+	"github.com/status-im/whisper/whisperv6"
```
adambabik:
```suggestion
	whisper "github.com/status-im/whisper/whisperv6"
```

services/shhext/mailservers/connmonitor.go
```diff
@@ -0,0 +1,85 @@
...
+	go func() {
+		events := make(chan whisperv6.EnvelopeEvent, whisperEventsBuffer)
+		sub := mon.whisper.SubscribeEnvelopeEvents(events)
+		for {
+			select {
+			case <-mon.quit:
+				sub.Unsubscribe()
+				mon.wg.Done()
+				return
+			case err := <-sub.Err():
```
adambabik, line 45: We can potentially have a while loop here.

services/shhext/mailservers/connmonitor.go
```diff
@@ -0,0 +1,85 @@
...
+				sub = mon.whisper.SubscribeEnvelopeEvents(events)
+			case ev := <-events:
+				node := mon.ps.Get(ev.Peer)
+				if node == nil {
+					continue
+				}
+				if ev.Event == whisperv6.EventMailServerRequestCompleted {
+					err := mon.updateRecord(ev.Peer)
+					if err != nil {
+						log.Error("unable to store that server was used", "peer", ev.Peer, "error", err)
```
adambabik:
```suggestion
						log.Error("unable to update store", "peer", ev.Peer, "error", err)
```

services/shhext/mailservers/utils.go
```diff
@@ -0,0 +1,52 @@
...
+		return nil
+	}
+	sort.Slice(records, func(i, j int) bool {
+		return records[i].LastUsed.After(records[j].LastUsed)
+	})
+	all := recordsToNodes(records)
+	if !records[0].LastUsed.IsZero() {
+		conn.Notify(all[:1])
+	}
+	conn.Notify(all)
```
adambabik, line 42: I don't quite get this fragment. First, we notify with the first item and then with all. Is this correct?

services/shhext/mailservers/cache.go
dshulyak: only exported items are marshaled. enode is unmarshaled from the key

services/shhext/mailservers/connmanager.go
dshulyak: it doesn't change any internal state of the connection manager, all operations are on local data. i can wrap all local data into a single struct though

services/shhext/mailservers/connmonitor.go
dshulyak: tbh i don't like this at all. the package is whisperv6 and all tools are importing it like that by default

services/shhext/mailservers/connmonitor.go
dshulyak: what do you mean?

services/shhext/mailservers/cache.go
adambabik: So the logic of json marshaler influence the structure of this struct? You can use `json:"-"` to be more explicit. 

I checked the usage once again and I think it makes sense if we consider `node` to be immutable after initialization and `LastUsed` mutable.

services/shhext/mailservers/connmanager.go
adambabik: Yeah, maybe that could help because it's very hard to reason about which parameters are mutable and what's the end result of these operations.

services/shhext/mailservers/connmonitor.go
adambabik: We have been doing it this way so far so for consistency I would stick to `whisper "github.com/status-im/whisper/whisperv6"`. 

That's a default behaviour because these tools are unaware of the fact that whisper module is poorly versioned. It should be "github.com/status-im/whisper/v6" I believe and it would be seen as `whisper`.

services/shhext/mailservers/connmonitor.go
adambabik: `sub` returns an error, it's handled with `sub = mon.whisper.SubscribeEnvelopeEvents(events)` and the new `sub` returns error again and so on. I am not sure if it's possible but looking at this fragment, it looks like it is.

services/shhext/mailservers/cache_test.go
```diff
@@ -0,0 +1,80 @@
...
+	return NewCache(db)
+}
+
+func containsNode(nodes []*enode.Node, node *enode.Node) error {
+	for _, n := range nodes {
+		if n.ID() == node.ID() {
+			return nil
+		}
+	}
+	return fmt.Errorf("nodes %+s do not containt  %s", nodes, node)
```
pedropombeiro:
```suggestion
	return fmt.Errorf("nodes %+s do not contain  %s", nodes, node)
```

services/shhext/mailservers/cache_test.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func TestReplaceRecords(t *testing.T) {
+	nodesNumber := 3
+	cache := newInMemCache(t)
+	nodes := make([]*enode.Node, nodesNumber)
+	// First round is a sanity check that record were written.
```
pedropombeiro:
```suggestion
	// First round is a sanity check that records were written.
```

services/shhext/mailservers/cache.go
dshulyak: i will add comments. in general nil made sense to me, because there is no unique state for recently created value.

services/shhext/mailservers/utils.go
dshulyak: yeah, the idea is that we need to prioritize certain connections. i wanted to keep this information outside of the connection management, but i am not sure that it will be possible if we want to prioritize connections based on latency.

services/shhext/mailservers/connmonitor.go
dshulyak: i guess i will exit goroutine. the thing is that this error shouldn't be raised at all. but it exists in the interface

dshulyak:
@PombeirP @adambabik addressed comments, please take a look

services/shhext/mailservers/utils.go
adambabik: Ok, sounds good but I would add a comment explaining why it is so. Also if we do `conn.Notify(all[:1])`, shouldn't the next call be `conn.Notify(all[1:])` or it does not matter?

adambabik (approved):

PR #1304, opened 2018-12-06
Ensure sent event is received from a mailserver

dshulyak:
In previous change i used same filtering for a handler that processes confirmations.
But if we are connected with peers that do not support confirmations on whisper level,
whisper may send "Sent" event when envelope is written to a socket. This is our old behaviour.

So, this PR allows to use confirmations from a mail servers and ensure that transition between
multiple version of peers will be smooth.


ghost:
### Pull Request Checklist

cammellos (approved):

PR #1306, opened 2018-12-06
Hash password before using it

cammellos:
The encryption password was not sanitized causing issues with special characters :woman_facepalming: 
Changes the behavior to take the `sha256` (used as an input string).

This is an hotfix, so already on release branch, so the scope for changes is limited, but I have set up a pattern so that further migrations of keys can be applied following convention.


ghost:
### Pull Request Checklist

services/shhext/service_test.go
```diff
@@ -92,7 +92,7 @@ func (s *ShhExtSuite) SetupTest() {
 			InstallationID:          "1",
 			DataDir:                 os.TempDir(),
```
dshulyak: i think it needs to use https://golang.org/pkg/io/ioutil/#TempDir for isolation. os.TempDir() returns default temporary directory, right?

services/shhext/service_test.go
```diff
@@ -106,6 +106,12 @@ func (s *ShhExtSuite) SetupTest() {
 	s.services[0].tracker.handler = newHandlerMock(1)
 }
 
+func (s *ShhExtSuite) TestInitProtocol() {
+	os.Remove("/tmp/1.v2.db")
```
dshulyak: i suggest to create temporary directory for service with ioutil.TempDir in Setup and remove it here. also it is not clear from the test name why directory is removed before InitProtocol

services/shhext/service.go
```diff
@@ -108,17 +109,27 @@ func (s *Service) InitProtocol(address string, password string) error {
...
 
-	if err := chat.MigrateDBFile(oldPath, newPath, password); err != nil {
+	if err := chat.MigrateDBFileV1(v0Path, v1Path, password); err != nil {
 		return err
 	}
 
-	persistence, err := chat.NewSQLLitePersistence(newPath, password)
+	if err := chat.MigrateDBFileV2(v1Path, v2Path, password, hashedPassword); err != nil {
+		// Remove db file as created with a blank password
+		os.Remove(v1Path)
```
pedropombeiro, line 129: Does this mean that if `MigrateDBFileV2` fails to rename `v1Path` into `v2Path` for some reason, we'll delete the source `v1Path` file? Why is that?

services/shhext/chat/sql_lite_persistence.go
```diff
@@ -86,13 +86,46 @@ func MigrateDBFile(oldPath string, newPath string, key string) error {
...
+		return err
+	}
+
+	// Migrate version that uses actual password
+	db, err := openDB(newPath, oldKey)
+	if err != nil {
+		return err
+	}
+
+	keyString := fmt.Sprintf("PRAGMA rekey= '%s'", newKey)
```
pedropombeiro:
```suggestion
	keyString := fmt.Sprintf("PRAGMA rekey = '%s'", newKey)
```

services/shhext/service.go
cammellos: basically it happened that those impacted would create a database with an empty password, so we are not able to rekey, it handles this case. Also this db `v1` can be safely erased, and we know that if it errors likely was never used.

services/shhext/chat/sql_lite_persistence.go
```diff
@@ -86,13 +86,46 @@ func MigrateDBFile(oldPath string, newPath string, key string) error {
 
 }
 
+func MigrateDBFileV2(oldPath string, newPath string, oldKey string, newKey string) error {
```
adambabik: Can't we have just one `MigrateDBFile` with this interface? It looks like `MigrateDBFileV1("/old/path", "/new/path", "my-key")` call can be replaced with `MigrateDBFileV2("/old/path", "/new/path", "ON", "my-key")`.

cammellos: yes, that's probably the same, I will take a look at it

services/shhext/service.go
pedropombeiro: Maybe we should add that context in the comment?

cammellos:
@adambabik @dshulyak @PombeirP updated!

services/shhext/service_test.go
```diff
@@ -104,6 +104,15 @@ func (s *ShhExtSuite) SetupTest() {
 		s.nodes[i] = stack
 	}
 	s.services[0].tracker.handler = newHandlerMock(1)
+
+	// Cleanup db file from previous runs
+	dbFilePath := fmt.Sprintf("%s/1.v2.db", os.TempDir())
```
dshulyak: this is not what i meant though. right now this test is not really isolated. for example two tests will conflict with each other. or leftover from previous test run may impact with new run. what i suggested is to use **ioutill**.TempDir, it will create a directory with random prefix instead of using root temporary directory

dshulyak (approved):

services/shhext/service_test.go
cammellos: oh I see, I misunderstood, I can change that, thanks

pedropombeiro (approved):

PR #1311, opened 2018-12-10
Add couple of e2e tests for connection manager

dshulyak:


adambabik:
This should be reviewed after #1303 is merged, correct? It looks like it contains changes from #1303.

dshulyak:
@adambabik yes you are right, it can be reviewed in parallel it if is convenient to review only single commit. but in general can be reviewed aftyer

PR #1312, opened 2018-12-10
add ExtendedKey.ChildForPurpose method to derive wallet or chat keys

gravityblast:
This is the first step to decouple the wallet and chat keys.
In this PR we add a method `ChildForPurpose` to `ExtendedKey` to derive a child key based on a purpose. 
Purposes are:

* `KeyPurposeWallet` ([BIP44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki) keys)
* `KeyPurposeChat` ([EIP1581](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1581.md) keys)

This method is not used yet, but it will be used by our version of `go-ethereum` that is currently calling the `BIP44Child` method of `ExtendedKey`.


extkeys/hdkey.go
```diff
@@ -41,6 +41,13 @@ import (
 
 // TODO make sure we're doing this ^^^^ !!!!!!
 
+type KeyPurpose int
+
+const (
+	KeyPurposeWallet KeyPurpose = iota
```
adambabik: I would start with `1` instead of `0` because it often leads to misunderstandings. `0` is a default `int` value and might be used by accident.

extkeys/hdkey.go
```diff
@@ -95,6 +103,22 @@ var (
...
+		HardenedKeyStart + 0,           // account
+		0,                              // 0 - public, 1 - private
+	}
+
+	// EthEIP1581ChatParentPath is EIP-1581 chat keys parent's derivation path
+	EthEIP1581ChatParentPath = []uint32{
+		HardenedKeyStart + 43,          // purpose
+		HardenedKeyStart + CoinTypeETH, // cointype set to ETH
+		HardenedKeyStart + 1581,        // EIP-1581 subpurpose
+		HardenedKeyStart + 0,           // key_type (chat)
```
adambabik: I wonder why we don't have `HardenedKeyStart + 0` in `EthBIP44ParentPath` but here it is so.

extkeys/hdkey.go
```diff
@@ -236,9 +260,28 @@ func (k *ExtendedKey) Child(i uint32) (*ExtendedKey, error) {
 	return child, nil
 }
 
+func (k *ExtendedKey) ChildForPurpose(p KeyPurpose, i uint32) (*ExtendedKey, error) {
```
adambabik, line 267: Missing comment.

extkeys/hdkey.go
```diff
@@ -236,9 +260,28 @@ func (k *ExtendedKey) Child(i uint32) (*ExtendedKey, error) {
...
+		return k.EthEIP1581ChatChild(i)
+	default:
+		return nil, ErrInvalidKeyPurpose
+	}
+}
+
 // BIP44Child returns Status CKD#i (where i is child index).
 // BIP44 format is used: m / purpose' / coin_type' / account' / change / address_index
+// BIP44Child is depracated in favour of EthBIP44Child
+// Param coinType is deprecatet; we override it to always use CoinTypeETH.
```
adambabik:
```suggestion
// Param coinType is deprecated; we override it to always use CoinTypeETH.
```

extkeys/hdkey.go
adambabik: Also, in this case, `0` means `key_type = chat`. If it happens that we need one more type, we would use `HardenedKeyStart + 1` and so on?

extkeys/hdkey_test.go
```diff
@@ -577,6 +577,48 @@ func TestBIP44ChildDerivation(t *testing.T) {
...
 
+func TestChildForPurpose(t *testing.T) {
+	masterKey, err := NewKeyFromString(masterPrivKey1)
+	if err != nil {
+		t.Error("NewKeyFromString: cannot create master extended key")
+	}
+
+	bip44Child, err := masterKey.EthBIP44Child(0)
+	if err != nil {
+		t.Error("Error dering BIP44-compliant key")
```
adambabik:
```suggestion
		t.Error("Error deriving BIP44-compliant key")
```

extkeys/hdkey_test.go
```diff
@@ -577,6 +577,48 @@ func TestBIP44ChildDerivation(t *testing.T) {
...
+	}
+
+	eip1581Child, err := masterKey.EthEIP1581ChatChild(0)
+	if err != nil {
+		t.Error("Error dering EIP1581-compliant key")
```
adambabik:
```suggestion
		t.Error("Error deriving EIP1581-compliant key")
```

extkeys/hdkey_test.go
```diff
@@ -577,6 +577,48 @@ func TestBIP44ChildDerivation(t *testing.T) {
...
+	}
+
+	walletChild, err := masterKey.ChildForPurpose(KeyPurposeWallet, 0)
+	if err != nil {
+		t.Error("Error dering BIP44-compliant key")
```
adambabik:
```suggestion
		t.Error("Error deriving BIP44-compliant key")
```

extkeys/hdkey_test.go
```diff
@@ -577,6 +577,48 @@ func TestBIP44ChildDerivation(t *testing.T) {
...
+	}
+
+	chatChild, err := masterKey.ChildForPurpose(KeyPurposeChat, 0)
+	if err != nil {
+		t.Error("Error dering EIP1581-compliant key")
```
adambabik:
```suggestion
		t.Error("Error deriving EIP1581-compliant key")
```

adambabik (changes requested):

extkeys/hdkey.go
gravityblast: @adambabik good point, I'm changing it!

extkeys/hdkey.go
gravityblast: > I wonder why we don't have HardenedKeyStart + 0 in EthBIP44ParentPath but here it is so.

As defined in the [BIP44 specs](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#path-levels), the 0 in the EthBIP44ParentPath is the `change` which is not hardened, so it starts with 0.

> Also, in this case, 0 means key_type = chat. If it happens that we need one more type, we would use HardenedKeyStart + 1 and so on?

Yes we will have a different index for each key_type, in the same way we have coin types for BIP44 keys. For example `60` is ETH etc..
Instead of using `0` I'm adding a constant

gravityblast:
@adambabik I updated the code and committed your suggestions

adambabik (approved):

adambabik:
@gravityblast looks good but the linter complains :(

PR #1313, opened 2018-12-10
Close both sides of the stream when client is done with reading response

dshulyak:
Libp2p keeps stream open if EOF wasn't seen and we called Close method.
The most important change is that reader now uses FullClose util, that will
wait for EOF character before closing the stream.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -811,7 +811,7 @@
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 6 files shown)


adambabik:
@dshulyak can you link a PR/commit with rendezvous changes?

dshulyak:
@adambabik https://github.com/status-im/rendezvous/commit/9e20b11affd0bf0591126a518f3e7b8aa057f88f

adambabik (approved):

PR #1326, opened 2018-12-17
Use selected wallet and chat accounts

gravityblast:
This PR prepares `status-go`  to decouple the wallet and chat keys, updating the `AccountManager` to have a `selectedWalletAccount` and a `selectedChatAccount`. 
For now they both contains the same `BIP44` key without breaking compatibility. 

TODO

* [x] rollback `VERSION` file


gravityblast:
@cammellos when you have time can you check this PR please? especially in places like `CreateContactCode`, `ProcessContactCode`, etc where we now use `selectedChatAccount`.

cammellos:
@gravityblast looks good to me, would you like me to test it locally against status-react?

t/e2e/whisper/whisper_test.go
```diff
@@ -109,7 +110,15 @@ func (s *WhisperTestSuite) TestSelectAccount() {
 	// select another account, make sure that previous account is wiped out from Whisper cache
 	s.False(whisperService.HasKeyPair(pubKey), "identity already present in whisper")
 	s.NoError(s.Backend.SelectAccount(address, TestConfig.Account1.Password))
-	s.True(whisperService.HasKeyPair(pubKey), "identity not injected into whisper")
+	// get the chat key of the selected account
```
adambabik: I would write it as a separate test `TestWalletAndChatPubKeyAreTheSame` and this test can be unchanged. 

Later, when we make the final change, this test will continue to pass and only `TestWalletAndChatPubKeyAreTheSame` will fail and will be replaced with a one verifying that the keys are different.

t/e2e/whisper/whisper_test.go
```diff
@@ -125,11 +134,22 @@ func (s *WhisperTestSuite) TestLogout() {
...
+
+	// select account
 	s.NoError(s.Backend.SelectAccount(address, TestConfig.Account1.Password))
-	s.True(whisperService.HasKeyPair(pubKey), "identity not injected into whisper")
+
+	// Get the chat account
+	selectedChatAccount, err := s.Backend.AccountManager().SelectedChatAccount()
+	s.NoError(err)
+
+	// Chat account public key should be the same key received when creating the account
```
adambabik: Keys being the same has nothing to do with `Logout` logic.

adambabik (approved):
LGTM 👍 Two testing fragments can be a bit clearer IMO.

gravityblast:
> @gravityblast looks good to me, would you like me to test it locally against status-react?

@cammellos if you can it would be great. I'm also preparing a build for the test team.

t/e2e/whisper/whisper_test.go
gravityblast: @adambabik for now `CreateAccount` returns only the wallet key. the first part of the test is testing that the returned key is not in whisper, so I wanted to be sure that the key we are testing is the same the we are using for chat. It will be removed in the next steps when we'll introduce a different key, so that part will fail and force us to update the test. 
If you prefer I can remove it

t/e2e/whisper/whisper_test.go
gravityblast: @adambabik ok I can create a new test for this, good idea!

t/e2e/whisper/whisper_test.go
gravityblast: I saw the other comment, so if we create a different test to check the 2 keys, it makes sense to remove from here!

gravityblast:
@adambabik I updated the code after your comments

adambabik:
@gravityblast thanks 👍

t/e2e/whisper/whisper_test.go
```diff
@@ -176,32 +191,65 @@ func (s *WhisperTestSuite) TestSelectedAccountOnRestart() {
...
+	s.Nil(selectedWalletAccount)
 
-	selectedAccount, err = s.Backend.AccountManager().SelectedAccount()
+	selectedChatAccount, err = s.Backend.AccountManager().SelectedChatAccount()
 	s.EqualError(account.ErrNoAccountSelected, err.Error())
-	s.Nil(selectedAccount)
+	s.Nil(selectedChatAccount)
+}
+
+func (s *WhisperTestSuite) TestWalletAndChatPubKeyAreTheSame() {
```
adambabik: Oh actually, can we put this into `accounts_test.go`? Because this is e2e test so it's not that often executed and this test does not require a running node.

adambabik (changes requested):

t/e2e/whisper/whisper_test.go
gravityblast: @adambabik I added this check in the account test when we call `SelectAccount`.
I would like to leave this one as well (I renamed it to `TestSelectedChatKeyIsUsedInWhisper`), so that we check that we inject the right key. what do you think?

account/accounts_test.go
```diff
@@ -254,14 +254,21 @@ func (s *ManagerTestSuite) TestSelectAccount() {
 			s.gethServiceProvider.EXPECT().AccountKeyStore().Return(testCase.accountKeyStoreReturn...).AnyTimes()
 			err := s.accManager.SelectAccount(testCase.address, testCase.password)
 			s.Equal(testCase.expectedError, err)
+
+			selectedWalletAccount, err := s.accManager.SelectedWalletAccount()
```
adambabik, line 258: I rather though about something like this:
```go
func (s *ManagerTestSuite) TestSelectedWalletAndChatAccountsAreEqual() {
  selectedWalletAccount, err := s.accManager.SelectedWalletAccount()
  s.NoError(err)
  selectedChatAccount, err := s.accManager.SelectedChatAccount()
  s.NoError(err)
  s.Equal(selectedWalletAccount.AccountKey, selectedChatAccount.AccountKey)
}
```

but we can also keep `func (s *WhisperTestSuite) TestSelectedChatKeyIsUsedInWhisper()`.

adambabik (approved):

gravityblast:
ok thank you! I left it in the existing account test now but I'm going to remove it soon with the next PR

PR #1330, opened 2018-12-20
Add CreateAccount RPC function

3ach:
Adds the CreateAccount functionality

Adds a CreateAccount RPC call which enables [this](https://github.com/status-im/status-react/issues/6875) issue to move forward. Also adds tests


jeluard:
@zachzundel Wow thanks for the effort!

account/accounts.go
```diff
@@ -408,3 +409,20 @@ func (m *Manager) AddressToDecryptedAccount(address, password string) (accounts.
 
 	return keyStore.AccountDecryptedKey(account, password)
 }
+
+func (m *Manager) CreateAddress() (address, pubKey, privKey string, err error) {
```
adambabik: As described in the issue https://github.com/status-im/status-react/issues/6875, this should be independent from local accounts. That's why I would not make this method a part of `Manager` but rather a separate file with a function.

adambabik: It should also have a comment explaining what it is and it's purpose.

account/accounts_test.go
```diff
@@ -412,3 +415,17 @@ func (s *ManagerTestSuite) TestAddressToDecryptedAccount() {
 		})
 	}
 }
+
+func (s *ManagerTestSuite) TestCreateAddress() {
+	addr, pub, priv, err := s.accManager.CreateAddress()
+	s.Equal(true, gethcommon.IsHexAddress(addr))
+	s.NoError(err)
```
adambabik: Swap with `s.Equal(true, gethcommon.IsHexAddress(addr))`. Checking `err` should be first.

account/accounts_test.go
```diff
@@ -412,3 +415,17 @@ func (s *ManagerTestSuite) TestAddressToDecryptedAccount() {
...
+
+	privECDSA, err := crypto.HexToECDSA(priv[2:])
```
adambabik: Why `priv[2:]` instead of `priv`?

account/accounts_test.go
```diff
@@ -412,3 +415,17 @@ func (s *ManagerTestSuite) TestAddressToDecryptedAccount() {
...
+	s.NoError(err)
+
+	pubECDSA := privECDSA.Public().(*ecdsa.PublicKey)
+	pubStr := hexutil.Encode(crypto.FromECDSAPub(pubECDSA))
+
+	s.Equal(pub, pubStr)
```
adambabik: I guess it should be `s.Equal(pubStr, pub)`. I would recommend to call the arguments `s.Equal(expectedPubKeyStr, pub)`.

services/status/api.go
```diff
@@ -65,3 +65,21 @@ func (api *PublicAPI) Signup(context context.Context, req SignupRequest) (res Si
...
+}
+
+type CreateAddressResponse struct {
+	Address string `json:"address"`
+	Pubkey  string `json:"pubkey"`
+	Privkey string `json:"privkey"`
+}
+
+// CreateAddress is an implementation of `status_createaccount` or `web3.status.createaccount` API
+func (api *PublicAPI) CreateAddress(context context.Context, req CreateAddressRequest) (res CreateAddressResponse, err error) {
```
adambabik:
```go
func (api *PublicAPI) CreateAddress(ctx context.Context) (res CreateAddressResponse, err error) {
```
should be enough.

services/status/api.go
```diff
@@ -65,3 +65,21 @@ func (api *PublicAPI) Signup(context context.Context, req SignupRequest) (res Si
...
+	if res.Address, res.Pubkey, res.Privkey, err = api.s.am.CreateAddress(); err != nil {
+		err = errors.New("could not create the address : " + err.Error())
```
adambabik:
```suggestion
		err = fmt.Errorf("could not create the address: %v", err)
```

services/status/api.go
```diff
@@ -65,3 +65,21 @@ func (api *PublicAPI) Signup(context context.Context, req SignupRequest) (res Si
 
 	return
 }
+
+type CreateAddressRequest struct {
+}
+
+type CreateAddressResponse struct {
```
adambabik, line 73: Missing comment.

account/accounts_test.go
3ach: Because HexToECDSA dies if the string starts with `0x`. Which is stupid, but how it is.

account/accounts.go
3ach: Separate from the `account` directory entirely? Or just a separate file in here? That would make sense but I'm not sure how I would implement all the RPC stuff again. This doesn't touch local accounts I don't think.

3ach: Wait, actually, I realize now what you mean. Sorry, change incoming.

3ach:
Thanks for the review @adambabik , I think I've addressed your concerns now.

services/status/api.go
```diff
@@ -65,3 +68,19 @@ func (api *PublicAPI) Signup(context context.Context, req SignupRequest) (res Si
...
+	Address string `json:"address"`
+	Pubkey  string `json:"pubkey"`
+	Privkey string `json:"privkey"`
+}
+
+// CreateAddress is an implementation of `status_createaccount` or `web3.status.createaccount` API
+func (api *PublicAPI) CreateAddress(context context.Context) (res CreateAddressResponse, err error) {
+	if res.Address, res.Pubkey, res.Privkey, err = account.CreateAddress(); err != nil {
+		err = fmt.Errorf("could not create the address:  %v", err)
```
pedropombeiro: nit:
```suggestion
		err = fmt.Errorf("could not create an address:  %v", err)
```

pedropombeiro (approved):

services/status/api.go
3ach: Oh, thanks, tired brain does words bad. Fixed!

account/wallet.go
```diff
@@ -0,0 +1,25 @@
...
+
+func CreateAddress() (address, pubKey, privKey string, err error) {
+	key, err := crypto.GenerateKey()
+	if err != nil {
+		return "", "", "", err
+	}
+
+	privKeyBytes := crypto.FromECDSA(key)
+	pubKeyBytes := crypto.FromECDSAPub(key.Public().(*ecdsa.PublicKey))
+	addressBytes := crypto.PubkeyToAddress(*key.Public().(*ecdsa.PublicKey))
```
gravityblast: @zachzundel if you want you can just use:
```
	pubKeyBytes := crypto.FromECDSAPub(&key.PublicKey)
	addressBytes := crypto.PubkeyToAddress(key.PublicKey)
```

gravityblast (approved):
I just added one small comment but the rest looks good!

account/wallet.go
```diff
@@ -0,0 +1,25 @@
+package account
```
adambabik: I would call this file `create_address.go` or `address.go` or put it into `utils.go`.

account/wallet_test.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCreateAddress(t *testing.T) {
+	addr, pub, priv, err := CreateAddress()
+	require.NoError(t, err)
+	require.Equal(t, gethcommon.IsHexAddress(addr), true)
+
+	privECDSA, err := crypto.HexToECDSA(priv[2:])
```
adambabik: The problem is that `hexutil` is used in `CreateAddress` while `crypto` uses `hex` from standard lib which does not expect `0x` prefix. 

Now, the question is if `CreateAddress` should return nicely formatted strings or instead it would return a private key and the whole formatting for JSON-RPC API response would happen in `func (api *PublicAPI) CreateAddress(context context.Context)`. I think that the second option is nicer.

Also, is the JSON-RPC method expected to return byte arrays encoded with `hexutils` and contain `0x` prefix?

account/wallet.go
3ach: Thanks!

account/wallet.go
3ach: Renamed!

account/wallet_test.go
3ach: The JSON-RPC method is not specified. I'm implementing a new method to address https://github.com/status-im/status-react/issues/6875.

account/wallet.go
adambabik: I still see `account/wallet.go` :D

account/wallet_test.go
adambabik: RIght but I am talking about following a convention which is used in other existing methods. 

But yeah, `hexutils` seems like a way to go, for example: `"result": "0xc94" // 1207`

adambabik: Let's merge as it is and we can have a follow-up.

adambabik (approved):
LGTM! Thanks @zachzundel 

Please change `account/wallet.go` file name and I will merge it.

3ach:
Sorry, forgot to push those changes. They're up now!

zachzundel:
Hey @adambabik just checking that you don't need any more from me to get this merged!

3ach:
Hey @adambabik just checking that you don't need any more from me to get this merged!

jeluard:
@adambabik  Is this part of a release already?

adambabik:
@jeluard should it be based on `0.17.10-beta.1` or `0.19.x` is fine?

jeluard:
@adambabik  What is the difference? Latest sounds good I guess. Can it be used with status-react develop?

adambabik:
I am not sure. In status-react I still see `0.17` in develop but we should upgrade. I am preparing `0.19` and if it won't work I can also do `0.17`.

adambabik:
`0.19` is ready: https://github.com/status-im/status-go/releases/tag/0.19.0-beta.2

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/status-im/status-go">status-im/status-go</a></strong></div><div>The Status module that consumes go-ethereum. Contribute to status-im/status-go development by creating an account on GitHub.</div></blockquote>

jeluard:
@adambabik Thanks! Not sure what is needed status-react side now but @cammellos offered to help :)

3ach:
There is an [open pull request](https://github.com/status-im/status-react/pull/6893) in status-react which must be merged in to work with the current version of status-go.

jeluard:
@zachzundel Your PR must account to this new `status-go` version.

3ach:
My PR did include the necessary commits, but you asked me to remove them. @cammellos did say that they could remove the problematic `status-go` commits when cutting its next version.

jeluard:
@zachzundel The `status-react` PR must point to a `status-go` release that includes the necessary dependencies. `status-go` version including those has just been released.

jeluard:
@adambabik Is this version available in `status-react` develop ?

adambabik:
@jeluard I don't think so because there is some old version now: https://github.com/status-im/status-react/commit/ad99ed2a28f17dc5e3ec95d42fb389aa2f8595c6. Not sure if that's not a mistake. cc @cammellos

cammellos:
@adambabik that commit does not include probably these changes, we haven't moved to 0.19 yet, but there's an outstanding PR that should include everything but the PN notifications changes, probably best to wait for that to be merged. 

It's not clear what the naming should be, as 0.19 indicates a breaking change, but PN notifications commit is always excluded, so it's more of a patched version of 0.17.

PR #1331, opened 2018-12-21
Allow multiple bundles, change n devices & bundle refresh

cammellos:
Change to support sending multiple bundles, as needed for group chats,
limit number of devices to 3 as already done in the UI and refresh
bundle daily.

Also changes the way we pull the key to use `SelectedKeyPairID`, so it's compatible with decoupling whisper key/wallet


services/shhext/api.go
```diff
@@ -488,29 +488,19 @@ func (api *PublicAPI) SendGroupMessage(ctx context.Context, msg chat.SendGroupMe
...
-			return err
-		}
 
-		privateKey, err = api.service.w.GetPrivateKey(string(keyBytes))
-		if err != nil {
-			return err
-		}
+	privateKeyID := api.service.w.SelectedKeyPairID()
+	if privateKeyID == "" {
+		return errors.New("No key selected")
```
```suggestion
		return errors.New("no key selected")
```

Errors should be lowercase.

services/shhext/chat/protocol.go
```diff
@@ -146,6 +146,19 @@ func (p *ProtocolService) HandleMessage(myIdentityKey *ecdsa.PrivateKey, theirPu
 		p.addedBundlesHandler(addedBundles)
 	}
 
+	// Process bundles
+	if bundles := protocolMessage.GetBundles(); bundles != nil {
```
adambabik: `range` should handle `nil` just fine so you can skip `if`.

cammellos:
@adambabik thanks for the feedback, updated!

pedropombeiro (approved):

PR #1333, opened 2018-12-25
start discovery after sending "node.ready" signal

rasom:
Currently, it takes ~2.1-2.4s on any device to receive "node.ready" signal after `Statusgo.StartNode` call. It seems that [`n.gethNode.Server().Self()`](https://github.com/status-im/status-go/blob/52a1bdfed669718a61093bbcc0aa8da917d029f9/node/status_node.go#L199) takes ~2s.

This PR is rather an attempt to emphasize what's the reason of that 2s delay, so if it totally doesn't make sense don't hesitate to close it :). Although already tested on a real android device and doesn't make any harm at first glance.

### DIFF
--- a/api/backend.go
+++ b/api/backend.go
@@ -135,7 +135,13 @@ func (b *StatusBackend) startNode(config *params.NodeConfig) (err error) {
 	services := []gethnode.ServiceConstructor{}
 	services = appendIf(config.UpstreamConfig.Enabled, services, b.rpcFiltersService())
 
-	if err = b.statusNode.Start(config, services...); err != nil {
+	if err = b.statusNode.StartWithOptions(config, node.StartOptions{
+		Services: services,
+		// The peers discovery protocols are started manually after
+		// `node.ready` signal is sent.
+		// It was discussed in https://github.com/status-im/status-go/pull/1333.
+		StartDiscovery: false,
+	}); err != nil {
 		return
 	}
 	signal.SendNodeStarted()
--- a/api/backend.go
+++ b/api/backend.go
@@ -166,6 +166,10 @@ func (b *StatusBackend) startNode(config *params.NodeConfig) (err error) {
 
 	signal.SendNodeReady()
 
+	if err := b.statusNode.StartDiscovery(); err != nil {
+		return err
+	}
+
 	return nil
 }
 
--- a/node/status_node.go
+++ b/node/status_node.go
@@ -43,6 +43,7 @@ var (
 	ErrNoRunningNode          = errors.New("there is no running node")
 	ErrAccountKeyStoreMissing = errors.New("account key store is not set")
 	ErrServiceUnknown         = errors.New("service unknown")
+	ErrDiscoveryRunning       = errors.New("discovery is already running")
 )
 
 // StatusNode abstracts contained geth node and provides helper methods to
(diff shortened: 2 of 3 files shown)


api/backend.go
```diff
@@ -166,6 +166,10 @@ func (b *StatusBackend) startNode(config *params.NodeConfig) (err error) {
 
 	signal.SendNodeReady()
 
+	if err := b.statusNode.StartDiscovery(); err != nil {
```
adambabik, line 175: If this can return an error, I would call it before `signal.SendNodeReady()`. It would be strange if this method fails but `node.ready` signal is sent.

Another option is starting the discovery mechanism in another thread and sending a signal if it fails (here is another question whether the UI should be bothered at all).

adambabik:
I think a little bit more detailed research is needed to figure out what exactly is slow. `n.gethNode.Server().Self()` should not really be a problem as it's a getter. In the method you linked, though, a record is created and signed which might be slower. On the other hand, it uses known cryptographic functions that are used pretty much everywhere.

api/backend.go
rasom: @adambabik, on the client side we use "node.ready" only to ensure that calls like `Login`, `CreateAccount` and `VerifyAccountPassword` can be executed. Currently, all three calls require running node (I doubt that the last one should really require running node btw).

So what we need here is to have a signal after which we can proceed with `Login` and `CreateAccount`. And that signal should be sent after a minimal set of operations really required before these calls. If that will be another signal, not "node.ready" - that's fine.

rasom:
> I think a little bit more detailed research is needed to figure out what exactly is slow. n.gethNode.Server().Self()

yes please, that would be great. As I mentioned before, it takes 2s on any kind of devices currently, at least when we run it using Status.

In general, as i mentioned in a comment above
> So what we need here is to have a signal after which we can proceed with Login and CreateAccount. And that signal should be sent after a minimal set of operations really required before these calls. If that will be another signal, not "node.ready" - that's fine.

The problem is that node starting + Login call can take more than 10s together and we can't send messages, add public keys (thus join new chats), etc before these operations are finished, though UI part is ready in 1-2s.

api/backend.go
adambabik: Thanks for the explanation!

dshulyak:
I agree that it looks totally confusing. Signing is not free, but i don't understand how it can take 2s on mobile device. I isolated signing function, and it costs 481373 ns per op. The other operation that we repeat every time is a ephemeral key creation for communication with rendezvous server, but it takes even less (135608 per op).  So, if it helps lets just use this approach.

dshulyak (approved):

rasom:
@dshulyak from what i see it also tales 2s on desktop

adambabik:
I will take a look at this PR because some tests fail.

dshulyak:
Weird. I also tried to time how long it takes to start a node using statusd binary. It was 38.102626ms before node.ready signal is sent, with rendezvous and discv5 enabled.

mandrigin:
@dshulyak can it be related to ARM architecture somehow? Can we repeat the test on a mobile device w/o react native?

adambabik:
One thing that I found out is that the interface suffers with this change... Discovery is a part of `StatusNode` and it's strange that when it's enabled in the config it does not start with the node but needs to be started separately.

I will think about a different approach which will send a signal earlier but won't make the interface looks odd.

PR #1335, opened 2018-12-30
Validate Whisper DataDir to be relative to the main DataDir

### LINKED ISSUE #1287: Validate consistency between `DataDir` and `WhisperConfig.DataDir` when using JSON configs
# Problem

We came across a subtle issue with setting `DataDir` when running statusd. The problem is that when you run statusd without `-dir` flag but set `DataDir` in the JSON config, `WhisperConfig.DataDir` will still have a default value that is set for `-dir`.

# Implementation

We should validate that if `DataDir` is provided in a JSON config and Whisper is enabled, `WhisperConfig.DataDir` should be provided as well.

# Acceptance Criteria

1. The described problem does not occur anymore,
2. It's not possible to provide `DataDir` and not `WhisperConfig.DataDir` if Whisper is enabled in a JSON config file.

# Notes

More details: https://github.com/status-im/infra-eth-cluster/pull/19

adambabik:
Enforces `WhisperConfig.DataDir` to be a subpath of `DataDir`.

This is a bit stronger condition than described in #1287 but there is no valid reason to make Whisper data directory outside of the main data directory.

Closes #1287

### DIFF
--- a/params/config.go
+++ b/params/config.go
@@ -539,8 +539,8 @@ func (c *NodeConfig) Validate() error {
 	// Whisper's data directory must be relative to the main data directory
 	// if EnableMailServer is true.
 	if c.WhisperConfig.Enabled && c.WhisperConfig.EnableMailServer {
-		if !strings.Contains(c.WhisperConfig.DataDir, c.DataDir) {
-			return fmt.Errorf("WhisperConfig.DataDir must be relative to DataDir")
+		if !strings.HasPrefix(c.WhisperConfig.DataDir, c.DataDir) {
+			return fmt.Errorf("WhisperConfig.DataDir must start with DataDir fragment")
 		}
 	}
 
--- a/params/config.go
+++ b/params/config.go
@@ -514,6 +514,7 @@ func loadConfigFromAsset(name string, config *NodeConfig) error {
 //
 //   Key: 'TestStruct.TestField' Error:Field validation for 'TestField' failed on the 'required' tag
 //
+// nolint: gocyclo
 func (c *NodeConfig) Validate() error {
 	validate := NewValidator()
 
--- a/params/config.go
+++ b/params/config.go
@@ -531,15 +531,16 @@ func (c *NodeConfig) Validate() error {
 		return fmt.Errorf("both UpstreamConfig and LightEthConfig are enabled, but they are mutually exclusive")
 	}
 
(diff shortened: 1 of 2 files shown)


params/config.go
```diff
@@ -534,6 +536,14 @@ func (c *NodeConfig) Validate() error {
 		return err
 	}
 
+	// Whisper's data directory must be relative to the main data directory
+	// if EnableMailServer is true.
+	if c.WhisperConfig.Enabled && c.WhisperConfig.EnableMailServer {
+		if !strings.Contains(c.WhisperConfig.DataDir, c.DataDir) {
```
pedropombeiro: Maybe it would be better to use something like [filepath.Rel](https://golang.org/src/path/filepath/path.go?s=7689:7740#L253) to check for this?

```suggestion
		if _, err = filepath.Rel(c.DataDir, c.WhisperConfig.DataDir); err != nil {
```

pedropombeiro (commented):
Added a comment. Also, can't we have unit test to verify that validation fails when the Whisper data dir is not contained in the root data dir?

params/config.go
adambabik: I tried it but the problem with `Rel()` is that it tries to create a path always: https://play.golang.org/p/2VS3zuRfGSf. We can specify two totally different paths and it will try to figure out a relative path.

adambabik:
> Also, can't we have unit test to verify that validation fails when the Whisper data dir is not contained in the root data dir?

Sure we can.

params/config.go
pedropombeiro: Wouldn't it then be enough to test that `Rel()` doesn't return something that starts with `..`?

adambabik: Yes but what's the point of using `Rel()` then? I don't see any advantage of that. We explicitly want the Whisper dir path to start with the main data dir... And I have just realised I should use `strings.HasPrefix` instead of `strings.Contains` :D

pedropombeiro: Yeah, that was one of the main reasons I was looking for an alternative. I guess if we call `Clean` on both paths and use `strings.HasPrefix`, that would do most of the same job as `Rel()`.

adambabik: I would not call `Clean` and move this responsibility on user to provide clean paths.

pedropombeiro (approved):

PR #1336, opened 2018-12-31
Add GetContactCode, JoinPublicChats, LeavePublicChats calls

cammellos:
This commit adds a JoinPublishChats & LeavePublicChats
endpoints.
It also moves the logic from api.go to service.go, so that logic is kept
in a single place and we can offer the same endpoints through bindings &
rpc in an easier way.

### DIFF
--- a/services/shhext/api.go
+++ b/services/shhext/api.go
@@ -313,12 +313,12 @@ func (api *PublicAPI) SendPublicMessage(ctx context.Context, msg chat.SendPublic
 }
 
 // JoinPublicChats join a set of public chats deriving the key from the chat name
-func (api *PublicAPI) JoinPublicChats(ctx context.Context, msg chat.JoinPublicChatRPC) ([]string, error) {
+func (api *PublicAPI) JoinPublicChats(ctx context.Context, msg chat.JoinPublicChatsRPC) ([]string, error) {
 	return api.service.JoinPublicChats(msg.Chats)
 }
 
 // JoinPublicChats join a set of public chats deriving the key from the chat name
-func (api *PublicAPI) LeavePublicChats(ctx context.Context, msg chat.JoinPublicChatRPC) error {
+func (api *PublicAPI) LeavePublicChats(ctx context.Context, msg chat.JoinPublicChatsRPC) error {
 	return api.service.LeavePublicChats(msg.Chats)
 }
 
--- a/services/shhext/api.go
+++ b/services/shhext/api.go
@@ -11,9 +11,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/rlp"
 
-	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
-	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/p2p/enode"
 	"github.com/status-im/status-go/mailserver"
@@ -145,29 +143,21 @@ type SyncMessagesResponse struct {
 
 // PublicAPI extends whisper public API.
 type PublicAPI struct {
-	service   *Service
-	publicAPI *whisper.PublicWhisperAPI
-	log       log.Logger
+	service *Service
+	log     log.Logger
 }
 
 // NewPublicAPI returns instance of the public API.
 func NewPublicAPI(s *Service) *PublicAPI {
 	return &PublicAPI{
-		service:   s,
-		publicAPI: whisper.NewPublicWhisperAPI(s.w),
-		log:       log.New("package", "status-go/services/sshext.PublicAPI"),
+		service: s,
+		log:     log.New("package", "status-go/services/sshext.PublicAPI"),
 	}
 }
 
 // Post shamelessly copied from whisper codebase with slight modifications.
 func (api *PublicAPI) Post(ctx context.Context, req whisper.NewMessage) (hash hexutil.Bytes, err error) {
-	hash, err = api.publicAPI.Post(ctx, req)
-	if err == nil {
-		var envHash common.Hash
-		copy(envHash[:], hash[:]) // slice can't be used as key
-		api.service.tracker.Add(envHash)
-	}
-	return hash, err
+	return api.service.Post(ctx, req)
 }
 
 func (api *PublicAPI) getPeer(rawurl string) (*enode.Node, error) {
@@ -308,24 +298,7 @@ func (api *PublicAPI) SyncMessages(ctx context.Context, r SyncMessagesRequest) (
 
 // GetNewFilterMessages is a prototype method with deduplication
 func (api *PublicAPI) GetNewFilterMessages(filterID string) ([]*whisper.Message, error) {
-	msgs, err := api.publicAPI.GetFilterMessages(filterID)
-	if err != nil {
-		return nil, err
-	}
-
-	dedupMessages := api.service.deduplicator.Deduplicate(msgs)
-
-	if api.service.pfsEnabled {
-		// Attempt to decrypt message, otherwise leave unchanged
-		for _, msg := range dedupMessages {
-
-			if err := api.processPFSMessage(msg); err != nil {
-				return nil, err
-			}
-		}
-	}
-
-	return dedupMessages, nil
+	return api.service.GetNewFilterMessages(filterID)
 }
 
 // ConfirmMessagesProcessed is a method to confirm that messages was consumed by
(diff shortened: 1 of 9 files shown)


services/shhext/chat/rpc.go
```diff
@@ -28,3 +29,8 @@ type SendGroupMessageRPC struct {
 	Payload hexutil.Bytes
 	PubKeys []hexutil.Bytes
 }
+
+// JoinPublicChatRPC represents the RPC payload for the JoinPublicChat RPC method
+type JoinPublicChatRPC struct {
```
adambabik: `JoinPublicChatRPC` but contains plural `Chats`. Should we change it to `JoinPublicChatsRPC`?

services/shhext/service_test.go
```diff
@@ -72,6 +73,8 @@ type ShhExtSuite struct {
 }
 
 func (s *ShhExtSuite) SetupTest() {
+	rand.Seed(time.Now().UnixNano())
```
adambabik: Wouldn't it be better to use the same seed for every test run?

services/shhext/service_test.go
```diff
@@ -116,9 +119,69 @@ func (s *ShhExtSuite) SetupTest() {
...
+	service := s.services[0]
+	address := "address-1"
+
+	err := service.Login(address, "`090///\nhtaa\rhta9x8923)$$'23")
+	s.Require().NoError(err)
+
+	s.NotNil(service.protocol)
+
+	err = service.Logout()
+
```
adambabik: Unneeded blank line.

services/shhext/service_test.go
```diff
@@ -116,9 +119,69 @@ func (s *ShhExtSuite) SetupTest() {
...
+	filters, err := service.JoinPublicChats([]string{publicChat})
+	s.Require().NoError(err)
+
+	// Just smoke test it for now, no messages will be fetched as we
+	// don't publish
+	_, err = service.GetNewFilterMessages(filters[0])
+	s.Require().NoError(err)
+
+	err = service.Logout()
+
```
adambabik: Not needed blank line.

services/shhext/service_test.go
```diff
@@ -116,9 +119,69 @@ func (s *ShhExtSuite) SetupTest() {
...
+	filters, err := service.JoinPublicChats(publicChats)
+	s.Require().NoError(err)
+
+	// Check filters are saved
+	_, found := service.loadedFilters["test-1"]
+	s.True(found)
+	_, found = service.loadedFilters["test-2"]
+	s.True(found)
+
+	s.Require().NotNil(filters)
```
adambabik: This should be after
```go
	filters, err := service.JoinPublicChats(publicChats)
	s.Require().NoError(err)
```

services/shhext/service.go
```diff
@@ -251,3 +276,323 @@ func (s *Service) Stop() error {
 	s.tracker.Stop()
 	return nil
 }
+
+// JoinPublicChats join a set of public chats deriving the key from the chat name
+func (s *Service) JoinPublicChats(chats []string) ([]string, error) {
+
```
adambabik: Unneeded blank line.

services/shhext/service.go
```diff
@@ -251,3 +276,323 @@ func (s *Service) Stop() error {
...
+	var response []string
+	if s.loadedFilters == nil {
```
adambabik: Any reason it can't be initialized in the constructor?

services/shhext/service.go
```diff
@@ -251,3 +276,323 @@ func (s *Service) Stop() error {
...
+		s.loadedFilters = make(map[string]LoadedFilter)
+	}
+
+	for _, chatID := range chats {
```
adambabik, line 287: It looks like we can create multiple filters for the same chats. Is that desired?

services/shhext/service.go
```diff
@@ -251,3 +276,323 @@ func (s *Service) Stop() error {
...
+	}
+	return nil
+}
+
+// Post shamelessly copied from whisper codebase with slight modifications.
+func (s *Service) Post(ctx context.Context, req whisper.NewMessage) (hash hexutil.Bytes, err error) {
+	hash, err = s.publicAPI.Post(ctx, req)
+	if err == nil {
+		var envHash common.Hash
+		copy(envHash[:], hash[:]) // slice can't be used as key
```
adambabik: Can we fix this fragment? This comment does not make sense and I would check if `common.Hash` can't be initialized directly from a byte slice instead of using `copy`.

services/shhext/service.go
```diff
@@ -251,3 +276,323 @@ func (s *Service) Stop() error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	dedupMessages := s.deduplicator.Deduplicate(msgs)
+
+	if s.pfsEnabled {
+		// Attempt to decrypt message, otherwise leave unchanged
+		for _, msg := range dedupMessages {
+
```
adambabik: Unneeded blank line.

services/shhext/service.go
```diff
@@ -38,9 +42,16 @@ type EnvelopeEventsHandler interface {
...
+// LoadedFilter is a record containing data about an added filter on whisper
+type LoadedFilter struct {
+	SymKeyID string
+	FilterID string
+}
+
 // Service is a service that provides some additional Whisper API.
 type Service struct {
 	w              *whisper.Whisper
+	publicAPI      *whisper.PublicWhisperAPI
```
adambabik, line 57: I don't really like this idea. PublicAPI can have a reference to a service because it uses it (the API is just a facade) but the other way around it seems odd.

If there is no easy way to remove it now, let's add a comment with `TODO` explaining that we need to get rid of this.

services/shhext/service.go
```diff
@@ -251,3 +276,323 @@ func (s *Service) Stop() error {
...
+	for chat := range s.loadedFilters {
+		chats = append(chats, chat)
+
+	}
+
+	return s.LeavePublicChats(chats)
+}
+
+func (s *Service) processPFSMessage(msg *whisper.Message) error {
+
```
adambabik: Unneeded blank line.

services/shhext/service.go
```diff
@@ -251,3 +276,323 @@ func (s *Service) Stop() error {
...
+	// Add unencrypted payload
+	msg.Payload = response
+
+	return nil
+}
+
+// buildPublicMessage sends a public chat message to the underlying transport
+func (s *Service) buildPublicMessage(msg chat.SendPublicMessageRPC) (*whisper.NewMessage, error) {
+
```
adambabik: Unneeded blank line.

services/shhext/service.go
cammellos: It is a bit more complex, currently `SshExt` had a reference to whisper public api.

So it's ShhExtPublicAPI -> WhisperPublicAPI.

Effectively this means that any logic that uses the WhisperPublicAPI needs to go through the RPC public API (i.e. Post/GetNewFilterMessages), so logic becomes scattered around two files arbitrarly (service/and public api currently both have business logic), and currently Service is not encapsulated (publicapi uses private fields freely `api.service.*`) and logic is in both places.

This changes the behavior to be:

ShhExtPublicAPI -> ShhExtService -> WhisperPublicAPI , which basically is saying:
ShhExtPublicAPI offers a public api (method signatures), Service has all the business logic and knows about wrhisper, while the publicAPI does not (you could potentially keep the same public interface and change the transport layer without touching PublicAPI).

I am happy to change it, but I think we need better defined boundaries, what do you think?

adambabik:
@cammellos I was looking into the code and started wondering why we exchange our public bundle when sending public messages? The bundle seems unused because the payload is not encrypted.

services/shhext/service.go
adambabik: > So it's ShhExtPublicAPI -> WhisperPublicAPI.

That's fine in my opinion because `ShhExtPublicAPI` acts as a decorator on some `WhisperPublicAPI` methods and adds some more. It makes sense. But the scattered logic is bad. Not sure why it ended up in this way.

> ShhExtPublicAPI -> ShhExtService -> WhisperPublicAPI , which basically is saying:
ShhExtPublicAPI offers a public api (method signatures), Service has all the business logic and knows about wrhisper, while the publicAPI does not (you could potentially keep the same public interface and change the transport layer without touching PublicAPI).

Everything about `ShhExtPublicAPI` makes sense and if it could only use public methods of `ShhExtService` that would be even better (better encapsulation). Finally, `ShhExtService` could be a programmable interface and `ShhExtPublicAPI` only as a facade needed by the RPC server.

What I would like to change is to have `ShhExtPublicAPI -> ShhExtService -> WhisperService`. Unless `WhisperService` is not really well implemented and a lot of logic is put inside `WhisperPublicAPI`... If this is the case, let's just put a comment with explanation. If not, we can create an issue and tackle it in a different PR (I would not do it in this one).

cammellos:
correct, it is not being used for public messages, it s just a way to
propagate the bundle ( in addition to others), so that if they send us a
private message, bundle info are available

On Dec 31, 2018 1:54 PM, "Adam Babik" <notifications@github.com> wrote:

@cammellos <https://github.com/cammellos> I was looking into the code and
started wondering why we exchange our public bundle when sending public
messages? The bundle seems unused because the payload is not encrypted.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/status-im/status-go/pull/1336#issuecomment-450642717>,
or mute the thread

adambabik:
Not sure if this is a tradeoff we should accept :D But it's a different discussion. Personally, I am for separating public and private/group messages logic as much as possible. I assume it's just bundle exchange optimization in this case.

adambabik:
@cammellos and a final question. Is there anywhere a written plan which describes moving to this protobuf-based protocol for the mobile and desktop Status clients? You have been working on it for a bit now but is it already used? I think some update on discuss would be helpful as many people started writing their own clients.

cammellos:
the propagation of the bundle is not strictly speaking part of the protocol (it is always optional), and not tied to private nor public messages (if anything it s more related to public messages as the bundle is public). eventually can be replaced by swarm/ipfs, or any other persistence storage, in addition to other channels

cammellos:
@adambabik by moving to protobuf you mean using these encryption layer? ( groip chats and pairing are already using it) or you mean changing transit to protobuf?

adambabik:
@cammellos I meant using this encryption layer. I understand that we will keep Transit format for a bit and the payload for this protobuf-based encryption layer will be actually encoded using Transit. 

After we move the whole messaging logic (except for payload) to status-go, and as the encryption layer is payload agnostic, it will be easier to remove Transit in the next steps (not sure if removing Transit is even decided).

cammellos:
cool, so it is used already by 0.9.32 for group chats and pairing, and is
read only compatible for 1 to 1, there s a pr out but not reviewed yet to
make it togglable for 1 to 1. there is still some ux work an i want more
user feedback before enabling it for good

services/shhext/service_test.go
cammellos: The only thing I was thinking is that if we start publishing on a public channel, we want the channel to be different at each test run, otherwise we'd get clashes

services/shhext/service.go
cammellos: That was my initial though, but looking at the `Post` method  https://github.com/ethereum/go-ethereum/blob/a4af734328d50b9ea89405c7e5050065a8087946/whisper/whisperv6/api.go#L232 for example, it seems the separation is not clear enough

services/shhext/service.go
cammellos: Not sure, 
basically it is used for the key of a map,
`common.Hash` is basically an `[32]byte`, https://github.com/ethereum/go-ethereum/blob/3e8184006189b57e2d998313ef0fa75429b0c5bc/common/types.go#L47 , I have changed it to use `setBytes` https://github.com/ethereum/go-ethereum/blob/3e8184006189b57e2d998313ef0fa75429b0c5bc/common/types.go#L109, is that ok?

services/shhext/service_test.go
adambabik: Hm that's not clear to me. So we create public channels using `rand` module? Where is this code?

services/shhext/service.go
adambabik: Right... Ok, so let's leave it this way then. We can add a comment explaining why a service imports PublicAPI.

EDITED: I see you did already, thanks!

services/shhext/service.go
adambabik: Yes, it's better. Looks good!

services/shhext/service.go
```diff
@@ -532,7 +536,9 @@ func (s *Service) GetNewFilterMessages(filterID string) ([]*whisper.Message, err
 
 // Login initialize the protocol and join the chat for advertising the bundle
 func (s *Service) Login(address string, password string) error {
+	s.loadedFilters = make(map[string]LoadedFilter)
 	return s.InitProtocol(address, password)
+
```
adambabik, line 541: This one empty line slipped in :)
```suggestion
```

adambabik (approved):

services/shhext/api.go
```diff
@@ -336,193 +309,32 @@ func (api *PublicAPI) ConfirmMessagesProcessed(messages []*whisper.Message) erro
...
-	whisperMessage := chat.PublicMessageToWhisper(msg, protocolMessage)
-	whisperMessage.SymKeyID = symKeyID
+// JoinPublicChats join a set of public chats deriving the key from the chat name
+func (api *PublicAPI) JoinPublicChats(ctx context.Context, msg chat.JoinPublicChatsRPC) ([]string, error) {
+	return api.service.JoinPublicChats(msg.Chats)
+}
 
-	// And dispatch
-	return api.Post(ctx, whisperMessage)
+// JoinPublicChats join a set of public chats deriving the key from the chat name
```
pedropombeiro, line 320:
```suggestion
// LeavePublicChats leaves a set of public chats deriving the key from the chat name
```

services/shhext/service.go
```diff
@@ -251,3 +279,324 @@ func (s *Service) Stop() error {
...
+		return nil, err
+	}
+	response = append(response, hash)
+
+	return response, nil
+}
+
+// SendPublicMessage sends a public chat message to the underlying transport
+func (s *Service) SendPublicMessage(ctx context.Context, msg chat.SendPublicMessageRPC) (hexutil.Bytes, error) {
+
```
pedropombeiro, line 399:
```suggestion
```

pedropombeiro (approved):
lgtm, added minor comments

services/shhext/service_test.go
cammellos: @adambabik below in the test, we use it to create a random string , which we treat as a publicChat which we then join

adambabik: > The only thing I was thinking is that if we start publishing on a public channel, we want the channel to be different at each test run, otherwise we'd get clashes

The better strategy would be to send messages, remember their hashes and request messages back checking if the expected hashes got received.

Also, the messages really shouldn't leave the local node so I don't see a reason to clash with anything.

If we can avoid using `rand` in tests, let's do it because it might lead to issues with debugging and reproducing failures as well as increases a chance to get flaky tests.

services/shhext/service.go
```diff
@@ -251,3 +279,324 @@ func (s *Service) Stop() error {
...
+			return nil, err
+		}
+
+		topic := chat.ToWhisperTopic(chatID)
+		topics := [][]byte{topic[:]}
+
+		filterOpts := &whisper.Filter{
+			KeySym:   symKey,
+			Topics:   topics,
+			AllowP2P: true,
```
gravityblast, line 310: `AllowP2P` allows to receive p2p messages that can be expired, do we need it here for public chats?

gravityblast: oh maybe yes if it's used for mail servers too

cammellos: correct, we use it for mailservers

iurimatias:
Iam correct in reading this is being done as a whisper extension? In that case should this type of thing really be done at this layer? stuff like joining&leaving channels should be done as a sub protocol and not part of whisper itself.

cammellos:
@iurimatias essentially we are just pushing some logic from status-react to status-go and exposing it through an RPC endpoint., as part of an effort to have status-react to act a simple UI, while status-go handling the interactions with whisper.
We are not changing the way whisper works or the way we currently join a public channel (generate symkey from public-chat-name, add filter), that remains the same, and we are still acting as a client of the whisper protocol.

You do have a point about the naming, as these are not meant to extend whisper, but rather to offer an API to other client that is aware of how status uses whisper, so probably worth changing as misleading.

iurimatias:
@cammellos that logic should be in a common client lib such as http://github.com/status-im/status-js not in (one possible) backend. Status-go should just be responsible for just being a good whisper node, not exposing more apis like this & dealing with arbitrary app protocol logic, that should be up to each client and or application built on top of the platform what type of messages they want to do.

cammellos:
@iurimatias there has been much discussion about this in the past, you might find this helpful https://discuss.status.im/t/a-status-api-proposal/220 . 

my understanding is that we first want to build a public api, and eventually carve it out into a separate library (probably similar to https://github.com/status-im/status-js) once we are happy with it, so that status-go is just a whisper node, while the api would handle the application logic (join chats etc), so it seems fairly similar to what you described.

Maybe someone else can join in the discussion @adambabik ?

<blockquote><div> Status.im</div><div><strong><a href="https://discuss.status.im/t/a-status-api-proposal/220">A Status API proposal</a></strong></div><div>During the Basel meeting, the idea of moving the communication protocol to status-go was brought up again. Weither some code should be on go or clojurescript side is a recurring topic of discussion at Status. Historically the rule of thumb was that status-go should only contain patches required to work on mobile and everything else should be on status-react side. The idea was that it should be easy to swap ethereum implementation.  This post is a proposal for the definition of a stable, document...</div></blockquote>
<blockquote><div> GitHub</div><div><strong><a href="https://github.com/status-im/status-js">status-im/status-js</a></strong></div><div>Status Javascript Client (WIP). Contribute to status-im/status-js development by creating an account on GitHub.</div></blockquote>

adambabik:
@iurimatias That's true that status-go has become much more than just a node (but you can't call it a Whisper node only, its role is also to sync the chain using LES). However, it's not that simple because status-go has never been built as a library but rather as a integral part of status-react. I am not saying that's good but I guess in early days it was just faster.

We want to extract the protocol outside of status-go. There will be a layer which will be responsible only for creating the payload (transport agnostic) and additionally Whisper transport adapters which will allow to exchange the payload. Finally, there will be a glue code (in status-go) which will add status protocol as a JSON-RPC API.

adambabik:
Is this PR still a thing or we can close it?

cammellos:
@adambabik we can close it for now, I don't have time to port status-react to the new endpoints yet, will re-open when I have some time

PR #1337, opened 2019-01-03
Accept status service config directly from json input

dshulyak:
Recently i extended status service config with more options and it seems impractical to duplicate all those options. So this change allows to provide status config directly.

I will also submit PR for status-react.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -235,7 +235,7 @@ func activatePersonalService(stack *node.Node, config *params.NodeConfig) error 
 }
 
 func activateStatusService(stack *node.Node, config *params.NodeConfig) error {
-	if !config.StatusServiceEnabled {
+	if !config.EnableStatusService {
 		logger.Info("Status service api is disabled")
 		return nil
 	}
@@ -327,17 +327,7 @@ func activateShhService(stack *node.Node, config *params.NodeConfig, db *leveldb
 		if err := ctx.Service(&whisper); err != nil {
 			return nil, err
 		}
-
-		config := &shhext.ServiceConfig{
-			DataDir:                 config.BackupDisabledDataDir,
-			InstallationID:          config.InstallationID,
-			Debug:                   config.DebugAPIEnabled,
-			PFSEnabled:              config.PFSEnabled,
-			MailServerConfirmations: config.MailServerConfirmations,
-		}
-
-		svc := shhext.New(whisper, shhext.EnvelopeSignalHandler{}, db, config)
-		return svc, nil
+		return shhext.New(whisper, shhext.EnvelopeSignalHandler{}, db, config.ShhextConfig), nil
 	})
 }
 
--- a/params/config.go
+++ b/params/config.go
@@ -180,9 +180,6 @@ type NodeConfig struct {
 
 	// DataDir is the file system folder the node should use for any data storage needs.
 	DataDir string `validate:"required"`
-	// BackupDisabledDataDir is the file system folder the node should use for any data storage needs that it doesn't want backed up.
-	BackupDisabledDataDir string `validate:"required"`
-	PFSEnabled            bool
 
 	// KeyStoreDir is the file system folder that contains private keys.
 	KeyStoreDir string `validate:"required"`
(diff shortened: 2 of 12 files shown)


node/node.go
```diff
@@ -327,17 +327,7 @@ func activateShhService(stack *node.Node, config *params.NodeConfig, db *leveldb
...
-			DataDir:                 config.BackupDisabledDataDir,
-			InstallationID:          config.InstallationID,
-			Debug:                   config.DebugAPIEnabled,
-			PFSEnabled:              config.PFSEnabled,
-			MailServerConfirmations: config.MailServerConfirmations,
-		}
-
-		svc := shhext.New(whisper, shhext.EnvelopeSignalHandler{}, db, config)
-		return svc, nil
+		return shhext.New(whisper, shhext.EnvelopeSignalHandler{}, db, &config.StatusServiceConfig), nil
```
adambabik: Any reason to pass `&config.StatusServiceConfig` by reference?

params/config.go
```diff
@@ -292,20 +292,28 @@ type NodeConfig struct {
...
+}
 
+// StatusServiceConfig defines options used by shhext service.
+type StatusServiceConfig struct {
+	// Enabled should be true if we want to use status service.
+	Enabled bool
+	// BackupDisabledDataDir is the file system folder the node should use for any data storage needs that it doesn't want backed up.
+	BackupDisabledDataDir string `validate:"required"`
+	// InstallationId id of the current installation
+	InstallationID string
```
adambabik: Are you sure PFS related stuff should be here? It looks like `StatusService` encapsulate options for `services/status` and `services/sshext`. Unless there is an idea to merge them.

node/node.go
dshulyak: i don't think so, it just was like that before my changes

params/config.go
dshulyak: yeah it would be cleaner to pass two separate configs, even internally same object uses them

dshulyak: it looks like it makes sense to keep a single config cause all PFS related functionality is kept under same shhext namespace. So splitting config into two would cause more confusion

params/config.go
```diff
@@ -292,20 +292,28 @@ type NodeConfig struct {
...
+	Debug          bool
```
adambabik: I would keep it named `DebugAPIEnabled`. `Debug` even in `StatusServiceConfig` is not descriptive enough.

adambabik (commented):
As `StatusServiceConfig` contains status service and shhext stuff and PFS maybe we should call it a bit differently? I am not convinced this change makes it more clear.

One proposal is to call it `ServicesConfig` (note plural) and prefix fields appropriately, for example, `StatusSvcEnabled` and `ShhextSvcPFSEnabled` and so on.

dshulyak:
> One proposal is to call it ServicesConfig (note plural) and prefix fields appropriately, for example, StatusSvcEnabled and ShhextSvcPFSEnabled and so on.

i actually missed that StatusService is in the separate namespace. I this variable used to enable shhext namespace. I will move it into the other config option.

dshulyak:
@adambabik take a look at this change please

params/config.go
```diff
@@ -292,20 +295,31 @@ type NodeConfig struct {
...
-	DebugAPIEnabled bool
-
 	// MailServerRegistryAddress is the MailServerRegistry contract address
 	MailServerRegistryAddress string
+}
 
+// ShhextConfig defines options used by shhext service.
+type ShhextConfig struct {
+	// BackupDisabledDataDir is the file system folder the node should use for any data storage needs that it doesn't want backed up.
+	BackupDisabledDataDir string `validate:"required"`
```
mandrigin, line 305: as in iOS back up? so it isn't in the Documents folder?
what is the difference between this and some temporary storage?

is it for temp files or for very sensitive data?

mandrigin (approved):
looks good, one question on naming

params/config.go
dshulyak: i guess this directory is for sensitive data. it is used for PFS service

mandrigin: so maybe we can call it something like `SensitiveDataDir` or something similar?

dshulyak: sounds ok to me. i can rename. @cammellos what do you think?

cammellos: I am ok with renaming it, to me personally is clearer to call it `backupdisableddatadir`, as that's what we are interested in (confidential information might also be stored in `~/Documents`, although they will be backed up), on android is literally called `no_backup`, but don't feel strongly about it to be honest

adambabik (approved):

adambabik:
@mandrigin @dshulyak is there a ticket in status-react to make it compatible?

PR #1341, opened 2019-01-07
mailserver: refactor mailserver's rate limiter

adambabik:
This is a prep PR to clean up some implementation details of the mailserver rate limiter. In the next PR, we want to include data pruning into the mail server but some naming was inconvenient and could cause some issues in the future.

Important changes:
- [x] make fields and type names more explicit.


ghost:
### Pull Request Checklist

mailserver/db_key.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+const (
+	// DBKeyLength is a size of the envelope key.
+	DBKeyLength = common.HashLength + timestampLength
+)
+
+// DBKey key to be stored on db.
```
mandrigin: typo
```suggestion
// DBKey key to be stored in a db.
```

mailserver/db_key.go
```diff
@@ -0,0 +1,44 @@
...
+	raw       []byte
+}
+
+// Bytes returns a bytes representation of the DBKey.
+func (k *DBKey) Bytes() []byte {
+	return k.raw
+}
+
+// NewDBKey creates a new DBKey with the given values.
+func NewDBKey(t uint32, h common.Hash) *DBKey {
```
mandrigin:
```suggestion
func NewDBKey(timestamp uint32, h common.Hash) *DBKey {
```

of alias uint32 as `timestamp` type, improves readability

mailserver/db_key.go
```diff
@@ -0,0 +1,44 @@
...
+	k.timestamp = t
+	k.hash = h
+	k.raw = make([]byte, DBKeyLength)
+	binary.BigEndian.PutUint32(k.raw, k.timestamp)
+	copy(k.raw[4:], k.hash[:])
+	return &k
+}
+
+// NewDBKeyFromBytes creates a DBKey from a byte slice.
+func NewDBKeyFromBytes(b []byte) *DBKey {
```
mandrigin: what if len(b) < 4 bytes?
maybe add validation and call it `ParseBytes`?

mailserver/mailserver.go
```diff
@@ -450,16 +388,16 @@ func (s *WMailServer) SyncMail(peer *whisper.Peer, request whisper.SyncMailReque
 // exceedsPeerRequests in case limit its been setup on the current server and limit
 // allows the query, it will store/update new query time for the current peer.
 func (s *WMailServer) exceedsPeerRequests(peer []byte) bool {
-	s.muLimiter.RLock()
-	defer s.muLimiter.RUnlock()
+	s.muRateLimiter.RLock()
+	defer s.muRateLimiter.RUnlock()
 
-	if s.limiter != nil {
+	if s.rateLimiter != nil {
```
mandrigin: I would invert this `if` to avoid unnecessary nesting

mandrigin (approved):
a few nitpicks, but in general looks good!

mailserver/db_key.go
mandrigin: or we can just accept `time.Time` and convert it to timestamp only for the `raw` calculation

pedropombeiro (approved):
lgtm once Igor's comments are addressed

mailserver/db_key.go
adambabik: > of alias uint32 as timestamp type, improves readability

It's only seemingly better because saying this you know that `timestamp` is `uint32` :) When one would like to call `NewDBKey` later, they will need to figure out what `timestamp` is just to see it's `uint32`. We had this discussion and if there is no methods for a custom type, it's rarely worth to do aliasing.

> or we can just accept time.Time and convert it to timestamp only for the raw calculation

In most of the use cases, it will mean converting `int64` or `uint32` to `time.Time` and then back to `uint32` in `NewDBKey`.

I like changing `t` to `timestamp`. I think it solves readability without introducing other issues.

mailserver/mailserver.go
adambabik: I can change it :)

mandrigin (approved):
:+1:

mailserver/db_key.go
adambabik: That was a bit more problematic because many methods needed to carry over this returned error. However, each call was already guarded with a check if this slice has a proper length. I solved it with adding `mustNewDBKeyFromBytes` which is unexported.

mailserver/mailserver.go
```diff
@@ -450,17 +388,19 @@ func (s *WMailServer) SyncMail(peer *whisper.Peer, request whisper.SyncMailReque
...
-		}
-		s.limiter.add(peerID)
+	s.muRateLimiter.RLock()
+	defer s.muRateLimiter.RUnlock()
+
+	if s.rateLimiter == nil {
+		return false
+	}
+
+	peerID := string(peer)
```
mandrigin, line 398: tiny nitpick, feel free to ignore

```suggestion
	if s.rateLimiter.IsAllowed(peerID) {
	    s.rateLimiter.Add(peerID)
	    return false
	}
	
	log.Info("peerID exceeded the number of requests per second")
	return true
     }
```

mandrigin (approved):

PR #1342, opened 2019-01-07
mailserver: build pruning into WMailServer

adambabik:
MailServer pruning was implemented as a separate command but that required stopping a mail server and executing the command manually. This change builds pruning into MailServer and can be set using `MailServerDataRetention` in `WhisperConfig`.

Merge https://github.com/status-im/status-go/pull/1341 first as it branches out from that PR.


mailserver/cleaner.go
```diff
@@ -1,30 +1,91 @@
...
+
+	period time.Duration
+	cancel chan struct{}
 }
 
-// NewCleanerWithDB returns a new Cleaner for db
-func NewCleanerWithDB(db dbImpl) *Cleaner {
-	return &Cleaner{
+// NewCleanerWithDB returns a new cleaner for db.
+func newCleanerWithDB(db dbImpl, retention time.Duration) *cleaner {
```
mandrigin: `newClearnerForDB`?

mailserver/cleaner.go
```diff
@@ -1,30 +1,91 @@
...
+		}
 	}
 }
 
-// Prune removes messages sent between lower and upper timestamps and returns how many has been removed
-func (c *Cleaner) Prune(lower, upper uint32) (int, error) {
+// Prune removes messages sent between lower and upper timestamps
+// and returns how many have been removed.
+func (c *cleaner) Prune(lower, upper uint32) (int, error) {
+	fmt.Printf("============= prune from %d to %d\n", lower, upper)
```
mandrigin: debug leftover? convert to log

mailserver/cleaner.go
```diff
@@ -1,30 +1,91 @@
...
+			return
+		}
 	}
 }
 
-// Prune removes messages sent between lower and upper timestamps and returns how many has been removed
-func (c *Cleaner) Prune(lower, upper uint32) (int, error) {
+// Prune removes messages sent between lower and upper timestamps
+// and returns how many have been removed.
+func (c *cleaner) Prune(lower, upper uint32) (int, error) {
```
mandrigin: do we ever call it with different lower bound? maybe we can remove this parameter?
also, why not accepting `time.Time` here, why uint32?


something like 
```
PruneEntriesOlderThan(time.Now().Add(-c.retention))
```

mandrigin (changes requested):
looks good overall, a few smaller questions

mailserver/cleaner.go
adambabik: I have usually seen `WithSomething` pattern if a parameter is assigned to a field in a struct.

mailserver/cleaner.go
adambabik: I can refactor it, I didn't really focus here on improving the interface. But yeah, it looks like the only call is here:
```go
lower := uint32(0)
upper := uint32(time.Now().Add(-c.retention).Unix())
if _, err := c.Prune(lower, upper); err != nil {
	log.Error("failed to prune data", "err", err)
}
```
so we can definitely change this method to accept a single param.

mailserver/cleaner.go
mandrigin: yeah, my idea was that your comment says that it is a cleaner *for* a db, which is logical.
to me, having something *with* something means that it uses the db for it's own reasons.
maybe we can call it just `NewDBCleaner` then it is obvious that it should accept a db somewhere in params?

mailserver/cleaner.go
mandrigin: yeah, and it makes sense to also make it accept `time.Time` too, that will avoid your `.Unix()` call when passing a param.

mailserver/cleaner.go
adambabik: That makes sense. `NewDBCleaner` seems like the most suitable option here.

mandrigin (approved):
👍 great job!

mandrigin (approved):

PR #1346, opened 2019-01-11
Unlock wallet and chat keys

gravityblast:
With this PR, in preparation to the full keys decoupling, I changed `lib/library.go` and `account.Manager` to treat wallet and chat addresses/pubkeys as they were two different physical keys in the keystore.

Creating and recovering an account now returns `walletAddress` and `walletPubKey`, but also `chatAddress` and `chatPubKey`,

`chatAddress` is the ethereum address of the chat key, which is only used as a key ID to unlock the key at login. 

Important changes:
- [x] CreateAccount returns wallet and chat  address and pub key
- [x] RecoverAccount returns wallet and chat  address and pub key
- [x] SelectAccount requires a walletAddress and chatAddress to unlock 2 different keys for the current account.

### DIFF
--- a/account/accounts.go
+++ b/account/accounts.go
@@ -51,12 +51,12 @@ func NewManager(geth GethServiceProvider) *Manager {
 // BIP44-compatible keys are generated: CKD#1 is stored as account key, CKD#2 stored as sub-account root
 // Public key of CKD#1 is returned, with CKD#2 securely encoded into account key file (to be used for
 // sub-account derivations)
-func (m *Manager) CreateAccount(password string) (address, pubKey, mnemonic string, err error) {
+func (m *Manager) CreateAccount(password string) (walletAddress, walletPubKey, chatAddress, chatPubKey, mnemonic string, err error) {
 	// generate mnemonic phrase
 	mn := extkeys.NewMnemonic()
 	mnemonic, err = mn.MnemonicPhrase(extkeys.EntropyStrength128, extkeys.EnglishLanguage)
 	if err != nil {
-		return "", "", "", fmt.Errorf("can not create mnemonic seed: %v", err)
+		return "", "", "", "", "", fmt.Errorf("can not create mnemonic seed: %v", err)
 	}
 
 	// Generate extended master key (see BIP32)
@@ -66,16 +66,19 @@ func (m *Manager) CreateAccount(password string) (address, pubKey, mnemonic stri
 	// for expert users, to be able to add a passphrase to the generation of the seed.
 	extKey, err := extkeys.NewMaster(mn.MnemonicSeed(mnemonic, ""))
 	if err != nil {
-		return "", "", "", fmt.Errorf("can not create master extended key: %v", err)
+		return "", "", "", "", "", fmt.Errorf("can not create master extended key: %v", err)
 	}
 
 	// import created key into account keystore
-	address, pubKey, err = m.importExtendedKey(extkeys.KeyPurposeWallet, extKey, password)
+	walletAddress, walletPubKey, err = m.importExtendedKey(extkeys.KeyPurposeWallet, extKey, password)
 	if err != nil {
-		return "", "", "", err
+		return "", "", "", "", "", err
 	}
 
-	return address, pubKey, mnemonic, nil
+	chatAddress = walletAddress
+	chatPubKey = walletPubKey
+
+	return walletAddress, walletPubKey, chatAddress, chatPubKey, mnemonic, nil
 }
 
 // CreateChildAccount creates sub-account for an account identified by parent address.
@@ -140,21 +143,24 @@ func (m *Manager) CreateChildAccount(parentAddress, password string) (address, p
 
 // RecoverAccount re-creates master key using given details.
 // Once master key is re-generated, it is inserted into keystore (if not already there).
-func (m *Manager) RecoverAccount(password, mnemonic string) (address, pubKey string, err error) {
+func (m *Manager) RecoverAccount(password, mnemonic string) (walletAddress, walletPubKey, chatAddress, chatPubKey string, err error) {
 	// re-create extended key (see BIP32)
 	mn := extkeys.NewMnemonic()
 	extKey, err := extkeys.NewMaster(mn.MnemonicSeed(mnemonic, ""))
 	if err != nil {
-		return "", "", ErrInvalidMasterKeyCreated
+		return "", "", "", "", ErrInvalidMasterKeyCreated
 	}
 
 	// import re-created key into account keystore
-	address, pubKey, err = m.importExtendedKey(extkeys.KeyPurposeWallet, extKey, password)
+	walletAddress, walletPubKey, err = m.importExtendedKey(extkeys.KeyPurposeWallet, extKey, password)
 	if err != nil {
 		return
 	}
 
-	return address, pubKey, nil
+	chatAddress = walletAddress
+	chatPubKey = walletPubKey
+
+	return walletAddress, walletPubKey, chatAddress, chatPubKey, nil
 }
 
 // VerifyAccountPassword tries to decrypt a given account key file, with a provided password.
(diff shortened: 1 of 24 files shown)


account/accounts_test.go
```diff
@@ -130,17 +131,25 @@ func TestManagerTestSuite(t *testing.T) {
...
 	require.NoError(t, err)
-	require.NotEmpty(t, addr)
-	require.NotEmpty(t, pubKey)
+	require.NotEmpty(t, chatAddr)
+	require.NotEmpty(t, chatPubKey)
+	require.NotEmpty(t, chatAddr)
+	require.NotEmpty(t, chatPubKey)
 	require.NotEmpty(t, mnemonic)
 
+	// Before the complete decouple of the keys, wallet and chat keys are the same
```
pedropombeiro:
```suggestion
	// Before the complete decoupling of the keys, wallet and chat keys are the same
```

account/accounts_test.go
```diff
@@ -252,15 +278,24 @@ func (s *ManagerTestSuite) TestSelectAccount() {
...
-			s.Equal(selectedWalletAccount.AccountKey, selectedChatAccount.AccountKey)
+			if testCase.expectedError == nil {
+				s.Equal(selectedWalletAccount.AccountKey, selectedChatAccount.AccountKey)
+				s.NoError(walletErr)
+				s.NoError(chatErr)
+			} else {
+				s.Nil(selectedWalletAccount)
+				s.Nil(selectedChatAccount)
+				s.Error(ErrNoAccountSelected)
+				s.Error(ErrNoAccountSelected)
```
pedropombeiro: Duplicate assert

services/status/api.go
```diff
@@ -54,14 +54,16 @@ type SignupRequest struct {
 
 // SignupResponse : json response returned by status_signup.
 type SignupResponse struct {
-	Address  string `json:"address"`
-	Pubkey   string `json:"pubkey"`
-	Mnemonic string `json:"mnemonic"`
+	Address     string `json:"address"`
```
pedropombeiro: Do we want to rename this to `WalletAddress` and just leave the JSON identifier as `address`?
```suggestion
	WalletAddress     string `json:"address"`
```

t/e2e/accounts/accounts_test.go
```diff
@@ -225,19 +243,27 @@ func (s *AccountsTestSuite) TestSelectedAccountOnRestart() {
 	s.NoError(s.Backend.StopNode())
 
 	// make sure that account is still selected
-	selectedAccount, err = s.Backend.AccountManager().SelectedWalletAccount()
+	selectedWalletAccount, err = s.Backend.AccountManager().SelectedWalletAccount()
+	s.NoError(err)
```
pedropombeiro: Do we want to continue the test if `SelectedWalletAccount` returns an error?
```suggestion
	s.Require().NoError(err)
```

t/e2e/whisper/whisper_test.go
```diff
@@ -47,7 +47,7 @@ func (s *WhisperTestSuite) TestWhisperFilterRace() {
 	s.True(ok, "identity not injected")
 
 	// account2
-	_, accountKey2, err := accountManager.AddressToDecryptedAccount(TestConfig.Account2.Address, TestConfig.Account2.Password)
+	_, accountKey2, err := accountManager.AddressToDecryptedAccount(TestConfig.Account2.WalletAddress, TestConfig.Account2.Password)
```
pedropombeiro: Why do we use the `WalletAddress` here?

pedropombeiro (approved):
lgtm, added a few comments

account/accounts.go
```diff
@@ -51,12 +51,12 @@ func NewManager(geth GethServiceProvider) *Manager {
 // BIP44-compatible keys are generated: CKD#1 is stored as account key, CKD#2 stored as sub-account root
 // Public key of CKD#1 is returned, with CKD#2 securely encoded into account key file (to be used for
 // sub-account derivations)
-func (m *Manager) CreateAccount(password string) (address, pubKey, mnemonic string, err error) {
+func (m *Manager) CreateAccount(password string) (walletAddress, walletPubKey, chatAddress, chatPubKey, mnemonic string, err error) {
```
adambabik: At this point, we should return a struct and an error. It's too many return parameters.

account/accounts.go
```diff
@@ -140,21 +143,24 @@ func (m *Manager) CreateChildAccount(parentAddress, password string) (address, p
 
 // RecoverAccount re-creates master key using given details.
 // Once master key is re-generated, it is inserted into keystore (if not already there).
-func (m *Manager) RecoverAccount(password, mnemonic string) (address, pubKey string, err error) {
+func (m *Manager) RecoverAccount(password, mnemonic string) (walletAddress, walletPubKey, chatAddress, chatPubKey string, err error) {
```
adambabik: The same here.

api/backend.go
```diff
@@ -421,17 +421,18 @@ func (b *StatusBackend) reSelectAccount() error {
 	return nil
 }
 
-// SelectAccount selects current account, by verifying that address has corresponding account which can be decrypted
-// using provided password. Once verification is done, decrypted key is injected into Whisper (as a single identity,
+// SelectAccount selects current wallet and chat accounts, by verifying that each address has corresponding account which can be decrypted
+// using provided password. Once verification is done, the decrypted chat key is injected into Whisper (as a single identity,
 // all previous identities are removed).
-func (b *StatusBackend) SelectAccount(address, password string) error {
+func (b *StatusBackend) SelectAccount(walletAddress, chatAddress, password string) error {
```
adambabik, line 427: Maybe we can create a type that represents an account? In our case it would be a struct that contains wallet and chat addresses? I am not sure about it yet but something tells me that we shouldn't double arguments in the public interface but rather wrap it into some logical entity...

t/e2e/accounts/accounts_test.go
```diff
@@ -136,17 +137,23 @@ func (s *AccountsTestSuite) TestRecoverAccount() {
...
+	s.T().Logf("Account created: {walletAddress: %s, walletKey: %s, chatAddress: %s, chatKey: %s, mnemonic:%s}",
+		walletAddress, walletPubKey, chatAddress, chatPubKey, mnemonic)
 
 	// try recovering using password + mnemonic
-	addressCheck, pubKeyCheck, err := s.Backend.AccountManager().RecoverAccount(TestConfig.Account1.Password, mnemonic)
+	walletAddressCheck, walletPubKeyCheck, chatAddressCheck, chatPubKeyCheck, err := s.Backend.AccountManager().RecoverAccount(TestConfig.Account1.Password, mnemonic)
 	s.NoError(err, "recover acc failed")
-	s.False(address != addressCheck || pubKey != pubKeyCheck, "incorrect accound details recovered")
+
+	s.True(walletAddress == walletAddressCheck &&
```
adambabik: With an `Account` struct, it could be written like this:
```go
account, mnemonic, err := s.Backend.AccountManager().CreateAccount(TestConfig.Account1.Password)
accountCheck, err := s.Backend.AccountManager().RecoverAccount(TestConfig.Account1.Password, mnemonic)
s.EqualValues(accountCheck, account)
```

gravityblast:
@adambabik I updated the code adding an `account.Info` type, that is returned after creating and recovering an account. Using the same struct also to select an account might be confusing, so if you agree I would update that in the next PRs when I'll introduce the login with keycard method. At that point we might need something like you said because logging in will happen via address/password or injecting a key for whisper and an address for the wallet. what do you think?

account/accounts.go
```diff
@@ -51,12 +51,13 @@ func NewManager(geth GethServiceProvider) *Manager {
 // BIP44-compatible keys are generated: CKD#1 is stored as account key, CKD#2 stored as sub-account root
 // Public key of CKD#1 is returned, with CKD#2 securely encoded into account key file (to be used for
 // sub-account derivations)
-func (m *Manager) CreateAccount(password string) (address, pubKey, mnemonic string, err error) {
+func (m *Manager) CreateAccount(password string) (*Info, string, error) {
```
adambabik: Up to you but returning a value instead of pointer should also work just fine. This is a very simple struct so pointer is probably not needed.

lib/types.go
```diff
@@ -64,10 +64,18 @@ func (e APIError) Error() string {
 
 // AccountInfo represents account's info.
 type AccountInfo struct {
-	Address  string `json:"address"`
-	PubKey   string `json:"pubkey"`
+	// Address is the wallet address
+	Address string `json:"address"`
```
adambabik: We keep it `Address` and `Pubkey` for backward compatibility? Do we want to change it in the future? If so, I would duplicate `Address` and `PubKey` and add `WalletAddress` and `WalletPubKey`. Then, status-react can upgrade in any moment.

adambabik: Also, in `SignupResponse` we changed the fields and break compatibility so it's a bit inconsistent.

services/status/api.go
```diff
@@ -54,18 +54,27 @@ type SignupRequest struct {
 
 // SignupResponse : json response returned by status_signup.
 type SignupResponse struct {
-	Address  string `json:"address"`
-	Pubkey   string `json:"pubkey"`
-	Mnemonic string `json:"mnemonic"`
+	WalletAddress string `json:"address"`
```
adambabik: Compatibility breaks here...?

gravityblast: it shouldn't break because the json field has still the same name, but I added now both address and walletAddress

lib/types.go
gravityblast: thank you @adambabik , I added both address and walletAddress so it's more clear

account/accounts.go
gravityblast: good idea, I updated the code

gravityblast:
thank you @adambabik, I updated the code after your comments

lib/types.go
```diff
@@ -64,10 +64,14 @@ func (e APIError) Error() string {
 
 // AccountInfo represents account's info.
 type AccountInfo struct {
-	Address  string `json:"address"`
-	PubKey   string `json:"pubkey"`
-	Mnemonic string `json:"mnemonic"`
-	Error    string `json:"error"`
+	Address       string `json:"address"`
```
adambabik:
```suggestion
	Address       string `json:"address"` // DEPRECATED
```

lib/types.go
```diff
@@ -64,10 +64,14 @@ func (e APIError) Error() string {
...
+	PubKey        string `json:"pubkey"`
```
adambabik:
```suggestion
	PubKey        string `json:"pubkey"` // DEPRECATED
```

adambabik (approved):
Sorry, I missed the fix. All looks great! I proposed two tiny suggestions :)

PR #1347, opened 2019-01-14
Revert "Change `NotifyUsers` to allow only data JSON element. Part of…

cammellos:
As discussed this revert notifyUsers change for now, we can re-apply with a different name in order to keep compatibility.


mandrigin (approved):
lgtm

pedropombeiro (approved):

PR #1348, opened 2019-01-15
Ensure that similar history requests won't be made using requestMessages

dshulyak:
This method will return error if multiple requests with same topics were made without any delay (by default 3 seconds, can be configured with ShhextConfig). With force=true in requests parameters described behavior will be ignored.

### DIFF
--- a/services/shhext/api.go
+++ b/services/shhext/api.go
@@ -80,6 +80,9 @@ type MessagesRequest struct {
 	// Timeout is the time to live of the request specified in seconds.
 	// Default is 10 seconds
 	Timeout time.Duration `json:"timeout"`
+
+	// Force ensures that requests will bypass enforced delay.
+	Force bool `json:"force"`
 }
 
 func (r *MessagesRequest) setDefaults(now time.Time) {
@@ -223,11 +226,17 @@ func (api *PublicAPI) RequestMessages(_ context.Context, r MessagesRequest) (hex
 	if err != nil {
 		return nil, err
 	}
+	hash := envelope.Hash()
+	if !r.Force {
+		err = api.service.requestsRegistry.Register(hash, r.Topics)
+		if err != nil {
+			return nil, err
+		}
+	}
 
 	if err := shh.RequestHistoricMessagesWithTimeout(mailServerNode.ID().Bytes(), envelope, r.Timeout*time.Second); err != nil {
 		return nil, err
 	}
-	hash := envelope.Hash()
 	return hash[:], nil
 }
 
--- a/services/shhext/requests.go
+++ b/services/shhext/requests.go
@@ -1,0 +1,83 @@
+package shhext
+
+import (
+	"fmt"
+	"hash/fnv"
+	"sync"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
+	whisper "github.com/status-im/whisper/whisperv6"
+)
+
+const (
+	// defaultRequestsDelay will be used in RequestsRegistry if no other was provided.
+	defaultRequestsDelay = 3 * time.Second
+)
+
+type requestMeta struct {
+	timestamp time.Time
+	lastUID   common.Hash
+}
+
+// NewRequestsRegistry creates instance of the RequestsRegistry and returns pointer to it.
+func NewRequestsRegistry(delay time.Duration) *RequestsRegistry {
+	return &RequestsRegistry{
+		delay:        delay,
+		uidToTopics:  map[common.Hash]common.Hash{},
+		byTopicsHash: map[common.Hash]requestMeta{},
+	}
+}
+
+// RequestsRegistry keeps map for all requests with timestamp when they were made.
+type RequestsRegistry struct {
+	mu           sync.Mutex
+	delay        time.Duration
+	uidToTopics  map[common.Hash]common.Hash
+	byTopicsHash map[common.Hash]requestMeta
+}
+
+// Register request with given topics. If request with same topics was made in less then configured delay then error
+// will be returned.
+func (r *RequestsRegistry) Register(uid common.Hash, topics []whisper.TopicType) error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	topicsHash := topicsToHash(topics)
+	if meta, exist := r.byTopicsHash[topicsHash]; exist {
+		if time.Since(meta.timestamp) < r.delay {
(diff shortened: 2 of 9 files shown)


ghost:
### Pull Request Checklist

vendor/github.com/status-im/whisper/whisperv6/message.go
```diff
@@ -75,6 +75,7 @@ type ReceivedMessage struct {
 
 	SymKeyHash   common.Hash // The Keccak256Hash of the key
 	EnvelopeHash common.Hash // Message envelope hash to act as a unique id
+	History      bool
```
mandrigin, line 78: Unnecessary change?

mandrigin:
@dshulyak it would be nice to see a few tests of this method if possible.

vendor/github.com/status-im/whisper/whisperv6/message.go
dshulyak: right, should not be here

cmd/statusd/flags_test.go
```diff
@@ -83,7 +83,7 @@ func TestStatusFlag(t *testing.T) {
 		}
 
 		require.Nil(t, err, msg)
-		require.Equal(t, s.enabled, c.StatusServiceEnabled, msg)
```
mandrigin: 👍 for the rename, it is much better now

mandrigin:
@dshulyak one question: why is that a separate method instead of it being an optional parameter for the existing method?

mandrigin (approved):

dshulyak:
> @dshulyak one question: why is that a separate method instead of it being an optional parameter for the existing method?

this is definitely an option. i don't have strong arguments for the separate method. it just seemed like a cleaner way to isolate this new feature. maybe that is not true.

mandrigin:
@dshulyak just "safe" is a bit generic word. I would use a parameter, like `force: true` to have a single method.
Then normal `shhext_requestMessages` will be limited, where `shhext_requestMessages(force = true)` will ignore the limitation.

services/shhext/api.go
```diff
@@ -178,7 +178,18 @@ func (api *PublicAPI) getPeer(rawurl string) (*enode.Node, error) {
...
+	return api.requestMessages(ctx, r, false)
+}
+
+// SafeRequestMessages provides same functionality as RequestMessages. But if similar request was made within
+// configured delay an error will be returned.
+func (api *PublicAPI) SafeRequestMessages(ctx context.Context, r MessagesRequest) (hexutil.Bytes, error) {
+	return api.requestMessages(ctx, r, true)
+}
+
+// RequestMessages sends a request for historic messages to a MailServer.
```
pedropombeiro:
```suggestion
// requestMessages sends a request for historic messages to a MailServer.
```

services/shhext/requests.go
```diff
@@ -0,0 +1,85 @@
...
+
+// Register request with given topics. If request with same topics was made in less then configured delay then error
+// will be returned.
+func (r *RequestsRegistry) Register(uid common.Hash, topics []whisper.TopicType) error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	topicsHash := topicsToHash(topics)
+	if meta, exist := r.byTopicsHash[topicsHash]; exist {
+		if time.Since(meta.timestamp) < r.delay {
+			return fmt.Errorf("another request with same topics was sent in less then %s ago. Please wait more time, or use shhext_requestMessages", r.delay)
```
pedropombeiro:
```suggestion
			return fmt.Errorf("another request with the same topics was sent less than %s ago. Please wait for a bit longer, or use shhext_requestMessages", r.delay)
```

pedropombeiro (approved):
lgtm

dshulyak:
variant with force looks better indeed. will apply

services/shhext/requests.go
```diff
@@ -0,0 +1,85 @@
...
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	topicsHash, exist := r.uidToTopics[uid]
+	if !exist {
+		return
+	}
+	delete(r.uidToTopics, uid)
+	meta := r.byTopicsHash[topicsHash]
+	// another request was made.
+	if meta.lastUID == uid {
```
adambabik, line 71: I guess `meta.lastUID == uid` means that there were no other requests? If so then the comment is inaccurate.

services/shhext/requests.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// topicsToHash returns non-cryptographic hash of the topics.
+func topicsToHash(topics []whisper.TopicType) common.Hash {
+	hash := fnv.New32()
+	for i := range topics {
+		hash.Write(topics[i][:]) // never returns error per documentation
+	}
+	rst := make([]byte, 0, 32)
+	rst = hash.Sum(rst)
```
adambabik: Won't `rst := hash.Sum(nil)` produce the same result without a need to specify the slice explicitly? Questions arise when looking at this fragment like "why `32`?" and "what if `New32` changes to something else" etc.

adambabik (approved):

adambabik:
> Then normal shhext_requestMessages will be limited, where shhext_requestMessages(force = true) will ignore the limitation.

This will introduce incompatibility, I believe. All existing calls to `shhext_requestMessages` will need to be fixed and send this additional param. It's fine but we need to remember about it and bump the `VERSION` properly.

mandrigin:
@adambabik yes, it will introduce incompatibility, but that is what we want because the current behaviour forces it to use too much trafic.

adambabik:
@mandrigin then it's cool :) wanted to make sure everyone is aware.

mandrigin:
@adambabik yep, that is important to tell explicitly, thanks for pointing this out!

services/shhext/api.go
```diff
@@ -178,7 +178,18 @@ func (api *PublicAPI) getPeer(rawurl string) (*enode.Node, error) {
...
+func (api *PublicAPI) requestMessages(_ context.Context, r MessagesRequest, ensureDelay bool) (hexutil.Bytes, error) {
```
mandrigin: hmm, isn't `ensureDelay` the opposite of `force`?

dshulyak: yes, but that code was without latest changes. with new changes i am using same method with force parameter

mandrigin:
linter looks unhappy :(

PR #1351, opened 2019-01-17
Ensure that all discovery compoenents are set to nil on stop

dshulyak:
Previously we always tried to stop discovery if it was enabled according to configuration. Now we check if it was started, because it can be started separately from a node.

### DIFF
--- a/node/status_node.go
+++ b/node/status_node.go
@@ -335,12 +335,13 @@ func (n *StatusNode) Stop() error {
 
 // stop will stop current StatusNode. A stopped node cannot be resumed.
 func (n *StatusNode) stop() error {
-	if n.discoveryEnabled() {
+	if n.isDiscoveryRunning() {
 		if err := n.stopDiscovery(); err != nil {
-			n.log.Error("Error stopping the PeerPool", "error", err)
+			n.log.Error("Error stopping the discovery components", "error", err)
 		}
 		n.register = nil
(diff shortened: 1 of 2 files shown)


node/status_node.go
```diff
@@ -337,10 +337,11 @@ func (n *StatusNode) Stop() error {
 func (n *StatusNode) stop() error {
 	if n.discoveryEnabled() {
 		if err := n.stopDiscovery(); err != nil {
-			n.log.Error("Error stopping the PeerPool", "error", err)
+			n.log.Error("Error stopping the discovery compoenents", "error", err)
```
mandrigin:
```suggestion
			n.log.Error("Error stopping the discovery components", "error", err)
```

mandrigin (approved):
lgtm, one typo though

rasom (approved):

adambabik (approved):

node/status_node.go
```diff
@@ -337,10 +337,11 @@ func (n *StatusNode) Stop() error {
 func (n *StatusNode) stop() error {
 	if n.discoveryEnabled() {
```
adambabik: Maybe we can check here if discovery is also started? Because with the recent changes to the API, it's not obvious. Discovery might be enabled in the config but it does not mean it was started with the node. I know it's not thread-safe but it's better than not having it at all.

Issue #238, opened 2017-08-07
Outline actual cross-projects architecture


tiabc:
Overall architecture is necessary to ensure we're all on the same page and where different integration problems may arise. This is also the first step to facilitate QA test `status-go`.

Infrastructure documentation would also be nice: how the project is built, how it communicates with `status-react` (and development workflow of that) and what our scaleway cluster is used for.

tiabc:
In slack:
> So high-level design is the following. `status-go` is cross-compiled for every platform using xgo. `status-react` calls it through JNI using signals defined in `library.go`. Those signals invoke `StatusAPI` which is our application Facade.
> 
> Under the hood, it contains different structures moved from `go-ethereum` and modified for our needs. We also have our customised version of `go-ethereum`:  https://github.com/status-im/go-ethereum/commits/status/1.6.7-stable and an issue to figure out which changes are still required and get rid of the rest: https://github.com/status-im/status-go/issues/238
> 
> As to your question about regression tests, we're only starting to let test team in our project and have an issue about that: https://github.com/status-im/status-go/issues/239
> 
> One important concept is our js jail. It's an isolated environment for JS code execution based on Otto VM. Jail consists of multiple jail cells - each refers to a particular chat in `status-react`. `status-react` (or dapp, user or user bot) can execute arbitrary JS in chat. All that calls are executed in our jail cells.
> Details on bindings `status-react` uses can be found here: https://github.com/status-im/status-go/wiki/Notes-on-Bindings
> 
> This is what I find most important things. You may glance through the ongoing issues on the zenhub board (once you have github access) and have a feeling what's currently going on. Meanwhile, you can familiarise yourself with pull requests.
> 
> Questions are welcome. By asking them you'll help to document the existing architecture from a newcomer's point of view.

divan:
Some initial design drawing is here: https://docs.google.com/drawings/d/1D2wukqYWockdMe3zQK4Y600nAIxp9weV5AdWCuCtzgI/edit

gruberb:
@divan Are those drawings available without having to have a  Google account?

divan:
@gruberb I’m afraid not. They’re also aren’t complete yet. I should probably review it and prepare for usage with wiki.

pedropombeiro:
@divan what should we do with this issue? It's been open for months now, should it be closed?

skyf0cker:
[...]

PR #1352, opened 2019-01-17
Add `SendDataNotification` method

pedropombeiro:
This PR adds again the changes to support the PN metadata leak fix PR (status-im/status-react#7268), but this time as a renamed method.

Part of status-im/status-react#7268

### DIFF
--- a/api/backend.go
+++ b/api/backend.go
@@ -7,7 +7,6 @@ import (
 	"math/big"
 	"sync"
 
-	fcmlib "github.com/NaySoftware/go-fcm"
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/log"
@@ -464,10 +463,12 @@ func (b *StatusBackend) SelectAccount(address, password string) error {
 }
 
 // NotifyUsers sends push notifications to users.
-func (b *StatusBackend) NotifyUsers(message string, payload fcmlib.NotificationPayload, tokens ...string) error {
-	err := b.newNotification().Send(message, payload, tokens...)
+func (b *StatusBackend) NotifyUsers(dataPayloadJSON string, tokens ...string) error {
+	log.Debug("sending push notification")
+
+	err := b.newNotification().Send(dataPayloadJSON, tokens...)
 	if err != nil {
-		b.log.Error("Notify failed", "error", err)
+		b.log.Error("NotifyUsers failed", "dataPayloadJSON", dataPayloadJSON, "error", err)
 	}
 
 	return err
--- a/api/backend.go
+++ b/api/backend.go
@@ -462,13 +462,23 @@ func (b *StatusBackend) SelectAccount(address, password string) error {
 	return nil
 }
 
-// NotifyUsers sends push notifications to users.
-func (b *StatusBackend) NotifyUsers(dataPayloadJSON string, tokens ...string) error {
-	log.Debug("sending push notification")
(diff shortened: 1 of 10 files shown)


mandrigin (approved):
👍

api/backend.go
```diff
@@ -473,6 +473,18 @@ func (b *StatusBackend) NotifyUsers(message string, payload fcmlib.NotificationP
 	return err
 }
 
+// SendMessageToUsers sends a message to users, which can be used to display a push notification.
+func (b *StatusBackend) SendMessageToUsers(dataPayloadJSON string, tokens ...string) error {
```
adambabik: As we expose it on `StatusBackend` maybe we can call this method less general? `SendMessageToUsers` indicates sending messages not notifications.

```suggestion
func (b *StatusBackend) SendNotifications(dataPayloadJSON string, tokens ...string) error {
```

lib/library.go
```diff
@@ -468,6 +468,46 @@ func NotifyUsers(message, payloadJSON, tokensArray *C.char) (outCBytes *C.char)
 	return
 }
 
+// SendMessageToUsers sends messages to users represented by given tokens.
+//export SendMessageToUsers
+func SendMessageToUsers(dataPayloadJSON, tokensArray *C.char) (outCBytes *C.char) {
```
adambabik: Similar issue here.

api/backend.go
adambabik: Also what is this `dataPayloadJSON`. Don't we want to make it a struct here? If it does not matter what's insight (a bit strange), then let's at least comment what should be in this payload.

mandrigin:
@PombeirP do we need this method if we merged the PN upgrade? Wouldn't it be better to revert the revert of NotifyUser's change?

pedropombeiro:
> @PombeirP do we need this method if we merged the PN upgrade? Wouldn't it be better to revert the revert of NotifyUser's change?

I asked Roman what he'd prefer and he said we should just add a second method just in case and remove the old one later, but it's true that it's much cleaner if we get rid of the cruft and just replace the method (even if we rename it).

api/backend.go
pedropombeiro: > SendMessageToUsers indicates sending messages not notifications.

In a sense we're indeed no longer sending notifications to users (we changed from notifications API to messaging API), but rather a generic FCM message that is captured by the RN app and then used to create a local PN. That was the reason for the name. We can rename it to `SendDataNotification` which more closely reflects what it does.

pedropombeiro: > Also what is this dataPayloadJSON. Don't we want to make it a struct here? If it does not matter what's insight (a bit strange), then let's at least comment what should be in this payload.

Yeah, this is up to status-react what they want to put in here, I added a comment giving a general overview of how it looks.

pedropombeiro: In fact, the only reason this method doesn't live in status-react is that the FCM library that exists for Clojure doesn't support the methods we need (at some point we can probably just fork it and create an upstream PR).

pedropombeiro:
Build for status-react: https://ci.status.im/job/status-go/job/parallel/238/ 

<blockquote><div><strong><a href="https://ci.status.im/job/status-go/job/parallel/238/">status-go » parallel #238 [Jenkins]</a></strong></div></blockquote>

api/backend.go
pedropombeiro: @adambabik is the current state OK or do I need to make changes?

adambabik (approved):
LGTM. Thanks for the code comments!

PR #1357, opened 2019-01-21
Introduce new blocking requestMessagesSync api request

dshulyak:
New api request block until expired or completed event is received from whisper backend. It accept configuration for retries, that has three options: max retries, base timeout and step timeout. Step timeout increases base timeout on every retry. 

Additionally connection manager now accept configuration option to wait for more then 1 expired event before changing a server.

I want to extend requestMessagesSync with an option to wait for switching a mail server, e.g. max server switches. That will ensure that server was switched that time before returning an error to a caller. I am not exactly sure how to wait for those switches, so that change will come as a separate PR.

### DIFF
--- a/services/shhext/api.go
+++ b/services/shhext/api.go
@@ -180,6 +180,58 @@ func (api *PublicAPI) getPeer(rawurl string) (*enode.Node, error) {
 	return enode.ParseV4(rawurl)
 }
 
+// RetryConfig specifies configuration for retries with timeout and max amount of retries.
+type RetryConfig struct {
+	BaseTimeout time.Duration
+	// StepTimeout defines duration increase per each retry.
+	StepTimeout time.Duration
+	MaxRetries  int
+}
+
+// RequestMessagesSync repeats MessagesRequest using configuration in retry conf.
+func (api *PublicAPI) RequestMessagesSync(conf RetryConfig, r MessagesRequest) error {
+	shh := api.service.w
+	events := make(chan whisper.EnvelopeEvent, 10)
+	sub := shh.SubscribeEnvelopeEvents(events)
+	defer sub.Unsubscribe()
+	var (
+		requestID hexutil.Bytes
+		err       error
+		retries   int
+	)
+	for retries <= conf.MaxRetries {
+		r.Timeout = conf.BaseTimeout + conf.StepTimeout*time.Duration(retries)
+		// FIXME this weird conversion is required because MessagesRequest expects seconds but defines time.Duration
+		r.Timeout = time.Duration(int(r.Timeout.Seconds()))
+		requestID, err = api.RequestMessages(context.Background(), r)
+		if err != nil {
+			return err
+		}
+		err = waitForExpiredOrCompleted(common.BytesToHash(requestID), events)
+		if err == nil {
+			return nil
+		}
+		retries++
+		api.log.Error("History request failed with %s. Making retry #%d", retries)
+	}
+	return fmt.Errorf("failed to request messages after %d retries", retries)
+}
+
+func waitForExpiredOrCompleted(requestID common.Hash, events chan whisper.EnvelopeEvent) error {
+	for {
+		ev := <-events
+		if ev.Hash != requestID {
+			continue
+		}
+		switch ev.Event {
+		case whisper.EventMailServerRequestCompleted:
+			return nil
+		case whisper.EventMailServerRequestExpired:
+			return errors.New("request expired")
+		}
+	}
+}
+
 // RequestMessages sends a request for historic messages to a MailServer.
 func (api *PublicAPI) RequestMessages(_ context.Context, r MessagesRequest) (hexutil.Bytes, error) {
 	api.log.Info("RequestMessages", "request", r)
--- a/services/shhext/mailservers/connmanager.go
+++ b/services/shhext/mailservers/connmanager.go
@@ -39,11 +39,12 @@ type p2pServer interface {
 }
 
 // NewConnectionManager creates an instance of ConnectionManager.
-func NewConnectionManager(server p2pServer, whisper EnvelopeEventSubscbriber, target int, timeout time.Duration) *ConnectionManager {
+func NewConnectionManager(server p2pServer, whisper EnvelopeEventSubscbriber, target, maxFailures int, timeout time.Duration) *ConnectionManager {
 	return &ConnectionManager{
 		server:           server,
 		whisper:          whisper,
 		connectedTarget:  target,
+		maxFailures:      maxFailures,
 		notifications:    make(chan []*enode.Node),
 		timeoutWaitAdded: timeout,
 	}
@@ -60,6 +61,7 @@ type ConnectionManager struct {
 	notifications    chan []*enode.Node
 	connectedTarget  int
 	timeoutWaitAdded time.Duration
+	maxFailures      int
 }
 
 // Notify sends a non-blocking notification about new nodes.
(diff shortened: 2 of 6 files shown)


services/shhext/mailservers/connmanager.go
```diff
@@ -106,19 +108,24 @@ func (ps *ConnectionManager) Start() {
...
 					delete(requests, ev.Hash)
 				case whisper.EventMailServerRequestExpired:
 					_, exist := requests[ev.Hash]
 					if !exist {
 						continue
 					}
+					failuresPerServer[ev.Peer]++
 					log.Debug("request to a mail server expired, disconncet a peer", "address", ev.Peer)
-					state.nodeDisconnected(ev.Peer)
+					if failuresPerServer[ev.Peer] == ps.maxFailures {
```
mandrigin: I would always use `>=` here for future-proofing (if we add `failuresPerServer[ev.Peer]++` somewhere else or by mistake)

mandrigin (approved):
looks good! also kudos for tests!

services/shhext/api.go
```diff
@@ -180,6 +180,60 @@ func (api *PublicAPI) getPeer(rawurl string) (*enode.Node, error) {
 	return enode.ParseV4(rawurl)
 }
 
+// RetryConf specifies configuration for retries with timeout and max amount of retries.
+type RetryConf struct {
```
pedropombeiro: Not a big deal, but shouldn't we call it `RetryConfig`, as in `ShhextConfig`?

services/shhext/api.go
```diff
@@ -180,6 +180,60 @@ func (api *PublicAPI) getPeer(rawurl string) (*enode.Node, error) {
...
+		hid, err = api.RequestMessages(context.Background(), r)
+		if err != nil {
+			return err
+		}
+		err = waitExpiredOrCompleted(common.BytesToHash(hid), events)
+		if err == nil {
+			return nil
+		}
+		retries++
+		api.log.Error("History request failed with %s. Making retry N %d", retries)
```
pedropombeiro:
```suggestion
		api.log.Error("History request failed with %s. Making retry #%d", retries)
```

services/shhext/mailservers/connmanager.go
```diff
@@ -106,19 +108,24 @@ func (ps *ConnectionManager) Start() {
 			case ev := <-events:
 				processPeerEvent(state, ev)
 			case ev := <-whisperEvents:
-				// TODO what about completed but with error? what about expired envelopes?
+				// TODO treat failued request the same way as expired
```
pedropombeiro:
```suggestion
				// TODO treat failed requests the same way as expired ones
```

services/shhext/service_test.go
```diff
@@ -449,3 +451,91 @@ func waitForHashInTracker(track *tracker, hash common.Hash, state EnvelopeState,
...
+	go func() {
+		msg, err := s.remoteRW.ReadMsg()
+		s.Require().NoError(err)
+		msg.Discard()
+	}()
+	s.Require().EqualError(s.localAPI.RequestMessagesSync(RetryConf{
+		BaseTimeout: time.Second,
+	}, MessagesRequest{
+		MailServerPeer: s.localNode.String(),
+	}), "failed to request messages after 1 retries")
```
pedropombeiro, line 512: Just curious, shouldn't this error message be something like `unexpectedly managed to retrieve messages`?

pedropombeiro (approved):

services/shhext/mailservers/connmanager.go
adambabik: Or even better set an invariant which will check and notify (`log.Error`?) in case `failuresPerServer[ev.Peer] > ps.maxFailures`.

services/shhext/api.go
```diff
@@ -180,6 +180,60 @@ func (api *PublicAPI) getPeer(rawurl string) (*enode.Node, error) {
...
+	}
+	return fmt.Errorf("failed to request messages after %d retries", retries)
+}
+
+func waitExpiredOrCompleted(hid common.Hash, events chan whisper.EnvelopeEvent) error {
```
```suggestion
func waitForExpiredOrCompleted(hid common.Hash, events chan whisper.EnvelopeEvent) error {
```

services/shhext/api.go
```diff
@@ -180,6 +180,60 @@ func (api *PublicAPI) getPeer(rawurl string) (*enode.Node, error) {
...
+}
+
+// RequestMessagesSync repeats MessagesRequest using configuration in retry conf.
+func (api *PublicAPI) RequestMessagesSync(conf RetryConf, r MessagesRequest) error {
+	shh := api.service.w
+	events := make(chan whisper.EnvelopeEvent, 10)
+	sub := shh.SubscribeEnvelopeEvents(events)
+	defer sub.Unsubscribe()
+	var (
+		hid     hexutil.Bytes
```
adambabik: What does `hid` stand for?

services/shhext/service_test.go
```diff
@@ -449,3 +451,91 @@ func waitForHashInTracker(track *tracker, hash common.Hash, state EnvelopeState,
...
+			attempt++
+			msg, err := s.remoteRW.ReadMsg()
+			s.Require().NoError(err)
+			if attempt < target {
+				msg.Discard()
+				continue
+			}
+			var e whisper.Envelope
+			s.Require().NoError(msg.Decode(&e))
+			p2p.Send(s.remoteRW, 125, whisper.CreateMailServerRequestCompletedPayload(e.Hash(), common.Hash{}, []byte{}))
```
adambabik: Magic number `125`.

adambabik (approved):

services/shhext/api.go
dshulyak: hash id :) probably better name is requestID

dshulyak:
Regarding counting mail server switches in this method. It makes flow of this method significantly harder to understand, and i don't see that it fixes any new issues. We can work around slow mail server or not responding mail server using generic retries and connection manager that switches mail servers based on aggregated stats.

We anyway will have to retry indefinitely so iterating over all mail servers will be done transparently and centralized way. As for the visible warnings to a user, it makes sense to show two of them:
- messages are not synced
- no connection to mail server

So, i suggest to retry this method all the time when there is a connection to any mail server. If this method returns error show a warning that messages are not synced, and retry this method again. Technically client can simply retry this method indefinitely, we just need to respond with meaningful message so that it can be shown as a warning to a user.

As for the solved issues:
1. Slow mail server will be fixed by providing sufficient number of retries to this method, so that timeout will gradually increase. E.g. we start from 5s timeout and increase it by 3s with each failure.  Bigger problem is that we don't track progress of this method, but it will be addressed differently.
2. Not responding mail server will be fixed because we will change mail server based on aggregated stats, e.g. if 2 requests failed we will switch to another mail server.
3. If there is no connected mail server we will be showing a warning and can even stop retrying this method during that time.

services/shhext/mailservers/connmanager.go
mandrigin: yeap, that might be a good idea

mandrigin:
I think it sounds reasonable, user just cares about these two scenarios.

mandrigin:
` retry indefinitely ` — I'm not sure I understand this bit. this code retries until maxRetries is reached. what exactly do you mean about indefinite retries?

dshulyak:
> retry indefinitely — I'm not sure I understand this bit. this code retries until maxRetries is reached. what exactly do you mean about indefinite retries?

i meant on the caller side, e.g. status-react will have to retry indefinitely anyway

mandrigin:
@dshulyak linter is unhappy :(

services/shhext/service_test.go
dshulyak: why? this one is the opposite, method failed to request messages and notifies caller about that

PR #1358, opened 2019-01-21
Login with keycard

gravityblast:
In this PR we allow logging in with keycard.
At login `status-react` creates a secure channel to communicate with the keycard and verifies the PIN.
If login with the keycard is successful, `status-react` inject the whisper key downloaded from the keycard to whisper. To do that, it calls `LoginWithKeycard` passing the whisper private key that will be injected in whisper. It also passes another key used as password for sqlite if PFS is enabled.

Important changes:
- [x] add `LoginWithKeycard` in `lib/library.go` 
- [x] add `InjectChatAccount` to `Backend` and use the received key for the whisper identity.
- [x] add `SetChatAccount` to account manager and use the received key for the `selectedChatAccount`

### DIFF
--- a/account/accounts.go
+++ b/account/accounts.go
@@ -249,7 +249,7 @@ func (m *Manager) SelectAccount(walletAddress, chatAddress, password string) err
 }
 
 // SetChatAccount initializes selectedChatAccount with privKey
-func (m *Manager) SetChatAccount(privKey *ecdsa.PrivateKey) error {
+func (m *Manager) SetChatAccount(privKey *ecdsa.PrivateKey) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
@@ -265,8 +265,6 @@ func (m *Manager) SetChatAccount(privKey *ecdsa.PrivateKey) error {
 		Address:    address,
 		AccountKey: key,
 	}
-
-	return nil
 }
 
 // SelectedWalletAccount returns currently selected wallet account
--- a/account/accounts.go
+++ b/account/accounts.go
@@ -248,8 +248,8 @@ func (m *Manager) SelectAccount(walletAddress, chatAddress, password string) err
 	return nil
 }
 
-// InjectChatAccount initializes selectedChatAccount with privKey
-func (m *Manager) InjectChatAccount(privKey *ecdsa.PrivateKey) error {
+// SetChatAccount initializes selectedChatAccount with privKey
+func (m *Manager) SetChatAccount(privKey *ecdsa.PrivateKey) error {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
--- a/account/accounts.go
+++ b/account/accounts.go
@@ -1,6 +1,7 @@
 package account
 
 import (
+	"crypto/ecdsa"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -14,6 +15,8 @@ import (
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/pborman/uuid"
+
 	"github.com/status-im/status-go/extkeys"
 )
 
@@ -241,6 +244,27 @@ func (m *Manager) SelectAccount(walletAddress, chatAddress, password string) err
 
 	m.selectedWalletAccount = selectedWalletAccount
 	m.selectedChatAccount = selectedChatAccount
+
+	return nil
+}
+
+// InjectChatAccount initializes selectedChatAccount with privKey
+func (m *Manager) InjectChatAccount(privKey *ecdsa.PrivateKey) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	address := crypto.PubkeyToAddress(privKey.PublicKey)
+	id := uuid.NewRandom()
+	key := &keystore.Key{
+		Id:         id,
+		Address:    address,
+		PrivateKey: privKey,
+	}
+
+	m.selectedChatAccount = &SelectedExtKey{
+		Address:    address,
+		AccountKey: key,
+	}
 
 	return nil
 }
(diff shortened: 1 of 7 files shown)


account/accounts.go
```diff
@@ -245,6 +248,27 @@ func (m *Manager) SelectAccount(walletAddress, chatAddress, password string) err
 	return nil
 }
 
+// InjectChatAccount initializes selectedChatAccount with privKey
```
adambabik: Maybe `SetChatAccount`? "Inject" suggests that it is added permanently or to some storage but it just sets `m.selectedChatAccount`.

api/backend.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
+	ethcrypto "github.com/ethereum/go-ethereum/crypto"
```
adambabik, line 12: Is `gethcrypto` alias needed?

api/backend.go
```diff
@@ -485,6 +486,52 @@ func (b *StatusBackend) SendDataNotification(dataPayloadJSON string, tokens ...s
...
+	if err != nil {
+		return err
+	}
+
+	chatAccount, err := b.accountManager.SelectedChatAccount()
+	if err != nil {
+		return err
+	}
+
+	whisperService, err := b.statusNode.WhisperService()
```
adambabik, line 507: Any interaction with Whisper is done here because we use the same key pair for chat and Whisper still, right? If so, let's add a comment explaining that.

api/backend.go
```diff
@@ -485,6 +486,52 @@ func (b *StatusBackend) SendDataNotification(dataPayloadJSON string, tokens ...s
...
+		return err
+	}
+
+	if whisperService != nil {
+		st, err := b.statusNode.ShhExtService()
+		if err != nil {
+			return err
+		}
+
+		if err := st.InitProtocol(chatAccount.Address.Hex(), encryptionKeyHex); err != nil {
```
adambabik, line 524: Is `InitProtocol` idempotent? What if the protocol is already initialized?

lib/library.go
```diff
@@ -327,6 +327,14 @@ func Login(address, password *C.char) *C.char {
 	return makeJSONResponse(err)
 }
 
+// LoginWithKeycard initializes an account with a chat key and encryption key used for PFS.
+// It purges all the previous identities from Whisper, and injects the key as shh identity.
```
adambabik, line 331: Haven't seen the code fragment that purges all previous identities from Whisper.

lib/library_test_utils.go
```diff
@@ -717,6 +723,53 @@ func testAccountSelect(t *testing.T) bool { //nolint: gocyclo
 	return true
 }
 
+func testLoginWithKeycard(t *testing.T) bool { //nolint: gocyclo
```
adambabik, line 726: I recommend putting this unit test in a file that contains tests for `StatusBackend`. `library.go` just calls `StatusBackend` method so here we should only test data serialization.

adambabik (changes requested):
Code looks nice and clean! I would just move it around per comments.

account/accounts.go
gravityblast: @adambabik yeah good idea!

api/backend.go
gravityblast: @adambabik  we also use `github.com/status-im/status-go/services/shhext/chat/crypto` so we need an alias

api/backend.go
gravityblast: @adambabik this part is the same we have in `SelectAccount`, but there we call `accountManager.SelectAccount` to set both wallet and chat account.
Would you prefer having a new method with this part so that we can call it from both `SelectAccount` and `SetChatAccount`?

api/backend.go
gravityblast: @adambabik  it initializes the `shhext`  service, and it's already called from `SelectAccount`, so every time a user logs in

lib/library.go
gravityblast: @adambabik I used some part of the comment from the `Login` function to show it does the same. Basically we call `whisperService.SelectKeyPair`, which reset the whisper key store and inject the new key.

lib/library_test_utils.go
gravityblast: @adambabik I added `TestBackendInjectChatAccount` in the backend tests. I also left this one in lib so that all the functions are tested in the same way, since we do the same for `Login`

gravityblast:
@adambabik I updated the code after your review and replied to some comments

gravityblast:
@adambabik I just fix a test that was failing on CI, would you like me to make any other change?

account/accounts.go
```diff
@@ -245,6 +248,27 @@ func (m *Manager) SelectAccount(walletAddress, chatAddress, password string) err
 	return nil
 }
 
+// SetChatAccount initializes selectedChatAccount with privKey
+func (m *Manager) SetChatAccount(privKey *ecdsa.PrivateKey) error {
```
adambabik: Returning `error` seems obsolete because it is never set.

api/backend.go
```diff
@@ -485,6 +486,52 @@ func (b *StatusBackend) SendDataNotification(dataPayloadJSON string, tokens ...s
 	return err
 }
 
+func (b *StatusBackend) InjectChatAccount(chatKeyHex, encryptionKeyHex string) error {
```
adambabik, line 490: Missing comment :) Also wondering if we should keep `Inject` here or change to `Set` 🤔

api/backend.go
adambabik: If it's the same code, let's extract it to a separate method 👍

lib/library_test_utils.go
adambabik: Hm. Ok, being consistent is actually better. We should refactor it at some point because when we introduce gomobile, we will need to have all logic in `StatusBackend` and tested there as well.

adambabik (approved):

account/accounts.go
gravityblast: @adambabik you are right, it was used before but with the new version it's not even used, I'm removing it

PR #1360, opened 2019-01-22
Require BackupDisabledDataDir only if PFSEnabled is true

dshulyak:


cammellos (approved):

adambabik (approved):
LGTM!

We can clean up `config_test.go` even further because many tests in there have `"BackupDisabledDataDir": "/some/dir"` while not dealing with PFS at all. It would be nice to remove these settings.

pedropombeiro (approved):

mandrigin (approved):

PR #1361, opened 2019-01-24
Lock packages to versions or revisions

pedropombeiro:
https://www.pivotaltracker.com/story/show/163451198

This PR attempts to lock dependencies which might float with time, so that in the future we have a good change to be able to build an old commit. Feel free to check the `Checks (2)` tab on this PR.

Important changes:
- Created releases where possible.
- We should update our repos (such as `migrate`, `doubleratchet`, etc.) to also depend as much as possible on tags (outside the scope of this PR).
- I've added the bot to our other Go repos to help our with the task.

Here'e a comparison of the before (left) and after (right) of `dep status`:
image

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -8,9 +8,9 @@
   pruneopts = "NUT"
   revision = "024ca6a2c5444c93980f558f91c35a2defebd362"
   source = "github.com/status-im/go-fcm"
+  version = "v1.0.0-status"
 
 [[projects]]
-  branch = "master"
   digest = "1:4cf11742b199ab3aacf26b00187e72f7ff8ba2145f363b74f295d54f098f79e4"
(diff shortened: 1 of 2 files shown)


mandrigin (approved):
oh, good job!

adambabik (approved):
LGTM

`Please lock the dependency with a tag/release.` -- a tag can also be force pushed. If anything like that will happen, `go mod` should complain or at least we will see `Gopkg.lock` being changed while having no changes in `Gopkg.toml`.

pedropombeiro:
>  -- a tag can also be force pushed. 

True, but that will require a more explicit action from the user, I believe it's not as likely as overwriting a commit (due to e.g. rebasing), so hopefully that's a safer alternative.

PR #1362, opened 2019-01-24
add shhext.InitProtocolWithEncryptionKey

gravityblast:
This PR splits the `shhext.InitProtocol` function in 2 different functions (`InitProtocolWithPassword `, `InitProtocolWithEncyptionKey `) to be able to use a password or an encryption key. 

When we log in with a Keycard, instead of using a password we use a specific encryption key generated by the card using the path defined in [EIP1581](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1581.md)

### DIFF
--- a/services/shhext/service_test.go
+++ b/services/shhext/service_test.go
@@ -22,6 +22,7 @@ import (
 	"github.com/stretchr/testify/suite"
 	"github.com/syndtr/goleveldb/leveldb"
 	"github.com/syndtr/goleveldb/leveldb/storage"
+	"golang.org/x/crypto/sha3"
 )
 
 const (
@@ -126,7 +127,14 @@ func (s *ShhExtSuite) SetupTest() {
 }
 
 func (s *ShhExtSuite) TestInitProtocol() {
-	err := s.services[0].InitProtocol("example-address", "`090///\nhtaa\rhta9x8923)$$'23")
+	pass := "`090///\nhtaa\rhta9x8923)$$'23"
+
+	err := s.services[0].InitProtocolWithPassword("example-address", pass)
+	s.NoError(err)
+
+	digest := sha3.Sum256([]byte(pass))
+	encKey := fmt.Sprintf("%x", digest)
+	err = s.services[0].InitProtocolWithEncyptionKey("example-address", encKey)
 	s.NoError(err)
 }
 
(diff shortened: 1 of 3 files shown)


corpetty:
If we go this route with making it possible to access Status without a password, then we need to be sure we make it common knowledge to not port your status seed phrase to other platforms if you use a keycard.  The further you expose the seed phrase the more vuln things become that won't be protected by a "something you know"

gravityblast:
hey @corpetty, thank you for the comment!
In this case we encrypt the database with an encryption key downloaded from the card instead of using the user password. and the user password is basically the PIN used to "unlock" the card. 
Do you see any problem doing that?

> to not port your status seed phrase to other platforms if you use a keycard.

why do you think so? maybe I'm missing something. In status we use the password to encrypt the database but we don't use it anymore to add entropy to the generation of the key.

corpetty:
The methodology you describe is safer than a password (hashed) encrypting everything, but if the seed phrase leaked through something outside status would bypass all of this.  This means we need to be sure to encourage users to not try and port their account elsewhere so our security guarantees aren't subverted by someone else's poor security practices.

gravityblast:
@corpetty ah ok so you mean importing an account to another client, and this client can potentially derive the encryption key and use it to decrypt the status database!
I'll start a conversation in #status-keycard.

thank you @corpetty !

cammellos (approved):
looks ok to me, maybe some goers could also have a look @adambabik @dshulyak ?

gravityblast:
good idea, thanks!

services/shhext/service_test.go
```diff
@@ -126,7 +127,12 @@ func (s *ShhExtSuite) SetupTest() {
 }
 
 func (s *ShhExtSuite) TestInitProtocol() {
-	err := s.services[0].InitProtocol("example-address", "`090///\nhtaa\rhta9x8923)$$'23")
+	err := s.services[0].InitProtocolWithPassword("example-address", "`090///\nhtaa\rhta9x8923)$$'23")
+	s.NoError(err)
+
+	digest := sha3.Sum256([]byte("`090///\nhtaa\rhta9x8923)$$'23"))
```
dshulyak, line 133: would be nice if those symbols were a variable. it is even hard to tell if they are the same

dshulyak (approved):

services/shhext/service_test.go
gravityblast: @dmitryn it was a variable but the linter complained saying "possible credentials in code" :-S

but in general they don't need to have the same value, it's just an example

PR #1363, opened 2019-01-25
fix deadlock on starting discovery with custom AdvertiseAddr

adambabik:
In `discoverNode()` method was called a public `Server()` method which acquires a lock. `discoverNode()` is also called from a method that acquires the same lock resulting in a deadlock.

After we merge this PR we need to apply it to `release/0.19` and `release/0.20` and deploy on servers.

This bug does not affect status-go on mobile and desktop.

cc @jakubgs

### DIFF
--- a/node/status_node.go
+++ b/node/status_node.go
@@ -215,7 +215,8 @@ func (n *StatusNode) discoverNode() (*enode.Node, error) {
 		return nil, nil
 	}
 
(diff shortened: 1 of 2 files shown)


mandrigin (approved):
looks good, thanks!

dshulyak (approved):

pedropombeiro (approved):

PR #1364, opened 2019-01-30
Use pagination in RequestMessagesSync

dshulyak:
This change enables pagination when RequestMessagesSync is used. It will make history requests with cursor if Limit parameter is passed by a caller.

By making single request smaller in size this change should prevent receiving same data if previous request timed out. I missed this solution before and was looking into much more complex one, that was tracking last timestamp per topic as soon as it was coming. But that solution wasn't compatible with current pagination. And secondly wasn't using bloom filter for making a request.

Caller will be able to track history updates because we are providing incremental updates (same signal will be sent on every limited request). So in case of failure request won't be restarted from scratch, caller will adjust timestamp based on received updates.


dshulyak:
Same stuff is already implemented in status-react. I will close this one for now

PR #1365, opened 2019-01-31
Resend envelope if we sent it but didn't get an ack

dshulyak:
TBD

### DIFF
--- a/services/shhext/envelopes.go
+++ b/services/shhext/envelopes.go
@@ -1,0 +1,204 @@
+package shhext
+
+import (
+	"sync"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/p2p/enode"
+	"github.com/status-im/status-go/params"
+	"github.com/status-im/status-go/services/shhext/mailservers"
+	whisper "github.com/status-im/whisper/whisperv6"
+)
+
+type deliveryState struct {
+	state    EnvelopeState
+	expired  int
+	envelope *whisper.Envelope
+}
+
+// EnvelopesMonitor monitors state of the envelopes delivery.
+type EnvelopesMonitor struct {
+	config params.MessageResendConfig
+
+	w                      *whisper.Whisper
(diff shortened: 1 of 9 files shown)


services/shhext/envelopes.go
```diff
@@ -0,0 +1,204 @@
...
+		whisper.EventBatchAcknowledged: mon.handleAcknowledgedBatch,
+	}
+
+	if handler, ok := handlers[event.Event]; ok {
+		handler(event)
+	}
+}
+
+func (mon *EnvelopesMonitor) handleEventEnvelopeSent(event whisper.EnvelopeEvent) {
+	if mon.mailServerConfirmation {
```
mandrigin, line 102: what if this is `false`?

services/shhext/envelopes.go
```diff
@@ -0,0 +1,204 @@
...
+
+func (mon *EnvelopesMonitor) checkResendEnvelope(hash common.Hash) {
+	mon.mu.Lock()
+	defer mon.mu.Unlock()
+	state, ok := mon.cache[hash]
+	if !ok || state.state == EnvelopeSent {
+		return
+	}
+	state.expired++
+	if state.expired == mon.config.MaxRetries {
```
mandrigin, line 145: `>=` here

services/shhext/envelopes.go
```diff
@@ -0,0 +1,204 @@
...
+		log.Debug("maximum number of retries reached", "hash", hash)
```
mandrigin, line 146: shouldn't we just `return` here?

mandrigin (changes requested):
in general, looks good. I have a few questions on logic in a few places.

services/shhext/envelopes.go
dshulyak: this code was moved due to refactoring. but in general this only ensures that we accept this event only when envelope was sent to mail server

dshulyak:
this needs to be implemented differently. we need to wrap original payload and create envelope every time we do a retry. when envelope is sealed the deadline is sealed as well, so we need to adjust it every time. i will create a separate issue

PR #1367, opened 2019-02-05
Add GetContactCode call, add DH flag

cammellos:
This PR does a few things:

1) Add a call GetContactCode to check whether we have a bundle for a
given user.

2) Add a DH flag to the API (non-breaking change), for those messages
that we want to target all devices (contact-requests for example).

3) Fixes a few small issues with installations, namely if for example a
messages is sent without a bundle (currently not done by any client),
we still infer installation info, so that we can communicate securely
and making it truly optional.

Please do not merge, I would like to make sure it works as expected before merging to develop (tested only locally).

### DIFF
--- a/api/backend.go
+++ b/api/backend.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"context"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"math/big"
@@ -551,6 +552,35 @@ func (b *StatusBackend) CreateContactCode() (string, error) {
 	bundle, err := st.GetBundle(selectedChatAccount.AccountKey.PrivateKey)
 	if err != nil {
 		return "", err
+	}
+
+	return bundle.ToBase64()
+}
+
+// GetContactCode return the latest contact code
+func (b *StatusBackend) GetContactCode(identity string) (string, error) {
+	st, err := b.statusNode.ShhExtService()
+	if err != nil {
+		return "", err
+	}
+
+	publicKeyBytes, err := hex.DecodeString(identity)
+	if err != nil {
+		return "", err
+	}
+
+	publicKey, err := ethcrypto.UnmarshalPubkey(publicKeyBytes)
+	if err != nil {
+		return "", err
+	}
+
+	bundle, err := st.GetPublicBundle(publicKey)
+	if err != nil {
+		return "", err
+	}
+
+	if bundle == nil {
+		return "", nil
 	}
 
 	return bundle.ToBase64()
--- a/lib/library.go
+++ b/lib/library.go
@@ -70,6 +70,24 @@ func ProcessContactCode(bundleString *C.char) *C.char {
 	return nil
 }
 
+// Get an X3DH bundle
+//export GetContactCode
+func GetContactCode(identityString *C.char) *C.char {
+	bundle, err := statusBackend.GetContactCode(C.GoString(identityString))
+	if err != nil {
+		return makeJSONResponse(err)
+	}
+
+	data, err := json.Marshal(struct {
+		ContactCode string `json:"code"`
+	}{ContactCode: bundle})
+	if err != nil {
+		return makeJSONResponse(err)
+	}
+
+	return C.CString(string(data))
+}
+
 //export ExtractIdentityFromContactCode
 func ExtractIdentityFromContactCode(bundleString *C.char) *C.char {
 	bundle := C.GoString(bundleString)
--- a/services/shhext/api.go
+++ b/services/shhext/api.go
@@ -422,7 +422,7 @@ func (api *PublicAPI) SendPublicMessage(ctx context.Context, msg chat.SendPublic
 }
 
 // SendDirectMessage sends a 1:1 chat message to the underlying transport
-func (api *PublicAPI) SendDirectMessage(ctx context.Context, msg chat.SendDirectMessageRPC) ([]hexutil.Bytes, error) {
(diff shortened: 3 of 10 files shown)


ghost:
### Pull Request Checklist

api/backend.go
```diff
@@ -7,6 +7,7 @@ import (
 	"math/big"
 	"sync"
 
+	"encoding/hex"
```
adambabik: Ideally it should be put in a group above where all built-in libraries are listed.

lib/library.go
```diff
@@ -70,6 +70,24 @@ func ProcessContactCode(bundleString *C.char) *C.char {
 	return nil
 }
 
+// Get an X3DH bundle
+//export GetContactCode
+func GetContactCode(identityString *C.char) *C.char {
```
adambabik, line 75: From now on, it also needs to be added to `mobile/`.

services/shhext/api.go
```diff
@@ -422,7 +422,7 @@ func (api *PublicAPI) SendPublicMessage(ctx context.Context, msg chat.SendPublic
 }
 
 // SendDirectMessage sends a 1:1 chat message to the underlying transport
-func (api *PublicAPI) SendDirectMessage(ctx context.Context, msg chat.SendDirectMessageRPC) ([]hexutil.Bytes, error) {
+func (api *PublicAPI) SendDirectMessage(ctx context.Context, msg chat.SendDirectMessageRPC) (hexutil.Bytes, error) {
```
adambabik, line 425: Is this change backward compatible?

services/shhext/chat/protocol.go
```diff
@@ -62,38 +64,27 @@ func (p *ProtocolService) BuildPublicMessage(myIdentityKey *ecdsa.PrivateKey, pa
...
-
-		payload, err := p.addBundleAndMarshal(myIdentityKey, protocolMessage, true)
-		if err != nil {
-			return nil, err
-		}
+func (p *ProtocolService) BuildDirectMessage(myIdentityKey *ecdsa.PrivateKey, publicKey *ecdsa.PublicKey, payload []byte) ([]byte, error) {
+	// Encrypt payload
+	encryptionResponse, err := p.encryption.EncryptPayload(publicKey, myIdentityKey, payload)
+	if err != nil {
+		p.log.Error("encryption-service", "error encrypting payload", err)
```
adambabik, line 71: `"encryption-service"` is already hinted during initialization: `log.New("package", "status-go/services/sshext.chat")`.

```suggestion
		p.log.Error("error encrypting payload when building direct message", "err", err)
```

services/shhext/api.go
cammellos: Yes, status-react is already compatible with both versions

lib/library.go
cammellos: done!

cammellos:
@adambabik addressed the feedback, would you mind having a look when you have time?
Thanks!

adambabik (approved):

services/shhext/api.go
```diff
@@ -574,10 +520,13 @@ func (api *PublicAPI) processPFSMessage(msg *whisper.Message) error {
 		handler := EnvelopeSignalHandler{}
 		handler.DecryptMessageFailed(keyString)
 		return nil
+	} else if err == chat.ErrNotProtocolMessage {
```
pedropombeiro: Maybe this whole comparison block should be converted to a `switch {}` block?

services/shhext/chat/encryption.go
```diff
@@ -450,46 +455,60 @@ func (s *EncryptionService) EncryptPayloadWithDH(theirIdentityKey *ecdsa.PublicK
...
-	if theirBundle == nil && !bytes.Equal(theirIdentityKeyC, ecrypto.CompressPubkey(&myIdentityKey.PublicKey)) {
+	if installationIds == nil && !bytes.Equal(theirIdentityKeyC, ecrypto.CompressPubkey(&myIdentityKey.PublicKey)) {
 		return s.EncryptPayloadWithDH(theirIdentityKey, payload)
 	}
 
 	response := make(map[string]*DirectMessageProtocol)
 
-	for installationID, signedPreKeyContainer := range theirBundle.GetSignedPreKeys() {
+	for _, installationID := range installationIds {
+		s.log.Debug("Processing installationID", "installationID", installationID)
```
pedropombeiro:
```suggestion
		s.log.Debug("Processing installation", "installationID", installationID)
```

services/shhext/chat/protocol.go
```diff
@@ -62,38 +64,27 @@ func (p *ProtocolService) BuildPublicMessage(myIdentityKey *ecdsa.PrivateKey, pa
...
+		DirectMessage:  encryptionResponse,
 	}
-	return response, nil
+
+	return p.addBundleAndMarshal(myIdentityKey, protocolMessage, true)
 }
 
-// BuildPairingMessage sends a message to our own devices using DH so that it can be decrypted by any other device.
-func (p *ProtocolService) BuildPairingMessage(myIdentityKey *ecdsa.PrivateKey, payload []byte) ([]byte, error) {
+// BuildDHMessage build a message with DH encryption so that it can be decrypted by any other device.
```
pedropombeiro:
```suggestion
// BuildDHMessage builds a message with DH encryption so that it can be decrypted by any other device.
```

pedropombeiro (approved):

services/shhext/api.go
cammellos: done!

PR #1368, opened 2019-02-06
Add a docker-compose.yml and Makefile for running a mailserver

jakubgs:
This should simplify running a mailserver for a less tech-savvy user. All you need to do is run:
```
cd _assets/compose/mailserver && make
```
In the repo directory.

Also adds a [README](https://github.com/status-im/status-go/tree/add-mailserver-compose/_assets/compose/mailserver).

### DIFF
--- a/_assets/compose/mailserver/Makefile
+++ b/_assets/compose/mailserver/Makefile
@@ -1,0 +1,72 @@
+GIT_ROOT = $(shell git rev-parse --show-toplevel)
+
+RED    := $(shell tput -Txterm setaf 1)
+GREEN  := $(shell tput -Txterm setaf 2)
+YELLOW := $(shell tput -Txterm setaf 3)
+RESET  := $(shell tput -Txterm sgr0)
+BOLD   := $(shell tput bold)
+
+# Settings
+export LOG_LEVEL = INFO
+export LISTEN_PORT ?= 30303
+export RPC_PORT ?= 8545
+export API_MODULES = eth,net,web3,admin
+export CONTAINER_NAME ?= status-go-mailserver
+export DATA_PATH ?= /var/tmp/status-go-mail
+export FLEET_NAME = eth.beta
+# Necessary to make mailserver available publicly
+export PUBLIC_IP ?= $(shell curl -s https://ipecho.net/plain)
+
+all: checks start show info enode
+
+checks:
+ifeq (, $(shell which docker-compose))
+	$(error No $(BOLD)docker-compose$(RESET) in your $$PATH. Please install it)
+endif
+ifeq (, $(shell which jq))
+	$(error No $(BOLD)jq$(RESET) in your $$PATH. Please install it)
+endif
+ifndef PUBLIC_IP
+	$(error PUBLIC_IP not set! Export it as environment variable)
+endif
+ifndef CONTAINER_NAME
+	$(error CONTAINER_NAME not set! Export it as environment variable)
+endif
+
+start: config
+	docker-compose up -d
+
+stop:
+	docker-compose down
+
+logs:
+	docker-compose logs -f -t --tail=100
+
+enode:
+	@curl -s -XPOST http://localhost:$(RPC_PORT)/ \
+		-H 'Content-type: application/json' \
+		-d '{"jsonrpc":"2.0","method":"admin_nodeInfo","params":[],"id":1}' \
+		| jq -r '.result.enode' \
+		| grep -oP '\Kenode://[^?]+'
+
+
+info:
+	@echo "$(GREEN)Your mailserver is listening on:$(RESET) $(BOLD)$(PUBLIC_IP):$(LISTEN_PORT)$(RESET)"
+	@echo "$(YELLOW)Make sure that address and port are available from the internet!$(RESET)"
+	@echo "$(GREEN)Your enode address is:$(RESET)"
+
+show:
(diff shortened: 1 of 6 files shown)


mandrigin (approved):
awesome!

cammellos (approved):

corpetty:
if attempting to run from a github clone (assuming you don't use go get properly, you get the following errors:

```
Current dir is /home/petty/GitHub/status/status-go, which seems to be different from your GOPATH.
Please, build status-go from GOPATH for proper build.
  GOPATH       = /home/petty/go
  Current dir  = /home/petty/GitHub/status/status-go
  Expected dir = /home/petty/go/src/github.com/status-im/status-go)
See https://golang.org/doc/code.html#GOPATH for more info
.  Stop.
```

gravityblast (approved):

jakubgs:
@corpetty ah yes, that's a Go thing. There is a check in that [Makefile at line 18](https://github.com/status-im/status-go/blob/develop/Makefile#L18):
```Makefile
EXPECTED_PATH=$(shell go env GOPATH)/src/github.com/status-im/status-go
ifneq ($(CURDIR),$(EXPECTED_PATH))
define NOT_IN_GOPATH_ERROR

Current dir is $(CURDIR), which seems to be different from your GOPATH.
Please, build status-go from GOPATH for proper build.
  GOPATH       = $(shell go env GOPATH)
  Current dir  = $(CURDIR)
  Expected dir = $(EXPECTED_PATH))
See https://golang.org/doc/code.html#GOPATH for more info

endef
$(error $(NOT_IN_GOPATH_ERROR))
endif
```
Which means every time - even if the target doesn't require it - this check is run and fails if repo was cloned to the wrong dir. Maybe it would be better for users if the instructions didn't say to run `make run-mailserver` but rather to do `cd _assets/compose/mailserver` and then `make`, to avoid this.

gravityblast:
> Maybe it would be better for users if the instructions didn't say to run `make run-mailserver` but rather to do `cd _assets/compose/mailserver` and then `make`, to avoid this.

yeah I think that would be easier for the user and more clear for the main project Makefile

corpetty:
I also seem to be getting an error with the `statusd` command that is run from the compose file.  `docker logs` only gives `statusd` usage prompts, not a running mailserver.

jakubgs:
Okay, changed the `config.json` target that generates it to just `config` so it's re-generated every time.
I also added checks for availability of `jq` and `docker-compose`:
```Makefile
checks:
ifeq (, $(shell which docker-compose))
	$(error No $(BOLD)docker-compoase$(RESET) in your $$PATH. Please install it)
endif
ifeq (, $(shell which jq))
	$(error No $(BOLD)jq$(RESET) in your $$PATH. Please install it)
endif
```

_assets/compose/mailserver/Makefile
```diff
@@ -0,0 +1,65 @@
...
+FLEETS = $(shell curl -s "$(FLEETS_URL)")
+BOOT_NODES = $(shell echo '$(FLEETS)' | jq '.fleets["$(FLEET_NAME)"].boot | map(.)')
+
+all: checks start show info enode
+
+checks:
+ifeq (, $(shell which docker-compose))
+	$(error No $(BOLD)docker-compoase$(RESET) in your $$PATH. Please install it)
+endif
+ifeq (, $(shell which jq))
```
adambabik, line 26: Isn't it going to return an error when running `make checks` because of `jq` used in `BOOT_NODES`?

_assets/compose/mailserver/Makefile
```diff
@@ -0,0 +1,65 @@
...
+	docker-compose up -d
+
+stop:
+	docker-compose down
+
+logs:
+	docker-compose logs -f -t --tail=100
+
+enode:
+	@docker-compose logs --tail=all | grep -oP '.*self="\Kenode://[^?]+' | sed 's/127.0.0.1/$(PUBLIC_IP)/'
```
adambabik: Another option might be to use IPC file and `admin.nodeInfo` method but it assumes IPC is created.

_assets/compose/mailserver/Makefile
```diff
@@ -0,0 +1,65 @@
...
+FLEETS_URL := https://fleets.status.im/
+FLEET_NAME := eth.beta
+FLEETS = $(shell curl -s "$(FLEETS_URL)")
+BOOT_NODES = $(shell echo '$(FLEETS)' | jq '.fleets["$(FLEET_NAME)"].boot | map(.)')
+
+all: checks start show info enode
+
+checks:
+ifeq (, $(shell which docker-compose))
+	$(error No $(BOLD)docker-compoase$(RESET) in your $$PATH. Please install it)
```
pedropombeiro:
```suggestion
	$(error No $(BOLD)docker-compose$(RESET) in your $$PATH. Please install it)
```

pedropombeiro (approved):

_assets/compose/mailserver/Makefile
jakubgs: You're right, I forgot to remove that part since we do not update config anyway.

_assets/compose/mailserver/Makefile
jakubgs: I thought this was simpler, but I guess I can try doing it with RPC.

jakubgs: Done.

_assets/compose/mailserver/Makefile
jakubgs: Fixed.

adambabik (approved):

PR #1370, opened 2019-02-07
Add build status notifications like in status-react

jakubgs:
Since I've created PR builds for separate platforms in Jenkins like for `status-react` in:
https://ci.status.im/job/status-go/job/prs/
I'm also adding notifications like in `status-react` using [`ghcmgr`](https://github.com/status-im/github-comment-manager).
I've implemented multi-repo handling here: https://github.com/status-im/github-comment-manager/commit/303079ea7d2dc68c3849299db366a3b39d38396b


adambabik:
LGTM 👍 

I only have a few questions about Jenkins integration. I wonder if it's needed to build each commit in Jenkins or instead trigger it with a GH comment. Another thing is that in this PR, for example, Jenkins builds are marked as failed but each step is green... Do you have any idea why it is so?

adambabik (approved):

jakubgs:
Ah yeah, I forgot to fix this since I was otherwise occupied. I forgot to use `${CI_DIR}` when accessing `ghcmgr.groovy`, the path was wrong.

Btw, glad to have you back Adam!

jakubgs:
Should be good and proper now.

adambabik:
Awesome!

mandrigin (approved):
I guess we need to make sure that we have all tests running as intended there and then get rid of Travis

adambabik:
The only difference I see is that Travis runs `make test-e2e-race` while Jenkins `test-e2e`.

jakubgs:
>Travis runs `make test-e2e-race` while Jenkins `test-e2e`

Should Jenkins run both? What's the difference?

adambabik:
@jakubgs no, just one of those. `test-e2e-race` runs tests with `-race` flag which slows them down but conduct checks race conditions as well.  `test-e2e` simply runs tests but without checking race conditions.

jakubgs:
Well, that added about ~13 minutes to each build. @adambabik okay with you?

adambabik:
That's super long :( 27 minutes for a build is IMO too long. Using `test-e2e` it is about 14 mins?

jakubgs:
Yeah, I'll revert it.

adambabik:
Thanks! We can think about running tests with checking race conditions daily and somehow notify us if they fail.

PR #1371, opened 2019-02-08
Change publishing envelopes behaviour

cammellos:
Before `processedEnvelopes` was always set to `0`, which lead to the behavior that envelopes would only be published when the size of the current bundle was too big, or when the count of the envelopes in a given bundle hit `limit`, and at the end of the loop.
 
Effectively this meant that for small `limit` it all worked ok, for large limits (i.e. 2000), we would publish all the envelopes in the range (say 0 -> 5000) in one go, and for medium size limit (`400`), the behavior was dependent on the size of the envelopes. In any case it correctly sent out all the envelopes in the original range, just not in the right pagination.

### DIFF
--- a/mailserver/mailserver.go
+++ b/mailserver/mailserver.go
@@ -294,9 +294,6 @@ func (s *WMailServer) DeliverMail(peer *whisper.Peer, request *whisper.Envelope)
 	)
 	requestProcessTimer.UpdateSince(start)
 
-	// bundles channel can be closed now because the processing finished.
-	close(bundles)
-
 	// Wait for the goroutine to finish the work. It may return an error.
 	if err := <-errCh; err != nil {
 		processRequestErrorsCounter.Inc(1)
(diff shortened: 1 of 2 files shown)


mailserver/mailserver.go
```diff
@@ -468,50 +474,63 @@ func (s *WMailServer) processRequestInBundles(
...
+	// Publish
+	for _, batch := range batches {
+		output <- batch
 	}
 
 	sentEnvelopesMeter.Mark(int64(processedEnvelopes))
 	sentEnvelopesSizeMeter.Mark(processedEnvelopesSize)
 
+	close(output)
+
```
mandrigin, line 544: I would add one more log entry that everything was sent successfully.

mailserver/mailserver.go
```diff
@@ -444,17 +438,29 @@ func (s *WMailServer) createIterator(lower, upper uint32, cursor []byte) iterato
...
 		bundle                 []*whisper.Envelope
 		bundleSize             uint32
+		batches                [][]*whisper.Envelope
 		processedEnvelopes     int
 		processedEnvelopesSize int64
 		nextCursor             []byte
 		lastEnvelopeHash       common.Hash
 	)
 
+	log.Debug("[mailserver:processRequestInBundles] processing request", "limit", limit)
```
mandrigin: to improve logging filtering, I would also add a `requestID` of a kind, for instance `requestID := fmt.Sprintf("%d-%d", time.Time.Unix(), rand.Intn(1000))` and add it to the logs

mandrigin (approved):

cammellos:
@mandrigin done

mandrigin (approved):

gravityblast (approved):

PR #1372, opened 2019-02-11
add LoginWithKeycard to mobile pkg

gravityblast:
This function is already present in `lib/library.go`, and now in the `mobile` pkg as well.


mandrigin (approved):

mandrigin:
I will build a release out of this branch to be able to re-launch the CI on status-react.

mandrigin:
merging because the change is trivial and was just forgotten

VERSION
```diff
@@ -1 +1 @@
-0.21.0g-beta.1
+0.21.0g-beta.2
```
adambabik, line 1: Any idea what does a letter `g` do in this version?

gravityblast: @adambabik I'm not sure tbh, I just bumped the version

gravityblast: maybe @mandrigin knows?

PR #1374, opened 2019-02-12
send tx with signature

gravityblast:
In case of a keycard account, the transaction is signed by the card and the signature is sent to `status-go` together with the original transaction.
The original transaction is needed to serialize the final transaction with signature.

Important changes:
- [x] add `SendTransactionWithSignature` to transactor

### DIFF
--- a/transactions/transactor.go
+++ b/transactions/transactor.go
@@ -69,6 +69,8 @@ func (t *Transactor) SendTransaction(sendArgs SendTxArgs, verifiedAccount *accou
 	return
 }
 
+// SendTransactionWithSignature receive a transaction and a signature, serialize them together and propage it to the network.
+// It's different from eth_sendRawTransaction because it receives a signature and not a serialized transaction with signature.
 func (t *Transactor) SendTransactionWithSignature(args SendTxArgs, sig []byte) (hash gethcommon.Hash, err error) {
 	chainID := big.NewInt(int64(t.networkID))
 	signer := types.NewEIP155Signer(chainID)
@@ -113,6 +115,7 @@ func (t *Transactor) SendTransactionWithSignature(args SendTxArgs, sig []byte) (
 	if err := t.sender.SendTransaction(ctx, signedTx); err != nil {
 		return hash, err
 	}
+
 	return signedTx.Hash(), nil
 }
 
--- a/transactions/transactor.go
+++ b/transactions/transactor.go
@@ -73,6 +73,10 @@ func (t *Transactor) SendTransaction(sendArgs SendTxArgs, verifiedAccount *accou
 // It's different from eth_sendRawTransaction because it receives a signature and not a serialized transaction with signature.
 // Since the transactions is already signed, we assume it was validated and used the right nonce.
 func (t *Transactor) SendTransactionWithSignature(args SendTxArgs, sig []byte) (hash gethcommon.Hash, err error) {
+	if !args.Valid() {
+		return hash, ErrInvalidSendTxArgs
+	}
+
 	chainID := big.NewInt(int64(t.networkID))
 	signer := types.NewEIP155Signer(chainID)
 
--- a/transactions/transactor.go
+++ b/transactions/transactor.go
@@ -69,6 +69,53 @@ func (t *Transactor) SendTransaction(sendArgs SendTxArgs, verifiedAccount *accou
 	return
 }
 
+func (t *Transactor) SendTransactionWithSignature(args SendTxArgs, sig []byte) (hash gethcommon.Hash, err error) {
+	chainID := big.NewInt(int64(t.networkID))
+	signer := types.NewEIP155Signer(chainID)
+
+	nonce := uint64(*args.Nonce)
+	to := *args.To
+	value := (*big.Int)(args.Value)
+	gas := uint64(*args.Gas)
+	gasPrice := (*big.Int)(args.GasPrice)
+	data := args.GetInput()
+
+	var tx *types.Transaction
+	if args.To != nil {
+		t.log.Info("New transaction",
+			"From", args.From,
+			"To", *args.To,
+			"Gas", gas,
+			"GasPrice", gasPrice,
+			"Value", value,
+		)
+		tx = types.NewTransaction(nonce, to, value, gas, gasPrice, data)
+	} else {
+		// contract creation is rare enough to log an expected address
+		t.log.Info("New contract",
+			"From", args.From,
+			"Gas", gas,
+			"GasPrice", gasPrice,
+			"Value", value,
+			"Contract address", crypto.CreateAddress(args.From, nonce),
+		)
+		tx = types.NewContractCreation(nonce, value, gas, gasPrice, data)
+	}
+
+	signedTx, err := tx.WithSignature(signer, sig)
+	if err != nil {
+		return hash, err
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), t.rpcCallTimeout)
+	defer cancel()
+
+	if err := t.sender.SendTransaction(ctx, signedTx); err != nil {
+		return hash, err
+	}
+	return signedTx.Hash(), nil
+}
+
 // make sure that only account which created the tx can complete it
 func (t *Transactor) validateAccount(args SendTxArgs, selectedAccount *account.SelectedExtKey) error {
 	if selectedAccount == nil {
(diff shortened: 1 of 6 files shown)


transactions/transactor.go
```diff
@@ -69,6 +69,97 @@ func (t *Transactor) SendTransaction(sendArgs SendTxArgs, verifiedAccount *accou
 	return
 }
 
+// SendTransactionWithSignature receive a transaction and a signature, serialize them together and propage it to the network.
+// It's different from eth_sendRawTransaction because it receives a signature and not a serialized transaction with signature.
+// Since the transactions is already signed, we assume it was validated and used the right nonce.
+func (t *Transactor) SendTransactionWithSignature(args SendTxArgs, sig []byte) (hash gethcommon.Hash, err error) {
```
adambabik, line 86: I have an impression that this method is already defined somewhere or one very similar to it. Did you create it from scratch or it's inspired by some other fragment?

gravityblast: @adambabik I think we have SendRawTransaction/SendSignedTransaction right? but it needs the full signed transaction and not only the signature. but from the keycard we receive only a signature and a tx not serialized

adambabik: Right, I am just thinking if we can generalize it or put some fragments into a separate function/method. Especially, the code around getting params and deciding whether we create a tx or contract. Also, getting a final nonce is tricky and it would be nice to encapsulate this logic.

gravityblast: @adambabik it's very similar to SendTransaction but:
- it doesn't set the nonce in case the remote nonce provider returns a higher one
- it doesn't sign the transaction because we already have the signature
- it doesn't alter any tx args because the signature would be invalid

gravityblast: @adambabik ok I can refactor the tx creation, but the nonce part is different so I would keep that separated for the 2 cases, what do you think?

adambabik: I see, `SendTransaction` is highly specified with a ton of tx args magic stuff. Ok, I guess there is no really a way to do something better here without refactoring `SendTransaction` which we don't want to do.

transactions/transactor_test.go
```diff
@@ -283,3 +283,52 @@ func (s *TransactorSuite) TestContractCreation() {
 	s.NoError(err)
 	s.Equal(crypto.CreateAddress(testaddr, 0), receipt.ContractAddress)
 }
+
+func (s *TransactorSuite) TestSendTransactionWithSignature_IncrementingNonce() {
```
adambabik: Any other tests planned?

gravityblast: I'm not 100% sure about the nonce logic I added. 
It can be more complicated than SendTransaction because we assument the tx nonce is correct,
so I wanted more feedback in that part. 
If that part is ok maybe I can add more tests on the nonce increment. 
Would you test something else?

adambabik: If you are sure that the correct nonce was used, I think we can skip the whole logic around nonce completely. How the nonce is retrieved when signing a transaction using a key card? Is it retrieved using the node on the smartphone or it can come from somewhere else?

gravityblast: that's not implemented yet. I'm going to work on that after this PR.
But I think it would be again similar to SendTransaction except for the last part. Instead of signing and sending I would hash and return. 
It will be again difficult to refactor SendTransaction because it has other things like the increment in the deferred function.

In this PR the problem could be that:

- status builds and hashes the transaction
- it waits for the keycard
- in the meantime another transaction is sent
- keycard signs the tx and calls SendTransactionWithSignature

In this case the tx nonce would not be the right one to increment, so I don't do increment it.
It's maybe a super edge case though.

adambabik: But I guess this transaction signed by the keycard won't be accepted anyway following this scenario, right? I guess you can't send another transaction with lower nonce (`* in the meantime another transaction is sent` will use a bumped nonce).

I see it this way:
1. One asks the app to prepare a tx giving some tx args (it does the same magic stuff as in `SendTransaction`),
2. It waits for the keycard to sign the tx,
3. The tx is pushed to the network.

This should be a blocking operation from the app perspective, I believe. A case when another transaction is introduced in the meantime would be only possible if someone did it from a different device. But then it's their fault and naturally the tx signed by the keycard would be invalid. It would need to be prepared once again with a new nonce to succeed.

gravityblast: > This should be a blocking operation from the app perspective

yeah it is from in UI, but they are still 2 different functions so I was just thinking about `status-go`.
Ok so you say that I just suppose the tx is correct, and I just increment the local nonce right?
I was just thinking what to do if tx nonce is different from local nonce or remote nonce. I can't change the tx nonce because it's already signed, should I fail and not sending it then?

adambabik: > Ok so you say that I just suppose the tx is correct, and I just increment the local nonce right?

Actually, I though about figuring out nonce before signing. I imagine one more method `PrepareTxArgs(args)` or something like that. It is called first, then the tx is signed using the keycard and `SendTransactionWithSignature` is called.

I would assume that from the UI/UX perspective it will be a single operation, however, it can always be interrupted because of various reason. So `PrepareTxArgs(args)` would already compute and set a proper nonce value in `t.localNonce` cache. If the sending is interrupted, no worries, a next nonce will be taken when a new transaction is created on the same device. There might be some gaps in nonces in that case but is that a problem? 

This will simplify `SendTransactionWithSignature` logic as it won't need to do anything with the nonce (and it can't really because the tx is already signed). 

What do you think?

gravityblast: > There might be some gaps in nonces in that case but is that a problem?

@adambabik yeah for sure the prepare transaction will have all the logic to check/set the current nonce.
But then the local nonce, not the transaction nonce,  should be incremented after the transaction is sent. so the current code is handling the increment for the local one, which is used for the next transaction. 
so we still need to increment it right?

adambabik: > But then the local nonce, not the transaction nonce, should be incremented after the transaction is sent.

I was suggesting we increment it immediately in the prepare transaction function. But not sure about it now...

Now, I think your suggestion is better so we would compute the nonce using the local nonce cache in the prepare tx function. When it is finally pushed to the network, the following scenarios might happen:
1. If the local and remote nonces are lower than the nonce from the tx, update the local nonce cache to the tx nonce (happy path scenario),
2. If the remote nonce is greater than the one from the tx, a different tx from the same address has just been sent, probably using a different device (people using the same wallet on different devices and doing something weird; the current tx we try to send will fail and the flow needs to be repeated),
3. If the local nonce is greater then the one from tx, something is off (it should never happen because it means that the app sent a different tx in the background).

Is there anything else that is possible?

gravityblast: I think that's it, those are the ones I thought about

gravityblast: the first one is is slightly different:

If local nonce and remote nonce are equal or lower, I send the tx and I set the local one to be `tx.nonce + 1`, so that the next nonce is the right one to be used for the next transaction.

adambabik: Is that what we do in `SendTransaction`? We need to stay consistent. I though that in the local nonce cache we keep the latest used nonce rather than the next available.

gravityblast: > Is that what we do in `SendTransaction`?
Yes that's what SentTransactions does, so I kept the same logic:

https://github.com/status-im/status-go/blob/cf52f2a0e09a92a768c2b791625e71907e82f895/transactions/transactor.go#L190-L196

adambabik: Interesting. But all clear now :)

gravityblast: thank you @adambabik  so shall we merge it like this?

adambabik: We can but I am still not comfortable with `if txNonce >= localNonce && txNonce >= remoteNonce {` condition. Honestly, I would return an error if `txNonce` is different than `localNonce`. I believe we don't allow overwriting transactions from the app, correct? So in theory that's the only valid scenario that `txNonce == localNonce`. We would increment it always if `t.sender.SendTransaction(ctx, signedTx)` is successful.

If `localNonce` is not set, this missing `PrepareTxArgs()` method would be responsible for initializing it.

adambabik (approved):
LGTM. I am playing super safe and still have some doubts expressed in https://github.com/status-im/status-go/pull/1374/files/#r256828714. But we can work on that later.

transactions/transactor_test.go
gravityblast: yeah it makes sense. I'll change it so that we send the tx if the nonce is the expected one.

gravityblast:
@adambabik it makes sense, I'll change that part!

gravityblast:
@adambabik I also modified to test to reflect what we said. I'll wait for CI and then merge it

PR #1375, opened 2019-02-13
Add confirm messages processed by ID

cammellos:
Currently PFS messages are decrypted and therefore modified before being
passed to the client. Which means that they are never deduped.
The irony is that they can only be decrypted once for now, as we delete the key once used, that's in the next PR :) 
This make IDs computation difficult, as we pass the whole object to the client and expect the object be passed back once
confirmed.
This changes the behavior allowing confirmation by ID, which is passed
to the client instead of the raw object.
This is a breaking change, but we can make status-react compatible
(backward and forward) before merging.


adambabik:
So this change also alters the whole architecture of deduping, right? Before, it was totally transparent from the client perspective. Now, with PFS messages, the client needs to send back IDs of processed messages. Why do we need confirmations for PFS messages and don't require them for other Whisper messages?

Also, I don't see how it works. It looks like a key is created based on `Payload` and `Topic`. In the case of PFS, `Payload` is always different so the same message will have different keys. Unless `Payload` is already decrypted PFS message... Is that the case?

cammellos:
@adambabik this does not change the architecture, as it wasn't transparent to the client, it still needed to confirm the messages (it makes an API call confirmMessagesProcessed). This changes the behavior to use IDs instead of passing back the whole message (so the id needs to be passed as well).

It works as such:

1) Stauts-go sends a bunch of whisper messages to status-react, which they get deduped by doing Hash(raw-message) 
2) Status-react process those messages and sends them back to status-go
3) Status-go computes the ID of the message Hash(raw-message) and adds it to the dedup cache

PFS messages are decrypted at the status-go layer, so Hash(raw-message) is different from Hash(decyrpted-message)

What this change is basically:
1) Status-go sends a bunch of whisper messages + a set of IDs `{message: decryptedmessage, id: Hash(raw-message)}`
2) Status-react process those messages, and passes to status-go a bunch of messageIds to status-go
3) Status-go adds the ids to the dedup cache

Essentially we compute the IDs only once before sending the message to the user, and we ask the user to send back the ID instead of the raw payload, as it is mutable in the case of PFS messages.

> In the case of PFS, `Payload` is always different so the same message will have different keys.

This is not quite true, not sure if I understand correctly , it will be different if the same content is encrypted twice (i.e a user re-sending a message), but it will be the same if you pull the message multiple times from the mailserver. Or I misunderstand what you mean?

adambabik:
Thanks for the explanation! Yeah, it makes totally sense to avoid going through the serialization once more when receiving the response from the client. 

> 1. Hash(raw-message)

You mean a raw Whisper message or raw Whisper message payload?

> This is not quite true, not sure if I understand correctly , it will be different if the same content is encrypted twice (i.e a user re-sending a message), but it will be the same if you pull the message multiple times from the mailserver. Or I misunderstand what you mean?

That's what I mean. Because I though it compares a hash of the payload which would disallow sending the same message one after another but it clearly is not the case.

cammellos:
Correct, it would be Hash(Payload of the raw whisper message) (same as it is now, that hasn't changed, which should work fine as long as the message has not been decrypted first (otherwise when deduping you would have to decrypt it again, and that might not be possible as we want to delete the key as soon as it is confirmed).

mandrigin (approved):

cammellos:
@adambabik could you have another look at this please? It would be ready to be merged otherwise. thanks!

services/shhext/chat/encryption.go
```diff
@@ -94,6 +101,33 @@ func (s *EncryptionService) getDRSession(id []byte) (dr.Session, error) {
 
 }
 
+func confirmationIDString(id []byte) string {
+	return fmt.Sprintf("%x", id)
```
adambabik: Nitpick: it's good but `hex.EncodeToString(id)` would be more idiomatic.

services/shhext/chat/encryption.go
```diff
@@ -94,6 +101,33 @@ func (s *EncryptionService) getDRSession(id []byte) (dr.Session, error) {
...
+
+// ConfirmMessagesProcessed confirms and deletes message keys for the given messages
+func (s *EncryptionService) ConfirmMessagesProcessed(messageIDs [][]byte) error {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+
+	for _, idByte := range messageIDs {
+		id := confirmationIDString(idByte)
+		confirmationData := s.messageIDs[id]
+		if confirmationData != nil {
```
adambabik: Suggestion to reduce number of indentation: 
```go
confirmationData, ok := s.messageIDs[id]
if !ok {
  continue
}
// Load session from store first
...
```

services/shhext/chat/encryption_test.go
```diff
@@ -866,3 +876,88 @@ func (s *EncryptionServiceTestSuite) TestRefreshedBundle() {
 	s.Equal(bobBundle2.GetSignedPreKeys()[bobInstallationID].GetSignedPreKey(), x3dhHeader2.GetId())
 
 }
+
+func (s *EncryptionServiceTestSuite) TestMessageConfirmation() {
```
adambabik, line 880: Do you think we could write this test as a table driven test?

services/shhext/chat/protocol.go
```diff
@@ -124,8 +124,13 @@ func (p *ProtocolService) GetPublicBundle(theirIdentityKey *ecdsa.PublicKey) (*B
 	return p.encryption.GetPublicBundle(theirIdentityKey)
 }
 
+// ConfirmMessagesProcessed confirms and deletes message keys for the given messages
+func (p *ProtocolService) ConfirmMessagesProcessed(messageIDs [][]byte) error {
```
adambabik, line 128: I can't find a reference to this method anywhere. What does call it?

adambabik (approved):
LGTM!

@cammellos do you think we should create a follow-up issue to refactor how deduplication works? I feel like it bleeds into some code parts that otherwise wouldn't know about something like deduplication.

services/shhext/chat/encryption_test.go
cammellos: I can try that, I will probably do in a separate PR if it's ok

services/shhext/chat/protocol.go
cammellos: :man_facepalming: forgot to wire things up, well spotted

PR #1378, opened 2019-02-18
Mark envelope as confirmed when it was received from a mail server

dshulyak:
Previously i didn't take into account that envelope might get delivered from a mail server faster then it is sent to it. For example, if we are connected with 2 peers, A and B. A is a mail server. B also connected with a mail server. In rare circumstances A will receive envelope from B and deliver it to us. Whisper will mark such envelope as "known" and will never try to re-deliver it to A.

This change is linked to a custom branch in our whisper repository. I will make a new whisper release once corresponding change is merged.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -46,7 +46,7 @@
 
 [[constraint]]
   name = "github.com/status-im/whisper"
-  version = "=v1.4.6"
+  branch = "env-received-event"
 
 [[constraint]]
   name = "golang.org/x/text"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -46,7 +46,7 @@
 
 [[constraint]]
   name = "github.com/status-im/whisper"
-  branch = "env-received-event"
+  version = "=v1.4.8"
 
 [[constraint]]
   name = "golang.org/x/text"
--- a/services/shhext/tracker.go
+++ b/services/shhext/tracker.go
@@ -1,7 +1,6 @@
 package shhext
 
 import (
-	"fmt"
 	"sync"
 
 	"github.com/ethereum/go-ethereum/common"
@@ -106,7 +105,6 @@ func (t *tracker) handleEvent(event whisper.EnvelopeEvent) {
 	}
 
 	if handler, ok := handlers[event.Event]; ok {
-		fmt.Println("received event")
 		handler(event)
 	}
 }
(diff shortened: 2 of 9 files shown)


Gopkg.toml
```diff
@@ -46,7 +46,7 @@
 
 [[constraint]]
   name = "github.com/status-im/whisper"
-  version = "=v1.4.6"
+  branch = "env-received-event"
```
mandrigin: just to mark that that it should be replaced by a tagged version

services/shhext/tracker.go
```diff
@@ -104,6 +106,7 @@ func (t *tracker) handleEvent(event whisper.EnvelopeEvent) {
 	}
 
 	if handler, ok := handlers[event.Event]; ok {
+		fmt.Println("received event")
```
mandrigin: looks like a leftover

mandrigin (changes requested):
2 small things, looks good otherwise

adambabik:
> For example, if we are connected with 2 peers, A and B. A is a mail server. B also connected with a mail server. In rare circumstances A will receive envelope from B and deliver it to us. Whisper will mark such envelope as "known" and will never try to re-deliver it to A.

I don't quite get this. Why does it matter whether a peer is a mail server or not? From Whisper perspective it does not matter, we will receive messages from both when being connected.

> In rare circumstances A will receive envelope from B and deliver it to us.

I wouldn't say it's rare. It's perfectly normal scenario. Do we assume anything about to how many and what kind of peers we are connected when dealing with confirmations?

dshulyak:
> I don't quite get this. Why does it matter whether a peer is a mail server or not? From Whisper perspective it does not matter, we will receive messages from both when being connected.

it matters because we want to receive a confirmation from a mail server. in normal circumstances we send an envelope to all peers, and wait for a mail server to send us a special confirmation (because it won't resend an envelope).

> I wouldn't say it's rare. It's perfectly normal scenario. 

our node is the original sender of an envelope in this example. thats why i said that this is rare. basically X -> B -> A -> X was delivered faster then X -> A. it is not that common. 
(X is the original sender)

> Do we assume anything about to how many and what kind of peers we are connected when dealing with confirmations?

we assume that we are connected with at least one mail server.

adambabik:
@dshulyak thanks for the explanation! All clear now.

mandrigin (approved):
Great!

mandrigin:
@adambabik @PombeirP @cammellos can you help reviewing it? It would be awesome to have it in a new status-go release for status-react.

mandrigin:
cc @gravityblast

pedropombeiro (approved):

PR #1379, opened 2019-02-18
add HashTransaction

gravityblast:
This PR adds a method `HashTransaction` use to validate a transaction and geneeate its hash. 
The hash will be sent to the keycard to be signed, and the signature will be sent to `SendTransactionWithSignature`.

### DIFF
--- a/transactions/transactor.go
+++ b/transactions/transactor.go
@@ -27,13 +27,12 @@ const (
 )
 
 type ErrBadNonce struct {
-	nonce       uint64
-	localNonce  uint64
-	remoteNonce uint64
+	nonce         uint64
+	expectedNonce uint64
 }
 
 func (e *ErrBadNonce) Error() string {
-	return fmt.Sprintf("bad nonce %d. local nonce: %d, remote nonce: %d", e.nonce, e.localNonce, e.remoteNonce)
+	return fmt.Sprintf("bad nonce. expected %d, got %d", e.expectedNonce, e.nonce)
 }
 
 // Transactor validates, signs transactions.
@@ -93,17 +92,7 @@ func (t *Transactor) SendTransactionWithSignature(args SendTxArgs, sig []byte) (
 	signer := types.NewEIP155Signer(chainID)
 
 	tx := t.buildTransaction(args)
-
-	var (
-		localNonce  uint64
-		remoteNonce uint64
-	)
-
 	t.addrLock.LockAddr(args.From)
-	if val, ok := t.localNonce.Load(args.From); ok {
-		localNonce = val.(uint64)
-	}
-
 	defer func() {
 		// nonce should be incremented only if tx completed without error
 		// and if no other transactions have been sent while signing the current one.
@@ -113,15 +102,13 @@ func (t *Transactor) SendTransactionWithSignature(args SendTxArgs, sig []byte) (
 		t.addrLock.UnlockAddr(args.From)
 	}()
 
-	ctx, cancel := context.WithTimeout(context.Background(), t.rpcCallTimeout)
-	defer cancel()
-	remoteNonce, err = t.pendingNonceProvider.PendingNonceAt(ctx, args.From)
+	expectedNonce, err := t.getTransactionNonce(args)
 	if err != nil {
 		return hash, err
 	}
 
-	if tx.Nonce() != localNonce || tx.Nonce() != remoteNonce {
-		return hash, &ErrBadNonce{tx.Nonce(), localNonce, remoteNonce}
+	if tx.Nonce() != expectedNonce {
+		return hash, &ErrBadNonce{tx.Nonce(), expectedNonce}
 	}
 
 	signedTx, err := tx.WithSignature(signer, sig)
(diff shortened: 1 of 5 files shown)


transactions/transactor.go
```diff
@@ -165,6 +145,82 @@ func (t *Transactor) SendTransactionWithSignature(args SendTxArgs, sig []byte) (
 	return signedTx.Hash(), nil
 }
 
+func (t *Transactor) HashTransaction(args SendTxArgs) (validatedArgs SendTxArgs, hash gethcommon.Hash, err error) {
```
dshulyak, line 129: i think it introduces several concurrency issues
- two concurrent transactions generated with this method will use same nonce
- local nonce won't get incremented if transaction generated with this method is submitted on chain

gravityblast: @dshulyak you are right, but in this specific case status-react will:

1 - call `HashTransaction`
2 - let the keycard sign the hash
3 - call `SendTransactionWithSignature`

Step 3 will increment the nonce. 
In theory from the app UI you won't be able to do anything else between step 1 and 3.
And also, `SendTransactionWithSignature` checks if the transaction that we are sending has the expected nonce, otherwise it fails, so we shouldn't have problems.

adambabik: Regarding concurrent `HashTransaction` calls with the same address, won't `t.addrLock.LockAddr(args.From)` prevent from using the same nonce?

EDITED: ok, I see the problem. They don't need to be concurrent actually to use the same nonce.

transactions/transactor.go
```diff
@@ -165,6 +145,82 @@ func (t *Transactor) SendTransactionWithSignature(args SendTxArgs, sig []byte) (
...
+	defer func() {
+		t.addrLock.UnlockAddr(args.From)
+	}()
+
+	var localNonce uint64
+	if val, ok := t.localNonce.Load(args.From); ok {
+		localNonce = val.(uint64)
+	}
+
+	var nonce uint64
```
adambabik: I would extract:
```go
func (t *Transactor) getTxNonce() {
    var (
        localNonce uint64
        nonce uint64
    )

    if val, ok := t.localNonce.Load(args.From); ok {
        localNonce = val.(uint64)
    }

    ctx, cancel := context.WithTimeout(context.Background(), t.rpcCallTimeout)
    defer cancel()
    nonce, err = t.pendingNonceProvider.PendingNonceAt(ctx, args.From)
    if err != nil {
        return validatedArgs, hash, err
    }
    // if upstream node returned nonce higher than ours we will use it, as it probably means
    // that another client was used for sending transactions
    if localNonce > nonce {
        nonce = localNonce
    }

    return nonce
}
```
Interesting that we don't use nonce from `SendTxArgs`.

Getting gas price and gas can also be extracted to separate method. They look the same here and in `validateAndPropagate`.

transactions/transactor.go
adambabik: Generally, I think that the flow with the keycard is fine. It's on the caller to make sure it does not execute concurrent transactions using `HashTransaction`.

dshulyak: >  ok, I see the problem. They don't need to be concurrent actually to use the same nonce.

why?

as i understand nonce is incremented in another operation called `SendTransactionWithSignature`, so the concurrency issue that i mentioned is a following one:
- HASH tx1
- HASH tx2
- SendTransactionWithSignature tx1
- SendTransactionWithSignature tx2

both tx1 and tx2 will use same nonce

i don't mind if whole flow is controlled by a client, btw there is no way to wait for this `2 - let the keycard sign the hash` in status-go?

gravityblast: yes you are right, it would be a problem meaning that tx1 and tx2 would have the same nonce but 
 SendTransactionWithSignature would fail with the second one. 
But as we said the process is managed by the client so from the UI you won't be able to do anything in between

transactions/transactor.go
gravityblast: I'll refactor it, I'm not sure if we want to touch the other methods right now since they are used from SendTransaction though.

gravityblast:
@adambabik  I'm going to refactor the method you mentioned, what about code climate? it says to refactor `buildTransaction` because it has 7 args but refactoring it would actually make it less readable I think

transactions/transactor.go
adambabik: We can do it in another PR but then let's create a GH issue.

adambabik:
@gravityblast alternatively, you can just create a struct that will represent all params passed to that method.

gravityblast:
yeah that's actually the struct used in that method, I'll see if there's another simple way to do it

gravityblast:
@adambabik I updated the code adding those small refactoring and changing some functions to make codeclimate happy.

transactions/transactor.go
dshulyak: yes, i understood that from the first comment :). and i have no issues with that. i was mostly trying to clarify why @adambabik said that tx's don't need to be concurrent

adambabik (approved):
LGTM!

Should we create a task to remove duplicated code (getting gas and gas price, for example) in `SendTransactionWithSignature` and `validateAndPropagate`?

gravityblast:
@adambabik I created an issue for that https://github.com/status-im/status-go/issues/1387

PR #1382, opened 2019-02-20
Split shhext.tracker into envelopes and mail monitors

dshulyak:
Preparatory refactoring that splits shhext.tracker into separate components that monitor envelopes and mail requests flow.


mandrigin (approved):
lgtm

services/shhext/envelopes.go
```diff
@@ -0,0 +1,168 @@
...
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/p2p/enode"
+	"github.com/status-im/status-go/services/shhext/mailservers"
+	whisper "github.com/status-im/whisper/whisperv6"
+)
+
+// EnvelopesMonitor responsible for monitoring whisper envelopes state.
```
pedropombeiro:
```suggestion
// EnvelopesMonitor is responsible for monitoring whisper envelope state.
```

pedropombeiro (approved):

services/shhext/envelopes.go
```diff
@@ -0,0 +1,168 @@
...
+type EnvelopesMonitor struct {
+	w                      *whisper.Whisper
+	handler                EnvelopeEventsHandler
+	mailServerConfirmation bool
```
adambabik, line 17: Does it mean that `EnvelopesMonitor` will send confirmation or what does this flag indicate?

services/shhext/envelopes.go
```diff
@@ -0,0 +1,168 @@
...
+
+	mailPeers *mailservers.PeerStore
+
+	wg   sync.WaitGroup
+	quit chan struct{}
+}
+
+// Start processing events.
+func (m *EnvelopesMonitor) Start() {
+	m.quit = make(chan struct{})
```
adambabik, line 31: Should we protect it from starting multiple times?

services/shhext/envelopes.go
```diff
@@ -0,0 +1,168 @@
...
+		case event := <-events:
+			m.handleEvent(event)
+		}
+	}
+}
+
+// handleEvent based on type of the event either triggers
+// confirmation handler or removes hash from tracker
+func (m *EnvelopesMonitor) handleEvent(event whisper.EnvelopeEvent) {
+	handlers := map[whisper.EventType]func(whisper.EnvelopeEvent){
```
adambabik, line 80: Switch statement wouldn't be better here?

services/shhext/envelopes.go
```diff
@@ -0,0 +1,168 @@
...
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	state, ok := m.cache[event.Hash]
+	// if we didn't send a message using extension - skip it
+	// if message was already confirmed - skip it
+	if !ok || state == EnvelopeSent {
+		return
+	}
+	log.Debug("envelope is sent", "hash", event.Hash, "peer", event.Peer)
```
adambabik, line 108: What it really means here is that an envelope was sent and a confirmation was received, correct?

services/shhext/envelopes.go
```diff
@@ -0,0 +1,168 @@
...
+	if event.Batch != (common.Hash{}) {
+		if _, ok := m.batches[event.Batch]; !ok {
+			m.batches[event.Batch] = map[common.Hash]struct{}{}
+		}
+		m.batches[event.Batch][event.Hash] = struct{}{}
+		log.Debug("waiting for a confirmation", "batch", event.Batch)
+	} else {
+		m.cache[event.Hash] = EnvelopeSent
+		if m.handler != nil {
```
adambabik, line 117: Maybe we could have `m.handler` as `noop` by default to avoid `if`s?

services/shhext/envelopes.go
```diff
@@ -0,0 +1,168 @@
...
+	defer m.mu.Unlock()
+
+	state, ok := m.cache[event.Hash]
+	// if we didn't send a message using extension - skip it
+	// if message was already confirmed - skip it
+	if !ok || state == EnvelopeSent {
+		return
+	}
+	log.Debug("envelope is sent", "hash", event.Hash, "peer", event.Peer)
+	if event.Batch != (common.Hash{}) {
```
adambabik, line 109: I don't quite get batch concept here. We handle an `EventEnvelopeSent` event and once it is has a batch field and it merely means that we need to wait for a confirmation and otherwise it actually marks an envelope as sent.

services/shhext/mailrequests.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/ethereum/go-ethereum/log"
+	whisper "github.com/status-im/whisper/whisperv6"
+)
+
+// EnvelopeState in local tracker
+type EnvelopeState int
+
+const (
+	// NotRegistered returned if asked hash wasn't registered in the tracker.
+	NotRegistered EnvelopeState = -1
```
adambabik, line 16: Any reason not to start with `iota + 1`?

services/shhext/mailrequests.go
```diff
@@ -0,0 +1,130 @@
...
+	// EnvelopePosted is set when envelope was added to a local whisper queue.
+	EnvelopePosted EnvelopeState = iota
+	// EnvelopeSent is set when envelope is sent to atleast one peer.
+	EnvelopeSent
+	// MailServerRequestSent is set when p2p request is sent to the mailserver
+	MailServerRequestSent
+)
+
+// MailRequestMonitor responsible for monitoring history request to mailservers.
+type MailRequestMonitor struct {
```
adambabik, line 26: It's quite similar to `EnvelopesMonitor`. I wonder if we can unify both but that might be a next PR.

dshulyak:
@adambabik this code was in tracker.go. i just moved it to 2 different files

services/shhext/mailrequests.go
dshulyak: the whole point was to split them because i am heavily changing EnvelopesMonitor.

services/shhext/envelopes.go
dshulyak: this flag means that we expect confirmations from mail server

services/shhext/envelopes.go
dshulyak: maybe, it was written some time ago like that

services/shhext/envelopes.go
dshulyak: it means that envelope was sent. in one case we are waiting for confirmation (like below) or we don't (this was older behavior)

services/shhext/envelopes.go
dshulyak: so the second condition  is a compatibility measure.

Issue #1381, opened 2019-02-18
Retry sending payload automatically if it expired


dshulyak:
# Problem

In current solution if envelope expires we send a signal to a caller, that we failed to deliver payload, and we wait for user actions.

# Implementation

Add a separate component that will retry to deliver payload at least couple of times before halting. Such component needs to track users payload by some unique ID and listen for envelope events. We will have to create new envelope every time expired event is received, because we need to adjust deadline that is set once envelope is sealed.

Once we reach maximum number of retries (some configurable number, e.g. 3) we can halt component completely or ignore envelope that we cannot deliver. I am in favor of former approach. We halt component and wait until new envelopes are getting confirmed, new connection appeared or just some longer time passed.

We cannot use hash of the envelope to confirm delivery to a caller now, because it will change on every retry. We will have to something generated on a caller side, or hash of the payload.

# Acceptance Criteria

Users payloads are getting re-delivered even if initially client failed to deliver them.

PR #1386, opened 2019-02-21
Re-send user payload if previously sent envelope wasn't acknowledged

dshulyak:
There are few changes merged into this PR. In the first one i removed debug API. It had only one method postSync, which allowed to wait for an envelope to get confirmed or expired. But it got into conflict with re-delivery approach and the only place where this API was used is in status-scale tests (they got abandoned anyway). So, i removed that code, if someone wants i can push those changes as a separate PR.

As for the re-delivery itself: when shhext_post called we will track user payload and original envelope ID. If that envelope expires before it get acknowledged, we will receive such event, and create new envelope to re-deliver payload. Signal that envelope expired will be sent only if we won't be able to deliver user payload with all attempts.

Maximum number of attempts can be configured with ShhextConfig.MaxMessageDeliveryAttempts options. Decided to go with single config option, but if we want different configuration per different message types we will have to add another API method. 

Envelope ID will be changing for every new envelope, cause it depends on the time when this envelope was sealed. But we need some unique way to distinguish a message, i used same approach as in our de-duplicator. I hashed the payload and the topic, and used that hash as a message ID. This message ID will be returned on shhext_post instead of envelope ID, and will be used in a signal to confirm message delivery or expiry. 

Closes # https://github.com/status-im/status-go/issues/1381


services/shhext/envelopes.go
```diff
@@ -10,15 +12,46 @@ import (
 	whisper "github.com/status-im/whisper/whisperv6"
 )
 
+func messageID(message whisper.NewMessage) common.Hash {
```
mandrigin, line 15: Let’s make it a separate type, otherwise maps look confusing. 

`type messageID common.Hash`

mandrigin (changes requested):

pedropombeiro (approved):
lgtm, once Igor's suggestion is implemented

services/shhext/envelopes.go
dshulyak: not sure if it worth it. i need to cast it in several place to common.Hash, and copy String() method. and i am using it only for messageToEnvelope map

i will add a comment which key is used for each map

mandrigin: okay

mandrigin (approved):

PR #1388, opened 2019-02-21
Fix deadlock in mailserver DeliverMail method

adambabik:
Fix an issue where `DeliverMail()` method could get into a deadlock when the connection with a peer was dropped but there was more bundles to deliver than buffered channel could store. In such a case, `WMailServer.processRequestInBundles()` would never return.

### DIFF
--- a/mailserver/mailserver.go
+++ b/mailserver/mailserver.go
@@ -295,6 +295,7 @@ func (s *WMailServer) DeliverMail(peer *whisper.Peer, request *whisper.Envelope)
 		iter,
 		bloom,
 		int(limit),
+		time.Minute,
 		bundles,
 		cancelProcessing,
 	)
@@ -369,6 +370,7 @@ func (s *WMailServer) SyncMail(peer *whisper.Peer, request whisper.SyncMailReque
 		iter,
 		request.Bloom,
 		int(request.Limit),
+		time.Minute,
 		bundles,
 		cancelProcessing,
 	)
@@ -450,6 +452,7 @@ func (s *WMailServer) processRequestInBundles(
 	iter iterator.Iterator,
 	bloom []byte,
 	limit int,
+	timeout time.Duration,
 	output chan<- []*whisper.Envelope,
 	cancel <-chan struct{},
 ) ([]byte, common.Hash) {
@@ -551,7 +554,7 @@ func (s *WMailServer) processRequestInBundles(
 			log.Error("[mailserver:processRequestInBundles] failed to push all batches",
 				"requestID", requestID)
 			break
-		case <-time.After(time.Minute):
+		case <-time.After(timeout):
 			log.Error("[mailserver:processRequestInBundles] timed out pushing a batch",
 				"requestID", requestID)
 			break
--- a/mailserver/mailserver.go
+++ b/mailserver/mailserver.go
@@ -212,18 +212,18 @@ func (s *WMailServer) Archive(env *whisper.Envelope) {
 func (s *WMailServer) DeliverMail(peer *whisper.Peer, request *whisper.Envelope) {
 	defer recoverLevelDBPanics("DeliverMail")
 
-	log.Info("Delivering mail", "peerID", peerIDString(peer))
+	log.Info("[mailserver:DeliverMail] delivering mail", "peerID", peerIDString(peer))
 
 	requestsMeter.Mark(1)
 
 	if peer == nil {
 		requestErrorsCounter.Inc(1)
-		log.Error("Whisper peer is nil")
+		log.Error("[mailserver:DeliverMail] peer is nil")
 		return
 	}
 	if s.exceedsPeerRequests(peer.ID()) {
 		requestErrorsCounter.Inc(1)
-		log.Error("Peer exceeded request per seconds limit", "peerID", peerIDString(peer))
+		log.Error("[mailserver:DeliverMail] peer exceeded the limit", "peerID", peerIDString(peer))
 		s.trySendHistoricMessageErrorResponse(peer, request, fmt.Errorf("rate limit exceeded"))
 		return
 	}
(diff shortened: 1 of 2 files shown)


mailserver/mailserver.go
```diff
@@ -274,47 +274,54 @@ func (s *WMailServer) DeliverMail(peer *whisper.Peer, request *whisper.Envelope)
...
 		close(errCh)
+		log.Info("[mailserver:DeliverMail] finished sending bundles", "counter", counter)
 	}()
 
 	start := time.Now()
 	nextPageCursor, lastEnvelopeHash := s.processRequestInBundles(
 		iter,
 		bloom,
 		int(limit),
+		time.Minute,
```
mandrigin: Can we extract it to a constant, it is in multiple places in this file.

mailserver/mailserver.go
```diff
@@ -361,7 +370,9 @@ func (s *WMailServer) SyncMail(peer *whisper.Peer, request whisper.SyncMailReque
 		iter,
 		request.Bloom,
 		int(request.Limit),
+		time.Minute,
```
mandrigin: constant?

mailserver/mailserver.go
```diff
@@ -531,15 +544,28 @@ func (s *WMailServer) processRequestInBundles(
...
 	for _, batch := range batches {
-		output <- batch
+		select {
+		case output <- batch:
+		// It might happen that during producing the batches,
+		// the connection with the peer goes down and
+		// the consumer of `output` channel exits prematurely.
+		// In such a case, we should stop pushing batches and exit.
+		case <-cancel:
+			log.Error("[mailserver:processRequestInBundles] failed to push all batches",
```
mandrigin: I would make this a `Warn` or even an `Info`, it sounds like it is more or less a normal and expected scenario, as opposed to the timeout in this loop.

mandrigin (approved):
Generally, looks very good!
Just a few minor changes, but feel free to ignore if you disagree.

mailserver/mailserver.go
adambabik: That's true, it might be a perfectly normal scenario when a peer looses a connection to a mailserver, especially when being on a 3G or something like that.

mailserver/mailserver.go
```diff
@@ -531,15 +544,28 @@ func (s *WMailServer) processRequestInBundles(
...
+				"requestID", requestID)
+			break
+		case <-time.After(timeout):
```
dshulyak, line 558: i noticed discussion about this... but it feels really weird. there is a finite conditions that needs to be taken into account. if consumer stopped - this loop needs to stop too, otherwise it should work. literally no reason to add such hack

mandrigin: No reason if this or underlying code will never change and never get any regressions in the future.

dshulyak: like why not to add timeout to every select block :) ?

mandrigin: I actually think that using lots of selects and channels with no timeouts is a bad coding practice, especially if your code supposed to be run for days and months with no restarts.

dshulyak: oh, really? i think you need to define concrete conditions when select statement will terminate. and it doesn't have to be limited by a timer. for instance if you look at p2p/server.go, most of the select statements never can be terminated by a timer, it would be just wrong

dshulyak: i can imagine different pattern, where whole DeliverMail procedure is time-limited, but it would be implemented differently

adambabik: I agree that this case should never happen and `case <-cancel` is here to cover stopped peers. However, we don't own devp2p stack and apparently we don't trust it. We can run this code for a few days and if this log never appears (which I believe is the case), this fragment can be removed.

> I actually think that using lots of selects and channels with no timeouts is a bad coding practice, especially if your code supposed to be run for days and months with no restarts.

If the code is well architected, I think timeouts are not needed in such cases. They may appear at some entry points handling incoming requests or explicitly timing out network connections if there is no way to timeout the connection itself due to some reason.

mandrigin: Sure. But if you have an assumption that some code should complete in max N minutes after some event, having a watchdog for it both helps to identifying issues and mitigating consequences.
For instance, no modern mobile OS will let you block the rendering loop for too long. Same is true for most modern desktop OSes. It wasn't true for Windows 95 though, and we all seen the consequences.

mandrigin: > If the code is well architected, I think timeouts are not needed in such cases.

Yeah, if there is no potential for a deadlock then it is obviously not needed. Unfortunately, in Go, when using channels you are asking for trouble :-P

dshulyak: > I agree that this case should never happen and `case <-cancel` is here to cover stopped peers. However, we don't own devp2p stack and apparently we don't trust it.

the bug wasn't in devp2p code though :)

> We can run this code for a few days and if this log never appears (which I believe is the case), this fragment can be removed.

yeah i think that it won't appear too. unless 1 minute is not enough to deliver data for some of the congested mail servers, like it was with HK mail servers. but if that happens, isn't this timeout is harmful?

adambabik: > the bug wasn't in devp2p code though :)

True but we still don't have a clear picture how devp2p behaves in the case of half-open TCP connections or network switches.

> but if that happens, isn't this timeout is harmful?

I don't see anything harmful here. It will close `bundles` channel which will exit goroutine processing it and `DeliverMail` will exit normally. The only bad thing that happens is that not all bundles (more precisely older envelopes) will get delivered.

dshulyak: > True but we still don't have a clear picture how devp2p behaves in the case of half-open TCP connections or network switches.

i see that either side will be eventually closed by read/write deadline. it seems pretty straightforward. what do you think is confusing?

adambabik: It seems :) here there is no explicit timeout when sending envelopes, so that part is confusing -- after what time it times out.

dshulyak: it will time out after write deadline passed since actual write started

adambabik: It's still implicit. Can you give me precise duration here?

There are a few logs so this condition is possible: https://kibana.status.im/goto/4b14a7de6e58a9237a18657beb94712b. On the other hand, these logs are old and comes from a period of time when we were deploying stuff.

dshulyak: it is hard to give precise duration. all writes are processed sequentially, and the deadline is counted from the time message hits socket. if there are no other writes - single write can take at most 20s.

> There are a few logs so this condition is possible: https://kibana.status.im/goto/4b14a7de6e58a9237a18657beb94712b. On the other hand, these logs are old and comes from a period of time when we were deploying stuff.

got it. but it is still possible on a slow connection. e.g. if there are 10 batches, and every single batch takes 19s to deliver.. in the end it will take longer than a minute

PR #1389, opened 2019-02-21
Fix wrong datadir in desktop

cammellos:
Desktop was wrongly passing a network dependent data-dir (mainnet_rpc).
The code migrates the db if present, moving it to the new location.

### DIFF
--- a/services/shhext/service.go
+++ b/services/shhext/service.go
@@ -162,6 +162,18 @@ func (s *Service) initProtocol(address, encKey, password string) error {
 		os.Remove(v4Path)
 	}
 
+	// Desktop was passing a network dependent directory, which meant that
+	// if running on testnet it would not access the right db. This copies
+	// the db from mainnet to the root location.
+	networkDependentPath := filepath.Join(s.dataDir, "ethereum", "mainnet_rpc", fmt.Sprintf("%s.v4.db", s.installationID))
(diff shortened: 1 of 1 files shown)


services/shhext/service.go
```diff
@@ -162,6 +162,18 @@ func (s *Service) initProtocol(address, encKey, password string) error {
 		os.Remove(v4Path)
 	}
 
+	// Desktop was passing a network dependent directory, which meant that
+	// if running on testnet it would not access the right db. This copies
+	// the db from mainnet to the root location.
+	networkDependentPath := filepath.Join(s.dataDir, "ethereum", "mainnet_rpc", fmt.Sprintf("%s.v4.db", s.installationID))
```
adambabik, line 168: Why was that file put in a wrong dir in the first place? Is this code just for backward compatibility?

cammellos: On desktop the directory for the database was defaulted to `DataDir`, which is network dependent (`ethereum/mainnet_rcp`), so that will be changed in status-react to pass a different directory, but here we move over the file if found, it will only work once status-react will pass the new directory (otherwise everything works as before).

cammellos: @adambabik would you mind reviewing again? Thanks!

adambabik (approved):

PR #1394, opened 2019-02-25
Fix `admin_nodeInfo` to provide the correct version again

mandrigin:
When we switched to gomobile, a few build flags were forgotten,
including the one that passes commit and version number.
This change fixes that.

(see https://github.com/status-im/status-react/issues/7522)


mandrigin:
It works after this fix: simulator screen shot - iphone 8 - 2019-02-25 at 15 52 26

pedropombeiro (approved):

adambabik (approved):

mandrigin:
(force-push was to bump status-go version in the same change)

cammellos (approved):

jakubgs (approved):

PR #1395, opened 2019-02-25
move running tests to a separate jenkins job

jakubgs:
We'll run them in parallel, should make things faster and allow us to drop Travis.

Status: WIP

### DIFF
--- a/_assets/ci/Jenkinsfile.tests
+++ b/_assets/ci/Jenkinsfile.tests
@@ -7,11 +7,6 @@ pipeline {
       defaultValue: 'develop',
       description: 'Name of branch to build.'
     )
-    booleanParam(
-      name: 'RELEASE',
-      defaultValue: false,
-      description: 'Enable to create build for release.',
-    )
   }
 
   options {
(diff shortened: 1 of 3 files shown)


mandrigin (approved):

mandrigin:
@jakubgs is it still WIP?

jakubgs:
@mandrigin I talked to @adambabik and he thinks that what we should do is run it all as one job. As in, first run the tests, then the builds. But that kinds breaks our workflow.

1. That means all builds will show up on github as one build
2. It increases the time to build artifacts by ~12 minutes
3. You would have to re-run the whole thing if one of the platform fails
4. It complicates the flow in Jenkinsfiles

I think it's better to run tests as a separate job. If we set the requirements for the PRs right we won't be able to merge without them passing.

mandrigin:
@jakubgs yeah, I tend to agree with you, because what you want to be able to do is to restart all the jobs separately. Some teams go as far as having unit-tests, lint, e2e as separate jobs too.
We have codeclimate as a separate job and the packages check. I don't think with status-go it will add any overhead to our CI.

mandrigin:
@adambabik what do you think?

_assets/ci/Jenkinsfile.tests
```diff
@@ -0,0 +1,69 @@
+pipeline {
```
adambabik, line 1: Doesn't it need to be used somewhere?

adambabik (approved):

_assets/ci/Jenkinsfile.tests
jakubgs: What do you mean?

adambabik: I am just wondering where and how this Jenkinsfile is used.

jakubgs: https://ci.status.im/job/status-go/job/prs/job/tests/view/change-requests/ - here

jakubgs: And you can see it below, as `jenkins/prs/tests`.

adambabik: Oh ok, configured directly in Jenkins. All clear.

PR #1398, opened 2019-02-27
expose more logs and metrics in DeliverMail

adambabik:
These changes expose more logs from `DeliverMail()` method and add an important metric that will measure the whole time needed to execute that method.

### DIFF
--- a/mailserver/mailserver.go
+++ b/mailserver/mailserver.go
@@ -213,7 +213,8 @@ func (s *WMailServer) Archive(env *whisper.Envelope) {
 func (s *WMailServer) DeliverMail(peer *whisper.Peer, request *whisper.Envelope) {
 	defer recoverLevelDBPanics("DeliverMail")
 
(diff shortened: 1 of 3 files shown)


mailserver/mailserver.go
```diff
@@ -334,6 +369,8 @@ func (s *WMailServer) SyncMail(peer *whisper.Peer, request whisper.SyncMailReque
 
 	defer recoverLevelDBPanics("SyncMail")
 
+	requestID := fmt.Sprintf("%d", time.Now().UnixNano())
```
mandrigin: to be absolutely paranoid, maybe we can make it like
```suggestion
	requestID := fmt.Sprintf("%d-%d", time.Now().UnixNano(), rand.Intn(1000))
```

mandrigin (approved):

mailserver/mailserver.go
adambabik: Why not :) I copied it from the previous implementation.

PR #1399, opened 2019-02-28
Use lumberjack library to rotate logs and add config that enables it

dshulyak:

### DIFF
--- a/params/config.go
+++ b/params/config.go
@@ -265,6 +265,15 @@ type NodeConfig struct {
 	// LogLevel defines minimum log level. Valid names are "ERROR", "WARN", "INFO", "DEBUG", and "TRACE".
 	LogLevel string `validate:"eq=ERROR|eq=WARN|eq=INFO|eq=DEBUG|eq=TRACE"`
 
+	// LogMaxBackups defines number of rotated log files that will be stored.
+	LogMaxBackups int
+
+	// LogMaxSize after current size is reached log file will be rotated
+	LogMaxSize int
+
+	// LogCompressRotated if true all rotated files will be gzipped.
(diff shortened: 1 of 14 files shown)


mandrigin (approved):
great changes!

adambabik (approved):
LGTM!

I would test this PR with iOS and Android before merging, if possible, because we used to have problems with log files permissions.

dshulyak:
Yeah trying to prepare a build for tests. I wonder though what kind of permissions issues we had. As i understand bindings have exactly same permissions to modify file system as a statusgo part. So what was solved by creating geth.log directly in native bindings?

params/config.go
```diff
@@ -378,6 +387,9 @@ func NewNodeConfigWithDefaults(dataDir string, networkID uint64, opts ...Option)
 	c.ListenAddr = ":30303"
 	c.LogEnabled = true
 	c.LogLevel = "INFO"
+	c.LogMaxSize = 100
```
mandrigin, line 390: is it megabytes, right?

mandrigin:
@dshulyak just do double-check, if I won't specify the log size limit when calling `StartNode`, but will specify other params, it will be limited to 100MB, right?

params/config.go
dshulyak: yes

mandrigin: thanks!

dshulyak:
> @dshulyak just do double-check, if I won't specify the log size limit when calling `StartNode`, but will specify other params, it will be limited to 100MB, right?

yes, by default 100mb, at most 2 last rotated logs will be kept and rotated logs will be compressed

PR #1401, opened 2019-02-28
validate signature size on SendTransactionWithSignature

gravityblast:
Check signature size to avoid panic from go-ethereum

### DIFF
--- a/transactions/transactor.go
+++ b/transactions/transactor.go
@@ -3,6 +3,7 @@ package transactions
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"math/big"
 	"sync"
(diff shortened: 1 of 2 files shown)


mandrigin (approved):
A test would be nice, but looks good otherwise.

gravityblast:
thanks @mandrigin , I added a test, waiting now for CI

transactions/transactor.go
```diff
@@ -24,8 +25,13 @@ const (
 	sendTxTimeout = 300 * time.Second
 
 	defaultGas = 90000
+
+	validSignatureSize = 65
```
adambabik, line 29: Can we take this value from go-ethereum package?

adambabik (approved):

transactions/transactor.go
gravityblast: @adambabik I see it's actually hardcoded there: 

https://github.com/status-im/status-go/blob/develop/vendor/github.com/ethereum/go-ethereum/core/types/transaction_signing.go#L196

PR #1402, opened 2019-03-01
Add a method to mobile bindings that allows to dump content of the logs file(s)

dshulyak:
This method allows to read content of the file that is current used for logging and return it's content to a caller. In future this method can be extended to read content of the files that were log rotated, and return compressed content.

It allows to get logs in situation when file system is accessible only by a mobile app.

### DIFF
--- a/exportlogs/logs.go
+++ b/exportlogs/logs.go
@@ -1,0 +1,34 @@
+package exportlogs
+
+import (
+	"fmt"
+	"io/ioutil"
+
+	"github.com/status-im/go-ethereum/common/hexutil"
+)
+
+// Log contains actual log content and filename. If content is gzipped Compressed will be set to true.
+type Log struct {
+	Filename   string
+	Content    []byte
+	Compressed bool
+}
+
+// ExportResponse contains all available logs
+type ExportResponse struct {
+	Error string
+	Logs  []Log
+}
+
+// ExportFromBaseFile reads log file and returns its content with some meta.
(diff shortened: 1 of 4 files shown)


mandrigin (approved):
Great feature!

I would rename `Dump` to `Export` everywhere, otherwise looks solid!

mobile/dumplogs.go
```diff
@@ -0,0 +1,34 @@
+package statusgo
+
+import (
+	"fmt"
+	"io/ioutil"
+
+	"github.com/status-im/go-ethereum/common/hexutil"
+)
+
+// Log containts actual log content and filename. If content is gzipped Compressed will be set to true.
```
pedropombeiro:
```suggestion
// Log contains actual log content and filename. If content is gzipped Compressed will be set to true.
```

pedropombeiro (approved):

adambabik (commented):
For features parity, we should also add `DumpNodeLogs` to `lib/`. Otherwise, the Clojure code needs to have some if-based logic detecting if a binding is available or not.

mandrigin:
@adambabik oh, great point!

dshulyak:
@adambabik @mandrigin sure, i will add this function to lib as well.
but can you clarify when we are using lib/mobile? i thought that we are making a complete switch to mobile, and don't really understand why do we need to keep both

mandrigin:
@dshulyak probably Desktop still uses `/lib`,maybe we should rename it.

PR #1404, opened 2019-03-05
Redirect log output to mobile system loggers

dshulyak:
Gomobile redirects output of the logger from std lib to mobile system loggers. For example in android log of the go application can be viewed with `adb logcat GoLog:I *:S`. 

This change wraps logger from std lib in a log.Handler interface.
sources: https://github.com/golang/mobile/blob/master/internal/mobileinit/mobileinit_android.go

### DIFF
--- a/logutils/stdhandler.go
+++ b/logutils/stdhandler.go
@@ -1,0 +1,17 @@
+package logutils
+
+import (
+	stdlog "log"
+
+	"github.com/ethereum/go-ethereum/log"
+)
+
+// NewStdHandler returns handler that uses logger from golang std lib.
+func NewStdHandler(fmtr log.Format) log.Handler {
+	return log.FuncHandler(func(r *log.Record) error {
+		line := fmtr.Format(r)
+		// 8 is a number of frames that will be skipped when log is printed.
+		// this is needed to show the file (with line number) where call to a logger was made
+		return stdlog.Output(8, string(line))
+	})
+}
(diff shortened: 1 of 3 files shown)


logutils/stdhandler.go
```diff
@@ -0,0 +1,15 @@
...
+	stdlog "log"
+
+	"github.com/ethereum/go-ethereum/log"
+)
+
+// NewStdHandler returns handler that uses logger from golang std lib.
+func NewStdHandler(fmtr log.Format) log.Handler {
+	return log.FuncHandler(func(r *log.Record) error {
+		line := fmtr.Format(r)
+		return stdlog.Output(2, string(line))
```
adambabik: Let's explain `2`.

adambabik (approved):

logutils/stdhandler.go
dshulyak: tbh that was a default value used for calldepth in log.Println/Printf. and i don't completely understand why it is 2. will try to figure that out :)

params/config.go
```diff
@@ -259,6 +259,9 @@ type NodeConfig struct {
 	// LogEnabled enables the logger
 	LogEnabled bool `json:"LogEnabled"`
 
+	// LogMobileSystem redirects log to android/ios system logger.
```
pedropombeiro:
```suggestion
	// LogMobileSystem enables log redirection to android/ios system logger.
```

pedropombeiro (approved):
Glad to hear we'll be able to see these logs in adb logcat!

logutils/stdhandler.go
dshulyak: turned out ethereum.Handler adds a lot of additional frames. so if someone will want to print where call to a logger was made we will have to skip 8 frames. will add explainer to file

PR #1411, opened 2019-03-11
Default limit & upper bound

cammellos:
We allow the user not to specify an upper bound, in which case it will
default to Time.Now() + 60 seconds, to accomodate for messages with
higher timestamp.

Limit is also defaulted to 2000 if not set.

We removed as well the 24 hours limitation as now all the requests will
come with a limit.

### DIFF
--- a/mailserver/mailserver.go
+++ b/mailserver/mailserver.go
@@ -40,7 +40,10 @@ import (
 
 const (
 	maxQueryRange = 24 * time.Hour
-	noLimits      = 0
+	defaultLimit  = 2000
+	// When we default the upper limit, we want to extend the range a bit
+	// to accommodate for envelopes with slightly higher timestamp, in seconds
+	whisperTTLSafeThreshold = 60
 )
 
 var (
@@ -263,6 +266,10 @@ func (s *WMailServer) DeliverMail(peer *whisper.Peer, request *whisper.Envelope)
 			"peerID", peerID,
 			"requestID", requestID)
 		lower, upper, bloom, limit, cursor, err = s.validateRequest(peer.ID(), request)
+	}
+
+	if limit == 0 {
+		limit = defaultLimit
 	}
 
 	if err != nil {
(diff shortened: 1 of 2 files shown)


mailserver/mailserver.go
```diff
@@ -535,7 +542,7 @@ func (s *WMailServer) processRequestInBundles(
 		lastEnvelopeHash = envelope.Hash()
 		processedEnvelopes++
 		envelopeSize := whisper.EnvelopeHeaderLength + uint32(len(envelope.Data))
-		limitReached := limit != noLimits && processedEnvelopes == limit
+		limitReached := processedEnvelopes == limit
```
dshulyak, line 545: it doesn't matter if we received less then a limit?

cammellos: what do you mean exactly?
If there are no more envelopes (less than the limit) we leave the loop, and never reach this condition, is this what you were thinking?

dshulyak:
> We allow the user not to specify an upper bound, in which case it will
default to Time.Now() + 60 seconds, to accomodate for messages with
higher timestamp.

do you suspect that we are missing some messages because sender time is slightly out of sync?

dshulyak (approved):

cammellos:
@dshulyak it does happen to me at least on the emulator, it gets out of sync fairly easily and I miss recent messages if going from offline->online.

mailserver/mailserver.go
dshulyak: yeah, sorry. i didn't realize that this part was reading from db until limit is reached..

mailserver/mailserver.go
```diff
@@ -40,7 +40,10 @@ import (
 
 const (
 	maxQueryRange = 24 * time.Hour
-	noLimits      = 0
+	defaultLimit  = 2000
+	// When we default the upper limit, we want to extend the range a bit
+	// to accomodate for envelopes with slightly higher timestamp
+	whisperTTLSafeThreshold = 60
```
adambabik, line 46: We should add that it is in seconds.

adambabik (approved):

PR #1412, opened 2019-03-12
add shhext_requestMessagesNew

adambabik:
I want to have a method to request historic messages and be it synchronous so that I don't need to listen to signals on my own.

This change adds `shhext_requestMessagesNew` which waits for a proper Whisper events regarding the current request and returns a cursor or an error. This is a new method any none of the existing ones has been been affected.

### DIFF
--- a/services/shhext/api.go
+++ b/services/shhext/api.go
@@ -106,6 +106,24 @@ func (r *MessagesRequest) setDefaults(now time.Time) {
 	}
 }
 
+func (r MessagesRequest) validate() error {
+	if r.From > r.To {
+		return fmt.Errorf("query range is invalid: from > to (%d > %d)", r.From, r.To)
+	}
+	return nil
+}
+
+// MessagesResponse is a response for shhext_requestMessages2 method.
+type MessagesResponse struct {
+	// Cursor from the response can be used to retrieve more messages
+	// for the previous request.
+	Cursor string `json:"cursor"`
+
+	// Error indicates that something wrong happened when sending messages
+	// to the requester.
+	Error string `json:"error"`
+}
+
 // SyncMessagesRequest is a SyncMessages() request payload.
 type SyncMessagesRequest struct {
 	// MailServerPeer is MailServer's enode address.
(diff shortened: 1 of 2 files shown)


dshulyak:
@adambabik so what is the difference with RequestMessagesSync ?

services/shhext/api.go
```diff
@@ -234,6 +252,102 @@ func waitForExpiredOrCompleted(requestID common.Hash, events chan whisper.Envelo
 	}
 }
 
+// RequestMessagesNew requests historic messages from a mail server
+// and blocks until the response from the mail server arrives.
+// nolint: gocyclo
+func (api *PublicAPI) RequestMessagesNew(ctx context.Context, r MessagesRequest) (MessagesResponse, error) {
+	log.Info("[shhext/PublicAPI.RequestMessagesNew] called", "request", r)
```
dshulyak: not sure why code from RequestMessages needs to be copied. Isn't it enough to call RequestMessages and then wait for the event in a loop?

dshulyak:
So it looks like that the main difference is that RequestMessagesSync doesn't return cursor. if that's the case i would suggest to change RequestMessagesSync. To me it looks like waitForExpiredOrCompleted needs to return that cursor from event, and thats it

adambabik:
@dshulyak yeah, it looks very similar. Does anyone use `RequestMessagesSync` or I can just change it now and it should not affect anyone?

services/shhext/api.go
adambabik: Yeah, that's true. I though I would need to change more but apparently it is almost the same + waiting for an event.

adambabik: But I think changing the result of `RequestMessagesSync` is a better idea if no one uses it and I won't break compatibility with anyone.

dshulyak:
@adambabik I think that no one uses it

services/shhext/api.go
```diff
@@ -234,6 +252,102 @@ func waitForExpiredOrCompleted(requestID common.Hash, events chan whisper.Envelo
 	}
 }
 
+// RequestMessagesNew requests historic messages from a mail server
+// and blocks until the response from the mail server arrives.
+// nolint: gocyclo
+func (api *PublicAPI) RequestMessagesNew(ctx context.Context, r MessagesRequest) (MessagesResponse, error) {
```
pedropombeiro: Doesn't calling this method `RequestMessagesSync` make more sense?

services/shhext/service_test.go
```diff
@@ -647,3 +648,88 @@ func (s *WhisperRetriesSuite) TestDeliveredFromFirstAttempt() {
 func (s *WhisperRetriesSuite) TestDeliveredFromSecondAttempt() {
 	s.testDelivery(2)
 }
+
+type RequestMessages2Suite struct {
```
pedropombeiro: Shouldn't this be called `RequestMessages2Suite` or `RequestMessagesSyncSuite`?

adambabik:
I decided to add a response struct to `RequestMessagesSync` method result.

Please review again.

dshulyak (approved):

mandrigin (approved):

Issue #1413, opened 2019-03-13
Support Goerli testnet


adambabik:
# Problem

In [go-ethereum v1.8.23](https://github.com/ethereum/go-ethereum/releases/tag/v1.8.23), a support for Görli testnet was added with a flag. status-go should also support this new test network.

# Implementation

1. Update our branch in [go-ethereum fork](https://github.com/status-im/go-ethereum/tree/patched/1.8) to v1.8.23,
2. Add Görli infura URL to `params/defaults.go`,
3. Add Görli network ID to `params/defaults.go`.

# Acceptance Criteria

Test sending transactions using Görli.

PR #1415, opened 2019-03-14
Goerli testnet

dshulyak:
I wasn't able to run test suite using goerli testnet yet. Faucet seems to be stuck, i got a confirmation from it that funds were allocated. But according to etherescan nothing was transacted to any of my addresses.

Closes: https://github.com/status-im/status-go/issues/1413
This change includes geth update https://github.com/status-im/status-go/pull/1414

### DIFF
--- a/params/defaults.go
+++ b/params/defaults.go
@@ -39,6 +39,10 @@ const (
 	// allow us avoid syncing node.
 	RinkebyEthereumNetworkURL = "https://rinkeby.infura.io/nKmXgiFgc2KqtoQ8BCGJ"
 
+	// GoerliEthereumNetworkURL is an open RPC endpoint to Goerli network
+	// Other RPC endpoints are available here: http://goerli.blockscout.com/
+	GoerliEthereumNetworkURL = "http://goerli.blockscout.com/"
+
 	// MainNetworkID is id of the main network
 	MainNetworkID = 1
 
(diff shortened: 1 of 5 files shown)


dshulyak:
It looks like that our e2e transactions suite verifies only that transactions were sent. So it basically doesn't make any difference which network is used. Also, the suite gets stuck on synchronization with any network except our mock, StatusChain.

params/defaults.go
```diff
@@ -39,6 +39,10 @@ const (
 	// allow us avoid syncing node.
 	RinkebyEthereumNetworkURL = "https://rinkeby.infura.io/nKmXgiFgc2KqtoQ8BCGJ"
 
+	// GoerliEthereumNetworkURL is an open RPC endpoint to Goerli network
+	// Other RPC endpoints are available here: http://goerli.blockscout.com/
+	GoerliEthereumNetworkURL = "http://goerli.blockscout.com/"
```
adambabik, line 44: So Infura does not support Goerli?

dshulyak: yeah, no rpc endpoint from them. they are going to setup one in the future. also they run one of the block signers for this poa network

dshulyak:
So what about this change? Are there any concerns?

adambabik (approved):

PR #1417, opened 2019-03-15
[WIP] add mailserver-tools command

adambabik:
`mailserver-tools` has a command `count` which returns a number of envelopes for a given topic.

We want to have a capability of verifying how new private chats topic splitting works.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,6 +106,11 @@ node-canary: ##@build Build P2P node canary using status-go deps
 	go build -i -o $(GOBIN)/node-canary -v -tags '$(BUILD_TAGS)' $(BUILD_FLAGS) ./cmd/node-canary/
 	@echo "Compilation done."
 
+mailserver-tools: ##@build Build mailserver-tools
+	go build -i -o $(GOBIN)/mailserver-tools -v -tags '$(BUILD_TAGS)' $(BUILD_FLAGS) ./cmd/mailserver-tools/
+	@echo "Compilation done."
+
+
 statusgo-cross: statusgo-android statusgo-ios
 	@echo "Full cross compilation done."
 	@ls -ld $(GOBIN)/statusgo-*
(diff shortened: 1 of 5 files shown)


adambabik:
The idea how to collect envelope counts per topic is to run this tool every e.g. an hour and store the result in Prometheus, for example. However, as it will be a short lived session, we need Prometheus Push Gateway to do that. @jakubgs can we run https://github.com/prometheus/pushgateway along our Prometheus instances?

mandrigin (approved):

jakubgs:
>can we run https://github.com/prometheus/pushgateway along our Prometheus instances?

@adambabik I guess we could do that, but why can't we use the normal pull way of getting all our metrics into Prometheus? Is it because you didn't make this tool as a service that could host a `/metrics` endpoint?

If that's the case than that's hardly an issue. It could be as simple as a `cron` script that generate a `/var/www/mailsever-envelopes/metrics` text file in Prometheus format that would be hosted by Nginx. Or do you have some other reason for not wanting to do that?

adambabik:
@jakubgs yeah, I didn't want to run it as a service all the time because there is no reason to do that.

I can work with generating metrics into a file if that's easier to consumer by Prometheus.

jakubgs:
As long as the tool returns a usable number, I can just wrap it in a bash script that will format it for Prometheus. No issue.

jakubgs:
What is the status of this? Is this on hold until we figure out the decision on the protocol breaking changes?

adambabik:
@jakubgs yes, when the implementation for partitioned topics is confirmed, this will be updated.

jakubgs:
I don't think this is stale.

jakubgs:
@adambabik is this still relevant?

adambabik:
Nope, it's already changed.

PR #1418, opened 2019-03-15
prefix git tag with v

adambabik:
In order to use status-go with Go modules, the git tags must start with a letter `v`.


jakubgs (approved):
lgtm

PR #1419, opened 2019-03-19
Add a script to bootstrap mail server

dshulyak:
Example output:
```
Create directory /home/dd/.status-node.
Create configuration file: /home/dd/.status-node/config.json.
Start docker container: 'status-mailserver-node' using image 'statusteam/status-go:latest'.
Mailserver enode: "enode://31ad24c7f482be2f11da81c29a93f09b4ccc732b3ed627f92842ac1915af5496c9712f30c023d5602cce48860614cb6da4fec923d90e2eca238eced476011a06@109.86.198.208:30303?discport=0"
```

### DIFF
--- a/_assets/mailserver/start.sh
+++ b/_assets/mailserver/start.sh
@@ -1,0 +1,72 @@
+#!/bin/bash
+
+MAILSERVER_IMAGE="${MAILSERVER_IMAGE:-statusteam/status-go:0.23.0-beta.6}"
+DEFAULT_HOST_DATA_DIR="$HOME/.status-node"
+HOST_DATA_DIR="${HOST_DATA_DIR:-"${DEFAULT_HOST_DATA_DIR}"}"
+DOCKER_CONTAINER="${DOCKER_CONTAINER:-status-mailserver-node}"
+EXTERNAL_ADDRESS=${EXTERNAL_ADDRESS:-}
+
+if [ ! -x "$(command -v docker)" ]; then
+  echo "Please install docker. For instructions refer to https://docs.docker.com/install/"
+  exit 1
+fi
+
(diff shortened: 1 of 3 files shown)


jakubgs:
We already have this... it's in https://github.com/status-im/status-go/tree/develop/_assets/compose/mailserver

_assets/mailserver/start.sh
```diff
@@ -0,0 +1,72 @@
...
+if [ ! -x "$(command -v docker)" ]; then
+  echo "Please install docker. For instructions refer to https://docs.docker.com/install/"
+  exit 1
+fi
+
+echo "Create directory ${HOST_DATA_DIR}."
+mkdir -p $HOST_DATA_DIR
+rm -f "$HOST_DATA_DIR/config.json"
+echo "Create configuration file: ${HOST_DATA_DIR}/config.json."
+cat <<EOF >> "$HOST_DATA_DIR/config.json"
```
jakubgs, line 18: Why not use the config in [`/config/cli/fleet-eth.beta.json`](/config/cli/fleet-eth.beta.json)?

dshulyak: I thought that it would be better if someone can run this script without checking out whole repository.

dshulyak:
> We already have this... it's in https://github.com/status-im/status-go/tree/develop/_assets/compose/mailserver

thanks, this is great. i wasn't aware and didn't look into that place obviously. it will require checking out whole repo, but i think this not a problem at this point.

i am going to close this change. @mandrigin do you see any issues with checking out whole repository and running make by an end user?

mandrigin:
@dshulyak at the moment we think that the user should have enough tech competency to do that, so no problem to me.

PR #1421, opened 2019-03-20
add HashMessage to backend, lib, and mobile.

gravityblast:
HashMessage is used by status-react to generate the hash of a message and sign it with a keycard.
The hash generated is the keccak256 of `fmt.Sprintf("\x19Ethereum Signed Message:\n%d%s", len(data), data)`

`go-ethereum`  doesn't have a constant for that string, so I used it hardcoded like in their code.

### DIFF
--- a/api/backend.go
+++ b/api/backend.go
@@ -709,3 +709,12 @@ func (b *StatusBackend) UpdateMailservers(enodes []string) error {
 	}
 	return st.UpdateMailservers(nodes)
 }
+
+// HashMessage calculates the hash of a message to be safely signed by the keycard
+// The hash is calulcated as
+//   keccak256("\x19Ethereum Signed Message:\n"${message length}${message}).
+// This gives context to the signed message and prevents signing of transactions.
+func (b *StatusBackend) HashMessage(data []byte) []byte {
+	msg := fmt.Sprintf("\x19Ethereum Signed Message:\n%d%s", len(data), data)
+	return ethcrypto.Keccak256([]byte(msg))
+}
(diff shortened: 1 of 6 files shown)


api/backend_test.go
```diff
@@ -382,3 +383,35 @@ func TestStartStopMultipleTimes(t *testing.T) {
...
+	require.NoError(t, err)
+	addr := crypto.PubkeyToAddress(key.PublicKey)
+
+	originalMessage := []byte{0x01, 0x02, 0x03}
+	hash := backend.HashMessage(originalMessage)
+
+	// simulate signature from external signer like a keycard
+	sig, err := crypto.Sign(hash, key)
+	require.NoError(t, err)
+	sig[64] += 27
```
adambabik: This definitely needs an explanation :P

api/backend.go
```diff
@@ -709,3 +709,12 @@ func (b *StatusBackend) UpdateMailservers(enodes []string) error {
 	}
 	return st.UpdateMailservers(nodes)
 }
+
+// HashMessage calculates the hash of a message to be safely signed by the keycard
+// The hash is calulcated as
+//   keccak256("\x19Ethereum Signed Message:\n"${message length}${message}).
+// This gives context to the signed message and prevents signing of transactions.
+func (b *StatusBackend) HashMessage(data []byte) []byte {
```
adambabik: It can be a standalone function in `api/` instead of `StatusBackend` method.

api/backend_test.go
gravityblast: yeah it's true, shall I add a comment pointing to https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go#L431 ?

api/backend.go
gravityblast: ok I'll move it!

gravityblast:
thank you @adambabik I changed the code after your comments

adambabik (approved):
LGTM 👏

PR #1422, opened 2019-03-20
[Chaos Unicorn] Implement ChaosModeSetUpstreamURL to dynamically change the upstream URL

adambabik:
The upstream URL of the RPC client can be changed dynamically without restarting the node. This is needed to properly simulate Infura being down during Chaos Unicorn Day.

### DIFF
--- a/mobile/status.go
+++ b/mobile/status.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
 	"unsafe"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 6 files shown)


mobile/status.go
```diff
@@ -604,3 +604,14 @@ func ExportNodeLogs() string {
 	}
 	return string(data)
 }
+
+// ChaosModeSetUpstreamURL changes the URL of the upstream RPC client.
+func ChaosModeSetUpstreamURL(url string) string {
```
mandrigin: I would pass a Boolean here and the just check the urls of private and public RPC endpoints.

mandrigin (approved):
One note, looks good otherwise.

adambabik:
@mandrigin I changed it to "on/off" instead of passing an URL.

mandrigin (approved):
Thank you! Looks ready to go now. Can you prepare a release?

adambabik:
@mandrigin just to confirm, we want to merge it to develop and create a regular release?

mandrigin:
I would base the release on 0.23.0-beta.8, so we have at least as possible regression scope.

PR #1423, opened 2019-03-21
forgot to update main README with link to instructions

jakubgs:
Forgot to do this in #1420 because I'm stupid like that.


Swader:
How about adding the run on ARM instructions too? https://our.status.im/bulletproofing-against-chaos-unicorns-with-status-on-arm/

jakubgs:
I don't know if the main `README` for this repo is the right place to advertise blog posts. @adambabik what do you think?

adambabik (approved):

adambabik (approved):

PR #1424, opened 2019-03-22
Notify users that envelope was discarded and retry sending it

dshulyak:
Changes in the whisper that enable this feature: https://github.com/status-im/whisper/pull/23

After some experimentation i decided to extend existing signal. This signal will be extended with a Message keyword. When we receive error from a peer that envelope was discarded because of the time sync issue - Message will be populated with appropriate error.

So, basically from the client perspective this change doesn't introduce any new steps in the flow of signals. Client still waits for an ExpiredSignal. But now we can provide more feedback why certain envelopes weren't delivered.

### DIFF
--- a/services/shhext/envelopes.go
+++ b/services/shhext/envelopes.go
@@ -195,8 +195,8 @@ func (m *EnvelopesMonitor) handleAcknowledgedBatch(event whisper.EnvelopeEvent) 
 	}
 	log.Debug("received a confirmation", "batch", event.Batch, "peer", event.Peer)
 	envelopeErrors, ok := event.Data.([]whisper.EnvelopeError)
-	if !ok {
-		log.Error("received unexpected data in the the confirmation event", "batch", event.Batch)
+	if event.Data != nil && !ok {
+		log.Warn("received unexpected data for the confirmation event", "batch", event.Batch)
 	}
 	failedEnvelopes := map[common.Hash]struct{}{}
 	for i := range envelopeErrors {
--- a/services/shhext/envelopes.go
+++ b/services/shhext/envelopes.go
@@ -2,6 +2,7 @@ package shhext
 
 import (
 	"context"
+	"errors"
 	"hash/fnv"
 	"sync"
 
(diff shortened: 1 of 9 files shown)


services/shhext/envelopes.go
```diff
@@ -193,7 +194,30 @@ func (m *EnvelopesMonitor) handleAcknowledgedBatch(event whisper.EnvelopeEvent)
 		log.Debug("batch is not found", "batch", event.Batch)
 	}
 	log.Debug("received a confirmation", "batch", event.Batch, "peer", event.Peer)
+	envelopeErrors, ok := event.Data.([]whisper.EnvelopeError)
+	if !ok {
+		log.Error("received unexpected data in the the confirmation event", "batch", event.Batch)
+	}
+	failedEnvelopes := map[common.Hash]struct{}{}
+	for i := range envelopeErrors {
```
pedropombeiro, line 202: Why not 

```suggestion
	for i, envelopeError := range envelopeErrors {
```

signal/events_shhext.go
```diff
@@ -14,6 +14,9 @@ const (
 	// to any peer
 	EventEnvelopeExpired = "envelope.expired"
 
+	// EventEnvelopeDiscarded is triggerd when envelope was discarded by a peer for some reason.
```
pedropombeiro:
```suggestion
	// EventEnvelopeDiscarded is triggered when the envelope was discarded by a peer for some reason.
```

pedropombeiro (approved):
lgtm, added a couple minor comments

services/shhext/envelopes.go
dshulyak: to avoid copying object. EnvelopeError is not a pointer, so full object will be copied if accessed this way

pedropombeiro: gotcha

dshulyak:
@mandrigin @adambabik can i get a second review here please?

adambabik (approved):

mandrigin (approved):

mandrigin (approved):

Issue #1325, opened 2018-12-17
Deploy docker image for status-go when doing a release


adambabik:
# Problem

Currently, we have a Jenkins job that creates status-go releases but this job does not pushes a docker image to Docker Hub.

# Implementation

When doing a release of status-go using Jenkins, a docker image should be a part of the release.

# Acceptance Criteria

1. Releasing status-go creates and pushes a Docker image to [Docker Hub](https://hub.docker.com/r/statusteam/status-go). It should push it with a released version as a tag and update `deployed-tests` tag.

cc @jakubgs

jakubgs:
Not stale, just forgot about it because it was not in the project.

jakubgs:
I consider this done, the builds run here:
https://ci.status.im/job/status-go/job/platforms/job/docker/

PR #1425, opened 2019-03-25
add Jenkinsfile.docker for pushing new images

jakubgs:
This provides a feature requested by  #1325 by adding `Jenkinsfile.docker` which by default builds a docker image and pushes it under a tag based of off contents of the `VERSION` file.
If the `RELEASE` checkbox is selected then it also pushes under the `deploy-test` tag to deploy the newly built image to the `eth.test` cluster.


adambabik (approved):

jakubgs:
@adambabik Do we want to run this for PR builds too?

adambabik:
No, let's do only releases.

jakubgs:
There is a `RELEASE` flag that isn't set during PR builds, which means only the image with tag from `VERSION` would be pushed. But it's fine if you don't want it then either, just clarifying.

PR #1426, opened 2019-03-26
hash typed data

gravityblast:
Added `HashTypedData` so that this hash can be signed with a keycard.

### DIFF
--- a/services/typeddata/hash.go
+++ b/services/typeddata/hash.go
@@ -19,8 +19,8 @@ var (
 	int256Type, _  = abi.NewType("int256", nil)
 )
 
-// Hash generates a hash of TypedData and verifies that chainId in the typed data matches currently selected chain.
-func Hash(typed TypedData, chain *big.Int) (common.Hash, error) {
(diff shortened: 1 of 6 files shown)


services/typeddata/hash.go
```diff
@@ -19,6 +19,15 @@ var (
 	int256Type, _  = abi.NewType("int256", nil)
 )
 
+// Hash generates a hash of TypedData and verifies that chainId in the typed data matches currently selected chain.
+func Hash(typed TypedData, chain *big.Int) (common.Hash, error) {
```
dshulyak: i would rename to something more verbose cause this method also performs some validation, and hashing usually doesn't

dshulyak (approved):

services/typeddata/hash.go
gravityblast: @dshulyak good idea, something like `ValidateAndHash`?

dshulyak: sounds fine :+1:

jakubgs:
I've merged #1427, please rebase this.

PR #1429, opened 2019-03-29
Feature/add get nodes from contract endpoint

cammellos:
Adds an endpoint to pull nodes from a contract, given a rpc endpoint & contract address.
I have also included the chaosmode update (only applied to a different branch), as otherwise status-react is not compatible.


dshulyak (approved):

api/backend.go
```diff
@@ -238,6 +242,32 @@ func (b *StatusBackend) CallRPC(inputJSON string) (string, error) {
...
+	if err != nil {
+		return response, err
+	}
+
+	contract, err := registry.NewNodes(gethcommon.HexToAddress(contractAddress), ethclient)
+	if err != nil {
+		return response, err
+	}
+
+	node, err := contract.Nodes(nil, big.NewInt(0))
```
dshulyak, line 262: it wasn't not obvious that you are fetching only the first one, isn't it better to request a certain number using rpc?

cammellos: oh, yes, I am just fetching one for testing as of now, I meant to fetch all but forgot, will add in the next PR!

PR #1433, opened 2019-04-01
Add history requests that allow to skip already received data

dshulyak:
After this API fully implemented it will allow to skip envelopes that client already received. The new API requires user to set which topics expected to be synced with a duration for each topic. For example, topic1=24 hours and topic2=7 days. After that application will generate and send requests for specified topics.

Concurrently we will be waiting for new envelopes from mail server. They will be arriving in the forward order, and we will be persisting every new timestamp on disk. In the case, if initially created request failed we will re-try it, but time range for the new request will be updated and request will progress. So, even if a connection is extremely flaky and slow we have a guarantee that history synchronization will progress.

Request between mail client and mail server will be performed with bloom filter. There will be no major changes in that client-server flow. The difference will be how that request is generated. At first when topics are received they will be grouped by the duration. So if we will have 2 different topics with same duration - we will make only one request with a bloom filter created from those topics. In case when duration is different, like 24 hours and 7 days - we will have to create different requests. I believe we do the same thing on status-react right now, but the implementation is somewhat different.

Also we will have to ensure that every request is finished before adding more topics to that request. This is required to ensure that all data will be downloaded for new topics without receiving duplicate data for old one. So every time we will receive request with new API - we will check if there are any pending requests that were terminated, exclude topics that are covered by such requests, and only then form new requests from all other topics. 

Example for a such condition: at first we requested messages starting from 0 to 10 and defined by bloom filter 1100. We know that the last message in this bloom filter was timestamped by 5, so if connection will be terminated/or app closed - we can safely restart request with same bloom filter and range from 5 to 10. At the same time new topic was added (doesn't matter if it is covered by existing bloom filter or not). We want to receive all data for this topic as well, from 0 to 10. As you see we cannot include this topic in the previous request, it will start from 5.

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -33,21 +33,22 @@ func NewMemoryDB() (*leveldb.DB, error) {
 	return leveldb.Open(storage.NewMemStorage(), nil)
 }
 
-// NewIsolatedDB returns instance that ensures isolated operations.
-func NewIsolatedDB(db *leveldb.DB, prefix storagePrefix) PrefixedLevelDB {
-	return PrefixedLevelDB{
+// NewDBNamespace returns instance that ensures isolated operations.
+func NewDBNamespace(db *leveldb.DB, prefix storagePrefix) LevelDBNamespace {
+	return LevelDBNamespace{
 		db:     db,
 		prefix: prefix,
 	}
 }
 
-// NewIsolatedMemoryDB wraps in memory leveldb with provided bucket.
-func NewIsolatedMemoryDB(prefix storagePrefix) (pdb PrefixedLevelDB, err error) {
+// NewMemoryDBNamespace wraps in memory leveldb with provided bucket.
+// Mostly used for tests. Including tests in other packages.
+func NewMemoryDBNamespace(prefix storagePrefix) (pdb LevelDBNamespace, err error) {
 	db, err := NewMemoryDB()
 	if err != nil {
 		return pdb, err
 	}
-	return NewIsolatedDB(db, prefix), nil
+	return NewDBNamespace(db, prefix), nil
 }
 
 // Key creates a DB key for a specified service with specified data
@@ -88,30 +89,30 @@ func Open(path string, opts *opt.Options) (db *leveldb.DB, err error) {
 	return
 }
 
-// PrefixedLevelDB database where all operations will be prefixed with a certain bucket.
-type PrefixedLevelDB struct {
+// LevelDBNamespace database where all operations will be prefixed with a certain bucket.
+type LevelDBNamespace struct {
 	db     *leveldb.DB
 	prefix storagePrefix
 }
 
-func (db PrefixedLevelDB) prefixedKey(key []byte) []byte {
+func (db LevelDBNamespace) prefixedKey(key []byte) []byte {
 	endkey := make([]byte, len(key)+1)
 	endkey[0] = byte(db.prefix)
 	copy(endkey[1:], key)
 	return endkey
 }
 
-func (db PrefixedLevelDB) Put(key, value []byte) error {
+func (db LevelDBNamespace) Put(key, value []byte) error {
 	return db.db.Put(db.prefixedKey(key), value, nil)
 }
 
-func (db PrefixedLevelDB) Get(key []byte) ([]byte, error) {
+func (db LevelDBNamespace) Get(key []byte) ([]byte, error) {
 	return db.db.Get(db.prefixedKey(key), nil)
 }
 
 // Range returns leveldb util.Range prefixed with a single byte.
 // If prefix is nil range will iterate over all records in a given bucket.
-func (db PrefixedLevelDB) Range(prefix, limit []byte) *util.Range {
+func (db LevelDBNamespace) Range(prefix, limit []byte) *util.Range {
 	if limit == nil {
 		return util.BytesPrefix(db.prefixedKey(prefix))
 	}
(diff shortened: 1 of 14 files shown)


dshulyak:
This change can be reviewed. I made necessary adjustments (e.g. https://github.com/status-im/whisper/pull/27) and now we will wait for confirmations from react to update history index once envelopes are processed and request is completed. I hope this is temporary solution, and can be changed after we will keep messages somewhere closer, and won't need multiple communication rounds.

db/db.go
```diff
@@ -20,8 +22,34 @@ const (
...
+	HistoryRequestBucket
 )
 
+// NewMemoryDB returns leveldb with memory backend prefixed with a bucket.
+func NewMemoryDB() (*leveldb.DB, error) {
+	return leveldb.Open(storage.NewMemStorage(), nil)
+}
+
+// NewIsolatedDB returns instance that ensures isolated operations.
+func NewIsolatedDB(db *leveldb.DB, prefix storagePrefix) PrefixedLevelDB {
```
mandrigin: maybe instead of "isolated" or "prefixed" we can use something that uses the word "namespace"...
like `NewDBNamespace(db, name)`.

db/history.go
```diff
@@ -0,0 +1,211 @@
...
+	"github.com/syndtr/goleveldb/leveldb/util"
+)
+
+var (
+	// ErrEmptyKey returned if key is not expected to be empty.
+	ErrEmptyKey = errors.New("TopicHistoryKey is empty")
+)
+
+// Interface is a common interface for DB operations.
+type Interface interface {
```
mandrigin: I would say that when called from outside `db.DB` looks better than `db.Interface`

db/history.go
```diff
@@ -0,0 +1,211 @@
...
+	if t.Current == zero && other.Current == zero {
+		return t.Duration == other.Duration
+	}
+	return t.Current == other.Current
+}
+
+// HistoryRequest is kept in the database while request is in the progress.
+// Stores necessary information to identify topics with associated ranges included in the request.
+type HistoryRequest struct {
+	db      Interface
```
mandrigin: why are there 2 DB instances, what is the difference, what are they for?

db/history.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+// Delete HistoryRequest from store and update every topic.
+func (req HistoryRequest) Delete() error {
+	for i := range req.histories {
+		th := &req.histories[i]
+		th.RequestID = common.Hash{}
+		th.Current = th.End
+		if err := th.Save(); err != nil {
+			return err
```
mandrigin, line 166: so, we don't delete it if any error occures?

also, I would prefer for this struct not to self-destruct and self-load (overwriting it's contents), but rather have these done on some other level, where `Load()` returns an instance of `*HistoryRequest`

services/shhext/api.go
```diff
@@ -493,6 +500,112 @@ func (api *PublicAPI) SendDirectMessage(ctx context.Context, msg chat.SendDirect
...
+	}
+
+	return hash, nil
+
+}
+
+// InitiateHistoryRequests is a stateful API for initiating history request for each topic.
+// Caller of this method needs to define only two parameters per each TopicRequest:
+// - Topic
+// - Duration before Now
```
mandrigin: this looks confusing... what does it mean exactly?

mandrigin (commented):
A few comments, but want to take a look at it more precisely one more time...

services/shhext/api.go
dshulyak: could be a bad wording :) what i meant by "Duration before Now" is for example: week before now, 24h before now

db/history.go
dshulyak: one instance per bucket. one for requests and another topic history index

db/db.go
dshulyak: i see Prefixed used for similar purposes in another project, and it is kinda consistent with "storagePrefix". not sure why i used Isolated in some cases

but i don't have anything against Namespace

services/shhext/api.go
mandrigin: I would call it something like `from` then.

db/history.go
mandrigin: so should we call them `requestsDB` and `topicHistoryIndexDB` then? :)

db/db.go
mandrigin: yeah, I know it is used every here and there, but I feel like a namespace is a more concise and good name if we want to isolate some data from some other data :)

db/history.go
dshulyak: yeah, so it is safe to not delete a request cause next time we will simply retry request with same topics and updated range

> also, I would prefer for this struct not to self-destruct and self-load (overwriting it's contents), but rather have these done on some other level, where Load() returns an instance of *HistoryRequest

you mean you prefer to update HistoryRequest outside of the Delete method? or something else?

mandrigin: it's like I would prefer `Load` and `Delete` to be performed by some struct outside this one

mandrigin: but I want too read this PR one more time before making any assumptions

0xc1c4da:
➤ Hester Bruikman commented:

mostly reliability

db/db.go
dshulyak: done

db/history.go
dshulyak: done

db/history.go
dshulyak: done

db/history.go
dshulyak: i moved topic updates when request is deleted to reactor. but attached histories are still loaded with a request. if it turns out to be a problem i can change that, but for now it makes easier to work with requests

services/shhext/api.go
dshulyak: changed wording

PR #1434, opened 2019-04-01
explicitly enable whisper for mailserver

jakubgs:
Looks like this didn't work properly without whisper enabled explicitly in the config.

### DIFF
--- a/_assets/compose/mailserver/Makefile
+++ b/_assets/compose/mailserver/Makefile
@@ -14,6 +14,8 @@ export API_MODULES    ?= eth,net,web3,admin
 export CONTAINER_NAME ?= status-go-mailserver
 export DATA_PATH      ?= /var/tmp/status-go-mail
 export FLEET_NAME     ?= eth.beta
+export REGISTER_TOPIC ?= whispermail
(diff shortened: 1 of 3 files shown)


mandrigin (approved):

jacqueswww (approved):

_assets/compose/mailserver/Makefile
```diff
@@ -14,6 +14,8 @@ export API_MODULES    ?= eth,net,web3,admin
 export CONTAINER_NAME ?= status-go-mailserver
 export DATA_PATH      ?= /var/tmp/status-go-mail
 export FLEET_NAME     ?= eth.beta
+export MAIL_TOPIC     ?= whispermail
```
adambabik: FYI this topic is for dynamic mail server discovery using our peers discovery mechanisms.

adambabik (approved):

_assets/compose/mailserver/Makefile
jakubgs: Ok, renamed variable to `REGISTER_TOPIC` and updated `README`.

PR #1435, opened 2019-04-04
Add sqlite store for whisper messages

dshulyak:
Implements MessageStore interface added in https://github.com/status-im/whisper/pull/25.

There is 2 implementations:
- storage based on encrypted sqlite
stores messages with envelope hash as a primary key, and whole received message as a json marshalled blob of bytes
- storage that sends confirmation that messages were persisted
since we need to know that message was persisted for correctness of the app i would prefer to send event from the wrapper that ensures that


sqlite/db.go
```diff
@@ -0,0 +1,111 @@
...
+// The reduced number of kdf iterations (for performance reasons) which is
+// currently used for derivation of the database key
+// https://github.com/status-im/status-go/pull/1343
+// https://notes.status.im/i8Y_l7ccTiOYq09HVgoFwA
+const kdfIterationsNumber = 3200
+
+const exportDB = "SELECT sqlcipher_export('newdb')"
+
+// EncryptDatabase encrypts an unencrypted database with key
+func EncryptDatabase(oldPath string, newPath string, key string) error {
```
cammellos: You probably don't need this `fn` I think it's only used by us as we had to migrate

dshulyak: right, i forgot to remove it

adambabik:
Can we link a user story to this PR or provide some context what problem we are solving? I hasn't been following closely what's going on in status-go repo but looking at any PR I should be able to get a full story why it is needed. I am unable to do so in this case.

dshulyak:
@adambabik more information https://github.com/status-im/status-go/pull/1433. basically message must be persisted before last seen timestamp in the index is updated. if we will update index before that and message will be removed from memory, we will never receive such message again

adambabik:
@dshulyak thanks, that makes it clear! :)

dshulyak:
It won't be needed. I will reuse confirmation that message was persisted that was used for deduplicator.

PR #1437, opened 2019-04-09
Deliver envelopes in forward order

dshulyak:
This is a requirement for https://github.com/status-im/status-go/pull/1433

Envelopes are delivered in order, from latest to newest. Returned cursor is a timestamp+envelope hash of the last delivered envelope. The difference is that we always iterate from lower to upper, and if cursor provided we will seek cursor and iterate from it till the end (upper bound). 

As i understand new flow will be incompatible with requests that will be in-flight at the moment of new mailserver deployment. Not sure how big is the issue, i would prefer not to add another query type for backward compatibility and just take the risk that some query will fail. What are people thoughts about that issue?

### DIFF
--- a/mailserver/db_key.go
+++ b/mailserver/db_key.go
@@ -53,13 +53,3 @@ func NewDBKeyFromBytes(b []byte) (*DBKey, error) {
 		hash:      common.BytesToHash(b[4:]),
 	}, nil
 }
-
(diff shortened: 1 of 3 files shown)


adambabik (approved):

dshulyak:
It hanged on review for a while, so i assume that people are fine that some query will fail (or not, depends on the amount of actual users during new deployment) :)

PR #1443, opened 2019-04-16
Add temporary sqlite storage for received whisper messages

dshulyak:
In order to guarantee that all messages were received and processed we are sending hash of the last envelope for matching on a client side. However this envelope is not always visible to a status-go user. On a client side all received envelopes are matched against topics with encryption keys, and on a mailserver side they are matched against bloom filter, built from those topics. And as a result client can be waiting for an envelope that was discarded on a lower level due to mismatch.

In order to resolve this issue we need to make that last envelope always visible to the application for processing. We obviously don't want to pass all envelopes to a user (status-react). The least invasive way is to persist messages when they are placed into whisper filters and notify users that message is added or discarded. This way even if last envelope is discarded we can be sure that all "required" envelopes are persisted and we can safely complete request.

But since we persisted messages before they were delivered to an app we need to guarantee that app will process them before they will disappear from storage. Cause next request won't include those messages. For this to work we won't remove messages from whisper filter when shh_getMessages is called. We will remove them only when user confirms that particular message is processed, e.g. we will reuse same mechanism currently used for deduplication.


sqlite/db.go
```diff
@@ -0,0 +1,46 @@
...
+// https://notes.status.im/i8Y_l7ccTiOYq09HVgoFwA
+const kdfIterationsNumber = 3200
+
+func openDB(path string, key string, kdfIter int) (*sql.DB, error) {
+	db, err := sql.Open("sqlite3", path)
+	if err != nil {
+		return nil, err
+	}
+
+	keyString := fmt.Sprintf("PRAGMA key = '%s'", key)
```
mandrigin, line 22: hmm, that looks like a perfect place for an SQL injection... is there any safer way to execute it? like maybe `db.Exec("PRAGMA key = ?", key)` or something like that

sqlite/db.go
```diff
@@ -0,0 +1,46 @@
...
+
+	if _, err = db.Exec("PRAGMA foreign_keys=ON"); err != nil {
+		return nil, err
+	}
+
+	if _, err = db.Exec(keyString); err != nil {
+		return nil, err
+	}
+
+	kdfString := fmt.Sprintf("PRAGMA kdf_iter = '%d'", kdfIter)
```
mandrigin, line 35: like in the previous one, is there any reason why we use `fmt.Sprintf` here instead of doing like here: https://github.com/status-im/status-go/pull/1443/files#diff-dad4e25e26c8c6c84ee658b3920ce82cR80

mandrigin (changes requested):
a few sql-related notes, looks good otherwise

adambabik:
What's the reason behind having a permanent storage vs in-memory? Why do we need to store decrypted message?

Isn't it enough to store just hashes of all p2p received envelopes and when the hash from the mail server response is received, we would just check whether the hash is in that list. If so, we confirm that the request succeeded otherwise not.

dshulyak:
> Isn't it enough to store just hashes of all p2p received envelopes and when the hash from the mail server response is received, we would just check whether the hash is in that list. If so, we confirm that the request succeeded otherwise not.

i dont think so. i covered my understanding in the issue description. message is persisted to guarantee that it will be processed later, we cannot update secondary index before that. let's say we updated secondary index and message was lost from memory being being processed, in this case will never receive it again, because we updated secondary index prematurely...

also i am not sure what persisting hashes will do. it is not very different from simply watching received envelopes

sqlite/db.go
dshulyak: i don't think that you can use arguments in pragma statements. somewhat hard to trace the whole flow due to c code use in the library. but i wasn't able to use ? or $1 in a statement

sqlite/db.go
mandrigin: okay, this one isn't very critical, but a comment about not being able to use ? in PRAGMA statements would be nice here and in `keyString`

sqlite/db.go
mandrigin: can we hash the key or make a hex-representation to make sure that we don't have any unsafe characters here? cc @cammellos

adambabik:
I read the description but it's still hard for me to imagine the whole complexity of the solution.

> In order to guarantee that all messages were received and processed we are sending hash of the last envelope for matching on a client side. However this envelope is not always visible to a status-go user. 

To be more precise, you talk about the hash in the mail server response: https://github.com/status-im/status-go/blob/v0.23.0-beta.10/mailserver/mailserver.go#L363, right?

The envelope with this hash might not be processed by the RPC caller but it always reaches the node. The mail server response is handled [here](https://github.com/status-im/whisper/blob/master/whisperv6/whisper.go#L1090-L1109) and each received envelope from the mail server is handled [here](https://github.com/status-im/whisper/blob/master/whisperv6/whisper.go#L984-L1029). Later, each envelope notifies [filter watchers and is pushed to the envelope feed](https://github.com/status-im/whisper/blob/master/whisperv6/whisper.go#L1253-L1259).

So, theoretically, we can subscribe to the feed and wait for the envelope with the hash from the mail server response to be received. Based on that we can decide whether the request was successful or timed out.

Of course, there is a problem here that filters processing is asynchronous, there are timeouts and loops so we would notify the client before it processes all messages but I think we can adjust to that. Another problem is that even though we might mark a request timed out, the envelopes might still pour in.

Do I miss anything in this reasoning?

dshulyak:
I considered another alternative. We expose additional signal when message is added/discarded in filters. Then status-react can complete request when it receives this signal with last envelope hash, and persisted all envelopes before this one. The order of messages is fine here, and there is no race issue as well. 

But then i cannot update history index in status-go without an additional confirmation from react that all envelopes were processed.

dshulyak:
> The envelope with this hash might not be processed by the RPC caller but it always reaches the node. The mail server response is handled here and each received envelope from the mail server is handled here. Later, each envelope notifies filter watchers and is pushed to the envelope feed.

unfortunately it is pushed in both cases, if it was delivered as a regular message or a history.

> So, theoretically, we can subscribe to the feed and wait for the envelope with the hash from the mail server response to be received. Based on that we can decide whether the request was successful or timed out.

right, i was going to listen for event that is posted even if message was discarded. but this event must be limited only to history delivery.

> Do I miss anything in this reasoning?

so the persistence on the go side comes in because we can only mark request as completed when messages are persisted. right? in case if we are persisting them only on react side, only react knows when all of them got persisted and we will need a confirmation that certain request is completed from react.

alternative to that is to split this flow in two. persist messages on go side, complete the request. and then safely deliver messages to react

adambabik:
> I considered another alternative. We expose additional signal when message is added/discarded in filters. Then status-react can complete request when it receives this signal with last envelope hash, and persisted all envelopes before this one. The order of messages is fine here, and there is no race issue as well.

I don't think it's true because of web3.js timeouts. Unless it will happen after some particular filter method is called which actually pushes the messages to the callback. However, it can still potentially be delayed between status-react and web3.js.

> But then i cannot update history index in status-go without an additional confirmation from react that all envelopes were processed.

What is history index?

dshulyak:
> I don't think it's true because of web3.js timeouts. Unless it will happen after some particular filter method is called which actually pushes the messages to the callback. However, it can still potentially be delayed between status-react and web3.js.

seems fine to me. after you receive signal with last envelope hash, you know that all other envelopes were added to a filter, which is guarded with mutex. so if later you do 1 call - you know that you can't miss any required messages. isn't right?

> What is history index?

timestamp that is updated every time when request completes.

dshulyak:
btw listening to a feed for last envelope is also a mess. you don't know which event will be received 1st, envelope.available with last envelope hash or request.complete. so you must track all this envelope.available somewhere and when request.complete is received check if last envelope hash was already received previously... the tricky part is when to clear that cache with envelope hashes.

it does seem that there is no satisfactory solution at all in this model...

adambabik:
> seems fine to me. after you receive signal with last envelope hash, you know that all other envelopes were added to a filter, which is guarded with mutex. so if later you do 1 call - you know that you can't miss any required messages. isn't right?

Right but between the time your receive a signal and actually persist messages a lot can happen. So here a confirmation from status-react is required and only then we can update the history index. But it seems like a pretty sane solution :)

So, the flow would look like this with [your alternative solution](https://github.com/status-im/status-go/pull/1443#issuecomment-483962457):
1. A mail server response is received with a last envelope hash,
2. We set up a signal in filters and if an envelope with the last hash is received, a signal is sent if message was discarded or processed,
3. status-react receives this signal and it should already have all historical messages,
4. status-react sends a confirmation for that signal to status-react that all messages are persisted,
5. status-go updates the history index.

All will be fine because we will have a connection between a particular mailserver request, response and signal so there shouldn't be any confusion.

dshulyak:
> Right but between the time your receive a signal and actually persist messages a lot can happen.

not really following.
1. signal received
2. peform a call for messages and persist them
3. complete request

all of that on single thread. if anything is interrupted then request  is not completed

> All will be fine because we will have a connection between a particular mailserver request, response and signal so there shouldn't be any confusion.

it doesn't seem fine to me at all. you can start receiving envelopes.available events from the next request.

i think i found a way how to ditch last envelope hash at all https://github.com/status-im/whisper/pull/27

sqlite/db.go
dshulyak: i don't think that i will end up using this storage, but we need to fix it in the original place if there are some concerns

mandrigin: what does "the original place" means? the PFS db?

mandrigin:
@dshulyak so we can close this PR then?

sqlite/db.go
dshulyak: yes, PFS db

dshulyak:
@mandrigin sure, i was holding it open because of the discussion

PR #1444, opened 2019-04-16
Expose vanilla ECDSA signing

### LINKED ISSUE #1438: Expose vanilla ECDSA signing
# Problem

eth_sign signs a message constructed like this: `\x19Ethereum Signed Message:\n" + message.length + message`. Swarm feeds require to sign only the message. Since the private key is only known to status-go it needs to be done there and therefore exposed through IPC.

# Implementation

Expose call similar to `.signMessage` but with `message` only

# Acceptance Criteria

The signature produced is valid and allows to update swarm feeds.

# Notes

This will be used for tribute to talk

mandrigin:
Closes #1438


api/backend.go
```diff
@@ -758,3 +758,20 @@ func (b *StatusBackend) UpdateMailservers(enodes []string) error {
 	}
 	return st.UpdateMailservers(nodes)
 }
+
+// SignHash exposes vanilla ECDSA signing for signing a message for Swarm
+func (b *StatusBackend) SignHash(hexEncodedHash string) (string, error) {
+	hash, err := hexutil.Decode(hexEncodedHash)
```
yenda, line 764: is it necessary?

yenda (approved):

api/backend.go
mandrigin: it transforms the string like `0xABABABABABABABABAB....` to an actuall array of bytes `[]byte{ 0xAB, 0xAB, 0xAB}` that we are are using for hashing.

the function itself receives and returns hex-strings: `0xABABABABAB` -> `0x12121212122121`

yenda: ok perfect then thanks for explaining

gravityblast (approved):

mandrigin:
@gravityblast @yenda can you test your code with this branch then? I'll merge if it works.

also, if you can provide some data for testcases like <hash> <private_key> <signature>, so I can add a few there, it would be nice...

gravityblast:
I'm still a little bit nervous about having this SignHash that can receive any transaction hash and sign it without showing a transaction to the user. I know there's a comment on the function though. Do you think it's fine? Or we can call it SignSwarmHash so that it's clear for future reader of the code?

yenda:
@gravityblast in status-react I'll call it `private-sign-hash`

yenda:
@mandrigin what is the current way to use a status-go branch in status-react?

mandrigin:
@yenda https://status.im/build_status/#Building-and-using-forks-of-status-go

account/accounts.go
```diff
@@ -466,3 +466,21 @@ func (m *Manager) unlockExtendedKey(address, password string) (*SelectedExtKey,
...
 }
+
+// SignHash returns the signature for specified hash for the selected chat account
+func (m *Manager) SignHash(hash []byte) ([]byte, error) {
+	errorFormat := "SignHash failed: %s"
+
+	chatAccount, err := m.SelectedChatAccount()
+
+	if err != nil {
+		return []byte{}, fmt.Errorf(errorFormat, err.Error())
```
adambabik:
```suggestion
		return nil, fmt.Errorf(errorFormat, err.Error())
```

account/accounts.go
```diff
@@ -466,3 +466,21 @@ func (m *Manager) unlockExtendedKey(address, password string) (*SelectedExtKey,
...
+	}
+
+	keyStore, err := m.geth.AccountKeyStore()
+	if err != nil {
+		return []byte{}, fmt.Errorf(errorFormat, "cannot get account keystore")
```
adambabik:
```suggestion
		return nil, fmt.Errorf(errorFormat, "cannot get account keystore")
```

account/accounts.go
```diff
@@ -466,3 +466,21 @@ func (m *Manager) unlockExtendedKey(address, password string) (*SelectedExtKey,
 
 	return selectedExtendedKey, nil
 }
+
+// SignHash returns the signature for specified hash for the selected chat account
+func (m *Manager) SignHash(hash []byte) ([]byte, error) {
+	errorFormat := "SignHash failed: %s"
+
+	chatAccount, err := m.SelectedChatAccount()
+
```
adambabik:
```suggestion
```

adambabik (commented):
Let's add unit tests for:
* `func (m *Manager) SignHash(hash []byte) ([]byte, error)`
* `func (b *StatusBackend) SignHash(hexEncodedHash string) (string, error)`

yenda:
@mandrigin if I call signHash with a digest I get this:

`(status/private-sign-hash "0x2e600e22f34fa9fca9f9f9369ea55f983eb740667b3e011a99e565f792a982b6" println)
nil
status-im.swarm.core> {"error":{"code":2,"message":"invalid character 'x' after top-level value"}}`

yenda:
@mandrigin it seems like it doesn't use the unlocked account. I get this:

```clojure
status-im.swarm.core> (status/private-sign-hash "0x2e600e22f34fa9fca9f9f9369ea55f983eb740667b3e011a99e565f792a982b6" println)
nil
status-im.swarm.core> {"error":"authentication needed: password or unlock"}
```

account/accounts.go
```diff
@@ -466,3 +466,21 @@ func (m *Manager) unlockExtendedKey(address, password string) (*SelectedExtKey,
...
+	if err != nil {
+		return []byte{}, fmt.Errorf(errorFormat, err.Error())
+	}
+
+	keyStore, err := m.geth.AccountKeyStore()
+	if err != nil {
+		return []byte{}, fmt.Errorf(errorFormat, "cannot get account keystore")
+	}
+
+	return keyStore.SignHash(chatAccount.GethAccount(), hash)
```
yenda: I think the key is not found here which returns ErrLocked
```golang
// SignHash calculates a ECDSA signature for the given hash. The produced
// signature is in the [R || S || V] format where V is 0 or 1.
func (ks *KeyStore) SignHash(a accounts.Account, hash []byte) ([]byte, error) {
	// Look up the key to sign with and abort if it cannot be found
	ks.mu.RLock()
	defer ks.mu.RUnlock()

	unlockedKey, found := ks.unlocked[a.Address]
	if !found {
		return nil, ErrLocked
	}
	// Sign the hash using plain ECDSA operations
	return crypto.Sign(hash, unlockedKey.PrivateKey)
}
```

yenda: ftr I am logged in in the app

yenda:
Also for unit tests I can provide data, digest and signature and what do you need, private key or 12 words for the account?

mandrigin:
I’d have both just in case

account/accounts.go
mandrigin: gotcha, I'll take a look today

mandrigin:
@yenda okay, I see that logging in doesn't actually unlock a chat account, it stays locked and the password is required to sign this hash... or we need to actually unlock the chat account on login.
@adambabik wdyt?

yenda:
@mandrigin we should unlock it, how does whisper work anyway if the account is not unlocked? you need the private key to decrypt the 1-1 messages afaik?

account/accounts.go
adambabik: I would not use `geth.AccountKeyStore()` at all. `chatAccount, err := m.SelectedChatAccount()` should be enough to get the private key (`chatAccount.AccountKey.PrivateKey`) and we can call `crypto.Sign(hash, unlockedKey.PrivateKey)` directly.

adambabik:
@mandrigin replied in https://github.com/status-im/status-go/pull/1444#discussion_r276208049. 

@yenda we don't use geth's key store in `Login` binding for example, only to create a key. We can do the same here.

mandrigin:
@adambabik gotcha, I'll take a look now...

mandrigin:
@yenda try now. and I'm still waiting for the test data.

account/accounts.go
mandrigin: yeah, I did exactly that.

yenda:
@mandrigin here are some test data, sorry I don't know how to get the 12 words

```
private key:  0xfacadefacadefacadefacadefacadefacadefacadefacadefacadefacadefaca
hash-digest: 0xa4735de5193362fe856416000105cdfa6ce56265607311cebae93b26e5adf438
signature: 0x176c971bae188c663614fc535ac9dbf62871dfeaadb38645809a510d28b3c4b0245415d5547c1b27f7cfea3341564f9c6981421144d3606b455346be69bd078c01

**Comment by mandrigin** [comment]
@yenda it passes the test on my machine...

**Comment by yenda** [comment]
@mandrigin works in status-react as well thank you very much!

**Comment by mandrigin** [comment]
@adambabik if you feel like more tests needed, let me know, otherwise, I think we can merge it.

**Comment by mandrigin** [comment]
ping @adambabik 🏓

File: VERSION [hunk 6]
```diff
@@ -1 +1 @@
-0.23.0-beta.10
+0.24.0-beta.10
```
adambabik: `0.23.0-beta.11` or `0.24.0-beta.0`

api/backend.go
```diff
@@ -758,3 +758,28 @@ func (b *StatusBackend) UpdateMailservers(enodes []string) error {
...
+
```
adambabik: So many spaces between each call and error handling :D

api/backend_test.go
```diff
@@ -386,6 +386,39 @@ func TestStartStopMultipleTimes(t *testing.T) {
...
+	backend := NewStatusBackend()
+	config, err := utils.MakeTestNodeConfig(params.StatusChainNetworkID)
+	require.NoError(t, err)
+
+	require.NoError(t, backend.StartNode(config))
+	defer func() {
+		require.NoError(t, backend.StopNode())
+	}()
+
+	testSignature := func(chatPrivKeyHex, hashHex, expectedSignatureHex string) {
```
adambabik: What's the reason of this? Usually a table driven approach is expected.

adambabik (approved):

api/backend.go
mandrigin: I love whitespaces!

api/backend_test.go
mandrigin: It is kind of table driven, but there is only one item to test... but it is easy to extend it later.

api/backend.go
mandrigin: adressed

api/backend_test.go
mandrigin: adressed

VERSION
mandrigin: fixed

mandrigin:
I have addressed the comments, will merge as soon as the CI passes

mandrigin:
@yenda you'll need STATUS_GO_VERSION set to 0.24.0-beta.0 to have this binding with status-go release in your code.

I'll do a release if I can :-D

PR #1445, opened 2019-04-17
Removing discovery-proxy from sources

dshulyak:
It was going to be used as proxy from discv5 to rendezvous for les nodes.
We are not using it at the moment, and not clear when we will be using it
and if we are going to use it all.

Additionally it has some memory leaks that need to be fixed, so it is better
to remove it for now and restore once/if we will need it again.


services/shhext/mailrequests.go
```diff
@@ -81,6 +81,7 @@ func (m *MailRequestMonitor) handleEnvelopeEvents() {
 // confirmation handler or removes hash from MailRequestMonitor
 func (m *MailRequestMonitor) handleEvent(event whisper.EnvelopeEvent) {
 	handlers := map[whisper.EventType]func(whisper.EnvelopeEvent){
+		whisper.EventEnvelopeAvailable:          m.handleEnvelopeAvailable,
```
adambabik: Seems like a leftover from some other work.

jakubgs (approved):
Thx!

PR #1446, opened 2019-04-23
Safely request previous history

dshulyak:
Includes https://github.com/status-im/status-go/pull/1433

Ranges now can be updated when new requests are created. Example:
1. In 1st request client asked for 24h of messages for topic 1. 
2. Request succeeded and all that data now stored locally.
3. In 2nd request client increases requested range to 48h, we don't want to request same 24h again,
so we split this requested range into two: 48h-24h and 24h - 0h and perform only the first one.
4. Once 48h-24h downloaded history index will be merged.

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -6,8 +6,10 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/syndtr/goleveldb/leveldb"
 	"github.com/syndtr/goleveldb/leveldb/errors"
+	"github.com/syndtr/goleveldb/leveldb/iterator"
 	"github.com/syndtr/goleveldb/leveldb/opt"
 	"github.com/syndtr/goleveldb/leveldb/storage"
+	"github.com/syndtr/goleveldb/leveldb/util"
 )
 
 type storagePrefix byte
@@ -20,7 +22,34 @@ const (
 	DeduplicatorCache
 	// MailserversCache is a list of mail servers provided by users.
 	MailserversCache
+	// TopicHistoryBucket isolated bucket for storing history metadata.
+	TopicHistoryBucket
+	// HistoryRequestBucket isolated bucket for storing list of pending requests.
+	HistoryRequestBucket
 )
+
+// NewMemoryDB returns leveldb with memory backend prefixed with a bucket.
+func NewMemoryDB() (*leveldb.DB, error) {
+	return leveldb.Open(storage.NewMemStorage(), nil)
+}
+
+// NewDBNamespace returns instance that ensures isolated operations.
+func NewDBNamespace(db *leveldb.DB, prefix storagePrefix) LevelDBNamespace {
+	return LevelDBNamespace{
+		db:     db,
+		prefix: prefix,
+	}
+}
+
+// NewMemoryDBNamespace wraps in memory leveldb with provided bucket.
+// Mostly used for tests. Including tests in other packages.
+func NewMemoryDBNamespace(prefix storagePrefix) (pdb LevelDBNamespace, err error) {
+	db, err := NewMemoryDB()
+	if err != nil {
+		return pdb, err
+	}
+	return NewDBNamespace(db, prefix), nil
+}
 
 // Key creates a DB key for a specified service with specified data
 func Key(prefix storagePrefix, data ...[]byte) []byte {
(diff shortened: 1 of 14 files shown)


db/db.go
```diff
@@ -20,8 +22,34 @@ const (
...
+	HistoryRequestBucket
 )
 
+// NewMemoryDB returns leveldb with memory backend prefixed with a bucket.
+func NewMemoryDB() (*leveldb.DB, error) {
+	return leveldb.Open(storage.NewMemStorage(), nil)
+}
+
+// NewIsolatedDB returns instance that ensures isolated operations.
+func NewIsolatedDB(db *leveldb.DB, prefix storagePrefix) PrefixedLevelDB {
```
mandrigin: isolated -> prefixed

db/db.go
```diff
@@ -20,8 +22,34 @@ const (
...
+	return PrefixedLevelDB{
+		db:     db,
+		prefix: prefix,
+	}
+}
+
+// NewIsolatedMemoryDB wraps in memory leveldb with provided bucket.
+func NewIsolatedMemoryDB(prefix storagePrefix) (pdb PrefixedLevelDB, err error) {
```
mandrigin: isolated -> prefixed

db/history.go
```diff
@@ -0,0 +1,209 @@
...
+	"github.com/syndtr/goleveldb/leveldb/util"
+)
+
+var (
+	// ErrEmptyKey returned if key is not expected to be empty.
+	ErrEmptyKey = errors.New("TopicHistoryKey is empty")
+)
+
+// Interface is a common interface for DB operations.
+type Interface interface {
```
mandrigin: db.Interface -> db.DB

services/shhext/api.go
```diff
@@ -418,6 +419,12 @@ func (api *PublicAPI) GetNewFilterMessages(filterID string) ([]dedup.Deduplicate
 // ConfirmMessagesProcessed is a method to confirm that messages was consumed by
 // the client side.
 func (api *PublicAPI) ConfirmMessagesProcessed(messages []*whisper.Message) error {
+	for _, msg := range messages {
+		err := api.service.historyUpdates.UpdateTopicHistory(msg.Topic, time.Unix(int64(msg.Timestamp), 0))
```
mandrigin: maybe we can just pass `msg` as a parameter there?

services/shhext/api.go
```diff
@@ -493,6 +500,112 @@ func (api *PublicAPI) SendDirectMessage(ctx context.Context, msg chat.SendDirect
 	return api.Post(ctx, whisperMessage)
 }
 
+// InitiateHistoryRequest type for initiating history requests from a peer.
+type InitiateHistoryRequest struct {
```
mandrigin: this mid-file struct definition looks a bit weird, especially taking into account that then you continue defining methods for another struct... that confuses when reading the file.

services/shhext/api.go
```diff
@@ -493,6 +500,112 @@ func (api *PublicAPI) SendDirectMessage(ctx context.Context, msg chat.SendDirect
...
+	return hash, nil
+
+}
+
+// InitiateHistoryRequests is a stateful API for initiating history request for each topic.
+// Caller of this method needs to define only two parameters per each TopicRequest:
+// - Topic
+// - Duration before Now
+// After that status-go will guarantee that request for this topic and date will be performed.
+func (api *PublicAPI) InitiateHistoryRequests(request InitiateHistoryRequest) ([]hexutil.Bytes, error) {
```
mandrigin: `InitiateHistoryRequestParams`, maybe?

db/db.go
dshulyak: comments are not fixed in this branch. and this PR is based on 1433

mandrigin (changes requested):

services/shhext/history.go
```diff
@@ -114,10 +137,69 @@ func (reactor *HistoryUpdateReactor) CreateRequests(topicRequests []TopicRequest
...
+	}
+	filtered = append(filtered,
+		GroupHistoriesByRequestTimespan(reactor.store, adjusted)...)
 	return RenewRequests(filtered, reactor.timeSource()), nil
 }
 
+// for every history that is not included in any request check if there are other ranges with such topic in db
+// if so check if they can be merged
+// if not adjust second part so that End of it will be equal to First of anothe
+func adjustRequestedRanges(store db.HistoryStore, histories []db.TopicHistory) ([]db.TopicHistory, error) {
```
mandrigin: can you add a diagram what does it do exactly?...

or maybe a separate unit-test for that function to actually show how it is used, because this algorithm is a bit hard to understand out of context.

also, tests for this particular function should be state-less and help with improving the code further

dshulyak: there are tests for every case in this method. all of them start with TestAdjustHistories

mandrigin (approved):

dshulyak:
@mandrigin added explanation for adjustRequestedRanges (in unit tests). let me know if some logic is still not clear

mandrigin (approved):

PR #1448, opened 2019-04-24
Bump version to 0.24.0-beta.1

dshulyak:


dshulyak:
@adambabik @mandrigin i noticed that release branches are not used anymore. Should we update doc and maybe say that it is optional to cherry-pick changes to release branch?

mandrigin (approved):

mandrigin:
I agree. @adambabik wdyt?

adambabik:
> @adambabik @mandrigin i noticed that release branches are not used anymore. Should we update doc and maybe say that it is optional to cherry-pick changes to release branch?

I think we need to automate and enforce it actually. Release branches are useful only when we work on more than one minor version. So, until this is done we can think of release branches as optional.

PR #1449, opened 2019-04-24
add a short doc about config.json

jakubgs:
During CUD some people complained that we didn't have docs about configuring `status-go`.

For easier reading: [add-config-doc/config](https://github.com/status-im/status-go/tree/add-config-doc/config)

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -1,0 +1,97 @@
+# Introduction
+
+This document describes the available options in the JSON config for `status-go`.
+
+The structure of the JSON config is defined in the [`params/config.go`](/params/config.go) file, which also contains detailed comments on meaning of each option. The `NodeConfig` struct defines the general configuration keys at the __root__ of the JSON file.
+
+If the descriptions of any options are too vague feel free to [open an issue](https://github.com/status-im/status-go/issues/new).
+
+Example config files can be viewed in the [`config/cli`](/config/cli) folder.
+
+# Important Sections
+
+The JSON config is separated into several sections. The most important ones are listed below.
+
+## `NodeConfig`
+
+The root of the JSON configuration.
+
+An example of most important settings would include:
+```json
+{
+    "NetworkID": 1,
+    "DataDir": "/tmp/status-go-data",
+    "NodeKey": "123qwe123qwe123qwe123",
+    "Rendezvous": true,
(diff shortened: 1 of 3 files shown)


config/README.md
```diff
@@ -0,0 +1,112 @@
...
+{
+    "WhisperConfig": {
+        "Enabled": true,
+        "EnableMailServer": true,
+        "LightClient": false,
+        "MailServerPassword": "status-offline-inbox"
+    }
+}
+```
+The `MailServerPassword` is used for symmetric encryption of history requests.
```
dshulyak, line 68: we can mention that this password will be used when mail server is added in status app.
@cammellos maybe you have a link to documentation that explains that?

dshulyak (approved):

config/README.md
```diff
@@ -0,0 +1,112 @@
+# Introduction
+
+This document describes the available options in the JSON config for `status-go`.
+
+The structure of the JSON config is defined in the [`params/config.go`](/params/config.go) file, which also contains details comments on meaning of each option.
```
adambabik: I would add that `NodeConfig` is the main struct one should start looking from.

config/README.md
```diff
@@ -0,0 +1,112 @@
...
+    "LogFile": "/var/log/status-go.log",
+    "LogLevel": "INFO",
+    "LogMaxSize": 200,
+    "LogMaxBackups": 5,
+    "LogCompressRotated": true
+}
+```
+Valid `LogLevel` settings are: `ERROR`, `WARN`, `INFO`, `DEBUG`, `TRACE`
+
+## `LightEthConfig`
```
adambabik: I would remove it because we don't support LES.

config/README.md
adambabik: I would add that `status-offline-inbox` is the password used by Status mail servers listed on https://fleets.status.im.

adambabik (approved):

config/README.md
jakubgs: Good idea.

config/README.md
jakubgs: Will do.

config/README.md
jakubgs: Less noise is always nice.

PR #1450, opened 2019-04-25
fix HashMessage to decode hex only if needed

gravityblast:
Update `HashMessage` to behave like `web3.eth.accounts.hashMessage`.
Like in the js implementation, it now decodes an hex string only if it starts with `0x` and if it's a valid hex.

### DIFF
--- a/api/utils.go
+++ b/api/utils.go
@@ -1,7 +1,9 @@
 package api
 
 import (
+	"encoding/hex"
 	"fmt"
+	"strings"
 
 	"github.com/ethereum/go-ethereum/crypto"
 )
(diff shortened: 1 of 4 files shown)


api/utils.go
```diff
@@ -21,7 +23,14 @@ func RunAsync(f func() error) <-chan error {
 // The hash is calulcated as
 //   keccak256("\x19Ethereum Signed Message:\n"${message length}${message}).
 // This gives context to the signed message and prevents signing of transactions.
-func HashMessage(data []byte) []byte {
+func HashMessage(message string) []byte {
+	data := []byte(message)
```
adambabik: Do we need this conversion? We do `%s` in `fmt.Sprintf` anyway. I guess we can do `data := message` and `data = string(value)`. Finally, we convert it again to a byte slice `crypto.Keccak256([]byte(msg))`.

Maybe:
```go
b := NewBufferString("\x19Ethereum Signed Message:\n")
_, err = b.WriteString(strconv.Itoa(len(data)))
_, err = b.Write(data) // if data is still []byte otherwise WriteString
return crypto.Keccak256(b.Bytes())
```

We should also have a unit test with a known hash for a given message.

gravityblast:
thank you @adambabik , I updated the code following the comments!

api/utils.go
```diff
@@ -21,7 +24,36 @@ func RunAsync(f func() error) <-chan error {
...
+		}
+		if _, err := buf.WriteString(message); err != nil {
+			return nil, err
+		}
+	}
+
+	return crypto.Keccak256(buf.Bytes()), nil
+}
+
+func decodeHexStrict(s string) (bool, []byte) {
```
adambabik: Usually, it's `([]byte, bool)`, like getting an element from a map.

adambabik (approved):
Great number of tests!

api/utils.go
gravityblast: oh you are right, I'm gonna change it!

PR #1452, opened 2019-04-29
Feature/network incentivisation

cammellos:
Add interactions with the contract & posting pings on whisper.

If started with an incentivisation config it will:
- try to register with the contract
- post whisper envelopes on a specific channel, that can be read by status client advertising its address
- check with the contract which nodes are available and collects pings from them
- vote based on the results of the pings

Still in progress and much to do, but I would like to merge as we go.


services/incentivisation/contract.go
```diff
@@ -0,0 +1,78 @@
...
+	tx, err := c.Vote(opts, joinNodes, removeNodes)
+	if err != nil {
+		return nil, err
+	}
+
+	for {
+		receipt, _ := c.client.TransactionReceipt(context.TODO(), tx.Hash())
+		if receipt != nil {
+			if receipt.Status == 0 {
+				return nil, errors.New("Receipt invalid")
```
pedropombeiro: Shouldn't it be
```suggestion
				return nil, errors.New("Invalid receipt")
```

services/incentivisation/service.go
```diff
@@ -0,0 +1,518 @@
...
+	privateKey      *ecdsa.PrivateKey
+	log             log.Logger
+	// The first round we will not be voting, as we might have incomplete data
+	initialSession uint64
+	// The current session
+	currentSession uint64
+	whisperPings   map[string][]uint32
+}
+
+// New returns a new Service
```
pedropombeiro:
```suggestion
// New returns a new incentivization Service
```

services/incentivisation/service.go
```diff
@@ -0,0 +1,518 @@
...
+}
+
+// Protocols returns a new protocols list. In this case, there are none.
+func (s *Service) Protocols() []p2p.Protocol {
+	return []p2p.Protocol{}
+}
+
+// APIs returns a list of new APIs.
+func (s *Service) APIs() []rpc.API {
+	s.log.Info("APIS CALLED")
```
pedropombeiro: Why all-caps here? Is it a remnant?

services/incentivisation/service.go
```diff
@@ -0,0 +1,518 @@
...
+	_, err = s.register()
+	if err != nil {
+		s.log.Error("error querying contract", "registered", err)
+		return err
+	}
+	return nil
+}
+
+// newSession checks if we are in a new session and updates the session if so
+func (s *Service) newSession() (bool, error) {
```
pedropombeiro: maybe `ensureSession` or `getSession` would better describe what the method does?

services/incentivisation/service.go
```diff
@@ -0,0 +1,518 @@
...
+	}
+
+	err = s.fetchMessages()
+	if err != nil {
+		return err
+	}
+
+	err = s.checkRegistered()
+	if err != nil {
+		s.log.Error("Could not check registered", "err", err)
```
pedropombeiro:
```suggestion
		s.log.Error("Could not check if node is registered with the contract", "err", err)
```

services/incentivisation/service.go
```diff
@@ -0,0 +1,518 @@
...
+			case <-s.quit:
+				s.ticker.Stop()
+				return
+			}
+		}
+	}()
+}
+
+func (s *Service) Start(server *p2p.Server) error {
+	s.log.Info("Incentivisation service started 2", "address", s.addressString(), "publickey", s.publicKeyString())
```
pedropombeiro: What is the 2 in `Incentivisation service started 2`?

services/incentivisation/service.go
```diff
@@ -0,0 +1,518 @@
...
+
+	response, err := s.contract.RegisterNode(&bind.TransactOpts{
+		GasLimit: gasLimit,
+		From:     auth.From,
+		Signer:   auth.Signer,
+	}, s.publicKeyBytes(), ip, s.config.Port)
+	if err != nil {
+		return false, err
+	}
+	s.log.Info("resposne from registered", "response", response)
```
pedropombeiro:
```suggestion
	s.log.Info("response from registered", "response", response)
```

services/incentivisation/service.go
```diff
@@ -0,0 +1,518 @@
...
+}
+
+func (s *Service) FetchEnodes() error {
+	one := big.NewInt(1)
+
+	activeNodeCount, err := s.contract.ActiveNodeCount(nil)
+	if err != nil {
+		return err
+	}
+	s.log.Info("fetched node count", "count", activeNodeCount)
```
pedropombeiro:
```suggestion
	s.log.Info("fetched active node count", "count", activeNodeCount)
```

services/incentivisation/service.go
```diff
@@ -0,0 +1,518 @@
...
+		if node.PublicKeyString() != s.publicKeyString() {
+			s.nodes[node.PublicKeyString()] = node
+		}
+	}
+
+	inactiveNodeCount, err := s.contract.InactiveNodeCount(nil)
+	if err != nil {
+		return err
+	}
+	s.log.Info("fetched node count", "count", inactiveNodeCount)
```
pedropombeiro:
```suggestion
	s.log.Info("fetched inactive node count", "count", inactiveNodeCount)
```

pedropombeiro (approved):
Lgtm, added some minor suggestions

services/incentivisation/service.go
cammellos: yes, will remove, thanks!

services/incentivisation/service.go
cammellos: testing docker images :) I will remove it

PR #1453, opened 2019-04-30
Replace request ID when same request is restarted

dshulyak:
Request ID must be replaced if same request is restarted.

### DIFF
--- a/services/shhext/history.go
+++ b/services/shhext/history.go
@@ -353,8 +353,8 @@ func (filter BloomFilterOption) ToMessagesRequestPayload() ([]byte, error) {
 		Bloom: filter.Filter,
 		// Client must tell the MailServer if it supports batch responses.
 		// This can be removed in the future.
-		Batch: true,
-		Limit: 10000,
+		//Batch: true,
+		Limit: 100000,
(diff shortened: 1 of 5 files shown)


services/shhext/history.go
```diff
@@ -353,8 +353,8 @@ func (filter BloomFilterOption) ToMessagesRequestPayload() ([]byte, error) {
 		Bloom: filter.Filter,
 		// Client must tell the MailServer if it supports batch responses.
 		// This can be removed in the future.
-		Batch: true,
-		Limit: 10000,
+		//Batch: true,
```
mandrigin: why is this commented out and limit is changed?

mandrigin (changes requested):
it looks like there is a leftover, but in general everything looks good

services/shhext/history.go
dshulyak: you are right, thanks. i was experimenting with it and thought that i discarded changes

pedropombeiro (approved):
lgtm once Igor's remark is addressed

mandrigin (approved):

PR #1454, opened 2019-05-01
Execute writes atomically only after request was processed without errors

dshulyak:
In order to avoid corrupted state due to unpredictable interrupt or storage error all writes will be performed atomically. It also makes it easier to reason about request flow.

I reused leveldb.Batch to enforce that. It doesn't provide any read isolation and doesn't allow to read its own uncommitted writes, so it shouldn't be used as a regular sql transaction.

### DIFF
--- a/db/tx.go
+++ b/db/tx.go
@@ -1,0 +1,40 @@
+package db
+
+import (
+	"github.com/syndtr/goleveldb/leveldb"
+	"github.com/syndtr/goleveldb/leveldb/iterator"
+	"github.com/syndtr/goleveldb/leveldb/util"
+)
+
+// LevelDBTx doesn't provide any read isolation. It allows to commit all writes atomically (put/delete).
+type LevelDBTx struct {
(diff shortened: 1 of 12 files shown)


db/tx.go
```diff
@@ -0,0 +1,40 @@
+package db
+
+import (
+	"github.com/syndtr/goleveldb/leveldb"
+	"github.com/syndtr/goleveldb/leveldb/iterator"
+	"github.com/syndtr/goleveldb/leveldb/util"
+)
+
+// LevelDBTx doesn't provide any read isolation. It allows to commit all writes atomically (put/delete).
```
pedropombeiro:
```suggestion
// LevelDBTx doesn't provide any read isolation. It allows committing all writes atomically (put/delete).
```

pedropombeiro (approved):
LGTM. Good call on leveraging `Context` for time source.

mandrigin:
the previous PR was merged, please rebase this one.

PR #1459, opened 2019-05-08
Publish rlp.RawValue instead of envelope

cammellos:
As part of a performance profiling of mailserver we noticed that most of
the resources on a query are spend decoding the whisper envelope.

This PR changes the way we store envelopes encoding the Topic into the
database key, so we can check that and we are able to publish the
envelope rawValue if it matches.

The change is backward compatible as only newly added envelopes will
have the new key, while old ones will have to be unmarshaled.

This is the first performance optimization, seems like that even if we go for postgres, we might want to allow users to use leveldb as well, as it's easier to setup (no setup required), so this is worth doing in any case.
We will only be able to see any improvement after 24hours of data have been saved (i.e there's no migration).

Currently deployed on `eth.test` if anyone wants to make some tests


ghost:
### Pull Request Checklist

dshulyak (approved):

mailserver/mailserver.go
```diff
@@ -475,16 +475,17 @@ func (s *WMailServer) exceedsPeerRequests(peer []byte) bool {
...
 
-	ku = NewDBKey(upper+1, emptyHash)
-	kl = NewDBKey(lower, emptyHash)
+	ku = NewDBKey(upper+1, emptyTopic, emptyHash)
+	kl = NewDBKey(lower, emptyTopic, emptyHash)
 
 	i := s.db.NewIterator(&util.Range{Start: kl.Bytes(), Limit: ku.Bytes()}, nil)
 	// seek to the end as we want to return envelopes in a descending order
-	if len(cursor) == DBKeyLength {
+	if len(cursor) == DBKeyLength || len(cursor) == DBKeyLength-whisper.TopicLength {
```
mandrigin: I'd leave old length as `DBKeyLengthLegacy` constant with a comment.

mailserver/mailserver.go
```diff
@@ -522,29 +523,41 @@ func (s *WMailServer) processRequestInBundles(
...
+				log.Error("[mailserver:processRequestInBundles] failed to decode RLP",
+					"err", decodeErr,
+					"requestID", requestID)
+				continue
+			}
+			envelopeBloom = envelope.Bloom()
 
-		if !whisper.BloomFilterMatch(bloom, envelope.Bloom()) {
+		} else {
+			envelopeBloom = whisper.TopicToBloom(whisper.BytesToTopic(key[timestampLength+common.HashLength:]))
```
mandrigin: can we have a `Topic()` method in `DBKey` struct to make this part a bit more readable?

mailserver/mailserver.go
```diff
@@ -522,29 +523,41 @@ func (s *WMailServer) processRequestInBundles(
...
-		var envelope whisper.Envelope
 
-		decodeErr := rlp.DecodeBytes(iter.Value(), &envelope)
-		if decodeErr != nil {
-			log.Error("[mailserver:processRequestInBundles] failed to decode RLP",
-				"err", decodeErr,
-				"requestID", requestID)
-			continue
-		}
+		var envelopeBloom []byte
```
mandrigin: I would move this declaration right before `if len(key).....`

mailserver/mailserver.go
```diff
@@ -522,29 +523,41 @@ func (s *WMailServer) processRequestInBundles(
...
+		rawValue := make([]byte, len(iter.Value()))
+		copy(rawValue, iter.Value())
+
+		key := iter.Key()
+
+		// Old key, we extract the topic from the envelope
+		if len(key) != DBKeyLength {
+			var envelope whisper.Envelope
```
mandrigin: maybe extract a method like `envelopeBloom = extractTopicFromEncodedEnvelope(rawValue)`

mailserver/mailserver.go
```diff
@@ -522,29 +523,41 @@ func (s *WMailServer) processRequestInBundles(
...
-		if !whisper.BloomFilterMatch(bloom, envelope.Bloom()) {
+		} else {
+			envelopeBloom = whisper.TopicToBloom(whisper.BytesToTopic(key[timestampLength+common.HashLength:]))
+		}
+		if !whisper.BloomFilterMatch(bloom, envelopeBloom) {
 			continue
 		}
 
-		lastEnvelopeHash = envelope.Hash()
+		lastEnvelopeHash = common.BytesToHash(key[timestampLength : common.HashLength+timestampLength])
```
mandrigin: again, a method like `DBKey.EnvelopeHash()` to hide this bytes array indexing.

mailserver/mailserver.go
```diff
@@ -608,16 +621,16 @@ func (s *WMailServer) processRequestInBundles(
 	return nextCursor, lastEnvelopeHash
 }
 
-func (s *WMailServer) sendEnvelopes(peer *whisper.Peer, envelopes []*whisper.Envelope, batch bool) error {
+func (s *WMailServer) sendEnvelopes(peer *whisper.Peer, envelopes []rlp.RawValue, batch bool) error {
```
mandrigin: `sendRawEnvelopes`?

services/shhext/api.go
```diff
@@ -750,7 +750,8 @@ func makeMessagesRequestPayload(r MessagesRequest) ([]byte, error) {
 	if err != nil {
 		return nil, fmt.Errorf("invalid cursor: %v", err)
 	}
-	if len(cursor) > 0 && len(cursor) != mailserver.DBKeyLength {
+	// We check against the old length for backward compatibility
+	if len(cursor) > 0 && len(cursor) != mailserver.DBKeyLength && len(cursor) != mailserver.DBKeyLength-whisper.TopicLength {
```
mandrigin: again, having a constant for the old `DBKeySize` will be useful here

mandrigin (changes requested):
In general, the code looks good. I'd love to see a few changes though, to improve readability.

cammellos:
@mandrigin ready to be reviewed again, thanks!

mandrigin (approved):
lgtm now

PR #1462, opened 2019-05-10
Allow multiple db implementations

cammellos:
This commit creates an interface to use with the db so that we can
abstract what kind of db we use, therefore allowing us to chose db based
on config.

It also adds support for postgres, which can be enabled in the config by specifying:

```
{
WhisperConfig:
{ DatabaseConfig:
{PGConfig: { Enabled: true, URI: "postgres://username:password@ip-address:port"}
}
}
```

One host is already running this code:

`mail-01.do-ams3.eth.test`

The DB is still filling up (started listening for envelopes on 2019/05/14 around 1 PM UTC), but it can already be used for testing (performance are not indicative as it's not a full database).

status: ready


mailserver/leveldb_impl.go
```diff
@@ -0,0 +1,113 @@
+package mailserver
```
mandrigin: I would call it `mailserver_db_leveldb.go` and `mailserver_db_postgres.go` instead.

jakubgs:
Is the schema created on first connection, or do I need to do some extra step?

Also, please update `config/README.md`.

cammellos:
@jakubgs I will update the README, thanks
the scheme is created on first connection, but I need the database set up for me (just an empty db + username & password)

jakubgs:
Any specific Postgres version you are expecting?

cammellos:
@jakubgs should work fine with any > 8.1

jakubgs:
Why the requirement for SSL? If the DB is local it's kinda pointless.
```
error="failed to register MailServer: open DB: pq: SSL is not enabled on the server"
```

cammellos:
?sslmode=disable in pg uri

jakubgs:
I've deployed `status-go` with PostgreSQL on `mail-01.do-ams3.eth.test`:
https://github.com/status-im/infra-eth-cluster/commit/1fbd6cf36e848dbdb88c665720b811cb0690843f

The connection worked, but I got no log message saying that it did. It might be a good idea to add an `INFO` level message saying it connected to the PostgreSQL database.
And I can see the schema has been created:
```
admin@mail-01.do-ams3.eth.test:~ % d exec -it statusd-mail-db psql -U whisper whisper
psql (9.6.13)
Type "help" for help.

whisper=# \dt
              List of relations
 Schema |       Name        | Type  |  Owner  
--------+-------------------+-------+---------
 public | envelopes         | table | whisper
 public | schema_migrations | table | whisper
(2 rows)
```

jakubgs:
Are you fine with just one host? Tell me if you want me to deploy this to the whole `eth.test` cluster.

cammellos:
just one host is fine for now, much thank yous! if i need to update the
code, it s the same process as updating the test fleet?

jakubgs:
I've used a specific tag for `mail-01.do-ams3.eth.test`:
https://github.com/status-im/infra-eth-cluster/commit/1fbd6cf36e848dbdb88c665720b811cb0690843f#diff-203160ce66994e01f63e2f353d098f36R3
```
statusd_tag: 'deploy-postgres'
```
So you can update it by doing:
```
make docker-image DOCKER_IMAGE_CUSTOM_TAG=deploy-postgres
docker push statusteam/status-go:deploy-postgres
```

cammellos:
goddamn you are efficient, thanks!

jakubgs:
No problem, poke me if you need anything else.

mailserver/mailserver.go
```diff
@@ -111,11 +90,22 @@ func (s *WMailServer) Init(shh *whisper.Whisper, config *params.WhisperConfig) e
 
 	// Open database in the last step in order not to init with error
 	// and leave the database open by accident.
-	database, err := db.Open(config.DataDir, nil)
-	if err != nil {
-		return fmt.Errorf("open DB: %s", err)
+	if config.DatabaseConfig.PGConfig.Enabled {
+		log.Info("Connecting to postgres database")
+		database, err := NewPostgresImpl(config)
```
mandrigin: I would prefer something like `NewPostgresDB` and `NewLevelDB`, no need to explicitly state that these are implementations, I think it is obvious anyway.

mailserver/mailserver_db.go
```diff
@@ -0,0 +1,36 @@
+package mailserver
+
+import (
+	whisper "github.com/status-im/whisper/whisperv6"
+	"time"
+)
+
+// dbImpl is an interface to abstract interactions with the db so that the mailserver
+// is agnostic to the underlaying technology used
+type dbImpl interface {
```
mandrigin: I think just `mailserver.db` interface name is okay already. Especially since the interface knows nothing about the implementation, idk if we should have `impl` as a part of the name.

mailserver/mailserver_db_leveldb.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/status-im/status-go/params"
+	whisper "github.com/status-im/whisper/whisperv6"
+	"github.com/syndtr/goleveldb/leveldb"
+	"github.com/syndtr/goleveldb/leveldb/errors"
+	"github.com/syndtr/goleveldb/leveldb/iterator"
+	"github.com/syndtr/goleveldb/leveldb/util"
+	"time"
+)
+
+type LevelDBImpl struct {
```
mandrigin: I'd say `type levelDB`: no need to state that that is an implementation and also no need to make it public, especially if the `db` interface isn't.

mailserver/mailserver_db_postgres.go
```diff
@@ -0,0 +1,224 @@
...
+
+	instance := &PostgresImpl{db: db}
+	if err := instance.setup(); err != nil {
+		return nil, err
+	}
+
+	return instance, nil
+}
+
+type PostgresImpl struct {
```
mandrigin: `type postgresdb`

mandrigin (changes requested):
overall the direction is right, but I think it needs a tiny bit more polishing here and there

cammellos:
@mandrigin ready to be reviewed, I kept the classes as public because the linter would complain, but otherwise I implemented all the changes.
Thanks!

mandrigin (approved):

PR #1466, opened 2019-05-17
Feature/topic negotiation

cammellos:
Add topic negotiation
    
This commit add topic negotiation to the protocol.
    
On receiving a message from a client with version >= 1, we will generate
a shared key using Diffie-Hellman. We will record also which
installationID has sent us a message.
    
This key will be passed back to the above layer, which will then use to
start listening to a whisper topic (the `chat` namespace has no
knowledge of whisper).
    
When sending a message to a set of InstallationIDs, we check whether we
have agreed on a topic with all of them, and if so, we will send on this separate topic, otherwise we fallback on discovery.
    
This change is backward compatible with other peers, as long as there is no downgrade of the app on the other side, but it changes some of the APIs so, but it's a breaking change for status-react, I'd rather focus on the status-react PR instead of making it backward compatible, but happy to do so if necessary (it's should be pretty simple, restore one endpoint and set ProtocolVersion to 0).

A few changes:
* Factored out the DB in a separate namespace as now it is
    being used by multiple services (TopicService and EncryptionService).
* Factored out multidevice management in a separate namespace
* Moved all the test to test the whole protoocl rather than just the encryption service
* moved all the filter management in status-go

I still would like to move some parts of status-react to status-go, but I might push them in separate PRs.
Also don't mind the logs please, used as I am still testing, I will remove them.
  
Sorry for the large PR, I have been carried away, happy to split it if necessary.

status: ready


services/shhext/api.go
```diff
@@ -427,13 +421,11 @@ func (api *PublicAPI) GetNewFilterMessages(filterID string) ([]dedup.Deduplicate
 
 	dedupMessages := api.service.deduplicator.Deduplicate(msgs)
 
-	if api.service.pfsEnabled {
-		// Attempt to decrypt message, otherwise leave unchanged
-		for _, dedupMessage := range dedupMessages {
+	// Attempt to decrypt message, otherwise leave unchanged
+	for _, dedupMessage := range dedupMessages {
```
mandrigin, line 426: we will most likely get rid of deduplication service when moving messaging to status-go side... jfyi.

services/shhext/api.go
```diff
@@ -474,99 +466,12 @@ func (api *PublicAPI) ConfirmMessagesProcessedByID(messageIDs [][]byte) error {
...
-		whisperMessage.PublicKey = destination
-		return &whisperMessage, nil
-	}
-
-	api.log.Debug("Sending on old discovery topic")
-	whisperMessage.Topic = whisperutils.DiscoveryTopicBytes
-	whisperMessage.PublicKey = destination
-
-	return &whisperMessage, nil
+	return api.service.SendPublicMessage(ctx, msg)
```
mandrigin: so you are planning to plug these method in to status-react already? that's good :-D

services/shhext/publisher/service.go
```diff
@@ -0,0 +1,446 @@
...
+		os.Remove(v3Path)
+	}
+
+	// Fix IOS not encrypting database
+	if err := appDB.EncryptDatabase(v3Path, v4Path, encKey); err != nil {
+		os.Remove(v3Path)
+		os.Remove(v4Path)
+	}
+
+	// Desktop was passing a network dependent directory, which meant that
```
mandrigin, line 120: shouldn't that be fixed in Desktop instead? looks off to me

mandrigin: cc @vkjr

services/shhext/publisher/service.go
```diff
@@ -0,0 +1,446 @@
...
+				}
+			case <-s.quit:
+				s.ticker.Stop()
+				return
+			}
+		}
+	}()
+}
+
+func (s *Service) perform() error {
```
mandrigin: hmm, why is this method doing nothing?

services/shhext/service.go
```diff
@@ -50,6 +37,7 @@ type EnvelopeEventsHandler interface {
 
 // Service is a service that provides some additional Whisper API.
 type Service struct {
+	*publisher.Service
```
mandrigin, line 40: Good usage of composition!

mandrigin (changes requested):
Looks good on a first glance. One thing I might be interested in is making this API as high-level as possible, to simplify integration with moving protocol to status-go.

services/shhext/publisher/service.go
cammellos: I am moving still some parts from status-react to status-go, filling it in as of now

services/shhext/publisher/service.go
cammellos: The code has just been moved, it was there before this PR, keeping in desktop would have been sensible, but decided otherwise in order to have a single place to handle migrations, eventually we can get rid of this (0.9.0 had this defect)

cammellos:
The API exported currently are:
`SendDirectMessage`

`SendPublicMessage`

`LoadFilters` (i.e `LoadChats` or `InitChats`)

`LoadFilter` (i.e `AddChat` )

`RemoveFilter` (i.e `RemoveChat`)

We use filters as currently all the chat code is still in status react, but no sym-key/addfilter will be done in status react.

Let me know if you'd like to make some changes

mandrigin:
@cammellos ah, okay, I thought much more is exposed.

api/backend.go
```diff
@@ -615,104 +614,49 @@ func appendIf(condition bool, services []gethnode.ServiceConstructor, service ge
...
 
-// ProcessContactCode process and adds the someone else's bundle
-func (b *StatusBackend) ProcessContactCode(contactCode string) error {
-	selectedChatAccount, err := b.AccountManager().SelectedChatAccount()
-	if err != nil {
-		return err
-	}
-
+// LoadFilters loads filter on sshext
+func (b *StatusBackend) LoadFilters(chats []*filter.Chat) ([]*filter.Chat, error) {
```
mandrigin: hmm, can it go through json-rpc instead? at least, in my attemts, I envision all the protocol stuff called through json-rpc, so bringing these together will make it more consistent.

mandrigin: also, no need to implement it in 2 places (mobile + library)

services/shhext/api.go
```diff
@@ -469,64 +466,22 @@ func (api *PublicAPI) ConfirmMessagesProcessedByID(messageIDs [][]byte) error {
...
-		return nil, ErrPFSNotEnabled
-	}
-	// To be completely agnostic from whisper we should not be using whisper to store the key
-	privateKey, err := api.service.w.GetPrivateKey(msg.Sig)
-	if err != nil {
-		return nil, err
-	}
-
-	publicKey, err := crypto.UnmarshalPubkey(msg.PubKey)
+	message, err := api.service.SendDirectMessage(msg.Sig, msg.PubKey, msg.DH, msg.Payload)
```
mandrigin: hmm, why `SendDirectMessage` and then `Post` it, it looks weird, looks like both methods are sending stuff to a network.

maybe `InitializeDirectMessage` + `Post` instead?

services/shhext/chat/multidevice/service.go
```diff
@@ -0,0 +1,93 @@
...
+		persistence: persistence,
+	}
+}
+
+type Service struct {
+	persistence Persistence
+	config      *Config
+}
+
+type IdentityAndIDPair [2]string
```
mandrigin: I would use a structure instead, unless there is a specific reason to not to.

services/shhext/chat/protocol.go
```diff
@@ -39,50 +57,110 @@ func (p *ProtocolService) addBundleAndMarshal(myIdentityKey *ecdsa.PrivateKey, m
...
+
+	for _, installation := range p.Installations[1:] {
+		if installation.Version < version {
+			version = installation.Version
+		}
+	}
+	return version
+}
+
+func (p *ProtocolMessageSpec) PartitionedTopic() bool {
```
mandrigin: to futureproof I would do an enum here instead of a boolean. `PARTITION_TOPIC_NO_SUPPORT`, `PARTITION_TOPIC_V1`, etc.

mandrigin (changes requested):
a few more smaller comments

cammellos:
@mandrigin addressed feeback, thanks!

services/shhext/api.go
```diff
@@ -469,64 +467,22 @@ func (api *PublicAPI) ConfirmMessagesProcessedByID(messageIDs [][]byte) error {
...
-	}
-
-	// This is transport layer agnostic
-	protocolMessage, err := api.service.protocol.BuildPublicMessage(privateKey, msg.Payload)
-	if err != nil {
-		return nil, err
-	}
-
-	symKeyID, err := api.service.w.AddSymKeyFromPassword(msg.Chat)
+	message, err := api.service.InitializePublicMessage(msg.Sig, msg.Chat, msg.Payload, false)
```
```suggestion
	message, err := api.service.CreatePublicMessage(msg.Sig, msg.Chat, msg.Payload, false)
```

services/shhext/api.go
```diff
@@ -469,64 +467,22 @@ func (api *PublicAPI) ConfirmMessagesProcessedByID(messageIDs [][]byte) error {
...
 }
 
 // SendDirectMessage sends a 1:1 chat message to the underlying transport
 func (api *PublicAPI) SendDirectMessage(ctx context.Context, msg chat.SendDirectMessageRPC) (hexutil.Bytes, error) {
-	if !api.service.pfsEnabled {
-		return nil, ErrPFSNotEnabled
-	}
-	// To be completely agnostic from whisper we should not be using whisper to store the key
-	privateKey, err := api.service.w.GetPrivateKey(msg.Sig)
+	message, err := api.service.InitializeDirectMessage(msg.Sig, msg.PubKey, msg.DH, msg.Payload)
```
```suggestion
	message, err := api.service.CreateDirectMessage(msg.Sig, msg.PubKey, msg.DH, msg.Payload)
```

Maybe it's just me but initialization suggests that the message is already created but that's not a case here.

services/shhext/chat/db/db.go
```diff
@@ -0,0 +1,244 @@
+package db
+
+import (
+	"database/sql"
+	"fmt"
+	"os"
+
+	sqlite "github.com/mutecomm/go-sqlcipher" // We require go sqlcipher that overrides default implementation
+	"github.com/status-im/migrate"
```
adambabik: I recently updated our fork of `github.com/status-im/migrate` to `v4` because it did not process my migrations properly. `v4` works fine.

services/shhext/chat/db/db.go
```diff
@@ -0,0 +1,244 @@
...
+// The default number of kdf iterations in sqlcipher (from version 3.0.0)
+// https://github.com/sqlcipher/sqlcipher/blob/fda4c68bb474da7e955be07a2b807bda1bb19bd2/CHANGELOG.md#300---2013-11-05
+// https://www.zetetic.net/sqlcipher/sqlcipher-api/#kdf_iter
+const defaultKdfIterationsNumber = 64000
+
+// The reduced number of kdf iterations (for performance reasons) which is
+// currently used for derivation of the database key
+// https://github.com/status-im/status-go/pull/1343
+// https://notes.status.im/i8Y_l7ccTiOYq09HVgoFwA
+const KdfIterationsNumber = 3200
```
adambabik, line 26: Do we use it anywhere?

services/shhext/chat/multidevice/service.go
```diff
@@ -0,0 +1,96 @@
...
+import (
+	"crypto/ecdsa"
+	"fmt"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/status-im/status-go/services/shhext/chat/protobuf"
+)
+
+type Installation struct {
+	ID      string
+	Version uint32
```
adambabik, line 14: Does it correspond to `ProtocolVersion` in `Config`? It would also be nice to see a comment explaining what `Installation` is.

services/shhext/chat/multidevice/service.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+type Config struct {
+	MaxInstallations int
+	ProtocolVersion  uint32
+	InstallationID   string
```
adambabik, line 20: Does it correspond to `Installation` and `ID` above?

services/shhext/chat/multidevice/sql_lite_persistence.go
```diff
@@ -0,0 +1,168 @@
...
+		latestVersion := installation.Version
+
+		err = stmt.QueryRow(identity, installation.ID).Scan(&oldEnabled, &oldVersion)
+		if err != nil && err != sql.ErrNoRows {
+			return err
+		}
+
+		// We update timestamp if present without changing enabled, only if this is a new bundle
+		// and we set the version to the latest we ever saw
+		if err != sql.ErrNoRows {
```
adambabik: Is this correct? We handle it already above `if err != nil && err != sql.ErrNoRows {`.

services/shhext/chat/multidevice/sql_lite_persistence.go
```diff
@@ -0,0 +1,168 @@
...
+			)
+			if err != nil {
+				return err
+			}
+			defer stmt.Close()
+		}
+
+	}
+
+	if err := tx.Commit(); err != nil {
```
adambabik, line 132: If the method returns earlier due to an error, for example, this tx will never be closed. It should be put at the top with `defer`.

services/shhext/chat/multidevice/sql_lite_persistence_test.go
```diff
@@ -0,0 +1,241 @@
+package multidevice
+
+import (
+	"database/sql"
+	"os"
+	"testing"
+
+	appDB "github.com/status-im/status-go/services/shhext/chat/db"
```
adambabik:
```suggestion
	chatDB "github.com/status-im/status-go/services/shhext/chat/db"
```

services/shhext/chat/protocol.go
```diff
@@ -5,30 +5,55 @@ import (
...
+
+const ProtocolVersion = 1
+const sharedSecretNegotiationVersion = 1
+const partitionedTopicMinVersion = 1
+
+type PartitionTopic int
+
+const (
+	PartitionTopicNoSupport PartitionTopic = iota
+	PartitionTopicV1        PartitionTopic = iota
```
adambabik: Are you sure it's correct? Also these consts are not really clear to me. `PartitionTopicNoSupport` means using the discovery topic?

services/shhext/chat/protocol.go
```diff
@@ -39,50 +64,112 @@ func (p *ProtocolService) addBundleAndMarshal(myIdentityKey *ecdsa.PrivateKey, m
...
+	// SharedSecret is a shared secret established among the installations
+	SharedSecret []byte
 }
 
-// BuildDirectMessage marshals a 1:1 chat message given the user identity private key, the recipient's public key, and a payload
-func (p *ProtocolService) BuildDirectMessage(myIdentityKey *ecdsa.PrivateKey, publicKey *ecdsa.PublicKey, payload []byte) ([]byte, error) {
+func (p *ProtocolMessageSpec) MinVersion() uint32 {
+
+	if len(p.Installations) == 0 {
+		return 0
```
adambabik: Magic const. Should we define `defaultMinVersion`?

services/shhext/chat/sharedsecret/persistence.go
```diff
@@ -0,0 +1,125 @@
...
+
+import (
+	"database/sql"
+	"strings"
+)
+
+type PersistenceService interface {
+	Add(identity []byte, secret []byte, installationID string) error
+	Get(identity []byte, installationIDs []string) (*Response, error)
+	All() ([][][]byte, error)
```
adambabik, line 15: If that's ok, we should put a comment explaining what is in each level of this multidimensional slice.

services/shhext/filter/service.go
```diff
@@ -0,0 +1,455 @@
...
+		return nil, err
+	}
+	return s.load(myKey, chat)
+}
+
+func ContactCodeTopic(identity string) string {
+	return "0x" + identity + "-contact-code"
+}
+
+// Get returns a negotiated filter given an identity
```
adambabik: But it returns `Chat` not a filter.

services/shhext/filter/service.go
```diff
@@ -0,0 +1,455 @@
...
+func (s *Service) GetNegotiated(identity *ecdsa.PublicKey) *Chat {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+
+	return s.chats[negotiatedID(identity)]
+}
+
+// GetByID returns a filter by chatID
```
adambabik: Returns `Chat`.

services/shhext/chat/db/db.go
cammellos: yes, it is used in this namespace, in one of the migrations and when we open the db

services/shhext/chat/multidevice/service.go
cammellos: Yes, I'll add a comment

services/shhext/chat/protocol.go
cammellos: `PartitionTopicNoSupport` means using the discovery topic, yes, what do you mean by correct in this case?

adambabik: Actually it works fine but I expected a different result. 

This is more idiomatic:
```go
const (
    PartitionTopicNoSupport PartitionTopic = iota 
    PartitionTopicV1 
)
```

services/shhext/chat/multidevice/service.go
cammellos: Yes

services/shhext/chat/multidevice/sql_lite_persistence.go
cammellos: Yes, it could be more explicit in this case, it would be:
`if err == nil` or if we swap the `if` and `else` `if err == sql.ErrNoRows` , probably it's easier, I will do that.

cammellos:
@adambabik addressed feedback, thanks. 
It's ready to be reviewed again, the only thing I haven't done is to update migrate, I will do that in a subsequent PR if that's ok.

cammellos:
@mandrigin @adambabik This is ready to be reviewed again, I am almost done with the PR on status-react side (still testing), but this should be ready to be merged if you are happy.

mandrigin (approved):
looks good to me

PR #1467, opened 2019-05-23
Download transfers starting from latest block header

dshulyak:
This change implements service that downloads transfers for specified ethereum address. Transfers include ethereum and erc20 transfers.

When user logs in we will start a reactor that will be periodically polling block chain for new blocks. Reactor will work transparently with any kind of node, it can be either local node or remote. Once we will receive header that wasn't previously seen we will check if there are any transfers attached to that block.

For ethereum we will compare balances between previous block and new one. If there is a difference we will download body of the block and iterate over transactions comparing recipient or signer of the transaction. If any will match such transaction will qualify as a transfer.

For erc20 we will use logs with erc20 Transfer events. Two separate queries will be sent for inbound and outbound events.

On pow chain reorgs are common thus we will have to check if new block parent matches to our previous known block. If they don't we will iterate backwards until we find common ancestor. All transfers that are attached to removed blocks will be also removed from database.

Transfers will be stored in sqlite database. It is not encrypted but path will be based on hex of the wallet address. 

API for client interaction described https://github.com/status-im/status-go/tree/71b6129b03b222e54880efe3132066a25e6c5e5e/services/wallet


dshulyak:
@yenda i am not sure if you got my messages in status. so, you can use ios/android builds from this branch for development. this version doesn't download historical transfers, but downloads new. 

described basic API in https://github.com/status-im/status-go/tree/71b6129b03b222e54880efe3132066a25e6c5e5e/services/wallet

api/backend.go
```diff
@@ -539,20 +555,45 @@ func (b *StatusBackend) SelectAccount(walletAddress, chatAddress, password strin
...
 	}
+	err = b.startWallet()
+	if err != nil {
+		return err
+	}
+	return nil
+}
 
+func (b *StatusBackend) startWallet() error {
+	if b.statusNode.Config().WalletConfig.Enabled {
```
adambabik: Inverse? 
```suggestion
	if !b.statusNode.Config().WalletConfig.Enabled {
	  return nil
	}
```

services/wallet/api.go
```diff
@@ -0,0 +1,30 @@
...
+type API struct {
+	s *Service
+}
+
+// GetTransfers returns transfers in range of blocks. If `end` is nil all transfers from `start` will be returned.
+// TODO(dshulyak) benchmark loading many transfers from database. We can avoid json unmarshal/marshal json if we will
+// return header, tx and receipt as a raw json.
+func (api *API) GetTransfers(ctx context.Context, start, end *big.Int) ([]Transfer, error) {
+	log.Debug("call to get transfers", "start", start, "end", end)
+	if api.s.db == nil {
```
adambabik, line 30: That's ok but even better if we can hide the presence of `db` from the `API` completely. So, maybe `api.s.GetTransfers(start, end)` and then serialize it if needed?

api/backend.go
```diff
@@ -539,20 +555,45 @@ func (b *StatusBackend) SelectAccount(walletAddress, chatAddress, password strin
...
+		wallet, err := b.statusNode.WalletService()
+		if err != nil {
+			return err
+		}
+		account, err := b.accountManager.SelectedWalletAccount()
+		if err != nil {
+			return err
+		}
+		path := path.Join(b.statusNode.Config().DataDir, fmt.Sprintf("wallet-%x.sql", account.Address))
+		err = wallet.StartReactor(path,
```
pedropombeiro: Why the use of `reactor` as a name instead of `monitor` (just seems a much more common name for this type of functionality)?

services/wallet/README.md
```diff
@@ -0,0 +1,121 @@
+Wallet API
+==========
+
+Wallet service starts a loop that watches for new transfers (eth and erc20).
+To correctly start a service two values need to be changed in the config:
```
pedropombeiro:
```suggestion
To correctly start the service, two values need to be changed in the config:
```

services/wallet/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+2. And expose wallet API with APIModules
+
+```
+{
+  APIModules: "eth,net,web3,peer,wallet",
+}
+```
+
+Wallet API
```
pedropombeiro: Again Wallet API section?

services/wallet/README.md
```diff
@@ -0,0 +1,121 @@
...
+      "gasUsed": "0x5208"
+    }
+  }
+]
+```
+
+Signals
+-------
+
+Two signals will are emmited:
```
pedropombeiro:
```suggestion
Two signals will are emitted:
```

services/wallet/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+1. Signal `newblock`
```
pedropombeiro:
```suggestion
1. `newblock` signal
```

services/wallet/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+Emmited when transfers from new block were added to the database. In this case block number if the number of this new block.
```
pedropombeiro:
```suggestion
Emitted when transfers from a new block were added to the database. In this case the block number is the number of this new block.
```

services/wallet/README.md
```diff
@@ -0,0 +1,121 @@
...
+{
+  "type": "wallet",
+  "event": {
+    "Type": "newblock",
+    "BlockNumber": 10
+  }
+}
+```
+
+2. Signal `reorg`.
```
pedropombeiro:
```suggestion
2. `reorg` signal
```

services/wallet/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+Emmited when part of blocks were removed. Starting from a given block number all transfers were removed.
```
pedropombeiro:
```suggestion
Emitted when part of blocks were removed. Starting from a given block number, all transfers were removed.
```

services/wallet/database.go
```diff
@@ -0,0 +1,254 @@
...
+	for _, t := range transfers {
+		_, err = insert.Exec(t.Transaction.Hash(), t.Header.Hash(), &JSONBlob{t.Transaction}, &JSONBlob{t.Receipt}, t.Type)
+		if err != nil {
+			return err
+		}
+	}
+	return err
+}
+
+// GetTransfers load transfers transfer betweeen two blocks.
```
pedropombeiro, line 160:
```suggestion
// GetTransfers returns transfers that happened between two blocks.
```

services/wallet/downloader.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// ETHTransferDownloader downloads regular eth transfers.
+type ETHTransferDownloader struct {
+	client  *ethclient.Client
+	address common.Address
+	signer  types.Signer
+}
+
+// GetTransfers checks is balance was changed between two blocks. And if so it downloads transfers
```
pedropombeiro:
```suggestion
// GetTransfers checks if the balance was changed between two blocks. if so, it downloads transfers
```

services/wallet/events.go
```diff
@@ -0,0 +1,19 @@
...
+
+import "math/big"
+
+// EventType type for event types.
+type EventType string
+
+const (
+	// EventNewBlock emitted when new block was added to the same canonical chan.
+	EventNewBlock = "newblock"
+	// EventReorg emitted when canonical chain was changed. In this case BlockNumber will be an earliest added block.
```
pedropombeiro:
```suggestion
	// EventReorg emitted when canonical chain was changed. In this case, BlockNumber will be a block added earlier.
```

services/wallet/service.go
```diff
@@ -0,0 +1,84 @@
...
+	reactor *Reactor
+	signals *SignalsTransmitter
+}
+
+// Start signals transmitter.
+func (s *Service) Start(*p2p.Server) error {
+	return s.signals.Start()
+}
+
+// StartReactor separately cause it requires known ethereum address, that will become available only after login.
```
pedropombeiro:
```suggestion
// StartReactor separately because it requires known ethereum address, which will become available only after login.
```

services/wallet/service.go
```diff
@@ -0,0 +1,84 @@
...
+		return nil
+	}
+	s.reactor.Stop()
+	if s.db == nil {
+		return nil
+	}
+	return s.db.Close()
+}
+
+// Stop reactor, signals  transmitter and close db.
```
pedropombeiro:
```suggestion
// Stop reactor, signals transmitter and close db.
```

services/wallet/transmitter.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/ethereum/go-ethereum/event"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/status-im/status-go/signal"
+)
+
+type publisher interface {
+	Subscribe(interface{}) event.Subscription
+}
+
+// SignalsTransmitter transmits received events as a wallet signals.
```
pedropombeiro:
```suggestion
// SignalsTransmitter transmits received events as wallet signals.
```

pedropombeiro (commented):
wow, lots of work there! Added some minor comments.

api/backend.go
```diff
@@ -474,7 +475,19 @@ func (b *StatusBackend) Logout() error {
 	default:
 		return err
 	}
-
+	if b.statusNode.Config().WalletConfig.Enabled {
+		wallet, err := b.statusNode.WalletService()
+		switch err {
+		case node.ErrServiceUnknown:
+		case nil:
+			err = wallet.StopReactor()
```
adambabik, line 483:
```suggestion
			return wallet.StopReactor()
```

api/backend.go
```diff
@@ -497,7 +510,10 @@ func (b *StatusBackend) reSelectAccount() error {
 	default:
 		return err
 	}
-
+	err = b.startWallet()
```
adambabik:
```suggestion
	return b.startWallet()
```

api/backend.go
```diff
@@ -539,20 +555,45 @@ func (b *StatusBackend) SelectAccount(walletAddress, chatAddress, password strin
 			return err
 		}
 	}
+	err = b.startWallet()
```
adambabik:
```suggestion
	return b.startWallet()
```

api/backend.go
adambabik:
```suggestion
		return wallet.StartReactor(path,
```

services/wallet/README.md
adambabik: We should add information what is an expected action after each signal. Whether a call to `wallet_getTransfers` should happen and with what params from received signal.

services/wallet/api.go
```diff
@@ -0,0 +1,30 @@
...
+		return nil, errors.New("wallet service is not initialized")
+	}
+	rst, err := api.s.db.GetTransfers(start, end)
```
adambabik:
```suggestion
	return api.s.db.GetTransfers(start, end)
```

services/wallet/database.go
```diff
@@ -0,0 +1,254 @@
...
+func (i *SQLBigInt) Value() (driver.Value, error) {
+	if !(*big.Int)(i).IsInt64() {
+		return nil, errors.New("not at int64")
+	}
+	return (*big.Int)(i).Int64(), nil
+}
+
+// JSONBlob type for marshaling/unmarshaling inner type to json.
+type JSONBlob struct {
+	Interface interface{}
```
adambabik:
```suggestion
	data interface{}
```

Because underlying value shouldn't be really exposed and `data` because it better describes the value.

services/wallet/database.go
```diff
@@ -0,0 +1,254 @@
...
+
+	insert, err = tx.Prepare("INSERT INTO transfers(hash, blk_hash, tx, receipt, type) VALUES (?, ?, ?, ?, ?)")
+	if err != nil {
+		return err
+	}
+	delete, err = tx.Prepare("DELETE FROM blocks WHERE hash = ?")
+	if err != nil {
+		return err
+	}
+	blocks, err = tx.Prepare("INSERT INTO blocks(hash,number,header) VALUES (?, ?, ?)")
```
adambabik:
```suggestion
	blocks, err = tx.Prepare("INSERT INTO blocks(hash, number, header) VALUES (?, ?, ?)")
```

services/wallet/database.go
```diff
@@ -0,0 +1,254 @@
...
+			&JSONBlob{transfer.Transaction}, &JSONBlob{transfer.Receipt})
+		if err != nil {
+			return nil, err
+		}
+		rst = append(rst, transfer)
+	}
+	return
+}
+
+// SaveHeader stores single header.
```
adambabik:
```suggestion
// SaveHeader stores a single header.
```

services/wallet/database.go
```diff
@@ -0,0 +1,254 @@
...
+func (db *Database) SaveHeader(header *types.Header) error {
+	_, err := db.db.Exec("INSERT INTO blocks(number, hash, header) VALUES (?, ?, ?)", (*SQLBigInt)(header.Number), header.Hash(), &JSONBlob{header})
+	return err
+}
+
+// SaveHeaders atomically stores list of headers.
```
adambabik:
```suggestion
// SaveHeaders stores a list of headers atomically.
```

services/wallet/database.go
```diff
@@ -0,0 +1,254 @@
...
+	)
+	tx, err = db.db.Begin()
+	if err != nil {
+		return
+	}
+	insert, err = tx.Prepare("INSERT INTO blocks(number,hash,header) VALUES (?,?,?)")
+	if err != nil {
+		return
+	}
+	defer func() {
```
adambabik, line 191: It is repeated a few times:
```suggestion
	defer func() { err = closeTx(tx) }() 
```

services/wallet/database.go
```diff
@@ -0,0 +1,254 @@
...
+}
+
+// GetHeaderByNumber selects header using block number.
+func (db *Database) GetHeaderByNumber(number *big.Int) (*types.Header, error) {
+	header := &types.Header{}
+	err := db.db.QueryRow("SELECT header FROM blocks WHERE number = ?", (*SQLBigInt)(number)).Scan(&JSONBlob{header})
+	if err == nil {
+		return header, nil
+	}
+	if err.Error() == errNoRows {
```
adambabik: This special case is not tested.

services/wallet/downloader.go
```diff
@@ -0,0 +1,159 @@
...
+			receipt, err := d.client.TransactionReceipt(ctx, tx.Hash())
+			if err != nil {
+				return nil, err
+			}
+			rst = append(rst, Transfer{Type: ethTransfer, Header: header, Transaction: tx, Receipt: receipt})
+			continue
+		}
+	}
+	if len(rst) == 0 {
+		return nil, errors.New("balance changed but no new transactions were found")
```
adambabik: Is this an error?

services/wallet/erc20/erc20.go
```diff
@@ -0,0 +1,353 @@
+// Code generated - DO NOT EDIT.
```
adambabik, line 1: How is it generated? Do we have gogenerate for it?

services/wallet/reactor.go
```diff
@@ -0,0 +1,204 @@
...
+	r.wg.Wait()
+	r.quit = nil
+}
+
+func (r *Reactor) loop() {
+	var (
+		// TODO(dshulyak) parametrize
+		// pow block on main chain is mined once per ~14 seconds
+		// but we for tests we are using clique chain with immediate block signer
+		ticker = time.NewTicker(1 * time.Second)
```
adambabik: Let's solve it now with `Reactor` param. We can default it using a map of networkID to tick time.

adambabik (commented):
I would also add more comments to methods (even private ones) explaining why a given method exists and what it does with all cases described, even if the cases are not special. This should be understandable by people who don't know how Ethereum blockchain works.

api/backend.go
dshulyak: i just found that Reactor is a common name for similar parts of applications across different projects and seems like a better name

services/wallet/api.go
dshulyak: i actually don't want to add more methods to Service. initially i wanted to pass db to API and make sure that it doesn't have access to Service at all. but then i had to create db after API is created

api/backend.go
dshulyak: i can't because i need to exit only if err != nil. if it is nil then Logout needs to happen

services/wallet/api.go
dshulyak: i prefer to return no transfers explicitly to avoid any confusion

services/wallet/erc20/erc20.go
dshulyak: i described in README.md which solidity snippet was used to generate this file. i generated it with abigen. i can try to make gogenerate statement for it

services/wallet/reactor.go
dshulyak: makes sense. will do it that way

services/wallet/downloader.go
dshulyak: yeah it is subtle but can happen. i query balance with block number and then query block body and because we are using infura we can hit different replica. it could be that we will hit replica A on first query and replica B on second. and block won't be the same there

but i realized that this is not the best way to test because gas spent will result in balance change as well

services/wallet/database.go
dshulyak: i tried. it looks much more readable the way it is now. note that i have to ignore error if tx is rollbacked and i need to pass error to closeTx to detemine how to close it

services/wallet/events.go
dshulyak: the point here is that there will be multiple blocks added in reorg, e.g. 4,5,6,7 and BlockNumber will be 4 - earliest added block. maybe there is a better word for it then earliest..

services/wallet/erc20/erc20.go
dshulyak: added file with gogenerate statement

services/wallet/database.go
dshulyak: done

services/wallet/database.go
dshulyak: done

services/wallet/README.md
dshulyak: that was pretty much the whole idea. added some details to readme

dshulyak:
@PombeirP @adambabik thank you guys for review. pushed update with address comments

api/backend.go
pedropombeiro: It's the first time I've seen it, it makes me think of a nuclear reactor :-)

services/wallet/events.go
pedropombeiro: `In this case, BlockNumber represents the first reorganized block.`?

services/wallet/api.go
adambabik: Not sure if it matters here but generally it's better to have business logic in a service and API being just a wrapper. Then, you can use the service in Go and the API layer if going through JSON-RPC. If you put everything in the API, it might be inefficient for Go due to limitation of the API methods.

If you don't think that is the case here, let's keep it.

services/wallet/database.go
adambabik: Ah right, it would look like `err = closeTx(tx, err)`. Not really clear.

adambabik (approved):

services/wallet/events.go
dshulyak: thanks, definitely better than what i wrote. i will change it in a followup PR to avoid rebase

services/wallet/api.go
dshulyak: i am not sure that i like when service implements all this logic. if i had to write multi-component service i would probably leave service to manage lifecycle and implement p2p interface. all other logic would be in sub-components, grouped by function for example Read and Send otr by objects that are used in each component. 

in general there is no problem using API directly in golang, one exception is when rpc.Subscription is returned. but if thats the issue there could be one more layer between service and api

gravityblast:
@dshulyak would it be possible to expose a method to start the service accepting a list of addresses instead of using `b.accountManager.SelectedWalletAccount()`?

This would enable this part to work with multi-account, where `status-react` knows a list of addresses/accounts associated to a user.

dshulyak:
@gravityblast 

> @dshulyak would it be possible to expose a method to start the service accepting a list of addresses instead of using b.accountManager.SelectedWalletAccount()?

sure, can be separate method. maybe it makes sense to change signature of the SelectAccount and it will accept multiple wallet addresses? i will simply pass them to my service. but whatever you prefer

gravityblast:
With multi-account there will be watch-only addresses too, that are not even stored in go. So I think if we can have something callable from status-react it would be better. For keycard as well, addresses are not stored in go and keys are on the card.

dshulyak:
> With multi-account there will be watch-only addresses too, that are not even stored in go. So I think if we can have something callable from status-react it would be better. For keycard as well, addresses are not stored in go and keys are on the card.

thats clear. i meant that we can receive list of addresses to watch using the same method (SelectAccount), doesn't matter where they are stored. but if it makes more sense to have separate method than sure let's add it

gravityblast:
@dshulyak ok so status-react will login passing "main address"/password and a list of addresses to watch that will be set in a slice inside account manager. would that be ok?

dshulyak:
> @dshulyak ok so status-react will login passing "main address"/password and a list of addresses to watch that will be set in a slice inside account manager. would that be ok?

yes

gravityblast:
ok it makes sense! if we can already setting up to be compatible with multiple accounts then I'll update it when I change the login part, thank you!

PR #1473, opened 2019-05-29
Downloaders for historical transfers

dshulyak:
developed on top of https://github.com/status-im/status-go/pull/1467

todo:
- better algorithm for finding eth transfers


dshulyak:
@yenda  you can take this branch instead of https://github.com/status-im/status-go/pull/1467. it will download all historical transfers. for erc20 i am using getLogs for range of blocks. and eth is downloaded by comparing balance between ranges of blocks. so eth downloader might be a bit slow.

dshulyak:
@PombeirP @adambabik since you guys already reviewed previous PR you can help me with this one :)

dshulyak:
will implement better algorithm for finding eth transfers in a followup PR

yenda:
@dshulyak I get this error a few seconds after login:
https://gist.github.com/yenda/5aa89b865af7ec752b5f34d32b615e43

here is the node config
https://gist.github.com/yenda/8fdcc165049b573c51d74cf45cb8f25e

dshulyak:
there is a bug in ethclient https://github.com/ethereum/go-ethereum/blob/master/ethclient/ethclient.go#L243-L251 it doesn't check an error. i pushed a workaround and will submit a fix for it

dshulyak:
made a fix here https://github.com/ethereum/go-ethereum/pull/19653

@yenda btw can you give me an address that you are using? i want to run it with my script

yenda:
@dshulyak `0x5DC6108dC6296b052BBd33000553AfE0Ea576b5E`

yenda:
I am not sure how to use `wallet_getTransfers`
In the json-rpc api block numbers are passed in hex, here it seems like it takes an integer

If I pass `[7773254 nil]` I get nothing, is it because the range start is the most recent block and end the oldest in the api? it is a bit confusing so the doc could be clarified on that
If I pass `[nil nil]` the app crashes https://gist.github.com/yenda/6c0ce03c321c05bc3ba354ef01e84248

This is all the params I tried with `0x5dc6108dc6296b052bbd33000553afe0ea576b5e`:
```
status-im.ethereum.json-rpc> (call {:method "wallet_getTransfers" :on-success println :params [783254 7773254]})
nil

status-im.ethereum.json-rpc> (call {:method "wallet_getTransfers" :on-success println :params [773254 7873254]})
nil

status-im.ethereum.json-rpc> (call {:method "wallet_getTransfers" :on-success println :params [7773254 7873254]})
nil

status-im.ethereum.json-rpc> (call {:method "wallet_getTransfers" :on-success println :params [7773254 nil]})
nilnil

status-im.ethereum.json-rpc> (call {:method "wallet_getTransfers" :on-success println :params [7873254 nil]})
nil

status-im.ethereum.json-rpc> (call {:method "wallet_getTransfers" :on-success println :params [7873346 nil]})
nil

status-im.ethereum.json-rpc> (call {:method "wallet_getTransfers" :on-success println :params [nil nil]})
crash
```

dshulyak:
> If I pass [7773254 nil] I get nothing, is it because the range start is the most recent block and end the oldest in the api? it is a bit confusing so the doc could be clarified on that

it gets all transfers from first block till the last if last is not provided query will have no limit. query works in forward order if second blocks is lower that the first probably nothing will be selected

as for the first one is nil i will handle that.

have you tried to query using 0 as a first parameter?

> In the json-rpc api block numbers are passed in hex, here it seems like it takes an integer

i will change it to hex

dshulyak:
examples:
```json
{"jsonrpc":"2.0","id":14,"method":"wallet_getTransfers","params":[0,20]}
{"jsonrpc":"2.0","id":14,"method":"wallet_getTransfers","params":[0,null]}
{"jsonrpc":"2.0","id":13,"method":"wallet_getTransfers","params":[0]}
```

dshulyak:
latest commit will accept hex encoding
```
{"jsonrpc":"2.0","id":5,"method":"wallet_getTransfers","params":["0x0"]}
```

yenda:
@dshulyak Nice it works, however is there a limit of 100 transactions? Because I have 176 transactions in the wallet but when I call `wallet_getTransfers` I only get 100.

edit: now 101

In the log the only thing I can see related to this is:
```
06-02 14:53:47.818  6785  6910 E GoLog   : WARN [06-02|12:53:47.818] failed to get latest block               number=5716892 error="not found"
```

yenda:
@dshulyak Also I noticed the following issues:

1. it's nitpicking but for consistency with the rest of the api, keys should start with a lowercase
2. when logging out, the node is stopped by status-react (as usual), but the node.stopped signal is never received so I can't login again. I don't see anything related in the logs

Here is a sequence of signal received (integers are block number obtained with subscription to newBlock and calling back on getBlockByHash:
```
node.started
node.started nil
node.ready
discovery.started
discovery.started nil
discovery.summary
discovery.summary
discovery.summary
discovery.summary
discovery.summary
subscriptions.data
mailserver.request.completed
5717008
wallet
wallet {:Type newblock, :BlockNumber 5717000}
wallet
wallet {:Type newblock, :BlockNumber 5717001}
subscriptions.data
5717010
wallet
wallet {:Type newblock, :BlockNumber 5717002}
subscriptions.data
5717011
wallet
wallet {:Type newblock, :BlockNumber 5717003}
discovery.stopped
discovery.stopped nil
wallet
wallet {:Type newblock, :BlockNumber 5717004}
wallet
wallet {:Type newblock, :BlockNumber 5717005}
wallet
wallet {:Type newblock, :BlockNumber 5717006}
wallet
wallet {:Type newblock, :BlockNumber 5717007}
wallet
wallet {:Type newblock, :BlockNumber 5717008}
wallet
wallet {:Type newblock, :BlockNumber 5717009}
wallet
wallet {:Type newblock, :BlockNumber 5717010}
wallet
wallet {:Type newblock, :BlockNumber 5717011}
wallet
wallet {:Type newblock, :BlockNumber 5717012}
```

3. From the api doc it says `Emmited when transfers from new block were added to the database. In this case block number if the number of this new block.`
So as expected once it has catched up with the regular subscription to newBlock with a callback to get blockNumber this is faster, but the problem is that it needs to catch up. Why is it related to wallet? I though of this event as an `ethereum.new-block` signal to avoid having to make a second call on each new block

4. After requesting history I see that there is a lot of signals with type history. It seems to iterate by 1000 block, is it because the dichotomic search isn't implemented yet and it just iterates by 1000 blocks?

5. I don't thing status-react has to know about these `History` signals, only if an old transaction is found

6. Since it seems `wallet_getTransfers` can return incomplete result I think it should return the range of blocks that was already scanned so that status-react knows it didn't get the full history yet

7. I keep receiving `History` blocks way past the block of my first historical transaction. It should stop when the balance is 0 and `eth_getTransactionCount` is 0 at a given block

yenda:
@dshulyak additionaly the transactions could ideally be returned in our internal format, the additional data isn't useful on status-react:

```
(let [inbound? (= address to)
        token    (get chain-tokens contractAddress
                      {:name tokenName
                       :symbol tokenSymbol
                       :decimals tokenDecimal
                       :address contractAddress})]
    {:value         value
     ;; timestamp is in seconds, we convert it in ms
     :timestamp     (str timeStamp "000")
     :symbol        (keyword tokenSymbol)
     :type          (if inbound?
                      :inbound
                      :outbound)
     :block         blockNumber
     :hash          hash
     :from          from
     :to            to
     :gas-limit     gas
     :gas-price     gasPrice
     :gas-used      gasUsed
     :nonce         nonce
     :data          input
     :error?        false
     :transfer      true
     :token         token}
```

this is compatible with what etherscan returns when polling it's api, which means you can optionally cheat on status-go for an initial history loading (maybe as an option to wallet config).

```
(def etherscan-api-key "DMSI4UAAKUBVGCDMVP3H2STAMSAUV7BYFI")

(defn- get-api-network-subdomain [chain]
  (case chain
    (:testnet) "api-ropsten"
    (:mainnet) "api"
    (:rinkeby) "api-rinkeby"))

(defn- get-transaction-url
  ([chain address] (get-transaction-url chain address false))
  ([chain address chaos-mode?]
   {:pre  [(keyword? chain) (string? address)]
    :post [(string? %)]}
   (let [network-subdomain (get-api-network-subdomain chain)]
     (if chaos-mode?
       "http://httpstat.us/500"
       (str "https://" network-subdomain
            ".etherscan.io/api?module=account&action=txlist&address=" address
            "&startblock=0&endblock=99999999&sort=desc&apikey=" etherscan-api-key
            "&q=json")))))

(defn- get-token-transaction-url
  ([chain address] (get-token-transaction-url chain address false))
  ([chain address chaos-mode?]
   {:pre  [(keyword? chain) (string? address)]
    :post [(string? %)]}
   (let [network-subdomain (get-api-network-subdomain chain)]
     (if chaos-mode?
       "http://httpstat.us/500"
       (str "https://" network-subdomain
            ".etherscan.io/api?module=account&action=tokentx&address=" address
            "&startblock=0&endblock=999999999&sort=asc&apikey=" etherscan-api-key
            "&q=json")))))
```

dshulyak:
@yenda 

> Nice it works, however is there a limit of 100 transactions? Because I have 176 transactions in the wallet but when I call wallet_getTransfers I only get 100.

hm for some reasons some transactions are not found then. do you expect me to find only pure eth transfers or any tx that spends gas or for example puts eth to multisig?

> it's nitpicking but for consistency with the rest of the api, keys should start with a lowercase

what keys?

UPD: keys in signals? i will change then

> when logging out, the node is stopped by status-react (as usual), but the node.stopped signal is never received so I can't login again. I don't see anything related in the logs

looks like something isn't stopped properly. i will debug

> From the api doc it says Emmited when transfers from new block were added to the database. In this case block number if the number of this new block.
So as expected once it has catched up with the regular subscription to newBlock with a callback to get blockNumber this is faster, but the problem is that it needs to catch up. Why is it related to wallet? I though of this event as an ethereum.new-block signal to avoid having to make a second call on each new block

thats a block that was indexed by wallet. i wanted to return it only if there were transfers, but then kept cause it is cheap to query for new transfers. so what t is the suggestion? i can hide it if it is not useful

> After requesting history I see that there is a lot of signals with type history. It seems to iterate by 1000 block, is it because the dichotomic search isn't implemented yet and it just iterates by 1000 blocks?

yes. erc20 requestd in bigger batches, eth in smaller. eth will be changed in next PR

> I don't thing status-react has to know about these History signals, only if an old transaction is found

done

> Since it seems wallet_getTransfers can return incomplete result I think it should return the range of blocks that was already scanned so that status-react knows it didn't get the full history yet

i think getTransfers should do only one thing. maybe another api call to query earliest known block will help

> I keep receiving History blocks way past the block of my first historical transaction. It should stop when the balance is 0 and eth_getTransactionCount is 0 at a given block

i will adjust it

dshulyak:
changes in last commit:

- More debug info when node is stopped
- low keys in events
- send history event only if new transfers were found

can you turn on debug logging for status-go? i wasn't able to repro 1st issue. but added couple of enhancements and logs when service is stopped

PR #1475, opened 2019-05-31
Faster downloader for historical eth transfers

dshulyak:
Implemented on top of https://github.com/status-im/status-go/pull/1473


dshulyak:
@yenda you can switch to this change now. it is rebased an includes last discussed changes.

yenda:
Here is a report for the logout bug:

https://gist.github.com/yenda/04c77cf8ba0504237918f8ffdaba4016

PR #1480, opened 2019-06-04
Download transfers for multiple addresses

dshulyak:
one thing that i want to add but lack some info is how to support adding accounts in runtime. i will add that once multi-accounts will be implemented in status-go


dshulyak:
@yenda you can use this change. i added all api changes to readme. while multi accounts are not implemented wallet service will be using single address for all transfers, but api should stay the same.

PR #1488, opened 2019-06-10
Integrate status-console-client into chatapi service

adambabik:
### Todo

- [x] fix the service so that the account is not needed immediately
- [x] enable PFS in `Messanger` from status-console-client
- [ ] ~~test the service and provide examples of JSON-RPC method calls~~ this is not possible without changing `status` service; it's not possible to properly select an account using JSON-RPC. It's possible only using bindings.


services/chatapi/api.go
```diff
@@ -0,0 +1,152 @@
...
+
+	contacts, err := api.s.messenger.Contacts()
+	if err != nil {
+		return result, fmt.Errorf("failed to load contacts: %v", err)
+	}
+
+	contactToChat := make(map[client.Contact][]*protocol.Message)
+
+	for _, c := range contacts {
+		messages, err := api.s.messenger.Messages(c, 0)
```
dshulyak, line 50: we will load whole db into memory. i think we need specific api to count unread messages per chat and load last msg per chat

services/chatapi/service.go
```diff
@@ -0,0 +1,84 @@
...
+			Service:   NewPrivateAPI(s),
+			Public:    true,
+		},
+	}
+}
+
+// Start is run when a service is started.
+func (s *Service) Start(server *p2p.Server) error {
+	// get the current account's private key
+	chatAccount, err := s.accountManager.SelectedChatAccount()
```
dshulyak: i think account is selected after node is started, i had this problem with wallet service

services/chatapi/api.go
adambabik: Yes, I will change that later.

services/chatapi/service.go
adambabik: How did you solve it?

dshulyak: i started part of app that requires key (address and decrypted db in my case)  when account is selected. the same way as for pfs

adambabik:
```
$ nc -U /Users/adamb/repos/status-go/test-dir/geth.ipc
{"jsonrpc":"2.0","method":"status_login", "params": [{"address": "REDACTED", "password": "REDACTED"}], "id":1}
{"jsonrpc":"2.0","id":1,"result":null}
{"jsonrpc":"2.0","method":"status_start", "params": [], "id":1}
{"jsonrpc":"2.0","id":1,"result":null}
{"jsonrpc":"2.0","method":"status_joinPublicChat", "params": ["status"], "id":1}
{"jsonrpc":"2.0","id":1,"result":null}
{"jsonrpc":"2.0","method":"status_chats", "params": [], "id":1}
{"jsonrpc":"2.0","id":1,"result":{"unviewed-messages-count":0,"chats":{}}}
```

One problem here is that when a public chat is joined, it should be returned in `status_chats`. Currently, a Whisper subscription is created and messages are loaded but the contact is not added and not returned. This needs to be fixed.

Issue: https://github.com/status-im/status-console-client/issues/99

adambabik:
It's outdated. We have https://github.com/status-im/status-protocol-go now.

PR #1497, opened 2019-06-21
Cache keys

cammellos:
Generating a symkey can take up to a second on slow devices, this commit
makes so that keys are saved once generated and stored in the database.


static/chat_db_migrations/1561059285_add_whisper_keys.up.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE whisper_keys (
+  chat_id TEXT NOT NULL PRIMARY KEY ON CONFLICT IGNORE,
```
dshulyak: nit: not null is not necessary with primary key

static/chat_db_migrations/1561059285_add_whisper_keys.up.sql
```diff
@@ -0,0 +1,4 @@
...
+  key BLOB NOT NULL
+)
```
dshulyak: without rowid can be used

services/shhext/filter/service.go
```diff
@@ -390,19 +401,36 @@ func (s *Service) loadContactCode(identity string) (*Chat, error) {
...
 	topics := [][]byte{topic}
 
-	symKeyID, err := s.whisper.AddSymKeyFromPassword(chatID)
-	if err != nil {
-		log.Error("SYM KEYN FAILED", "err", err)
-		return nil, err
-	}
+	symKey, ok := s.keys[chatID]
+	if ok {
+		log.Debug("Loading from cache")
```
dshulyak: maybe add a chat id to log?

dshulyak (approved):

static/chat_db_migrations/1561059285_add_whisper_keys.up.sql
dshulyak: @cammellos any reason not to? not trying to enforce, just curious why

cammellos: no reason, didn't think of it, but we don't need `rowid` so I have added it now, thanks!

services/shhext/filter/persistence.go
```diff
@@ -0,0 +1,60 @@
...
+type SQLLitePersistence struct {
+	db *sql.DB
+}
+
+func NewSQLLitePersistence(db *sql.DB) *SQLLitePersistence {
+	return &SQLLitePersistence{db: db}
+}
+
+func (s *SQLLitePersistence) All() (map[string][]byte, error) {
+
```
adambabik:
```suggestion
```

services/shhext/filter/persistence.go
```diff
@@ -0,0 +1,60 @@
...
+			return nil, err
+		}
+		keys[chatID] = key
+	}
+
+	return keys, nil
+}
+
+func (s *SQLLitePersistence) Add(chatID string, key []byte) error {
+	statement := "INSERT INTO whisper_keys(chat_id,key) VALUES(?,?)"
```
adambabik:
```suggestion
	statement := "INSERT INTO whisper_keys(chat_id, key) VALUES(?, ?)"
```

adambabik (approved):

PR #1499, opened 2019-06-24
Move installations to status-go

cammellos:
This commit moves installations management/storage to status-go.
We remove the native binding and provide RPC endpoints to set the
metadata and return a list of our own installations.


services/shhext/chat/multidevice/sql_lite_persistence.go
```diff
@@ -52,20 +53,110 @@ func (s *SQLLitePersistence) GetActiveInstallations(maxInstallations int, identi
 
 }
 
+// GetInstallations returns all the installations for a given identity
+// we both return the installations & the metadata
+func (s *SQLLitePersistence) GetInstallations(identity []byte) ([]*Installation, error) {
+	installationMap := make(map[string]*Installation)
+	var installations []*Installation
+
+	// We query both tables as sqlite does not support outer joins
```
dshulyak: it seems not to support only full outer join, which is what you need as it seems. i was using only inner but internet says so :)

services/shhext/chat/multidevice/sql_lite_persistence.go
```diff
@@ -52,20 +53,110 @@ func (s *SQLLitePersistence) GetActiveInstallations(maxInstallations int, identi
...
+	}
+	defer installationsStmt.Close()
+
+	installationRows, err := installationsStmt.Query(identity)
+	if err != nil {
+		return nil, err
+	}
+
+	for installationRows.Next() {
+		var installationID string
```
dshulyak: you can use a single multiline var statement, var ( ... )

services/shhext/publisher/service.go
```diff
@@ -188,29 +187,87 @@ func (s *Service) GetBundle(myIdentityKey *ecdsa.PrivateKey) (*protobuf.Bundle,
...
+// SetInstallationMetadata sets the metadata for our own installation
+func (s *Service) SetInstallationMetadata(installationID string, data *multidevice.InstallationMetadata) error {
 	if s.protocol == nil {
 		return errProtocolNotInitialized
 	}
 
-	return s.protocol.DisableInstallation(myIdentityKey, installationID)
+	privateKeyID := s.whisper.SelectedKeyPairID()
+	if privateKeyID == "" {
+		return errors.New("no key selected")
```
dshulyak: it is common to define static variable on top, e.g. ErrNoKeySelect = ...
it looks it is also repeated several times

services/shhext/chat/multidevice/service.go
```diff
@@ -7,11 +7,28 @@ import (
...
+	ID string `json:"id"`
 	// The last known protocol version of the device
-	Version uint32
+	Version uint32 `json:"version"`
+	// Enabled is whether the installation is enabled
+	Enabled bool `json:"enabled"`
+	// Timestamp is the last time we saw this device
+	Timestamp int64 `json:"timestamp"`
+	// InstallationMetadata
+	InstallationMetadata *InstallationMetadata `json:"metadata"`
```
dshulyak, line 31: not clear why it is stored in a separate table, can you add a comment somewhere?

dshulyak (approved):

services/shhext/chat/multidevice/service.go
cammellos: I will add a comment, thanks, basically sometimes we have only metadata, and we don't have other info on the device, in such cases, we want to store only the metadata, so I thought it would be best to keep in a separate table.

services/shhext/chat/multidevice/persistence.go
```diff
@@ -8,5 +8,9 @@ type Persistence interface {
 	// DisableInstallation disable the installation.
 	DisableInstallation(identity []byte, installationID string) error
 	// AddInstallations adds the installations for a given identity, maintaining the enabled flag
-	AddInstallations(identity []byte, timestamp int64, installations []*Installation, defaultEnabled bool) error
+	AddInstallations(identity []byte, timestamp int64, installations []*Installation, defaultEnabled bool) ([]*Installation, error)
```
adambabik, line 11: Hm what's the difference between `installations` taken as an argument and returned?

services/shhext/chat/multidevice/sql_lite_persistence.go
```diff
@@ -52,20 +55,116 @@ func (s *SQLLitePersistence) GetActiveInstallations(maxInstallations int, identi
...
+		)
+		if err != nil {
+			return nil, err
+		}
+		installationMap[installationID] = &Installation{
+			ID:                   installationID,
+			Version:              version,
+			Enabled:              enabled,
+			Timestamp:            timestamp,
+			InstallationMetadata: &InstallationMetadata{},
```
adambabik, line 99: Do we need to initialize it? If we always want to have a zero value, maybe it would be better not to use a pointer?

services/shhext/chat/multidevice/sql_lite_persistence.go
```diff
@@ -52,20 +55,116 @@ func (s *SQLLitePersistence) GetActiveInstallations(maxInstallations int, identi
...
+	}
+	defer installationsStmt.Close()
+
+	installationRows, err := installationsStmt.Query(identity)
+	if err != nil {
+		return nil, err
+	}
+
+	for installationRows.Next() {
+		var (
```
adambabik, line 79: `var installation Installation` and then `&installation.ID` wouldn't work?

services/shhext/chat/multidevice/persistence.go
cammellos: we return only those installations that we actually inserted, I'll add a comment

services/shhext/chat/multidevice/sql_lite_persistence.go
cammellos: Only in this endpoint, other endpoint don't return the metadata as it's not needed, so unless you feel strongly I would keep it as a pointer

adambabik: No, if neither case covers 100%, let's keep it. I would just add a comment in this line why we do it this way as it's a bit odd and in a normal situation, it would be skipped.

PR #1500, opened 2019-06-25
add multi-account support

### LINKED ISSUE #1479: multi-account support from main HD Wallet (GO)
# Problem
We want to have multi-account support (multiple ethereum accounts for each user),
and custom derivation paths for HD Wallets (another issue willll be opened for imported accounts).

Each user has accounts derived from the main HD Wallet + imported accounts.
The main HD wallet is the one defined by the generated/imported mnemonic phrase.

From the main HD Wallet we can derive and save extended keys that we can use to derive other
accounts without the need to store the master key.

When users generate or import a HD Wallet, we want to store one or many root extended keys,
used to derive child keys.

## Example

* The user opens Status for the first time
* A master key is generated and kept in memory in status-go
* status-react requests a list of n addresses derived from the BIP44 parent key (m/44'/60'/0'/0), or any custom path.
* status-go returns a list of addresses so that status-react can show them to the user
* status-react tells status-go to save a list of keys:
  * the BIP44 parent key (m/44'/60'/0'/0)
  * the first BIP44 account (m/44'/60'/0'/0/0)
  * other BIP44 keys if needed
* status-go saves the requested keys and remove the master key from memory
* later status-react can call status-go to unlock the BIP44 parent key and derive other BIP44 accounts

# Definitions
**Extended key**: a private key together with its chain code. As described in the BIP32,
an extended key can be used to derive child keys (with their own chain code).

**Normal key**: a private key without a chain code. A normal key cannot derive BIP32 child keys.

# Implementation
The current implementation has 1 keystore json file for each user.
Each keystore file is encrypted with the user's password.
The standard json keystore file has only one normal key.
Our implementation extends that format to have two keys:

* the standard normal key (that we derive with m/44'/60'/0'/0/0)
 This key is currently used for wallet and chat.
* an extended key (that we currently derive with m/44'/60'/0'/0/1)
 The initial idea of this key was to use it to derive sub-accounts.

My proposal is to keep only one key for each keystore file.
We can still have the two "extended key" and "key" fields,
but they are basically the same key.
Key is the normal key of the keystore file.
ExtendedKey is the same key with its chain code, so it allows to derive child keys.
ExtendedKey can be empty if a user imports a normal key.

When a user creates the first account status-react can asks status-go to save 4 keystore files:

* m/44'/60'/0'/0/0
 The first wallet key.
* m/44'/60'/0'/0
 The BIP44 parent key. This key allows deriving other wallet accounts.
* m/43'/60'/1581'/0'/0'
 Whisper key
* m/43'/60'/1581'/0'
 Whisper parent key. This key allows deriving other chat keys

The master key is never saved. So other subtrees are not available for future use, unless the user regenerate the master key importing the mnemonic phrase.

When we create a new random account we generate a new extended master key,
so we can derive any key, and any extended child key can derive again other keys and so on.
The same thing applies if the user is recovering an account or importing one with the mnemonic phrase.

The same process can be used for custom derivation.
status-react will pass another root path instead of the standard BIP44 path.

Importing raw private keys or standard keystore json file will create only normal keys,
not extended keys, so it won't be possible to derive child keys from those accounts.

The 4 keystore files above will be encrypted with the same user's password.
In the future we could allow having different passwords for some accounts for the same user.
status-react can save a reference to the first account and try to unlock it
every time a password is needed. So that it can verify the password before storing/encrypting a new account.

status-react needs to keep a mapping of all the created accounts by address for each user. (maybe it’s already partially implemented like this).
Example (pseudo code and names just for reference):

```
{
user1 => {
address1 => {
origin: mainHDWallet, // it will be recovered with the mnemonic phrase
type: extendedKey,
path: "m/44'/60'/0'/0"
// BIP44 accounts parent. can be instantiated to derive other child keys
},
address2 => {
origin: mainHDWallet, // it will be recovered with the mnemonic phrase
type: extendedKey,
path: "m/44'/60'/0'/0/0"
// BIP44 accounts 0. can derive child keys
// but normally we don't want to derive child keys from BIP44 accounts.
// An advance user can anyway use it to derive child keys.
},
address3 => {
origin: mainHDWallet, // it will be recovered with the mnemonic phrase
type: extendedKey,
path: "m/1’/2’/3" // custom path. It can derive child keys
},
address4 => {
origin: imported, // from raw private key or keystore json file. it won't be recovered with the mnemonic phrase
type: normalKey,
path: unknown,
// it cannot be used to derive keys
},
user2 => {
...
},
...
}
```

Each one of the 4 keys has an address.
Every time status-react needs to use of the the keys it will refer to them by address and password to unlock them.

status-go will export 2 groups of functions

1 - Import Account (for normal keys):
 1a. import raw private key
 1b. import keystore json file

Both 1a and 1b will generate normal keys, it will not be possible to derive keys. It will only be possible to encrypt and store them.

2 - Create account from extended key:
 1a. New - it instantiates a new random extended master key (used for the first account)
 1b. From mnemonic - it instantiates an extended master key from the mnemonic phrase (used to recover and to add accounts from a different mnemonic).
 1c. From existing key already stored - it will unlock an existing keystore file (imported or created previously)
 and instantiates the extended key store inside.
 If the selected keystore contains only a normal key it will return an error.

These 3 steps will instantiate an ExtendedKey in memory.
It will be a master key for 1a and 1b, or another extended key for 1c.
At this point status-react can:

* ask for n addresses from a specific root path (relative to the instantiated key).
for example with n=3 and path=m/44'/60'/0'/0 it will return the addresses for these accounts:

```
m/44'/60'/0'/0/0
m/44'/60'/0'/0/1
m/44'/60'/0'/0/2
```

This allows status-react to show addresses to the user before storing the key.

* store 1 or multiple keys specifying a specific path. For example, if the master key is instantiated:
* Store(m/44'/60'/0'/0) - this will allow status-react to save a reference to
the BIP44 root to be able to instantiate it in the future to
derive other child keys
* Store(m/44'/60'/0'/0/i) - store the BIP44 account at index i, to be used as a normal wallet key.

### LINKED ISSUE #1481: importing accounts
# Problem

Depends on #1479

After implementing #1479, users will be able to have multiple ethereum accounts derived from their main HD Wallet (recoverable from their mnemonic phrase).

We want to give users the ability to import other accounts by specifying the private key or importing a keystore json file. 

# Implementation

We should add two functions to:

- import a raw hex private key
- import a json keystore file

Both the above function will parse the input and transform the key in our keystore format and store them encrypted with the user's password.

The imported keys are normal keys, not extended keys, so it won't be possible to derive other child keys from them.

A part from deriving child keys, they will be the same of all the other keys stored previously.

# Acceptance Criteria

It's possible to import raw hex private keys and

# Next steps

Importing additional keys keys from mnemonic phrases

gravityblast:
Add `statusaccounts` RPC API to manage accounts and enable multi-account.

Closes #1479
Closes #1481 

This PR adds native functions to use the multi-accounts generator to be able to generate new accounts, derive addresses, derive child keys, and store them in keystore files.

All keys are kept in memory until `MultiAccountStoreAccount` or `MultiAccountStoreDerivedAccounts` are called.
If the generated or imported key has an extended key, the API allows to derive child keys from it.

### TODO:

* [x] remove accounts on login
* [x] unit tests for generator struct
* [x] e2e tests
* [x] move vendor changes to geth repo
* [x] check deps imported from status-im/go-ethereum


gravityblast:
@yenda can you try these builds? /cc @guylouis 
You need to add `EnableStatusAccountsService=true` in the config, and the available methods are:

* statusaccounts_generate(mnemonicLength, n) # generates n accounts with mnemonic length `mnemonicLength`
* statusaccounts_deriveAddresses(accountID, [path1, path2, etc...])
* statusaccounts_storeDerivedAccounts(accountID, "password", [path1, path2, etc...])
* statusaccounts_storeAccount(accountID, "password")
* statusaccounts_ImportPrivateKey(privateKeyHex)
* statusaccounts_ImportJSONKey(jsonString, passwordToDecrypt)

I'm going to add more docs in an issue but at least you can try to see if it works

gravityblast:
@yenda I added:

* statusaccounts_ImportPrivateKey(privateKeyHex)
* statusaccounts_ImportJSONKey(jsonString, passwordToDecrypt)

you can download again the latest builds
/cc @guylouis

gravityblast:
@yenda I updated the builds and some method signatures. I added them to the PR description

flexsurfer:
@gravityblast it would be nice to have something like `statusaccounts_generate_and_derive` which will return accounts, default wallet accounts and whisper accounts, i don't see any reason to implement this in status-react, so instead of 6 calls and logic for them we'll have only 1

gravityblast:
good idea @flexsurfer , I added that method in the last commit and added its signature in the PR description

account/utils.go
```diff
@@ -6,11 +6,13 @@ import (
...
 	"github.com/ethereum/go-ethereum/accounts/keystore"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/status-im/status-go/extkeys"
 )
 
 // errors
 var (
-	ErrInvalidAccountAddressOrKey = errors.New("cannot parse address or key to valid account address")
+	ErrInvalidAccountAddressOrKey  = errors.New("cannot parse address or key to valid account address")
+	ErrInvalidMnemonicPhraseLength = errors.New("mnemonic phrase length; valid lengths are 12, 15, 18, 21, and 24")
```
pedropombeiro: Maybe should be
```suggestion
	ErrInvalidMnemonicPhraseLength = errors.New("invalid mnemonic phrase length; valid lengths are 12, 15, 18, 21, and 24")
```

account/utils.go
```diff
@@ -68,3 +70,15 @@ func ToAddress(accountAddress string) *common.Address {
 
 	return &to.Address
 }
+
+// MnemonicPhraseLengthToEntropyStrenght returns the entropy strength for a given mnemonic length
+func MnemonicPhraseLengthToEntropyStrenght(length int) (extkeys.EntropyStrength, error) {
```
pedropombeiro:
```suggestion
func MnemonicPhraseLengthToEntropyStrength(length int) (extkeys.EntropyStrength, error) {
```

api/backend.go
```diff
@@ -536,7 +558,16 @@ func (b *StatusBackend) SelectAccount(walletAddress, chatAddress, password strin
 
 	b.AccountManager().RemoveOnboarding()
 
-	err := b.accountManager.SelectAccount(walletAddress, chatAddress, password)
+	statusAccountsService, err := b.statusNode.StatusAccountsService()
```
pedropombeiro: Shouldn't this just be a call to `resetStatusAccountsService()`?

params/config.go
```diff
@@ -349,6 +349,9 @@ type NodeConfig struct {
 	// BrowsersConfig extra configuration for browsers.Service.
 	BrowsersConfig BrowsersConfig
 
+	// StatusAccountsConfig add extra configuration for the StatusAccountsService.
```
pedropombeiro:
```suggestion
	// StatusAccountsConfig adds extra configuration for the StatusAccountsService.
```

params/config.go
```diff
@@ -374,6 +377,11 @@ type BrowsersConfig struct {
 	Enabled bool
 }
 
+// StatusAccountsConfig extra configuration for the StatusAccountsService.
```
pedropombeiro:
```suggestion
// StatusAccountsConfig defines extra configuration for the StatusAccountsService.
```

t/e2e/accounts/accounts_test.go
```diff
@@ -82,6 +83,58 @@ func (s *AccountsTestSuite) TestAccountsList() {
...
+	s.NotNil(keyStore)
+
+	// create a private key
+	privateKey, err := crypto.GenerateKey()
+	s.NoError(err)
+
+	// import as normal account
+	password := "test-password-2"
+	addr, err := s.Backend.AccountManager().ImportNormalAccount(privateKey, password)
+	s.NoError(err)
```
pedropombeiro: Shouldn't these be `s.Require().NoError(err)`? There's no point in running the next code if this one failed, right?

pedropombeiro (dismissed):

gravityblast:
thank you @PombeirP I updated the code following your suggestions!

account/accounts.go
```diff
@@ -303,6 +304,42 @@ func (m *Manager) Logout() {
 	m.selectedChatAccount = nil
 }
 
+func (m *Manager) ImportNormalAccount(privateKey *ecdsa.PrivateKey, password string) (common.Address, error) {
```
adambabik: What is `normal` account? Let's add a comment.

extkeys/hdkey.go
```diff
@@ -365,9 +365,14 @@ func (k *ExtendedKey) Neuter() (*ExtendedKey, error) {
 	}, nil
 }
 
+// IsZeroed returns true if key is nil or empty
+func (k *ExtendedKey) IsZeroed() bool {
```
adambabik, line 369: Do we need to export it?

adambabik, line 369:
```suggestion
func (k *ExtendedKey) isZero() bool {
```

services/statusaccounts/account.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (a GeneratedAccountInfo) toGeneratedAndDerived(derived map[string]AccountInfo) GeneratedAndDerivedAccountInfo {
+	return GeneratedAndDerivedAccountInfo{
+		GeneratedAccountInfo: a,
+		Derived:              derived,
+	}
+}
+
+// GeneratedAndDerivedAccountInfo contains GeneratedAccountInfo and derived AccountInfo mapped by derivation path.
+type GeneratedAndDerivedAccountInfo struct {
```
adambabik: All these structs could be probably handled with one and `json:"field,omitempty"`. Or not really?

services/statusaccounts/generator.go
```diff
@@ -0,0 +1,288 @@
...
+
+var (
+	errAccountNotFoundByID          = errors.New("account not found")
+	errAccountCannotDeriveChildKeys = errors.New("selected account cannot derive child keys")
+	errAccountManagerNotSet         = errors.New("account manager not set")
+)
+
+type generator struct {
+	am       *staccount.Manager
+	accounts map[string]*account
```
adambabik: Should we guard `accounts` with a mutex?

services/statusaccounts/generator.go
```diff
@@ -0,0 +1,288 @@
...
+
+func (g *generator) addAccount(acc *account) string {
+	id := uuid.NewRandom().String()
+	g.accounts[id] = acc
+
+	return id
+}
+
+// Reset resets the accounts map removing all the accounts from memory.
+func (g *generator) Reset() {
```
adambabik: It looks like the only exported method :)

services/statusaccounts/path_decoder.go
```diff
@@ -0,0 +1,224 @@
...
+
+func newPathDecoder(path string) (*pathDecoder, error) {
+	d := &pathDecoder{
+		s: path,
+		r: strings.NewReader(path),
+	}
+
+	err := d.reset()
+
+	return d, err
```
adambabik: Maybe `return nil, err` in case of an error?

services/statusaccounts/path_decoder.go
```diff
@@ -0,0 +1,224 @@
+package statusaccounts
```
adambabik: It looks like a nice piece of code to be a separate package/lib in the root dir of the project. But we can think about it later.

dshulyak:
@gravityblast this service will be available only after login with main account, right?

gravityblast:
@dshulyak  it should always be available, it's also used to generate/recover an account

dshulyak:
@gravityblast that's might be a problem. first, according to @yenda we are not using StartNode without Login. So this service might not be available if you need it to generate or recover account. Previously account was generated using native bindings, no?

But also i am moving settings for accounts to status-go, all settings will be stored in the database encrypted using password. So as you understand they won't be available before actual login, and we won't be able to start a node before login. That's clearly incompatible with your change, so we need to figure out smth

on more general note, we currently do quite weird stuff with node startup. Several services are actually requiring password to decrypt their databases. Therefore we start them when node starts, and then inject databases when user provides password. If we could merge Login with StartNode this flow will be much better.

flexsurfer:
It would be great if status-go could manage node by itself, it's a mess to manage it from status-react, so yeah would be great to have generate and login methods available without starting node from status-react

yenda:
I'd say a node should be started when you login, or just finished creating an account. The rest (login, create account, list accounts...) should be done via native bindings

gravityblast:
> I'd say a node should be started when you login, or just finished creating an account. The rest (login, create account, list accounts...) should be done via native bindings

@dshulyak  @yenda  everything was in native bindings but we moved them to RPC call because we said that it would have been easier for status-react, that's why we did it like these.

Do we want to move them back to native bindings then?

extkeys/hdkey.go
gravityblast: @adambabik it's used in generator.go to check if the account has an extended key or not

gravityblast: @adambabik following `btcutil`  and other comments in our code base it looks like "zeroed" is the term used to described an empty extended key

services/statusaccounts/generator.go
gravityblast: yes it's the only one used from another package, which is the backend during Logout

services/statusaccounts/path_decoder.go
gravityblast: @adambabik it comes from the pkg I wrote for keycard-go:

https://github.com/status-im/keycard-go/tree/develop/derivationpath

I didn't want to add a dependency but if we want we can import it later

gravityblast:
thank you @adambabik , I updated the code after your comments. we are discussing now if we want to move it back to native binding instead of having a service

dshulyak:
@gravityblast i think we should. if we are moving settings from status-react to status-go and they will have to be encrypted then we can't start a node before decrypting them. hence node should be started after login, and after account is created.

it is probably true that it is more convenient to work with rpc, but i think it is less important here.

additionally both status-react and status-go codebase will benefit from merging login/start node and logout/stop node. all status-go services will be nicely coupled with login, so that we won't need to perform "double" start, when on first start you start a service and on second start you start a database

gravityblast:
that's ok for me! I will work on a separate branch to be able to test it and when everything is fine for status-react I will merge it to this one

services/statusaccounts/path_decoder.go
adambabik: I don't see anything wrong with that. It would be `github.com/status-im/keycard-go/derivationpath` so looks fine. However, `keycard-go` should be tagged properly and watched for breaking changes.

gravityblast:
@dshulyak @flexsurfer I created a new branch https://github.com/status-im/status-go/tree/multi-account-native-bindings

I started CI so that you @flexsurfer can have a build to test when it finishes:
https://ci.status.im/job/status-go/job/manual/436/

The functions to call have the same name starting with MultiAccount, you can see them here:
https://github.com/status-im/status-go/blob/multi-account-native-bindings/lib/multiaccount.go

Each one receive a JSON with some parameters defined in the same file, for example `MultiAccountGenerateParams`, `MultiAccountGenerateAndDeriveAddressesParams`, etc...

gravityblast:
I merged the multi-account native bindings branch to this one and updated the PR description.
It's ready to merge after another review of the code. I think @flexsurfer already tested it from status-react.
/cc @adambabik @guylouis

dshulyak:
@gravityblast nice, what about bindings in `mobile` module? i thought they are used for android/ios

gravityblast:
> @gravityblast nice, what about bindings in mobile module? i thought they are used for android/ios

thank you @dshulyak , it was in #1525 but not here. I added it now!

mobile/multiaccount.go
```diff
@@ -0,0 +1,167 @@
+package statusgo
+
+import (
+	"encoding/json"
+)
+
+// MultiAccountGenerateParams are the params sent to MultiAccountGenerate.
+type MultiAccountGenerateParams struct {
```
dshulyak, line 8: does it make sense to define all Params only once and use them in mobile and lib modules?

dshulyak: looks like the only difference between lib and mobile are types for input and output values. i would consider moving this code to separate module and reusing that code in mobile and in lib with simple facade to ensure proper types (or import mobile from lib if thats possible)

gravityblast:
>looks like the only difference between lib and mobile are types for input and output values. i would consider moving this code to separate module and reusing that code in mobile and in lib with simple facade to ensure proper types (or import mobile from lib if thats possible)

@dshulyak I changed lib to use the params from the mobile pkg

dshulyak (approved):
@gravityblast i meant that you could also do smth like this

```
func MultiAccountCreate(params *C.Char) *C.Char {
   return C.CString(mobile.MobultieAccountCreate(C.GoString(params))
}
```

lib/library_test_utils.go
```diff
@@ -157,6 +160,40 @@ func testExportedAPI(t *testing.T, done chan struct{}) {
...
+			testMultiAccountGenerateAndDerive,
+		},
+		{
+			"MultiAccount - Import/Store",
+			testMultiAccountImportStore,
+		},
+	}
+
+	for _, test := range tests {
+		t.Logf("=== RUN   %s", test.name)
```
adambabik: Why not `t.Run()`?

adambabik (approved):

gravityblast:
>@gravityblast i meant that you could also do smth like this
>func MultiAccountCreate(params *C.Char) *C.Char {
>   return C.CString(mobile.MobultieAccountCreate(C.GoString(params))
>}

Each package defines and initialize an instance of the `StatusBackend`.
I know it could be refactored but it would change everything and not only this PR. And I hope in the future to have just 1 of the 2 packages instead of having repeated code.

lib/library_test_utils.go
gravityblast: @adambabik I just used the same cod ewe already use in `testExportedAPI` I didn't want to change too much the style of those tests

gravityblast:
@PombeirP I can't merge because you requested some changes. Are those ok for you now?

PR #1502, opened 2019-06-25
Sync messages in a loop until error or no cursor

### LINKED ISSUE #1496: Syncing mailservers needs some love
# Problem

- [x] in the README.md there is no mention of `Limit` in the request
- [x] there is no error passing from mailserver (https://github.com/status-im/whisper/pull/29)
- [x] unknown behaviour if `Limit` is `0` (https://github.com/status-im/whisper/pull/29)
- [x] syncing envelopes does not follow cursor. We need an option which will follow the cursor. This is needed to sync long periods of time

cc @annadanchenko @jakubgs

adambabik:
### Problem

The current message syncing was implemented without following the cursor so the caller needed to manually read the cursor from the response and issue another call. That was ok if a mailserver was down for a short period of time as usually all messages are received in one batch.

However, when one want to sync more data, this is a big inconvenience.

### Solution

Follow the cursor automatically and finish the syncing when there is an error or no cursor meaning all data was downloaded.

Closes #1496

cc @annadanchenko @jakubgs


services/shhext/service.go
```diff
@@ -179,3 +181,46 @@ func (s *Service) Stop() error {
...
+
+	// Wait for the response which is received asynchronously as a p2p packet.
+	// This packet handler will send an event which contains the response payload.
+	events := make(chan whisper.EnvelopeEvent, 1024)
+	sub := s.w.SubscribeEnvelopeEvents(events)
+	defer sub.Unsubscribe()
+
+	// Add explicit timeout context, otherwise the request
+	// can hang indefinitely if not specified by the sender.
+	// Sender is usually throug netcat or some bash tool
```
pedropombeiro:
```suggestion
	// Sender is usually through netcat or some bash tool
```

pedropombeiro (approved):

adambabik:
Update after https://github.com/status-im/whisper/pull/29 is merged.

services/shhext/service.go
```diff
@@ -179,3 +181,46 @@ func (s *Service) Stop() error {
...
+		case event := <-events:
+			if event.Event != whisper.EventMailServerSyncFinished {
+				continue
+			}
+
+			log.Info("received EventMailServerSyncFinished event", "data", event.Data)
+
+			var ok bool
+
+			resp, ok = event.Data.(whisper.SyncEventResponse)
```
dshulyak, line 215: event from concurrent sync request can be received, is it possible to compare request id?

adambabik: Not really, there are no IDs that would allow that. But also a single requester shouldn't make more than one call. I think we can assume that safely.

services/shhext/service.go
```diff
@@ -179,3 +181,46 @@ func (s *Service) Stop() error {
 
 	return s.Service.Stop()
 }
+
+func (s *Service) syncMessages(ctx context.Context, mailServerID []byte, r whisper.SyncMailRequest) (resp whisper.SyncEventResponse, err error) {
+	err = s.w.SyncMessages(mailServerID, r)
```
jakubgs, line 186: I thought `SyncMessages` returned `(SyncMessagesResponse, error)`, not just `error`, isn't this wrong?

jakubgs (approved):
NICE

services/shhext/service.go
adambabik: It depends which method as there are at least two with the same name :D This one comes from `Whisper` struct and it only returns an error.

jakubgs: Makes sense.

PR #1504, opened 2019-06-26
Refactor publisher service so that it can be used by status-console-client

adambabik:
We want to move all stuff related to the protocol to a single place. Currently, it's being collected in status-console-client's `protocol` package. 

The messaging logic is distributed between status-go and status-react. We don't want to simply duplicate it in status-console-client, instead, we want to make status-go services responsible for some messaging functionality generic and use it temporarily in status-console-client's `protocol` package. Eventually, status-console-client `protocol` package should not depend on status-go at all.

To achieve this, the first step is to generalize the publisher service so that it can be used by status-console-client.

### DIFF
--- a/services/shhext/publisher/service.go
+++ b/services/shhext/publisher/service.go
@@ -5,6 +5,10 @@ import (
 	"crypto/ecdsa"
 	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/log"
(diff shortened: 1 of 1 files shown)


services/shhext/publisher/service.go
```diff
@@ -69,17 +70,20 @@ func New(config *Config, w *whisper.Whisper) *Service {
 }
 
 // InitProtocolWithPassword creates an instance of ProtocolService given an address and password used to generate an encryption key.
+// TODO(adam): replace with SetPFSProtocolService()
```
cammellos, line 73: as a side note, we should stop calling the protocol `PFS` as it handles encryption/multidevices etc, and some messages have no PFS (DH), so we should probably move away from the naming

adambabik:
Closing in favour of #1508.

PR #1508, opened 2019-06-28
Publisher refactor

adambabik:
The goal is to remove all dependencies on status-go/status-react specific stuff like signals from publisher package. All that specific code should be encapsulated in `services/shhext` exclusively.


messaging/publisher/publisher_test.go
```diff
@@ -164,12 +151,7 @@ func (s *ServiceTestSuite) TestTopic() {
...
 
 	// We receive the contact code
-	dedupMessage2 := dedup.DeduplicateMessage{
-		DedupID: []byte("1"),
-		Message: message2,
-	}
-
-	err = s.alice.ProcessMessage(dedupMessage2)
+	err = s.alice.ProcessMessage(message2, []byte("1"))
 	s.Require().NoError(err)
```
adambabik: In this line, I am getting:
```
                 no payload
```
@cammellos, any idea why is it so? It's a bit strange because it's a public message and `newMessage2, err := s.bob.sendContactCode()` succeeds.

cammellos: That's correct, basically in those messages there's no payload, as we don't transfer any data other than the actual bundles, we should probably not consider it an error as it's a perfectly valid message

adambabik: Ah ok that makes sense. We indeed do not send any message just contact code. The only thing is that it worked before my changes... :D

messaging/publisher/publisher.go
```diff
@@ -0,0 +1,455 @@
...
+		p.log.Debug("Not a protocol message", "err", err)
+		return ErrNoProtocolMessage
+	}
+
+	response, err := p.protocol.HandleMessage(privateKey, publicKey, protocolMessage, msgID)
+	switch err {
+	case nil:
+		msg.Payload = response
+	case chat.ErrDeviceNotFound:
+		// TODO(adam): why is this condition needed? What's the different if it it's false?
```
adambabik: @cammellos can you please help me with this one? And two lines below, I don't understand what it means if public keys are different in this case.

cammellos: `ErrDeviceNotFound` is thrown when we receive a message that is not targeted to our device.

Our own devices are manage explicitly, meaning that if we see a new device with our own public key (say A1 & A2), we will not be targeting `A2` by default, it needs to be explicitly enabled by the user (by calling `EnableInstallation(deviceID)`. That's for security reason, as we expect the user to be in control of his own devices, and he knows which one to trust (so if a user steals our own key, we won't be disclosing our own data to the new device).
That means that if we received a message from our own public key, not for our device, it's because the user *explicitly* did not want our device to receive it (because it's not a paired device).

When it comes to devices we don't own (say B1 and B2), those are managed implicitly, as there's no source of truth, so, A1 will send to B1 & B2 even if they are not paired, as we don't really know which one to trust.
So if we receive a message from B that is not targeting our device, we want to let B know that we are also to be targeted (which is what this signal does, upon receiving a message that is not targeted to us, we will reply with our bundle, so we are included in the next message).

Not sure if it's clear

cammellos: Also, this signal will not need to be sent to the client (any client) anymore, eventually this can all be handled in the backend (`status-go` or whatever), basically upon seeing that we are not targeted, we just reply with a message, and we send maximum a message every x hours (we don't want to reply to any message, otherwise it's too much traffic)

adambabik: Ok, so for the case where the public keys (from the message and our) are the same, first case, we can return `ErrNotPairedDevice` which probably gonna be only logged and skipped.

For the second case, maybe `ErrInvalidDeviceTarget` can be returned and in such a case, the signal will be sent.

Does it sound legitimate? I would even put this logic into `HandleMessage` to return a well specified error.

cammellos: yes, sounds much better!

adambabik: Actually, the second case is covered:
```go
        msg := msgs[s.config.InstallationID]
	if msg == nil {
		msg = msgs[noInstallationID]
	}

	// We should not be sending a signal if it's coming from us, as we receive our own messages
	if msg == nil && *theirIdentityKey != myIdentityKey.PublicKey {
		return nil, ErrDeviceNotFound
	}
```

I also realised that the first option is not gonna happen because as you said we don't send a message to not paired device. However, if some other implementation did that, we will have a panic because the code above is followed by `payload := msg.GetPayload()` and `msg` being `nil`.

I will cover only this case with `ErrNotPairedDevice `.

cammellos: > I also realised that the first option is not gonna happen because as you said we don't send a message to not paired device

It actually does happen because of the way whisper works, essentially we use the whisper pk to target messages, so anyone with your pk will receive the message:
```
{
  installation-id-1: encrypted-payload1
  installation-id-2: encrypted-payload2
}
```

So any of your installations will receive this, but only `installation-1` and `installation-2` will be able to decrypt the content, while `installation-3` will receive the message, but will have no content that can be decrypted

adambabik: Ah right, I did it correctly (https://github.com/status-im/status-go/pull/1508/commits/5bfcdbe9c8a325eac3bb65614df15c7144a0f786#diff-e155aad09e72a37bdf6f38432d5b9ab6R241) but described it in a wrong way. If a message is sent from device 1 or 2, of course device 3 will receive it but the installation won't match. Without the change mentioned above it could panic or this case is handled somewhere else earlier?

cammellos: Interesting, it definitely does not panic as we would have noticed, but it's not handled, so just by looking at the code it should panic.

would it be possible that this always return true? ` *theirIdentityKey != myIdentityKey.PublicKey` is using underneath `BigInt` and possibly is comparing pointers instead (as I understand it should use `Cmp`). Status react also makes the same check so we never noticed.

cammellos (approved):
looks ok to me, a bit hard to see what has changed in publisher (maybe we could use `git mv`? )
but as I understand the main changes are:

1) Initialization of the db has been moved out of publisher into shhextservice
2) Changed the inteface of process message so that it does not take a Dedup struct, but rather two args
3) Signals moved away from publisher

Anything else? (apart from various refactoring)
As I understand this is not a breaking change for status-react, correct?

adambabik:
@cammellos good point, I will try use `git mv` but generally nothing changed. I just removed `Init*` methods and added one to set dependencies.

Your summary is correct. There are no other changes except adding `chat.ErrNotPairedDevice`.

Yes, this should be backward compatible from status-react point of view.

messaging/publisher/publisher.go
adambabik: Ah, that might be a case. This is never true... In the console client, we check if explicitly with `Cmp` to avoid that: https://github.com/status-im/status-console-client/blob/6009b73feb651d0df793f758c1e630b99af4b051/protocol/client/stream.go#L81. I will fix it.

PR #1510, opened 2019-07-01
Refactor messaging

adambabik:
Simplify the structure of `messaging/` and `messaging/chat` packages.


cammellos (approved):

messaging/chat/protocol.go
```diff
@@ -197,17 +200,52 @@ func (p *ProtocolService) BuildDHMessage(myIdentityKey *ecdsa.PrivateKey, destin
...
 
-	theirIdentityKey, err := ExtractIdentity(bundle)
-	p.log.Debug("Processing bundle", "bundle", bundle)
+	installations, fromOurs, err := p.recoverInstallationsFromBundle(myIdentityKey, bundle)
 	if err != nil {
 		return nil, err
 	}
 
-	return p.multidevice.ProcessPublicBundle(myIdentityKey, theirIdentityKey, bundle)
+	// TODO(adam): why do we add installations using identity obtained from GetIdentity()
```
cammellos, line 214: It should be identical, as `theirIdentityKey` should match (is the key recovered from the signature, but it is compare in it's string form against crypto.DecompressPubkey(bundle.GetIdentity())), best guess is that probably some fixtures are not there in the tests, I can have a quick look

cammellos (approved):

messaging/chat/protocol.go
adambabik: I left `TODO` but it would be great to take a look.

PR #1511, opened 2019-07-02
Add personal topic & use signals for messages

cammellos:
This commits does 2 things:
1) Adds a personal topic, which is not currently needed but might help out with migrations
2) Change the behavior to send signals in status-react, this currently is a breaking change as polling from a filter will remove any outstanding messages.


messaging/filter/service.go
```diff
@@ -137,9 +148,29 @@ func (s *Service) Init(chats []*Chat) ([]*Chat, error) {
...
+			return
+		}
+	}
+}
+
 // Stop removes all the filters
 func (s *Service) Stop() error {
 	var chats []*Chat
+
+	close(s.quit)
```
adambabik: Maybe we can put it at the top as it gets in the middle of collecting `chats`? 

Also, in theory, when `s.quite` is being closed, `<-ticker.C` could return and messages are processed. Not sure whether that's ok or we should wait for that loop to finish before we continue with removing chats.

messaging/filter/service.go
```diff
@@ -484,6 +535,46 @@ func (s *Service) addAsymmetricFilter(keyAsym *ecdsa.PrivateKey, chatID string,
...
+		}
+	}
+
+	return response
+}
+
+func (s *Service) getMessagesForChat(chatID string) *Messages {
+	response := &Messages{}
+
+	response.Chat = s.chats[chatID]
```
adambabik, line 560: No mutex here? It's in `getMessages()` above.

messaging/filter/service.go
```diff
@@ -137,9 +148,29 @@ func (s *Service) Init(chats []*Chat) ([]*Chat, error) {
...
+		}
+	}
+}
+
 // Stop removes all the filters
 func (s *Service) Stop() error {
 	var chats []*Chat
+
+	close(s.quit)
 	for _, chat := range s.chats {
```
adambabik, line 176: No mutex to guard `s.chats` access?

messaging/filter/service.go
cammellos: it's only called from `getMessages()` so it should be ok not to lock here as taken care from the calling function (`getMessages()`), is that correct?

messaging/filter/service.go
cammellos: It should be ok, because those messages won't be confirmed and will just be fetched back once logging in again, if we add a lock as pointed above it should also take care of any race condition

cammellos: I mean,  I will put it at the top btw

cammellos:
@adambabik addressed feedback, thanks!

messaging/filter/service.go
adambabik: Correct but it's not obvious if another method would like to use this one. Not a big deal.

adambabik (approved):

PR #1512, opened 2019-07-02
Add api method wallet_getTokensBalances

### LINKED ISSUE #1507: Multiaccounts: wallet_checkDerivationPaths method
# Problem

When importing an account in the multi account wallet from a seed phrase, the app needs to show the user paths that are already used an contain funds

# Implementation

We would need a function that given a seed phrase, returns as efficiently as possible the balance of assets if there is any for all known derivation path, potentially we could call it like this:
`wallet_checkDerivationPaths(seedPhrase, ERC20Contracts, derivationPaths)`

and it would return map with derivation paths as keys and maps of contractAddress and balance as values

dshulyak:
New method was added that accepts list of tokens and accounts and returns balances for every pair.

The PR is big mainly because of the bindings, i can make it smaller by crafting request manually, but it will be less transparent.

closes: https://github.com/status-im/status-go/issues/1507

### DIFF
--- a/services/wallet/README.md
+++ b/services/wallet/README.md
@@ -155,6 +155,35 @@ Returns avaiable transfers in a given range.
 
 Objects in the same format.
 
+#### wallet_getTokensBalances
+
+Returns tokens balances mapping for every account. See section below for the response example.
+
+##### Parameters
+
+- `accounts` `HEX` - list of ethereum addresses encoded in hex
+- `tokens` `HEX` - list of ethereum addresses encoded in hex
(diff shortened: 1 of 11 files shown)


adambabik (approved):

dshulyak:
@yenda can you please confirm that you are ok with this api https://github.com/status-im/status-go/blob/853fc901c00cb3b02b5101e847308eb8bbac4866/services/wallet/README.md#wallet_gettokensbalances ?

PR #1514, opened 2019-07-04
Feature/handle device not found

cammellos:
When receiving a message from someone not targeting our device,
we reply with an empty message that includes our own devices, so next
time they send a message they will include our device.

This is currently done in status-react, moving down to status-go

depends on #1511 so only the last commit is relevant


messaging/filter/service.go
```diff
@@ -48,23 +49,33 @@ type Chat struct {
...
-	mutex       sync.Mutex
-	keys        map[string][]byte
+	whisper       *whisper.Whisper
+	secret        *sharedsecret.Service
+	chats         map[string]*Chat
+	persistence   Persistence
+	mutex         sync.Mutex
+	keys          map[string][]byte
+	quit          chan struct{}
+	onNewMessages func([]*Messages)
```
adambabik: This worries me a bit that we do any message retrieval/processing in `filter`. Why do we need this?

messaging/filter/service.go
```diff
@@ -137,12 +148,36 @@ func (s *Service) Init(chats []*Chat) ([]*Chat, error) {
 	return allChats, nil
 }
 
+func (s *Service) Start(checkPeriod time.Duration) {
```
adambabik: Who is the caller that provides `onNewMessages` callback? Maybe we can implement this in the caller. `filter` shouldn't be as simple as possible?

services/shhext/service.go
```diff
@@ -177,7 +179,28 @@ func (s *Service) initProtocol(address, encKey, password string) error {
...
+			signalMessages = append(signalMessages, signalMessage)
+			dedupMessages, err := s.processReceivedMessages(chatMessages.Messages)
+			if err != nil {
+				log.Error("could not process messages", "err", err)
+				continue
+			}
+
+			signalMessage.Messages = dedupMessages
+		}
+		handler.NewMessages(signalMessages)
```
adambabik: Can we move `handler` declaration here?

messaging/filter/service.go
cammellos: In status-react we poll through web3 for each filter, this changed the behavior so that the polling is done at status-go level, and there's no interaction with web3 (or whisper) anymore in status-react. Instead a signal is sent with multiple messages from different filters, for performance reasons (roughly 50% performance improvement).
We can change the interface that I don't mind, as long as it's compatible with status-react.

messaging/filter/service.go
cammellos: Sure, currently shhext service is passing a callback, called whenever new messages are available

static/chat_db_migrations/1562227874_add_device_acks_metadata.up.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE device_not_found_acks (
```
adambabik: Can we add a comment here what this table contains? Like: "it contains information to which identities and when we send info about our devices".

messaging/publisher/persistence.go
```diff
@@ -5,8 +5,10 @@ import (
 )
 
 type Persistence interface {
-	Get() (int64, error)
-	Set(int64) error
+	GetLastPublished() (int64, error)
+	SetLastPublished(int64) error
+	GetLastAcked(identity []byte) (int64, error)
```
adambabik, line 12: Do we ever delete acks? If after `ErrDeviceNotFound`, we receive a message from the same identity including our device, I think we shouldn't send them an empty message periodically anymore. Also, does not it seem like a duplication of some logic in contact code where we publish information about our bundles/devices?

cammellos: We never delete acks, but we only reply with an Ack when we have `ErrDeviceNotFound`, it's not sent periodically.

It's similar in purpose to SendContactCode, but it's only triggered when we know that the other user does not know about our device (or has a stale list), it offers some redundancy as we can't make much guarantees on whisper/mailservers.

Basically it works this way:

1) I send you a message not including your device -> You reply with an Ack letting me know
2) I send you another message not including your device -> If the cooldown period has passed (2 hours), I send back an ack, otherwise nothing
3) I send you a message including your device -> Nothing is sent back
4) I send you a message not including your device -> same as point 2

adambabik: > We never delete acks, but we only reply with an Ack when we have ErrDeviceNotFound, it's not sent periodically.

Ah right. Not sure it stuck in my head it's periodical. 

Yeah, the flow is clear. Question is whether we need a table to stor this information or it could be in memory 🤔 Possibly someone could create new identities and keep forcing us to send acks over and over and filling up the database as it's never cleared up.

cammellos: That's a good idea, although having it in memory might trigger a lot of sending messages on startup, when we retrieve data from the mailserver, but it is an option ( I would use a table for now just to be on the safe side).

The attack over acks is a concern, but in that case is much simpler for an attacker to send a legit message each time with a new identity, and fill that database up, as we would save the new messages. 

I would keep it as it for now not to introduce too many changes (status-react currently uses the same method), but I will add a comment, I think it's a good idea to store in memory (and possibly just rate-limit the sending to avoid the scenario when we start the app), and I will integrate this with status-react and next week I'll add a inmemory persistence service and we can test things out, what do you reckon?

cammellos: actually the more I think of it , the more I think you are right and should be in memory, so I will change it next week, contact-code as well, no reason not to be as well, thanks for the suggestion!

adambabik: That sounds good! 👍

adambabik (approved):

messaging/publisher/persistence.go
cammellos: Updated, I have only left `lastAcked` in memory, so we can try it out for a while, and if it works fine, I'll remove as well `lastPublished`. thanks for the suggestion

Issue #1463, opened 2019-05-14
[Onboarding] Store keys with the 6-digit pincode


flexsurfer:
It should be possible to use 6-digit pincode instead of a password 

**Actual behaviour**
5 accounts are generated and 5 public keys sent to status-react -> user chooses one -> enters password -> chosen account is stored with the password

**Expected behaviour**
5 accounts are generated and 5 public keys sent to status-react -> user chooses one -> enters 6-digit pincode -> chosen account is stored with the pincode

**Implementation notes**
There is a pincode doc describing updated account flows: https://docs.google.com/document/d/1IrjvvAuAQbYsr3xqZvXfgAeeyDZRl_x9KRxWnLcwVuI/edit?ts=5cc831fb
When using pins instead of a password, Ethereum key pair will have to be encrypted/decrypted using a keypair generated inside device's TEE, as pins do not have enough enthropy. This needs to happen on Go side as having access to it from JS context is insecure due to multiple dependencies.

`status-go` should receive an object with 3 methods (`initialize`, `encrypt`, `decrypt`) with native implementations for iOS/Android.

An Android implementation but for JS already exists in our fork of react-native-keychain (https://github.com/status-im/react-native-keychain/commit/43e5512cabb8ee064fd9e503be943dcf2c7d7669), as for iOS, inspiration can be taken from https://github.com/keybase/go-keychain.

Also a nice overview is given [here](https://www.notion.so/Encryption-in-a-secure-element-ccb8778651024440a81f4f1e626a4342)

rachelhamlin:
@gravityblast @siphiuel hmm found this issue in Wrike and don't know how to organize it. Is it still relevant?

vitvly:
Yes it is, i've been working on the status-react counterpart to this.

0xc1c4da:
➤ Nabil Naghdy commented:

We should probably update the date for this then as well

dshulyak:
@flexsurfer are you the author of the doc? i have a couple of questions, maybe you can help

>  3. The app (status-go) creates an Ethereum HD key pair (eE, dE) and encrypts it with e0
    5. The app (status-go) uses Ethereum encryption key (eE) to encrypt the db.

we cannot use asymmetric key to encrypt database. sqlite doesn't support it in any way, and i don't understand how it can. do you propose to derive secret from eE? what should i use as a key material?

> 2. The app (status-react) creates a key pair in the secure element (e0, d0)

[...]

PR #1515, opened 2019-07-04
[WIP] Implement keystore using keychain interface

dshulyak:
ref https://github.com/status-im/status-go/issues/1463

### DIFF
--- a/keystore/keystore.go
+++ b/keystore/keystore.go
@@ -1,0 +1,129 @@
+package keystore
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/ethereum/go-ethereum/accounts/keystore"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/pborman/uuid"
+	"github.com/status-im/status-go/extkeys"
+	"github.com/status-im/status-go/keychain"
+)
+
+func NewKeystore(directory string, keychain keychain.Keychain) *Keystore {
+	return &Keystore{
+		directory: directory,
+		keychain:  keychain,
+	}
+}
+
+type Keystore struct {
+	directory string
+	keychain  keychain.Keychain
+}
+
+func (k *Keystore) ImportKey(ekey *extkeys.ExtendedKey, auth string) (common.Address, error) {
+	// TODO zero ecdsa key and marshalled bytes
+	key, err := newKeyForPurposeFromExtendedKey(extkeys.KeyPurposeWallet, ekey)
+	if err != nil {
+		return common.Address{}, err
+	}
+	path := filepath.Join(k.directory, key.Address.String())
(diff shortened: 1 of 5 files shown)


keystore/keystore.go
```diff
@@ -0,0 +1,129 @@
...
+func (k *Keystore) Accounts() (rst []common.Address, err error) {
+	files, err := ioutil.ReadDir(k.directory)
+	if err != nil {
+		return nil, err
+	}
+	for _, f := range files {
+		if f.IsDir() {
+			continue
+		}
+		parts := strings.Split(f.Name(), "/")
```
adambabik, line 88: https://golang.org/pkg/os/#PathSeparator

keystore/keystore.go
```diff
@@ -0,0 +1,129 @@
...
+		if len(parts) < 2 {
```
adambabik, line 89: Why do we do that?

keystore/keystore.go
```diff
@@ -0,0 +1,129 @@
...
+			continue
+		}
+		rst = append(rst, common.HexToAddress(parts[len(parts[-1])]))
```
adambabik, line 92: Isn't this account file a JSON file we can get an address from? It looks odd we get this from a file name.

dshulyak: that file will be encrypted with key from keychain and we will store only encrypted version on disk

dshulyak: btw it is not fully implemented, i will check what additional info ethereum keystore keeps and will adjust keystore accordingly. i just needed something for an e2e test

keystore/keystore.go
dshulyak: i guess i can just take last par

keystore/keystore.go
gravityblast: @dshulyak @adambabik  are we going to encrypt the full json file or do we want to use the normal keystore file encryption used by go-ethereum but using the keychain key instead of the normal password?

dshulyak: we can encrypt only private key bytes. what is normal keystore file encryption? i think they encrypt private key using key derived from password and then put that in file with more metadata.

but this PR is on hold right now, more work is required with ios and android. not sure if i abandon it or not

gravityblast: > i think they encrypt private key using key derived from password and then put that in file with more metadata.

yes exactly, I thought we were using the keychain key as password so that we didn't need to implement more things

dshulyak: the idea was to encrypt ethereum keys with keychain key. using a key as a password to keystore is an interesting option that will simplify implementation for sure. do you know if it is possible to freely export key generated on keychain?

for android i see this excerpt from https://developer.android.com/training/articles/keystore

> Key material never enters the application process. When an application performs cryptographic operations using an Android Keystore key, behind the scenes plaintext, ciphertext, and messages to be signed or verified are fed to a system process which carries out the cryptographic operations. If the app's process is compromised, the attacker may be able to use the app's keys but cannot extract their key material (for example, to be used outside of the Android device).

PR #1516, opened 2019-07-05
Refactor publisher

cammellos:
Refactoring publisher to make it easier to play with console client.


messaging/publisher/publisher.go
```diff
@@ -205,6 +216,15 @@ func (p *Publisher) LoadFilter(chat *filter.Chat) ([]*filter.Chat, error) {
 func (p *Publisher) RemoveFilters(chats []*filter.Chat) error {
 	return p.filter.Remove(chats)
 }
+func (p *Publisher) ProcessNegotiatedSecret(secrets []*sharedsecret.Secret) {
+	for _, secret := range secrets {
+
```
adambabik:
```suggestion
```

services/shhext/service.go
```diff
@@ -253,39 +240,13 @@ func (s *Service) processReceivedMessages(messages []*whisper.Message) ([]dedup.
...
-				FilterID: chat.FilterID,
-				Identity: chat.Identity,
-				Topic:    chat.Topic,
-			}
-
-			filters = append(filters, filter)
-		}
-		if len(filters) != 0 {
-			handler := PublisherSignalHandler{}
-			handler.WhisperFilterAdded(filters)
```
adambabik, line 286: We don't need to inform status-react about this because we will keep all filters on status-go side, right? Instead, we will pass inform about messages through a signal?

adambabik (approved):

services/shhext/service.go
cammellos: Correct, this is not needed anymore, it was kept there so it would be backward compatible, but code has been merged now in status-react so not needed

PR #1517, opened 2019-07-09
[service/wallet] Several changes in API after feedback

dshulyak:
- Timestamp from block header stored in blocks table and added to each transfer
- From field is computed from signature to offload this computation from client side
- `history` event is added back, so that client can ignore historical blocks when watching
only for new blocks
-  block number and timestamp are marshalled in hex. consistent with ethereum data structures


yenda (approved):
Integrated and tested in status-react

PR #1523, opened 2019-07-11
Add browsers API

dshulyak:
three methods:

## browsers_getBrowsers
returns collection of objects

```json
[
{
"id": "1",
"name": "first",
"timestamp": "0x64",
"dapp": true
},
{
"id": "2",
"name": "second",
"timestamp": "0x0",
"dapp": false
},
{
"id": "3",
"name": "third",
"timestamp": "0x0",
"dapp": false,
"historyIndex": "0x3",
"history": [
"hist1",
"hist2"
]
}
]

```

## browsers_addBrowser
accepts object in the format above

## browsers_deleteBrowser
accepts `id`


dshulyak:
wallet_getBrowsersTransit
---
```transit
[["~#b1",["1","first",100,true,0,[]]],["^0",["2","second",0,false,0,[]]],["^0",["3","third",0,false,3,["hist1","hist2"]]]]
```

flexsurfer:
why `wallet_` ?

yenda:
@flexsurfer it was just to save time for the POC

services/browsers/api_test.go
```diff
@@ -0,0 +1,130 @@
...
+			Timestamp: hexutil.Uint64(50),
+		},
+		{
+			ID:        "3",
+			Name:      "third",
+			Dapp:      true,
+			Timestamp: hexutil.Uint64(100),
+		},
+	}
+	// insert in reverse order for a clean test
```
pedropombeiro: This comment doesn't seem to reflect what it is doing.

services/browsers/database.go
```diff
@@ -0,0 +1,137 @@
...
+
+import (
+	"database/sql"
+
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/status-im/status-go/services/browsers/migrations"
+	"github.com/status-im/status-go/sqlite"
+)
+
+// Database sql wrapper for operations with wallet objects.
```
pedropombeiro: should be 
```suggestion
// Database sql wrapper for operations with browser objects.
```
?

pedropombeiro (approved):

services/browsers/api_test.go
dshulyak: right, forgot to cleanup

dshulyak:
@adambabik @cammellos @PombeirP do you see any downsides in creating one database per each simple/logicless service? i am adding api for dapps permissions and then there are several other simple things and thinking if i should use single database for all of them.

the only thing that bothers me is that it is probably faster to open a single database then several, and more open databases mean more memory for connections. but given that we have a single connection per database it is probably not an issue. what do you think?

cammellos:
@dshulyak I think it's ok, also I would not keep stuff that is status-react specific (browsers) in the main db (unless we leave that choice to the application). In terms of performance, not sure, but sounds reasonable.
I noticed though that the db is not encrypted, we should probably encrypt everything with sqlcipher, unless we really don't care about that info.

services/browsers/README.md
```diff
@@ -0,0 +1,48 @@
...
+#### browsers_addBrowser
+
+Stores browser in the database.
+All fields are specified below:
+
+```json
+{
+  "id": "1",
+  "name": "first",
+  "timestamp": "10,
```
yenda: missing `"`

dshulyak: actually the first one should be removed, integer marshalled without braces

dshulyak:
@cammellos 

> @dshulyak I think it's ok, also I would not keep stuff that is status-react specific (browsers) in the main db (unless we leave that choice to the application). In terms of performance, not sure, but sounds reasonable.

what you meant by a main db? thats in case if all things will be in the same database?

>  I noticed though that the db is not encrypted, we should probably encrypt everything with sqlcipher, unless we really don't care about that info.

i think database is encrypted, why do you say so?

cammellos:
> i think database is encrypted, why do you say so?

@dshulyak you are right, I missed that we imported our own package `status-go/sqlite`

> what you meant by a main db? thats in case if all things will be in the same database?

Yes, I was agreeing with you, I would not keep everything in a single DB, we can have a db per service, or if that's a performance problem 2 dbs, one with data that is shared across any client (messages, contacts etc), and one with data that is status-react specific (browsers, dapp-permissions etc)

dshulyak:
> Yes, I was agreeing with you, I would not keep everything in a single DB, we can have a db per service, or if that's a performance problem 2 dbs, one with data that is shared across any client (messages, contacts etc), and one with data that is status-react specific (browsers, dapp-permissions etc)

clear now, thanks for clarification. i guess i will go now with one database per service and if we will notice degradation in node startup or smth else it will be easy to adjust that

services/browsers/database.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	return &Database{db: db}, nil
+}
+
+type Browser struct {
+	ID           string   `json:"id"`
+	Name         string   `json:"name"`
+	Timestamp    uint64   `json:"timestamp"`
+	Dapp         bool     `json:"dapp"`
+	HistoryIndex int      `json:"historyIndex,omitempty"`
```
adambabik: What is `HistoryIndex` and `History` for?

adambabik (approved):

services/browsers/database.go
yenda: history is a vector of URLs and historyIndex tells you which one is the current one. However @dshulyak it would be great if it could be `history-index`, `dapp?`, `browser-id` so we can use the data directly?

PR #1525, opened 2019-07-17
Multi account login

### LINKED ISSUE #1489: multi-account login (GO)
# Problem

depends on #1479 

The current implementation has  Login function that unlock a key used for both chat and wallet and keeps it in memory. 
Decide which key we want to use for login (probably the key used for chat) and update the code to keep only the whisper key in memory.

### LINKED ISSUE #1490: update SendTransaction to unlock the sender key (GO)
# Problem

depends on #1479 

The current implementation of SendTransaction is using the key unlocked at login.
With multi-account stauts-react can specify the key to be unlocked and used using the "from" field.

### LINKED ISSUE #1492: update onboarding to be compatible with multi-account (GO)
# Problem

Depends on #1479

After implementing  #1479, we should update the onboarding process (#1456) to be compatible with multi-account and allow status-react to specify the derivation paths to be stored.

gravityblast:
depends on #1500 

This PR enables the use of multi-account already merged in #1500 .
The main change is removing the selectedWalletAccount from memory. 
Only the selectedChatAccount key remains in memory. 
For tx signing an account will be unlocked and used when needed.

Important changes:
- [x] remove selectedWalletAccount
- [x] add mainAccountAddress to be returned to `web3.eth.getAccounts()`
- [x] add watchAddresses, a list of addresses to watch when the wallet service starts.

Closes #1489 
Closes #1490 
Closes #1492

### DIFF
--- a/account/accounts.go
+++ b/account/accounts.go
@@ -32,6 +32,8 @@ var (
 	ErrOnboardingAccountNotFound      = errors.New("cannot find onboarding account with the given id")
 )
 
+var zeroAddress = common.Address{}
+
 // GethServiceProvider provides required geth services.
 type GethServiceProvider interface {
 	AccountManager() (*accounts.Manager, error)
@@ -47,8 +49,9 @@ type Manager struct {
 	accountsGenerator *generator.Generator
 	onboarding        *Onboarding
 
-	selectedWalletAccount *SelectedExtKey // account that was processed during the last call to SelectAccount()
-	selectedChatAccount   *SelectedExtKey // account that was processed during the last call to SelectAccount()
+	selectedChatAccount *SelectedExtKey // account that was processed during the last call to SelectAccount()
+	mainAccountAddress  common.Address
+	watchAddresses      []common.Address
 }
 
 // NewManager returns new node account manager.
@@ -101,66 +104,6 @@ func (m *Manager) CreateAccount(password string) (Info, string, error) {
 	info.ChatPubKey = info.WalletPubKey
 
 	return info, mnemonic, nil
-}
-
-// CreateChildAccount creates sub-account for an account identified by parent address.
-// CKD#2 is used as root for master accounts (when parentAddress is "").
-// Otherwise (when parentAddress != ""), child is derived directly from parent.
-func (m *Manager) CreateChildAccount(parentAddress, password string) (address, pubKey string, err error) {
-	m.mu.Lock()
-	defer m.mu.Unlock()
-
-	keyStore, err := m.geth.AccountKeyStore()
-	if err != nil {
-		return "", "", err
-	}
-
-	if parentAddress == "" && m.selectedWalletAccount != nil { // derive from selected account by default
-		parentAddress = m.selectedWalletAccount.Address.Hex()
-	}
-
-	if parentAddress == "" {
-		return "", "", ErrNoAccountSelected
-	}
-
-	account, err := ParseAccountString(parentAddress)
-	if err != nil {
-		return "", "", ErrAddressToAccountMappingFailure
-	}
-
-	// make sure that given password can decrypt key associated with a given parent address
-	account, accountKey, err := keyStore.AccountDecryptedKey(account, password)
-	if err != nil {
-		return "", "", fmt.Errorf("%s: %v", ErrAccountToKeyMappingFailure.Error(), err)
-	}
-
-	parentKey, err := extkeys.NewKeyFromString(accountKey.ExtendedKey.String())
-	if err != nil {
-		return "", "", err
-	}
-
-	// derive child key
-	childKey, err := parentKey.Child(accountKey.SubAccountIndex)
-	if err != nil {
-		return "", "", err
-	}
-	if err = keyStore.IncSubAccountIndex(account, password); err != nil {
-		return "", "", err
-	}
-	accountKey.SubAccountIndex++
-
-	// import derived key into account keystore
-	address, pubKey, err = m.importExtendedKey(extkeys.KeyPurposeWallet, childKey, password)
-	if err != nil {
-		return
-	}
-
-	// update in-memory selected account
-	if m.selectedWalletAccount != nil {
-		m.selectedWalletAccount.AccountKey = accountKey
-	}
-
-	return address, pubKey, nil
 }
 
 // RecoverAccount re-creates master key using given details.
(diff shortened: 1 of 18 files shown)


account/accounts.go
```diff
@@ -272,15 +212,24 @@ func (m *Manager) SetChatAccount(privKey *ecdsa.PrivateKey) {
 	}
 }
 
-// SelectedWalletAccount returns currently selected wallet account
-func (m *Manager) SelectedWalletAccount() (*SelectedExtKey, error) {
+// MainAccountAddress returns currently selected watch addresses.
+func (m *Manager) MainAccountAddress() (common.Address, error) {
```
dshulyak, line 230: is this address always the same for the single account or it can be changed based on users preferences?

gravityblast: @dshulyak  it's passed from status-react to the Login function, so I think it will be configurable from there

dshulyak: so we cannot use it as a unique account identifier?

gravityblast: yes we can use any id to identify the "user account", but this is just an ethereum address used as main account address for the user

gravityblast: @dshulyak it's only used as account to be returned from web3.eth.getAccounts

api/backend.go
```diff
@@ -605,11 +636,13 @@ func (b *StatusBackend) startBrowsers(password string) error {
...
 	}
-	account, err := b.accountManager.SelectedWalletAccount()
+
+	mainAccountAddress, err := b.accountManager.MainAccountAddress()
 	if err != nil {
 		return err
 	}
-	path := path.Join(b.statusNode.Config().DataDir, fmt.Sprintf("browsers-%x.sql", account.Address))
+
+	path := path.Join(b.statusNode.Config().DataDir, fmt.Sprintf("browsers-%x.sql", mainAccountAddress))
```
dshulyak, line 647: @gravityblast the reason i was asking if `mainAccountAddress` can change later is that if it will then new db will be instantiated, which is quite a bad thing :)

gravityblast: @dshulyak you are right! it doesn't now, but it will be possible to change it in the future. 
So maybe in your PR #1530 we can add a unique ID and use that one instead of mainAccountAddress?

dshulyak: sounds good

account/utils.go
```diff
@@ -14,6 +16,54 @@ var (
 	ErrInvalidMnemonicPhraseLength = errors.New("invalid mnemonic phrase length; valid lengths are 12, 15, 18, 21, and 24")
 )
 
+type LoginParams struct {
+	ChatAddress    common.Address   `json:"chatAddress"`
+	Password       string           `json:"password"`
+	MainAccount    common.Address   `json:"mainAccount"`
+	WatchAddresses []common.Address `json:"watchAddresses"`
```
dshulyak, line 23: chat address, main account and watch addresses are stored in status-react in current version? @flexsurfer @yenda 

can you please specify what else is stored in status-react related to multiaccount?

flexsurfer: currently only chat address and main account stored in status-react

dshulyak: where should i get watch addresses then?

i am trying to understand how this flow will change when multiaccount db state will be in status-go

gravityblast: I think we are still not using the multi-account functions. 
Before your PR I expect status-react to derive and store new addresses and send them as watchAddresses during login. together with other watch only addresses.

account/accounts.go
```diff
@@ -309,13 +257,14 @@ func (m *Manager) SelectedChatAccount() (*SelectedExtKey, error) {
...
-// Logout clears selectedWalletAccount.
+// Logout clears selected accounts.
 func (m *Manager) Logout() {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
 	m.accountsGenerator.Reset()
-	m.selectedWalletAccount = nil
+	m.mainAccountAddress = zeroAddress
+	m.watchAddresses = make([]common.Address, 0)
```
dshulyak: = nil ?

account/utils.go
yenda: I'm not a big fan of the terminology here, it's going to need some changes but probably not in this PR because we don't have the time for that.

I think WatchAddresses here is refering to the list of accounts that status-react is passing to status-go that need to be watched by the wallet service.
This should be simply called accounts. Each multiaccount contains X accounts. Chat accounts and wallet accounts are technically the same thing it's just that one is loaded in memory and not the other.

@dshulyak for status-go db I would use an account db (but not the one for the user accounts, in status-react we have sorted this out by calling user accounts multiaccounts to avoid any confusion, ideally status-go should do the same or find a similar solution) and have some attributes like "load-at-startup?", "watch-only?" to distinguish these different types of accounts wdyt?

account/accounts.go
```diff
@@ -387,40 +336,17 @@ func (m *Manager) Accounts() ([]gethcommon.Address, error) {
 	m.mu.RLock()
 	defer m.mu.RUnlock()
 
-	am, err := m.geth.AccountManager()
+	_, err := m.geth.AccountManager()
```
dshulyak: @gravityblast  you can remove this line right?

can you clarify why we don't read other addresses from manager now?

gravityblast:
@dshulyak do you think we can merge this PR before the one that moves the db to go?
I'll be on holiday next weeks so I won't be able to update this to be compatible but I think @flexsurfer needs this to be merged to finish his part.

dshulyak:
@gravityblast definitely we should merge it, i am already doing some changes on top of your PR https://github.com/status-im/status-go/pull/1546

dshulyak (approved):

api/backend.go
```diff
@@ -564,37 +585,46 @@ func (b *StatusBackend) SelectAccount(walletAddress, chatAddress, password strin
...
-	account, err := b.accountManager.SelectedWalletAccount()
+
+	watchAddresses := b.accountManager.WatchAddresses()
+	mainAccountAddress, err := b.accountManager.MainAccountAddress()
 	if err != nil {
 		return err
 	}
-	path := path.Join(b.statusNode.Config().DataDir, fmt.Sprintf("wallet-%x.sql", account.Address))
+
+	allAddresses := append(watchAddresses, mainAccountAddress)
```
dshulyak: it will extend original b.accountManager.watchAddress if internal arrays capacity is bigger than current length.

i would suggest to copy watch addresses before modifying them

account/accounts.go
gravityblast: this methods is called from `web3.eth.getAccounts()`.
For now we never had child accounts so we always returned the only address we have.
Now we added more accounts, also to let people having more addresses for privacy reasons. 
I wouldn't return all of them to a dapp. We also discussed to have a default address option in the future to allow the user decide which address to expose. So I thought to do it like this.
What do you think about it?

api/backend.go
gravityblast: good catch thank you @dshulyak

gravityblast:
thank you @dshulyak I updated the code after the comments

account/accounts.go
dshulyak: thanks, that makes sense

so you are saying that web3.eth.getAccounts is using manager.Accounts? can you show where it is called? if we will use our own instance of keystore will it affect that (https://github.com/status-im/status-go/pull/1546) ?

gravityblast: @dmitryn here we define the handler for `eth_accounts`

https://github.com/status-im/status-go/blob/develop/api/backend.go#L425

PR #1526, opened 2019-07-17
Integrate status-protocol-go

adambabik:
Let's bring https://github.com/status-im/status-protocol-go to status-go!


cammellos (commented):
@adambabik is this compatible with status-react or requires some changes? (I see `LoadFilters` has been removed, what should we use instead?)

services/shhext/service.go
```diff
@@ -168,74 +160,74 @@ func (s *Service) initProtocol(address, encKey, password string) error {
...
 		return err
 	}
 
-	multideviceConfig := &multidevice.Config{
-		InstallationID:   s.config.InstallationID,
-		ProtocolVersion:  chat.ProtocolVersion,
-		MaxInstallations: maxInstallations,
+	// TODO: v4Path is the current database file. We should figure out
+	// how to split it into multiple files and migrate data.
+	// TODO: pass onNewMessagesHandler.
```
adambabik: That's not needed since messages are handled in `retrieveMessagesLoop()`.

services/shhext/service.go
```diff
@@ -313,13 +301,18 @@ func (s *Service) Stop() error {
 	s.requestsRegistry.Clear()
 	s.envelopesMonitor.Stop()
 	s.mailMonitor.Stop()
-	if s.filter != nil {
-		if err := s.filter.Stop(); err != nil {
-			log.Error("Failed to stop filter service with error", "err", err)
+
+	if s.cancelMessenger != nil {
```
adambabik, line 336: But filters are not removed. Bring it back.

services/shhext/service.go
```diff
@@ -168,74 +160,74 @@ func (s *Service) initProtocol(address, encKey, password string) error {
...
 	}
 
-	multideviceConfig := &multidevice.Config{
-		InstallationID:   s.config.InstallationID,
-		ProtocolVersion:  chat.ProtocolVersion,
-		MaxInstallations: maxInstallations,
+	// TODO: v4Path is the current database file. We should figure out
+	// how to split it into multiple files and migrate data.
+	// TODO: pass onNewMessagesHandler.
+	messenger, err := protocol.NewMessenger(
```
adambabik, line 190: Provide an option to avoid storing messages.

adambabik:
@cammellos I haven't had a chance to build Status with this release. There are a few places missing which has `TODO` added. 

`LoadFilters` needs to be brought back because I thought that the initial list of chats can be provided where the `status.Messenger` is initialized but that does not seem possible. 

One huge part is the fact that we don't have a single database anymore. There are like three now or so. The tables and schemas are the same, although, when someone upgrades, we need to move data to these new databases. Or we can just copy them under proper names which is hacky but trivial.

Other than that, it looks like everything as it was.

cammellos:
@adambabik cool, no worries

> Or we can just copy them under proper names which is hacky but trivial.

I would probably lean towards this (or the one that requires the least amount of effort), once we are in v1 we can throw away the code as we don't need to support upgrades, but for now there's value in having an upgrade path so we can more easily detect if there's any issue.
what do you reckon?

adambabik:
@cammellos I brought back pretty much all APIs so that minimal changes on status-react side should be required.

cammellos (approved):

cammellos:
Thanks @adambabik! let me know if you are already aware of some changes required in status-react, tomorrow I can spin up the environment and see how it goes and we can take it from there

adambabik:
@cammellos there is only one API removed (`loadFilter` but `loadFilters` stayed). Except for that, all should be done. Including copying the database file into two locations (I need to add unit tests for that actually).

cammellos:
Testing this, I will report some of the issues I find here:

1) Messages are not marked as confirmed in the UI. Basically if you send a message in a public chat, it will be sent, but the spinner does not go away.
2) No one-to-one messages, as far as I can tell. Sending a one to one message to a device running this version of status-go does not open a chat
3) Sending any direct message results in `chat ID is empty`
4) Opening a one-to-one chat results in `
```
W ReactNativeJS: WARN [status-im.ethereum.json-rpc:48] - :json-rpc/error shhext_loadFilters :params [[{:C
hatID "0x04c7b7a7afbc804adda01cf71ed8bd8b868d2059e9633d5bf36b4021f949f9936c5c21cda2ba7e54f0907cce4978a40f1268fd8d3e62a173ca155a1ce72b91fb3f", :OneToOne true, :Identity "04c7b7a7afbc804adda01cf71ed8bd8b868d2059e9633d5bf36b4021f949f9936c5c21cda2ba7e54f0907cce4978a40f1268fd8d3e62a173ca155a1ce72b91fb3f"} {:ChatID "0x04c7b7a7afbc804adda01cf71ed8bd8b868d2059e9633d5bf36b4021f949f9936c5c21cda2ba7e54f0907cce4978a40f1268fd8d3e62a173ca155a1ce72b91fb3f", :OneToOne true, :Identity "04c7b7a7afbc804adda01cf71ed8bd8b868d2059e9633d5bf36b4021f949f9936c5c21cda2ba7e54f0907cce4978a40f1268fd8d3e62a173ca155a1ce72b91fb3f"}]] :error {:code -32000, :message "hex string without 0x prefix"}
```
5) We receive a signal each time we poll messages, even if no messages are actually retrieved (before we would not send a signal if no messages are available, minor point, but might help a bit with performance)

I haven't checked in develop if they are present but I'd assume no, I can take a look at some of these at some point as well, but maybe you could have a look, thanks!

adambabik:
> Sending any direct message results in chat ID is empty

This assumes that there is a `ChatID` specified by the caller. It is defined now with a method `func (m SendDirectMessageRPC) ID() string { return m.Chat }`. Any idea why `SendDirectMessageRPC.Chat` is not specified? We can remove this requirement when using `SendRaw` because it is required when creating a transit-encoded message from scratch.

> Opening a one-to-one chat results in

Can you tell me which JSON-RPC method is run? 

> We receive a signal each time we poll messages, even if no messages are actually retrieved (before we would not send a signal if no messages are available, minor point, but might help a bit with performance)

I will fix this.

adambabik:
@cammellos 
1. Not sure how this works. Is it related to `shhext_confirmMessagesProcessedByID`?
2. Could it be a problem with not listening on the discovery topic? It's turned off now. I need to add `protocol.WithGenericDiscoveryTopicSupport()` option for `protocol.NewMessenger()` constructor in order to enable it.

3, 4 and 5 are fixed.

cammellos:
1) I figured out why messages are not marked as `Sent`
The reason is that we should use `api.Post` https://github.com/status-im/status-go/blob/19c26b129dc0a81caa10923fb3b8b36e2da05535/services/shhext/api.go#L192 for sending whisper messages, so `SendRaw` should be not be sending the message but returning the actual whisper newMessage, then we can call `api.Post`, similarly as it was done in  https://github.com/status-im/status-go/blob/19c26b129dc0a81caa10923fb3b8b36e2da05535/services/shhext/api.go#L486 .
Alternatively we could bring the tracking code (`EnvelopesMonitor`) in `status-protocol-go` but that looks like much bigger task. Would you mind having a look at this? Thanks!

Still not sure about `ConfirmMessagesProcessed` I don't think it's related, possibly something different.

cammellos:
I also figured out 
> Not sure how this works. Is it related to shhext_confirmMessagesProcessedByID?

Basically we use the `DeduplicationID` to confirm messages, and that was passed to the encryption layer (we need to confirm the processing to remove key material). This has been changed to use the `Hash` of the whisper message, so when trying to confirm messages it won't find it an not key material will be removed, so that's definitely something we need to address for the security audit.

Probably is ok to do it in a separate task, as we might want to change how things work, as passing the dedup id is a bit cumbersome and there's no real reason to use that, any ID is fine as long as it's passed to the client.

adambabik:
> Alternatively we could bring the tracking code (EnvelopesMonitor) in status-protocol-go but that looks like much bigger task. 

Yeah, there should be a more generic solution figured out for stuff like deduplication and this monitor. Something like pre-hooks and post-hooks that can be provided by the caller but the order of execution would be handled by status-protocol-go.

(1) is fixed in https://github.com/status-im/status-go/pull/1526/commits/c2ce4a438d6ad4dd7708138ada4081af631c4b04.

services/shhext/rpc.go
```diff
@@ -4,21 +4,63 @@
...
+
+func (m JoinRPC) PublicName() string {
+	if len(m.PubKey) > 0 {
+		return ""
+	}
+	return m.Chat
+}
+
+func (m JoinRPC) PublicKey() *ecdsa.PublicKey {
+	if len(m.PubKey) > 0 {
```
pedropombeiro, line 61: @adambabik isn't this logic reversed?

PR #1527, opened 2019-07-17
Change to support status-react #8136 [Onboarding] Recovery

pacamara:
Change to support https://github.com/status-im/status-react/issues/8136 

If RecoverAccount called with empty password, delete recovered key from keystore immediately.
This enables testing recoverability of an account before user supplies a password, per the new UX flow in https://github.com/status-im/status-react/issues/8136


dmitryn:
Closing as Recovery was implemented without these changes.

PR #1530, opened 2019-07-22
Accounts data management

### LINKED ISSUE #1552: Store sub accounts information in status-go
# Problem
With multiaccoutns functionality user can derive mutliple keys to use with the status account.
Types:

* main (when it is used?)
* wallet (used for signing transactions)
* watch (transactions history will be watched for this type)
* another type?

Main and wallet accounts should be unique, e.g. you can't choose another main without disabling previously selected.

This information will be used when user logs in to fetch required sub accounts from database and pass them to components that needs them. For instance, wallet address will be passed to accounts manager, and watch address will be passed to transfers watcher.

# Implementation
We will add one table that will store list of addresses with types that are specified above.
And api with following methods:

## accounts_saveSubAccounts
Accepts list of accounts, e.g.

```json
[
{
"address": "0x0100000000000000000000000000000000000000",
"main": true,
"wallet": true,
"watch": false
},
{
"address": "0x0200000000000000000000000000000000000000",
"main": false,
"wallet": false,
"watch": true
}
]

```

Preserves constrains specified above.

## accounts_getSubAccounts
returns all accounts in the format above.

# Acceptance Criteria
Account list stored in status-go, used during login and can be modified.

@flexsurfer @yenda take a look please, especially i need to clarify types and api

dshulyak:
closes: https://github.com/status-im/status-go/issues/1552


dshulyak:
@yenda i think you can give it a try. PR is not finished yet, but API is ok for first version.

There is a new endpoint `OpenAccounts(datadir string)` that returns list of accounts, like

```json
[{"name:"Test", "address": "0xhex"}]
```

And `SaveAccountAndLogin(accountData, password, configJSON string)`.
This second endpoint needs to be called only once when account is created. It will store account and node config in db and login after that.

Login now additionally starts the node, and Logout stops the node.

dshulyak:
Configuration will probably stay in unencrypted database. When user logs in i need to know datadir to open an encrypted db, but currently datadir is stored in node config, so in order to login the configuration must be already fetched. It can be fixed by splitting configuration into private and public information, but i prefer not to it and will keep config in plain sql.

dshulyak:
@yenda you can use this branch once you can proceed with accounts integration. i will try to prepare documentation for API changes, but if you will start sooner than that then just ping me

yenda:
@dshulyak starting now, will ping you if I need help

yenda:
@dshulyak I forgot that we need a `photo-path` String field that contains the base64 encoded profile picture of the user in the multiaccount db

yenda:
lgtm

yenda (approved):

dshulyak:
@yenda so you already have necessary changes for status-react? i think it is better to merge this change when status-react will be ready to merge, otherwise, status-go will be temporarily incompatible with react

yenda:
@dshulyak @cammellos https://github.com/status-im/status-react/pull/8725#issuecomment-523000966 can be merged please merge

Issue #1533, opened 2019-07-23
Merge logs from status-protocol-go into geth.log


adambabik:
# Problem

status-protocol-go uses a different logger which logs to stderr by default. In status-go, we collect all logs into a file pointed in the config: https://github.com/status-im/status-go/blob/develop/params/config.go#L307-L308.

We need to combine these two loggers into a single log stream for debug purposes.

# Implementation

TBD

# Acceptance Criteria

* status-protocol-go logs are collected together with status-go logs.

adambabik:
Closing with https://github.com/status-im/status-go/pull/1526/commits/94ba02724ea22fa00ff6d9dfce2cfded835fb07e.

PR #1534, opened 2019-07-24
add an adapter for zap and geth loggers

adambabik:
status-protocol-go uses https://github.com/uber-go/zap as a logger. status-go uses geth's log package. These two must work together so that any logs created by status-protocol-go are visible in status-go binaries.

Relates to #1533


cammellos (approved):

dshulyak:
Why not ethereum logger though? :) app seems small enough to use unified logger across the stack

adambabik:
@dshulyak geth logger is not a separate project but rather an internal thing that can change any time. We wanted a lightweight, stable, configurable logger in status-protocol-go and be independent of geth internal packages as much as possible.

PR #1542, opened 2019-07-25
add multi-account LoadAccount, ImportMnemonic, and Reset functions

gravityblast:
Expose multi account  `LoadAccount` , `ImportMnemonic`, and `Reset` methods.

### DIFF
--- a/account/generator/README.md
+++ b/account/generator/README.md
@@ -1,0 +1,43 @@
+# Account Generator
+
+The Account Generator is used to generate, import, derive child keys, and store accounts.
+It is instantiated in the `account.Manager` struct and it's accessible from the `lib` and `mobile`
+package through functions with the `MultiAccount` prefix:
+
+* MultiAccountGenerate
+* MultiAccountGenerateAndDeriveAddresses
+* MultiAccountImportMnemonic
+* MultiAccountDeriveAddresses
+* MultiAccountStoreDerivedAccounts
+* MultiAccountImportPrivateKey
+* MultiAccountStoreAccount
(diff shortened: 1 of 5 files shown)


gravityblast:
@flexsurfer I added here the Load and Reset functions.
You need to call reset in case you load an account and then the user cancel the action instead of derive and store a new account.

flexsurfer:
@gravityblast could you elaborate on why we need LoadAccount ? how can it be used?

flexsurfer:
and what about generate ? if user generates 5 accounts, goes back and generate 5 more ? do we need to reset when he goes back ?

gravityblast:
@flexsurfer yes I'll write everything in a readme file so that it can always be in the repo and I'll post it here

gravityblast:
@flexsurfer 
> and what about generate ? if user generates 5 accounts, goes back and generate 5 more ? do we need to reset when he goes back ?

All the multi-account methods are used to generate derive etc, but only StoreDerived and StoreAccount are actually saving keys to the filesystem.
You can call generate as many time as you want. but if you restart the application the keys are gone, they were just in memory until you store them.
After storing them, you don't need to multi-account functions anymore, until you want to load one to derive more child keys.

gravityblast:
you don't need to reset, but you could do it. you can just use the new 5  ids, but there will be 10 keys in memory.

gravityblast:
@flexsurfer , I added a README file, can you check it and tell me if it makes sense please?

adambabik (approved):

PR #1545, opened 2019-07-26
Don't omit history index when the value is 0

### LINKED ISSUE #1544: history-index is not returned by browser service
# Problem

history-index key is not returned by browser service when calling `browsers_getBrowsers`

# Acceptance Criteria

history-index key is returned as well

dshulyak:
closes: https://github.com/status-im/status-go/issues/1544

### DIFF
--- a/services/browsers/database.go
+++ b/services/browsers/database.go
@@ -35,7 +35,7 @@ type Browser struct {
 	Name         string   `json:"name"`
 	Timestamp    uint64   `json:"timestamp"`
 	Dapp         bool     `json:"dapp?"`
-	HistoryIndex int      `json:"history-index,omitempty"`
+	HistoryIndex int      `json:"history-index"`
 	History      []string `json:"history,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


services/browsers/database.go
```diff
@@ -35,7 +35,7 @@ type Browser struct {
 	Name         string   `json:"name"`
 	Timestamp    uint64   `json:"timestamp"`
 	Dapp         bool     `json:"dapp?"`
-	HistoryIndex int      `json:"history-index,omitempty"`
+	HistoryIndex *int     `json:"history-index,omitempty"`
```
adambabik: Wouldn't it be enough to remove `omitempty`? Or we still need to distinguish between `0` and no history index?

dshulyak: i think it is correct to distinguish between 0 and no history. history index should be consistent with the history array. e.g. if there is no array or no items in array - history index 0 is confusing

adambabik: In theory, no `history` might be considered that there is no history instead of looking at the index. I guess that would be the same? Just `*int` looks odd in this case but all is correct :)

dshulyak: i guess it makes sense, a consumer will have to double check that index is not out of bounds anyway

gravityblast (approved):

PR #1546, opened 2019-07-26
Initialize keystore using keydir passed from a client

### LINKED ISSUE #1543: Account manager shouldn't depend on started ethereum node
# Problem
Since we are moving configuration to status-go, we will be able to start a node only after user provided password. Configuration stored in encrypted database and we want to open it only once on actual login. However some functionality that requires key generation depends on node to be started. Particularly status-go account manager requires these two objects to be available when user imports generated key:

```
type GethServiceProvider interface {
AccountManager() (*accounts.Manager, error)
AccountKeyStore() (*keystore.KeyStore, error)
}
```

The only thing we need to make them decoupled from geth node is to instantiate them ourselves using correct keydir. And copy some code from ethereum: https://github.com/ethereum/go-ethereum/blob/master/node/config.go#L466-L528.

# Implementation
Instantiate account manager and account keystore using keydir and store them on the account manager object. Keydir can be provided by status-react on a first native binding call to status-go, for example using `OpenAccounts` call.

# Acceptance Criteria
It should be possible to use account manager without started ethereum node.

@gravityblast @adambabik what do you think about this proposal?
cc @yenda

dshulyak:
closes: https://github.com/status-im/status-go/issues/1543


account/accounts.go
```diff
@@ -55,15 +51,29 @@ type Manager struct {
...
 // NewManager returns new node account manager.
-func NewManager(geth GethServiceProvider) *Manager {
-	manager := &Manager{
-		geth: geth,
-	}
-
-	accountsGenerator := generator.New(manager)
-	manager.accountsGenerator = accountsGenerator
+func NewManager() *Manager {
+	m := &Manager{}
```
adambabik, line 54:
```suggestion
	m := &Manager{
		accountsGenerator: generator.New(m),
	}
```

adambabik (approved):

account/accounts.go
dshulyak: there is a circular reference, generator.New takes pointer to Manager as an input

yenda:
Amazing! Does this mean I can retest and this time shouldn't need a running node to generate account?

dshulyak:
@yenda i had some issues but now i think you can use this change. There is additional binding with signature `InitKeystore(keydir string) string`. It must be called before Login and StartNode.

We probably will use OpenAccounts to open dbs and initialize keystore at the same time but i wanted to make this change decoupled from accounts work.

yenda:
Ok so this morning I am going to integrate this on current status-react, my expectations is that I should be able to use InitKeystore(keydir string) string instead of StartNode when the app starts

dshulyak:
@yenda yes that should be expected. i also integrated this change into the branch https://github.com/status-im/status-go/pull/1530. but it would be awesome if you can integrate it without that branch so i can merge this one earlier

yenda:
@dshulyak lgtm, I integrated and tested on iOS and Android, minor detail I used the rootDir as param, which works, but should I use the keystoreDir instead? seems like it's just rootDir + "keystore" appended so not a big change

dshulyak:
@yenda i would prefer to use separate directory for keystore. using rootdir probably won't cause any issues (e.g. overlapping files) but separate directory just seems cleaner and safer to me

PR #1553, opened 2019-07-30
Conslidate protocol databases

adambabik:
This change uses a single database file for the protocol database.

Todo:
- [x] update status-protocol-go version.


cammellos (approved):

cammellos:
Thanks!

Currently nightlies have a bug as `status-protocol-go` creates a non namespaced `messages.sql`, and so creating two accounts fails, as they won't have the same key.

This will fix it, so we might want to get it merged as soon as possible so we can attempt to fix nightly, also outstanding prs needs to be updated.

Issue #1537, opened 2019-07-24
Refactor wallet database to store data for multiple networks


dshulyak:
# Problem

Having a separate wallet db is necessary because the application was designed to store data from a single network. Storing transfers from multiple networks will bork the whole flow.

But also separate wallet db incurs overhead when a node is started (~500ms with data in it). So it will be beneficial for an app if wallet db can be merged with other databases.

# Implementation

- Add network id to relevant tables in wallet database
- Ensure that network id is used for all relevant queries when wallet service works with wallet database
- adjust uniqueness indexes

# Acceptance Criteria

Data from multiple networks can be stored in the same wallet db.

dshulyak:
was closed with https://github.com/status-im/status-go/pull/1564

PR #1540, opened 2019-07-25
Consolidate browsers/permissions/accounts/wallet db in a single database

dshulyak:
closes: https://github.com/status-im/status-go/issues/1536

PR #1564, opened 2019-08-08
 Wallet database isolated by the network id

dshulyak:
Wallet database refactored so that every query ensures isolation by the network id.
Network id provided when database object is created, thus it is transparent to other parts
of the wallet module.

Also every uniqueness index is changed to ensure that it doesn't prevent adding
object with same id but from a different network.

closes: https://github.com/status-im/status-go/issues/1537


adambabik (approved):

dshulyak:
Included in https://github.com/status-im/status-go/pull/1540

PR #1568, opened 2019-08-09
Add fields necessary for a login with keycard

### LINKED ISSUE #1567: Account fields for keycard available before login
# Problem

Based on https://github.com/status-im/status-react/pull/8725 and conversation with @yenda, there is a set of Account fields which are accessible before login (like address, photo and name).

Login with keycard requires additional fields from Account entity to be accessible before login:

* `keycard-pairing` String
* `keycard-key-uid` String

dshulyak:
closes: https://github.com/status-im/status-go/issues/1567


dshulyak:
@dmitryn @yenda @gravityblast added fields will be stored in unencrypted database. are we sure that it is fine?

dmitryn:
>added fields will be stored in unencrypted database. are we sure that it is fine?

I think so. Unless you steal physical card, there is not much you can do with keycard-pairing and keycard-key-uid data alone.
/cc @bitgamma

bitgamma:
Yeah the pairing is only needed for MITM protection. It is not great if it is leaked, so if it were encrypted using the Android facilities it would be better. Some (most?) phone should have a secure element to store credentials. However the pairing key does not provide any access to funds on card so it is not critical. You need the card and the PIN.

dshulyak:
Thanks for comments, we don't have the tools right now to get key material from secure element. I will open an issue to track that

adambabik (approved):

PR #1569, opened 2019-08-13
Watch new accounts aftter they were saved to accounts table

### LINKED ISSUE #1566: Wallet should be tracking accounts created after login
# Problem

Any account that is created after user logged in won't be tracked by a wallet service. Currently list of accounts is passed when user logs in and after that there is no way to add more accounts to a wallet service.

# Implementation

Once https://github.com/status-im/status-go/pull/1530 is merged we will have a service with an rpc method `accounts_saveAccounts`. When it is called we must notify wallet service. Wallet will check for any new accounts and restart wallet.Reactor with new list of accounts. Note that we don't watch every account separately so whole reactor needs to be restarted.

Adding new account to wallet service shouldn't block `accounts_saveAccounts` method. There should be a simple component in wallet service that manages reactor updates and provides a communication channel to other services. Accounts can be passed either using a channel to a service or triggering a channel after accounts stored in db.

# Acceptance Criteria

New accounts can be added to a wallet tracker after user logged in.

@flexsurfer does it sound fine to you if we will use accounts_saveAccounts to start listening for new addresses in the wallet?

dshulyak:
After accounts_saveAccounts persisted accounts in the db successfully it will notify wallet service that it can start watching for new addresses. 
Wallet will restart whole reactor component to ensure that every address will have the same head of the chain after new accounts are synced.

closes: https://github.com/status-im/status-go/issues/1566

### DIFF
--- a/api/backend.go
+++ b/api/backend.go
@@ -15,6 +15,7 @@ import (
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	ethcrypto "github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/ethereum/go-ethereum/event"
 	"github.com/ethereum/go-ethereum/log"
 	gethnode "github.com/ethereum/go-ethereum/node"
 	"github.com/ethereum/go-ethereum/p2p/enode"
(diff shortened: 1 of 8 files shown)


yenda:
I integrated it in my branch and it works nicely

yenda (approved):

yenda:
@dshulyak actually I think I found an issue, seems like when I send a transaction from one account to the other, I only receive a transfer for one of the account (the :wallet true one) and not the other. It is fine for token transfers but not for eth

yenda:
Also I still get this error:
https://github.com/status-im/status-react/issues/8759

even though the account is saved and the key is stored (but it works after relogin)

dshulyak:
> @dshulyak actually I think I found an issue, seems like when I send a transaction from one account to the other, I only receive a transfer for one of the account (the :wallet true one) and not the other. It is fine for token transfers but not for eth

thanks, there is definitely a bug. i will fix it in a separate pr

> https://github.com/status-im/status-react/issues/8759

not sure what is the issue there, will take a look

yenda:
> > [status-im/status-react#8759](https://github.com/status-im/status-react/issues/8759)
> 
> not sure what is the issue there, will take a look

Note that after logout/login the issue is gone

dshulyak:
> > > [status-im/status-react#8759](https://github.com/status-im/status-react/issues/8759)
> > 
> > 
> > not sure what is the issue there, will take a look
> 
> Note that after logout/login the issue is gone

ok,so there is already an issue https://github.com/status-im/status-go/issues/1575

Issue #1562, opened 2019-08-07
Share the database with Messenger from status-protocol-go


adambabik:
# Problem

We have at least three sqlite databases being opened when the app starts:
- db for browsers
- db for permissions
- db for wallet.

Each separate encrypted sqlite database file requires non-trivial amount of time to be opened. This slows down the startup of the application.

# Implementation

1. Open a single encrypted sqlite database file which will be shared between various services and libraries.
2. Collect migrations in one place similarly to https://github.com/status-im/status-protocol-go/blob/master/migrations.go and apply in one go before the app passes the initial setup. This is to detect any problems early.

Because there is no separation of database files between individual components, there is still a threat that two components define tables with the same names. This can be mitigated to some extent by always applying all migrations for unit and integration tests of the components or at least for an object managing the state of a component.

Another problem is a table storing the migrations metadata. By default, [`schema_migrations`](https://github.com/status-im/migrate/blob/master/database/sqlite3/sqlite3.go#L21) is used. However, when a single database is shared between different libraries, like in our case where a single database is shared between status-go and status-protocol-go, each library should use a different migrations table to avoid conflicts.

# Acceptance Criteria

* A single database instance and database file is shared between status-go services and status-protocol-go library.
* Migrations are properly applied in status-go services as well as in libraries using custom migrations tables to avoid conflicts.
* status-go individual components do not open a database. They receive a `*sql.DB` instance only.
* All unit and integrations tests relying on a database should use one with all migrations from a given library applied.

cc @dshulyak @cammellos

dshulyak:
Well this is a duplicate. There is this issue https://github.com/status-im/status-go/issues/1536 which is closed by https://github.com/status-im/status-go/pull/1540. I am just waiting for Eric to merge integrate main part

dshulyak:
Also wallet is not straightforward to merge with others. Cause there is network specific information. So this needs to be solved https://github.com/status-im/status-go/issues/1537

adambabik:
@dshulyak I was not aware of that PR. It looks good but there are a few things missing described in this issue:
- no custom name for the migrations table,
- `Messenger` from status-protocol-go also needs to share this instance.

dshulyak:
you are right those things are not in the pr

adambabik:
That's cool, though. We can do it in a separate one, it makes more sense and provides better visibility.

PR #1570, opened 2019-08-13
Hash password before passing it as a pragma

dshulyak:
issue in react: https://github.com/status-im/status-react/issues/8702

### DIFF
--- a/sqlite/sqlite.go
+++ b/sqlite/sqlite.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"fmt"
 
+	"golang.org/x/crypto/sha3"
+
 	_ "github.com/mutecomm/go-sqlcipher" // We require go sqlcipher that overrides default implementation
 )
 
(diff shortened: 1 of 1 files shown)


cammellos:
@dshulyak looks good to me, one thing, how does this fit with  https://github.com/status-im/status-go/issues/1562 ? if we will share the database we then won't need to do this here I take (or at leas we should be using the same hashing :) )

dshulyak:
> if we will share the database we then won't need to do this here I take 

@cammellos we will have to do it somewhere. why not here? 
as i understand we will need to open database and then pass it to appdatabase module and protocol database module where migrations will be applied. if you want to use another sqlite module (from protocol lib) i definitely have nothing against it, but technically i don't see any difference

cammellos:
@dshulyak sure by all means, I meant that probably we might want to handle the two issues together, but it's all fine by me. Seems like though this might not be strictly necessary, as status-react will pass the hashed password, https://github.com/status-im/status-react/issues/8702#event-2553731323 , what do you think? I guess we could still keep it, no harm in it

dshulyak:
that makes sense. one thing is that ethereum accounts manager, that we are using in account module, was using password for kdf. but if we are ignoring backward compatibility it can just start using hash i guess

dshulyak:
@cammellos as i understand this is an optional change, please let me know if i misunderstood smth

PR #1571, opened 2019-08-18
Add Metadata to messages, expose new messenger methods

cammellos:
This commits modifies deduplicator so that it takes a `StatusMessage`
instead of `WhisperMessage` and also returns a `Metadata` field which is
then passed back by the client when confirming messages, which fixes the
issue we had with not confirming pfs messages.


services/shhext/service.go
```diff
@@ -235,23 +235,44 @@ func (s *Service) retrieveMessagesLoop(tick time.Duration, cancel <-chan struct{
...
+
+			for _, messages := range chatWithMessages {
+				for _, message := range messages {
+					messageIDs = append(messageIDs, message.ID.String())
+				}
+			}
+
+			existingMessages, err := s.messenger.MessagesExist(messageIDs)
+			if err != nil {
+				log.Error("failed to chek existing messages", "err", err)
```
pedropombeiro:
```suggestion
				log.Error("failed to check existing messages", "err", err)
```

services/shhext/service.go
```diff
@@ -235,23 +235,44 @@ func (s *Service) retrieveMessagesLoop(tick time.Duration, cancel <-chan struct{
...
 				signalMessage := &signal.Messages{
 					Chat:     chat,
 					Error:    nil, // TODO: what is it needed for?
-					Messages: s.deduplicator.Deduplicate(retrievedMessages),
+					Messages: s.deduplicator.Deduplicate(messages),
+				}
+
+				var newMessages []dedup.DeduplicateMessage
+
+				// Remove already saved messages
```
pedropombeiro: The comment says we're removing messages, but the method below is appending messages, which is a bit confusing. Maybe the comment could be clearer?

pedropombeiro (approved):

services/shhext/dedup/deduplicator.go
```diff
@@ -2,6 +2,9 @@ package dedup
 
 import (
 	"github.com/ethereum/go-ethereum/log"
+
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	protocol "github.com/status-im/status-protocol-go/v1"
```
adambabik: I would avoid leaking this dependency here and instead...

services/shhext/dedup/deduplicator.go
```diff
@@ -35,18 +44,26 @@ func NewDeduplicator(keyPairProvider keyPairProvider, db *leveldb.DB) *Deduplica
...
 // specified filter.
-func (d *Deduplicator) Deduplicate(messages []*whisper.Message) []DeduplicateMessage {
+func (d *Deduplicator) Deduplicate(messages []*protocol.StatusMessage) []DeduplicateMessage {
+
 	result := make([]DeduplicateMessage, 0)
 	selectedKeyPairID := d.keyPairProvider.SelectedKeyPairID()
 
 	for _, message := range messages {
-		if has, err := d.cache.Has(selectedKeyPairID, message); !has {
+		whisperMessage := message.TransportMessage
```
adambabik: ...do this transformation by the caller. Then we could have the signature like this:
```go
func (d *Deduplicator) Deduplicate(messages []*DeduplicateMessage)
```

and just do this part:
```go
// redacted
    // redacted
    message.Metadata.DedupID = d.cache.KeyToday(selectedKeyPairID, message.Message)
}
```

EDIT: actually, we can't because not all messages are returned. So the signature would need to look like this at best: 
```go
func (d *Deduplicator) Deduplicate(messages []*DeduplicateMessage) []*DeduplicateMessage
```

adambabik (approved):

services/shhext/dedup/deduplicator.go
cammellos: done!

dshulyak:
@cammellos are you going to merge this/any other changes today? Eric told me that his accounts branch is ready to be merged, so i plan to merge and release necessary status-go changes. it is still possible that status-go won't be compatible for some period, so maybe it makes sense to coordinate other merges

cammellos:
@dshulyak I believe it's still under testing https://github.com/status-im/status-react/pull/8725 , but as soon as it's tested successfully feel free to merge , @yenda can quickly then update `status-go` with the hash from develop if necessary. I won't merge my changes until https://github.com/status-im/status-react/pull/8763 is also ready to merge (I will follow same procedure, if pr in status-react is tested ok, I will merge this in, update status-react pr with new hash, and merge that), does that sound good to you?

dshulyak:
sounds good, i guess new release is indeed not necessary

cammellos:
@dshulyak @yenda I have merged this PR and https://github.com/status-im/status-react/pull/8763 , as tested successfully, you might have to rebase against this changes

PR #1577, opened 2019-08-21
Make stop node part of the logout blocking

dshulyak:
Caller needs to be sure that previous node is stopped and all previous state is cleared before next login. 

As alternative we can add a signal that node is stopped and the caller will have to wait for that signal.

### DIFF
--- a/lib/library.go
+++ b/lib/library.go
@@ -404,8 +404,7 @@ func Logout() *C.char {
 	if err != nil {
 		makeJSONResponse(err)
 	}
-	api.RunAsync(statusBackend.StopNode)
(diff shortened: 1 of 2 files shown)


yenda (approved):

adambabik (approved):

dshulyak:
@yenda is it safe to merge this change? will you make a necessary adjustements on status-react side?

yenda:
I already made a PR that just update status-go and it works without changes. I could still reproduce the bug we were trying to fix but I didn't have time yet to check the logs and see if the cause is still the same. I'd say this is good to merge already anyway

PR #1587, opened 2019-08-28
Statefull login with keycard

### LINKED ISSUE #1576: LoginWithKeycard should start node
# Problem

`LoginWithKeycard` now results in `{\"error\":\"there is no running node\"}`

# Implementation

Start the node like during `Login` call https://github.com/status-im/status-go/blob/be9c55bc16ca035ca9bfc0586b6c9298a91bdcd4/lib/library.go#L356

dshulyak:
closes: https://github.com/status-im/status-go/issues/1576

@gravityblast please a take a look if this change make sense. i plan to be mainly unavailable starting from next week, so it would be good if we can resolve this issue before that)

The proposal is to call `SaveAccountAndLoginWithKey(accountData, password, configJSON, keyHex string)` on first login (when user created account),

And then use `LoginWithKeycard(accountData, password, keyHex string)` for regular login.

accountData is a multiaccount struct, the same that is used for login flow without keycard.
configJSON is node configuration json generated from default by status-react.
password is anything that can be used for secure database encryption (i don't know if accounts with keycard are using passwords)
keyHex is a chat key.

@dmitryn does it work for you?

### DIFF
--- a/mobile/status.go
+++ b/mobile/status.go
@@ -376,11 +376,50 @@ func InitKeystore(keydir string) string {
 	return makeJSONResponse(err)
 }
 
+// SaveAccountAndLoginWithKeycard saves account in status-go database..
+func SaveAccountAndLoginWithKeycard(accountData, password, configJSON, keyHex string) string {
+	var account multiaccounts.Account
+	err := json.Unmarshal([]byte(accountData), &account)
+	if err != nil {
+		return makeJSONResponse(err)
+	}
+	var conf params.NodeConfig
+	err = json.Unmarshal([]byte(configJSON), &conf)
+	if err != nil {
+		return makeJSONResponse(err)
+	}
+	api.RunAsync(func() error {
+		log.Debug("starting a node, and saving account with configuration", "address", account.Address)
+		err := statusBackend.SaveAccountAndStartNodeWithKey(account, &conf, password, keyHex)
+		if err != nil {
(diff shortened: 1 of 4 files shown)


mobile/status.go
```diff
@@ -376,11 +376,50 @@ func InitKeystore(keydir string) string {
...
+	})
+	return makeJSONResponse(nil)
+}
+
 // LoginWithKeycard initializes an account with a chat key and encryption key used for PFS.
 // It purges all the previous identities from Whisper, and injects the key as shh identity.
-func LoginWithKeycard(chatKeyData, encryptionKeyData string) string {
-	err := statusBackend.InjectChatAccount(chatKeyData, encryptionKeyData)
-	return makeJSONResponse(err)
+func LoginWithKeycard(accountData, password, keyHex string) string {
```
gravityblast, line 406: does a keycard user have a password too?

gravityblast:
@dshulyak LGTM so far, I just added a question

gravityblast (approved):

mobile/status.go
dshulyak: i think so, basically i need any string to use for database encryption. i assume the situation was the same for realm

dmitryn (commented):
Thanks @dshulyak That should work

dmitryn:
Would like to have this PR merged so i can try it on status-react side

PR #1589, opened 2019-08-28
update status-protocol-go

adambabik:
There is one braking change. `shhext_chats` takes no arguments and returns all chats always.


cammellos:
@adambabik this has now been already included, currently develop points to top of master in status-protocol-go, I think it's safe to close this one.

PR #1595, opened 2019-09-02
Add mailservers service

### LINKED ISSUE #1578: Service to store mailservers
# Problem

To remove realm from status-react, we need a service to save and get mailservers associated with an account

# Implementation

```
{:id        :string
:name      {:type     :string}
:address   {:type     :string}
:password  {:type     :string :optional true}
:fleet     {:type     :string}}
```

We need the following rpc methods:
   "mailservers_addMailserver"
   "mailservers_getMailservers"
   "mailservers_deleteMailserver"

# Acceptance Criteria

Service is integrated on status-react and replaces the current realm table

adambabik:
This change adds a mailservers service which exposes three new JSON-RPC methods: `mailservers_addMailserver `, `mailservers_getMailservers` and `mailservers_deleteMailserver`.

There is no new migration created. If we already released a version with browsers, permissions etc., a new migration must be created. (@yenda do you know if that happened?)

Closes #1578


appdatabase/migrations/sql/0001_app.up.sql
```diff
@@ -75,3 +75,11 @@ sync INT,
 FOREIGN KEY(network_id,blk_number) REFERENCES blocks(network_id,number) ON DELETE CASCADE,
 CONSTRAINT unique_mapping_for_account_to_block_per_network UNIQUE (address,blk_number,network_id)
 );
+
+CREATE TABLE IF NOT EXISTS mailservers (
+    id VARCHAR PRIMARY KEY,
+    name VARCHAR NOT NULL,
```
adambabik, line 81: @yenda should `name` be unique?

cammellos (approved):

appdatabase/migrations/sql/0001_app.up.sql
yenda: it wasn't in realm

yenda: regarding migrations no need

PR #1596, opened 2019-09-03
Add Mailserver Request Gaps service

cammellos:
This commits add a new service used to store mailserver request gaps,
and provides 4 endpoints:

```
AddMailserverRequestGaps(gaps []MailserverRequestGap) error
GetMailserverRequestGaps(chatID string) []MailserverRequestGap
DeleteMailserverRequestGaps(ids []string) error
DeleteMailserverRequestGapsByChatID(id string) error
```


adambabik (approved):

adambabik:
@cammellos actually, I think we can merge these JSON-RPC methods into `mailservers` service. In the issues, they all have prefix `mailservers_*`. We would avoid creating that many services and I think all of these JSON-RPC methods are needed or not, there is no subset that is ever useful.

cammellos:
@adambabik sure, do you mean having all the code under `mailservers` namespace or rather just have the endpoint exposed as `mailserver_...` ? I am fine with both, don't have a strong opinion

adambabik:
@cammellos thought about having code in `services/mailservers` so one namespace/package/directory. This implies exposing all endpoints as `mailservers_XXX`.

cammellos:
@adambabik sounds good, will do that

Issue #1520, opened 2019-07-09
Verify ENS name client side


rachelhamlin:
Moving from react [#8541](https://github.com/status-im/status-react/issues/8541)

## User Story
As a user, I want to make sure ENS name displayed in chat and profile has been verified.

## Description
**Type:** Feature

ENS name is sent as part of chat messages and account details. It is displayed as is and could be spoofed.

Proper solution involves checking the ENS validity client side. For performance reason it's been agreed that it should be done status-go side.

rachelhamlin:
Is the required change clear @PombeirP?

If it's straightforward enough, perhaps we could get it done in the next week. If so, we can include all of @jeluard's ENS work in the upcoming release (native registration + default display in chat).

pedropombeiro:
@rachelhamlin I'd just add the details @jeluard mentioned, i.e. that the data that is returned from status-go will omit the ENS name if it couldn't be verified.

jeluard:
@PombeirP Some more details: the message payload contains a `:content` map, itself optionally containing a `:name` string. This name is the full ENS name (e.g. `julien.stateofus.eth`, `pombeirp.eth`).

At the status-react level we need to know if this name is really controlled by the message sender. So the payload should be modified in such a way that the `:name` string can be trusted. The simplest would be to just drop it if it has been spoofed.

Checking for name validity is a simple ENS contract `call` (I can provide guidance if needed).

rachelhamlin:
@flexsurfer you mentioned that we need this for the code freeze. Still true based on today's discussion?

@PombeirP could we assign someone to this?

pedropombeiro:
I think I should be able to take it.

flexsurfer:
i'm not sure about code freeze, but it must be in v1 for sure if we want ens names in chats, but it's blocked by chat api, we can implement only after chat api will be done on go side

rachelhamlin:
Noted. Should we consider this a blocker to releasing ENS resolution @flexsurfer @jeluard?

jeluard:
It's definitely a security concern. Not entirely clear if it should prevent ENS being released in the pre-v1 release. 
@corpetty WDYT?

Also note that the check could be implemented status-react side, if needed.

rachelhamlin:
Note: we can get a contributor to hide the ENS chat resolution feature for beta release v0.14, so this does not need to be done before that. Will come back later on the level of priority for v1.

pedropombeiro:
[...]

PR #1620, opened 2019-09-27
Add VerifyENSName endpoint

pedropombeiro:
This PR adds the `VerifyENSName` endpoint, to be used in https://github.com/status-im/status-go/issues/1520 to add some guarantees on user name authenticity.

It imports a new library to create namehash and invoke the `Pubkey` contract method  https://github.com/wealdtech/go-ens.


pedropombeiro:
The test that is failing doesn't seem to be related to the changes in this PR.

api/backend.go
```diff
@@ -555,6 +558,53 @@ func (b *StatusBackend) GetNodesFromContract(rpcEndpoint string, contractAddress
...
+func (b *StatusBackend) VerifyENSName(ensName, publicKeyStr, rpcEndpoint, contractAddress string) (bool, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), contractQueryTimeout)
+	defer cancel()
+
+	pubKeyBytes, err := hex.DecodeString(publicKeyStr)
+	if err != nil {
+		return false, err
+	}
+
+	expectedPubKey, err := btcec.ParsePubKey(pubKeyBytes, btcec.S256())
```
adambabik: The same can be done with `crypto.UnmarshalPubkey()` where `crypto` is `github.com/ethereum/go-ethereum/crypto`.

api/backend.go
```diff
@@ -555,6 +558,53 @@ func (b *StatusBackend) GetNodesFromContract(rpcEndpoint string, contractAddress
...
+		return false, err
+	}
+	x, y, err := resolver.PubKey()
+	if err != nil {
+		b.log.Error("error while resolving public key from ENS name", "ensName", ensName, "error", err)
+		return false, err
+	}
+
+	// Assemble the bytes returned for the pubkey
+	pb := make([]byte, btcec.PubKeyBytesLenUncompressed)
```
adambabik: I would try to stick to `github.com/ethereum/go-ethereum/crypto`. Which means doing something like this:
```go
// crypto/elliptic
// github.com/ethereum/go-ethereum/crypto
pubKeyBytes := elliptic.Marshal(crypto.S256(), new(big.Int).SetBytes(x), new(big.Int).SetBytes(y))
return bytes.Equal(expectedPubKeyBytes, pubKeyBytes)
```

yenda:
Is it an just an endpoint or does it also verify on new messages?

pedropombeiro:
> Is it an just an endpoint or does it also verify on new messages?

For now this is just an enpoint. However I think I'll need to move this code to status-protocol-go and do the whole implementation there. Neither status-go nor status-react should care about verifying the authenticity of the sender.

PR #1625, opened 2019-10-02
add script for easier setup of mailserver

jakubgs:
People often complained that running a mailserver is too hard, some of those people complained that they don't want to or can't use Docker, so this is an attempt at providing and easy way of running `statusd` as a user `systemd` service.

The main thing it adds is a `make run-mailserver-systemd` which builds, configures, and starts a `statusd` mailserver instance using a `systemd` user service definition.

Not entirely sure if I put the `sdnotify.Stopping()` call in the right place.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,5 +68,4 @@ Session.vim
 /cmd/*/.ethereum/
 
 # created for running container
-_assets/compose/mailserver/config.json
 _assets/compose/bootnode/keys
--- a/cmd/statusd/main.go
+++ b/cmd/statusd/main.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/log"
 	gethmetrics "github.com/ethereum/go-ethereum/metrics"
+	"github.com/okzk/sdnotify"
 	"github.com/status-im/status-go/api"
 	"github.com/status-im/status-go/logutils"
 	nodemetrics "github.com/status-im/status-go/metrics/node"
@@ -134,6 +135,16 @@ func main() {
 		return
 	}
 
+	err = sdnotify.Ready()
+	if err == sdnotify.ErrSdNotifyNoSocket {
+		logger.Debug("sd_notify socket not available")
+	} else if err != nil {
+		logger.Warn("sd_notify READY call failed", "error", err)
+	} else {
+		// systemd aliveness notifications, affects only Linux
+		go startSystemDWatchdog()
+	}
+
 	// handle interrupt signals
 	interruptCh := haltOnInterruptSignal(backend.StatusNode())
 
@@ -168,6 +179,9 @@ func main() {
 	if gethNode != nil {
 		// wait till node has been stopped
 		gethNode.Wait()
+		if err := sdnotify.Stopping(); err != nil {
+			logger.Warn("sd_notify STOPPING call failed", "error", err)
+		}
 	}
 }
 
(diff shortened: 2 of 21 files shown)


cmd/statusd/main.go
```diff
@@ -189,6 +195,15 @@ func setupLogging(config *params.NodeConfig) {
 	}
 }
 
+// loop for notifying systemd about process being alive
+func startSystemDWatchdog() {
+	tick := time.Tick(30 * time.Second)
+	for {
```
adambabik:
```suggestion
	for range time.Tick(30 * time.Second) {
	    sdnotify.Watchdog()
	}
```

cmd/statusd/main.go
```diff
@@ -133,6 +134,10 @@ func main() {
 		logger.Error("Node start failed", "error", err)
 		return
 	}
+	
+	// systemd aliveness notifications
+	go startSystemDWatchdog()
+	sdnotify.Ready()
```
adambabik: What if we do not have it as a systemd service? This can run wherever like macOS as well?

cmd/statusd/main.go
jakubgs: Done.

cmd/statusd/main.go
jakubgs: This has only an effect on Linux:
https://github.com/okzk/sdnotify/blob/master/notify_linux.go
On other systems it does nothing:
https://github.com/okzk/sdnotify/blob/master/notify.go
This is achieved using Go [build constraints](https://golang.org/pkg/go/build/#hdr-Build_Constraints):
```golang
// +build !linux
```

jakubgs:
This is weird. I tried running `make statusgo` on a MacOS host and it just gets stuck:
```
 $ make statusgo     
go build -i -o /Users/administrator/repos/status-go/build/bin/statusd -v -tags '' -ldflags ' -X main.buildStamp=2019-10-03.10:10:06 -X github.com/status-im/status-go/params.Version=0.33.0-beta.1 -X github.com/status-im/status-go/params.GitCommit=40e66e61 -X github.com/status-im/status-go/vendor/github.com/ethereum/go-ethereum/metrics.EnabledStr=true' ./cmd/statusd
...
go: finding golang.org/x/sys v0.0.0-20190801041406-cbf593c0f2f3
go: finding github.com/DATA-DOG/go-sqlmock v1.3.3
```
And just stops, no CPU usage, nothing, and this is on `develop`, not on this branch.

jakubgs:
There's a separate `statusgo-ios` target, but that's for use with `gomobile` and for building for Android.

adambabik:
@jakubgs do `make vendor` if you added any new dependencies in this PR and commit `vendor/` changes.

jakubgs:
Calling `make vendor` fails:
```
 $ make vendor
go mod tidy
warning: ignoring symlink /home/sochan/work/status-go
go mod vendor
warning: ignoring symlink /home/sochan/work/status-go
go: no dependencies to vendor
modvendor -copy="**/*.c **/*.h" -v
Whoops, cannot find vendor/modules.txt, first run `go mod vendor` and try again
make: *** [Makefile:303: vendor] Error 1
```
And it seems to have removed the `vendor` directory entirely:
```
 $ ls -l vendor
ls: cannot access 'vendor': No such file or directory
```
Is this intended behavior?

adambabik:
`warning: ignoring symlink /home/sochan/work/status-go` -- do you call it from a symlink? It might require to be called from the directory.

Also, if you upgrade to Go 1.13 you can put Go Modules compatible project inside `$GOPATH` so that no symlink is needed anymore.

jakubgs:
So I fixed the PR with the results of `make vendor` but the build on a MacOS hosts still just gets stuck.
@adambabik could you try out building this branch on your Mac?

adambabik:
@jakubgs which version of Go is there? Works fine on my localhost. I can debug further on the build host but I need to know which experiences this problem.

jakubgs:
The version on CI hosts is:
```
administrator@macos-01.ms-eu-dublin.ci.misc:~ % go version
go version go1.12.5 darwin/amd64
```

.gitignore
```diff
@@ -68,5 +68,5 @@ Session.vim
 /cmd/*/.ethereum/
 
 # created for running container
-_assets/compose/mailserver/config.json
+config.json
```
adambabik:
```suggestion
mailserver-config.json
```

adambabik (approved):

.gitignore
jakubgs: I think I'd just rather move the generated config to the `DATA_PATH` in `/var/tmp/status-go-mail`.

jakubgs:
Updated and with added `README.md` files for new way of running things and a `MAILSERVER.md`.

PR #1627, opened 2019-10-04
Update for status-protocol-go API changes

pedropombeiro:
This PR integrates the changes from https://github.com/status-im/status-protocol-go/pull/88.

This PR is part of a group of 3 PRs (https://github.com/status-im/status-protocol-go/pull/88, https://github.com/status-im/status-go/pull/1627, https://github.com/status-im/status-console-client/pull/128) and we should only merge when we're happy with the 3 of them.


cammellos:
@PombeirP this is a non breaking change correct?

pedropombeiro:
> @PombeirP this is a non breaking change correct?

Correct.

cammellos (approved):

pedropombeiro:
@cammellos on second thought, there is one aspect that can be considered a breaking change, so we might want to not merge this as-is. I've commented out the envelope monitor in status-protocol-go. We have 2 alternatives: mark this as WIP and continue with Nimbus effort, or I can spend some more time trying to isolate the envelope monitor as I did for the rest of the services. The reason I didn't touch the envelope monitor is that I'm not sure whether we want to port that implementation to Nimbus.

cammellos:
@PombeirP yes, we need envelopes monitor otherwise all messages in status-react will be marked as pending and never confirmed, probably best to port it before merging

pedropombeiro:
Ready for final review. This is now a non-breaking change. cc @cammellos

cammellos (approved):

PR #1628, opened 2019-10-05
Remove xgo builds

adambabik:
We don't need xgo anymore because we switched to gomobile for iOS and Android and building from source for Status Desktop.


adambabik:
@jakubgs in which repo do we keep content of our website? I need to change fifth point on https://status.im/build_status/status_go.html.

jakubgs:
Right here:
https://github.com/status-im/status.im/blob/5edafcbbf41/source/build_status/status_go.md#L106-L113

jakubgs (approved):
Love removing stuff.

adambabik:
Docs changed: https://github.com/status-im/status.im/pull/411

PR #1633, opened 2019-10-09
create a /metrics endpoint for Prometheus

jakubgs:
Since Geth `1.9` [introduced a Prometheus metrics endpoint](https://blog.ethereum.org/2019/07/10/geth-v1-9-0/#metrics-collection) and @adambabik has upgraded `status-go` to use Geth `1.9` in 26880b83d7cf4e950ac11eb8b14464adf10f3cb8 it would make sense to start using that endpoint.

This is done for two reasons:

* Allows us to drop the [geth_exporter](https://github.com/status-im/geth_exporter) service for exporting metrics
* Implement a new `whiper_mobile_peers_total` metrics for https://github.com/status-im/infra-misc/issues/26

I got the default `9305` port from [Prometheus default port allocations](https://github.com/prometheus/prometheus/wiki/Default-port-allocations).

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -1,0 +1,32 @@
+package metrics
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/metrics"
+	"github.com/ethereum/go-ethereum/metrics/prometheus"
+)
+
+// Server runs and controls a HTTP pprof interface.
+type Server struct {
+	server *http.Server
+}
+
+func NewMetricsServer(port int, r metrics.Registry) *Server {
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", prometheus.Handler(r))
+	p := Server{
+		server: &http.Server{
+			Addr:    fmt.Sprintf(":%d", port),
+			Handler: mux,
+		},
+	}
+	return &p
+}
+
+// Listen starts the HTTP server in the background.
+func (p *Server) Listen() {
+	log.Info("metrics server stopped", "err", p.server.ListenAndServe())
+}
--- a/cmd/statusd/main.go
+++ b/cmd/statusd/main.go
@@ -17,6 +17,7 @@ import (
 	gethmetrics "github.com/ethereum/go-ethereum/metrics"
 	"github.com/status-im/status-go/api"
 	"github.com/status-im/status-go/logutils"
+	"github.com/status-im/status-go/metrics"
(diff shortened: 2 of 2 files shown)


metrics/metrics.go
```diff
@@ -0,0 +1,35 @@
...
+		server: &http.Server{
+			Addr:    fmt.Sprintf(":%d", port),
+			Handler: mux,
+		},
+	}
+	return &p
+}
+
+// Go starts the HTTP pprof in the background.
+func (p *MetricsServer) Go() {
```
adambabik:
```suggestion
func (p *MetricsServer) Listen() {
    log.Info("metrics server stopped", "err", p.server.ListenAndServe())
}
```

Also, in this case, it's totally fine for the caller to take the responsibility to call it in a goroutine rather than do it here.

So it would be executed like this:
```go
go metrics.NewMetricsServer(metricsPort, gethmetrics.DefaultRegistry).Listen()
```

jakubgs: Sure, why not.

metrics/metrics.go
```diff
@@ -0,0 +1,32 @@
...
+			Addr:    fmt.Sprintf(":%d", port),
+			Handler: mux,
+		},
+	}
+	return &p
+}
+
+// Listen starts the HTTP server in the background.
+func (p *MetricsServer) Listen() {
+	log.Info("metrics server stopped", "err", p.server.ListenAndServe())
```
pedropombeiro, line 31: Shouldn't this be:
```suggestion
	log.Info("metrics server started", "err", p.server.ListenAndServe())
```

pedropombeiro (approved):

metrics/metrics.go
adambabik: It can be both. It can error immediately because, for example, the port is already used, it can error due to a signal after serving HTTP for a while and so on. I would keep `stopped` as it's more common or rephrase it completely.

jakubgs: But.... it's already using `stopped`... I don't get what you want me to do here.

adambabik: Pedro proposed `started` instead. I don't want you to do anything, IMO it's good as it is. I think no one has strong opinion about this so it's good to merge.

adambabik:
Linter is not happy: https://ci.status.im/blue/organizations/jenkins/status-go%2Fprs%2Ftests/detail/PR-1633/2/pipeline

metrics/metrics.go
jakubgs: Ah, you're right, I'm blind I guess, didn't even notice Pedro there.

adambabik (approved):

PR #1652, opened 2019-10-23
String payload

cammellos:
Add a string payload, which will be used by status-react as it avoid having to convert `hex` to `utf8`, improving performance.

### DIFF
--- a/services/shhext/service.go
+++ b/services/shhext/service.go
@@ -213,6 +213,7 @@ func (s *Service) retrieveMessagesLoop(tick time.Duration, cancel <-chan struct{
 							Message: message.TransportMessage,
 						}
 						dedupMessage.Message.Payload = message.DecryptedPayload
+						dedupMessage.Payload = string(message.DecryptedPayload)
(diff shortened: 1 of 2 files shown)


services/shhext/service.go
```diff
@@ -213,6 +213,7 @@ func (s *Service) retrieveMessagesLoop(tick time.Duration, cancel <-chan struct{
 							Message: message.TransportMessage,
 						}
 						dedupMessage.Message.Payload = message.DecryptedPayload
```
adambabik, line 215: Do we need this then?

cammellos: yes, we need still timestamp, but working on related code (I might need to make some other changes), in the next PR I will just drop that and add the `timestamp` field directly.

cammellos:
@adambabik @PombeirP could you please have a look at this one? thanks!

adambabik (approved):

pedropombeiro (approved):

cammellos:
Closing this as already merged in different pr.

PR #1656, opened 2019-10-28
make sure Version is always set in loadNodeConfig()

jakubgs:
This should resolve the issue with missing `status-go` version in the App:
https://github.com/status-im/status-react/issues/9270

### DIFF
--- a/params/version_test.go
+++ b/params/version_test.go
@@ -1,0 +1,19 @@
+package params_test
+
+import (
+	"io/ioutil"
+	"path/filepath"
+	"regexp"
+	"testing"
+)
+
+func TestVersionFormat(t *testing.T) {
+	data, err := ioutil.ReadFile(filepath.Join("..", "VERSION"))
+	if err != nil {
+		t.Error("unable to open VERSION file")
+	}
+	matched, _ := regexp.Match(`^\d+\.\d+\.\d+(-[.\w]+)?\n?$`, data)
+	if !matched {
+		t.Error("version in incorrect format")
+	}
+}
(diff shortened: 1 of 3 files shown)


params/version_test.go
```diff
@@ -0,0 +1,17 @@
+package params_test
+
+import (
+	"io/ioutil"
+	"strings"
+	"testing"
+)
+
+func TestVersionFormat(t *testing.T) {
+	data, err := ioutil.ReadFile("../VERSION")
```
adambabik:
```suggestion
	data, err := ioutil.ReadFile(filepath.Join("..", "VERSION"))
```

params/version_test.go
```diff
@@ -0,0 +1,17 @@
...
+	if err != nil {
+		t.Error("unable to open VERSION file")
+	}
+	if strings.HasPrefix(string(data), "v") {
```
adambabik: Maybe we can check with regex?

params/version_test.go
jakubgs: Sure.

params/version_test.go
jakubgs: Ok.

params/version_test.go
```diff
@@ -0,0 +1,19 @@
...
+	"regexp"
+	"testing"
+)
+
+func TestVersionFormat(t *testing.T) {
+	data, err := ioutil.ReadFile(filepath.Join("..", "VERSION"))
+	if err != nil {
+		t.Error("unable to open VERSION file")
+	}
+	matched, _ := regexp.Match(`^\d+\.\d+\.\d+(-[.\w]+)?\n?$`, data)
```
adambabik, line 15: Is the new line required? `^\d+\.\d+\.\d+(-[.\w]+)?$` should apply per line, right?

adambabik (approved):
LGTM! One minor suggestion which I am not sure of.

params/version_test.go
jakubgs: It did not work without the newline.

PR #1660, opened 2019-11-06
Add parsed message

cammellos:
This commits add a field (parsedMessage) to the json payload sent to
status-react.
This field is the parsed version of the transit message.
The code is all in dedup, I will re-organize it once we made all the
necesseary changes.


services/shhext/service.go
```diff
@@ -214,6 +214,8 @@ func (s *Service) retrieveMessagesLoop(tick time.Duration, cancel <-chan struct{
 						}
 						dedupMessage.Message.Payload = message.DecryptedPayload
 						dedupMessage.Payload = string(message.DecryptedPayload)
+						dedupMessage.ParsedMessage = message.ParsedMessage
```
adambabik, line 217: I believe it's known and it will be optimized later but the payload is tripled:
- `dedupMessage.Message.Payload`
- `dedupMessage.Payload`
- `dedupMessage.ParsedMessage` in decoded form

adambabik (approved):

services/shhext/service.go
cammellos: Yes, I think we can remove now Message.Payload, but I will double check

PR #1671, opened 2019-11-12
DeleteAccount method

rasom:
DeleteAccount method is added, it deletes account from
multiaccounts database and also removes account's DB file.
The method is needed for keycard reseting functionality.
https://github.com/status-im/status-react/issues/9229

### DIFF
--- a/api/backend.go
+++ b/api/backend.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"os"
 	"path/filepath"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


api/backend.go
```diff
@@ -1052,3 +1067,23 @@ func (b *StatusBackend) SignHash(hexEncodedHash string) (string, error) {
...
+	err := b.closeAppDB()
+	if err != nil {
+		return err
+	}
+	if len(b.rootDataDir) == 0 {
+		return errors.New("root datadir wasn't provided")
+	}
+
+	path := accountDBPath(b.rootDataDir, address)
+	err = os.RemoveAll(path)
```
cammellos: you can just do `return os.RemoveAll(path)` I believe

lib/library.go
```diff
@@ -383,6 +383,24 @@ func SaveAccountAndLogin(accountData, password, configJSON, subaccountData *C.ch
 	return makeJSONResponse(nil)
 }
 
+// DeleteAccount delete account's database and delete account from multiaccount database
+//export DeleteAccount
+func DeleteAccount(addressChars *C.char) *C.char {
```
cammellos, line 388: Can we use RPC in this case? I believe the consensus is to use RPC when poissble

rasom: hm, yes I think we can

api/backend.go
```diff
@@ -173,6 +174,20 @@ func (b *StatusBackend) SaveAccount(account multiaccounts.Account) error {
...
+func (b *StatusBackend) DeleteAccount(address common.Address) error {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	if b.multiaccountsDB == nil {
+		return errors.New("accoutns db wasn't initialized")
+	}
+	return b.multiaccountsDB.DeleteAccount(address)
+}
+
+func accountDBPath(root string, address common.Address) string {
```
adambabik, line 187: We can move it at the bottom of the file.

PR #1672, opened 2019-11-12
add tcp-pinger for mailservers

### LINKED ISSUE #9394: Select mailservers based on ping
### User Story

Currently the mailservers from the fleet are selected randomly, which results in me getting connected to the Hong Kong mailservers almost every time when I'm in Europe and should be getting the european ones.

### Description

This results in a really bad experience for users with slow sync times and bad latency.

### Acceptance Criteria

The simplest solution would be to just do an ICMP ping to a few mailserver, group them into response time buckets, and pick a random one from the fastest bucket.

### Notes

Example ping output:
```
 $ for DC in "do-ams3" "gc-us-central1-a" "ac-cn-hongkong-c"; do \
     ping -c1 mail-01.${DC}.eth.beta.statusim.net | grep 'time='; \
   done 
64 bytes from 206.189.243.162 (206.189.243.162): icmp_seq=1 ttl=55 time=29.1 ms
64 bytes from 210.19.188.35.bc.googleusercontent.com (35.188.19.210): icmp_seq=1 ttl=50 time=127 ms
64 bytes from 47.91.156.93 (47.91.156.93): icmp_seq=1 ttl=47 time=279 ms
```
As you can see for me from Poland the ranking of access time is simple:

1. Amsterdam - `29.1 ms`
2. United States - `127 ms`
3. Hong Kong - `279 ms`

jakubgs:
This is an initial version of implementation of an RPC call for checking latency of mailservers.

Partially resolves https://github.com/status-im/status-react/issues/9394.

Status: __READY__


jakubgs:
Currently this doesn't seem to work as I'm unable to make and RPC request:
```bash
curl -s localhost:8545 -H 'content-type: application/json' \
    -d '{"jsonrpc":"2.0","method":"mailservers_Ping","params":[["1.1.1.1:53"], 1000],"id":1}' | jq
```
```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": -32601,
    "message": "the method mailservers_Ping does not exist/is not available"
  }
}
```
NOTE: Turns our the call name needs to be lowercase: `mailservers_ping`.

services/mailservers/tcp_ping.go
```diff
@@ -0,0 +1,41 @@
+package mailservers
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/status-im/status-go/tcpping"
```
adambabik: I would call it `github.com/status-im/status-go/rtt`

services/mailservers/tcp_ping.go
```diff
@@ -0,0 +1,41 @@
...
+		Address: r.Address,
+		Latency: &r.Latency,
+		PingErr: &err,
+	}
+}
+
+func (a *API) Ping(ctx context.Context, pq PingQuery) ([]PingResult, error) {
+	timeout := time.Duration(pq.TimeoutMs) * time.Millisecond
+	// run the checks concurrently
+	results, err := tcpping.CheckHosts(pq.Addresses, timeout)
```
adambabik: If `err` is here, I believe we can return immediately with `return nil, err`.

services/mailservers/tcp_ping.go
```diff
@@ -0,0 +1,41 @@
...
+
+type PingQuery struct {
+	Addresses []string `json:"addresses"`
+	TimeoutMs int      `json:"timeoutMs"`
+}
+
+type PingResult struct {
+	Address string  `json:"address"`
+	Latency *int    `json:"latency"`
+	PingErr *string `json:"error"`
```
adambabik: This is totally ok. I was mistaken, having `PingErr error` will always return `"Error": {}` when marshalled to JSON unless you define a custom error type like this: https://play.golang.org/p/qlZEjqORxlO

services/mailservers/tcp_ping.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func fromPingResultToJSON(r tcpping.Result) PingResult {
+	err := fmt.Sprint(r.PingErr)
```
adambabik:
```suggestion
	err := r.PingErr.Error()
```

services/mailservers/tcp_ping.go
```diff
@@ -0,0 +1,41 @@
...
+	TimeoutMs int      `json:"timeoutMs"`
+}
+
+type PingResult struct {
+	Address string  `json:"address"`
+	Latency *int    `json:"latency"`
+	PingErr *string `json:"error"`
+}
+
+func fromPingResultToJSON(r tcpping.Result) PingResult {
```
adambabik:
```suggestion
func toPingResult(r tcpping.Result) PingResult {
```
maybe? `fromPingResult` is odd when `PingResult` is returned :D

tcpping/tcpping.go
```diff
@@ -0,0 +1,95 @@
...
+	return fmt.Sprintf(
+		"Result{address: %s, latency: %d, error: %s}\n",
+		cr.Address, cr.Latency, cr.PingErr)
+}
+
+func runCheck(c *tcp.Checker, address string, resChan chan Result, timeout time.Duration) {
+	// mesaure RTT
+	start := time.Now()
+	// TCP Ping
+	err := c.CheckAddr(address, timeout)
```
adambabik: Handle error here immediately.

tcpping/tcpping.go
```diff
@@ -0,0 +1,95 @@
...
+	PingErr error
+}
+
+func (cr *Result) String() string {
+	return fmt.Sprintf(
+		"Result{address: %s, latency: %d, error: %s}\n",
+		cr.Address, cr.Latency, cr.PingErr)
+}
+
+func runCheck(c *tcp.Checker, address string, resChan chan Result, timeout time.Duration) {
```
adambabik:
```suggestion
func runCheck(c *tcp.Checker, address string, resChan chan<- Result, timeout time.Duration) {
```

tcpping/tcpping.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// return results
+	resChan <- Result{
+		Address: address,
+		Latency: latency,
+		PingErr: err,
+	}
+}
+
+func waitForResults(resChan chan Result, expected int, timeout time.Duration) ([]Result, error) {
```
adambabik:
```suggestion
func waitForResults(ch <-chan Result) (result []Result) {
    for item := range ch {
        result = append(result, item)
    }
    return
}
```

tcpping/tcpping.go
```diff
@@ -0,0 +1,95 @@
...
+	c := tcp.NewChecker()
+
+	// stop the checking loop when function exists
+	ctx, stopChecker := context.WithCancel(context.Background())
+	defer stopChecker()
+
+	// loop that queries Epoll and pipes events
+	go func() {
+		// TODO handle error from this?
+		c.CheckingLoop(ctx)
```
adambabik:
```suggestion
		errCh <- c.CheckingLoop(ctx)
```

where `errCh := make(chan error, 1)`

tcpping/tcpping.go
```diff
@@ -0,0 +1,95 @@
...
+	}()
+
+	// channel for returning results from concurrent checks
+	var resChan = make(chan Result)
```
adambabik:
```suggestion
	resChan := make(chan Result, len(addresses))
```

tcpping/tcpping.go
```diff
@@ -0,0 +1,95 @@
...
+
+	for i := 0; i < len(addresses); i++ {
+		go runCheck(c, addresses[i], resChan, timeout)
+	}
+
```
adambabik:
```suggestion
close(resChan)
```

tcpping/tcpping.go
```diff
@@ -0,0 +1,95 @@
...
+	// add a bit extra time to let TCP connections time out
+	waitTimeout := timeout + (100 * time.Millisecond)
+
+	// wait for the results for all addresses
+	return waitForResults(resChan, len(addresses), waitTimeout)
```
adambabik:
```suggestion
	return waitForResults(resChan)
```

No need timeout in `waitForResult` because `runCheck` goroutines have timeout and will exit before anyway. Also, they will run in parallel so the whole set should not run longer than `timeout`.

tcpping/tcpping.go
jakubgs: But I don't want to return a `Result` struct with the error, so it can be returned. Stopping here makes no sense because the RPC call response wont include info about why it failed.

services/mailservers/tcp_ping.go
jakubgs: done

services/mailservers/tcp_ping.go
jakubgs: sure

services/mailservers/tcp_ping.go
jakubgs: k

tcpping/tcpping.go
jakubgs: neat

tcpping/tcpping.go
jakubgs: much cleaner, nice

tcpping/tcpping.go
jakubgs: But where should I read from that `errCh`? In `waitForResults()`? After it finishes?

tcpping/tcpping.go
jakubgs: done

tcpping/tcpping.go
jakubgs: done

tcpping/tcpping.go
jakubgs: Sure, after i added a `sync.WaitGroup` before the `close()`.

tcpping/tcpping.go
jakubgs: Ok, found a solution:
```go
func waitForResults(errCh <-chan error, resCh <-chan Result) (results []Result, err error) {
	for {
		select {
		case err = <-errCh:
			return nil, err
		case res, ok := <-resCh:
			if !ok {
				return
			}
			results = append(results, res)
		}
	}
}
```

services/mailservers/tcp_ping.go
jakubgs: done

rtt/rtt.go
```diff
@@ -0,0 +1,103 @@
...
+	tcp "github.com/status-im/tcp-shaker"
+)
+
+type Result struct {
+	Address string
+	Latency int
+	PingErr error
+}
+
+func (cr *Result) String() string {
```
adambabik: Where do you use it?

rtt/rtt.go
```diff
@@ -0,0 +1,103 @@
...
+			case tcp.ErrConnect:
+				err = errors.Wrap(err, "unable to connect")
+			default:
+				err = errors.Wrap(err, "unknown error")
+			}
+		}
+	}
+
+	// return results
+	ch <- Result{
```
adambabik: Hm I don't see any goroutine here. You can just `return Result{...}` without using channels.

rtt/rtt.go
```diff
@@ -0,0 +1,103 @@
...
+		// TODO handle error from this?
+		errCh <- c.CheckingLoop(ctx)
+	}()
+
+	var wg sync.WaitGroup
+	for i := 0; i < len(addresses); i++ {
+		wg.Add(1)
+		go func(address string) {
+			defer wg.Done()
+			runCheck(c, address, resCh, timeout)
```
adambabik:
```suggestion
			resCh <- runCheck(c, address, timeout)
```

rtt/rtt.go
```diff
@@ -0,0 +1,103 @@
...
+	resCh := make(chan Result, len(addresses))
+
+	// stop the checking loop when function exists
+	ctx, stopChecker := context.WithCancel(context.Background())
+	defer stopChecker()
+
+	// loop that queries Epoll and pipes events
+	go func() {
+		// TODO handle error from this?
+		errCh <- c.CheckingLoop(ctx)
```
adambabik, line 76: Does it return ever if there is no error?

rtt/rtt.go
```diff
@@ -0,0 +1,103 @@
...
+			runCheck(c, address, resCh, timeout)
+		}(addresses[i])
+	}
+	// wait for all the routines to finish
+	wg.Wait()
+	// close the pipe
+	close(resCh)
+
+	// wait for the results for all addresses
+	return waitForResults(errCh, resCh)
```
adambabik, line 97:
```suggestion
	if err := <-errCh; err != nil {
	    return nil, err
	}
	return <-collectResults(resCh)
```

Or even do:

```go
errCh := make(chan error, 1)
doneChecks := make(chan struct{})
resCh := make(chan Result, len(addresses))

go func() {
		select {
		case err := <-c.CheckingLoop(ctx):
				if err != nil {
					errCh <- err
				}
		case <-doneChecks:
		}
}()

// ...

var result []Result
go func() {
    for item := range resCh {
        result = append(result, item)
    }
}()

// ...

wg.Wait()
close(doneChecks)
close(resCh)

if err := <-errCh; err != nil {
    return nil, err
}
return result, nil
```

It for 99% can be simplified.

services/mailservers/tcp_ping.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+type PingResult struct {
+	Address string  `json:"address"`
+	Latency *int    `json:"latency"`
+	PingErr *string `json:"error"`
+}
+
+func toPingResult(r rtt.Result) PingResult {
+	var err *string = nil
```
adambabik:
```suggestion
	var err *string
```

services/mailservers/tcp_ping.go
```diff
@@ -0,0 +1,47 @@
...
+	results, err := rtt.CheckHosts(pq.Addresses, timeout)
+	if err != nil {
+		return nil, err
+	}
+	// conver to json format
+	jsonResults := make([]PingResult, len(results))
+	for i := range results {
+		jsonResults[i] = toPingResult(results[i])
+	}
+	return jsonResults, err
```
adambabik:
```suggestion
	return jsonResults, nil
```

rtt/rtt.go
jakubgs: I was using it when testing stuff, can drop it now.

rtt/rtt.go
jakubgs: Well, the context manager should close it, so I guess it does. Is that a problem?

services/mailservers/tcp_ping.go
jakubgs: k

rtt/rtt.go
jakubgs: So I guess it returns a `nil` because it's killed by the `context` thing.

rtt/rtt.go
jakubgs: Adjusted.

rtt/rtt.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+type Result struct {
+	Addr  string
+	RTTMs int
+	Err   error
+}
+
+// tcp-sharer doesn't export TimeoutError type
+type TimeoutError interface {
```
adambabik:
```suggestion
type timeoutError interface {
```

rtt/rtt.go
```diff
@@ -0,0 +1,99 @@
...
+	tcp "github.com/status-im/tcp-shaker"
+)
+
+type Result struct {
+	Addr  string
+	RTTMs int
+	Err   error
+}
+
+// tcp-sharer doesn't export TimeoutError type
```
adambabik:
```suggestion
// TimeoutError indicates an error due to TCP connection timeout. 
// tcp-sharer returns an error implementing this interface in such a case.
```

adambabik (approved):

adambabik:
Error in the CI:
```
[2019-11-13T17:13:30.666Z] # github.com/status-im/status-go/vendor/github.com/status-im/tcp-shaker

[2019-11-13T17:13:30.666Z] vendor/github.com/status-im/tcp-shaker/socket_linux.go:89:29: timeout.Milliseconds undefined (type time.Duration has no field or method Milliseconds)
```

jakubgs:
Android seems broken, not sure why:
```
vendor/github.com/status-im/tcp-shaker/socket_linux.go:89:29: timeout.Milliseconds undefined
```
Something similar goes for iOS:
```
vendor/github.com/status-im/tcp-shaker/socket_darwin.go:92:29: cannot use timeout.Nanoseconds() (type int64) as type int32 in field value
```

adambabik:
```
vendor/github.com/status-im/tcp-shaker/socket_linux.go:89:29: timeout.Milliseconds undefined
```
This is added in Go 1.13 and I think we still use 1.12 in the CI.

```
vendor/github.com/status-im/tcp-shaker/socket_darwin.go:92:29: cannot use timeout.Nanoseconds() (type int64) as type int32 in field value
```

https://sourcegraph.com/github.com/golang/sys/-/blob/plan9/syscall_plan9.go#L258 it's clearly `int32` so you need to do the conversion:
```go
tsTimeout := unix.Timespec{Nsec: int32(timeout.Nanoseconds())}
```

jakubgs:
Yeah, I get it, but why are you looking at a `plan9` file?

jakubgs:
Now I have no idea why iOS fails or the linting in tests...

adambabik:
```
[2019-11-14T01:00:09.599Z] vendor/github.com/status-im/tcp-shaker/socket_darwin.go:79:3: cannot use uint32(fd) (type uint32) as type uint64 in field value

[2019-11-14T01:00:09.599Z] vendor/github.com/status-im/tcp-shaker/socket_darwin.go:92:29: cannot use int32(timeout.Nanoseconds()) (type int32) as type int64 in field value
```
Problem with iOS. Sorry, I looked in a wrong file. But generally there are some issues with types so it just needs fixing as told.

And linter now... Did you update Go in the CI?

jakubgs:
I did, to `1.13.4`: https://github.com/status-im/infra-ci/commit/558247f9

jakubgs:
What I'm getting from this is that `gomobile` is building for both `arm` and `arm64` which use different types for the `Ident` attribute of `Kevent_t`:
```go
type Kevent_t struct {
	Ident  uint32
	Filter int16
	Flags  uint16
	Fflags uint32
	Data   int32
	Udata  *byte
}
```
http://www.golang.so/src/syscall/ztypes_darwin_arm.go
```go
type Kevent_t struct {
	Ident  uint64
	Filter int16
	Flags  uint16
	Fflags uint32
	Data   int64
	Udata  *byte
}
```
http://www.golang.so/src/syscall/ztypes_darwin_arm64.go
But I think the solution is to use the `SetKevent()` method:
```go
func SetKevent(k *Kevent_t, fd, mode, flags int) {
	k.Ident = uint64(fd)
	k.Filter = int16(mode)
	k.Flags = uint16(flags)
}
```
http://www.golang.so/src/syscall/syscall_darwin_arm.go
http://www.golang.so/src/syscall/syscall_darwin_arm64.go

jakubgs:
Same goes for `Timespec`, they have different types on `arm` and `arm64`:
```go
type Timespec struct {
	Sec  int32
	Nsec int32
}
```
http://www.golang.so/src/syscall/ztypes_darwin_arm.go
```go
type Timespec struct {
	Sec  int64
	Nsec int64
}
```
http://www.golang.so/src/syscall/ztypes_darwin_arm64.go

There is a helper method but it's not exported and it uses strict types:
```go
func setTimespec(sec, nsec int64) Timespec {
	return Timespec{Sec: int32(sec), Nsec: int32(nsec)}
}
```
http://www.golang.so/src/syscall/syscall_darwin_arm.go

jakubgs:
Found the solution to `Timespect` in https://github.com/golang/go/issues/23078 which is the `TimeToTimespec` method:
```go
func TimeToTimespec(t time.Time) (Timespec, error)
```
https://godoc.org/golang.org/x/sys/unix#NsecToTimespec

jakubgs:
@adambabik I managed to fix the build errors, but I have no clue what the linter issues are about.

Also, I tried building it with `status-react` via Nix but I'm getting:
```
rtt/rtt.go:31:24: elapsed.Milliseconds undefined (type time.Duration has no field or method Milliseconds)
```
Which makes no sense because the method clearly exists:
```go
func (d Duration) Milliseconds() int64
```
https://golang.org/pkg/time/#Duration.Milliseconds
Gonna need to look into that too.

jakubgs:
I think this issue - https://github.com/golang/go/issues/28564 - indicates that the `Milliseconds()` method for `time.Duration` was added only in Go `1.13`, and we are still on `1.12` in `status-react`:
https://github.com/status-im/status-react/blob/2000532c/nix/derivation.nix#L20
I have attempted to upgrade our `nixpkgs` to include `1.13` but it was quite difficult. For now I'm just going to do a simple calculation so we don't have to use that method.

jakubgs:
I ran a build of `status-react` on a branch called `test-tcp-ping` and it succeeded for all architectures:
https://ci.status.im/job/status-react/job/manual/516/
So we can consider this almost done, the only thing remaining is fixing the test/linting errors.

adambabik:
I have no idea why it started complaining about returned C types. In 153124f, this error is ignored. I reported this behaviour in https://github.com/golangci/golangci-lint/issues/858. There is a simple case how to reproduce it.

jakubgs:
@adambabik thanks for fixing that, then I guess we are ready to merge unless you think there's anything else this might be missing.

I added a small `rtt/README.md` to explain usage. Seemed like the right place to put it.

services/mailservers/tcp_ping.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("%s:%d", net.IP(ip4).String(), tcp), nil
+}
+
+func parseEnodes(enodes []string) (map[string]*PingResult, []string) {
+	// parse enode addreses into normal host + port addresses
+	results := make(map[string]*PingResult, len(enodes))
+	toPing := make([]string, 0)
```
adambabik:
```suggestion
	var toPing []string
```

services/mailservers/tcp_ping.go
```diff
@@ -0,0 +1,98 @@
...
+
+	// run the checks concurrently
+	results, err := rtt.CheckHosts(toPing, timeout)
+	if err != nil {
+		return nil, err
+	}
+
+	// set ping results
+	for i := range results {
+		r := results[i]
```
adambabik, line 93: We should check if `r != nil`.

adambabik (approved):

services/mailservers/tcp_ping.go
jakubgs: Not sure how it could be `nil`, but I've added a `continue` if it is.

jakubgs: Not sure how it could be `nil`, but I've added a `continue` if it is.

PR #1673, opened 2019-11-13
Update whisper to v1.6.0-beta.0 and introduce rate limiting

adambabik:
This introduces Whisper rate limiting per IP and peer ID. It is not enforced but only to collect metrics.


node/node.go
```diff
@@ -324,6 +324,11 @@ func activateShhService(stack *node.Node, config *params.NodeConfig, db *leveldb
 
 		whisperService := whisper.New(whisperServiceConfig)
 
+		if config.WhisperConfig.EnableRateLimiter {
+			r := whisperRateLimiter(config.ClusterConfig.StaticNodes)
```
adambabik: It does not work, it does not contain nodes from a fleet.

pedropombeiro (approved):

PR #1684, opened 2019-11-20
Move to monorepo structure

pedropombeiro:
This PR moves status-go into a monorepo structure by folding in status-protocol-go into the `protocol` subdirectory.

Changes:
- `github.com/status-im/status-protocol-go` -> `github.com/status-im/status-go/protocol`
- `package protocol` -> `package v1protocol`
- `package statusproto` -> `package protocol`


pedropombeiro:
This PR seems to be ready for review, only CodeClimate is failing for some unknown reason, but that doesn't prevent merging.

jakubgs (changes requested):
I don't really see why we would put all the normal contents of `status-go` in a `status-go` subfolder. If we are treating this as a mono-repo into which we want to integrate `status-protocol-go` then we shouldn't be modifying the structure of `status-go`.

I say simply put contents of `status-protocol-go` at the root of this repo in a `protocol` folder and rename the package from `statusproto` to just `protocol`. That will make it `github.com/status-im/status-go/protocol`, which is much cleaner. @adambabik thoughts?

pedropombeiro:
@jakubgs yeah, at first I didn't like seeing status-protocol-go getting lost among the dozens of existing folders, but I guess it makes more sense to leave status-go at the top level. I've pushed that change.

jakubgs (approved):

pedropombeiro:
Merging once build passes

PR #1685, opened 2019-11-21
Add KeyUID to IdentifiedAccountInfo

rasom:
This field will be used as a deterministic id of multiaccount on
status-react side instead of master key address. The reason why
it is calculated as sha256 of public key is that this way is
already used on keycard side and it will simplify integration.

related to https://github.com/status-im/status-react/issues/9435


account/generator/account.go
```diff
@@ -25,9 +26,12 @@ func (a *account) toAccountInfo() AccountInfo {
 
 func (a *account) toIdentifiedAccountInfo(id string) IdentifiedAccountInfo {
 	info := a.toAccountInfo()
+	keyUID := sha256.Sum256(crypto.FromECDSAPub(&a.privateKey.PublicKey))
+	keyUIDHex := statusproto.EncodeHex(keyUID[:])
 	return IdentifiedAccountInfo{
 		AccountInfo: info,
 		ID:          id,
+		KeyUID:      keyUIDHex,
```
adambabik, line 34: Why not `KeyID`? And what is `id` in this context? Can we add a comment because it gets confusing why we suddenly have two IDs.

account/generator/account.go
```diff
@@ -48,7 +52,8 @@ type AccountInfo struct {
 // IdentifiedAccountInfo contains AccountInfo and the ID of an account.
 type IdentifiedAccountInfo struct {
 	AccountInfo
-	ID string `json:"id"`
+	ID     string `json:"id"`
```
adambabik: Let's explain what it is and how `KeyUID` (maybe rename to `KeyID`) is different from `ID`.

adambabik (commented):
Implementation is fine just asked for a bit more explanation.

account/generator/account.go
rasom: I agree that it might be `KeyID`, the thing is that the choice of the name has been done already, and it is widely used in many repos
https://github.com/status-im/react-native-status-keycard/search?q=uid&unscoped_q=uid
https://github.com/status-im/keycard-cli/search?q=uid&unscoped_q=uid
https://github.com/status-im/status-keycard/search?q=uid&unscoped_q=uid
https://github.com/status-im/status-keycard-java/search?q=uid&unscoped_q=uid
https://github.com/status-im/status-react/search?q=keycard-key-uid&unscoped_q=keycard-key-uid

Even in this repo, although as `keycard-key-uid`:
https://github.com/status-im/status-go/search?q=keycard-key-uid&unscoped_q=keycard-key-uid

The problem now is that we either need to change it to `KeyID`(`key-id`) everywhere or some of repos (almost likely `status-react` and `status-go` atm) will have both and this won't make things easier to maintain and understand, raising the question "What's the difference?" each time you encounter both names.

account/generator/account.go
rasom: We can rename it to `KeyID` but keep it as `keyUid` in json, would that work for you?

account/generator/account.go
adambabik: Got it, thanks for the explanation. Let's keep it as it is then and only explain what it is and how it is different from `ID`.

rasom:
@adambabik added a comment as you requested and also added another commit to rename `keycard-key-uid` to `key-uid` in multiaccs database. Let me know if that comment is not sufficient enough or database change should be filed in a separate PR.

adambabik (approved):

PR #1695, opened 2019-11-23
Create a home submodule for Eth node bridges

pedropombeiro:
This PR paves the way for the Nimbus backend work. It moves stuff that is not intrinsic to the protocol into a new sub-module called `eth-node`. Notable changes:
- renamed StatusBackend to GethStatusBackend (to make room for future NimbusStatusBackend)
- status-go/crypto -> status-go/eth-node/crypto
- some go-ethereum/crypto -> status-go/eth-node/crypto
- status-go/protocol/types -> status-go/eth-node/types
- status-go/protocol/transport/whisper/types -> status-go/eth-node/types
- status-go/protocol/bridge -> status-go/eth-node/bridge
- `build-nimbus.sh` now lives in `eth-node` instead of in the consuming modules.

What I'm not happy about in this PR:
- We need a top-level `types.Node` instance which will be responsible for handing out appropriate services (e.g. a Nimbus or Geth implementation of ENSVerifier, depending on the node type). So far so good, but ENSVerifier uses a zap logger, and currently I'm passing it as an argument to `Node.NewENSVerifier`, and that is not so great, because we're tying our interface to an implementation detail.


adambabik:
@PombeirP any idea why moving stuff resulted in such an increase in additions `+2,002,428 −7,548`?

pedropombeiro:
@adambabik it looks like `go mod` pulled in new stuff into `protocol/vendor` (like `vendor/github.com/ethereum/go-ethereum`) which for some reason wasn't there before. Geth shouldn't even be in protocol anymore, only a single test references it at this point, and even if I comment out the test `make vendor` will add that folder again. Any ideas?

UPDATE: Adam just let me know that it's not necessary to vendor in the submodules, so that does get rid of all those additions. Commiting fix.

cammellos (approved):

eth-node/Makefile
```diff
@@ -0,0 +1,24 @@
...
+
+test:
+	go test ./...
+.PHONY: test
+
+lint:
+	golangci-lint run -v
+.PHONY: lint
+
+vendor:
```
jakubgs: I guess we don't need this.

jakubgs (approved):
NICE.

PR #1703, opened 2019-11-25
Update Travis badge in README to point to develop branch (only)

gliptak:


jakubgs (approved):
I think this makes sense, what do you think @adambabik ?

adambabik (approved):

PR #1706, opened 2019-11-26
Move Message to protobuf

cammellos:
* Use a single Message type `v1/message.go` and `message.go` are the same now, and they embed `protobuf.ChatMessage`

* Use `SendChatMessage` for sending chat messages, this is basically the old `Send` but a bit more flexible so we can send different message types (stickers,commands), and not just text.

* Remove dedup from services/shhext. Because now we process in status-protocol, dedup makes less sense, as those messages are going to be processed anyway, so removing for now, we can re-evaluate if bringing it to status-go or not.

* Change the various retrieveX method to a single one:
`RetrieveAll` will be processing those messages that it can process (Currently only `Message`), and return the rest in `RawMessages` (still transit). The format for the response is:
`Chats`: -> The chats updated by receiving the message
`Messages`: -> The messages retrieved (already matched to a chat)
`Contacts`: -> The contacts updated by the messages
`RawMessages` -> Anything else that can't be parsed, eventually as we move everything to status-protocol-go this will go away.

I have verified status-react works with these changes and status-console-client. Now they use the same code paths for sending/receiving messages, which will help us to check compatibility of features when developing status-go/protocol.

status-react PR:
https://github.com/status-im/status-react/pull/9547

status-console-client PR:
https://github.com/status-im/status-console-client/pull/135

There's one test I could not fix, not sure why, TestHistoricMessages, it seems I am not able to mock it anymore, any idea?

`RetrieveAll` is not quite a large function, I would refactor once is finalized, as there's still code to be added (all the various `ContactRequest` etc)

status: ready


cammellos:
Should have been a draft PR , forgot

protocol/protobuf/message.proto
```diff
@@ -0,0 +1,172 @@
...
+  repeated MembershipUpdate updates = 2;
+  // An optional chat message
+  ChatMessage message = 3;
+}
+
+message MembershipUpdate {
+  // Signature of the concatenated array bytes of the events
+  string signature = 1;
+  // The protobuf encoded events
+  repeated bytes events = 2;
```
adambabik: Why this one is not of `GroupChatEvent` type?

protocol/protobuf/message.proto
```diff
@@ -0,0 +1,172 @@
...
+    COMMAND_REQUEST = 4;
+    EMOJI = 5;
+  }
+}
+
+message ApplicationMetadataMessage {
+  // Signature of the payload field
+  bytes signature = 1;
+  // This is the encoded protobuf of the application level message
+  bytes payload = 2;
```
adambabik, line 58: This is encoded `StatusMessage`?

protocol/protobuf/message.proto
```diff
@@ -0,0 +1,172 @@
...
+    PAIR_INSTALLATION = 3;
+    SYNC_INSTALLATION = 4;
+  }
+}
+
+message StatusMessage {
+  oneof payload {
+    ApplicationMetadataMessage application_metadata_message = 1;
+    // Datasync messages will contain an array of Message
+    bytes datasync_message = 2;
```
adambabik: It's `bytes` because we don't know data sync wrapping?

protocol/protobuf/message.proto
```diff
@@ -0,0 +1,172 @@
+syntax = "proto3";
```
adambabik, line 1: I think we can split this file into logical units like `sync.proto`, `message.proto` etc.

protocol/protobuf/message.proto
```diff
@@ -0,0 +1,172 @@
...
+  // The protobuf encoded events
+  repeated bytes events = 2;
+}
+
+message StickerMessage {
+  string hash = 1;
+  int32 pack = 2;
+}
+
+message ChatMessage {
```
adambabik, line 10: It's not clear how this is used and its relationship with `StatusMessage`.

protocol/protobuf/message.proto
cammellos: I haven't settled on this, but basically it's because if we make it`bytes`, we can more easily change signature scheme without breaking compatibility (also it simplifies signing).

How it is now is that we concatenate each field and we sign the concatenated JSON fields.
i.e (simplified):

```
Sign(field_a + field_b + field_c)
```
and we send the signature.

The issue with this approach is that if we add `field_d`, it will break compatibility (unless we propagate 2 different signatures, one with and one without), as old client will not be using the new field for signature.

One way to overcome this, is to sign the encoded payload (same as we do for the whole message), so that any change inside the payload won't break the signature, at the expenses of having to keep the event payload for propagation.

protocol/protobuf/message.proto
cammellos: We can import them I guess

protocol/protobuf/message.proto
cammellos: No, this is the application layer message, so `ChatMessage`/`SyncInstallation` etc. We use the encoded version for signature reasons (so that any change inside the payload won't break signature scheme or won't be requiring any backward compatibility code)

protocol/protobuf/message.proto
cammellos: I'll add a comment, but basically this is what is in encoded in `ApplicationMetadataMessage` (one of the possible values, which can be `SyncInstallation` `GroupMembershipUpdate` etc

cammellos:
@adambabik @PombeirP this is ready to be reviewed, sorry for the large PR but it required many changes

protocol/message.go
```diff
@@ -1,72 +1,159 @@
...
+		MessageType:      m.MessageType,
+		Sticker:          m.GetSticker(),
+	}
+
+	return json.Marshal(item)
+}
+
+func (m *Message) UnmarshalJSON(data []byte) error {
+	type Alias Message
+	aux := &struct {
```
adambabik:
```suggestion
	aux := struct {
```

protocol/message_processor_test.go
```diff
@@ -33,22 +34,18 @@ type MessageProcessorSuite struct {
...
-		ContentT:  "text/plain",
-		MessageT:  "public-group-user-message",
-		Clock:     154593077368201,
-		Timestamp: 1545930773682,
-		Content: v1protocol.Content{
-			ChatID: "testing-adamb",
-			Text:   "abc123",
-		},
-	}
+	s.testMessage = Message{}
```
adambabik:
```suggestion
	s.testMessage = Message{
		Text: "abc123",
		ChatId: "testing-adamb",
		ContentType: 0, // should be const
		MessageType: 2, // should be const
		Clock: 154593077368201,
		Timestamp: 1545930773682
	}
```

protocol/messenger.go
```diff
@@ -624,187 +624,297 @@ func (m *Messenger) Contacts() ([]*Contact, error) {
 	return m.persistence.Contacts()
 }
 
-func (m *Messenger) Send(ctx context.Context, chatID string, data []byte) ([][]byte, error) {
-	logger := m.logger.With(zap.String("site", "Send"), zap.String("chatID", chatID))
+func timestampInMs() uint64 {
+	return uint64(time.Now().UnixNano()) / (uint64(time.Millisecond) / uint64(time.Nanosecond))
```
```suggestion
	return uint64(time.Now().UnixNano() / int64(time.Millisecond))
```

protocol/messenger.go
```diff
@@ -816,48 +926,130 @@ func (m *Messenger) RetrieveRawAll() (map[transport.Filter][]*v1protocol.StatusM
...
 				if msg.ParsedMessage != nil {
-					if textMessage, ok := msg.ParsedMessage.(v1protocol.Message); ok {
-						textMessage.Content = v1protocol.PrepareContent(textMessage.Content)
-						msg.ParsedMessage = textMessage
+					if textMessage, ok := msg.ParsedMessage.(*protobuf.ChatMessage); ok {
+						receivedMessage := &Message{
+							ChatMessage: *textMessage,
+						}
+						receivedMessage.PrepareContent()
+						receivedMessage.ID = messageID
```
adambabik: We could put these lines into `receivedMessage := &Message{...}`.

protocol/v1/status_message.go
```diff
@@ -147,26 +163,17 @@ func (m *StatusMessage) HandleApplicationMetadata() error {
 }
 
 func (m *StatusMessage) HandleApplication() error {
-	value, err := decodeTransitMessage(m.DecryptedPayload)
+	// Try protobuf first
+	message := &protobuf.ChatMessage{}
```
adambabik:
```suggestion
	var message protobuf.ChatMessage
```

protocol/v1/status_message.go
```diff
@@ -147,26 +163,17 @@ func (m *StatusMessage) HandleApplicationMetadata() error {
...
+
+	err := proto.Unmarshal(m.DecryptedPayload, message)
```
adambabik:
```suggestion
	err := proto.Unmarshal(m.DecryptedPayload, &message)
```

protocol/v1/status_message.go
```diff
@@ -147,26 +163,17 @@ func (m *StatusMessage) HandleApplicationMetadata() error {
...
-		m.MessageType = MembershipUpdateMessageT
-	case PairMessage:
-		m.MessageType = PairMessageT
-		// By default we null the parsed message field, as
-		// otherwise is populated with the raw transit and we are
-		// unable to marshal in case it contains maps
-		// as they have type map[interface{}]interface{}
-	default:
-		m.ParsedMessage = nil
+		m.ParsedMessage = message
```
adambabik, line 174: This way here we will have `nil` instead of a pointer to an empty struct. Not sure if this changes much but later it can be checked with `m.ParsedMessage != nil` which might be useful.

protocol/message_processor_test.go
cammellos: this one would be a bit uglier, as those fields are embedded from `protobuf` so would look something like:
```

s.testMessage = Message{
   ChatMessage: protobuf.ChatMessage{
   .....
   }
}

```

Which one is preferred or there's some other way?

cammellos:
@adambabik addressed feedback and left a question, thanks!

protocol/message_processor_test.go
adambabik: Not bad, it's just one struct nested. Up to you. I generally prefer literal. We should definitely replace `0` and `2` with constants.

adambabik (approved):

protocol/message_processor_test.go
cammellos: will do, thanks

protocol/protobuf/pubkey.go
```diff
@@ -1,11 +1,11 @@
-package applicationmetadata
+package protobuf
 
 import (
 	"crypto/ecdsa"
-	"github.com/status-im/status-go/eth-node/crypto"
+	"github.com/ethereum/go-ethereum/crypto"
```
pedropombeiro: Ideally we would keep geth away from status-protocol dependencies. What's the reason for going back to using `ethereum/go-ethereum/crypto`?

protocol/v1/status_message.go
```diff
@@ -25,26 +26,44 @@ const (
...
+	TransportLayerSigPubKey *ecdsa.PublicKey `json:"-"`
 	// ApplicationMetadataLayerPubKey contains the public key provided by the application metadata layer
-	ApplicationMetadataLayerSigPubKey *ecdsa.PublicKey
+	ApplicationMetadataLayerSigPubKey *ecdsa.PublicKey `json:"-"`
+}
+
+// Temporary JSON marshaling for those messages that are not yet processed
+// by the go code
+func (m *StatusMessage) MarshalJSON() ([]byte, error) {
+	type Alias StatusMessage
```
pedropombeiro: `Alias` doesn't seem to be used anywhere. Leftover?

protocol/protobuf/pubkey.go
cammellos: I think I have missed it in the rebase, well spotted

pedropombeiro (approved):

PR #1709, opened 2019-11-29
Add a way to run bootnode under systemd

jakubgs:
After reviewing some stuff for https://github.com/status-im/design/issues/40 I decided to refactor the way we start a mailserver under systemd. I've made it so it works the same as the container ones, now they also use `Makefile`s.

In addition I've implemented the same thing for running Status bootnodes. But in order to streamline that process I've also implemented the `-genkey` and `-writeaddress` flags for `bootnode` service.


cmd/bootnode/main.go
```diff
@@ -69,6 +83,10 @@ func main() {
 			log.Crit("unable to convert decoded hex into ecdsa.PrivateKey", "data", key, "error", err)
 		}
 	}
+	if *writeAddr {
+		fmt.Printf("%x\n", crypto.FromECDSAPub(&nodeKey.PublicKey)[1:])
```
adambabik, line 88: Why omitting the first byte? It's that compressed/uncompressed info?

jakubgs: I just copied it from `go-ethereum` repo:
https://github.com/ethereum/go-ethereum/blob/c2d65d34/cmd/bootnode/main.go#L90-L93
Any idea why they did it like that?

jakubgs: Oh, I think they are just removing the `0x` from it so it's ready to be used in an `enode://` address.

adambabik: No, it's a bytes slice. `0x` is added when formatting to a string. It's about that uncompressed byte: https://github.com/golang/go/blob/go1.13.1/src/crypto/elliptic/elliptic.go#L310. LGTM but we can add a comment why we do `[1:]`. I assume this byte is omitted in enode ID...?

jakubgs: Okay, I added a comment explaining it. Okay to merge?

adambabik (approved):

PR #1713, opened 2019-12-02
Add ValidateMnemonic method and update tests

### LINKED ISSUE #1663: Method for bip39 mnemonic validation
# Problem

https://github.com/status-im/status-react/issues/9050

We want to recover only valid mnemonics (words from the dictionary and checksum is valid)

# Implementation

validateMnemonic(mnemonicString)

words from the dictionary and checksum are valid

acolytec3:
Add new method to validate checksum as part of BIP39 mnemonic validation

Important changes:
- Adds new ValidateMnemonic that returns error messages indicating cause of when a mnemonic seed phrase fails validation
 - This method is called by the current `ValidMmenonic` method so the current API stays the same

Closes #1663


extkeys/mnemonic.go
```diff
@@ -201,11 +201,11 @@ func (m *Mnemonic) MnemonicPhrase(strength EntropyStrength, language Language) (
...
 }
 
-// ValidMnemonic validates mnemonic string
-func (m *Mnemonic) ValidMnemonic(mnemonic string, language Language) bool {
+// ValidateMnemonic validates that all words from a mnemonic string are in wordlist and that checksum is valid
+func (m *Mnemonic) ValidateMnemonic(mnemonic string, language Language) error {
 	wordList, err := m.WordList(language)
 	if err != nil {
-		return false
+		return errors.New("Invalid language specified")
```
andremedeiros: Errors in Golang aren't capitalized

extkeys/mnemonic.go
```diff
@@ -216,17 +216,27 @@ func (m *Mnemonic) ValidMnemonic(mnemonic string, language Language) bool {
...
 		}
 	}
 
-	return true
+	return nil
+}
+
+// ValidMnemonic validates mnemonic string
+func (m *Mnemonic) ValidMnemonic(mnemonic string, language Language) bool {
+
```
andremedeiros: This method can read:

```go
func (m *Mnemonic) ValidMnemonic(mnemonic string, language Language) bool {
	return m.ValidateMnemonic(mnemonic, language) == nil
}
```

extkeys/mnemonic.go
```diff
@@ -24,6 +24,7 @@ import (
 // https://github.com/trezor/trezor-crypto/blob/master/bip39.c
 // https://github.com/trezor/python-mnemonic/blob/master/mnemonic/mnemonic.py
 // https://github.com/bitpay/bitcore-mnemonic/blob/master/lib/mnemonic.js (used in eth-lightwallet.js)
+// https://github.com/tyler-smith/go-bip39/blob/master/bip39.go
```
andremedeiros, line 27: 👏

andremedeiros:
Great progress so far!

acolytec3:
@andremedeiros Thanks!  All I've got so far is basically porting over most of the code from the reference implementation you linked to earlier and fixed a bug in my new testing code.  I'm getting segmentation faults when I run the tests now so will work on debugging now.

acolytec3:
@andremedeiros @gravityblast I've made some progress.  I fixed all the logic bugs I introduced when I first added the checksum validation and it now builds and will run.  That said, the new test logic I added to `TestMnemonicPhrase` to validate the logic is still failing for some of the phrases in the `mnemonic_vectors.json`.  It seems to work for all of the English, Russian, Italian, Simplified Chinese, some of the Spanish, and some of the French.  It fails for all of the Traditional Chinese and Japanese and some each of the Spanish and French.  Is it possible it's the mnemonics that are flawed in this case or do you think this indicates I still introduced some bug somewhere in the comparison logic?  I don't know the history of all the test data in that set used for testing.

Here's a gist with all the output of the test that validates the Mnemonic phrase - https://gist.github.com/acolytec3/31da77abc6fd7ffc7f8ab61e7e843a7a

acolytec3:
Follow-up here, I found one more typo bug related to incorrect capitalization in the `chinese_traditional` language name that fixes validation for that set of mnemonics.  It's still not working for Japanese or most of the French and Spanish ones and the error that ValidateMnemonic is returning is "words not found in dictionary" so I need to do some further research there since it seems to be passing for a couple of the French and Spanish vectors.

acolytec3:
@gravityblast @andremedeiros I've updated this [gist](https://gist.github.com/acolytec3/31da77abc6fd7ffc7f8ab61e7e843a7a) with some output that highlights the words that the ValidateMnemonic method is not able to find in the predefined `wordLists`.  I also did some debugging in VSCode and set breakpoints where it parses the constructs the `wordMap` to do the lookups for validating each word is in the dictionary and here is what I've found:
* For Japanese - it appears that there are words in some of the precomputed seed vectors that just aren't in the Japanese language dictionary.  For example - あおそら(aosora - sky) is in one of the seed vectors and it's not in the Japanese dictionary.
* For French and Spanish, it looks like Go is having trouble parsing the dictionaries at all.  For example, `ábaco` is identified as a word that doesn't appear in the Spanish dictionary but it's the very first word in the [Spanish dictionary](https://github.com/status-im/status-go/blob/429d2b4291f460a8bced8880ef2f5cdd92bd419d/extkeys/mnemonic.go#L103).  When I look at the `wordMap` I construct to do the word lookups for the mnemonic, the debugger  shows a bunch of garbage in the wordMap variable and some IO errors.

So, all that to say, I think my code is working correctly and the issue is the way I updated `mnemonic_test.go` to run the ValidateMnemonic method against each of the precomputed seed vectors. The ValidateMnemonic is passing for tests where it's generating a new Mnemonic and the majority of the precomputed vectors except where outlined above.

 Thoughts on whether to just remove that piece of the test?  Not sure on the way forward here.

gravityblast:
hey @acolytec3, thanks for your work!

> For French and Spanish, it looks like Go is having trouble parsing the dictionaries at all. For example, ábaco is identified as a word that doesn't appear in the Spanish dictionary but it's the very first word in the Spanish dictionary. When I look at the wordMap I construct to do the word lookups for the mnemonic, the debiggerto shows a bunch of garbage in the wordMap variable and some IO errors.

I checked the vector and found out that the encoding of some words are not the same compared to the list we have in go. For example in French we have the word dénicher:

in the vector is: `dénicher`
in the go list is: `dénicher`

but if you try even with javascript,  `"dénicher" == "dénicher"` is `false`.
The `e`s with the accent are different: 

`"é" == "é"`

In fact I see they are different bytes:
```
irb(main):010:0> "é".bytes
=> [101, 204, 129]
irb(main):011:0> "é".bytes
=> [195, 169]
```

I think the best way is comparing them to the original list from the BIP to understand which ones are the right ones.

acolytec3:
@gravityblast so what's the next step? Do I need to just update the dictionaries in our `mnemonic.go` file to match the words from the original BIP proposal? Trying to understand if this character encoding mismatch is related to my PR or just a data issue in the seed vectors.

gravityblast:
I don't think it's related to this PR because you didn't touch the list nor the vector right?
The best way should be understanding which one is different and fix it.

flexsurfer:
@acolytec3 just to make sure you're aware we need this method to be available in status-react, probably you could work on this one as well https://github.com/status-im/status-react/issues/9050, so we can test it before merging status-go side

acolytec3:
Correct.  I just added a step in the test that validates the seed vectors to run the `ValidateMnemonic` function as part of the test and that wasn't there before so the words in the seed vectors weren't being compared against the word lists before as far as I can tell.  If I comment out that new section, the tests run fine.  [This](https://github.com/status-im/status-go/blob/429d2b4291f460a8bced8880ef2f5cdd92bd419d/extkeys/mnemonic_test.go#L66) section right here is still executing the `ValidateMnemonic` method I added in my PR as part of the test of validating that a newly generated mnemonic and that test is passing which indicates that the checksum validation I added is working correctly.  

So, to my original question, should I remove that additional section of testing scope since it's not proving anything?  We could run `ValidateMnemonic` against all of the English seed vectors and it passes for all of them and then add a negative test for one of the Japanese ones where it throws a "word not found in wordlist" error.  

Beyond that, is there anything else we need in this PR since my understanding from @andremedeiros is just that we're adding a checksum validation to the current `ValidateMnemonic` scope to ensure that any time status-react passes in a seed phrase, we're validating the checksum as well as the pre-existing checks already there to check if the words are in the dictionary or not.

acolytec3:
I've spent an hour poking around the codebase trying to understand how individual methods are exposed to status-react in the API without much success.  Do I need to add some new function under `GethStatusBackEnd` in the API that wraps the `ValidMnemonic` method?  As far as I can tell, `ValidMnemonic` isn't exposed as a method called by status-react today so wondering if I need to add a wrapper function under `GethStatusBackEnd` to call this particular method.

adambabik:
@acolytec3 you can use https://github.com/status-im/status-go/blob/develop/mobile/status.go#L226-L248 as an example. It should look more or less like this:
```go
func ValidateMnemonic(mnemonic string) string {
    err := statusBackend.ValidateMnemonic(mnemonic)
    // or err := statusBackend.AccountManager().ValidateMnemonic(mnemonic)
    return makeJSONResponse(err)
}
```

extkeys/mnemonic.go
```diff
@@ -53,14 +54,14 @@ const (
 
 // Languages is a list of supported languages for which mnemonic keys can be generated
 var Languages = [...]string{
-	"English",
```
andremedeiros: Why did this change?

extkeys/mnemonic.go
```diff
@@ -216,17 +232,61 @@ func (m *Mnemonic) ValidMnemonic(mnemonic string, language Language) bool {
...
+	if checksum.Cmp(computedChecksum) != 0 {
+		return errors.New("checksum for mnemonic seed is invalid")
+	}
+
+	return nil
+}
+
+// ValidMnemonic validates mnemonic string
+func (m *Mnemonic) ValidMnemonic(mnemonic string, language Language) bool {
+	return m.ValidateMnemonic(mnemonic, language) == nil
```
andremedeiros, line 289: Needs to implement @adambabik's suggested change.

extkeys/mnemonic.go
```diff
@@ -216,17 +232,61 @@ func (m *Mnemonic) ValidMnemonic(mnemonic string, language Language) bool {
...
+	}
+
+	b := big.NewInt(0)
+
 	// Check if all words belong in the wordlist
 	for i := 0; i < numOfWords; i++ {
-		if !contains(wordList, words[i]) {
-			return false
+
+		wordIndex, found := wordMap[words[i]]
```
andremedeiros: The pattern is usually

```go
wordIndex, ok := wordmap[words[i]]
if !ok {
	return errors.New("not all words are found in dictionary")
}
```

andremedeiros (changes requested):
Just a couple minor changes and some questions. Really excited to merge this! 🎉

extkeys/mnemonic.go
acolytec3: I'd changed them to lower-case to match the case of the languages specified in the seed vectors when I was running the `ValidateMnemonic` test against all of the seed vectors.  This can be changed back as needed I think.

acolytec3:
> @acolytec3 you can use https://github.com/status-im/status-go/blob/develop/mobile/status.go#L226-L248 as an example. It should look more or less like this:
> 
> ```go
> func ValidateMnemonic(mnemonic string) string {
>     err := statusBackend.ValidateMnemonic(mnemonic)
>     // or err := statusBackend.AccountManager().ValidateMnemonic(mnemonic)
>     return makeJSONResponse(err)
> }
> ```

I've done the below locally and it seems to compile.  I will do some testing to see if I can call this method successfully or not.  Can I interact with the statusd server and call all the functions identified in `status.go` using geth if I enable IPC mode?
```
func ValidateMnemonic(mnemonic string) string {
	err := extkeys.ValidateMnemonic(mnemonic, "english")
	return makeJSONResponse(err)
}
```

extkeys/mnemonic.go
acolytec3: Understood, will be updated in next commit.

acolytec3:
I have a question about how to use this forked version of status-go with a local version of status-react so I can test it with the front-end.  I've followed [these steps](https://status.im/build_status/#Locally-built-status-go-dependency) and tried to start the clojurescript build in status-react with `make startdev-android-real STATUS_GO_SRC_OVERRIDE=$GOPATH/my/local/status-go/directory` and it goes through the whole process of compiling status-go and then fails at the very end with the below error output
```
# github.com/status-im/status-go/mobile
go/src/github.com/status-im/status-go/mobile/status.go:711:9: undefined: extkeys.ValidateMnemonic
/nix/store/n2yhfjmk1qxl9vs4i378lik5vmrqjb1g-gomobile-20190719-d2bd2a2-bin/bin/gomobile: go build -v -ldflags -X github.com/status-im/status-go/params.GitCommit=unknown -X github.com/status-im/status-go/params.Version=develop -s -w -buildmode=c-shared -o=/build/gomobile-work/android/src/main/jniLibs/armeabi-v7a/libgojni.so gobind failed: exit status 2

builder for '/nix/store/qh4v606li2kgyvrzkybdz8lwbs09rwmz-status-go-develop-unknown-android.drv' failed with exit code 1
error: build of '/nix/store/qh4v606li2kgyvrzkybdz8lwbs09rwmz-status-go-develop-unknown-android.drv' failed
Makefile:211: recipe for target '_watch-android-real' failed
make[1]: *** [_watch-android-real] Error 100
make[1]: Leaving directory '/home/jim/status-react'
make[1]: Entering directory '/home/jim/status-react'
Unknown target device '_startdev-android-real'. Supported targets:
startdev-android-avd
startdev-android-genymotion
startdev-android-real
startdev-desktop
startdev-ios-real
startdev-ios-simulator
Makefile:340: recipe for target '_unknown-startdev-target-_startdev-android-real' failed
make[1]: *** [_unknown-startdev-target-_startdev-android-real] Error 1
make[1]: Leaving directory '/home/jim/status-react'
Makefile:346: recipe for target '_startdev-android-real' failed
make: *** [_startdev-android-real] Error 2
```
It looks like it's not compiling from the right source because when I run `make status-go` on the branch with the code from this PR in it, it compiles just fine and I can run `build/bin/statusd` without issue.

acolytec3:
@gravityblast @andremedeiros @adambabik Any suggestions on how to fix the above error.  Further research on my side says that it's an import error where `status.go` isn't correctly importing `extkeys`.   I'm getting a `ValidateMnemonic not declared by package extkeys` even though go.mod is pointing to the correct local extkeys directly in the status-go root directoy.  Is there something obvious that I'm missing about how package imports work?

extkeys/mnemonic.go
```diff
@@ -216,17 +232,61 @@ func (m *Mnemonic) ValidMnemonic(mnemonic string, language Language) bool {
...
+	b := big.NewInt(0)
+
 	// Check if all words belong in the wordlist
 	for i := 0; i < numOfWords; i++ {
-		if !contains(wordList, words[i]) {
-			return false
+
+		wordIndex, ok := wordMap[words[i]]
+		if !ok {
+			return errors.New("not all words are found in dictionary")
```
adambabik:
```suggestion
			return fmt.Errorf("word %s not found in the dictionary", words[i])
```

extkeys/mnemonic.go
```diff
@@ -216,17 +232,61 @@ func (m *Mnemonic) ValidMnemonic(mnemonic string, language Language) bool {
...
 		}
+		var wordBytes [2]byte
+		binary.BigEndian.PutUint16(wordBytes[:], uint16(wordIndex))
+		b = b.Mul(b, rightShift11BitsDivider)
+		b = b.Or(b, big.NewInt(0).SetBytes(wordBytes[:]))
+	}
+
+	checksum := big.NewInt(0)
+	checksumMask := wordLengthChecksumMasksMapping[len(words)]
```
adambabik: This can return `nil` and it's not checked in the line below. This can result in panic. We either make sure this case is tested or we do a check.

```suggestion
	checksumMask := wordLengthChecksumMasksMapping[numOfWords]
```

adambabik:
@acolytec3 the problem is that `ValidateMnemonic` is not exported in `extkeys`. This is a method of `Mnemonic` struct. I suggest to replace it in `status.go`:
```go
func ValidateMnemonic(mnemonic, lang string) string {
        m := extkeys.NewMnemonic()
	err := m.ValidateMnemonic(mnemonic, lang)
	return makeJSONResponse(err)
}
```

extkeys/mnemonic.go
acolytec3: We check earlier in the code to verify that the number of words equals one of the acceptable values and return an error if not.
```
if numOfWords%3 != 0 || numOfWords < 12 || numOfWords > 24 {
		return errors.New("mnemonic contains an invalid number of words")
	}
```

Is it still possible we get a `nil` back from this line?

acolytec3:
@adambabik Thanks!  Completely missed that my function was part of `Mnemonic`.  Will correct in the next commit.

acolytec3:
@adambabik I made the changes you suggested in `status.go`and it appears to build again (at least my local environment.  Looks like jenkins still can't build it for Android or iOS though so let me know .  That said, status-react still throws an error along the same lines `m.ValidateMnemonic undefined (type *extkeys.Mnemonic has no field or method ValidateMnemonic)` when I try to use my local version of status-go with  `make startdev-android-real STATUS_GO_SRC_OVERRIDE=$GOPATH/my/local/status-go/directory` and I'm assuming that must be related to the android build failing?

adambabik:
@acolytec3 it might be because iOS and Android uses `vendor/` dir because gomobile does not support Go Modules properly yet. Can you try run `make modvendor-install` and then `make vendor`?

acolytec3:
@adambabik Thanks, that seems to have solved it.  I'll start working on the status-react side now.

@gravityblast @andremedeiros Is there anything else for this PR?  The android/ios builds are failing for Jenkins so not sure if there's something more that needs to be corrected.

extkeys/mnemonic.go
acolytec3: This is done in latest commit. Let me know if it needs anything further.

extkeys/mnemonic.go
acolytec3: Do you want me to change this back?

extkeys/mnemonic.go
adambabik: No, unless one shuffles the code around :)

extkeys/mnemonic.go
adambabik: Are these seed vectors in this repo? Or it was a temporary change and tested with a code that is not included in this PR? I would minimize the number of changes if not necessary.

adambabik (approved):

adambabik:
@acolytec3 you need to run `make modvendor-install` and then `make vendor` and then commit all changed files from `vendor/` to make the CI tests pass.

acolytec3:
@adambabik Regarding the language case and the question of seed vectors, in [this commit[(https://github.com/status-im/status-go/pull/1713/commits/c0034966e1a51f78d658e545e494bd36eada98a8), I changed them so that I could add a new section of the `mnemonic_test.go` that validated all of the seed phrases [here](https://github.com/status-im/status-go/blob/develop/extkeys/mnemonic_vectors.json).  That said, some of them have words not in the BIP39 dictionaries and there were some I/O errors related to words with certain accents on them in Italian and French where the word appears to be the same as the dictionary but it's encoded differently in the seed vector.  Regardless, I removed that section of the test since fixing those seed vectors to be valid based on checksum/dictionary checks is beyond the scope of the PR.  I'll revert the language back to UpperCase.

acolytec3:
I ran `make modvendor-install` and `make vendor` and then fixed the merge commits (as best I understood them).

adambabik (approved):

acolytec3:
Are y'all ready for me to squash?

andremedeiros:
Do it!

adambabik:
And someone from the org needs to merge it. @andremedeiros can you approve this PR so that I can do it?

acolytec3:
@adambabik @andremedeiros After squashing, looks like the tests are failing on Jenkins.  What's the error?

andremedeiros:
They're green here, so I'm approving.

andremedeiros (approved):

PR #1714, opened 2019-12-02
Limit the upper bounds of how many pending peers we track

andremedeiros:
This change introduces an upper bound on how many pending peers we track, in order to prevent unbounded growth.

Closes status-im/trailofbits-audit#38
Closes status-im/trailofbits-audit#39

### DIFF
--- a/peers/topicpool.go
+++ b/peers/topicpool.go
@@ -24,6 +24,10 @@ const (
 // to get the maximum number of cached peers allowed.
 var maxCachedPeersMultiplier = 1
 
+// maxPendingPeersMultiplier peers max limit will be multiplied by this number
+// to get the maximum number of pending peers allowed.
+var maxPendingPeersMultiplier = 2
+
 // TopicPoolInterface the TopicPool interface.
 type TopicPoolInterface interface {
 	StopSearch(server *p2p.Server)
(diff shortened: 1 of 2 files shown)


peers/topicpool.go
```diff
@@ -85,12 +91,14 @@ type TopicPool struct {
 
 	pendingPeers         map[enode.ID]*peerInfoItem // contains found and requested to be connected peers but not confirmed
 	discoveredPeersQueue peerPriorityQueue          // priority queue to find the most recently discovered peers; does not containt peers requested to connect
+	discoveredPeers      map[enode.ID]bool          // remembers which peers have already been discovered and are enqueued
```
adambabik, line 94: Unit tests to verify we manage this set properly would be nice.

peers/topicpool.go
```diff
@@ -101,12 +109,32 @@ func (t *TopicPool) addToPendingPeers(peer *peerInfo) {
 		peerInfo: peer,
 		index:    notQueuedIndex,
 	}
+
+	if t.maxPendingPeers == 0 || t.maxPendingPeers > len(t.pendingPeers) {
```
adambabik: Plus unit test for this one maybe.

cammellos: `0` means unbound correct? If so we should probably add a comment explaining how values are interpreted

andremedeiros: This is kind of a point of contention that I should have left a PR comment on.

For some reason the tests expicitly set limits to 0, both upper and lower bounds, so I'm not sure what the right thing to do is here.

On one hand, I want unbounded to be an option still, on the other hand, 0 is a magic value that I'm not fond of and maybe we should update the tests.

What do you two think?

peers/topicpool.go
andremedeiros: Agreed. On it.

peers/topicpool.go
cammellos: I feel in general `-1` is better to specify no limits, as there's no ambiguity, while 0 might mean literally "don't have any", although in this particular case is a bit painful to have exactly `-1` as it's multiplied (could be less than 0 though), but I don't have a strong opinion, both 0 or -1 is fine by me, as long as there's a comment/test

andremedeiros: Do we update tests too to specify limits above 0? Some of these tests actually rely on nodes being remembered _while_ setting limits at 0.

andremedeiros:
@adambabik @cammellos please re-review.

corpetty (approved):

corpetty (approved):

adambabik (approved):

PR #1717, opened 2019-12-05
APIs to Get and Create custom Tokens

andremedeiros:
Paired with Eric


go.mod
```diff
@@ -32,15 +39,19 @@ require (
 	github.com/okzk/sdnotify v0.0.0-20180710141335-d9becc38acbd
 	github.com/pborman/uuid v1.2.0
 	github.com/pkg/errors v0.8.1
-	github.com/prometheus/client_golang v1.2.1
+	github.com/prometheus/client_golang v1.0.0
```
adambabik: Something went wrong here. I recommend checking out `go.mod` from develop and just run  `go mod tidy` to get minimal update and if all looks good `make vendor`.

services/wallet/database.go
```diff
@@ -273,6 +273,66 @@ SELECT blocks.hash, blk_number FROM accounts_to_blocks JOIN blocks ON blk_number
...
+func (db *Database) AddCustomToken(token Token) (err error) {
+	var (
+		tx     *sql.Tx
+		insert *sql.Stmt
+	)
+	tx, err = db.db.Begin()
+	if err != nil {
+		return
+	}
+	insert, err = tx.Prepare("INSERT OR REPLACE INTO TOKENS (network_id, address, name, symbol, decimals, color) VALUES (?, ?, ?, ?, ?, ?)")
```
adambabik: I think this should be after `defer`.

yenda (approved):

services/wallet/database.go
```diff
@@ -273,6 +273,67 @@ SELECT blocks.hash, blk_number FROM accounts_to_blocks JOIN blocks ON blk_number
 	return nil, err
 }
 
+type Token struct {
+	Address  common.Address `json:"address"`
+	Name     string         `json:"name"`
+	Symbol   string         `json:"symbol"`
+	Decimals uint           `json:"decimals"`
```
cammellos: What is `Decimal` exactly? I think the other fields are self-explanatory, but maybe worth leaving a comment describing `Decimals`

services/wallet/database.go
```diff
@@ -273,6 +273,67 @@ SELECT blocks.hash, blk_number FROM accounts_to_blocks JOIN blocks ON blk_number
...
+	Color    string         `json:"color"`
+}
+
+func (db *Database) GetCustomTokens() ([]*Token, error) {
+	rows, err := db.db.Query(`SELECT address, name, symbol, decimals, color FROM tokens WHERE network_id = ?`, db.network)
```
cammellos, line 289: You need to close `rows` otherwise it will hang eventually.
`defer rows.Close()`
although close might return an error, generally we ignore it, otherwise before any return

services/wallet/database.go
```diff
@@ -273,6 +273,67 @@ SELECT blocks.hash, blk_number FROM accounts_to_blocks JOIN blocks ON blk_number
...
+
+	return rst, nil
+}
+
+func (db *Database) AddCustomToken(token Token) (err error) {
+	var (
+		tx     *sql.Tx
+		insert *sql.Stmt
+	)
+	tx, err = db.db.Begin()
```
cammellos: no need of a transaction here if you are making a single call

cammellos (changes requested):

services/wallet/database.go
adambabik: It's important to do it AFTER `err` check otherwise `rows` might be missing.

services/wallet/database.go
```diff
@@ -273,6 +273,67 @@ SELECT blocks.hash, blk_number FROM accounts_to_blocks JOIN blocks ON blk_number
...
+	rows, err := db.db.Query(`SELECT address, name, symbol, decimals, color FROM tokens WHERE network_id = ?`, db.network)
+
+	if err != nil {
+		return nil, err
+	}
+
+	var rst []*Token
+	for rows.Next() {
+		token := &Token{}
+		err = rows.Scan(&token.Address, &token.Name, &token.Symbol, &token.Decimals, &token.Color)
```
adambabik:
```suggestion
		err := rows.Scan(&token.Address, &token.Name, &token.Symbol, &token.Decimals, &token.Color)
```

services/wallet/database.go
```diff
@@ -273,6 +273,67 @@ SELECT blocks.hash, blk_number FROM accounts_to_blocks JOIN blocks ON blk_number
...
+
```
adambabik:
```suggestion
```

services/wallet/database.go
```diff
@@ -273,6 +273,67 @@ SELECT blocks.hash, blk_number FROM accounts_to_blocks JOIN blocks ON blk_number
...
+	Address  common.Address `json:"address"`
+	Name     string         `json:"name"`
+	Symbol   string         `json:"symbol"`
+	Decimals uint           `json:"decimals"`
+	Color    string         `json:"color"`
+}
+
+func (db *Database) GetCustomTokens() ([]*Token, error) {
+	rows, err := db.db.Query(`SELECT address, name, symbol, decimals, color FROM tokens WHERE network_id = ?`, db.network)
+
```
adambabik:
```suggestion
```

adambabik (changes requested):

adambabik (approved):
Two nitpicks :)

services/wallet/database.go
andremedeiros: Why do we want to shadow `err` here?

services/wallet/database.go
```diff
@@ -273,6 +273,55 @@ SELECT blocks.hash, blk_number FROM accounts_to_blocks JOIN blocks ON blk_number
...
+		rst = append(rst, token)
+	}
+
+	return rst, nil
+}
+
+func (db *Database) AddCustomToken(token Token) (err error) {
+	var insert *sql.Stmt
+	insert, err = db.db.Prepare("INSERT OR REPLACE INTO TOKENS (network_id, address, name, symbol, decimals, color) VALUES (?, ?, ?, ?, ?, ?)")
```
cammellos: don't need to split `var` , just `insert, err :=...`

cammellos (approved):

services/wallet/database.go
cammellos: oh I see, named parameter

PR #1727, opened 2019-12-09
Handle membership update message

cammellos:

1) Handle membership updates using protobuf and adds the relevant
endpoints.
2) Store in memory a map of chats + contacts for faster lookups, which
are then flushed to disk on each update
3) Validate incoming messages

Sorry for the large pr, but you know, v1 :)


protocol/chat.go
```diff
@@ -5,12 +5,24 @@ import (
...
+	"math/rand"
+	"time"
 
 	"github.com/status-im/status-go/eth-node/crypto"
 	"github.com/status-im/status-go/eth-node/types"
+	"github.com/status-im/status-go/protocol/protobuf"
 	v1protocol "github.com/status-im/status-go/protocol/v1"
 )
 
+var chatColors = []string{
```
adambabik, line 16: Do they have some meaning? If so we can add comments.

protocol/chat.go
```diff
@@ -207,18 +204,23 @@ func CreateOneToOneChat(name string, publicKey *ecdsa.PublicKey) Chat {
 }
 
 func CreatePublicChat(name string) Chat {
+	rand.Seed(time.Now().Unix())
```
adambabik: Isn't it enough to call it once or don't call it at all? Not sure what's the penalty hit for doing it over and over in `CreatePublicChat`.

protocol/message_handler.go
```diff
@@ -1,68 +1,127 @@
...
 
+	"github.com/status-im/status-go/eth-node/crypto"
+	"github.com/status-im/status-go/eth-node/types"
+	"github.com/status-im/status-go/protocol/protobuf"
 	v1protocol "github.com/status-im/status-go/protocol/v1"
 )
 
-type persistentMessageHandler struct {
-	persistence *sqlitePersistence
+var defaultSystemMessagesTranslations = map[protobuf.MembershipUpdateEvent_EventType]string{
```
adambabik: I would probably just create a function:
```go
type membershipUpdateArgs struct {
    From string
    Name string
    Members []string
    Member string
}

func membershipUpdateMessage(e v1protocol.MembershipUpdateEvent) string {
    switch e.Type {
    case protobuf.MembershipUpdateEvent_CHAT_CREATED:
        return fmt.Sprintf("%s created the group %s", e.From, e.Name)
    ...
    }
    
}
```

protocol/message_handler.go
```diff
@@ -1,68 +1,127 @@
...
-func newPersistentMessageHandler(persistence *sqlitePersistence) *persistentMessageHandler {
-	return &persistentMessageHandler{persistence: persistence}
+func Tsprintf(format string, params map[string]string) string {
+	for key, val := range params {
+		format = strings.Replace(format, "{{"+key+"}}", fmt.Sprintf("%s", val), -1)
+	}
+	return fmt.Sprintf(format)
+}
+
+func eventToSystemMessage(e v1protocol.MembershipUpdateEvent, translations map[protobuf.MembershipUpdateEvent_EventType]string) *Message {
```
adambabik: Which should simplify this to:
```go
text := membershipUpdateMessage(e)
```

protocol/protobuf/application_metadata_message.proto
```diff
@@ -0,0 +1,22 @@
...
+package protobuf;
+
+message ApplicationMetadataMessage {
+  // Signature of the payload field
+  bytes signature = 1;
+  // This is the encoded protobuf of the application level message, i.e ChatMessage
+  bytes payload = 2;
+
+  // The type of protobuf message sent
+  MessageType message_type = 3;
```
adambabik:
```suggestion
  Type type = 3;
```

Does it have the same meaning?

protocol/message_handler.go
cammellos: They need to be passed in as config though, as we need to support multiple languages, it's not pretty, but I think it's useful to generate system-messages in status-go (so that also console client can use them and does not have to re-implement for example), but in status-react we support multiple languages, so that's the best I can come up with.

otherwise we could have passed system messages to status-react, then s.r. would translate and save them, but is more convoluted, as you can't just pass them as messages, you need to have a new type as you need to carry some of the info that a message does not have.

A third option would be to generate system messages in status-react (as before) and save them, but then any client needs to do something similar, and I think it's quite useful, and you have this mix of processing between s.r. and s.go.

protocol/chat.go
cammellos: I think call it once at program start is enough, you are right.

I believe if you don't call it, it will always have the same seed, so the chats will always be created with the same color order, and you'd be hitting the first colors much more.

cammellos:
@adambabik thanks for the feedback, addressed, the only part that I did not address is the translations, as we need to support multiple languages, and they need to be passed (or at least we need to have the option) by status-react. Not sure if there's a better or alternative way. Thanks!

protocol/message_handler.go
adambabik: Kinda understand but don't really have a full picture :) 

Maybe we can extract it to a separate file then? I feel like it can still be improved but the first step is to maybe better encapsulate and separate it from the main logic.

cammellos: yes, I can put it in a `group_chat_system_messages.go`, it should be a bit clearer, thanks

cammellos: separated

adambabik (approved):

PR #1731, opened 2019-12-13
Add Commands

cammellos:
This commit adds handling of Transaction commands.

It includes some changes for V1: 
- Handling contact updates
- Handling sync installation


eth-node/core/types/transaction.go
```diff
@@ -0,0 +1,39 @@
...
+		nonce:      nonce,
+		amount:     amount,
+		gasLimit:   gasLimit,
+		gasPrice:   gasPrice,
+		data:       data,
+		checkNonce: checkNonce,
+	}
+}
+
+func (m Message) From() types.Address { return m.from }
```
adambabik, line 32: We don't need getters. I think you can safely export these fields in `Message`.

eth-node/core/types/transaction.go
```diff
@@ -0,0 +1,39 @@
...
+	from       types.Address
+	nonce      uint64
+	amount     *big.Int
+	gasLimit   uint64
+	gasPrice   *big.Int
+	data       []byte
+	checkNonce bool
+}
+
+func NewMessage(from types.Address, to *types.Address, nonce uint64, amount *big.Int, gasLimit uint64, gasPrice *big.Int, data []byte, checkNonce bool) Message {
```
adambabik, line 19: Following the convention, `New` should return a pointer. You won't need this at all if you export `Message` fields.

protocol/chat.go
```diff
@@ -203,6 +203,13 @@ func oneToOneChatID(publicKey *ecdsa.PublicKey) string {
 	return types.EncodeHex(crypto.FromECDSAPub(publicKey))
 }
 
+func OneToOneFromPublicKey(pk *ecdsa.PublicKey) (*Chat, error) {
```
adambabik:
```suggestion
func OneToOneFromPublicKey(pk *ecdsa.PublicKey) *Chat {
```

protocol/go.mod
```diff
@@ -23,6 +23,7 @@ require (
 	github.com/russolsen/transit v0.0.0-20180705123435-0794b4c4505a
 	github.com/status-im/doubleratchet v3.0.0+incompatible
 	github.com/status-im/migrate/v4 v4.6.2-status.2
+	github.com/status-im/status-go v0.38.0 // indirect
```
adambabik: Should not be here.

protocol/message_builder.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"crypto/ecdsa"
+
+	"github.com/status-im/status-go/eth-node/crypto"
+	"github.com/status-im/status-go/eth-node/types"
+	"github.com/status-im/status-go/protocol/identity/alias"
+	"github.com/status-im/status-go/protocol/identity/identicon"
+)
+
+func buildBasicMessage(message *Message, chat *Chat, key *ecdsa.PublicKey) error {
```
adambabik:
```suggestion
func extendMessageFromChat(message *Message, chat *Chat, key *ecdsa.PublicKey) error {
```

protocol/message_validator.go
```diff
@@ -3,9 +3,92 @@ package protocol
 import (
 	"errors"
 	"github.com/status-im/status-go/protocol/protobuf"
+	"strconv"
 	"strings"
 )
 
+func ValidateReceivedSendTransaction(message *protobuf.SendTransaction) error {
```
adambabik, line 31: Maybe having something like this would be useful:
```go
type Validator struct {
    message interface{}
}

func NewValidator(message interface{}) *Validator {
    return &Validator{message: message}
}

func (v *Validator) Validate() error {
    switch vm := v.message.(type) {
    case *protobuf.SendTransaction:
        return v.validateSendTransaction()
    case *protobuf.RequestAddressForTransaction
        return ...
    }
}
```

Or even have a single function `func ValidateMessage(message interface{}) error` with this switch inside. This way from all the places you could call `ValidateMessage`.

protocol/migrations/sqlite/000001_init.up.db.sql
```diff
@@ -1,31 +1,90 @@
+CREATE TABLE IF NOT EXISTS chats (
+id VARCHAR PRIMARY KEY ON CONFLICT REPLACE,
```
adambabik: Indentation missing.

protocol/transaction_validator.go
```diff
@@ -0,0 +1,289 @@
...
+
+const (
+	transferFunction        = "a9059cbb"
+	tokenTransferDataLength = 68
+	transactionHashLength   = 66
+)
+
+type TransactionValidator struct {
+	persistence *sqlitePersistence
+	addresses   map[string]bool
```
adambabik, line 27: If checking only existence:
```suggestion
	addresses   map[string]struct{}
```

eth-node/core/types/transaction.go
cammellos: these are copied from `geth` should we not keep them the same for consistency?

eth-node/core/types/transaction.go
cammellos: same here, all this is `geth` stuff, what should we do? copy and modify or just copy as it is?

protocol/message_validator.go
cammellos: I might do that in the next PR if that's ok

cammellos:
@adambabik ready to be reviewed again, thanks!

protocol/message.go
```diff
@@ -19,13 +19,49 @@ type QuotedMessage struct {
...
+	// Address is the address sent with the command
+	Address string `json:"address"`
+	// Contract is the contract address for ERC20 tokens
+	Contract string `json:"contract"`
+	// Value is the value as a string sent
+	Value string `json:"value"`
+	// TransactionHash is the hash of the transaction
+	TransactionHash string `json:"transactionHash"`
+	// CommandState is the state of the command
+	CommandState CommandState
```
adambabik: We don't want to have `json:"commandState"` here? Or omit it?

cammellos: Forgot to change it, I was testing with the capital letter, good point

eth-node/core/types/transaction.go
adambabik: Oh sorry. If it's copied then it's fine.

protocol/message_processor.go
```diff
@@ -171,13 +171,41 @@ func (p *messageProcessor) sendPrivate(
...
+	messageType protobuf.ApplicationMetadataMessage_Type,
+) ([]byte, error) {
+	p.logger.Debug("sending private message", zap.Binary("recipient", crypto.FromECDSAPub(recipient)))
+
+	wrappedMessage, err := p.wrapMessageV1(data, messageType)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to wrap message")
+	}
+
+	messageID := v1protocol.MessageID(&p.identity.PublicKey, wrappedMessage)
```
adambabik: Nitpick: It's used at the bottom so we can push it just before `p.transport.Track(...)`.

protocol/message_validator.go
```diff
@@ -2,11 +2,126 @@ package protocol
...
 	"errors"
+	"strconv"
 	"strings"
 
 	"github.com/status-im/status-go/protocol/protobuf"
 )
 
+func ValidateReceivedPairInstallation(message *protobuf.PairInstallation) error {
+	if message.Clock == 0 {
+		return errors.New("Clock can't be 0")
```
adambabik: Errors should be lower case so that later they are joined in this manner: `error1 due to reason 1: error2 due to reason 2`.

```suggestion
		return errors.New("clock can't be 0")
```

protocol/message_validator.go
```diff
@@ -2,11 +2,126 @@ package protocol
...
+	}
+
+	if len(strings.TrimSpace(message.Name)) == 0 {
+		return errors.New("Name can't be empty")
```
adambabik:
```suggestion
		return errors.New("name can't be empty")
```

protocol/message_validator.go
```diff
@@ -2,11 +2,126 @@ package protocol
...
+	}
+
+	if len(strings.TrimSpace(message.DeviceType)) == 0 {
+		return errors.New("Device type can't be empty")
```
adambabik:
```suggestion
		return errors.New("device type can't be empty")
```

protocol/message_validator.go
```diff
@@ -2,11 +2,126 @@ package protocol
...
+	}
+
+	if len(strings.TrimSpace(message.InstallationId)) == 0 {
+		return errors.New("InstallationId can't be empty")
```
adambabik:
```suggestion
		return errors.New("installationId can't be empty")
```

protocol/persistence.go
```diff
@@ -347,6 +348,95 @@ func (db sqlitePersistence) Contacts() ([]*Contact, error) {
...
+	// Encode recipients
+	var encodedRecipients bytes.Buffer
+	encoder := gob.NewEncoder(&encodedRecipients)
+
+	if err := encoder.Encode(pubKeys); err != nil {
+		return err
+	}
+
+	_, err := db.db.Exec(`INSERT INTO raw_messages(
+                id,
```
adambabik: Indentation :)

services/shhext/service.go
```diff
@@ -179,6 +188,83 @@ func (s *Service) retrieveMessagesLoop(tick time.Duration, cancel <-chan struct{
 	}
 }
 
+type VerifyTransactionClient struct {
```
adambabik: It looks like it can be private.

services/shhext/service.go
```diff
@@ -179,6 +188,83 @@ func (s *Service) retrieveMessagesLoop(tick time.Duration, cancel <-chan struct{
...
+				log.Error("failed to retrieve accounts", "err", err)
+			}
+			var wallets []types.Address
+			for _, account := range accounts {
+				if account.Wallet {
+					wallets = append(wallets, types.BytesToAddress(account.Address.Bytes()))
+				}
+			}
+
+			response, err := s.messenger.ValidateTransactions(client, wallets)
```
adambabik: It should take a context as an argument. The scenario is that the request fires but `cancel` channel is closed. The request should be interrupted and the loop should be exited immediately. Currently, it will wait for the request to finish.

protocol/transaction_validator.go
```diff
@@ -0,0 +1,317 @@
...
+	Signature []byte
+	From      *ecdsa.PublicKey
+}
+
+func NewTransactionValidator(addresses []types.Address, persistence *sqlitePersistence, logger *zap.Logger) *TransactionValidator {
+	addressesMap := make(map[string]bool)
+	for _, a := range addresses {
+		addressesMap[strings.ToLower(a.Hex())] = true
+	}
+	logger.Debug("Checking addresses", zap.Any("addrse", addressesMap))
```
adambabik, line 52:
```suggestion
	logger.Debug("Checking addresses", zap.Any("addresses", addressesMap))
```

protocol/transaction_validator.go
```diff
@@ -0,0 +1,317 @@
...
+
+	data := transaction.Data()
+	if len(data) != tokenTransferDataLength {
+		return nil, errors.New(fmt.Sprintf("wrong data length: %d", len(data)))
+	}
+
+	functionCalled := hex.EncodeToString(data[:4])
+
+	if functionCalled != transferFunction {
+		return &VerifyTransactionResponse{}, nil
```
adambabik: This line looks suspicious. Is it correct?

protocol/transaction_validator.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+
+	actualContractAddress := strings.ToLower(transaction.To().Hex())
+
+	if parameters.Contract != "" && actualContractAddress != parameters.Contract {
+		return &VerifyTransactionResponse{}, nil
```
adambabik: This one as well.

protocol/transaction_validator.go
```diff
@@ -0,0 +1,317 @@
...
+	if len(message.Data()) != 0 {
+		t.logger.Debug("Validating token")
+		return t.validateTokenTransfer(parameters, message)
+	} else {
+		t.logger.Debug("Validating eth")
+		return t.validateEthereumTransfer(parameters, message)
+	}
+}
+
+func (t *TransactionValidator) ValidateTransactions(client EthClient) ([]*VerifyTransactionResponse, error) {
```
adambabik: What's the reason to pass `client` here instead of the constructor `NewTransactionValidator`?

protocol/messenger.go
```diff
@@ -52,14 +50,20 @@ type Messenger struct {
...
 	shutdownTasks              []func() error
 	systemMessagesTranslations map[protobuf.MembershipUpdateEvent_EventType]string
 	allChats                   map[string]*Chat
 	allContacts                map[string]*Contact
-	mutex                      sync.Mutex
+	allInstallations           map[string]*multidevice.Installation
+	modifiedInstallations      map[string]bool
+	installationID             string
+
+	Response *MessengerResponse
```
adambabik: Unused?

protocol/messenger.go
```diff
@@ -295,7 +292,7 @@ func NewMessenger(
 	encryptionProtocol := encryption.New(
 		database,
 		installationID,
-		c.onNewInstallationsHandler,
+		onNewInstallationsHandler,
```
adambabik, line 304: This can be a method instead.

protocol/messenger.go
```diff
@@ -1205,149 +1313,256 @@ func (m *Messenger) SendChatMessage(ctx context.Context, message *Message) (*Mes
...
+	Message protobuf.ChatMessage
+	// MessageID is the ID of the message
+	MessageID string
+	// WhisperTimestamp is the whisper timestamp of the message
+	WhisperTimestamp uint64
+	// Contact is the contact associated with the author of the message
+	Contact *Contact
+	// PublicKey is the public key of the author of the message
+	PublicKey *ecdsa.PublicKey
+}
```
adambabik, line 1557:
```suggestion
}

```

protocol/messenger.go
```diff
@@ -1205,149 +1313,256 @@ func (m *Messenger) SendChatMessage(ctx context.Context, message *Message) (*Mes
...
+type ReceivedMessageState struct {
+	// State on the message being processed
+
```
```suggestion
```

protocol/messenger.go
```diff
@@ -1205,149 +1313,256 @@ func (m *Messenger) SendChatMessage(ctx context.Context, message *Message) (*Mes
...
+	CurrentMessageState *CurrentMessageState
+	AllChats            map[string]*Chat
```
```suggestion

	AllChats            map[string]*Chat
```

protocol/messenger.go
```diff
@@ -1750,3 +1919,693 @@ func (m *Messenger) VerifyENSNames(rpcEndpoint, contractAddress string, ensDetai
...
+	if err != nil {
+		logger.Error("Error validating", zap.Error(err))
+		return nil, err
+	}
+	for _, validationResult := range responses {
+		var message *Message
+		var chat *Chat
+		var ok bool
+		chatID := types.EncodeHex(crypto.FromECDSAPub(validationResult.Transaction.From))
+		chat, ok = m.allChats[chatID]
```
```suggestion
		chat, ok := m.allChats[chatID]
```
I don't see a reason to forward declare `chat` and `ok`.

protocol/messenger.go
```diff
@@ -1750,3 +1919,693 @@ func (m *Messenger) VerifyENSNames(rpcEndpoint, contractAddress string, ensDetai
...
+		message.ContentType = protobuf.ChatMessage_TRANSACTION_COMMAND
+		message.LocalChatID = chatID
+		message.OutgoingStatus = ""
+
+		clock, timestamp := chat.NextClockAndTimestamp()
+		message.Clock = clock
+		message.Timestamp = timestamp
+		message.Text = "Transaction received"
+
+		message.ID = validationResult.Transaction.MessageID
```
adambabik, line 2596: Maybe there can be something like `extendMessageFromTxValidation()` similar to `extendMessageFromChat`?

protocol/transaction_validator.go
cammellos: Yes, it returns `Valid: false` which means the transaction is not valid and it should not be validated anymore

protocol/transaction_validator.go
cammellos: mainly because messenger takes care of initializing it (they share persistence), and does not have access to `EthClient` (which I understand we want to use interfaces in `protocol` and let `status-go` give a concrete implementation), do you think is ok?

protocol/transaction_validator.go
adambabik: I see. So maybe at the top we can define something like this:
```go
invalidResp = &VerifyTransactionResponse{Valid: false}
// ...
return invalidResp, nil
```
This should make it explicit what we are doing here.

Also, we don't want to provide any details why it's invalid?

protocol/transaction_validator.go
adambabik: So, it's like this:
```
// in shhext service
client := ...
messenger.ValidateTransactions(client)

// in Messenger
func (m *Messenger) ValidateTransactions(client) {
    validator := NewTransactionValidator()
    responses, err := validator.ValidateTransactions(client)
}
```
So, it looks like it's totally fine to pass `client` to `NewTransactionValidator`.

But that's just a technicality which does not change anything except for making `TransactionValidator` methods a bit simpler.

> which I understand we want to use interfaces in protocol and let status-go give a concrete implementation

That's correct.

--- 

Alternative thing is that `NewMessanger` would take an option `WithEthClient` and the client would be stored in `Messenger`'s field. That would probably make the `Messenger` interface the most clear but not sure we can have `EthClient` instantiated where the `Messenger` is. 

Also, `EthClient` on `Messenger` could be optional and validating txs would always fail if the client is not provided. Would that be helpful if someone does not take care about txs? I am thinking here of bots or bridges which only care about public messages.

protocol/transaction_validator.go
cammellos: cool, will do that, thanks.
For now we won't provide any details (not used by the client), but we might do eventually

protocol/messenger.go
cammellos: correct me if I am wrong, but this is relaying on the fact that `messenger` is a pointer (it's basically a closure), as it has not been initialized yet, so having it a method would work, but then we would need to pass a instantiated instance of messenger, which relies on encryption object , which in turns needs to be passed to messenger to instantiate :) 

so if I am not wrong we would have to first initialize messenger with some fields, then initialize encryptor passing the messenger function, and then set encryptor on messenger, which is a bit complicated, what do you think?

protocol/transaction_validator.go
cammellos: going for the latter option

protocol/messenger.go
cammellos: I'd wait a bit to make sure implementation settles before refactoring, as this might still have issues/request for changes etc, if that's ok

cammellos:
@adambabik ready for another round of review, thanks!

protocol/messenger.go
adambabik: Sorry, I did not notice that `onNewInstallationsHandler` is passed to the `encryption.New`. I though it's the `NewMessenger` constructor. You are right, this is good and less complicated. 

The only alternative I can think of is extracting the installations to a separate manager and share it with `encryptionProtocol` and `Messenger` but it does not sound like a task for now :)

adambabik (approved):

PR #1732, opened 2019-12-13
Drop peers when rate limit is exceeded

adambabik:


oskarth:
@adambabik do we have any data on how rate limiting limits behave in practice?

adambabik:
@oskarth we have some numbers from the `eth.test`. Not sure about what you ask specifically. Generally, I imagine that at some point we will gradually introduce rate limiting for nodes and mobile clients to find the optimal configuration. This PR adds a capability, the configuration is dynamic per node.

PR #1733, opened 2019-12-13
add missing shhclient to whisper

jakubgs:
Needed for https://github.com/status-im/statusd-bots/tree/use-go-modules.

### DIFF
--- a/whisper/shhclient/client.go
+++ b/whisper/shhclient/client.go
@@ -22,7 +22,7 @@ import (
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/rpc"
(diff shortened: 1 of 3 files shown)


adambabik (approved):

whisper/shhclient/client.go
```diff
@@ -0,0 +1,193 @@
...
+
+package shhclient
+
+import (
+	"context"
+
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/rpc"
+	whisperv6 "github.com/status-im/status-go/whisper"
```
adambabik: Oh actually, we can replace this:
```suggestion
	"github.com/status-im/status-go/whisper"
```

and replace `whisperv6` with `whisper`.

PR #1738, opened 2019-12-16
add accounts_deleteAccount method

yenda:
add deleteAccount method for accounts as well as created and updated timestamps so they can be returned ordered to status-react


rasom:
@yenda so we don't remove account's database, keys, only entry from multiaccs database? is that enough?

yenda:
@rasom it's to delete an account not a multiaccount

appdatabase/migrations/sql/0001_app.up.sql
```diff
@@ -12,7 +12,9 @@ storage TEXT,
 pubkey BLOB,
 path TEXT,
 name TEXT,
-color TEXT
+color TEXT,
+created_at DATETIME,
```
adambabik:
```suggestion
created_at DATETIME NOT NULL,
```

appdatabase/migrations/sql/0001_app.up.sql
```diff
@@ -12,7 +12,9 @@ storage TEXT,
...
+updated_at DATETIME
```
adambabik:
```suggestion
updated_at DATETIME NOT NULL
```

appdatabase/migrations/sql/0001_app.up.sql
adambabik: This column can benefit from having an index.

services/accounts/accounts.go
```diff
@@ -29,3 +30,12 @@ func (api *API) SaveAccounts(ctx context.Context, accounts []accounts.Account) e
 func (api *API) GetAccounts(ctx context.Context) ([]accounts.Account, error) {
 	return api.db.GetAccounts()
 }
+
+func (api *API) DeleteAccount(ctx context.Context, address common.Address) error {
+	err := api.db.DeleteAccount(address)
+	if err != nil {
+		return err
+	}
+	api.feed.Send(address)
```
adambabik: Do we need this? `api.feed` takes various arguments, not sure what it is useful for.

appdatabase/migrations/sql/0001_app.up.sql
andremedeiros: I can't imagine the volume would make much of a difference, but I'm on it.

andremedeiros (approved):

adambabik (approved):

PR #1744, opened 2019-12-18
[WIP] status-im/status-react#9203 BalanceAt caching

rasom:
status-im/status-react#9203

what's already fixed here:
- we do not request history twice for the same address as we did
- we do not request balance for a specific block more then once, depending on circumstances we could do this up 23-24 times for a specific block, but in average it reduces the number of `eth_getBalance` requests in 3-4 times

status: ready

### DIFF
--- a/api/geth_backend.go
+++ b/api/geth_backend.go
@@ -907,11 +907,20 @@ func (b *GethStatusBackend) startWallet() error {
 	for i, addr := range watchAddresses {
 		allAddresses[1+i] = common.Address(addr)
 	}
+
+	uniqAddressesMap := map[common.Address]struct{}{}
+	uniqAddresses := []common.Address{}
+	for _, address := range allAddresses {
+		if _, ok := uniqAddressesMap[address]; !ok {
+			uniqAddressesMap[address] = struct{}{}
+			uniqAddresses = append(uniqAddresses, address)
+		}
+	}
+
 	return wallet.StartReactor(
 		b.statusNode.RPCClient().Ethclient(),
-		allAddresses,
-		new(big.Int).SetUint64(b.statusNode.Config().NetworkID),
-	)
+		uniqAddresses,
+		new(big.Int).SetUint64(b.statusNode.Config().NetworkID))
 }
 
 // InjectChatAccount selects the current chat account using chatKeyHex and injects the key into whisper.
--- a/services/wallet/balance_cache.go
+++ b/services/wallet/balance_cache.go
@@ -1,0 +1,53 @@
+package wallet
+
+import (
+	"context"
+	"math/big"
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+type balanceCache struct {
+	// cache maps an address to a map of a block number and the balance of this particular address
+	cache map[common.Address]map[*big.Int]*big.Int
+	lock  sync.RWMutex
+}
+
+func (b *balanceCache) readCachedBalance(account common.Address, blockNumber *big.Int) *big.Int {
+	b.lock.RLock()
+	defer b.lock.RUnlock()
+
+	return b.cache[account][blockNumber]
+}
+
+func (b *balanceCache) addBalanceToCache(account common.Address, blockNumber *big.Int, balance *big.Int) {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
+	_, exists := b.cache[account]
+	if !exists {
+		b.cache[account] = make(map[*big.Int]*big.Int)
+	}
+	b.cache[account][blockNumber] = balance
+}
+
+func (b *balanceCache) BalanceAt(client BalanceReader, ctx context.Context, account common.Address, blockNumber *big.Int) (*big.Int, error) {
+	cachedBalance := b.readCachedBalance(account, blockNumber)
+	if cachedBalance != nil {
+		return cachedBalance, nil
+	}
+	balance, err := client.BalanceAt(ctx, account, blockNumber)
+	if err != nil {
+		return nil, err
+	}
+	b.addBalanceToCache(account, blockNumber, balance)
+
+	return balance, nil
+}
+
+func newBalanceCache() *balanceCache {
+	return &balanceCache{
+		cache: make(map[common.Address]map[*big.Int]*big.Int),
+	}
+}
--- a/services/wallet/commands.go
+++ b/services/wallet/commands.go
@@ -14,11 +14,12 @@ import (
 )
 
 type ethHistoricalCommand struct {
-	db      *Database
-	eth     TransferDownloader
-	address common.Address
-	client  reactorClient
-	feed    *event.Feed
+	db           *Database
+	eth          TransferDownloader
+	address      common.Address
+	client       reactorClient
+	balanceCache *balanceCache
+	feed         *event.Feed
 
 	from, to *big.Int
 }
@@ -47,7 +48,7 @@ func (c *ethHistoricalCommand) Run(ctx context.Context) (err error) {
 	defer cancel()
 	concurrent := NewConcurrentDownloader(ctx)
 	start := time.Now()
-	downloadEthConcurrently(concurrent, c.client, c.eth, c.address, c.from, c.to)
+	downloadEthConcurrently(concurrent, c.client, c.balanceCache, c.eth, c.address, c.from, c.to)
 	select {
 	case <-concurrent.WaitAsync():
 	case <-ctx.Done():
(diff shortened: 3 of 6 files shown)


api/geth_backend.go
```diff
@@ -888,9 +888,19 @@ func (b *GethStatusBackend) startWallet() error {
 	allAddresses := make([]common.Address, len(watchAddresses)+1)
 	allAddresses[0] = mainAccountAddress
 	copy(allAddresses[1:], watchAddresses)
+
+	uniqAddressesMap := map[common.Address]bool{}
```
adambabik:
```suggestion
	uniqAddressesMap := map[common.Address]struct{}{}
```

api/geth_backend.go
```diff
@@ -888,9 +888,19 @@ func (b *GethStatusBackend) startWallet() error {
...
+	uniqAddresses := []common.Address{}
+	for _, address := range allAddresses {
+		if _, value := uniqAddressesMap[address]; !value {
```
adambabik:
```suggestion
		if _, ok := uniqAddressesMap[address]; !ok {
```

api/geth_backend.go
```diff
@@ -888,9 +888,19 @@ func (b *GethStatusBackend) startWallet() error {
...
+			uniqAddressesMap[address] = true
```
adambabik:
```suggestion
			uniqAddressesMap[address] = struct{}{}
```

services/wallet/concurrent.go
```diff
@@ -45,17 +45,56 @@ type TransferDownloader interface {
 	GetTransfersByNumber(context.Context, *big.Int) ([]Transfer, error)
 }
 
+var balancesCache = map[common.Address]map[*big.Int]*big.Int{}
```
adambabik:
```suggestion
var (
    balancesCache = make(map[common.Address]map[*big.Int]*big.Int)
    balancesCacheLock sync.RWMutex
)
```

Is it ok if this cache is not persistent?

Another approach would be that this cache belongs to something that calls `downloadEthConcurrently`. It can then assign the reference to the cache to `ConcurrentDownloader`.

services/wallet/concurrent.go
```diff
@@ -45,17 +45,56 @@ type TransferDownloader interface {
...
+var balanaceMapLock = sync.RWMutex{}
+
+func readCachedBalance(account common.Address, blockNumber *big.Int) (*big.Int, bool) {
+	balanaceMapLock.RLock()
+	defer balanaceMapLock.RUnlock()
+
+	balance, exists := balancesCache[account][blockNumber]
```
adambabik:
```suggestion
	return balancesCache[account][blockNumber]
```

services/wallet/concurrent.go
```diff
@@ -45,17 +45,56 @@ type TransferDownloader interface {
...
+	return balance, exists
```
adambabik:
```suggestion
```

services/wallet/concurrent.go
```diff
@@ -45,17 +45,56 @@ type TransferDownloader interface {
...
+}
+
+func writeCachedBalance(account common.Address, blockNumber *big.Int, balance *big.Int) {
+	balanaceMapLock.Lock()
+	defer balanaceMapLock.Unlock()
+
+	_, exists := balancesCache[account]
+	if !exists {
+		balancesCache[account] = map[*big.Int]*big.Int{}
```
adambabik:
```suggestion
		balancesCache[account] = make(map[*big.Int]*big.Int)
```

services/wallet/concurrent.go
```diff
@@ -45,17 +45,56 @@ type TransferDownloader interface {
...
+
+func readCachedBalance(account common.Address, blockNumber *big.Int) (*big.Int, bool) {
+	balanaceMapLock.RLock()
+	defer balanaceMapLock.RUnlock()
+
+	balance, exists := balancesCache[account][blockNumber]
+	return balance, exists
+}
+
+func writeCachedBalance(account common.Address, blockNumber *big.Int, balance *big.Int) {
```
adambabik:
```suggestion
func addBalanceToCache(account common.Address, blockNumber *big.Int, balance *big.Int) {
```
because write something a bit different.

adambabik (commented):
Generally, I would try to put the cache into a struct and assign an owner of it.

adambabik:
Sorry, I didn't notice it's WIP...

services/wallet/concurrent.go
rasom: > Is it ok if this cache is not persistent?

yes it is ok, this is to prevent requesting the balance from the same block during initial scan, not like we want to persist the whole history of these calls

services/wallet/concurrent.go
rasom: if I do so how do I check whether map contains the value? because I also have to change signature of the function and it wouldn't return that `ok` value or something

rasom:
> Generally, I would try to put the cache into a struct and assign an owner of it.

@adambabik not like I really understand what you mean :D I could add the last thing I wanted to have in this PR (checking of `eta_getTransactionCount` in case if balance was zero in both `from` and `to` blocks of given blocks range we check for transactions) and then you could rewrite it the way it pleases golang dev eye, because for me it's kind of impossible task at the moment

rasom:
@adambabik 
Hey Adam, could you take a look again? I hope it looks a bit better now, let me know what has to be changed.

services/wallet/balance_cache.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"context"
+	"math/big"
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+type BalanceCache struct {
+	cache map[common.Address]map[*big.Int]*big.Int
```
adambabik, line 13: Let's describe it:
```suggestion
	// cache maps an address to a map of a block number and the balance of this particular address
	cache map[common.Address]map[*big.Int]*big.Int
```

services/wallet/balance_cache.go
```diff
@@ -0,0 +1,55 @@
...
+	lock  sync.RWMutex
+}
+
+func (b *BalanceCache) readCachedBalance(account common.Address, blockNumber *big.Int) (*big.Int, bool) {
+	b.lock.RLock()
+	defer b.lock.RUnlock()
+
+	balance, exists := b.cache[account][blockNumber]
```
adambabik:
```suggestion
	return b.cache[account][blockNumber]
```

services/wallet/balance_cache.go
```diff
@@ -0,0 +1,55 @@
...
+		b.cache[account] = make(map[*big.Int]*big.Int)
+	}
+	b.cache[account][blockNumber] = balance
+}
+
+func (b *BalanceCache) BalanceAt(client BalanceReader, ctx context.Context, account common.Address, blockNumber *big.Int) (*big.Int, error) {
+	cachedBalance, exists := b.readCachedBalance(account, blockNumber)
+	if exists {
+		return cachedBalance, nil
+	} else {
```
adambabik: You can skip `else` because you return in `if`.

services/wallet/balance_cache.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"context"
+	"math/big"
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+type BalanceCache struct {
```
adambabik: Looks like we don't need to export it:
```suggestion
type balanceCache struct {
```

services/wallet/balance_cache.go
```diff
@@ -0,0 +1,55 @@
...
+		if err != nil {
+			return nil, err
+		}
+		b.addBalanceToCache(account, blockNumber, balance)
+
+		return balance, nil
+	}
+}
+
+func NewBalanceCache() *BalanceCache {
```
adambabik: It would be nice to put it right below `type BalanceCache struct { ... }` declaration.

Also, we don't need to export it:
```suggestion
func newBalanceCache() *balanceCache {
```

services/wallet/balance_cache.go
```diff
@@ -0,0 +1,55 @@
...
+	return &BalanceCache{
+		cache: make(map[common.Address]map[*big.Int]*big.Int),
+		lock:  sync.RWMutex{},
```
adambabik: Not needed. It's zero value is the same.

services/wallet/commands.go
```diff
@@ -321,9 +323,10 @@ func (c *controlCommand) fastIndex(ctx context.Context, to *DBHeader) error {
 		}
 		group.Add(erc20.Command())
 		eth := &ethHistoricalCommand{
-			db:      c.db,
-			client:  c.client,
-			address: address,
+			db:           c.db,
+			client:       c.client,
+			balanceCache: NewBalanceCache(),
```
adambabik:
```suggestion
			balanceCache: newBalanceCache(),
```

adambabik (commented):
Looks great! 👏 A few details that can be improved.

rasom:
@adambabik should be fixed now

PR #1745, opened 2019-12-18
Pre-nimbus changes

pedropombeiro:
This PR implements some changes required by the Nimbus branch, but which are not dependent on that branch and do not break develop. By merging early to develop, we'll avoid merge conflicts later on.

### DIFF
--- a/eth-node/bridge/nimbus/node.go
+++ b/eth-node/bridge/nimbus/node.go
@@ -75,14 +75,10 @@ func NewNodeBridge() Node {
 }
 
 func (n *nimbusNodeWrapper) StartNimbus(privateKey *ecdsa.PrivateKey, listenAddr string, staging bool) error {
-	retVal := n.routineQueue.Send(func(c chan<- interface{}) {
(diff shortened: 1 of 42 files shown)


eth-node/bridge/nimbus/node.go
```diff
@@ -18,46 +19,136 @@ import (
...
+
+func NewNodeBridge() Node {
+	c := make(chan Node, 1)
+	go func(c chan<- Node, delay time.Duration) {
+		runtime.LockOSThread()
+		defer runtime.UnlockOSThread()
+
+		n := &nimbusNodeWrapper{
+			routineQueue: 			NewRoutineQueue(),
+			tid: 								syscall.Gettid(),
```
cammellos: I see it weirdly indented, but I think it might be github interface

cammellos (approved):

eth-node/bridge/nimbus/node.go
pedropombeiro: There were some tabs there for some reason, fixed it.

PR #1746, opened 2019-12-18
Put settings in a table with schema

yenda:
Move settings table schema from a key-value store to a one row table with many columns.

We now save the first raw with initial data in saveAccountAndLogin and follow up saveSetting calls are only saving one setting at a time.


multiaccounts/accounts/database.go
```diff
@@ -50,44 +124,107 @@ func (db Database) Close() error {
...
+  address,
+  dapps_address,
+  eip1581_address,
+  installation_id,
+  key_uid,
+  keycard_instance_uid,
+  keycard_paired_on,
+  keycard_pairing,
+  latest_derived_path,
+  last_updated,
```
andremedeiros: I'd leave this field for last

multiaccounts/accounts/database.go
```diff
@@ -36,6 +73,43 @@ type Account struct {
...
+	RememberSyncingChoice  *bool `json:remember-syncing-choice,omitempty`
+	ShowName               *bool `json:show-name,omitempty`
+	SigningPhrase          string `json:signing-phrase`
+	StickerPacksInstalled  *json.RawMessage `json:stickers-packs-installed,omitempty`
+	StickersRecentStickers *json.RawMessage `json:stickers-recent-stickers,omitempty`
+	SyncingOnMobileNetwork *bool `json:syncing-on-mobile-network,omitempty`
+	Usernames              *[]string `json:usernames,omitempty`
+	WalletRootAddress      common.Address `json:wallet-root-address,omitempty`
+	WalletSetupPassed      *bool `json:wallet-setup-passed,omitempty`
+	WalletVisibleTokens    *[]string `json:wallet-visible-tokens,omitempty`
```
andremedeiros: Are we sure this is the correct syntax for naming fields?

yenda: yes I want to avoid having to rename all fields on status-react side

api/backend_test.go
```diff
@@ -28,6 +28,22 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+var (
+	settings  = accounts.Settings{
```
corpetty: where did these creds come from?  just curious

yenda: from an account I created with status a few days ago. they are also in the rfc-002

appdatabase/migrations/bindata.go
```diff
@@ -88,7 +90,7 @@ func _0001_appDownSql() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "0001_app.down.sql", size: 387, mode: os.FileMode(0644), modTime: time.Unix(1574780240, 0)}
+	info := bindataFileInfo{name: "0001_app.down.sql", size: 387, mode: os.FileMode(0644), modTime: time.Unix(1575559353, 0)}
```
corpetty: https://github.com/status-im/trailofbits-audit/issues/10

corpetty: file permissions too broad

appdatabase/migrations/bindata.go
```diff
@@ -108,7 +110,7 @@ func _0001_appUpSql() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "0001_app.up.sql", size: 3088, mode: os.FileMode(0644), modTime: time.Unix(1576507278, 0)}
+	info := bindataFileInfo{name: "0001_app.up.sql", size: 3088, mode: os.FileMode(0664), modTime: time.Unix(1576664357, 0)}
```
corpetty: same issue with file permissions

appdatabase/migrations/bindata.go
```diff
@@ -128,7 +130,7 @@ func _0002_tokensDownSql() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "0002_tokens.down.sql", size: 19, mode: os.FileMode(0644), modTime: time.Unix(1575999124, 0)}
+	info := bindataFileInfo{name: "0002_tokens.down.sql", size: 19, mode: os.FileMode(0644), modTime: time.Unix(1576510356, 0)}
```
corpetty: file permissions

appdatabase/migrations/bindata.go
```diff
@@ -148,11 +150,51 @@ func _0002_tokensUpSql() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "0002_tokens.up.sql", size: 248, mode: os.FileMode(0644), modTime: time.Unix(1575999124, 0)}
+	info := bindataFileInfo{name: "0002_tokens.up.sql", size: 248, mode: os.FileMode(0644), modTime: time.Unix(1576510356, 0)}
```
corpetty: file permissions

appdatabase/migrations/bindata.go
```diff
@@ -148,11 +150,51 @@ func _0002_tokensUpSql() (*asset, error) {
...
+	)
+}
+
+func _0003_settingsDownSql() (*asset, error) {
+	bytes, err := _0003_settingsDownSqlBytes()
+	if err != nil {
+		return nil, err
+	}
+
+	info := bindataFileInfo{name: "0003_settings.down.sql", size: 118, mode: os.FileMode(0644), modTime: time.Unix(1576697070, 0)}
```
corpetty: ding ding

appdatabase/migrations/bindata.go
```diff
@@ -148,11 +150,51 @@ func _0002_tokensUpSql() (*asset, error) {
...
+	)
+}
+
+func _0003_settingsUpSql() (*asset, error) {
+	bytes, err := _0003_settingsUpSqlBytes()
+	if err != nil {
+		return nil, err
+	}
+
+	info := bindataFileInfo{name: "0003_settings.up.sql", size: 1340, mode: os.FileMode(0644), modTime: time.Unix(1576767632, 0)}
```
corpetty: ding ding

appdatabase/migrations/bindata.go
```diff
@@ -168,7 +210,7 @@ func docGo() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "doc.go", size: 74, mode: os.FileMode(0644), modTime: time.Unix(1574780240, 0)}
+	info := bindataFileInfo{name: "doc.go", size: 74, mode: os.FileMode(0644), modTime: time.Unix(1575559353, 0)}
```
corpetty: one more

corpetty (changes requested):
There is also the `os.MkdirAll` issue from the audit (https://github.com/status-im/trailofbits-audit/issues/41).  Maybe this should be a diff PR, but fixing the file permissions in this one is a start in the right direction.

andremedeiros (approved):

multiaccounts/accounts/database.go
```diff
@@ -49,44 +128,114 @@ func (db Database) Close() error {
...
+  preview_privacy,
+  public_key,
+  signing_phrase,
+  wallet_root_address,
+  synthetic_id
+) VALUES (
+?,?,?,?,?,?,?,?,?,?,
+?,?,?,?,?,?,?,?,?,?,
+'id')`,
+		&sqlite.JSONBlob{s.Address},
```
cammellos: Not familiar with this, could it not be just `s.Address` ?

multiaccounts/accounts/database.go
```diff
@@ -49,44 +128,114 @@ func (db Database) Close() error {
...
-		if err != nil {
-			return nil, err
-		}
-		rst[typ] = buf
-	}
-	return rst, nil
+func (db *Database) GetSettings() (Settings, error) {
+	var s Settings
+	err := db.db.QueryRow("SELECT address, chaos_mode, currency, current_network, custom_bootnodes, custom_bootnodes_enabled, dapps_address, eip1581_address, fleet, hide_home_tooltip, installation_id, key_uid, keycard_instance_uid, keycard_paired_on, keycard_pairing, last_updated, latest_derived_path, log_level, mnemonic, name, networks, notifications_enabled, photo_path, pinned_mailserver, preferred_name, preview_privacy, public_key, remember_syncing_choice, show_name, signing_phrase, stickers_packs_installed, stickers_recent_stickers, syncing_on_mobile_network, usernames, wallet_root_address, wallet_set_up_passed, wallet_visible_tokens FROM settings WHERE synthetic_id = 'id'").Scan(
+		&sqlite.JSONBlob{&s.Address},
```
cammellos: not sure you need to wrap everything in `sqlite.JSONBlob`

multiaccounts/accounts/database.go
```diff
@@ -49,44 +128,114 @@ func (db Database) Close() error {
...
-func (db *Database) GetConfig(typ string, value interface{}) error {
-	return db.db.QueryRow("SELECT value FROM settings WHERE type = ?", typ).Scan(&sqlite.JSONBlob{value})
+func (db *Database) SaveSetting(typ string, value interface{}) error {
+	setting, ok := settingsFields[typ]
+	if !ok {
+		return ErrInvalidConfig
+	}
+
+	stmt := fmt.Sprintf("UPDATE settings SET %s = ? WHERE synthetic_id = 'id'", setting)
+	_, err := db.db.Exec(stmt, &sqlite.JSONBlob{value})
```
cammellos: I think this statement forces us to use `JSONBlob` everywhere.
We should first ensure that things are stored correctly, i.e. If a save a string `hello` it should be stored as `hello`, not as `"hello"`, otherwise that might be annoying in the future.

So we should unwrap the type here of `interface{}` and have the right statement for each type, 
https://tour.golang.org/methods/16
that way we can ensure that things are stored in the right way and we can get rid of `JSONBlob{}`

cammellos (changes requested):

multiaccounts/accounts/database.go
cammellos: ah I see what we are trying to do,
I think we should just have a case on `typ` and parse the `interface{}` appropriately in the right type, so it gets stored correctly in the db, and we can get rid of the various `sqlite.JSONBlob`, might look boring but contains the complexity

yenda:
@adambabik could you have a look? Ideally we'd want this branch to be merged today

cammellos (approved):

yenda:
@corpetty is your change still requested?

corpetty (approved):

multiaccounts/accounts/database.go
```diff
@@ -49,44 +91,196 @@ func (db Database) Close() error {
...
+	)
 
-func (db *Database) GetConfigBlobs(types []string) (map[string]json.RawMessage, error) {
-	// it expands number of bind vars to the number of types. sql interface doesn't allow to bypass it without modifying driver.
-	// expansion can be hidden in a function that modifies string but better to make it explicitly.
-	query := fmt.Sprintf("SELECT type, value FROM settings WHERE type IN (?%s)", strings.Repeat(",?", len(types)-1)) // nolint: gosec
-	args := make([]interface{}, len(types))
-	for i := range types {
-		args[i] = types[i]
+	switch setting {
```
adambabik, line 153: You can replace all this with some rules I believe:
```go
func normalizeSettingName(name string) string {
    name = strings.ReplaceAll(name, "?", "")
    name = strings.ReplaceAll(name, "-", "_")
    // other corner cases
    return name
}

columnName := normalizeSettingName(setting)
update, err := db.db.Prepare(
    fmt.Sprintf("UPDATE settings SET %s = ? WHERE synthetic_id = 'id'", columnName)
)
if err != nil {
    return err
}
_, err = update.Exec(toSQLValue(name, value))
```

multiaccounts/accounts/database.go
```diff
@@ -35,6 +36,47 @@ type Account struct {
 	Color     string         `json:"color"`
 }
 
+type Settings struct {
+	// required
+	Address                types.Address    `json:"address"`
+	ChaosMode              bool             `json:"chaos-mode?,omitempty"`
+	Currency               string           `json:"currency,omitempty"`
+	CurrentNetwork         string           `json:"networks/current-network"`
+	CustomBootnodes        *json.RawMessage `json:"custom-boot-nodes,omitempty"`
```
adambabik, line 45: `json.RawMessage` is alias of `[]byte` so it's already a pointer because slices are pointers. `json.RawMessage` should be enough.

multiaccounts/accounts/database.go
```diff
@@ -35,6 +36,47 @@ type Account struct {
...
+	CustomBootnodesEnabled *json.RawMessage `json:"custom-boot-nodes-enabled,omitempty"`
+	DappsAddress           types.Address    `json:"dapps-address"`
+	EIP1581Address         types.Address    `json:"eip1581-address"`
+	Fleet                  *string          `json:"fleet,omitempty"`
```
adambabik, line 49: Any reason it is a pointer while other string fields are not?

multiaccounts/accounts/database.go
yenda: I made this switch to avoid the SQL injection warning. It's fine like this

multiaccounts/accounts/database.go
yenda: Because it is nullable others are not. The whole things works well in status-react. I only have an issue with the e2e test

multiaccounts/accounts/database.go
adambabik: If done with `fmt.Sprintf` before passing to `Prepare()`, there should be no SQL injection warnings. We can keep it this way, it's just a lot of manual work that computer could do.

multiaccounts/accounts/database.go
yenda: it's because it's not just json.RawMessage it can also be nil, when calling getSettings if it's not a pointer you get:
```
        	            	sql: Scan error on column index 4, name "custom_bootnodes": unsupported Scan, storing driver.Value type <nil> into type *json.RawMessage
```

PR #1749, opened 2019-12-19
fix sticker pack with id = 0

flexsurfer:
fix sticker pack with id = 0


cammellos (commented):
can we disallow pack with 0? or is too much  of a change? ( there s no way to distinguish not set from 0 in golang).

Otherwise is fine

flexsurfer:
this ID in the contract, so we can't change it

cammellos (approved):

flexsurfer:
could someone merge please, thanks :)

adambabik (approved):

PR #1752, opened 2019-12-20
Abstract the keystore

pedropombeiro:
This PR abstracts the Keystore so that it can rely on the geth implementation or the Nimbus implementation (upcoming). It introduces two new types: `types.Key` and `types.Account` that we should use instead of the Geth versions. It also pulls in the `DecryptKey` function from geth.


account/accounts_geth.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"github.com/ethereum/go-ethereum/accounts"
+	"github.com/status-im/status-go/account/generator"
+)
+
+// GethManager represents account manager interface.
+type GethManager struct {
+	Manager
+
+	manager *accounts.Manager
```
adambabik:
```suggestion
	gethAccManager *accounts.Manager
```

account/accounts_geth.go
```diff
@@ -0,0 +1,42 @@
...
+package account
+
+import (
+	"github.com/ethereum/go-ethereum/accounts"
+	"github.com/status-im/status-go/account/generator"
+)
+
+// GethManager represents account manager interface.
+type GethManager struct {
+	Manager
```
adambabik:
```suggestion
	*Manager
```
would make more sense I think. If `GethManager` was used as a value, it could produce some strange results (`Manager` will be copied).

account/accounts.go
```diff
@@ -37,8 +35,7 @@ var zeroAddress = types.Address{}
 // Manager represents account manager interface.
 type Manager struct {
```
adambabik, line 36: Do we need to keep it exported?

account/accounts_geth.go
```diff
@@ -0,0 +1,42 @@
...
+
+	manager *accounts.Manager
+}
+
+// NewManager returns new node account manager.
+func NewManager() *GethManager {
+	m := &GethManager{}
+	m.accountsGenerator = generator.New(m)
```
adambabik: Then here you would need to do:
```suggestion
	m.Manager = &Manager{accountsGenerator: generator.New(m)}
```

account/accounts_geth.go
pedropombeiro: Good point!

account/accounts.go
pedropombeiro: For the time being yes, since it's being used in other packages like `multiaccounts` or `accounts`. If the opportunity arises we should have an interface and make the implementation private.

pedropombeiro:
Thanks for the review Adam! I've addressed the feedback.

Makefile
```diff
@@ -176,6 +176,10 @@ prepare-release: clean-release
 clean-release:
 	rm -rf $(RELEASE_DIR)
 
+gofmt:
```
adambabik, line 179: Up to you, but there are many unrelated files changed because of this.

adambabik (approved):

PR #1759, opened 2019-12-28
Integrate Waku service

adambabik:
Introduce Waku service into the Status Node. It can be enabled by providing `NodeConfig.WakuConfig.Enabled = true`.

Other changes:
- `EnableNTPSync` moved to `NodeConfig` and `true` by default (`EnableNTPSync` is gone from `WhisperConfig`),
- `gethEnvelopeWrapper` supports Whisper and Waku envelopes,
- `Envelope` interface has more getters,
- `MailServer` interface in Whisper and Waku changed slightly,
- There are separate implementations of `MailServer` for Waku and Whisper now,
- mailserver request's `Limit` is enforced to maximum `1000`.


eth-node/bridge/geth/envelope.go
```diff
@@ -2,29 +2,112 @@ package gethbridge
...
 	"github.com/status-im/status-go/eth-node/types"
+	"github.com/status-im/status-go/waku"
 	"github.com/status-im/status-go/whisper/v6"
 )
 
 type gethEnvelopeWrapper struct {
-	envelope *whisper.Envelope
+	shhEnvelope  *whisper.Envelope
+	wakuEnvelope *waku.Envelope
 }
```
jakubgs, line 11: Wouldn't it make more sense to create an `type Envelope interface` defining what methods `waku.Envelop` and `whisper.Envelope` should implement to be of that type?

eth-node/bridge/geth/envelope.go
```diff
@@ -2,29 +2,112 @@ package gethbridge
...
 
-// NewGethEnvelopeWrapper returns an object that wraps Geth's Envelope in a types interface
-func NewGethEnvelopeWrapper(e *whisper.Envelope) types.Envelope {
+// NewWhisperEnvelopeWrapper returns an object that wraps Geth's Whisper Envelope in a types interface.
+func NewWhisperEnvelopeWrapper(e *whisper.Envelope) types.Envelope {
```
jakubgs: This is only used once in `services/shhext/api.go`. Why not use `whisper.Enveloper` there directly?

mailserver/mailserver.go
```diff
@@ -17,28 +17,30 @@
...
+	"github.com/status-im/status-go/waku"
 	"github.com/status-im/status-go/whisper/v6"
-
-	prom "github.com/prometheus/client_golang/prometheus"
 )
 
 const (
 	maxQueryRange = 24 * time.Hour
-	defaultLimit  = 2000
+	defaultLimit  = 1000
```
jakubgs: Maybe a comment explaining purpose of this limit? Also, the word "default" doesn't really describe well what this does, maybe renaming it would be a good idea.

jakubgs (approved):
Looks good from what I've seen, but I'll do another pass on the weekend or Monday.

eth-node/bridge/geth/envelope.go
adambabik: Because `whisper.Envelope` does not implement `Envelope` interface. An adapter is needed, that's why we have `gethEnvelopeWrapper`. What I should do is to split Whisper and Waku wrappers.

eth-node/bridge/geth/envelope.go
adambabik: There is such a thing: https://github.com/status-im/status-go/blob/develop/eth-node/types/envelopes.go#L5-L8

node/node.go
```diff
@@ -350,6 +365,51 @@ func activateShhService(stack *node.Node, config *params.NodeConfig, db *leveldb
...
+
+	err = stack.Register(func(ctx *node.ServiceContext) (node.Service, error) {
+		return createWakuService(ctx, &config.WakuConfig, &config.ClusterConfig)
+	})
+	if err != nil {
+		return
+	}
+
+	// Register eth-node node bridge
+	// TODO: what to do with it? Shouldn't it be configured when a node is created?
```
adambabik: cc @PombeirP

pedropombeiro: Yeah, this should move out of `activateShhService`/`activateWakuService` because it's not really related to Whisper/Waku.

eth-node/bridge/geth/envelope.go
jakubgs: Yes, that would make sense.

eth-node/bridge/geth/envelope.go
jakubgs: I see.

jakubgs (approved):
LGTM, but I still don't get what's the purpose of `eth-node/bridge/geth/envelope.go`. Maybe you can explain it to me sometime. It seems pointless.

adambabik:
@jakubgs it's similar to all other type generalizations. The goal is that `types.Envelope` is an interface used by the whole `protocol/` and all `services/` instead of concrete type. The concrete type can't be used because it can be either go-ethereum or Nimbus and they are two different types.

oskarth:
Assuming this PR is targeting the Waku spec, it'd be useful to outline which version of the spec it implements, along with an gotchas (didn't implement X). At an interface level, not implementation "service" level. That would make reviewing for correctness easier, and make sure we don't miss things such as:
- mailserver topic interest
- bridging
- handshake stuff?
- other unknowns/small changes in spec

https://specs.vac.dev/waku/waku.html
https://specs.vac.dev/waku/wms.html
https://specs.vac.dev/waku/wedf.html

also cc @decanus as spec writing and gopher

PR #1775, opened 2020-01-08
faster tx fetching with less request

rasom:
WIP

### Transactions or Transfers
 Technically we have a history of transfers, not transactions history. 
In case if transaction contains more than one transfer related to current
address we show these transfers as separate entries in the list. There 
is a small issue which related to this, if there is more than one 
transfer, only one of them will have a correct etherscan link.

### How it worked before this PR on multiaccount creation:

* On multiacc creation we scanned chain for eth and erc20 transfers. For 
 each address of a new empty multiaccount this scan required 
  1. two `eth_getBalance` requests to find out that there is no any 
 balance change between zero and the last block, for eth transfers 
  1. and `chain-size/100000` (currently ~100) `eth_getLogs` requests, 
 for erc20 transfers
* For some reason we scanned an address of the chat account as well, and 
 also accounts were not deduplicated. So even for an empty multiacc we 
 scanned chain twice for each chat and main wallet addresses, in result
 app had to execute about 400 requests. 
* As mentioned above, `eth_getBalance` requests were used to check if 
 there were any eth transfers, and that caused empty history in case
 if user already used all available eth (so that both zero and latest 
 blocks show 0 eth for an address). There might have been transactions 
 but we wouldn't fetch/show them.
* There was no upper limit for the number of rpc requests during the 
 scan, so it could require indefinite number of requests; the scanning 
 algorithm was written so that we persisted the whole history of 
 transactions or tried to scan form the beginning again in case of 
 failure, giving up only after 10 minutes of failures. In result 
 addresses with sufficient number of transactions would never be fully 
 scanned and during these 10 minutes app could use gigabytes of 
 internet data. 
* Failures were caused by `eth_getBlockByNumber`/`eth_getBlockByHash`
 requests. These requests return significantly bigger responses than
`eth_getBalance`/`eth_transactionsCount` and it is likely that 
 execution of thousands of them in parallel caused failures for 
 accounts with hundreds of transactions. Even for an account with 12k
 we could successfully determine blocks with transaction in a few 
 minutes using `eth_getBalance` requests, but `eth_getBlock...`
 couldn't be processed for this acc.
* There was no caching for for `eth_getBalance` requests, and this 
 caused in average 3-4 times more such requests than is needed.

### How it works now on multiaccount creation:

* On multiacc creation we scan chain for last ~30 eth transactions and
 then check erc20 in the range where these eth transactions were found. 
 For an empty address in multiacc this means:
  1. two `eth_getBalance` transactions to determine that there was no 
 balance change between zero and the last block; two 
`eth_transactionsCount` requests to determine there are no outgoing
 transactions for this address; total 4 requests for eth transfers
  1. 20 `eth_getLogs` for erc20 transfers. This number can be lowered, 
 but that's not a big deal
* Deduplication of addresses is added and also we don't scan chat 
 account, so a new multiacc requires ~25 (we also request latest block 
 number and probably execute a few other calls) request to determine 
 that multiacc is empty (comparing to ~400 before)
* In case if address contains transactions we:
  1. determine the range which contains 20-25 outgoing eth/erc20 
 transactions. This usually requires up to 10 `eth_transactionCount`
 requests
  1. then we scan chain for eth transfers using `eth_getBalance` and
`eth_transactionCount` (for double checking zero balances)
  1. we make sure that we do not scan db for more than 30 blocks with 
 transfers. That's important for accounts with mostly incoming
 transactions, because the range found on the first step might 
 contain any number of incoming transfers, but only 20-25 outgoing
 transactions
  1. when we found ~30 blocks in a given range, we update initial 
 range `from` block using the oldest found block
  1. and now we scan db for erc20transfers using `eth_getLogs`
`oldest-found-eth-block`-`latest-block`, we make not more than 20 calls
  1. when all blocks which contain incoming/outgoing transfers for a 
 given address are found, we save these blocks to db and mark that 
 transfers from these blocks are still to be fetched
  1. Then we select latest ~30 (the number can be adjusted) blocks from 
 these which were found and fetch transfers, this requires 3-4 
 requests per transfer. 
  1. we persist scanned range so that we know were to start next time
  1. we dispatch an event which tells client that transactions are found
  1. client fetches latest 20 transfers
* when user presses "fetch more" button we check if app's db contains next
 20 transfers, if not we scan chain again and return transfers after

### known problems (or potential problems):

* we do not merge scanned ranges, that means if 20-40 and 0-20 ranges
 were scanned we store both in app's db
* and then when we restart app we check recent history, and in case if
 there are more then 30 eth transfers (previously unknown to app) it 
 might happen that we will not the whole range from the last known 
 block till latest, which will create a gap in ranges which wont be 
 filled later. This is a problem but very unlikely to be reproduced 
 if user uses Status to send transactions
* even now on initial scan the multiacc which contains more than 20 
 transfers will require ~5MB of internet data for scanning, which still
 might be expensive for some users. Thus we need to ask the user 
 whether they want to scan chain on mobile network, as we do with chat 
 history
* there is no 1000 request per minute limit
* rpc documentations is not updated


api/geth_backend.go
```diff
@@ -907,11 +907,20 @@ func (b *GethStatusBackend) startWallet() error {
 	for i, addr := range watchAddresses {
 		allAddresses[1+i] = common.Address(addr)
 	}
+
+	uniqAddressesMap := map[common.Address]struct{}{}
+	uniqAddresses := []common.Address{}
+	for _, address := range allAddresses {
```
andremedeiros: This work can be done on the above loop.

appdatabase/database.go
```diff
@@ -9,7 +9,7 @@ import (
 
 // InitializeDB creates db file at a given path and applies migrations.
 func InitializeDB(path, password string) (*sql.DB, error) {
-	db, err := sqlite.OpenDB(path, password)
+	db, err := sqlite.OpenUnecryptedDB(path)
```
andremedeiros: Make sure this change gets reverted.

appdatabase/migrations/sql/0001_app.up.sql
```diff
@@ -56,8 +56,8 @@ sender VARCHAR NOT NULL,
 receipt BLOB,
 log BLOB,
 type VARCHAR NOT NULL,
-FOREIGN KEY(network_id,blk_hash) REFERENCES blocks(network_id,hash) ON DELETE CASCADE,
```
andremedeiros, line 59: Why were these removed?

services/wallet/api.go
```diff
@@ -28,7 +28,7 @@ type API struct {
 // TODO(dshulyak) benchmark loading many transfers from database. We can avoid json unmarshal/marshal if we will
 // read header, tx and receipt as a raw json.
 func (api *API) GetTransfers(ctx context.Context, start, end *hexutil.Big) ([]TransferView, error) {
-	log.Debug("call to get transfers", "start", start, "end", end)
```
andremedeiros: Make sure this change gets reverted.

services/wallet/api.go
```diff
@@ -45,7 +45,7 @@ func (api *API) GetTransfers(ctx context.Context, start, end *hexutil.Big) ([]Tr
 
 // GetTransfersByAddress returns transfers for a single address between two blocks.
 func (api *API) GetTransfersByAddress(ctx context.Context, address common.Address, start, end *hexutil.Big) ([]TransferView, error) {
-	log.Debug("call to get transfers for an address", "address", address, "start", start, "end", end)
```
andremedeiros, line 48: Make sure this change gets reverted.

services/wallet/database.go
```diff
@@ -320,13 +526,71 @@ func (db *Database) DeleteCustomToken(address common.Address) error {
...
+			return nil, err
+		}
+
+		res[loaded] = count
+	}
+
+	return res, nil
+}
+
+func (db *Database) getTransfersStats(address common.Address) int64 {
```
andremedeiros: Needs error checking.

services/wallet/database.go
```diff
@@ -320,13 +526,71 @@ func (db *Database) DeleteCustomToken(address common.Address) error {
 	return err
 }
 
+func (db *Database) getBlocksStats(address common.Address) (map[int64]int64, error) {
```
andremedeiros: Needs error checking.

services/wallet/database.go
```diff
@@ -320,13 +526,71 @@ func (db *Database) DeleteCustomToken(address common.Address) error {
...
+		err = rows.Scan(&res)
+		if err != nil {
+			return 0
+		}
+	}
+
+	return res
+}
+
+func (db *Database) GetTransfersStats(address common.Address) (map[int64]int64, int64) {
```
andremedeiros: Needs to pass potential error as return value.

services/wallet/api.go
```diff
@@ -60,6 +60,157 @@ func (api *API) GetTransfersByAddress(ctx context.Context, address common.Addres
...
 }
 
+type StatsView struct {
+	BlocksStats    map[int64]int64 `json:"blocksStats"`
+	TransfersCount int64           `json:"transfersCount"`
+}
+
+// GetTransfersStats
+func (api *API) GetTransfersStats(ctx context.Context, address common.Address) (*StatsView, error) {
+	blocksStats, transfersCount := api.s.db.GetTransfersStats(address)
```
andremedeiros: Needs error checking (after implementing passing error up from these calls.)

services/wallet/api.go
```diff
@@ -60,6 +60,157 @@ func (api *API) GetTransfersByAddress(ctx context.Context, address common.Addres
...
+
+	return &StatsView{
+		BlocksStats:    blocksStats,
+		TransfersCount: transfersCount,
+	}, nil
+}
+
+// GetTransfersFromBlock
+func (a *API) GetTransfersFromBlock(ctx context.Context, address common.Address, block *hexutil.Big) ([]TransferView, error) {
+	log.Info("call to get transfers by block", "address", address, "block", block)
```
andremedeiros: Should be `log.Debug`

services/wallet/api.go
```diff
@@ -60,6 +60,157 @@ func (api *API) GetTransfersByAddress(ctx context.Context, address common.Addres
...
+	if a.s.db == nil {
+		return nil, ErrServiceNotInitialized
+	}
+
+	blocksByAddress := make(map[common.Address][]*big.Int)
+	blocksByAddress[address] = []*big.Int{block.ToInt()}
+	ctxt := context.Background()
```
andremedeiros: Ignoring the context that was passed. Pass `ctx` below.

services/wallet/api.go
```diff
@@ -60,6 +60,157 @@ func (api *API) GetTransfersByAddress(ctx context.Context, address common.Addres
...
+	if err != nil {
+		return nil, err
+	}
+
+	return castToTransferViews(rst), nil
+}
+
+// GetTransfersByAddressPage returns transfers for a single address between two blocks.
+func (a *API) GetTransfersByAddressPage(ctx context.Context, address common.Address, beforeBlock, pageSize *hexutil.Big) ([]TransferView, error) {
+	log.Info("call to get transfers for an address", "address", address, "block", beforeBlock, "pageSize", pageSize)
```
andremedeiros: Should be `log.Debug`

services/wallet/api.go
```diff
@@ -60,6 +60,157 @@ func (api *API) GetTransfersByAddress(ctx context.Context, address common.Addres
...
+		return nil, ErrServiceNotInitialized
+	}
+	rst, err := a.s.db.GetTransfersByAddressAndPage(address, beforeBlock.ToInt(), pageSize.ToInt().Int64())
+	if err != nil {
+		return nil, err
+	}
+
+	transfersCount := big.NewInt(int64(len(rst)))
+	if pageSize.ToInt().Cmp(transfersCount) == 1 {
+		log.Info("page size is bigger than number of fetched transfers, try to fetch more transfers")
```
andremedeiros: Should be `log.Debug`

services/wallet/api.go
```diff
@@ -60,6 +60,157 @@ func (api *API) GetTransfersByAddress(ctx context.Context, address common.Addres
...
+
+		ctxt := context.Background()
```
andremedeiros: Ignores the passed in context. Should pass `ctx` below.

services/wallet/api.go
```diff
@@ -60,6 +60,157 @@ func (api *API) GetTransfersByAddress(ctx context.Context, address common.Addres
...
+			db:            a.s.db,
+			chain:         a.s.reactor.chain,
+			client:        a.s.client,
+			balanceCache:  balanceCache,
+			feed:          a.s.feed,
+			fromByAddress: fromByAddress,
+			toByAddress:   toByAddress,
+		}
+
+		err = blocksCommand.Command()(ctxt)
```
andremedeiros: Should read:

```go
if err = blocksCommand.Command()(ctxt); err != nil {
    return nil, err
}
```

services/wallet/balance_cache.go
```diff
@@ -0,0 +1,90 @@
...
+type balanceCache struct {
+	// cache maps an address to a map of a block number and the balance of this particular address
+	cache            map[common.Address]map[*big.Int]*big.Int
+	requestCounter   map[common.Address]uint
+	cacheHitsCounter map[common.Address]uint
+	lock             sync.RWMutex
+}
+
+func (b *balanceCache) readCachedBalance(account common.Address, blockNumber *big.Int) *big.Int {
+	b.lock.RLock()
```
andremedeiros: 👏

services/wallet/balance_cache.go
```diff
@@ -0,0 +1,90 @@
...
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+type balanceCache struct {
+	// cache maps an address to a map of a block number and the balance of this particular address
+	cache            map[common.Address]map[*big.Int]*big.Int
+	requestCounter   map[common.Address]uint
+	cacheHitsCounter map[common.Address]uint
+	lock             sync.RWMutex
```
andremedeiros: This is usually called `rw`

services/wallet/concurrent.go
```diff
@@ -40,42 +45,202 @@ func (r *Result) Get() []Transfer {
...
+			return nil
+		})
+
+	}
+
+	select {
+	case <-c.WaitAsync():
+	case <-ctx.Done():
+		log.Error("eth downloader is stuck")
+		return nil, nil, errors.New("eth downloader is stuck")
```
andremedeiros: Consider making this error a "constant"

services/wallet/concurrent.go
```diff
@@ -40,42 +45,202 @@ func (r *Result) Get() []Transfer {
...
+	}
+
+	if c.Error() != nil {
+		log.Error("failed to dowload transfers using concurrent downloader", "error", c.Error())
+		return nil, nil, c.Error()
```
andremedeiros: Errors should be wrapped with context.

services/wallet/concurrent.go
```diff
@@ -40,42 +45,202 @@ func (r *Result) Get() []Transfer {
...
+	return c.GetRanges(), c.GetBlocks(), nil
+}
+
+func findBlocksWithEthTransfers(parent context.Context, client BalanceReader, cache *balanceCache, downloader TransferDownloader, account common.Address, low, high *big.Int) (from *big.Int, blocks []*big.Int, err error) {
+	ranges := [][]*big.Int{{low, high}}
+	minBlock := big.NewInt(low.Int64())
+	blocks = []*big.Int{}
+	var lvl = 1
+	for len(ranges) > 0 && lvl <= 30 {
+		log.Info("check blocks ranges", "lvl", lvl, "ranges len", len(ranges))
```
andremedeiros: Should be `log.Debug`

services/wallet/concurrent.go
```diff
@@ -40,42 +45,202 @@ func (r *Result) Get() []Transfer {
...
+		lvl = lvl + 1
```
andremedeiros:
```go
lvl++
```

services/wallet/concurrent.go
```diff
@@ -40,42 +45,202 @@ func (r *Result) Get() []Transfer {
...
+		newRanges, newBlocks, err := checkRanges(parent, client, cache, downloader, account, ranges)
+		blocks = append(blocks, newBlocks...)
+		if err != nil {
```
andremedeiros, line 178: Shouldn't `err` be checked before appending?

services/wallet/concurrent.go
```diff
@@ -40,42 +45,202 @@ func (r *Result) Get() []Transfer {
...
+			return nil, nil, err
+		}
+
+		log.Info("found new ranges", "account", account, "lvl", lvl, "new ranges len", len(newRanges))
```
andremedeiros: Should be `log.Debug`

services/wallet/concurrent.go
```diff
@@ -40,42 +45,202 @@ func (r *Result) Get() []Transfer {
...
+
+func downloadEthConcurrently(c *ConcurrentDownloader, client BalanceReader, cache *balanceCache, downloader TransferDownloader, account common.Address, low, high *big.Int) {
 	c.Add(func(ctx context.Context) error {
 		if low.Cmp(high) >= 0 {
 			return nil
 		}
 		log.Debug("eth transfers comparing blocks", "low", low, "high", high)
-		lb, err := client.BalanceAt(ctx, account, low)
+		lb, err := cache.BalanceAt(client, ctx, account, low)
+		//lb, err := client.BalanceAt(ctx, account, low)
```
andremedeiros: Remove dead code.

services/wallet/concurrent.go
```diff
@@ -40,42 +45,202 @@ func (r *Result) Get() []Transfer {
...
 		if err != nil {
 			return err
 		}
-		hb, err := client.BalanceAt(ctx, account, high)
+		hb, err := cache.BalanceAt(client, ctx, account, high)
+		//hb, err := client.BalanceAt(ctx, account, high)
```
andremedeiros: Remove dead code.

services/wallet/concurrent.go
```diff
@@ -40,42 +45,202 @@ func (r *Result) Get() []Transfer {
...
+			}
+			if ln == hn {
+				log.Debug("transaction count is also equal", "low", low, "high", high)
+				return nil
+			}
 		}
 		if new(big.Int).Sub(high, low).Cmp(one) == 0 {
-			transfers, err := downloader.GetTransfersByNumber(ctx, high)
+			c.PushBlock(high)
+			/*transfers, err := downloader.GetTransfersByNumber(ctx, high)
```
andremedeiros: Remove dead code.

services/wallet/concurrent_test.go
```diff
@@ -4,7 +4,7 @@ import (
 	"context"
 	"errors"
 	"math/big"
-	"sort"
+	//"sort"
```
andremedeiros: Remove unused import.

services/wallet/concurrent_test.go
```diff
@@ -105,11 +105,11 @@ func TestConcurrentEthDownloader(t *testing.T) {
 		},
 	} {
 		t.Run(tc.desc, func(t *testing.T) {
-			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			/*ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
andremedeiros: Should re-enable test and figure out why it's failing

services/wallet/database.go
```diff
@@ -154,6 +209,157 @@ func (db *Database) GetTransfersByAddress(address common.Address, start, end *bi
...
+func (db *Database) RemoveBlockWithTransfer(address common.Address, block *big.Int) error {
+	query := `DELETE FROM blocks_with_transactions 
+	WHERE address = ? 
+	AND blk_number = ? 
+	AND network_id = ?`
+
+	_, err := db.db.Exec(query, address, (*SQLBigInt)(block), db.network)
+
+	if err != nil {
+		log.Info("block wasn't removed", "block", block, "error", err)
```
andremedeiros: This might need to be a `log.Error`

services/wallet/database_test.go
```diff
@@ -15,11 +14,12 @@ import (
 func setupTestDB(t *testing.T) (*Database, func()) {
 	tmpfile, err := ioutil.TempFile("", "wallet-tests-")
 	require.NoError(t, err)
+	t.Log(tmpfile.Name())
```
andremedeiros: Remove when done.

services/wallet/database_test.go
```diff
@@ -15,11 +14,12 @@ import (
...
 	db, err := appdatabase.InitializeDB(tmpfile.Name(), "wallet-tests")
 	require.NoError(t, err)
 	return NewDB(db, 1777), func() {
 		require.NoError(t, db.Close())
-		require.NoError(t, os.Remove(tmpfile.Name()))
+		//require.NoError(t, os.Remove(tmpfile.Name()))
```
andremedeiros: Uncomment when done.

services/wallet/database_test.go
```diff
@@ -68,6 +68,29 @@ func TestDBHeaderDoesntExist(t *testing.T) {
 	require.False(t, rst)
 }
 
+func TestDBProcessBlocks(t *testing.T) {
```
andremedeiros, line 50: Test should be rewritten to assert state.

services/wallet/downloader.go
```diff
@@ -111,24 +111,30 @@ func (d *ETHTransferDownloader) getTransfersInBlock(ctx context.Context, blk *ty
 				if err != nil {
 					return nil, err
 				}
-				if isTokenTransfer(receipt.Logs) {
-					log.Debug("eth downloader found token transfer", "hash", tx.Hash())
-					continue
+				transactionLog := getTokenLog(receipt.Logs)
+				transactionType := ethTransfer
+
+				log.Info("eth downloader found transfer", "block", blk.Number(), "hash", tx.Hash())
```
andremedeiros: Should be `log.Debug`

services/wallet/downloader.go
```diff
@@ -111,24 +111,30 @@ func (d *ETHTransferDownloader) getTransfersInBlock(ctx context.Context, blk *ty
...
+				if transactionLog != nil {
+					log.Info("eth downloader found token transfer", "hash", tx.Hash())
```
andremedeiros: Should be `log.Debug`

services/wallet/downloader.go
```diff
@@ -111,24 +111,30 @@ func (d *ETHTransferDownloader) getTransfersInBlock(ctx context.Context, blk *ty
...
 					Transaction: tx,
 					From:        from,
-					Receipt:     receipt})
+					Receipt:     receipt,
+					Log:         transactionLog})
 
 			}
 		}
 	}
+	log.Info("getTransfersInBlock found", "len", len(rst))
```
andremedeiros: Should be `log.Debug`

services/wallet/downloader.go
```diff
@@ -234,6 +240,26 @@ func (d *ERC20TransfersDownloader) transfersFromLogs(parent context.Context, log
 	return concurrent.Get(), concurrent.Error()
 }
 
+func (d *ERC20TransfersDownloader) blocksFromLogs(parent context.Context, logs []types.Log, address common.Address) ([]*big.Int, error) {
+	concurrent := NewConcurrentDownloader(parent)
+	for i := range logs {
```
andremedeiros, line 284:
```go
for _, log := range logs {

}
```

services/wallet/downloader.go
```diff
@@ -234,6 +240,26 @@ func (d *ERC20TransfersDownloader) transfersFromLogs(parent context.Context, log
...
+		}
+		concurrent.Add(func(ctx context.Context) error {
+			concurrent.PushBlock(big.NewInt(int64(l.BlockNumber)))
+			return nil
+		})
+	}
+	select {
+	case <-concurrent.WaitAsync():
+	case <-parent.Done():
+		return nil, errors.New("logs downloader stuck")
```
andremedeiros: Consider making this error a "constant"

services/wallet/reactor.go
```diff
@@ -93,6 +87,19 @@ func (r *Reactor) Start(accounts []common.Address) error {
...
+
+	return ctl
+}
+
+// Start runs reactor loop in background.
+func (r *Reactor) Start(accounts []common.Address) error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	if r.group != nil {
+		return errors.New("already running")
```
andremedeiros: consider making error a "constant".

Also this could gain from using `sync.Once` but it's probably outside the scope of this PR

andremedeiros (dismissed):
Great fucking work. Couple little nitpicks and fixes, but it looks solid.

appdatabase/database.go
rasom: sure

appdatabase/migrations/sql/0001_app.up.sql
rasom: because I used a different table for this PR, as I didn't want to touch code which is not related to scanning. I can rename `blocks_with_transfers` to `blocks` now and put this key back. 

The difference is that `blocks` was used for saving blocks while app is running (besides scanning), and we stored even blocks which did not contain transaction we want to fetch, and it feels like unnecessary thing which might also create a lot of garbage in local db.

services/wallet/api.go
rasom: The reason I used `Info` instead of `Debug` is that it was too noisy. There should be better way to filter these logs, but I'm not aware of it. Except of adding some custom string to all logs of interest, but that would cause almost the same issue.

services/wallet/downloader.go
rasom: this is actually copied code, I will change it but I kind of can't take credit for it https://github.com/status-im/status-go/blob/0b403ce25bf25be672f9b39ba97384cb2b72b442/services/wallet/downloader.go#L215 :)

services/wallet/api.go
rasom: I'm going to remove this method, this was rather for testing purposes, it is useless otherwise

services/wallet/concurrent.go
rasom: again, I'll do that, but that's the code which already exists in project

services/wallet/concurrent.go
rasom: it should be, although `blocks` is "lost" in case of error anyway

services/wallet/reactor.go
rasom: > consider making error a "constant".

https://github.com/status-im/status-go/blob/c8a911ebd1e5f6c567d8ff6f3700309b988f0d42/services/wallet/reactor.go#L79

ok 

> Also this could gain from using sync.Once

yep it could, if I knew what it means :D

services/accounts/accounts.go
```diff
@@ -20,6 +21,7 @@ type API struct {
 }
 
 func (api *API) SaveAccounts(ctx context.Context, accounts []accounts.Account) error {
+	log.Info("save accounts")
```
adambabik:
```suggestion
	log.Info("[AccountsAPI::SaveAccounts]")
```

services/wallet/api.go
```diff
@@ -60,6 +60,139 @@ func (api *API) GetTransfersByAddress(ctx context.Context, address common.Addres
...
 }
 
+type StatsView struct {
+	BlocksStats    map[int64]int64 `json:"blocksStats"`
+	TransfersCount int64           `json:"transfersCount"`
+}
+
+// GetTransfersFromBlock
+func (a *API) GetTransfersFromBlock(ctx context.Context, address common.Address, block *hexutil.Big) ([]TransferView, error) {
+	log.Debug("call to get transfers by block", "address", address, "block", block)
```
adambabik:
```suggestion
	log.Debug("[WalletAPI:: GetTransfersFromBlock] get transfers from block", "address", address, "block", block)
```

services/wallet/api.go
```diff
@@ -60,6 +60,139 @@ func (api *API) GetTransfersByAddress(ctx context.Context, address common.Addres
...
+	rst, err := a.s.db.GetTransfersByAddress(address, block.ToInt(), block.ToInt())
+	if err != nil {
+		return nil, err
+	}
+
+	return castToTransferViews(rst), nil
+}
+
+// GetTransfersByAddressPage returns transfers for a single address between two blocks.
+func (a *API) GetTransfersByAddressPage(ctx context.Context, address common.Address, beforeBlock, pageSize *hexutil.Big) ([]TransferView, error) {
```
adambabik: What is `pageSize`?

services/wallet/api.go
```diff
@@ -60,6 +60,139 @@ func (api *API) GetTransfersByAddress(ctx context.Context, address common.Addres
...
+			feed:          a.s.feed,
+			fromByAddress: fromByAddress,
+			toByAddress:   toByAddress,
+		}
+
+		if err = blocksCommand.Command()(ctx); err != nil {
+			return nil, err
+		}
+
+		blocks, err := a.s.db.GetBlocksByAddress(address, 40)
```
adambabik: Let's move this magic number to a const.

services/wallet/api.go
```diff
@@ -60,6 +60,139 @@ func (api *API) GetTransfersByAddress(ctx context.Context, address common.Addres
...
+	}
+
+	return castToTransferViews(rst), nil
+}
+
+type BlockView struct {
+	Block *big.Int `json:"number"`
+}
+
+func castToBlockViews(blocks []*big.Int) []BlockView {
```
adambabik: What's the purpose of that? Instead of an array of ints, we have an array of objects `{ "block": value }`.

services/wallet/transfers_query.go
```diff
@@ -45,16 +46,26 @@ func (q *transfersQuery) FilterEnd(end *big.Int) *transfersQuery {
...
 		q.added = true
-		q.buf.WriteString(" blocks.number <= ?")
+		q.buf.WriteString(" blk_number <= ?")
 		q.args = append(q.args, (*SQLBigInt)(end))
 	}
 	return q
 }
 
+func (q *transfersQuery) FilterLoaded(loaded int) *transfersQuery {
+
```
adambabik:
```suggestion
```

adambabik:
It still would use some clean up (lines commented out etc.).

adambabik:
@rasom I wonder if it would be possible to have like a fake client with stubs that would help to test the logic of this? How do you test whether it yields the correct result for some more complicated cases?

andremedeiros:
> I wonder if it would be possible to have like a fake client with stubs

In Ruby we have VCRs where it records requests and their responses, is there a similar tool in Go that can be used here?

adambabik:
@andremedeiros there is [httptest](https://golang.org/pkg/net/http/httptest/#Server) but it's low-level. 

After giving it some thoughts, ideally, we could have something like `ChainData` provider which would hide the fact that we use an HTTP client at all. We could stub `ChainData` with some fake data describing a scenario and this tx fetching code would call it to calculate the output. In this way, we could test corner cases like balance equal `0` even though having a history of transactions or assert the number of calls.

andremedeiros:
@adambabik I like the idea. How strongly do you feel about shipping this vs incurring some debt for post v1?

adambabik:
@andremedeiros happy to do it later. Let's create an issue and specify what kind of scenarios we need to cover.

andremedeiros:
@adambabik thanks for being flexible on this. I'd love to spend some time with you looking at the other tech debt that we have here and budget time to fix that too.

services/wallet/api.go
rasom: it is useless, leftover

services/wallet/api.go
rasom: that's limit actually, renamed

andremedeiros (approved):

PR #1781, opened 2020-01-10
specify gomarkdown/markdown version to fix conflicts

jakubgs:
This should solve the conflict I have when building [`status-im/matterbridge`](https://github.com/status-im/matterbridge):
```
go: github.com/status-im/status-go/protocol@v1.1.0 requires
    github.com/gomarkdown/markdown@v0.0.0-20191209105822-e3ba6c6109ba: invalid version: unknown revision e3ba6c6109ba
```
And this due to `matterbridge` using `ToHTML()`:
```
# github.com/42wim/matterbridge/bridge/helper
bridge/helper/helper.go:184:20: undefined: markdown.ToHTML
```
Also bumped verison to `0.39.2`.


VERSION
```diff
@@ -1 +1 @@
-0.39.1
+0.39.2
```
adambabik, line 1: I already bumped it.
```suggestion
0.39.3
```

protocol/go.mod
```diff
@@ -4,7 +4,7 @@ go 1.13
 
 replace github.com/ethereum/go-ethereum v1.9.5 => github.com/status-im/go-ethereum v1.9.5-status.7
 
-replace github.com/gomarkdown/markdown => github.com/status-im/markdown v0.0.0-20191209105822-e3ba6c6109ba
+replace github.com/gomarkdown/markdown v0.0.0-20191104174740-4d42851d4d5a => github.com/status-im/markdown v0.0.0-20191209105822-e3ba6c6109ba
```
adambabik, line 7: We need the same in `go.mod` in the root dir.

adambabik (commented):
Please rebase.

VERSION
adambabik: Ups, sorry I did not.

protocol/go.mod
jakubgs: Not sure why since it works without it but okay.

adambabik (approved):

PR #1073, opened 2018-06-29
Update to golangci-lint 1.8.1. Get rid of standalong misspell linter.

pedropombeiro:
For some reason `govet` is complaining about missing `lib/C` and `signal/C` folders. The workaround here was to ignore any subfolders of those 2 folders, but alternatively we could run the linter once for govet (without those folders) and a second time for the rest (although we'd lose some performance gains by doing that).

PR #1785, opened 2020-01-11
bump version to 0.39.3

jakubgs:
I'm trying to fix the build issues with my changes #1073. I tried building with them but it fails with:

```
$ go get github.com/status-im/status-go@330d177de6554eee61a85d9eff7582c99fbbfcad
go: finding github.com 330d177de6554eee61a85d9eff7582c99fbbfcad
go: finding github.com/status-im/status-go 330d177de6554eee61a85d9eff7582c99fbbfcad
go: finding github.com/status-im 330d177de6554eee61a85d9eff7582c99fbbfcad
go: downloading github.com/status-im/status-go v0.39.3-0.20200111184124-330d177de655
go: extracting github.com/status-im/status-go v0.39.3-0.20200111184124-330d177de655
go get: github.com/status-im/status-go@v0.39.3-0.20200111184124-330d177de655 requires
github.com/status-im/status-go/protocol@v1.1.0 requires
github.com/gomarkdown/markdown@v0.0.0-20191209105822-e3ba6c6109ba: invalid version: unknown revision e3ba6c6109ba
```

I assume I have to update `protocol` and tag that with `protocol/v1.1.1` but I tried that and it failed anyway. I found out that for some reason `protocol` depends on `status-go` and `status-go` depends on `protocol`:
https://github.com/status-im/status-go/blob/330d177de6554eee61a85d9eff7582c99fbbfcad/go.mod#L41
https://github.com/status-im/status-go/blob/330d177de6554eee61a85d9eff7582c99fbbfcad/protocol/go.mod#L26
This creates some kind of weird loop. I really think we need to drop the whole submodules thing because it creates so many issues it's beyond retarded.


jakubgs:
Also, I can't really find how exactly `status-go/protocol` uses `status-go`, it doesn't seem to.

adambabik:
I run `go mod why -m github.com/status-im/status-go` in `protocol/` and it does not need it. It is removed when run `go mod tidy` in `protocol/`.

At the same time, I agree that currently submodules cause more harm than good. I will prepare a PR which gets rid of them. We can always introduce them later.

adambabik:
I believe this was solved.

PR #1789, opened 2020-01-13
Report correct number of pruned rows in Postgres mailserver

adambabik:
When using Postgres mailsever, the reported pruned number of rows is always 0. This change uses `RowsAffected` to provide the correct number.

## Note

Before merging, the query `DELETE FROM envelopes WHERE id BETWEEN $1 AND $2` should be verified.

cc @cammellos @jakubgs

### DIFF
--- a/mailserver/mailserver_db_postgres.go
+++ b/mailserver/mailserver_db_postgres.go
@@ -171,11 +171,15 @@ func (i *PostgresDB) Prune(t time.Time, batch int) (int, error) {
 	}
 	defer stmt.Close()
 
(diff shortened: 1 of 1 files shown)


jakubgs:
Oh, so it was the counter that was wrong? Are we actually pruning?
I was kinda looking forward to a `created_at` column...

adambabik:
I confirmed it works. It looks like it compares byte-by-byte and the ID's prefix is a big-endian encoded timestamp.

jakubgs (approved):
THX

PR #1791, opened 2020-01-13
Sync installation messages & contact requests

cammellos:
This is the last messages to move to protobuf.

A few other things:

- Fixes an issue with group chats (we use clock values from the chat now, so that system messages are correctly ordered)
- Fixes an issue with serialization of stickers, as we don't wan't to `omitempty` as 0 is a valid value for packs
- Change interface of a currently not used endpoint `SendTransaction`


protocol/messenger.go
```diff
@@ -966,6 +966,19 @@ func (m *Messenger) LeaveGroupChat(ctx context.Context, chatID string) (*Messeng
 }
 
 func (m *Messenger) saveChat(chat *Chat) error {
+	_, ok := m.allChats[chat.ID]
+	if !ok && chat.Active {
+		if err := m.Join(*chat); err != nil {
```
adambabik: It seems like `saveChat` has more than one responsibility here. Why do we need to join here?

protocol/messenger.go
```diff
@@ -1328,6 +1336,13 @@ func (m *Messenger) SendContactUpdates(ctx context.Context, ensName, profileImag
 	m.mutex.Lock()
 	defer m.mutex.Unlock()
 
+	myID := types.EncodeHex(crypto.FromECDSAPub(&m.identity.PublicKey))
```
adambabik: Maybe there should be a helper function called `contactIDFromPublicKey` or something? It is repeated a few times in this PR.

protocol/messenger.go
cammellos: we can be explicit about when to join a chat, but basically every time we save a new chat we want to join it (if active).
We can also split it, but basically we will be duplicating a bit of code, in at least 3 places: 

- when status-react calls `SaveChat`->`Join`

- when we receive a one to one message `SaveChat`->`Join`

- when we receive a `SyncPublicChatMessage` `SaveChat`->`Join` 

So I felt having it always is probably safer (I can't see a case where you would want to `Save` a chat but not join if active), but happy to go with anything to be honest

adambabik: So maybe we can have a public method `SaveAndJoin()` which is implemented as:
```go
func (m *Messenger) SaveAndJoin(chat *Chat) error {
    if err := m.saveChat(chat); err != nil {
        return err
    }
    return m.join(chat)
}

func (m *Messenger) Join(chat Chat) error {
    return m.join(chat)
}
```

cammellos: Actually, I have removed that bit, it's not needed as of now as filters are managed explicitly by status-react.

Once we go around moving filters management in status-go, we might re-think and go for something as you described.

cammellos:
@adambabik ready to be reviewed again, thanks!

go.mod
```diff
@@ -8,7 +8,7 @@ replace github.com/Sirupsen/logrus v1.4.2 => github.com/sirupsen/logrus v1.4.2
 
 replace github.com/docker/docker => github.com/docker/engine v1.4.2-0.20190717161051-705d9623b7c1
 
-replace github.com/gomarkdown/markdown v0.0.0-20191104174740-4d42851d4d5a => github.com/status-im/markdown v0.0.0-20191209105822-e3ba6c6109ba
+replace github.com/gomarkdown/markdown => github.com/status-im/markdown v0.0.0-20191209105822-e3ba6c6109ba
```
adambabik: Can we keep it with the version?

protocol/go.mod
```diff
@@ -4,7 +4,7 @@ go 1.13
 
 replace github.com/ethereum/go-ethereum v1.9.5 => github.com/status-im/go-ethereum v1.9.5-status.7
 
-replace github.com/gomarkdown/markdown v0.0.0-20191104174740-4d42851d4d5a => github.com/status-im/markdown v0.0.0-20191209105822-e3ba6c6109ba
+replace github.com/gomarkdown/markdown => github.com/status-im/markdown v0.0.0-20191209105822-e3ba6c6109ba
```
adambabik: Same here? Otherwise, it breaks a matterbridge.

protocol/message.go
```diff
@@ -159,10 +163,16 @@ func (m *Message) MarshalJSON() ([]byte, error) {
 		Timestamp:         m.Timestamp,
 		ContentType:       m.ContentType,
 		MessageType:       m.MessageType,
-		Sticker:           m.GetSticker(),
 		CommandParameters: m.CommandParameters,
 	}
 
+	sticker := m.GetSticker()
+	if m.GetSticker() != nil {
```
adambabik:
```suggestion
	if sticker := m.GetSticker(); sticker != nil {
```

protocol/message.go
```diff
@@ -159,10 +163,16 @@ func (m *Message) MarshalJSON() ([]byte, error) {
 		Timestamp:         m.Timestamp,
 		ContentType:       m.ContentType,
 		MessageType:       m.MessageType,
-		Sticker:           m.GetSticker(),
 		CommandParameters: m.CommandParameters,
 	}
 
+	sticker := m.GetSticker()
```
adambabik:
```suggestion
```

protocol/messenger.go
```diff
@@ -1400,7 +1415,8 @@ func (m *Messenger) sendContactUpdate(ctx context.Context, chatID, ensName, prof
 		return nil, err
 	}
 
-	if !contact.IsAdded() {
+	if !contact.IsAdded() && contact.ID != contactIDFromPublicKey(&m.identity.PublicKey) {
+
```
```suggestion
```

protocol/messenger.go
```diff
@@ -2647,3 +2749,8 @@ func (m *Messenger) ValidateTransactions(ctx context.Context, addresses []types.
 	}
 	return &response, nil
 }
+
+func contactIDFromPublicKey(key *ecdsa.PublicKey) string {
```
adambabik: Maybe in `contact.go` instead?

cammellos:
@adambabik thanks for fixing the issue with markdown, I have address the remaining feedback.
cheers

adambabik (approved):

Issue #1677, opened 2019-11-15
Nimbus node in status-go


pedropombeiro:
# Problem
In the long run, we want to replace Geth with Nimbus. The goal of this issue is to allow a side-by-side implementation of Nimbus that can be selected at runtime, through a new field in `params.NodeConfig`. The minimalistic scenario to support is:

1. Creating an in-device account
1. Adding a public chat
1. Sending and receiving a message in the public chat
1. Logging out

# Implementation
<details>
<summary>Known steps towards feature implementation.</summary>

* [x] Cross-compile Nimbus for Android
  * [x] arm64
  * [x] x86 (for Android emulator)
* [x] Link Nimbus Android build with status-go
* [x] Test app deployed to emulator
* [x] Implement `AddPeer`
* [ ] Connect status-go chat account with Nimbus
  * [x] Extract geth types from status-go's `account.Manager` (e.g. `keystore.KeyStore`, `accounts.Manager`)
  * [ ] Create Nimbus version of the geth KeyStore type
* [ ] Connect status-go with Nimbus RPC
  * [x] Connect to local services (`shhext`, `account`, etc)
  * [ ] Implement/adapt routing
  * [ ] Implement missing RPC functions
* [x] Create Nimbus version of `ShhExtService`
* [ ] Integrate Nimbus time drift correction service.
* [ ] ...

</details>

I've added logging to `mobile/status.go` methods that are called by status-react to get a sense of the minimalistic API that would enable this scenario. The logs are here: [log.txt](https://github.com/status-im/status-go/files/3931678/log.txt).

<details>
<summary>List of required functions</summary>
Based on these logs, we can see that the functions required in `status.go` are:

_Note: Required functions are prefixed with a `*`_

* `*`[`AddPeer`](https://github.com/status-im/status-go/blob/9a3ed980c9dcbc6e55fcba843224e447b06eb08f/lib/library.go#L646)
* `*`[`CallPrivateRPC`](https://github.com/status-im/status-go/blob/9a3ed980c9dcbc6e55fcba843224e447b06eb08f/lib/library.go#L192)
* Implemented already in status-go:
  * [`settings_saveConfig`](https://github.com/status-im/status-go/blob/da5876f6324659a28e56b5f5c91a034c010570e2/services/accounts/settings.go#L20)
  * Mailserver-related (all map to DB operations so not an issue, they are still called even if `MailserversConfig.Enabled = false`):
[...]

PR #1794, opened 2020-01-15
Nimbus backend initial implementation

pedropombeiro:
The intent of this PR is to merge the Nimbus work done so far in a way that doesn't impact the way we're using status-go (with geth).

Changes:
  - Replaced `mobile/backend.go` with a pared-down Nimbus version (non-implemented methods will panic for the time being);
  - Still using a couple of Geth services instead of Nimbus replacements (waiting on @kdeme here):
    - keystore while we work to expose the Nimbus version;
    - RPC client/server classes;
  - Tests have not been adapted for Nimbus;
  - Some services have been disabled for Nimbus (mailservers, wallet, personal, discovery);

Thought process/architecture behind the change;
  - status-react will leverage Nix to build a `libnimbus.a` lib and feed it to status-go;
  - The idea is to fork logic at the backend level with the entry point used by status-react (`mobile/backend.go`);
  - We'll do compile-time backend selection (controlled by feature flag in `.env` files). Runtime selection is possible given some time investment;
  - Make it possible to iterate with local repos without needing to commit Nimbus/status-go changes;

Part of https://github.com/status-im/status-go/issues/1677


Makefile
```diff
@@ -282,7 +293,8 @@ ci: lint canary-test test-unit test-e2e ##@tests Run all linters and tests at on
 ci-race: lint canary-test test-unit test-e2e-race ##@tests Run all linters and tests at once + race
 
 clean: ##@other Cleanup
-	rm -fr build/bin/* mailserver-config.json
+	rm -fr build/bin/* mailserver-config.json vendor/github.com/status-im/nimbus
```
jakubgs, line 294: wat?

account/accounts_nimbus.go
```diff
@@ -6,23 +6,24 @@ import (
...
+	m.Manager = &Manager{accountsGenerator: generator.New(m)}
 	return m
 }
 
-// InitKeystore sets key manager and key store.
-func (m *Manager) InitKeystore(keydir string) error {
-	m.mu.Lock()
-	defer m.mu.Unlock()
+// // InitKeystore sets key manager and key store.
+// func (m *Manager) InitKeystore(keydir string) error {
```
jakubgs, line 22: Please don't commit commented out stuff like this.

account/keystore_nimbus.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/ethereum/go-ethereum/accounts/keystore"
+
+	gethbridge "github.com/status-im/status-go/eth-node/bridge/geth"
+	// nimbusbridge "github.com/status-im/status-go/eth-node/bridge/nimbus"
+	"github.com/status-im/status-go/eth-node/types"
+)
+
+// makeAccountManager creates ethereum accounts.Manager with single disk backend and lightweight kdf.
+// If keydir is empty new temporary directory with go-ethereum-keystore will be intialized.
+func makeAccountManager(keydir string) (manager *accounts.Manager, err error) {
```
jakubgs, line 19: We're doing a LOT of code duplication with stuff like this. The only difference I see here is the name of the temporary directory which is `nimbus-keystore`. Do we really have to duplicate all this code just to change that?

I guess the main difference is the `gethbridge` used. If that's the case then why can't we parametrize it to accept an interface that can accept both types of bridges?

api/nimbus_backend.go
```diff
@@ -0,0 +1,1027 @@
...
+	defer log.Info("Status backend initialized", "backend", "nimbus", "version", params.Version, "commit", params.GitCommit)
+
+	statusNode := node.NewNimbus()
+	accountManager := account.NewGethManager()
+	// transactor := transactions.NewTransactor()
+	// personalAPI := personal.NewAPI()
+	// rpcFilters := rpcfilters.New(statusNode)
+	return &nimbusStatusBackend{
+		statusNode:     statusNode,
+		accountManager: accountManager,
```
jakubgs, line 90: Same here. All this code duplication just because we want to use a different `AccountManager`? Is that really wise?

jakubgs (commented):
I think a call to explain this PR would be cool.

account/accounts_nimbus.go
pedropombeiro: This is kind of a special case, where we're committing a partially implemented feature that's hidden behind a feature flag. I'm happy to discuss it on a call. Maybe you'd like to join the handover call that I'll be having with @andremedeiros?

jakubgs (approved):
Approving based on conversation I had with Pedro on Thursday. I understand that many of those changes are in a sense temporary and it is intentional that we are avoiding refactoring in order to simplify reaching functionality parity step by step.

PR #1803, opened 2020-01-17
Split shhext into shhext and wakuext

### LINKED ISSUE #1800: Test Waku and Whisper enabled together in a single node
# Problem

If we decide to run Waku and Whisper in a single node, we need to make sure they can coexists.

# Implementation

Write tests which will confirm this setup works as expected.

# Acceptance Criteria

1. Waku and Whisper can coexists within a single node.

adambabik:
From now on, `services/ext` is an abstract implementation. `services/shhext` is a Whisper extension and `services/wakuext` is Waku extension.

Closes #1800


services/shhext/api.go
```diff
@@ -187,103 +34,72 @@ type PublicAPI struct {
...
-	return enode.ParseV4(rawurl)
-}
-
-// RetryConfig specifies configuration for retries with timeout and max amount of retries.
-type RetryConfig struct {
-	BaseTimeout time.Duration
-	// StepTimeout defines duration increase per each retry.
-	StepTimeout time.Duration
-	MaxRetries  int
+// Post is used to send one-to-one for those who did not enabled device-to-device sync,
```
cammellos: This is actually not used anymore

cammellos (approved):
Post,SendDirectMessage/SendPublicMessage are not used anymore, so could be removed, but we can do that in another PR

services/shhext/api.go
adambabik: Awesome, I will delete it.

pedropombeiro:
@adambabik what happened to `shhext/history.go`?

adambabik:
@pombeiro This is not used by status-react so I got rid of it.

jakubgs (approved):
Thx

PR #1804, opened 2020-01-17
Peg clock value to whisper timestamp

cammellos:
This commit pegs the clock value to max +120 seconds from the whisper
timestamp.
In this way we avoid having a client increase arbitrarily the chat clock value.

Validating that the clock value is not too much in the past would be useful, but currently that would break `datasync` (depending on the range), as `datasync` re-transmit the original message in a new whisper message, so it's a case where the difference between the two is ok, but it also means that we are exposed to malicious insertion of messages in the past (there's no way to know if a message has been genuinely re-transmitted by datasync or the user maliciously inserted a message in the history).

One  way to mitigate this is to have a UX solution, whereby those messages are for example marked and displayed with a warning.


protocol/chat.go
```diff
@@ -207,39 +207,39 @@ func oneToOneChatID(publicKey *ecdsa.PublicKey) string {
 	return types.EncodeHex(crypto.FromECDSAPub(publicKey))
 }
 
-func OneToOneFromPublicKey(pk *ecdsa.PublicKey) *Chat {
+func OneToOneFromPublicKey(pk *ecdsa.PublicKey, timesource clockValueTimesource) *Chat {
```
adambabik: This is a public method so for the convenience `clockValueTimesource` should be also public.

protocol/message_validator.go
```diff
@@ -2,17 +2,52 @@ package protocol
 
 import (
 	"errors"
+	"math/big"
 	"strconv"
 	"strings"
 
 	"github.com/status-im/status-go/protocol/protobuf"
+	"github.com/status-im/status-go/protocol/v1"
```
adambabik, line 9: We are in `github.com/status-im/status-go/protocol`. This is not needed.

protocol/message_validator.go
```diff
@@ -2,17 +2,52 @@ package protocol
...
+const maxWhisperDriftMs uint64 = 120000
+
+func validateClockValue(clock uint64, whisperTimestamp uint64) error {
+	if clock == 0 {
 		return errors.New("clock can't be 0")
 	}
 
+	c := new(big.Int).SetUint64(clock)
+	w := new(big.Int).SetUint64(whisperTimestamp)
+	result := new(big.Int).Sub(c, w)
```
adambabik: If we only subtract lower from higher, we should never overflow and we wouldn't need to use `big.Int`, right?

protocol/message_validator.go
```diff
@@ -2,17 +2,52 @@ package protocol
...
+
+	return nil
+}
+
+func ValidateMembershipUpdateMessage(message *protocol.MembershipUpdateMessage, timeNowMs uint64) error {
+
+	for _, e := range message.Events {
+		// We only compare in one direction for membership update as they are relayed
+		// without the original whisper timestamp
+		if e.ClockValue > timeNowMs && e.ClockValue-timeNowMs > maxWhisperDriftMs {
```
adambabik:
```suggestion
		if err := validateClockValue(e.ClockValue, timeNowMs); err != nil {
```

protocol/messenger.go
```diff
@@ -1307,7 +1304,7 @@ func (m *Messenger) SendChatMessage(ctx context.Context, message *Message) (*Mes
 		return nil, errors.New("Chat not found")
 	}
 
-	err := extendMessageFromChat(message, chat, &m.identity.PublicKey)
+	err := extendMessageFromChat(message, chat, &m.identity.PublicKey, m.transport)
```
adambabik: I would recommend to have a getter on `Messenger`:
```suggestion
	err := extendMessageFromChat(message, chat, &m.identity.PublicKey, m.getTimeSource())
```
so that when the implementation changes, you will only need to change it in one place, `getTimeSource` method. It looks odd that we pass `transport` to `extendMessageFromChat` which is a helper.

protocol/message_validator.go
cammellos: How do I import `MembershipUpdateMessage` otherwise? I tried removing it, but without it wont' compile, is there another way? (in `messenger.go` we import I  believe)

cammellos:
@adambabik addressed feedback, 

I actually had to change validating that the clock value is not too much in the past as that would break datasync, an explanation is above.

Thanks!

protocol/message_validator_test.go
```diff
@@ -72,13 +87,15 @@ func (s *MessageValidatorSuite) TestValidateRequestAddressForTransaction() {
 
 func (s *MessageValidatorSuite) TestValidatePlainTextMessage() {
 	testCases := []struct {
-		Name    string
-		Valid   bool
-		Message protobuf.ChatMessage
+		Name             string
+		WhisperTimestamp uint64
+		Valid            bool
```
pedropombeiro, line 92: I know this is not directly related to this PR (other than the fact that this PR added more error types), but instead of true/false, should we compare with the actual expected error message?

pedropombeiro (approved):

protocol/message_validator_test.go
cammellos: I don't mind either approaches, basically `false` here is to say: `This is not to be validated anymore`, while `Error` means that it can be retried. Probably not in this PR, but you think matching the errors explicitly is a better way? If so I can push the change in a separate PR.
thanks!

protocol/message_validator.go
adambabik: Ah sorry. It is needed. I though I deleted this comment...

protocol/transport/whisper/whisper_service.go
```diff
@@ -360,6 +361,10 @@ func (a *WhisperServiceTransport) Track(identifiers [][]byte, hash []byte, newMe
 	}
 }
 
+func (a *WhisperServiceTransport) GetCurrentTime() uint64 {
```
adambabik, line 365: Let's add a comment or a unit test that makes sure it returns unix timestamp in milliseconds.

adambabik (approved):

PR #1805, opened 2020-01-17
Add support for request messages by topics

adambabik:
This change adds querying mailservers by topics. It is only supported in waku subprotocol and using PostgreSQL maislerver implementation.

This is the implementation of [Waku Mailserver spec](https://specs.vac.dev/waku/mailserver.html) v0.2


mailserver/mailserver_db_postgres.go
```diff
@@ -79,26 +82,36 @@ func (i *PostgresDB) BuildIterator(query CursorQuery) (Iterator, error) {
 		// We disable security checks as we need to use string interpolation
 		// for this, but it's converted to 0s and 1s so no injection should be possible
 		/* #nosec */
-		stmtString = fmt.Sprintf("SELECT id, data FROM envelopes where id >= $1 AND id < $2 AND bloom & b'%s'::bit(512) = bloom ORDER BY ID DESC LIMIT $3", toBitString(query.bloom))
+		stmtString = fmt.Sprintf(`
+			SELECT id, data 
+			FROM envelopes 
+			WHERE id >= $1 AND id < $2 AND (bloom & b'%s'::bit(512) = bloom OR topic = any($3))
```
cammellos: I would split the two queries, unless there's an actual use case for specifying both `bloom` and `topic`, reason being that `OR` queries are tricky to index, and we might get degraded performance.

Say something like:
```
if len(topics) > 0 {
  // perform topic query 
} else {
  // perform bloom query
}
```

this way we can create an index on `id` and topics if it does not exists already.

oskarth:
Is this PR closing mailserver spec v0.2 (as it currently stands)? Can we make this explicit? https://specs.vac.dev/waku/mailserver.html

adambabik:
@oskarth yes, it does. One question:
- Do we want to keep it as `topic-interest` or `topics`? I guess it does not really matter but I think it would be nice if the implementation align with the field names used in the spec.

oskarth:
The spec says topic-interest, do you want to change it? We can, but lets make sure it happens in spec repo first. That way @kdeme can also verify.

kdeme:
I'm ok with the name change. Removes confusion between `topic-interest` packet and `topic-interest` parameters.

adambabik:
Spec changes: https://github.com/vacp2p/specs/pull/91

cammellos (approved):

PR #1809, opened 2020-01-22
fix: close resultsets so we don't leak them

andremedeiros:
This PR fixes some cases where resultsets weren't being closed, which
meant that they were slowly leaking.


services/permissions/database.go
```diff
@@ -95,6 +95,7 @@ func (db *Database) GetPermissions() (rst []DappPermissions, err error) {
 		}
 		dapps[perms.Name] = &perms
 	}
+	rows.Close()
```
cammellos: I think you want to close it just after you check for `err`, it might return early and not close

services/browsers/database.go
```diff
@@ -100,6 +100,7 @@ func (db *Database) GetBrowsers() (rst []*Browser, err error) {
 		browsers[browser.ID] = &browser
 		rst = append(rst, &browser)
 	}
+	rows.Close()
```
cammellos: similar here, best to close them on line  `92`

cammellos (changes requested):
Good catch! There's a couple where we close `rows` a bit too late, and the fn might early return before, otherwise looks good

services/permissions/database.go
adambabik: `defer` ftw!

Samyoul:
@cammellos would you mind re-reviewing this? I've addressed the outstanding issues and i believe that it is ready for merging.

jakubgs (approved):
Nice. This was in limbo for too long.

cammellos (approved):

Samyoul:
I just want to keep our repo clean, and this was such a simple tweak that I just did it.

PR #1810, opened 2020-01-22
Fix ens names and contact

cammellos:
Currently the wrong pk format was used to query the in-memory contacts,
this commit corrects the behavior.


yenda (approved):

yenda:
@cammellos is it rebased on latest develop? In which case could you also bump version to 0.39.11 so we merge in status-react along with mailserver fixes?

PR #1812, opened 2020-01-23
bump to 0.40.0

rasom:
@adambabik @andremedeiros 

There is a breaking change in db schema, should it be 0.40.0 then? Or we keep 0.39.x before release?


cammellos:
It should be `0.40.0` if there's a breaking change, it does not matter so much as I understand this is already tested and ready to go , but any breaking change should bump the 0.x version, likely until we are a bit more stable

cammellos (approved):

rasom:
@cammellos changed to 0.40.0

flexsurfer (approved):

PR #1813, opened 2020-01-23
Fix wrong field being logged in  EnvelopesMonitor

pedropombeiro:
Jakub mentioned the following error showed up in a `develop` build:

```
E GoLog   : ERROR[01-23|14:45:57.832] received unexpected data in the the confirmation event namespace=EnvelopesMonitor batch=
```

The `Data` field is not of an expected type and unfortunately we're logging the wrong field there (`Batch`). This PR fixes the logging code.


jakubgs (approved):
Don't you need to run `make vencor` after changing this because of retarded sub-modules?

pedropombeiro:
Indeed I do, good catch!

PR #1824, opened 2020-01-31
Verify ENS in the background

cammellos:
Currently ENS are verified explicitly by status-react, this is not ideal
as if that fails it will have to be explicilty retried in status-react.
This commits changes that behavior so that ENS are verified in a loop
and updated if new messages are received.

status: ready


flexsurfer:
please do not merge until  https://github.com/status-im/status-go/pull/1825  will be merged

protocol/ens.go
```diff
@@ -0,0 +1,57 @@
...
+
+	if c.ENSVerified {
+		return false
+	}
+
+	if c.ENSVerificationRetries >= maxRetries {
+		return false
+	}
+
+	if now < c.ENSVerifiedAt+ENSBackoffTimeMs*c.ENSVerificationRetries*uint64(math.Exp2(float64(c.ENSVerificationRetries))) {
```
jakubgs: A mouthful like this would be nice to be replaced with a `verifiedRecentlyEnough()` helper, something like that. Possibly with a comment about what the `math.Exp2()` is giving us here.

protocol/message_handler.go
```diff
@@ -337,6 +337,17 @@ func (m *MessageHandler) HandleChatMessage(state *ReceivedMessageState) error {
 	state.ModifiedChats[chat.ID] = true
 	state.AllChats[chat.ID] = chat
 
+	contact := state.CurrentMessageState.Contact
+	if overrideENSName(contact, receivedMessage.EnsName, receivedMessage.Clock) {
```
jakubgs: This would probably make more sense as `hasENSNameChanged()`, and maybe add comment like `This should trigger ENS re-verification in VerifyENSNames()`.

params/config.go
```diff
@@ -527,6 +527,9 @@ type ShhextConfig struct {
 	// IMPORTANT: It should always be mainnet unless used for testing
 	VerifyENSURL string
 
+	// VerifyENSContractAddress is the address of the contract used to verify ENS
+	VerifyENSContractAddress string
```
jakubgs, line 532: Maybe add comment explaining that `No default is provided and if not set ENS resolution is disabled` or similar.

jakubgs (commented):
Overall looks good, just few small comments. My main one would be:

Why are we adding ENS stuff into `protocol` package?
Wouldn't it make sense to make a separate `ens` package?
Or at least `protocol/ens`?

protocol/message_handler.go
```diff
@@ -337,6 +337,17 @@ func (m *MessageHandler) HandleChatMessage(state *ReceivedMessageState) error {
...
+		contact.ENSVerifiedAt = 0
+		contact.ENSVerified = false
+		contact.ENSVerificationRetries = 0
+		contact.LastENSClockValue = receivedMessage.Clock
```
jakubgs: What do you think about adding something like:
```go
func (*c Contact) ResetENSVerification(clockValue uint64) {
	contact.ENSVerifiedAt = 0
	contact.ENSVerified = false
	contact.ENSVerificationRetries = 0
	contact.LastENSClockValue = clockValue
}
```
In `protocol/contact.go`.

cammellos:
Thanks for the review, addressed the feedback:

> Why are we adding ENS stuff into protocol package?

The protocol naming is not very descriptive of the responsibilities, basically the `protocol/` package takes care of the handling and state management of the app, and this happens in reaction to a protocol event (receiving a message from a user).

I would move it to `protocol/ens.go`, but currently there would be some circular dependencies as most of the data structures are top level `Contact` for example, and `ENS` depends on `Contact`, while `Messenger` depends on both `Contact` and `ENS`, so we should move first `Contact` to his own namespace and then have `ENS` and  `Messenger` both depend on it, but I'd rather handle this once adam is back, as he is the one who has a better feel on how things should be organized, if that's ok.

jakubgs (approved):

PR #1831, opened 2020-02-03
docs: clarify purpose and describe ports

jakubgs:
Some people were asking about purpose of ports so I'm clarifying it in the docs.

Also make bootnode use `30301` instead of `30303` by default to avoid clash/confusion with Mailserver.

### DIFF
--- a/BOOTNODE.md
+++ b/BOOTNODE.md
@@ -1,5 +1,14 @@
 # Description
 
+A "Bootnode" is a match-maker for all kinds of nodes. For example our [Mailserver](/MAILSERVER.md). It helps a new node find other nodes it can communicate with. It bootstraps connections between them.
+
+A Bootstrap node is a very static thing and should not change its IP. Their addresses are either hardcoded into the node binary or provided via flags or config files on process start.
+
+# Service Ports
+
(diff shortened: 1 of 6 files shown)


BOOTNODE.md
```diff
@@ -1,5 +1,14 @@
 # Description
 
+A "Bootnode" is a match-maker for Whisper nodes like a [Mailserver](/MAILSERVER.md). It helps a new node find other nodes it can communicate with. It bootstraps connections between them.
```
adambabik:
```suggestion
A "Bootnode" is a match-maker for any type of nodes. It helps a new node find other nodes it can communicate with. It bootstraps connections between them.
```

BOOTNODE.md
```diff
@@ -1,5 +1,14 @@
...
+
+A Bootstrap node is a very static thing and should not change it's IP. Their addresses are either hardcoded into the node binary or provded via flags or config files on process start.
```
adambabik:
```suggestion
A Bootstrap node is a very static thing and should not change its IP. Their addresses are either hardcoded into the node binary or provided via flags or config files on process start.
```

BOOTNODE.md
jakubgs: Fixed.

BOOTNODE.md
jakubgs: Fixed.

adambabik (approved):

PR #1832, opened 2020-02-04
Initialize protocol when waku active

cammellos:
Some fixes to get waku mode working in the app.


jakubgs (approved):

api/geth_backend.go
```diff
@@ -886,8 +901,24 @@ func (b *GethStatusBackend) injectAccountIntoServices() error {
 		return err
 	}
 
-	if whisperService != nil {
-		st, err := b.statusNode.ShhExtService()
+	wakuService, err := b.statusNode.WakuService()
+
+	switch err {
+	case node.ErrServiceUnknown: // Whisper was never registered
```
adambabik:
```suggestion
	case node.ErrServiceUnknown: // Waku was never registered
```

api/geth_backend.go
```diff
@@ -886,8 +901,24 @@ func (b *GethStatusBackend) injectAccountIntoServices() error {
...
+	case nil:
+		if err := wakuService.DeleteKeyPairs(); err != nil { // err is not possible; method return value is incorrect
+			return err
+		}
+		b.selectedAccountShhKeyID, err = wakuService.AddKeyPair(identity)
```
adambabik: Let's changed it to `selectedAccountKeyID`. We should also make sure it's not possible for a client to have both Whisper and Waku started. It's fine for the relay node.

PR #1835, opened 2020-02-06
Remove submodules

adambabik:
At the current stage, submodules, especially for `protocol` and `eth-node` packages, bring more troubles than advantages.

The biggest problem is that we use `replace`s in the main `go.mod` which makes status-go act differently depending whether it is used as a main program vs a library in other projects.


protocol/contact.go
```diff
@@ -58,7 +58,7 @@ type Contact struct {
 	SystemTags []string `json:"systemTags"`
 
 	DeviceInfo    []ContactDeviceInfo `json:"deviceInfo"`
-	TributeToTalk string              `json:"tributeToTalk,omitEmpty"`
+	TributeToTalk string              `json:"tributeToTalk,omitempty"`
```
adambabik, line 61: Potentially breaking change but tests pass. cc @cammellos

protocol/message.go
```diff
@@ -93,7 +93,7 @@ type Message struct {
 
 	// Replace indicates that this is a replacement of a message
 	// that has been updated
-	Replace   string           `json:"replace,omitEmpty"`
+	Replace   string           `json:"replace,omitempty"`
```
adambabik, line 96: Potentially breaking change cc @cammellos

protocol/message.go
```diff
@@ -130,10 +129,10 @@ func (m *Message) MarshalJSON() ([]byte, error) {
 		ParsedText        json.RawMessage                  `json:"parsedText"`
 		LineCount         int                              `json:"lineCount"`
 		Text              string                           `json:"text"`
-		ChatId            string                           `json:"chatId"`
+		ChatID            string                           `json:"chatId"`
 		LocalChatID       string                           `json:"localChatId"`
 		Clock             uint64                           `json:"clock"`
-		Replace           string                           `json:"replace,omitEmpty"`
+		Replace           string                           `json:"replace"`
```
adambabik, line 135: A unit test was failing so I removed it. `omitEmpty` did not work, it is `omitempty`. cc @cammellos

protocol/messenger.go
```diff
@@ -74,12 +74,12 @@ type RawResponse struct {
 }
 
 type MessengerResponse struct {
-	Chats         []*Chat                     `json:"chats,omitEmpty"`
-	Messages      []*Message                  `json:"messages,omitEmpty"`
-	Contacts      []*Contact                  `json:"contacts,omitEmpty"`
-	Installations []*multidevice.Installation `json:"installations,omitEmpty"`
+	Chats         []*Chat                     `json:"chats,omitempty"`
```
adambabik, line 77: Potentially breaking change cc @cammellos.

protocol/contact.go
cammellos: This is ok, I believe, tributeToTalk is not used

protocol/message.go
cammellos: This is also likely ok

protocol/message.go
cammellos: good to know, I obviously sprinkled the codebase with the wrong version :)

protocol/messenger.go
cammellos: This is also fine

cammellos (approved):

PR #1837, opened 2020-02-07
Make Messenger.TimeSource public

adambabik:
`github.com/status-im/status-go/protocol` exports a few `Chat` constructors which require a `TimeSource`, however, there is no way to get `TimeSource` from `Messenger`.

This change exports `Messenger.TimeSource` which provides `TimeSource` used by `Messenger` to send messages.

### DIFF
--- a/protocol/messenger.go
+++ b/protocol/messenger.go
@@ -126,6 +126,8 @@ type config struct {
 
 type Option func(*config) error
 
+// WithSystemMessagesTranslations is required for Group Chats which are currently disabled.
+// nolint: unused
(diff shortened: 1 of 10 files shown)


protocol/messenger.go
```diff
@@ -126,13 +126,6 @@ type config struct {
 
 type Option func(*config) error
 
-func WithSystemMessagesTranslations(t map[protobuf.MembershipUpdateEvent_EventType]string) Option {
```
cammellos, line 129: This is currently not used, but will be used once we enable group chats, but we can remove it for the time being and re-introduce it, no worries

cammellos (approved):

PR #1851, opened 2020-02-13
add GetContactByID to Messenger

jakubgs:
This will allow me to more efficiently get ENS names for contacts in our bridge.

Otherwise I'd need to use `Contacts()` and iterate through a list, which is stupid.

### DIFF
--- a/protocol/messenger.go
+++ b/protocol/messenger.go
@@ -1147,8 +1147,18 @@ func (m *Messenger) Contacts() []*Contact {
 	for _, contact := range m.allContacts {
 		contacts = append(contacts, contact)
 	}
-
 	return contacts
+}
+
+// GetContactByID assumes pubKey includes 0x prefix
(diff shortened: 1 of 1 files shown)


cammellos (approved):

protocol/messenger.go
```diff
@@ -1147,10 +1147,20 @@ func (m *Messenger) Contacts() []*Contact {
...
 	return contacts
 }
 
+// GetContactByID assumes pubKey includes 0x prefix
+func (m *Messenger) GetContactByID(pubKey string) (*Contact, error) {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+	contact, ok := m.allContacts[pubKey]
+	if !ok {
+		return nil, errors.New("No contact found")
```
```suggestion
		return nil, errors.New("no contact found")
```

adambabik (approved):

protocol/messenger.go
jakubgs: Fixed.

PR #1861, opened 2020-02-17
Add statusUpdate code

### LINKED ISSUE #1853: Implement code 22 for exchanging options
Implement code 22 for exchanging options

# Problem
Currently we have no way to exchange topic-interest, and we need a way to exchange that information among peers.

# Solution

The proposed solution is to have a new code `22`, which exchange an `ALIST` through `RLP`, which may contain any option that is currently exchanged in the handshake (therefore including topic-interest).

All the options are optional, and we assume that the field has not changed if not present.

For a description of how topic-interest and bloom filters are handled refer to https://notes.status.im/bloom-filter-topic-interest-waku?view .

The handshake will also be changed to allow for the absence of any field, which otherwise will be defaulted as follow:

- for PoW, 0
- light node off
- full bloom filter
- no topic-interest

cammellos:
This commit adds a new code `statusUpdate = 2` which replaces:

`powRequirementCode`
`bloomFilterExCode`
`rateLimitingCode`

Tested against status-react and currently deployed on `eth.test` fleet

Closes https://github.com/status-im/status-go/issues/1853


waku/peer.go
```diff
@@ -277,6 +313,21 @@ func (p *Peer) bloomMatch(env *Envelope) bool {
 	return p.fullNode || BloomFilterMatch(p.bloomFilter, env.Bloom())
 }
 
+func (p *Peer) topicInterestMatch(env *Envelope) bool {
+	p.topicInterestMu.Lock()
```
oskarth, line 310: What does this and the other similar locks protect against? Perhaps a comment here would be useful, especially in case there are other instances where the class of race conditions (or whatever) can occur

waku/waku.go
```diff
@@ -56,15 +55,18 @@ type Bridge interface {
...
-	RestrictLightClientsConn bool    // Restrict connection between two light clients
-	SyncAllowance            int     // Maximum time in seconds allowed to process the waku-related messages
+	MaxMsgSize               uint32             // Maximal message length allowed by the waku node
+	EnableConfirmations      bool               // Enable sending message confirmations
+	MinPow                   float64            // Minimal PoW required by the waku node
+	MinPowTolerance          float64            // Minimal PoW tolerated by the waku node for a limited time
+	BloomFilter              []byte             // Bloom filter for topics of interest for this node
+	BloomFilterTolerance     []byte             // Bloom filter tolerated by the waku node for a limited time
+	TopicInterest            map[TopicType]bool // Topic interest for this node
+	TopicInterestTolerance   map[TopicType]bool // Topic interest tolerated by the waku node for a limited time
```
oskarth, line 65: *see below*

waku/waku.go
```diff
@@ -56,15 +55,18 @@ type Bridge interface {
...
+	BloomFilterMode          bool               // Whether we should match against bloom-filter only
+	LightClient              bool               // Light client mode enabled does not forward messages
+	RestrictLightClientsConn bool               // Restrict connection between two light clients
+	SyncAllowance            int                // Maximum time in seconds allowed to process the waku-related messages
```
oskarth, line 69: Update: looked at where it is used, makes sense. Do you think should perhaps make its way to the spec? If nothing else as a recommendation.

oskarth:
Nice! Didn't go through Go code in great detail but looks reasonable, left some questions

waku/doc.go
```diff
@@ -50,13 +48,14 @@ const (
...
-	keyIDSize       = 32                     // in bytes
-	BloomFilterSize = 64                     // in bytes
-	flagsLength     = 1
+	TopicLength      = 4                      // in bytes
+	signatureLength  = crypto.SignatureLength // in bytes
+	aesKeyLength     = 32                     // in bytes
+	aesNonceLength   = 12                     // in bytes; for more info please see cipher.gcmStandardNonceSize & aesgcm.NonceSize()
+	keyIDSize        = 32                     // in bytes
+	BloomFilterSize  = 64                     // in bytes
+	MaxTopicInterest = 10000
```
kdeme, line 57: In the specification this is currently set to `1000`

waku/handshake.go
```diff
@@ -18,14 +18,61 @@ import (
...
+	if s.RateLimits != nil && s.RateLimits.IsZero() {
+		s.RateLimits = nil
+	}
+
+	return s
+}
+
+func (s statusOptions) WithDefaults() statusOptions {
+	if s.PoWRequirement == nil {
+		var pow uint64 = 0
```
kdeme: How about we set the default of the peer its PoW to the same value as the node its default? 
Or some other default value, but not 0?

waku/handshake.go
```diff
@@ -18,14 +18,61 @@ import (
...
+		s.LightNodeEnabled = &lightNodeEnabled
+	}
+
+	if s.ConfirmationsEnabled == nil {
+		confirmationsEnabled := false
+		s.ConfirmationsEnabled = &confirmationsEnabled
+	}
+
+	if s.RateLimits == nil {
+		s.RateLimits = &RateLimits{}
```
kdeme: What does this default result in?

waku/peer.go
cammellos: Basically `map`s (or slices) are not thread safe, so we need to acquire a lock before accessing the datastructure, as they might get accessed them from different threads, I will leave a comment

waku/waku.go
cammellos: yes, makes sense, although I take is up to each client to decide how strict they want to be

waku/peer.go
```diff
@@ -224,7 +258,7 @@ func (p *Peer) broadcast() error {
 	envelopes := p.host.Envelopes()
 	bundle := make([]*Envelope, 0, len(envelopes))
 	for _, envelope := range envelopes {
-		if !p.marked(envelope) && envelope.PoW() >= p.powRequirement && p.bloomMatch(envelope) {
+		if !p.marked(envelope) && envelope.PoW() >= p.powRequirement && p.topicOrBloomMatch(envelope) {
```
kdeme, line 252: Wait, do I understand correctly that it will apply both topic and bloom filter matching? It should only apply one or the other. Topics prioritised if they are set. Similar as I think is done in `topicInterestOrBloomMatch`.

waku/waku.go
kdeme: I wouldn't specify specific timeouts, it really depends on clients and there use cases. But I guess recommendations do no harm. We currently have something like this mentioned for timeout on the handshake

kdeme (changes requested):

waku/doc.go
cammellos: I think that's a bit tight for our use case, I can see how some busy clients might be listening on 500 topics, I thought we settled for 10k initially to be on the safe side

waku/handshake.go
cammellos: I believe it means no rate limits, but @adambabik can confirm

waku/doc.go
kdeme: ok, you know the usage of topics better than me, but lets not forget to update this in the spec then.

waku/config.go
```diff
@@ -22,6 +22,7 @@ package waku
 type Config struct {
 	MaxMessageSize           uint32  `toml:",omitempty"`
 	MinimumAcceptedPoW       float64 `toml:",omitempty"`
+	BloomFilterMode          bool    `toml:",omitempty"` // when true, we only match against topic interest
```
adambabik:
```suggestion
	BloomFilterMode          bool    `toml:",omitempty"` // when true, we match evenlopes against a bloom filter
```

waku/handshake.go
```diff
@@ -18,14 +18,61 @@ import (
...
-	LightNodeEnabled     bool        `rlp:"key=2"`
-	ConfirmationsEnabled bool        `rlp:"key=3"`
-	RateLimits           RateLimits  `rlp:"key=4"`
+	LightNodeEnabled     *bool       `rlp:"key=2"`
+	ConfirmationsEnabled *bool       `rlp:"key=3"`
+	RateLimits           *RateLimits `rlp:"key=4"`
 	TopicInterest        []TopicType `rlp:"key=5"`
 }
 
+func (s statusOptions) WithoutDefaults() statusOptions {
```
adambabik: This is a bit odd. Can't we create `statusOptions` without defaults in the first place?

waku/handshake.go
```diff
@@ -18,14 +18,61 @@ import (
...
+	}
+
+	if s.RateLimits != nil && s.RateLimits.IsZero() {
+		s.RateLimits = nil
+	}
+
+	return s
+}
+
+func (s statusOptions) WithDefaults() statusOptions {
```
adambabik: I think this is a bit confusing because a default state for this node is different from a default state for a peer. Maybe instead we could have two functions.

waku/peer.go
cammellos: I though it was doing that, as we don't set `fullNode` to true if `topicInterest` is set, but just looking at the function `BloomFilterMatch` is matches if bloom filter is nil, regardless on whether `fullNode` is set. I'll add a test to ensure the behavior is correct.

go.mod
```diff
@@ -42,15 +43,16 @@ require (
 	github.com/status-im/migrate/v4 v4.6.2-status.2
 	github.com/status-im/rendezvous v1.3.0
 	github.com/status-im/status-go/extkeys v1.1.0
-	github.com/status-im/status-go/waku v1.2.0
```
adambabik: Where did it go?

waku/handshake.go
```diff
@@ -48,24 +95,34 @@ var keyFieldIdx = func() map[string]int {
...
 	v := reflect.ValueOf(o)
-	optionsList := make([]interface{}, 0, v.NumField())
+	var optionsList []interface{}
 	for i := 0; i < v.NumField(); i++ {
-		value := v.Field(i).Interface()
-		key, ok := idxFieldKey[i]
-		if !ok {
-			continue
+		field := v.Field(i)
+		if !field.IsNil() {
```
adambabik, line 100:
```suggestion
		if field.CanInterface() {
```
Might be even safer as it will never panic.

waku/peer.go
```diff
@@ -94,17 +100,21 @@ func (p *Peer) handshake() error {
...
+			TopicInterest:        p.host.TopicInterest(),
+			RateLimits:           &rateLimits,
+			LightNodeEnabled:     &isLightNode,
+			ConfirmationsEnabled: &confirmationsEnabled,
 		}
-		opts.SetPoWRequirementFromF(p.host.MinPow())
-		errc <- p2p.SendItems(p.ws, statusCode, ProtocolVersion, opts)
+		pow := p.host.MinPow()
+		if pow != 0 {
+			opts.SetPoWRequirementFromF(pow)
```
adambabik: Shouldn't we examine other options as well? For example, if we use bloom mode, we shouldn't even set `topic-interest`, right?

waku/peer.go
```diff
@@ -277,6 +315,29 @@ func (p *Peer) bloomMatch(env *Envelope) bool {
...
+}
+
+func (p *Peer) topicOrBloomMatch(env *Envelope) bool {
+	p.topicInterestMu.Lock()
+	topicInterestMode := p.topicInterest != nil
+	p.topicInterestMu.Unlock()
+
+	if topicInterestMode {
+		return p.topicInterestMatch(env)
+	} else {
```
adambabik: This `else` can be omitted.

waku/peer.go
```diff
@@ -277,6 +315,29 @@ func (p *Peer) bloomMatch(env *Envelope) bool {
...
+	defer p.topicInterestMu.Unlock()
+
+	if p.topicInterest == nil {
+		return false
+	}
+
+	return p.fullNode || p.topicInterest[env.Topic]
+}
+
+func (p *Peer) topicOrBloomMatch(env *Envelope) bool {
```
adambabik, line 325: I would explain here or somewhere else how we treat `len(topicInterest) == 0` and zero bloom filter. It is not obvious looking at the code.

waku/waku.go
```diff
@@ -1203,6 +1275,62 @@ func (w *Waku) add(envelope *Envelope, isP2P bool) (bool, error) {
...
+	return true, nil
+}
+
+func (w *Waku) topicInterestOrBloomMatch(envelope *Envelope) (bool, error) {
+	w.settingsMu.RLock()
+	topicInterestMode := !w.settings.BloomFilterMode
+	w.settingsMu.RUnlock()
+
+	if topicInterestMode {
+		topicInterestMatch, err := w.topicInterestMatch(envelope)
```
adambabik:
```suggestion
		match, err := w.topicInterestMatch(envelope)
```

waku/waku.go
```diff
@@ -1203,6 +1275,62 @@ func (w *Waku) add(envelope *Envelope, isP2P bool) (bool, error) {
...
+		if err != nil {
+			return false, err
+		}
+		if topicInterestMatch {
```
adambabik:
```suggestion
		if match {
```

waku/waku.go
```diff
@@ -815,7 +899,19 @@ func (w *Waku) GetSymKey(id string) ([]byte, error) {
 func (w *Waku) Subscribe(f *Filter) (string, error) {
 	s, err := w.filters.Install(f)
 	if err == nil {
-		w.updateBloomFilter(f)
+		w.settingsMu.RLock()
```
adambabik: I would create a new method and have here only something like this:
```go
w.updateSettingsForFilter(f)
```

Is there a code where we update bloom filter/topic interest after unsubscribing from a filter?

waku/waku.go
```diff
@@ -256,6 +261,71 @@ func (w *Waku) SetBloomFilter(bloom []byte) error {
...
+
+	w.settingsMu.Lock()
+	w.settings.TopicInterest = topicInterestMap
+	// Setting topic interest resets bloom filter
+	w.settings.BloomFilter = nil
+	w.settingsMu.Unlock()
+	w.notifyPeersAboutTopicInterestChange(topicInterest)
+
+	go func() {
+		// allow some time before all the peers have processed the notification
```
adambabik, line 334: This should exit early if Waku is stopped. There is a channel to handle it.

waku/waku.go
```diff
@@ -256,6 +261,71 @@ func (w *Waku) SetBloomFilter(bloom []byte) error {
...
+		topicInterest[i] = topic
+		i++
+	}
+	return topicInterest
+}
+
+// updateTopicInterest adds a new topic interest
+// and informs the peers
+func (w *Waku) updateTopicInterest(f *Filter) error {
+
```
adambabik:
```suggestion
```

adambabik (commented):
One more general thing. We distinguish explicitly `fullNode`. Should we do the same with `emptyNode`? Currently, it's a bit hard to navigate how the logic behaves when we have an `emptyNode`.

waku/handshake.go
adambabik: Correct. I will start working on full rate limiting support and the idea is to verify it using `RateLimits.IsZero() == false`.

waku/handshake.go
cammellos: I think it looks like it does something slightly different, did you mean something like:
`if field.CanInterface() && !field.IsNil()` ? i.e is the field public and is not nil?

waku/peer.go
cammellos: currently that's not an issue, as `topicInterest`and bloom are never used in conjunction (setting one sets the other to `nil`), but we can be explicit about it

waku/handshake.go
cammellos: we can, it's useful for the testing, as it compares the size and so we need to strip out all the `zero` options, some of them are dynamically set, and it's a lot of `ifs` :) , in the actual code we use it only once, I would still encapsulate that logic in a function though, but we can create it without the defaults

cammellos:
> One more general thing. We distinguish explicitly fullNode. Should we do the same with emptyNode
I can try to see if it makes it better, I would not introduce another piece of state as that requires further state management, although checking for `emptyNode` is a bit redundant for example (just for topic-interest for simplicity):

`emptyNode() { return !m.fullNode && len(m.topicInterest) == 0}`
`matchTopic(topic) { return m.fullNode || !m.emptyNode() && m.topicInterest[topic]}`

So the check for `emptyNode` in this case is redundant, although more explicit, what do you think?

waku/handshake.go
cammellos: I would probably set it to a default rather than the host PoW, just to make it a bit simpler, we could use the current minpow used by the nodes? (can't remember exactly what it is, 0.0002 or something?)

waku/handshake.go
adambabik: Sorry, I misunderstood `CanInterface`. We only have exported fields here so we can do both but `!field.IsNil()` is good as well 👍

waku/handshake.go
adambabik: What we use is practically 0. Especially in terms of calculating it. We run some tests once and I think anything below 0.2 or 0.02 was close to 0. We use [`0.002`](https://github.com/status-im/status-go/blob/44aa31398107f2dcb224cf6e689896aed0079227/protocol/transport/message.go#L9).

waku/peer.go
adambabik: Right but when I look at this I would wonder: "I use bloom mode but in the code I still propagate `topic-interest`. Why is it so? Shouldn't I only send my bloom filter" :) 

Also, when receiving options we do `peerOptions.TopicInterest != nil`. But if we always send `topic-interest`, after RLP decoding it won't be `nil`, it will be an empty slice so we never reach `else if peerOptions.BloomFilter != nil `. Did I miss anything?

waku/handshake.go
kdeme: > I would probably set it to a default rather than the host PoW, just to make it a bit simpler, we could use the current minpow used by the nodes? (can't remember exactly what it is, 0.0002 or something?)

That's what I meant.

waku/waku.go
```diff
@@ -1410,6 +1556,31 @@ func (w *Waku) expire() {
 	}
 }
 
+func (w *Waku) ToStatusOptions() statusOptions {
```
adambabik: Do we need to export it?

adambabik:
@cammellos re the empty node. I think it is fine if we go implicit but somewhere in the code (close to `fullNode` prop or in the setters) we should add a comment explaining how an empty node situation is handled and what "empty node" means. Especially, saying that it is a perfectly valid scenario.

cammellos:
@adambabik @kdeme all the feedback should be addressed, could you please review again? thanks!

kdeme (approved):

adambabik (approved):
LGTM. Please check why the CI fails. Also, let's make sure `waku/` is tested in the CI pipeline. 

`whisper/` after removal of `go.mod` might also start to be tested but as far as I remember the tests were not passing after forking it from go-ethereum. I will take a look.

waku/waku_test.go
```diff
@@ -278,7 +278,6 @@ func TestSymKeyManagement(t *testing.T) {
 	var err error
 	var k1, k2 []byte
 	w := New(nil, nil)
-	id1 := string("arbitrary-string-1")
```
adambabik, line 281: This is causing tests to fail. 
```
[2020-02-20T11:54:35.582Z] # github.com/status-im/status-go/waku [github.com/status-im/status-go/waku.test]
[2020-02-20T11:54:35.582Z] waku/waku_test.go:283:2: undefined: id1
[2020-02-20T11:54:35.582Z] waku/waku_test.go:288:24: undefined: id1
```

They pass on develop for waku. The tests for Whisper fail on develop (confirmed) and I will fix them.

PR #1863, opened 2020-02-18
Add mark as read endpoint

cammellos:
add a mark as read endpoint, and removes some unused endpoints


protocol/messenger.go
```diff
@@ -2047,7 +2022,23 @@ func (m *Messenger) MarkMessagesSeen(chatID string, ids []string) error {
...
-// DEPRECATED: required by status-react.
+func (m *Messenger) MarkAllRead(chatID string) error {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+	chat, ok := m.allChats[chatID]
+	if !ok {
+		return errors.New("chat not found")
+	}
+
+	err := m.persistence.MarkAllRead(chatID)
```
adambabik, line 2033: Marking all as read and saving the chat is not an atomic operation. Is this ok?

cammellos: Worth making it atomic I guess, will change that

protocol/messenger.go
```diff
@@ -1658,21 +1658,6 @@ func (m *Messenger) syncContact(ctx context.Context, contact *Contact) error {
 	return m.saveChat(chat)
 }
 
-// SendRaw takes encoded data, encrypts it and sends through the wire.
-// DEPRECATED
-func (m *Messenger) SendRaw(ctx context.Context, chat Chat, data []byte) ([]byte, error) {
```
jakubgs, line 1663: Why are we removing this?

services/shhext/api_nimbus.go
```diff
@@ -265,30 +265,6 @@ func (api *NimbusPublicAPI) Post(ctx context.Context, newMessage types.NewMessag
...
-		Name: msg.Chat,
-	}
-	return api.service.messenger.SendRaw(ctx, chat, msg.Payload)
-}
-
-// SendDirectMessage sends a 1:1 chat message to the underlying transport
-// Message's payload is a transit encoded message.
-// It's important to call NimbusPublicAPI.afterSend() so that the client receives a signal
-// with confirmation that the message left the device.
-func (api *NimbusPublicAPI) SendDirectMessage(ctx context.Context, msg SendDirectMessageRPC) (types.HexBytes, error) {
```
jakubgs, line 283: Why are we removing this?

services/ext/api.go
```diff
@@ -194,30 +194,6 @@ func (api *PublicAPI) ConfirmMessagesProcessedByID(messageConfirmations []*Metad
...
-		Name: msg.Chat,
-	}
-	return api.service.messenger.SendRaw(ctx, chat, msg.Payload)
-}
-
-// SendDirectMessage sends a 1:1 chat message to the underlying transport
-// Message's payload is a transit encoded message.
-// It's important to call PublicAPI.afterSend() so that the client receives a signal
-// with confirmation that the message left the device.
-func (api *PublicAPI) SendDirectMessage(ctx context.Context, msg SendDirectMessageRPC) (types.HexBytes, error) {
```
jakubgs, line 212: Why are we removing this?

protocol/messenger.go
```diff
@@ -2000,37 +1982,30 @@ func (m *Messenger) ConfirmMessagesProcessed(messageIDs [][]byte) error {
 	return nil
 }
 
-// DEPRECATED: required by status-react.
 func (m *Messenger) MessageByID(id string) (*Message, error) {
```
jakubgs, line 1985: Why are they not deprecated anymore?

protocol/messenger.go
cammellos: Not used anymore

services/shhext/api_nimbus.go
cammellos: Not used anymore

services/ext/api.go
cammellos: Not used anymore

protocol/messenger.go
cammellos: I think the idea is was that we will build a new API from scratch with non-deprecated methods, after a discussion we agreed on not marking endpoint as deprecated until we have a suitable replacement or there are no more use cases for it. Currently this endpoints are used by status-react. (also before we had two tables for messages, now there's only one)

jakubgs: Makes sense.

jakubgs (approved):

adambabik (approved):

PR #1866, opened 2020-02-20
make apis private

cammellos:


adambabik (commented):
@cammellos looking at this fragment:
```go
if whitelist[api.Namespace] || (len(whitelist) == 0 && api.Public) {
```
it means that we have `shhext` and `wakuext` in `HTTPModules`. Is that intentional?

cammellos:
@adambabik 

No idea, I hardly understand what it means ;)
>  shhext and wakuext in HTTPModules.
If that means that they are accessible through HTTP (therefore any website we visit with the browser can access them), I would say that's not desirable, all the endpoints in `shhext` and `wakuext` should not be accessible by anyone but status-react, the same goes for `shh` I think, currently I see no reason to expose it

adambabik:
@cammellos 

> I would say that's not desirable, all the endpoints in shhext and wakuext should not be accessible by anyone but status-react, the same goes for shh I think, currently I see no reason to expose it

Yeah, that's my reasoning as well. So I would like to get into the bottom of things why these services are exposed.

We expose `GenerateConfig` binding and it calls `NewNodeConfig` which provides:
```
APIModules:       "eth,net,web3,peer,wallet",
```
There is no `shh` or `shhext` but there is `web3` and I don't know how is this handled on status-react side.

Also, can we examine whether `APIModules` is specified by status-react in the config? (`HTTPModules` is derived from `APIModules` which is a string and `HTTPModules` is simply a slice of `APIModules`).

whisper/message_test.go
```diff
@@ -351,7 +351,7 @@ func TestRlpEncode(t *testing.T) {
 	}
 
 	var decoded Envelope
-	rlp.DecodeBytes(raw, &decoded)
+	rlp.DecodeBytes(raw, &decoded) // nolint: errcheck
```
adambabik: We can disable `errcheck` for `./whisper`. Check out https://github.com/status-im/status-go/blob/develop/.golangci.yml#L76-L79.

It will be something like:
```
- path: whisper/.*_test.go
  linters:
    - errcheck
    - gosec
```

adambabik (approved):

adambabik (changes requested):
One more thing. I noticed you removed `go.{mod,sum}` from `whisper`. It means we need to change imports everywhere from `github.com/status-im/status-go/whisper/v6` to `github.com/status-im/status-go/whisper` and eventually get rid of `github.com/status-im/status-go/whisper/v6 v6.2.1` from the main `go.mo`.

cammellos:
@adambabik reverted have whisper separated for now, easier for the hottfix, ready to be reviewed again

adambabik (approved):

PR #1875, opened 2020-02-25
Add example to demonstrate Whisper-Waku bridge

adambabik:


jakubgs (approved):
I like how this shows usage of RPC to do all of those things. Neat.

PR #1878, opened 2020-02-28
Feature/upgrade extkeys

cammellos:
upgrade extkeys to go 1.13


go.mod
```diff
@@ -42,7 +42,7 @@ require (
 	github.com/status-im/markdown v0.0.0-20200210164614-b9fe92168122
 	github.com/status-im/migrate/v4 v4.6.2-status.2
 	github.com/status-im/rendezvous v1.3.0
-	github.com/status-im/status-go/extkeys v1.1.0
+	github.com/status-im/status-go/extkeys v1.1.2
```
adambabik, line 45: You created this tag from a commit that is not in master yet, right?

cammellos: yes, I don't know what's the best, as I still need to test it, so I need to push & tag, if I rebase I can keep, otherwise I will discard and push an updated one

adambabik: Not sure how you test it but for the time you test a PR, you can add `replace` directive in `go.mod` like `replace github.com/status-im/status-go/extkeys => ./extkeys`. And when you done, you remove it, merge a PR and bump the version in the very next PR.

PR #1880, opened 2020-02-28
Stop fetching new blocks while app is not active

rasom:
Saves some traffic (blocks with transactions are heavy) and infura
requests.

status: ready


services/wallet/commands.go
```diff
@@ -292,6 +293,10 @@ func (c *newBlocksTransfersCommand) checkDelay(parent context.Context, nextHeade
 }
 
 func (c *newBlocksTransfersCommand) Run(parent context.Context) (err error) {
+	if !c.service.IsAppActive() {
```
andremedeiros: Will `c.from` be the same on the next invocation?

rasom: yes sure, it is like last known block here. The idea is that we do not touch the way how this command recovers after the long delay before fetching next block, we just "create" this delay

api/geth_backend.go
```diff
@@ -783,6 +783,14 @@ func (b *GethStatusBackend) AppStateChange(state string) {
 	b.log.Info("App State changed", "new-state", s)
 	b.appState = s
 
+	wallet, err := b.statusNode.WalletService()
+	if err != nil {
+		log.Error("Can't get wallet service", "error", err)
+		return
+	}
+
+	wallet.SetAppActive(s == appStateForeground)
```
adambabik: I have a different approach -- couldn't we start and stop wallet service when the app state changes? That would prevent from spreading `Service` between all structs in `services/wallet`.

rasom: I'm just not sure how to do this so that it properly recovers after restart

rasom: let me try

adambabik: It depends how well `wallet.Service` is built. In theory, it has `Start()` and `Stop()` so it should work out of the box:

```go
if (s == appStateForeground) {
  err := wallet.Start(b.GethNode().Server())
  // check for err
  wallet.StartReactor(...) // this is missing from Start and it's fine if it needs to be separate
} else if (s == appStateBackground) {
  wallet.Stop()
}
```

But we need an integration test (in `service_test.go`) to verify it. Immediate thing I noticed is that `Start` does not do `StartReactor` so that would need to be called manually.

rasom: @adambabik does it look better now?

api/geth_backend.go
```diff
@@ -783,6 +783,30 @@ func (b *GethStatusBackend) AppStateChange(state string) {
 	b.log.Info("App State changed", "new-state", s)
 	b.appState = s
 
+	if s == appStateForeground {
+		wallet, err := b.statusNode.WalletService()
+		if err != nil {
+			return
```
adambabik, line 790: We should log this with `log.Error` because it's a big deal. I am even wondering whether we should bubble this error to status-react through signals.

api/geth_backend.go
```diff
@@ -783,6 +783,30 @@ func (b *GethStatusBackend) AppStateChange(state string) {
...
+		}
+		err = wallet.Start(b.statusNode.Server())
+		if err != nil {
+			return
```
adambabik: Same here.

api/geth_backend.go
```diff
@@ -783,6 +783,30 @@ func (b *GethStatusBackend) AppStateChange(state string) {
...
+		}
+
+		err = b.startWallet()
+		if err != nil {
+			return
```
adambabik: And here.

adambabik (commented):
@rasom yeah, looks great! We just need error handling.

adambabik (commented):
We should also make sure a sequence of calls `Wallet.Start/Stop` works properly.

rasom:
> We should also make sure a sequence of calls Wallet.Start/Stop works properly.

@adambabik what's expected? How do we check that?

adambabik:
@rasom It's simple. Write a new test in https://github.com/status-im/status-go/blob/develop/services/wallet/service_test.go which has the following sequence of calls:
```go
func TestServiceStartStop(t *testing.T) {
  s := NewService(...)
  require.NoError(t, s.Start())

  // act on it, i.e. make it do some job and verify the output
  
  require.NoError(t, s.Stop())
  require.NoError(t, s.Start())

  // act on it again, make sure the output is the same as above
}
```

I can add such a unit test if it's something tricky.

rasom:
@adambabik added test, is that enough?

adambabik (approved):

andremedeiros (approved):

PR #1893, opened 2020-03-05
do not return private key in the result for VerifyAccountPassword

### LINKED ISSUE #1882: do not return private key in the result for VerifyAccountPassword
do not return private key in the result for VerifyAccountPassword

0kok0:
Add readKey parameter to manager to return key in VerifyAccountPassword iff explicitly set by the caller. Default behaviour is only verifying the passphrase.

Closes #1882

### DIFF
--- a/api/geth_backend.go
+++ b/api/geth_backend.go
@@ -683,6 +683,29 @@ func (b *GethStatusBackend) HashTypedData(typed typeddata.TypedData) (types.Hash
 	return types.Hash(hash), err
 }
 
+func (b *GethStatusBackend) VerifyWalletAccount(address, password string) error {
+	config := b.StatusNode().Config()
+
+	db := accounts.NewDB(b.appDB)
+	exists, err := db.AddressExists(types.HexToAddress(address))
+	if err != nil {
+		b.log.Error("failed to query db for a given address", "address", address, "error", err)
+		return err
+	}
+
+	if !exists {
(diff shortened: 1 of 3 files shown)


api/geth_backend.go
```diff
@@ -697,7 +697,7 @@ func (b *GethStatusBackend) getVerifiedWalletAccount(address, password string) (
 		b.log.Error("failed to get a selected account", "err", transactions.ErrInvalidTxSender)
 		return nil, transactions.ErrAccountDoesntExist
 	}
-
+    b.accountManager.ReadKey = true
```
adambabik: I guess whether to read a key or not should come from a caller so a caller of `getVerifiedWalletAccount`, right? Eventually, we want the mobile app to decide whether to return a key or now.

I would change the interface of `accountManager` to be explicit:
```go
func (b *GethStatusBackend) verifyWalletAccount(address, password string) error {
    // ...
    return b.accountManager.VerifyAccount(config.KeyStoreDir, address, password)
}

func (b *GethStatusBackend) verifyAndReturnWalletAccount(address, password string) (*account.SelectedExtKey, error) {
    // ...
    key, err := b.accountManager.VerifyAccountAndReturnKey(config.KeyStoreDir, address, password)
    // ...
}
```

0kok0: > from a caller so a caller of getVerifiedWalletAccount, right?

Ok sure, will make the change

flexsurfer:
thanks @0kok0 and sorry for the confusion, there is no private key in results

PR #1897, opened 2020-03-09
Fix listen flag

adambabik:


cmd/statusd/main.go
```diff
@@ -32,10 +32,6 @@ const (
 	serverClientName = "Statusd"
 )
 
-var (
-	buildStamp = "N/A" // rely on linker: -ldflags -X main.buildStamp"
```
jakubgs, line 36: Why exactly are we dropping this?

cmd/statusd/main.go
```diff
@@ -65,7 +61,7 @@ var (
 			[]string{params.FleetProd, params.FleetStaging, params.FleetTest}, params.FleetProd,
 		),
 	)
-	listenAddr = flag.String("addr", ":30303", "address to bind listener to")
+	listenAddr = flag.String("addr", "", "address to bind listener to")
```
jakubgs, line 64: What was wrong about providing a default value of `":30303"`?

cmd/statusd/main.go
adambabik: It's unused. I have never examined this setting and it's preventing from caching more layers when building a docker image.

jakubgs (approved):

PR #1906, opened 2020-03-17
add README for metrics

jakubgs:
Because it wasn't clear how we collect metrics about connected Status peers.


metrics/README.md
```diff
@@ -0,0 +1,44 @@
...
+* `promhttp.HandlerFor()` - Our own custom metrics from this package.
+* `gethprom.Handler(reg)` - Geth metrics defined in [`metrics`](https://github.com/ethereum/go-ethereum/tree/master/metrics)
+
+By calling both we can extend existing metrics.
+
+# Metrics
+
+We add a few extra metrics on top of the normal Geth ones in [`node/metrics.go`](./node/metrics.go):
+
+* `p2p_peers_count` - Current numbers of peers split by name.
```
adambabik, line 20: Some of these metrics are redundant, for example `p2p_peers_count`, `p2p_peers_absolute`, and `p2p_peers_changes` carry the same information but provide different labels.

metrics/README.md
```diff
@@ -0,0 +1,44 @@
...
+* `p2p_peers_changes` - Current changes to connected peers.
+* `p2p_peers_absolute` - Absolute number of connected peers.
+* `p2p_peers_max` - Maximum number of peers that can connect.
+
+The `p2p_peers_count` metrics includes 3 labels:
+
+* `type` - Set to `StatusIM` for mobile and `Statusd` for demon.
```
adambabik:
```suggestion
* `type` - Set to `StatusIM` for mobile and `Statusd` for daemon.
```

adambabik (approved):

metrics/README.md
jakubgs: Yeah, that's because I was copying legacy stuff. Do you think the `absolute` and `changes` metrics are useless? If so we can get rid of them and the `computeMetrics()` method.

adambabik: I think we can get rid of `_changes`. `_absolute` has this advantage over `_count` that it counts everything while `_count` parses names and skips peers. So only `_changes` seem to be really obsolete.

jakubgs: Good point. I'll drop it.

adambabik (approved):

PR #1908, opened 2020-03-20
Ignore full node flag when in topic mode

cammellos:
When in topicInterest mode the flag `fullNode` should be ignored.


adambabik (approved):

adambabik:
For the record:
1. Waku peer is a full node by default: https://github.com/status-im/status-go/blob/v0.49.1/waku/peer.go#L84,
2. `fullNode` field is changed only in [`setBloomFilter`](https://github.com/status-im/status-go/blob/v0.49.1/waku/peer.go#L340) method,
3. `setBloomFilter` method is called only if [bloom filter is set in `statusOptions`](https://github.com/status-im/status-go/blob/v0.49.1/waku/peer.go#L169-L177).

To avoid other bugs in situations where the code relies on the `fullNode` field, I suggest setting `fullNode = false` always in the `setTopicInterest` method.

adambabik (approved):

PR #1909, opened 2020-03-22
Clarify situation between lib and mobile

adambabik:


cammellos (approved):

oskarth (approved):
Out of curiosity, if library is deprecated, how come it is still in the code base?

adambabik:
@oskarth it has been used by the Status Desktop when it was in active development. If this project is completely ditched (I heard the opposite) then we can remove it. If we start working on it again, it needs to be reviewed and aligned with `mobile/`.

PR #1914, opened 2020-03-24
Add change group chat name endpoint

cammellos:
Add an endpoint to change group chat name and allow leaving without removing the chat.


protocol/messenger.go
```diff
@@ -823,6 +823,70 @@ func (m *Messenger) AddMembersToGroupChat(ctx context.Context, chatID string, me
 	return &response, m.saveChat(chat)
 }
 
+func (m *Messenger) ChangeGroupChatName(ctx context.Context, chatID string, name string) (*MessengerResponse, error) {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
```
adambabik, line 830: The lock's scope is too large. For example, `event.Sign` can take significant time and it does not make sense to block for that long any other operations.

I see it's in other methods like this. We can keep it for consistency and fix in a different PR.

protocol/messenger.go
```diff
@@ -823,6 +823,70 @@ func (m *Messenger) AddMembersToGroupChat(ctx context.Context, chatID string, me
...
+
+	var response MessengerResponse
```
adambabik: We can move it down to where it is actually used.

protocol/messenger.go
```diff
@@ -823,6 +823,70 @@ func (m *Messenger) AddMembersToGroupChat(ctx context.Context, chatID string, me
...
+	logger := m.logger.With(zap.String("site", "ChangeGroupChatName"))
```
adambabik: We can move it to the top of the method in order not to clutter the logic.

protocol/messenger.go
```diff
@@ -823,6 +823,70 @@ func (m *Messenger) AddMembersToGroupChat(ctx context.Context, chatID string, me
...
+	logger.Info("Changing group chat name", zap.String("chatID", chatID), zap.String("name", name))
+	chat, ok := m.allChats[chatID]
+	if !ok {
+		return nil, errors.New("can't find chat")
```
adambabik: `errors.New("can't find chat")` can be defined as a variable.

protocol/messenger.go
```diff
@@ -823,6 +823,70 @@ func (m *Messenger) AddMembersToGroupChat(ctx context.Context, chatID string, me
 	return &response, m.saveChat(chat)
 }
 
+func (m *Messenger) ChangeGroupChatName(ctx context.Context, chatID string, name string) (*MessengerResponse, error) {
```
adambabik, line 825: No unit test?

adambabik (commented):
> and allow leaving without removing the chat.

What will be the consequences?

cammellos:
@adambabik addressed feedback
> What will be the consequences?

None, the user will still be able to read old messages, won't receive new ones

adambabik (approved):

flexsurfer (approved):

protocol/messenger.go
```diff
@@ -821,6 +822,71 @@ func (m *Messenger) AddMembersToGroupChat(ctx context.Context, chatID string, me
...
+	}
+
+	encodedMessage, err := m.processor.EncodeMembershipUpdate(group, nil)
+	if err != nil {
+		return nil, err
+	}
+	_, err = m.dispatchMessage(ctx, &RawMessage{
+		LocalChatID:         chat.ID,
+		Payload:             encodedMessage,
+		MessageType:         protobuf.ApplicationMetadataMessage_MEMBERSHIP_UPDATE_MESSAGE,
```
jakubgs, line 869: Maybe renaming this event to something like `GROUP_STATE_CHANGE` would make it clearer as to what it does.

protocol/messenger.go
```diff
@@ -821,6 +822,71 @@ func (m *Messenger) AddMembersToGroupChat(ctx context.Context, chatID string, me
...
+	clock, _ := chat.NextClockAndTimestamp(m.getTimesource())
+	// Add members
+	event := v1protocol.NewNameChangedEvent(name, clock)
+	event.ChatID = chat.ID
+	err = event.Sign(m.identity)
+	if err != nil {
+		return nil, err
+	}
+
+	err = group.ProcessEvent(event)
```
jakubgs, line 852: A comment saying something like `// Update in-memory stuctures` would be nice.

protocol/messenger.go
```diff
@@ -821,6 +822,71 @@ func (m *Messenger) AddMembersToGroupChat(ctx context.Context, chatID string, me
...
+	if err != nil {
+		return nil, err
+	}
+
+	recipients, err := stringSliceToPublicKeys(group.Members(), true)
+	if err != nil {
+		return nil, err
+	}
+
+	encodedMessage, err := m.processor.EncodeMembershipUpdate(group, nil)
```
jakubgs, line 862: A comment explaining that this encodes the whole state of thee group to be sent, not just name change.

protocol/messenger.go
```diff
@@ -953,15 +1019,15 @@ func (m *Messenger) ConfirmJoiningGroup(ctx context.Context, chatID string) (*Me
 	return &response, m.saveChat(chat)
 }
 
-func (m *Messenger) LeaveGroupChat(ctx context.Context, chatID string) (*MessengerResponse, error) {
+func (m *Messenger) LeaveGroupChat(ctx context.Context, chatID string, remove bool) (*MessengerResponse, error) {
```
jakubgs, line 1023: This change should probably be a separate commit/pr.

protocol/messenger_test.go
```diff
@@ -874,6 +874,67 @@ func (s *MessengerSuite) TestRetrieveTheirPrivateGroupChat() {
 	s.Require().NotNil(actualChat.LastMessage)
 }
 
+// Test receiving a message on an existing private group chat
+func (s *MessengerSuite) TestChangeNameGroupChat() {
+	var response *MessengerResponse
+	theirMessenger := s.newMessenger(s.shh)
+	response, err := s.m.CreateGroupChatWithMembers(context.Background(), "id", []string{})
```
jakubgs: Maybe rename `"id"` to `"old-name"` to make it clear that second argument is the group chat name.

protocol/messenger_test.go
```diff
@@ -874,6 +874,67 @@ func (s *MessengerSuite) TestRetrieveTheirPrivateGroupChat() {
...
+	// Retrieve their messages so that the chat is created
+	err = tt.RetryWithBackOff(func() error {
+		var err error
+		response, err = theirMessenger.RetrieveAll()
+		if err == nil && len(response.Chats) == 0 {
+			err = errors.New("chat invitation not received")
+		}
+		return err
+	})
+	s.Require().NoError(err)
```
jakubgs, line 903: What's the difference between `s.NoError(err)` and `s.Require().NoError(err)`?

protocol/messenger_test.go
```diff
@@ -874,6 +874,67 @@ func (s *MessengerSuite) TestRetrieveTheirPrivateGroupChat() {
...
+
+	_, err = theirMessenger.ConfirmJoiningGroup(context.Background(), ourChat.ID)
+	s.NoError(err)
+
+	err = tt.RetryWithBackOff(func() error {
```
jakubgs, line 909: Add a comment as above like `// Retrieve messages to verify group was joined`. Maybe this could be a helper method called `retrieveOrError()` or something like that. Since we do it three times.

protocol/messenger_test.go
cammellos: `s.NoError` will continue the test, while `require().NoError()` will stop execution, which is what probably you want most of the times

PR #1916, opened 2020-03-24
Update eth-node README

adambabik:


oskarth (approved):
Cheers!

oskarth:
Does that mean that go-ethereum is completely decoupled or is it still intertwined in some way?

Is there or have there been any plans for a mocking service, i.e. neither geth nor nimbus but an empty implementation for e.g. testing?

adambabik:
> Does that mean that go-ethereum is completely decoupled or is it still intertwined in some way?

I think there might be still some places where this needs to be done. @andremedeiros do you remember what Pedro achieved finally with regards to running status-go with Nimbus? Was it possible to send messages or only connect to peers?

> Is there or have there been any plans for a mocking service, i.e. neither geth nor nimbus but an empty implementation for e.g. testing?

Our integration tests were also altered to use this package so in theory they can also be duplicated and run with geth and nimbus. However, it's not status-go's responsibility to verify geth and nimbus are compatible. Nimbus in status-go requires an adapter anyway (because originally everything was tailored to geth) and this adapter can be tested to match geth behaviour.

Regarding mocking services, I don't see much value in this. Firstly, running nodes in memory is cheap. Secondly, if we want to cover some more complex business logic with unit tests, it's better and easier to write this logic in a way where it is independent from a node at all.

PR #1921, opened 2020-03-30
Explain how to run Waku node in README

adambabik:


_examples/README.md
```diff
@@ -2,6 +2,31 @@
...
+  "HTTPEnabled": true,
+  "HTTPHost": "localhost",
+  "HTTPPort": "8545",
+  "WakuConfig": {
+    "Enabled": true
+  }
+}
+```
+
+This command will start a Waku node using the `eth.prod` fleet:
```
oskarth, line 20: Is this transparent to non Status people? What alternatives are there to running with `eth.prod` fleet? E.g. adding a node like in geth

oskarth (commented):
Generally looks good to me, one question/suggestion

_examples/README.md
adambabik: Currently, there is no way to run status-go without selecting a fleet. Of course, if one wants to use their own nodes, it's possible by providing a proper config overriding properties defined in, for example, https://github.com/status-im/status-go/blob/develop/config/cli/fleet-eth.prod.json.

_examples/README.md
```diff
@@ -2,6 +2,31 @@
...
 
+## Run Waku node
+
+Running Waku node is a matter of a correct configuration. To enable Waku and JSON-RPC HTTP interface use:
+```shell script
+{
+  "APIModules": "waku",
+  "HTTPEnabled": true,
+  "HTTPHost": "localhost",
+  "HTTPPort": "8545",
```
kdeme: Can't be a string

```suggestion
  "HTTPPort": 8545,
```

_examples/README.md
kdeme, line 20: The default waku.json there does not seem to open HTTP RPC endpoint. Only ipc. Looks like that only works if you also provide HTTPHost.
Perhaps make it clear that you should use the new config and not the one of the examples.

kdeme (changes requested):
I can't mark it in the diff as it was not changed, but line 72 should hold the symKeyID that matches the one of the result when generating from password.

Thus change `98999c238e3747b7562674a86d450d531eca616d288a500268878e90848bfe4e` with `1e07adfcb80c9e9853fb2c4cce3d91c17edd17ab6e950387833d64878fe91624`.

_examples/README.md
adambabik: > Running Waku node is a matter of a correct configuration. To enable Waku and JSON-RPC HTTP interface use

I think this is pretty clear in the context of this README.

But I can also update `*.json` files to enable HTTP.

adambabik: Done

adambabik:
@kdeme fixed

kdeme (commented):
OK, looks good. But my remark of the wrong symKeyID on line 72 still stands

adambabik:
@kdeme sorry, I forgot about that comment. Fixed.

PR #1926, opened 2020-04-02
Fix setup-dev target

adambabik:
Relevant changes in the documentation: https://github.com/status-im/status.im/pull/492

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,7 +163,7 @@ endif
 install-os-dependencies:
 	_assets/scripts/install_deps.sh
 
-setup-dev: setup-build mock-install install-os-dependencies gen-install ##@other Prepare project for development
+setup-dev: lint-install mock-install modvendor-install gen-install tidy install-os-dependencies ##@other Prepare project for development
 
 setup-build: lint-install release-install gomobile-install ##@other Prepare project for build
 
@@ -239,7 +239,8 @@ modvendor-install:
 	GO111MODULE=off go get -u github.com/adambabik/modvendor
 
 mock-install: ##@other Install mocking tools
-	go get github.com/golang/mock/mockgen@latest
+	# keep in sync with go.mod and github.com/golang/mock
+	go get github.com/golang/mock/mockgen@v1.4.1
 
 mock: ##@other Regenerate mocks
 	mockgen -package=fake         -destination=transactions/fake/mock.go             -source=transactions/fake/txservice.go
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -163,7 +163,7 @@ endif
 install-os-dependencies:
 	_assets/scripts/install_deps.sh
 
-setup-dev: setup-build mock-install install-os-dependencies gen-install ##@other Prepare project for development
+setup-dev: lint-install mock-install modvendor-install gen-install tidy install-os-dependencies ##@other Prepare project for development
 
 setup-build: lint-install release-install gomobile-install ##@other Prepare project for build
```
jakubgs, line 169: Is this even necessary anymore?

jakubgs (changes requested):
Fails for me:
```
 > make setup-dev
curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | BINARY=golangci-lint bash -s -- -d -b /bin v1.21.0
golangci/golangci-lint info checking GitHub for tag 'v1.21.0'
golangci/golangci-lint debug http_download https://github.com/golangci/golangci-lint/releases/v1.21.0
golangci/golangci-lint info found version: 1.21.0 for v1.21.0/linux/amd64
golangci/golangci-lint debug downloading files into /tmp/tmp.huS0uoGPHX
golangci/golangci-lint debug http_download https://github.com/golangci/golangci-lint/releases/download/v1.21.0/golangci-lint-1.21.0-linux-amd64.tar.gz
golangci/golangci-lint debug http_download https://github.com/golangci/golangci-lint/releases/download/v1.21.0/golangci-lint-1.21.0-checksums.txt
install: cannot create regular file '/bin/golangci-lint': Permission denied
make: *** [Makefile:290: lint-install] Error 1
```

Makefile
adambabik: Jenkins still uses it.

adambabik:
@jakubgs hm so `GOPATH` is actually required still. In my case, I have `GOPATH` and it writes to `/Users/adamb/go/bin/golangci-lint`. In your case, you don't have it and it tries to write it to `/bin/golangci-lint`. Should we bring requirement of having `GOPATH` in the docs?

jakubgs:
Adding another step in docs is ugly. Don't newer versions of Go just default to `~/go`?
```
 > go env | grep GOPATH
GOPATH="/home/sochan/go"
```
They do. Then just do `GOPATH ?= $(HOME)/go` somewhere at the top of the `Makefile` so it defaults to that if `GOPATH` is not set(that's what `?=` does).

jakubgs:
Btw, please fix the shebang in `_assets/scripts/install_deps.sh`, it should be:
```
#!/usr/bin/env bash
```

adambabik:
@jakubgs updated

jakubgs (approved):
LGTM!

PR #1928, opened 2020-04-06
Return count of updated messages in MarkMessagesSeen

cammellos:
The frontend needs to know if the messaged marked as seen was actually
seen or not, as we might not have it loaded in the frontend and we need to update the counter client side.


protocol/persistence_legacy.go
```diff
@@ -479,7 +479,13 @@ func (db sqlitePersistence) MarkMessagesSeen(chatID string, ids []string) error
 	q := "UPDATE user_messages SET seen = 1 WHERE id IN (" + inVector + ")" // nolint: gosec
 	_, err = tx.Exec(q, idsArgs...)
 	if err != nil {
-		return err
+		return 0, err
+	}
+
+	var count uint64
+	row := tx.QueryRow("SELECT changes();")
```
adambabik, line 486: That's cool!

protocol/messenger.go
```diff
@@ -1999,20 +1999,20 @@ func (m *Messenger) DeleteMessagesByChatID(id string) error {
 	return m.persistence.DeleteMessagesByChatID(id)
 }
 
-func (m *Messenger) MarkMessagesSeen(chatID string, ids []string) error {
+func (m *Messenger) MarkMessagesSeen(chatID string, ids []string) (uint64, error) {
```
adambabik, line 2005:
```suggestion
// MarkMessagesSeen marks messages with `ids` as seen in the chat `chatID`.
// It returns the number of affected messages or error. If there is an error,
// the number of affected messages is always zero.
func (m *Messenger) MarkMessagesSeen(chatID string, ids []string) (uint64, error) {
```

adambabik (approved):

PR #1930, opened 2020-04-07
set archived_envelopes_total on start based on DB count

### LINKED ISSUE #1923: Implement metrics for currenly stored envelopes
# Problem

In relation to recent [postmortem](https://github.com/status-im/infra-docs/blob/master/postmortems/20200327_mailserver_db_issues.md) of DB issues we need metrics for currently stored envelopes in the DB. Current metrics cover only envelopes archived since the process started.

# Implementation

We'll have to periodically query the DB for a count of all envelopes. Pretty simple.

# Acceptance Criteria

I can track currently stored envelopes in Grafana.

jakubgs:
This change adds an `envelopesCount()` method that will be called in constructors for both DBs in order to set the initial value of `archivedEnvelopesGauge` to that of the count. Those are `NewLevelDB()` and `NewPostgresDB()` respectively.

It also changes the archive metrics to be labeled with `db` set to either DB name(Postgres) or the path(LevelDB). This way we can distinguish between `waku` and `whisper` metrics.

Resolves: #1923

### DIFF
--- a/mailserver/mailserver_db.go
+++ b/mailserver/mailserver_db.go
@@ -6,6 +6,9 @@ import (
 	"github.com/status-im/status-go/eth-node/types"
 )
 
+// every this many seconds check real envelopes count
+const envelopeCountCheckInterval = 60
+
 // DB is an interface to abstract interactions with the db so that the mailserver
 // is agnostic to the underlying technology used
 type DB interface {
--- a/mailserver/mailserver_db_leveldb.go
+++ b/mailserver/mailserver_db_leveldb.go
@@ -18,7 +18,9 @@ import (
 
 type LevelDB struct {
 	// We can't embed as there are some state problems with go-routines
-	ldb *leveldb.DB
+	ldb  *leveldb.DB
+	name string
+	done chan struct{}
 }
 
 type LevelDBIterator struct {
@@ -68,7 +70,41 @@ func NewLevelDB(dataDir string) (*LevelDB, error) {
 		log.Info("database is corrupted trying to recover", "path", dataDir)
 		db, err = leveldb.RecoverFile(dataDir, nil)
 	}
-	return &LevelDB{ldb: db}, err
+
+	instance := LevelDB{
+		ldb:  db,
+		name: dataDir, // name is used for metrics labels
+		done: make(chan struct{}),
+	}
+
+	// initialize the metric value
+	instance.updateArchivedEnvelopesCount()
+	// checking count on every insert is inefficient
+	go func() {
+		for {
+			select {
+			case <-instance.done:
+				return
+			case <-time.After(time.Second * envelopeCountCheckInterval):
+				instance.updateArchivedEnvelopesCount()
+			}
+		}
+	}()
+	return &instance, err
+}
+
+// GetEnvelope get an envelope by its key
+func (db *LevelDB) GetEnvelope(key *DBKey) ([]byte, error) {
+	defer recoverLevelDBPanics("GetEnvelope")
+	return db.ldb.Get(key.Bytes(), nil)
+}
+
+func (db *LevelDB) updateArchivedEnvelopesCount() {
+	if count, err := db.envelopesCount(); err != nil {
+		log.Warn("db query for envelopes count failed", "err", err)
+	} else {
+		archivedEnvelopesGauge.WithLabelValues(db.name).Set(float64(count))
+	}
 }
 
 // Build iterator returns an iterator given a start/end and a cursor
@@ -81,13 +117,6 @@ func (db *LevelDB) BuildIterator(query CursorQuery) (Iterator, error) {
 		i.Seek(query.cursor)
 	}
 	return &LevelDBIterator{i}, nil
-}
-
-// GetEnvelope get an envelope by its key
-func (db *LevelDB) GetEnvelope(key *DBKey) ([]byte, error) {
-	defer recoverLevelDBPanics("GetEnvelope")
-
-	return db.ldb.Get(key.Bytes(), nil)
 }
 
 // Prune removes envelopes older than time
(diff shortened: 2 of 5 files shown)


mailserver/mailserver_db.go
```diff
@@ -10,6 +10,8 @@ import (
 // is agnostic to the underlying technology used
 type DB interface {
 	Close() error
+	// EnvelopesCount returns current count of envelopes
+	EnvelopesCount() (int, error)
```
adambabik:
```suggestion
	EnvelopesCount() (int64, error)
```

adambabik: Btw do we need this to be a part of the interface?

jakubgs: What's the harm?

jakubgs: I guess the argument that an interface should be as small as possible makes sense to me.

mailserver/mailserver_db_leveldb.go
```diff
@@ -68,7 +68,14 @@ func NewLevelDB(dataDir string) (*LevelDB, error) {
...
 	}
-	return &LevelDB{ldb: db}, err
+	instance := LevelDB{ldb: db}
+
+	count, err := instance.envelopesCount()
+	if err == nil {
+		archivedEnvelopesCounter.Add(float64(count))
+	}
+
+	return &instance, err
```
adambabik, line 93: Do we want to surface this error? This means that an error when collecting metrics prevents from creating an instance of `LevelDB`. This is highly unlikely that `envelopesCount` returns an error and `LevelDB` instance is useful but it may happen.

So, if `envelopesCount` returns an error, you should either log it and ignore or return it but then return `nil` as the first value.

Currently, it's also inconsistent with `NewPostgresDB` constructor.

mailserver/mailserver_db_postgres.go
```diff
@@ -190,6 +196,23 @@ func (i *PostgresDB) Prune(t time.Time, batch int) (int, error) {
...
+	statement := "SELECT count(1) FROM envelopes"
+	stmt, err := i.db.Prepare(statement)
+	if err != nil {
+		return 0, err
+	}
+	defer stmt.Close()
+
+	var count int
+
+	if err = stmt.QueryRow(stmt).Scan(&count); err != nil {
```
adambabik:
```suggestion
	var count int
	return count, stmt.QueryRow(stmt).Scan(&count)
```

mailserver/mailserver_db_leveldb.go
jakubgs: Sure,I was thinking about ignoring it too. But if `count()` fails then the DB is clearly fucked, so it's better to fail the whole DB init.

mailserver/mailserver_db_postgres.go
jakubgs: Neater. Fixed.

mailserver/mailserver_db_leveldb.go
```diff
@@ -68,7 +68,16 @@ func NewLevelDB(dataDir string) (*LevelDB, error) {
 		log.Info("database is corrupted trying to recover", "path", dataDir)
 		db, err = leveldb.RecoverFile(dataDir, nil)
 	}
-	return &LevelDB{ldb: db}, err
+	instance := LevelDB{ldb: db}
+
+	// Initialize the archive counter
+	count, err := instance.envelopesCount()
+	if err != nil {
+		return nil, err
```
adambabik: I though we wanted not to fail but log it only in the case of the error because it's not a crucial operation.

mailserver/mailserver_db_leveldb.go
```diff
@@ -68,7 +68,16 @@ func NewLevelDB(dataDir string) (*LevelDB, error) {
...
+	}
+	archivedEnvelopesCounter.Add(float64(count))
+
+	return &instance, err
```
adambabik, line 93: If `err != nil`, it should be `return nil, err`.

adambabik (approved):

jakubgs:
Changed to run in a Go routine every minute to take into account envelope pruning routine in metrics.

mailserver/mailserver_db_leveldb.go
```diff
@@ -68,7 +68,30 @@ func NewLevelDB(dataDir string) (*LevelDB, error) {
 		log.Info("database is corrupted trying to recover", "path", dataDir)
 		db, err = leveldb.RecoverFile(dataDir, nil)
 	}
-	return &LevelDB{ldb: db}, err
+	instance := LevelDB{ldb: db}
+
+	// checking count on every insert is inefficient
+	go func() {
```
adambabik, line 83: We should kill this goroutine when `Close` is called.

jakubgs:
I don't see a better way of checking for LevelDB close state than a custom `closed` value.

They don't expose [`isClosed()`](https://github.com/syndtr/goleveldb/blob/758128399b1df3a87e92df6c26c1d2063da8fabe/leveldb/db_state.go#L229) nor do they throw `ErrClosed` when calling `NewIterator()`.

mailserver/mailserver_db_leveldb.go
```diff
@@ -68,7 +69,35 @@ func NewLevelDB(dataDir string) (*LevelDB, error) {
...
+		for {
+			if instance.closed {
```
adambabik: Exiting from a goroutine should be done using channels, usually:
```go
for {
    select {
    case <-instance.done:
        return
    case <-time.After(time.Second * envelopeCounterCheckInterval):
        instance.updateArchivedEnvelopesCounter()
    }
}
```

where `close` is `close chan struct{}` initialized in the constructor `instance: make(chan struct{})`.

jakubgs: I think making the loop depend on the state of the DB is better, and I see no problem with checking DB state by checking type of error. But both LevelDB and SQL library don't expose properly those things. So I might as well change it as you described, though I don't think it's any better.

adambabik (approved):
LGTM

It could be improved further:
- making `Close` idempotent as it will panic now when the `done` channel is already closed
- `done` field should be protected with a mutex

jakubgs:
I've modified to make `Close()` idempotent, but I also had to make the archive metrics because on the fleet we run two DBs along side each other, for `whisper` and `waku`. Here's the result:
```
 $ curl -s localhost:9305/metrics | grep '^mailserver_archived_envelopes'
mailserver_archived_envelopes_total{db="waku"} 3960
mailserver_archived_envelopes_total{db="whisper"} 7954
```
In case of LevelDB it uses the full path:
```
 $ curl -s localhost:9090/metrics | grep '^mailserver_archived_envelopes_total'
mailserver_archived_envelopes_total{db="/var/tmp/status-go-mailserver/waku"} 483
mailserver_archived_envelopes_total{db="/var/tmp/status-go-mailserver/wnode"} 483
```

adambabik (approved):

Issue #1929, opened 2020-04-06
Waku Status message not according to spec


kdeme:
# Problem
I've noticed while interop testing status-go with Nimbus wakunode that the handshake fails. 
I see two problems:

1. The keys from the `StatusOptions` struct get character encoded. So resulting rlp contains 48, 49, 50, etc. instead of 0, 1, 2, etc. I think the ABNF part of the spec dictates the actual integer values (if you compare it with other parts).
(see spec: https://github.com/vacp2p/specs/blob/master/specs/waku/waku.md#abnf-specification vs https://github.com/status-im/status-go/blob/develop/waku/handshake.go#L22 )

```
Reading rlp data of packet: StatusOptions
list{
list{
byte 48
blob(8) [3F50624DD2F1A9FC]
}
list{
byte 50
""
}
list{
byte 51
""
}
list{
byte 52
list{
""
""
""
}
}
}
```

1. Status-go disconnects from Nimbus wakunode because of issue 1. I didn't look too much in the code as to why, but I can confirm that the handshake succeeds when I adjust Nimbus wakunode to start the keys from 48.

```
DEBUG[04-06|17:37:43.797] Adding p2p peer id=cc746081f8465015 conn=inbound addr=127.0.0.1:59768 peers=3 name="Nimbus waku node"
DEBUG[04-06|17:37:43.797] confirm peer added ID=cc746081f8465015
DEBUG[04-06|17:37:43.797] peer added event peer=cc746081f8465015843c5c751a25ea1115cb5f4798974d4134ff3f39590cc823 inbound=true
DEBUG[04-06|17:37:43.798] Removing p2p peer id=cc746081f8465015 conn=inbound addr=127.0.0.1:59768 peers=2 duration=669.82µs req=false err="p [cc746081f8465015843c5c751a25ea1115cb5f4798974d4134ff3f39590cc823]: failed to decode status packet: rlp: call of ListEnd not positioned at EOL"
DEBUG[04-06|17:37:43.798] confirm peer dropped ID=cc746081f8465015
```

Why is 2. a problem? This is perhaps not directly written down in the spec( https://github.com/vacp2p/specs/blob/master/specs/waku/waku.md#status), but the idea was that when new k:v pairs in `StatusOptions` get added and the client doesn't know about these, then these get ignored. This was done for forward compatibility. 

So IMO, even with this bug, I think the handshake should still pass (fyi, according to Nimbus side, the handshake succeeds).

[...]

PR #1931, opened 2020-04-07
Waku handshake RLP key variable initialisation

Samyoul:
## Why make the change?

This PR is the first attempt at resolving the issue reported here https://github.com/status-im/status-go/issues/1929, however this PR has been superseded by https://github.com/status-im/status-go/pull/1938.

## What's changed?

This PR is now is essentially a refactor of variable initialisation and the declaration of a new type. 

### Refactor

Variable declaration

- `waku.idxFieldKey map[int]statusOptionKey{}`
- `waku.keyFieldIdx map[statusOptionKey]int{}`

### Added

Type
- `waku.statusOptionKey`

Functions
- `waku.initRLPKeyFields()`
- `waku.init()`

Tests
- `TestInitRLPKeyFields()`

This test explicitly checks that the variable initialisation matches the expected parse result of the `statusOptions` struct.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,7 @@ Session.vim
 /.idea/
 /.vscode/
 /cmd/*/.ethereum/
+*.iml
 
 # created for running container
 _assets/compose/bootnode/keys
--- a/waku/handshake.go
+++ b/waku/handshake.go
@@ -6,12 +6,15 @@ import (
 	"io"
 	"math"
 	"reflect"
+	"strconv"
 	"strings"
 
 	"github.com/ethereum/go-ethereum/rlp"
 )
 
 var defaultMinPoW = math.Float64bits(0.001)
+
+type statusOptionKey int64
 
 // statusOptions defines additional information shared between peers
 // during the handshake.
@@ -26,6 +29,9 @@ type statusOptions struct {
 	ConfirmationsEnabled *bool       `rlp:"key=3"`
 	RateLimits           *RateLimits `rlp:"key=4"`
 	TopicInterest        []TopicType `rlp:"key=5"`
+
+	idxFieldKey map[int]statusOptionKey
+	keyFieldIdx map[statusOptionKey]int
 }
 
 // WithDefaults adds the default values for a given peer.
@@ -57,11 +63,15 @@ func (o statusOptions) WithDefaults() statusOptions {
 	return o
 }
 
-var idxFieldKey = make(map[int]string)
-var keyFieldIdx = func() map[string]int {
-	result := make(map[string]int)
-	opts := statusOptions{}
-	v := reflect.ValueOf(opts)
+//var idxFieldKey = make(map[int]statusOptionKey)
+//var keyFieldIdx = map[statusOptionKey]int{}
+
+func (o *statusOptions) parseStatusOptionKeys() error {
+	kfi := make(map[statusOptionKey]int)
+	ifk := make(map[int]statusOptionKey)
+
+	v := reflect.ValueOf(o)
+
 	for i := 0; i < v.NumField(); i++ {
 		// skip unexported fields
 		if !v.Field(i).CanInterface() {
@@ -72,12 +82,31 @@ var keyFieldIdx = func() map[string]int {
 		if rlpTag == "" {
 			continue
 		}
-		key := strings.Split(rlpTag, "=")[1]
-		result[key] = i
-		idxFieldKey[i] = key
+
+		keys := strings.Split(rlpTag, "=")
+		// skip rlp tags that cannot be split by "="
+		// TODO Do we want to throw an error here if the length is not exactly 2?
+		if len(keys) < 2 {
+			continue
+		}
+
+		// parse keys[1] as an int
+		rkey, err := strconv.ParseInt(keys[1], 10, 64)
+		if err != nil {
+			return err
+		}
+
+		// typecast rkey to be of statusOptionKey type
+		key := statusOptionKey(rkey)
+		kfi[key] = i
+		ifk[i] = key
 	}
-	return result
-}()
+
+	o.keyFieldIdx = kfi
+	o.idxFieldKey = ifk
+
+	return nil
+}
 
 func (o statusOptions) PoWRequirementF() *float64 {
 	if o.PoWRequirement == nil {
@@ -99,7 +128,7 @@ func (o statusOptions) EncodeRLP(w io.Writer) error {
 		field := v.Field(i)
 		if !field.IsNil() {
 			value := field.Interface()
-			key, ok := idxFieldKey[i]
+			key, ok := o.idxFieldKey[i]
 			if !ok {
 				continue
 			}
@@ -117,6 +146,8 @@ func (o *statusOptions) DecodeRLP(s *rlp.Stream) error {
 		return fmt.Errorf("expected an outer list: %v", err)
 	}
 
+	o.parseStatusOptionKeys()
+
 	v := reflect.ValueOf(o)
 
 loop:
@@ -130,7 +161,7 @@ loop:
 		default:
 			return fmt.Errorf("expected an inner list: %v", err)
 		}
-		var key string
+		var key statusOptionKey
 		if err := s.Decode(&key); err != nil {
 			return fmt.Errorf("invalid key: %v", err)
 		}
@@ -138,17 +169,17 @@ loop:
 		// It might happen when there is a new peer
 		// which supports a new option with
 		// a higher index.
-		idx, ok := keyFieldIdx[key]
+		idx, ok := o.keyFieldIdx[key]
 		if !ok {
 			// Read the rest of the list items and dump them.
 			_, err := s.Raw()
 			if err != nil {
-				return fmt.Errorf("failed to read the value of key %s: %v", key, err)
+				return fmt.Errorf("failed to read the value of key %d: %v", key, err)
 			}
 			continue
 		}
 		if err := s.Decode(v.Elem().Field(idx).Addr().Interface()); err != nil {
-			return fmt.Errorf("failed to decode an option %s: %v", key, err)
+			return fmt.Errorf("failed to decode an option %d: %v", key, err)
 		}
 		if err := s.ListEnd(); err != nil {
 			return err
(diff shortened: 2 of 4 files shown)


waku/handshake.go
```diff
@@ -6,12 +6,50 @@ import (
...
 	"math"
 	"reflect"
-	"strings"
+	"strconv"
 
 	"github.com/ethereum/go-ethereum/rlp"
 )
 
-var defaultMinPoW = math.Float64bits(0.001)
+type statusOptionKey uint
```
adambabik:
```suggestion
// statusOptionKey is a current type used in statusOptions as a key.
type statusOptionKey uint
```

waku/handshake.go
```diff
@@ -6,12 +6,50 @@ import (
...
+
+type statusOptionKeyType uint
```
adambabik:
```suggestion
// statusOptionKeyType is a type of a statusOptions key used for a particular instance of statusOptions struct.
type statusOptionKeyType uint
```

waku/handshake.go
```diff
@@ -6,12 +6,50 @@ import (
...
+
+type statusOptionKeyToType struct {
+	Idx  int
+	Key  statusOptionKey
+	Type statusOptionKeyType
+}
+
+const (
+	sOKTS statusOptionKeyType = iota // Status Option Key Type String
```
adambabik:
```suggestion
	sOKTS statusOptionKeyType = iota + 1 // Status Option Key Type String
```

Avoid zero value as it's impossible to tell whether it's an uninitialized variable or actual value.

waku/handshake.go
```diff
@@ -158,6 +206,58 @@ loop:
...
+
+	if o.keyTypeMapping.keyFieldIdx == nil {
+		o.keyTypeMapping.keyFieldIdx = make(map[statusOptionKey]*statusOptionKeyToType)
+	}
+
+	o.keyTypeMapping.idxFieldKey[ktt.Idx] = ktt
+	o.keyTypeMapping.keyFieldIdx[ktt.Key] = ktt
+}
+
+func (k statusOptionKeyToType) decodeStream(s *rlp.Stream) error {
```
adambabik:
```suggestion
func (k *statusOptionKeyToType) decodeStream(s *rlp.Stream) error {
```

Shouldn't it be a pointer receiver? You assign values to `k` below.

waku/handshake.go
```diff
@@ -6,12 +6,50 @@ import (
...
+}
+
+const (
+	sOKTS statusOptionKeyType = iota // Status Option Key Type String
+	sOKTU                            // Status Option Key Type Uint
+)
+
+var (
+	defaultMinPoW = math.Float64bits(0.001)
+	idxFieldKey   = map[int]statusOptionKey{
```
adambabik: I gave it a though and we kinda rely on internal Go implementation here. It should not change but who knows...

I would do it this way:
1. When encoding, go through all fields with `rlp:"key=N"`.
2. When decoding, create a method `statusOptions.setField(key, value)` which would set a value for a field based on `key`, i.e. for a field with `rlp:"key={key}"`.

waku/handshake.go
```diff
@@ -26,6 +64,7 @@ type statusOptions struct {
 	ConfirmationsEnabled *bool       `rlp:"key=3"`
 	RateLimits           *RateLimits `rlp:"key=4"`
 	TopicInterest        []TopicType `rlp:"key=5"`
+	keyTypeMapping       keyTypeMapping
```
adambabik: Based on the comment above, we would not need this. We would only need a field `keyType statusOptionKeyType`.

waku/handshake.go
Samyoul: Good point

waku/handshake.go
Samyoul: Ah, yes. Thank you. I can't believe I missed this.

waku/handshake.go
Samyoul: >I gave it a though and we kinda rely on internal Go implementation here. It should not change but who knows...

Ok, I can implement the `init()` func we discussed earlier.

waku/handshake.go
Samyoul: Isn't this presuming that all the keys are of the same type? What if some are `string` and some are `uint`?

Samyoul: Or would there be a `statusOptionKeyType` for each `statusOptionKey`?

adambabik: > Isn't this presuming that all the keys are of the same type? What if some are string and some are uint?

This is not possible. The current version always sends keys as strings and this will always send keys as ints unless it received keys as strings then it will also reply with string keys.

waku/handshake.go
```diff
@@ -26,6 +53,39 @@ type statusOptions struct {
...
+		if !v.Field(i).CanInterface() {
+			continue
+		}
+		rlpTag := v.Type().Field(i).Tag.Get("rlp")
+		// skip fields without rlp field tag
+		if rlpTag == "" {
+			continue
+		}
+
+		keys := strings.Split(rlpTag, "=")
```
adambabik, line 55: I would add a check:
```go
if len(keys) != 2 || keys[0] != "key" {
    panic("invalid value of \"rlp\" tag, expected \"key=N\" where N is uint")
}
```

waku/handshake.go
```diff
@@ -26,6 +53,39 @@ type statusOptions struct {
 	ConfirmationsEnabled *bool       `rlp:"key=3"`
 	RateLimits           *RateLimits `rlp:"key=4"`
 	TopicInterest        []TopicType `rlp:"key=5"`
+	keyTypeMapping       keyTypeMapping
+}
+
+// initFLPKeyFields initialises the values of `idxFieldKey` and `keyFieldIdx`
+func initRLPKeyFields() error {
```
adambabik: Nice 👍

waku/handshake.go
Samyoul: Ok. I will sort this.

waku/handshake.go
```diff
@@ -97,15 +135,41 @@ func (o statusOptions) EncodeRLP(w io.Writer) error {
...
+			continue
+		}
+
+		value := field.Interface()
+		key, ok := idxFieldKey[i]
+		if !ok {
+			continue
+		}
+
+		var val []interface{}
```
adambabik: I don't quite get it down here. I would see something like:
```go
var keyValue interface{}
if o.keyType == sOKTS {
    keyValue = fmt.Sprint(key)
} else {
    keyValue = key
}

if value != nil {
    optionsList = append(optionsList, []interface{}{keyValue, value})
}
```

Or if you want to keep `statusOptionKeyToType`:
```go
ktt := statusOptionKeyToType{Key: key, Type: o.typ}
if value != nil {
    optionsList = append(optionsList, []interface{}{ktt, value}) // statusOptionKeyToType must implement `EncodeRLP`
}
```

Samyoul: I'm completely removing `statusOptionKeyToType` so all of this will be gone completely and work more like you've suggested. 😄

waku/handshake.go
Samyoul: Yes. I had a similar check in previous versions of this functionality but for some reason it is gone... Or maybe I just imagined it. Either way I 100% agree and will implement this.

Samyoul:
https://github.com/status-im/status-go/pull/1938 makes this branch redundant. Marking this for deletion.

Waiting to see if there is any code that is reusable. Probably the init functionality.

waku/handshake.go
Samyoul: @cammellos FYI, I'd like to keep this functionality in waku. I think that it is more sustainable to have variables initialised using an actual init function.

I can revert all the changes implemented by this branch and then keep all the changes related to this specific change.

cammellos: sure, sounds good to me

waku/handshake.go
```diff
@@ -11,7 +11,14 @@ import (
 	"github.com/ethereum/go-ethereum/rlp"
 )
 
-var defaultMinPoW = math.Float64bits(0.001)
+// statusOptionKey is a current type used in statusOptions as a key.
+type statusOptionKey string
```
Samyoul, line 15: I've explicitly declared the RLP options key type here. In `waku/1` this should be changed to be `uint`

waku/handshake.go
```diff
@@ -28,6 +35,35 @@ type statusOptions struct {
...
+
+		keys := strings.Split(rlpTag, "=")
+
+		if len(keys) != 2 || keys[0] != "key" {
+			panic("invalid value of \"rlp\" tag, expected \"key=N\" where N is uint")
+		}
+
+		// typecast key to be of statusOptionKey type
+		keyFieldIdx[statusOptionKey(keys[1])] = i
+		idxFieldKey[i] = statusOptionKey(keys[1])
```
Samyoul, line 63: In `waku/1` the key needs to be parsed into a `uint` before it can be used as a `statusOptionKey`

waku/handshake.go
```diff
@@ -28,6 +35,35 @@ type statusOptions struct {
...
+		// skip fields without rlp field tag
+		if rlpTag == "" {
+			continue
+		}
+
+		keys := strings.Split(rlpTag, "=")
+
+		if len(keys) != 2 || keys[0] != "key" {
+			panic("invalid value of \"rlp\" tag, expected \"key=N\" where N is uint")
+		}
```
Samyoul, line 59: @cammellos Any objection to throwing a panic here, rather than throwing an error? I think a panic needs to be given, and in my mind it can be here or in the package `init()`.

.gitignore
```diff
@@ -66,6 +66,7 @@ Session.vim
 /.idea/
 /.vscode/
 /cmd/*/.ethereum/
+*.iml
```
Samyoul, line 69: @cammellos I've added my IDE file to the gitignore in this pull. Any objection?

waku/handshake.go
cammellos: sure, panic looks fine

.gitignore
cammellos: no objections :)

Samyoul:
Hey @cammellos and @jakubgs I'm ready for this to be merged into `develop`. When you get some time could you pass your eye over this 100% essential refactor 😄.

cammellos (approved):

cammellos:
I will quickly test this against the production app and merge if the test is successful.

cammellos:
tested successfully

PR #1933, opened 2020-04-08
Add replies to messages

cammellos:
Currently replies to messages are handled in status-react.
    This causes some issues with the fact that sometimes replies might come
    out of order, they might be offloaded to the database etc.
    
This commit changes the behavior so that status-go always returns the
    replies, and in case a reply comes out of order (first the reply, later
    the message being replied to), it will include in the messages the
    updated message.
    
It also adds some fields (RTL,Replace,LineCount) to the database which
    were not previously saved, resulting in some potential bugs.
    
The method that we use to pull replies is currently a bit naive, we just
    pull all the message again from the database, but has the advantage of
    being simple. It will go through performance testing to make sure
    performnace are acceptable, if so I think it's reasonable to avoid some
    complexity.


protocol/messenger.go
```diff
@@ -2812,6 +2822,20 @@ func (m *Messenger) ValidateTransactions(ctx context.Context, addresses []types.
 	return &response, nil
 }
 
+func (m *Messenger) addResponses(messages []*Message) ([]*Message, error) {
```
adambabik: What this method does is actually pulling messages not adding them to the messenger/store as a name suggests. Should we rename it? 

Or it can be a separate function:
```go
messageAndResponseIDs := pickMessageAndResponseIDs(messageState.Response.Messages)
messagesWithResponses, err := m.persistence.MessagesByIDs(messageAndResponseIDs)
```

protocol/persistence_legacy.go
```diff
@@ -313,6 +325,51 @@ func (db sqlitePersistence) MessagesExist(ids []string) (map[string]bool, error)
 	return result, nil
 }
 
+func (db sqlitePersistence) MessagesByIDs(ids []string) ([]*Message, error) {
+	if len(ids) == 0 {
+		return nil, nil
+	}
+
+	idsArgs := make([]interface{}, 0, len(ids))
```
adambabik, line 333: I don't think this is necessary. Can't we just pass `ids` to `Query()`?

cammellos: just checked, actually not sure we can, as `Query` expects `[]interface{}` rather than `[]string`,
I have made the conversion a bit faster (not using append,  and using index instead, although not sure it makes a difference here as the array is allocated already), is there any other way to do the conversion?

cammellos:
@adambabik thanks for the feedback, I have addressed the name change, but didn't change the array construction because of the type conversion, not sure if there's something better?
Thanks!

andremedeiros (commented):
Just checking, should the `down` migration be written too? If we migrate up, down, and up again, it will fail in its current state, right?

protocol/persistence_legacy.go
adambabik: You are right, sorry for the confusion. And no, there is no other way to case it :( It's the best way.

adambabik (approved):

cammellos:
@andremedeiros we should, but unfortunately sqlite does not support dropping columns, so effectively  this is a one way migration (it would be possible to write a migration with `down` by creating another table in up, copying all the data and in down do the opposite, but it's fiddly and might take a bit longer). Currently we don't actually really use `down` migrations, we would probably migrate UP if any issue occurs.

andremedeiros:
> sqlite does not support dropping columns

I... uh... ok.

VERSION
```diff
@@ -1 +1 @@
-0.51.0
+0.51.1
```
jakubgs: #1935 already bumped it to `0.51.1`.

jakubgs (approved):
LGTM, thought I wish we had a doc explaining all the fields we use. Things like `Replace` or `RTL` aren't very descriptive to the uninitiated.

cammellos:
Thanks, added some descriptions in `protocol/message.go`

PR #1934, opened 2020-04-08
Update protocol README

adambabik:


protocol/README.md
```diff
@@ -1,6 +1,18 @@
 # status-go/protocol
 
-This is the Status Protocol implementation in Go.
+This is an implementation of the [secure transport](https://specs.status.im/spec/5) and [payloads](https://specs.status.im/spec/6) which are a part of [the Status Client specification](https://specs.status.im/spec/1).
 
+This implementation uses SQLite and [SQLCipher](github.com/mutecomm/go-sqlcipher) for persistent storage.
```
jakubgs, line 5: That's kinda confusing, Because we use LevelDB and PorsgreSQL for persistent storage too. My understanding is that this is for storing of stuff like accounts, ens names, chat messages received, and such. Could you maybe clarify that to distinguish it from mailserver storage?

jakubgs (approved):

protocol/README.md
adambabik: This is exclusively about `protocol` package as this README is in `protocol/` dir. I would not mention here mailservers as it's unrelated.

jakubgs: You don't have to mention mailserver. You just have to explain WHAT is being stored.

PR #1935, opened 2020-04-09
Increment retry count when ENS verification fails

cammellos:
We did not increment the counter when verification failed, which meant
that we would not backoff and retry more often that we should have.

cc @flexsurfer


flexsurfer (approved):

adambabik (approved):
LGTM

I created https://github.com/status-im/status-go/issues/1936 to fix the lock issue we have.

jakubgs (approved):

PR #1938, opened 2020-04-16
Upgrade to waku/1

cammellos:
This commit upgrades to waku-1.

Now the nodes will advertise both `waku/0` and `waku/1` protocol
version, so compatibility is kept.
The old behavior of `waku/0` is maintained.

### DIFF
--- a/waku/handshake.go
+++ b/waku/handshake.go
@@ -2,11 +2,8 @@ package waku
 
 import (
 	"errors"
-	"fmt"
 	"io"
 	"math"
-	"reflect"
-	"strings"
 
 	"github.com/ethereum/go-ethereum/rlp"
 )
@@ -20,12 +17,12 @@ var defaultMinPoW = math.Float64bits(0.001)
 // In the case of RLP, options should be serialized to an array of tuples
 // where the first item is a field name and the second is a RLP-serialized value.
 type statusOptions struct {
-	PoWRequirement       *uint64     `rlp:"key=0"` // RLP does not support float64 natively
-	BloomFilter          []byte      `rlp:"key=1"`
-	LightNodeEnabled     *bool       `rlp:"key=2"`
-	ConfirmationsEnabled *bool       `rlp:"key=3"`
-	RateLimits           *RateLimits `rlp:"key=4"`
-	TopicInterest        []TopicType `rlp:"key=5"`
+	PoWRequirement       *uint64
+	BloomFilter          []byte
+	LightNodeEnabled     *bool
+	ConfirmationsEnabled *bool
+	RateLimits           *RateLimits
+	TopicInterest        []TopicType
 }
 
 // WithDefaults adds the default values for a given peer.
@@ -57,27 +54,13 @@ func (o statusOptions) WithDefaults() statusOptions {
 	return o
 }
 
-var idxFieldKey = make(map[int]string)
-var keyFieldIdx = func() map[string]int {
-	result := make(map[string]int)
-	opts := statusOptions{}
-	v := reflect.ValueOf(opts)
-	for i := 0; i < v.NumField(); i++ {
-		// skip unexported fields
-		if !v.Field(i).CanInterface() {
-			continue
-		}
-		rlpTag := v.Type().Field(i).Tag.Get("rlp")
-		// skip fields without rlp field tag
-		if rlpTag == "" {
-			continue
-		}
-		key := strings.Split(rlpTag, "=")[1]
-		result[key] = i
-		idxFieldKey[i] = key
-	}
-	return result
-}()
+func (o statusOptions) EncodeRLP(w io.Writer) error {
+	return errors.New("not implemented")
+}
+
+func (o *statusOptions) DecodeRLP(s *rlp.Stream) error {
+	return errors.New("not implemented")
+}
 
 func (o statusOptions) PoWRequirementF() *float64 {
 	if o.PoWRequirement == nil {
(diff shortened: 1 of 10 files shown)


waku/handshake.go
```diff
@@ -20,12 +17,12 @@ var defaultMinPoW = math.Float64bits(0.001)
...
-	ConfirmationsEnabled *bool       `rlp:"key=3"`
-	RateLimits           *RateLimits `rlp:"key=4"`
-	TopicInterest        []TopicType `rlp:"key=5"`
+	PoWRequirement       *uint64
+	BloomFilter          []byte
+	LightNodeEnabled     *bool
+	ConfirmationsEnabled *bool
+	RateLimits           *RateLimits
+	TopicInterest        []TopicType
 }
```
Samyoul, line 26: What is the purpose of this struct? It seems to be a skeleton that has no uses.

waku/handshake.go
```diff
@@ -87,80 +70,9 @@ func (o statusOptions) PoWRequirementF() *float64 {
...
-}
-
 func (o statusOptions) Validate() error {
-	if len(o.TopicInterest) > 1000 {
-		return errors.New("topic interest is limited by 1000 items")
+	if len(o.TopicInterest) > 10000 {
+		return errors.New("topic interest is limited by 10000 items")
 	}
 	return nil
 }
```
Samyoul, line 78: Does `waku/1` introduce a higher topic limit?

waku/peer.go
```diff
@@ -100,12 +102,13 @@ func (p *Peer) stop() {
...
+func (p *Peer) handshakeV0() error {
+	p.logger.Info("V0 Handshake", zap.Binary("peerID", p.ID()))
 	// Send the handshake status message asynchronously
 	errc := make(chan error, 1)
-	opts := p.host.toStatusOptions()
+	opts := p.host.toStatusOptionsV0()
 	go func() {
-		errc <- p2p.SendItems(p.ws, statusCode, ProtocolVersion, opts)
+		errc <- p2p.SendItems(p.ws, statusCode, ProtocolVersion0, opts)
 	}()
```
Samyoul, line 112: This section is a prime candidate for abstraction. I can implement the code for this if required / desired.

I think that the `peer.handshake()` api shouldn't change. the underlying mechanisms of this function should choose the type of handshake version. Better yet, base the abstraction on the `statusOptions` level and keep the versioning restricted to what needs to be changed. This should hopefully keep the code [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)er.

The functionality in `Waku.toStatusOptions()` needs to change, so I'd recommend that `Waku` host and peer have a function that allows them to determine the highest common version they both support. The chosen version should be used as the basis for generating options from an options implementation of an interface.

Any time we are writing things like `functionV0`, `functionV1` is a "code smell" that tells us we need to be implementing an interface and abstracting the functionality.

Samyoul (changes requested):
I can implement some of these changes to give you my thoughts on how this should work.

waku/handshake.go
cammellos: Already `waku/0` should have had the higher topic limit https://github.com/vacp2p/specs/blob/master/specs/waku/waku.md , I think it was left off from the implementation.

waku/peer.go
cammellos: I think talking about code smell and DRY is simplistic in this context, rather we should have a conversation about how to handle API versioning, as there are distinct strategies (both widely used).

One is to do as you propose, which is fairly traditional, 
we keep a single namespace/method and have conditional switches in the same codepath (i.e a single handshake method that handles both v0/v).
This has the advantage to have obviously less code and improve re-usability and being DRYer, although complexity increases as you add support to multiple versions `(v1/v2)`.

Another approach is to have completely separate code path so two new namespaces `v0/` and `v1/` (or `peerv1`/ `peerv2`), with a lot of code duplication.
This brings some advantages. 
API behavior is clearer, the individual code is simpler, fixing things in `v0` will not break things in `v1` , decommissioning/introducing a new version is a no-brainer (remove namespace/create a new one). Note that this is what actually whisper  is using  https://github.com/ethereum/go-ethereum/tree/34d94e22d9c9f3c2373a398269ecbba0139827d2/whisper . Also advocated by the likes of Rich Hickey https://www.youtube.com/watch?v=oyLBGkS5ICk and the widely used in microservices development.
The disadvantage as you point out is code duplication, which increases the cost of maintaining the code if changes needs to be made on both versions.
 In our case also we will need to "bridge" `v0` and `v1` so that can be done by sharing some state, or more explicitly as we did with `waku/whisper`, depending on the level of segregation of the two code paths.

The decision of which strategy to follow I think needs to be more nuanced than code smell/dry, we should take into consideration advantages and tradeoffs of both methods, precedence and specifically how it applies to this project (for example Spotify I believe tends to keep codebases separate for their apis, as they have to maintain older version for long as embedded devices are often non-upgradable, how long do we have to support `v0` ourselves?).

At this point in time I don't have a strong opinion either ways, I tend to favor the second approach as it's an already established pattern in `ethereum-go`, and there's a hint of that under `protocol/v1` and more generally I like to make things explicit when it comes to API versioning, but at the end of the day I am sure we can have good results with both.

What's your opinion?
 
cc @andremedeiros

Samyoul: Thank you for laying out those pros and cons. I had an assumption that there may be other reasons for your choice of implementation, I am happy to understand them better now.

On reflection your reasoning for implementing your current pattern seems to be more maintainable for phasing in and out versions, which is basically our goal in this case.

>how long do we have to support v0 ourselves?

This is a good point, I suspect that we will be phasing out `waku/0` in the near future. Maybe only a `status-go` version or two before all peers are using at least `waku/1`.

Question. Where do we want to draw the line for API versioning? At the package level? On the `Waku` struct level? Etc

I can see that you've chosen to implement the versioning on all the code that touches `statusOptions`. Should this be the starting point?

cammellos: > This is a good point, I suspect that we will be phasing out waku/0 in the near future. Maybe only a status-go version or two before all peers are using at least waku/1.

That's a difficult question to answer, we tried many times :) 
Basically we can decommission one version once no mobile phone (i.e client installation) is actually running it, which is a bit tricky to tell, I'd say roughly that's 3 months after release of `v1`, but in the company there have been some proponent of the "never break" approach.
Myself I think it's fine to drop support once we notice that no one is actually connecting using `waku/0` for a week or two.

> Question. Where do we want to draw the line for API versioning? At the package level? On the Waku struct level? Etc

That's difficult to say actually, the temptation was to have completely different namespaces, a la `waku0` and `waku1`, but I think at this point is a bit unnecessary and just makes things a bit more complicated then they ought to.
Having two separate namespaces would mean having completely separated state, but in our case an envelope sent through `waku/0` should be relied to `waku/1` peers and vice versa. 
So if we go for complete separation we'd have to implement some bridging, similarly as we did with `whisper`, which sounds like an overkill.

One way that seems tempting to me is to have say `peerv0/` and `peerv1/` namespaces and define an interface, for example:

```
type Peer interface {
  handleStatusUpdateCode error
  handleP2PMessageCode error
  handshake error 
  notifyAboutPoWRequirementChange error
}
```

and then have two concrete implementation `peerv0` and `peerv1`.

We can factor out some common code that does not change into either embedded structs or some other form.

> I can see that you've chosen to implement the versioning on all the code that touches statusOptions. Should this be the starting point?

No no :) I only coded this without thinking much, my main aim was to make sure running `waku/0` and `waku/1` together is possible and works (basically I wanted to build a client https://github.com/status-im/status-react/pull/10344 that is compatible with both `waku/0` and `waku/1`, which seems like it's working), as that's not something we ever tried, so although some code here we might want to use, it's not meant to be kept or give an indication on how things should be.

What do you think is the best way to approach this?

Samyoul: >Myself I think it's fine to drop support once we notice that no one is actually connecting using `waku/0` for a week or two.

For what it is worth, Adam felt that this should be the approach. I suppose it doesn't affect the code too much to have clearly labelled legacy code, that can be optionally removed at some future date.

Perhaps we can add clear failing errors that report "version incompatibility please update your client" messages.

---

>That's difficult to say actually, the temptation was to have completely different namespaces, a la waku0 and waku1, but I think at this point is a bit unnecessary and just makes things a bit more complicated then they ought to.

I agree with this.

>One way that seems tempting to me is to have say peerv0/ and peerv1/ namespaces and define an interface, 
>
>and then have two concrete implementation peerv0 and peerv1.
>
>We can factor out some common code that does not change into either embedded structs or some other form.

Yes, I agree. This is the would make the most sense. And any future versions may embed the functionality if needed, or may not if a future change is required.

>No no :) I only coded this without thinking much, my main aim was to make sure running waku/0 and waku/1 together is possible

Haha 😆 I thought that was the case, but I wasn't sure how to ask the question. I know this is only a draft so, we're getting functionality first and then making it pretty after.

---

>What do you think is the best way to approach this?

Basically I think that this depends on where we draw the line for where versioning should occur inside the waku package.

I agree that versioning the entire package is overkill.

Looking through the code `Peer` probably is the best place to make the distinction between versions. Though this is a hard question to answer as it will be a nightmare of a refactor if we need to version at a higher level at some future point. I think that automated interop testing will be 100% essential from this point forward.

Your suggestion of a `Peer` interface is as far as I can see the best solution in this case where we only want to change the RLP key types, but don't want to turn the whole package upside-down.

Samyoul:
@cammellos is this a dead branch now?

PR #1943, opened 2020-04-20
Don't pass contacts without custom fields to the client

cammellos:
When receiving a message with save a contact in the database in order to
avoid re-calculating image/profile.
This contact is then passed to the client, which can negatively impact
performance.

This commit changes the behavior so that only those contacts that have
some custom fields (have been explicitly added by the user, have been
blocked by the user, have sent a contact request or have a verified ens
name) are passed to the client.


protocol/messenger_test.go
```diff
@@ -1191,7 +1191,7 @@ func (s *MessengerSuite) TestBlockContact() {
 		Name:        "contact-name",
 		Photo:       "contact-photo",
 		LastUpdated: 20,
-		SystemTags:  []string{"1", "2"},
+		SystemTags:  []string{contactAdded, contactRequestReceived},
```
Samyoul, line 1194: I like the use of constants rather than the inline declaration. 👍

Samyoul (approved):
LGTM

protocol/messenger.go
```diff
@@ -2009,8 +2021,8 @@ func (m *Messenger) handleRetrievedMessages(chatWithMessages map[transport.Filte
 		}
 	}
 
-	if len(messageState.Response.Contacts) > 0 {
-		err = m.persistence.SaveContacts(messageState.Response.Contacts)
+	if len(contactsToSave) > 0 {
+		err = m.persistence.SaveContacts(contactsToSave)
```
jakubgs, line 2025: Wouldn't using `messageState.ModifiedContacts` here make sense?

protocol/messenger_installations_test.go
```diff
@@ -250,28 +250,6 @@ func (s *MessengerInstallationSuite) TestSyncInstallation() {
...
 	s.Require().NotNil(statusChat)
 
-	var actualContact *Contact
-	for _, c := range allContacts {
-		if c.ID == contact.ID {
-			actualContact = c
-		}
-	}
-
+	actualContact := allContacts[0]
```
jakubgs: Why are we just picking the first one here? I assume it's the only one with custom fields, but could we make this more explicit?

jakubgs (approved):
Makes sense to me.

protocol/messenger.go
cammellos: `ModifiedContacts` contains only the ID of the one that have  been modified, we need to hydrate them (pull them from `allContacts`)

protocol/messenger_installations_test.go
Samyoul: @jakubgs how would you propose making this more explicit? Adding a comment? Changing the variable name?

Samyoul: @cammellos is there a check that the `allContacts` slice has a len > 0?

cammellos: Yes, we would throw an error if the slice has length `0` (in `RetryWithBackOff` we return an error until `len(x) == 1`)

PR #1947, opened 2020-04-22
prepare for waku/1

cammellos:
# Why make the change?

As discussed previously, the way we will move across versions is to maintain completely separate 
codebases and eventually remove those that are not supported anymore.

This has the drawback of some code duplication, but the advantage is that is more
explicit what each version requires, and changes in one version will not 
impact the other, so we won't pile up backward compatible code. 
This is the same strategy used by `whisper` in go ethereum and is influenced by
https://www.youtube.com/watch?v=oyLBGkS5ICk .

All the code that is used for the networking protocol is now under `v0/`.
Some of the common parts might still be refactored out.
The main namespace `waku` deals with `host`->`waku` interactions (through RPC), 
while `v0` deals with `waku`->`remote-waku` interactions.

In order to support `v1`, the namespace `v0` will be copied over, and changed to 
support `v1`. Once `v0` will be not used anymore, the whole namespace will be removed.

This PR does not actually implement `v1`, I'd rather get things looked over to
make sure the structure is what we would like before implementing the changes.

# What has changed?

- Moved all code for the common parts under `waku/common/` namespace
- Moved code used for bloomfilters in `waku/common/bloomfilter.go`
- Removed all version specific code from `waku/common/const` (`ProtocolVersion`, status-codes etc)
- Added interfaces for `WakuHost` and `Peer` under `waku/common/protocol.go`

# Things still to do

Some tests in `waku/` are still testing by stubbing components of a particular version (`v0`).
I started moving those tests to instead of stubbing using the actual component, which increases
the testing surface. Some other tests that can't be easily ported should be likely moved under
`v0` instead. Ideally no version specif code should be exported from a version namespace (for 
example the various codes, as those might change across versions). But this will be a work-in-progress.

Some code that will be common in `v0`/`v1` could still be extract to avoid duplication, and duplicated only
when implementations diverge across versions.


waku/common/filter.go
```diff
@@ -252,3 +250,8 @@ func IsPubKeyEqual(a, b *ecdsa.PublicKey) bool {
 	// the curve is always the same, just compare the points
 	return a.X.Cmp(b.X) == 0 && a.Y.Cmp(b.Y) == 0
 }
+
+// ValidatePublicKey checks the format of the given public key.
+func ValidatePublicKey(k *ecdsa.PublicKey) bool {
+	return k != nil && k.X != nil && k.Y != nil && k.X.Sign() != 0 && k.Y.Sign() != 0
+}
```
Samyoul: I presume you chose to move this here because `filter.go` has the most call to the `ValidatePublicKey()` function. I've spotted a number of functions like this that don't really belong to any one file. I wonder if moving them to a `helper.go` may be a more suitable place for them. What do you think?

waku/common/message.go
```diff
@@ -487,3 +458,32 @@ func (store *MemoryMessageStore) Pop() ([]*ReceivedMessage, error) {
...
+
+// ContainsOnlyZeros checks if the data contain only zeros.
+func ContainsOnlyZeros(data []byte) bool {
+	for _, b := range data {
+		if b != 0 {
+			return false
+		}
+	}
+	return true
+}
```
Samyoul: This may also be a similar case to https://github.com/status-im/status-go/pull/1947/files#r415707643 . What do you think?

waku/common/filter.go
cammellos: yes, that makes more sense, I can gather them up and just add to `helper.go`

waku/common/randomid.go
```diff
@@ -0,0 +1,20 @@
...
+	buf, err := GenerateSecureRandomData(KeyIDSize)
+	if err != nil {
+		return "", err
+	}
+	if !ValidateDataIntegrity(buf, KeyIDSize) {
+		return "", fmt.Errorf("error in generateRandomID: crypto/rand failed to generate random data")
+	}
+	id = common.Bytes2Hex(buf)
+	return id, err
+}
```
Samyoul: Maybe another case of https://github.com/status-im/status-go/pull/1947/files#r415707643 ?

waku/common/rate_limiter_test.go
```diff
@@ -173,3 +174,20 @@ type mockRateLimiterHandler struct {
...
+}
+
+func (p *TestWakuPeer) IP() net.IP {
+	return p.peer.Node().IP()
+}
+
+func (p *TestWakuPeer) ID() []byte {
+	id := p.EnodeID()
+	return id[:]
+}
```
Samyoul: This seems to be the only concrete implementation of the new `RateLimiterPeer` interface. Do we not implement `RateLimiterPeer` anywhere else in the code? Or does the `RateLimiterPeer` interface exist only for testing purposes?

waku/common/validators.go
```diff
@@ -0,0 +1,13 @@
...
+// which is the simplest and the most common bug.
+func ValidateDataIntegrity(k []byte, expectedSize int) bool {
+	if len(k) != expectedSize {
+		return false
+	}
+	if expectedSize > 3 && ContainsOnlyZeros(k) {
+		return false
+	}
+	return true
+}
```
Samyoul: This feels like a case of https://github.com/status-im/status-go/pull/1947/files#r415707643, but there is a known pattern of placing validation functions into `validators.go` so I think this case makes sense to me.

waku/v0_test.go
```diff
@@ -81,7 +85,7 @@ type TestData struct {
 }
 
 type TestNode struct {
-	shh     *Waku
+	waku    *Waku
```
Samyoul, line 88: Good catch.

waku/common/rate_limiter_test.go
cammellos: `Peer` also implements it, before it was actually using the whole `Peer` as an argument, so I initially kept it as it is, but I feel like `RateLimiter` should impose the least amount of requirements on its arguments.

I have actually removed `EnodeID()` as not used (locally, not pushed yet), so basically all `RateLimiter` cares about is something that answer to:

```
func (p *TestWakuPeer) IP() net.IP
func (p *TestWakuPeer) ID() []byte
```

Which intuitively seems to make sense to me, as it is basically a tuple IP/ID, all the info we need to rate limit something.

waku/waku.go
```diff
@@ -1030,246 +1019,84 @@ func (w *Waku) Stop() error {
 
 // HandlePeer is called by the underlying P2P layer when the waku sub-protocol
 // connection is negotiated.
-func (w *Waku) HandlePeer(peer *p2p.Peer, rw p2p.MsgReadWriter) error {
+func (w *Waku) HandlePeer(p2pPeer *p2p.Peer, rw p2p.MsgReadWriter) error {
 	// Create the new peer and start tracking it
-	wakuPeer := newPeer(w, peer, rw, w.logger.Named("waku/peer"))
+	var peer common.Peer = v0.NewPeer(w, p2pPeer, rw, w.logger.Named("waku/peer"))
```
Samyoul, line 1019: Should we be calling a concrete implementation of the `NewPeer()` function? This could be a good place to implement factory pattern.

Something like:

`var peer common.Peer = PeerFactory(w, p2pPeer, rw, w.logger.Named("waku/peer"))`

Then `PeerFactory()` could also be responsible for finding the highest version of compatibility and passing out the concrete `common.Peer` implementation.

cammellos: sure, makes sense, I will implement that in the same PR as we introduce `v1` code if that's ok

waku/waku.go
```diff
@@ -145,25 +145,25 @@ func New(cfg *Config, logger *zap.Logger) *Waku {
...
+		Name:    v0.Name,
+		Version: uint(v0.Version),
+		Length:  v0.NumberOfMessageCodes,
 		Run:     waku.HandlePeer,
 		NodeInfo: func() interface{} {
 			return map[string]interface{}{
-				"version":        ProtocolVersionStr,
+				"version":        v0.VersionStr,
 				"maxMessageSize": waku.MaxMessageSize(),
 				"minimumPoW":     waku.MinPow(),
```
Samyoul, line 163: I imagine this is the place where all the supported versions are declared? I think I remember from your draft that this was an array of `p2p.Protocol`, this could be the main section that any potential `PeerFactory{}` uses to determine which peer to instantiate.

waku/v0/init.go
```diff
@@ -0,0 +1,5 @@
+package v0
+
+func Init() {
+	initRLPKeyFields()
+}
```
Samyoul: We don't need to export this function, `init()` will fire automatically any time that Go accesses the package.

cammellos: ah, nice, I didn't know that, thanks!

waku/waku.go
```diff
@@ -110,7 +110,7 @@ type Waku struct {
 
 // init initialises the waku package
 func init() {
-	initRLPKeyFields()
+	v0.Init()
 }
```
Samyoul: This isn't needed now, see https://github.com/status-im/status-go/pull/1947/files#r415753463 . If we are moving the variable initialisation into the version packages, then they can be loaded automatically when the specific package is accessed.

waku/common/rate_limiter_test.go
Samyoul: Ah ok, I see that. That's a good implementation.

Samyoul (approved):
Apart from the fairly minor suggestions we've discussed, everything looks great. Looking forward to seeing the v1 implementation.

PR #1948, opened 2020-04-22
Set chat active on being re-invited

cammellos:
If the user deletes/leaves a group chat, the chat is set as not active.
This means that if we are re-invited to the chat it won't be shown to
the user.

This commit changes this behavior so that if we are re-invited to the
chat it is set as active again.


protocol/chat.go
```diff
@@ -150,7 +150,16 @@ func (c *Chat) MembersAsPublicKeys() ([]*ecdsa.PublicKey, error) {
...
 }
 
-func (c *Chat) updateChatFromProtocolGroup(g *v1protocol.Group) {
+func (c *Chat) updateChatFromProtocolGroup(myID string, g *v1protocol.Group) {
+
+	// Check if we were already in the chat
+	var hadJoined bool
+	for _, member := range c.Members {
+		if myID == member.ID {
+			hadJoined = true
```
jakubgs: Wouldn't it be cleaner to have a `func (c *Chat) HasMember(memberId string) bool` method?

protocol/messenger_test.go
```diff
@@ -936,6 +936,73 @@ func (s *MessengerSuite) TestChangeNameGroupChat() {
...
+
+	_, err = theirMessenger.ConfirmJoiningGroup(context.Background(), ourChat.ID)
+	s.NoError(err)
+
+	// Wait for join group event
+	err = tt.RetryWithBackOff(func() error {
+		var err error
+		response, err = s.m.RetrieveAll()
+		if err == nil && len(response.Chats) == 0 {
+			err = errors.New("no joining group event received")
```
jakubgs: We use this `RetryWithBackOff()` block a lot in these tests, wouldn't it make sense to refactor it out as a helper for waiting with a custom error message as argument?

jakubgs (approved):

protocol/chat.go
```diff
@@ -150,7 +150,16 @@ func (c *Chat) MembersAsPublicKeys() ([]*ecdsa.PublicKey, error) {
 	return stringSliceToPublicKeys(publicKeys, true)
 }
 
-func (c *Chat) updateChatFromProtocolGroup(g *v1protocol.Group) {
+func (c *Chat) updateChatFromProtocolGroup(myID string, g *v1protocol.Group) {
```
Samyoul: The only real issue I've got is the name of the function. It should probably change as it is no longer `updateChatFromProtocolGroup(g *v1protocol.Group)` its now something like `updateChanges(myID string, g *v1protocol.Group)`.

What do you think?

protocol/chat.go
Samyoul: I thought this, but a `HasMember()` method also introduces an extra iteration over the `c.Members`

Samyoul: Ah sorry ignore me. I'm looking at the wrong place in the code 🤦

Samyoul: `func (c *Chat) HasMember(memberId string) bool` would be an improvement.

Issue #1944, opened 2020-04-20
Add waku package README.md docs


Samyoul:
# Problem

The `waku` package has no docs in the repo (notably neither do a significant number of the other packages, but I don't want to turn this issue into a formidable mammoth 🐘 😆). This package needs some docs, particularly with the merging of https://github.com/status-im/status-go/pull/1931 and the finalisation of https://github.com/vacp2p/specs/pull/113 and https://github.com/status-im/status-go/pull/1938 there will be an increased necessity for this package to have good documentation.

# Implementation

Write some `README.md` docs for the `waku` package.

# Acceptance Criteria

The docs are a `README.md` file that detail or reference details of the `waku` package's functionality, and the impact of the Waku specs has on the implementation of this package.

# Notes

- See issue https://github.com/status-im/status-go/issues/1929 for details about requirement to make breaking changes to the `waku` package
- See PR https://github.com/status-im/status-go/pull/1931 for details about the first attempt to resolve the above issue, and the rationale for implementing explicit variable init.
- See PR https://github.com/vacp2p/specs/pull/113 for details about updating the Waku specs as a method of fixing the above issue.
- See PR https://github.com/status-im/status-go/pull/1938 for details about implementing the above proposed Waku specs.

# Future Steps

Any significant change to the `waku` package will require updates to these docs.

PR #1949, opened 2020-04-22
Waku package README.md documentation

Samyoul:
Documentation for the `waku` package. See #1944


waku/README.md
```diff
@@ -0,0 +1,287 @@
...
+- [Waku package files](#waku-package-files)
+
+## What is Waku?
+
+Waku is a communication protocol for sending messages between Dapps. Waku is a fork of the [Ethereum Whisper subprotocol](https://github.com/ethereum/wiki/wiki/Whisper), although not directly compatible with Whisper, both Waku and Whisper subprotocols can communicate [via bridging](https://github.com/vacp2p/specs/blob/master/specs/waku/waku-1.md#backwards-compatibility).
+
+Waku was [created to solve scaling issues with Whisper](https://discuss.status.im/t/fixing-whisper-for-great-profit/1419) and [currently diverges](https://github.com/vacp2p/specs/blob/master/specs/waku/waku-1.md#differences-between-shh6-and-waku1) from Whisper in the following ways:
+
+- RLPx subprotocol is changed from `shh/6` to `waku/1`.
+- Light node capability is added.
```
cammellos: I believe light node was also available in whisper https://github.com/ethereum/go-ethereum/blob/510b6f90db406b697610fe0ff2eee66d173673b2/whisper/whisperv6/whisper.go#L291 , not sure if it's exactly the same, but the core concept seems to be (does not forward any envelope)

waku/README.md
```diff
@@ -0,0 +1,287 @@
...
+
+--- 
+
+### `filter.go`
+
+[`filter.go`](./common/filter.go) is home to `Filter{}` which represents a waku filter.
+
+#### Usage
+
+A `status-go` node will install / register filters through RPC calls from a client (eg `status-react`). When the client installs a filter, it will specify 2 things:
```
cammellos: This is a bit more complicated:
it allows to specify a `PoW` ( https://github.com/status-im/status-go/blob/4d00656c41909ccdd80a8a77a0982bd66f74d29e/waku/common/filter.go#L37 ) as well, which is the minimum `PoW` for that filter, a public key for the sender (`Src` https://github.com/status-im/status-go/blob/4d00656c41909ccdd80a8a77a0982bd66f74d29e/waku/common/filter.go#L33 )  and whether we accept `P2P` envelopes on this filter https://github.com/status-im/status-go/blob/4d00656c41909ccdd80a8a77a0982bd66f74d29e/waku/common/filter.go#L38 .
We currently don't use `Src` at all.
I believe we always set `P2P` to `true` as we always want to receive `P2P` envelopes on those filters (only from trusted peers).

And we kind of use .`PoW`:

Whisper forced you to listen to a filter if you wanted to publish to it (for light client nodes I am not sure that makes much sense, but we'd have to look into it a bit deeper).

In `status-react` each user listen to the topic `hash(pk)`, so if I want to send a message to `hash(pk1)` I will need to listen to that topic as well, but I don't really want to receive envelopes. So we listen to that topic, but we set the `PoW` to 1 so that all envelopes are discarded.

Not sure we want to document all of these (we could), but otherwise I would just change `it will specify 2 things` to `it will specify at least 2 things` :)

cammellos (approved):
Awesome work!

waku/README.md
Samyoul: Thanks for the feedback. Would that mean that [the specs](https://github.com/vacp2p/specs/blob/master/specs/waku/waku-1.md#differences-between-shh6-and-waku1) require clarification? As the specs say that light node capability is new.

I'll dig in the code a bit to see where the divergence is.

waku/README.md
Samyoul: It is tricky to know where the line needs to be drawn on the docs. I want to give the read enough information so that they can understand the purpose but at the same time not be so overwhelmed with information that they are confused.

I will at least change the statement that `it will specify at least 2 things` and then detail the optional fields.

waku/README.md
cammellos: oh, yes, looks like the capability is not new actually, nice catch, maybe worth updating the specs if that's the case

waku/README.md
```diff
@@ -0,0 +1,287 @@
...
+
+It is worth noting that each function of `PublicWakuAPI{}` received an unused `context.Context` parameter. This is originally passed in way higher up the food-chain and without significant refactoring is not a simple thing to remove / change. Mobile bindings depend on the ability to pass in a context.
+
+---
+
+### `config.go`
+
+[`config.go`](./config.go) is home to the `Config{}` struct and the declaration of `DefaultConfig`.
+
+`Config{}` is used to initialise the settings of an instantiated `Waku{}`. `waku.New()` creates a new instance of a `Waku{}` and takes a `Config{}` as a parameter, if nil is passed instead of an instance of `Config{}`, `DefaultConfig` is used.
```
andremedeiros, line 102: What are the possible config values and what are they defined as with regards to defaults?

waku/README.md
```diff
@@ -0,0 +1,287 @@
...
+
+### Version `init.go`
+
+`init.go` is home to the version sub-package's initialisation, and is used to initialise struct based variables at runtime. 
+
+---
+
+### Version `message.go`
+
+`message.go` // TODO - I don't really understand what this does.
```
andremedeiros: - [ ] Complete this

waku/README.md
Samyoul: Raised here https://github.com/vacp2p/specs/pull/117 . If you don't mind I will wait to see what the feedback is from this PR.

waku/README.md
cammellos: `message.go` is used for what we call message confirmations:

In the UI when a message is sent, it is marked initially as "Pending" and eventually as "Sent".

In order to trigger the transition from pending to "sent", there are these "message confirmations".

Each peer on receiving a message will send back one of these message responses https://github.com/status-im/status-go/blob/4d00656c41909ccdd80a8a77a0982bd66f74d29e/waku/v1/message.go#L31 .

The waku host will check that this is a mailserver and if it's been selected by the user.
If so the message will be marked as "Sent" in the UI.

https://github.com/status-im/specs/blob/master/docs/stable/3-whisper-usage.md has a section on it.

As a side not, we should not be doing any versioning at this level any more (the struct has a `Version` field), something to remove once we decide to move to `waku/2` for example.

waku/README.md
```diff
@@ -0,0 +1,287 @@
...
+
+Waku is a communication protocol for sending messages between Dapps. Waku is a fork of the [Ethereum Whisper subprotocol](https://github.com/ethereum/wiki/wiki/Whisper), although not directly compatible with Whisper, both Waku and Whisper subprotocols can communicate [via bridging](https://github.com/vacp2p/specs/blob/master/specs/waku/waku-1.md#backwards-compatibility).
+
+Waku was [created to solve scaling issues with Whisper](https://discuss.status.im/t/fixing-whisper-for-great-profit/1419) and [currently diverges](https://github.com/vacp2p/specs/blob/master/specs/waku/waku-1.md#differences-between-shh6-and-waku1) from Whisper in the following ways:
+
+- RLPx subprotocol is changed from `shh/6` to `waku/1`.
+- Light node capability is added.
+- Optional rate limiting is added.
+- Status packet has following additional parameters: light-node, confirmations-enabled and rate-limits
+- Mail Server and Mail Client functionality is now part of the specification.
```
jakubgs, line 20: Worst naming ever, but I guess I will never be able to push through the renaming of this to `history node`.

Samyoul: He who dares wins. I support this change and would lobby for its implementation.

If we can agree on the better name. I would propose `persistence node`, or `echo node`.

One thing I am 100% in agreement with, the Mailserver is a stupid name.

jakubgs: Yeah, makes you think of technology from 1970s, not something developed in 2020s.

waku/README.md
Samyoul: So according to the feedback I got, the Vac team feel that making this change to the specs is not desirable. I will make an amendment to the docs to reflect the discrepancy between Waku light nodes being a direct fork of Whisper light nodes and simultaneously considered a new feature.

waku/README.md
Samyoul: Addressed with https://github.com/status-im/status-go/pull/1949/commits/223f4d26937d41b9f7a1acb6336581f1157fb290

waku/README.md
Samyoul: Addressed with https://github.com/status-im/status-go/pull/1949/commits/493308ebf152a414d66ed962d321f8d690a459ab

waku/README.md
Samyoul: Addressed with https://github.com/status-im/status-go/pull/1949/commits/39bc04b1958a3bc0b7161613608ee572bab8aa8a

waku/README.md
Samyoul: Resolved with https://github.com/status-im/status-go/pull/1949/commits/bc17559d4284f4d3bbdee51d6dc9dd7a63bacf7e

Samyoul: @cammellos @andremedeiros 

**I'm going to implement the following:**

- Move send bundle to Peer
- change v*/message.go to v*/message_response.go - I know that I originally merged message response functionality into message, but after the `waku/1` refactor it makes sense to move it back.

**I've got some questions:**

- Should `MultiVersionResponse` and `Version1MessageResponse` be merged into a single struct MessageResponse? Or at least renamed to be more obvious what they are.
- Should the structs in v*/message be unexported as they are only consumed inside the version subpackage?
- Should Waku "message response" be renamed as "message confirmation"? It is called a message confirmation in the docs, but also a message response, it makes more sense as confirmation because a message response can be anything that responds to a message including a confirmation, a confirmation is explicitly a confirmation.
- Should MessagesResponse be renamed to MessageResponse
- Should we track new features for the next Waku version? If so where would be a good place?

See :
>Versioning at the MessageResponse level (see the struct Version field) should be phased out and defer to the subpackage's version number. Consider removal once we decide to move to a new major Waku version (i.e. waku/2).

cammellos: > Should MultiVersionResponse and Version1MessageResponse be merged into a single struct MessageResponse? Or at least renamed to be more obvious what they are.

I would say so, if feasible one single struct makes sense.

> Should the structs in v*/message be unexported as they are only consumed inside the version subpackage?

Yes

> Should Waku "message response" be renamed as "message confirmation"?

I'd think so, not sure we use it for anything other than messages confirmations

> Should we track new features for the next Waku version? If so where would be a good place?

I'd think that would be the specs repo, as any new feature should be driven by the specs I take, what do you think?

Samyoul: I agree. I was thinking about making the renaming and merging changes now in the open tidy PR, but on reflection I think that we need consensus on the use of terminology and the direction we want to take improvements to the specs.

I will open a spec issue on this subject once I'm done [updating the status specs](https://github.com/status-im/specs/issues/61).

waku/README.md
Samyoul: @jakubgs this may interest you https://discuss.status.im/t/wherefore-art-thou-mailserver-treatise-on-waku-terminology/1664

Samyoul:
@cammellos, @jakubgs, @andremedeiros Would you mind giving this PR a final review. I believe it is ready for merging, but I'd like the new updates to be checked.

cammellos:
looks good, thanks!

PR #1950, opened 2020-04-23
Refactor/waku.doc

Samyoul:
# Why make the change?

`waku/doc.go` is a hangover from https://github.com/ethereum/go-ethereum/blob/master/whisper/whisperv6/doc.go and has become a bit of a dumping ground for any miscellaneous waku functionality or structs.

Additionally [the golang docs convention](https://blog.golang.org/godoc) reserves the `doc.go` file for housing only docs.

>another convention for packages that need large amounts of introductory documentation: the package comment is placed in its own file, doc.go, which contains **only** those comments and a package clause.

# What has changed?

- Removal of `doc.go` and `doc_test.go`
- Moved package consts to `const.go`
- Moved `EnvelopeError{}` to `envelope.go` 
- Renamed `mailserver_response.go` to `mailserver.go`
- Moved `MailServer{}` interface to `mailserver.go`
- Moved `MailServerResponse{}` to `mailserver.go`
- Moved `MessagesRequest{}`, `MessagesResponse{}`, `MultiVersionResponse{}`, `Version1MessageResponse{}` and `NewMessagesResponse()` to `message.go`
- Moved `TestEncodeDecodeVersionedResponse{}` to `message_test.go`
- Moved `RateLimits{}` to `rate_limiter.go`

The same changes were also applied to the `whisper` package.


waku/message_test.go
```diff
@@ -182,9 +184,6 @@ func TestMessageSeal(t *testing.T) {
 	params.TTL = 1
 
 	env := NewEnvelope(params.TTL, params.Topic, msg, time.Now())
-	if err != nil {
-		t.Fatalf("failed Wrap with seed %d: %s.", seed, err)
-	}
```
Samyoul, line 187: Just so I'm not trying to sneak this through, I want to point out the I've removed this because the condition is always false because err is always nil. The err is tested on line 181 and doesn't change after that check and before the check I've removed.

waku/message_test.go
```diff
@@ -495,3 +494,15 @@ func TestValidateAndParseSizeOfPayloadSize(t *testing.T) {
...
 	}
 }
+
+func TestEncodeDecodeVersionedResponse(t *testing.T) {
+	response := NewMessagesResponse(common.Hash{1}, []EnvelopeError{{Code: 1}})
+	b, err := rlp.EncodeToBytes(response)
+	require.NoError(t, err)
+
+	var mresponse MultiVersionResponse
+	require.NoError(t, rlp.DecodeBytes(b, &mresponse))
```
Samyoul, line 505: I've also changed `bytes` to `b` to remove the type name / variable name collision.

cammellos (approved):
Linter fails, but looks good!

Samyoul:
File formatting lint fail. The worst of all lint fails 😭

waku/message_test.go
```diff
@@ -26,9 +26,11 @@ import (
 	"testing"
 	"time"
```
Samyoul, line 28: @cammellos the linter is complaining that this line is failing on file formatting grounds https://ci.status.im/blue/organizations/jenkins/status-go%2Fprs%2Ftests/detail/PR-1950/1/pipeline . I don't see anything at all wrong with it. 

Are you able to see any thing? Does your `gofmt` give anything different?

Samyoul:
@cammellos [codeclimate is complaining](https://codeclimate.com/github/status-im/status-go/pull/1950) also because there is duplication of code between Whisper and Waku. I know we are in the middle of refactoring everything, is it possible to ignore these warnings from codeclimate and address it in a separate PR?

cammellos:
@Samyoul yes, codeclimate is not required, so they can safely be ignored.
I tried  `goimports -w message_test.go` but did not show any diff, maybe some difference in versions?

jakubgs (approved):
LGTM, but do we care that `whisper/doc.go` has the same issue?

Samyoul:
>do we care that `whisper/doc.go` has the same issue?

Yes, I care. @jakubgs would it make sense to update this in a new PR?

jakubgs:
Don't see the point of a separate PR. Separate commit is just fine.

Samyoul:
@cammellos , I'm going to restart the tests and see if that makes any kind of difference (it has before in race condition fails).

I'm working on the `whisper/doc.go` refactor, once done I'll commit and see if there's any change to the mood the linter is in 😆 .

jakubgs (approved):
Nice.

waku/message_test.go
Samyoul: resolved with https://github.com/status-im/status-go/pull/1950/commits/04413b9e31e79230f55531cd34cd1481225da993

PR #1956, opened 2020-04-29
Implement waku/1

cammellos:
### Why make the change?
This implements waku/1 which is a breaking change as waku/0 diverged
from the specs.

Tested against v1.3 both in waku/0 and waku/1 mode.

### What has changed?

* Added v1 namespace
* Changed the testing code to test from the outside rather than stubbing
p2p specific messages, so that any version specific code is under `vx`
namespaces
* Split waku vs waku/x test code
* Added a test suite that can be used for different versions

### Things still to do (probably in a separate PR)
I kept the tests we have for the versioned aspect of waku, probably we
want to add some or change other to provide the optimal test coverage, but I would do this in a separate PR.

There's also a few places where we don't check for errors in the tests as it's a bit tricky (when we run the main loop), that probably need sorting out at some point , but I didn't want to complicate the tests too much, so open to suggestions.

### Things still to do in this PR
We talked about factory pattern, currently we don't have to look into `caps` directly as we can pass directly a version and an handler to `devp2p` and it will call the right one, so it seems like it's the best place to make that decision:

```
waku.protocols = []p2p.Protocol{{
Name: v0.Name,
Version: uint(v0.Version),
Length: v0.NumberOfMessageCodes,
Run: waku.HandlePeerV0,
NodeInfo: func() interface{} {
return map[string]interface{}{
"version": v0.VersionStr,
"maxMessageSize": waku.MaxMessageSize(),
"minimumPoW": waku.MinPow(),
}
},
},
{
Name: v1.Name,
Version: uint(v1.Version),
Length: v1.NumberOfMessageCodes,
Run: waku.HandlePeerV1,
NodeInfo: func() interface{} {
return map[string]interface{}{
"version": v1.VersionStr,
"maxMessageSize": waku.MaxMessageSize(),
"minimumPoW": waku.MinPow(),
}
},
},
}
```

Where `handlePeerVx` are just calling `HandlePeer` with the right one.

```
func (w *Waku) handlePeerV0(p2pPeer *p2p.Peer, rw p2p.MsgReadWriter) error {
return w.HandlePeer(v0.NewPeer(w, p2pPeer, rw, w.logger.Named("waku/peerv0")), rw)
}

func (w *Waku) handlePeerV1(p2pPeer *p2p.Peer, rw p2p.MsgReadWriter) error {
return w.HandlePeer(v1.NewPeer(w, p2pPeer, rw, w.logger.Named("waku/peerv1")), rw)
}
```

I looked into passing a func to HandlePeer, but it seems like it's more verbose then just having separate functions.

We talked about a factory pattern, how do you think we could/should fit this into the code @Samyoul ?


waku/simulation_test.go
```diff
@@ -397,31 +394,6 @@ func sendMsg(t *testing.T, expected bool, id int) {
...
-		t.Fatalf("failed Wrap with seed %d.", seed)
-	}
-
-	p := v0.NewPeer(nil, nil, nil, nil)
-	p.Mark(env)
-	if !p.Marked(env) {
-		t.Fatalf("failed mark with seed %d.", seed)
-	}
-}
-
```
Samyoul, line 412: Removing whole tests make me nervous 😄 . I see this test in https://github.com/status-im/status-go/pull/1956/files#diff-da14d6cb8f94efda56c670a48d575551R76 but I don't see an equivalent test in `waku/v0`

waku/v1/message.go
```diff
@@ -0,0 +1,55 @@
...
+// MultiVersionResponse allows to decode response into chosen version.
+type MultiVersionResponse struct {
+	Version  uint
+	Response rlp.RawValue
+}
+
+// DecodeResponse1 decodes response into first version of the messages response.
+func (m MultiVersionResponse) DecodeResponse1() (resp common.MessagesResponse, err error) {
+	return resp, rlp.DecodeBytes(m.Response, &resp)
+}
```
Samyoul, line 22: I know the original that this was copied from doesn't have any tests, but this could be a good opportunity to introduce some / one.

waku/v1/message.go
```diff
@@ -0,0 +1,55 @@
...
+	err = rw.WriteMsg(p2p.Msg{
+		Code:    messagesCode,
+		Size:    uint32(len(data)),
+		Payload: bytes.NewBuffer(data),
+	})
+	if err != nil {
+		return
+	}
+	return crypto.Keccak256Hash(data), nil
+}
```
Samyoul, line 55: I know the original that this was copied from doesn't have any tests, but this could be a good opportunity to introduce some / one.

waku/v1/statusoptions.go
```diff
@@ -0,0 +1,212 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/ethereum/go-ethereum/rlp"
+
+	"github.com/status-im/status-go/waku/common"
+)
+
+// statusOptionKey is a current type used in StatusOptions as a key.
+type statusOptionKey uint64
```
Samyoul, line 18: 🎉

waku/v1/statusoptions.go
```diff
@@ -0,0 +1,212 @@
...
+
+		keys := strings.Split(rlpTag, "=")
+
+		if len(keys) != 2 || keys[0] != "key" {
+			panic("invalid value of \"rlp\" tag, expected \"key=N\" where N is uint")
+		}
+		key, err := strconv.ParseUint(keys[1], 10, 64)
+		if err != nil {
+			panic("could not parse \"rlp\" key")
+		}
```
Samyoul, line 91: 🎉

Samyoul:
>We talked about a factory pattern, how do you think we could/should fit this into the code @Samyoul ?

Let me look over the code and think about the best way to implement this.

jakubgs:
I wish this was two commits. One copying Waku v0 stuff to v1, and another changing the v1 copy.
With as it is right now it's kinda hard to tell what changed.

cammellos:
@jakubgs yes, that's a good idea, I can split it off in 2 commits, the changes are minimal

cammellos:
@jakubgs done, the first commit wire things up, and the second implements the specific `waku/1` changes, should be clearer now.

waku/v1/message.go
jakubgs: At least in a separate commit. Separate PR would be better.

waku/simulation_test.go
cammellos: Yes, I only moved it to `v1`, I will add it to `v0` as well

cammellos:
@Samyoul @jakubgs I have added the tests (last commit, `message.go` had actually one test that was testing the two main functions, I have added one for `sendBundle`). It's ready to be reviewed again I believe.
Once is ok we can merge https://github.com/status-im/status-go/pull/1957 first and I can rebase against it.

Samyoul:
@cammellos Ok, I'm taking a look now. I've got some flaky test issues also ... I wonder if they are the same problem.

jakubgs (approved):
I'm no specialist in Waku or Whisper but I didn't see anything jarring.

waku/waku.go
```diff
@@ -151,28 +152,37 @@ func New(cfg *Config, logger *zap.Logger) *Waku {
...
+			Length:  v1.NumberOfMessageCodes,
+			Run:     waku.handlePeerV1,
+			NodeInfo: func() interface{} {
+				return map[string]interface{}{
+					"version":        v1.VersionStr,
+					"maxMessageSize": waku.MaxMessageSize(),
+					"minimumPoW":     waku.MinPow(),
+				}
+			},
+		},
```
Samyoul, line 180: My suggestion of using factory pattern was based on a fundamental misunderstanding of your eventual approach. Now looking at this final implementation I can see that the version will be selected by the consuming code.

Do we want to introduce in another PR the selection functionality (the code that selects the highest mutual version between peers)?

Samyoul (approved):
Looks good. This package will be so cool once it is finished with it 🌟

waku/waku.go
cammellos: I think that's already handled by devp2p, it will already do that behind the scene and call the right handler (specified in the `Run` field)

Samyoul: Oooh, that is fancy! Cool.

Samyoul:
https://github.com/status-im/status-go/pull/1957 is merged

Samyoul:
@cammellos can I make a quick change suggestion?

Filename change:

`common/helper.go` > `common/helpers.go` (This is my fault and a tpyo that was implemented on my suggestion)
`v*/statusoptions.go` > `v*/status_options.go` (obviously this applies to `v0` and `v1`)

waku/v0/message_test.go
```diff
@@ -40,3 +41,29 @@ func TestEncodeDecodeVersionedResponse(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, response.Response.Hash, v1resp.Hash)
 }
+
+func TestSendBundle(t *testing.T) {
+	rw1, rw2 := p2p.MsgPipe()
+	defer func() { handleError(t, rw1.Close()) }()
+	defer func() { handleError(t, rw2.Close()) }()
```
Samyoul, line 48: These look nicer than the way I was implementing the same functionality. Inline is so fine 😎

services/wakuext/api_test.go
```diff
@@ -303,8 +303,8 @@ func (s *WakuNodeMockSuite) SetupTest() {
 	pkey, err := crypto.GenerateKey()
 	s.Require().NoError(err)
 	node := enode.NewV4(&pkey.PublicKey, net.ParseIP("127.0.0.1"), 1, 1)
-	peer := p2p.NewPeer(node.ID(), "1", []p2p.Cap{{"shh", 6}})
 	rw1, rw2 := p2p.MsgPipe()
+	peer := v0.NewPeer(w, p2p.NewPeer(node.ID(), "1", []p2p.Cap{{"shh", 6}}), rw2, nil)
```
Samyoul, line 307: Should `"shh"` be `"waku"`?

cammellos:
Tested against nimbus, release app (1.3, waku/0) and PR running this code (waku/1)

PR #1967, opened 2020-05-12
Re-introduce deprecated method of requesting messages

cammellos:
During the move to waku/1 we removed handing of a deprecated format.
Turns out that no client is using the new format and the only live
client is using the deprecated format.
This commit re-introduces the functionality.


corpetty:
How do we properly deprecate it from here?

cammellos:
@corpetty unclear, I am not sure why it was deprecated, it's a bit less explicit but the other method seems to be basically equivalent.
In general, all the versioning/deprecations go with waku versioning, `waku/2` will drop support for the deprecated method for example.

jakubgs (approved):

Samyoul (approved):

Samyoul:
I can't see where all this was removed, could you provide the link to the commit that took it out?

I can see https://github.com/status-im/status-go/commit/aa7f5915873b8cd7a0a06662fa80dbf93aeec32f#diff-1769c0979f800a1b8cc0f3ed15e9e1bbL1237-L1242

cammellos:
> I can't see where all this was removed, could you provide the link to the commit that took it out?
> 
> I can see [aa7f591#diff-1769c0979f800a1b8cc0f3ed15e9e1bbL1237-L1242]

Yes, that's the commit, I have added a link to the commit description.
https://github.com/status-im/status-go/commit/aa7f5915873b8cd7a0a06662fa80dbf93aeec32f#diff-ea5e44cf3db4a12b3a2a246c7fa39602R290

PR #1969, opened 2020-05-13
Add support for images

cammellos:
## Why make the changes?

Adding support for images in the status-protocol as per https://github.com/status-im/specs/pull/107

## What has changes?

- Added support for parsing images and sending images in `protocol/`
- lowered the `PoWTarget` and `MinimumPoW`
- Refactored a few methods from `message_processor.go`
- Removed incentivisation service (wasn't used anymore and was relying on some duplicated PoW const)

I have split the changes in commit, there's a better description in each commit message.


Samyoul:
Really great work! I'll give this a full review later this afternoon.

VERSION
```diff
@@ -1 +1 @@
-0.53.1
+0.53.2
```
Samyoul: https://github.com/status-im/status-go/pull/1969/commits/bb7c6f049059aafdf647e628e6cb33766202217d seems to suggest that this change would be a kind of breaking change. Should this be at least a minor version change?

params/config.go
```diff
@@ -633,6 +602,16 @@ func (c *NodeConfig) UpdateWithMobileDefaults() {
 	if c.APIModules == "" {
 		c.APIModules = "net,web3,eth"
 	}
+
+	// Override defaultMinPoW passed by the mobile
```
Samyoul:
```suggestion
	// Override defaultMinPoW passed by the client
```

Samyoul:
@cammellos I'm sad that the test image file is not a little kitten 🐱

protocol/images/type.go
```diff
@@ -0,0 +1,43 @@
...
+		return protobuf.ImageMessage_JPEG
+	} else if png(buf) {
+		return protobuf.ImageMessage_PNG
+	} else if gif(buf) {
+		return protobuf.ImageMessage_GIF
+	} else if webp(buf) {
+		return protobuf.ImageMessage_WEBP
+	}
+
+	return protobuf.ImageMessage_UNKNOWN_IMAGE_TYPE
```
Samyoul: Although functionally identical to an if/else chain a `switch` is more idiomatic for implementing a choice from a list of options.

```suggestion
	switch {
	case jpeg(buf):
		return protobuf.ImageMessage_JPEG
	case png(buf):
		return protobuf.ImageMessage_PNG
	case gif(buf):
		return protobuf.ImageMessage_GIF
	case webp(buf):
		return protobuf.ImageMessage_WEBP
	default:
		return protobuf.ImageMessage_UNKNOWN_IMAGE_TYPE
	}
```

protocol/message.go
```diff
@@ -211,6 +220,45 @@ func isRTL(s string) bool {
...
+		mime = "png"
+	case protobuf.ImageMessage_JPEG:
+		mime = "jpeg"
+	case protobuf.ImageMessage_WEBP:
+		mime = "webp"
+	case protobuf.ImageMessage_GIF:
+		mime = "gif"
+	default:
+		return errors.New("image format not supported")
+	}
```
Samyoul: Is it possible to have this code be a function? That would allow for this functionality to be testable.

Something like

```go
func getImageMessageMIME(i protobuf.ImageMessage) (string, error) {
	switch i.Type {
	case protobuf.ImageMessage_PNG:
		return "png", nil
	case protobuf.ImageMessage_JPEG:
		return "jpeg", nil
	case protobuf.ImageMessage_WEBP:
		return "webp", nil
	case protobuf.ImageMessage_GIF:
		return "gif", nil
	default:
		"", return errors.New("image format not supported")
	}
}
```

Usage:

```go
mime, err := getImageMessageMIME(image)
```

protocol/message_processor.go
```diff
@@ -428,12 +428,23 @@ func messageSpecToWhisper(spec *encryption.ProtocolMessageSpec) (*types.NewMessa
...
+// calculatePoW returns the PoWTarget to be used.
+// We check the size and arbitrarely set it to a lower PoW if the packet is
+// greater than 50KB. We do this as the defaultPoW is to high for clients to send
+// large messages.
+func calculatePoW(payload []byte) float64 {
+	if len(payload) > largeSizeInBytes {
+		return whisperLargeSizePoW
+	}
+	return whisperDefaultPoW
+}
```
Samyoul: I know this is necessary. But is this open to abuse? I could negate the PoW limitations applied to text by sending huge text messages. Potentially only theoretical.

Samyoul (approved):
Looks good, just a few suggestions.

VERSION
cammellos: Yes, good point, hopefully no one notices that is breaking, but it is technically breaking as file greater than `50K` (if ever sent), won't be received anymore.

protocol/images/type.go
cammellos: didn't know you can have function in a case statement!

protocol/message_processor.go
cammellos: You could, but you would not have much benefit from it, as the message just won't be received by older clients and will be received by newer, which probably is not achieving much.

The bit that is open for abuse is the fact that we lowered the PoW for the whole fleet, so you could instead send  messages quicker than before, as they would need a lower PoW, and those would be received by newer clients, so spamming is easier with this changes.

Having said that, with the PoW we had before was already trivial sending many messages in quick succession (you can send 500 in a few seconds), so yes, it's making things worse, but things were not great before :) 
Spamming is something we haven't quite figure out, but before deploying this probably best to run some tests with the differences.

protocol/images/type.go
Samyoul: Case statements in Go compile down into if else statements, so you can run anything that results in a bool.

protocol/message_processor.go
Samyoul: Ok, thank you. That is a helpful insight.

node/geth_node.go
```diff
@@ -497,49 +487,6 @@ func createWakuService(ctx *node.ServiceContext, wakuCfg *params.WakuConfig, clu
 	return w, nil
 }
 
-// activateIncentivisationService configures Whisper and adds it to the given node.
-func activateIncentivisationService(stack *node.Node, config *params.NodeConfig) (err error) {
```
jakubgs, line 501: Why are we removing this tho? I don't see this mentioned in PR description.

cammellos: Last point in the PR description

params/config.go
```diff
@@ -633,6 +602,16 @@ func (c *NodeConfig) UpdateWithMobileDefaults() {
...
 		c.APIModules = "net,web3,eth"
 	}
+
+	// Override defaultMinPoW passed by the client
+	if c.WakuConfig.Enabled {
+		c.WakuConfig.MinimumPoW = WakuMinimumPoW
+	}
+
+	if c.WhisperConfig.Enabled {
+		c.WakuConfig.MinimumPoW = WhisperMinimumPoW
```
jakubgs: I'm confused, why are we setting `WakuConfig` to a `WhisperMinimumPoW`?

cammellos: ouch, that's a good reason to be confused :) well spotted

params/defaults.go
```diff
@@ -22,13 +22,13 @@ const (
 	DefaultGas = 180000
 
 	// WhisperMinimumPoW amount of work for Whisper message to be added to sending queue
-	WhisperMinimumPoW = 0.001
+	WhisperMinimumPoW = 0.000002
```
jakubgs, line 26: Would it make sense to add a comment here about us enforcing a minimum?
https://github.com/status-im/status-go/blob/a21f8a39bc4b786999011f88f519bf518fbfa299/params/config.go#L84-L85

params/defaults.go
```diff
@@ -22,13 +22,13 @@ const (
...
 
 	// WhisperTTL is time to live for messages, in seconds
 	WhisperTTL = 120
 
 	// WakuMinimumPoW amount of work for Whisper message to be added to sending queue
-	WakuMinimumPoW = 0.001
+	WakuMinimumPoW = 0.000002
```
jakubgs, line 33: Same.

params/defaults.go
cammellos: added

jakubgs: That's nice, but I meant in `params/config.go`. Where most people would look for explanations of options.

cammellos: added in the right place as well :)

jakubgs: Cheers!

protocol/message_processor.go
```diff
@@ -428,12 +428,23 @@ func messageSpecToWhisper(spec *encryption.ProtocolMessageSpec) (*types.NewMessa
...
-		PowTarget: whisperPoW,
+		PowTarget: calculatePoW(payload),
 		PowTime:   whisperPoWTime,
 	}
 	return newMessage, nil
 }
 
+// calculatePoW returns the PoWTarget to be used.
+// We check the size and arbitrarely set it to a lower PoW if the packet is
+// greater than 50KB. We do this as the defaultPoW is to high for clients to send
```
jakubgs:
```suggestion
// greater than 50KB. We do this as the defaultPoW is too high for clients to send
```

protocol/messenger.go
```diff
@@ -1288,7 +1291,7 @@ func (m *Messenger) dispatchPairInstallationMessage(ctx context.Context, spec *R
 	var err error
 	var id []byte
 
-	id, err = m.processor.SendPairInstallation(ctx, &m.identity.PublicKey, spec.Payload, spec.MessageType)
+	id, err = m.processor.SendPairInstallation(ctx, &m.identity.PublicKey, spec)
```
jakubgs, line 1298: I like how you reduced number of arguments for this and `SendPrivate`.

jakubgs (approved):
Nice.

PR #1982, opened 2020-05-22
Load identicon & alias in chat.

cammellos:
### Why make the changes?
Mainly performance, those fields are almost always present in the
database but they are re-calculated on load by the client as it does not
have necessarily access to it.

### What has changed?

* Remove `_legacy` persistence namespaces as it's a vestige of the
initial move frmo status-react to status-go
* Pulling chats is now a join with contacts to add contact & alias


jakubgs (approved):
I wish the renaming was done in a separate commit, but that's just a nitpick. I see no issues here.

Samyoul (approved):
Looks good. No issues at all.

PR #1983, opened 2020-06-02
Generate bindings dynamically

cammellos:
Before we had two directories `lib/` and `mobile/` that generated
respectively the bindings for desktop and ios/android.

This needed to be kept in sync and there was a fair amount of code
duplication, plus some missing methods on one or the other side.

I have made a change so the whole `lib/` namespace is generated by
parsing the `AST` of `mobile`, and bindings are generated before
compiling.

Operationally is fairly crude, it creates a file in the build directory, and passes it to go build.


Makefile
```diff
@@ -111,8 +111,10 @@ statusgo-ios: ##@cross-compile Build status-go for iOS
 	@echo "iOS framework cross compilation done in build/bin/Statusgo.framework"
 
 statusgo-library: ##@cross-compile Build status-go as static library for current platform
+	mkdir -p $(GOBIN)/statusgo-lib
+	go run cmd/library/* > $(GOBIN)/statusgo-lib/main.go
```
jakubgs: Maybe a brief comment referencing `cmd/library/README.md`?

jakubgs (approved):
Absolutely disgusting, but if it works it works.

Can we get a `cmd/library/README.md` explaining what this does and why for any poor souls that stumble upon this?

Also, how do we know that the resulting file is valid Go code? Some sanity check with `gofmt` maybe?

cmd/library/const.go
```diff
@@ -0,0 +1,15 @@
+package main
+
+const prelude = `package main
+// #include <stdlib.h>
```
jakubgs, line 6: Maybe add a comment here:
```go
// generated by cmd/library tool to keep in sync with mobile
```

cammellos:
@jakubgs done, I have added a step in the jenkins file for linux that generates and compiles the bindings, not sure if it's the best place, but essentially we want to generate the bindings on CI and fail if they don't compile.

jakubgs (approved):

PR #1985, opened 2020-06-03
use EncodeToString to print []byte variables

jakubgs:
If we don't use it we get shit like:
```
[mailserver:DeliverMail] request failed validaton
  peerID="[113 189 115 37 242 24 204 13 148 190 2 73 121 249 83 175 21 18 22 254 138 157 128 45 142 254 46 252 132 209 109 135]"
  requestID=0x1b2574d73c33c6c68aadfd81746f195d629edeed403403b8e55fcd6e0e1bfde7
  err="query range is invalid: from > to (4165960798 > 3600145796)"
```
In logs, which not conducive to searching logs by peer ID.

### DIFF
--- a/mailserver/mailserver.go
+++ b/mailserver/mailserver.go
@@ -138,7 +138,7 @@ func (s *WhisperMailServer) DeliverMail(peerID []byte, req *whisper.Envelope) {
 		log.Debug(
 			"[mailserver:DeliverMail] failed to decode request",
 			"err", err,
-			"peerID", peerID,
+			"peerID", types.BytesToHash(peerID),
 			"requestID", req.Hash().String(),
 		)
 		payload, err = s.decompositeRequest(peerID, req)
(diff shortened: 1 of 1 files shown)


cammellos (approved):

Samyoul (approved):

jakubgs:
I found a better way of doing this using the already provided way:
https://github.com/status-im/status-go/blob/a1f2761fc055b2e15a415a89e0cd4aa93abc4ecb/eth-node/types/hash.go#L82-L88
Which provides it's own `String()` function:
https://github.com/status-im/status-go/blob/a1f2761fc055b2e15a415a89e0cd4aa93abc4ecb/eth-node/types/hash.go#L51-L55
My understanding it that `String()` will be called implicitly when being formatted for printing.

Samyoul (approved):
As long as your printer expects a type that implements the stringer interface then you are all good.

Native `fmt.` printers will try to call the `String()` func if the `%s` type is declared.

jakubgs:
Btw, this section is weird:
https://github.com/status-im/status-go/blob/a1f2761fc055b2e15a415a89e0cd4aa93abc4ecb/mailserver/mailserver.go#L136-L156
Seems like the two `if`s and error prints are redundant...

jakubgs:
Fuck it, waku just uses `decodeRequest`, we don't care about whisper anymore. Fuck whisper.

PR #1986, opened 2020-06-04
Adjust index to use `local_chat_id`

cammellos:
Create different index for filtering
    
The index for message was fairly inefficient as it was only using the
    cursor, as it was referring to the old `chat_id` field.
    
This meant that newer messages would be fetched much faster then older
    messages.
    
The index has been changed so that now it includes `local_chat_id`
    (which is currently used for filtering), and not using `hide`.
    
The reason being is that `hide` is a low cardinality index, so there's
    no performance benefit to have it in, also it's mostly ignored by the
    query planner.
    
This commit also adds the missing migrations, we generated the file, but
    the source was missing, probably I forgot to add them in a rebase. They
    have been generated from the migration file, using `RestoreAsset`.


Samyoul (approved):
Looks good

Samyoul:
Should these down migration be empty?

[1588665364_add_image_data.down.sql](https://github.com/status-im/status-go/blob/696553bee6aa24aceafc82a959986c289a767544/protocol/migrations/sqlite/1588665364_add_image_data.down.sql)
[1589365189_add_pow_target.down.sq](https://github.com/status-im/status-go/blob/696553bee6aa24aceafc82a959986c289a767544/protocol/migrations/sqlite/1589365189_add_pow_target.down.sql)

cammellos:
> Should these down migration be empty?


Kind of, basically sqlite does not have a drop column statement, you can only add to an existing table.

The proper way to handle migrations would be to create a temporary table, copy over all the data, drop the original table, and rename the new table to the old one. 
Similarly could be done for the `down` migration.

That's fairly heavy handed for our use case, and operationally it might cause more issues. 
We never run down migrations, because we don't have control over the deployment of our code (users might install the app, or might not), so we would always roll forward.

So for simplicity sake, it's easier to leave down migrations empty, although I'd be ok with writing a proper migration, if you feel is better (I don't have a strong opinion myself).

Samyoul:
>We never run down migrations, because we don't have control over the deployment of our code (users might install the app, or might not), so we would always roll forward.

This is a very good point. Hmm, I suppose the only benefit to down migrations would be if the user was downgrading their application ... although the down migrations can only exist in the versions that introduced the up migration, so wouldn't get used in a downgrade either.

PR #1987, opened 2020-06-04
expose getContactByID method

cammellos:
Used by the desktop team.

### DIFF
--- a/protocol/messenger.go
+++ b/protocol/messenger.go
@@ -1234,14 +1234,11 @@ func (m *Messenger) Contacts() []*Contact {
 }
 
 // GetContactByID assumes pubKey includes 0x prefix
-func (m *Messenger) GetContactByID(pubKey string) (*Contact, error) {
+func (m *Messenger) GetContactByID(pubKey string) *Contact {
 	m.mutex.Lock()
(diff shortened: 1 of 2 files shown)


protocol/messenger.go
```diff
@@ -1234,14 +1234,11 @@ func (m *Messenger) Contacts() []*Contact {
...
+func (m *Messenger) GetContactByID(pubKey string) *Contact {
 	m.mutex.Lock()
 	defer m.mutex.Unlock()
-	contact, ok := m.allContacts[pubKey]
-	if !ok {
-		return nil, errors.New("no contact found")
-	}
-	return contact, nil
+
+	return m.allContacts[pubKey]
```
jakubgs, line 1241: So why exactly are we dropping the error here?

cammellos: Before we returned error if the contact was not in memory (i.e not in database), desktop prefers not to return an error in this case, and just return `null`, and it's only used by desktop

Samyoul (approved):

protocol/messenger.go
jakubgs: Makes sense.

jakubgs (approved):

PR #1989, opened 2020-06-08
Implement rate limiting by bytes

cammellos:
##  Why make this change?
    
Currently we only rate limit by number of messages.
This works well if we want to limit a large amount of small messages but
    breaks down if sending a smaller amount of large messages.
    
This commit extends the current code to limit by size as well, setting
 the default to 1MB per second, which should be enough.
    
 ##  What has changed?
    
 - Rate limiter for waku only

Feature: https://github.com/status-im/status-react/issues/10775
Specs: https://github.com/vacp2p/specs/pull/141


waku/common/rate_limiter.go
```diff
@@ -101,26 +102,36 @@ func (h *DropPeerRateLimiterHandler) ExceedIPLimit() error {
...
 
 var defaultPeerRateLimiterConfig = PeerRateLimiterConfig{
-	LimitPerSecIP:      10,
-	LimitPerSecPeerID:  5,
-	WhitelistedIPs:     nil,
-	WhitelistedPeerIDs: nil,
+	PacketLimitPerSecIP:     10,
+	PacketLimitPerSecPeerID: 5,
+	BytesLimitPerSecIP:      1048576, // 2MB
+	BytesLimitPerSecPeerID:  1048576, // 2MB
```
jakubgs: 1MB, not 2MB.

waku/common/rate_limiter.go
```diff
@@ -228,30 +243,45 @@ func (r *PeerRateLimiter) Decorate(p RateLimiterPeer, rw p2p.MsgReadWriter, runL
...
 	if enodeIDSliceContains(r.whitelistedPeerIDs, id) {
 		return false
 	}
-	return r.peerIDThrottler.Halt(id.String(), 1, r.LimitPerSecPeerID)
+
+	var packetLimiterResponse bool
+	var bytesLimiterResponse bool
+
+	if r.PacketLimitPerSecPeerID != 0 {
+		packetLimiterResponse = r.packetPeerIDThrottler.Halt(id.String(), 1, r.PacketLimitPerSecPeerID)
```
jakubgs: Do we even need separate `Throttler` intances for IDs and IPs? They are different identifiers, they wouldn't clash.
I guess it makes the code a bit cleaner, and possible makes key searches shorter.

jakubgs (approved):
I'm wondering if throttling by Peer ID has any point considering how easy it is for spammer to change that.

Also, how would throttling appear to an App user? Do we have any handling on App side for that?

params/config.go
```diff
@@ -177,13 +177,21 @@ type WakuConfig struct {
...
-	RateLimitPeerID int64
+	PacketRateLimitPeerID int64
+
+	// BytesRateLimitIP sets the limit on the number of bytes per second
+	// from a given IP.
+	BytesRateLimitIP int64
+
+	// BytesRateLimitPeerID sets the limit on the number of bytes per second
+	// from a given peer ID.
+	BytesRateLimitPeerID int64
```
Samyoul, line 194: Just a minor thing, but I really like this spacing style for consts, vars and fields. 👍

Samyoul (approved):
Looks really good.

waku/common/rate_limiter.go
cammellos: Good idea

cammellos:
> I'm wondering if throttling by Peer ID has any point considering how easy it is for spammer to change that.
> 
> Also, how would throttling appear to an App user? Do we have any handling on App side for that?

Currently we don't really handle it, limits are set so high that we should not be reaching them anyway, but that's something we should work on.

Regarding spam, this is mostly just for preventing a single user with the status app to spam images, unlikely to stop anyone more sophisticated.

waku/common/rate_limiter.go
```diff
@@ -190,7 +200,7 @@ func (r *PeerRateLimiter) Decorate(p RateLimiterPeer, rw p2p.MsgReadWriter, runL
 			if p != nil {
 				peerID = p.ID()
 			}
-			if halted := r.throttlePeer(peerID); halted {
+			if halted := r.throttlePeer(peerID, packet.Size); halted {
```
andremedeiros, line 203:
```go
if r.throttlePeer(peerID, packet.Size) {
```

Wouldn't this be idiomatic?

andremedeiros (approved):

andremedeiros:
Can this be merged?

cammellos:
I think it's fine to merge, yes. This needs to be tested by qa before going live, or rate limiting should be set to an higher value.
Currently is 1mb per second , which we should not be hitting, but not sure for our own nodes.

Issue #1937, opened 2020-04-14
Compress public key for chat


hesterbruikman:
# Problem
Chat key is currently shown (as used) as the public key in full. This is unnecessarily daunting. Compressing the public key used for chat can make it less daunting, more space efficient in the UI as well as reducing file size and visual complexity of the QR code.

# Implementation
Implement two functions:
- decompress public key - anywhere the full public key is required 
- request compressed key - places where the key is shown or shared in the UI

Use 02 and 03 for compressed and maintain 04 for decompressed key

status-react should always receive the decompressed key (no changes to current implementation) unless the compressed key is requested. Changes for this will be documented in a separate issue and include:
- `Profile`  > `Share icon`
- Invite friends url from `Invite friends` button and `+ -button` > `Invite friends`
- Other user `Profile details` and `Profile details` > `Share icon`

# Acceptance Criteria
Documentation is updated if needed: https://specs.status.im/spec/2

# Notes
for guidance:
https://stackoverflow.com/questions/43629265/deriving-an-ecdsa-uncompressed-public-key-from-a-compressed-one
https://bitcoin.stackexchange.com/questions/69315/how-are-compressed-pubkeys-generated
https://crypto.stackexchange.com/questions/8914/ecdsa-compressed-public-key-point-back-to-uncompressed-public-key-point

# Future Steps
[Implementation in status-react](https://github.com/status-im/status-react/issues/10325) to request compressed key

gravityblast:
for reference, since we use `go-ethereum` we can use its functions to compress and decompress keys:

https://github.com/ethereum/go-ethereum/blob/master/crypto/secp256k1/secp256.go#L126

Samyoul:
## Some Code

A little update on this. I've spent a little bit of time experimenting with this and here is the play code I wrote:

```go
func Test(t *testing.T) {
	pk := "04261c55675e55ff25edb50b345cfb3a3f35f60712d251cbaaab97bd50054c6ebc3cd4e22200c68daf7493e1f8da6a190a68a671e2d3977809612424c7c3888bc6"

	// decode hex string to bytes
	pkb, err := hex.DecodeString(pk)
	if err != nil {
		t.Error(err)
	}

	// encode bytes in to variously base encodings
	b32pk := base32.StdEncoding.EncodeToString(pkb)
	b58pk := base58.Encode(pkb)
	b64pk := base64.StdEncoding.EncodeToString(pkb)

	// create crypto public key from decoded bytes
[...]

PR #1990, opened 2020-06-09
Feature/key compression

Samyoul:
## What has changed?

I've introduced to the public binding functionality that will compress and decompress public keys of a variety of encoding and key types. This functionality supports all major byte encoding formats and the following EC public key types:

- `secp256k1` pks
- `bls12-381 g1` pks
- `bls12-381 g2` pks

## Why make the change?

We want shorter public (chat) keys and we want to be future proof and encoding agnostic. See the issue here https://github.com/status-im/status-go/issues/1937


Samyoul:
FYI, I've got some specifications to write for this, as the functions are not straightforward to use and require some background knowledge of `multibase` encoding, `multicodec` key types and `multiformat/varints`.

I'm writing these next, I won't merge this functionality until I've committed the specs.

---

**EDIT**

Spec PR here https://github.com/status-im/specs/pull/137

cammellos (approved):
Looks good, the only thing I would change is the naming,
We use extensively `CompressPubKey` in the codebase (which goes from the secp256k uncompressed to compressed), so it might get a bit confusing which one does what,
we could have a separate namespace `api/utils/multikeyformat` or change the name of the function to reflect the fact that is compressing and using a multikey format (any naming is good with me)

arnetheduck:
I'm a but confused by the functions in this PR - basically, compressed and uncompressed are serialization formats - the typical way to use keys in an application is to keep them in a format native to the library that's used to verify signatures etc - basically, you hold the binary representation of the big number in memory, so you can perform arithmetic on it, and each library (like libsecp256k1) has its own way of dividing it up the big number into something computers can chew on - some use 64-bit limbs, some use smaller limbs etc - it's up to the library and a bit of a black box, but you don't want to be converting back and forth between serialized and native format when you use the key.

Once you have a native key, you serialize them to compressed or uncompressed formats, then maybe decorate them with multiformat if the situation demands it - whenever they have to be displayed to the user, sent over the wire or written to storage

Thus, I find it a bit confusing have functions that convert straight from from one serialization format to another - the number of combinations will then grow exponentially for every new serialization type (there's DER, ethereum-style, discv5-style etc, and that's just for secp keys)

Samyoul:
@arnetheduck Thanks for your input. The functions in this PR are only for external consumption, they provide a convenient endpoint for the `status-go` client to compress and decompress keys solely for the purpose of presenting the keys to the application user.

For example `status-react` would require the compressed version of a pk for presenting a compact QR code. The use cases of compact pks are detailed in the sister issue https://github.com/status-im/status-react/issues/10325

- `Onboarding` > `Choose a chat name`
- `Profile` > `Header`
- `Profile` > `Share icon` > `QR code popover`
- `Invite friends` url from `Invite friends` button and `+ -button` > `Invite friends`
- Other user `Profile details` 
- `Profile details` > `Share icon` > `QR code popover`

Samyoul:
@cammellos 

>we could have a separate namespace api/utils/multikeyformat or change the name of the function to reflect the fact that is compressing and using a multikey format (any naming is good with me)

This is a good point, I'll make the change for this.

arnetheduck:
same as the spec comment, you're serializing to compressed or uncompressed format, not "compressing" or "decompressing" - when switching between the two formats, "coverting" is a more accurate term.

arnetheduck:
> they provide a convenient endpoint

yes - but even then one might consider exposing two functions "serialize" and "parse" instead - this is more natural for the operation that's being performed - https://github.com/bitcoin-core/secp256k1/blob/master/include/secp256k1.h#L365 is the "standard" API for the functionality you're exposing here.

Samyoul:
Most libraries call this functionality "compress" and "decompress", or "from_compressed", "to_compressed", "from_decompressed" and "to_decompressed". The common parlance for this functionality is compress and decompress.

Even the underlying functions that this PR depends on use the terms "compress" and "decompress".

If you think it would be a better idea and more easily understood by the users of this API to call these functions

- `MultiformatSerialisePublicKey`
- `MultiformatDeserialisePublicKey`

I don't have an objection to making this change.

PR #1991, opened 2020-06-17
Use the Hash of the IP address instead of the actual IP address

cammellos:
For privacy reasons/compliance we want to keep the least amount
of information in our application.

https://discuss.status.im/t/status-app-privacy-policy-refresh/1762/3

This hashes the IP addresses before using them as key, so that the clear
IP address is never stored.

### DIFF
--- a/waku/common/rate_limiter.go
+++ b/waku/common/rate_limiter.go
@@ -25,10 +25,11 @@ import (
 	"net"
 	"time"
 
-	"github.com/tsenart/tb"
+	"golang.org/x/crypto/sha3"
 
 	"github.com/ethereum/go-ethereum/p2p"
 	"github.com/ethereum/go-ethereum/p2p/enode"
+	"github.com/tsenart/tb"
 )
 
 var errRateLimitExceeded = errors.New("rate limit has been exceeded")
@@ -186,6 +187,14 @@ func (r *PeerRateLimiter) Decorate(p RateLimiterPeer, rw p2p.MsgReadWriter, runL
 				// this relies on <nil> being the string representation of nil
 				// as IP() might return a nil value
 				ip = p.IP().String()
+
+				// If not the special <nil> case, take the hash of the IP address
+				// so it's not stored on our side
+				if ip != "<nil>" {
+					h := make([]byte, 64)
+					sha3.ShakeSum256(h, []byte(ip))
+					ip = fmt.Sprintf("%x", h)
+				}
 			}
 			if halted := r.throttleIP(ip, packet.Size); halted {
 				for _, h := range r.handlers {


waku/common/rate_limiter.go
```diff
@@ -186,6 +187,14 @@ func (r *PeerRateLimiter) Decorate(p RateLimiterPeer, rw p2p.MsgReadWriter, runL
...
 				// as IP() might return a nil value
 				ip = p.IP().String()
+
+				// If not the special <nil> case, take the hash of the IP address
+				// so it's not stored on our side
+				if ip != "<nil>" {
+					h := make([]byte, 64)
+					sha3.ShakeSum256(h, []byte(ip))
+					ip = fmt.Sprintf("%x", h)
+				}
```
jakubgs, line 197: Could we just make this into a `obfuscateIp` helper?
```go
func obfuscateIp(ip string) string {
	if ip == "<nil>" {
		return ip
	}
	h := make([]byte, 64)
	sha3.ShakeSum256(h, []byte(ip))
	return fmt.Sprintf("%x", h)
}
```

jakubgs (changes requested):
Good idea, but anything that can be its own function should be.

waku/common/rate_limiter.go
cammellos: Yes, can I just call it `hashIPAddress` instead?

jakubgs: whatever

Samyoul (approved):
Aside from addressing what @jakubgs has raised, this PR looks good.

waku/common/rate_limiter.go
```diff
@@ -25,10 +25,11 @@ import (
 	"net"
 	"time"
 
-	"github.com/tsenart/tb"
+	"golang.org/x/crypto/sha3"
 
 	"github.com/ethereum/go-ethereum/p2p"
 	"github.com/ethereum/go-ethereum/p2p/enode"
+	"github.com/tsenart/tb"
```
Samyoul, line 32: Ah, I think that the linter is complaining that your imports aren't grouped by repo.

```suggestion
	"golang.org/x/crypto/sha3"

	"github.com/ethereum/go-ethereum/p2p"
	"github.com/ethereum/go-ethereum/p2p/enode"

	"github.com/tsenart/tb"
```

Samyoul: The linter gives the worst messages for this kind of linting issue.

cammellos:
Is this even needed then? 
As I understand, this is just a box ticking exercise (it's trivial to build a list of the IP addresses hash), but technically we would not be storing IP addressing in memory. We can add a salt generated at runtime if needed, but in terms of privacy is identical (if an attacker gets access to the memory it will be able to access the hashes and the salt).

cc @hesterbruikman

Samyoul:
> (if an attacker gets access to the memory it will be able to access the hashes and the salt

Would the attacker also have access to the private key(s) if memory access is compromised?

cammellos:
> Would the attacker also have access to the private key(s) if memory access is compromised?

Yes, at least chat keys

hesterbruikman:
Missed a lot of communication around this. Originally what I wanted to understand was:
>  I’d like to understand if rate limiting by IP address puts us in the controller category. Are IP addresses, stored, for what time?

If a hash of the IP address is stored (rather than the IP address itself). That might make a difference. I wasn't aware though that the IP address is only stored in memory, which in itself might make a difference. Let me verify both cases with the legal counsel. 

Objective is to see what we can do to not store any personal identifiable information (IP addresses). This a) avoids having to jump through legal hoops, reduce the number of boxes to tick, and b) pushes for what we can do to increase privacy. If hashes don't make a difference for `b`, `a` remains, pending on whether this makes a difference compared to storing in memory

waku/common/rate_limiter.go
andremedeiros: I think what the linter wants is:

```
- Imports from stdlib

- Imports from 3rd party vendors

- Imports from your own project
```

cyanlemons:
>We can add a salt generated at runtime if needed, but in terms of privacy is identical (if an attacker gets access to the memory it will be able to access the hashes and the salt).

@cammellos There is a slight benefit to adding a salt even if it gets compromised. You are immune to rainbow tables. SHA256 is rather common, and I wouldn't be surprised if someone already precomputed the hash of every single IPv4 address. If a salt is used, an attacker that gains access to the hashes would be required to recompute the full table every time.

PR #1992, opened 2020-06-17
Audio voice files

cammellos:
https://github.com/status-im/specs/pull/138 for the specs

## Why make this change?
    
We are adding support of audio recorded files, similarly to how we did
    with images
    
## What has changed?
    
- Added protobuf definition, only AAC supported
- Added migrations to store files
- Fixed an issue with nil pointer when transaction would fail to be created, causing the application to crash


cammellos:
@0kok0 adding you as I can't find a tag for security

protocol/protobuf/chat_message.pb.go
```diff
@@ -350,10 +426,16 @@ type ChatMessage_Image struct {
...
 
+type ChatMessage_Audio struct {
+	Audio *AudioMessage `protobuf:"bytes,11,opt,name=audio,proto3,oneof"`
+}
+
 func (*ChatMessage_Sticker) isChatMessage_Payload() {}
 
 func (*ChatMessage_Image) isChatMessage_Payload() {}
 
+func (*ChatMessage_Audio) isChatMessage_Payload() {}
```
jakubgs, line 448: What is this and how does this work?

jakubgs (approved):
I don't see anything wrong with this.

protocol/protobuf/chat_message.pb.go
cammellos: That's magic, it's true meaning has been lost in the mists of time.
Some say that reciting that magic incantation on a full moon will create a path between you and another being, and through that path they'll hear your voice (after clicking on an icon).
Others say that this is just what a compiled protobuf declaration looks like, and that is automatically generated, so not to be questioned.
It is up to you who to believe.

jakubgs: Weird but okay...

0kok0:
Still reviewing this - don't merge pls

0kok0 (approved):
Looks good to me

protocol/message.go
```diff
@@ -253,6 +261,37 @@ func (m *Message) parseImage() error {
 	return nil
 }
 
+// parseAudio check the message contains an audio, and if so
+// it creates the a base64 encoded version of it.
```
Samyoul: tiny tpyo

```suggestion
// it creates a base64 encoded version of it.
```

protocol/persistence.go
```diff
@@ -31,6 +31,10 @@ func (db sqlitePersistence) SaveChat(chat Chat) error {
 
 func (db sqlitePersistence) SaveChats(chats []*Chat) error {
 	tx, err := db.db.BeginTx(context.Background(), &sql.TxOptions{})
+	if err != nil {
+		return err
+	}
+
```
Samyoul, line 38: Nice early return

protocol/messenger.go
```diff
@@ -1412,6 +1413,26 @@ func (m *Messenger) SendChatMessage(ctx context.Context, message *Message) (*Mes
 		message.Payload = &protobuf.ChatMessage_Image{Image: &image}
 
 	}
+
+	if len(message.AudioPath) != 0 {
+		file, err := os.Open(message.AudioPath)
```
Samyoul, line 1388: Is it possible for `if len(message.ImagePath) != 0` and `if len(message.AudioPath) != 0` to both be true?

protocol/message_validator.go
```diff
@@ -209,5 +209,22 @@ func ValidateReceivedChatMessage(message *protobuf.ChatMessage, whisperTimestamp
...
+			return errors.New("no audio content")
+		}
+		if len(audio.Payload) == 0 {
+			return errors.New("audio payload empty")
+		}
+
+		if audio.Type == protobuf.AudioMessage_UNKNOWN_AUDIO_TYPE {
+			return errors.New("audio type unknown")
+		}
+	}
```
Samyoul, line 227: This looks like it should be a switch statement on `message.ContentType` with dedicated type validator functions or errors for

- `protobuf.ChatMessage_UNKNOWN_CONTENT_TYPE`
- `protobuf.ChatMessage_TRANSACTION_COMMAND`
- `protobuf.ChatMessage_STICKER`
- `protobuf.ChatMessage_IMAGE`
- `protobuf.ChatMessage_AUDIO`

protocol/messenger.go
Samyoul: Looks like these tests are also sub-functions, would it make sense to break them out into their own dedicated function?

protocol/message_validator.go
jakubgs: if you're going that far then you might as well make the handler of each type it's own function to keep the `switch` concise.

PR #1993, opened 2020-06-18
Separate keystore dir per multiaccount

rasom:
As part of https://github.com/status-im/status-react/issues/10195, having all keys in a separate dir will simplify account removing. Also it will allow to have same accounts to added via private key/seed in wallet in different multiaccs (as keys will be stored in separate dirs).

This commit provides `MigrateKeyStoreDir` method which moves multiacc's keys from a common keystore directory to multiacc specific directory.

status: ready


api/geth_backend.go
```diff
@@ -305,6 +308,90 @@ func (b *GethStatusBackend) startNodeWithAccount(acc multiaccounts.Account, pass
...
+		return checkFile(path, fileInfo)
+	})
+	if err != nil {
+		return nil, fmt.Errorf("cannot traverse key store folder: %v", err)
+	}
+
+	return paths, nil
+}
+
+func (b *GethStatusBackend) MigrateKeyStoreDir(acc multiaccounts.Account, password, oldDir, newDir string) error {
```
andremedeiros, line 308: Is this supposed to be idempotent?

rasom: Yep it safe to be called many times (if that's what you are asking about).

api/geth_backend.go
```diff
@@ -305,6 +308,90 @@ func (b *GethStatusBackend) startNodeWithAccount(acc multiaccounts.Account, pass
...
 
+func getKeyFilePathsByAddresses(keyStoreDir string, addresses map[string]struct{}) ([]string, error) {
+	paths := []string{}
+
+	checkFile := func(path string, fileInfo os.FileInfo) error {
+		if fileInfo.IsDir() {
+			return nil
+		}
+
+		rawKeyFile, e := ioutil.ReadFile(path)
```
cammellos: `err` is more idiomatic, instead of `e`

api/geth_backend.go
```diff
@@ -305,6 +308,90 @@ func (b *GethStatusBackend) startNodeWithAccount(acc multiaccounts.Account, pass
...
+		}
+
+		var accountKey struct {
+			Address string `json:"address"`
+		}
+		if e := json.Unmarshal(rawKeyFile, &accountKey); e != nil {
+			return fmt.Errorf("failed to read key file: %s", e)
+		}
+
+		address := types.HexToAddress("0x" + accountKey.Address).Hex()
```
cammellos: Not sure I understand here, are you basically normalizing the address by parsing it and converting it back to hex? 
If so I would leave a comment

cammellos (approved):
Version bump is missing, left a few minor comments, otherwise looks good

api/geth_backend.go
```diff
@@ -305,6 +308,90 @@ func (b *GethStatusBackend) startNodeWithAccount(acc multiaccounts.Account, pass
...
+
+	for _, account := range accounts {
+		addresses[account.Address.Hex()] = struct{}{}
+	}
+
+	paths, err := getKeyFilePathsByAddresses(oldDir, addresses)
+
+	if err != nil {
+		return err
+	}
```
Samyoul: Little formatting tweak. It's idiomatic to have the err check directly below the function that generates an error.

```suggestion
	paths, err := getKeyFilePathsByAddresses(oldDir, addresses)
	if err != nil {
		return err
	}
```

api/geth_backend.go
Samyoul: Is this a kind of validation of the address?

Samyoul (changes requested):
Good work. I believe that this code needs some tests adding, as currently there aren't any tests the new functionality added.

api/geth_backend.go
rasom: I actually just copied code from here https://github.com/status-im/status-go/blob/8b61d92d5834215fb06532c527b0bead7b4625d9/account/accounts.go#L142 ` ¯\_(ツ)_/¯`

api/geth_backend.go

api/geth_backend.go

api/geth_backend.go
rasom: @cammellos @Samyoul let me know if there is a better way to write this code, but at least it works

Samyoul: @rasom Would you mind breaking this section of code out into a dedicated `verifyKeyFile()` function and adding some tests specifically for this functionality? 😊 Thank you

rasom: @Samyoul I added one tiny test for `MigrateKeyStoreDir`, it is a bit puzzling for me how to juggle with slice references (I suppose it would be necessary with `verifyKeyFile` )

rasom: let me know if you want to see something else there, meanwhile the whole thing is tested manually here https://github.com/status-im/status-react/pull/10847

Samyoul (approved):
Thank you very much

PR #1996, opened 2020-06-25
Add mute/unmute functionality

cammellos:
## Why make this change?

We will be adding soon support for notifications, a first step is to
allow the user to selectively mute or unmute chats.

## What has changed?

- Added migration for mute/unmute
- Messages from muted chats won't change the unviewedMessagesCount and
will be marked automatically as seen


cammellos:
Not relevant anymore after a chat with UX

PR #1999, opened 2020-07-07
Emoji Reactions

Samyoul:
## What has changed?
I've added two types of waku message payload:

* `EmojiReaction`
* `EmojiReactionRetraction`

Along with:

* protobuf definitions
* JSON RPC (un)marshalling
* data validation
* SQLite migrations
* SQLite getting and putting

## Why make the change?
To add emoji reaction functionality see:

* https://github.com/status-im/status-react/issues/7118
* https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=1239%3A0
* https://github.com/status-im/status-react/pull/10912

## Helpful reference

* [#1992 - Audio voice files](https://github.com/status-im/status-go/pull/1992/files)

## New Roadmap

* [x] Undo initial implementation
* [x] Implement new protobufs
  * EmojiReaction
  * EmojiReactionRetraction
* [x] Implement new ApplicationMetadataMessage
  * EMOJI_REACTION
  * EMOJI_REACTION_RETRACTION
* [x] Implement emoji reaction table creation migrations
* [x] Implement **Sending** API
  * [x] SendEmojiReaction(message_id, emoji_id) emoji_reaction_id
  * [x] SendEmojiReactionRetraction(emoji_reaction_id) emoji_reaction_retraction_id
* [x] Implement **Receiving** API
  * Update RetrieveAll() to return with new `EmojiReaction` field `MessengerResponse{EmojiReaction {EmojiID string, Retracted bool }}`
* [x] Implement Emoji Reaction persistence handling
* [ ] Implement new Emoji Reaction API **getter** for chat and messages
  * [x] `EmojiReactions(chatID, cursor string, limit int) []EmojiReaction {MessageID string, EmojiID string, Retracted bool}`
  * [ ] `EmojiReactionsFor(messages []string)`

## Specs
https://github.com/status-im/specs/pull/145


Samyoul:
[💜 1999 🦄 ](https://www.youtube.com/watch?v=rblt2EtFfC4)

Samyoul:
Specs pending review https://github.com/status-im/specs/pull/145

cammellos:
I would probably have a separate protobuf message for this, similarly to what you have done, but not as a content type of a chat message:

```
message EmojiReaction {
  string message_id = 1;
  Type type = 2;

  enum Type {
    UNKNOWN_EMOJI_REACTION_TYPE = 0;
    LOVE = 1;
    THUMBS_UP = 2;
    THUMBS_DOWN = 3;
    LAUGH = 4;
    SAD = 5;
    ANGRY = 6;
  }
}
```
With the addition of a `clock/timestamp` as we might need to be able to order the reaction from the same user on the same post, also to give the reaction an unique id.

There are a few issues with having it as a new content type:

- Old client will see for any reaction/retraction a separate message saying `Upgrade to the latest version to see a nice reaction!`, in this case I would probably expect the best UX is to just ignore reactions for older clients.

- Having reactions interleaved with messages will make querying a bit more cumbersome. When pulling messages from the database you will need to exclude reactions (https://github.com/status-im/status-go/blob/6d5a93287b2c1b4d4d5d1178a3ecec870bf18b9e/protocol/message_persistence.go#L395), as you want to pull 20 messages, and reactions are not ordered in the message stream (a recent reaction might be referring to an older message).

So I would probably store reactions in a different table altogether, and provide different endpoints to send reactions, something like `SendReactionr`/`RetractReaction`.

There's also the question on how to surface this information to the client, which is a bit tricky.

As far as I can see now they come in the messages signal, which works only for the current chat, but as soon as the user open/closes a chat all the reactions are gone (kind of, you might get some in the messages pulled from the db, but we fetch 20 only).

I suppose desired behavior is that we always show all the reactions for a given message.

If stored in a different table, you can JOIN against the reactions table and count them up when loading messages, or we can denormalize the count on the message table, or we can pull them asynchronously through a different endpoint (for performance reasons).

And lastly there's the issue on how do we signal to the client that an new reaction/retraction is in, currently is using the `messages` field, but probably I would have a separate field, in order to distinguish the two.

what do you think?

Samyoul:
Thank you @cammellos , this is the kind of feedback I was hoping to receive.

>as you want to pull 20 messages, and reactions are not ordered in the message stream (a recent reaction might be referring to an older message).

I was not aware of this, and makes a lot of sense for changing the entire approach.

>If stored in a different table, you can JOIN against the reactions table and count them up when loading messages, or we can denormalize the count on the message table, or we can pull them asynchronously through a different endpoint (for performance reasons).

I'd maybe go with a "asynchronously through a different endpoint". That would make the logic cleaner and also allow the client to chose whether to receive reactions or not.

>And lastly there's the issue on how do we signal to the client that an new reaction/retraction is in, currently is using the messages field, but probably I would have a separate field, in order to distinguish the two.

I got a bit stuck in my implementation, I wasn't able to track where the signal for a new message is broadcast from, so I piggybacked on the `ChatMessage type. I agree with you that ideally reactions should be treated as a new category of message and not a sub-category of `ChatMessage`. Do you know where the "new message" signal is sent from in the code?

cammellos:
> I'd maybe go with a "asynchronously through a different endpoint". That would make the logic cleaner and also allow the client to chose whether to receive reactions or not.

Sounds like an option, we probably will need to liaise a bit with the client as there are potential issues with this approach.
I am thinking of when you receive new messages, and you don't want to make a query for each single message.
But you could have a cool off period in the UI to batch requests, so it can be overcome, so I think it looks like a sound first step and we don't impact performance by adding yet another join.

> I got a bit stuck in my implementation, I wasn't able to track where the signal for a new message is broadcast from,

Yes, `messages.new` it should be renamed actually, up to you if you want to do this in the PR (it's an easy change but breaking for clients, but it's easy to fix). It was named messages.new because all the logic was in status-react, so it was just handling the marshaled payloads to status-react, but now a name such as `updates` would probably described better the purpose.

Basically `messages.new` contains a `MessengerResponse`.  https://github.com/status-im/status-go/blob/6d5a93287b2c1b4d4d5d1178a3ecec870bf18b9e/protocol/messenger.go#L80 .
So in order to add reactions you can just add a field to this struct and it will be serialized correctly.

protocol/protobuf/service.go
```diff
@@ -4,7 +4,7 @@ import (
 	"github.com/golang/protobuf/proto"
 )
 
-//go:generate protoc --go_out=. ./chat_message.proto ./application_metadata_message.proto ./membership_update_message.proto ./command.proto ./contact.proto ./pairing.proto
+//go:generate protoc --go_out=. ./chat_message.proto ./application_metadata_message.proto ./membership_update_message.proto ./command.proto ./contact.proto ./pairing.proto ./emoji_reaction.proto
```
andremedeiros: I absolutely hate that go:generate doesn't support globs.

Samyoul:
@cammellos I am very sorry to hand this over to you, following on from what we discussed on Friday I have done additional work on parsing messages and determining the chat type. On Saturday I realised that the processor.EncodeMembershipUpdate() functionality requires that a message is sent when a private group message is sent. I've abstracted most of the `ChatMessage` specific code to use the new `ChatEntity` interface, I've had to abstract part of the protobufs handling for groupmemberupdate to accept a "chat entity" rather than a message.

Please see what work still remains as far as I am aware. I will be available for questions this week, I may not be immediately available, but I will make special effort to get back to you ASAP.

cc @andremedeiros, @hesterbruikman as I know you are very eager for this PR to be completed and merged as it is preventing 1.5's release.

# Global

None of the new functions have new exposed API methods

Sending emoji reactions

`SendEmojiReaction(chat_id, message_id, emoji_id) MessengerResponse`
`SendEmojiReactionRetraction(emoji_reaction_id) MessengerResponse`

Getting emoji reactions from network:

**Done**
```
RetrieveAll()
MessengerResponse{
    []EmojiReaction {
		emoji_reaction_id: string
		message_id: string
		emoji_id: int
		from: string
		retracted: bool
    }
}
```

Getting emoji reactions from db:

`EmojiReactions(chatID, cursor string, limit int) []EmojiReaction`
`EmojiReactionsFor(message_ids []string) []EmojiReaction` *optional*

# protocol/message_handler.go

## HandleMembershipUpdate()
Need to add a new field to messageState.CurrentMessageState called EmojiReaction to handle emoji reactions coming from a membership update

## HandleEmojiReaction()
This function is not fully functional, I believe we need to add to emoji reaction to the state.Response.EmojiReactions for later processing

## HandleEmojiReactionRetraction()
Create

# protocol/message_persistence.go

Requires two query functions to get emoji reactions by chat id and by a list of message ids

## EmojiReactions(chatID, cursor string, limit int) []EmojiReaction

create function that gets the all the emoji reactions for all messages that match the chat id and cursor

## EmojiReactionsFor(messages []string) []EmojiReaction

create

protocol/v1/membership_update_message.go
```diff
@@ -86,13 +98,18 @@ func MembershipUpdateMessageFromProtobuf(raw *protobuf.MembershipUpdateMessage)
 	return &MembershipUpdateMessage{
 		ChatID:  raw.ChatId,
 		Events:  events,
-		Message: raw.Message,
+		Message: raw.GetMessage(),
```
Samyoul: Need to also handle an EmojiReaction field

protocol/v1/status_message.go
```diff
@@ -160,165 +161,76 @@ func (m *StatusMessage) HandleApplicationMetadata() error {
...
+		m.ParsedMessage = nil
+		log.Printf("[message::DecodeMessage] could not decode %T: %#x, err: %v", pb, m.Hash, err.Error())
+	} else {
+		rv = reflect.ValueOf(ptr)
+		m.ParsedMessage = rv.Elem()
+		return nil
+	}
+
 	return nil
 }
```
Samyoul: I haven't yet implemented any tests on this func, and I'm not 100% is will always parse pointers and values as intended. It should work fine, but I'm not 100%

cammellos:
@Samyoul I have implemented what was left, there wasn't much left to do, the changes I have made are:

- Added `EmojiReactionsByChatID` endpoint
- Changed the ID scheme to be `HexEncode(Keccak(From-MessageID-EmojiType))` so that we can gracefully handle multiple likes/unlikes/likes (last one wins, we drop the history, as I don't think we are interested in it)
- Removed the emojretraction message, and instead use a `retracted` flag. This is because the same amount of information needs to be carried, so it's easier to just have a flag.
- Set local-chat-id 

I have added an integration test in `messenger_emoji_test.go`. 
I have updated the specs as well https://github.com/status-im/specs/pull/145/files and this should be ready for review.

cc @andremedeiros @jakubgs

Samyoul:
Thank you @cammellos , I really appreciate this.

> Changed the ID scheme to be HexEncode(Keccak(From-MessageID-EmojiType)) so that we can gracefully handle multiple likes/unlikes/likes (last one wins, we drop the history, as I don't think we are interested in it)

I saw the commit with this and I wondered what its purpose was.

>Removed the emojretraction message, and instead use a retracted flag. This is because the same amount of information needs to be carried, so it's easier to just have a flag.

In this context it makes sense. I think that this is a good idea.

---

Is it typical for co-authors to review to review a PR? Should I give my review of the parts you've finished off?

protocol/v1/membership_update_message.go
Samyoul: @cammellos am I mistaken in thinking that this section needs to also handle `EmojiReaction`s?

Something like:

```
Message: raw.GetMessage(),
EmojiReaction: raw.GetEmojiReaction(),
```

cammellos: oh good point, I missed that, thanks!

cammellos:
> Is it typical for co-authors to review to review a PR? Should I give my review of the parts you've finished off?

I'd say yes, mainly because there aren't many of us left for reviews, so every review counts :)

protocol/v1/status_message.go
```diff
@@ -175,241 +176,90 @@ func (m *StatusMessage) HandleApplicationMetadata() error {
...
-			m.ParsedMessage = message
+		return m.unmarshalProtobufData(new(protobuf.PushNotificationResponse))
+	case protobuf.ApplicationMetadataMessage_EMOJI_REACTION:
+		return m.unmarshalProtobufData(new(protobuf.EmojiReaction))
+	case protobuf.ApplicationMetadataMessage_PUSH_NOTIFICATION_REGISTRATION:
+		// This message is a bit different as it's encrypted, so we pass it straight through
+		v := reflect.ValueOf(m.DecryptedPayload)
+		m.ParsedMessage = &v
+		return nil
+	}
```
Samyoul, line 240: That's interesting

Samyoul (commented):
@cammellos , I've reviewed all the changes you've made and they all look good to me. Unfortunately GitHub won't let me approve this PR because I'm an author. We need someone else to approve this.

cammellos:
@jakubgs would you mind approving for @Samyoul  (or reviewing if you'd like)?
thanks!

jakubgs:
Sure, let me review it. I intended to do so but I got busy with the F-Droid stuff.

protocol/common/chat_entity.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"crypto/ecdsa"
+
+	"github.com/golang/protobuf/proto"
+
+	"github.com/status-im/status-go/protocol/protobuf"
+)
+
+type ChatEntity interface {
```
jakubgs, line 13: Could we add a comment here explaining how this is supposed to encompass `Message` and `EmojiReaction`?

protocol/message_persistence.go
```diff
@@ -488,6 +490,72 @@ func (db sqlitePersistence) MessageByChatID(chatID string, currCursor string, li
...
+			FROM
+				emoji_reactions e
+			WHERE NOT(e.retracted)
+			AND
+			e.local_chat_id = ?
+			AND
+			e.message_id IN
+			(SELECT id FROM user_messages m WHERE NOT(m.hide) AND m.local_chat_id = ? %s
+			ORDER BY substr('0000000000000000000000000000000000000000000000000000000000000000' || m.clock_value, -64, 64) || m.id DESC LIMIT ?)
+			LIMIT 100
```
jakubgs: This is weird. If we cap the query at N results then the result might not include our own reaction, which would make the UI appear like you can vote again.

protocol/message_persistence.go
```diff
@@ -488,6 +490,72 @@ func (db sqlitePersistence) MessageByChatID(chatID string, currCursor string, li
...
+			&emojiReaction.Type,
+			&emojiReaction.MessageId,
+			&emojiReaction.ChatId,
+			&emojiReaction.LocalChatID,
+			&emojiReaction.Retracted)
+		if err != nil {
+			return nil, err
+		}
+
+		result = append(result, &emojiReaction)
```
jakubgs, line 560: Returning the whole list seems like a real overkill. This will get very heavy in threads that have loads of reactions on loads of messages. A more sensible response would just include a count and a `bool` telling you if you are in the list. Especially since we'll be re-sending this list every time we want to update the shown count.

Might be better to do it in a separate PR though.

jakubgs (approved):
Doing this in a distributed way makes this so over-complicated it's painful.

Samyoul:
@jakubgs , it is isn't it. It's so funny how different a decentralised approach for something so apparently simple as emoji reactions compares to what would be implemented using a traditional set up.

PR #2001, opened 2020-07-13
DeleteMultiaccount method

rasom:
status-im/status-react#10195

status: ready


api/backend_test.go
```diff
@@ -614,3 +614,79 @@ func TestLoginWithKey(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, crypto.PubkeyToAddress(chatKey.PublicKey), extkey.Address)
 }
+
+func TestDeleteMulticcount(t *testing.T) {
+	backend := NewGethStatusBackend()
+
+	rootDataDir, err := ioutil.TempDir("", "test-keystore-dir")
```
cammellos, line 621: Error checking, also run `defer` as soon as you check for the error, otherwise if `require.NoError(t,err)` fails on line 623, it won't run it.

api/geth_backend.go
```diff
@@ -178,6 +179,40 @@ func (b *GethStatusBackend) SaveAccount(account multiaccounts.Account) error {
...
+	for _, path := range dbFiles {
+		if _, err := os.Stat(path); err == nil {
+			err = os.Remove(path)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	err = os.RemoveAll(keyStoreDir)
```
cammellos: you can just:
`return os.RemoveAll(keyStoreDir)`

cammellos (approved):
looks good, just a couple of minor comments

api/backend_test.go
rasom: fixed, thanks

api/geth_backend.go

PR #2006, opened 2020-07-28
dont use a pointer for public methods

cammellos:
A bug was introduced due to the fact that a pointer structure was modified and re-used. 
This commit changes the public interface to copy-by-value so that we don't have to worry about this happening in the future.
A regression test has been added as it wasn't picked up in the first place.


protocol/messenger_installations_test.go
```diff
@@ -241,3 +241,54 @@ func (s *MessengerInstallationSuite) TestSyncInstallation() {
...
+		"installation not received",
+	)
+
+	s.Require().NoError(err)
+	actualInstallation := response.Installations[0]
+	s.Require().Equal(bob2.installationID, actualInstallation.ID)
+	err = bob1.EnableInstallation(bob2.installationID)
+	s.Require().NoError(err)
+
+	// send a message from bob1 to alice, it should be received on both bob1 and bob2
```
jakubgs, line 276: Why should it be received on both bob1 and bob2 if it's being sent to alice?

jakubgs (commented):
I get that this was an issue, but why? What was it causing?

protocol/messenger_installations_test.go
cammellos: because bob1 is paired with bob2, basically if you send a message from your IOS device to someone, it will show on your android device as well

cammellos:
@jakubgs basically if you want to set a different key to send messages, the field `Sender` was set by the caller. 
This field was then modified by this function if not set.
If `RawMessage` was reused (such as in sending to a paired message), the dirty field would be persisted, hitting an error condition as we don't support sending with datatasync and setting this field.

```
rawMessage = {Sender: nil, Payload: blah, Datasync: true}

sendToAlice(rawMessage,aliceKey)
// This would set `Sender` to BobIdentity
sendToPairedDevice(rawMessage) // Here both Datasync & Sender have been set, which will hit the error condition as we don't support setting  sender and datasync
```

Basically we should not modify anything if passed by reference, unless the behavior is clear (it returns nothing for example).

jakubgs (approved):
Thanks for explaining.

PR #2009, opened 2020-07-30
Add topic and contact code propagation

cammellos:
## What has changed

- A topic is passed to the push notification server, this is needed as two version of the app (PR & Release for example), use different topics.
- Removed handlers and moved to subscriptions in encryption layer, this makes initialization of messenger simpler and gives more control over the handling
- Propagate push notification details with contact code
- Have client specify type of token & topic
- Minor fixes

 ## Why make this change

Some of these are just bug fixes, other are features.

We propagate push notification info with contact code in order to minimize traffic to the push notification server and reduce the delay when sending push notifications.

We have to specify a topic to support multiple version of the app (PR & release for example)

I have changed to subscriptions in the encryption layer so it's easier to initialize messenger instances and logic is better encapsulated inside messenger, it also simplifies building the payload for the contact code. Callbacks are still necessary for message processor for now, as we need to synchronously listen to the negotiated topic before sending a message.


protocol/common/message_processor.go
```diff
@@ -479,6 +510,16 @@ func (p *MessageProcessor) sendDataSync(ctx context.Context, publicKey *ecdsa.Pu
...
+	// The shared secret needs to be handle before we send a message
+	// otherwise the topic might not be set up before we receive a message
+	if p.handleSharedSecrets != nil {
+		err := p.handleSharedSecrets([]*sharedsecret.Secret{messageSpec.SharedSecret})
+		if err != nil {
+			return err
+		}
+
+	}
+
```
Samyoul, line 522: If the shared secret needs to be handled before we send a message, should an error be returned if the handler is not set?

protocol/pushnotificationclient/client.go
```diff
@@ -302,6 +318,53 @@ func (c *Client) HandlePushNotificationRegistrationResponse(publicKey *ecdsa.Pub
 	return c.persistence.UpsertServer(server)
 }
 
+// processQueryInfo takes info about push notifications and validates them
+func (c *Client) processQueryInfo(clientPublicKey *ecdsa.PublicKey, serverPublicKey *ecdsa.PublicKey, info *protobuf.PushNotificationQueryInfo) error {
+	// make sure the public key matches
+	if !bytes.Equal(info.PublicKey, common.HashPublicKey(clientPublicKey)) {
+		c.config.Logger.Warn("reply for different key, ignoring")
+		return errors.New("reply for a different key, ignoreing")
```
Samyoul:
```suggestion
		return errors.New("reply for a different key, ignoring")
```

protocol/pushnotificationclient/client.go
```diff
@@ -1245,6 +1307,48 @@ func (c *Client) handleAllowedKeyList(publicKey *ecdsa.PublicKey, allowedKeyList
 	return ""
 }
 
+func (c *Client) MyPushNotificationQueryInfo() ([]*protobuf.PushNotificationQueryInfo, error) {
```
Samyoul, line 1355: Does the "My" from `MyPushNotificationQueryInfo` refer to the node's info rather than a peer's info?

I know there are a lot of references to "my" and "their", are these placeholder terms? I kind of like the use of "my and "their" but we also have the concept of "node" and "peer".

Does "my" and "their" have a different meaning / use case than "node" and "peer"?

Samyoul (approved):
Looks good

protocol/common/message_processor.go
cammellos: I take that in some cases you might not want to set up an handler, say fire & forget messages, in the case of scripts for example.

protocol/pushnotificationclient/client.go
cammellos: In this case it refers to the identity you started the `messenger` with, which is not equivalent to node nor peer I think.

My understanding is that node generally refers to the `geth` node started, while peer generally refers to the `waku` peers connected.

In this cases is neither, as the identity might change, but the peer will stay the same (we login with a different account).

What about `CurrentIdentityPushNotificationInfo`?

Samyoul: Ok, that is a reasonable distinction. I'm imagining what a future newcomer would think when looking at this code in the future.

`CurrentIdentityPushNotificationInfo` I think is more descriptive, I would say it is a better choice.

Does "their" refer to PN peers?

PR #2013, opened 2020-08-07
Feature/group chat invitation

flexsurfer:
issue: https://github.com/status-im/status-react/issues/10179

Flow:
1) admin generates and shares a link for a group chat `https://join.status.im/g/chatID&chat%20name&adminPK`
2) any status user opens a link, a chat is opened, there is only one option - request membership
3) user requests membership, we send a 1-1 GroupChatInvitation message to adminPK from the link
4) admin can reject, in that case we send GroupChatInvitation back with reject flag
5) admin can accept in that case we just use the flow for regular group chat invitation using userPK from the membership request message

Implementation:
Added:

ext service

```go
SendGroupChatInvitationRequest(ctx Context, chatID string, adminID string, message )
SendGroupChatInvitationRejection(ctx Context, invitationRequestID string) (*protocol.MessengerResponse, error)
```

sql

`group_chat_invitations` table

message

```protobuf
message GroupChatInvitation {
string chat_id = 1;

string introduction_message = 2;

bool rejection = 3;
}
```

Thanks for your review and feedback


protocol/group_chat_invitation.go
```diff
@@ -0,0 +1,65 @@
...
+
+	// From is a public key of the author of the invitation.
+	From string `json:"from,omitempty"`
+
+	// SigPubKey is the ecdsa encoded public key of the invitation author
+	SigPubKey *ecdsa.PublicKey `json:"-"`
+
+	// LocalChatID is the chatID of the local chat (one-to-one are not symmetric)
+	LocalChatID string `json:"localChatId"`
+}
```
Samyoul: I think that this code was probably copied from the emoji reaction (that change wasn't done by me), but idiomatically tagged struct fields are grouped together at the top and then untagged fields.

```suggestion
type GroupChatInvitation struct {
	protobuf.GroupChatInvitation

	// From is a public key of the author of the invitation.
	From string `json:"from,omitempty"`

	// LocalChatID is the chatID of the local chat (one-to-one are not symmetric)
	LocalChatID string `json:"localChatId"`

	// SigPubKey is the ecdsa encoded public key of the invitation author
	SigPubKey *ecdsa.PublicKey `json:"-"`
}
```

protocol/message_handler.go
```diff
@@ -722,3 +722,57 @@ func (m *MessageHandler) HandleEmojiReaction(state *ReceivedMessageState, pbEmoj
...
+
+	//chat, err := m.matchChatEntity(emojiReaction, state.AllChats, state.Timesource)
+	/*if err != nil {
+		return err // matchChatEntity returns a descriptive error message
+	}*/
+
+	// Set local chat id
+	//not sure what to do here, LocalChatID is really confusing
+	//groupChatInvitation.LocalChatID = chat.ID
+
```
Samyoul: I suppose this is just some code to help you with developing. Maybe needs to be removed

```suggestion
```

protocol/message_handler.go
```diff
@@ -722,3 +722,57 @@ func (m *MessageHandler) HandleEmojiReaction(state *ReceivedMessageState, pbEmoj
...
+	logger.Debug("Handling group chat invitation")
+
+	/*if chat.LastClockValue < pbEmojiR.Clock {
+		chat.LastClockValue = pbEmojiR.Clock
+	}*/
+
+	//state.ModifiedChats[chat.ID] = true
+	//state.AllChats[chat.ID] = chat
+
```
Samyoul: Same here

```suggestion
```

protocol/message_handler.go
```diff
@@ -722,3 +722,57 @@ func (m *MessageHandler) HandleEmojiReaction(state *ReceivedMessageState, pbEmoj
...
+	// save emoji reaction
```
Samyoul:
```suggestion
	// save group chat invitation
```

protocol/protobuf/group_chat_invitation.proto
```diff
@@ -0,0 +1,15 @@
...
+  // chat_id the ID of the private group chat the message belongs to, for query efficiency the chat_id is stored in the db even though the
+  // target message also stores the chat_id
+  string chat_id = 1;
+
+  string invitation_id = 2;
+
+  string introduction_message = 3;
+
+  bool rejection = 4;
+}
```
Samyoul, line 25: I notice that there is no clock on this protobuf. Did you decide against ordering the group chat invitations?

protocol/message_validator.go
```diff
@@ -249,3 +249,19 @@ func ValidateReceivedEmojiReaction(emoji *protobuf.EmojiReaction, whisperTimesta
 
 	return nil
 }
+
+func ValidateReceivedGroupChatInvitation(invitation *protobuf.GroupChatInvitation, whisperTimestamp uint64) error {
+	/*if err := validateClockValue(emoji.Clock, whisperTimestamp); err != nil {
+		return err
+	}*/
+
```
Samyoul: If you are not using a clock, this should be removed.

```suggestion
```

protocol/group_chat_invitation.go
```diff
@@ -0,0 +1,65 @@
...
+func (g GroupChatInvitation) GetSigPubKey() *ecdsa.PublicKey {
+	return g.SigPubKey
+}
+
+// GetProtoBuf returns the struct's embedded protobuf struct
+// this function is required to implement the ChatEntity interface
+func (g GroupChatInvitation) GetProtobuf() proto.Message {
+	return &g.GroupChatInvitation
+}
+
```
Samyoul: This looks like you are implementing the `ChatEntity` interface. I don't think that you need to do this.

protocol/messenger.go
```diff
@@ -846,6 +848,133 @@ func (m *Messenger) ChangeGroupChatName(ctx context.Context, chatID string, name
...
+	}
+
+	//why do we need this code?
+	spec.ID = types.EncodeHex(id)
+	spec.SendCount++
+	err = m.persistence.SaveRawMessage(&spec)
+	if err != nil {
+		return nil, err
+	}
+
```
Samyoul: This is used for auto resend functionality.

protocol/messenger.go
```diff
@@ -1778,7 +1907,8 @@ type ReceivedMessageState struct {
 	ExistingMessagesMap map[string]bool
 	// EmojiReactions is a list of emoji reactions for the current batch
 	// indexed by from-message-id-emoji-type
-	EmojiReactions map[string]*EmojiReaction
+	EmojiReactions       map[string]*EmojiReaction
+	GroupChatInvitations map[string]*GroupChatInvitation
```
Samyoul, line 2114: Add a comment for this new field

Samyoul (commented):
Looks good. Thank you, really good work.🥇 

I've got some questions and suggestions, once I understand better the user story I think I'll be able to confidently approve your PR.

Samyoul:
>there is no chat for these new messages, so we send 1-1 message but without chat, so im not sure how it will work without chat

I'd add them to an admin user's request log. Which will look like a list of action items.

>i use same message for request and reject and replace it in the table, but if in emojis author of the message the same , here authours will be different, not sure if its ok

This smells funny. If you are saving data but one element will be inconsistent, then I think you need to change your structure.

Something like `GroupChatApplicationRequest` and `GroupChatApplicationResponse`

```protobuf
message GroupChatApplicationRequest {
  // chat_id the ID of the private group chat the message belongs to, for query efficiency the chat_id is stored in the db even though the
  // target message also stores the chat_id
  string chat_id = 1;

  string invitation_id = 2;

  string introduction_message = 3;
} 
```

```protobuf
message GroupChatApplicationResponse{
  // chat_id the ID of the private group chat the message belongs to, for query efficiency the chat_id is stored in the db even though the
  // target message also stores the chat_id
  string chat_id = 1;

  string application_id = 2;

  enum response_status= 3;
} 
```

protocol/message_handler.go
flexsurfer: yeah i still not sure if we need this code or not

protocol/protobuf/group_chat_invitation.proto
flexsurfer: i believe for clock we need chat but we don't have it

protocol/message_validator.go
flexsurfer: yes its still wip

protocol/messenger.go
flexsurfer: will it be resent correctly without chat?

flexsurfer:
> Something like GroupChatApplicationRequest and GroupChatApplicationResponse
but in that case i can't use 
```
type GroupChatInvitation struct {
	protobuf.GroupChatInvitation

```
i need two structs then ?

protocol/group_chat_invitation.go
flexsurfer: could you elaborate please?

flexsurfer:
> I'd add them to an admin user's request log. Which will look like a list of action items.

not sure i understand this

protocol/group_chat_invitation.go
flexsurfer: do you mean i don't need MarshalJSON ?

protocol/group_chat_invitation.go
```diff
@@ -0,0 +1,60 @@
...
+	// From is a public key of the author of the invitation.
+	From string `json:"from,omitempty"`
+
+	// SigPubKey is the ecdsa encoded public key of the invitation author
+	SigPubKey *ecdsa.PublicKey `json:"-"`
+}
+
+// ID is the Keccak256() contatenation of From-ChatId
+func (g GroupChatInvitation) ID() string {
+	return types.EncodeHex(crypto.Keccak256([]byte(fmt.Sprintf("%s%s", g.From, g.ChatId))))
```
cammellos, line 29: This is non unique for multiple invitation from the same user, not sure if that's relevant, but worth pointing out

protocol/group_chat_invitation.go
```diff
@@ -0,0 +1,60 @@
...
+	"fmt"
+
+	"github.com/golang/protobuf/proto"
+
+	"github.com/status-im/status-go/eth-node/crypto"
+	"github.com/status-im/status-go/eth-node/types"
+	"github.com/status-im/status-go/protocol/protobuf"
+)
+
+// Invitation represents an group chat invitation request from a user in the application layer, used for persistence, querying and
```
Samyoul:
```suggestion
// Invitation represents a group chat invitation request from a user in the application layer, used for persistence, querying and
```

services/ext/api.go
```diff
@@ -233,6 +241,18 @@ func (api *PublicAPI) ChangeGroupChatName(ctx Context, chatID string, name strin
...
+}
+
+func (api *PublicAPI) GetGroupChatInvitations() ([]*protocol.GroupChatInvitation, error) {
+	return api.service.messenger.GetGroupChatInvitations()
+}
+
+func (api *PublicAPI) SendGroupChatInvitationRejection(ctx Context, invitationRequestID string) (*protocol.MessengerResponse, error) {
+	return api.service.messenger.SendGroupChatInvitationRejection(ctx, invitationRequestID)
+}
+
```
Samyoul, line 251: Should there be a `SendGroupChatInvitationAcceptance` method? How does the admin allow a user to join the group?

protocol/protobuf/group_chat_invitation.proto
Samyoul: Clock is basically a number. Items ordered by clock will show in date order. If items are ordered by clock descending then the items will be shown from oldest to newest, if ordered ascending the items will be shown newest to oldest.

protocol/protobuf/group_chat_invitation.proto
```diff
@@ -0,0 +1,13 @@
...
+package protobuf;
+
+message GroupChatInvitation {
+  // chat_id the ID of the private group chat the message belongs to, for query efficiency the chat_id is stored in the db even though the
+  // target message also stores the chat_id
+  string chat_id = 1;
+
+  string introduction_message = 2;
+
+  bool rejection = 3;
```
Samyoul: What is the default state of a group chat invitation?

If the state is false, how does the application distinguish the default state from a rejected state?
If the state is true, how does the application distinguish the default state from an accepted state?

Samyoul (commented):
Thank you for you work so far on this. I've added some questions to your PR.

In addition to the questions on the code, how are chat invitation links revoked?

services/ext/api.go
flexsurfer: regular flow for an invitation is used, so admin just sends a regular invitation

protocol/protobuf/group_chat_invitation.proto
flexsurfer: there is no accepted state, only two states requested and rejected

flexsurfer:
@cammellos for some reason after relogin i receive invitation message again in status-react, what could be the reason?

protocol/messenger.go
```diff
@@ -917,6 +944,134 @@ func (m *Messenger) ChangeGroupChatName(ctx context.Context, chatID string, name
...
+	if err != nil {
+		return nil, err
+	}
+
+	id, err := m.processor.SendPrivate(ctx, userpk, spec)
+	if err != nil {
+		return nil, err
+	}
+
+	//why do we need this code?
```
cammellos: We save messages that we send for retrying purpose, as we don't want to marshal/unmarshal again

protocol/messenger.go
```diff
@@ -1121,6 +1276,24 @@ func (m *Messenger) LeaveGroupChat(ctx context.Context, chatID string, remove bo
 	return &response, m.saveChat(chat)
 }
 
+func (m *Messenger) RemoveGroupChat(chatID string) (*MessengerResponse, error) {
```
cammellos: Currently we let the user handle this, using `SaveChat` setting directly `Active` to `false`, as `RemoveX` has some more behavior attached to it (we delete also the messages and set `DeletedAt`), so probably I would just handle this in status-react for now, or make a generic endpoint `RemoveChat` that deletes also the messages & set `DeletedAtClockValue`, otherwise it might be a bit confusing.

protocol/message_handler.go
```diff
@@ -54,10 +55,19 @@ func (m *MessageHandler) HandleMembershipUpdate(messageState *ReceivedMessageSta
 		return err
 	}
 
-	if chat == nil {
+	if chat == nil || len(chat.InvitationAdmin) > 0 {
```
cammellos, line 60: Here what's happening is that you create a group chat from an invitation (link) and you have received a response, correct? so the chat is there, but does not have the full events. 
Is that correct?
I would leave a comment here explaining this case, as it might not be immediately clear

protocol/messenger.go
```diff
@@ -722,6 +724,23 @@ func (m *Messenger) CreateGroupChatWithMembers(ctx context.Context, name string,
 	return &response, m.saveChat(&chat)
 }
 
+func (m *Messenger) CreateGroupChatFromInvitationLink(name string, chatID string, adminPK string) (*MessengerResponse, error) {
```
cammellos: Here we directly mention a `Link`, but there's not much code that deals with links here, 
so I would either remove the reference to `Link` (`CreateGroupChatFromInvitation`) or move the code for generating/parsing the link here, and not in status-react.
We should take into consideration that desktop as well should use the same methods, so we should keep the complexity of the frontend to a minimum, in order to re-use.
Up to you as you know best what the complexity of the frontend is.

protocol/message_handler.go
```diff
@@ -722,3 +731,32 @@ func (m *MessageHandler) HandleEmojiReaction(state *ReceivedMessageState, pbEmoj
...
+		return err
+	}
+
+	groupChatInvitation := &GroupChatInvitation{
+		GroupChatInvitation: pbGHInvitations,
+		SigPubKey:           state.CurrentMessageState.PublicKey,
+	}
+
+	if groupChatInvitation.Rejection {
+		groupChatInvitation.From = types.EncodeHex(crypto.FromECDSAPub(&m.identity.PublicKey))
```
cammellos, line 771: I am not sure I understand this,
why do we set `From` to us if it's a rejection?
That would change the ID of the invitation

protocol/message_handler.go
```diff
@@ -722,3 +731,32 @@ func (m *MessageHandler) HandleEmojiReaction(state *ReceivedMessageState, pbEmoj
 
 	return nil
 }
+
+func (m *MessageHandler) HandleGroupChatInvitation(state *ReceivedMessageState, pbGHInvitations protobuf.GroupChatInvitation) error {
+	logger := m.logger.With(zap.String("site", "HandleGroupChatInvitation"))
```
cammellos, line 757: Here you can check back if you have requests in the state so you don't process them multiple times, as messages are received multiple times.

cammellos (commented):
Amazing work, I am really impressed!!! It's good to see more people working on on the whole feature.

I have left some comments, overall looks good, just some clarification.
One thing I am not sure is the ID scheme (`from` + `chatId`), we can probably make it work, but it comes with some caveats, but we'll see.

protocol/group_chat_invitation.go
flexsurfer: yes, that's intentional, we should have more than one from same user

protocol/messenger.go
flexsurfer: thanks, yeah forgot to remove that comment

protocol/messenger.go
flexsurfer: could you elaborate on `just handle this in status-react ` ?

protocol/message_handler.go
flexsurfer: yes if there InvitationAdmin field means that chat created from link, comment will be useful for sure, thanks

protocol/messenger.go
flexsurfer: ok thanks, i'll remove Link, because links parsing logic is too general to move it in this PR

protocol/message_handler.go
flexsurfer: From is always author of the request, so i use the same invitation instance and change its state , but from it always the same and means the initial author of the request, probably i need to rename the field

flexsurfer: yes i was not sure about this code as well, i feel like something wrong here but not sure how to improve that

protocol/messenger.go
flexsurfer: got it thanks

protocol/messenger.go
```diff
@@ -917,6 +944,134 @@ func (m *Messenger) ChangeGroupChatName(ctx context.Context, chatID string, name
...
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+
+	invitationR, err := m.persistence.InvitationByID(invitationRequestID)
+	if err != nil {
+		return nil, err
+	}
+
+	invitationR.Rejection = true
+
```
Samyoul: Looking at this code you pull the invitation request from the db, which contains the `From` / `source` of the original requester, and then set the `Rejection` flag to true.

So there is no need to overwrite the `From` field when handling an incoming rejection message?

protocol/message_handler.go
Samyoul: The `&m.identity.PublicKey` refers to the identity of the current user, in the [rejection sending code](https://github.com/status-im/status-go/pull/2013/files/6c5e2f8fef9de45d64496e698ac81febbad2d4b2#r476370135) you set the `From` field to the original requester.

Did you find that the returning rejection message had a different `From` key than you expected?

protocol/messenger.go
flexsurfer: `From / source` is the PK of invitation author it's not changed

flexsurfer:
@Samyoul @cammellos addressed all comments, added clock and replaced rejection flag by state

flexsurfer:
@Samyoul @cammellos status-react has been tested and ready to be merged

cammellos (approved):
It'd be good to have some tests for the functionality, but we can do it in a second time, great work!

PR #2015, opened 2020-08-11
implement Makefile target for builing status-go as a shared library

michaelsbradleyjr:
A shared library is useful for libstatus consumers that experience collision of exported symbols during static linking.

It was discovered that on Windows `"-Wl,--export-all-symbols"` needs to be supplied in `CGO_LDFLAGS` when building the shared library else the expected symbols will not be found in `libstatus.dll`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,22 @@ PRE_RELEASE := "1"
 RELEASE_TYPE := $(shell if [ $(PRE_RELEASE) = "0" ] ; then echo release; else echo pre-release ; fi)
 GOLANGCI_BINARY=golangci-lint
 
+ifeq ($(OS),Windows_NT)     # is Windows_NT on XP, 2000, 7, Vista, 10...
+ detected_OS := Windows
+else
+ detected_OS := $(strip $(shell uname))
+endif
+
+ifeq ($(detected_OS),Darwin)
+ GOBIN_SHARED_LIB_EXT := dylib
+else ifeq ($(detected_OS),Windows)
+ # on Windows need `--export-all-symbols` flag else expected symbols will not be found in libstatus.dll
+ GOBIN_SHARED_LIB_CGO_LDFLAGS := CGO_LDFLAGS="-Wl,--export-all-symbols"
+ GOBIN_SHARED_LIB_EXT := dll
+else
+ GOBIN_SHARED_LIB_EXT := so
+endif
+
 help: ##@other Show this help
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -8,6 +8,21 @@ PRE_RELEASE := "1"
...
+ifeq ($(OS),Windows_NT)     # is Windows_NT on XP, 2000, 7, Vista, 10...
+ detected_OS := Windows
+else
+ detected_OS := $(strip $(shell uname))
+endif
+
+ifeq ($(detected_OS),Darwin)
+ GOBIN_SHARED_LIB_EXT := dylib
+else ifeq ($(detected_OS),Windows)
+ GOBIN_SHARED_LIB_CGO_LDFLAGS := CGO_LDFLAGS="-Wl,--export-all-symbols"
```
jakubgs, line 21: Why is this necessary only for windows?

jakubgs (approved):
I don't see anything wrong with this, but why do you need a shared library tho?

Makefile
michaelsbradleyjr: I discovered the need for that flag on Windows a few months ago when attempting to hook status-go (built as a shared library) to Clojure via JNA.  Without it, *none* of the expected symbols could be found in `libstatus.dll`. When creating this target for the sake of nim-status/-client, I tried compiling for Windows without the flag and had the same problem.

I searched and experimented for hours back in May before stumbling on this solution in some Stack Overflow comment (didn't save the link unfortunately). There may be a better way, but it works and I'm not aware of any downsides.

michaelsbradleyjr:
> I don't see anything wrong with this, but why do you need a shared library tho?

We've hit a snag where some Nim libraries used by nim-status/-client rely on C libs that export symbols that have the same names as symbols encountered when statically linking status-go, which causes compilation to fail. It was found that dynamically linking to status-go is a workaround. Maybe there's a better way? Definitely open to suggestions.

Makefile
jakubgs: Can we add a comment explaining why that's there then? Or explain it in the commit message?

michaelsbradleyjr: Sure, I'll add a comment and reword the commit message to include an explanation.

michaelsbradleyjr: Done.

jakubgs: Nice.

cammellos (approved):

PR #2017, opened 2020-08-17
Fix mark messages seen

cammellos:
If a chat had a last message the persistence call `Chat()` would not
decode `LastMessage` correctly.
This commit fixes the issue.

related issue in status-react: https://github.com/status-im/status-react/issues/11068


flexsurfer (approved):

hesterbruikman:
Check ahead of release 1.6 which is to include advertiser flow for referral program only (based on new contract) plus only high-prio fixes that would otherwise go into a `1.5.1`. Therefore it is likely based of `1.5` instead of `develop`.

Does this PR include any dependencies on `develop` that are not in [1.5 branch](https://github.com/status-im/status-react/commits/release/1.5.x)?

cammellos:
@hesterbruikman this PR has no dependencies on develop that are not on the 1.5 branch, we might have to cherry pick, but other that that is finkle

Samyoul:
Why didn't the persistence call `Chat()` decode `LastMessage` correctly? I don't quite understand why this would be the case?

PR #2018, opened 2020-08-18
Handle wallet initialization

cammellos:
StartWallet was called before service initialization.
After the recent changes this call was moved after initialization, but
the geth system automatically start services.
This meant that `IsStarted()` returned true, although the reactor was
not started, and only after calling `StopWallet()` and `StartWallet()`
again the system would reach the right state.

This commit changes the behavior so that we only check whether the
reactor has been started when calling `IsStarted()` and we allow
multiple calls to `Start()` on the signal service, which won't return an
error (it's a noop if callled multiple times).


flexsurfer (approved):

services/wallet/transmitter.go
```diff
@@ -24,7 +24,8 @@ type SignalsTransmitter struct {
 // Start runs loop in background.
 func (tmr *SignalsTransmitter) Start() error {
 	if tmr.quit != nil {
-		return errAlreadyRunning
+		// already running, nothing to do
+		return nil
```
Samyoul, line 28: This is nice

Samyoul (approved):

PR #2021, opened 2020-08-20
Allow node to start a push notification server

cammellos:
This commit allows a node to start a push notification server.
If the config is set it will start a messenger with a corresponding pn
server.

### DIFF
--- a/cmd/statusd/main.go
+++ b/cmd/statusd/main.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/okzk/sdnotify"
 	"golang.org/x/crypto/ssh/terminal"
 
@@ -20,12 +21,16 @@ import (
 	gethmetrics "github.com/ethereum/go-ethereum/metrics"
 
 	"github.com/status-im/status-go/api"
+	"github.com/status-im/status-go/appdatabase"
+	gethbridge "github.com/status-im/status-go/eth-node/bridge/geth"
+	"github.com/status-im/status-go/eth-node/crypto"
 	"github.com/status-im/status-go/logutils"
 	"github.com/status-im/status-go/metrics"
 	nodemetrics "github.com/status-im/status-go/metrics/node"
 	"github.com/status-im/status-go/node"
 	"github.com/status-im/status-go/params"
 	"github.com/status-im/status-go/profiling"
+	"github.com/status-im/status-go/protocol"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


protocol/pushnotificationserver/server.go
```diff
@@ -21,6 +21,7 @@ const encryptedPayloadKeyLength = 16
 const defaultGorushURL = "https://gorush.status.im"
 
 type Config struct {
+	Enabled bool
```
jakubgs, line 24: Doesn't this kind break how we do things? All config structs are in `params/config.go`, why should this one be here?

jakubgs: Same for `defaultGorushURL`, should that be in `params/defaults.go`?

cmd/statusd/main.go
```diff
@@ -150,6 +155,50 @@ func main() {
 		return
 	}
 
+	if config.PushNotificationServerConfig.Enabled {
+		if config.NodeKey == "" {
+			logger.Error("node key needs to be set if running a push notification server")
+			return
```
jakubgs, line 161: Why does it need a `NodeKey` set though? Why would the generated one be bad? Because it would change at restart?

cammellos: Yes, node key should be stable, otherwise client will not be able to communicate with the server anymore

protocol/pushnotificationserver/server.go
cammellos: I don't know, `Messenger` config is in messenger, as other configs that are more "local", generally I am not sure it's a good pattern to have everything in `params/config`. 

I think the config should be close to where it is used, for better encapsulation and clarity, I don't see why `protocol/pushnotificationclient` namespace should depend on a top level `params/config` namespace, and I'd be vary of scattering logic around.
What do you think?

cmd/statusd/main.go
```diff
@@ -150,6 +155,50 @@ func main() {
...
+			protocol.WithDatabase(db),
+		}
+
+		messenger, err := protocol.NewMessenger(identity, gethbridge.NewNodeBridge(backend.StatusNode().GethNode()), installationID.String(), options...)
+		if err != nil {
+			logger.Error("failed to create messenger", "error", err)
+			return
+		}
+
+		err = messenger.Start()
```
jakubgs, line 197: Okay, we start the messenger, but where do we start the norifications server?

cammellos: That's based on the config, the messenger will start it if passed `PushNotificationServer` -> `Enabled` on start

protocol/pushnotificationserver/server.go
jakubgs: Right, so we already have exceptions. I'd say it should be one or the other, either all in `params/config.go` or close to where it's used, as you say. Since we already have a mix there's no fixing this now without doing a lot of busywork that doesn't provide much value. Ugh.

@Samyoul what do you think? I guess it's fine. I just need to give up on consistency in this package.

cmd/statusd/main.go
jakubgs: Makes sense.

cmd/statusd/main.go
jakubgs: I see, could we add a comment here making that clear?

cammellos: done

jakubgs (approved):
The inconsistencies in how we do configs is what it is. It's fine.

PR #2022, opened 2020-08-20
Enable push notifications by default

cammellos:
depends on #2022 

This PR does a few things:
- Enable sending push notifications by default
- Process push notifications at-most-once on the server side
- Filter by chatID push notifications on server side
- Set default push notifications servers when starting the app


config/cli/fleet-eth.prod.json
```diff
@@ -23,6 +23,17 @@
...
+      "04c42f368a23fa98ee546fd247220759062323249ef657d26d357a777443aec04db1b29a3a22ef3e7c548e18493ddaf51a31b0aed6079bd6ebe5ae838fcfaf3a49",
+      "04ee2b53b0ace9692167a410514bca3024695dbf0e1a68e1dff9716da620efb195f04a4b9e873fb9b74ac84de801106c465b8e2b6c4f0d93b8749d1578bfcaf03e",
+      "042c8de3cbb27a3d30cbb5b3e003bc722b126f5aef82e2052aaef032ca94e0c7ad219e533ba88c70585ebd802de206693255335b100307645ab5170e88620d2a81",
+      "047aa648d6e855950b2e3d3bf220c496e0cae4adfddef3e1e6062e6b177aec93bc6cdcf1282cb40d1656932ebfdd565729da440368d7c4da7dbd4d004b1ac02bf8",
+      "0430211cbd81c25f07b03a0196d56e6ce4604bb13db773ff1c0ea2253547fafd6c06eae6ad3533e2ba39d59564cfbdbb5e2ce7c137a5ebb85e99dcfc7a75f99f55",
+      "04e85f1d4209f2f99da801af18db8716e584a28ad0bdc47fbdcd8f26af74dbd97fc279144680553ec7cd9092afe683ddea1e0f9fc571ebcb4b1d857c03a088853d",
+      "048a64b3c349a2e0ef4a32ea49609ed6eb3364be1110253c20adc17a3cebbc39a219e5d3e13b151c0eee5d8e0f9a8ba2cd026014e67b41a4ab7d1d5dd67ca27427",
+      "0444160e22e8b42bd32a06c1532165fa9e096eebedd7fa6d6e5f8bbef0440bc4a4591fe3651be68193a7ec029021cdb496cfe1d7f9f1dc69eb99226e6f39a7a5d4"
+    ],
```
Samyoul, line 36: Curious more than anything. These look like a list of public keys, how are they defined? Were they just randomly generated? Where are the corresponding private keys stored (if stored at all)?

protocol/messenger.go
```diff
@@ -78,6 +78,7 @@ type Messenger struct {
 	installationID             string
 	mailserver                 []byte
 	database                   *sql.DB
+	quit                       chan struct{}
```
Samyoul, line 81: Why is this a `struct{]` and not a bool for example?

Does the `quit` field have a write event? I can't see it?

protocol/pushnotificationclient/client.go
```diff
@@ -287,15 +314,47 @@ func (c *Client) Reregister(contactIDs []*ecdsa.PublicKey, mutedChatIDs []string
...
 }
 
+// pickDefaultServesr picks n servers at random
+func (c *Client) pickDefaultServers(servers []*ecdsa.PublicKey) []*ecdsa.PublicKey {
+	// shuffle and pick n at random
+	shuffledServers := make([]*ecdsa.PublicKey, len(servers))
+	copy(shuffledServers, c.config.DefaultServers)
+	mrand.Shuffle(len(shuffledServers), func(i, j int) {
+		shuffledServers[i], shuffledServers[j] = shuffledServers[j], shuffledServers[i]
+	})
```
Samyoul, line 325: `mrand` doesn't seem to have its seed set. This will default to the same seed each time.

Samyoul (approved):
Lot's of good work, the only point of concern is setting the `mrand` seed

config/cli/fleet-eth.prod.json
cammellos: PN servers will run the same as mailservers, so these are the public keys of the mailservers (you can check the enode://.. above, and they should be the same)

protocol/messenger.go
cammellos: not sure, I always seen that being used, I think it's because of memory https://stackoverflow.com/questions/52035390/why-using-chan-struct-when-wait-something-done-not-chan-interface

protocol/pushnotificationclient/client.go
cammellos: ah didn't know, good point!

protocol/messenger.go
cammellos: it's closed in `Shutdown()` https://github.com/status-im/status-go/pull/2022/files#diff-5df3979bc10f2ff16f7942ec557947ecR570

protocol/pushnotificationclient/client.go
Samyoul: It's a quirk of go, in go the random libraries aren't random by default.

protocol/messenger.go
Samyoul: That's interesting. It feels hacky, but I see why one would use `struct{}`.

config/cli/fleet-eth.prod.json
jakubgs: Why are these just public keys and not full enodes?

cammellos:
because push notifications servers are status clients, same as a status any
other client, it s not through devp2p, so you communicate with them through
waku and not directly

protocol/pushnotificationclient/client.go
```diff
@@ -60,13 +60,21 @@ const pushNotificationBackoffTime int64 = 2
 // RegistrationBackoffTime is the step of the exponential backoff
 const RegistrationBackoffTime int64 = 15
 
+type ServerType int
+
+const (
+	ServerTypeDefault = iota + 1
```
jakubgs, line 69: This is some black magic.

Samyoul: iota is one of my favourite features of go.

PR #2023, opened 2020-08-20
[#11046] Add local contact names

flexsurfer:
added nickname to contacts table and sync contact if nickname has changed


protocol/messenger.go
```diff
@@ -1215,7 +1223,7 @@ func (m *Messenger) saveContact(contact *Contact) error {
 	contact.Identicon = identicon
 	contact.Alias = name
 
-	if m.isNewContact(contact) {
+	if m.isNewContact(contact) || m.isNicknameHasChanged(contact) {
```
cammellos: `hasNicknameChanged` is probably a better name

protocol/message_handler.go
```diff
@@ -209,6 +209,8 @@ func (m *MessageHandler) HandleSyncInstallationContact(state *ReceivedMessageSta
 		}
 		contact.Photo = message.ProfileImage
 		contact.LastUpdated = message.Clock
+		contact.Nickname = message.Nickname
```
cammellos: Could you also update the tests (there should be one in `messenger_installations_test.go`) in order to make sure the nickname is also synced?

cammellos (changes requested):
looks good, just maybe adding test.
Also I would probably rename `nickname` to `localName` or `localContactName` (something so it's clear that is set locally, rather than remotely), but not sure what a good name would be

Samyoul (dismissed):
Great work @flexsurfer ! You're really doing some great stuff in `status-go`.

This PR looks great, the only suggestion I'd make is to rename references to `Nickname` to `LocalNickname` to communicate that the nickname is only for the local user.

In the case that we implement display names we want to make a clear distinction between the two.

Alternatively add some documentation to the code to make explicit that the nickname is for local purposes.

flexsurfer:
@cammellos, @Samyoul thanks, addressed all comments

protocol/messenger_installations_test.go
```diff
@@ -249,6 +250,7 @@ func (s *MessengerInstallationSuite) TestSyncInstallation() {
 	s.Require().NotNil(statusChat)
 
 	s.Require().True(actualContact.IsAdded())
+	s.Require().True(actualContact.LocalNickname == "Test Nickname")
```
cammellos:
```suggestion
	s.Require().Equal("Test Nickname", actualContact.LocalNickname)
```

cammellos (approved):
Good job! just added a suggestion

Samyoul:
@flexsurfer Great work. Congrats on the merge.

I think that this functionality would also need an update to the specs, probably in [payloads syncinstallationcontact](https://github.com/status-im/specs/blob/master/docs/stable/6-payloads.md#syncinstallationcontact)

PR #2025, opened 2020-08-26
Honor FullNode config in waku

cammellos:
previously FullNode would only result in setting a bloom filter to full.

This behavior caused issues as you effectively cannot install a filter
on a FullNode, as it would advertise the new topic/bloom filter and stop
receiving all the messages.

This caused an issue when running push notifications servers together
with mailservers, also the behavior is a bit counter-intuitive as I
would expect the FullNode config to be honored no matter of what filters
are installed.

This is useful as well for desktop, when we might want to run a full node, but still use the node as a client.


waku/waku.go
```diff
@@ -1209,6 +1249,20 @@ func (w *Waku) SetBloomFilterMode(mode bool) {
 	// Recalculate and notify topic interest or bloom, currently not implemented
 }
 
+func (w *Waku) SetFullNode(set bool) {
+	w.settingsMu.Lock()
+	w.settings.FullNode = set
+	w.settingsMu.Unlock()
```
Samyoul, line 1255: I like this. No deferring the unlock, simple unlock immediately after use.

Samyoul (approved):
Looks good. An interesting bug.

PR #2026, opened 2020-08-27
Local notifications service

### LINKED ISSUE #11101: In-app Push Notifications for transactions
https://github.com/status-im/status/issues/5

As a user, I want to receive notification for actions that happen in the app which are not in my current focus, so thus I can instantly react and switch to the new actions.

Add push notification to specific transactions and accounts.

This change will also include the migration to new flow for building the notifications inside status go. Initially pushing only the transaction notifications.

Status go will keep a service listening to new messages (later) and transactions (now). When they happen, inside status go we build the notification and send it to the app using the new signal `notification`.

```
{
"id": "A unique notification id",
"title": "Notification title",
"body": "Notification message", (Could be an object if we want to build i18n strings during handling of signal)
"platform": "Name of platform if the notification is platform-specific",
"category": "category name, to show actions like reply, (Could be also defined on the build layer based on body fields)",
"image": "Url to show the image if present,"
"dl": "deep link to app screen",
}
```

On Android, a thin overlay over native notification module will keep a background service and show local notifications.

On iOs, we can also write a thin overlay in native code, or we can use directly ClojureScript as the OS does not keep our app in the background for a long time.

The status go knows the current-chat-id and can cancel the sending of specific notification if the user is currently in the chat.

The watching of transactions can work similarly to the current implementation https://github.com/status-im/status-react/blob/08ff2edb0be798e5499c8c90024dd01b6d55865b/src/status_im/ethereum/transactions/core.cljs#L88

Status go will hold also all the state values (subscribed/unsubscribed) for specific notification types and chats/addresses.
One way to keep them as a table of `[type[enum], topic[object], preference[enum ]]`

Example:
`[chat,{(chat-type) private, chat-id, (message-type)mention},subscribed]`
`["chat", {"private", "0x0...", "new-message"}, subscribed]`

Note that currently, there is an active task to improve watching transactions #9987

*All the technical details are just an opinion and not a specification

Ferossgp:
Fixes https://github.com/status-im/status-react/issues/11101

The Local Notification service is a thin layer that processes events from other services and builds a notification from it. Current PR implements only watching for transactions, by watching all `newblock` event from wallet service.


api/geth_backend.go
```diff
@@ -886,18 +898,19 @@ func (b *GethStatusBackend) AppStateChange(state string) {
 				return
 			}
 		}
-	} else if s == appStateBackground && !b.forceStopWallet {
```
Ferossgp, line 891: How we approach this? We need the wallet running in the background so we can watch transactions

services/local-notifications/core.go
```diff
@@ -0,0 +1,322 @@
...
+
+	if preference.Enabled {
+		s.StartWalletWatcher()
+	}
+
+	return nil
+}
+
+// StartWalletWatcher - Forward wallet events to notifications
+func (s *Service) StartWalletWatcher() {
```
Ferossgp, line 201: Is the approach used here to subscribe to the wallet correct in terms of code style?

services/local-notifications/core.go
```diff
@@ -0,0 +1,322 @@
...
+	if err != nil {
+		log.Debug("Could not select To account by address", "error", err)
+	}
+
+	body := notificationBody{
+		State:       state,
+		From:        transfer.From,
+		To:          transfer.Address,
+		FromAccount: from,
+		ToAccount:   to,
```
Ferossgp, line 147: The account here is not transformed into JSON 🤔

services/local-notifications/core.go
```diff
@@ -0,0 +1,322 @@
...
+		state = outbound
+	}
+
+	from, err := s.accountsDB.GetAccountByAddress(types.Address(transfer.From))
+
+	if err != nil {
+		log.Debug("Could not select From account by address", "error", err)
+	}
+
+	to, err := s.accountsDB.GetAccountByAddress(types.Address(transfer.To))
```
Ferossgp, line 130: We use different types for the address in the Multiaccount module and the Wallet, implementation is the same as seems like the one used in `"github.com/status-im/status-go/eth-node/types"` is a fork from `"github.com/ethereum/go-ethereum/common"`. IMO we should use everywhere the `go-ethereum` types

appdatabase/migrations/sql/0014_local_notifications_preferences.up.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE local_notifications_preferences (
```
Ferossgp: Empirically, I think that with a 3-dimensional table we can handle all types of notifications.

services/wallet/transmitter.go
```diff
@@ -16,7 +16,6 @@ type publisher interface {
 // SignalsTransmitter transmits received events as wallet signals.
 type SignalsTransmitter struct {
 	publisher
-
 	wg   sync.WaitGroup
 	quit chan struct{}
 }
```
Samyoul: Just a little point, the go convention is to separate sync related struct fields by a line from the other struct fields. So this line was most likely intentional.

services/local-notifications/core.go
```diff
@@ -0,0 +1,333 @@
...
+func pushMessage(notification *Notification) {
+	log.Info("Pushing a new push notification", "info", notification)
+	signal.SendLocalNotifications(notification)
+}
+
+func (s *Service) buildTransactionNotification(rawTransfer wallet.Transfer) *Notification {
+	log.Info("Handled a new transfer in buildTransactionNotification", "info", rawTransfer)
+
+	var deeplink string
+	var state = undetermined
```
cammellos: You can probably do:
`state := undetermined`

services/local-notifications/core.go
cammellos: I believe this was done as a part of the previous effort to move to nimbus, but now it has been discontinued. 
We could probably just remove all the types and go back to `eth-node/types`, as it seems like it's not going to be picked up the work.

cammellos (commented):
Looks good to me, just a few observations, @Samyoul might have a better idea on what's idiomatic in go. As already discussed it would be good to have better test coverage, but amazing work @Ferossgp !

services/local-notifications/core.go
Samyoul: Nice early returns, ~~but I do have a question~~

I did have a question, but I'd misread the code, I saw the add to waitgroup as a wait call 😬

services/local-notifications/core.go
```diff
@@ -0,0 +1,338 @@
...
+						Type:                      string(event.Type),
+						BlockNumber:               event.BlockNumber,
+						Accounts:                  []common.Address(event.Accounts),
+						NewTransactionsPerAccount: map[common.Address]int(event.NewTransactionsPerAccount),
+						ERC20:                     bool(event.ERC20),
+					})
+				}
+			}
+		}
+	}()
```
Samyoul, line 249: What is the functional difference between this implementation and the below implementation?

```suggestion
	for {
		select {
		case <-s.walletTransmitter.quit:
			sub.Unsubscribe()
			return
		case err := <-sub.Err():
			// technically event.Feed cannot send an error to subscription.Err channel.
			// the only time we will get an event is when that channel is closed.
			if err != nil {
				log.Error("wallet signals transmitter failed with", "error", err)
			}
			return
		case event := <-events:
				if event.Type == wallet.EventNewBlock {
				s.transmitter.publisher.Send(TransactionEvent{
					Type:                      string(event.Type),
					BlockNumber:               event.BlockNumber,
					Accounts:                  []common.Address(event.Accounts),
					NewTransactionsPerAccount: map[common.Address]int(event.NewTransactionsPerAccount),
					ERC20:                     bool(event.ERC20),
				})
			}
		}
	}
```

We start a subroutine but block the parent function by adding to the wait group.

So either we don't need to wait and run a routine, or we do need the routine but we don't need to wait.

services/local-notifications/core.go
```diff
@@ -0,0 +1,338 @@
...
+			case err := <-sub.Err():
+				if err != nil {
+					log.Error("Local notifications transmitter failed with", "error", err)
+				}
+				return
+			case event := <-events:
+				s.transactionsHandler(event)
+			}
+		}
+	}()
```
Samyoul, line 293: Same for this code. If the parent function's progress is blocked by the subroutine and no other work is performed, do we need to have a subroutine?

Samyoul (commented):
Really good job, generally a great PR. 🎉🎆

~~I've got some questions about your use of `go func`s, but these are quite tricky things when starting with go.~~

I did have a question, but I'd read the code wrong. Soooo, I have no questions.

Samyoul (approved):

cammellos (approved):

PR #2028, opened 2020-08-27
Don't log chat as it makes status-desktop blow up

cammellos:
For some reason when calling saveChat from desktop with `lastMessage`
set to null, a sigsev is received.
The issue seems to be in logFormat
https://github.com/ethereum/go-ethereum/blob/05280a7ae3f47adc8aeb9130c7f5404a42fb3a55/log/format.go#L356
which for some reason blows up if passed a nil pointer (`lastMessage`).
Can't replicate on any other platform (ios,android,nodejs,status-console-clien) or running it locally, but hey,
this fixes the issue on desktop.

`¯\_(ツ)_/¯ ¯\_(ツ)_/¯ ¯\_(ツ)_/¯  ¯\_(ツ)_/¯ ¯\_(ツ)_/¯ ¯\_(ツ)_/¯ ¯\_(ツ)_/¯ ¯\_(ツ)_/¯`


jakubgs (approved):
Damn, that's dumb.

vitvly:
This was also causing issues during channel loading for status-react+nim-status builds in https://github.com/status-im/status-react/pull/10936. Just rebuilt iOS with the latest status-go version and the problem is gone.

No issues on Android though.

Thank you so much @cammellos !

PR #2029, opened 2020-08-31
Drop messages with text longer than 4096 characters

vkjr:
Drop messages with text longer than 4096 characters

New check added to `ValidateReceivedChatMessage` function along with test for it.

Closes TBD


protocol/message_validator.go
```diff
@@ -163,6 +164,11 @@ func ValidateReceivedChatMessage(message *protobuf.ChatMessage, whisperTimestamp
 		return errors.New("text can't be empty")
 	}
 
+	const maxLength = 4096
```
cammellos: probably this should be better placed at the top of the file, and maybe given a slightly different name:
`const maxChatMessageTextLength = 4096`

protocol/message_validator_test.go
```diff
@@ -193,6 +193,21 @@ func (s *MessageValidatorSuite) TestValidatePlainTextMessage() {
...
... magnam aliquam quaerat voluptatem. Ut enim ad minima veniam, quis nostrum exercitationem ullam corporis suscipit laboriosam, nisi ut aliquid ex ea commodi consequatur? Quis autem vel eum iure reprehenderit qui in ea voluptate velit esse quam nihil molestiae consequatur, vel illum qui dolorem eum fugiat quo voluptas nulla pariatur?Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo inventore veritatis et quasi architecto beatae vitae dicta sunt explicabo. Nemo enim ipsam voluptatem quia voluptas sit aspernatur aut odit aut fugit, sed quia consequuntur magni dolores eos qui ratione voluptatem sequi nesciunt. Neque porro quisquam est, qui dolorem ipsum quia dolor sit amet, consectetur, adipisci velit, sed quia non numquam eius modi tempora incidunt ut labore et dolore magnam aliquam quaerat voluptatem. Ut enim ad minima veniam, quis nostrum exercitationem ullam corporis suscipit laboriosam, nisi ut aliquid ex ea commodi consequatur? Quis autem vel eum iure reprehenderit qui in ea voluptate velit esse quam nihil molestiae consequatur, vel illum qui dolorem eum fugiat quo voluptas nulla pariatur?",
```
cammellos: probably best if we generate the string, something like
```
Text: fmt.Sprintf("%x", make([]byte, maxLength/2 + 1)) // %x will take 2 spaces for each byte
```

cammellos (approved):
Looks all good, only a couple of minor points!

protocol/message_validator_test.go
vkjr: Thanks! For sure, this is a much better approach :)

PR #2030, opened 2020-09-01
Add push notifications for mentions

cammellos:
### Why make this change?

This change allows users to receive push notification for mentions from public chat they have joined.

### What has changed?
- `protocol/message.go` moved to `protocol/common/message.go` as used now by `pushnotificationclient` namespace
- added and stored `Mentions` field in new message
- modified protobuf to have a `BlockMentions` field to indicate that we don't want to receive any push notification for mentions
- modified protobuf to have a `AllowedMentionsChatList` field which is the shake256 of the public chats we joined
- added settings for block mentions
- we stop sending messages in private group chats for user who have not joined, this was a pre-v1 backward compatibility behavior, which is not needed anymore and causes issues when sending push notifications

Specs PR: https://github.com/status-im/specs/pull/148


peers/peerpool_test.go
```diff
@@ -30,64 +30,36 @@ import (
...
-	}
-
-	l, err := net.ListenTCP("tcp", addr)
-	if err != nil {
-		return 0, err
-	}
-	defer l.Close()
-	return l.Addr().(*net.TCPAddr).Port, nil
-}
-
```
Samyoul, line 47: Ah, this is the fix for the racey test. Good job.

protocol/chat.go
```diff
@@ -111,6 +112,16 @@ func (c *Chat) MembersAsPublicKeys() ([]*ecdsa.PublicKey, error) {
 	return stringSliceToPublicKeys(publicKeys, true)
 }
 
+func (c *Chat) JoinedMembersAsPublicKeys() ([]*ecdsa.PublicKey, error) {
+	var publicKeys []string
+	for _, item := range c.Members {
+		if item.Joined {
+			publicKeys = append(publicKeys, item.ID)
+		}
+	}
```
Samyoul: No problem with the logic of this, but the human readability is more intuitive if the `item` is named after what it represents.

Then you get a nice line `if member.Joined`. At least that's my preference.

```suggestion
	for _, member := range c.Members {
		if member.Joined {
			publicKeys = append(publicKeys, member.ID)
		}
	}
```

protocol/message_handler.go
```diff
@@ -218,19 +218,19 @@ func (m *MessageHandler) HandleSyncInstallationContact(state *ReceivedMessageSta
...
 	}
 
 	chat := CreatePublicChat(chatID, state.Timesource)
 
 	state.AllChats[chat.ID] = &chat
 	state.ModifiedChats[chat.ID] = true
 
-	return nil
+	return true, nil
 }
```
Samyoul: This is an interesting fix. Would this function ever return an error? It doesn't seem that it would.

protocol/messenger.go
```diff
@@ -3275,12 +3309,13 @@ func (m *Messenger) EnableSendingPushNotifications() error {
 	return nil
 }
 
-func (m *Messenger) addedContactsAndMutedChatIDs() ([]*ecdsa.PublicKey, []string) {
+func (m *Messenger) pushNotificationOptions() *pushnotificationclient.RegistrationOptions {
```
Samyoul, line 3521: I like this rename

protocol/persistence_test.go
```diff
@@ -322,14 +325,14 @@ func TestDeleteMessagesByChatID(t *testing.T) {
 	err = insertMinimalMessage(p, "2")
 	require.NoError(t, err)
 
-	m, _, err := p.MessageByChatID("chat-id", "", 10)
+	m, _, err := p.MessageByChatID(testPublicChatID, "", 10)
```
Samyoul, line 328: I like the change to a high level var 👍

protocol/pushnotificationclient/client.go
```diff
@@ -797,6 +829,109 @@ func (c *Client) handleMessageSent(sentMessage *common.SentMessage) error {
...
+			time.Sleep(300 * time.Millisecond)
+			continue
+		} else if err != nil {
+			return nil, err
+		}
+
+		return message, nil
+	}
+	return nil, sql.ErrNoRows
+}
```
Samyoul, line 853: We're presuming that an incoming message id exists or at least should exist, otherwise the application will sleep for ~3 seconds. That seems reasonable, were we getting circumstances where the calls for messages were returning nil when the message should be in the db?

protocol/pushnotificationclient/client.go
```diff
@@ -797,6 +829,109 @@ func (c *Client) handleMessageSent(sentMessage *common.SentMessage) error {
...
+
+// handlePublicMessageSent handles public messages, we notify only on mentions
+func (c *Client) handlePublicMessageSent(sentMessage *common.SentMessage) error {
+	// We always expect a single message, as we never batch them
+	if len(sentMessage.MessageIDs) != 1 {
+		return errors.New("batched public messages not handled")
+	}
```
Samyoul, line 860: If we never batch the messages, should `SentMessage.MessageIDs` be just an int? Or are batch messages used in other contexts?

protocol/common/message.go
```diff
@@ -294,10 +301,33 @@ func (m *Message) parseAudio() error {
...
+	// only on entering we fetch, otherwise we go on
+	if !entering {
+		return ast.GoToNext
+	}
+	switch n := node.(type) {
+	case *ast.Mention:
+		v.mentions = append(v.mentions, string(n.Literal))
+	}
+	return ast.GoToNext
+}
```
Samyoul, line 320: Maybe add a note to state that this is an implementation of the `github.com/status-im/markdown/ast.NodeVisitor` interface. I got confused initially trying to understand the purpose of this struct and the Visit() function

Samyoul (approved):
Great Job @cammellos . A few questions and recommendations, but nothing that should block the merge of this PR

protocol/pushnotificationclient/client.go
cammellos: public chats are never batched but private & group might be batched, yes

protocol/pushnotificationclient/client.go
cammellos: it's new code, but yes, essentially it might happen, as we schedule first, dispatch and then save, and subscribers are notified asynchronously

PR #2032, opened 2020-09-09
Soft blacklist peer ids

cammellos:
### Why make this change?

This change is useful so we can blacklist peer ids, without actually
disconnecting them, so that it's harder for them to tell that they are
being blacklisted. Envelopes will just be dropped.

### What has changed?

In waku we pass a config with a list of hex-encoded peer ids, and we
check that against newly received envelopes.
If so we just drop the envelope and return

### DIFF
--- a/waku/waku_test.go
+++ b/waku/waku_test.go
@@ -31,9 +31,13 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/p2p"
+	"github.com/ethereum/go-ethereum/p2p/enode"
 
+	"github.com/status-im/status-go/eth-node/types"
 	"github.com/status-im/status-go/waku/common"
 	v0 "github.com/status-im/status-go/waku/v0"
+	v1 "github.com/status-im/status-go/waku/v1"
 )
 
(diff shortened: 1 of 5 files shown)


waku/waku_test.go
```diff
@@ -1065,6 +1068,31 @@ func TestTopicInterest(t *testing.T) {
...
+	cfg := &Config{
+		SoftBlacklistedPeerIDs: []string{types.EncodeHex(peer.ID())},
+	}
+	w2 := New(cfg, nil)
+
+	envelopeError, err = w2.OnNewEnvelopes([]*common.Envelope{envelope}, peer)
+	require.NoError(t, err)
+	// Since it's blacklisted, it will just drop envelopes, keep the connection open
+	require.Nil(t, envelopeError)
+}
```
Samyoul, line 1109: Would it make sense to test that the peers are still connected?

Samyoul (approved):
Really smart idea. I like the silent blacklisting.

waku/waku_test.go
cammellos: Connection is handled at the peer level (so v0/v1), in this case as long as we don't return `err` from `OnNewEnvelopes` the peer will not be disconnected, so we should be safe as implicitly tested

PR #2034, opened 2020-09-09
[mentions] ParsedText in quoted message

rasom:
ParsedText is needed to properly display mention in quoted message.

status: ready


cammellos (approved):

VERSION
```diff
@@ -1 +1 @@
-0.60.0
+0.60.1
```
Samyoul: https://github.com/status-im/status-go/pull/2033/files already has a version bump to `0.60.1` maybe set to `0.60.2`?

Samyoul (approved):

VERSION
cammellos: Yep, merged, need to upgrade version

PR #2035, opened 2020-09-10
Add `allow permission request` field to settings

vkjr:
New settings field added that is needed for [PR](https://github.com/status-im/status-react/pull/11034)
When it is false - webpages can't request permission to use camera. When it is true - they can.


cammellos (approved):

VERSION
```diff
@@ -1 +1 @@
-0.61.0
+0.61.1
```
cammellos: Newline at the end of the file, the tests are complaining

vkjr: added 👌

PR #2036, opened 2020-09-10
Add index on seen & hide long messages

cammellos:
This commit does two things:
1) Add an index on seen & update only the not-seen messages in the query
2) Hide long messages in the database, as that's likely spam


flexsurfer (approved):

protocol/migrations/sqlite/1599720851_add_seen_index_remove_long_messages.down.sql
```diff
@@ -0,0 +1,2 @@
+DROP INDEX seen_local_chat_id_idx;
+UPDATE user_messages SET hide = 0 WHERE length(text) > 4096;
```
rasom, line 2: Did we have `hide` before this commit? Does it mean that long messages which were hidden for a different reason will become visible? I realise that's not a big deal, just asking

cammellos: we did have hide before and it's correct that they would be shown, although we only hide transaction messages, so that should be not an issue.

The thing is that we never run down migrations, as we have no control over the deployment, so this is effectively unused.

rasom: ok, thanks

PR #2037, opened 2020-09-10
Add API functions to store pending transaction data

richard-ramos:
This PR adds functions to insert/retrieve/delete pending transactions containing the transaction hash, address, type of transaction (what originated it) and a data attribute which can be used to store extra information related to the transaction.

This is useful to check the state of pending transactions when login in the app, (like when you register an ENS username, or buy a sticker pack) because as soon as you close the app, this pending state is gone.

With this PR the flow to check a pending transaction state would be like this, using a ENS registration as example:

1. User register ENS username
2. StorePendingTransaction is called, with the following values: 
  - The transaction hash
  - The wallet address
  - Type: `"RegisterENS"`
  - Block number at the moment of posting the transaction
  - Data: the ens username "someuser.stateofus.eth"  
  (if the transaction was to buy a sticker pack, the type would be "BuyStickerPack" and the data the sticker pack id)
3. UI will display the username as being pending (cannot be used in the meantime)

In the signal handling block,  when a `new-block` wallet signal is received
  - Call `GetPendingTransactionsByAddress( account )`
  - If a pending transaction is returned, call `eth_getTransactionReceipt` using the transaction hash of the pending transaction
  - if there is a receipt, check if the status is 0x1 to know if the transaction did not revert, and remove the pending state of the username obtained from the data attribute of the pending transaction
  - if there is no receipt, the transaction is still pending

On login,
  - Call `GetPendingTransactions( )`, this will return all pending transactions that were stored
  - If a pending transaction is returned, call `eth_getTransactionReceipt` using the transaction hash of the pending transaction
  - if there is a receipt, check if the status is 0x1 to know if the transaction did not revert, and remove the pending state of the username obtained from the data attribute of the pending transaction
  - if there is no receipt, the transaction is still pending

The block number can be used to identify if a transaction is taking too much time to be processed by comparing it against the current block number, to then suggest the user the possibility of cancelling the transaction or speeding it up.


services/wallet/database_test.go
```diff
@@ -204,3 +204,53 @@ func TestCustomTokens(t *testing.T) {
...
+	defer stop()
+
+	trx := PendingTransaction{
+		TransactionHash: common.Hash{1},
+		BlockNumber:     big.NewInt(1),
+		Address:         common.Address{1},
+		Type:            RegisterENS,
+		Data:            "someuser.stateofus.eth",
+	}
+
```
Samyoul: Tests make me happy 😄

appdatabase/migrations/sql/0012_pending_transactions.up.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE IF NOT EXISTS pending_transactions (
+  network_id UNSIGNED BIGINT NOT NULL,
```
Samyoul, line 2: +1 for `network_id`

appdatabase/migrations/sql/0012_pending_transactions.up.sql
```diff
@@ -0,0 +1,10 @@
...
+  transaction_hash VARCHAR NOT NULL,
+  blk_number BIGINT NOT NULL,
+  address VARCHAR NOT NULL,
```
Samyoul: Could you add a `to_address` and a `from_address` to allow for both in bound and out bound pending transactions to be stored?

services/wallet/database.go
```diff
@@ -583,6 +583,80 @@ func (db *Database) DeleteCustomToken(address common.Address) error {
...
+	RegisterENS    PendingTrxType = "RegisterENS"
+	ReleaseENS     PendingTrxType = "ReleaseENS"
+	SetPubKey      PendingTrxType = "SetPubKey"
+	BuyStickerPack PendingTrxType = "BuyStickerPack"
+)
+
+type PendingTransaction struct {
+	TransactionHash common.Hash    `json:"transactionHash"`
+	BlockNumber     *big.Int       `json:"blockNumber"`
+	Address         common.Address `json:"address"`
```
Samyoul: Same as the DB schema, could you add a `ToAddress` and a `FromAddress`?

Samyoul (changes requested):
Congratulations on your first PR to `status-go`! 🎉

Really clean work.

I initially I had a request to change your methods names to include "Outbound" to make a distinction between inbound and outbound transactions. However if we include the transaction `From` field we will be able to use this code for both inbound and outbound pending transactions, along with query capability.

Example get all pending transactions from my address, get all pending transactions to my address.

Samyoul (approved):
Looks great, thanks for making that change.

PR #2040, opened 2020-09-14
Dont allow disabling waku

Ferossgp:
As the whisper will not be used anymore, there is no need to keep `waku-enabled` in the user settings. https://github.com/status-im/status-react/issues/11092


appdatabase/migrations/sql/0013_dont_disable_waku.down.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE settings DROP COLUMN waku_enabled;
```
cammellos: There's no drop column in sqlite, I know, it's weird :) You can leave this migration empty

services/accounts/settings.go
```diff
@@ -16,6 +16,11 @@ type SettingsAPI struct {
 }
 
 func (api *SettingsAPI) SaveSetting(ctx context.Context, typ string, val interface{}) error {
+	// NOTE(Ferossgp): Backward compatibility, skip this for older clients instead of returning error
```
cammellos: Maybe worth adding the VERSION in the comment, so we know when it's safe to remove. Very neat!

cammellos (approved):
Just a few things, but awesome work!

PR #2043, opened 2020-09-15
[#2042] Wrong ERC20 decimal rounding in "Set max"

### LINKED ISSUE #2042: Wrong ERC20 decimal rounding in "Set max"
# Problem

https://github.com/status-im/status-react/issues/10060

flexsurfer:
fixes: https://github.com/status-im/status-go/issues/2042


services/wallet/balance_test.go
```diff
@@ -101,27 +101,22 @@ func (s *BalancesSuite) TestBalanceEqualPerAccount() {
...
 
 func (s *BalancesSuite) TestNoBalances() {
-	result, err := GetTokensBalances(context.Background(), s.client, s.accounts, s.tokens)
+	_, err := GetTokensBalances(context.Background(), s.client, s.accounts, s.tokens)
 	s.Require().NoError(err)
-	for _, account := range s.accounts {
-		for _, token := range s.tokens {
-			s.Require().Equal(zero.Int64(), result[account][token].Int64())
-		}
-	}
```
Samyoul: This test seems to be testing if an account has no balance, why would we remove the part of the test that checks the balance?

services/wallet/balance_test.go
```diff
@@ -61,13 +61,13 @@ func (s *BalancesSuite) SetupTest() {
 	}
 }
 
-func (s *BalancesSuite) TestBalanceEqualPerToken() {
+/*func (s *BalancesSuite) TestBalanceEqualPerToken() {
```
Samyoul, line 64: Why are we commenting out this code?

If this test is failing, we should find the reason and fix it.

Samyoul (dismissed):
Hey @flexsurfer, thank you for your PR. I don't have any problem with replacing `big.Int` with `hexutil.Big`, I've just got some questions about the changes you've made to the tests.

services/wallet/balance_test.go
flexsurfer: hey @Samyoul so tests are failing with 
```
=== RUN   TestBalancesSuite/TestNoBalances
panic: reflect: call of reflect.Value.Type on zero Value

goroutine 365 [running]:
reflect.Value.Type(0x0, 0x0, 0x0, 0x4ce8aa0, 0x8)
```
thats why i asked you yo take a look

flexsurfer: this PR isnt ready i need your help :)

Samyoul: Ah, I'm very sorry. Thank you for clarifying that. Ok, let me see what is causing the error.

Does the error give any more stack trace? Any line numbers?

flexsurfer:
```
GOROOT=/usr/local/Cellar/go/1.14.6/libexec #gosetup
GOPATH=/Users/andrey/go #gosetup
/usr/local/Cellar/go/1.14.6/libexec/bin/go test -c -o /private/var/folders/xn/r9ts4cb53t938r29hy320l8w0000gn/T/___go_test_github_com_status_im_status_go_services_wallet github.com/status-im/status-go/services/wallet #gosetup
# gopkg.in/olebedev/go-duktape.v3
In file included from _cgo_export.c:4:
debugger.go:22:13: warning: unused function '_duk_debugger_attach' [-Wunused-function]
/usr/local/Cellar/go/1.14.6/libexec/bin/go tool test2json -t /private/var/folders/xn/r9ts4cb53t938r29hy320l8w0000gn/T/___go_test_github_com_status_im_status_go_services_wallet -test.v
=== RUN   TestBalancesSuite
=== RUN   TestBalancesSuite/TestInterrupted
=== RUN   TestBalancesSuite/TestNoAccounts

Process finished with exit code 1
=== RUN   TestBalancesSuite/TestNoBalances
panic: reflect: call of reflect.Value.Type on zero Value

goroutine 416 [running]:
reflect.Value.Type(0x0, 0x0, 0x0, 0x4ce8aa0, 0x8)
	/usr/local/Cellar/go/1.14.6/libexec/src/reflect/value.go:1879 +0x183
github.com/ethereum/go-ethereum/accounts/abi.set(0x0, 0x0, 0x0, 0x4e869e0, 0xc0ab5e1080, 0x16, 0xc0ab3d9640, 0x400f558)
	/Users/andrey/go/pkg/mod/github.com/status-im/go-ethereum@v1.9.5-status.9/accounts/abi/reflect.go:83 +0x5a
github.com/ethereum/go-ethereum/accounts/abi.set(0x4dd69a0, 0xc0ab5720c8, 0x196, 0x4e869e0, 0xc0ab5e1080, 0x16, 0xc0006935a0, 0x4d9cc40)
	/Users/andrey/go/pkg/mod/github.com/status-im/go-ethereum@v1.9.5-status.9/accounts/abi/reflect.go:88 +0x413
github.com/ethereum/go-ethereum/accounts/abi.unpack(0xc0ab3d9870, 0x4ca09a0, 0xc0ab5720c8, 0x4e869e0, 0xc0ab5e1080, 0xc0ab5720c8, 0x4e869e0)
	/Users/andrey/go/pkg/mod/github.com/status-im/go-ethereum@v1.9.5-status.9/accounts/abi/argument.go:132 +0xccd
github.com/ethereum/go-ethereum/accounts/abi.Arguments.unpackAtomic(0xc009de9b00, 0x1, 0x4, 0x4ca09a0, 0xc0ab5720c8, 0x4e869e0, 0xc0ab5e1080, 0x1, 0x1)
	/Users/andrey/go/pkg/mod/github.com/status-im/go-ethereum@v1.9.5-status.9/accounts/abi/argument.go:216 +0x1fc
github.com/ethereum/go-ethereum/accounts/abi.Arguments.Unpack(0xc009de9b00, 0x1, 0x4, 0x4ca09a0, 0xc0ab5720c8, 0xc0ab5fc3e0, 0x20, 0x20, 0x44c0d98, 0xc039038050)
	/Users/andrey/go/pkg/mod/github.com/status-im/go-ethereum@v1.9.5-status.9/accounts/abi/argument.go:102 +0x17d
github.com/ethereum/go-ethereum/accounts/abi.ABI.Unpack(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/andrey/go/pkg/mod/github.com/status-im/go-ethereum@v1.9.5-status.9/accounts/abi/abi.go:87 +0x2a4
github.com/ethereum/go-ethereum/accounts/abi/bind.(*BoundContract).Call(0xc009dcb6b0, 0xc0ab3d9f20, 0x4ca09a0, 0xc0ab5720c8, 0x4e9943a, 0x9, 0xc000be0e30, 0x1, 0x1, 0x5224920, ...)
	/Users/andrey/go/pkg/mod/github.com/status-im/go-ethereum@v1.9.5-status.9/accounts/abi/bind/base.go:164 +0x35a
github.com/status-im/status-go/services/wallet/ierc20.(*IERC20Caller).BalanceOf(0xc000342158, 0xc000be0f20, 0x8a8f54cd27b6c21a, 0xb3bb2294c736cd41, 0xc0efa778e8, 0xc0006933b0, 0x400ece6, 0x0)
	/Users/andrey/go/src/github.com/status-im/status-go/services/wallet/ierc20/ierc20.go:226 +0x104
github.com/status-im/status-go/services/wallet.GetTokensBalances.func1(0x5224920, 0xc00049e840, 0x400814b, 0xc0ab40e540)
	/Users/andrey/go/src/github.com/status-im/status-go/services/wallet/balance.go:40 +0x14f
github.com/status-im/status-go/services/wallet.(*AtomicGroup).Add.func1(0xc00049e880, 0xc009d9d680)
	/Users/andrey/go/src/github.com/status-im/status-go/services/wallet/async.go:114 +0x6e
created by github.com/status-im/status-go/services/wallet.(*AtomicGroup).Add
	/Users/andrey/go/src/github.com/status-im/status-go/services/wallet/async.go:112 +0x66

```

flexsurfer: @Samyoul

Samyoul:
@flexsurfer See the latest commit. https://github.com/status-im/status-go/pull/2043/commits/dc099b3c8dbb5edca0033a7dd81998f24c2578aa

services/wallet/ierc20/ierc20.go
```diff
@@ -220,11 +220,11 @@ func (_IERC20 *IERC20CallerSession) Allowance(owner common.Address, spender comm
 // Solidity: function balanceOf(address account) constant returns(uint256)
 func (_IERC20 *IERC20Caller) BalanceOf(opts *bind.CallOpts, account common.Address) (*hexutil.Big, error) {
 	var (
-		ret0 = new(*hexutil.Big)
+		ret0 = new(*big.Int)
 	)
 	out := ret0
 	err := _IERC20.contract.Call(opts, out, "balanceOf", account)
-	return *ret0, err
+	return (*hexutil.Big)(*ret0), err
```
flexsurfer, line 228: in that case problem won't be solved, because we need to write call response to `hexutil.Big`

flexsurfer: @Samyoul

Samyoul: Hey @flexsurfer , are you sure? When I was experimenting with this code, the assignment code makes a `*big.Int`.

Does my suggested change not give the desired results?

I basically pull out the raw `big.Int` data from the `Call` function and then I type assert the new `big.Int` value to a `hexutil.Big`.

---

If you use the following code

```go
func (_IERC20 *IERC20Caller) BalanceOf(opts *bind.CallOpts, account common.Address) (*hexutil.Big, error) {
	var (
		ret0 = new(hexutil.Big)
	)
	out := &ret0
	err := _IERC20.contract.Call(opts, out, "balanceOf", account)
	return ret0, err
}
```

You get the following error:

`abi: cannot umarshal *big.Int in to hexutil.Big`

What this tells me is that the reflection code in the `Set` functionality parses the incoming contract data as `*big.Int`, so if we need to use a `hexutil.Big` higher up the process type asserting the data would be the best solution.

The only exception, would be if the `Call` function uses the `encoding.TextUnmarshaler` interface, but I can't see the use of `UnmarshalText()` in the accounts package.

flexsurfer: hey @Samyoul so initial issue in status-react that we receive rounded value , my guess we can't use big.Int because it rounds the value and should use hexutil.Big, but im not sure, so if we will use big.Int in Call, result still will be rounded ? or do you know why we get rounded value in status-react ?

Samyoul: Yeah, I read through the bug report. I don't think that the problem is with `big.Int` as `big.Int` shouldn't be rounding at all, it is designed for very big numbers.

`hexutil.Big` is a wrapper for `big.Int` with some additional un/marshalling functionality, so unless the `Call` function uses one of the additional interface implementations there is nothing that `hexutil.Big` would do that would fix the problem.

My suggestion is that you do some experimenting and print out the value of a contract at various points in the `status-go` process flow to see if the value is always rounded, or if the value gets rounded in a function inside `status-go`. Does the `status-go` API endpoint return a rounded value?

cammellos: Currently as far as  I can see, `Big.Int` is used as a number in Javascript :
https://github.com/status-im/status-react/blob/0448edb509b705bb48bfefacef38bbaf2cc114de/src/status_im/wallet/core.cljs#L146

That looks like a problem to me, as we might be hitting `Number.MAX_SAFE_INTEGER`  (9007199254740991 on my browser).

We should have the JSON from the RPC return a string and keep that as a string in javascript (or any non-lossy conversion, byte buffers etc), as any conversion lossy might be hitting the limit if we use numbers.

On ours side we probably want to marshalJSON as strings, safer.

cammellos: In our case it will be now a string of hex bytes https://github.com/ethereum/go-ethereum/blob/9a39c6bcb1fe11a252ac832d995a72e5f2210fbe/common/hexutil/json.go#L39 I believe, which is fine if that's what we want.
If we use math/big int it will be converted to a number, loosing precision when parsed in javascript
https://play.golang.org/p/K6CxuHzhI1S , maybe that's the issue

flexsurfer: @Samyoul you are right, it works, i fixed tests , could you help with lint please

cammellos (approved):

cammellos:
@Samyoul I believe tests have been fixed, feel free to review again, thanks!

PR #2046, opened 2020-09-24
Communities 

cammellos:
This commit add the basics for community management.

- Create a community
- Import/Export a community
- Send an invitation message
- Add/Remove users to a community
- Create channels within a community

I have left the commit split in order to facilitate the review process.


cammellos:
@Samyoul @jakubgs This should be ready for review. 
I will merge after Profile pics are merged, but it can already be reviewed, mind I have pulled some of the protobufs from there and will be replaced once profile pics is merged.
Thanks!

VERSION
```diff
@@ -1 +1 @@
-0.66.0
+0.66.1
```
Samyoul: I feel like Communities should be a semver `minor` level, rather than a `patch` level change

protocol/chat.go
```diff
@@ -301,22 +333,11 @@ func CreateGroupChat(timesource TimeSource) Chat {
...
-			b, err = types.DecodeHex(item)
-		} else {
-			b, err = hex.DecodeString(item)
-		}
-		if err != nil {
-			return nil, err
-		}
-		result[idx], err = crypto.UnmarshalPubkey(b)
+		var err error
+		result[idx], err = common.HexToPubkey(item)
```
Samyoul, line 340: I really like the neatness of this function

protocol/messenger_response.go
```diff
@@ -2,20 +2,28 @@ package protocol
...
+	Communities      []*communities.Community        `json:"communities,omitempty"`
+	CommunityChanges []*communities.CommunityChanges `json:"communitiesChanges,omitempty"`
+	Filters          []*transport.Filter             `json:"filters,omitempty"`
+	RemovedFilters   []string                        `json:"removedFilters,omitempty"`
 }
 
 func (m *MessengerResponse) IsEmpty() bool {
-	return len(m.Chats) == 0 && len(m.Messages) == 0 && len(m.Contacts) == 0 && len(m.Installations) == 0 && len(m.Invitations) == 0
+
+	return len(m.Chats)+len(m.Messages)+len(m.Contacts)+len(m.Installations)+len(m.Invitations)+len(m.EmojiReactions)+len(m.Communities)+len(m.CommunityChanges)+len(m.Filters)+len(m.RemovedFilters)+len(m.RemovedChats) == 0
```
Samyoul, line 26: Ah, I like this,

services/ext/api.go
```diff
@@ -316,6 +317,53 @@ func (api *PublicAPI) SetInstallationMetadata(installationID string, data *multi
...
+}
+
+func (api *PublicAPI) CreateCommunityChat(orgID string, c *protobuf.CommunityChat) (*protocol.MessengerResponse, error) {
+	return api.service.messenger.CreateCommunityChat(orgID, c)
+}
+
+func (api *PublicAPI) InviteUserToCommunity(orgID, userPublicKey string) (*protocol.MessengerResponse, error) {
+	return api.service.messenger.InviteUserToCommunity(orgID, userPublicKey)
+}
+
```
Samyoul: I normally don't mind Exported functions not having a description comment, but on API endpoints, descriptions serve as a self documenting API. Do you think it would be beneficial to add description comments to these functions?

Samyoul (commented):
So I've reviewed all the peripheral code and have no comments of substance. Looks good. I've reviewed 32 / 45 files, the remaining is the actually core of `protocol/communities`.

Question. What is a "grant"? It pops up as a new field in the protobufs, but I don't understand its role (which may be more obvious after I read the `protocol/communities code`)

cammellos:
@Samyoul addressed feedback thanks!

Grant is described here: https://github.com/status-im/specs/pull/151/files

Basically because we can't rely on receiving messages in order, we have a `grant` that is given by the community creator to the invited user, which will be carried on when posting.
In this way other clients can optimistically accept the message if a grant is valid, in case the updated community description has not been fetched yet.
The feature is not fully implemented yet, but there's room for it in the codebase.

Samyoul:
Ah thank you. So a grant is like an attestation or permission in the form of a signed message from a known admin of an organisation, that's a nice solution.

protocol/protobuf/chat_identity.proto
```diff
@@ -14,6 +14,14 @@ message ChatIdentity {
 
   // images is a string indexed mapping of images associated with an identity
   map<string, IdentityImage> images = 3;
+
+  // display name is the user set identity, valid only for organisations
+  string display_name = 4;
+
+  // description is the user set description, valid only for organisations
+  string description = 5;
+
```
Samyoul, line 23: I like these as although they are noted as being organisations only, they can be used in a user context.

protocol/communities/community.go
```diff
@@ -0,0 +1,706 @@
...
+}
+
+func (o *Community) AcceptRequestToJoin(pk *ecdsa.PublicKey) (*protobuf.CommunityRequestJoinResponse, error) {
+
+	return nil, nil
+}
+
+func (o *Community) DeclineRequestToJoin(pk *ecdsa.PublicKey) (*protobuf.CommunityRequestJoinResponse, error) {
+	return nil, nil
+}
```
Samyoul: Do these functions need a `TODO` adding?

protocol/communities/community.go
```diff
@@ -0,0 +1,706 @@
...
+
+	// Store request to join
+	return nil
+}
+
+func (o *Community) IsAdmin() bool {
+	return o.config.PrivateKey != nil
+}
+
+func (o *Community) handleRequestJoinWithChatID(signer *ecdsa.PublicKey, request *protobuf.CommunityRequestJoin) error {
```
Samyoul: `signer` isn't used in this function

protocol/communities/community.go
```diff
@@ -0,0 +1,706 @@
...
+
+	var chat *protobuf.CommunityChat
+	chat, ok := o.config.CommunityDescription.Chats[request.ChatId]
```
Samyoul: `chat` is explicitly declared and then immediately implicitly redeclared. I know that this is the only way to declare `ok` but as there is no code that relies on `chat` being conditionally set it is not necessary to explicitly declare it.

protocol/communities/community.go
Samyoul: Ah, is this a work in progress?

protocol/communities/community.go
```diff
@@ -0,0 +1,706 @@
...
+	}
+
+	if chat.Permissions.EnsOnly && len(request.EnsName) == 0 {
+		return ErrCantRequestAccess
+	}
+
+	return nil
+}
+
+func (o *Community) handleRequestJoinWithoutChatID(signer *ecdsa.PublicKey, request *protobuf.CommunityRequestJoin) error {
```
Samyoul: `signer` not used in this function. But the function also looks like a work in progress, do we need to add a `TODO`?

protocol/communities/manager.go
```diff
@@ -0,0 +1,361 @@
...
+
+	err = m.persistence.SaveCommunity(community)
+	if err != nil {
+		return nil, err
+	}
+
+	return community, nil
+}
+
+func (m *Manager) HandleCommunityInvitation(signer *ecdsa.PublicKey, invitation *protobuf.CommunityInvitation, payload []byte) (*Community, error) {
```
Samyoul, line 226: `signer` and `invitation` are not used here. Is this a work in progress?

protocol/communities/manager_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+	actualCommunity := communities[0]
+	if bytes.Equal(community.ID(), communities[1].ID()) {
+		actualCommunity = communities[1]
+	}
+
+	s.Require().Equal(community.ID(), actualCommunity.ID())
+	s.Require().Equal(community.PrivateKey(), actualCommunity.PrivateKey())
+	s.Require().True(proto.Equal(community.config.CommunityDescription, actualCommunity.config.CommunityDescription))
+}
```
Samyoul, line 61: There seems to be a lot of functionality that `Manager` has that isn't tested here. Things like adding importing and exporting a community and handling description messages.

protocol/communities/persistence.go
```diff
@@ -0,0 +1,131 @@
...
+
+func (p *Persistence) queryCommunities(query string) ([]*Community, error) {
+	var response []*Community
+
+	rows, err := p.db.Query(query)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
```
Samyoul: A tiny little thing, `rows.Close()` returns an error, so if anything goes wrong with closing the `rows` the error will never be handled.

If we care about this, something like below would work to catch the error.

```suggestion
func (p *Persistence) queryCommunities(query string) (response []*Community, err error) {
	rows, err := p.db.Query(query)
	if err != nil {
		return nil, err
	}
	defer func() {
		err = rows.Close()
	}()

```

protocol/communities_messenger_test.go
```diff
@@ -0,0 +1,514 @@
...
+	options := []Option{
+		WithCustomLogger(s.logger),
+		WithMessagesPersistenceEnabled(),
+		WithDatabaseConfig(tmpFile.Name(), ""),
+		WithDatasync(),
+	}
+	return s.newMessengerWithOptions(shh, privateKey, options)
+}
+
+func (s *MessengerCommunitiesSuite) newMessenger(shh types.Waku) *Messenger {
```
Samyoul: `shh` is not used in this function, I can see that this suite manages shh so probably a hangover from other test suites that don't manage the protocol type.

protocol/message_handler.go
```diff
@@ -324,6 +330,81 @@ func (m *MessageHandler) HandlePairInstallation(state *ReceivedMessageState, mes
...
+	if err != nil {
+		return err
+	}
+	state.AllCommunities[community.IDString()] = community
+
+	return nil
+}
+
+// HandleWrappedCommunityDescriptionMessage handles a wrapped community description
+func (m *MessageHandler) HandleWrappedCommunityDescriptionMessage(state *ReceivedMessageState, payload []byte) (*communities.Community, error) {
```
Samyoul: `state` is not used in this function

Samyoul (approved):
Great work. Congratulations. My feedback is nothing major, mostly related to either unused parameters or how we handle certain erros.

PR #2048, opened 2020-09-25
Migrate journals files as well

cammellos:
In some instances where the db wasn't correctly closed the journal files had to be moved has well.


rasom (approved):

Samyoul (approved):
Ah, I should have reviewed this one first. The rune count changes make more sense.

PR #2051, opened 2020-09-29
Profile Images

Samyoul:
## What has changed
**Added** new `waku_ext` api methods

* `GetIdentityImages(keyUID string)`
  * Gets all identity images for the given keyUID
* `GetIdentityImage(keyUID, name string)`
  * Gets the identity image for the given keyUID and name
* `StoreIdentityImage(keyUID, filepath string, ax, ay, bx, by int)`
  * parses, generates and stores identity images for given keyUID
* `DeleteIdentityImage(keyUID)`
  * Deletes all identity images for a given keyUID

**Added** new protobufs:

```protobuf
// ChatIdentity represents the user defined identity associated with their public chat key
message ChatIdentity {
// Lamport timestamp of the message
uint64 clock = 1;

// ens_name is the valid ENS name associated with the chat key
string ens_name = 2;

// images is a string indexed mapping of images associated with an identity
map<string, IdentityImage> images = 3;
}

// ProfileImage represents data associated with a user's profile image
message IdentityImage {

// payload is a context based payload for the profile image data,
// context is determined by the `source_type`
bytes payload = 1;

// source_type signals the image payload source
SourceType source_type = 2;

// image_type signals the image type and method of parsing the payload
ImageType image_type =3;

// SourceType are the predefined types of image source allowed
enum SourceType {
UNKNOWN_SOURCE_TYPE = 0;

// RAW_PAYLOAD image byte data
RAW_PAYLOAD = 1;

// ENS_AVATAR uses the ENS record's resolver get-text-data.avatar data
// The `payload` field will be ignored if ENS_AVATAR is selected
// The application will read and parse the ENS avatar data as image payload data, URLs will be ignored
// The parent `ChatMessageIdentity` must have a valid `ens_name` set
ENS_AVATAR = 2;
}
}
```

**Removed** multi-account photo path field

**Added**`status-go` side image decode/encode and manipulation

## Why make the change
Allows users to have profile images that are light weight and do not require large bandwidth capabilities

## Road map

* [x] **Image handling**
  * [x] Image decode functionality
  * [x] Image manipulation
    * [x] Crop
    * [x] Resize
  * [x] Image encode
    * [x] Render to best file size functionality
  * [x] Refactor and merge `protocol` image functionality into `images` package
* [x] **Image persistence**
  * [x] User's Identity image db migration
  * [x] CRUD for User's identity image in encrypted db
  * [x] `ChatIdentity` last published to public chat register
  * [x] ~~Contact table migration to change column name to `image_uri`~~
* [x] **Protocol**
  * [x] Create `ChatIdentity` protobuf
  * [x] Refactor `ImageType` to Enum
    * [x] Refactor all dependencies
  * [x] Publish `ChatIdentity` to public channel
    * Use large image (240 px)
  * [x] Publish `ChatIdentity` to public chat
    * Use small image (80 px)
  * [x] Read `ChatIdentity` data from public channel
  * [x] Read `ChatIdentity` data from public chat
  * [x] ~~Include user `ChatIdentity` info in sync data~~
    * Include in account data sync
  * [x] ~~Include contact `ChatIdentity` info in sync data~~
    * include in contact data sync
* [x] **API**
  * [x] Profile Image CRUD
  * [x] Trigger for publishing `ChatIdentity`
  * [x] Trigger for updating user's `ChatIdentity` data sync

## TODOs

* Remove photoPath from settings
* Implement a fix for edge case race condition for dropped sync messages in `ChatIdentity` handler
* Handle ENS in `ChatIdentity`
* Handle not attempting to attach a ChatIdentityImage if the account has no images


images/identity.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func (i IdentityImage) MarshalJSON() ([]byte, error) {
+	uri, err := i.GetDataURI()
+	if err != nil {
+		return nil, err
+	}
+
+	temp := struct {
+		KeyUID       string `json:"key_uid"`
```
cammellos: we generally use `camelCase` for json, though we are not super consistent, as sometimes we use `kebab-case` (clojurism). And sometimes `snake_case` but that's mainly protobuf default serialization.
Ideally eventually we go for `camelCase` everywhere

multiaccounts/accounts/database.go
```diff
@@ -108,6 +108,7 @@ func (db Database) Close() error {
 	return db.db.Close()
 }
 
+// TODO remove photoPath from settings
```
cammellos, line 125: This is still a TODO or has been resolved?

protocol/message_handler.go
```diff
@@ -799,3 +801,67 @@ func (m *MessageHandler) HandleGroupChatInvitation(state *ReceivedMessageState,
...
+	logger := m.logger.With(zap.String("site", "HandleChatIdentity"))
+	contact := state.CurrentMessageState.Contact
+	chat, ok := state.AllChats[contact.ID]
+	if !ok {
+		chat = OneToOneFromPublicKey(state.CurrentMessageState.PublicKey, state.Timesource)
+		// We don't want to show the chat to the user
+		chat.Active = false
+	}
+
+	logger.Info("Handling contact update")
```
cammellos, line 809: This probably should read differently

protocol/messenger.go
```diff
@@ -377,6 +389,131 @@ func (m *Messenger) handleSendContactCode() error {
...
+	}
+	shouldPublishChatIdentity, err := m.shouldPublishChatIdentity(chat.ID)
+	if err != nil {
+		return err
+	}
+	if !shouldPublishChatIdentity {
+		return nil
+	}
+
+	ci, err := m.createChatIdentity("public-chat")
```
cammellos: maybe we could use const/enums for `public-chat` and `private-chat`?

protocol/messenger.go
```diff
@@ -377,6 +389,131 @@ func (m *Messenger) handleSendContactCode() error {
...
+}
+
+// shouldPublishChatIdentity returns true if the last time the ChatIdentity was attached was more than 24 hours ago
+func (m *Messenger) shouldPublishChatIdentity(chatID string) (bool, error) {
+	lp, err := m.persistence.GetWhenChatIdentityLastPublished(chatID)
+	if err != nil {
+		return false, err
+	}
+
+	return *lp == 0 || time.Now().Unix()-*lp > 24*60*60, nil
```
cammellos: we could put the interval in a var here

cammellos (commented):
Looks good to me, only some minor comments

multiaccounts/accounts/database.go
Samyoul: Still a TODO. I'm compiling a list of the TODOs in the description of this PR, then I'll make an issue for each. I thought it would be better to get the main body of this functionality into `develop` then incrementally improve it

protocol/messenger.go
Samyoul: Yeah, that's good point. This actually went through my mind at the time of writing the code, then I added it to a mental list of todos that got forgot `shame_emoji`

protocol/message_handler.go
Samyoul: Good point, could be very confusing

multiaccounts/database.go
```diff
@@ -76,3 +132,88 @@ func (db *Database) DeleteAccount(keyUID string) error {
...
+		}
+
+		iis = append(iis, ii)
+	}
+
+	return iis, nil
+}
+
+func (db *Database) GetIdentityImage(keyUID, it string) (*images.IdentityImage, error) {
+	rows, err := db.db.Query("SELECT key_uid, name, image_payload, width, height, file_size, resize_target FROM identity_images WHERE key_uid = ? AND name = ?", keyUID, it)
```
cammellos: You can use `QueryRow` here, is probably more idiomatic, so you don't have to loop, you just need to make sure you return `nil` if the error is `ErrNoRows`

cammellos (approved):

PR #2052, opened 2020-10-01
Force migration if dirty

cammellos:
Some users have reported an issue with migrations:

https://github.com/status-im/status-react/issues/11249

Essentially if a migration is in place and the app is closed while the db is being migrated the migration will fail.

I have added some code that will retry the migration if that has failed. 
We should wrap migrations from now on inside transactions so that they can be cleanly retried.
I had to change a specific migration (the one failing) so that if retried it will succeed.


churik:
Tested on Android 10 with error from https://github.com/status-im/status-react/issues/11249 - after applying patch can login normally.
- biometric is enabled
- group, 1-1 is preserved - including audio, images, replies, stickers
- contacts are preserved

jakubgs (approved):

PR #2059, opened 2020-10-20
Link previews support added

vkjr:
Added database fields and statusgo functions needed for link preview support


cammellos (changes requested):
Thanks @vkjr everything looks good!
Two minor points:

We tend to use RPC instead of modifying native module unless necessary (so you would add a function in `services/ext/api.go` instead, and use `callPrivateRPC on the module), I am not sure what's the rationale, but it has been the convention for now, so maybe worth moving them there (as well/or only there). Probably the reason is that we can expose them through http/rpc apis if we want to do so.


I would probably rename `CanAskToPreviewLink` to `link_previews_enabled/disabled` as it seems clearer to me, what do you think?

Otherwise looks good

vkjr:
@cammellos, thanks will fix that!

jrainville (commented):
I tested this branch in status desktop (nim-status-client) and it works well, but it seems like not all youtube links work.

On mobile, sharing youtube links look like this: `https://youtu.be/ID`
Note, it's a shorter url using `.be`, but it leads to the `youtube.com` alternative.

Anyway, I thought I'd let you know to see if you could add it as well as a youtube alt url

protocol/common/message.go
```diff
@@ -325,11 +330,34 @@ func extractMentions(parsedText ast.Node) []string {
 	return visitor.mentions
 }
 
+type LinkNodeVisitor struct {
```
cammellos: You probably want to modify the existing class, above, to also save links, in the same way you did, so we avoid vising each node twice

vkjr: Good idea, thanks! 
It is still wip, I'll re-request review when ready :)

vkjr:
@cammellos, review notes addressed - RPC used, links collected with single visitor, added support for `youtu.be` domain, tests and lint complaints fixed

cammellos (approved):
Awesome work!
Looks all good, two things that we might want to take into consideration:

1) On upgrade, the new field `link_preview_request_enabled` will be set to `NULL` on existing clients. This causes problem with some field types as they are non-nullable (`bool` is non nullable), and it will throw an error on deserializing from the db (it's 100% the case with `string`, not sure with bool). To avoid this you can simply add 
`UPDATE settings SET link_preview_request_enabled = 1` to the migration, so that existing clients will have that field set for sure.
2) You might want to extra the domain for easier lookup from the links, so that `Links` is []Links{{Domain: domain, URL: url}}, but that can be done on the client side as @flexsurfer prefers (for perf reason is probably better to have it done in status-go)

Otherwise it looks good!

vkjr:
@cammellos, doesn't `BOOLEAN DEFAULT TRUE` makes `link_preview-request_enabled` non-null?

cammellos:
> @cammellos, doesn't `BOOLEAN DEFAULT TRUE` makes `link_preview-request_enabled` non-null?

Yes, though for existing rows the field won't change, it will not update older rows, only new ones, 100% this would crash on strings, bool I am not so sure though.

vkjr:
@cammellos, ah I see, thanks!

vkjr:
@cammellos, I've updated database migration file.
As for passing links to `status-react`, I'd suggest to leave them as is. Because in `status-react` we don't really extract and use domain name. We only check if link starts with the whitelisted domain. this check won't go away and won't improve performance. Also we can't make this check beforehand in `status-go` and store results in db, because whitelist can be updated in newer versions and stored db records will become incorrect. Wdyt?

cammellos:
>  We only check if link starts with the whitelisted domain.
That should be cheap enough, sure, it's fine to leave as it is!

cammellos:
@vkjr Generally we wait until the corresponding PR on status-react (or desktop) has been tested,  to make sure that no changes are necessary, and also ensures that `status-go` is always releasable (we had issues in the past where status-go was not releasable as there were either some issues on the status-go code or status-react code integrating it), but it's probably safe this time, though next time it's probably best to wait.

You want to create a release and update your branch to point to the newly tagged,
basically just:
`git tag v0.62.15` on develop
`git push origin v0.62.15`
and later you can use the tag in `scripts/update-status-go.sh` 
Well done on the PR amazing work!

vkjr:
@cammellos, got it, thanks!

PR #2060, opened 2020-10-20
profile status updates

flexsurfer:
profile status updates


cammellos (commented):
Overall looks good, I think you might have to handle a few more cases when introducing a new chat type:

https://github.com/status-im/status-go/blob/4ecca1169d2e0a07b4e0cf5a7a3f4044216c678f/protocol/messenger.go#L505 

Here you can add the id to `publicChatIDs` so that it creates the right filter.

https://github.com/status-im/status-go/blob/4ecca1169d2e0a07b4e0cf5a7a3f4044216c678f/protocol/messenger.go#L699

https://github.com/status-im/status-go/blob/4ecca1169d2e0a07b4e0cf5a7a3f4044216c678f/protocol/messenger.go#L676

https://github.com/status-im/status-go/blob/4ecca1169d2e0a07b4e0cf5a7a3f4044216c678f/protocol/messenger.go#L1660

https://github.com/status-im/status-go/blob/4ecca1169d2e0a07b4e0cf5a7a3f4044216c678f/protocol/messenger.go#L3799

Other then that it all looks ok, nice work!

flexsurfer:
thanks @cammellos , fixed !

cammellos:
Oh, sorry forgot to mention, it would be good to have  some tests as well, I can help you with that if you need, just essentially make sure you can send and receive a message

flexsurfer:
@cammellos yeah i added one https://github.com/status-im/status-go/pull/2060/files#diff-4af24fb726ed01702762a369d7d6e509b09eb942d0cba81cad895e609c2041beR376 do we need more ?

cammellos:
@flexsurfer missed it, sorry, that's probably enough, it would have been also good to test the receiving of the message, but probably that's enough as it's very similar to public chat. thanks

cammellos (approved):

PR #2064, opened 2020-10-30
feat: add bookmarks to the settings

jrainville:
Add bookmark support to be used in the browser of both mobile and desktop.

This is really simple, it adds the `bookmarks` field in Settings that is a JSON blob that is used as an array like so:

```js
bookmarks: [
{
name: "My Bookmark",
url: "http://...."
}
]
```

Let me know if I need to add the "prototype" of the bookmark objects somewhere so that we have some kind of doc.


jrainville:
Closing for now as I'll redo the PR using a different table since the settings are more limited

PR #2069, opened 2020-11-03
Split datasync messages in batches

cammellos:
When sending messages in quick succession, it might be that multiple
messages are batched together in datasync, resulting in a single large
payload.
This commit changes the behavior so that we can pass a max-message-size
and we split the message in batches before sending.

A more elegant way would be to split at the transport layer (i.e
waku/whisper), but that would be incompatible with older client.

We can still do that eventually to support larger messages.

Tested PR in status-react https://github.com/status-im/status-react/pull/11361


protocol/datasync/transport.go
```diff
@@ -9,15 +9,24 @@ import (
...
 	"github.com/vacp2p/mvds/transport"
+	"go.uber.org/zap"
 
 	datasyncpeer "github.com/status-im/status-go/protocol/datasync/peer"
 )
 
 var errNotInitialized = errors.New("Datasync transport not initialized")
 
+// payloadTagSize is the tag size for the protobuf.Payload message which is number of fields * 2 bytes
+var payloadTagSize = 14
```
jakubgs, line 20: Hopefully we won't forget to increment this when the number of fields changes.

protocol/datasync/transport.go
```diff
@@ -39,21 +50,100 @@ func (t *NodeTransport) Watch() transport.Packet {
...
+		if len(offer)+currentSize+1 > maxSizeBytes {
+			if newPayload.IsValid() {
+				response = append(response, newPayload)
+			}
+			newPayload = &protobuf.Payload{Offers: [][]byte{offer}}
+			currentSize = len(offer) + payloadTagSize + 1
+		} else {
+			newPayload.Offers = append(newPayload.Offers, offer)
+			currentSize += len(offer)
+		}
```
jakubgs, line 112: Can't help to notice that this whole `for` block is almost identical to the one for `payload.Requests` and `playload.Acks`. Couldn't we make a helper function and call if in all three `for` loops? The third one is a bit different so it might not be worth generalizing it, but at least these three loops could be. Or do you think it wouldn't make much of a difference?

If it has a good name it could make it more clear about what's happening in those loops.

jakubgs (approved):

protocol/datasync/transport.go
cammellos: I have the feeling that it will be more complex, as we'd have to generalize offer different types/methods, which is something that golang probably does not do very well, so I'd be inclined to keep it as it is, dumb but simple, with some duplication

jakubgs: That's fair, if a bit ugly.

PR #2070, opened 2020-11-03
Log disconnect/connect event

cammellos:
Marketing was relying on mailserver entries for checking the time a
give peer spent on the app.

This was not accurate as the assumption was that a peer would "ping" a
mailserver every 15s, which is not the case, it only hit the mailserver
as it comes from an "offline" state.

This commit changes the log level of two entries so that we have
connect/disconnect time for a given peer, which should be enough to
calculate roughly the time a peer has been online if connected to our
fleet.


waku/waku.go
```diff
@@ -1083,7 +1083,7 @@ func (w *Waku) HandlePeer(peer common.Peer, rw p2p.MsgReadWriter) error {
 	w.peers[peer] = struct{}{}
 	w.peerMu.Unlock()
 
-	w.logger.Debug("handling peer", zap.String("id", types.EncodeHex(peer.ID())))
+	w.logger.Info("handling peer", zap.String("id", types.EncodeHex(peer.ID())))
```
jakubgs: Wrong field name in log. It's `id` and should be `peerID` as per Logstash parsing rules:
https://github.com/status-im/infra-hq/blob/0a91d7eef311d3e25f25ff3098a7d6964db60013/ansible/roles/logstash/templates/pipe/50-filter-statusd.conf#L14-L20

cammellos: done!

jakubgs (approved):

Issue #1978, opened 2020-05-15
Consistently flaky Jenkins unit tests


Samyoul:
# Problem
Jenkins unit tests will consistently fail ~~even if all the tests pass~~ (there was a failing test, I just didn't notice them in the full test log). The tests can be passed mostly by retesting and making no changes to the code.

* [Recent tests](https://ci.status.im/blue/organizations/jenkins/status-go%2Fprs%2Ftests/activity)
* [Specific test that failed at the very end of unit test](https://ci.status.im/blue/organizations/jenkins/status-go%2Fprs%2Ftests/detail/PR-1969/12/pipeline)
* [Full log](https://ci.status.im/blue/rest/organizations/jenkins/pipelines/status-go/pipelines/prs/pipelines/tests/branches/PR-1969/runs/12/nodes/73/steps/76/log/?start=0)
<details>
<summary>Sample log data</summary>

```text
[2020-05-14T14:46:52.615Z] + make test-unit
[...]

PR #2071, opened 2020-11-03
Add bookmarks table for the browser

jrainville:
Add bookmark support to be used in the browser of both mobile and desktop.

It add a new Table called `bookmarks` which contains a `name`, `url` and `imageUrl` fields, all strings.

The `imageUrl`  field is populated by status-go itself using the [besticon library](https://github.com/mat/besticon)


services/browsers/database.go
```diff
@@ -122,3 +124,72 @@ func (db *Database) DeleteBrowser(id string) error {
...
+func (db *Database) StoreBookmark(bookmark Bookmark) (Bookmark, error) {
+	insert, err := db.db.Prepare("INSERT OR REPLACE INTO bookmarks (url, name, image_url) VALUES (?, ?, ?)")
+
+	if err != nil {
+		return bookmark, err
+	}
+
+	// Get the right icon
+	finder := besticon.IconFinder{}
+	icons, iconError := finder.FetchIcons(bookmark.Url)
```
cammellos: You might want to log the error here, just in case we have some issues

appdatabase/migrations/sql/0017_bookmarks.up.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE IF NOT EXISTS bookmarks (
+  url VARCHAR NOT NULL,
+  name BIGINT NOT NULL,
```
cammellos: That's interesting, it clearly works as it's tested, is it intentional to have name as BIGINT and not a VARCHAR?

services/browsers/database.go
```diff
@@ -122,3 +124,72 @@ func (db *Database) DeleteBrowser(id string) error {
...
+
+	if iconError == nil {
```
cammellos: you can probably merge the two ifs
`if iconError == nil && len(icons) > 0`
but up to you, either are fine by me

cammellos (approved):
Thanks for the PR! looks all good to me, just a few questions/comments

cammellos:
Tests are not building for some reason:
```
[2020-11-03T17:10:17.903Z] FAIL	github.com/status-im/status-go/services/browsers [build failed]
```
You can run them by going into `services/browsers` and run `go test`

appdatabase/migrations/sql/0017_bookmarks.up.sql
jrainville: Oops, not at all. Copypaste error. I'll fix

jrainville:
@cammellos thanks for the review. I updated my code with your suggestions and fixed the test

cammellos:
@jrainville thank you, some linting issues I believe https://ci.status.im/blue/organizations/jenkins/status-go%2Fprs%2Ftests/detail/PR-2071/3/pipeline 
also if you haven't done so, you will want to increase the `VERSION` file, we use semver-ish a breaking change you would bump the `y` of `x.y.z` while non-breaking (such as this one), you'd bump the `z` only

jrainville:
The tests seem to fail on a waku test, which I haven't touched at all. Is it known to sometime fail for no reason?

cammellos:
We have some flaky tests, we are trying to address them but for the time being we compiled a list https://github.com/status-im/status-go/issues/1978 here.
You can restart the tests by going to `Details` and you should see a `Restart` for the step. I have restarted it myself now.

jrainville:
Ok, good to know. it passed now. I'll squash my commits now that everything seems ok

PR #2072, opened 2020-11-04
timeline

flexsurfer:
timeline


protocol/messenger.go
```diff
@@ -502,7 +502,7 @@ func (m *Messenger) Init() error {
 			continue
 		}
 		switch chat.ChatType {
-		case ChatTypePublic, ChatTypeProfile:
+		case ChatTypePublic, ChatTypeProfile, ChatTypeTimeline:
```
cammellos: This is used to create filters, do you need a filter for a timeline chat?

protocol/messenger.go
```diff
@@ -3800,7 +3815,7 @@ func (m *Messenger) encodeChatEntity(chat *Chat, message common.ChatEntity) ([]b
 		if err != nil {
 			return nil, err
 		}
-	case ChatTypePublic, ChatTypeProfile:
+	case ChatTypePublic, ChatTypeProfile, ChatTypeTimeline:
```
cammellos: Same here, do you ever save messages for a timeline chat?

protocol/messenger.go
flexsurfer: no thanks

protocol/messenger.go
flexsurfer: no thanks

flexsurfer:
thanks @cammellos fixed

cammellos (approved):

cammellos:
@flexsurfer looks good to me, there's some linting issues

PR #2073, opened 2020-11-04
Allow receiving old messages after rejoining a pubchat

rasom:
needed for https://github.com/status-im/status-react/issues/10362 

status: ready


protocol/message_handler.go
```diff
@@ -355,7 +355,7 @@ func (m *MessageHandler) HandleChatMessage(state *ReceivedMessageState) error {
 	}
 
 	// If deleted-at is greater, ignore message
-	if chat.DeletedAtClockValue >= receivedMessage.Clock {
+	if chat.DeletedAtClockValue >= receivedMessage.Clock && chat.ChatType != ChatTypePublic {
```
cammellos: There's already a method I believe:
`chat.Public()`

cammellos (approved):
Minor style change, we could also have a test to make sure that even if deletedAt is set you receive messages in a public chat, not that critical but would be nice

protocol/message_handler.go
rasom: fixed, thanks

rasom:
@cammellos added test. I thought about this but was too lazy.

PR #2074, opened 2020-11-06
fix timeline

flexsurfer:
fix timeline


protocol/messenger.go
```diff
@@ -501,6 +501,11 @@ func (m *Messenger) Init() error {
 		if !chat.Active {
 			continue
 		}
+
+		if chat.ChatType == ChatTypeTimeline {
```
cammellos: You can join in the condition above, 
```
if !chat.Active || chat.Timeline() {
   continue
}
```

cammellos (approved):

PR #2076, opened 2020-11-06
Increase verified count when ENS name does not verify

cammellos:
In some cases no error is returned but the name is not verified.
Before this ENS names would get verified again every 30s.
This commit changes the logic so that if they fail to verify without
error they will still exponentially backoff.


flexsurfer (approved):

VERSION
```diff
@@ -1 +1 @@
-0.63.4
```
flexsurfer: weird, but version in develop is 0.63.5

cammellos: oh, I might not have rebased, thanks!

PR #2079, opened 2020-11-13
timeline reactions

flexsurfer:
timeline reactions


cammellos (approved):
Added linting fix, looks good to me, could you add a test for the new functionality please?

PR #2080, opened 2020-11-17
Fix nil pointer local notifications

cammellos:
added check for nil value


api/geth_backend.go
```diff
@@ -912,7 +912,8 @@ func (b *GethStatusBackend) AppStateChange(state string) {
 			return
 		}
 
-		if !localNotifications.IsWatchingWallet() {
+		// If we have no local notifications, force wallet stop, otherwise check if it's watching the wallet
+		if localNotifications != nil || (localNotifications != nil && !localNotifications.IsWatchingWallet()) {
```
flexsurfer: first on should be localNotifications **==** nil  ?

cammellos: dang

flexsurfer (approved):

PR #2081, opened 2020-11-17
[wallet] Reduce number of RPC requests

rasom:
- Wallet service is not started on foreground event on status-go side
  anymore, it leaves a client side opportunity to decide whether new
  blocks should be watched.
- `watchNewBlocks` parameter is added to `StartWallet`.
- Some requests are removed/moved to the place where they are necessary.

status: ready


api/geth_backend.go
```diff
@@ -882,26 +882,7 @@ func (b *GethStatusBackend) AppStateChange(state string) {
 	b.log.Info("App State changed", "new-state", s)
 	b.appState = s
 
-	if s == appStateForeground && !b.forceStopWallet {
```
rasom, line 885: Wallet service is not started on foreground on `status-go` side anymore, it leaves a client side opportunity to decide whether new blocks should be watched.

api/geth_backend.go
```diff
@@ -947,7 +928,7 @@ func (b *GethStatusBackend) StopWallet() error {
 	return nil
 }
 
-func (b *GethStatusBackend) StartWallet() error {
+func (b *GethStatusBackend) StartWallet(watchNewBlocks bool) error {
```
rasom, line 931: `watchNewBlocks` parameter is added to `StartWallet`

services/wallet/commands.go
```diff
@@ -731,12 +732,6 @@ func (c *controlCommand) Run(parent context.Context) error {
 	if target.Cmp(zero) <= 0 {
 		target = zero
 	}
-	ctx, cancel = context.WithTimeout(parent, 3*time.Second)
```
rasom, line 734: This request is needed only if watching of new block is enabled, thus moved to `if c.watchNewBlocks { ... }` block

services/wallet/iterative.go
```diff
@@ -75,16 +74,8 @@ func (d *IterativeDownloader) Next(parent context.Context) ([]*DBHeader, *big.In
 		log.Error("failed to get transfer in between two bloks", "from", from, "to", to, "error", err)
 		return nil, nil, nil, err
 	}
-	// use integers instead of DBHeader
-	ctx, cancel := context.WithTimeout(parent, 5*time.Second)
```
rasom, line 79: Some unnecessary request, most likely a leftover

cammellos (approved):

PR #2083, opened 2020-11-20
custom nonce

flexsurfer:
Add option for user to cancel a pending transaction. #11215
Add option for user to increase the gas payment for a pending transaction #11214


cammellos (approved):
Version bump is missing, otherwise looks good

Issue #2068, opened 2020-11-03
Mailserver requests are still using bloom filters


cammellos:
# Problem

After investigating  https://github.com/status-im/status-react/issues/11351 
we noticed that mailservers requests are still using bloom filters, regardless of whether waku is using bloom filter mode.

Mailservers don't support querying by topic at the moment. https://github.com/status-im/status-go/blob/3f63b0c23cbe32c2331cb6201ee5cb5255ce9353/mailserver/mailserver.go#L1002 

and topics is never passed to mailserver:

https://github.com/status-im/status-go/blob/3f63b0c23cbe32c2331cb6201ee5cb5255ce9353/services/ext/api.go#L599

My understanding from previous conversations was that it was implemented, but it's not the case.

We should understand performance of querying by topics when a large amount of topics is passed in (500/1000 etc), to make sure the query is performant.

churik:
Reproduced during testing release 17/11/2020

Logs from Jarrad device: [Status-debug-logs.zip](https://github.com/status-im/status-go/files/5553613/Status-debug-logs.zip)

PR #2084, opened 2020-11-23
[Fixes: #2068] Pass topics for mailserver

cammellos:
We were not actually passing the topics in the request, therefore we
were using bloom filter for query, which resulted in long syncing times
for some users.


services/ext/api.go
```diff
@@ -610,6 +613,16 @@ func MakeMessagesRequestPayload(r MessagesRequest) ([]byte, error) {
 	return rlp.EncodeToBytes(payload)
 }
 
+func topicsToByteArray(topics []types.TopicType) [][]byte {
```
jakubgs, line 616: Why not define this as a method for `MessagesRequest` in it's own file? This way it could be reused in the future.
Or is that a premature optimization thing?

jakubgs (approved):
lgtm

services/ext/api.go
cammellos: Probably early, these files very rarely changes, but worth doing if next time we re-use it

PR #2085, opened 2020-11-24
Fix datasync retry

cammellos:
This commit fixes a bug on the mvds library where the nextEpoch would be
incorrectly summed to the retry time, resulting in messages not being
retried, or retried much less frequently the longer the app was running.
It also updates the retry timing to backoff exponentially at multiple of
30 seconds.

https://github.com/vacp2p/mvds/pull/86


vendor/github.com/vacp2p/mvds/node/node.go
```diff
@@ -545,7 +545,7 @@ func (n *Node) insertSyncState(groupID *state.GroupID, messageID state.MessageID
 
 func (n *Node) updateSendEpoch(s state.State) state.State {
 	s.SendCount += 1
-	s.SendEpoch += n.nextEpoch(s.SendCount, n.epoch)
+	s.SendEpoch = n.nextEpoch(s.SendCount, n.epoch)
```
Samyoul, line 548: Such a tiny little thing can cause all kinds of chaos. 💥

Samyoul (approved):
Looks good.

Just curious, why did you remove the debug logging?

cammellos:
@Samyoul I was debugging an issue and it's mostly just noise, I wanted to put them on `TRACE` but `DEBUG` is the lowest level.

PR #2086, opened 2020-12-01
Add SendChatMessages endpoints

cammellos:
This commit adds an endpoint to batch the sending of messages.
This is useful to simplify client logic when sending a batch of messages
and ensuring the correct order in the message stream.

It currently implements only what's needed, and naively return an error
if any of the messages fail.


Samyoul (approved):

Samyoul:
Nice work. I like the splitting out of the message response logic into a dedicated file, with dedicated tests 👍

jakubgs (approved):

PR #2087, opened 2020-12-02
[wallet] Dispatch event on tx history fetching failure

rasom:
needed for https://github.com/status-im/status-react/issues/11502 

Currently if `controlCommand` fails it will be restarted infinitely.
If `controlCommand` failed three times in a row for any reason we dispatch an event about failure. There might be different reasons but usually it is either bad connection or RPC endpoint which doesn't allow history requests.

status: ready


services/wallet/commands.go
```diff
@@ -830,6 +852,19 @@ func (c *controlCommand) Run(parent context.Context) error {
 	return err
 }
 
+func (c *controlCommand) NewError(err error) bool {
```
cammellos, line 855: Why 3 exactly? could you add a description of the behavior that we want to achieve please?

rasom: sure

rasom: 3 is arbitrary here

rasom: @cammellos added two sentences to PRs description, let me know if that's enough

cammellos (approved):

PR #2088, opened 2020-12-02
Resend emoji reactions when expired

vkjr:
This PR implements silent resending of emoji reactions, see [this issue](https://github.com/status-im/status-react/issues/11020)

To make this happen:
- `Messenger` got ability to listen notifications on envelope events. `NewMessenger()` now creates a wrapper-object `EnvelopeEventsInterceptor` and sets it to the envelopes monitor instead of original events handler.
- `RawMessage` struct got `Expired` field and `raw_messages` db table extended to save info about messages that didn't reach destination
- `Messenger`'s functions `processExpiredMessages()` and `processSentMessages()` update messages state in db
- `resendExpiredEmojiReactions()` gets raw messages from db and sends them again up to 3 times. This one called when `Messenger` goes online or when some messages expired.


cammellos (changes requested):
Overall looks really good!
One thing I would change, is that currently we resend all of them on `resendExpiredEmojiReactions`, and we don't wait for some backoff time.

So instead of re-sending them on `Start` or on `Expired`, I would start a go-routine in a loop, sleeping at interval (say 1s), and when it wakes up run `resendExpiredEmojiReactions`.
We also want to space the retries, say first retry at 30s, second at 60s, then 120s. 
It's a pattern we use already in the codebase, I can show you exactly where, we can have a chat tommorrow.
Other than that looks all really good!

vkjr:
@camellos, changes made to PR:
- `LastSent` field now sets in `RawMessage` on sending
- `watchExpiredEmojis()` resends emojis once in a second
- `shouldResendEmojiReaction()` implements condition that permits resending emoji only after more and more time passes. 

We were also discussing that information about message expiration can be ignored but I checked its implementation and found that it is not related to receiving acknowledgement, so it is quite reliable so Expired still used to fetch less records from DB. wdyt?

vkjr:
@cammellos, I got rid of expired field in favor of using `Sent` and `SendCount`

protocol/messenger.go
```diff
@@ -248,11 +291,86 @@ func NewMessenger(
...
+	if message.Sent {
+		return false, errors.New("Should resend only non-sent messages")
+	}
+
+	if message.SendCount > emojiResendMaxCount {
+		return false, nil
+	}
+	//exponential backoff depends on how many attempts to send message already made
+	backoff := time.Duration(math.Pow(2, float64(message.SendCount-1))) * emojiResendMinDelay * time.Second
+	backoffElapsed := time.Now().After(time.Unix(int64(message.LastSent), 0).Add(backoff))
```
cammellos: Here you probably want to pass a `TimeSource`, which is an `NTP` synced time, otherwise you might have inconsistencies if the clock of the user is skewed.
in messenger you can just pass as an argument:
`shouldSendEmojiReaction(message, m.getTimesource())`
and then call `t.getCurrentTime()` which returns the current time in unix milliseconds, so you can make this comparison using only `ints`

protocol/messenger.go
```diff
@@ -1691,6 +1833,7 @@ func (m *Messenger) dispatchMessage(ctx context.Context, spec common.RawMessage)
 	}
 	spec.ID = types.EncodeHex(id)
 	spec.SendCount++
+	spec.LastSent = uint64(time.Now().Unix())
```
cammellos: Same here, you probably want to use `m.getTimesource` so it's NTP synchronize

cammellos (commented):
Looks good, just a few minor comments about `getTimesource` but otherwise all good!

cammellos (approved):
Great job!

PR #2094, opened 2020-12-11
feat: allow getPreviewLinkData for all whitelisted links

emizzle:
Only YouTube links were supported in `getPreviewLinkData` previously.

Now, any whitelisted links can have their data retreived using this function, returning the true content type (by examining the header bytes of the content).

feat: add tenor.com and giphy.com to whitelisted urls list


protocol/urls/urls.go
```diff
@@ -87,9 +97,23 @@ func GetLinkPreviewData(link string) (previewData LinkPreviewData, err error) {
...
-	case "youtube.com", "www.youtube.com", "youtu.be":
-		return GetYoutubePreviewData(link)
+	hostname := strings.ToLower(url.Hostname())
+	youtubeHostnames := []string{"youtube.com", "www.youtube.com", "youtu.be"}
+	for _, youtubeHostname := range youtubeHostnames {
+		if youtubeHostname == hostname {
+			return GetYoutubePreviewData(link)
+		}
+	}
+	for _, site := range LinkPreviewWhitelist() {
```
cammellos, line 112: I would probably consider `Tenor` and `giphy` special cases, as they are image websites, so I would be more explicit in handling them.
Either I would use the case statement we had before:
```
case "youtube.com", "www.youtube.com"....
case ""
```
But I take your concern is that you want to have subdomains as well (otherwise case statement should be fine).

If that's the case, maybe we can do something like:
```
if strings.HasSuffix(hostname, site.Address) && site.ImageWebsite {
   return GetImageWebsitePreviewData(link) 
}
```
and you'd have to add `ImageWebsite` bool to the struct

So that if someone adds a new website, that is not an image website, the function still throws an error, as we haven't implemented pulling `title`/`description` etc

What do you think?
Otherwise looks good to me, thank you!

emizzle: @cammellos, updated as per your suggestion.

cammellos (approved):

protocol/urls/urls.go
```diff
@@ -18,24 +19,38 @@ type LinkPreviewData struct {
...
 	ThumbnailURL string `json:"thumbnailUrl"`
+	ContentType  string `json:"contentType"`
 }
 
 type Site struct {
-	Title   string `json:"title"`
-	Address string `json:"address"`
+	Title        string `json:"title"`
+	Address      string `json:"address"`
+	ImageSite    bool   `json:imageSite`
```
cammellos:
```suggestion
	ImageSite    bool   `json:"imageSite"`
```

cammellos (approved):
All good, thanks for the contribution! 
Would you mind bumping the VERSION file as well?
Let me know when you are happy with it and I can merge it.

emizzle:
> All good, thanks for the contribution!
> Would you mind bumping the VERSION file as well?
> Let me know when you are happy with it and I can merge it.

@cammellos VERSION bumped and rebased against latest `develop`

emizzle:
> All good, thanks for the contribution!
> Would you mind bumping the VERSION file as well?
> Let me know when you are happy with it and I can merge it.

@cammellos VERSION bumped, and I've rebased against latest develop.

cammellos:
@emizzle linter is complaining, (I can fix it for you), but in general you can fix this warnings by running:
```
goimports -w -local 'github.com/ethereum/go-ethereum,github.com/status-im/status-go,github.com/status-im/markdown' {filename}
```
Are you ok then with merging this?

PR #2099, opened 2020-12-16
Remove incorrect envelope sent event

vkjr:
Currently when we send envelope to peer that doesn't answer with acknowledgements, envelope treated as "sent". 
That is incorrect and fixed.

Also flag was renames `mailserverConfirmations` -> `awaitOnlyMailServerConfirmations` to make its intention clear. [(it wasn't before)](https://github.com/status-im/status-go/pull/1382/files#r258946919)


vkjr:
Converting back to draft while figuring out difference between local and CI test results

vkjr:
@cammellos, could you please take a look when have a minute?

vkjr:
Closed in favor of #2102

PR #2100, opened 2020-12-17
Expand Local Notifications to support multiple Notification types

Samyoul:
Expands local notifications functionality that currently handles only Transaction notifications, to also handle notifications for new incoming messages.

See notes https://notes.status.im/notifications-android


services/local-notifications/core.go
```diff
@@ -101,6 +112,53 @@ func NewService(appDB *sql.DB, network uint64) *Service {
...
+		}
+
+	default:
+		return nil, fmt.Errorf("unknown NotificationType '%s'", n.Body.GetType())
+	}
+
+	alias := struct {
+		ID            common.Hash  `json:"id"`
+		Platform      float32      `json:"platform,omitempty"`
+		Body          []byte       `json:"body"`
```
cammellos: I think the issue is that you declare `Body` as `[]byte`, meaning that it will be `base64` encoded in the `JSON`.

If you want it to serialize as  JSON (i.e `"body": {"something": ...}`) 
I believe you have to use `json.RawMessage`
https://golang.org/pkg/encoding/json/#RawMessage

which is used for fields that are already serialized

Samyoul: Yes, this was part of the problem. `json.RawMessage` was useful for delaying json unmarshalling and handling pre-marshalled data.

But the problem I found was because the `Body.Type` had not been set and `signal.send()` was swallowing the error that gave the marshalling failure.

services/local-notifications/core.go
```diff
@@ -354,3 +469,19 @@ func (s *Service) Protocols() []p2p.Protocol {
...
+			BodyType: TypeMessage,
+			Category: "", // TODO what category do we want?
+			Deeplink: "", // TODO find what if any Deeplink should be used here
+			Image:    "", // TODO do we want to attach any image data contained on the MessageBody{}?
+		})
+	}
+
+	// sends notifications messages to the OS level application
+	pushMessages(ns)
+}
```
Samyoul: @cammellos @vkjr the `Notification` struct has a couple of fields that look helpful for a building the OS notification, I've asked some questions in the code would you mind give your opinion on these?

cammellos: "Category" could be `newMessage`? and deep link I will let you know, we can sort it out at a later stage. Image similarly, for now it would be enough just to display text messages.

cammellos (approved):
looks good to me, we can start plugging it into status-react and use it!
Version needs bumping btw.

vkjr (approved):

PR #2102, opened 2020-12-18
Remove incorrect envelope sent event

vkjr:
Currently when we send envelope to peer that doesn't answer with acknowledgements, envelope treated as "sent". 
That is incorrect and fixed.

Misprint fixed in aknowledgement event that was a reason why acknowledgement didn't process even when received

Also flag was renames `mailserverConfirmations` -> `awaitOnlyMailServerConfirmations` to make its intention clear. [(it wasn't before)](https://github.com/status-im/status-go/pull/1382/files#r258946919)


cammellos (approved):
Looks all good to me, good find spotting this!

vkjr:
blocked by #2115

vkjr:
Per discussion with @cammellos changes splitted between prs #2128 and #2129 for easier upgrade.

PR #2106, opened 2020-12-28
[wallet] Show PNs right after enabling

### LINKED ISSUE #11535: Push notification about incoming tx is received only after reopening the app
# Bug Report

## Problem

If User receives an incoming tx while in the foreground, he will not receive a push notification about this until the application is reopened.

#### Expected behavior

Push notification for incoming tx is received while in the foreground without reopening the app.

#### Actual behavior

Reopening is required. 

### Reproduction

Receiver:
- Enable notifications (background service and wallet txs notifications)
- Keep Status in the foreground

Sender:
- Send some ETH or tokens to Receiver from Wallet view

Receiver:
- Put the app to the background
- Reopen the app

### Additional Information

- Status version: Nightly 10/12

cc @rasom

rasom:
fixes status-im/status-react#11535

So the idea is to start "wallet watcher" regardless of settings to allow it to handle `maxKnownBlocks` event on startup. In this case "watcher" is aware starting from which block fetched transaction should be treated as new and requires PNs.

status: ready


api/geth_backend.go
```diff
@@ -886,29 +886,6 @@ func (b *GethStatusBackend) AppStateChange(state string) {
 	b.log.Info("App State changed", "new-state", s)
 	b.appState = s
 
-	if s == appStateBackground {
```
rasom, line 889: This is not needed anymore, we always starts/stop wallet from the client at the moment

cammellos (approved):

PR #2108, opened 2021-01-04
Add ClearHistory & DeactivateChat methods

cammellos:
Adds 2 methods, clear history and deactivate chat.
These are functionalities that are already used in the client, but done through a mixture of different endpoints.
This commit consolidates the logic so it can be done on a single call.
It also adds a few endpoints:
- AddContact
- RemoveContact

But those are still wip.
It moves also a bunch of messenger methods on their own file for clarity.


jakubgs:
Why is this adding `AddContact` and `RemoveContact` if they are still work in progress?
Could we at least do it as a separate commit or are the changes connected to the other changes?

protocol/contact.go
```diff
@@ -83,6 +82,16 @@ func (c Contact) IsBlocked() bool {
...
 }
 
+func (c *Contact) Remove() {
+	var newSystemTags []string
+	for _, tag := range newSystemTags {
+		if tag != contactAdded {
+			newSystemTags = append(newSystemTags, tag)
+		}
+	}
+	c.SystemTags = newSystemTags
```
jakubgs, line 94: I don't get what this method is doing.

protocol/message_persistence.go
```diff
@@ -1131,3 +1150,78 @@ func (db sqlitePersistence) InvitationByID(id string) (*GroupChatInvitation, err
...
+	err := db.saveChat(tx, *chat)
+	if err != nil {
+		return err
+	}
+
+	return db.clearHistory(chat, currentClockValue, tx)
+}
+
+func (db sqlitePersistence) clearHistory(chat *Chat, currentClockValue uint64, tx *sql.Tx) error {
+	// Set deleted at clock value
```
jakubgs: That's not a very helpful comment, we can see in this line that's what happens:
```go
chat.DeletedAtClockValue = currentClockValue
```
But why does it need to be set? That's what the comment should say.

protocol/messenger.go
```diff
@@ -2089,59 +2089,40 @@ func (m *Messenger) DeleteChat(chatID string) error {
...
+		return nil, err
 	}
 
-	contact.Identicon = identicon
-	contact.Alias = name
-
-	if m.isNewContact(contact) || m.hasNicknameChanged(contact) {
-		err := m.syncContact(context.Background(), contact)
+	if chat.Public() {
+		err := m.reregisterForPushNotifications()
```
jakubgs, line 2109: "reregister"? Shouldn't this be called "deregister"? Or am I missing what this actually does?

protocol/messenger.go
```diff
@@ -2089,59 +2089,40 @@ func (m *Messenger) DeleteChat(chatID string) error {
...
-
-	m.allContacts[contact.ID] = contact
+	m.allChats[chatID] = chat
 
-	// Reregister only when data has changed
-	if shouldReregisterForPushNotifications {
-		return m.reregisterForPushNotifications()
-	}
+	response.Chats = []*Chat{chat}
+	// TODO: Remove filters
```
jakubgs, line 2118: Is this TODO important?

jakubgs (approved):
I don't see anything major wrong with this. Just some tiny nitpicks.

protocol/contact.go
cammellos: It's removing the `newSystemTag` from a contact (effectively removing the contact). I will add a comment

protocol/messenger.go
cammellos: it's re-registering effectively, as the fact that a contact has been removed means that we need to re-register with the PN server as our options have changed, I can add a comment

PR #2110, opened 2021-01-07
Fix communities migration

cammellos:
In some instances the communities migration would be skipped but not
marked as `dirty`.

This commit addresses the issue by:
- Making sure that if dirty is set the migration is not skipped but
replayed
- If the version is on the communities migration and dirty is false, we
check for the presence of the communities table. If not present we
replay the communities migration.
- Make community_id field in user_messages nullable

It also removes all the `down` migration, as we can't use them
effectively, as explained in the README.md added.


Samyoul (approved):
Looks good. This looks like it was a pain to debug.

Question do we want to remove all down migrations in the whole repo? Like you say there isn't really a valid/safe process for down migrating.

cammellos:
@Samyoul probably at some point it might be worth it, though some migrations we can keep (`mailserver` as those are using postgresql)

jakubgs (approved):
I don't know shit about Go SQL migrations but I like the `README` you added and the contents of `protocol/sqlite/ad_hoc_migration.go` make sense to me.

PR #2111, opened 2021-01-08
Send abridged version of history with messages in group chats

cammellos:
When sending a message in a private group chat we send the whole history
for redundancy and allow out-of-order processing.
This can be very expensive in some chats, resulting in long delay when
sending a message and calculating the POW.
This commit improves the performance  by only forwarding the events
necessary for the user to be able to construct a group chat and process
correctly the message.


vkjr (approved):

VERSION
```diff
@@ -1 +1 @@
-0.68.0
+0.68.1
```
Samyoul: I notice that your message caching also has the same version number. I don't know how much this matters,

protocol/v1/membership_update_message.go
```diff
@@ -314,6 +314,78 @@ func (g Group) Events() []MembershipUpdateEvent {
...
+				events = append(events, event)
+			}
+		}
+
+	}
+
+	// Reverse events
+	for i, j := 0, len(events)-1; i < j; i, j = i+1, j-1 {
+		events[i], events[j] = events[j], events[i]
+	}
```
Samyoul, line 384: 😄 reminds me of algorithm practice.

Samyoul (approved):
Looks great.

cammellos:
Closing as tested in a separate PR manually tested by QA, feedback has been addressed there

PR #2112, opened 2021-01-08
Cache waku messages

cammellos:
This commit re-introduces a feature that we lost during the migration to status-go.
Messages are cached for a couple of days if processed correctly by status-go, to avoid performance issues.
It's at-leat-once processing.


protocol/messenger.go
```diff
@@ -1216,11 +1223,10 @@ func (m *Messenger) CreateGroupChatWithMembers(ctx context.Context, name string,
 	m.allChats[chat.ID] = &chat
 
 	_, err = m.dispatchMessage(ctx, common.RawMessage{
```
vkjr: I'd suggest to wrap similar calls to function like `dispatchMembershipUpdateMessage` to avoid simultaneous changes later

protocol/transport/cache.go
```diff
@@ -0,0 +1,81 @@
+package transport
```
vkjr: I'm wondering if `cache` is a best naming for this functionality. I would suspect from name that it contains some messages for quick access. But in fact we keep track of processed messages here. wdyt?

vkjr (approved):

protocol/transport/cache.go
cammellos: If we don't mind being verbose, we could call it:
`ProcessedMessageIDsCache` 
what do you think?

vkjr: super) I love verbose names)

protocol/messenger.go
```diff
@@ -2875,7 +2882,10 @@ func (m *Messenger) handleRetrievedMessages(chatWithMessages map[transport.Filte
 
 	logger := m.logger.With(zap.String("site", "RetrieveAll"))
 	for _, messages := range chatWithMessages {
+		var processedMessages []string
 		for _, shhMessage := range messages {
+			// Indicates tha all messages in the batch have been processed correctly
+			messageProcessed := true
```
Samyoul: Maybe rename this var to `allMessagesProcessed` ?

protocol/messenger.go
```diff
@@ -2917,6 +2927,7 @@ func (m *Messenger) handleRetrievedMessages(chatWithMessages map[transport.Filte
 					c, err := buildContact(publicKey)
 					if err != nil {
 						logger.Info("failed to build contact", zap.Error(err))
+						messageProcessed = false
 						continue
 					}
```
Samyoul: Part of me thinks that we may need to wrap this failure functionality, future changes may forget to include the new `messageProcessed = false` line.

We could maintain the `messageProcessed` state in the Messenger struct and call an unexported func a something like:

```
handledMessageProcessFail(msg string, err error, l logger) {
  l.Info(msg, err)
  m.allMessagesProcessed = false
}
```

This could be over engineering. I think if we add more than just logging and setting a var to false, 100% this should be wrapped, but at the moment we're sort of on the border.

Maybe leave it as it is but keep an eye on if we need to add more functionality to a failed message process.

Samyoul (approved):
Looks good. Only minor comments about naming and code reuse

PR #2114, opened 2021-01-11
Add request/approve joining of community

cammellos:
This commit adds endpoint for requesting/approving to join a community.

There's also a few changes:

- ENS name resolving has been separated, this is needed to have communties ENS resolution (not implemented currently), and avoid locking
- CreateChat methods returns pointers for consistency
- Join returns the filters created so they can be passed to the frontend and we can move more logic in status-go


protocol/common/message_processor.go
```diff
@@ -186,6 +203,39 @@ func (p *MessageProcessor) SendGroup(
...
+	}
+
+	messageID := v1protocol.MessageID(&rawMessage.Sender.PublicKey, wrappedMessage)
+	rawMessage.ID = types.EncodeHex(messageID)
+
+	// Notify before dispatching, otherwise the dispatch subscription might happen
+	// earlier than the scheduled
+	p.notifyOnScheduledMessage(rawMessage)
+
+	// When SkipEncryption is set we don't pass the message to the encryption layer
```
vkjr: probably leftover from `sendPrivate()` code

protocol/communities/community_test.go
```diff
@@ -716,6 +720,7 @@ func (s *CommunitySuite) configOnRequest() Config {
 	description := s.emptyCommunityDescription()
 	description.Permissions.Access = protobuf.CommunityPermissions_ON_REQUEST
 	return Config{
+		MemberIdentity:       &s.identity.PublicKey,
```
vkjr: Probably it worth adding function like `newConfig(s.identity, description)`

protocol/communities/community.go
```diff
@@ -294,17 +401,19 @@ func (o *Community) RemoveUserFromOrg(pk *ecdsa.PublicKey) (*protobuf.CommunityD
...
 	}
 
+	o.increaseClock()
+
 	return o.config.CommunityDescription, nil
 }
 
 // TODO: this should accept a request from a user to join and perform any validation
-func (o *Community) AcceptRequestToJoin(pk *ecdsa.PublicKey) (*protobuf.CommunityRequestJoinResponse, error) {
+func (o *Community) AcceptRequestToJoin(pk *ecdsa.PublicKey) (*protobuf.CommunityRequestToJoinResponse, error) {
```
vkjr: Should these stay as `TODO` for now?

protocol/communities/community.go
```diff
@@ -423,8 +532,8 @@ func (o *Community) HandleCommunityDescription(signer *ecdsa.PublicKey, descript
 	return response, nil
 }
 
-// HandleRequestJoin handles a request, checks that the right permissions are applied and returns an CommunityRequestJoinResponse
-func (o *Community) HandleRequestJoin(signer *ecdsa.PublicKey, request *protobuf.CommunityRequestJoin) error {
+// HandleRequestToJoin handles a request, checks that the right permissions are applied
+func (o *Community) HandleRequestToJoin(signer *ecdsa.PublicKey, request *protobuf.CommunityRequestToJoin) error {
```
vkjr: There is a comment at the end of this function - `Store request to join`, but there is no logic for storing. 
I see that `HandleCommunityRequestToJoin` calls this function and in fact stores request if there are no errors.
So I would suppose that purpose of the function is to check if request is correct for the community. `HandleRequestToJoin` name doesn't really reflect function's purpose so I suggest to rename it.
For example to something like `ValidateRequestToJoin()`

In general, I think that any function that starts from `handle..`, `process..`, `manage..` is a good candidate for renaming because those verbs are non-informative and complicate understanding. Probably exception is when function is very highlevel and it is not possible to describe it more precisely.

protocol/communities/manager.go
```diff
@@ -209,21 +270,51 @@ func (m *Manager) HandleCommunityDescriptionMessage(signer *ecdsa.PublicKey, des
 		}
 	}
 
-	_, err = community.HandleCommunityDescription(signer, description, payload)
+	changes, err := community.HandleCommunityDescription(signer, description, payload)
```
vkjr: Maybe worth renaming `HandleCommunityDescription` -> `FindCommunityChanges` ?

protocol/communities/manager.go
```diff
@@ -209,21 +270,51 @@ func (m *Manager) HandleCommunityDescriptionMessage(signer *ecdsa.PublicKey, des
...
+
+	return &CommunityResponse{
+		Community: community,
+		Changes:   changes,
+	}, nil
 }
 
 // TODO: Finish implementing this
-func (m *Manager) HandleCommunityInvitation(signer *ecdsa.PublicKey, invitation *protobuf.CommunityInvitation, payload []byte) (*Community, error) {
+func (m *Manager) HandleCommunityInvitation(signer *ecdsa.PublicKey, invitation *protobuf.CommunityInvitation, payload []byte) (*CommunityResponse, error) {
```
vkjr, line 318: If `TODO` still valid maybe we can add more details on what should be implemented.
is `save grant` few lines lower is a leftover?

protocol/messenger.go
```diff
@@ -788,106 +805,56 @@ func (m *Messenger) handleEncryptionLayerSubscriptions(subscriptions *encryption
 	}()
 }
 
-func (m *Messenger) publishOrg(org *communities.Community) error {
-	m.logger.Debug("publishing org", zap.String("org-id", org.IDString()), zap.Any("org", org))
-	payload, err := org.MarshaledDescription()
-	if err != nil {
-		return err
-	}
+func (m *Messenger) handleENSVerified(records []*ens.VerificationRecord) {
```
vkjr, line 808: Maybe `saveContactsFromVerifiedEns`?

protocol/messenger_response.go
```diff
@@ -9,70 +11,199 @@ import (
...
+	var communities []*communities.Community
+	for _, c := range r.communities {
+		communities = append(communities, c)
+	}
+	return communities
 }
 
 func (m *MessengerResponse) IsEmpty() bool {
-	return len(m.Chats)+len(m.Messages)+len(m.Contacts)+len(m.Installations)+len(m.Invitations)+len(m.EmojiReactions)+len(m.Communities)+len(m.CommunityChanges)+len(m.Filters)+len(m.RemovedFilters)+len(m.RemovedChats)+len(m.Notifications)+len(m.MailserverTopics)+len(m.Mailservers)+len(m.MailserverRanges) == 0
+	return len(m.chats)+len(m.Messages)+len(m.Contacts)+len(m.Installations)+len(m.Invitations)+len(m.EmojiReactions)+len(m.communities)+len(m.CommunityChanges)+len(m.Filters)+len(m.RemovedFilters)+len(m.removedChats)+len(m.MailserverTopics)+len(m.Mailservers)+len(m.MailserverRanges)+len(m.Notifications)+len(m.RequestsToJoinCommunity) == 0
```
vkjr: Looks scary) I'd suggest to split by lines

protocol/messenger_response.go
```diff
@@ -9,70 +11,199 @@ import (
...
+			m.Messages = append(m.Messages, overrideMessage)
 		}
 	}
 }
 
-// overrideMessages append new messages and override existing ones in response.Messages
-func (m *MessengerResponse) overrideMessages(messages []*common.Message) {
-	for _, overrideMessage := range messages {
+// overrideFilters append new filters and override existing ones in response.Filters
+func (m *MessengerResponse) overrideFilters(filters []*transport.Filter) {
```
vkjr: overrideFilters and overrideRemovedFilters are identical except used used field. Can be implemented as single function with pointer to field passed

vkjr (approved):

PR #2116, opened 2021-01-14
fix protocol.MessageNotificationBody marshalling

rasom:
status: ready

### DIFF
--- a/services/local-notifications/core.go
+++ b/services/local-notifications/core.go
@@ -137,7 +137,7 @@ func (n *Notification) MarshalJSON() ([]byte, error) {
 
 	switch n.BodyType {
 	case TypeTransaction:
-		if nb, ok := n.Body.(*notificationBody); ok {
+		if nb, ok := n.Body.(notificationBody); ok {
 			body, err = json.Marshal(nb)
 			if err != nil {
 				return nil, err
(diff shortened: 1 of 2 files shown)


cammellos (approved):

services/local-notifications/core.go
```diff
@@ -475,10 +475,11 @@ func (s *Service) IsStarted() bool {
 }
 
 func SendMessageNotifications(mnb []protocol.MessageNotificationBody) {
+	log.Info("sending message notification", "notification", mnb)
 	var ns []*Notification
 	for _, n := range mnb {
 		ns = append(ns, &Notification{
-			Body:     n,
+			Body:     &n,
 			BodyType: TypeMessage,
```
cammellos: Here you are taking the address of a temporary variable, and it won't work.
You need to take the address from the array:

```
for i, n := range mnb {
   ...
   Body: &mnb[i],
  ... 
}
```
Though not sure the type is the right one.

rasom: well I went another way, can you check it please?

cammellos: probalby is best to have some consistency, either we always pass a pointer or never, so I'd say we either modify the above to have an actual object, or pass a pointer in yours instead of the object

cammellos (approved):

PR #2119, opened 2021-01-18
Skip wrapping emojis in private group chats

cammellos:
This commit fixes a couple of issues:
1) Emojis were sent to any member of the group chat, regardless of
whether they joined
2) We don't want to wrap emojis, as there's no need to do so, only
messages are to be wrapped


vkjr (approved):

protocol/messenger.go
```diff
@@ -4488,7 +4495,7 @@ func (m *Messenger) SendEmojiReactionRetraction(ctx context.Context, emojiReacti
 	return &response, nil
 }
 
-func (m *Messenger) encodeChatEntity(chat *Chat, message common.ChatEntity) ([]byte, error) {
+func (m *Messenger) encodeChatEntity(chat *Chat, message common.ChatEntity, wrapGroupMessages bool) ([]byte, error) {
```
Samyoul, line 4498: Changing the signature of this function to handle a conditional feels like the `wrapGroupMessages` state should be a property of either the `Chat` or the `message`, rather than stored in external bool. WDYT?

cammellos: `chat` is not enough to decide, as we wrap messages but not emojis in group chat.
We could though implement it in `chatEntity`, I might change that, good idea.
Thanks

cammellos: Actually, is not so simple :) 
if we want to resend the message, we don't have access to the `ChatEntity` anymore, as we only store the raw data (we don't want to encode/reencode as protobuf is not deterministic), so getting access to the original chatentity would make things a bit hairy , do you mind if we keep it as it is?

Samyoul: Isn't an implementing `ChatEntity` struct already a parsed protobuf? At the moment it is only an instance of `common.Message` or `EmojiReaction`, which seem to require a parsed protobuf type.

EDIT : Isn't `encodeChatEntity` called before the raw message is stored for resend?

cammellos: I think you are right, I got confused, I'll have a look tomorrow, thanks!

cammellos:
Closing as tested in a separate PR manually tested by QA, feedback has been addressed there

PR #2121, opened 2021-01-18
Clean topics that we don't listen to

cammellos:
There was a bug on status-react where it would save filters that were
not listened to.
This commit adds a task to clean up those filters as they might result
in long syncing times.


protocol/messenger.go
```diff
@@ -430,9 +433,27 @@ func (m *Messenger) Start() error {
 	m.watchConnectionChange()
 	m.watchExpiredEmojis()
 	m.watchIdentityImageChanges()
+	if err := m.cleanTopics(); err != nil {
+		return err
+	}
 	return nil
 }
 
+// cleanTopics remove any topic that does not have a Listen flag set
```
vkjr, line 464: This function takes filters from transport and instructs mailserver database to remove some. Looks like it doesn't work with anything messenger-specific. I would suggest not to make it a part of Messenger. Maybe it better fits Transport? Wdyt?

cammellos: I was thinking that eventually mailserver logic should be wrapped in messenger, as status-react should not be directly worrying about filters (it does now as we haven't fully migrated the code from status-react to status-go).
Essentially the flow should be for example:

1) Status-react creates a chat through messenger `CreateChat` endpoint
2) Messenger creates a chat, filter and mailserver topics and requests messages from mailserver

Instead currently we have

1) Status react creates a chat through messenger
2) Status-react ask messenger to create filters
3) Status-react saves the topics derived from filters to the database

So for now I agree it would make more sense to have it separate, but eventually mailserver logic should be handled by messenger, as it's dependent on creating chats etc (that's something I'd like to work next as we keep having issue with it)

What do you think?

vkjr: Thanks for explanation! With that in mind it absolutely has sense)

vkjr (approved):

protocol/transport/filters_manager.go
```diff
@@ -230,6 +231,33 @@ func (s *FiltersManager) Remove(filters ...*Filter) error {
...
+	if err := s.service.UnsubscribeMany(filterIDs); err != nil {
+		return err
+	}
+
+	for _, filter := range filters {
+		if filter.SymKeyID != "" {
+			s.service.DeleteSymKey(filter.SymKeyID)
+		}
+		delete(s.filters, filter.ChatID)
+	}
```
Samyoul, line 256: This uses 3 for loops to go through the same information. Once to find if we don't listen, once to unsubscribe and once to delete the filters. Would the below do the same with less loops?

```suggestion
	for _, f := range filters {
		if !f.Listen {
			filterIDs = append(filterIDs, f.FilterID)
			filters = append(filters, f)
			
			if err := s.service.Unsubscribe(f.FilterID); err != nil {
				return err
			}
			
			if filter.SymKeyID != "" {
				s.service.DeleteSymKey(filter.SymKeyID)
			}
			delete(s.filters, filter.ChatID)
		}
	}
```

Samyoul (approved):
Looks good. Only one small question of optimisation.

protocol/transport/filters_manager.go
cammellos: Not quite, because `Unsubscribe` trigger a waku message, so that would be `n` messages, so it's best to collect and unsubscribe all in one go, producing a single waku message, and after, we want to delete filters only if we managed to unsubscbribe, otherwise we might be in some inconsistent state, so not sure we can avoid the looping

PR #2123, opened 2021-01-20
🐱 Giphy and  Tenor preview data

shivekkhurana:
Giphy and Tenor links are links to HTML sites not the actual gif source. This PR parses valid Giphy/Tenor links and use their oembed services to get gif sources from links to HTML. 

This will help us show the gifs in chat messages: https://github.com/status-im/status-react/issues/11580

### DIFF
--- a/protocol/urls/urls.go
+++ b/protocol/urls/urls.go
@@ -22,12 +22,16 @@ type GiphyOembedData struct {
 	ProviderName string `json:"provider_name"`
 	Title        string `json:"title"`
 	URL          string `json:"url"`
+	Height       int `json:"height"`
+	Width        int  `json:"width"`
 }
 
 type TenorOembedData struct {
 	ProviderName string `json:"provider_name"`
 	ThumbnailURL string `json:"thumbnail_url"`
 	AuthorName   string `json:"author_name"`
+	Height       int `json:"height"`
+	Width        int `json:"width"`
 }
 
 type LinkPreviewData struct {
@@ -35,6 +39,8 @@ type LinkPreviewData struct {
 	Title        string `json:"title" meta:"og:title"`
 	ThumbnailURL string `json:"thumbnailUrl" meta:"og:image"`
 	ContentType  string `json:"contentType"`
+	Height       int `json:"height"`
+	Width        int `json:"width"`
 }
 
 type Site struct {
@@ -172,6 +178,8 @@ func GetGiphyPreviewData(link string) (previewData LinkPreviewData, err error) {
 	previewData.Title = oembedData.Title
 	previewData.Site = oembedData.ProviderName
 	previewData.ThumbnailURL = oembedData.URL
+	previewData.Height = oembedData.Height
+	previewData.Width = oembedData.Width
 
 	return previewData, nil
 }
@@ -231,6 +239,8 @@ func GetTenorPreviewData(link string) (previewData LinkPreviewData, err error) {
 	previewData.Title = oembedData.AuthorName // Tenor Oembed service doesn't return title of the Gif
 	previewData.Site = oembedData.ProviderName
 	previewData.ThumbnailURL = oembedData.ThumbnailURL
+	previewData.Height = oembedData.Height
+	previewData.Width = oembedData.Width
 
 	return previewData, nil
 }
@@ -248,7 +258,7 @@ func GetLinkPreviewData(link string) (previewData LinkPreviewData, err error) {
 		return GetYoutubePreviewData(link)
 	case "github.com":
 		return GetGithubPreviewData(link)
-	case "giphy.com":
+	case "giphy.com", "media.giphy.com":
 		return GetGiphyPreviewData(link)
 	case "gph.is":
 		return GetGiphyShortURLPreviewData(link)
--- a/protocol/urls/urls.go
+++ b/protocol/urls/urls.go
@@ -179,6 +179,7 @@ func GetGiphyPreviewData(link string) (previewData LinkPreviewData, err error) {
 // Giphy has a shortener service called gph.se, the oembed service doesn't work with shortened urls,
 // so we need to fetch the long url first
 func GetGiphyLongURL(shortURL string) (longURL string, err error) {
+	// nolint: gosec
 	res, err := http.Get(shortURL)
 
 	if err != nil {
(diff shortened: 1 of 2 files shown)


protocol/urls/urls.go
```diff
@@ -140,6 +212,12 @@ func GetLinkPreviewData(link string) (previewData LinkPreviewData, err error) {
 	if "github.com" == hostname {
```
cammellos: Probably at this point a `switch/case` statement is more idiomatic.

I think we can have a single switch statement for all:
```
switch hostname {
   case "github.com": 
    ....
    case "giphy": 
    ...
    case "tenor.com":
    ...
    case "youtube.com", "youtube.be", ....:
     ... 
    default:
     	return previewData, fmt.Errorf("Link %s isn't whitelisted. Hostname - %s", link, url.Hostname())
}

```
Since all the cases in the list are actually well defined, we don't need anymore to iterate on the list

cammellos (commented):
Looks all good, nice work!
The only thing I'd change is to have a switch statement for the multiple cases, since that's all covered I believe.
nice gifs choice for the tests :)

shivekkhurana:
Giphy just returned a different link to the image source:

```go test -run='TestGetLinkPreviewData|TestGetGiphyPreviewData|TestGetTenorPreviewData' .
--- FAIL: TestGetGiphyPreviewData (0.27s)
    urls_test.go:45: 
        	Error Trace:	urls_test.go:45
        	Error:      	Not equal: 
        	            	expected: "https://media1.giphy.com/media/lcG3qwtTKSNI2i5vst/giphy.gif"
        	            	actual  : "https://media2.giphy.com/media/lcG3qwtTKSNI2i5vst/giphy.gif"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-https://media1.giphy.com/media/lcG3qwtTKSNI2i5vst/giphy.gif
        	            	+https://media2.giphy.com/media/lcG3qwtTKSNI2i5vst/giphy.gif
        	Test:       	TestGetGiphyPreviewData
FAIL
FAIL	github.com/status-im/status-go/protocol/urls	1.470s
FAIL

Go-Test exited abnormally with code 1 at Wed Jan 20 17:23:26
```

`media2` instead of `media1`. The tests should be more resilient.

I'm new to Golang. Any ideas about how to handle this are appreciated.

protocol/urls/urls.go
shivekkhurana: I have made the change, the tests pass, but can you still check it once?
If it's okay, please resolve this comment.

protocol/urls/urls_test.go
```diff
@@ -30,3 +30,40 @@ func TestGetLinkPreviewData(t *testing.T) {
...
 
 }
+
+func TestGetGiphyPreviewData(t *testing.T) {
+	validGiphyLink := "https://giphy.com/gifs/FullMag-robot-boston-dynamics-dance-lcG3qwtTKSNI2i5vst"
+	previewData, err := GetGiphyPreviewData(validGiphyLink)
+	bostonDynamicsEthGifData := LinkPreviewData{
+		Site:         "GIPHY",
+		Title:        "Boston Dynamics Yes GIF by FullMag - Find & Share on GIPHY",
+		ThumbnailURL: "https://media1.giphy.com/media/lcG3qwtTKSNI2i5vst/giphy.gif",
```
cammellos, line 46: If they return different subdomains, I would probably the path and just match against the path, similarly as we do in the `switch/case` statement, not as effective, but should do.

cammellos (approved):

cammellos:
Ah you want as well to bump `VERSION`, patch version probably, or minor, up to you.

Samyoul (approved):
Great work on your first PR to status-go @shivekkhurana , looks good. The only point on the code is that the go convention for errors is that they start with lowercase and do not end with any punctuation.

The only other point I'd make is related to our general approach to 3rd party content host and not directly related to your work. I dont like recipient side link unfurling, even with a white list, unless we give users the option to opt out of unfurling and/or package unfurling data into a protobuf and send via Waku.

Ps this is less of a comment on your work and more of pointer on maintaining our application's privacy features. 😁

shivekkhurana:
> The only point on the code is that the go convention for errors is that they start with lowercase and do not end with any punctuation.

Did I break this convention somewhere in this PR ? I tried looking but wasn't able to find anything. 

> The only point on the code is that the go convention for errors is that they start with lowercase and do not end with any punctuation.

Moral Dilemma 101.

cammellos:
> The only other point I'd make is related to our general approach to 3rd party content host and not directly related to your work. I dont like recipient side link unfurling, even with a white list, unless we give users the option to opt out of unfurling and/or package unfurling data into a protobuf and send via Waku.

Currently the user needs to opt-in to unfurl links, they are all disabled by default, though I agree with you that sender side link preview is generally safer and more established.

protocol/urls/urls.go
```diff
@@ -123,35 +136,84 @@ func GetGithubPreviewData(link string) (previewData LinkPreviewData, err error)
...
-func GetLinkPreviewData(link string) (previewData LinkPreviewData, err error) {
+func GetGiphyOembed(url string) (data GiphyOembedData, err error) {
+	oembedLink := fmt.Sprintf(GiphyOembedLink, url)
+
+	jsonBytes, err := GetURLContent(oembedLink)
 
-	url, err := url.Parse(link)
 	if err != nil {
-		return previewData, fmt.Errorf("Cant't parse link %s", link)
+		return data, fmt.Errorf("Can't get bytes from Giphy oembed response at %s", oembedLink)
```
Samyoul: Example, this error begins with an upper case letter

protocol/urls/urls.go
```diff
@@ -123,35 +136,84 @@ func GetGithubPreviewData(link string) (previewData LinkPreviewData, err error)
...
 
-	hostname := strings.ToLower(url.Hostname())
-	youtubeHostnames := []string{"youtube.com", "www.youtube.com", "youtu.be"}
-	for _, youtubeHostname := range youtubeHostnames {
-		if youtubeHostname == hostname {
-			return GetYoutubePreviewData(link)
-		}
+	err = json.Unmarshal(jsonBytes, &data)
+	if err != nil {
+		return data, fmt.Errorf("Can't unmarshall json")
```
Samyoul: This error too, to be clear the convention is error messages not the variable names.

protocol/urls/urls.go
Samyoul, line 145: This is an example, to be clear the convention applies to error text not the error variable.

protocol/urls/urls.go
shivekkhurana: Understood. I was looking at the error variable name. Do you want me to change it? The errors started with an uppercase letter in other parts of this package.

Samyoul: Hey sorry for the late reply, I didnt get pinged about your question and I haven't been reading my email this week.

What I normally do is if I spot something like this in preexisting code I make changes through the package. Yes please make these changes, it doesn't look like there are many, otherwise I would say just to leave them for a dedicated clean up.

shivekkhurana: I'll make the changes to this package.

Samyoul (approved):
Great work @shivekkhurana, looks good to me!

PR #2124, opened 2021-01-20
compose: parametrize docker image, bind to v0.64.3

jakubgs:
While working with @jinhojang6 on better node running instructions we found out that `latest` tag was borked.

Using `latest` tag is dangerous for non-technical users.
And updating `latest` tag willy-nilly is also bad.


cammellos (approved):

jinhojang6 (approved):

jakubgs:
I'm gonna assume the tests failing has nothing to do with my makefile changes, because that doesn't make sense.

PR #2125, opened 2021-01-23
fix "Pending" status for messages in private group chat

qfrank:
Fix "Pending" status for messages in private group chat, here is relevant [issue](https://github.com/status-im/status-react/issues/11555) .

When there is only one joined member(the creator) in private group chat, we won't really send the chat message out, instead we just mark OutgoingStatus as sent.


Samyoul (approved):
Looks good. Thank you for the PR.

Just a little note, it looks like your PR has failed the below e2e test:

```
[2021-01-24T02:41:18.782Z] --- FAIL: TestERC20Transfers (28.85s)
[2021-01-24T02:41:18.782Z]     --- PASS: TestERC20Transfers/TestInRange (9.15s)
[2021-01-24T02:41:18.782Z]     --- PASS: TestERC20Transfers/TestInboundEvent (5.36s)
[2021-01-24T02:41:18.782Z]     --- PASS: TestERC20Transfers/TestNoEvents (4.68s)
[2021-01-24T02:41:18.782Z]     --- FAIL: TestERC20Transfers/TestOutboundEvent (9.64s)
[2021-01-24T02:41:18.782Z]         downloader_test.go:248: 
[2021-01-24T02:41:18.782Z]             	Error Trace:	downloader_test.go:248
[2021-01-24T02:41:18.782Z]             	Error:      	Received unexpected error:
[2021-01-24T02:41:18.782Z]             	            	context deadline exceeded
[2021-01-24T02:41:18.782Z]             	Test:       	TestERC20Transfers/TestOutboundEvent
[2021-01-24T02:41:18.782Z] FAIL
[2021-01-24T02:41:18.782Z] FAIL	github.com/status-im/status-go/services/wallet	145.794s
```

Which looks flaky to me, so rerun the e2e and I think this PR will pass.

protocol/messenger.go
```diff
@@ -2343,6 +2343,12 @@ func (m *Messenger) SendChatMessages(ctx context.Context, messages []*common.Mes
 	return &response, nil
 }
 
+func (m *Messenger) isOnlyMyselfJoinedChat(chat *Chat, rawMessage common.RawMessage) bool {
```
cammellos: mmh, not sure exactly, but I think the logic should be simpler, 
we should mark the message as `sent` only if 
1) Is a private group chat 
and 
2) The message is not sent to anyone (i.e Recipients are nil or 0)

If we send to someone (even ourselves, in case we have paired messages), then we should wait for the the dispatched event and so it should no be marked as `sent`.

So I think the logic should simply be:

```
  return chat.ChatType == ChatTypePrivateGroupChat && len(rawMessage.Recipients) == 0
```

Also I would name the function differently, something like `propagatedToTheNetwork`/`dispatchedOverTheNetwork`, i.e something that indicates that we want to know whether the message has been sent through waku or not.
Maybe @Samyoul has a good idea about naming it.

cammellos (changes requested):
Thanks, looks good, only some changes in logic

protocol/messenger.go
Samyoul: Hmm, funny one, the func is positioned after the despatch but the question is not dependant on any altered state during the despatch. ie we can know before despatch if the message has a recipient, so asking a question past tense implies a change.

Maybe something like `isDispatchable()` would be better, as this naming is only a question of state rather than any actions performed.

Samyoul: Also this leaves the function to be expanded and used in other contexts, rather than in this specific case.

cammellos: It's dependent on altered state in some sense though, if you look at `dispatch`, we set `Recipients` in case of `ChatTypePrivateGroupChat`.
We could re-calculate it as we have `chat` and the `MessageType`, but it's a bit expensive, so it would be probably better to piggy-back on the value that has been calculated in `dispatch`.
What do you think?

Samyoul: Ah good point, I see that. Hmm, the only thing that I'd say is that this function is agnostic to whether a message has actually been despatched, it only cares if it could be despatched.

And alternative path could be to add a field on the `RawMessage` struct called `Despatched` which will be set by the despatcher `(p *MessageProcessor) sendPrivate` and can be queried at various points in the `Message`'s lifetime.

I suppose it depends on whether we want to know has the message **been** despatched or just if it **can be** despatched.

But if we add a `Despatched` field to the `RawMessage{}` struct, and set it to true only if `(p *MessageProcessor) sendPrivate` is successfully fired on the message, then we could do without this function and change the logic in `sendChatMessage()` to something like

```go
if chat.ChatType == ChatTypePrivateGroupChat && !rawMessage.Recipients.Despatched {
	message.OutgoingStatus = common.OutgoingStatusSent
}
```

cammellos: ah good idea,
actually `RawMessage` has already a `Sent` field, and we use that for re-transmission, so it makes sense to set that flag if it's not going out, so we don't even try to re-send it.
After we can check that field to set the `OutgoingStatus` i.e
```
if rawMessage.Sent {
  message.OutgoingStatus = common.OutgoingStatusSent
} else {
message.OutgoingStatus = common....Pending
}
```

protocol/messenger.go
```diff
@@ -2350,27 +2350,33 @@ func (m *Messenger) dispatchMessage(ctx context.Context, spec common.RawMessage)
 			spec.Recipients = spec.Recipients[:n]
 		}
 
+		// We won't really send the message out if there's no recipients
```
cammellos, line 2357: nice comment

cammellos (approved):
Looks good to me, @Samyoul what do you think?

Samyoul:
Yes, looks good to me.

PR #2126, opened 2021-01-25
Pass topics to mailserver cursor

cammellos:
The topics field was not passed to the mailserver, which meant that
queries were still using the old bloom filter.
Hopefully this is the last place where we need to pass this.

While investigating https://github.com/status-im/status-react/issues/11620#issuecomment-766669449
that issue came up.


jakubgs (approved):
Nice.

Samyoul (approved):

PR #2128, opened 2021-01-26
Confirmation setting fixed for mailservers

vkjr:
Currently mailservers don't send confirmations even when corresponding setting specified in config file. 
This PR fixes problem


params/config.go
```diff
@@ -211,6 +211,9 @@ type WakuConfig struct {
 
 	// SoftBlacklistedPeerIDs is a list of peer ids that should be soft-blacklisted (messages should be dropped but connection kept)
 	SoftBlacklistedPeerIDs []string
+
+	// when true, sends message confirmations
+	EnableConfirmations bool
```
Samyoul: A minor thing, go convention is that documentation comments begin with the name of the element they are describing https://blog.golang.org/godoc

```suggestion
	// EnableConfirmations when true, sends message confirmations
	EnableConfirmations bool
```

Samyoul (approved):
Looks good. Only one tiny little point

params/config.go
vkjr: didn't know this, thanks! :)

PR #2129, opened 2021-01-26
Confirmations processing fixed on client side

vkjr:
Fixes:
- don't treat envelope as successfully sent right after we send it to peer that doesn't support confirmations
- fixed misprint in event name (fixes confirmations receiving)
- flag renamed `mailserverConfirmations` -> `awaitOnlyMailServerConfirmations` to make its intention clear. [(it wasn't before)](https://github.com/status-im/status-go/pull/1382/files#r258946919)


Samyoul (approved):
Nice new naming.

protocol/transport/waku/envelopes.go
```diff
@@ -156,10 +157,8 @@ func (m *EnvelopesMonitor) handleEvent(event types.EnvelopeEvent) {
 }
 
 func (m *EnvelopesMonitor) handleEventEnvelopeSent(event types.EnvelopeEvent) {
-	if m.mailServerConfirmation {
-		if !m.isMailserver(event.Peer) {
-			return
-		}
+	if m.awaitOnlyMailServerConfirmations && !m.isMailserver(event.Peer) {
+		return
 	}
```
Samyoul: I like this refactor.

PR #2136, opened 2021-01-28
Make sure tokens balance can be fetched

rasom:
needed for https://github.com/status-im/status-react/issues/11674

We need to set to wallet `Ethclient` before we check history the first time as it might not happen on mobile network.


api/geth_backend.go
```diff
@@ -628,6 +628,11 @@ func (b *GethStatusBackend) startNode(config *params.NodeConfig) (err error) {
 	}); err != nil {
 		return
 	}
+	wallet, err := b.statusNode.WalletService()
+	if err != nil {
+		return err
+	}
+	wallet.SetClient(b.statusNode.RPCClient().Ethclient())
```
Samyoul: The var `wallet` collides with the package `github.com/status-im/status-go/services/wallet`, although this code is compliable it makes any reference to "wallet" in this function ambiguous. An easy fix is to rename the var to `walletService` or `w` or something cooler.

Samyoul (approved):
Looks good, just a var / package name collision and we need a VERSION bump 😄

api/geth_backend.go

services/wallet/service.go
```diff
@@ -51,6 +51,11 @@ func (s *Service) GetFeed() *event.Feed {
 	return s.feed
 }
 
+// SetClient sets ethclient.
+func (s *Service) SetClient (client *ethclient.Client) {
```
Samyoul:
```suggestion
func (s *Service) SetClient(client *ethclient.Client) {
```

Samyoul (commented):
Try this

PR #2137, opened 2021-01-29
Fix deadlock in peerpoool

cammellos:
This is a bit complicated, so:

1) Peerpool was subscribing to `event.Feed`, which is a global event
emitter for ethereum.
2) The p2p.Server was publshing on `event.Feed`, this triggered in the
same routine a publish on `event.Feed`.
3) Peerpool was listening to `event.Feed`, react on it, and in the same
routine, trigger some code on p2p.Server that would publish on
`event.Feed`

This meant that if the size of the channel was unbufferred, it would deadlock, as
peerPool would not be consuming when it would publish (the same go
routine publishes and listen effectively, through a lot of indirection
and non-buffered channels, p2p.Server->event.Feed)

The channel though was a buffered channel with size 10, and this
meant that most of the times is fine.

The issue is that peerpool is not the only producer to this channel.
So it's possible that while is processing an event, the buffer would
fill up, and it would hange trying to publish, and nobody is listening
to the channel, hanging EVERYTHING.

At least that's what I think, needs to be tested, but definitely an issue.

I kept the code changes to a minimum, this code is a bit hairy, but it's
fairly critical so I don't want to make too many changes.

### DIFF
--- a/peers/peerpool.go
+++ b/peers/peerpool.go
@@ -30,6 +30,7 @@ var (
 type PoolEvent string
 
 const (
+	immediately = 0 * time.Minute
 	// expirationPeriod is an amount of time while peer is considered as a connectable
 	expirationPeriod = 60 * time.Minute
 	// discoveryRestartTimeout defines how often loop will try to start discovery server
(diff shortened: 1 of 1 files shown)


jakubgs (approved):
Pretty scary stuff. Changing this code could easily break things even further.

PR #2144, opened 2021-02-05
Fix peerpool test

cammellos:
With the introduction of the new non-blocking code, events for the
peerpool might come out-of-order. That's generally not an issue, but it
made tests fail.
I have changed the code so that order is more consistent (It's still
theoretically possible that a stop signal would arrive out of order in a
real scenario, but impact is low and I don't want to change this code
too much).


shivekkhurana (approved):
Looks good to me.

PR #2145, opened 2021-02-05
[wallet] Merge blocks ranges when possible

rasom:
- old existing ranges are merged when wallet service is started
- a new range is merged with an existing one if possible

This will decrease the number of entries in blocks_range table as
currently it can grow indefinitely (@flexsurfer reported 23307 entries).

This change is also needed for further optimisations of RPC usage.

status: ready


services/wallet/database.go
```diff
@@ -790,31 +789,199 @@ func insertBlocksWithTransactions(creator statementCreator, account common.Addre
...
+	if len(ranges) <= 1 {
+		return nil
+	}
+
+	tx, err = db.db.Begin()
+	if err != nil {
+		return err
+	}
+
+	defer func() {
```
cammellos: I would use named return values here `(err error)`, because although it works fine in this case, if any code is added and does, something like:
`a, err := ...`
it would shadow `err` and the code below would stop working (for that `err`). 

If you use a named parameter, it would warn you and should be caught by either the compiler or the linter.

services/wallet/database.go
```diff
@@ -790,31 +789,199 @@ func insertBlocksWithTransactions(creator statementCreator, account common.Addre
...
+		_ = tx.Rollback()
+	}()
+
+	initValue := big.NewInt(-1)
+	prevFrom := big.NewInt(-1)
+	prevTo := big.NewInt(-1)
+	hasMergedRanges := false
+	newRanges := []*BlocksRange{}
+	deletedRanges := []*BlocksRange{}
+	for idx, blocksRange := range ranges {
```
cammellos, line 868: would you mind extracting this and adding a test please?

api/geth_backend.go
```diff
@@ -1261,6 +1261,11 @@ func (b *GethStatusBackend) startWallet(watchNewBlocks bool) error {
 		}
 	}
 
+	err = wallet.MergeBlocksRanges(allAddresses, b.statusNode.Config().NetworkID)
```
cammellos, line 1264: how expensive is this operation for large accounts?

rasom: It depends on device and the number of ranges, a bit hard to estimate... I thought about limiting it to 100-200 ranges per run, just to make sure we do not block the app for a few seconds in the worst case

services/wallet/database.go
rasom: sure

services/wallet/database.go
rasom: ok

api/geth_backend.go
cammellos: what about running it in the background, any implication?

rasom: Right now it is run on starting wallet, and if there is no new transactions we will try to upsert a new scanned range quite soon after start. I guess the slowest part of that merging would be db transaction and it might lock table for write operations (at least I suppose it does) and thus insertion (or update) of a new range will be delayed. I'm not sure that's a big deal, also not sure whether that blocking will be noticeable in the app, but it looks like we will just move it from one place to another.

cammellos: ok, so we don't wait for `startWallet` to login etc , correct?
If we don't wait, yes, running in the background is not useful

rasom: we don't wait and also soon we will run it only after user opens wallet tab so it might not happen at all :)

services/wallet/database.go
rasom: done

services/wallet/database.go
rasom: tests added

cammellos (approved):

services/wallet/database.go
```diff
@@ -790,31 +789,206 @@ func insertBlocksWithTransactions(creator statementCreator, account common.Addre
...
+	return ranges, nil
+}
+
+func getNewRanges(ranges []*BlocksRange) ([]*BlocksRange, []*BlocksRange) {
+	initValue := big.NewInt(-1)
+	prevFrom := big.NewInt(-1)
+	prevTo := big.NewInt(-1)
+	hasMergedRanges := false
+	newRanges := []*BlocksRange{}
+	deletedRanges := []*BlocksRange{}
```
Samyoul: Although this will compile fine, it is best to avoid an empty slice declaration via a literal, instead use a nil slice declaration.

See https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

```suggestion
	var newRanges []*BlocksRange
	var deletedRanges []*BlocksRange
```

Samyoul (approved):
Apart from the minor choice between an empty string verses a nil declaration, This looks good to me.

services/wallet/database.go
rasom: fixed

PR #2146, opened 2021-02-09
go-ethereum: Drop throttling of inbound connections

jakubgs:
This fix puts an end to a saga that essentially start during the Status Prague Meetup at the end of October 2018. At the time we were experiencing massive issues with `Connecting...` spinners in the app in the venue we rented. We were pulling our hairs out what to do and we could not find the cause of the issue at the time.

Three months later I deployed the following change: https://github.com/status-im/infra-eth-cluster/commit/63a13eed

Which used `iptables` to map the `443` port onto our `30504` Status node port using `PREROUTING` chain and `REDIRECT` jump in order to fix issues people have been complaining about when using WiFi networks in various venues: https://github.com/status-im/status-react/issues/6351

Our thinking when trying to resolve the reported issue assumed that some networks might block outgoing connections on non-standard ports other than the usual `80`(HTTP)/`443`(HTTPS) which would disrupt Status connectivity. While this fix could have indeed helped a few edge cases, what it really did was cause the Status node to stop seeing actual public IPs of the clients.

Buy __pure accident__ this change caused the code we inherited from `go-ethereum` implementation of DevP2P protocol to stop throttling new incoming connections, because the IP as which they appeared was a `172.16.0.0/12` network address of the Docker bridge.

The `go-ethereum` code used the `!netutil.IsLAN(remoteIP)` check to avoid throttling connections from local addresses, which included the local Docker bridge address:
https://github.com/status-im/status-go/blob/8268083008449b08f28861791c5dfeb1053d425f/vendor/github.com/ethereum/go-ethereum/p2p/netutil/net.go#L36

The fix intended to target a small number of networks with fortified firewall configuration accidentally resolved our issues with
`Connecting...` prompts that our application showed us en masse during our Prauge Meetup. Part of the reason for that is that venues like that normally give out local IP addresses and use NAT to translate them onto the only public IP address they possess.

Since out application is supposed to be usable from within networks behind NAT like airport WiFi networks for example, it makes no sense to keep the inbound connection throttle time implemented in `go-ethereum`.

Research issue: https://github.com/status-im/infra-eth-cluster/issues/35
Depends on: https://github.com/status-im/go-ethereum/pull/91

## Note 

I'm leaving `inboundThrottleTime` in because it's used to calculate value for `dialHistoryExpiration` in:
`vendor/github.com/ethereum/go-ethereum/p2p/dial.go`

I believe reducing that value once we deploy this change should also increase the speed with which the Status application is able to reconnect to a node that was temporarily unavailable, instead waiting the 5*30 seconds.


jakubgs:
But as pointed out by @cammellos , this is not a proper long term fix, since it would be overwritten if we ever decide to upgrade something under `vendor` and that would cause the fix to be overwritten. We might have to make our own fork of `go-ethereum`.

jakubgs:
Adjusted to make use of our patched `go-ethereum` fork: https://github.com/status-im/go-ethereum/pull/91

cammellos (approved):

jakubgs:
I've deployed a new image from this branch to `eth.staging` and it appears to be green so far:
https://canary.infra.status.im/service/2/

jakubgs:
I've merged https://github.com/status-im/go-ethereum/pull/91 and updated this PR to use the new `v1.9.5-status.10` tag.

jakubgs:
I've also deployed this change to one prod host yesterday `mail-03.do-ams3.eth.prod` - just for shits and giggles:
```
admin@mail-03.do-ams3.eth.prod:~ % d
CONTAINER ID   NAMES               IMAGE                                 CREATED        STATUS
b4ede113683b   statusd-mail-node   statusteam/status-go:deploy-staging   11 hours ago   Up 11 hours
0356ef1e56c7   statusd-mail-db     postgres:9.6-alpine                   11 hours ago   Up 11 hours
```
And as far as I can tell it's working just fine. No canary alerts all night. And the peers graphs looks okay:

image

jakubgs:
Though I do see a spike in delivery failures, but that happened around 4am, way after I deployed the change:

image

Samyoul (approved):

jakubgs:
I think the delivery failures might be related to the `mail-01.do-ams3.eth.test` node that didn't have the `443` port properly configured and would cause timeouts.

jakubgs:
After restart I can still see some spiking in delivery failures:

image

I will try to build an image off of `v0.69.3` plus this change to see what happens.

jakubgs:
But it is weird how only `mail-03` has this spike, specifically in `historic_msg_resp` type failures in delivery:

image

jakubgs:
We normally get some `process` type delivery failures, but the `historic_msg_resp` is new, here's 7 days:

image

Maybe it's just a coincidence. This is the code:

https://github.com/status-im/status-go/blob/220f7b15be7548a30c896cc99cb0d8cbc0be259d/mailserver/mailserver.go#L1127-L1139

jakubgs:
I'm still seeing it:

image

I don't get it. It was happening when I applied this change straight on top of `v0.69.3`. I don't get how this change can cause elevated delivery failures.

jakubgs:
And now I'm seeing `process` type failures:

image

Code:
https://github.com/status-im/status-go/blob/220f7b15be7548a30c896cc99cb0d8cbc0be259d/mailserver/mailserver.go#L789-L800

jakubgs:
Based on the logs the error that happens is `p2p: read or write on closed message pipe` but the number isn't huge:

image

https://kibana.infra.status.im/goto/d8d9668b81ee921a2f313f32e5128131

jakubgs:
I adjusted the graph to actually use `rate()` and it looks much more normal:

image

jakubgs:
To be fair other hosts have delivery failure spikes too:

image

But the one on `mail-03.do-ams3.eth.prod` is considerably bigger:

image

jakubgs:
In general I think the spikes are normal:

image

I've also deployed the staging image to `mail-03.gc-us-central1-a.eth.prod` and it seems fine so far:

image

This might be a red herring.

jakubgs:
As far as I can tell this was a coincidence because the `` host has not been showing any delivery errors for 5 hours now:

image

So I'm, going to merge this, tag it, and deploy to prod.

PR #2149, opened 2021-02-11
[wallet] Set initial blocks range for a new empty account

rasom:
For a newly created account `0 - latest` range will be set as scanned to avoid making unnecessary RPC requests. It is highly unlikely that a new account has any tx history anyway. 
Currently we execute 32 `etc_*` RPC requests when create a new account. This number will be reduced to 3 requests: one to get latest block number, one to get ETH balance and one to get SNT balance. There is no point in checking balance as well but this will be fixed on the client side. 

status: ready


services/wallet/database.go
```diff
@@ -789,6 +789,16 @@ func insertBlocksWithTransactions(creator statementCreator, account common.Addre
 	return nil
 }
 
+func (db *Database) UpsertRange(account common.Address, network uint64, from *big.Int, to *big.Int) error {
+	log.Info("upsert range", "account", account, "network id", network, "from", from, "to", to)
```
cammellos: maybe debug?

services/wallet/service.go
```diff
@@ -182,3 +183,26 @@ func mapToList(m map[common.Address]struct{}) []common.Address {
 func (s *Service) IsStarted() bool {
 	return s.started
 }
+
+func (s *Service) SetIntialBlocksRange(network uint64) error {
```
cammellos: typo, it should be `SetInitial`

cammellos (approved):
looks good, just a few minor neatpicks

services/wallet/database.go
rasom: sure

services/wallet/service.go
rasom: fixed

Samyoul (approved):
Looks good

PR #2150, opened 2021-02-12
"Delivered" status for messages sent with datasync

vkjr:
Status-go part of [this issue](https://github.com/status-im/status-react/issues/11439)

**Implementation:**
- when a message sent via datasync we remember an id of a sent message within `RawMessage` and save it in the database
- when we receive a new message from datasync we take list of datasync message ids that were acknowledged (acks), search for corresponding raw messages in database, set `OutgoingStatus` to delivered and notify client with new `message.delivered` event

**Refactoring:**
- datasync `Handle()` renamed to `UnwrapPayloadsAndAcks()` to specify purpose. Also logic of returning non-datasync message moved to `MessageProcessor`
- `HandleDatasync()` function taken out from `StatusMessage` because it doesn't work with any of struct's fields. Moved to `MessageProcessor`and renamed to `unwrapDatasyncMessage()`


protocol/common/message_processor.go
```diff
@@ -396,30 +398,57 @@ func (p *MessageProcessor) SendPublic(
 	return messageID, nil
 }
 
+// нandleDatasync tries to unwrap message as datasync one and in case of success
```
cammellos: I believe you might have renamed the function but not the comment

protocol/datasync/datasync.go
```diff
@@ -24,28 +25,35 @@ func New(node *datasyncnode.Node, transport *NodeTransport, sendingEnabled bool,
...
+// UnwrapPayloadsAndAcks tries to unwrap datasync message and return messages payloads
+// and acknowledgements for previously sent messages
+func (d *DataSync) UnwrapPayloadsAndAcks(sender *ecdsa.PublicKey, payload []byte) ([][]byte, [][]byte, error) {
 	var payloads [][]byte
+	var acks [][]byte
 	logger := d.logger.With(zap.String("site", "Handle"))
 
 	datasyncMessage, err := unwrap(payload)
 	// If it failed to decode is not a protobuf message, if it successfully decoded but body is empty, is likedly a protobuf wrapped message
-	if err != nil || !datasyncMessage.IsValid() {
```
cammellos, line 33: This I would be a bit careful with

I believe the way protobuf works is that it might throw a deserialization error in some cases, which might indicate the message needs to be passed along, that might happen in case of incompatible message types and clashes, though I have been struggling to replicate (I believe I did see this in the logs at some point).

In our case datasync uses `5001`, `5002` etc, so there should not be any clashes and should never error,  so if we are confident, that looks ok, otherwise is a bit risky as if there's any clash messages might be discarded, and it's worth considering whether to err on the side of caution and just pass the message along if error or not a valid datasync message.

protocol/migrations/sqlite/1612870480_add_datasync_id.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE raw_messages ADD COLUMN datasync_id TEXT default "";
```
cammellos: Probably I would just use bytes here, any reason to use `TEXT`? 
the easiest way is to use `[]byte` and `BLOB`, (`types.HexBytes as a type` will serialize nicely to hex, but we won't need it if it's never surfaced to the user)
Also it takes away the issue of updating raw message table, we noticed that it can be slow to update messages on slow devices, so best to leave the `UPDATE` statement below out and deal with `nullable` fields (in case you use a BLOB `[]byte` is nullable so no issue there)

cammellos: Also, probably best to create an index on the new field, so that we can efficiently query

cammellos (commented):
Nice work!

Basically it's all good, I would probably consider using `BLOB` for datasync id instead of `TEXT`, and make it nullable so we don't have to `UPDATE` as it might run slow on some devices.

The other thing I am not sure about is not passing the datasync message in case of errors, it should be safe I think, but a bit risky, I would probably be a bit more cautious and maybe leave for a later change, but up to you.

protocol/datasync/datasync.go
vkjr: @cammellos, If I correctly understand, you are not sure if it is ok that I skipped the piece of logic when we don't return message which apperead to be non-datasync.
If so, I actually processed this case but in other place. I think that datasync modules shouldn't take this decision, so moved it to processor to make more obvious:
https://github.com/status-im/status-go/pull/2150/files#diff-853e917c9371c30a01a1d22490f565efeb58ec11e15510ddf1c4f20619b22d15R451

wdyt?

p.s. hope I got you correctly..

cammellos: ah nice, much better, I missed it!

protocol/messenger_config.go
```diff
@@ -14,6 +14,10 @@ import (
 	"github.com/status-im/status-go/services/mailservers"
 )
 
+type MessengerSignalsHandler interface {
+	MessageDelivered(string, string)
+}
+
```
Samyoul: I'm curious why you use an interface here, there doesn't seem to be any other structs that implement this except for a purpose created struct. Would something like the following also suit the situation, maybe better?

```suggestion
type MessengerSignalsHandler func(string, string)
```

**EDIT**

I didn't think that my example was very clear so I've coded up what I'm thinking here https://github.com/status-im/status-go/commit/42ebc9bf1fae321f320e5a2cc8f8bd527aa0597d

Samyoul (approved):
Overall great work. I just have a small question about the use of an interface type over a func type

protocol/messenger_config.go
vkjr: @Samyoul, thanks for reviewing!
I used an interface because I thought it will be more consistent with existing `services/ext/signals.go` content. All functions there grouped logically with the help of empty structs. Since `delivered` is a first messenger's signal and probably not the last one,  same logic were used. I created empty struct and function of it.
And in `services/ext/service.go` I took `EnvelopeSignalHandler{}` as an example and used `MessengerSignalsHandler{}` similarly.

Overall this empty structs approach seemed unfamiliar to me, probably due to lack of golang experience but I think it will work well for Messenger when more signals added. And most probably they will be added when more logic moved out of `status-react`. 

Wdyt?

Samyoul: Thank you for your reasoning. I can see your approach.

Personally I think that an interface with a single struct implementation wrapping a single function is unnecessary complexity. `ext.EnvelopeEventsHandler{}` struct is similar that it has no state fields, but the `ext.EnvelopeEventsHandler{}` interface does have multiple methods and multiple implementations (it also confusingly has 4 declarations), and the `protocol.EnvelopeEventsInterceptor{}` implementation has state fields.

> And most probably they will be added when more logic moved out of status-react.

This is the best reason for keeping your current approach, if we plan on implementing multiple message signal handler functions in the near future then it makes sense to use a struct to load in all the functions at the same time.

Would the `MessengerSignalsHandler {}` interface have more than one implementation?

vkjr: >Personally I think that an interface with a single struct implementation wrapping a single function is unnecessary complexity

Totally agree)

>Would the MessengerSignalsHandler {} interface have more than one implementation?

Probably not. Maybe some can be added for testing if needed.

@cammellos, do you know if Messenger will need more signals to client when logic moved out of `status-react`? 
If there are no events we can think about right now, I'd be happy to remove unnecessary interface and use the approach suggested by @Samyoul

cammellos: mmh, tough question :)
I can't think of anything at the moment, signals tends to be pretty stable, though maybe when we move some mailserver logic we might have to create new signals, hard to say.
I don't mind either approaches here myself, so it's completely up to you two, as you prefer (I tend to favor Sam's point, but there's little complexity added and we might add more signals, so I think both are acceptable)

vkjr: Thanks! If we aren't sure - I'd also prefer to simplify the code :)

PR #2151, opened 2021-02-12
Private Profile Photos

Samyoul:
## What has changed?

**Settings**

I've added a new setting `profilePicturesShowTo` which has three options for whom should see the user's profile picture:

- `ProfilePicturesShowToContactsOnly` - shows only to the user's contacts *default*
- `ProfilePicturesShowToEveryone` - shows their profile image to everyone
- `ProfilePicturesShowToNone` - shows their profile image to no-one, even if an image is set.

I've also added functionality that does the following:

**On Sending**

- If setting `profilePicturesShowTo` is set to `ProfilePicturesShowToContactsOnly`
- Create an AES encryption key and encrypt all the user's identity image payloads
- Replace the plain payload with the encrypted payload
- Encrypt the encryption AES key using ECDH AES encryption for each of the user's contacts
- Attach all the encrypted encryption AES key payloads to the `IdentityImage` protobuf
- Send the protobuf as usual

**On Receiving**

- For every `ChatIdentity` that is received attempt to decrypt any attached `IdentityImages`
  - loop through all encrypted encryption AES key payloads
  - If a encrypted encryption AES key payload can be decrypted using an ECDH key between sender and recipient
  - attempt to decrypt the encrypted image payload
  - on success replace the encrypted payload with the plain decrypted payload
- After the decryption process any images that remain encrypted are discarded and not processed any further
- Otherwise process the images as usual

## Why make the change?

In line with offering a privacy first product images and the concerns raised in https://github.com/status-im/status/issues/56 this change was introduced to ensure a user could use profile images without sharing their image with everyone they interacted with on Status if they so choose not to.


cammellos (approved):
Looks good to me, the only question I have, what happens to old client if they receive an encrypted image?
If they drop it/don't display it, it's all fine, if they show some garbled picture then we might want to use a separate field for the encrypted image in protobuf.

Samyoul:
> Looks good to me, the only question I have, what happens to old client if they receive an encrypted image?

Excellent point. So the image payload is passed by the `HandleChatIdentity` func to `images.GetPayloadDataURI()` and would almost 100% get a `image format not supported` error which would trigger a `logger.Warn("failed to handle ChatIdentity", zap.Error(err))` and set `allMessagesProcessed` to false. Which I think is safe behaviour.

---

Although that would trigger for any image payload that is encrypted on the `ChatIdentity`. At the moment a `ChatIdentity` only has a total of 2 potential image payloads (`thumbnail` and `large`), but in the future this may not be the case (some maybe encrypted and other not ... maybe).

I've also realised that a single image parsing error would cause all of the attached images to be rejected, I've added a `TODO` to this functionality I'll address it in another PR.

Also I've realised that a single `IdentityImage` parse failure would result in the entire `ChatIdentity` parse to fail, that needs addressing also.

Samyoul:
Awaiting a successful manual test of https://github.com/status-im/status-react/pull/11768 before merge of this PR

Samyoul:
Failing Scenarios

```
{
"chat context type": "private-chat", 
"profile picture Show Settings": "ShowToEveryone", 
"profile picture View Settings": "ViewFromContactsOnly", 
"bob in Alice's Contacts": false,
"alice in Bob's Contacts": true
}
expect to receive a profile pic? {"result": true}
```

[Full Logs](https://gist.github.com/Samyoul/9bcebe7858ba9ae139811a62eb901d85)

---

```
{
"chat context type": "private-chat", 
"profile picture Show Settings": "ShowToEveryone", 
"profile picture View Settings": "ViewFromEveryone", 
"bob in Alice's Contacts": true, 
"alice in Bob's Contacts": false
}
expect to receive a profile pic?{"result": true}
```

[Full Logs](https://gist.github.com/Samyoul/b35430936277f4fbb1872a7485919b84)

---

```
{
"chat context type": "private-chat", 
"profile picture Show Settings": "ShowToContactsOnly", 
"profile picture View Settings": "ViewFromEveryone", 
"bob in Alice's Contacts": true, 
"alice in Bob's Contacts": false
}
expect to receive a profile pic?{"result": true}
```

[Full Logs](https://gist.github.com/Samyoul/21ac0f56afe40e0ab7856b4f265fb45f)

---


```
{
"chat context type": "private-chat", 
"profile picture Show Settings": "ShowToContactsOnly", 
"profile picture View Settings": "ViewFromContactsOnly", 
"bob in Alice's Contacts": true,
"alice in Bob's Contacts": true
}
expect to receive a profile pic?{"result": true}
```

[Full Logs](https://gist.github.com/Samyoul/7fd78c45f9fa3b5e72a299de8a33d0c6)

---

protocol/messenger.go
```diff
@@ -600,14 +608,24 @@ func (m *Messenger) publishContactCode() error {
 		m.logger.Error("could not build contact code advertisement", zap.Error(err))
 	}
 
+	m.logger.Debug("initing cca if not set")
```
cammellos: You might want to remove this logs

cammellos (approved):
looks good, just some extra logs

protocol/messenger.go
Samyoul: Ah, I thought I'd got them all.

PR #2153, opened 2021-02-15
Fix image messages being dropped in profile chats

cammellos:
It looks like profile chats are created as public chats, and the code
will drop messages for it.

This commit fixes the issue by checking for the "@" prefix in profile
chat names, until we fix the issue by migrating those chats.

we should probably move the creation code to status-go to avoid these kind of issues, or create a dedicated endpoint.


protocol/chat.go
```diff
@@ -93,7 +94,9 @@ func (c *Chat) PublicKey() (*ecdsa.PublicKey, error) {
 }
 
 func (c *Chat) Public() bool {
-	return c.ChatType == ChatTypePublic
+	// It looks like status-react created profile chats as public chats
+	// so for now we need to check for the presence of "@" in their chatID
+	return c.ChatType == ChatTypePublic && !strings.HasPrefix(c.ID, "@")
```
flexsurfer: im not sure this will work, because updates should still be treated as public chats , so probbaly better to move this condition outside where we drop images

cammellos: good point, pushed a fix, thanks

flexsurfer (approved):

PR #2154, opened 2021-02-16
[wallet] Avoid some unnecessary RPC requests

rasom:
Three optimisations useful for an account with no history.
- avoid making RPC request for `zero - zero` range
- avoid checking of nonce for a lower block in the range if it is zero in a higher block 
- on `wallet_getTransfersByAddress` scanning of history is skipped if zero block is already reached
- no need to fetch block num before fetching token balances

status: ready

### DIFF
--- a/services/wallet/commands.go
+++ b/services/wallet/commands.go
@@ -41,7 +41,7 @@ func (c *ethHistoricalCommand) Command() Command {
 func (c *ethHistoricalCommand) Run(ctx context.Context) (err error) {
 	start := time.Now()
 	totalRequests, cacheHits := c.balanceCache.getStats(c.address)
-	log.Info("balance cache before checking range", "total", totalRequests, "cached", totalRequests-cacheHits)
+	log.Info("balance cache before checking range", "total", totalRequests, "cached", totalRequests-cacheHits, "from", c.from, "to", c.to)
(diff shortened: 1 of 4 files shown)


services/wallet/commands.go
```diff
@@ -648,6 +648,10 @@ func findFirstRange(c context.Context, account common.Address, initialTo *big.In
 	to := initialTo
 	goal := uint64(20)
 
+	if from.Cmp(to) == 0 {
+		return to, nil
+	}
+
 	firstNonce, err := client.NonceAt(c, account, to)
```
rasom, line 655: without new `if` block this request was executed although there is no point in in checking `0 - 0` range

services/wallet/concurrent.go
```diff
@@ -111,11 +111,16 @@ func checkRanges(parent context.Context, client reactorClient, cache BalanceCach
 			if lb.Cmp(hb) == 0 {
 				log.Debug("balances are equal", "from", from, "to", to)
 
-				ln, err := client.NonceAt(ctx, account, from)
+				hn, err := client.NonceAt(ctx, account, to)
 				if err != nil {
 					return err
 				}
-				hn, err := client.NonceAt(ctx, account, to)
+				if hn == 0 {
```
rasom, line 118: If nonce is zero at higher block then no sense in checking lower block's nonce

cammellos (approved):

PR #2142, opened 2021-02-03
Adding Transactions messages to the localnotifications

Samyoul:
## What's Changed?

I've added functionality that creates `MessageNotificationBody`s for validated transaction messages and adds the notifications to the `localnotification` queue to be sent to the OS.

Also I've moved the original functionality that created `MessageNotificationBody`s from new chat messages to the bottom of the `handleRetrievedMessages` function so that we only notify about new messages and are not dependant on specific handlers for adding `MessageNotificationBody` to `Response.Notifications` list.

Closes #2141

## Note

**DO NOT MERGE** until https://github.com/status-im/status-go/pull/2130 has been merged.

PR #2157, opened 2021-02-22
Filter message PNs on status go side

rasom:
status: ready


services/local-notifications/core.go
```diff
@@ -474,16 +476,33 @@ func (s *Service) IsStarted() bool {
 	return s.started
 }
 
-func SendMessageNotifications(mnb []protocol.MessageNotificationBody) {
+func showNotification(publicKey ecdsa.PublicKey, n protocol.MessageNotificationBody) bool {
```
cammellos: Personally I would move this logic closer to `messenger.go`,
as it's very contextual (it will deal with `chats`, `contacts`, `communities` etc).
But up to you, either ways it's ok.
Could you please add a test for this functionality?

Samyoul: I have some code in https://github.com/status-im/status-go/pull/2142 that will give this code a better home.

We could either add it below `message.New = true` :

https://github.com/status-im/status-go/blob/f9d08bf9c272022f9e4a06082aa8acf9c17beb85/protocol/messenger.go#L3259-L3268

Or inside :

https://github.com/status-im/status-go/blob/f9d08bf9c272022f9e4a06082aa8acf9c17beb85/protocol/messenger.go#L2756-L2779

rasom: code is moved

protocol/messenger.go
```diff
@@ -2807,14 +2807,23 @@ func (r *ReceivedMessageState) addNewMessageNotification(m *common.Message) erro
...
 
-	r.Response.Notifications = append(
-		r.Response.Notifications,
-		MessageNotificationBody{
-			Message: m,
-			Contact: r.AllContacts[contactID],
-			Chat:    r.AllChats[m.ChatId],
-		},
-	)
+	chat := r.AllChats[m.ChatId]
```
cammellos: you should use straight away `localChatID`, not `ChatId` :)

protocol/messenger.go
```diff
@@ -2917,7 +2926,7 @@ func (m *Messenger) handleRetrievedMessages(chatWithMessages map[transport.Filte
 							allMessagesProcessed = false
 							continue
 						}
-
+						// TODO(rasom)
```
cammellos: TODO?

protocol/messenger.go
```diff
@@ -3318,6 +3331,30 @@ func (m *Messenger) handleRetrievedMessages(chatWithMessages map[transport.Filte
 	return messageState.Response, nil
 }
 
+func showNotification(publicKey ecdsa.PublicKey, n MessageNotificationBody, responseTo *common.Message) bool {
+	if n.Chat != nil && n.Chat.ChatType == ChatTypeOneToOne {
+		return true
+	}
+
+	publicKeyString := common.PubkeyToHex(&publicKey)
+	mentioned := false
```
cammellos, line 3336: same as `var mentioned bool` btw, either are fine

cammellos (approved):

protocol/messenger.go
rasom: fixed

protocol/messenger.go
rasom: fixed

PR #2158, opened 2021-02-26
feat: introduce new `joined` property in `Chat` struct

0x-r4bbit:
This property is useful for clients to know when a channel or chat
was joined so they can use that to calculate the order of channels
and chats shown in applications.

Changes include a new `joined` property on the `Chat` struct,
as well as adjustments in the persistence layer to retreive and
update chat data in the database.

In addition there's a migration script that alters the existing
chat table to introduce a new column for the `joined` field.

Important changes:

Closes #


0x-r4bbit:
>  Have you updated the documentation, if impacted (e.g. docs.status.im)?

To be fair, that link gives a 404

Samyoul (commented):
Hey @PascalPrecht thanks for the PR.

Where the time for the `Joined` field is set? At the moment this implementation seems to only ever save a `0` value.

0x-r4bbit:
Because "joining a chat" is essentially "saveChat" with certain parameters, it's the client's job to send a `joined` value along with it. So in other words, in Status Desktop for example, this is happening in our `join()` API: https://github.com/status-im/status-desktop/blob/d4bf29085c7aafbc0aaf72b6b6b748ce010451c5/src/status/chat.nim#L100-L101

Samyoul:
Ah, I see.

Samyoul (approved):

cammellos (changes requested):
I think with `joined` you need to `Scan` a `NullInt`, as it will be `null` for existing databases (setting `DEFAULT` to `0` will only work on new rows).
So it will probably throw an error on existing databases post-migration.

You can either use `sql.NullInt` or add a statement to the migration `UPDATE chats SET joined = 0` so that all the existing chats have a value, in that case the code does not have to be changed.

Could you also rebase to the latest develop? (make sure that your migration is the latest, otherwise it might be skipped), and bum the VERSION please?

Thanks!

0x-r4bbit:
@cammellos I've updated the commit. re: the `VERSION`: Since what I did is rather a feature, I've incremented the `minor` version. Let me know if you aren't following SemVer and prefer a patch udpate.

cammellos (approved):

0x-r4bbit:
Failing tests seem to be unrelated to this commit

cammellos:
some tests are flaky, you can re run them frim Jenkins, i am afk, but
@samyoul can assist

0x-r4bbit:
Re-ran the tests on jenkins, still failing due to missing test file for `shhClient`
Also, looks like I need to rebase this one once more

cammellos:
ok, I will rebase it and merge it once ready

cammellos:
@PascalPrecht I have rebased it on feat/add-joined-prop-to-chat-2 if you want to copy over the branch (I can't push on this one as it's on your repo)

0x-r4bbit:
@cammellos Done!

But it **seems** your branch needs rebase as well. Will do it now.. :D

PR #2160, opened 2021-03-01
Use json.Marshaler for notifications body

cammellos:
This commit introduces the following changes:

- `local-notifications` require as body an interface complying with
`json.Marshaler`
- removed unmarshaling of `Notifications` as not used (we only Marshal
notifications)
- `protocol/messenger.go` creates directly a `Notification` instead of
having an intermediate format
- moves `transaction` code to its own file in transactions

This is needed to make it easier to create non-message notifications (for communities)


rasom (approved):

services/local-notifications/types.go
```diff
@@ -0,0 +1,10 @@
+package localnotifications
+
+const (
+	CategoryTransaction            PushCategory = "transaction"
+	CategoryMessage                PushCategory = "newMessage"
+	CategoryCommunityRequestToJoin              = "communityRequestToJoin"
+
+	TypeTransaction NotificationType = "transaction"
+	TypeMessage     NotificationType = "message"
+)
```
Samyoul, line 10: I like this, breakout the types of notification. One thing I notice is that `CategoryCommunityRequestToJoin              = "communityRequestToJoin"` doesn't have a type, it is just a string, is this deliberate?

protocol/messenger_response.go
```diff
@@ -219,6 +230,24 @@ func (r *MessengerResponse) AddChats(chats []*Chat) {
...
+func (r *MessengerResponse) ClearNotifications() {
+	r.notifications = nil
+}
+
+func (r *MessengerResponse) AddNotifications(notifications []*localnotifications.Notification) {
+	for _, c := range notifications {
+		r.AddNotification(c)
+	}
+}
+
```
Samyoul, line 250: 👍 these are needed now

protocol/local_notifications.go
```diff
@@ -1,9 +1,111 @@
...
+
+func NewMessageNotification(id string, message *common.Message, chat *Chat, contact *Contact) *localnotifications.Notification {
+	body := &NotificationBody{
+		Message: message,
+		Chat:    chat,
+		Contact: contact,
+	}
+
+	return body.toMessageNotification(id)
+}
```
Samyoul: Would `id` ever be different from the `common.Message.ID`?

protocol/contact.go
```diff
@@ -26,6 +26,22 @@ type ContactDeviceInfo struct {
...
+	if c.ENSVerified {
+		return c.Name
+	}
+
+	return c.Alias
+}
+
+func (c *Contact) CanonicalImage() string {
+	return c.Identicon
+}
```
Samyoul, line 43: Would a profile picture be considered a CanonicalImage? Or are we keeping them private?

signal/events_shhext.go
```diff
@@ -158,6 +157,6 @@ func SendWhisperFilterAdded(filters []*Filter) {
 	send(EventWhisperFilterAdded, WhisperFilterAddedSignal{Filters: filters})
 }
 
-func SendNewMessages(response *statusproto.MessengerResponse) {
-	send(EventNewMessages, response)
+func SendNewMessages(obj json.Marshaler) {
+	send(EventNewMessages, obj)
 }
```
Samyoul, line 162: Can any `json.Marshaler` be sent, or only a specific subset of `json.Marshaler`?

Samyoul (approved):
Great. Just a few implementation questions

services/local-notifications/types.go
cammellos: nope, missed it, thanks :)

protocol/contact.go
cammellos: yes, I haven't fully implemented for now as I can't get them to display in push notification for some reason, but eventually the profile picture would be the canonical image (if present)

signal/events_shhext.go
cammellos: any could be sent, as long as it explicitly implement `MarshalJSON`

Samyoul: My question was more, *should* the `obj` be limited to a subset of `json.Marshaler` incase we wanted to control what exactly could be sent here. Something like:

```go
type SendAbleEventMessage interface {
  json.Marshaler
  SomeOtherMethodUniqueToThisTypeOfStruct()
}
```

But on thinking about it maybe `SendNewMessages()` should be very general and can received anything, and we control what is sent upstream. I may be over thinking this one.

cammellos: Ah, I think it's fine for now, we only send `MessengerResponse` and I don't think we will extend it much, the only reason it's made generic is that otherwise I was getting into issues with circular dependencies :)

PR #2161, opened 2021-03-01
[wallet] Store account's last known balance and nonce

rasom:
On a regular wallet account with some tx history it will save 6 requests per hour (15 requests per hour vs 21 requests per hour). Not a big deal, but on larger scale it might be noticeable (for 10k accounts it will save `6 * 24 * 10k = ~1.4m` requests per day) 

status: ready


appdatabase/migrations/sql/0019_blocks_ranges_extra_data.down.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE blocks_ranges DROP COLUMN balance;
```
cammellos: that's not supported on sqlite, so I would just get rid of the `down` migration. 
Otherwise if you are really keen on having one you would have to create a temp table, copy over the data without the column, drop the initial one and move it over. Probably not worth as we don't run `down` migrations anyway.

services/wallet/commands.go
```diff
@@ -1091,8 +1094,14 @@ func (c *findAndCheckBlockRangeCommand) Run(parent context.Context) (err error)
 			}
 		}
 
-		log.Debug("saving headers", "len", len(uniqHeaders), "address")
-		err = c.db.ProcessBlocks(address, newFromByAddress[address], c.toByAddress[address], uniqHeaders)
+		lastBlockNumber := c.toByAddress[address]
+		log.Info("saving headers", "len", len(uniqHeaders), "lastBlockNumber", lastBlockNumber, "balance", c.balanceCache.ReadCachedBalance(address, lastBlockNumber), "nonce", c.balanceCache.ReadCachedNonce(address, lastBlockNumber))
```
cammellos: Do you want to keep this as info? (just making sure it's intentional)

cammellos (approved):
Looks good, just a couple of minor points

appdatabase/migrations/sql/0019_blocks_ranges_extra_data.down.sql
rasom: > Otherwise if you are really keen on having one

not really :D

I will remove it, thanks!

services/wallet/commands.go
rasom: No it is not, will revert to debug, thanks

rasom:
@cammellos @Samyoul 
I had to add `SQLBigIntBytes` type which allows to store big.Int as BLOB in the database. Otherwise balance can easily be bigger than int64 and that caused issues with e2e tests. Let me know if there is a better name for it than `SQLBigIntBytes` or it should be handled differently.

cammellos (approved):

PR #2162, opened 2021-03-01
[wallet] Nonce range cache

rasom:
In a nutshell: if we know that nonce=5 at block 10 and that nonce=5 at block 20 the means we also know that for all blocks between 10 and 20 nonce=5. 

On average it saves 10-12% of `eth_getTransactionCount` requests (50-60 per run) on fetching tx history. This number can vary depending on number of incoming transactions in a specific range (more incoming transactions means it is more likely nonce will hit this "cache")

status: ready


cammellos (approved):

Samyoul (approved):
Looks good

PR #2168, opened 2021-03-08
[wallet] RPC usage stats

rasom:
status: ready


cammellos (approved):

services/rpcstats/service.go
```diff
@@ -0,0 +1,44 @@
...
+// New returns a new Service.
+func New() *Service {
+	return &Service{}
+}
+
+// APIs returns a list of new APIs.
+func (s *Service) APIs() []rpc.API {
+	return []rpc.API{
+		{
+			Namespace: "status",
```
Samyoul: I think that an API called `status` in an app called **Status** should be something very central to the API. Would you consider `rpcstats` as an alternative name for the rpc stats api namespace?

services/wallet/service.go
```diff
@@ -34,7 +34,7 @@ type Service struct {
 	db                  *Database
 	reactor             *Reactor
 	signals             *SignalsTransmitter
-	client              *ethclient.Client
+	client              reactorClient
```
Samyoul: I notice that the previous implementation used a pointer to an `*ethclient.Client`, did you deliberately chose to not have a struct pointer for your reactorClient? This comment also relates to the other cases in this PR.

Samyoul (approved):
Great! No major issues just a question about pointers and API namespace naming

services/rpcstats/service.go
rasom: Changed to `rpcstats` and so we have two methods `rpcstats_getStats` and `rpcstats_reset`.

services/wallet/service.go
rasom: `reactorClient` was an interface, not a struct and thus pointer wouldn't make sense there as I understand, right? But it was useless there and I replaced it in the last commit with `*walletClient` which is a struct pointer.

services/rpcstats/stats.go
```diff
@@ -0,0 +1,48 @@
...
+
+type RPCUsageStats struct {
+	total            uint
+	counterPerMethod map[string]uint
+	rw               sync.RWMutex
+}
+
+var stats *RPCUsageStats
+
+func getInstace() *RPCUsageStats {
```
vkjr: small misprint

rasom: danke

PR #2169, opened 2021-03-11
Expand confirmations to private group chat messages

cammellos:
This commit expands the confirmation mechanism to allow private group
chat messages to be confirmed, both in a `everyone confirmed` or `at least one confirmed` mode:

Changes:

- Added a separate table for message confirmations as group chat
messages have same messageID but multiple datasyncID
- Removed DataSyncID from raw message (I haven't removed the column name
as it can't be done in sqlite without copying over the table)


protocol/common/raw_messages_persistence.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+
+	return ids, nil
+}
+
+// MarkAsConfirmed marks all the messages with dataSyncID as confirmed and returns
+// the messageIDs that can be considered confirmed.
+// If atLeastOne is set it will return messageid if at least once of the messages
+// sent has been confirmed
+func (db RawMessagesPersistence) MarkAsConfirmed(dataSyncID []byte, atLeastOne bool) (messageIDs []types.HexBytes, err error) {
```
vkjr: @cammellos I'm wondering if it is possible to get more than one messageID as a result of this function?

My understanding so far was that we can have same messageID with different related datasyncIDs (private group case), but we can't have different messageIDs related to the same datasyncID. And that means we won't get more than one message confirmed after receiving ack-ed dataSyncID. So logic of this function can be simpler.

Please, correct me if I'm wrong

protocol/common/raw_messages_persistence.go
```diff
@@ -0,0 +1,262 @@
...
+		}
+		ids = append(ids, id)
+	}
+
+	return ids, nil
+}
+
+// MarkAsConfirmed marks all the messages with dataSyncID as confirmed and returns
+// the messageIDs that can be considered confirmed.
+// If atLeastOne is set it will return messageid if at least once of the messages
```
vkjr, line 184: misprint: once -> one

protocol/common/raw_messages_persistence.go
cammellos: Oh you are right,
I got confused by the fact that messages can be batched in a single datasync message, but `AppendMessage` returns the ID of the "inner" datasync message.
We can probably simplify, well spotted!

protocol/common/message_processor.go
```diff
@@ -262,7 +264,19 @@ func (p *MessageProcessor) sendPrivate(
...
-		rawMessage.DataSyncID = datasyncID
+		// We don't need to receive confirmations from our own devices
+		if !IsPubKeyEqual(recipient, &p.identity.PublicKey) {
+			confirmation := &RawMessageConfirmation{
+				DataSyncID: datasyncID,
+				MessageID:  messageID,
+				PublicKey:  crypto.CompressPubkey(recipient),
+			}
+
+			err = p.persistence.InsertConfirmation(confirmation)
```
vkjr: I'd suggest to rename this function because it gives false feeling that we already got confirmation. 
Maybe it can be named like `InsertAwaitedConfirmation()`

To be honest, I think that probably even `raw_message_confirmations` isn't ideal naming for database table, because what we really store - is info about datasync messages - their ids, confirmation times, correspondence to messageids. And info about confirmations - is what we deduct from it using additional external info (`atLestOne` flag). But I don't have a good suggestion on how to change this, so not going to be pesky about 😁

vkjr (approved):

protocol/common/raw_messages_persistence.go
```diff
@@ -205,50 +204,39 @@ func (db RawMessagesPersistence) MarkAsConfirmed(dataSyncID []byte, atLeastOne b
 	}
 
 	// Select any tuple that has a message_id with a datasync_id = ? and that has just been confirmed
-	rows, err := tx.Query(`SELECT message_id,confirmed_at FROM raw_message_confirmations WHERE message_id IN (SELECT message_id FROM raw_message_confirmations WHERE datasync_id = ?)`, dataSyncID)
+	rows, err := tx.Query(`SELECT message_id,confirmed_at FROM raw_message_confirmations WHERE message_id = (SELECT message_id FROM raw_message_confirmations WHERE datasync_id = ? LIMIT 1)`, dataSyncID)
```
vkjr, line 207: Can we simplify this to:
`SELECT message_id,confirmed_at FROM raw_message_confirmations WHERE datasync_id = ?`
?

cammellos: I don't think that would work, essentially if you take a group chat message we will have 
3 identical `messageID` and 3 different `datasyncID`.
We need to check that all 3 have been confirmed (depending on `atLeastOne`, but suppose `atLeatOne` is `false` for now), basically make sure that all of them have been confirmed.

vkjr: Ah, I got the idea, thanks for explanation!

protocol/datasync/transport.go
```diff
@@ -165,6 +166,8 @@ func splitPayloadInBatches(payload *protobuf.Payload, maxSizeBytes int) []*proto
 
 // CalculateSendTime calculates the next epoch
 // at which a message should be sent.
+// We randomize it a bit so that not all messages are sent on the same epoch
 func CalculateSendTime(count uint64, time int64) int64 {
-	return time + int64(uint64(math.Exp2(float64(count-1)))*backoffInterval*offsetToSecond)
+	return time + int64(uint64(math.Exp2(float64(count-1)))*backoffInterval*offsetToSecond) + int64(rand.Intn(30)) // nolint: gosec
+
```
Samyoul, line 172: Just a little point about `math/rand`, if you want the result to be random you need to set a source of randomness, otherwise your results will always be the same.

See https://gobyexample.com/random-numbers

Samyoul (approved):
Nice work. ~~Just a little note about `math/rand`~~ Does not apply in this context.

PR #2170, opened 2021-03-15
🎭 📊 Anonymous Metrics V0

shivekkhurana:
### Overview
This PR lays the foundation for tracking our user behaviour in an opt-in fashion:
- data is collected and stored locally (in cache or encrypted database)
- any personally identifiable information is not collected
- all data stays offline unless the user explicitly allows us to transfer (the transfer will take place over Waku, and is not a part of this PR)
- any metric being saved to the DB **has to match a pre-defined shape**, this ensures that we do not collect sensitive data by mistake (nice error messages are raised if the developer mistakenly captures extra data)

### Design
There are three layers to the system:
- Buffer layer
- Validation layer and 
- Write to db layer

A client can send data that adheres to predefined shapes to the Go backend. This data is saved to a buffered channel. If the number of data points in the buffer reaches 8 or if the service is stopped,  the data is flushed to local encrypted SQLite database. 

In later stages, a system will take this data and post it to our servers over Waku (after opt-in). There is also a method to view all recored data.

### Audit
The feature is designed with keeping audit-ability in mind. All the data being collected can be checked by just reading one fileL `status-go/appmetrics/validators.go`. Validators are JSON Schema Specs that define the shape of data being collected. Anything that doesn't adheres to this shape is rejected before write to db.


_examples/waku.json
```diff
@@ -1,5 +1,5 @@
 {
-	"APIModules": "waku",
+	"APIModules": "waku,appmetrics",
```
cammellos: if you don't need this change we can revert

appdatabase/migrations/sql/0019_metrics.up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS app_metrics (
+       id INTEGER PRIMARY KEY AUTOINCREMENT,
```
cammellos: If you don't need an autoincrement field is best not to use one for perf reasons https://stackoverflow.com/questions/52700893/sqlite-best-practices-about-using-autoincrement  
You can use the implicit `rowid` if you need a unique identified, unless you really need an autoincrement field.

appmetrics/database.go
```diff
@@ -0,0 +1,161 @@
...
+		_ = tx.Rollback()
+	}()
+
+	insert, err = tx.Prepare("INSERT INTO app_metrics (event, value, app_version, operating_system) VALUES (?, ?, ?, ?)")
+	if err != nil {
+		return err
+	}
+
+	for _, metric := range appMetrics {
+		_, err := insert.Exec(metric.Event, metric.Value, metric.AppVersion, metric.OS)
```
cammellos: I think here you might be shadowing `err`, and what's going to happen is that fi `err` on 135 is not nil, it will still try to commit the transaction, because `err` (the one declared at the beginning of the function and in the `defer` block is nil).

So you likely want to just do:
`_, err = insert.Exec...`

appmetrics/database.go
```diff
@@ -0,0 +1,161 @@
...
+
+		if err != nil {
+			return err
```
cammellos: generally if you use named parameters, it's clear if you just always use `return` without explicitly `return err`, the linter will be able to catch non-used errors like the one above.

appmetrics/database.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		return appMetrics, err
+	}
+	defer rows.Close()
+	for rows.Next() {
+		metric := AppMetric{}
+		err := rows.Scan(&metric.Event, &metric.Value, &metric.AppVersion, &metric.OS)
+
+		if err != nil {
+			return appMetrics, err
```
cammellos: Better to return nil in this case, as it's a reference and it's more idiomatic

appmetrics/database.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+	}
+	return
+}
+
+func (db *Database) GetAppMetrics(limit int, offset int) (appMetrics []AppMetric, err error) {
+	rows, err := db.db.Query("SELECT event, value, app_version, operating_system FROM app_metrics LIMIT ? OFFSET ?", limit, offset)
+
+	if err != nil {
+		return appMetrics, err
```
cammellos: Better to return nil in this case, as array are refs and it's more idiomatic

services/appmetrics/api.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func (api *API) ValidateAppMetrics(ctx context.Context, appMetrics []appmetrics.AppMetric) error {
+	log.Info("[AppMetricsAPI::ValidateAppMetrics]")
+	return api.db.ValidateAppMetrics(appMetrics)
+}
+
+func (api *API) SaveAppMetrics(ctx context.Context, appMetrics []appmetrics.AppMetric) error {
+	log.Info("[AppMetricsAPI::SaveAppMetrics]")
+	chanFull := len(api.metricsBufferedChan) == cap(api.metricsBufferedChan)
```
cammellos: Honestly I would not bother having a buffer here, I doubt it's going to impact performance and it's probably just added complexity, this will be run on a separate thread so won't impact the UI thread or the JS thread.

services/appmetrics/api_test.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func TestValidateAppMetrics(t *testing.T) {
+	db, close := setupTestDB(t)
+	defer close()
+	api := NewAPI(db, make(chan appmetrics.AppMetric, 8))
+
+	validMetrics := []appmetrics.AppMetric{appmetrics.AppMetric{
+		Event:      "go/test1",
+		Value:      json.RawMessage(`"some-string"`),
```
cammellos: Do you have a link to the client side code? I would like to check if `RawMessage` is the right type here, it makes sense if you are storing raw json in a field (i.e a `[]byte` that is already some encoded json), but from the example above, it seems you are storing plain strings

cammellos (changes requested):
Overall looks good, well done!
Left some questions/comments

appdatabase/migrations/sql/0019_metrics.up.sql
shivekkhurana: SQLITE `rowids` are recycled when data is deleted from the table. And we will, at some point delete data from this table. 

I don't know if it matters though. Can give it more thought. Leaving this open for now.

services/appmetrics/api_test.go
shivekkhurana: The field is flexible enough to store any JSON shape. I used strings here for testing purposes.

services/appmetrics/api.go
shivekkhurana: So we just write to DB each time an event occurs? 

We don't really know the frequency of these events, but we know that this frequency will grow over time.

services/appmetrics/api_test.go
cammellos: Yes, that's clear, but the question is, do you need `RawMessage` at all? Currently, as far as I can see, you are not using any of the features of it, you are getting a `RawMessage`, but converting it back to a `string`, so it makes more sense (unless I am missing something), to use a `string` in this case.
If you are just sending some stringified JSON, I don't think you need to use `RawMessage`, but I'd have to see the client side code to make sure.

services/appmetrics/api.go
cammellos: I'd say so, unless write are multiple times per second (4/5) I don't think it will have an impact performance wise, the RPC call in the UI thread and handling of serialization is likely more expensive, so any batching is best done on the client if performance are a concern (I'd have to time it, but we know UI/Cljoure is much slower than go code, so you'll be likely hitting your bottleneck on the client before having issue with db writes)

appdatabase/migrations/sql/0019_metrics.up.sql
shivekkhurana: Okay, I looked into it. There is no benefit in uniquely identifying a metric. Dropping `AUTOINCREMENT`.

appmetrics/database.go
shivekkhurana: Do you mean `return nil, err`?

services/appmetrics/api.go
shivekkhurana: The way we have built this, might lead to 2-5 calls at once, followed by a few seconds of inactivity. 

Since we use this in conjunction with Re-frame event loop, we can easily be in a situation where we decide to capture 3 events, which get fired almost immediately. I'm trying to think of an example. 

One of the reasons why I pushed batching to Go side is because it can be re-used by Desktop team.

services/appmetrics/api_test.go
shivekkhurana: It's a raw message because it could be a string or a json object. If we set it to a string and send the value `{"hello": "world"}`, the function will reject it. 

The problem is that a JS object cannot include stringified JSON (I mentioned this on the call). If it's a string, and try to send stringified JSON, it will throw up (it was actually a string in the first iteration).

appmetrics/database.go
shivekkhurana: I tried using `return` but the linter started complaining that `err` is being shadowed. This is what I have done instead, is it fine?

Screenshot 2021-03-16 at 4 55 03 PM

appmetrics/database.go
shivekkhurana: Not sure if this is what you suggested, but if I get rid of the explicit `return`, the linter complains that the `err` is defined but never used:

Screenshot 2021-03-16 at 4 58 16 PM

services/appmetrics/api_test.go
shivekkhurana: This is the call I used to check the integration:
```clojure
(fx/defn check
  {:events [::check]}
  [_]
  {::json-rpc/call [{:method "appmetrics_validateAppMetrics"
                     :params [[{:event "go/test1"
                                :val "some-string" ;; could also be a map like {:a "b" :c {:d "e"}}
                                :app_version "1.12"
                                :os "android"}]]
                     :on-success prn
                     :on-error prn}]})
```

appmetrics/database.go
cammellos: yes, you also need to remove the initialization,
`:=` should be `=`, otherwise you will be shadowing `err`

appmetrics/database.go
cammellos: no, that would give you the same issue, `err` will be `nil` in the `defer` (the one declared at the beginning) and it will try to commit the transaction, you just have to not shadow `err` and re-use it instead:

`_, err = insert.Exec` (notice that is `=` and not `:=`, which is the issue in your code)

appmetrics/database.go
cammellos: yes

services/appmetrics/api.go
cammellos: Probably you want to batch those calls in status-react if you are making them, I would just "debounce" them, and have this endpoint rather receive an array and insert as soon as it receives them.
We have seen this before, basically the bottleneck it's status-react/the bridge, if you make too many calls, status-go will never be hit hard enough (provided that the workload is small on status-go of course, like in this case), but the bridge will choke

services/appmetrics/api_test.go
cammellos: Oh I see, so you want to have a variable type there, I thought you were always sending stringified JSON or a string.

Ok, so if you are always sending stringified JSON, you can almost certainly use a string, you client code should be:

`:val (if (string? a) a (js/JSON.stringify a))`

That will not throw an error in status-go (my understanding is that you will not parse it)

If though you actually want to send raw-object, then yes, you need to use `RawMessage` to prevent go to parse it.

shivekkhurana: okay, let me try this out.

Samyoul:
I love the emojis in your PR names 💥 🥇

appmetrics/validators.go
```diff
@@ -0,0 +1,27 @@
...
+	"properties": map[string]interface{}{
+		"view_id": map[string]interface{}{
+			"type":      "string",
+			"maxLength": 16,
+		},
+		"params": map[string]interface{}{
+			"type": "object",
+			"properties": map[string]interface{}{
+				"screen": map[string]interface{}{
+					"enum": []string{"allowed-screen-name"},
```
Samyoul, line 14: Should this be a list of allowed screen names for use in production?

services/appmetrics/api_test.go
shivekkhurana: - `string` instead of `json.RawMessage` works:

Screenshot 2021-03-17 at 12 59 03 PM

- it didn't work the last time I tried because I was adding newline characters in stringified JSON and Insomnia probably did some magic before sending the request
- I can change it back to string, but is there any extra cost of using `json.RawMessage`? Because sending pure json makes it easy to test via Insomnia (compared to strings, which are fragile and hard to manage)

Samyoul (approved):
Really great! Just a question on the list of enums

services/appmetrics/api_test.go
cammellos: It's a bit less obvious what's happening when using `RawMessage` but I don't mind, maybe leave a comment on the field so it's clear (i.e we use `RawMessage` so that we can send any type of data and not just strings)

shivekkhurana: I'll add a comment.

appmetrics/validators.go
shivekkhurana: No, getting rid of this.

shivekkhurana: Actually, if I remove this, the tests will fail. We need at least one element in the list to run tests.

appmetrics/database.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+type AppMetricValidationError struct {
+	Metric AppMetric
+	Errors []gojsonschema.ResultError
+}
+
+const (
+	// Events for testing the system
+	TestEvent1 AppMetricEventType = "go/test1"
```
cammellos: It's a bit odd to have testing code not in `_test.go` file, can we avoid it and move it to the `_test.go` file?
If they are not used other than in test, is there value in testing with them? I'd probably test using the `NavigateToCofx` event type or use a schema specifically for testing (you'd have to pass it to `Validate` function in this case)
@Samyoul what's idiomatic in this cases?

services/appmetrics/api.go
```diff
@@ -0,0 +1,31 @@
...
+func NewAPI(db *appmetrics.Database) *API {
+	return &API{db: db}
+}
+
+type API struct {
+	db *appmetrics.Database
+}
+
+func (api *API) ValidateAppMetrics(ctx context.Context, appMetrics []appmetrics.AppMetric) error {
+	log.Info("[AppMetricsAPI::ValidateAppMetrics]")
```
cammellos: maybe `Debug`?

appdatabase/migrations/sql/0020_metrics.up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS app_metrics (
+       id INTEGER PRIMARY KEY,
```
cammellos: I would probably leave a comment here as we are aliasing `rowid` and it might not be immediately clear if you don't know the behavior of sqlite (or we can just drop it and use rowid, but that's up to you).

appmetrics/database.go
```diff
@@ -0,0 +1,159 @@
...
+		_ = tx.Rollback()
+	}()
+
+	insert, err = tx.Prepare("INSERT INTO app_metrics (event, value, app_version, operating_system) VALUES (?, ?, ?, ?)")
+	if err != nil {
+		return err
+	}
+
+	for _, metric := range appMetrics {
+		_, err = insert.Exec(metric.Event, metric.Value, metric.AppVersion, metric.OS)
```
cammellos, line 131: not checking for error here, which might result in the first event erroring and the next overriding the error

cammellos (commented):
Some minor comments,
I see tests are failing with 
```
[2021-03-17T08:27:09.774Z]         	            	unable to parse file 0019_metrics.up.sql
```

On my machine `make generate` fixes the issue, have you tried? (I can run it and push if it does not work on your machine)

appdatabase/migrations/sql/0020_metrics.up.sql
shivekkhurana: We can drop it. The id is not important.

appmetrics/database.go
shivekkhurana: Added a check for err.

appmetrics/database.go
shivekkhurana: There is no value in custom test events. I added them in the first place because I wanted to test JSON Schema. These should go.

shivekkhurana:
I did `make generate` again before pushing this last commit. Observing the tests now.

cammellos (approved):
Looks good to me, great work! last thing, don't forget to bump `VERSION`!

PR #2173, opened 2021-03-18
📖 Add appmetrics overview doc

shivekkhurana:


Samyoul (approved):
Cool @shivekkhurana. There's a bit of a debate about where exactly these kinds of docs should live.

There are two main places that they should appear. One is my opinion and the other is a matter of president.

- A README.md in the feature package (my opinion), but that is the best place to put documentation about a specific feature. Personally I believe we should also have an index doc in `docs/` or even in the root README.md that can have links to the other existing docs in the repo.
- An entry in https://github.com/status-im/specs , which is where the inner workings of the status app is described ... roughly.

@cammellos I'd welcome your input on this.

shivekkhurana:
@Samyoul I pressed the merge button without realising that you had left a comment, sorry!! I can revert this if you want something changed. Please let me know.

PR #2174, opened 2021-03-18
[wallet] WatchTransaction method

rasom:
Currently for watching transaction's status the next algorithm is used:
1. Each new block with all transaction is fetched via `eth_getBlockByNumber` (once per 10 seconds)
2. two `eth_getLogs` requests are executed for erc20 transfers

This pr changes it to `eth_getTransactionByHash` every 10 seconds, thus three times less requests and much less traffic. 

Note: `reorgSafetyDepth` now always returns 0. The reason is that otherwise we will not be able to fetch transaction using range scan right after confirmation. Because of this one test is commented out and will be removed later. Reorgs checking code is a bit complicated and almost useless, it will be removed in a separate PR. In case if we will ever need to handle reorgs in future there is a different approach from what we have. But it is still unlikely that we will need it.  
 
status: ready


cammellos (approved):

Samyoul (approved):
Looks good. Sorry for the late review

PR #2175, opened 2021-03-19
Ban user from community

cammellos:
This commit introduces a new endpoint to ban a user from a community.

Messages from this user will be ignored.


vkjr (approved):

protocol/communities/community.go
```diff
@@ -322,6 +322,24 @@ func (o *Community) hasMember(pk *ecdsa.PublicKey) bool {
 	return member != nil
 }
 
+func (o *Community) BannedUser(pk *ecdsa.PublicKey) bool {
```
Samyoul: Maybe change the function name to `IsBanned` or `UserIsBanned`? At the moment the function reads like a Getter/Setter.

protocol/communities/community.go
```diff
@@ -322,6 +322,24 @@ func (o *Community) hasMember(pk *ecdsa.PublicKey) bool {
...
+	o.mutex.Lock()
+	defer o.mutex.Unlock()
+	return o.bannedUser(pk)
+
+}
+
+func (o *Community) bannedUser(pk *ecdsa.PublicKey) bool {
```
Samyoul: A matter of curiosity, why is the receiver `o` for the Type `Community`? It doesn't matter too much, the convention is usually to match the receiver name to the Type's name ie `c` to `Community`. `o` seems a interesting choice and I feel like there must be a story here 😄 .

protocol/protobuf/communities.proto
```diff
@@ -40,6 +40,7 @@ message CommunityDescription {
   CommunityPermissions permissions = 3;
   ChatIdentity identity = 5;
   map<string,CommunityChat> chats = 6;
+  repeated string ban_list = 7;
```
Samyoul, line 43: I can't think of a better way of doing this, but oh boy this list could grow!

protocol/requests/ban_user_from_community.go
```diff
@@ -0,0 +1,27 @@
...
+
+var ErrBanUserFromCommunityInvalidCommunityID = errors.New("ban-user-from-community: invalid community id")
+var ErrBanUserFromCommunityInvalidUser = errors.New("ban-user-from-community: invalid user id")
+
+type BanUserFromCommunity struct {
+	CommunityID types.HexBytes `json:"communityId"`
+	User        types.HexBytes `json:"user"`
+}
+
+func (j *BanUserFromCommunity) Validate() error {
```
Samyoul: Here's another funny receiver name. I can't see the pattern, it feels like some kind of puzzle. 🧠

Samyoul (approved):
Lovely. Just a few readability suggestions

protocol/communities/community.go
vkjr: I'd bet that was "organisation" at start :)

Samyoul: I think you are probably right.

cammellos: correct :)

protocol/protobuf/communities.proto
cammellos: yes, that might become eventually an issue as we are limited by 700K for community description, we'll have to split the message in batches eventually

PR #2176, opened 2021-03-19
[wallet] Skip transaction on chain id mismatch

rasom:
needed for https://github.com/status-im/status-react/issues/11902

Sometimes eth_getBlockByNumber returns txs with chainId which is not
equal to chain's id. That caused an error and tx fetching was
interrupted. From now on such txs will be skipped.

status: ready


cammellos (approved):

services/wallet/downloader_test.go
```diff
@@ -55,6 +55,7 @@ func (s *ETHTransferSuite) SetupTest() {
 	db, stop := setupTestDB(s.Suite.T())
 	s.dbStop = stop
 	s.downloader = &ETHTransferDownloader{
+		chain:  big.NewInt(1337),
```
Samyoul, line 58: 😎

Samyoul (approved):

services/wallet/downloader_test.go
rasom: Yeah, I was lazy. But it will work :D

PR #2177, opened 2021-03-23
Retrieving community info from mailserver

vkjr:
status-go part of [community link unfurling issue](https://github.com/status-im/status-react/issues/11775)

Created endpoint for retrieving info about community from mailserver: `RequestCommunityInfoFromMailserver()`
When called it installs new filter and makes request to mailserver. 
On community descritption messages arrival filter deinstalled and community info passed to status-react with `CommunityInfoFoundSignal` struct

p.s. I returned back `MessengerSignalHandler` because now we have 2 signals from messenger and there may be more.


protocol/messenger_communities.go
```diff
@@ -463,3 +464,55 @@ func (m *Messenger) RemoveUserFromCommunity(id types.HexBytes, pkString string)
...
+		return nil
+	}
+
+	//send signal to client that message status updated
+	community, err := m.communitiesManager.GetByIDString(communityID)
+	if err != nil || community == nil {
+		return fmt.Errorf("can't get community by ID: %v", err)
+	}
+
+	m.config.messengerSignalsHandler.CommunityInfoFound(community.Name(), community.Description(), community.MembersCount(), community.Verified())
```
cammellos: I would probably pass the whole community here, as we will likely show more data (the channels etc), so if you just pass the community object is going to be easier I think.

cammellos (approved):
Looks good, nice work!

jakubgs:
The `linux-05` host had issue with just OpenJDK `jre-headless` package being installed resulting in:
```
exec: "javac": executable file not found in $PATH
```
Fixed by installing `openjdk-11-jdk-headless` in https://github.com/status-im/infra-ci/commit/a51ed1a0 and making it default.

PR #2180, opened 2021-03-24
add PinMessage and PinnedMessage

gravityblast:
This PR adds a new protocol message called `PinMessage`.

Issue on status-react:
https://github.com/status-im/status-react/issues/11699

It also adds 2 new methods to the messenger APIs:

* `PinnedMessageByChatID` (similar to `MessageByChatID`)
* `SendPinMessage` 

`SendPinMessage` receives a `PinMessage`  object, similar to how `SendChatMessage` receives a `Message` object:

```
type PinMessage struct {
	protobuf.PinMessage

	// ID calculated as keccak256(compressedAuthorPubKey, data) where data is unencrypted payload.
	ID        string `json:"id"`
	MessageID string `json:"messageID"`
	// WhisperTimestamp is a timestamp of a Whisper envelope.
	WhisperTimestamp uint64 `json:"whisperTimestamp"`
	// From is a public key of the user who pinned the message.
	From string `json:"from"`
	// The chat id to be stored locally
	LocalChatID string           `json:"localChatId"`
	SigPubKey   *ecdsa.PublicKey `json:"-"`
	// Identicon of the author
	Identicon string `json:"identicon"`
	// Random 3 words name
	Alias string `json:"alias"`
}
```

The `protobuf.PinMessage` has a field `Pinned bool`, so the `PinMessage` json object can specify `pinned: [true|false]` to pin and unpin message.


protocol/message_handler.go
```diff
@@ -265,6 +265,59 @@ func (m *MessageHandler) HandleSyncInstallationPublicChat(state *ReceivedMessage
...
+	// Set the LocalChatID for the message
+	pinMessage.LocalChatID = chat.ID
+
+	if chat.LastClockValue < message.Clock {
+		chat.LastClockValue = message.Clock
+	}
+
+	state.Response.PinMessages = append(state.Response.PinMessages, pinMessage)
+
+	chat.Active = false
```
cammellos: This does not look right, it might hide the chat whenever a pinned message is received

protocol/message_persistence.go
```diff
@@ -807,6 +893,76 @@ func (db sqlitePersistence) SaveMessages(messages []*common.Message) (err error)
...
+				message.Clock,
+				message.Pinned,
+			}
+			_, err = insertStmt.Exec(allValues...)
+			if err != nil {
+				return
+			}
+		case nil:
+			// found, update if current message is more recent, otherwise skip
+			if existingTimestamp < message.Timestamp {
```
cammellos: We tend to use `Clock` and not timestamp, as timestamp can result in some issues of consistency, while clock is a logical clock

protocol/messenger_pin_messages.go
```diff
@@ -0,0 +1,116 @@
...
+// SendPinMessage sends the PinMessage to the corresponding chat
+func (m *Messenger) SendPinMessage(ctx context.Context, message *common.PinMessage) (*MessengerResponse, error) {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+	return m.sendPinMessage(ctx, message)
+}
+
+// SendPinMessages takes a array of pin messages and sends it based on the corresponding chats
+func (m *Messenger) SendPinMessages(ctx context.Context, messages []*common.PinMessage) (*MessengerResponse, error) {
+	m.mutex.Lock()
```
cammellos: I think `SentPinMessage` is probably enough, I don't think status-react will need to pin multiple messages in a single RPC call, so maybe we don't need to have it

protocol/messenger_pin_messages.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	message.ID = rawMessage.ID
+
+	err = m.persistence.SavePinMessages([]*common.PinMessage{message})
+	if err != nil {
+		return nil, err
+	}
+
+	response.PinMessages = []*common.PinMessage{message}
```
cammellos: We might want to use the `AddPinnedMessage` pattern on `response`, as that's what we are moving towards

protocol/messenger_pin_messages.go
```diff
@@ -0,0 +1,116 @@
...
+	// TODO: needed?
+	// response.Messages, err = m.pullMessagesAndResponsesFromDB([]*common.Message{message})
```
cammellos: I don't think so

protocol/messenger_response.go
```diff
@@ -191,6 +194,21 @@ func (r *MessengerResponse) overrideRemovedFilters(filters []*transport.Filter)
 	}
 }
 
+func (r *MessengerResponse) overridePinMessages(messages []*common.PinMessage) {
```
cammellos: Here we might want to use the patter we use with communities, so instead of an array use a `map`, as eventually we'd like to have them all using the same

cammellos (commented):
first pass, some comments :)

jrainville:
I tried to build this branch to start implementing pinned messages in Desktop, but I get the following build error:
```
 github.com/status-im/status-go/services/ext
services/ext/api.go:487:2: not enough arguments to return
services/ext/api.go:487:30: api.service.messenger.SendPinMessages undefined (type *"github.com/status-im/status-go/protocol".Messenger has no field or method SendPinMessages)
make[1]: *** [Makefile:143: statusgo-shared-library] Error 2
```

Also, question relative to the API: to add a pinned message, I have to call `SendPinMessage`? If so, what are the arguments I need to pass, I didn't understand well from the code.

Finally, is there a way to unpin a message?

briansztamfater:
> I tried to build this branch to start implementing pinned messages in Desktop, but I get the following build error:
> 
> ```
>  github.com/status-im/status-go/services/ext
> services/ext/api.go:487:2: not enough arguments to return
> services/ext/api.go:487:30: api.service.messenger.SendPinMessages undefined (type *"github.com/status-im/status-go/protocol".Messenger has no field or method SendPinMessages)
> make[1]: *** [Makefile:143: statusgo-shared-library] Error 2
> ```
> 
> Also, question relative to the API: to add a pinned message, I have to call `SendPinMessage`? If so, what are the arguments I need to pass, I didn't understand well from the code.
> 
> Finally, is there a way to unpin a message?

@jrainville  I got the same results while building for mobile but now with the latest changes the build succeeded. Try again with the latest changes and the build will probably pass!

gravityblast:
@jrainville I pushed a commit to fix the build problem and added more info in the description about the message structure and how to pin/unpin.

jrainville:
I confirm that it builds now. Thanks for that and the new info in the description 💯

gravityblast:
@jrainville @briansztamfater I forced-pushed after a rebase to have all the new changes from `develop`.

protocol/message_persistence.go
```diff
@@ -543,6 +543,91 @@ func (db sqlitePersistence) MessageByChatID(chatID string, currCursor string, li
...
+		result = append(result, pinnedMessage)
+		cursors = append(cursors, cursor)
+	}
+
+	var newCursor string
+	if len(result) > limit {
+		newCursor = cursors[limit]
+		result = result[:limit]
+	}
+	return result, newCursor, nil
```
Samyoul: The `cursors` slice maybe nil at this point, as there is no check on the slice's state.

```suggestion
	if len(result) > limit  && cursors != nil  {
		newCursor = cursors[limit]
		result = result[:limit]
	}
	return result, newCursor, nil
```

protocol/messenger_pin_message_test.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func (s *MessengerPinMessageSuite) newMessenger(shh types.Waku) *Messenger {
+	privateKey, err := crypto.GenerateKey()
+	s.Require().NoError(err)
+
+	messenger, err := newMessengerWithKey(s.shh, privateKey, s.logger, nil)
+	s.Require().NoError(err)
+	return messenger
+}
```
Samyoul: The `MessengerPinMessageSuite` provides the `types.Waku` via `s.shh`, so the `shh` parameter is never used

```suggestion
func (s *MessengerPinMessageSuite) newMessenger() *Messenger {
	privateKey, err := crypto.GenerateKey()
	s.Require().NoError(err)

	messenger, err := newMessengerWithKey(s.shh, privateKey, s.logger, nil)
	s.Require().NoError(err)
	return messenger
}
```

protocol/messenger_pin_messages.go
```diff
@@ -0,0 +1,89 @@
...
+	return m.sendPinMessage(ctx, message)
+}
+
+func (m *Messenger) sendPinMessage(ctx context.Context, message *common.PinMessage) (*MessengerResponse, error) {
+	var response MessengerResponse
+
+	// A valid added chat is required.
+	chat, ok := m.allChats[message.ChatId]
+	if !ok {
+		return nil, errors.New("Chat not found")
```
Samyoul: Tiny little thing, go convention is not to start an error message with a capital letter.

```suggestion
		return nil, errors.New("chat not found")
```

Samyoul (changes requested):

Samyoul (approved):
There only minor things that I spotted.

gravityblast:
thank you @Samyoul , I'm going to check your comments and update the code

jrainville:
@gravityblast I think the `SendPinMessage` api is missing the `ChatId` property. I couldn't call that API until I manually hardcoded my chatId in the go file. The API seems to only have the localChatId, but not the ChatId.

jrainville:
Also, is it possible that `pinnedMessageByChatID` is not part of the API? I don't see it in api.go and also, I get `the method wakuext_pinnedMessageByChatID does not exist/is not available`

gravityblast:
@jrainville sorry I though you had access to messageByChatID, but I see the right API should be `ChatMessages`.
so I added now `ChatPinnedMessages`, that you should see as `wakuext_chatPinnedMessages`

gravityblast:
> @gravityblast I think the SendPinMessage api is missing the ChatId property. I couldn't call that API until I manually hardcoded my chatId in the go file. The API seems to only have the localChatId, but not the ChatId.

@jrainville does it work if you call it `chat_id`? I see it's auto generated by protobuf

jrainville:
@gravityblast Cool, both work now. Thanks!

cammellos:
@gravityblast @briansztamfater I have attempted a rebase here https://github.com/status-im/status-go/tree/feature/pinned-messages-2 if you want to use it for the time being (I just made sure it compiles but haven't checked much more)

I have updated the timestamp of the migration as now in develop there's a couple of newer ones and this one needs to come last.

jrainville:
Everything works very well now, but it seems like there is no signal when there is a pin (or unpin) that is sent. I receive a signal with the chat and the last message, but it doesn't specify `pinned` or anything relative to that.

It would be super helpful if when a pin or unpin was done, it would send a message signal with the `pinned` flag update in it.

Let me know if it's doable. Thanks!

gravityblast:
@cammellos I also rebased and forced pushed here, /cc @briansztamfater @jrainville

gravityblast:
@jrainville ok I'll update it to send a signal as well

briansztamfater:
@gravityblast @jrainville  Not sure but thinking about edge cases, I'm wondering if we have a very old pinned message that is not loaded in the chat, we would also need to load the reactions for that specific message. Do we have any API method to get reactions for specific messageId(s)?

jrainville:
As far as I can see, that API doesn't exist. 
You raise a good point. That API could be useful.
However, I wonder what happens if the message and pin is very old (more than one month), you wouldn't be able to fetch it in the first place right?

gravityblast:
yes for now if you don't have the message itself you can't see it as a pinned message.
we can start with this and then see if we need some changes

jrainville:
Tested the latest commit on desktop and everything works now. Signals, fetching, pinning and unpining all work in 1-1 chats and group chats.

Can't wait for this to be merged 🙌

PR #2182, opened 2021-03-30
Add v4 endpoints for sign typed data

cammellos:
Adds `V4` endpoints for compatibility with metamask.


Samyoul (approved):
Looks great. The only teeny tiny thing is that your consts are declared at the bottom of your test file rather than at the top, but that is not a major thing.

jakubgs (approved):
lgtm

PR #2187, opened 2021-04-03
👤 Add random sessionId when API starts

### LINKED ISSUE #2183: Add a session id and time since last session event to anonymous metrics
# Problem
Since our metrics data has no user data attached to it, we are limited in the number of interactions we can measure. For example we can check how many times was the app opened or how many times did we send a message.

But this will not tell us anything about the active vs passive user behaviours. We can have a user who uses the app to communicate, and another one who only reads the messages.

By adding a session id, that get's reset each time the app is started, we'll have more information as to the actions a user takes while using the app. And since this session id is random, and changes every session, it cannot be linked back to the user, so no privacy is lost.

This session id will help us answer questions like:
- Of all the sessions, how many people checked the wallet tab
- Of all the sessions, sent a message 
- Of all the sessions, started a group chat

A session id adds more detail without exposing the individual behaviour. 

Apart from the session id, we also wish to add a time since last session event. This event will contain no personally identifiable information, nor the session id (and hence cannot be linked to other sessions). This event will tell us how often do users open the app.

# Implementation
1. Add a session id that is populated automatically at Go end. This session id should reset each time the node starts.
2. Add a new event type - time since last session id. This could be a special event that is computed at the backend and saved to the db. The frontend can just make an api call. Alternatively, this could be added automatically, when the session starts. Time since last session should be null for first time users.

# Acceptance Criteria
1. All events should have a session id, generated once per session
2. A new event type (validator) for `time-since-last-session` event

# Notes
CC: @hesterbruikman

shivekkhurana:
Closes #2183

### DIFF
--- a/services/appmetrics/api.go
+++ b/services/appmetrics/api.go
@@ -3,16 +3,19 @@ package appmetrics
 import (
 	"context"
 
+	"github.com/pborman/uuid"
+
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/status-im/status-go/appmetrics"
 )
 
 func NewAPI(db *appmetrics.Database) *API {
-	return &API{db: db}
+	return &API{db: db, sessionId: uuid.NewRandom().String()}
 }
 
 type API struct {
-	db *appmetrics.Database
+	db        *appmetrics.Database
+	sessionId string
 }
 
 func (api *API) ValidateAppMetrics(ctx context.Context, appMetrics []appmetrics.AppMetric) error {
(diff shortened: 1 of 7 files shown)


cammellos (approved):

services/appmetrics/api.go
```diff
@@ -3,16 +3,19 @@ package appmetrics
 import (
 	"context"
 
+	"github.com/pborman/uuid"
```
Samyoul, line 6: This repo uses the Google UUID package. Does this package have any special functionality missing from the Google package?

Samyoul (changes requested):

services/appmetrics/api.go
shivekkhurana: There is nothing special, we only need a UUID. I just discovered a usage in `account/accounts.go` and replicated it.

Samyoul: Ah, that's annoying. So we've already got multiple dependencies that serve the same purpose 😬, ok.

Flawless PR, then. 😆

Samyoul (approved):

Samyoul (approved):

Issue #2192, opened 2021-04-12
Medium.com link unfurling is broken


shivekkhurana:
# Problem
Medium has probably changed how they deliver content, this could be temporary but the tests are failing.

In our tests, we use [this article](https://medium.com/the-bitcoin-podcast-blog/a-look-at-the-status-im-ico-token-distribution-f5bcf7f00907) to verify unfurling. When opened in the browser, the article looks fine. 

But when opened using an `httpClient`, a 404 page is returned. This is also related to Medium moving to subdomains like `shivekkhurana.medium.com`. This used to be `medium.com/@shivekkhurana` earlier.

Here are the 404 responses returned by `httpClient.Get` for two medium urls:
* [test2.txt](https://github.com/status-im/status-go/files/6294912/test2.txt)
* [test3.txt](https://github.com/status-im/status-go/files/6294913/test3.txt)

# Implementation
Understand the new Medium routing, update unfurling code, enable Medium.com unfurling again.

# Acceptance Criteria
The app should be able to unfurl Medium links.

PR #2193, opened 2021-04-12
❌ Disable Medium.com unfurling

shivekkhurana:
Temporary measure to disable Medium.com link unfurling, because something has changed on Medium's end, and my tests were failing because of that.

* Medium Issue: https://github.com/status-im/status-go/issues/2192
* Companion React PR: https://github.com/status-im/status-react/pull/11980

### DIFF
--- a/protocol/urls/urls.go
+++ b/protocol/urls/urls.go
@@ -99,11 +99,13 @@ func LinkPreviewWhitelist() []Site {
 			Address:   "github.com",
 			ImageSite: false,
 		},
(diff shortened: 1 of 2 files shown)


vkjr (approved):

vkjr:
@shivekkhurana, maybe test for medium and its mention in whitelist better to be commented out with `TODO` note than removed. Just to be a reminder against `temprorary` being `forever` :) wdyt?

PR #2197, opened 2021-04-13
[wallet] Pending transfers API

rasom:
This pr updates methods related to pending transactions. See more details in `services/wallet/README.md`

status: ready


cammellos:
@rasom could you elaborate on the differences for pending transactions? Ideally we have only one as it's hard to tell the difference between the two just by looking at the code
thanks

rasom:
@cammellos @richard-ramos 
On desktop side it is used for watching transactions, specifically in case when app is closed before transaction is confirmed. 
That's the list of transaction "types" https://github.com/status-im/status-desktop/blob/fb8380a861d89063734217404fa5244901353fbc/src/status/libstatus/types.nim#L206-L210

That's fields used for `pending_transactions` table
```sql
transaction_hash VARCHAR NOT NULL,
blk_number BIGINT NOT NULL,  -- apparently that's a current block number, used as a timestamp
from_address VARCHAR NOT NULL, 
to_address VARCHAR NOT NULL,
type VARCHAR NOT NULL, -- one of those transaction types
data TEXT -- some additional information specific for transaction type, 
-- that's not actually transaction's data field
```

On mobile side we need it for showing pending transactions in tx history list after re-login. So `pending_transfers` table contains fields we need for showing transfer in the list:
```sql
hash VARCHAR NOT NULL,
timestamp UNSIGNED BIGINT NOT NULL,
from_address VARCHAR NOT NULL,
to_address VARCHAR NOT NULL,
symbol VARCHAR,  -- ETH or some ERC20 token, for instance SNT 
gas_price BLOB NOT NULL,
gas_limit BLOB NOT NULL,
value BLOB NOT NULL,
data TEXT, -- actual tx's data  
```

So if we want to merge it, I would probably remove `blk_number` at all and used `timestamp` instead, renamed `pending_transactions`'s `data` to something like `additional_data` and then `symbol`, `value`,`gas_price`,  `gas_limit `, `type` to be `null`.

Merged table would look like
```sql
hash VARCHAR NOT NULL,
timestamp UNSIGNED BIGINT NOT NULL,
from_address VARCHAR NOT NULL,
to_address VARCHAR NOT NULL,
symbol VARCHAR, 
gas_price BLOB,
gas_limit BLOB,
value BLOB,
data TEXT,
type VARCHAR, 
additional_data TEXT,
```

Would that work?

cammellos:
@rasom thanks, 
the purpose seems to be similar (i.e you make a transaction from the status app, you want to keep track of it), so I would merge it.
Mind that sqlite does not support removing or renaming fields (you'd have to re-create a table, copy over the data etc).
But in general, it would make sense to have only a single table, as far as I can see, but you will have better context, so ultimately it's up to you.
thanks

Samyoul (approved):
Looks good.

rasom:
@cammellos @richard-ramos updated, ready for review

richard-ramos (approved):

michaelsbradleyjr:
Change of the key name in the JSON from `transactionHash` to `hash` made for some (eventual) runtime surprises, though to be fair it's (sort of) documented in the PR. 😄

PR #2181, opened 2021-03-29
More granular locking on Messenger

Samyoul:
Hopefully addresses issues reported in https://github.com/status-im/status-go/issues/1936.

PR #2199, opened 2021-04-14
Activity center, messages from contacts only and remove messenger lock

cammellos:
This PR incorporate 3 changes.

- Add activity center
- Add option for messages from contacts only
- Remove messenger locks

Normally they would be 3 separate PRs, but because we want to release this quickly, the QA team is testing them together, so it makes sense to have a single PR incorporating all 3.


protocol/activity_center.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+type ActivityCenterType int
+
+const (
+	ActivityCenterNotificationTypeNewOneToOne = iota + 1
+	ActivityCenterNotificationTypeNewPrivateGroupChat
+)
+
+var ErrInvalidActivityCenterNotification = errors.New("invalid activity center notification")
```
vkjr, line 21: I think a comment explaining what an "activity center" concept is would be helpful

protocol/communities_messenger_test.go
```diff
@@ -956,6 +956,11 @@ func (s *MessengerCommunitiesSuite) TestShareCommunity() {
 	s.Require().NotNil(response)
 	s.Require().Len(response.Messages, 1)
 
+	// Add bob to contacts so it does not go on community center
```
vkjr: did you mean "activity center"?

protocol/message_handler.go
```diff
@@ -166,11 +203,20 @@ func (m *MessageHandler) handleCommandMessage(state *ReceivedMessageState, messa
 	if err := message.PrepareContent(); err != nil {
 		return fmt.Errorf("failed to prepare content: %v", err)
 	}
-	chat, err := m.matchChatEntity(message, state.AllChats, state.Timesource)
+	chat, err := m.matchChatEntity(message, state.AllChats, state.AllContacts, state.Timesource)
 	if err != nil {
 		return err
 	}
 
+	allowed, err := m.allowMessageFrom(state.AllContacts, state.CurrentMessageState.Contact.ID, chat)
```
vkjr: I'd suggest to rename function to `isMessageAllowedFrom`, because `allowMessageFrom` sound as an instruction to allow, not as a question

protocol/message_handler.go
```diff
@@ -241,25 +303,26 @@ func (m *MessageHandler) HandleSyncInstallationContact(state *ReceivedMessageSta
...
 		contact.LocalNickname = message.LocalNickname
 
-		state.ModifiedContacts[contact.ID] = true
-		state.AllContacts[contact.ID] = contact
+		state.ModifiedContacts.Store(contact.ID, true)
+		state.AllContacts.Store(contact.ID, contact)
 	}
 
-	state.AllChats[chat.ID] = chat
+	// TODO(samyoul) remove storing of an updated reference pointer?
```
vkjr, line 305: @cammellos, there are similar todos on lines 395, 435, 606 and more in other files. Is it a question to @Samyoul that should be answered before this PR merge? Or it can be decided after? If after - I would suggest to create a dedicated issue, because we already have around hundred todos in code and these ones also risk to become a "part of the ship, part of the crew" 🏴‍☠️

protocol/message_handler.go
```diff
@@ -498,11 +583,28 @@ func (m *MessageHandler) HandleChatMessage(state *ReceivedMessageState) error {
 		return err
 	}
 
-	// Set chat active
-	chat.Active = true
+	// If the chat is not active, create a notification in the center
```
vkjr, line 581: Looks like a good candidate for  separate `addActivityCenterNotification` function

protocol/messenger_locker.go
```diff
@@ -0,0 +1,224 @@
+package protocol
```
vkjr: I'd suggest to rename this file to `messenger_maps` because `locker` sounds like some new entity and it never mentioned again inside the file.

protocol/messenger_locker.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	cm.sm.Range(nf)
+}
+
+func (cm *contactMap) Delete(contactID string) {
+	cm.sm.Delete(contactID)
+}
+
+/*
+|--------------------------------------------------------------------------
```
vkjr: Haven't seen this style of commenting before. Any reason to introduce it?

protocol/activity_center_persistence.go
```diff
@@ -0,0 +1,273 @@
...
+		}
+
+		notifications = append(notifications, notification)
+	}
+
+	return latestCursor, notifications, nil
+
+}
+func (db sqlitePersistence) buildActivityCenterQuery(tx *sql.Tx, cursor string, limit int, ids []types.HexBytes) (string, []*ActivityCenterNotification, error) {
+	args := []interface{}{}
```
Samyoul: A little thing, generally an empty slice should be declared via a variable rather than via a literal. It is a small optimisation as you don't cause memory allocation until the slice is written to.

```suggestion
	var args []interface{}
```

protocol/chat.go
```diff
@@ -165,10 +165,17 @@ func (c *Chat) HasMember(memberID string) bool {
...
+			return true
+		}
+	}
 
-	// Check if we were already in the chat
-	hadJoined := c.HasMember(myID)
+	return false
+}
+
+func (c *Chat) updateChatFromGroupMembershipChanges(myID string, g *v1protocol.Group) {
```
Samyoul: It's a bit hard to see in the diff, but it doesn't seem that the `myID` param is used any more.

protocol/message_handler.go
Samyoul: These comments are written by me, they come from https://github.com/status-im/status-go/pull/2181, which is a bit of WIP / proof of concept / experiment.

To be honest I'm not sure what to do about these, we update a reference and then save the reference back in the sync.Map. In all of these cases the reference was originally pulled from the sync.Map, so there is no need to save the reference back into the map. However leaving the `Store()` does make it explicit that the referenced chat is updated into the global map.

A dedicated issue for this is a good idea.

protocol/messenger.go
```diff
@@ -74,6 +74,7 @@ var messageCacheIntervalMs uint64 = 1000 * 60 * 60 * 48
 // because installations are managed by the user.
 // Similarly, it needs to expose an interface to manage
 // mailservers because they can also be managed by the user.
+// TODO we may want to change the maps into sync.Maps if we start getting unexpected locks or weird collision bugs
```
Samyoul: Ah, this can be removed now I suppose. Do you want me to update my branch?

protocol/message_handler.go
cammellos: We can create a separate issue, could you please open one @Samyoul ?
Thanks

protocol/messenger_locker.go
cammellos: cc @Samyoul

protocol/messenger_activity_center.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	if len(chats) != 0 {
+		err := m.saveChats(chats)
+		if err != nil {
+			return nil, err
+		}
+	}
+	m.mutex.Unlock()
+	return response, nil
+}
```
Samyoul: You've got early returns in this function what would make this unlock unreachable.

```suggestion
func (m *Messenger) processAcceptedActivityCenterNotifications(notifications []*ActivityCenterNotification) (*MessengerResponse, error) {
	m.mutex.Lock()
	defer m.mutex.Unlock()
	
	response := &MessengerResponse{}
	var chats []*Chat
	for _, notification := range notifications {
		if notification.ChatID != "" {
			chat, ok := m.allChats.Load(notification.ChatID)
			if !ok {
				// This should not really happen, but ignore just in case it was deleted in the meantime
				m.logger.Warn("chat not found")
				continue
			}
			chat.Active = true

			chats = append(chats, chat)
			response.AddChat(chat)
		}
	}
	if len(chats) != 0 {
		err := m.saveChats(chats)
		if err != nil {
			return nil, err
		}
	}
	return response, nil
}
```

cammellos:
@vkjr addressed feedback, thanks!

protocol/messenger_locker.go
Samyoul: Yeah, good point, this file used to contain a custom mutex for `Messenger` but was gradually changed to only include custom sync.Maps.

vkjr (approved):

protocol/messenger_locker.go
Samyoul: Generally when writing a file with a lot of repetitive structs I introduce these kinds of comments to form visual delimiter, these comments make reading through the file easier to see where the different sections of code are.

Samyoul (approved):
Great. Mutex without a deferred unlock needs attention.

protocol/messenger.go
cammellos: I'll do it in this branch, thanks

protocol/messenger_activity_center.go
cammellos: nice catch!!!

protocol/messenger_locker.go
vkjr: I'm totally up for divider lines, maybe something like `/* --------- */` But think it would be convenient to put types description right before declarations to fit go documentation conventions. For example for later use with `godoc`

Samyoul: Yeah, that's true

protocol/message_handler.go
Samyoul: https://github.com/status-im/status-go/issues/2203

jakubgs (approved):
Couldn't find anything wrong. But I'm no Go wizard.

PR #2200, opened 2021-04-14
Whisper cleanup, step 1

vkjr:
Refactorings:

- removed `protocol/transport/whisper` directory
- get rid of transport interface
- removed transport/waku/mailserver.go becase its content fits waku_service
- EnvelopesMonitor, FiltersManager and migration files moved from `transport/waku` to `transport`
- renamed `waku_service` -> `transport`
- receivers for `Transport` named `t`, for `FiltersManager` named `f`


protocol/transport/envelopes_monitor.go
```diff
@@ -1,11 +1,27 @@
...
 )
 
+// EnvelopeState in local tracker
+type EnvelopeState int
+
+const (
+	// NotRegistered returned if asked hash wasn't registered in the tracker.
+	NotRegistered EnvelopeState = -1
+	// EnvelopePosted is set when envelope was added to a local waku queue.
+	EnvelopePosted EnvelopeState = iota
```
Samyoul: It is best to avoid zero value consts as they can collide with nil value ints.

```suggestion
	EnvelopePosted EnvelopeState = iota + 1
```

Samyoul (approved):
Looks good, and good to see this tidy.

cammellos:
@vkjr we can now merge this and the corresponding status-react PR

PR #2201, opened 2021-04-14
Added new make migration command

Samyoul:
## What's changed

A new `make` command

```sh
$ make migration DESC=add_new_columns_to_table
created appdatabase/migrations/sql/1618416974_add_new_columns_to_table.up.sql
```

## Why make the change

It is annoying to manually make a new migration file every time I want to make a new unique migration. This is better 😄

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -367,3 +367,10 @@ clean-mailserver-systemd: ##@Easy Clean your systemd service for running a mails
 
 clean-mailserver-docker: ##@Easy Clean your Docker container running a mailserver
 	@cd _assets/compose/mailserver && $(MAKE) clean
+
+DEFAULT_MIGRATION_PATH := appdatabase/migrations/sql/
+UNIX_TIMESTAMP := $(shell date +%s)
+migration:
+	@echo >> $(DEFAULT_MIGRATION_PATH)$(UNIX_TIMESTAMP)_$(DESC).up.sql
+	@echo created $(DEFAULT_MIGRATION_PATH)$(UNIX_TIMESTAMP)_$(DESC).up.sql
+	## TODO(samyoul) option to set and/or overwrite migration path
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -367,3 +367,10 @@ clean-mailserver-systemd: ##@Easy Clean your systemd service for running a mails
 
 clean-mailserver-docker: ##@Easy Clean your Docker container running a mailserver
 	@cd _assets/compose/mailserver && $(MAKE) clean
+
+DEFAULT_MIGRATION_PATH := appdatabase/migrations/sql/
+UNIX_TIMESTAMP := $(shell date +%s)
+migration:
```
jakubgs, line 374: If the variables are specific to the `make` target, make them actually specific to the target:
```suggestion
migration: DEFAULT_MIGRATION_PATH := appdatabase/migrations/sql/
migration: UNIX_TIMESTAMP := $(shell date +%s)
migration:
```
https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html

Makefile
```diff
@@ -367,3 +367,10 @@ clean-mailserver-systemd: ##@Easy Clean your systemd service for running a mails
...
+	@echo >> $(DEFAULT_MIGRATION_PATH)$(UNIX_TIMESTAMP)_$(DESC).up.sql
```
jakubgs: Why `echo`? You're not actually putting anything in it as far as I can tell, then just use `touch`:
```suggestion
	@touch $(DEFAULT_MIGRATION_PATH)$(UNIX_TIMESTAMP)_$(DESC).up.sql
```

Makefile
```diff
@@ -367,3 +367,10 @@ clean-mailserver-systemd: ##@Easy Clean your systemd service for running a mails
...
+	@echo created $(DEFAULT_MIGRATION_PATH)$(UNIX_TIMESTAMP)_$(DESC).up.sql
```
jakubgs: What's the point of this echo? You can achieve pretty much the same thing by just removing the `@` from `touch`.

Samyoul: Yeah, but I wanted the word "created" to show... This'll be fine.

Makefile
Samyoul: `touch` is better, yeah. Changed

Makefile
Samyoul: Done

jakubgs (approved):

PR #2204, opened 2021-04-15
Added custom path and default protocol path

Samyoul:
## What has changed

Improved the `make migration` command to:

- Replace `DESC` flag with `D`
- Added an over-writable default `PATH` flag
- Added default protocol migration target `make migration-protocol`

## Why make the change

Because it is useful to generate a migration in any directory you'd like.

## Usage

```
$ make migration D="added_columm_to_table"
touch appdatabase/migrations/sql/1618502237_added_columm_to_table.up.sql

$ make migration D="added_columm_to_table" PATH="some/custom/path"
touch some/custom/path/1618502237_added_columm_to_table.up.sql

$ make migration-protocol D="add_goats_to_boats"
touch protocol/migrations/sqlite/1618502450_add_goats_to_boats.up.sql
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -368,8 +368,11 @@ clean-mailserver-systemd: ##@Easy Clean your systemd service for running a mails
 clean-mailserver-docker: ##@Easy Clean your Docker container running a mailserver
 	@cd _assets/compose/mailserver && $(MAKE) clean
 
-## TODO(samyoul) option to set and/or overwrite migration path
-migration: DEFAULT_MIGRATION_PATH := appdatabase/migrations/sql/
-migration: UNIX_TIMESTAMP := $(shell date +%s)
+UNIX_TIMESTAMP := $(shell date +%s)
+migration: PATH := appdatabase/migrations/sql
 migration:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -368,8 +368,11 @@ clean-mailserver-systemd: ##@Easy Clean your systemd service for running a mails
 clean-mailserver-docker: ##@Easy Clean your Docker container running a mailserver
 	@cd _assets/compose/mailserver && $(MAKE) clean
 
-## TODO(samyoul) option to set and/or overwrite migration path
-migration: DEFAULT_MIGRATION_PATH := appdatabase/migrations/sql/
-migration: UNIX_TIMESTAMP := $(shell date +%s)
+UNIX_TIMESTAMP := $(shell date +%s)
```
jakubgs: If you put it here it will execute every time you run any `make` target. And only two targets need it.
```
$(UNIX_TIMESTAMP)
$(shell date +%s)
```
It's the same length, just use the `shell date +%s` command directly in the target command.

Samyoul: Resolved

jakubgs (approved):

PR #2211, opened 2021-04-22
[WIP] Add user-facing explainer

hesterbruikman:
- Added a user-facing explainer at the top of the doc as we link to this file from the UI.
- Moved all content that was already in the doc to below the divider with the heading 'How it works'

### DIFF
--- a/_docs/app-metrics.md
+++ b/_docs/app-metrics.md
@@ -13,7 +13,7 @@ Status is an open-source platform. To verify that the app only collects and shar
 
 In detail this means the app can collect and share:
 - Navigation events
-- Screens that is navigated to
+- Screens that are navigated to
 - Action events (for example button clicks, list swipes, switch toggles, etc)
 - Time events are created
 - Operating System
--- a/_docs/app-metrics.md
+++ b/_docs/app-metrics.md
@@ -5,7 +5,7 @@ _______
 
 _______
 
-Starting release 1.14, the Status mobile app asks to share privacy-preserving, end-to-end encrypted data about how you use Status. Data is only ever shared if you opt in to doing so, you can review all data before it is sent and it is shared over Status' peer-to-peer network (Waku), just like a 1:1 message.
+Starting release 1.14, the Status mobile app asks to share data about how you use Status. Data is only ever shared if you opt in to doing so, you can review all data before it is sent and it is shared over Status' peer-to-peer network (Waku), just like a 1:1 message.
 
 Sharing data is strictly opt-in, and can be changed at any times in Settings in the app.
 
@@ -27,7 +27,7 @@ In detail this means the app can collect and share:
 - Time since last batch sent
 
 For any data to be collected and shared, it needs to meet the rules set in the validator script. By policy of Status' Core Contributors, rules set in the validator script need to be assessed as posing a non-existent to low threat by being in:
-- local storage over time **(Specify what worst case can be and when this occurs)**
+- local storage over time
 - aggregated storage from Status over time
 
 ### What will never be shared
--- a/_docs/app-metrics.md
+++ b/_docs/app-metrics.md
@@ -1,5 +1,5 @@
 
-# Anonymous usage data
+# Status usage data
 Starting release 1.14, the Status mobile app asks to share anonymous end-to-end encrypted data about how you use Status. Data is only ever shared if you opt in to doing so, you can review all data before it is sent and it is shared anonymously over Waku, just like a 1:1 message.
 
 Sharing data is strictly opt-in, and can be changed at any times in Settings in the app.
@@ -23,7 +23,7 @@ In detail this means the app can collect and share:
 - Time since last data message sent
 
 ### What will never be shared
-No data will be shared unless you opt-in. Furthermore, Status commits to never collect identifiable data in its broadest sense. This means that we will never collect anything that can be linked back to you. Including but not limited to:
+No data will be shared unless you opt-in. Furthermore, Status commits to never collect identifiable data in its broadest sense. This means that we will never collect anything that we believe can be linked back to you. Including but not limited to:
 - IP addresses
 - Random name
 - Chat key
@@ -39,7 +39,7 @@ No data will be shared unless you opt-in. Furthermore, Status commits to never c
 While we employ a ‘can’t be evil’ approach. We highly recommend that you employ a ‘don’t trust, but verify’ approach. In this case, the above data can technically be logged, stored and shared. In order to verify that this does not happen, view the rules defined to store metrics data in the [source code](https://github.com/status-im/status-go/blob/develop/appmetrics/validators.go).
 
 ### Purpose of the data
-Anonymous, aggregated data is used to inform product development. Given our principles, the type of data that is collected and its anonymous nature, there is no incentive for Status or other parties to use the data for any other purpose.
+Aggregated data is used to inform product development. Given our principles, the type of data that is collected and its rudimentary nature, there is no incentive for Status or other parties to use the data for any other purpose.
 
 
 ### Viewing the data
--- a/_docs/app-metrics.md
+++ b/_docs/app-metrics.md
@@ -49,7 +49,7 @@ The above data can technically be logged, stored and shared. In order to verify 
 Moreover, while we made a best effort to make data collection as privacy-preserving as possible, we cannot guarantee that there are no ways to de-anonymous collected data. **If you find yourself at risk if your identity were to be uncovered: don't enable sharing data.**
 
 ### Purpose of the data
-Aggregated data is used to inform product development. Given our principles, the type of data that is collected and its rudimentary nature, there is no incentive for Status or other parties to use the data for any other purpose.
+Aggregated data is used to inform product development. Given our principles, the type of data that is collected and its rudimentary nature, there is no incentive for Status or other parties to use the data for any other purpose. Status will never use these data for profit.
 
 
 ### Viewing the data
--- a/_docs/app-metrics.md
+++ b/_docs/app-metrics.md
@@ -5,7 +5,7 @@ Starting release 1.14, the Status mobile app asks to share anonymous end-to-end 
 Sharing data is strictly opt-in, and can be changed at any times in Settings in the app.
 
 ### What is shared (Opt-in)
-Status is an open-source platform. To verify that the app only collects and shares the events and metadata listed below, you can view the rules defined to store metrics data in the [source code](https://github.com/status-im/status-go/blob/develop/appmetrics/validators.go).
+Status is an open-source platform. To verify that the app only collects and shares the events and metadata listed below, you can view the rules set to store metrics data in the [source code](https://github.com/status-im/status-go/blob/develop/appmetrics/validators.go).
 
 - Your interactions with the app like clicks and screenviews
 - Background activity and internal processes
@@ -21,6 +21,10 @@ In detail this means the app can collect and share:
 - Batch ID
 - Time since last session
 - Time since last data message sent
+
+For any data to be collected and shared, it needs to meet the rules set in the validator script. By policy of Status' Core Contributors, rules set in the validator script need to be assessed as posing a non-existent to low threat by being in:
+- local storage over time **(Specify what worst case can be and when this occurs)**
+- aggregated storage from Status over time
 
 ### What will never be shared
 No data will be shared unless you opt-in. Furthermore, Status commits to never collect identifiable data in its broadest sense. This means that we will never collect anything that we believe can be linked back to you. Including but not limited to:
--- a/_docs/app-metrics.md
+++ b/_docs/app-metrics.md
@@ -52,8 +52,6 @@ As of writing, we plan to make a Dashboard that shows the aggregated data public
 _________
 
 # How it works
-## App Metrics
-
 `appmetrics` is a way to capture and transfer app usage data, anonymously over Waku to a Status managed key.
 
 
--- a/_docs/app-metrics.md
+++ b/_docs/app-metrics.md
@@ -9,7 +9,7 @@ Status is an open-source platform. To verify that the app only collects and shar
 
 - Your interactions with the app like clicks and screenviews
 - Background activity and internal processes
-- Settings and pre=eferences
+- Settings and preferences
 
 In detail this means the app can collect and share:
 - Navigation events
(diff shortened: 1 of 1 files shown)


_docs/app-metrics.md
```diff
@@ -1,4 +1,58 @@
...
+Starting release 1.14, the Status mobile app asks to share anonymous end-to-end encrypted data about how you use Status. Data is only ever shared if you opt in to doing so, you can review all data before it is sent and it is shared anonymously over Waku, just like a 1:1 message.
+
+Sharing data is strictly opt-in, and can be changed at any times in Settings in the app.
+
+### What is shared (Opt-in)
+Status is an open-source platform. To verify that the app only collects and shares the events and metadata listed below, you can view the rules defined to store metrics data in the [source code](https://github.com/status-im/status-go/blob/develop/appmetrics/validators.go).
+
+- Your interactions with the app like clicks and screenviews
+- Background activity and internal processes
+- Settings and pre=eferences
```
Samyoul:
```suggestion
- Settings and preferences
```

_docs/app-metrics.md
```diff
@@ -1,4 +1,58 @@
...
+
+In detail this means the app can collect and share:
+- Navigation events
+- Screens that is navigated to
```
Samyoul:
```suggestion
- Screens that are navigated to
```

_docs/app-metrics.md
```diff
@@ -1,4 +1,58 @@
...
+- Action events (for example button clicks, list swipes, switch toggles, etc)
+- Time events are created
```
Samyoul:
```suggestion
- The time events are created
```

_docs/app-metrics.md
```diff
@@ -1,4 +1,58 @@
...
+- Operating System
+- App version
+- Session key
```
Samyoul: This is just a random string, I think that `id` sounds less scary
```suggestion
- Session id
```

_docs/app-metrics.md
```diff
@@ -1,4 +1,58 @@
...
+### Purpose of the data
+Anonymous, aggregated data is used to inform product development. Given our principles, the type of data that is collected and its anonymous nature, there is no incentive for Status or other parties to use the data for any other purpose.
+
+
+### Viewing the data
+As of writing, we plan to make a Dashboard that shows the aggregated data public; Similar to metrics.status.im. As soon as this Dashboard is available we will provide a link here. The data that is stored on your device and shared can be viewed through the interface of the app.
+
+
+
+
```
shivekkhurana: Is there a purpose to this whitespace?

_docs/app-metrics.md
```diff
@@ -1,4 +1,58 @@
...
+- App version
+- Session key
+- Time since last session
+- Time since last data message sent
+
+### What will never be shared
+No data will be shared unless you opt-in. Furthermore, Status commits to never collect identifiable data in its broadest sense. This means that we will never collect anything that can be linked back to you. Including but not limited to:
+- IP addresses
+- Random name
+- (Regular) Chat key
```
Samyoul: What does "(Regular)" mean

_docs/app-metrics.md
shivekkhurana: Session key is not apt. The term session generally links to a logged in user. But we use session key to group events. Not to identify users. Also this session key is randomly generated. Something like `batch id` will be more suitable.

Samyoul (approved):
Looks good Hester I've made some tpyo / minor suggestions, but apart from that congrats on your first PR in `status-go`

shivekkhurana (changes requested):

_docs/app-metrics.md
hesterbruikman: Good call. Thanks

hesterbruikman: Will opt for Batch ID as @shivekkhurana suggested, opposed to 'session' indeed batch is more accurate

_docs/app-metrics.md
hesterbruikman: It refers in this case to the Chat key you use for conversations. I'll leave it to 'Chat key', the user doesn't need to know another key is used as identifier on Waku to send a batch. Thanks

_docs/app-metrics.md
hesterbruikman: Nope, I'll remove:)

_docs/app-metrics.md
```diff
@@ -1,5 +1,57 @@
...
+- Public Ethereum addresses
+- Account balance(s)
+- Account history
+- ENS name
+- Input field entries (browser address bar, chat key/ENS entry)
+- Any content you generate (images, messages, profile picture)
+- Contacts or other (favorite) lists (wallet and browser favorites)
+- Chat, group, community memberships
+
+While we employ a ‘can’t be evil’ approach. We highly recommend that you employ a ‘don’t trust, but verify’ approach. In this case, the above data can technically be logged, stored and shared. In order to verify that this does not happen, view the rules defined to store metrics data in the [source code](https://github.com/status-im/status-go/blob/develop/appmetrics/validators.go).
```
arnetheduck: Hm, I believe this goes too far in the promises it makes - basically all metrics data is toxic and enables ways to be evil for someone that makes the effort to do so - the contract between us and the user is basically one of trust: they trust that we will keep the collected and stored data private and not exploit novel big data approaches to de-anonymize them.

De-anonymizing is based on finding correlated patterns - navigation events in particular are fairly easy to link to particular pages you have visited, and the more data they contain (clicks, interactions, time), the more likely it is that you can narrow them down matching more particular information like transactions. 

IE if you buy something and we record the timestamp, and you navigate and interact with a pattern first consistent with the transfers screen then in the browser with etherscan to look at the transaction status, this is enough information to create a "user profile" around that event and start correlating it with other actions / transactions.

It's incredibly hard to keep promises like this, to the point where it might easier to say that while we strive to keep the data private and anonymous, we cannot guarantee that a determined entity will make alternative use of it. For example, if you're a journalist or living in a country with restrictive laws or aggressive, you _most definitely don't under any circumstances_ want to have metrics enabled.

Here's a [fun example](https://ethresear.ch/t/packetology-validator-privacy/7547) of how this is done for ethereum itself - similar to the promises above, IP addresses and identities are not collected and exposed in ethereum - nevertheless it's trivial to link identities to nodes, which in turn enables grafting attacks against individual stakers should someone make the effort.

hesterbruikman: I see what you mean. Jrhea's work is a clear example. How about:

> "The above data can technically be logged, stored and shared. In order to verify that this does not happen, view the rules defined to store metrics data in the [source code](https://github.com/status-im/status-go/blob/develop/appmetrics/validators.go). 

> Moreover, while we made a best effort to make data collection as privacy-preserving as possible, we cannot guarantee that there are no ways to de-anonymous collected data. **If you find yourself at risk if your identity is uncovered: don't enable sharing data.**"

I'd also repeat that last sentence at the top. @shivekkhurana related to this: I think it's best if we also remove the 4th bullet on the `About sharing data` screen in the UI. "Usage data cannot be associated with your IP address"

_docs/app-metrics.md
```diff
@@ -1,5 +1,57 @@
...
+- Settings and preferences
+
+In detail this means the app can collect and share:
+- Navigation events
+- Screens that are navigated to
+- Action events (for example button clicks, list swipes, switch toggles, etc)
+- The time events are created
+- Operating System
+- App version
+- Batch ID
```
arnetheduck, line 28: what's a batch id?

hesterbruikman: Data is stored and send in batches to limit data that is stored over time. A batch is a collection of data that is removed after transmission. Batch ID is there to view events against an interval. High number of events in a batch implies active use, low number of events in a batch implies passive use. It's somewhat similar to a Session ID but not the same as we don't have a definition of a session that triggers transmission.

cc @Samyoul @shivekkhurana to clarify if needed

_docs/app-metrics.md
arnetheduck: yep, that's good.

> In order to verify that this does not happen, view the rules defined to store metrics data in the source code.

Hm, for the general, I suspect the most important thing to verify is that the build has been signed off by us - already by including this feature we're setting up a trust relationship between Status and the user, which in and of itself is a target  - the next step of course is to read the source code, but if I were to choose a weak link, it would probably be to make the user believe they're running the code they're being presented with, while implementing my own rules. This goes for everything the user does with Status the app of course, so maybe it's already covered elsewhere.

hesterbruikman: So the idea being, someone forks to build, changes the validator rules, adds events and distributes the build. I don't think we currently have a counter measure for that. We show build information inside the app as a general mitigation to check your build (very minimal, I wouldn't know how to use it), but you can't view that information before you are asked to opt-in to metrics. We can add it to the UI; I strongly doubt it'll be effective along the wealth of information presented already. I'll bring this to #core and the design team too see if there are any suggestions on how to handle this threat

hesterbruikman: fyi, the UI links directly to the rules (this url https://github.com/status-im/status-go/blob/develop/appmetrics/validators.go) so you can view them. That doesn't resolve the case you describe though, because obviously that url can be changed and you'd need to notice that you viewing rules in a different repo

arnetheduck: It's a requirement for most app stores that builds are signed and there are ways to verify this signature - I'd be surprised if it isn't covered already - the more economically profitable way to modify the app is to simply send all funds to some account ;) It might be a good thing to mention though when the feature is enabled.

hesterbruikman: It's covered as far as I understand the function of `app version` in about is to verify the signature. However, you can only do this after you're asked to share data.

_docs/app-metrics.md
```diff
@@ -1,5 +1,57 @@
...
+Aggregated data is used to inform product development. Given our principles, the type of data that is collected and its rudimentary nature, there is no incentive for Status or other parties to use the data for any other purpose.
+
+
+### Viewing the data
+As of writing, we plan to make a Dashboard that shows the aggregated data public; Similar to metrics.status.im. As soon as this Dashboard is available we will provide a link here. The data that is stored on your device and shared can be viewed through the interface of the app.
+
+_________
+
+# How it works
 `appmetrics` is a way to capture and transfer app usage data, anonymously over Waku to a Status managed key.
```
corpetty: another use of the term "anonymous" which might need to be changed if we're moving away from that language and implicit guarantees.

_docs/app-metrics.md
```diff
@@ -1,5 +1,57 @@
...
+- ENS name
+- Input field entries (browser address bar, chat key/ENS entry)
+- Any content you generate (images, messages, profile picture)
+- Contacts or other (favorite) lists (wallet and browser favorites)
+- Chat, group, community memberships
+
+While we employ a ‘can’t be evil’ approach. We highly recommend that you employ a ‘don’t trust, but verify’ approach. In this case, the above data can technically be logged, stored and shared. In order to verify that this does not happen, view the rules defined to store metrics data in the [source code](https://github.com/status-im/status-go/blob/develop/appmetrics/validators.go).
+
+### Purpose of the data
+Aggregated data is used to inform product development. Given our principles, the type of data that is collected and its rudimentary nature, there is no incentive for Status or other parties to use the data for any other purpose.
```
corpetty: Even though we feel there is no incentives to use it for profit, we should also explicitly state that we will never do so. Many data platforms didn't realize they were data platform until they had the treasure in hand, which then pushed a change to make profit from it.

corpetty (commented):
Imo, we should have EVERY intention of performing whatever machine learning we can on the data we collect and analyze to make the product better. With that in light, the effort to not consume data that can be used to identify someone should be strong and on-going. 

If you're going to collect metrics, use them to the very best of your ability. If you don't want an outcome from that data, then spend a bunch of time up front reasoning about how it could ever possibly be used. 

In the event that it becomes what you sought out to not collect, purge it and inform the users.

arnetheduck:
> Imo, we should have EVERY intention of performing whatever machine learning we can on the data we collect and analyze to make the product better. With that in light, the effort to not consume data that can be used to identify someone should be strong and on-going.
> 
> If you're going to collect metrics, use them to the very best of your ability. If you don't want an outcome from that data, then spend a bunch of time up front reasoning about how it could ever possibly be used.
> 
> In the event that it becomes what you sought out to not collect, purge it and inform the users.

That's actually a more fun approach, I'm game - demonstrating possibilities is a powerful way to raise awareness of the potential of metrics analysis in general - we can label it as a social experiment for those that want to enable metrics on their phones

hesterbruikman:
> 
> 
> > Imo, we should have EVERY intention of performing whatever machine learning we can on the data we collect and analyze to make the product better. With that in light, the effort to not consume data that can be used to identify someone should be strong and on-going.
> > If you're going to collect metrics, use them to the very best of your ability. If you don't want an outcome from that data, then spend a bunch of time up front reasoning about how it could ever possibly be used.
> > In the event that it becomes what you sought out to not collect, purge it and inform the users.
> 
> That's actually a more fun approach, I'm game - demonstrating possibilities is a powerful way to raise awareness of the potential of metrics analysis in general - we can label it as a social experiment for those that want to enable metrics on their phones

Both great ideas. It does mean that we need an owner to do exhaustive machine learning analysis. I suspect analyzing the data only with the objective of improving will not surface anything we did not seek out to collect. cc @johnlea-quiup @guylouis

_docs/app-metrics.md
hesterbruikman: Thanks, replaced with 'end-to-end encrypted over Waku'

_docs/app-metrics.md
hesterbruikman: Good point. Added: "Status will never use these data for profit."

_docs/app-metrics.md
shivekkhurana: The batch id is updated each time a user logs in. So it gives us an idea about what the user did in a particular sitting without exposing any identity.

churik:
@hesterbruikman is PR still relevant?

hesterbruikman:
Thanks for checking @churik . I'll figure out how to add a minor edit such that we don't refer to any specific release and make it clear that metrics has not been rolled out.

With those changes, the PR is still relevant. It removes inaccuracies in the previous file. In the off chance that we ever end up including metrics in a dedicated test build, we will want these changes to be merged.

hesterbruikman:
@Samyoul Could you please assist by taking a 2nd look and merging this doc? 

The doc still documents the metrics project and if anyone bumps into it it needs the added disclaimer. Also in the off chance we ever use the tool in dedicated apk builds for user testing we'd need the document updates to be in place.

Samyoul:
Hey @hesterbruikman I've gone through the doc and it looks all fine, I've added links to all the relevant PRs because they have lots of additional information. I'll also fix the PR fails.

Samyoul:
Ok, all these fails are not failing now. 🎉 , so I think this PR is ready for merging

hesterbruikman:
> Ok, all these fails are not failing now. 🎉 , so I think this PR is ready for merging

thank you!

PR #2212, opened 2021-04-23
Ensure timeline chat

cammellos:
If the user is asking for timeline data, we ensure it exists before pulling messages.


protocol/messenger.go
```diff
@@ -3055,7 +3055,16 @@ func (m *Messenger) MessageByChatID(chatID, cursor string, limit int) ([]*common
 		return nil, "", err
```
flexsurfer, line 3055: won't it return error here if no chat ?

flexsurfer (approved):

protocol/chat.go
```diff
@@ -33,7 +33,10 @@ const (
 	ChatTypeCommunityChat
 )
 
-const pkStringLength = 68
+const (
+	timelineChatID = "@timeline70bd746ddcc12beb96b2c9d572d0784ab137ffc774f5383e50585a932080b57cca0484b259e61cecbaa33a4c98a300a"
```
jakubgs, line 38: What is this magical string?

protocol/messenger.go
cammellos: no, returns nil

protocol/chat.go
jakubgs: A comment would be nice explaining the purpose.

cammellos: done, is a magical string basically, can be anything, as long as it's a const

jakubgs (approved):

flexsurfer:
replaced by https://github.com/status-im/status-react/pull/12037

PR #2223, opened 2021-05-05
🪟🗝️ Expose password reset function

shivekkhurana:
API to re-encrypt database using a new password. Will be helpful in preserving the database during multiaccount->keycard migration and to create password less accounts.

Considerations:
- Will this work with accounts before EIP55 ?
- The API re-encrypts the database and keystore files, what happens if one step fails? How to manage errors without data loss?
- This implementation bypasses Status's account generation process and uses Geth directly. There might be some consequences.

Todo:
- Check password security conditions on the backend


corpetty:
I would recommend we have some password strength validation included, following as much of the policy set [here](https://github.com/status-im/status-security/blob/master/drafts/Policy-Authentication.md) as possible.

shivekkhurana:
Make sure this works for accounts that were introduced before [EIP55](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md). There are test cases in `account/accounts_test.go` file.

shivekkhurana:
we are using the `EncryptKey` fn directly from Geth bypassing Status APIs. @gravityblast raised a concern that:
> I'm not sure it's the best way to do it, because we are basically skipping some of our APIs
> so if we change the implementation in the future it can be tricy to find
> but I'm not sure if there's a better way to do it

To be considered when this is ready.

sqlite/sqlite.go
```diff
@@ -137,3 +137,18 @@ func OpenUnecryptedDB(path string) (*sql.DB, error) {
...
 }
+
+func ChangeEncryptionKey(path, key, newKey string) error {
+	db, err := openDB(path, key)
+
+	if err != nil {
+		return err
+	}
+
+	resetKeyString := fmt.Sprintf("PRAGMA rekey = '%s'", newKey)
```
0kok0, line 148: hey @shivekkhurana if a password manager is used to generate newKey which includes control chars, this could lead to a db crash (https://play.golang.org/p/rHetXaO9-9e). This function needs to handle/escape and to be tested for control chars in the newKey

account/accounts.go
```diff
@@ -477,3 +480,100 @@ func (m *Manager) MigrateKeyStoreDir(oldDir, newDir string, addresses []string)
...
+		return reEncryptedKey, fmt.Errorf("convert to crypto json error: %v", e)
+	}
+
+	decryptedKey, e := keystore.DecryptKey(rawKey, pass)
+	if e != nil {
+		return reEncryptedKey, fmt.Errorf("decryption error: %v", e)
+	}
+
+	n := int(cryptoJSON.KDFParams["n"].(float64))
+	p := int(cryptoJSON.KDFParams["p"].(float64))
```
0kok0, line 499: index error if n or p ar not present for some reason. Could you point me to the generator of KDF params?

shivekkhurana: We use this [generator fn](https://github.com/status-im/status-go/blob/develop/account/generator/generator.go#L267) which calls [ImportSingleExtendedKey](https://github.com/status-im/go-ethereum/blob/v1.9.5-status.12/accounts/keystore/keystore.go#L469) and finally [importKey](https://github.com/status-im/go-ethereum/blob/v1.9.5-status.12/accounts/keystore/keystore.go#L522).

Can these `n` and `p` values be different depending on how the account was setup?

sqlite/sqlite.go
shivekkhurana: There is a sanitation process that fixes issues like the one you pointed. I checked with this password `111111""""""` and it worked fine. 

But this is a good point. I'm currently writing a validation function on the backend. It will implement a subset of password conditions pointed out by Corey above.

account/accounts.go
0kok0: Yes, there's plenty of ways to generate BIP44 keys and users might actually import them from unknown sources

shivekkhurana: Added a check on existence of `n` and `p`:

```go
	if cryptoJSON.KDFParams["n"] == nil || cryptoJSON.KDFParams["p"] == nil {
		return reEncryptedKey, fmt.Errorf("Unable to determine `n` or `p`: %v", e)
	}
```
https://github.com/status-im/status-go/pull/2223/files#diff-0a5c50ab0de0200125520596dd297061369d42a646ede3641296789c621961afR495

account/accounts.go
```diff
@@ -477,3 +480,109 @@ func (m *Manager) MigrateKeyStoreDir(oldDir, newDir string, addresses []string)
...
+		tempWritePath := filepath.Join(tempKeyDirPath, fileInfo.Name())
+		e = ioutil.WriteFile(tempWritePath, reEncryptedKey, fileInfo.Mode().Perm())
+		if e != nil {
+			return fmt.Errorf("unable write key file: %v", e)
+		}
+
+		return nil
+	}
+
+	parts := strings.Split(keyDirPath, string(os.PathSeparator))
```
gravityblast: you could also use directly `filepath.Split` and `Join`
https://golang.org/pkg/path/filepath/#Split

api/geth_backend.go
```diff
@@ -459,6 +459,19 @@ func (b *GethStatusBackend) ImportUnencryptedDatabase(acc multiaccounts.Account,
...
+	dbPath := filepath.Join(b.rootDataDir, fmt.Sprintf("%s.db", keyUID))
+	config := b.StatusNode().Config()
+	keyDir := config.KeyStoreDir
+
+	err := b.accountManager.ReEncryptKeyStoreDir(keyDir, password, newPassword)
+	if err != nil {
+		return fmt.Errorf("ReEncryptKeyStoreDir error: %v", err)
+	}
+
+	return appdatabase.ChangeDatabasePassword(dbPath, password, newPassword)
```
gravityblast: do we need to somehow manage the case where the keystore is re-encrypted but it fails to change the db password?

gravityblast (approved):
great job!
I added a couple of comments. 
The only edge cases I see are:

* key re-encrypted but changing db password fails. so the user needs both the 2 passwords but we don't have a way now to recover
* keydir is renamed to backupDir and something happens like the phone turns off.

do you think there's a way to recover those?

account/accounts.go
gravityblast: I found that one of the `geth` commands is using some constants for those values when the user changes the password:

https://github.com/ethereum/go-ethereum/blob/d8ff53dfb8a516f47db37dbc7fd7ad18a1e8a125/cmd/ethkey/changepassword.go#L73-L77

account/accounts.go
```diff
@@ -477,3 +480,109 @@ func (m *Manager) MigrateKeyStoreDir(oldDir, newDir string, addresses []string)
...
+	if err != nil {
+		// if this happens, then the app is probably bricked, because the keystore won't exist anymore
+		// try to restore from backup
+		_ = os.Rename(backupKeyDirPath, keyDirPath)
+		return fmt.Errorf("unable to rename tempKeyDirPath to keyDirPath: %v", err)
+	}
+
+	// remove temp and backup folders and their contents
+	os.RemoveAll(tempKeyDirPath)
+	os.RemoveAll(backupKeyDirPath)
```
0kok0: Nice cleanup of  toxic cryptographic waste! os.RemoveAll can't error?

0xAverageUser (approved):

0kok0 (approved):

api/geth_backend.go
shivekkhurana: It should be handled, I'll update it to reset the password to original password.

account/accounts.go
shivekkhurana: It can, will catch it but I don't think we should throw that error because the process went through. 
I will retry deleting files.

0kok0: Yupp, i wouldn't throw here either but log.error that the user needs to remove the folder manually while they're present

account/accounts.go
shivekkhurana: `filepath.Split` was just what I needed!

shivekkhurana:
@gravityblast 
> key re-encrypted but changing db password fails. so the user needs both the 2 passwords but we don't have a way now to recover

Handled this by changing the password back to original password if db re-encryption fails

> keydir is renamed to backupDir and something happens like the phone turns off.

Maybe. If the folder was not deleted, it can be recovered definitely. But we need some UI to initiate that recovery. 

This PR will be merged but will not go live (will be hidden behind a feature flag). I'll raise this with design team and see if we can have some recovery UI. I also feel that reset password is potentially destructive, so we should warn users to backup seed phrase before reseting password.

PR #2229, opened 2021-05-14
feat: Add edit communities

emizzle:
Allow Communities to be edited, including display name, description, color, image, membership, and permissions.

Important changes:

### DIFF
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -495,7 +495,7 @@ func (o *Community) Edit(description *protobuf.CommunityDescription) {
 	o.config.CommunityDescription.Identity.Description = description.Identity.Description
 	o.config.CommunityDescription.Identity.Color = description.Identity.Color
 	o.config.CommunityDescription.Identity.Images = description.Identity.Images
-	o.config.CommunityDescription.Permissions = description.Permissions
+	o.increaseClock()
 }
 
 func (o *Community) Join() {
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -490,6 +490,14 @@ func (o *Community) BanUserFromCommunity(pk *ecdsa.PublicKey) (*protobuf.Communi
 	return o.config.CommunityDescription, nil
 }
 
+func (o *Community) Edit(description *protobuf.CommunityDescription) {
+	o.config.CommunityDescription.Identity.DisplayName = description.Identity.DisplayName
+	o.config.CommunityDescription.Identity.Description = description.Identity.Description
+	o.config.CommunityDescription.Identity.Color = description.Identity.Color
+	o.config.CommunityDescription.Identity.Images = description.Identity.Images
+	o.config.CommunityDescription.Permissions = description.Permissions
+}
+
 func (o *Community) Join() {
 	o.config.Joined = true
 }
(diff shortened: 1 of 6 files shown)


vkjr:
Few changes:
- added `EditCommunity` request type
- added check that `EditCommunity` invoked by admin
- added publishing community after change

vkjr:
@Samyoul, @cammellos, could you please review?

protocol/communities/community.go
```diff
@@ -490,6 +490,14 @@ func (o *Community) BanUserFromCommunity(pk *ecdsa.PublicKey) (*protobuf.Communi
 	return o.config.CommunityDescription, nil
 }
 
+func (o *Community) Edit(description *protobuf.CommunityDescription) {
+	o.config.CommunityDescription.Identity.DisplayName = description.Identity.DisplayName
+	o.config.CommunityDescription.Identity.Description = description.Identity.Description
+	o.config.CommunityDescription.Identity.Color = description.Identity.Color
+	o.config.CommunityDescription.Identity.Images = description.Identity.Images
+	o.config.CommunityDescription.Permissions = description.Permissions
```
cammellos: I would not allow to edit permissions, as that's likely to require some more logic for it to work

protocol/communities/manager.go
```diff
@@ -173,13 +183,66 @@ func (m *Manager) CreateCommunity(description *protobuf.CommunityDescription) (*
...
+		newDescription.Permissions.Access = community.config.CommunityDescription.Permissions.Access
+	}
+	// If the image wasn't edited, use the existing one
+	// NOTE: This will NOT allow deletion of the community image; it will need to
+	// be handled separately.
+	if request.Image == "" {
+		newDescription.Identity.Images = community.config.CommunityDescription.Identity.Images
+	}
+	// TODO: handle delete image (if needed)
+	newDescription.Clock++
```
cammellos: I don't see clock being copied over/saved in the community, or am I missing something?

cammellos (commented):
Looks good mostly, just a couple of questions

protocol/communities/manager.go
vkjr: Great catch, thanks! Yes, we shouldn't adjust clock on description (because it also has incorrect clock). Instead we can call `increaseClock()` right in a community `Edit()` function.

cammellos (approved):

vkjr:
@iurimatias, any reason to merge this PR? Related mobile and desktop PRs have the same issue with syncing community changes, so it is not fully working yet.

iurimatias:
@vkjr so previously @cammellos had told me that if a PR in status-go had at least one approval I could go ahead and merge it.
Since there was an approval from Andrea himself, and your comments seemed to have been addressed (they say `outdated` and there is commits afterwards), and there were no requests for changes state either, I went ahead and merged it.

iurimatias:
@vkjr ah further more, it seems that unfortunatly I merged that right before or around the time that @anastasiyaig posted that issue, unfortunate timing

vkjr:
@iurimatias, I see, thanks for explanation!
Probably `status-go` PRs need additional check in PR checklist, something like "Tested on the client"

PR #2231, opened 2021-05-18
Add mailserver logic

cammellos:
This PR does a few things:

### Move mailserver logic to status-go

The logic to query mailservers and which topics to query has been moved to status-go.
Before status-react would keep track of the filters that were passed from status-go and query them, handling the response, and creating gaps and ranges. 
The logic has now been simplified and ported to status-go.
Ranges are not used anymore. 
Gaps are now just simple messages.
Both collections are not used anymore and can be removed (in a separate PR).
All the new logic is in `protocol/messenger_mailserver.go`, which handles querying, making sure that timestamps are up to date, create gaps messages and make sure the user allows syncing with mailservers, depending on the connection.

### Fixes some issues with topics management on the non-deprecated request from mailservers

There are currently 2 ways to query mailservers. The deprecated way is the one that status-react used, and it would encrypt the payload with a symmetric key.
The non-deprecated way didn't use a sym key and sent the request in "plain-text" (it's encrypted by dev-p2p), and this was used by the new code querying for communities.
It turned out that though the non-deprecated method did not support querying by topics, and only supported querying by bloom filter, which is likely the reason why requests were taking longer than expected.
This has been fixed, and requires a deployment to prod (it's in staging now).

### Removes some unused mailserver code

This PR also removes some of the old whisper mailserver code, so that we only support waku mailservers.

### Other changes

- Removes some unused whisper code

- Removes some unused nimbus code

Sorry for the large PR but it would be difficult to test it otherwise.


vkjr:
136 changed files, wow!)
@cammellos, could you please provide as detailed info on PR as possible to facilitate understanding during the review. Also, that would be helpful for those who aren't going to read the code but would like to understand the overall changes.

`Move mailserver logic to status-go` - what particular pieces of logic?
`Makes old topics/ranges obsolete, as they are not used anymore` - what changed in code and what should be used now?
`Removes some unused mailserver code` - like what?
`Adds a script to populate the db for testing` - how and when it can be used by other devs?

Thanks!

cammellos:
> 136 changed files, wow!)
> @cammellos, could you please provide as detailed info on PR as possible to facilitate understanding during the review. Also, that would be helpful for those who aren't going to read the code but would like to understand the overall changes.
> 
> `Move mailserver logic to status-go` - what particular pieces of logic?
> `Makes old topics/ranges obsolete, as they are not used anymore` - what changed in code and what should be used now?
> `Removes some unused mailserver code` - like what?
> `Adds a script to populate the db for testing` - how and when it can be used by other devs?
> 
> Thanks!

Yes, good point, I should have done that before :)
I'll fix a few issues with tests and I will provide a suitable description.
Thank you

cammellos:
@vkjr updated the description, thanks!

jakubgs (commented):
Are you sure you want to add that `cmd/node-canary/node-canary` binary?

ghost:
`vendor/github.com/stretchr/objx/Gopkg.lock` changed. Pinging @adambabik and @corpetty

jakubgs (commented):
I've went through most of it and it appears that 90% is just removing or replacing whisper stuff, which looks correct. The other 10% will need some more scrutiny tomorrow.

arnetheduck:
it might make sense to split out the removal of the Nimbus integration to a separate PR as this might come in handy in a not-too-distant future - in general, it is useful for the nimbus team as a reference for what functionality the app is using and removing it is a regression

cammellos:
> it might make sense to split out the removal of the Nimbus integration to a separate PR as this might come in handy in a not-too-distant future - in general, it is useful for the nimbus team as a reference for what functionality the app is using and removing it is a regression

By removing whisper we'll have to change the nimbus integration and that would be extra work, so I'd avoid it (the integration was already broken as not maintained), we can point out at a commit where the integration is in a working state for reference, would that be ok?

arnetheduck:
It depends more on what the core team sees as useful - Ethereum is generally aimed to upgrade to PoS that comes with light sync by the end of the year with eth1 and eth2 merging, and it would make sense that some of that effort is driven from the core app - it's around that time that nimbus-eth1 is expected to near production quality as well - we can actually either use desktop or core to drive the direction of nimbus' integration, whichever is easier. 

A clean PR shows what the integration looked like basically without the mess, and potentially makes it easier for core to rejoin the merge effort when/if it becomes relevant. If it's obsolete and not representative, it's indeed not that useful really.

multiaccounts/accounts/database.go
```diff
@@ -555,6 +555,24 @@ func (db *Database) GetNotificationsEnabled() (bool, error) {
 	return result, err
 }
 
+func (db *Database) GetUseMailservers() (bool, error) {
```
Samyoul: This function returns a bool but starts with `Get`, perhaps this function needs to be renamed into a question. Maybe `CanUseMailservers`?

multiaccounts/accounts/database.go
```diff
@@ -555,6 +555,24 @@ func (db *Database) GetNotificationsEnabled() (bool, error) {
...
+	var result bool
+	err := db.db.QueryRow("SELECT use_mailservers FROM settings WHERE synthetic_id = 'id'").Scan(&result)
+	if err == sql.ErrNoRows {
+		return result, nil
+	}
+	return result, err
+}
+
+func (db *Database) GetSyncingOnMobileNetwork() (bool, error) {
```
Samyoul: This function returns a bool but starts with `Get`, perhaps rename it to something like `CanSyncOnMobileNetwork` ?

protocol/chat.go
```diff
@@ -35,6 +35,9 @@ const (
 
 const pkStringLength = 68
 
+// timelineChatID is a magic constant id for your own timeline
+const timelineChatID = "@timeline70bd746ddcc12beb96b2c9d572d0784ab137ffc774f5383e50585a932080b57cca0484b259e61cecbaa33a4c98a300a"
```
Samyoul, line 39: 🧙

protocol/common/message.go
```diff
@@ -69,6 +69,12 @@ type CommandParameters struct {
 	Signature []byte `json:"signature"`
 }
 
+// GapParameters is the From and To indicating the missing period in chat history
+type GapParameters struct {
+	From uint32 `json:"from,omitempty"`
+	To   uint32 `json:"to,omitempty"`
```
Samyoul, line 75: Are these clock values?

cammellos:
> It depends more on what the core team sees as useful - Ethereum is generally aimed to upgrade to PoS that comes with light sync by the end of the year with eth1 and eth2 merging, and it would make sense that some of that effort is driven from the core app - it's around that time that nimbus-eth1 is expected to near production quality as well - we can actually either use desktop or core to drive the direction of nimbus' integration, whichever is easier.
> 
> A clean PR shows what the integration looked like basically without the mess, and potentially makes it easier for core to rejoin the merge effort when/if it becomes relevant. If it's obsolete and not representative, it's indeed not that useful really.

I think the best reference is probably a link to once the functionality was working, as it has been chipped away/not maintained so the last remaining are not going to be representative, I will find the has and send it over.

protocol/requests/clear_history.go
```diff
@@ -0,0 +1,19 @@
...
+	"errors"
+)
+
+var ErrClearHistoryInvalidID = errors.New("clear-history: invalid id")
+
+type ClearHistory struct {
+	ID string `json:"id"`
+}
+
+func (j *ClearHistory) Validate() error {
```
Samyoul: This received should match the struct name `c *ClearHistory` or `ch *ClearHistory`

protocol/requests/create_profile_chat.go
```diff
@@ -0,0 +1,19 @@
...
+	"errors"
+)
+
+var ErrCreateProfileChatInvalidID = errors.New("create-public-chat: invalid id")
+
+type CreateProfileChat struct {
+	ID string `json:"id"`
+}
+
+func (j *CreateProfileChat) Validate() error {
```
Samyoul: Ah, all these receivers have the same issue

Samyoul (approved):
Looks good, just naming convention things. Well done on all the deletions 🎉

protocol/common/message.go
cammellos: Timestamps actually, although they are fairly similar

PR #2237, opened 2021-05-24
Set timestamp on group chat invitation

cammellos:
Set timestamps on group chat invitations.


protocol/messenger.go
```diff
@@ -1224,6 +1224,11 @@ func (m *Messenger) CreateGroupChatFromInvitation(name string, chatID string, ad
 	chat.Name = name
 	chat.InvitationAdmin = adminPK
 
+	timestamp := uint32(chat.Timestamp / 1000)
```
vkjr: Wondering if we can move these changes into `CreateGroupChat()`?
Or not every created chat has the same initial timestamps logic?

vkjr (approved):

protocol/messenger.go
cammellos: Good point, turns out that we always set them, changing, thanks!

PR #2239, opened 2021-05-24
Add mention notification to activity center

briansztamfater:
This PR adds a notification to activity center when user is mentioned in a group / community chat.
Also, added a `message` prop to the `ActivityCenterNotification` to store the message where the user was mentioned.

### DIFF
--- a/protocol/activity_center.go
+++ b/protocol/activity_center.go
@@ -1,6 +1,7 @@
 package protocol
 
 import (
+	"crypto/ecdsa"
 	"errors"
 
 	"github.com/status-im/status-go/eth-node/types"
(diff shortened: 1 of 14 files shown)


protocol/activity_center.go
```diff
@@ -44,3 +47,23 @@ func (n *ActivityCenterNotification) Valid() error {
...
 }
+
+func showMentionActivityCenterNotification(publicKey ecdsa.PublicKey, message *common.Message, chat *Chat, responseTo *common.Message) bool {
+	if chat != nil && !chat.Active {
+		return false
+	}
+
+	publicKeyString := common.PubkeyToHex(&publicKey)
+	mentioned := false
+	for _, mention := range message.Mentions {
```
cammellos: We changed the code, I believe now you can just check for:
`message.Mentioned`

cammellos (approved):
looks good, just a minor comment, well done!

protocol/activity_center.go
briansztamfater: Good to know, will change it :)

PR #2242, opened 2021-05-26
Add unviewed mentions count

cammellos:
This commit add a field to count the unviewed mentions in a given chat.
The usage pattern is identical to `UnviewedMessagesCount`


protocol/message_handler.go
```diff
@@ -233,6 +233,9 @@ func (m *MessageHandler) handleCommandMessage(state *ReceivedMessageState, messa
 	// Increase unviewed count
 	if !common.IsPubKeyEqual(message.SigPubKey, &m.identity.PublicKey) {
 		chat.UnviewedMessagesCount++
+		if message.Mentioned {
```
vkjr: Maybe we can extract similar pieces of work with unviewed messages and notifications from `handleCommandMessage` and `HandleChatMessage`?

vkjr (approved):

PR #2246, opened 2021-05-27
feat: edit messages

richard-ramos:
Adds the required functionality to edit a message.

To edit an already sent message, a new message must be sent with the `replace` attribute containing the message to edit, and the content type must be EDIT. You can only edit your own messages.


protocol/messenger_edit_message_test.go
```diff
@@ -0,0 +1,150 @@
...
+	_, err := theirMessenger.Start()
+	s.Require().NoError(err)
+
+	theirChat := CreateOneToOneChat("Their 1TO1", &s.privateKey.PublicKey, s.m.transport)
+	err = theirMessenger.SaveChat(theirChat)
+	s.Require().NoError(err)
+
+	inputMessage := buildTestMessage(*theirChat)
+
+	// TODO: send edit before original message
```
richard-ramos: @Samyoul @cammellos:
For this test I need that `s.m` receives an edited message before receiving the original message. 
How can I achieve this?

richard-ramos:
For some reason the CI is failing with  `protocol/messenger.go:2064:13: sender.Equal undefined (type *ecdsa.PublicKey has no field or method Equal)`
even tho that function is available according to https://golang.org/pkg/crypto/ecdsa/#PublicKey.Equal

protocol/message_handler.go
```diff
@@ -684,6 +684,31 @@ func (m *MessageHandler) HandleChatMessage(state *ReceivedMessageState) error {
...
+		}
+
+		m.logger.Debug("Handling edit content type")
+
+		originalMessage, err := m.persistence.MessageByID(receivedMessage.Replace)
+
+		if err != nil {
+			// if the OG message has not been received yet, do nothing
+			if err == common.ErrRecordNotFound {
+				state.Response.pendingEdits = append(state.Response.pendingEdits, receivedMessage)
```
cammellos: I think you can just save it directly here, instead of adding it to response, as the client does not need to be notify regarding pending edits.

cammellos (commented):
@richard-ramos we might be running an older version of ecdsa/go?

For the test, the easiest way is to use the private methods of `messenger` probably, instead of sending the message, you can call directly `HandleChatMessage` on `s.m.handler` with the some data you pass in

I left a couple of comments as well, sorry if it's too early to review. 
Also we need to validate the user can modify a message (so we need to make sure the author of the original message is the same author of the edited one)

richard-ramos:
I attempted to implement a solution to edit messages being received before the original message, as well as checking if the edit message received was edited by the author in status-go and found it to be difficult. Ended up implementing it in the frontend here: https://github.com/status-im/status-desktop/blob/fb1340011b8181f955d1065076333493948d3ce9/src/app/chat/view.nim#L466-L506

Here's a video of the functionality for editing messages:

https://user-images.githubusercontent.com/1106587/120389879-02885100-c2fb-11eb-85e5-3ae8d8154efb.mp4

https://github.com/status-im/status-desktop/pull/2625/

cammellos:
> I attempted to implement a solution to edit messages being received before the original message, as well as checking if the edit message received was edited by the author in status-go and found it to be difficult. Ended up implementing it in the frontend here: https://github.com/status-im/status-desktop/blob/fb1340011b8181f955d1065076333493948d3ce9/src/app/chat/view.nim#L466-L506
> 
> Here's a video of the functionality for editing messages:
> 
>  editing-messages-2021-06-01_16.52.14.mp4 
> [status-im/status-desktop#2625](https://github.com/status-im/status-desktop/pull/2625)

Thanks!

We'll probably try to write the logic in status-go ourselves for performance reasons, do you want us to take over this PR?

richard-ramos:
Replaced by https://github.com/status-im/status-go/pull/2267

PR #2249, opened 2021-05-28
Allow setting mailserver sync value

cammellos:
This commit adds a setting for mailserver, that allows the client to
specify the syncing period.
Also fixes an issue with public chats and `DeletedAtClockValue`


protocol/messenger_chats.go
```diff
@@ -67,7 +68,11 @@ func (m *Messenger) CreatePublicChat(request *requests.CreatePublicChat) (*Messe
 
 	// We set the synced to, synced from to the default time
 	if !willSync {
-		timestamp := uint32(m.getTimesource().GetCurrentTime()/1000) - defaultSyncInterval
+		defaultSyncPeriod, err := m.settings.GetDefaultSyncPeriod()
```
vkjr: I'd suggest extracting this piece of code to function like `InitChatSyncFields()`or something like that because we already use it 4 times.

protocol/messenger_mailserver.go
```diff
@@ -170,17 +167,24 @@ func (m *Messenger) syncChat(chatID string) (*MessengerResponse, error) {
...
+	if err != nil {
+		return 0, err
+	}
+	return uint32(m.getTimesource().GetCurrentTime()/1000) - defaultSyncPeriod, nil
 }
 
 // calculateSyncPeriod caps the sync period to the default
-func (m *Messenger) calculateSyncPeriod(period uint32) uint32 {
-	d := m.defaultSyncPeriod()
+func (m *Messenger) calculateSyncPeriod(period uint32) (uint32, error) {
```
vkjr: Maybe lets rename it to `capToDefaultSyncPeriod()` for better readability, wdyt?

protocol/messenger_mailserver.go
```diff
@@ -170,17 +167,24 @@ func (m *Messenger) syncChat(chatID string) (*MessengerResponse, error) {
 	return m.syncFilters(filters)
 }
 
-func (m *Messenger) defaultSyncPeriod() uint32 {
-	return uint32(m.getTimesource().GetCurrentTime()/1000) - defaultSyncInterval
+func (m *Messenger) defaultSyncPeriod() (uint32, error) {
```
vkjr: It is a bit confusing that Messenger's `defaultSyncPeriod()` returns different value than settings' `GetDefaultSyncPeriod()`. I'd suggest to rename to `defaultSyncPeriodBackFromNow()`.. or any less clumsy name that would tell there is a difference :)

vkjr (approved):

PR #2250, opened 2021-06-02
add PinnedBy to PinnedMessage

### LINKED ISSUE #2234: Pinned message `from` is the message author instead of the person who pinned
When receiving the pinned message in real time using the signal, the `from`  value is correctly the pub key of the person who pinned, but when using `chatPinnedMessages`, the returned messages don't have the right `from` value. In that case,  the `from` is the person who originally posted the message.

Either the `from` should be the person who pinned when calling `chatPinnedMessages` or either we should add a new property called something like `pinnedBy`.

Here is an example of a message that was pinned by account2, but the `from` shows the OP (account 1)
```
"messages": [
      {
        "id": "0x016bda8ebfb9c6f410b31f34c28aa01defdc7e7cb87dc5551e4dd61a326b34b6",
        "whisperTimestamp": 1621531275853,
        "from": "0x04c56a973a4082e0586837f955d4a09725007ab2148d6eb51484b5f8c32bb67a15f7866ae671a612a73df5ca936070f204af7a33f99e876cb38b20a749a196fb0e",
        "alias": "Oblong Circular Ling",
        "identicon": "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAADIAAAAyCAYAAAAeP4ixAAAAiklEQVR4nOzYwQmEQBAF0d1lcxHzMwzzE6PRuzjS4IjFp95xmEvRl6Z/nxCG0BhCYwiNITQxIf/qx3WZt7P3YZy+b/w/ipmIITSG0BhCYwiNITTlXaul1051V8xEDKExhMYQGkNoDKEp3Yyu9NqdqverlpiJGEJjiB4SMxFDaAyhMYTGEJo9AAD//zoXF6FykVuGAAAAAElFTkSuQmCC",
        "seen": true,
        "outgoingStatus": "delivered",
        "quotedMessage": null,
        "rtl": false,
        "parsedText": [
          {
            "type": "paragraph",
            "children": [
              {
                "literal": "new"
              }
            ]
          }
        ],
        "lineCount": 0,
        "text": "new",
        "chatId": "0x04b133e7b1d7617665c6f897dff88e9d83bd582270085f5260e09bb60aa8c5ad25eea384d599de155d33c479dbbb88bef0c39d31e304f68c792ede7d3b7eecb706",
        "localChatId": "0x04b133e7b1d7617665c6f897dff88e9d83bd582270085f5260e09bb60aa8c5ad25eea384d599de155d33c479dbbb88bef0c39d31e304f68c792ede7d3b7eecb706",
        "clock": 1621531275853,
        "replace": "",
        "responseTo": "",
        "ensName": "",
        "sticker": null,
        "commandParameters": null,
        "timestamp": 1621531275853,
        "contentType": 1,
        "messageType": 1
      }
  ]
  ```

### PR DESCRIPTION
**Pull request by gravityblast**
Add a `PinnedBy` attribute to the `PinMessage` struct.

Closes #2234

### CODE REVIEW

**Review by cammellos — APPROVED** [review]

**Review by jrainville — COMMENTED** [review]
Looks good. I'll try to test it in desktop today

**Comment by jrainville** [comment]
I just tested and `pinnedBy` is not set in `chatPinnedMessages` nor the signal. Maybe it was supposed to be in the `from`, but the `from` is still the author of the message

**Comment by gravityblast** [comment]
@jrainville @briansztamfater I fixed it with the last commits. The problem was that the `PinnedMessage` struct actually had the new attribute, but the embedded struct generated by protobuf overrides the marshal method adding only its fields.
This is visible only from the public API that is not testable now, but for sure we need to find a way to test those parts as well.
To avoid hacks on that struct I updated the response to be like this:

```
{
    "pinnedMessages": [
        {
            "message": {
                "id": "...",
                ...
            },
            "pinnedAt": 1622667109565,
            "pinnedBy": "0x..."
        }
    ],
    "cursor": ""
}
```

/cc @cammellos if you want to review it again after this changes

jrainville:
Just letting you know it works now!
It is a breaking change for normal message parsing, but it was a simple fix on our side

gravityblast:
@jrainville are we still testing it or can we merge it?

Samyoul (approved):

jrainville:
@gravityblast It's all good on our end. Merge as soon as you can!

PR #2252, opened 2021-06-02
fix order of migration

cammellos:
Last migration included had a timestamp that was older than the latest migration in the develop branch.
We should probably have a script that prevents pushing/commit if that's the case, as it's really hard to notice when rebasing.


vkjr (approved):

vkjr:
@cammellos, did it introduce any problems? I thought that if changes are independent, it should be ok to migrate in any order?
Oh... it probably breaks already existing databases?

cammellos:
@vkjr it does not break the database, but the migration is skipped, so querying will fail as the field is not there

PR #2253, opened 2021-06-02
Sync Communities

Samyoul:
## What's Changed

Add functionality for syncing communities across a user's paired devices.

## Why Make the change?

As part of a wider project to increase the things that are synced, this PR add communities to the list of things that are now synced, see https://github.com/status-im/status-react/issues/12145

Full e2e testing here https://github.com/status-im/status-react/pull/12248


Samyoul:
Needs update to handle muted functionality https://github.com/status-im/status-go/pull/2271

gravityblast (approved):

gravityblast:
LGTM

I often see the migrations and bindata files changing a little bit, maybe based on the version of the cli used, is there a way to lock it to a specific version ? or maybe we can just be sure we use the same one, but when I did it it changed even more things.

Samyoul:
Awaiting results of https://github.com/status-im/status-react/pull/12248

Samyoul:
QA e2e tests pass https://github.com/status-im/status-react/pull/12248#issuecomment-892489957 🎉

protocol/communities/community.go
```diff
@@ -167,7 +168,7 @@ func (o *Community) Name() string {
 	return ""
 }
 
-func (o *Community) Description() string {
+func (o *Community) DescriptionText() string {
```
cammellos, line 172: Good change!

protocol/communities/persistence.go
```diff
@@ -81,6 +112,61 @@ func (p *Persistence) JoinedCommunities(memberIdentity *ecdsa.PublicKey) ([]*Com
...
+		// Community specific fields
+		var publicKeyBytes, privateKeyBytes, descriptionBytes []byte
+		var joined, verified, muted bool
+
+		// Request to join specific fields
+		var rtjID, rtjCommunityID []byte
+		var rtjPublicKey, rtjENSName, rtjChatID sql.NullString
+		var rtjClock, rtjState sql.NullInt64
+
+		err := rows.Scan(
```
cammellos: I think here you might be shadowing the original `err`.
So if this `err` is not nil, but `err` above is nil, it will be overwritten by line 135 I think,
so we should probably have:

```
err = rows.Scan(
```
so that we don't shadow it

protocol/messenger.go
```diff
@@ -1791,13 +1791,19 @@ func (m *Messenger) ReSendChatMessage(ctx context.Context, messageID string) err
 }
 
 func (m *Messenger) hasPairedDevices() bool {
+	logger := m.logger.Named("hasPairedDevices")
```
cammellos, line 1794: Probably left overs :)

protocol/messenger.go
```diff
@@ -2280,6 +2282,42 @@ func (m *Messenger) syncContact(ctx context.Context, contact *Contact) error {
 	return m.saveChat(chat)
 }
 
+func (m *Messenger) syncCommunity(ctx context.Context, community *communities.Community) error {
+	logger := m.logger.Named("syncCommunity")
+	logger.Debug("fired")
```
cammellos: some leftovers

cammellos (approved):

PR #2255, opened 2021-06-04
Upsert one to one chat on mute

cammellos:
Before, status-react would make sure a chat existed before muting, now
the logic is all in the endpoint.


protocol/messenger.go
```diff
@@ -3112,12 +3112,28 @@ func (m *Messenger) MarkAllRead(chatID string) error {
...
 // on new messages from this chat
 func (m *Messenger) MuteChat(chatID string) error {
+	// Only one to one chan be muted
+	publicKey, err := common.HexToPubkey(chatID)
+	if err != nil {
+		return err
+	}
+
 	chat, ok := m.allChats.Load(chatID)
 	if !ok {
```
vkjr, line 3116: Wondering why chat absence is not an error during muting? Is it some case for synced devices?

vkjr (approved):

protocol/messenger.go
cammellos: No, it's the case where you for example see someone in a public chat, and click on profile, and decide to mute them, without opening a chat with them.
In this case, the chat is not created, but we need to create on and mute it.

vkjr: Got it, thanks!

PR #2257, opened 2021-06-09
Add replies to activity center

jrainville:
Adds the replies to the activity center as a new type that is very similar (basically the same data) as Mentions.

I decided to still make them separate types because the front-end can handle them differently (with a filter for example), so it's good to know that it's not exactly the same type.

I reused and modified the `showMentionActivityCenterNotification` so that it is now called `showMentionOrReplyActivityCenterNotification` and returns the type of notification it detected. If it returns -1, it is not a notification (aka `false` in the old function)


protocol/activity_center.go
```diff
@@ -48,14 +49,19 @@ func (n *ActivityCenterNotification) Valid() error {
...
+		return ActivityCenterNotificationTypeMention
 	}
 
-	return false
+	publicKeyString := common.PubkeyToHex(&publicKey)
+	if responseTo != nil && responseTo.From == publicKeyString {
+		return ActivityCenterNotificationTypeReply
+	}
+
+	return -1
```
cammellos: In this case, probably you want to use `0`, and make it a constant, above in the  `iota` block, you can change it to:
```
const (	
ActivityCenterNotificationNoType = iota
ActivityCenterNotificationTypeNewOneToOne	
ActivityCenterNotificationTypeNewPrivateGroupChat
```

But probably at this point you could also return multiple values, like `(bool,AcitvityCenterType)` and return 0 if `bool` is `false`, probably more idiomatic, though not sure, maybe @Samyoul  has an opinion on the matter

protocol/activity_center.go
```diff
@@ -48,14 +49,19 @@ func (n *ActivityCenterNotification) Valid() error {
 
 }
 
-func showMentionActivityCenterNotification(publicKey ecdsa.PublicKey, message *common.Message, chat *Chat, responseTo *common.Message) bool {
-	if chat != nil && !chat.Active {
-		return false
+func showMentionOrReplyActivityCenterNotification(publicKey ecdsa.PublicKey, message *common.Message, chat *Chat, responseTo *common.Message) int {
+	if chat == nil || !chat.Active || (!chat.CommunityChat() && !chat.PrivateGroupChat()) {
```
cammellos, line 54: Just wondering, why excluding private group chats or community chats?

cammellos (commented):
Overall looks good! just a couple of questions on idiomatic usage and functionality

protocol/activity_center.go
jrainville: I reused the same condition Brian added and put it at he start.

Unless I messed up, it should be that we only add mentions and replies from group chats and community chats.

cammellos: oh I missed that, I think we should support all kind of chats probably, I'd leave it out for now, and unless something is not working we can change it

protocol/activity_center.go
jrainville: Updated. Thanks for the comment

cammellos (approved):

protocol/activity_center.go
jrainville: > I think we should support all kind of chats probably

Yeah I think so too. We can discuss it with the rest of the team and change it.

protocol/activity_center.go
Samyoul: This is a fine approach.

Idiomatically enums are set with `iota + 1` because a nil int == `0`, but in this case `ActivityCenterNotificationNoType` is effectively the nil state, so the approach is good.

protocol/activity_center.go
```diff
@@ -15,9 +15,11 @@ import (
 type ActivityCenterType int
 
 const (
-	ActivityCenterNotificationTypeNewOneToOne = iota + 1
+	ActivityCenterNotificationNoType = iota
+	ActivityCenterNotificationTypeNewOneToOne
 	ActivityCenterNotificationTypeNewPrivateGroupChat
 	ActivityCenterNotificationTypeMention
+	ActivityCenterNotificationTypeReply
 )
```
Samyoul: I see that these consts are used as enums of the `ActivityCenterType` type but there is no type assertion on the consts.

Easily implemented though, see below:

```go
const (
	ActivityCenterNotificationNoType ActivityCenterType = iota
	ActivityCenterNotificationTypeNewOneToOne
	ActivityCenterNotificationTypeNewPrivateGroupChat
	ActivityCenterNotificationTypeMention
	ActivityCenterNotificationTypeReply
)
```

Samyoul (approved):
Looks good, only a tiny issue that seems to have existed in the code before this PR

PR #2258, opened 2021-06-10
go-waku: RequestAllHistoricMessages

richard-ramos:


protocol/transport/transport.go
```diff
@@ -405,35 +405,46 @@ func (t *Transport) cleanFiltersLoop() {
...
-) (cursor []byte, err error) {
-
-	r := createMessagesRequest(from, to, previousCursor, topics)
-
-	events := make(chan types.EnvelopeEvent, 10)
-	sub := t.waku.SubscribeEnvelopeEvents(events)
-	defer sub.Unsubscribe()
+) (cursor []byte, storeCursor *types.StoreRequestCursor, err error) {
+	r := createMessagesRequest(from, to, previousCursor, previousStoreCursor, topics)
+	if t.waku.Version() == 2 {
```
cammellos: I would probably split the two implementation completely, at the cost of some code duplication, so have two different toplevel methods (RequestMessagesV2/RequestMessagesV1), to avoid increased complexity, though not sure what's best, 
what do you think? Maybe @vkjr or @Samyoul have some opinion

vkjr: I also would prefer different implementations here because we have only `r := createMessagesRequest()` line in common

Samyoul: Yes, I agree. A different implementations per version. Increased divergence between V1 and V2 will lead to more conditional complexity in the code.

iurimatias (approved):

iurimatias:
we tested this on Desktop to confirm it was confirm (works with wakuv2.test fleet), also still works with v1 as expected

PR #2259, opened 2021-06-16
feat: add invited_by to the notification in the activity center

jrainville:
Adds `invited_by` to the notification in the activity center.

It is used to know who invited you to a group chat.


protocol/message_handler.go
```diff
@@ -192,6 +192,7 @@ func (m *MessageHandler) createMessageNotification(chat *Chat, messageState *Rec
 		Name:        chat.Name,
 		LastMessage: chat.LastMessage,
 		Type:        notificationType,
+		InvitedBy:   messageState.CurrentMessageState.Contact.ID,
```
cammellos: It needs to be a `sql.NullString` because on existing rows it will be null, and will throw an error on migrated devices.

cammellos (changes requested):
Looks all good, apart from the `NullString`
I am not sure `InvitedBy` is the best name, as it makes sense mostly on private group chats, maybe `From` or `Author` is a bit more generic, but `InvitedBy` works as well.

protocol/message_handler.go
jrainville: Done. I renamed and added the NullString here https://github.com/status-im/status-go/pull/2259/commits/20444ceab16451f2523e878eb6b19d2295624d5a#diff-6d77ca93a9f35007250fed5af54fb5916c180cda472c625f1a7fb68ea5e7630aR71
Let me know if I did it correctly

cammellos (approved):
Looks good, thanks!

PR #2264, opened 2021-06-22
Refactoring: rename MessageProcessor to MessageSender

vkjr:
I noticed that class names `MessageProcessor` and `MessageHandler` confuse me as a code reader. They both used in `Messenger` and it is not fully clear what is a functional difference between `processing` and `handling`. 
But In fact `MessageProcessor` deals with messages sending and `MessageHandler` - with messages receiving.

So this small refactoring renames `MessageProcessor` -> `MessageSender`. That should make the difference from `handler` more intuitive.


cammellos (approved):

vkjr:
@cammellos, wondering if we can merge this since very simple renaming was done and it shouldn't affect any functionality. And the issue with ios build in status-react [PR](https://github.com/status-im/status-react/pull/12242) is known.

cammellos:
@vkjr yes, no need to test simple refactors like this, even if there's an issue they are simple to fix

PR #2266, opened 2021-06-23
feat: add api to delete a community channel

jrainville:
Adds the API and manager function to delete a community channel

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -341,6 +341,35 @@ func (m *Manager) EditChat(communityID types.HexBytes, chatID string, chat *prot
 	return community, changes, nil
 }
 
+func (m *Manager) DeleteChat(communityID types.HexBytes, chatID string) (*Community, *protobuf.CommunityDescription, error) {
+	community, err := m.GetByID(communityID)
+	if err != nil {
+		return nil, nil, err
+	}
+	if community == nil {
+		return nil, nil, ErrOrgNotFound
+	}
+
+	// Remove communityID prefix from chatID if exists
+	if strings.HasPrefix(chatID, communityID.String()) {
+		chatID = strings.TrimPrefix(chatID, communityID.String())
+	}
+	description, err := community.DeleteChat(chatID)
+	if err != nil {
+		return nil, nil, err
+	}
+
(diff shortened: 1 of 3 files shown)


cammellos:
@jrainville there's a missing functionality I believe,
we need to notify the user that chats have been removed (in status-go).

For example when we leave a community, we add the removed chats in https://github.com/status-im/status-go/blob/8ac498a5fec39a9e4cfc3df52fe79a05843a7c68/protocol/messenger_communities.go#L367 

Similarly, we want to notify the user that a chat has been removed if the user has joined the chat,
probably the best place to do that is here:
https://github.com/status-im/status-go/blob/8ac498a5fec39a9e4cfc3df52fe79a05843a7c68/protocol/messenger_communities.go#L744

What do you think?

jrainville:
@cammellos I just tried and the channel gets deleted in the other app as well. I know I put the `AddRemovedChat` already. Maybe the other line is not needed?

protocol/communities/manager.go
```diff
@@ -334,6 +334,30 @@ func (m *Manager) EditChat(communityID types.HexBytes, chatID string, chat *prot
...
 
+func (m *Manager) DeleteChat(communityID types.HexBytes, chatID string) (*Community, *protobuf.CommunityDescription, error) {
+	community, err := m.GetByID(communityID)
+	if err != nil {
+		return nil, nil, err
+	}
+	if community == nil {
+		return nil, nil, ErrOrgNotFound
+	}
+	description, err := community.DeleteChat(chatID)
```
richard-ramos, line 345:
```
// Remove communityID prefix from chatID if exists
if strings.HasPrefix(chatID, communityID.String()) {
	chatID = strings.TrimPrefix(chatID, communityID.String())
}
```

richard-ramos, line 357:
```suggestion

        // Remove communityID prefix from chatID if exists
	if strings.HasPrefix(chatID, communityID.String()) {
		chatID = strings.TrimPrefix(chatID, communityID.String())
	}

	description, err := community.DeleteChat(chatID)
```

richard-ramos (approved):

iurimatias:
tested in https://github.com/status-im/status-desktop/pull/2757

PR #2267, opened 2021-06-23
Edit messages

cammellos:
Add edit messages functionality.

It's a separate message type, only used to edit text messages.

It also begins to move away from having a separate message processor, as effectively they were sharing state and it was just a bit clunky.


protocol/edit_message.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/status-im/status-go/protocol/protobuf"
+)
+
+// EditMessage represents an edit of a message from a user in the application layer, used for persistence, querying and
+// signaling
+type EditMessage struct {
+	protobuf.EditMessage
+
+	ID string `json:"id",omitempty"`
```
vkjr: Does edit messages has own ids counter or this is an id of original message that should be edited? Probably it worth mentioning in comments

protocol/edit_message.go
```diff
@@ -0,0 +1,49 @@
+package protocol
```
vkjr, line 1: Currently two classes implementing `ChatEntity` interfaces located in `protocol/common` folder (Message, PinMessage). And two in `protocol` (EmojiReaction, EditMessage). Maybe we'd better group them?

protocol/message_handler.go
```diff
@@ -31,31 +28,10 @@ const (
 
 var ErrMessageNotAllowed = errors.New("message from a non-contact")
 
-type MessageHandler struct {
```
vkjr, line 34: I suggest renaming file `message_handler` -> `messenger_handler` to align with other files working on `Messenger` struct

protocol/messenger.go
```diff
@@ -1296,8 +1292,8 @@ func (m *Messenger) RemoveMemberFromGroupChat(ctx context.Context, chatID string
 	chat.updateChatFromGroupMembershipChanges(group)
 
 	response.AddChat(chat)
```
vkjr: Looking at number of same edits in different placers, seems that this actions sequence can benefit from being in a separate function:
```
reponse.AddChat()
reponse.AddMessages()
m.persistence.SaveMessages()
m.saveChat()
```

protocol/messenger_messages.go
```diff
@@ -0,0 +1,108 @@
+package protocol
```
vkjr, line 1: Will there be more functions moved to this file? Otherwise `messenger_messages.go` name sounds counterintuitive. Since there are only a couple of functions related particularly to editing

vkjr:
Great PR, I love the refactorings! :)

richard-ramos (approved):

Samyoul (approved):

protocol/messenger_messages.go
cammellos: there will be, yes, I haven't moved all of them to avoid too many changes

vkjr (approved):

PR #2274, opened 2021-07-02
send/receive ChatID in TransactionCommand messages

### LINKED ISSUE #10994: Creating "myself" 1-1 chat in Activity center on a synchronized device after sending address request in 1-1 chat
# Bug Report

## Problem

If send ETH to any other user from Device 1, then on paired Device 2 will be created a 1-1 chat with “myself” (with the same account from which the Address request was sent). 

#### Expected behavior

"Myself" 1-1 chat on a synchronized device isn't created.

#### Actual behavior

ezgif com-resize (2)

### Reproduction

Device 1:
- Sync with Device 2
- Open 1-1 chat with any other user -> Send transaction
- Enter valid ETH amount and send Address request
- Check chats on Device 2

### Additional information

- Status version: Nightly 24/07
- Operating System: Android 8, 10

bitgamma:
Fixes https://github.com/status-im/status-react/issues/10994


protocol/messenger_handler.go
```diff
@@ -687,7 +687,7 @@ func (m *Messenger) HandleRequestAddressForTransaction(messageState *ReceivedMes
 			Clock:       command.Clock,
 			Timestamp:   messageState.CurrentMessageState.WhisperTimestamp,
 			Text:        "Request address for transaction",
-			ChatId:      contactIDFromPublicKey(&m.identity.PublicKey),
+			ChatId:      messageState.CurrentMessageState.Message.GetChatId(),
```
cammellos: Looks good, I would only leave a comment on both that validation of `ChatId` is done in `handleCommandMessage`

cammellos (approved):
Looks good, just left a comment, thanks!

protocol/messenger_handler.go
bitgamma: thanks, added the comment.

bitgamma:
@cammellos unfortunately the initial fix was not enough to solve the problem. With the initial fix, the 1-to-1 chat to yourself is indeed not created but since the outgoing message didn't have any chat id set, the message was simply not synced. I now added chat id to outgoing commands and will test with status-react if this fixes the issue completely.

This means however that if the message originated from a non-fixed app, the message will not be synced (it will be lost because of no chat id). Probably not a big deal and slightly better than having it synced as a transaction/request to yourself.

cammellos (approved):
looks good!

PR #2275, opened 2021-07-09
Upgrade geth to 1.10.4

cammellos:
This commit upgrades geth to 1.10.4, in order to get ready for the London fork.

Because of the many changes in geth, there are quite a few changes required:

- Services have changed interface, and are now Lifecycles
- I have moved all the services in StatusNode, as before they were scattered around between status-node,geth_backend and geth_node
- Upgraded a bunch of libraries (go-ens, iconv)
- Removed some unused ens contract code, only used in testing
- Removed mailserver registry code, not used anymore
- Fixed some upgrade issues
- Upgraded to the new london signer
- Added a populate db script from a different PR, useful for testing
- Removed e2e tests, as most of them seems to be testing node usage outside of mobile/desktop

Because of the time constraints I had to cut some corners and remove some of the tests.

A lot of it was testing blockchain synchronization, which we don't use, or using other functionality (status api), that we never used.


node/rpc.go
```diff
@@ -0,0 +1,30 @@
+package node
+
+import (
+	"reflect"
+	"unicode"
+)
+
+// formatName converts to first character of name to lowercase.
+func formatName(name string) string {
```
Samyoul: I think that this function could have a more descriptive name, though I personally appreciate the mystery of `formatName`

```suggestion
func firstCharToLower(name string) string {
```

peers/cotopicpool.go
```diff
@@ -65,7 +66,14 @@ func (t *cacheOnlyTopicPool) ConfirmAdded(server *p2p.Server, nodeID enode.ID) {
...
 	if peer, ok := t.connectedPeers[nodeID]; ok {
-		t.removeServerPeer(server, peer)
+		// NOTE: removeServerPeer removes the server peer immediately.
+		// which means the next discovery.summary is not going to include
+		// the peer.
+		// We leave some time so that we ensure the signal is propagated
+		go func() {
+			time.Sleep(200)
+			t.removeServerPeer(server, peer)
+		}()
```
Samyoul, line 76: 🐱‍💻 Hacker cat approves of your hacks

Samyoul (approved):
Looks really good. Great job.

PR #2276, opened 2021-07-10
feat: user status

richard-ramos:
**StatusMessage** spec
Status messages are used to indicate to other nodes the current status of an user account. Clients are expected to use the status_type to visually represent the status of the users that belong to a community, or the status of an added contact. Devices receiving a status message from another paired device should update their own status if the received message has a newer clock field value.
```
message StatusMessage {
  uint64 clock = 1
  
  StatusType status_type = 2;
    
  string custom_text = 3
  
  enum StatusType {
    UNKNOWN_STATUS_TYPE = 0;
    ONLINE = 1;
    DO_NOT_DISTURB = 2;
  }
}
```
Clients MUST specify clock, status_type, and optionally custom_emoji and custom_text. For better developer experience this specifications RECOMMENDS creating a function that a user would call to set their current status, and then, the client will periodically send a status message every 5 minutes containing the latest user's status. The custom_emoji and custom_text are optional values that can be used to extend the user status with extra content with a free form short messaje and Unicode emoji

Clients MUST sanitize the payload before accessing its content, in particular the `custom_text` field

Status messages are sent into specific topics:
- With communities, a topic derived from the community-id with the suffix -ping concatenated.
- To know the status of a contact, their updates will be sent in their "contact-code" topic

A filter is created in the clients using the topics created for a community and contacts to subscribe to the status messages. 

Since these messages can be thought as the equivalent of a heartbeat message, clients can visually change the representation for the online status depending on the time that has passed since the last time a status message has been received for some user account, and mark a user as offline if the user has not sent a status message in a considerable amount of time

**Signal**
`messages.new` has new `event` attributes: 
- `statusUpdates` containing the latest contacts and community member status updates
- `currentStatus` which is included if the current user status changes in another paired device

**Settings**
Two new settings are expected to be created on the user's database setting table:
- `send-status-updates?`, a boolean value indicating whether the status will be sent or not. This feature is opt-out, so the default value is true
- `current-user-status`, an object containing the latest user status

**RPC methods**
- `wakuext_setUserStatus`: with the payload `[status int, customText string]`. This is used to change the user status. The `customText` parameter is an optional string that should not exceed 140 characters


protocol/messenger_status_updates.go
```diff
@@ -0,0 +1,247 @@
...
+			case <-time.After(5 * time.Minute):
+				m.sendCurrentUserStatus()
+			case <-m.quit:
+				return
+			}
+		}
+	}()
+}
+
+func (m *Messenger) SetUserStatus(newStatus int, newCustomText string) error {
```
alaibe: Don't you want to also save it in this case?

protocol/messenger_status_updates.go
```diff
@@ -0,0 +1,247 @@
...
+		return err
+	}
+
+	joinedCommunities, err := m.communitiesManager.Joined()
+	if err != nil {
+		return err
+	}
+	for _, community := range joinedCommunities {
+		rawMessage.LocalChatID = community.StatusUpdatesChannelID()
+		_, err = m.sender.SendPublic(context.Background(), rawMessage.LocalChatID, rawMessage)
```
pinebit: you create many background contexts here in this function.
since they all belong to the same request, it shall be the same context instance passed as a function parameter.
this, in particular, ensures better unit testing.

protocol/persistence.go
```diff
@@ -843,3 +843,60 @@ func (db sqlitePersistence) SaveWhenChatIdentityLastPublished(chatID string, has
...
+	)
+
+	return err
+}
+
+func (db sqlitePersistence) CleanOlderStatusUpdates() error {
+	now := time.Now()
+	oneHourAgo := now.Add(time.Duration(-1) * time.Hour)
+	_, err := db.db.Exec(`DELETE FROM status_updates WHERE clock < ?`,
+		int(oneHourAgo.Unix()),
```
pinebit: should be `uint64(...)` for consistency

pinebit (commented):
Added a couple of comments. Overall this looks great.

protocol/messenger_status_updates.go
```diff
@@ -0,0 +1,247 @@
...
+
+	return m.sendUserStatus(*currStatus)
+}
+
+func (m *Messenger) HandleStatusUpdate(state *ReceivedMessageState, statusMessage protobuf.StatusUpdate) error {
+	if err := ValidateStatusUpdate(statusMessage); err != nil {
+		return err
+	}
+
+	currentStatus, err := m.GetCurrentUserStatus()
```
cammellos: This looks like can go inside the next `if` block, so we save a database query in the most likely scenario (i.e the status is not coming from us)

protocol/messenger_status_updates.go
```diff
@@ -0,0 +1,247 @@
...
+	if err != nil {
+		m.logger.Debug("Error obtaining latest status", zap.Error(err))
+		return err
+	}
+
+	if common.IsPubKeyEqual(state.CurrentMessageState.PublicKey, &m.identity.PublicKey) { // Status message is ours
+		if currentStatus.Clock >= statusMessage.Clock || (currentStatus.StatusType == int(statusMessage.StatusType) && currentStatus.CustomText == statusMessage.CustomText) {
```
cammellos: This does not look exactly correct, I think there might be instances where it might fail.

Thinking of a case like:

Say we have 2 paired devices:

1) Device 1 sets status  `S1` (assume type is always the same) to `Clock: 0, CustomText: "A"`
2) Device 2 receives the update, goes offline
3) Device 1 sets status `S2` to `Clock: 1, customText: "B"`
4) Device 1 sets status `S3` to `Clock: 2, customText: "A"`
5) Device 2 goes back online and receive first  `S3`, which will be ignored as per code above. Then receives `S2` and it will be set to `S2`, but it should be `S3` (similarly if a message is dropped)

I think we should only trust `Clock`, and discard if clock is less (or equal, up to us), as it's effectively an updated status, as we can't guarantee that there were changes in between.

protocol/status_update.go
```diff
@@ -0,0 +1,18 @@
+package protocol
+
+import "github.com/status-im/status-go/protocol/protobuf"
+
+type UserStatus struct {
+	PublicKey  string `json:"public-key,omitempty"`
```
cammellos: We are moving away from `kebab-case` (albeit very slowly :)), as `camelCase` is generally more widely accepted

multiaccounts/accounts/database.go
```diff
@@ -454,7 +465,10 @@ func (db *Database) GetSettings() (Settings, error) {
 		&s.WalletSetUpPassed,
 		&s.WalletVisibleTokens,
 		&s.WakuBloomFilterMode,
-		&s.WebViewAllowPermissionRequests)
+		&s.WebViewAllowPermissionRequests,
+		&sqlite.JSONBlob{Data: &s.CurrentUserStatus},
```
cammellos, line 469: I'd avoid using a blob, as it's difficult to migrate in case of schema changes (although this looks like a fairly safe usage), probably I'd have either a separate table, or flatten the field in settings, though if we are confident it won't cause issues in the future we are probably ok, up to you

cammellos (commented):
awesome work as always, left a few questions/comments

protocol/messenger_status_updates.go
richard-ramos: It will be done inside `m.sendUserStatus()` in line 209

appdatabase/migrations/sql/1625872445_user_status.up.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE settings ADD COLUMN current_user_status BLOB;
+ALTER TABLE settings ADD COLUMN send_status_updates BOOLEAN DEFAULT TRUE;
+
```
cammellos: You will need also to set the existing rows to `TRUE`, as otherwise they will be `nil`/`false` (existing rows won't be impacted by defaulting a new field)

cammellos (approved):

iurimatias:
tested again, we found an issue with the migration and it was fixed.

PR #2277, opened 2021-07-13
add account conversion

bitgamma:
As part of implementing https://github.com/status-im/status-react/issues/11777 we need a way to convert a regular account into a Keycard account. The added function does

1. Add the keycard-pairing field to the multiaccount (unrelated sidenote, this field is basically used as a boolean because the actual value is ignored in status-react)
2. Add the keycard-instance_uid, keycard-paired_on and keycard-pairing field to the account settings
3. Re-encrypts the database
4. Removes the stored keys


shivekkhurana (approved):
L00ks g00d.

PR #2279, opened 2021-07-15
Delete messages

jrainville:
Makes it possible to delete messages.

Adds the `Deleted` property to Messages as well as a new table, `user_messages_deletes`, to keep track of deleted messages and remove them once they come in.

The delete feature for now doesn't really delete the, but rather hides them using the previously implemented `hide` field.

~~I will add tests in another commit. You can start to review straight away.~~ tests added


Samyoul (commented):
This all looks great. Would it be possible to add some tests to this PR?

jrainville:
> Would it be possible to add some tests to this PR?

Yup, I'm fixing some UI issues then adding the tests

jrainville:
I added the tests. There is just one issue that when you delete a message, you'd also like to delete the activity notification(s) that are related to that message, but I don't know how to do that, because the `activity_center_notifications` table doesn't have the messageId and the `message` field is just a blob.

protocol/messenger_messages.go
```diff
@@ -82,6 +82,71 @@ func (m *Messenger) EditMessage(ctx context.Context, request *requests.EditMessa
...
+func (m *Messenger) DeleteMessageAndSend(ctx context.Context, messageID string) (*MessengerResponse, error) {
+	message, err := m.persistence.MessageByID(messageID)
+	if err != nil {
+		return nil, err
+	}
+
+	if message.From != common.PubkeyToHex(&m.identity.PublicKey) {
+		return nil, ErrInvalidEditOrDeleteAuthor
+	}
+
```
richard-ramos, line 95: Should the user be allowed to delete this kind of messages? (dunno if it's possible to send a message of this type)
```
	0:  "UNKNOWN_CONTENT_TYPE",
	3:  "STATUS",
	5:  "TRANSACTION_COMMAND",
	6:  "SYSTEM_MESSAGE_CONTENT_PRIVATE_GROUP",
	10: "SYSTEM_MESSAGE_GAP",
```

cammellos: No, none of them, I am not sure about `STATUS`, I don't think it's used?

jrainville: Nice catch.

I used a whitelist approach, that way, if we ever add a new message type that cannot be deleted, it will be denied. I feel like it's better to block deleting more than letting through a delete that would break the app (so stability over usability in case of new types)

richard-ramos (approved):

shivekkhurana:
Please create a version tag when you update the `VERSION` file. Tags are needed for mobile to fetch the correct version.
I have created one for this merge [here](https://github.com/status-im/status-go/releases/tag/v0.83.2).

PR #2281, opened 2021-07-20
use 23/WAKU2-TOPICS format for content topics

richard-ramos:


vitvly (approved):

wakuv2/common/topic.go
```diff
@@ -60,3 +63,26 @@ func (t TopicType) MarshalText() ([]byte, error) {
...
+func (t TopicType) ContentTopic() string {
+	enc := make([]byte, len(t)*2)
+	hex.Encode(enc, t[:])
+	return "/waku/1/" + string(enc) + "/rlp"
+}
+
+func V2TopicToV1Topic(s string) (*TopicType, error) {
+	p := strings.Split(s, "/")
+
+	if len(p) != 5 || p[1] != "waku" || p[2] != "1" || p[4] != "proto" {
```
michaelsbradleyjr:
```suggestion
	if len(p) != 5 || p[1] != "waku" || p[2] != "1" || p[4] != "rlp" {
```

I may be confused, but per the discussion yesterday with the vac folks, it seems like `waku/1` app receivers and senders using waku v2 would be expecting/indicating `"rlp"` instead of `"proto"` for the encoding.

@jm-clius thoughts?

richard-ramos: I'll comit the suggestion in order for it to match what I have on line 71, but I'm still not sure whether I should use `rlp` or `proto` since I'm not using RLP encoding when WakuV2 is enabled

jm-clius: Yeah, the current design for `waku/1` as an "application" over `waku/2` is to use `rlp` as the encoding portion of the content topic. This is to indicate how the payload itself is used. For bridging the Waku v2 payload must contain the encrypted `data` portion of the Waku v1 envelope as per https://rfc.vac.dev/spec/15/#bridge

PR #2287, opened 2021-07-25
feat: wakuv2 peer persistence

richard-ramos:


wakuv2/waku.go
```diff
@@ -165,15 +170,35 @@ func New(nodeKey string, cfg *Config, logger *zap.Logger) (*Waku, error) {
...
 	connStatusChan := make(chan node.ConnStatus, 100)
 
 	if cfg.KeepAliveInterval == 0 {
 		cfg.KeepAliveInterval = DefaultConfig.KeepAliveInterval
 	}
 
+	libp2pOpts := node.DefaultLibP2POptions
+	libp2pOpts = append(libp2pOpts, libp2p.BandwidthReporter(waku.bandwidthCounter))
+
+	if appdb != nil && cfg.PersistPeers {
```
alaibe: I think we could do:
```
if cfg.PersistPeers {
  if appdb == nil {
    return nil, fmt.Errorf(¨A db connection must be provided in order to persist the peers") // or a better message
  }
  ...rest of logic
```

That way we will always respect the config and not have a state where we want to persist the peers and not doing it

wakuv2/persistence/sqlcipher.go
```diff
@@ -0,0 +1,93 @@
+package persistence
```
alaibe: nit: why is this file named sqlcipher and not query.go?
I don't see any sqlcipher related items here?

alaibe (approved):
LGTM, but pls see the 2 comments below

wakuv2/persistence/sqlcipher.go
richard-ramos: originally it had, but I removed the code but forgot to rename the file! :)

PR #2286, opened 2021-07-23
feat: upload/download rate for waku v1 messages

richard-ramos:

PR #2288, opened 2021-07-26
Wakuv2 peer stats

### LINKED ISSUE #2280: Add peer connection status support for Wakuv2
# Problem

Since Wakuv2 support has been merged into status-go, we don't have peer information available. Hence mobile client has to temporarily include a hardcoded conditional, whereby we always assume we're connected to some peers if using Wakuv2.

# Implementation

Now that go-waku has some support for peer management merged (https://github.com/status-im/go-waku/pull/30), it is necessary to make it available to status-go.

Also, as the Waku spec for peer management is now available (https://rfc.vac.dev/spec/27), some changes might be needed in go-waku to reflect the spec.

vitvly:
Adds support for go-waku's connection status channel receiving connectivity status updates for wakuv2 peers.

Fixes #2280

### DIFF
--- a/signal/events_wakuv2.go
+++ b/signal/events_wakuv2.go
@@ -1,0 +1,12 @@
+package signal
+
+const (
+	// EventPeerStats is sent when peer is added or removed.
+	// it will be a map with capability=peer count k/v's.
+	EventPeerStats = "wakuv2.peerstats"
+)
+
+// SendPeerStats sends discovery.summary signal.
+func SendPeerStats(peerStats interface{}) {
+	send(EventPeerStats, peerStats)
+}
--- a/wakuv2/waku.go
+++ b/wakuv2/waku.go
@@ -50,6 +50,7 @@ import (
 	"github.com/status-im/go-waku/waku/v2/protocol/relay"
 
 	"github.com/status-im/status-go/eth-node/types"
+	"github.com/status-im/status-go/signal"
 	"github.com/status-im/status-go/wakuv2/common"
 
 	node "github.com/status-im/go-waku/waku/v2/node"
@@ -145,6 +146,9 @@ func New(nodeKey string, cfg *Config, logger *zap.Logger) (*Waku, error) {
 		return nil, fmt.Errorf("failed to setup the network interface: %v", err)
 	}
 
+	connStatusChan := make(chan node.ConnStatus)
+
+	keepAliveInt := 1
 	waku.node, err = node.New(context.Background(),
 		node.WithLibP2POptions(
 			libp2p.BandwidthReporter(waku.bandwidthCounter),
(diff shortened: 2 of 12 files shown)


wakuv2/waku.go
```diff
@@ -139,13 +140,21 @@ func New(nodeKey string, cfg *Config, logger *zap.Logger) (*Waku, error) {
...
 		node.WithWakuRelay(wakurelay.WithMaxMessageSize(int(waku.settings.MaxMsgSize))),
 		node.WithWakuStore(false), // Mounts the store protocol (without storing the messages)
+		node.WithConnStatusChan(connStatusChan),
 	)
 
+	go func() {
+		for c := range connStatusChan {
+			signal.SendPeerStats(c)
+		}
+	}()
```
richard-ramos: I think you'll have an import cycle issue, not sure signals should be emited in this file.
Also the go routine is never stopped. Perhaps something like this could be used? (there are similar examples of code in the repo)

```
go func(){
  for {
    select {
       case <-w.quit:
	   return
	case c <-connStatusChan:
           signal.SendPeerStats(c)
      }
}
}()
```

signal/events_wakuv2.go
```diff
@@ -0,0 +1,12 @@
...
+
+const (
+	// EventPeerStats is sent when peer is added or removed.
+	// it will be a map with capability=peer count k/v's.
+	EventPeerStats = "wakuv2.peerstats"
+)
+
+// SendPeerStats sends discovery.summary signal.
+func SendPeerStats(peerStats interface{}) {
+	send(EventPeerStats, peerStats)
```
richard-ramos, line 11: Is `peerStats` translated to JSON correctly? I was checking  the definition for ConnStatus which I think is the data sent
```
type ConnStatus struct {
	IsOnline   bool
	HasHistory bool
}
```
and i did not see that the attributes had metadata like `json:"nameOfTheField,omitempty"``.  
Is this done automatically if that info is missing?

vitvly: Good catch! I missed those annotations. But it's translated fine - in status-react i receive a proper JSON.

wakuv2/waku.go
vitvly: Fixed the goroutine, thanks for pointing this out. What will the import cycle issue entail? Are you suggesting moving signals part to a separate module?

richard-ramos: The build will not complete due to failing in the lint step:
```
[2021-07-26T16:10:50.760Z] + make lint
[2021-07-26T16:10:50.760Z] lint
[2021-07-26T16:11:03.479Z] level=warning msg="[runner] Can't run linter goanalysis_metalinter: deadcode: analysis skipped: errors in package: [-: import cycle not allowed in test]"
[2021-07-26T16:11:06.430Z] level=warning msg="[runner] Can't run linter unused: buildir: failed to load package ens: could not load export data: no export data for \"github.com/status-im/status-go/protocol/ens [github.com/status-im/status-go/protocol/common.test]\""
[2021-07-26T16:11:06.430Z] level=error msg="Running error: buildir: failed to load package ens: could not load export data: no export data for \"github.com/status-im/status-go/protocol/ens [github.com/status-im/status-go/protocol/common.test]\""
[2021-07-26T16:11:06.430Z] make: *** [Makefile:304: lint] Error 3
script returned exit code 2
```

I had this issue with a different status-go PR https://github.com/status-im/status-go/pull/2286/commits  and had to reorganize the code just to get past that issue. You can see here in the highlighted code https://github.com/status-im/status-go/pull/2286/commits/ea3038a10b92a264529b7724539442715b5bc326#diff-bf66678b3e92c93f707a87660f5b33bf6b46d9a26c5eca024e816b1da20f3b62L199-L211 how I had to remove the go routine from waku.go and move it to a different layer. Probably something similar will be needed here

vitvly: Ah, ok, strange though, i see that Jenkins built the PR fine, also didn't see the problem appearing with `make statusgo` and with status-react builds.

vitvly: Cryptic error message from `make lint` in any case.

cammellos (approved):

PR #2290, opened 2021-07-27
Set timeout on mailserver requests

cammellos:
Set timeout on mailservers requests, as QA team noticed that sometimes syncing hangs.


richard-ramos (approved):
LGTM!

Is it necessary to do something in the frontend for when there's a timeout?

jrainville (approved):
LGTM

PR #2297, opened 2021-08-03
re-apply signTypedData fix

bitgamma:
update go-ethereum to reapply a fix to signTypedData which allows handling chainID both as int and string


cammellos (approved):
Thanks for spotting this, and sorry for leaving it out!

PR #2299, opened 2021-08-05
Status public API

cammellos:
Adds a public api endpoint: `status_communityInfo`, to retrieve synchronously community info, used by the community curation dapp.


services/status/service.go
```diff
@@ -0,0 +1,82 @@
...
+// Stop is run when a service is stopped.
+func (s *Service) Stop() error {
+	return nil
+}
+
+type PublicAPI struct {
+	service *Service
+}
+
+func (p *PublicAPI) Ping() string {
```
bitgamma: maybe it would be nice to have a comment on why this is needed?

bitgamma (approved):

services/status/service.go
cammellos: I'll leave a comment, it's actually not needed, I added it as we have some contractors working on it so they can test they are using the right version, but good point, thanks!

Samyoul (approved):
Ping Pong

PR #2306, opened 2021-08-09
Support to audio/ogg.

pinebit:
This adds support to audio/ogg for voice messages.
The idea is to migrate to this format, instead of currently adopted AAC.


protocol/messenger.go
```diff
@@ -2035,6 +2035,8 @@ func (m *Messenger) sendChatMessage(ctx context.Context, message *common.Message
 		audioMessage.Payload = payload
 		audioMessage.Type = audio.Type(payload)
 		message.Payload = &protobuf.ChatMessage_Audio{Audio: audioMessage}
+
+		_ = file.Close()
```
pinebit, line 2044: this is bug fix, not directly related to the new feature, since this code is unused.

Samyoul: Perhaps you can defer the file close so that the file will be guaranteed to be closed when the func ends
```suggestion
		defer file.Close()
```

Samyoul (approved):
Looks good, little suggestion on the `file.Close`

protocol/messenger.go
pinebit: it is deferred at the top of the scope.
the problem is that below it tries to delete this file, and since defer is not executed yet, the file remains locked and this results in error.

Samyoul: Ah I see it. I didn't see the top of the scope.

Samyoul (approved):

igor-sirotin:
Closing as too old. Feel free to re-open and rebase.

PR #2310, opened 2021-08-12
fix: delete messages when the signal arrives before the OG msg

richard-ramos:


protocol/message_persistence.go
```diff
@@ -298,6 +299,11 @@ func (db sqlitePersistence) tableUserMessagesAllValues(message *common.Message)
 		}
 	}
 
+	hide := false
```
cammellos: I would probably hide the message in the database as we set the `Deleted` flag, or you think it's better to keep them separate? What do you think?
Also, would you mind adding tests for the two scenarios please? (Receiving a Delete first, and after the message, and making sure that LastMessage is updated).
Thanks!

richard-ramos: I added it here because I saw that SaveMessages was being called when the messages are added to the response, and SaveMessages overwrite the messages we have in the database:
https://github.com/status-im/status-go/blob/develop/protocol/messenger.go#L3048-L3054

Perhaps we should instead skip adding the message to the response if the Delete is received first?

cammellos: > Perhaps we should instead skip adding the message to the response if the Delete is received first?

mmh, interesting, so we should probably save the message in those cases, since we might want the whole history (or we might not :), that depends on what we want to do with deleted messages), but definitely we should not add it to the response, I agree

richard-ramos:
I made the requested changes and added some extra checks to the test unit

cammellos (approved):
Awesome! Thank you!

PR #2316, opened 2021-08-16
preload chats

flexsurfer:
we want to first preload N chats and then load all


protocol/messenger_chats.go
```diff
@@ -9,12 +9,13 @@ import (
...
-func (m *Messenger) Chats() []*Chat {
+func (m *Messenger) Chats(num int) []*Chat {
 	var chats []*Chat
-
+	var i = 0
 	m.allChats.Range(func(chatID string, chat *Chat) (shouldContinue bool) {
 		chats = append(chats, chat)
-		return true
+		i++
+		return num == 0 || i < num
```
Samyoul: I can see what you are doing here, but I think it is introducing a bit of an anti-pattern. Giving a zero parameter for all chats is a bit wonky. In my mind there are 2 other options.

a change to the function to allow for no parameter, using a variadic parameter. 

```go
func (m *Messenger) Chats(num ...int) []*Chat {
	var chats []*Chat
	var i int
	
	m.allChats.Range(func(chatID string, chat *Chat) (shouldContinue bool) {
		chats = append(chats, chat)
		i++
		return len(num) == 0 || i < num[0]
	})

	return chats
}
```

Can be called 

`s.m.Chats()` or `s.m.Chats(4)`

OR

A dedicated function that is designed specifically to return `n` Chats. And a parameter of 0 will return 0 Chats.

Samyoul: Also would it be possible to add some tests for the new behaviour of this function or new dedicated function?

flexsurfer: good points ,thanks

flexsurfer:
@Samyoul done

Samyoul (approved):
Looks good 👍

PR #2319, opened 2021-08-17
Dismiss activity center notifications for deleted messages

briansztamfater:
`status-go` part for fixing https://github.com/status-im/status-react/issues/12449

Dismiss activity center notifications for deleted messages

Includes changes from #2317, only last commit should be reviewed

### DIFF
--- a/protocol/activity_center_persistence.go
+++ b/protocol/activity_center_persistence.go
@@ -17,6 +17,52 @@ func (db sqlitePersistence) DeleteActivityCenterNotification(id []byte) error {
 	return err
 }
 
+func (db sqlitePersistence) DeleteActivityCenterNotificationForMessage(chatID string, messageID string) error {
+	var tx *sql.Tx
+	var err error
+
+	tx, err = db.db.BeginTx(context.Background(), &sql.TxOptions{})
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err == nil {
+			err = tx.Commit()
+			return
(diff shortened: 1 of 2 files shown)


protocol/activity_center_persistence.go
```diff
@@ -17,6 +17,44 @@ func (db sqlitePersistence) DeleteActivityCenterNotification(id []byte) error {
...
+	var ids []types.HexBytes
+
+	for _, notification := range notifications {
+		if notification.Message != nil && notification.Message.ID == messageID {
+			ids = append(ids, notification.ID)
+		}
+	}
+
+	if len(ids) > 0 {
+		return db.DismissActivityCenterNotifications(ids)
```
briansztamfater: When this is executed it _seems_ to work but after that can't retrieve any messages from any chat, app does not load chat messages. Any clue what I could be doing wrong? cc @cammellos @Samyoul

briansztamfater: Nevermind, DB connection was being closed after this DB query execution.

cammellos (approved):
It would be good to have a test covering the new behavior, but I understand if we are in a rush for release, maybe just a reminder to add it?

briansztamfater:
> It would be good to have a test covering the new behavior, but I understand if we are in a rush for release, maybe just a reminder to add it?

@cammellos Sure thing. I'll work on tests for these changes on another PR.

PR #2332, opened 2021-08-24
revert removal of Rendezvous config and nodes

jakubgs:
Since we've deployed `v0.84.0` to `eth.staging` we can start using it. This is specifically important for Status Node Program participants. With their participation we can really test the new node version before pushing it out to `eth.prod` fleet.

For this reason I'm making `eth.staging` the default fleet used by `gen_config.sh` script.
We can update it back to `eth.prod` once we roll out the new node version to it.


churik:
@jakubgs is PR still relevant?

jakubgs:
Good question. I don't think it is.

PR #2333, opened 2021-08-24
feat: obtain external address for rendezvous

richard-ramos:
Currently, if rendezvous is enabled, and AdvertiseAddr is not added to the configuration, the rendezvous logic will use the IP address returned by geth, and that IP address might be localhost or private and the node will not be discoverable. To avoid this situation we hardcode an external IP address in the node configuration, however this is not ideal, since the external IP address might be dinamic.

This PR changes this logic so if the IP address returned by geth or the value in AdvertiseAddr are local/private, the node will do a request to a random rendezvous server, and this server will reply back with the IP address the node used to connect to it.

### DIFF
--- a/discovery/rendezvous.go
+++ b/discovery/rendezvous.go
@@ -129,6 +129,7 @@ func (r *Rendezvous) MakeRecord(srv *ma.Multiaddr) (record enr.Record, err error
 			parsedIP := net.ParseIP(ipAddr)
 			if parsedIP != nil {
 				ip = parsedIP
+				log.Info("node's external IP address", "ipAddr", ipAddr)
 			} else {
 				log.Error("invalid ip address obtained from rendezvous server", "ipaddr", ipAddr, "err", err)
 			}
--- a/discovery/rendezvous.go
+++ b/discovery/rendezvous.go
@@ -104,7 +104,7 @@ func (r *Rendezvous) Stop() error {
 	return nil
 }
 
-func (r *Rendezvous) MakeRecord(srv ma.Multiaddr) (record enr.Record, err error) {
+func (r *Rendezvous) MakeRecord(srv *ma.Multiaddr) (record enr.Record, err error) {
 	r.recordMu.Lock()
 	defer r.recordMu.Unlock()
 	if r.record != nil {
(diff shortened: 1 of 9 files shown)


jakubgs (approved):
It looks good, but I'd also include an adjustment to the comment in `params/config.go`:
https://github.com/status-im/status-go/blob/83394e0ed17265cb20ab3e56fe3de13ca4057267/params/config.go#L341-L343
To indicate that:

* Config value is specifically required by Rendezvous protocol
* Is optional, especially if the host IPs can change
* Can be useful when in rare cases when Rendezvous nodes return wrong IP

Also, we should wait for review from @cammellos and @Samyoul but they are out today.

jakubgs:
Also, could you explain how this plays with use of `AdvertiseAddr`. I'm looking at this code:
https://github.com/status-im/status-go/blob/83394e0ed17265cb20ab3e56fe3de13ca4057267/node/get_status_node.go#L256-L263
And I'm a but confused what's happening here.

richard-ramos:
When AdvertiseAddr is empty, `discoverNode()` will return a `enode.Node` with the enr that geth uses. (When I did the test on my local machine, the IP address that geth was returning in this enr was 127.0.0.1):
https://github.com/status-im/status-go/blob/83394e0ed17265cb20ab3e56fe3de13ca4057267/vendor/github.com/ethereum/go-ethereum/p2p/server.go#L410-L419

Then, this `enode.Node` is passed to `NewRendezvous` to be used later to obtain the IP Address/TCP port/UDP port:
https://github.com/status-im/status-go/blob/develop/discovery/rendezvous.go#L33-L41
https://github.com/status-im/status-go/blob/develop/discovery/rendezvous.go#L119-L121

Later, when rendezvous starts, it will attempt to register the node with a rendezvous server with `Register` which executes `MakeRecord` to create a new ENR instead of reusing the one from geth  ¯\\\_(ツ)_/¯
https://github.com/status-im/status-go/blob/88f6ee9040671063234ee340df83cff8f61e001a/discovery/rendezvous.go#L167-L170

Notice in the definition of MakeRecord that in that moment `r.record` will be null because status-go uses `NewRendezvous` instead of `NewRendezvousWithENR`, so it will then attemptthen to create a enr, reusing the information of the `enode.Node` created in `discoverNode()`, and then, it will evaluate the node IP address to know if it should determine the external IP address or not.

jakubgs (approved):
Nice. Thanks for the explanation and the comment.

jakubgs:
@cammellos @Samyoul can you check this out?

Samyoul (approved):
This all looks good

PR #2335, opened 2021-08-25
fix activechats ordering

flexsurfer:
fix activechats ordering


vitvly (approved):

protocol/messenger.go
```diff
@@ -991,6 +993,12 @@ func (m *Messenger) Init() error {
 		}
 
 		m.allChats.Store(chat.ID, chat)
+
+		if i < 60 {
```
jakubgs, line 1000: Magic number? What is it?

jakubgs (changes requested):
Magic numbers are bad.

protocol/messenger.go
flexsurfer: Magic number

flexsurfer: this function is a temporary workaround, the entire implementation is not perfect, I could add a description but...it won't explain anything :)

flexsurfer: in will only add more questions :)

jakubgs (approved):
Thanks for the comment.

PR #2349, opened 2021-09-02
feat: enable wallet without network binding

alaibe:
Initial work to enable wallet to accommodate multiple network. In this PR:
- [X] Add network manager as well as network record
- [X] Build eth client based on the network
- [X] Remove binding between network and wallet (no more dependency injection of the rpc client)
- [X] Pass chainID when calling the database instead of having the chainID being in the struct of the db
- [X] Maintain current functionality by using the chain id passed when building the service
- [X] Add API to CRUD networks
- [X] Update current API to work with multiple network => it means adding a lookup for all enabled network and do similar work
** [X] Token
** [X] Transaction
** [X] Networks
** [X] Crypto on ramps
** [X] Opensea
- [X] Documentation
- [x] Handle bootstrap network in config


services/wallet/bigint/big_int.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+func (b *BigInt) UnmarshalJSON(p []byte) error {
+	if string(p) == "null" {
+		return nil
+	}
+	z := new(big.Int)
+	_, ok := z.SetString(strings.Trim(string(p), "\""), 10)
+	if !ok {
+		return fmt.Errorf("not a valid big integer: %s", "123")
```
richard-ramos:
```suggestion
		return fmt.Errorf("not a valid big integer: %s", string(p))
```

services/wallet/transfer/view.go
```diff
@@ -9,16 +9,41 @@ import (
...
+	GasUsed              hexutil.Uint64 `json:"gasUsed"`
+	Nonce                hexutil.Uint64 `json:"nonce"`
+	TxStatus             hexutil.Uint64 `json:"txStatus"`
+	Input                hexutil.Bytes  `json:"input"`
+	TxHash               common.Hash    `json:"txHash"`
+	Value                *hexutil.Big   `json:"value"`
+	From                 common.Address `json:"from"`
+	To                   common.Address `json:"to"`
+	Contract             common.Address `json:"contract"`
+	NetworkID            uint64         `json:"network_id"`
```
richard-ramos:
```suggestion
	NetworkID            uint64         `json:"networkId"`
```

services/wallet/network/network.go
```diff
@@ -0,0 +1,185 @@
...
+	RPCURL                 string `json:"rpc_url"`
+	BlockExplorerURL       string `json:"block_explorer_url,omitempty"`
+	IconURL                string `json:"icon_url,omitempty"`
+	NativeCurrencyName     string `json:"native_currency_name,omitempty"`
+	NativeCurrencySymbol   string `json:"native_currency_symbol,omitempty"`
+	NativeCurrencyDecimals uint64 `json:"native_currency_decimals"`
+	IsTest                 bool   `json:"is_test"`
+	Layer                  uint64 `json:"layer"`
+	Enabled                bool   `json:"enabled"`
+}
```
richard-ramos: should we use camelCase instead of snake_case in these annotations?

services/wallet/token.go
```diff
@@ -0,0 +1,123 @@
...
+type Token struct {
+	Address common.Address `json:"address"`
+	Name    string         `json:"name"`
+	Symbol  string         `json:"symbol"`
+	Color   string         `json:"color"`
+	// Decimals defines how divisible the token is. For example, 0 would be
+	// indivisible, whereas 18 would allow very small amounts of the token
+	// to be traded.
+	Decimals uint   `json:"decimals"`
+	ChainID  uint64 `json:"chain_id"`
```
richard-ramos:
```suggestion
	ChainID  uint64 `json:"chainId"`
```

richard-ramos (approved):
Great work! :clap:

services/wallet/network/network.go
alaibe: Will do, TBH I don't know what I should use I've seen both. :confused:

services/wallet/bigint/big_int.go
alaibe: Good catch, I did a copy/paste without looking in details :see_no_evil:

jakubgs:
Hmmmm:
```
[2021-09-09T13:15:44.475Z] FAIL    github.com/status-im/status-go/appmetrics [build failed]
```
```
[2021-09-09T13:15:03.177Z] appmetrics/database_test.go:66:83: conversion from int to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)
```

alaibe:
Thanks @jakubgs @richard-ramos for the help on solving the test!

PR #2353, opened 2021-09-07
chats list performance

flexsurfer:
chats list performance


protocol/messenger_chats.go
```diff
@@ -13,15 +13,38 @@ func (m *Messenger) Chats() []*Chat {
...
+					text = chat.LastMessage.Text
+				}
+				chat.LastMessage = &common.Message{
+					ChatMessage: protobuf.ChatMessage{
+						ContentType: chat.LastMessage.ContentType,
+						Text:        text},
+				}
+			}
+
+			chat.Members = nil
```
cammellos: This is a bit dangerous, 
I think you are resetting the field for good, as we use a pointer, so it would be gone from memory, I think what you want to do is to use a custom marshaler for chats, where you can do this, or make a copy of the chat (probably neater to use a custom marshaler)

cammellos (changes requested):
I think this might cause bugs, as setting `Members = nil` will impact other parts of the code

protocol/messenger_chats.go
flexsurfer: thanks, created new ChatPreview struct

flexsurfer:
ready for review @cammellos

cammellos (approved):
Don't forget to bump the version!

PR #2354, opened 2021-09-08
feat: Add more attributes from opensea

alaibe:
Add sale / orders information as well as collection traits

### DIFF
--- a/services/wallet/opensea.go
+++ b/services/wallet/opensea.go
@@ -50,6 +50,24 @@ type OpenseaTrait struct {
 	DisplayType string     `json:"display_type"`
 }
 
+type OpenseaPaymentToken struct {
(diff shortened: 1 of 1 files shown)


richard-ramos (commented):
Looks good, my only observation would be regarding the usage of snake_case instead of using camelCase. According to https://github.com/status-im/status-go/pull/2276#discussion_r670419278 we should be using camelCase.

alaibe:
@richard-ramos Have a look to opensea API they are using snake_case this is why the json metadata is snake case too:

e.g: https://api.opensea.io/api/v1/assets?owner=0xdd0d6dc129e73e240c1e7bf0926394237aa4a427

Khushboo-dev-cpp (approved):

PR #2358, opened 2021-09-08
drop incorrect index on user_messages_edits and create index on user_messages_deletes

michaelsbradleyjr:
In `protocol/migrations/sqlite/1625762506_add_deleted_messages.up.sql` an index is created on `user_messages_edits` but it should instead be created on `user_messages_deletes`.

Drop the incorrect index and create the correct one.

---

I installed `go-bindata` via `make gen-install`, and I ran `go generate` within the `protocol/migrations/sqlite` directory.

Let me know if I missed a step or did something incorrectly and I'll fix it ASAP.


Samyoul:
Hey @michaelsbradleyjr the filename of your migration should have the `*.up.sql` suffix otherwise it won't be picked up by the migrator.

There are also a helpful make commands that will prefix with a unix timestamp and the correct suffixes for you,

- `make migration D=<MIGRATION NAME>`
  - Outputs to `appdatabase/migrations/sql/<MIGRATION FILE>`
- `make migration-protocol D=<MIGRATION NAME>`
  - Outputs to `protocol/migrations/sqlite/<MIGRATION FILE>`

michaelsbradleyjr:
Sorry to leave this hanging, I didn't notice the email with @Samyoul's reply and just now thought to check back on this PR.

@Samyoul I'm happy to add the `.up.sql` suffix. When I created the file at the time, I saw there was already a file named [`1628280060_create-usermessages-index.sql`](https://github.com/status-im/status-go/blob/fix/deletes-index/protocol/migrations/sqlite/1628280060_create-usermessages-index.sql) (lacking the suffix) and it was picked up by the migrator, so I named mine similarly since it likewise only deals with an index.

Should I rename both files so they end with `.up.sql` or only mine?

Samyoul:
Hey @michaelsbradleyjr that's interesting. Are you able to see that the new indexes are created? I can see that `make generate` picks up the migration file to create a comm, but without the `.up.sql` suffix the migrator can't know if a migration is up or down.

I've checked this migration and the one you mention directly in the db, it isn't being picked up.

Screenshot 2021-09-30 at 16 16 16

Screenshot 2021-09-30 at 16 18 00

michaelsbradleyjr:
> Are you able to see that the new indexes are created?

No, actually! Good catch.

I had conflated seeing the incorrect index added in `1625762506_add_deleted_messages.up.sql` (in a command-line dump of the schema) with the index added in `1628280060_create-usermessages-index.sql`.

So I should change the name of `1628280060_create-usermessages-index.sql` to `1628280060_create-usermessages-index.up.sql` and `1631129360_fix-deletedmessages-index.sql` to `1631129360_fix-deletedmessages-index.up.sql`, and that's it?

Samyoul:
>So I should change the name of 1628280060_create-usermessages-index.sql to 1628280060_create-usermessages-index.up.sql and 1631129360_fix-deletedmessages-index.sql to 1631129360_fix-deletedmessages-index.up.sql, and that's it?

Yes, that should be enough for the migrator to pick up the new migrations

michaelsbradleyjr:
Done!

churik:
@michaelsbradleyjr is PR still relevant?

igor-sirotin:
Closing as too old. Feel free to re-open and rebase.

PR #2359, opened 2021-09-09
feat: Rpc client manage multiple eth client

alaibe:
Make the rpc client chain id aware.

In order to make a call with a different id, it needs to be added as a params, example:
```
{
      "jsonrpc": "2.0",
      "chainId": 3
      "method": "methodName",
      "params": {}
}
```

By default the chainID used is the one from the config network ID and the URL url is the one from the upstream URL


richard-ramos (approved):

iurimatias:
tested in https://github.com/status-im/status-desktop/pull/3474

PR #2367, opened 2021-09-16
🖼 Set any url as profile picture (useful for NFTs)

shivekkhurana:
Introduces an RPC api on multiaccounts endpoint to add any url as a profile picture.

In case the URL is a square image, it is not cropped. If it's rectangular, it's cropped to get the largest square with centre same as the centre of the rectangle. 

This feature is required for setting NFTs as profile picture.


alaibe (approved):

images/decode.go
```diff
@@ -27,6 +30,20 @@ func Decode(fileName string) (image.Image, error) {
 	return decodeImageData(fb, file)
 }
 
+func DecodeFromURL(path string) (image.Image, error) {
+	res, err := http.Get(path)
```
richard-ramos: Consider adding a timeout:
```suggestion
	client := http.Client{
		Timeout: 5 * time.Second,
	}
	res, err := client.Get(path)
```

images/decode.go
```diff
@@ -27,6 +30,20 @@ func Decode(fileName string) (image.Image, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
```
richard-ramos, line 44:
```suggestion
defer func() {
	if err := res.Body.Close(); err != nil {
		log.Error("failed to close profile pic http request body", "err", err)
	}
}()
```

richard-ramos (commented):
Good work!
I only have some minor observations

images/decode.go
shivekkhurana: This suggestions feature is cool.

richard-ramos (approved):

Samyoul (approved):
Functionally, this work is great. I do have one observation, does the user know that they are accessing a 3rd party service directly and therefore exposing their IP address?

shivekkhurana:
@Samyoul 
> Functionally, this work is great. I do have one observation, does the user know that they are accessing a 3rd party service directly and therefore exposing their IP address?

Before the user is able to view collectibles, they explicitly agree to the fact that their IP will be exposed. This function (given the current ui) will be called only after the user's permission is collected. 

However this function doesn't assume the permission and can be called anytime.

Samyoul:
@shivekkhurana This  what I assumed, we don't really police these settings in `status-go` we just assume that `status-react` is doing it for us. My concern is that this functionality can be used for any remote image source, not necessarily NFTs. Although in the case where a user is explicitly selecting an image from a remote source, I'd assume that they understand that the app will call a remote 3rd party server and expose their IP ... I suppose in that case if we added functionality to select any remote image source that a permission needs to be granted in the settings managed by `status-react`

shivekkhurana:
>  I suppose in that case if we added functionality to select any remote image source that a permission needs to be granted in the settings managed by status-react

If we add a functionality to select *any* image from a URL, we will add this permission to status-react. As of now, the user can *only* select an NFT as PFP by clicking on it (after explicitly enabling it).

PR #2369, opened 2021-09-17
separate mobile and desktop in the ckecklist

flexsurfer:
separate mobile and desktop in the ckecklist


anastasiyaig:
this is awesome, hope it will bring more value than we currently have using this check list

anastasiyaig (approved):

PR #2370, opened 2021-09-19
chore: add soname to libstatus.so

richard-ramos:
Required by https://github.com/status-im/status-desktop/pull/3512. Do not merge until that PR is merged too.

This is required for creating a valid .deb file, otherwise `lintian` will complain with the following error:
`E: status-desktop: sharedobject-in-library-directory-missing-soname usr/lib/libstatus.so`
 
Also, it's a good practice for shared libraries to include the soname.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,7 @@ else ifeq ($(detected_OS),Windows)
  GOBIN_SHARED_LIB_EXT := dll
 else
  GOBIN_SHARED_LIB_EXT := so
+ GOBIN_SHARED_LIB_CGO_LDFLAGS := CGO_LDFLAGS="-Wl,-soname,libstatus.so.0"
 endif
 
 help: ##@other Show this help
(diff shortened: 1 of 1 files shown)


richard-ramos:
@michaelsbradleyjr what do you think if we make this change in macOS too? It might be helpful to avoid having to use install_name_tool while building desktop

michaelsbradleyjr:
I'm unfamiliar with `-soname` do you know if it's supported on macOS?

michaelsbradleyjr:
I'm not getting any hits for a search for `soname` in `man clang` and `man ld` on macOS (10.15). It's possible the `-rpath`option serves a similar role... I need to come to a better understanding of `-soname`'s purpose.

richard-ramos:
In macos the equivalent would be the flag -install_name

michaelsbradleyjr:
`-install_name` seems like a good option. I'm googl'ing to see if someone's done a writeup like, *"When to prefer install_name over rpath in macOS"*, but haven't found anything yet.

michaelsbradleyjr:
Some info here:

https://www.mikeash.com/pyblog/friday-qa-2009-11-06-linking-and-install-names.html

https://wincent.com/wiki/@executable_path,_@load_path_and_@rpath

cammellos (approved):

PR #2372, opened 2021-09-21
contacts performance 🚀

flexsurfer:
contacts performance


protocol/messenger_pin_messages.go
```diff
@@ -73,24 +73,6 @@ func (m *Messenger) sendPinMessage(ctx context.Context, message *common.PinMessa
 }
 
 func (m *Messenger) PinnedMessageByChatID(chatID, cursor string, limit int) ([]*common.PinnedMessage, string, error) {
-	chat, err := m.persistence.Chat(chatID)
```
cammellos, line 76: Is this intentional?

flexsurfer: yes, we don't have pinned messages for timeline

protocol/messenger_contact_update_test.go
```diff
@@ -95,7 +95,6 @@ func (s *MessengerContactUpdateSuite) TestReceiveContactUpdate() {
 	receivedContact := response.Contacts[0]
 	s.Require().Equal(theirName, receivedContact.Name)
 	s.Require().False(receivedContact.ENSVerified)
-	s.Require().True(receivedContact.HasBeenAdded())
```
Samyoul, line 98: Should we still check that the contact is added?

protocol/messenger_test.go
```diff
@@ -270,24 +270,23 @@ func (s *MessengerSuite) TestInit() {
 				key, err := crypto.GenerateKey()
 				s.Require().NoError(err)
 				contact := Contact{
-					ID:         types.EncodeHex(crypto.FromECDSAPub(&key.PublicKey)),
-					Name:       "Some Contact",
-					SystemTags: []string{contactAdded, contactBlocked},
+					ID:      types.EncodeHex(crypto.FromECDSAPub(&key.PublicKey)),
+					Name:    "Some Contact",
+					Blocked: true,
```
Samyoul, line 276: If a contact is blocked would they also be added?

protocol/contact.go
```diff
@@ -148,27 +105,20 @@ func BuildContactFromPublicKey(publicKey *ecdsa.PublicKey) (*Contact, error) {
...
 
 	contact := &Contact{
 		ID:        publicKeyString,
 		Alias:     alias.GenerateFromPublicKey(publicKey),
-		Identicon: identicon,
+		Identicon: newIdenticon,
 	}
 
 	return contact, nil
 }
```
Samyoul, line 113: If we are looking for optimisation, we could offload some computation work by storing the identicon payload in the DB and retrieving it rather than building it every time we want access to the contact.

This would come at the cost of an increased DB size, but an identicon never changes so it is the perfect thing to store.

Samyoul (approved):
I don't see anything breaking about this PR. I've got a couple of questions but nothing that impacts on the general principle and implementation of what you've done here. Nice work

protocol/contact.go
flexsurfer: yeah, we're thinking about that, but will be addressed in another PR

protocol/messenger_test.go
flexsurfer: yes

richard-ramos:
Aren't contacts being created for any new message received?  I noticed when I logged out from an account, and logged in again that some `wakuext_chatMessages` results don't include the `alias`. And when querying the database, I see the number of rows in `contacts` does not match the number of different `source` in the `user_messages` table:

image

Notice in the screenshot below how most messages are missing the alias:
image

image


Also, isn't the "identicon" being returned anymore as part of the messages?
image

flexsurfer:
this is the essence of this PR, to reduce the number of entities in contacts db, so we remove entities in migration and do not add new later, so contacts table doesn't grow that much. The problem was that we added contact for every message, and because of spam for users who actively use the app, contacts table grows up to 30k entities. So now if there is no alias you have to generate it. In the next PR we want to use HTTP server for identicons and I know we probably want to drop public chats and aliases (but the latter is not certain)

richard-ramos:
> In the next PR we want to use HTTP server for identicons

Very cool! would it be possible for that HTTP server to return the user profile picture too? for example, it could be really useful to have the following URLs available:
```
http://localhost:8080/identicon/0x.......
http://localhost:8080/image/large/0x.......
http://localhost:8080/image/thumbnail/0x.......

# this returns the thumbnail if the user is a contact, otherwise returns the identicon (and it's also aware of the setting to display profile pictures for non-contacts)
http://localhost:8080/auto/0x.......  
```

flexsurfer:
yes

PR #2378, opened 2021-09-22
feat: add signals that are emitted before/after history messages are requested

richard-ramos:


iurimatias (approved):

iurimatias:
tested in https://github.com/status-im/status-desktop/pull/3667

PR #2383, opened 2021-09-24
feat: add API to mark activity center messages as unread

0x-r4bbit:
As raised in https://github.com/status-im/status-desktop/issues/3665, there's no way in Status Deskop to mark activity center notifications as unread. Not only is this not implemented in Status Desktop's UI, there's also no API to mark notifications as unread.

This commit adds a new `markActivityCenterNotificationsAsUnread` to the protocol API, so marking notifications as unread can be implemented by status clients.

### DIFF
--- a/protocol/activity_center_persistence.go
+++ b/protocol/activity_center_persistence.go
@@ -382,6 +382,20 @@ func (db sqlitePersistence) MarkActivityCenterNotificationsRead(ids []types.HexB
 
 }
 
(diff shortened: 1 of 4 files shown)


jrainville (approved):

0x-r4bbit:
> Have you updated the documentation, if impacted (e.g. docs.status.im)?

Seems like the bot generates a link that's broken.

PR #2384, opened 2021-09-26
Store Highlight field for identify new chats

### LINKED ISSUE #10087: Identify new chats on the list
# Problem

There's no visible way to distinguish new chats that appeared on your list. A message from a new contact or an invitation to join a group chat might go unnoticed.

## Implementation

Screenshot 2020-02-26 at 14 48 10

To identify new chats on a list, please change the background colour of the list item to our `Light Blue`. This should persist until you open the chat for the first time, after which it returns to usual white. This affects new private chats that appeared on your list and new group chats.

## Acceptance Criteria

New chats are identified with the light blue background until they're opened.

## Notes

Thanks!

### LINKED ISSUE #12650: The chat list is missing the last message, in the case of community Invite
# Problem

Previously app used to show the last message with the invited community name, currently, it is showing blank.

## Implementation
<div align="center">
Screenshot
Screenshot
</div>

Parveshdhull:
fixes https://github.com/status-im/status-react/issues/10087
fixes https://github.com/status-im/status-react/issues/12650

This PR stores highlight field which are required for Identify new chats

## status-react PR:
https://github.com/status-im/status-react/pull/12654


Samyoul (commented):
Hey @Parveshdhull thank you for your PR. Looks good.

I think that naming the field `Highlight` may be a little meta, what we really want to communicate is whether the chat is new or if we've seen it before. So I'd recommend changing the field to `SeenBefore` or `IsNew`. From there we can determine the behaviour based on the state.

protocol/migrations/sqlite/1632567059_add_highlight_column_in_chats.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE chats ADD COLUMN highlight BOOLEAN DEFAULT TRUE;
```
Samyoul: `BOOLEAN DEFAULT TRUE` This will produce unwelcome results, this change will update all existing chats with the highlight field set to `TRUE`. Meaning all chats will be highlighted

Parveshdhull: Great Catch 👏, Thanks

Parveshdhull:
> Hey @Parveshdhull thank you for your PR. Looks good.
> 
> I think that naming the field `Highlight` may be a little meta, what we really want to communicate is whether the chat is new or if we've seen it before. So I'd recommend changing the field to `SeenBefore` or `IsNew`. From there we can determine the behaviour based on the state.

Hi @Samyoul, Thanks for the review. I made the change you requested.
Also, `IsNew` sounded great in starting, but now I included a new case where chat is being highlighted for community invites, even though chat is not new, so that will create some confusion, I think 🤔

cammellos (approved):
lgtm :+1:

PR #2385, opened 2021-09-27
Upgrade status-go ethereum

cammellos:
Restores ip throttling, which we left over during the upgrade.


Samyoul (approved):

jakubgs (approved):
Yep, we used to have this fix but we lost it during Go-Ethereum upgrade:

* https://github.com/status-im/status-go/pull/2146
* https://github.com/status-im/go-ethereum/pull/91

Thanks!

PR #2395, opened 2021-10-01
revert mentions for chat preview

flexsurfer:
revert mentions for chat preview


ghost:
### Pull Request Checklist

protocol/messenger_chats.go
```diff
@@ -51,11 +53,15 @@ func (m *Messenger) ChatsPreview() []*ChatPreview {
...
 				chatPreview.ContentType = chat.LastMessage.ContentType
 				if chat.LastMessage.ContentType == protobuf.ChatMessage_TEXT_PLAIN {
-					if len(chat.LastMessage.Text) > 200 {
-						chatPreview.Text = chat.LastMessage.Text[:200]
+					if len(chat.LastMessage.Text) > 100 {
+						chatPreview.Text = chat.LastMessage.Text[:100]
 					} else {
 						chatPreview.Text = chat.LastMessage.Text
 					}
+					if strings.Contains(chatPreview.Text, "@0x") {
```
cammellos: I think there's a field already, `Mentions` that you should be able to check, something like `len(chatPreview.Mentions)>0`

cammellos: on `LastMessage`

flexsurfer: i mean if there is no visible mention we don't need to send it, we want to send only if it's in the first 100 chars

cammellos: Oh I see, you are right, I missed that. You might still want to do it performance wise though, as 
`len(chatPreview.LastMessage.Mentions) > 0 && strings.Contains(chatPreview.Text, "@0x")` should be faster in most cases. Though not sure if it's a necessary optimization.

cammellos: What about markdown btw?

flexsurfer: this PR is only for mentions

flexsurfer: markdown is not so critical in the preview imho

flexsurfer: added GetSimplifiedText for markdown, thanks

flexsurfer:
ping @cammellos

cammellos (approved):

PR #2397, opened 2021-10-05
feat: add batch information and request ID in history requests

richard-ramos:
With the idea of adding a progress bar in desktop, I added a request id and emit a signal each time a mailserver request batch is processed, as well as some attributes that could be used to guesstimate the progress


iurimatias:
@richard-ramos can you rebase please

iurimatias (approved):

PR #2405, opened 2021-10-12
[pairing] Sync read messages

rasom:
status: ready


ghost:
### Pull Request Checklist

Samyoul (approved):
All this looks good. What happens if the chat receives a message with a clock value of before the chat `ReadMessagesAtClockValue`?

protocol/migrations/sqlite/1633952865_add_clock_read_at_column_in_chats.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE chats ADD COLUMN read_messages_at_clock_value INT NOT NULL DEFAULT 0;
```
cammellos: This might cause some issue on migrations, as the value would be null for existing rows (or it might throw an error on migration), and then we try to put a value in a non nullable field.
Safer if you drop the NOT NULL and update immediately after (or use nullable values in the query).

```
ALTER TABLE chats ADD COLUMN read_messages_at_clock_value INT DEFAULT 0;
UPDATE chats SET read_messages_at_clock_value = 0;
```

rasom:
@Samyoul it is marked as `seen`

protocol/migrations/sqlite/1633952865_add_clock_read_at_column_in_chats.up.sql
rasom: done, thanks

PR #2407, opened 2021-10-13
Add endpoint for paginating on both sides

cammellos:
it's still a bit wip, but I would like to merge to avoid rebasing, as it moves some endpoints to `messenger_messages.go`


Samyoul (approved):

churik:
@cammellos is PR still relevant?

PR #2408, opened 2021-10-16
Online status indicator

Parveshdhull:
This PR further extends the User Status feature:

Original PR/specs: https://github.com/status-im/status-go/pull/2276

### Important fields
- current-user-status - This field contains the value of visibility status of the current user
- send-status-updates? - This field controls if a user should broadcast their status or not

### Important functions
- wakuext_setUserStatus - Use this function to set current user status. This will also update clock value
- settings_saveSetting - Use this function to modify `send-status-updates?` field.

### Flow
Status Updates are received using `messages.new` signals. For contacts and also for members of joined communities.

### Status-Types and requirements
- AUTOMATIC
			_To Send_ - `AUTOMATIC` status ping every 5 minutes
			_Display_ - Online for up to 5 minutes from the last clock, after that Offline
- ALWAYS_ONLINE
			_To Send_ - `ALWAYS_ONLINE` status ping every 5 minutes
			_Display_ - Online for up to 2 weeks from the last clock, after that Offline
- INACTIVE
			_To Send_ - A single `INACTIVE` status ping
			_Display_ - Offline forever

Note: Only send pings if the user interacted with the app in the last x minutes

#### status-react PR:
https://github.com/status-im/status-react/pull/12716


richard-ramos (approved):

protocol/messenger_status_updates.go
```diff
@@ -199,7 +199,10 @@ func (m *Messenger) SetUserStatus(ctx context.Context, newStatus int, newCustomT
 		return fmt.Errorf("custom text shouldn't be longer than %d", maxStatusMessageText)
 	}
 
-	if newStatus != int(protobuf.StatusUpdate_ONLINE) && newStatus != int(protobuf.StatusUpdate_DO_NOT_DISTURB) {
+	if newStatus != int(protobuf.StatusUpdate_AUTOMATIC) &&
```
cammellos, line 202: Here you can simplify, basically protobuf will default to `0` (i.e `UNKNOWN_STATUS_TYPE`), if the status is not any of the one defined in the protobuf, so you could in theory check for `if newStatus == UNKNOW...`.
Though if anyone adds one to the protobufs, and forgets to update this code, then it would try to process it, so this is a bit safer I suppose.

protocol/persistence.go
```diff
@@ -878,9 +878,9 @@ func (db sqlitePersistence) InsertStatusUpdate(userStatus UserStatus) error {
 
 func (db sqlitePersistence) CleanOlderStatusUpdates() error {
 	now := time.Now()
-	oneHourAgo := now.Add(time.Duration(-1) * time.Hour)
+	twoWeeksAgo := now.Add(time.Duration(-1*24*14) * time.Hour)
```
cammellos: I would probably define `24*14`  somewhere in a var

cammellos (approved):

Parveshdhull:
**PR Status:** Waiting for QA's approval on *React PR* before merging 👍

PR #2410, opened 2021-10-20
Send profile picture in private chat

cammellos:
2 commits:

1) We send pictures in private chat, so that the user receiving the message has more info on whether they want to open the chat, before is put in the activity center.

2) Fixes an issue with migrations, it seems like we deployed at some point a waku2 migration out of order, which results in the account being locked if waku2 is selected (we replicated on 1.16, and it only affects users who installed 1.14 -> 1.15 -> 1.16). 
The fix is to add the same migration with an `IF NOT EXISTS` clause, so that both client are able to run it.


ghost:
### Pull Request Checklist

Samyoul (approved):

PR #2413, opened 2021-10-21
Backup contacts through waku

cammellos:
Add backing up of contacts
    
We periodically check whether we should backup contacts through waku.
    
Currently it's not tied to any event, it will only schedule a backup on
 a tick, or through the provided API endpoint.

The protobuf message sent over is just a wrapper around the sync contact message. We batch the number of contacts to avoid a too large protobuf.

A signal is sent to the frontend when a backup is performed.

It also gets ride of the old `SaveContact` endpoint in favor of `AddContact`/`UnblockContact`..., so it's a breaking change.

cc @richard-ramos


protocol/messenger_backup.go
```diff
@@ -0,0 +1,161 @@
...
+					continue
+				}
+				if !enabled {
+					m.logger.Debug("backup not enabled, skipping")
+					continue
+				}
+
+				lastBackup, err := m.lastBackup()
+				if err != nil {
+					m.logger.Error("failed to fetch last backup time")
```
richard-ramos, line 56: Is it okay to not `continue` if there's an error fetching the last backup time? the next condition would always be true and backup the data.

protocol/messenger_backup.go
```diff
@@ -0,0 +1,161 @@
...
+				}
+
+				now := time.Now().Unix()
+				if uint64(now) <= backupIntervalSeconds+lastBackup {
+					m.logger.Debug("not backing up")
+					continue
+				}
+				m.logger.Debug("backing up data")
+
+				_, err = m.BackupData(context.Background())
```
richard-ramos: Would it make sense to use here a context.WithTimeout? or is this operation fast enough to not worry about it?

cammellos: probably best to give it a timeout, good point

protocol/messenger_backup.go
cammellos: ah no, missed the `continue` statement, good catch!

cammellos:
@richard-ramos putting this back in draft, there's a bit more work required for syncing contacts that added us. thanks for reviewing in the meantime!

cammellos:
@Samyoul @richard-ramos ready to be reviewed again, thanks!

protocol/protobuf/pairing.proto
```diff
@@ -3,6 +3,13 @@ syntax = "proto3";
 option go_package = "./;protobuf";
 package protobuf;
 
+message Backup {
```
Samyoul, line 6: Little name suggestion, we may wish to backup more than just contacts.

```suggestion
message ContactsBackup {
```

protocol/messenger_backup_test.go
```diff
@@ -0,0 +1,272 @@
...
+
+func (s *MessengerBackupSuite) TestBackupContactsGreaterThanBatch() {
+	bob1 := s.m
+	// Create bob2
+	bob2, err := newMessengerWithKey(s.shh, bob1.identity, s.logger, nil)
+	s.Require().NoError(err)
+	_, err = bob2.Start()
+	s.Require().NoError(err)
+
+	// Create 2 contacts
```
Samyoul: Looks like you create more than just 2 contacts here

protocol/messenger_backup_test.go
```diff
@@ -0,0 +1,272 @@
...
+		s.Require().Equal(actualContacts[0].ID, contactID2)
+		s.Require().Equal(actualContacts[1].ID, contactID1)
+	}
+	lastBackup, err := bob1.lastBackup()
+	s.Require().NoError(err)
+	s.Require().NotEmpty(lastBackup)
+	s.Require().Equal(clock, lastBackup)
+}
+
+func (s *MessengerBackupSuite) TestBackupContactsGreaterThanBatch() {
```
Samyoul, line 140: I came here looking for this test

Samyoul (approved):
This looks great.

One point I've got is actually naming related. Do we intend on having other kinds of backup? There are a number of functions that have the name "backup" or "handleBackup", but probably should be named "contactBackup" to be explicit.

The only other point I've got it that this feature is not quite a backup, functionally it works like a backup but only because history nodes store the messages and keep them alive in the network. But this is still a kind of ephemeral storage. I know you know this. I'm thinking of a case where the user device is destroyed or lost but the user doesn't reactivate their status keys until 32 days after the last backup. The restore from backup in this case would fail.

I like the implementation, its a bit like keeping a balloon in the air above a bed of nails 😆 . Are we working towards hardening the backup data?

cammellos:
@richard-ramos This has been tested and merged, mind that is a breaking change, let me know if you need any help with desktop

PR #2418, opened 2021-11-02
http server for images

flexsurfer:
[EXPERIMENT] http server for images


protocol/images/server.go
```diff
@@ -0,0 +1,108 @@
...
+		s.logger.Error("no publicKey")
+		return
+	}
+	pk := pks[0]
+	image, err := identicon.Generate(pk)
+	if err != nil {
+		s.logger.Error("could not generate identicon")
+	}
+
+	w.Header().Set("Content-Type", "image/png")
```
richard-ramos, line 116: Would it make sense to add headers like these so the browser keeps the identicons in cache?:
```go
w.Header().Set("Cache-Control", "max-age:290304000, public")
w.Header().Set("Expires", time.Now().AddDate(60, 0, 0).Format(http.TimeFormat))
```

protocol/images/server.go
```diff
@@ -0,0 +1,108 @@
...
+func (s *identiconHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	pks, ok := r.URL.Query()["publicKey"]
+	if !ok || len(pks) == 0 {
+		s.logger.Error("no publicKey")
+		return
+	}
+	pk := pks[0]
+	image, err := identicon.Generate(pk)
+	if err != nil {
+		s.logger.Error("could not generate identicon")
```
richard-ramos, line 113: Could be interesting to add a default image placeholder here for cases where the identicon could not be generated

protocol/images/server.go
```diff
@@ -0,0 +1,108 @@
...
+	db     *sql.DB
+}
+
+func NewServer(db *sql.DB, logger *zap.Logger) *Server {
+	return &Server{db: db, logger: logger}
+
+}
+
+func (s *Server) Start() error {
+	listener, err := net.Listen("tcp", ":0")
```
richard-ramos:
```suggestion
	listener, err := net.Listen("tcp", "localhost:0")
```

0kok0: Would it also be possible to use net.http.ListenAndServeTLS instead or a [tls.listener](https://pkg.go.dev/crypto/tls#NewListener)? TLS with self signed certs would harden the architecture here, also it wouldn't be a problem if at some point in time the server is moved out of the local application.

cammellos: Thanks for the review, sorry for the late reply, I missed the comment.
I'll look into it, we can definitely use a self signed certificate in the go code, but the tricky bit is to have the client accept a self signed certificate, I'll have a look.
So far I was looking at https://developer.apple.com/documentation/foundation/url_loading_system/handling_an_authentication_challenge/performing_manual_server_trust_authentication for IOS, but I'll have to spend a bit more time researching.

protocol/images/server.go
```diff
@@ -0,0 +1,189 @@
...
+		DNSNames:              []string{"localhost"},
+		KeyUsage:              x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+		BasicConstraintsValid: true,
+		IsCA:                  true,
+	}
+
+	derBytes, err := x509.CreateCertificate(rand.Reader, &template, &template, &priv.PublicKey, priv)
+	if err != nil {
+		log.Fatalf("Failed to create certificate: %v", err)
```
cammellos: Do you not want to return in this case? similarly below, maybe we should just make the FN return an error and abort if we cant' create?

bitgamma: right, I forgot about that, sorry. I don't expect errors to ever happen since there is no user input among the parameters, but better safe than sorry. Fixed now.

bitgamma:
TODO: (reminder to self) bump the VERSION file right before mering

cammellos (approved):

bitgamma:
@qoqobolo ISSUE 5 should be fixed now

0kok0 (commented):
neat work on the tls

PR #2424, opened 2021-11-05
feat: Push messages to telemetry server when receiving messages

alaibe:
Add push to telemetry server when receiving message, this needs to be enabled via settings


appdatabase/migrations/sql/1635942153_add_telemetry_server_url_to_settings.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE settings ADD COLUMN telemetry_server_url VARCHAR NOT NULL DEFAULT "";
```
cammellos: I think this might be problematic on upgrades, as the row will have a NULL value (defaults are not applied retrospectively).
Definitely you will have issues on upgrade with the code below, as it will try to read a `NULL` into a non-nullable `string`.
Generally you have 2 options, either you use a `sql.NullString` and check if it's valid when scanning the row, or you can just make sure it's always there, by setting it in the migration, so add a line:
`UPDATE settings SET telemetry_server_url = ""`

That way you are guaranteed that it's not null on existing clients.

telemetry/client.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (c *Client) PushReceivedMessages(messages map[transport.Filter][]*types.Message) {
+	c.logger.Debug("Pushing received messages to telemetry server")
+	url := fmt.Sprintf("%s/received-messages", c.serverURL)
+	for filter, messages := range messages {
+		for _, message := range messages {
+			postBody, _ := json.Marshal(map[string]interface{}{
+				"chatId":         filter.ChatID,
+				"messageHash":    string(message.Hash),
```
cammellos: I'd check that `string(message.Hash)` gives us the `0x...` representation, as that's the one we probably want as "canonical"

cammellos (approved):
Looks good, just left a few minor points.

Some observations:

This will catch any message that has been received and successfully pulled from the filters.
  We are also not logging the application `messageID`.
That's probably fine for now, but we might want to expand eventually to also log the `messageID` if it makes it to that point.

One thing that we are not logging though and it might be a source of messages been dropped is those messages that made it through to the waku layer but they were dropped there, so I think we should log messages as soon as they come in to waku.

The reason that they might be dropped is for example if they don't match any filter, or there's a timestamp mismatch (too old or too recent), so we could eventually log those as well as we might not have the full picture, but we can address separately.

telemetry/client.go
alaibe: The message hash doesn't look like 0x.... I could move the push after this line:
```
statusMessages, acks, err := m.sender.HandleMessages(shhMessage, true)
```

So that we have the message id available at that time and use this. Let me know what you think

alaibe:
> The reason that they might be dropped is for example if they don't match any filter, or there's a timestamp mismatch (too old or too recent), so we could eventually log those as well as we might not have the full picture, but we can address separately. 

I think we will need more push than one, cause only one datapoint cannot be enough to understand a problem. As soon as we have the first result we can start to add more datapoint depending on the result. if result are good it means that the messages are loss after this one, if they are not good it means that we are already loosing data up to that point. Does it makes sense?

If you think this one won't provide good value, I can move it to another place (just not sure where) feel free to suggest ;)

appdatabase/migrations/sql/1635942153_add_telemetry_server_url_to_settings.up.sql
alaibe: Good point, I added the update in the migration

telemetry/client.go
cammellos: We probably want to have both, you can use `types.EncodeToHex(message.Hash)` to have the hex encoded representation.
Pushing after handle message is useful as well, thought we might drop them in between :) but that's unlikely, in general for the first pass I would push it after, so we can log the application message ID as well

cammellos:
> > The reason that they might be dropped is for example if they don't match any filter, or there's a timestamp mismatch (too old or too recent), so we could eventually log those as well as we might not have the full picture, but we can address separately.
> 
> I think we will need more push than one, cause only one datapoint cannot be enough to understand a problem. As soon as we have the first result we can start to add more datapoint depending on the result. if result are good it means that the messages are loss after this one, if they are not good it means that we are already loosing data up to that point. Does it makes sense?
> 
> If you think this one won't provide good value, I can move it to another place (just not sure where) feel free to suggest ;)

I think it's ok for the first pass, I agree, we can see how it goes and add further if we still have data loss and no explanation

telemetry/client.go
alaibe: Alright, I will move the call to after and add the message ID, thanks for the input ;)

alaibe: Ok cool, this is change has been made, we now log the message id (kept the hash) and also the node name in case of mobile/desktop running for the same account

richard-ramos (approved):

PR #2428, opened 2021-11-11
Allow sending messages offline (non-datasync messages)

rasom:
status: ready


protocol/messenger.go
```diff
@@ -139,9 +139,12 @@ func (interceptor EnvelopeEventsInterceptor) EnvelopeSent(identifiers [][]byte)
 		err := interceptor.Messenger.processSentMessages(ids)
 		if err != nil {
 			interceptor.Messenger.logger.Info("Messenger failed to process sent messages", zap.Error(err))
+		} else {
+			interceptor.EnvelopeEventsHandler.EnvelopeSent(identifiers)
 		}
+	} else {
```
cammellos, line 146: This I am not sure I understand, would you mind leaving a comment to explain why it's an `if or else`?

rasom: Soooo.... right now `processSentMessages` can return error when connection state is offline which means we don't want to execute `interceptor.EnvelopeEventsHandler.EnvelopeSent` but if there is no `else` it will be called anyway. I can move connection state check to `EnvelopeSent` but having a separate `if` or `else` is probably unavoidable in this case.

cammellos: I think it's fine, maybe if you could leave the explanation in a comment in the code that's enough, thank you

rasom: ok

protocol/messenger.go
```diff
@@ -415,32 +422,42 @@ func (m *Messenger) processSentMessages(ids []string) error {
...
 	}
 
-	if message.SendCount > emojiResendMaxCount {
+	if message.SendCount > messageResendMaxCount {
 		return false, nil
 	}
 
 	//exponential backoff depends on how many attempts to send message already made
-	backoff := uint64(math.Pow(2, float64(message.SendCount-1))) * emojiResendMinDelay * uint64(time.Second)
+	backoff := uint64(math.Pow(2, float64(message.SendCount-1))) * messageResendMinDelay * uint64(time.Second.Milliseconds())
```
Samyoul, line 455: Why is the timing done in milliseconds and not nanoseconds? Just curious.

Samyoul (approved):
I am happy to see the generalisation of a specific function.

protocol/messenger.go
rasom: Probably because `t.GetCurrentTime()` and `message.LastSent` are in milliseconds. Previously there was a mix of milliseconds and nanoseconds  because of which this code never worked as expected (the message would be resent only in ~350 days).

protocol/messenger.go
rasom: @cammellos done

PR #2429, opened 2021-11-12
Ensure owner is removed from community when leaving community

0x-r4bbit:
Prior to this commit, when a community owner leaves her community,
status-go would just set the `joined` flag to `false` without updating the
members list. This results in wrong data as the number of members doesn't
match the reality.

This commit removes the owner of the community from the members list
when she's leaving the community.

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -726,8 +726,15 @@ func (m *Manager) LeaveCommunity(id types.HexBytes) (*Community, error) {
 	if community == nil {
 		return nil, ErrOrgNotFound
 	}
+	if community.IsAdmin() {
+		_, err = community.RemoveUserFromOrg(m.identity)
+		if err != nil {
+			return nil, err
+		}
+	}
 	community.Leave()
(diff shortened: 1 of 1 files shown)


protocol/communities/manager.go
```diff
@@ -726,8 +726,10 @@ func (m *Manager) LeaveCommunity(id types.HexBytes) (*Community, error) {
 	if community == nil {
 		return nil, ErrOrgNotFound
 	}
+  community.RemoveUserFromOrg(m.identity)
```
0x-r4bbit: Notice that I'm not handling possible errors here. When this API is called we know that the user exists in the org so removal should always go fine.

alaibe: nit: github shows some indentation issue

alaibe (approved):

protocol/communities/manager.go
0x-r4bbit: Fixed!

cammellos (approved):

PR #2436, opened 2021-11-22
Reject contact request endpoint

cammellos:
Adds RejectContactRequest endpoint as used by desktop.

cc @PascalPrecht


protocol/requests/reject_contact_request.go
```diff
@@ -0,0 +1,21 @@
...
+	"errors"
+
+	"github.com/status-im/status-go/eth-node/types"
+)
+
+var ErrRejectContactRequestInvalidID = errors.New("reject-contact-request: invalid id")
+
+type RejectContactRequest struct {
+	ID types.HexBytes `json:"id"`
+}
```
0x-r4bbit, line 13: Out of curiosity: It seems `BlockContact` and `UnblockContact` APIs as an example` take just and `id`, while this API requires a struct. Do you think there's gonna be more properties added in the future for this type, or should mentioned APIs also ask for a serialized struct?

cammellos: It's mostly for consistency and it's easier to document them, they could take an ID as well. 
I don't really expect them to take more arguments, but who knows :) 
I don't have a strong opinion though, either is fine by me

0x-r4bbit: Alrighty okay, thanks for clarifying!

0x-r4bbit (approved):

PR #2438, opened 2021-11-22
feat: waku2 - use identify protocol

richard-ramos:
**BREAKING CHANGE**

This would use identify protocol to determine what protocols are supported by the waku2 peers, instead of having separate buckets in the configuration.

The node config in desktop and mobile should be changed from:

```js
ClusterConfig: {
    RelayNodes: ["/ip4/127.0.0.1/udp/1234",  "enstree:somednsdiscoveryurl..."],
    StoreNodes: ["/ip4/127.0.0.1/udp/1234",  "enstree:somednsdiscoveryurl..."],
    FilterNodes: ["/ip4/127.0.0.1/udp/1234",  "enstree:somednsdiscoveryurl..."],
    LightpushNodes: ["/ip4/127.0.0.1/udp/1234",  "enstree:somednsdiscoveryurl..."],
    WakuRendezvousNodes: ["/ip4/127.0.0.1/udp/1234",  "enstree:somednsdiscoveryurl..."]
}
```

to

```js
ClusterConfig: {
    WakuNodes: ["/ip4/127.0.0.1/udp/1234",  "enstree:somednsdiscoveryurl..."]
}
```

cc: @jm-clius @oskarth


wakuv2/waku.go
```diff
@@ -317,51 +303,96 @@ func (w *Waku) dnsDiscover(enrtreeAddress string, protocol libp2pproto.ID, apply
...
+
+	err = identifyService.Host.Connect(ctx, *peerInfo)
+	if err != nil {
+		log.Error("could not extract peerinfo", zap.Any("ma", ma), zap.Error(err))
+		return
+	}
+
+	conns := identifyService.Host.Network().ConnsToPeer(peerInfo.ID)
+	if len(conns) == 0 {
+		return // No connection
```
alaibe, line 424: shall we return an error in that case?

alaibe (approved):

wakuv2/waku.go
richard-ramos: hm, maybe logging an error but returning an error i think it's not necessary. I'm not sure how would I handle a case where the connection to a peer was succesful, yet libp2p Network() says that there's no connection to that peer

churik:
@richard-ramos still relevant?

richard-ramos:
@churik yes. I'll rebase it today

PR #2441, opened 2021-11-24
feat: auto message

alaibe:
Send message every 2 minutes to status-bot chat id


richard-ramos (approved):

cammellos (changes requested):
Just to understand, is this only for testing? 
We have a bot already that is sending as an external command, we could expand that one.
In `cmd/ping-community`

alaibe:
@cammellos yes, this is for testing,this is related to our last call where we want each desktop user to have a flag that if enable, that will push a message every tick to a channel. At least this is what I understood.

I tried to see if i could reuse the command but as this should be started based on the flag I couldn t find a way to reuse the ping command, maybe I missed something, feel free to indicate me what I could do better

protocol/messenger_auto_message.go
```diff
@@ -0,0 +1,71 @@
...
+const chatID = "status-bot"
+
+func (m *Messenger) AutoMessageEnabled() (bool, error) {
+	return m.settings.AutoMessageEnabled()
+}
+
+func (m *Messenger) startAutoMessageLoop() {
+	enabled, err := m.AutoMessageEnabled()
+	if err != nil {
+		m.logger.Error("[auto message] failed to start auto message loop", zap.Error(err))
```
cammellos, line 27: I would make it return an error, this might indicate an actual problem (db is closed etc)

protocol/messenger_auto_message.go
```diff
@@ -0,0 +1,71 @@
...
+				timestamp := time.Now().Format(time.RFC3339)
+
+				msg := &common.Message{}
+				msg.Text = fmt.Sprintf("%d\n%s", count, timestamp)
+				msg.ChatId = chatID
+				msg.LocalChatID = chatID
+				msg.ContentType = protobuf.ChatMessage_TEXT_PLAIN
+				resp, err := m.SendChatMessage(context.Background(), msg)
+				if err != nil {
+					m.logger.Error("[auto message] failed to send message", zap.Error(err))
```
cammellos, line 52: You might want `continue` here, otherwise i'ts likely to panic

cammellos (commented):
I think it needs a bit of error handling, otherwise looks good

protocol/messenger_auto_message.go
```diff
@@ -0,0 +1,74 @@
...
+					continue
+				}
+				signal.SendNewMessages(resp)
+
+				err = m.UpdateMessageOutgoingStatus(msg.ID, common.OutgoingStatusDelivered)
+				if err != nil {
+					m.logger.Error("[auto message] failed to mark message as delivered", zap.Error(err))
```
cammellos, line 59: Do you want to send a signal even if this fails? or is it better to continue in this case?

cammellos (approved):

PR #2443, opened 2021-11-25
Sync activity center 

bitgamma:
Needed for https://github.com/status-im/status-react/issues/12571

Note to self: bump VERSION before merging


Parveshdhull (approved):

protocol/messenger_handler.go
```diff
@@ -139,8 +139,10 @@ func (m *Messenger) HandleMembershipUpdate(messageState *ReceivedMessageState, c
 		wasUserAdded = !existingGroup.IsMember(ourKey) &&
 			updateGroup.IsMember(ourKey)
 
+		joined := group.HasJoined(ourKey)
+
 		// Reactivate deleted group chat on re-invite from contact
-		chat.Active = chat.Active || (isActive && wasUserAdded)
+		chat.Active = chat.Active || (isActive && (wasUserAdded || joined))
```
Parveshdhull: Hi @bitgamma, Great work 👏
So group invite chat is only added to the home screen when joined. So if a device opens a group invite from the activity center and doesn't join it, wouldn't that be still shown in the activity center of the synced device?🤔

bitgamma: Hi @Parveshdhull thanks for the review. Yes, I think this is the desired behavior cause it is not moved to the home screen of the main device either for what I have seen

Parveshdhull: I didn't tried syncing yet, but looks like opening group invite is adding to home screen

https://user-images.githubusercontent.com/17097240/143572400-a3e7b5b3-3aeb-4ace-ac75-b9f04ee6721d.MP4

bitgamma: I must have done something wrong when testing then. The fix is even simpler in that case cause I can just make it active when syncing the message read event (now I explicitly filter one to one chats to exclude group chats. Thanks for noticing that!

PR #2444, opened 2021-11-26
fix: iterate through message with batch topics

alaibe:
When iterating, reuse the list of topics rather than all topics. This should reduce the size of the data we receive


cammellos (approved):
don't forget to bump `VERSION`, well spotted!

PR #2445, opened 2021-11-26
Wait for p2p messages to be processed

cammellos:
We now wait for all messages to be processed before marking the request as confirmed, so in case of crashes/quitting the app, those messages will be refetched from the mailserver.


protocol/transport/transport.go
```diff
@@ -219,15 +220,19 @@ func (t *Transport) RetrieveRawAll() (map[Filter][]*types.Message, error) {
...
-		// Don't pull from filters we don't listen to
-		if !filter.Listen {
-			continue
-		}
 		msgs, err := t.api.GetFilterMessages(filter.FilterID)
 		if err != nil {
 			logger.Warn("failed to fetch messages", zap.Error(err))
 			continue
 		}
+		// Don't pull from filters we don't listen to
```
alaibe, line 227: is the comment still relevant? it seems we do pull now. Also should we remove the code line 222? because we will pull for all filter?

cammellos: not relevant anymore, you are right

cammellos: and yes, I'll remove the code, thanks!

alaibe (approved):
LGTM!

Samyoul (approved):

PR #2448, opened 2021-11-29
fix: error reordering chats when destination category position is the same as original chatId position

richard-ramos:


Parveshdhull (approved):
**Tested** - working great 👍,
Thank you very much for fixing it quickly ❤️

PR #2449, opened 2021-11-30
feat: Mark messages as confirmed in waku2

alaibe:
Waku2 implementation

### DIFF
--- a/wakuv2/waku.go
+++ b/wakuv2/waku.go
@@ -123,6 +123,9 @@ type Waku struct {
 
 	envelopeFeed event.Feed
 
+	storeMsgIDs   map[gethcommon.Hash]bool // Map of the currently processing ids
+	storeMsgIDsMu sync.RWMutex
+
 	timeSource func() time.Time // source of time for waku
 
 	logger *zap.Logger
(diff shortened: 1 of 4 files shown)


wakuv2/waku.go
```diff
@@ -969,7 +974,14 @@ func (w *Waku) processQueue() {
 		case <-w.quit:
 			return
 		case e := <-w.msgQueue:
```
cammellos, line 976: is this queue for any messages? or only p2p messages? (i.e pulled from a store node/mailserver)

alaibe: AFAIK this queue is for any messages: pulled from mailserver or P2P.

@richard-ramos could confirm if I am not wrong

richard-ramos: Yes, it's for any message.

cammellos: Ok, so mind that you will not receive a "mailserver" completed signal until there are no more messages in processing, not just store/mailserver messages. In theory it might never end, if the queue is kept busy.

alaibe: Gotcha, I will exclude the mailserver messages from the `map`

cammellos: > Gotcha, I will exclude the mailserver messages from the map

You mean the opposite right? (the code looks correct)

alaibe: Synced in PM, Yeah I mean the opposite, also did a bit of renaming

richard-ramos (approved):

PR #2451, opened 2021-12-01
refactor(timeline) Removed timeline feature

alexandraB99:
Since timeline feature is depecated, added condition
in status-go to not save timeline and profile chats
in order to avoid displaying public chats of this type.

Relates to dekstop [#4177](https://github.com/status-im/status-desktop/issues/4177)

### DIFF
--- a/protocol/messenger.go
+++ b/protocol/messenger.go
@@ -1126,14 +1126,16 @@ func (m *Messenger) Init() error {
 			continue
 		}
 
-		m.allChats.Store(chat.ID, chat)
+                if (!chat.Timeline() && chat.ChatType != ChatTypeProfile) {
+                        m.allChats.Store(chat.ID, chat)
+                }
 
-		if !chat.Active || chat.Timeline() {
-			continue
-		}
+                if !chat.Active || chat.Timeline() {
+                        continue
+                }
 
-		switch chat.ChatType {
-		case ChatTypePublic, ChatTypeProfile:
+                switch chat.ChatType {
+                case ChatTypePublic, ChatTypeProfile:
 			publicChatIDs = append(publicChatIDs, chat.ID)
 		case ChatTypeCommunityChat:
 			publicChatIDs = append(publicChatIDs, chat.ID)


protocol/messenger.go
```diff
@@ -1126,14 +1126,16 @@ func (m *Messenger) Init() error {
 			continue
 		}
 
-		m.allChats.Store(chat.ID, chat)
+                if (!chat.Timeline() && chat.ChatType != ChatTypeProfile) {
+                        m.allChats.Store(chat.ID, chat)
+                }
```
0x-r4bbit, line 1131: I'm not familiar with the  `allChats.Store()` API, but I assume this affects which chats will be returned by the RPC API.  I think we need to be careful here because this will affect mobile as well. While I assume the timeline will be deprecated on a protocol level, desktop and mobile releases aren't in sync. 

So we should:

1. Either ensure with mobile that they won't be affected by this change in an unexpected way OR
2. Do what I originally proposed in https://github.com/status-im/status-desktop/pull/4198#discussion_r760106468.

> This means, what we'd need to do, is kind of the reverse of what is being done in step 3).
Instead of filling the gaps and add missing profile update channels, we should remove all chats in status-go that are of type Profile, so when we hit step 4), we don't actually create filters for those anymore.

As I'm reading this now, I see that this was probably misinterpreted. I didn't mean literally changing code in status-go, but removing channels from status-go via APIs called in desktop. That way we can safely remove the timeline feature in desktop without affecting mobile, nor the protocol in general.

0x-r4bbit (commented):
@alexandraB99 I think there was a misunderstanding with regards to "removing profile/timeline channels from status-go". Tried to clarify it with inline comments.

Lemme know what you think.

Otherwise, if we get green light from mobile, then this change is okay as well.

/cc @cammellos

protocol/messenger.go
0x-r4bbit: Added some more concrete comments here: https://github.com/status-im/status-desktop/pull/4198#discussion_r760931782

cammellos:
mmh, not sure we have properly made a decision about this, that's not something we were involved with, so I would hold on for this one.

0x-r4bbit:
Thanks for clarifying feedback!

John-44:
> mmh, not sure we have properly made a decision about this, that's not something we were involved with, so I would hold on for this one.

Copying comment from Discord: John_ — Today at 10:23
@cammellos remember we discussed removing this feature when communities launches a few months back?  The argument for keeping it in mobile for now is that for mobile, while we still have public unmoderated group chats, keeping it in would have very low maintenance overhead.  For desktop the situation is very different - if they want to keep this feature in, they would have to refactor it as part of the refactoring effort, and this seems like a waste of dev resource when we know we are going to pull this feature out of mobile in the future.  Given desktop is in beta, I don't think desktop not investing in refactoring this feature (and therefor removing it) while it remains in mobile until the launch of communities is necessarily a problem.  If it is felt that this (not having the feature in desktop, but the feature living on in mobile for a bit longer) is a problem, my instinct would be to remove it from mobile as well.  

@cammellos I've sent you an invite for this afternoon so we can discuss this again, then we can loop back on this thread.

John-44:
Screenshot of discussion from Discord on this topic

2021-12-02 11_02_33-mobile-ui - Discord

0x-r4bbit:
Thanks @John-44 for the transparency.

Let me just re-emphasize and also add to this that the change in status-go here was a misunderstanding(https://github.com/status-im/status-go/pull/2451#discussion_r760926144) :

> As I'm reading this now, I see that this was probably misinterpreted. I didn't mean literally changing code in status-go, but removing channels from status-go via APIs called in desktop. That way we can safely remove the timeline feature in desktop without affecting mobile, nor the protocol in general.

So if we go down the route I've proposed initial (although, probably wrongly articulated) the change here isn't needed at all and mobile can do its thing without being affected by this.

alexandraB99:
Confirming, this change is not needed anymore. I would close this PR if no objections.

0x-r4bbit:
Closing this as those changes aren't necessary

PR #2453, opened 2021-12-02
Add nil check for DirectMessageProtocol object

vitvly:


cammellos (approved):

protocol/encryption/protocol.go
```diff
@@ -477,12 +477,14 @@ func (p *Protocol) HandleMessage(
 		if dmProtocol == nil {
 			dmProtocol = encryptedMessage[noInstallationID]
 		}
-		hrHeader := dmProtocol.HRHeader
-		if hrHeader != nil && hrHeader.SeqNo == 0 {
-			// Payload contains hash ratchet key
-			err = p.encryptor.persistence.SaveHashRatchetKey(hrHeader.GroupId, hrHeader.KeyId, message)
-			if err != nil {
-				return nil, err
+		if dmProtocol != nil {
```
flexsurfer, line 480: it could be just else ?

vitvly: Not really, as `dmProtocol` can still be nil.

vitvly: There is an assignment in previous if.

flexsurfer: oh right

PR #2455, opened 2021-12-02
Decline pending group invites from user, when user is blocked

### LINKED ISSUE #12839: Decline non-answered group chat invitation when a user is blocked
# Feature Issue

leftover from https://github.com/status-im/status-react/issues/12108

As a user
When I block someone
I want existing group chat invitation to be declined
So that it does not clutter my home screen

### Description

When a user is blocked, existing group chat invitation that have not been accepted  should also be removed.

Parveshdhull:
fixes https://github.com/status-im/status-react/issues/12839

### Summary
PR implements a feature, which auto declines pending group invites from a user when he/she is blocked

#### status-react PR:
https://github.com/status-im/status-react/pull/12878


VERSION
```diff
@@ -1 +1 @@
-0.91.14
+0.91.15
```
cammellos: Looks like there's a breaking change in the API, we generally bump the `y` version for those, so it would be `0.92.0`

protocol/messenger_contacts.go
```diff
@@ -358,6 +358,28 @@ func (m *Messenger) BlockContact(contactID string) ([]*Chat, error) {
...
+		return nil, err
+	}
+	response.AddChats(chats)
+
+	response, err = m.DeclineAllPendingGroupInvitesFromUser(response, contactID)
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = m.persistence.AcceptAllActivityCenterNotificationsFromUser(contactID)
```
cammellos: Out of curiosity, why accepting them rather then dismissing them?

Parveshdhull: Yes, dismissing sounds better, will do that 👍, Thanks

Samyoul (approved):

PR #2457, opened 2021-12-03
drop VERSION file, use git tags instead

jakubgs:
In theory this should make it clear which change is which...

We can't do the same in `status-react` because there we use `git ls-remote` which does not have access to dates.


bitgamma (approved):
lgtm

igor-sirotin:
Fixed here: https://github.com/status-im/status-go/pull/5735

PR #2460, opened 2021-12-06
fix: remove mailserver topic when deactivating a chat

richard-ramos:
This fixes an issue where re-joining a chat does not load the messages again from the mailserver


richard-ramos:
Might be a good idea to test this with status-react!

0x-r4bbit (approved):
Changes LGTM but it seems like this is failing some tests

0x-r4bbit:
Seems to be still failing..

0x-r4bbit:
Once squashed, this seems to be gtg

0x-r4bbit (approved):

cammellos (approved):
Version change is missing, otherwise look good, would you mind testing it works as expected before merging?

richard-ramos:
I did test on deskop. But I do not know how to test this status-go version in status-react

cammellos:
@richard-ramos desktop is fine, if it works there it should work in status-react (in most cases), so no need to test there.
Thanks!

PR #2463, opened 2021-12-10
Send a contact advertisement when the profile picture is removed

richard-ramos:


saledjenic (approved):

protocol/messenger.go
```diff
@@ -719,11 +719,16 @@ func (m *Messenger) attachChatIdentity(cca *protobuf.ContactCodeAdvertisement) e
...
+		hash = crypto.Keccak256([]byte(m.account.Identicon))
+	} else {
+		hash = img.Hash()
 	}
 
-	err = m.persistence.SaveWhenChatIdentityLastPublished(contactCodeTopic, img.Hash())
+	err = m.persistence.SaveWhenChatIdentityLastPublished(contactCodeTopic, hash)
 	if err != nil {
 		return err
 	}
```
Samyoul, line 734: This seems like a bit of an antipattern. If we have no image we should probably save no hash rather than the hash of the identicon

Maybe something like below? What do you think?
```go
	if img == nil {
		return m.persistence.SaveWhenChatIdentityLastPublished(contactCodeTopic, []byte{})
	} else {
		return m.persistence.SaveWhenChatIdentityLastPublished(contactCodeTopic, img.Hash())
	}
```

Samyoul (changes requested):
I think that this approach is a bit prone to false positives. Simply omitting IdentityImages in a ChatIdentity shouldn't trigger a deletion from the db in my opinion.

I think that we need an explicit protobuf message type for removing IdentityImages from the db of other users.

If we are implementing this kind of feature we should also add tests to ensure we don't introduce any unintentional behaviour

richard-ramos:
Good idea. I'll change the code so we have a new message type instead of just sending empty images

PR #2465, opened 2021-12-14
feat: desktop browser provider

richard-ramos:


services/provider/api.go
```diff
@@ -0,0 +1,432 @@
+package provider
```
alaibe: Is provider not too abstract? Maybe web3provider?

services/provider/api.go
```diff
@@ -0,0 +1,432 @@
...
+	} else {
+		return api.web3Call(request)
+	}
+
+	/*
+
+		/*
+
+
+		   if data.payload.rpcMethod == "eth_sendTransaction":
```
alaibe: i can see we have same technique!

services/provider/api_test.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+}
+
+func setupTestAPI(t *testing.T) (*API, func()) {
+	db, cancel := createDB(t)
+
+	keyStoreDir, err := ioutil.TempDir(os.TempDir(), "accounts")
+
+	// Creating a dummy status node to simulate what it's done in get_status_node.go
+	upstreamConfig := params.UpstreamRPCConfig{
```
alaibe: so that means you do call for real?

services/provider/api.go
```diff
@@ -0,0 +1,432 @@
...
+		MessageID: request.MessageID,
+		Result: JSONRPCResponse{
+			JSONRPC: "2.0",
+			ID:      request.Payload.ID,
+			Result:  result,
+		},
+	}, nil
+}
+
+func (api *API) getVerifiedWalletAccount(address, password string) (*account.SelectedExtKey, error) {
```
alaibe: I need to duplicate this function and it is already a copy/paste of an existing function gethbackend. I think we should find a way to not duplicate that. We have an account module already. Shall we move that there?

services/provider/api_test.go
richard-ramos: yeah!

alaibe (approved):

PR #2467, opened 2021-12-16
Feat/ens provider

alaibe:
Add ens service based on desktop needs

### DIFF
--- a/services/ens/api.go
+++ b/services/ens/api.go
@@ -1,0 +1,321 @@
+package ens
+
+import (
+	"context"
+	"encoding/binary"
+	"encoding/hex"
+	"fmt"
+	"math/big"
+
+	"github.com/ipfs/go-cid"
+	"github.com/multiformats/go-multibase"
+	"github.com/multiformats/go-multihash"
+	"github.com/pkg/errors"
+	"github.com/wealdtech/go-multicodec"
+
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/status-im/status-go/rpc"
+)
+
+func NewAPI(rpcClient *rpc.Client) *API {
+	return &API{
+		contractMaker: &contractMaker{
+			rpcClient: rpcClient,
+		},
+	}
+}
+
+type uri struct {
+	Scheme string
+	Host   string
+	Path   string
+}
+
+type publicKey struct {
+	X [32]byte
+	Y [32]byte
+}
+
+type API struct {
+	contractMaker *contractMaker
+}
+
+func (api *API) Resolver(ctx context.Context, chainID uint64, username string) (*common.Address, error) {
+	err := validateENSUsername(username)
+	if err != nil {
+		return nil, err
+	}
+
+	registry, err := api.contractMaker.newRegistry(chainID)
+	if err != nil {
+		return nil, err
+	}
+
+	callOpts := &bind.CallOpts{Context: ctx, Pending: false}
+	resolver, err := registry.Resolver(callOpts, nameHash(username))
+	if err != nil {
+		return nil, err
+	}
+
+	return &resolver, nil
+}
+
+func (api *API) OwnerOf(ctx context.Context, chainID uint64, username string) (*common.Address, error) {
+	err := validateENSUsername(username)
+	if err != nil {
+		return nil, err
+	}
+
+	registry, err := api.contractMaker.newRegistry(chainID)
+	if err != nil {
+		return nil, err
+	}
+
+	callOpts := &bind.CallOpts{Context: ctx, Pending: false}
+	owner, err := registry.Owner(callOpts, nameHash(username))
+	if err != nil {
+		return nil, nil
+	}
+
+	return &owner, nil
+}
+
+func (api *API) ContentHash(ctx context.Context, chainID uint64, username string) ([]byte, error) {
+	err := validateENSUsername(username)
+	if err != nil {
+		return nil, err
+	}
+
+	resolverAddress, err := api.Resolver(ctx, chainID, username)
+	if err != nil {
+		return nil, err
+	}
+
+	resolver, err := api.contractMaker.newPublicResolver(chainID, resolverAddress)
+	if err != nil {
+		return nil, err
+	}
+
+	callOpts := &bind.CallOpts{Context: ctx, Pending: false}
+	contentHash, err := resolver.Contenthash(callOpts, nameHash(username))
+	if err != nil {
+		return nil, nil
+	}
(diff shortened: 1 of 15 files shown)


services/ens/README.md
```diff
@@ -0,0 +1,48 @@
...
+}
+```
+
+#### browsers_getBrowsers
+
+Reads all browsers, returns in the format specified above. List is sorted by timestamp.
+
+#### browsers_deleteBrowser
+
+Delete browser from database. Accepts browser `id`.
```
richard-ramos:
```suggestion
```

services/ens/api.go
```diff
@@ -0,0 +1,312 @@
...
+	callOpts := &bind.CallOpts{Context: ctx, Pending: false}
+	contentHash, err := resolver.Contenthash(callOpts, nameHash(username))
+	if err != nil {
+		return nil, nil
+	}
+
+	return contentHash, nil
+}
+
+func (api *API) PublicKeyOf(ctx context.Context, chainID uint64, username string) ([32]byte, [32]byte, error) {
```
richard-ramos: I wonder if it would be a good idea to instead of returning X and Y, to return instead a `ecdsa.PublicKey`?

services/ens/api.go
```diff
@@ -0,0 +1,312 @@
...
+	price, err := registrar.GetPrice(callOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	return price, nil
+}
+
+func (api *API) Release(ctx context.Context, chainID uint64, from string, gasPrice *big.Int, gasLimit uint64, password string, username string) (string, error) {
+	err := validateENSUsername(username)
```
richard-ramos: It's probably a better idea to receive a `transactions.SendTxArgs` instead of  from `string, gasPrice *big.Int, gasLimit uint64, password string`. `SendTxArgs` also has supports for other trx fields like the Nonce, and EIP1559 MaxFeePerGas  and MaxPriorityFeePerGas

richard-ramos: If you use the transactions.SendTxArgs, you'd need a mapping function to convert from that struct to a `bind.TransactOpts`

services/ens/api.go
```diff
@@ -0,0 +1,312 @@
...
+	txOpts := &bind.TransactOpts{
+		From: common.HexToAddress(from),
+		Signer: func(addr common.Address, tx *types.Transaction) (*types.Transaction, error) {
+			// return types.SignTx(tx, types.NewLondonSigner(chainID), selectedAccount.AccountKey.PrivateKey)
+			return nil, nil
+		},
+		GasPrice: gasPrice,
+		GasLimit: gasLimit,
+	}
+	tx, err := registrar.Release(txOpts, nameHash(username))
```
richard-ramos: I noticed that when a sendTransaction is executed in status-go, these functions are called:
```
hash, err = api.s.transactor.SendTransaction(sendArgs, verifiedAccount)
if err != nil {
	return
}
go rpcFiltersSrvc.TriggerTransactionSentToUpstreamEvent(hash)
```
The first part seems to be covered by `registrar.Release(txOpts, nameHash(username))`, but it probably makes sense to call the `TriggerTransactionSentToUpstreamEvent` go routine. IIRC it's used so the wallet knows about the transaction and you eventually receive a signal once the transaction is mined

services/ens/contracts.go
```diff
@@ -0,0 +1,63 @@
...
+	"errors"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/status-im/status-go/rpc"
+	"github.com/status-im/status-go/services/ens/registrar"
+	"github.com/status-im/status-go/services/ens/resolver"
+)
+
+var resolversByChainID = map[uint64]common.Address{
+	1: common.HexToAddress("0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e"),
```
richard-ramos, line 13: Missing:
```
Ropsten, at 0x112234455c3a32fd11230c42e7bccd4a84e02010.
Rinkeby, at 0xe7410170f87102df0055eb195163a03b7f2bff4a.
Goerli, at 0x112234455c3a32fd11230c42e7bccd4a84e02010.
```

richard-ramos: Missing the chainId for Ropsten, Rinkeby and Goerli,  (the address is the same)

richard-ramos: Ah no wait. I'm confused. Here it says https://docs.ens.domains/ens-deployments that `0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e` is a registrar. Maybe it's a case of wrong variable name?

services/ens/contracts.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+var registrarsByChainID = map[uint64]common.Address{
```
richard-ramos:
```suggestion
var usernameRegistrarsByChainID = map[uint64]common.Address{
```

richard-ramos: (this is so we know this refers to status username registrar, and not referring to the ens registrars)

services/ens/contracts.go
```diff
@@ -0,0 +1,63 @@
...
+	1: common.HexToAddress("0xDB5ac1a559b02E12F29fC0eC0e37Be8E046DEF49"),
+}
+
+type contractMaker struct {
+	rpcClient *rpc.Client
+}
+
+func (c *contractMaker) newRegistry(chainID uint64) (*resolver.ENSRegistryWithFallback, error) {
+	if _, ok := resolversByChainID[chainID]; !ok {
+		return nil, errors.New("not available for chainID")
```
richard-ramos:
```go
var ErrorNotAvailableOnChainID = errors.New("not available for chainID")
```

```suggestion
		return nil, ErrorNotAvailableOnChainID
```

services/ens/contracts.go
```diff
@@ -0,0 +1,63 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return resolver.NewPublicResolver(*resolverAddress, backend)
+}
+
+func (c *contractMaker) newUsernameRegistrar(chainID uint64) (*registrar.UsernameRegistrar, error) {
+	if _, ok := registrarsByChainID[chainID]; !ok {
+		return nil, errors.New("not available for chainID")
```
richard-ramos:
```suggestion
		return nil, ErrorNotAvailableOnChainID
```

services/ens/api.go
alaibe: Commented for now til I come back to this

services/ens/api.go
alaibe: :+1:

richard-ramos (approved):

PR #2468, opened 2021-12-17
Set joined=message.clock on syncing pubchat creation

rasom:
needed for https://github.com/status-im/status-react/issues/12913

In case if current timestamp is set to `chat.joined` but there is a newer `SyncChatRemoved` message, the latter might be ignored as `chat.joined` bigger than `message.clock`. 

status: ready


Parveshdhull (approved):

bitgamma (approved):

protocol/messenger_handler.go
```diff
@@ -421,8 +421,9 @@ func (m *Messenger) HandleSyncInstallationPublicChat(state *ReceivedMessageState
 	chat := existingChat
 	if !ok {
 		chat = CreatePublicChat(chatID, state.Timesource)
+		chat.Joined = int64(message.Clock)
 	} else {
-		existingChat.Joined = int64(state.Timesource.GetCurrentTime())
+		existingChat.Joined = int64(message.Clock)
```
Parveshdhull, line 426: chat.Joined ?

Parveshdhull (changes requested):

protocol/messenger_handler.go
rasom: what do you mean?

Parveshdhull: Oh sorry, I should have included more details.

We are modifying `existingChat.Joined` but we are not using it 🤔, should we also change this into `chat.Joined`? then we can just also remove if else check.

Parveshdhull (changes requested):

protocol/messenger_handler.go
Parveshdhull: Oh sorry, I should have included more details.

We are modifying `existingChat.Joined` but we are not using it 🤔, should we change this into `chat.Joined` ? then, we can also remove if else check

rasom: both are pointers to the same value, so I doubt there is any difference

Parveshdhull (approved):

PR #2469, opened 2021-12-21
feat: LoginWithConfig

richard-ramos:
Adds a new funtion to login with an account passing a node configuration that will override any specified attribute from the stored account nodeconfig


vendor/modules.txt
```diff
@@ -184,6 +184,8 @@ github.com/huin/goupnp/httpu
 github.com/huin/goupnp/scpd
 github.com/huin/goupnp/soap
 github.com/huin/goupnp/ssdp
+# github.com/imdario/mergo v0.3.12
+github.com/imdario/mergo
```
alaibe, line 188: is it normal to have it commented and not commented?

alaibe (approved):

vendor/modules.txt
richard-ramos: No idea :thinking:

PR #2470, opened 2021-12-21
feat: extract node config from settings to individual tables

richard-ramos:
In this PR instead of storing the `node_config` in a column on the `settings` table, it's stored in a set of different tables. This means that any new node settings that we create could be manipulated easily without having to worry about having the node configuration stored as a blob.

For example, imagine we create a new setting like this:
```js
// NodeConfig json
{
...
   MyCustomConfig {
       enabled: true,
       someValue: "1234"
   }
...
}
```
Instead of having to modify the serialized node config on the settings table, we can create a table and add a row with the configuration:
```sql
CREATE TABLE my_custom_config (
    enabled BOOLEAN DEFAULT false,
    someValue VARCHAR NOT NULL DEFAULT "",
    synthetic_id VARCHAR DEFAULT 'id' PRIMARY KEY
) WITHOUT ROWID;

INSERT INTO my_custom_config VALUES(true, "1234", "id");
```

And then modify the function `func (db *Database) loadNodeConfig(tx *sql.Tx) (*params.NodeConfig, error) ` so the `params.NodeConfig` object is populated with the values from this new table.

It includes also logic to migrate the node configuration of existing accounts to these new tables. This operation is done when executing `func (db *Database) GetNodeConfig() (*params.NodeConfig, error)`.

For this, the migrations have been separated into 2 folders:
1 - migrationsprevnodecfg which contains all the .sql files up to the moment when the node config was still embedded as a blob inside the settings table
2 - migrations which should contain .sql files that assume that the node config has been extracted into its own set of tables

In addition to that, structs like NodeConfig and Network have been moved to different packages to avoid import cycles


multiaccounts/accounts/node_config.go
```diff
@@ -0,0 +1,711 @@
...
+const WakuRendezvousNodes = "waku_rendezvous"
+const DiscV5BootstrapNodes = "discV5boot"
+
+func (db *Database) nodeConfigWasMigrated(tx *sql.Tx) (migrated bool, err error) {
+	row := tx.QueryRow("SELECT exists(SELECT 1 FROM node_config)")
+	switch err := row.Scan(&migrated); err {
+	case sql.ErrNoRows:
+		return migrated, nil
+	case nil:
+		return migrated, nil
```
Samyoul: It is a matter of preference, but you can group cases in a switch case. If you prefer the original way that's completely cool, but I just thought I'd highlight this other way.

```suggestion
	case sql.ErrNoRows, nil:
		return migrated, nil
```

multiaccounts/accounts/node_config.go
```diff
@@ -0,0 +1,711 @@
...
+			return nil, err
+		}
+		nodecfg.WakuConfig.SoftBlacklistedPeerIDs = append(nodecfg.WakuConfig.SoftBlacklistedPeerIDs, peerID)
+	}
+
+	return nodecfg, nil
+}
+
+func (db *Database) GetNodeConfig() (*params.NodeConfig, error) {
+	tx, err := db.db.BeginTx(context.Background(), &sql.TxOptions{})
```
Samyoul: When does this `tx` call `.Commit()`?

Samyoul (approved):
Looks really good, just a couple of little pointers / questions

multiaccounts/accounts/node_config.go
richard-ramos: good catch!

PR #2471, opened 2021-12-22
feat: ens API use hex return

alaibe:
Update return from bigint to hex for better compatibility with nim


richard-ramos (approved):

jrainville (commented):
Would this be a breaking change for mobile?

alaibe:
@jrainville this is a new API I pushed yesterday, mobile do not use it yet

jrainville (approved):
Approved then!

PR #2474, opened 2021-12-23
chore: execute nodecfg migration as part of db migrations

richard-ramos:
In this PR, the nodeconfig extraction is executed as part of the migrations workflow. 

For this, the migrations have been separated into 2 folders:
1 - `migrationsprevnodecfg` which contains all the .sql files up to the moment when the node config was still embedded as a blob inside the `settings` table
2 - `migrations` which should contain .sql files that assume that the node config has been extracted into its own set of tables

In addition to that, structs like `NodeConfig` and `Network` have been moved to different packages to avoid import cycles


alaibe (approved):
Code looks good to me. One question regarding the prevnodeconfig. What stop us to only have one previous.sql in the same directory as all the future migration?
I don t think we would ever edit this file.

Samyoul:
Why does nodeconfig extraction migrations require separating the migrations into "before" and "after" directories? Is there something special about these migrations? 🙂

richard-ramos:
> Why does nodeconfig extraction migrations require separating the migrations into "before" and "after" directories? Is there something special about these migrations? slightly_smiling_face

In PR https://github.com/status-im/status-go/pull/2470 I extracted the node configuration to separate tables. This is an operation that is executed when GetNodeConfig() is called, which verifies if the node configuration was moved from `settings` table to `node_config` tables (and it happens after `InitializeDB()` is executed). However, if we were to create a .sql migration to manipulate the existing node configuration, we can't safely assume that the node config information was moved to the new tables, so it's likely that the .sql migration would fail silently (since any UPDATE or DELETE would affect 0 rows)

With this PR I fix this issue because I force the node config migration to execute as part of the InitializeDB() function, meaning that any new .sql migrations can expect that the node config tables will have data, meaning it's possible to execute any sql statement to manipulate these rows

richard-ramos:
Also, unfortunately the migration dependency we are using does not allow executing code migrations, only .sql, and after looking at its implementation, it seems to me that modifying it to support this functionality is more complex than just having a separate prev-nodeconfig migration folder, and a post-nodeconfig migration folder.

I do think that if in the future we decide to continue extracting blobs from the settings table to separate tables, it's probably a good idea to extend the functionality of the migrations dependency

Samyoul:
Ah I see now, thank you. This implementation will also unfortunately introduce a breaking change in the application, by moving the migration directory the migration library will create a new migrations table and re-execute all the migrations again, basically wiping the db.

If you inspect a db after the appdatabase migrations and the protocol migrations have been ran you can see 2 migrations tables, these tables are defined by the directory that the migrations are in.

---

All this said, I 100% agree with you and support moving away from JSON blobs in the db, it is inefficient and a incongruent with the rest of the application's data storage patterns.

richard-ramos:
> Ah I see now, thank you. This implementation will also unfortunately introduce a breaking change in the application, by moving the migration directory the migration library will create a new migrations table and re-execute all the migrations again, basically wiping the db.

Curiously enough, I see that the table `status_go_schema_migrations` is used by 2 separate migrations: `appdatabase/migrations` and `multiaccounts/migrations`, without observing the data being wiped out. Could you help me confirm if this behaviour you are describing still happens?

Samyoul:
@richard-ramos it doesn't happen, I am mistaken about this issue. The migration table is explicitly defined in the code, for some reason I thought that the migration table was generated from the migrations directory.

Samyoul (approved):

PR #2475, opened 2021-12-31
Waku v2 rendezvous for mobile

vitvly:
This adds go-waku.test fleet and modifies fleet update script slightly to account for different naming (`libp2p` and `waku`). Expected to drop script modification once fleet naming is consistent.


jakubgs (approved):
Looks good, but I agree that it would make more sense to refresh the naming of services and be consistent across fleets.

churik:
@siphiuel still relevant?

vitvly:
discv5 is now used instead of rendezvous

PR #2478, opened 2022-01-05
Sync Settings

Samyoul:
Adds functionality to synchronise settings between paired devices.

Also implements a refactor of settings handling to a setting register to allow more complex behaviour to be attached to specific settings.

---

**TODO :**

- [ ] [Remove node config from the settings flow / API](https://github.com/status-im/status-go/pull/2478#discussion_r793581556)


Parveshdhull (approved):
Awesome Implementation 👏

appdatabase/database.go
```diff
@@ -71,3 +72,30 @@ func EncryptDatabase(oldPath, newPath, password string) error {
...
+			return "", err
+		}
+
+		// The "main" database is the one we care about
+		if category == "main" {
+			return filename, nil
+		}
+	}
+	if err := rows.Err(); err != nil {
+		return "", nil
```
richard-ramos: Should the err be returned here?
```suggestion
		return "", err
```

multiaccounts/settings/database.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+
+	if setting.ValueHandler() != nil {
+		value, err = setting.ValueHandler()(value)
+		if err != nil {
+			return err
+		}
+	}
+
+	// TODO(samyoul) this is ugly as hell need a more elegant solution
```
richard-ramos, line 190: I agree with this comment. There's a "settings_saveNodeConfig" RPC method. Maybe we could just remove the node config as a valid setting, and update mobile and desktop to use that rpc method instead of settings_saveSetting for updating the node config?  

Doing this change in desktop shouldnt take long, I'd say that it should be the same case in mobile.

multiaccounts/settings/database.go
```diff
@@ -0,0 +1,299 @@
...
+
+	err := db.db.QueryRow(query).Scan(&result)
+	if err != nil {
+		return 0, err
+	}
+
+	return result, nil
+}
+
+func (db *Database) SetSettingLastSynced(column string, clock uint64) error {
```
richard-ramos: What do you think of changing this function so it's not exported, to avoid the possibility of it being called by another package, and having the package needing to know a DB column name (which IMO it's an implementation detail that should only be known by this package).

```suggestion
func (db *Database) setSettingLastSynced(column string, clock uint64) error {
```

If you still prefer to have this function exported, maybe It can be changed so it receives a SettingField instead of the column name? and the the column value could be extracted with `theSetting.GetDbName()`

richard-ramos (approved):
Very cool PR!

appdatabase/database.go
Samyoul: 🤦, thank you nice catch

multiaccounts/settings/database.go
Samyoul: Ah, you've pin pointed my problem. Node config now operates outside of the normal settings flow, because it isn't classed as a setting anymore.

Yes, I agree with your suggestion, I'll make the change on mobile, and removed node config from the settings API.

multiaccounts/settings/database.go
Samyoul: I agree with this also. I think I'll implement both your suggestions because I'd like to make it hard to mess with the sync clocks be accident.

Samyoul:
Experienced flaky tests:

[#2590 - TestEncryptionServiceTestSuite/TestMaxKeep](https://github.com/status-im/status-go/issues/2590)
[#2596 - TestMessengerPushNotificationSuite/TestReceivePushNotificationFromContactOnly](https://github.com/status-im/status-go/issues/2596)
[#2597 - TestTransactionValidatorSuite/TestValidateTransactions](https://github.com/status-im/status-go/issues/2597)

richard-ramos (approved):

PR #2481, opened 2022-01-05
feat: desktop mailserver cycle

richard-ramos:
This PR moves the logic for the mailserver cycle from desktop to status-go. Future iterations might even include an automatic call to request the mailserver history, but in the meanwhile this should be handled on Desktop.

The following API methods are added:
- "wakuext_startMailserverCycle". Should be called on login
- "wakuext_disconnectActiveMailserver". Should be called when "wakuext_requestAllHistoricMessages" returns an error. In the future this API method will be removed, once status-go handles the history requests by itself.

Two new signals are added:
- "mailserver.available": triggered as soon as a mailserver becomes available. This signal should be used to know when to request the mailserver history.
- "mailserver.changed": triggered when the mailserver is switched (this happens when the mailserver cycle is executed, and the current mailserver is disconnected). This signal is useful to display the current mailserver.

**Note:** There is a some duplication of code because we have to handle both WakuV1 and WakuV2, but this is something that will be solved by itself in medium term once we migrate to wakuv2 and remove wakuv1 code.
In a separate PR the code for handling pinned mailservers will be added.

### DIFF
--- a/services/ext/api.go
+++ b/services/ext/api.go
@@ -909,6 +909,14 @@ func (api *PublicAPI) RequestAllHistoricMessages() (*protocol.MessengerResponse,
 	return api.service.messenger.RequestAllHistoricMessages()
 }
 
+func (api *PublicAPI) StartMailserverCycle() error {
+	return api.service.messenger.StartMailserverCycle()
+}
+
+func (api *PublicAPI) DisconnectActiveMailserver() {
+	api.service.messenger.DisconnectActiveMailserver()
+}
+
 // Echo is a method for testing purposes.
 func (api *PublicAPI) Echo(ctx context.Context, message string) (string, error) {
 	return message, nil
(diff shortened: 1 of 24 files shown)


services/ext/api.go
```diff
@@ -909,6 +909,14 @@ func (api *PublicAPI) RequestAllHistoricMessages() (*protocol.MessengerResponse,
 	return api.service.messenger.RequestAllHistoricMessages()
 }
 
+func (api *PublicAPI) StartMailserverCycle() error {
```
alaibe: Can this be started when we start the messenger?

alaibe (commented):
Code LGTM, I just have a question regarding the API. Does the client needs to know about the mailserver at all!. Can this not be started when starting the messenger?

alaibe (approved):

PR #2482, opened 2022-01-06
feat: Add token store API

alaibe:
Add the token store to the wallet API such that the list of token are shared between all clients


Khushboo-dev-cpp (approved):
Still new to this but looks good to me :)

richard-ramos (approved):
:100:

flexsurfer:
@alaibe @richard-ramos did you consider having a file that status-go can read? to do not build status-go each time we add new token

PR #2485, opened 2022-01-10
fix: Send transaction is not handled correctly during a request transaction flow

Khushboo-dev-cpp:
fix: Send transaction is not handled correctly during a request transaction flow

Added flow to update message when send transaction event is received.

### DIFF
--- a/protocol/messenger_handler.go
+++ b/protocol/messenger_handler.go
@@ -21,6 +21,7 @@ import (
 )
 
 const (
+	sendTransactionMessage                      = "Transaction sent"
 	transactionRequestDeclinedMessage           = "Transaction request declined"
 	requestAddressForTransactionAcceptedMessage = "Request address for transaction accepted"
 	requestAddressForTransactionDeclinedMessage = "Request address for transaction declined"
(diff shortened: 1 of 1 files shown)


protocol/messenger_handler.go
```diff
@@ -1071,19 +1072,42 @@ func (m *Messenger) HandleSendTransaction(messageState *ReceivedMessageState, co
...
-		From:            messageState.CurrentMessageState.PublicKey,
-		RetryCount:      0,
+
+	oldMessage, err := m.persistence.MessageByID(command.Id)
+	if err != nil {
+		return err
+	}
+	if oldMessage == nil {
+		return errors.New("message not found")
+	}
```
alaibe, line 1082: I think it is one or the other. If you have err != nil you oldMessage will not be nil

protocol/messenger_handler.go
```diff
@@ -1071,19 +1072,42 @@ func (m *Messenger) HandleSendTransaction(messageState *ReceivedMessageState, co
...
+	// Hide previous message
+	err = m.persistence.HideMessage(command.Id)
+	if err != nil {
+		return err
 	}
-	m.logger.Info("Saving transction to validate", zap.Any("transaction", transactionToValidate))
+	oldMessage.Replace = command.Id
 
-	return m.persistence.SaveTransactionToValidate(transactionToValidate)
+	return m.handleCommandMessage(messageState, oldMessage)
```
alaibe, line 1110: so you called a function that will late save the tx?

Khushboo-dev-cpp: it will update the message with new status

protocol/messenger_handler.go
Khushboo-dev-cpp: Actually followed what's done in HandleDeclineRequestTransaction. not sure if there is a chance where these two can be both hit?

alaibe: Going deep down into the finder:
```
err = db.tableUserMessagesScanAllFields(row, &message)
	switch err {
	case sql.ErrNoRows:
		return nil, common.ErrRecordNotFound
	case nil:
		return &message, nil
	default:
		return nil, err
	}
```

I think we can assume that.

protocol/messenger_handler.go
alaibe: Out of curiosity, is there any other usage of SaveTransactionToValidate, if none, we could remove this function too?

alaibe: Also I look in the transaction validator, should we not send the event from there?

cammellos:
@Khushboo-dev-cpp thanks for the fix!
Could you please add a test that would fail with the old behavior and is fixed by the changes introduced please?
Also, could you briefly explain what the issue was?
Thanks!

protocol/messenger_handler.go
Khushboo-dev-cpp: thanks for checking. will do  :)

protocol/messenger_handler.go
Khushboo-dev-cpp: > Out of curiosity, is there any other usage of SaveTransactionToValidate, if none, we could remove this function too?

Dont see any other usage. Will check and remove

Khushboo-dev-cpp:
> @Khushboo-dev-cpp thanks for the fix!
> Could you please add a test that would fail with the old behavior and is fixed by the changes introduced please?
> Also, could you briefly explain what the issue was?
> Thanks!

@cammellos Sure let me try and write a test or updated the existing TestRequestTransaction for this.

To briefly explain what the issue is -
When a user request transaction -> the state is not updated in case the transaction is sent by the user.
In case of a decline of the "Request Transaction" the message list is updated with the correct data.
What I found is that in case of a response "Transaction Sent" the function was updating the message list with this update, so I updated that part.

Do let me know if you think there could be a better way to proceed with this, @alaibe also pointed out "transaction_validator" would that be a better place for this. I am very new to this part of the code so I am be completely off track

Khushboo-dev-cpp:
This is a non issue the test was performed under ropsten and reason the status is not updated is that -> we set the VerifyTransactionURL  to mainnet 
https://github.com/status-im/status-go/blob/f1569e4bde50e993611288a59b7a415a010cbfa8/params/config.go#L614
This is  because -> the URL for verifying transaction is a separate piece of config, we always point it to mainnet, and we don't want to change it if the user changes network, as otherwise someone could trick them and send a ropsten transaction instead of a mainnet
So if any one wants to test for development please set
"VerifyTransactionChainID": 3 &
 "VerifyTransactionURL": "https://ropset.infura.io/v3/" & INFURA_TOKEN_RESOLVED here
 https://github.com/status-im/status-desktop/blob/7f058f3141dee2e0b2a66037501475c3b1825968/src/app_service/common/account_constants.nim#L156 
 and everything should work as expected. 
 
 Closing this PR as it not required

PR #2486, opened 2022-01-12
fix: mailserver response time

richard-ramos:
RTTMs were being set nil when the response time was less than 1ms


jrainville (approved):
It would crash for me without this change!

PR #2492, opened 2022-01-18
feat: move create accounts to status go

alaibe:
Move all account generation/addition to status-go

### DIFF
--- a/multiaccounts/accounts/database.go
+++ b/multiaccounts/accounts/database.go
@@ -765,6 +765,12 @@ func (db *Database) GetChatAddress() (rst types.Address, err error) {
 	return
 }
 
+func (db *Database) GetLatestDerivedPath() (uint, error) {
+	var result uint
+	err := db.db.QueryRow("SELECT latest_derived_path FROM settings WHERE synthetic_id = 'id'").Scan(&result)
+	return result, err
+}
+
 func (db *Database) GetAddresses() (rst []types.Address, err error) {
 	rows, err := db.db.Query("SELECT address FROM accounts ORDER BY created_at")
 	if err != nil {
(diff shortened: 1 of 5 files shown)


multiaccounts/accounts/database.go
```diff
@@ -492,6 +492,10 @@ func (db *Database) SaveSetting(setting string, value interface{}) error {
 	return err
 }
 
+func (db *Database) GetNodeConfig(nodecfg interface{}) error {
```
richard-ramos: This column is not used anymore. There's a GetNodeConfig() available in nodecfg package.

services/accounts/accounts.go
```diff
@@ -37,3 +48,131 @@ func (api *API) GetAccounts(ctx context.Context) ([]accounts.Account, error) {
...
+	ctx context.Context,
+	mnemonic string,
+	address string,
+	password string,
+	name string,
+	color string,
+) error {
+	mnemonicNoExtraSpaces := strings.Join(strings.Fields(mnemonic), " ")
+
+	_, err := api.manager.VerifyAccountPassword(api.config.KeyStoreDir, address, password)
```
richard-ramos: I think we should be able to avoid having to pass `address` as a parameter. I think `wallet_root_address` in `settings` table contains it.  (same comment applies to `AddAccountWithPrivateKey`, and `GenerateAccount`)

richard-ramos (changes requested):

multiaccounts/accounts/database.go
alaibe: Oops bad rebase!!

richard-ramos (approved):

PR #2499, opened 2022-01-23
read-indicator

jo-mut:
Functionality to process read receipts. 

This sends a message to other user in a one to one chat letting the user know that their message has been seen


appdatabase/migrationsprevnodecfg/sql/1647852385_add_read_receipts_enabled_to_settings.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE settings ADD COLUMN read_receipts_enabled BOOLEAN DEFAULT false;
```
richard-ramos: This migration file should be placed inside  appdatabase/migrations/sql/  instead of this folder  (migrationsprevnodecfg should not be modified since these are the sql files that were run previous to migrating `node_config` column out of the settings table)

richard-ramos:
```suggestion
ALTER TABLE settings ADD COLUMN read_receipts_enabled BOOLEAN DEFAULT false;
UPDATE settings SET read_receipts_enabled = 0;
```

richard-ramos (changes requested):

appdatabase/migrationsprevnodecfg/sql/1647852385_add_read_receipts_enabled_to_settings.up.sql
jo-mut: could be why i have not been able to make this work. Let me do that. Thanks

richard-ramos: This PR might help with an issue that happens when you logout and try to login again: https://github.com/status-im/status-go/pull/2504

Samyoul: For future reference there is a make command to help with making new migrations:

`make migration D=add_column_to_settings `

which will create the following file:
`appdatabase/migrations/sql/1642593284_add_column_to_settings.up.sql`

PR #2500, opened 2022-01-24
feat: introduce communities settings

### LINKED ISSUE #2495: Introduce flag to enable/disable message history archive support in communities
This is a sub task of the Community History Archives MVP.

**Draft spec**: https://github.com/status-im/specs/pull/164
**Roadmap / Tasks**: https://hackmd.io/@status-desktop/B1naRjxh_/%2FNeX8-q49TtqN18i4VkXRFw

---

There are three different cases that need to be covered:

1. We need a general flag in status-go to turn community history archive support on/off. This is needed so mobile clients won't participate in serving community history archives.
2. Community owners are expected to always provide community history archives, however this should still be possible to turn this off. So we need to add some sort of flag in the user settings
3. Community member nodes may or may not participate in serving community history archives. They can either just request archives, or also seed archives. Ideally they all seed, but still, this needs to be configurable. This can be either done using the same flag mentioned in 2), or for more control, we could introduce an additional one

0x-r4bbit:
This commit introduces new communities settings that are needed for the
upcoming Community History Archive support.

Namely, we need to persist whether community owners will seed message
history archives and whether community members are interested in
fetching such archives.

The settings are defined as follows:

1. `community_id` - The id of the community the settings belong to
2. `message_archive_seeding_enabled` - Whether message history archives
   will be created and seeded for this community
3. `message_archive_fetching_enabled` - Whether message history archives
will fetched for this community

The data is persisted in a newly introduced `communities_settings`
table.

Furthermore, the RPC API that returns `Settings` for the logged in
account will now include a `communities-settings` field of type:
`[]CommunitySettings`.

The community settings in the database will be added, updated an deleted
depending whether the user joins, leaves or edits the community in
question.

### Tasks

- [x] Write tests
- [x] Correct formatting
- [ ] Migration script to add community settings for existing communities

Closes #2495


multiaccounts/accounts/database.go
```diff
@@ -559,6 +560,9 @@ func (db *Database) GetSettings() (Settings, error) {
 		&s.AutoMessageEnabled,
 	)
 
+  communitiesSettings, err := db.GetCommunitiesSettings()
+  s.CommunitiesSettings = communitiesSettings
```
0x-r4bbit: Since `communities_settings` live in a separate table, I'm explicitly querying/assigning it here.

Let me know if this should not happen here and whether we prefer a separate RPC API so clients can request that data.

params/config.go
```diff
@@ -504,6 +504,12 @@ type Network struct {
 	Enabled                bool   `json:"enabled"`
 }
 
+type CommunitySettings struct {
+  CommunityID                     string          `json:"communityId"`
+  MessageArchiveFetchingEnabled   bool            `json:"messageArchiveFetchingEnabled"`
+  MessageArchiveSeedingEnabled    bool            `json:"messageArchiveSeedingEnabled"`
+}
```
0x-r4bbit, line 511: This might not be the proper place to put this type. 

Eventually, we probably need a type for enabling/disabling community history archive support in the CLI as well. Might as well leverage this type for that, which is why I put it here.

Happy to put it somewhere else.

protocol/messenger_communities.go
```diff
@@ -137,6 +138,17 @@ func (m *Messenger) JoinCommunity(ctx context.Context, communityID types.HexByte
 		return nil, err
 	}
 
+  communitySettings := params.CommunitySettings{
+    CommunityID: communityID.String(),
+    MessageArchiveFetchingEnabled: true,
+    MessageArchiveSeedingEnabled: false,
+  }
```
0x-r4bbit: These are the defaults when joining a community.

If could also introduce a `GetDefaultCommunitySettings` API if you prefer.

protocol/requests/create_community_request.go
```diff
@@ -17,17 +18,19 @@ var (
...
+	Emoji                         string                               `json:"emoji"`
+	Membership                    protobuf.CommunityPermissions_Access `json:"membership"`
+	EnsOnly                       bool                                 `json:"ensOnly"`
+	Image                         string                               `json:"image"`
+	ImageAx                       int                                  `json:"imageAx"`
+	ImageAy                       int                                  `json:"imageAy"`
+	ImageBx                       int                                  `json:"imageBx"`
+	ImageBy                       int                                  `json:"imageBy"`
+  MessageArchiveSeedingEnabled  bool                                 `json:"messageArchiveSeedingEnabled,omitempty"`
+  MessageArchiveFetchingEnabled  bool                                 `json:"messageArchiveFetchingEnabled,omitempty"`
```
0x-r4bbit: I made those two `omitempty` so we can potentially put these changes into production without having the need for desktop and/or mobile to do anything in particular.

0x-r4bbit:
>  Have you updated the documentation, if impacted (e.g. docs.status.im)?

There's a pending PR for a protocol spec addition that eventually needs these changes: https://github.com/status-im/specs/pull/164

0x-r4bbit:
The linting error caused in this PR: https://ci.status.im/blue/organizations/jenkins/status-go%2Fprs%2Ftests/detail/PR-2500/4/pipeline#step-60-log-3 seems to be unrelated to my changes.

0x-r4bbit:
As discussed offline with @cammellos, introduction of `communities_settings` for existing communities (prior to this PR/commit) can possibly be done in an SQL migration script.

I'll give that a spin, so please don't merge this yet.

protocol/requests/create_community_request.go
```diff
@@ -89,3 +92,14 @@ func (c *CreateCommunity) ToCommunityDescription() (*protobuf.CommunityDescripti
 	}
 	return description, nil
 }
+
+func (c *CreateCommunity) ToCommunitySettings(publicKey string) (params.CommunitySettings, error) {
```
richard-ramos, line 96: Is an error expected? if not, it can be removed from the return parameters

multiaccounts/accounts/database.go
```diff
@@ -864,3 +868,55 @@ func (db *Database) ENSName() (string, error) {
 	}
 	return "", err
 }
+
+func (db *Database) SaveCommunitySettings(communitySettings params.CommunitySettings) error {
```
richard-ramos, line 872: In protocol/communities/persistence.go live all the functions related to communities CRUD operations. What do you think of having these community settings functions live there too?

appdatabase/migrations/sql/1642771235_add_communities_settings.up.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE communities_settings (
+  community_id TEXT PRIMARY KEY ON CONFLICT REPLACE,
```
richard-ramos, line 2: In protocol/migrations/sqlite/1605075346_add_communities.up.sql  the community_id is defined as 
`  id BLOB NOT NULL PRIMARY KEY ON CONFLICT REPLACE`.  Consider using the same data type (blob)

multiaccounts/accounts/database.go
```diff
@@ -864,3 +868,55 @@ func (db *Database) ENSName() (string, error) {
...
+    message_archive_fetching_enabled
+  ) VALUES (?, ?, ?)`,
+		communitySettings.CommunityID,
+		communitySettings.MessageArchiveSeedingEnabled,
+		communitySettings.MessageArchiveFetchingEnabled,
+	)
+	if err != nil {
+		return err
+	}
+	return nil
```
richard-ramos, line 885:
```suggestion
	return err
```

multiaccounts/accounts/database.go
```diff
@@ -559,6 +560,9 @@ func (db *Database) GetSettings() (Settings, error) {
 		&s.AutoMessageEnabled,
 	)
 
+	communitiesSettings, err := db.GetCommunitiesSettings()
+	s.CommunitiesSettings = communitiesSettings
```
richard-ramos, line 564: If there's an `err`, it might end up being overwritten by a succesful `db.getCommunitiesSetting()` response

```suggestion
        if err != nil {
            return s, err
        }

	communitiesSettings, err := db.GetCommunitiesSettings()
	s.CommunitiesSettings = communitiesSettings
```

appdatabase/migrations/sql/1642771235_add_communities_settings.up.sql
0x-r4bbit: Good call!

multiaccounts/accounts/database.go
0x-r4bbit: Good catch!!

multiaccounts/accounts/database.go
0x-r4bbit: Happy to move it over!

Main reason I put them here was because I consider them more "settings"  related to the account. Like.. we could also call the entire thing "node_settings" if we want (once we get to actually supporting all the settings like throttling etc). In that sense it wouldn't necessarily be community specific anymore.

But yea, I totally see where you're coming from, so I'll move it over for now. We can still migrate it to somewhere else once this stuff gets broader responsibility.

protocol/requests/create_community_request.go
0x-r4bbit: It's not, wanted to stay consistent. But happy to change this!

0x-r4bbit:
Closing in favour of #2574

PR #2503, opened 2022-01-26
enable ens service by default

flexsurfer:
enable ens service by default


richard-ramos (commented):
I think we should also remove the `ENSConfig` attribute from the config, since it will not do anything.

flexsurfer:
thanks @richard-ramos , done

richard-ramos (approved):

alaibe (approved):

PR #2505, opened 2022-01-26
feat: Move Tenor Api to status_go

Khushboo-dev-cpp:

Move tenor API calls to status_go

Important changes:


appdatabase/migrations/sql/1901001000_gif_api_key_setting.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE settings ADD COLUMN gif_api_key TEXT NOT NULL DEFAULT "";
```
richard-ramos, line 1:
```suggestion
ALTER TABLE settings ADD COLUMN gif_api_key TEXT NOT NULL DEFAULT "";
UPDATE settings SET gif_api_key = "";
```

services/gif/gif.go
```diff
@@ -0,0 +1,67 @@
...
+    var defaultParams = fmt.Sprintf("&media_filter=minimal&limit=50&key=%s",tenorApiKey)
+    tenorApiKey, error := api.GetTenorAPIKey()
+    if error!=nil {
+      return "", fmt.Errorf("Error in reading TenorAPIKey: %v", error)
+    }
+
+    for currentRetry < MAX_RETRY {
+    response, error = api.httpClient.Get(baseUrl+path+defaultParams)
+    if error != nil {
+    fmt.Printf("can't get content from path %s", path)
```
richard-ramos: Probably should be replaced by a log instead of Printf, also, the error variable should be added to this log
Also please rename `error` to `err`. (To reuse the `err` return parameter)

richard-ramos (changes requested):

appdatabase/migrations/sql/1901001000_gif_api_key_setting.up.sql
Khushboo-dev-cpp: Done :)

services/gif/gif.go
Khushboo-dev-cpp: Done :)

services/gif/gif.go
```diff
@@ -0,0 +1,71 @@
...
+func (api *API) GetTenorAPIKey() (value string, err error) {
+	log.Info("[GifAPI::getTenorAPIKey]")
+	apiKey, err := api.db.GifAPIKey()
+	if err != nil {
+		return "", err
+	}
+	return apiKey, err
+}
+
+func (api *API) GetContentWithRetry(path string) (value string, err error) {
```
alaibe, line 53: That looks more like an internal API that should not be exposed. Could we have more dedicated endpoint?

Khushboo-dev-cpp: hmmm would something like getContent be better?

Khushboo-dev-cpp: Done @alaibe have also moved the recents and favourites here

richard-ramos (approved):

services/gif/gif.go
```diff
@@ -0,0 +1,108 @@
...
+
+	return string(data), nil
+}
+
+func (api *API) FetchGifs(path string) (value string, err error) {
+	log.Info("[GifAPI::fetchGifs]")
+	return api.GetContentWithRetry(path)
+}
+
+func (api *API) UpdateRecentGifs(updatedGifs json.RawMessage) (err error) {
```
alaibe, line 87: rather than RawMessage, could we have typed struct here?

alaibe (approved):
One more thing, do you think you could add test

services/gif/gif.go
Khushboo-dev-cpp: Done and for unit tests I have created a different task so that this lands in the meantime :)
https://github.com/status-im/status-go/issues/2508

Samyoul:
This migration `1901001000` will cause a problem it is dated way into the future. Other migrations will not trigger if this migration is present

PR #2512, opened 2022-01-31
Dont overwrite delivered status

cammellos:
We should not override a "delivered" status on re-transmission.


Samyoul (approved):

rasom (approved):

protocol/messenger.go
```diff
@@ -471,7 +471,7 @@ func (m *Messenger) processSentMessages(ids []string) error {
 			return errors.Wrapf(err, "Can't save raw message marked as sent")
 		}
 
-		err = m.UpdateMessageOutgoingStatus(id, "sent")
+		err = m.UpdateMessageOutgoingStatus(id, common.OutgoingStatusSent)
```
Samyoul, line 488: Forgot to say, die string literals die!

services/mailservers/fleet.go
```diff
@@ -0,0 +1,133 @@
+package mailservers
+
+func DefaultMailservers() []Mailserver {
+
+	return []Mailserver{
+		Mailserver{
+			ID:      "mail-01.ac-cn-hongkong-c.eth.prod",
+			Address: "enode://606ae04a71e5db868a722c77a21c8244ae38f1bd6e81687cc6cfe88a3063fa1c245692232f64f45bd5408fed5133eab8ed78049332b04f9c110eac7f71c1b429@47.75.247.214:443",
+			Fleet:   "eth.prod",
```
richard-ramos: Is it possible to use the consts from params/cluster.go instead of strings:
```
const (
	FleetUndefined  = ""
	FleetProd       = "eth.prod"
	FleetStaging    = "eth.staging"
	FleetTest       = "eth.test"
	FleetWakuV2Prod = "wakuv2.prod"
	FleetWakuV2Test = "wakuv2.test"
	FleetStatusTest = "status.test"
)
```

protocol/messenger_mailserver.go
```diff
@@ -669,6 +722,31 @@ func (m *Messenger) LoadFilters(filters []*transport.Filter) ([]*transport.Filte
...
+func (m *Messenger) ToggleUseMailservers(value bool) error {
+	err := m.settings.SetUseMailservers(value)
+	if err != nil {
+		return err
+	}
+
+	if value {
+		m.cycleMailservers()
+		return nil
+	}
```
richard-ramos, line 780: In a future PR, we could call instead `m.StartMailserverCycle()` and also implement a `m.StopMailserverCycle()` which should stop these goroutines:
```
go m.checkMailserverConnection()
go m.updateWakuV1PeerStatus()
go m.updateWakuV2PeerStatus()
```
That way we ensure that the mailserver cycle is not really running.

richard-ramos (changes requested):
`EnableMailserverCycle` should be removed from params/config.go and nodecfg/node_config.go, and a migration should be created to drop `enable_mailserver_cycle` from `shhext_config` table

richard-ramos (approved):

PR #2510, opened 2022-01-28
[#12783] Improve order of mailserver requests

rasom:
needed for status-im/status-react#12783

Before syncing, topics are sorted by priority. In case if all topics have zero priority requests are executed as before. Otherwise, topics with higher priority are requested in first N batches (currently three batches with 1, 5, 10 topics. That's arbitrary, emphasis is on the first batch), and the rest of topics are requested as before.

`chat.ReadMessagesAtClockValue` is used to prioritise topics before historical requests as it is always updated when the user opens chat (at least in mobile version). We can add a separate field for this but I'm not sure this is necessary.
 
status: ready

PR #2517, opened 2022-02-02
[#12783] fix index out of range when there are no filters

rasom:
fix index out of range issue mentioned here 
https://github.com/status-im/status-go/pull/2510#issuecomment-1027912364

### DIFF
--- a/protocol/messenger_mailserver.go
+++ b/protocol/messenger_mailserver.go
@@ -288,7 +288,7 @@ func (m *Messenger) syncFilters(filters []*transport.Filter) (*MessengerResponse
 	prioritizedBatches := getPrioritizedBatches()
 	currentBatch := 0
 
-	if filters[0].Priority == 0 {
+	if len(filters) == 0 || filters[0].Priority == 0 {
 		currentBatch = len(prioritizedBatches)
 	}
 


protocol/messenger_mailserver.go
```diff
@@ -288,7 +288,7 @@ func (m *Messenger) syncFilters(filters []*transport.Filter) (*MessengerResponse
 	prioritizedBatches := getPrioritizedBatches()
 	currentBatch := 0
 
-	if filters[0].Priority == 0 {
+	if len(filters) == 0 || filters[0].Priority == 0 {
```
Khushboo-dev-cpp, line 291: wouldn't this still call filters[0].Priority and crash?

Khushboo-dev-cpp (commented):
would this still crash?

protocol/messenger_mailserver.go
rasom: I would need to read some golang specific docs to be 100% sure but usually the order of evaluation of `or` is so that the second expression is skipped when the first one is equal to `true`. And I suppose that should work in our case. 
So if  `len(filters) == 0` (which means that `filters` is empty), then the second shouldn't be executed. Otherwise we are safe to use `filters[0].Priority == 0`.

Khushboo-dev-cpp: aaah okay. I am new to golang, I think in C++ it would fail. approving it :)

Khushboo-dev-cpp (approved):
approved  :)

protocol/messenger_mailserver.go
rasom: @Khushboo-dev-cpp quick googling showed that it works exactly the same way in c++ (I mean skipping of the rest of `or` when the first expression is `true`), I bet it is going to be a challenge to find some more or less useful language which has a different evaluation order for this case. Well, maybe I'm wrong... :)

PR #2518, opened 2022-02-02
feat: stickerpacks service

richard-ramos:
The following methods will be available with this PR:

- `stickers_market(chainID)`: returns all the stickerpacks available in the smart contract, not including those already installed or pending (to avoid doing requests to infura for information that we already have)
- `stickers_installed()`: return the stickerpacks already installed
- `stickers_install(chainID, packID)`: installs a sticker pack. This does not check if you have bought the stickerpack before.
- `stickers_uninstall(packID)`: uninstalls a stickerpack
- `stickers_recent()`: returns the list of last 24 stickers used
- `stickers_addRecent({"packID": ..., "hash": .... })`: Prepends a sticker to the list of recently used stickers 
- `stickers_pending()`: returns the list of sticker packs pending transaction confirmation
- `stickers_removePending(packID)`: removes a pending stickerpack
- `stickers_buyEstimate(chainID, transactionParams, packID)`: estimate gas cost of buying a stickerpack
- `stickers_buy(chainID, transactionParams, packID, password)`: create a transaction to buy a stickepack. Adds the stickerpack automatically to the list of pending stickerpacks, and returns the transactionhash

Methods that return stickerpacks will contain objects like these:
```js
{
    id: "123", // ID of the sticker pack
    name: "Lorem Ipsum", // name of the stickerpack
    "author": "JohnDoe", // author of the stickerpack
    "owner": "0x1234....abcd", // address of the owner of the stickerpack
    "price": "10000000", // price of the stickerpack in gwei
    "preview": "https://............", //   URL of the preview of the stickerpack
    "thumbnail": "https://......", // URL of the thumbnail for the stickerpack
    "status": 1,  	// 0: available for buying, 1: installed, 2: purchased, 3: pendingConfirmation
    "stickers": [
      ...
      {
         "packID": "123", // ID of the pack for this sticker
         "url": "https://......", // URL of the sticker image
         "hash": "abcd...eef" // Hash of the sticker
      },
      ...
     ]
}
```
Notice that in the case of the stickerpack's `preview` and `thumbnail` attributes, there's no need to translate hashes to URLs as these are done automatically in status-go. The same happens for the stickers, since they now include a "url" attribute.

*WARNING*: There's also a migration in which the recent stickers are set to null since the structure for recent packs changes. This is because previously it was an array of hashes, without the packID, so if you wanted to send a recent sticker, you needed to look for which stickerpack contains a sticker with an specific hash, while with this PR, since the setting will contain an object `{"packID": ..., "hash": .... }`, it becomes easier to know which pack to use.

Buying a stickerpack has not been tested yet. Desktop has not still implemented the changes necessary to use this new service


multiaccounts/accounts/database.go
```diff
@@ -760,6 +760,11 @@ func (db *Database) GetMessagesFromContactsOnly() (bool, error) {
 	return result, err
 }
 
+func (db *Database) GetInstalledStickerPacks() (rst *json.RawMessage, err error) {
```
alaibe, line 763: can we type and avoid raw json

multiaccounts/accounts/database.go
```diff
@@ -760,6 +760,11 @@ func (db *Database) GetMessagesFromContactsOnly() (bool, error) {
...
+	err = db.db.QueryRow("SELECT stickers_packs_installed FROM settings WHERE synthetic_id = 'id'").Scan(&rst)
+	return
```
alaibe, line 765: is the return necessary?

services/stickers/contracts.go
```diff
@@ -0,0 +1,98 @@
...
+	1: common.HexToAddress("0x12824271339304d3a9f7e096e62a2a7e73b4a7e7"), // mainnet
+	3: common.HexToAddress("0x6CC7274aF9cE9572d22DFD8545Fb8c9C9Bcb48AD"), // ropsten
+}
+
+var stickerPackByChainID = map[uint64]common.Address{
+	1: common.HexToAddress("0x110101156e8F0743948B2A61aFcf3994A8Fb172e"), // mainnet
+	3: common.HexToAddress("0xf852198d0385c4b871e0b91804ecd47c6ba97351"), // ropsten
+}
+
+// TODO: refactor SNT. This is also used in ens service
```
alaibe: yes I think we need a shared place where we put all the contract addresses

services/stickers/api.go
```diff
@@ -0,0 +1,648 @@
...
+}
+
+func (api *API) Uninstall(packID uint64) error {
+	installedPacks, err := api.Installed()
+	if err != nil {
+		return err
+	}
+
+	if _, exists := installedPacks[uint(packID)]; !exists {
+		return nil // Pack is not installed
```
alaibe: shall we return an error here?

services/stickers/api.go
```diff
@@ -0,0 +1,648 @@
...
+}
+
+func (api *API) Install(chainID uint64, packID uint64) error {
+	installedPacks, err := api.Installed()
+	if err != nil {
+		return err
+	}
+
+	if _, exists := installedPacks[uint(packID)]; exists {
+		return nil // Pack is already installed
```
alaibe: same as before, maybe an error here?

multiaccounts/accounts/database.go
alaibe: nvm, I can see that you type in the api

services/stickers/api.go
```diff
@@ -0,0 +1,648 @@
...
+	callOpts := &bind.CallOpts{Context: api.ctx, Pending: false}
+
+	balance, err := stickerPack.BalanceOf(callOpts, common.Address(account))
+	if err != nil {
+		return nil, err
+	}
+
+	var tokenIDs []*big.Int
+	for i := uint64(0); i < balance.Uint64(); i++ {
+		tokenID, err := stickerPack.TokenOfOwnerByIndex(callOpts, common.Address(account), new(big.Int).SetUint64(i))
```
alaibe: Shall we move that to a coroutine?

services/stickers/api.go
```diff
@@ -0,0 +1,648 @@
...
+		if err != nil {
+			return nil, err
+		}
+		tokenIDs = append(tokenIDs, tokenID)
+	}
+
+	var purchasedPackIDs []*big.Int
+	for _, tokenID := range tokenIDs {
+		packID, err := stickerPack.TokenPackId(callOpts, tokenID)
```
alaibe: same question, move to coroutine?

multiaccounts/accounts/database.go
richard-ramos: Sadly, yes.

services/stickers/contracts.go
richard-ramos: I was thinking of extracting all the smart contracts from both ens and stickerpacks into a common folder

services/stickers/api.go
richard-ramos: Good idea!

contracts/contracts.go
```diff
@@ -0,0 +1,122 @@
+package contracts
```
alaibe, line 1: FYI there is also contracts in wallet, but we can move then later

services/stickers/api.go
```diff
@@ -0,0 +1,543 @@
...
+
+	tokenIDs, err := api.getTokenOwnerOfIndex(chainID, account, balance)
+	if err != nil {
+		return nil, err
+	}
+
+	return api.getTokenPackIDs(chainID, tokenIDs)
+}
+
+func hashToURL(hash []byte) (string, error) {
```
alaibe, line 234: We cannot reuse ens resource url function?

alaibe (approved):

PR #2523, opened 2022-02-08
DRAFT: Feat/wallet sync

Jonathansoufer:
This PR implements a sync mechanism for wallet accounts creation and update. 

This enables users to sync wallet accounts changes through paired devices.

Closes #


ghost:
Hey @Jonathansoufer, and thank you so much for making your first pull request in status-go! :heart: Please help us make your experience better by filling out this brief questionnaire https://goo.gl/forms/uWqNcVpVz7OIopXg2

ghost:
### Pull Request Checklist

multiaccounts/accounts/database.go
```diff
@@ -937,6 +940,6 @@ func (db *Database) GetAccountLastSynced(address types.Address) (uint64, error)
 func (db *Database) SetAccountLastSynced(address types.Address, clock uint64) error {
 	query := "UPDATE accounts SET clock = ? WHERE address = ? AND clock < ?"
 
-	_, err := db.db.Exec(query, address, clock)
+	_, err := db.db.Exec(query, clock, address, clock)
```
Jonathansoufer, line 943: Why passing clock twice?

protocol/communities_messenger_test.go
```diff
@@ -1188,45 +1188,65 @@ func (s *MessengerCommunitiesSuite) TestSyncCommunity() {
...
+		Address: types.Address{0x03},
+		Name: "Alice watch only 2",
+		Color: "blue",
+		Type: accounts.AccountTypeWatch,
 	}
 
-	// Create a main acount on alice
-	s.NoError(s.alice.settings.SaveAccounts(mainAccount))
+	// Create a main account on alice
+	s.NoError(s.alice.settings.SaveAccounts([]accounts.Account{mainAccount}))
```
Jonathansoufer, line 1210: is this change a code pattern related one or does this change the behaviour itself? I see you passed [] directly here instead of on the variable.

protocol/communities_messenger_test.go
```diff
@@ -1188,45 +1188,65 @@ func (s *MessengerCommunitiesSuite) TestSyncCommunity() {
...
 	s.Len(acc2, 1, "Must have 1 main account")
 
+	// Check account values match the expected values
+	s.Require().Equal(mainAccount.Address, acc2[0].Address)
+	s.Require().Equal(mainAccount.Name, acc2[0].Name)
+	s.Require().Equal(mainAccount.Color, acc2[0].Color)
+	s.Require().Equal(mainAccount.Type, acc2[0].Type)
+
 	// Pair devices
-	err = alicesOtherDevice.SetInstallationMetadata(alicesOtherDevice.installationID, &multidevice.InstallationMetadata{
```
Jonathansoufer: Same here. is it a pattern or functionality change?

multiaccounts/accounts/database.go
Samyoul: `query := "UPDATE accounts SET clock = ? WHERE address = ? AND clock < ?"`

The query has 3 positional parameters marked with `?`, we need to give a value for each of the parameters in the query.

`SET clock = ?` requires the incoming clock value
`WHERE address = ?` requires the address
`clock < ?` requires the clock again

protocol/communities_messenger_test.go
Samyoul: Yeah, this is just a code pattern change, and mainly so that we can compare the values of the accounts to the values taken from the db. Your approach was 100% valid, and using your approach you could chose to compare the values like:

```go
// Check account values match the expected values
	s.Require().Equal(mainAccount[0].Address, acc1[0].Address)
	s.Require().Equal(mainAccount[0].Name, acc1[0].Name)
	s.Require().Equal(mainAccount[0].Color, acc1[0].Color)
	s.Require().Equal(mainAccount[0].Type, acc1[0].Type)
```

It is an awkward setup because the main function `SaveAccounts` expects an array of Account, but we just want to add one at a time.

PR #2525, opened 2022-02-09
feat: unified chat API pt. 1

richard-ramos:
See https://notes.status.im/vciGwqsKTA6Fnid0lqnEnQ for a summary of why these changes are needed, as well as methods that will be created.

Some notes:
- ChatIDs for chats that belong to a community will not be prefixed with the community ID
- There is no need to merge the chats from a community with the information from "wakuext_chats" as this is done automatically in these methods
- 1:1 chats will not have a chat name. This is because this information must be obtained from the contacts

### getChats
Returns all the chats and communities the user has joined
```json
{"method":"chat_getChats","params":[]}
```

### getChat
Returns all the information for an specific chat
```json
{"method":"chat_getChat", "params":["the_community_id", "the_chat_id"]}
```
The community_id in the case of 1:1/group/public chats can either be "" or the user's public key


services/ext/chat_accesors.go
```diff
@@ -0,0 +1,328 @@
...
+	Type       ChannelGroupType                         `json:"channelGroupType"`
+	Name       string                                   `json:"name"`
+	Images     map[string]images.IdentityImage          `json:"images"`
+	Color      string                                   `json:"color"`
+	Chats      map[string]*Chat                         `json:"chats"`
+	Categories map[string]communities.CommunityCategory `json:"categories"`
+	EnsName    string                                   `json:"ensName"`
+}
+
+func (api *PublicAPI) GetChats(parent context.Context) (map[string]ChannelGroup, error) {
```
alaibe: To really differentiate with previous api. Shall we create a chat service? That will also help differentiate old vs new?

Samyoul:
@richard-ramos I really like the work you are doing on upgrading the chat APIs. The only point that I've got is the hardcoded reference to communityID via positional params.

> The community_id in the case of 1:1/group/public chats can either be "" or the user's public key

Do you think we could use named parameters? The JSON RPC spec https://www.jsonrpc.org/specification#parameter_structures allows for named parameters that can be omitted. This would allow for a more flexible and understandable API, rather than forcing the user to always give a communityID.

What do you think?

richard-ramos:
> @Samyoul: Do you think we could use named parameters? The JSON RPC spec https://www.jsonrpc.org/specification#parameter_structures allows for named parameters that can be omitted. This would allow for a more flexible and understandable API, rather than forcing the user to always give a communityID.

Great idea! I was not aware that named parameters was a thing with JSON RPC. Will do some tests to see how it works. Thank you!

services/chat/api_send_messages.go
```diff
@@ -0,0 +1,159 @@
...
+
+type SendMessageResponse struct {
+	Chat     *Chat             `json:"chat"`
+	Messages []*common.Message `json:"messages"`
+}
+
+func (api *API) SendSticker(ctx context.Context, communityID types.HexBytes, chatID string, packID int32, hash string, responseTo string) (*SendMessageResponse, error) {
+	ensName, _ := api.s.accountsDB.GetPreferredUsername()
+
+	msg := &common.Message{
```
cammellos, line 23: Probably I would move the logic in messenger, so we keep the API fairly logic less, what do you think?
It would look something like:
```
SendSticker(....) (...) {
  response, err := ap.s.messenger.SendSticker(....)
  if err ...
  }
}

File: services/chat/api_send_messages.go [hunk 3]
```diff
@@ -0,0 +1,159 @@
...
+		contentType = protobuf.ChatMessage_EMOJI
+	}
+
+	return contentType
+}
+
+func (api *API) SendMessage(ctx context.Context, communityID types.HexBytes, chatID string, text string, responseTo string) (*SendMessageResponse, error) {
+	ensName, _ := api.s.accountsDB.GetPreferredUsername()
+
+	msg := &common.Message{
```
cammellos, line 80: Similarly, I would probably move this into messenger for the most part, and just relay the arguments to messenger, but up to you

services/chat/api_send_messages.go
richard-ramos: I can do that change, I was just worried about increasing the number of functions in Messenger, but I guess I can just put all these in a new file.

iurimatias (approved):

PR #2527, opened 2022-02-10
disable changing settings while registering for push notifications

Parveshdhull:
Currently, we are also changing settings for notifications while registering/un-registering for push notifications.
But while logging out we should only un-register from notifications, settings should remain the same so that we can again register after login.

@richard-ramos  will this affect desktop?

### status-react pr
https://github.com/status-im/status-react/pull/13028


richard-ramos:
> will this affect desktop?

no, desktop does not use the push notification feature

cammellos (approved):

PR #2528, opened 2022-02-10
feat: add opensea api key

alaibe:
Add opensea api key and config

### DIFF
--- a/services/wallet/service.go
+++ b/services/wallet/service.go
@@ -13,7 +13,7 @@ import (
 )
 
 // NewService initializes service instance.
-func NewService(db *sql.DB, rpcClient *rpc.Client, accountFeed *event.Feed) *Service {
+func NewService(db *sql.DB, rpcClient *rpc.Client, accountFeed *event.Feed, openseaAPIKey string) *Service {
 	cryptoOnRampManager := NewCryptoOnRampManager(&CryptoOnRampOptions{
 		dataSourceType: DataSourceStatic,
 	})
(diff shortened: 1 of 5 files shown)


services/wallet/service.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 // NewService initializes service instance.
-func NewService(db *sql.DB, rpcClient *rpc.Client, accountFeed *event.Feed) *Service {
+func NewService(db *sql.DB, rpcClient *rpc.Client, accountFeed *event.Feed, openseaAPIKey string) *Service {
```
Samyoul, line 16: Is the key being introduced to the application at build time?

alaibe: In the case of desktop, the key is injected via jenkins at build: https://github.com/status-im/status-jenkins-lib/pull/38

Samyoul: Ok, injected into the client build. Thank you.

Samyoul (approved):

richard-ramos (approved):

services/wallet/service.go
flexsurfer: @alaibe what if we inject it directly in status-go ? overwise we have to inject it in mobile and desktop and pass it in config

alaibe: We talk about this solution with @cammellos as an intermediate solution for mobile until mobile use loginWithConfig

cammellos: Yes, we should drive this config from the client (status-react), as desktop does, but we need to swap the endpoint, but that's a risky change for us and not hotfix material. 
For now, we can fallback on build-inject time key if the config is empty and then move to use `loginWithConfig`.
Though I doubt we will hotfix this, so we might as well just use `loginWithConfig` and release in 1.19

PR #2530, opened 2022-02-10
Share image

jo-mut:
Share image message


protocol/requests/share_image_message.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/status-im/status-go/eth-node/types"
+)
+
+var ErrShareMessageInvalidID = errors.New("share image message: invalid id")
+var ErrShareMessageEmptyUsers = errors.New("share image message: empty users")
+
+type ShareImageMessage struct {
+	MessageID   string           `json:"id"`
+	Users       []types.HexBytes `json:"users"`
+	ContentType int32            `json:"contentType"`
```
cammellos: Do you need to pass `ContentType` from the client? Is it not always an image?

protocol/messenger_chats.go
```diff
@@ -575,3 +575,45 @@ func (m *Messenger) clearHistory(id string) (*MessengerResponse, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	var messages []*common.Message
+	for _, pk := range request.Users {
+		message := &common.Message{}
+		message.ChatId = pk.String()
+		message.Payload = msg.Payload
+		message.ContentType = protobuf.ChatMessage_ContentType(request.ContentType)
```
cammellos: Do we not know the content type based on the fact that you are calling "ShareImageMessage"?

cammellos (commented):
Left a few comments, it would also be good if you created tests verifying the behavior, it will help develop faster in status-go if you use the tests to verify it works. But you can send it to QA earlier in case they find some issues.

protocol/requests/share_image_message.go
jo-mut: yeah i actually don't have to

jo-mut:
I will write the test and also remove any unnecessary code given its known its an image

Samyoul (approved):
Nice PR. Apart from the items that Andrea mentions, I don't see anything that needs attention.

jo-mut:
> Nice PR. Apart from the items that Andrea mentions, I don't see anything that needs attention.

Thank you @Samyoul.

PR #2537, opened 2022-02-14
feat: add RequestCommunityInfoFromMailserverAsync

jrainville:
This function doesn't wait for the response from mailservers, so the client has to rely on signals

This was added because the basic `RequestCommunityInfoFromMailserver` could take a lot of time to resolve, and it would either freeze the UI or, in the caser of a different thread, prevent the app from closing since the thread wouldn't close until the function finished.

I also remove `RequestCommunityInfoFromMailserverSync`, because it was literally just calling "normal" `RequestCommunityInfoFromMailserver`. So places that called `RequestCommunityInfoFromMailserverSync` now call `RequestCommunityInfoFromMailserver`.


jrainville:
This new endpoint is used in this desktop PR https://github.com/status-im/status-desktop/pull/4793

protocol/messenger_communities.go
```diff
@@ -746,21 +746,21 @@ func (m *Messenger) BanUserFromCommunity(request *requests.BanUserFromCommunity)
...
 }
 
-// RequestCommunityInfoFromMailserverSync installs filter for community and requests its details
-// from mailserver. It will wait for a response and return it if any community is found
-func (m *Messenger) RequestCommunityInfoFromMailserverSync(communityID string) (*communities.Community, error) {
-	return m.requestCommunityInfoFromMailserver(communityID)
+// RequestCommunityInfoFromMailserverAsync installs filter for community and requests its details
+// from mailserver. When response received it will be passed through signals handler
+func (m *Messenger) RequestCommunityInfoFromMailserverAsync(communityID string) error {
+	_, err := m.requestCommunityInfoFromMailserver(communityID, true)
```
cammellos: Why not do the opposite so we don't have to negate it? so passing here false, above true, and then you don't need to swap `async`, we could call it `waitForResponse` as it's called in the code, what do you think?

protocol/messenger_communities.go
```diff
@@ -792,12 +792,16 @@ func (m *Messenger) requestCommunityInfoFromMailserver(communityID string) (*com
...
 		nil,
 		filter,
-		true)
+		!async)
 
 	if err != nil {
 		return nil, err
 	}
 
+	if async {
```
cammellos: I think you still want to return `err`, if the call is async it will return early, but say there's an error and the call is not made at all, you probably want to be notified

jrainville: The error check is done already on the lines above (797), unless I misunderstood your point?

protocol/messenger_communities.go
jrainville: Fixed

protocol/messenger_communities.go
jrainville: Updated to merge the `err` and `waitForResponse` checks

richard-ramos (approved):

protocol/messenger_communities.go
cammellos: oh wait, maybe I did miss something, but basically, always return the error

protocol/messenger_communities.go
```diff
@@ -792,9 +792,9 @@ func (m *Messenger) requestCommunityInfoFromMailserver(communityID string) (*com
 		nil,
 		nil,
 		filter,
-		true)
+		waitForResponse)
 
-	if err != nil {
+	if err != nil || !waitForResponse {
```
cammellos: here you want to always return the error, sorry for the confusion if I asked you to change it before, I must have misread it.

Essentially:


``` 
if err != nil {
  return nil, err
}

if !waitForResponse {
return nil, nil
}

```

cammellos: or I guess it's the same as you wrote, I need to go to bed :) nevermind

jrainville: Alright, it's back to the way it was before, the way you commented

protocol/messenger_communities.go
```diff
@@ -792,12 +792,16 @@ func (m *Messenger) requestCommunityInfoFromMailserver(communityID string) (*com
...
 		filter,
-		true)
+		waitForResponse)
 
 	if err != nil {
 		return nil, err
 	}
 
+	if !waitForResponse {
+		return nil, nil
```
osmaczko, line 802: In this case, the filter for this community will not be removed after the community is fetched. Is that desired behavior?

PR #2540, opened 2022-02-17
Android mute notifications

jo-mut:
Disables sending notifications to muted chats


cammellos (approved):
Don't forget to bump the `VERSION` file

jo-mut:
what do you mean by bump the version file

cammellos:
@jo-mut when you make a change to status-go, you should increase version here:

https://github.com/status-im/status-go/blob/develop/VERSION

the format is 0.x.y where `y` is for non-breaking changes, and `x` is for breaking changes.
So say the version is currently 0.95.1
If you add a breaking change -> 0.96.0
If you add a non-breaking change (such as your case, though is arguable) -> 0.95.2

jo-mut:
oooh cool i understand now

Samyoul (approved):

PR #2541, opened 2022-02-18
feat: display name

richard-ramos:
This feature lets the users set their own display name instead of using the three word name. ([spec](https://github.com/status-im/feature-specs/pull/42))

Part of https://github.com/status-im/status-desktop/issues/4783 

### Summary

- New RPC method: `wakuext_setDisplayName(displayName)`. Should be used in the following scenarios:
  - When __existing__ users sets their display name for the first time
  - When an user wants to change their display name

- When creating a new account, ideally the settings json must have a `display-name` attribute containing the display name of the user. This value should be obtained during onboarding process.


ghost:
### Pull Request Checklist

protocol/messenger_handler.go
```diff
@@ -1420,6 +1431,18 @@ func (m *Messenger) HandleChatIdentity(state *ReceivedMessageState, ci protobuf.
 		return err
 	}
 
+	contact := state.CurrentMessageState.Contact
+
+	if err = ValidateDisplayName(ci.DisplayName, ci.Version); err != nil {
```
cammellos: I think this could be problematic, it's a breaking change and it enforces users to set a username before sending messages, I had a word with John and I think the plan for rolling out the feature are incompatible with this change, but I'll ping you directly,  just to ask a few questions.

richard-ramos:
As suggested by @cammellos , I removed all the breaking changes, and made setting the display name optional

iurimatias (approved):

PR #2542, opened 2022-02-18
refactor: remove identicons

richard-ramos:
With this PR, Identicons are removed from status-go ([spec](https://github.com/status-im/feature-specs/pull/42))

Part of https://github.com/status-im/status-desktop/issues/4783 

Set as draft because I still have to determine if the identicons will be removed in the login screen as well


protocol/contact.go
```diff
@@ -62,13 +61,12 @@ type Contact struct {
 	// Ethereum address of the contact
 	Address string `json:"address,omitempty"`
 	// ENS name of contact
-	Name string `json:"name,omitempty"`
+	EnsName string `json:"ensName,omitempty"`
```
0x-r4bbit, line 64: This looks like this is a breaking change

0x-r4bbit (approved):
Changes look good so far to me.
Just raising awareness that this will have some reverb effects on other downstream projects like Status Desktop and StatusQ with regards to breaking changes.

We need to be careful when updating desktop to this version, it might need some coordination.

cammellos:
@richard-ramos do we have to remove identicon in this PR? It makes is harder to merge, since it means that we also need to implement display name in status-react, but designs are not yet ready, so it would not be mergeable until then.
If we keep both display name and identicon, we can merge this and status-react can keep using identicons until it's ready to integrate display names, and desktop can start using display names straight away, what do you think?

I also have some comments about versioning and display name, but I need to ask John for the details first.

richard-ramos:
I agree with both of you, @cammellos and @PascalPrecht. This PR shouldnt be merged until work in the mobile app and in desktop is complete. I can periodically rebase the changes from `develop` branch into this PR, so when the time comes when work in both apps is complete, we can merge it.

This PR is not strictly necessary for the display name work, but can be seen more of a code cleaning PR since it will remove code that eventually will not be used across all status clients

alaibe (approved):

PR #2551, opened 2022-02-27
refactor: reuse nodeconfig instead of loading it from DB

richard-ramos:
Avoids having to reload the nodeconfig from the DB if it's already available in GethStatusBackend


Samyoul (approved):
Yes, I like it.

PR #2552, opened 2022-02-28
feat: add service for checking if there are new versions available

richard-ramos:
Adds a new API for checking if there is a new version available:

```
{"method":"updates_check","params":[chainId, ensName, currentVersion]}
```
For example, for desktop, we have setup `desktop.status.eth` contenthash to point to an IPFS folder.
If we assume that the current version is `0.1.0`, and we want to verify using mainnet's ENS contracts, the following request can be done:
```
{"method":"updates_check","params":[1, "desktop.status.eth", "0.1.0"]}
```

The IPFS folder should contain besides the executables/installers,  a file called `VERSION` which should contain the following json, indicating the latest version available:
```js
{
  "version": "1.0.0"
}
```

status-go will verify if the latest version is newer than the current, and emit the following signal:
```
{"type":"update.available","event":{"available":true,"version":"1.1.1", "url": "https://....."}}
```
where `available` indicates if there's a new version available or not; `version` indicates the latest version number (with semver format), and `url` the IPFS url of the folder containing the latest version


alaibe:
The service looks good I think but this PR contains a bit more (all the sticker code), maybe rebase so that it is easer to review

alaibe (approved):

richard-ramos:
@churik 
This PR adds a new RPC service that's currently only used in status-desktop. Do you think it would still require testing in mobile? or would being tested in desktop be enough to determine if there are problems with it?

churik:
If it just adds a new service and didn't change the existing code, that will be fine I believe. 
Thank you for your explanation!
I've put the label in order to follow the process.

anastasiyaig:
https://github.com/status-im/status-desktop/pull/5969 corresponding PR in desktop repo

anastasiyaig:
@richard-ramos hey there, so i tested this with desktop and when i do this request 

```
{"method":"updates_check","params":[1, "desktop.status.eth", "0.1.0"]}
```

i get in response

```
{"jsonrpc": "2.0", "id":0, "result":null}
```

is that expected?

Screenshot 2022-06-02 at 13 35 37

PR #2555, opened 2022-03-02
Pass opensea key from env

rasom:
status: ready


node/status_node_services.go
```diff
@@ -52,6 +52,7 @@ var (
 	// ErrRPCClientUnavailable is returned if an RPC client can't be retrieved.
 	// This is a normal situation when a node is stopped.
 	ErrRPCClientUnavailable = errors.New("JSON-RPC client is unavailable")
+	OpenseaKeyFromEnv string
```
cammellos, line 58: Maybe add a comment that the variable is injected at build time, so that's it's more explicit

cammellos (approved):

cammellos:
@jakubgs we need to inject the variable at build time, so when compiling we need do something like:
```
go build -ldflags "-X node.OpenseaKeyFromEnv=$OPENSEA_KEY"
```

is that a change that needs to be made on your side or should be in this PR?

rasom:
@cammellos I added it on status-react side, @churik tested it seems working. I requested @jakubgs's review there 

https://github.com/status-im/status-react/pull/13144

node/status_node_services.go
rasom: ok

rasom: @cammellos done

node/status_node_services.go
```diff
@@ -52,6 +52,8 @@ var (
 	// ErrRPCClientUnavailable is returned if an RPC client can't be retrieved.
 	// This is a normal situation when a node is stopped.
 	ErrRPCClientUnavailable = errors.New("JSON-RPC client is unavailable")
+	// Opensea Key passed from on env var during build time
```
cammellos: Linting will complain, it needs to be the same name

PR #2559, opened 2022-03-03
Removing similar emojis

hesterbruikman:
Removing emojis that are difficult to distinguish

Removes 226 emojis


osmaczko (approved):
Thaks! :)

PR #2571, opened 2022-03-08
feat: add address to permission

alaibe:
Add address to dapps permission


appdatabase/migrations/sql/1646659233_add_address_to_dapp_permisssion.up.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE IF NOT EXISTS dapps_2 (
+    id INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
+    name TEXT,
+    address TEXT NOT NULL default ""
+);
```
richard-ramos, line 6: Since you are already using a primary key, i think you can add a `WITHOUT ROWID`

appdatabase/migrations/sql/1646659233_add_address_to_dapp_permisssion.up.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE IF NOT EXISTS dapps_2 (
+    id INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
+    name TEXT,
```
richard-ramos, line 3: Would it make sense to add a unique constraint over (name,address)?

services/permissions/database.go
```diff
@@ -35,27 +37,55 @@ func (db *Database) AddPermissions(perms DappPermissions) (err error) {
 		}
 		_ = tx.Rollback()
 	}()
+	dRows, err := tx.Query("SELECT id FROM dapps where name = ? AND address = ?", perms.Name, perms.Address)
+	defer dRows.Close()
+	if err != nil {
+		return
+	}
```
richard-ramos:
```suggestion
	dRows, err := tx.Query("SELECT id FROM dapps where name = ? AND address = ?", perms.Name, perms.Address)
	if err != nil {
		return
	}
	defer dRows.Close()
```

richard-ramos (approved):
Nice!

appdatabase/migrations/sql/1646659233_add_address_to_dapp_permisssion.up.sql
alaibe: Nope, it is not possible (just tried)

services/permissions/database.go
```diff
@@ -35,27 +37,56 @@ func (db *Database) AddPermissions(perms DappPermissions) (err error) {
...
 	if err != nil {
 		return
 	}
-	_, err = dInsert.Exec(perms.Name)
-	dInsert.Close()
+	defer pDelete.Close()
+	_, err = pDelete.Exec(id)
 	if err != nil {
 		return
 	}
```
richard-ramos, line 77:
```suggestion
	_, err = pDelete.Exec(id)
	if err != nil {
		return
	}
	defer pDelete.Close()
```

richard-ramos (approved):

PR #2572, opened 2022-03-08
Intrododuce `NodeConfig.TorrentConfig` and CLI flags

### LINKED ISSUE #2563: Add node option to enable/disable torrent client / community history archive protocol
This includes:

1. Introduce a new CLI flag to status-cli to enable community history archive protocol (should be `false` by default), plus a CLI option to configure a desired port that the torrent client should listen on
2. Make the same available as `NodeConfig` parameters (`TorrentConfig`) so apps that initialize the status node in a non-cli way have an option to configure this too.
3. Ensure these configs are stored in a newly introduced `torrent_config` table

Furthermore it should include the following defaults:

```go
type TorrentConfig struct {
  Enabled: bool,
  Port: int, // 9025 or whatever
  DataDir: string, // default data dir + "/archivedata"
  TorrentDir: string, // default data dir + "/torrents"
}
```

0x-r4bbit:
This commit introduces a new `TorrentConfig` as per #2563 with dedicated
default values and new options/flags for the status-go CLI.

Since it's part of `NodeConfig`, which is stored per user in the
database, this commit also adds a migration script that adds a new
`torrent_config` table and database APIs to insert and retreive the
torrent config.

Closes #2563


richard-ramos (approved):

0x-r4bbit:
> ### Pull Request Checklist
> * [ ]  Have you updated the documentation, if impacted (e.g. [docs.status.im](https://docs.status.im/docs/build_status_go.html))?
> * [ ]  Have you tested changes with mobile?
> * [ ]  Have you tested changes with desktop?

I have not added any documentation and didn't test this with mobile, but I did with desktop.

Also updated the PR to include some tests.

params/config_test.go
```diff
@@ -59,13 +64,23 @@ func TestNewConfigFromJSON(t *testing.T) {
...
 		"DataDir": "` + tmpDir + `",
 		"KeyStoreDir": "` + tmpDir + `",
-		"NoDiscovery": true
+		"NoDiscovery": true,
+    "TorrentConfig": {
+      "Port": 9025,
+      "Enabled": false,
+      "DataDir": "` + tmpDir + `/archivedata",
+      "TorrentDir": "` + tmpDir + `/torrents"
+    }
```
0x-r4bbit, line 73: The `make lint-fix` task doesn't seem to take test files into account.

params/config_test.go
```diff
@@ -280,6 +295,21 @@ func TestNodeConfigValidate(t *testing.T) {
...
+				"NetworkId": 1,
+				"DataDir": "/some/dir",
+				"KeyStoreDir": "/some/dir",
+        "TorrentConfig": {
+          "Enabled": true,
+          "Port": 9025,
+          "DataDir": "",
+          "TorrentDir": ""
+        }
+      }`,
```
0x-r4bbit, line 310: Same here.
Let me know if this is crucial and should be updated.

params/config_test.go
Samyoul: I use `go fmt` on any file that the linter complains about. 90% of the time it will resolve the issue.

go.mod
```diff
@@ -73,7 +73,6 @@ require (
 	golang.org/x/crypto v0.0.0-20211202192323-5770296d904e
 	golang.org/x/image v0.0.0-20210220032944-ac19c3e999fb
 	golang.org/x/tools v0.1.2 // indirect
-	google.golang.org/protobuf v1.27.1
```
Samyoul, line 77: Was this intentional?

protocol/messenger.go
```diff
@@ -211,6 +212,7 @@ func (interceptor EnvelopeEventsInterceptor) MailServerRequestExpired(hash types
 
 func NewMessenger(
 	nodeName string,
+	torrentConfig *params.TorrentConfig,
```
Samyoul: Perhaps you could pass the TorrentConfig in with the opts []Option. You'll see a lot of other configs are passed in using this. It will also prevent the need to pass nil into a new Messenger instance

Samyoul (approved):
In the main solid submission, just a couple of points.

go.mod
0x-r4bbit: I **believe** this was done by `make vendor`, changes are however that it was me when I tried to make something work.

Samyoul:
You've got a fail on `TestEncryptionServiceTestSuite/TestMaxKeep`, that looks flaky to me. Rerun the test and it should work out fine.

Reported here https://github.com/status-im/status-go/issues/2590

protocol/messenger.go
0x-r4bbit: Much better indeed. This is done.

go.mod
0x-r4bbit: Can confirm that this was done by `make vendor`

0x-r4bbit:
@Samyoul I've updated the PR according to your comments and rebased it on top of latest `develop`.

Samyoul:
Look good @PascalPrecht

Issue #2568, opened 2022-03-08
Handle magnet link messages


0x-r4bbit:
Detect messages of magnet link message type coming from magnet link channel of joined communities. Ensure they have been signed by community.

Then:

1. Pass magnetlink to torrent client and download `index` file
2. Determine which data pieces to download
3. Download individual archives and turn them into waku messages
4. Update database and keep track of which archives have been downloaded already
5. Replace application messages in database for given time range with archive messages
6. Seed torrent data

PR #2576, opened 2022-03-09
Introduce communty history archive routine 

### LINKED ISSUE #2567: Implement archive tasks interval routine
This routine is about creating torrents, seeding them and distributing the magnet links.
Essentially the flow is as follows:

1. Figure out the start date to create archives for. This is important as it isn't as straightforward as one might think. It can be one of the following:
1.1 The last archive "end date". Which is the timestamp at which the last archive date range ends
1.2. If there's none because no archive was created yet, take the oldest stored waku message (#2515) and use that as a starting point
1.3. If there's no waku messages stored, check if there are any provided by the store nodes (30 days old) and fetch those first - this then becomes the starting point
2. Create archive torrent from waku messages (partitioned by an interval of 7 days)
3. Seed the torrent
4. Create magnet link and distribute it

^  The above routine needs to run in several scenarios with slight adjustments:

1. After bootstrap
1.1. Check if admin communities exist, then run archive protocol routine, also take into account that the last archive was created less then `interval` time ago, so wait until next `interval` before created/seeding the next one. Seed the current one until then
1.2. Then kick off interval and run routine periodically
2. Community creation
2.1. Kick off routine interval after creation if community settings (#2564) say so
3 .Community editing
3.1. Check if archive support has been turned off/on, then start/stop torrent creation/seeding accordingly
4. Community import
4.1. This makes a user a community owner, so the node needs to fetch waku messages of last 30 days then run routine
5. Leaving community
5.1. Need to drop torrents stop any intervals and delete community settings

0x-r4bbit:
This introduces logic needed to:

1. Create `WakuMessageArchive`s and and indices from store waku messages
2. History archive torrent data to disk and create `.torrent` file from
   that
3. Seed and unseed history archive torrents as necessary
4. Starting/stopping the torrent client
5. Enabling/disabling community history support for individual
   components and starting/stopping the routine intervals accordingly

This **does not** yet handle magnet links (#2568)

Closes #2567


0x-r4bbit:
This still needs tests.

protocol/communities/manager.go
```diff
@@ -15,23 +20,36 @@ import (
...
 	"github.com/status-im/status-go/protocol/protobuf"
 	"github.com/status-im/status-go/protocol/requests"
+	"github.com/status-im/status-go/protocol/transport"
 )
 
+var defaultAnnounceList = [][]string{
+	{"udp://tracker.opentrackr.org:1337/announce"},
+	{"udp://tracker.openbittorrent.com:6969/announce"},
+}
+var pieceLength = 1024
```
0x-r4bbit: This value is for testing purposes and needs to be updated

protocol/messenger.go
```diff
@@ -151,11 +151,12 @@ type peerStatus struct {
...
-	activeStoreNode  *peer.ID    // For usage with wakuV2
-	peers            map[string]peerStatus
-	events           chan *p2p.PeerEvent
-	subscription     event.Subscription
+	activeMailserver          *enode.Node // For usage with wakuV1
+	activeStoreNode           *peer.ID    // For usage with wakuV2
+	peers                     map[string]peerStatus
+	events                    chan *p2p.PeerEvent
+	subscription              event.Subscription
+	availabilitySubscriptions map[chan struct{}]chan string
```
0x-r4bbit: This was my attempt to create a way to subscribe inside status-go to whether a mailserver is available.

There's probably a better way to do this, please let me know.

protocol/messenger.go
```diff
@@ -659,6 +661,18 @@ func (m *Messenger) Start() (*MessengerResponse, error) {
...
 		}
+
+		if m.config.torrentConfig.Enabled {
+			adminCommunities, err := m.communitiesManager.Created()
+			if err == nil && len(adminCommunities) > 0 {
+				mailserverAvailable, unsubscribe := m.SubscribeMailserverAvailable()
+				go func() {
+					<-mailserverAvailable
+					close(unsubscribe)
+					m.InitHistoryArchiveTasks(adminCommunities)
```
0x-r4bbit: Here you can see how I wait for `mailserverAvailable` being emitted before we kick off the history archive tasks.

Again, there's probably a nicer more "go-like" way to do this.

protocol/messenger_communities.go
```diff
@@ -139,7 +142,7 @@ func (m *Messenger) JoinCommunity(ctx context.Context, communityID types.HexByte
 
 	communitySettings := communities.CommunitySettings{
 		CommunityID:                  communityID.String(),
-		HistoryArchiveSupportEnabled: true,
```
0x-r4bbit: Ah, this should've probably been `false` for now in the previous commit as well.

protocol/messenger_communities.go
```diff
@@ -1106,6 +1137,220 @@ func (m *Messenger) handleSyncCommunity(messageState *ReceivedMessageState, sync
...
+		}
+	}
+
+	for _, c := range communities {
+
+		if c.Joined() {
+			settings, err := m.communitiesManager.GetCommunitySettingsByID(c.ID())
+			if err != nil {
+				log.Println("failed to get community settings", err)
+				m.logger.Debug("failed to get community settings", zap.Error(err))
```
0x-r4bbit, line 1189: I've doubled the logs here so I can see something in stdout. Happy to remove them for final version, but it's very helpful to see what's going on in realtime..

protocol/messenger_communities.go
```diff
@@ -18,6 +19,8 @@ import (
 	"github.com/status-im/status-go/protocol/transport"
 )
 
+var messageArchiveInterval = 5 * time.Minute
```
0x-r4bbit: This should be 7 * time.Day once this goes to production, but for testing a shorter interval is needed

protocol/messenger_mailserver.go
```diff
@@ -263,7 +263,7 @@ func getPrioritizedBatches() []int {
 	return []int{1, 5, 10}
 }
 
-func (m *Messenger) syncFilters(filters []*transport.Filter) (*MessengerResponse, error) {
+func (m *Messenger) syncFiltersFrom(filters []*transport.Filter, lastRequest uint32) (*MessengerResponse, error) {
```
0x-r4bbit, line 335: This is my attempt to have `syncFilters` provide a way to control how far we want to go back in time to sync from. 

If this is not a good way to do this, please let me know. Basically we need a way to say "Fetch all messages from store node from last 30 days"

protocol/messenger_mailserver.go
```diff
@@ -439,6 +448,10 @@ func (m *Messenger) syncFilters(filters []*transport.Filter) (*MessengerResponse
 	return response, nil
 }
 
+func (m *Messenger) syncFilters(filters []*transport.Filter) (*MessengerResponse, error) {
+	return m.syncFiltersFrom(filters, 0)
+}
```
0x-r4bbit, line 522: Notice here how the original `syncFilters` semantics stay intact

protocol/messenger_mailserver_cycle.go
```diff
@@ -648,6 +651,23 @@ func (m *Messenger) checkMailserverConnection() {
...
+		}
+	}
+}
+
+func (m *Messenger) SubscribeMailserverAvailable() (chan string, chan struct{}) {
+	c := make(chan string)
+	cancel := make(chan struct{})
+	m.mailserverCycle.availabilitySubscriptions[cancel] = c
+	return c, cancel
+}
```
0x-r4bbit: This is the other part of me trying to get a signal inside status-go that a mailserver is available.

Most likely better ways to do this. Appreciate input

protocol/communities/manager.go
```diff
@@ -957,6 +1045,490 @@ func (m *Manager) GetAdminCommunitiesChatIDs() (map[string]bool, error) {
...
+		if from.Equal(endDate) || from.After(endDate) {
+			break
+		}
+		log.Println("Creating message archive for partition (", partition, "). From: ", from, " To: ", to)
+		messages, err := m.persistence.GetWakuMessagesByFilterTopic(topics, uint64(from.Unix()), uint64(to.Unix()))
+		if err != nil {
+			return err
+		}
+
+		wakuMessageArchive := m.createWakuMessageArchive(from, to, messages, topicsAsByteArrays)
```
0x-r4bbit, line 1328: Don't create archive if no messages for this time range exist

protocol/communities/manager.go
```diff
@@ -957,6 +1045,490 @@ func (m *Manager) GetAdminCommunitiesChatIDs() (map[string]bool, error) {
...
+	if err != nil {
+		return err
+	}
+
+	err = os.WriteFile(m.torrentFile(communityID.String()), metaInfoBytes, 0644)
+	if err != nil {
+		return err
+	}
+
+	err = m.persistence.UpdateLastMessageArchiveEndDate(communityID, uint64(from.Unix()))
```
0x-r4bbit: This has a bug. We need to check whether we already have one to update it. Otherwise add it.

0x-r4bbit: This has a bug. We need to check whether we already have one to update it. Otherwise add it.

protocol/communities/manager.go
```diff
@@ -18,23 +21,37 @@ import (
...
 	"github.com/status-im/status-go/protocol/protobuf"
 	"github.com/status-im/status-go/protocol/requests"
+	"github.com/status-im/status-go/protocol/transport"
 )
 
+var defaultAnnounceList = [][]string{
+	{"udp://tracker.opentrackr.org:1337/announce"},
+	{"udp://tracker.openbittorrent.com:6969/announce"},
+}
+var pieceLength = 100 * 1024
```
0x-r4bbit, line 36: Going with 100kb pieces now but we probably want to change this @staheri14 I believe you wanted to propose a good piece size?

protocol/communities/manager.go
0x-r4bbit: This is now addressed above in L:1284

protocol/messenger_communities.go
```diff
@@ -18,6 +19,9 @@ import (
 	"github.com/status-im/status-go/protocol/transport"
 )
 
+// 7 days interval
+var messageArchiveInterval = 7 * 24 * time.Hour
```
0x-r4bbit, line 22: For testing purposes, it might make sense to change this for now or at least make it configurable.

protocol/messenger_communities.go
```diff
@@ -1098,6 +1130,226 @@ func (m *Messenger) handleSyncCommunity(messageState *ReceivedMessageState, sync
...
+				log.Println("failed to get community chats filters", err)
+				m.logger.Debug("failed to get community chats filters", zap.Error(err))
+				continue
+			}
+
+			if len(filters) == 0 {
+				log.Println("no filters or chats for this community", c.IDString())
+				m.logger.Debug("no filters or chats for this community")
+				continue
+			}
```
0x-r4bbit, line 1206: Actually, this is an issue... When we create a community it won't have any filters because it won't have any chats yet. This means, when `InitHistoryArchiveTasks` is called, it won't schedule the interval. We should schedule it here.

protocol/messenger_mailserver_cycle.go
```diff
@@ -650,3 +651,18 @@ func (m *Messenger) getPinnedMailserver() (*mailservers.Mailserver, error) {
...
+		l := len(m.mailserverCycle.availabilitySubscriptions)
+		m.mailserverCycle.availabilitySubscriptions = m.mailserverCycle.availabilitySubscriptions[:l-1]
+	}
+}
+
+func (m *Messenger) SubscribeMailserverAvailable() chan struct{} {
+	c := make(chan struct{})
+	m.mailserverCycle.availabilitySubscriptions = append(m.mailserverCycle.availabilitySubscriptions, c)
+	return c
+}
```
0x-r4bbit, line 668: @Samyoul I've updated this as per our discussions. `EmitMailserverAvailable` now merely emits a signal and closes directly without waiting for a cancel signal from a second channel

protocol/communities/manager.go
```diff
@@ -961,6 +1043,512 @@ func (m *Manager) GetAdminCommunitiesChatIDs() (map[string]bool, error) {
...
+
+	if exists {
+		log.Println("History archive tasks interval already runs for community: ", id)
+		m.logger.Debug("History archive tasks interval already runs for community: ", zap.Any("id", id))
+		return
+	}
+
+	cancel := make(chan struct{})
+	m.historyArchiveTasks[id] = cancel
+	m.historyArchiveTasksWaitGroup.Add(1)
```
0x-r4bbit, line 1178: @Samyoul Here I'm making use of `waitGroup`s. Lemme know if this should be done differently.

protocol/communities/manager.go
```diff
@@ -961,6 +1043,512 @@ func (m *Manager) GetAdminCommunitiesChatIDs() (map[string]bool, error) {
...
+			err = m.CreateAndSeedHistoryArchive(community.ID(), topics, lastArchiveEndDate, to, interval)
+			if err != nil {
+				m.logger.Debug("failed to create and seed history archive", zap.Error(err))
+				continue
+			}
+			dispatchMagnetlink(community.ID())
+		case <-cancel:
+			m.UnseedHistoryArchiveTorrent(community.ID())
+			delete(m.historyArchiveTasks, id)
+			m.historyArchiveTasksWaitGroup.Done()
```
0x-r4bbit, line 1218: @Samyoul We're decreasing the waitgroup counter here

protocol/communities/manager.go
```diff
@@ -961,6 +1043,512 @@ func (m *Manager) GetAdminCommunitiesChatIDs() (map[string]bool, error) {
...
+}
+
+func (m *Manager) StopHistoryArchiveTasksIntervals() {
+	for _, t := range m.historyArchiveTasks {
+		close(t)
+	}
+	// Stoping archive interval tasks is async, so we need
+	// to wait for all of them to be closed before we shutdown
+	// the torrent client
+	m.historyArchiveTasksWaitGroup.Wait()
```
0x-r4bbit, line 1231: @Samyoul And wait for it to be done here, making `StopHistoryArchiveTasksIntervals` a somewhat synchronous API

protocol/messenger.go
```diff
@@ -623,6 +625,7 @@ func (m *Messenger) Start() (*MessengerResponse, error) {
 
 	m.handleEncryptionLayerSubscriptions(subscriptions)
 	m.handleCommunitiesSubscription(m.communitiesManager.Subscribe())
+	m.handleCommunitiesHistoryArchivesSubscription(m.communitiesManager.Subscribe())
```
0x-r4bbit, line 633: I'm reusing the existing community manager subscriptions to signal community history events as well

protocol/messenger_communities.go
```diff
@@ -59,6 +63,56 @@ func (m *Messenger) publishOrgInvitation(org *communities.Community, invitation
...
+					)
+				}
+
+				if sub.NoHistoryArchivesCreatedSignal != nil {
+					m.config.messengerSignalsHandler.NoHistoryArchivesCreated(
+						sub.NoHistoryArchivesCreatedSignal.CommunityID,
+						sub.NoHistoryArchivesCreatedSignal.From,
+						sub.NoHistoryArchivesCreatedSignal.To,
+					)
+				}
```
0x-r4bbit, line 93: These look a bit unfortunate. We emit the `History*Signal`'s on the `Subscription` and then use the attached data to send them to connected clients by destructing the attached data, which is then reconstructed again.

The reason I do this is because we already need the `History*Signal`s anyways to inform connected clients, so it made sense to reuse them here in the subscription.

Lemme know if this should be done differently.

protocol/communities/manager.go
```diff
@@ -67,8 +89,13 @@ func NewManager(identity *ecdsa.PublicKey, db *sql.DB, logger *zap.Logger, verif
...
 type Subscription struct {
-	Community   *Community
-	Invitations []*protobuf.CommunityInvitation
+	Community                      *Community
+	Invitations                    []*protobuf.CommunityInvitation
+	CreatingHistoryArchivesSignal  *signal.CreatingHistoryArchivesSignal
+	HistoryArchivesCreatedSignal   *signal.HistoryArchivesCreatedSignal
+	NoHistoryArchivesCreatedSignal *signal.NoHistoryArchivesCreatedSignal
+	HistoryArchivesSeedingSignal   *signal.HistoryArchivesSeedingSignal
+	HistoryArchivesUnseededSignal  *signal.HistoryArchivesUnseededSignal
```
0x-r4bbit, line 97: I could build a separate `Subscription` type for this, but I thought there's no reason not to put those events in here.

0x-r4bbit:
I know this is a large PR, but would be cool if we could get it reviewed @cammellos @Samyoul 

It's been through a bunch of iterations (and have to go through another one as there were changes in `develop` that force a rebase here, also because of the linter needing an upgrade), if you need any info or a walk-through of this, please let me know

protocol/messenger_communities.go
0x-r4bbit: This has been addressed by now..

protocol/messenger_communities.go
```diff
@@ -1095,6 +1172,223 @@ func (m *Messenger) handleSyncCommunity(messageState *ReceivedMessageState, sync
...
+	log.Println("Dispatching magnet link for community ", community.IDString())
+	_, err = m.sender.SendPublic(context.Background(), chatID, rawMessage)
+	if err != nil {
+		return err
+	}
+
+	err = m.communitiesManager.UpdateCommunityDescriptionMagnetlinkMessageClock(community.ID(), magnetLinkMessage.Clock)
+	if err != nil {
+		return err
+	}
```
0x-r4bbit, line 1329: We're updating the magnet link clock in the community description. This is needed so that, when users receive community invites, they get the latest magnetlink clock (otherwise, the moment they'll join, they'll try to process all past magnetlink messages they receive)

protocol/messenger.go
```diff
@@ -657,6 +660,17 @@ func (m *Messenger) Start() (*MessengerResponse, error) {
...
+		adminCommunities, err := m.communitiesManager.Created()
+		if err == nil && len(adminCommunities) > 0 {
+			available := m.SubscribeMailserverAvailable()
+			go func() {
+				<-available
+				m.InitHistoryArchiveTasks(adminCommunities)
+			}()
+		}
+	}
+
```
Samyoul: Ok, I've found the fix for the failing test.

```suggestion
	if m.config.torrentConfig != nil && m.config.torrentConfig.Enabled {
		adminCommunities, err := m.communitiesManager.Created()
		if err == nil && len(adminCommunities) > 0 {
			available := m.SubscribeMailserverAvailable()
			go func() {
				<-available
				m.InitHistoryArchiveTasks(adminCommunities)
			}()
		}
	}

```

0x-r4bbit: Argh... I'm sorry you had to go after this. Thank you!

Samyoul: It's ok, these sorts of bugs are very hard to find. When a package uses `recover` it makes discovery so tricky.

protocol/messenger_communities.go
```diff
@@ -1095,6 +1172,223 @@ func (m *Messenger) handleSyncCommunity(messageState *ReceivedMessageState, sync
...
+			filters, err := m.communitiesManager.GetCommunityChatsFilters(c.ID())
+			if err != nil {
+				log.Println("failed to get community chats filters", err)
+				m.logger.Debug("failed to get community chats filters", zap.Error(err))
+				continue
+			}
+
+			if len(filters) == 0 {
+				log.Println("no filters or chats for this community starting interval", c.IDString())
+				m.logger.Debug("no filters or chats for this community starting interval")
```
Samyoul: I notice a number of double logs like this are these for you own development flow?

protocol/protobuf/communities.proto
```diff
@@ -43,6 +43,7 @@ message CommunityDescription {
   map<string,CommunityChat> chats = 6;
   repeated string ban_list = 7;
   map<string,CommunityCategory> categories = 8;
+  uint64 magnetlink_clock = 9;
```
Samyoul: Just a naming suggestion
```suggestion
  uint64 archive_magnetlink_clock = 9;
```
In case in future communities use magnet links for other things.

Samyoul (approved):
Really cool PR. This must have taken so much time to do. Very thorough work.

I would replace all usages of `log` with the uber logger library, just as a matter of consistency and so we know the source of our logs.

I also debated with myself about suggesting breaking out the torrent related logic from the `Manager` into a dedicated `TorrentManager`, but I am not sure it is really necessary.

I really like that you chose to use the `waitgroup`, and it is implemented well.

---

Also sorry it took so long to review, there was just a lot going on in this PR.

0x-r4bbit:
Removed all usages of `log.Println` and replaced them with `zap.logger`.
Although it was very nice and useful to get the live logs about how the archive protocol behaves. Makes it significantly harder to monitor now.

I've also updated the protobuf property according to you suggestion.

I have not moved the torrent related APIs to a separate manager yet. Happy to do that in the future but I'd prefer landing this PR before introducing another set of bigger changes. Been rebasing this one for a month now :D

PR #2582, opened 2022-03-10
feat: Added emoji params for a wallet account

Khushboo-dev-cpp:


richard-ramos (approved):

services/accounts/accounts.go
```diff
@@ -50,11 +50,12 @@ func (api *API) DeleteAccount(ctx context.Context, address types.Address) error
 	return api.db.DeleteAccount(address)
 }
 
-func (api *API) AddAccountWatch(ctx context.Context, address string, name string, color string) error {
+func (api *API) AddAccountWatch(ctx context.Context, address string, name string, emoji string, color string) error {
```
alaibe: pls add the new parameter to the end. Otherwise this will break the API for other client

Because default is string value is empty string, it means that if mobile use this API they will inverse emoji and color while if it is at the end they will just set emoji to empty string

same comment for all the api endpoint

alaibe (changes requested):
Requesting change only to avoid the breaking change

Khushboo-dev-cpp:
alll done @alaibe thanks for pointing that out!

alaibe (approved):
Nice work

PR #2585, opened 2022-03-14
Handle history archive magnetlink messages

### LINKED ISSUE #2568: Handle magnet link messages

0x-r4bbit:
This introduces the ability for status notes to handle community
history archive magnetlinks. To make this work, a few things are needed:

1. A new database table has been introduced to store message archive
   hashes. This is necessary so status nodes can determine whether or
   not they need to download a certain archive
2. The messenger's `handleRetrievedMessages()` has been exteded to take
   magnetlink messages into account
3. New APIs were added to download torrent data given a magnetlink and
   also to extract messages from downloaded archives, which are then
   later fed to `handleRetrievedMessages`

Closes #2568


0x-r4bbit:
This needs tests and some fine-tuning but should be in a reviewable shape.

protocol/communities/manager.go
```diff
@@ -1455,6 +1482,130 @@ func (m *Manager) IsSeedingHistoryArchiveTorrent(communityID types.HexBytes) boo
...
+	ml, _ := metainfo.ParseMagnetUri(magnetlink)
+
+	log.Println("Adding torrent from magnetlink for community: ", id)
+	torrent, err := m.torrentClient.AddMagnet(magnetlink)
+	if err != nil {
+		return nil, err
+	}
+	m.torrentTasks[id] = ml.InfoHash
+
+	<-torrent.GotInfo()
```
0x-r4bbit: So here we're waiting for the magnetlink to emit its metadata (this is needed for us to know what files the torrent will have, so having this info is crucial).

During testing, I've experienced that it sometimes takes a long time to receive a signal, at which point nothing happens here.

protocol/communities/manager.go
```diff
@@ -1455,6 +1482,130 @@ func (m *Manager) IsSeedingHistoryArchiveTorrent(communityID types.HexBytes) boo
...
+	files := torrent.Files()
+
+	indexFile := files[1]
```
0x-r4bbit: We always expect exactly two files `data` and `index`. So accessing `[1]` is safe, but we should probably still dynamically determine, in case we'll add more files to these torrent sin the future (maybe due to https://github.com/status-im/specs/pull/167)

protocol/messenger_handler.go
```diff
@@ -626,6 +627,63 @@ func (m *Messenger) HandleCommunityInvitation(state *ReceivedMessageState, signe
...
+					log.Println("failed to write history archive messages to database", err)
+					m.logger.Debug("failed to write history archive messages to database", zap.Error(err))
+					return
+				}
+			}()
+
+			return m.communitiesManager.UpdateMagnetlinkMessageClock(id, clock)
+		}
+	}
+	return nil
```
0x-r4bbit, line 750: Originally I had most of this logic inside `communitiesManaegr`, however because we need access to `handleRetrievedMessages()` I've moved this here. Lemme know if there's a different way to get access to `messenger` APIs inside `communitiesManager` otherwise.

protocol/message_persistence.go
```diff
@@ -1853,7 +1853,7 @@ func (db sqlitePersistence) GetDeletes(messageID string, from string) ([]*Delete
 }
 
 func (db sqlitePersistence) SaveEdit(editMessage EditMessage) error {
-	_, err := db.db.Exec(`INSERT INTO user_messages_edits (clock, chat_id, message_id, text, source, id) VALUES(?,?,?,?,?,?)`, editMessage.Clock, editMessage.ChatId, editMessage.MessageId, editMessage.Text, editMessage.From, editMessage.ID)
+	_, err := db.db.Exec(`INSERT OR REPLACE INTO user_messages_edits (clock, chat_id, message_id, text, source, id) VALUES(?,?,?,?,?,?)`, editMessage.Clock, editMessage.ChatId, editMessage.MessageId, editMessage.Text, editMessage.From, editMessage.ID)
```
0x-r4bbit: This is probably not what we want. I noticed that, when replaying history messages into the DB, message edits were causing SQL errors, saying that the entries have to be unique. So I've added this here to satisfy it.

0x-r4bbit:
Please review the last commit of this PR

protocol/communities/manager.go
cammellos: Can it be that `files` is of different length? (say 1 or 0), even in cases where the torrent is malformed.
If that's the case, best to error check and return an error, since a panic would just crash the app, and if someone manages to push a malformed torrent for example, it would just crash all clients.

0x-r4bbit: Yea good point. I'll update it so it'll search for an `index` file and if it doesn't exist it'll return early as the torrent is malformed

protocol/message_persistence.go
0x-r4bbit: @cammellos Would be cool if you could confirm whether this here is okay

protocol/communities/manager.go
0x-r4bbit: Updated

protocol/message_persistence.go
cammellos: As far as I can see, it should not matter, since all the fields are not modified after insertion, so it would just be replacing the same message. 
Though if you make it not return an error, the code will process the edit multiple times, not sure whether that is an issue (in theory it should be fine), but we need to make sure we test edits before merging, just to make sure.

0x-r4bbit: As per offline discussion, I wanted to changes this to leave out the `OR REPLACE` and check for the error, which can then be interpreted as "don't process this message".

While doing this, I realized I couldn't reproduce the error anymore. So I'll just remove this change for now.

protocol/communities/manager.go
```diff
@@ -87,6 +88,25 @@ func NewManager(identity *ecdsa.PublicKey, db *sql.DB, logger *zap.Logger, verif
...
+	return len(md)
+}
+
+func (md archiveMDSlice) Swap(i, j int) {
+	md[i], md[j] = md[j], md[i]
+}
+
+func (md archiveMDSlice) Less(i, j int) bool {
+	return md[i].from > md[j].from
+}
```
0x-r4bbit, line 108: This is later used when we receive the archive index to make it sortable. This is needed to ensure latest archives are always downloaded first (instead of random).

protocol/communities/manager.go
```diff
@@ -1526,6 +1596,154 @@ func (m *Manager) IsSeedingHistoryArchiveTorrent(communityID types.HexBytes) boo
...
+
+	var archiveIDs []string
+
+	archiveHashes := make(archiveMDSlice, 0, len(index.Archives))
+
+	for hash, metadata := range index.Archives {
+		archiveHashes = append(archiveHashes, &archiveMetadata{hash: hash, from: metadata.Metadata.From})
+	}
+
+	sort.Sort(archiveHashes)
```
0x-r4bbit: Here, we're creating the slice of the archive metadata to make it sortable and then sort it, making newer archives come first.

protocol/communities/manager.go
```diff
@@ -1526,6 +1596,154 @@ func (m *Manager) IsSeedingHistoryArchiveTorrent(communityID types.HexBytes) boo
 	return ok && torrent.Seeding()
 }
 
+func (m *Manager) DownloadHistoryArchivesByMagnetlink(communityID types.HexBytes, magnetlink string) ([]string, error) {
+
+	id := communityID.String()
+	ml, _ := metainfo.ParseMagnetUri(magnetlink)
```
richard-ramos: Is it possible that ParseMagnetUri fails if the magnet link somehow is incorrect?
The reason for the question is that it seems the a possible error here is being ignored.

protocol/messenger.go
```diff
@@ -2917,7 +2917,7 @@ func (m *Messenger) RetrieveAll() (*MessengerResponse, error) {
 		return nil, err
 	}
 
-	return m.handleRetrievedMessages(chatWithMessages)
+	return m.handleRetrievedMessages(chatWithMessages, true)
```
richard-ramos, line 3004: Wouldnt this cause everyone to store the waku messages automatically if they're are admins of a community?
Will this be an option in the future, or should admins be required always to store the messages? 
i.e. I might want to store messages if i'm using my desktop machine, but on mobile i might not.

protocol/messenger.go
```diff
@@ -3672,6 +3672,15 @@ func (m *Messenger) handleRetrievedMessages(chatWithMessages map[transport.Filte
 							continue
 						}
 
+					case protobuf.CommunityMessageArchiveMagnetlink:
+						logger.Debug("Handling CommunityMessageArchiveMagnetlink")
+						magnetlinkMessage := msg.ParsedMessage.Interface().(protobuf.CommunityMessageArchiveMagnetlink)
+						err = m.HandleHistoryArchiveMagnetlinkMessage(messageState, publicKey, magnetlinkMessage.MagnetUri, magnetlinkMessage.Clock)
+						if err != nil {
+							logger.Warn("failed to handle CommunityMessageArchiveMagnetlink", zap.Error(err))
+							continue
```
richard-ramos, line 3681:
```suggestion
							allMessagesProcessed = false
							continue
```

protocol/communities/manager.go
0x-r4bbit: Yeap good point!

protocol/messenger.go
0x-r4bbit: Right so I had a talk with @John-44 about this and we decided that admins should always store every community message as waku message. However we didn't talk about how this should behave on mobile (where the torrent protocol is turned off by default). So this is a valid concern..

One alternative would be to always put storing the waku messages behind the same flag of whether the protocol is enabled or not. That would apply the same on mobile and desktop.

Just means that, if admins turn off the protocol for whatever reason, they will not store messages as waku messages, resulting in them not being part of the archives.

0x-r4bbit: So to clarify, we're already storing waku messages (also on mobile). This flag here has just been introduced for when we call `handleRetrievedMessages` to replay archive messages into the DB. In that case, we don't want to also store waku messages (in which case the flag is `false`)

protocol/communities/manager.go
```diff
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"fmt"
 	"os"
+	"sort"
```
0x-r4bbit, line 8: This comment here is just to ensure this file is being reviewed, as github collapses it due to the amount of changes.

multiaccounts/accounts/database.go
```diff
@@ -219,5 +219,5 @@ func (db *Database) AddressExists(address types.Address) (exists bool, err error
 }
 
 func (db *Database) GetNodeConfig() (*params.NodeConfig, error) {
-	return nodecfg.GetNodeConfig(db.db)
+	return nodecfg.GetNodeConfigFromDB(db.db)
```
0x-r4bbit: I needed to make this change because `GeNodeConfig` no longer exists. @richard-ramos is that related to you changes?

richard-ramos: yes, it was surprising that this wasnt caught by the CI

protocol/communities/manager.go
```diff
@@ -1530,6 +1600,157 @@ func (m *Manager) IsSeedingHistoryArchiveTorrent(communityID types.HexBytes) boo
...
 
+func (m *Manager) DownloadHistoryArchivesByMagnetlink(communityID types.HexBytes, magnetlink string) ([]string, error) {
+
+	id := communityID.String()
+	ml, err := metainfo.ParseMagnetUri(magnetlink)
+	if err != nil {
+		return nil, err
+	}
+
+	m.logger.Debug("Adding torrent via magnetlink for community", zap.Any("id", id), zap.Any("magnetlink", magnetlink))
```
cammellos: You can use strings for zap `zap.String("id", id.String())` or `zap.String("magnet", magnetlink.String()), so you get the hexencoded version (otherwise is going to be base encoded)

protocol/communities/manager.go
cammellos: What happens if you are offline? We should probably timeout this if it could hang forever

protocol/communities/manager.go
```diff
@@ -1530,6 +1600,157 @@ func (m *Manager) IsSeedingHistoryArchiveTorrent(communityID types.HexBytes) boo
...
+		// We're dealing with a malformed torrent, so don't do anything
+		return nil, nil
+	}
+
+	indexFile := files[i]
+
+	m.logger.Debug("Downloading history archive index")
+	indexFile.Download()
+	for {
+		if indexFile.BytesCompleted() == indexFile.Length() {
```
cammellos: that's a  busy loop, you want to wait you should use a different pattern (use a channel,select/polling etc)

protocol/communities/manager.go
```diff
@@ -1530,6 +1600,157 @@ func (m *Manager) IsSeedingHistoryArchiveTorrent(communityID types.HexBytes) boo
...
+			continue
+		}
+		if hasArchive {
+			continue
+		}
+
+		startIndex := int(metadata.Offset) / pieceLength
+		endIndex := startIndex + int(metadata.Size)/pieceLength
+
+		m.logger.Debug("Downloading data for message archive", zap.Any("hash", hash))
```
cammellos: we use lowercase for log entries

protocol/communities/manager.go
```diff
@@ -1530,6 +1600,157 @@ func (m *Manager) IsSeedingHistoryArchiveTorrent(communityID types.HexBytes) boo
...
+		data := totalData[metadata.Offset : metadata.Offset+metadata.Size-metadata.Padding]
+
+		err := proto.Unmarshal(data, archive)
+		if err != nil {
+			m.logger.Debug("Failed to unmarshal WakuMessageArchive", zap.Error(err))
+			continue
+		}
+
+		for _, message := range archive.Messages {
+			filter := m.transport.FilterByTopic(message.Topic)
```
cammellos, line 1743: what are you checking exactly here? meaning, you are checking the topic matches, but generally that's not enough to ensure the message is for you and you should process it. 
Which messages should we process and in which case they should not be processed?

protocol/messenger_handler.go
```diff
@@ -683,6 +684,63 @@ func (m *Messenger) HandleCommunityInvitation(state *ReceivedMessageState, signe
...
+
+	id := types.HexBytes(crypto.CompressPubkey(communityPubKey))
+	settings, err := m.communitiesManager.GetCommunitySettingsByID(id)
+	if err != nil {
+		m.logger.Debug("Couldn't get community settings for community with id: ", zap.Any("id", id))
+		return err
+	}
+
+	if m.config.torrentConfig != nil && m.config.torrentConfig.Enabled && settings.HistoryArchiveSupportEnabled {
+		signedByOwnedCommunity, err := m.communitiesManager.IsAdminCommunity(communityPubKey)
```
cammellos, line 699: At a first glance, I think we should be using the key of the community to send this messages, rather than the admins? (there's arguments for both, but worth considering)

protocol/messenger_handler.go
```diff
@@ -683,6 +684,63 @@ func (m *Messenger) HandleCommunityInvitation(state *ReceivedMessageState, signe
...
+			return err
+		}
+		lastClock, err := m.communitiesManager.GetMagnetlinkMessageClock(id)
+		if err != nil {
+			return err
+		}
+		// We are only interested in a community archive magnet link
+		// if it originates from a community that the current account is
+		// part of and doesn't own the private key at the same time
+		if !signedByOwnedCommunity && joinedCommunity && clock >= lastClock {
```
cammellos, line 714: Wondering why we are not interested if we are the admin, is the assumption that we would have the same messages therefore is not useful/is redundant?

protocol/messenger_handler.go
```diff
@@ -683,6 +684,63 @@ func (m *Messenger) HandleCommunityInvitation(state *ReceivedMessageState, signe
...
+				}
+
+				messagesToHandle, err := m.communitiesManager.ExtractMessagesFromHistoryArchives(id, downloadedArchiveIDs)
+				if err != nil {
+					log.Println("failed to extract history archive messages", err)
+					m.logger.Debug("failed to extract history archive messages", zap.Error(err))
+					return
+				}
+
+				_, err = m.handleRetrievedMessages(messagesToHandle, false)
```
cammellos: This looks a bit dangerous, we currently don't run `handleRetrieveMessages` in parallel, so weird things might happen

protocol/messenger_handler.go
cammellos: You can use pub/sub, but we can have a look at it later

protocol/messenger_handler.go
0x-r4bbit: So the message with the magnet link **is** signed by the community key.

However this part of the code is handling the magnet link messages. So what we're checking here is: Am I the admin/owner of this community? If yes, then I'm not interested in handling the magnet link because I'm expected to be node that has created the magnet link message from a history archive in the first place.

In other words: only members of the community are interested in this, not admins/owners

protocol/messenger_handler.go
0x-r4bbit: As mentioned above, the admin is the one that publishes the history archives via magnetlinks, so that node already has the history. Only members are downloading torrent data

protocol/communities/manager.go
0x-r4bbit: This function (`ExtractMessagesFromHistoryArchive`) aims to create a `map[transport.Filter][]*types.Message` from a downloaded archive, so they can be passed to `handleRetrievedMessages`.

Because we're only getting the topic from each archive message, we need to figure out what filter they belong to, so we can create the correct return type.
All messages extracted from an archive (and passed to `handleRetrievedMessages`) are considered to be processed.

protocol/messenger_handler.go
0x-r4bbit: Hm.. I wonder what such weird things could be, given that `handleRetrieveMessages` has various checks for incoming messages and handling out of order messages etc.

Any suggestions on how else this should be done?

cammellos: First thing that comes to mind is to have a channel, and handle retrieved messages it's called on it, so you have 2 producers (polling from the network, and torrent) and a single consumer, a bit more complex, but we don't have to think about the code running in parallel, as it might lead to difficult to find bugs.

protocol/communities/manager.go
```diff
@@ -63,8 +65,16 @@ func NewManager(identity *ecdsa.PublicKey, db *sql.DB, logger *zap.Logger, verif
 		}
 	}
 
+	lc := zap.NewDevelopmentConfig()
+	lc.OutputPaths = []string{"stdout"}
+	archiveLogger, err := lc.Build()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create a archive logger")
+	}
```
0x-r4bbit: I've added a new logger so we get some output in stdout when archives are created/downloaded (easier for debugging, kinda hard when stuff only gets written to geth.log)

protocol/communities/manager.go
```diff
@@ -1530,6 +1609,164 @@ func (m *Manager) IsSeedingHistoryArchiveTorrent(communityID types.HexBytes) boo
...
+	torrent, err := m.torrentClient.AddMagnet(magnetlink)
+	if err != nil {
+		return nil, err
+	}
+	m.torrentTasks[id] = ml.InfoHash
+
+	select {
+	case <-time.After(20 * time.Second):
+		return nil, errors.New("torrent has timed out")
+	case <-torrent.GotInfo():
```
0x-r4bbit, line 1621: @cammellos I've added a time out of 20 seconds here in case `torrent.GotInfo()` doesn't emit anything

protocol/communities/manager.go
```diff
@@ -1530,6 +1609,164 @@ func (m *Manager) IsSeedingHistoryArchiveTorrent(communityID types.HexBytes) boo
...
+
+		indexFile := files[i]
+		indexFile.Download()
+		m.archiveLogger.Debug("downloading history archive index")
+
+		ticker := time.NewTicker(200 * time.Millisecond)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ticker.C:
```
0x-r4bbit: To make the busy loop less busy, we're now using a 200 ms ticker, which should be more than enough.

protocol/messenger_handler.go
0x-r4bbit: Okay I'll see if I can make that happen. Thanks for your input!

cmd/statusd/main.go
```diff
@@ -412,6 +414,13 @@ func retrieveMessagesLoop(messenger *protocol.Messenger, tick time.Duration, can
 				logger.Error("failed to retrieve raw messages", "err", err)
 				continue
 			}
+		case messagesToHandle := <-messageArchivesDownloaded:
+			_, err := messenger.HandleRetrievedMessages(messagesToHandle, false)
+			if err != nil {
+				logger.Error("failed to write history archive messages to database", err)
+				continue
+			}
```
0x-r4bbit: @cammellos lemme know if this is what you had in mind

protocol/communities/persistence.go
```diff
@@ -443,6 +443,11 @@ func (p *Persistence) GetWakuMessagesByFilterTopic(topics []types.TopicType, fro
 	return messages, nil
 }
 
+func (p *Persistence) HasCommunityArchiveInfo(communityID types.HexBytes) (exists bool, err error) {
+	err = p.db.QueryRow(`SELECT count(1) FROM communities_archive_info WHERE community_id = ?`, communityID.String()).Scan(&exists)
```
richard-ramos: Not sure if it's worth it, but depending on the number of rows in `communities_archive_info`, `EXISTS` is more efficient
```suggestion
	err = p.db.QueryRow(`SELECT EXISTS (SELECT 1 FROM communities_archive_info WHERE community_id = ?)`, communityID.String()).Scan(&exists)
```

protocol/messenger_handler.go
```diff
@@ -683,6 +684,58 @@ func (m *Messenger) HandleCommunityInvitation(state *ReceivedMessageState, signe
...
+func (m *Messenger) HandleHistoryArchiveMagnetlinkMessage(state *ReceivedMessageState, communityPubKey *ecdsa.PublicKey, magnetlink string, clock uint64) error {
+
+	id := types.HexBytes(crypto.CompressPubkey(communityPubKey))
+	settings, err := m.communitiesManager.GetCommunitySettingsByID(id)
+	if err != nil {
+		m.logger.Debug("Couldn't get community settings for community with id: ", zap.Any("id", id))
+		return err
+	}
+
+	if m.config.torrentConfig != nil && m.config.torrentConfig.Enabled && settings.HistoryArchiveSupportEnabled {
```
richard-ramos: `settings` can be null if communities_settings does not contain the `id`

protocol/messenger_handler.go
```diff
@@ -683,6 +684,58 @@ func (m *Messenger) HandleCommunityInvitation(state *ReceivedMessageState, signe
...
+		if err != nil {
+			return err
+		}
+		// We are only interested in a community archive magnet link
+		// if it originates from a community that the current account is
+		// part of and doesn't own the private key at the same time
+		if !signedByOwnedCommunity && joinedCommunity && clock >= lastClock {
+
+			m.communitiesManager.UnseedHistoryArchiveTorrent(id)
+			go func() {
```
richard-ramos, line 718: Is this a 'slow' operation?  what happens if an user logouts while the goroutine is still executing? If it's important for the goroutine to finish before logout, maybe it's a good idea to add a `sync.WaitGroup` in messenger and wait for it in `func (m *Messenger) Shutdown() (err error)`  (or maybe this is already handled somewhere else?)

richard-ramos (approved):
Left some comments, but in general LGTM! :medal_military:

richard-ramos:
Please update the VERSION file

protocol/messenger_handler.go
0x-r4bbit: This is a very good point. Adding this now

protocol/messenger_handler.go
0x-r4bbit: Good catch

protocol/communities/persistence.go
0x-r4bbit: Thanks, this is cool!

0x-r4bbit:
@cammellos @richard-ramos thanks for all your comments.

- I've updated the SELECT query for checking the existence of archive hash info
- I've added a waitgroup to ensure shutting down the client waits for downloading torrent data to finish
- I've reduced the time interval for checking whether the download of the torrent index file is done to 100ms
- I've introduced a mutex lock for `handleRetrievedMessages` to ensure it doesn't conflict when executed via message archive protocol

protocol/messenger_handler.go
cammellos: You also need to publish the response, otherwise the client will be in inconsistent state, https://github.com/status-im/status-go/blob/f66ad3d3dd754a8c2e789e3339c5bf0b970129c1/services/ext/service.go#L210 , have you tested this on desktop to see if changes are reflected on the frontend?

churik:
If follow the processes (which are still TBD, but better to check twice) we need to check it on the mobile client-side briefly and on the desktop side as well. 
Only in case, you're absolutely sure that it will not introduce any regression issues we can merge it as it is
cc @anastasiyaig

anastasiyaig:
@churik u may test it on mobile side and provide feedback. It was not yet tested on desktop, @elina2015 please check this today

protocol/messenger_handler.go
0x-r4bbit: In desktop, what we do is we listen for the `HistoryDownloaded` signal, and if that has happened we refetch messages from status-go for the current active channel.

0x-r4bbit:
> @churik u may test it on mobile side and provide feedback. It was not yet tested on desktop

Just to let you know, this feature is completely turned off on mobile, so the only thing to test/check there is that it indeed doesn't create archives, nor downloads them.

qoqobolo:
> Just to let you know, this feature is completely turned off on mobile, so the only thing to test/check there is that it indeed doesn't create archives, nor downloads them.

Hey, thank you for the clarification!

But for context: in such cases, this is not really testing of the feature itself. It was decided (_at least until there’s a better solution, suggestions are still welcome and appriciated_) to create and test clones of Go pull requests in client repositories (Mobile and Desktop) in order to avoid issues that sometimes arise after PRs are merged without any testing on the Go side ([discussion here](https://discord.com/channels/624205794384281629/626738033701683200/971340906001219624)).
So we just run e2e tests, sometimes do some smoke-testing using the corresponding PR in our repo, and then approve the Go PR.

Regarding this particular PR: looks good from the Mobile side, no issues found.

PR #2589, opened 2022-03-15
feat: Added new endpoint to get password strength info

### LINKED ISSUE #4980: password validation

noeliaSD:
Here my first contribution to `status-go`!

### What does the PR do
By integrating `zxcvbn` module, it has been added a new endpoint to get password strength quality information like Entropy, CrackTime, CrackTimeDisplay, Score, MatchSequence and CalcTime.

Together with [PR #5044](https://github.com/status-im/status-desktop/pull/5044) and [PR #8](https://github.com/status-im/nim-status-go/pull/8) it closes [#4980](https://github.com/status-im/status-desktop/issues/4980)


services/ext/api.go
```diff
@@ -1045,3 +1047,12 @@ func topicsToBloom(topics ...types.TopicType) []byte {
 func TopicsToBloom(topics ...types.TopicType) []byte {
 	return topicsToBloom(topics...)
 }
+
+// GetPasswordStrength uses zxcvbn module and generates a JSON containing information about the quality of the given password
+// (Entropy, CrackTime, CrackTimeDisplay, Score, MatchSequence and CalcTime).
+// userInputs argument can be whatever list of strings like user's personal info or site-specific vocabulary that zxcvbn will
+// make use to determine the result.
+// For more details on usage see https://github.com/status-im/zxcvbn-go
+func (api *PublicAPI) GetPasswordStrength(request *requests.GetPasswordStrength) scoring.MinEntropyMatch {
```
alaibe: Don t we have an auth service or an account part where we could have this function?
@richard-ramos do u know more?

cammellos: Also, worth checking if the node is running when you need to use this endpoint, if it's pre-login it might not be, in which case we generally use c-bindings

richard-ramos: There's this services/accounts/accounts.go "accounts_*" API, but like @cammellos mentions this probably should be done as a c-bindinng instead since the RPC server is not available before login, The function should be moved to mobile/status.go instead.

richard-ramos:
Execute  `make vendor` so the dependncies are automagically copied into vendor/ folder

noeliaSD:
Please review it again @richard-ramos, @alaibe and @cammellos. I hope it addresses all your comments! Thanks!

richard-ramos (approved):

alaibe (approved):
Good job

PR #2594, opened 2022-03-17
fix: move visual-identity servce to protocol/identity

osmaczko:
It is required to be called before RPC server is running on client side


osmaczko:
tested in: https://github.com/status-im/status-desktop/pull/5083

richard-ramos (approved):

PR #2599, opened 2022-03-22
feat: Added new PasswordStrengthScore method

### LINKED ISSUE #5096: Integrate password validation (zxcvbn lib) in new password screens
Use new _PrivacyStore_ method `getPasswordStrengthScore` and link it to the new password strength bar value.

noeliaSD:
### What does the PR do
By integrating `zxcvbn` module, it has been added a new method to get the specific password strength score information.

Closes [#5096](https://github.com/status-im/status-desktop/issues/5096) together with related `nim-go` and `desktop` PRs.

**NOTE**: Added specific get score method althought there is already `PasswordStrength` one, that provides the same but extended information, due to an issue parsing JSON  response in the _NIM_ site since the `zxcvbn`returns big int crack time numbers when the password lenght is big and nim json cannot parse the response correctly. See the following  [post](https://github.com/nim-lang/Nim/issues/15413).

### DIFF
--- a/mobile/status.go
+++ b/mobile/status.go
@@ -14,6 +14,7 @@ import (
 	signercore "github.com/ethereum/go-ethereum/signer/core"
 
 	"github.com/status-im/zxcvbn-go"
+	"github.com/status-im/zxcvbn-go/scoring"
 
 	"github.com/status-im/status-go/api"
 	"github.com/status-im/status-go/api/multiformat"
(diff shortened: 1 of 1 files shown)


mobile/status.go
```diff
@@ -755,23 +756,53 @@ func ImageServerTLSCert() string {
...
+	if err != nil {
+		return makeJSONResponse(err)
 	}
 
-	err := json.Unmarshal([]byte(paramsJSON), &params)
+	data, err := json.Marshal(zxcvbn.PasswordStrength(requestParams.Password, requestParams.UserInputs))
+	if err != nil {
+		return makeJSONResponse(fmt.Errorf("Error marshalling to json: %v", err))
+	}
+	return string(data)
```
alaibe, line 785: Should it not be makeJSONResponse?

same question for function GetPasswordStrengthScore

alaibe (approved):

mobile/status.go
alaibe: nvm

jrainville (approved):
Looks good.

> due to an issue parsing JSON response in the NIM site since the zxcvbnreturns big int crack time numbers when the password lenght is big and nim json cannot parse the response correctly

I'm not sure if I fully understand this. You mean that in Nim we have issues parsing the response because the Int returned is too big? 
If that's the case, we can use the `stint` library, which is used to parse big numbers.

If we can, having only one function for Strength and Score would be great

noeliaSD:
> Looks good.
> 
> > due to an issue parsing JSON response in the NIM site since the zxcvbnreturns big int crack time numbers when the password lenght is big and nim json cannot parse the response correctly
> 
> I'm not sure if I fully understand this. You mean that in Nim we have issues parsing the response because the Int returned is too big? If that's the case, we can use the `stint` library, which is used to parse big numbers.
> 
> If we can, having only one function for Strength and Score would be great



> Looks good.
> 
> > due to an issue parsing JSON response in the NIM site since the zxcvbnreturns big int crack time numbers when the password lenght is big and nim json cannot parse the response correctly
> 
> I'm not sure if I fully understand this. You mean that in Nim we have issues parsing the response because the Int returned is too big? If that's the case, we can use the `stint` library, which is used to parse big numbers.
> 
> If we can, having only one function for Strength and Score would be great

Let me try to explain it better.. And I've realised the [post](https://github.com/nim-lang/Nim/issues/15413) link was wrong so I can understand why you cannot understand my comment!! 

I firstly created the generic `GetPasswordStrength` method from `status-go` that incorporates the complete information the library we are using provides. It provides  entropy, crack time, score, match sequence... Then I created all the chain in `nim-status-go` and in `status-desktop` passing a `json` with all that information and in some point in `status-desktop` it was got the `score` that now it is just the only info that is of our interest. 
Then, testing some passwords with a considerable length, where for example, its crack time starts to be huge, the json parser in the nim site raised exceptions regarding the integer values received so looking at the internet I saw the above post and seems the nim json library has some limitations so I decided  to create that new method that just only retreives the score and then we can manage it well in nim but keeping both methods still we have the door open in `status-go` to get a better and more complete information.

I hope it is more clear now and it has sense :)

jrainville:
@noeliaSD Aaaah yes it's much clearer now.

I already approved, so here is a thumbs up instead 👍

PR #2606, opened 2022-03-24
Profile picture sync

vitvly:
Related to https://github.com/status-im/status-react/issues/11558


protocol/messenger.go
```diff
@@ -2526,6 +2526,49 @@ func (m *Messenger) ShareImageMessage(request *requests.ShareImageMessage) (*Mes
 	return response, nil
 }
 
+func (m *Messenger) SyncProfilePicture(ctx context.Context, request *requests.SyncProfilePicture) error {
```
Samyoul: This functions seems to be triggered externally from the API, but profile pictures should be synced in response to 2 main user behaviours.

1. When the user first pairs and syncs their device see `func (m *Messenger) SyncDevices`
1. When the user updates their profile picture

If we keep this pattern we won't need to implement separate sync API methods for each item of sync

protocol/messenger.go
```diff
@@ -2526,6 +2526,49 @@ func (m *Messenger) ShareImageMessage(request *requests.ShareImageMessage) (*Mes
...
+	if !m.hasPairedDevices() {
+		return nil
+	}
+
+	if err := request.Validate(); err != nil {
+		return err
+	}
+
+	image, err := m.multiAccounts.GetIdentityImage(request.KeyUID, request.Name)
```
Samyoul: I'd recommend syncing all images at the same time, rather than syncing each type of image per message. This way, the sync functionality will be less likely to miss a new type of identity image associated with the user's key.

protocol/messenger_handler.go
```diff
@@ -418,6 +418,19 @@ func (m *Messenger) HandleSyncInstallationContact(state *ReceivedMessageState, m
 	return nil
 }
 
+func (m *Messenger) HandleSyncProfilePicture(state *ReceivedMessageState, message protobuf.SyncProfilePicture) error {
```
Samyoul: This function should have a clock that comes from `(m *Messenger) getTimesource()`, see usage of `(m *Messenger) getLastClockWithRelatedChat()`.

Once you have the clock value you can compare the time the images were generated against the time of the images in the database, this will mean that only the most recent image changes will be saved on a receiving device.

protocol/protobuf/pairing.proto
```diff
@@ -119,3 +119,13 @@ message SyncClearHistory {
...
+
+message SyncProfilePicture {
+  string key_uid = 1;
+  string name = 2;
+  bytes  payload = 3;
+	uint32 width = 4;
+	uint32 height = 5;      
+	uint32 file_size = 6;
+	uint32 resize_target = 7;
+}
```
Samyoul: You could add the sync clock to the `SyncProfilePicture` protobuf and then batch all your `SyncProfilePicture`s together with something like:

```proto
message SyncProfilePicture {
  string key_uid = 1;
  string name = 2;
  bytes payload = 3;
  uint32 width = 4;
  uint32 height = 5;      
  uint32 file_size = 6;
  uint32 resize_target = 7;
  uint64 clock = 8;
}

message SyncProfilePictures {
  repeated SyncProfilePicture pictures = 1;
}
```

protocol/messenger.go
vitvly: Got it, thanks, fixed.

protocol/protobuf/pairing.proto
vitvly: Done.

protocol/messenger.go
vitvly: Made required changes, thanks for pointing this out.

protocol/messenger_handler.go
vitvly: Added `Clock` field to `identity_images` table.

Samyoul:
Hey @siphiuel, great changes, I can see that identity images now have a clock value, but I wasn't able to spot where the clock is used. For example where is the clock compared and then accepted or rejected?

protocol/messenger_handler.go
```diff
@@ -418,6 +418,37 @@ func (m *Messenger) HandleSyncInstallationContact(state *ReceivedMessageState, m
...
+	}
+	dbImageMap := make(map[string]*images.IdentityImage)
+	for _, img := range dbImages {
+		dbImageMap[img.Name] = img
+	}
+	idImages := make([]*images.IdentityImage, len(message.Pictures))
+	i := 0
+	for _, message := range message.Pictures {
+		dbImg := dbImageMap[message.Name]
+		if dbImg != nil && dbImg.Clock > int(message.Clock) {
```
vitvly: @Samyoul here is the clock comparison

protocol/messenger_sync_profile_picture_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+	s.Require().NoError(err)
+
+	syncedImages, err := theirMessenger.multiAccounts.GetIdentityImages(keyUID)
+	s.Require().NoError(err)
+	s.Require().Equal(2, len(syncedImages))
+
+	// Check that we don't update images with earlier clock values
+
+	for _, img := range iis {
```
vitvly, line 146: @Samyoul and here we test that images with older clock values are discarded

protocol/messenger_handler.go
Samyoul: 🤦‍♂️

Samyoul: I'm so blind, thank you and sorry

Samyoul (approved):
Great, thank you.

multiaccounts/migrations/sql/1648646095_image_clock.down.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE identity_images DROP COLUMN clock;
```
Parveshdhull, line 1: hi @siphiuel, Thank you very much for great PR.
We probably not going to use down migrations, but if we ever use this will not work
https://stackoverflow.com/questions/8442147/how-to-delete-or-add-column-in-sqlite

Parveshdhull: hi @siphiuel, Thank you very much for great PR.
We probably not going to use down migrations, but if we ever use this will not work
https://stackoverflow.com/questions/8442147/how-to-delete-or-add-column-in-sqlite

vitvly: @Parveshdhull thanks, i didn't know that! What would you suggest - removing the `down` script?

Parveshdhull: yes, removing sounds good to me.

vitvly: I've actually added a workaround as already employed by previous scripts - using a temp table.

PR #2607, opened 2022-03-25
Feat/wallet sync rebase

msuess:
rebased version of #2523 for review before force-pushing


vitvly:
Made some changes to the test and added subscribe/watch functionality for account changes. Maybe makes sense to refactor the latter in a separate PR, as there's quite a lot in common between bookmarks/images/accounts.
@Samyoul @cammellos can you please review?

vitvly:
@cammellos rebased with `develop` and 1.18 changes

api/geth_backend.go
```diff
@@ -851,6 +851,7 @@ func (b *GethStatusBackend) CallRPC(inputJSON string) (string, error) {
 // CallPrivateRPC executes public and private RPC requests on node's in-proc RPC server.
 func (b *GethStatusBackend) CallPrivateRPC(inputJSON string) (string, error) {
 	client := b.statusNode.RPCClient()
+	b.log.Info("backend.CallPrivateRPC")
```
cammellos: probably we don't need this in logs ;)

multiaccounts/accounts/database.go
```diff
@@ -70,19 +75,19 @@ func (db Database) Close() error {
...
-	accounts := []Account{}
+	accounts := []*Account{}
 	pubkey := []byte{}
 	for rows.Next() {
-		acc := Account{}
+		acc := &Account{}
 		err := rows.Scan(
 			&acc.Address, &acc.Wallet, &acc.Chat, &acc.Type, &acc.Storage,
-			&pubkey, &acc.Path, &acc.Name, &acc.Emoji, &acc.Color, &acc.Hidden, &acc.DerivedFrom)
+			&pubkey, &acc.Path, &acc.Name, &acc.Emoji, &acc.Color, &acc.Hidden, &acc.DerivedFrom, &acc.Clock)
```
cammellos, line 131: We need to be careful on upgrade, since the value will be Null on existing accounts, generally either you set it in the migration, or you use `sql.NullInt64`

node/status_node_services.go
```diff
@@ -581,6 +581,7 @@ func (b *StatusNode) CallPrivateRPC(inputJSON string) (string, error) {
 		return "", ErrRPCClientUnavailable
 	}
 
+	b.log.Info("### node service callprivaterpc")
```
cammellos: log

services/accounts/accounts.go
```diff
@@ -39,24 +39,25 @@ type DerivedAddress struct {
 	AlreadyCreated bool           `json:"alreadyCreated"`
 }
 
-func (api *API) SaveAccounts(ctx context.Context, accounts []accounts.Account) error {
+func (api *API) SaveAccounts(ctx context.Context, accounts []*accounts.Account) error {
 	log.Info("[AccountsAPI::SaveAccounts]")
-	err := api.db.SaveAccounts(accounts)
+	log.Info("### SaveAccounts")
```
cammellos: log

cammellos (commented):
Looks good, the only thing that I think needs to be looked at is the Null value in clock.

I think I prefer to avoid subscription to account changes and create an endpoint in messenger so that there's no circular dependency between messenger/accounts service, but that's a larger change so we can do (if we want) in a separate pr.
Otherwise looks great!

multiaccounts/accounts/database.go
vitvly: In the migration sql we specify `ADD COLUMN clock INT NOT NULL DEFAULT 0;`. So it should always be 0 on existing accounts. Not sure though.
I've tried changing the `Account.Clock` type to `sql.NullInt64` before noticing the `NOT NULL` constraint and then ran some tests in the `api` suite. Turned out that lines like `db.SaveAccounts([]*accounts.Account{{Address: address}})` begin to fail, because by default `sql.NullInt64` will try to insert a NULL value which is forbidden.

protocol/messenger.go
```diff
@@ -2716,7 +2716,96 @@ func (m *Messenger) SyncDevices(ctx context.Context, ensName, photoPath string)
...
+func (m *Messenger) watchAccountListChanges() {
+	m.logger.Debug("watching account changes")
+
+	channel := m.settings.SubscribeToAccountChanges()
+
+	go func() {
+		for {
+			select {
+			case accounts := <-channel:
+				err := m.syncWallets(accounts)
```
Samyoul, line 2736: Do we need to check if we've seen this account before? I think that wallet api passes a lot of things through this feed, I could be mistaken, but I've seen that the wallet API makes these kind of checks.

protocol/messenger_handler.go
```diff
@@ -1712,3 +1713,61 @@ func (m *Messenger) updateUnviewedCounts(chat *Chat, mentioned bool) {
 		chat.UnviewedMentionsCount++
 	}
 }
+
+func (m *Messenger) HandleSyncWalletAccount(state *ReceivedMessageState, message protobuf.SyncWalletAccounts) error {
```
Samyoul, line 1729: Do we need to also send data to the account changes feed?

Samyoul (approved):
Looks good, just a couple of questions that may have an impact on QA's results.

Samyoul:
>Maybe makes sense to refactor the latter in a separate PR, as there's quite a lot in common between bookmarks/images/accounts.

@siphiuel definitely do it 👍

multiaccounts/accounts/database.go
vitvly: @cammellos do you think `NOT NULL` constraint is enough?

protocol/messenger_handler.go
vitvly: What is that account changes feed? What is it used for? (pardon my ignorance)

protocol/messenger.go
vitvly: Thanks, this is interesting. So this is invoked only from `SaveAccounts`/`DeleteAccounts`, its only purpose is send a list of changed accounts to other paired devices. We might repeatedly see the same account if it e.g. was deleted and then re-added again. What negative consequences might arise from this?

services/accounts/accounts.go
vitvly: Removed.

node/status_node_services.go
vitvly: Removed.

api/geth_backend.go
vitvly: Removed.

multiaccounts/accounts/database.go
cammellos: I can't remember what's the behavior for INT, maybe that works fine, it would have to be tested (create account with previous version, upgrade app, login).
Regarding nullInt:
Only when scanning you would use `sql.NullInt64` and then assign if not null to `Account.Clock` similarly to https://github.com/status-im/status-go/blob/997708bc82c68c842512700ee4a6ccdb04411f94/protocol/persistence.go#L276

churik:
Tested in https://github.com/status-im/status-react/pull/13308
@siphiuel is it possible to test the future itself on the status-desktop side or for desktop only regression is worth checking?

vitvly:
@churik this will need implementation on Desktop side in order to handle status-go responses containing synced wallet accounts. What is the proper procedure for this?

alaibe:
Thanks for enabling this sync, I have a question regarding the result, if I import an account with private key, it is also added to the keystore, I believe this is not sync? And how will I be able to make transaction with the account?

The question is also valid for account imported with seed phrase

vitvly:
@alaibe sync for other types of accounts is in progress, this PR is for watch-only addresses (i hope this answers your questions, not sure if i understood them properly).

alaibe:
@siphiuel totally, thanks for clarifying, I wasn t sure it was for watch only addresses

PR #2610, opened 2022-03-25
fix(chat-api): fix getChats functions

jrainville:
The getChats had some issues where it errored or returned bad values. Some small fixes.

### DIFF
--- a/services/chat/api.go
+++ b/services/chat/api.go
@@ -171,7 +171,7 @@ func (api *API) GetChats(ctx context.Context) (map[string]ChannelGroup, error) {
 		}
 
 		for _, chat := range channels {
-			if chat.CommunityID == community.IDString() {
+			if chat.CommunityID == community.IDString() && chat.Active {
 				c, err := api.toAPIChat(chat, community, pubKey)
 				if err != nil {
 					return nil, err
@@ -285,10 +285,6 @@ func (api *API) toAPIChat(protocolChat *protocol.Chat, community *communities.Co
 	}
 	chat.Members = chatMembers
 
-	if chat.CommunityID == "" {
-		chat.CommunityID = pubKey
-	}
-
 	return chat, nil
 }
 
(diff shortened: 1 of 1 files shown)


services/chat/api.go
```diff
@@ -285,10 +285,6 @@ func (api *API) toAPIChat(protocolChat *protocol.Chat, community *communities.Co
 	}
 	chat.Members = chatMembers
 
-	if chat.CommunityID == "" {
-		chat.CommunityID = pubKey
-	}
```
jrainville, line 290: Removed this because even though it makes sense that personal chanels are in the user's "community", having it named `communityId` can cause confusion, as well as causes some breaking changes where some code can look at `communityId` to see if a channel is inside a community

services/chat/api.go
```diff
@@ -340,7 +336,8 @@ func (chat *Chat) populateCommunityFields(community *communities.Community) erro
 
 	commChat, exists := community.Chats()[chat.ID]
 	if !exists {
-		return ErrChatNotFound
+		// Skip unknown community chats. They might be channels that were deleted
+		return nil
```
jrainville, line 340: Just explaining again that somehow, the Chats we got from the DB contained old channels that were deleted from the Community, so this condition retuned an error.
I'm not sure if it's a bug that the DB contains old channel, but skipping them does the trick anyway.

services/chat/api.go
```diff
@@ -353,7 +350,6 @@ func (chat *Chat) populateCommunityFields(community *communities.Community) erro
 	chat.Permissions = commChat.Permissions
 	chat.Emoji = commChat.Identity.Emoji
 	chat.Name = commChat.Identity.DisplayName
-	chat.Color = commChat.Identity.Color
```
jrainville, line 356: The color is already handled by the Chat color coming from the previous call. Setting it here again sometimes came up with an empty color.

alaibe (approved):

PR #2611, opened 2022-03-27
feat: add IPFS rate limiter for downloading stickers and use http server for retrieving stickers

richard-ramos:


nodecfg/node_config.go
```diff
@@ -35,7 +35,7 @@ func nodeConfigWasMigrated(tx *sql.Tx) (migrated bool, err error) {
 
 type insertFn func(tx *sql.Tx, c *params.NodeConfig) error
 
-func insertNodeConfig(tx *sql.Tx, c *params.NodeConfig) error {
+func upgradeNodeConfig(tx *sql.Tx, c *params.NodeConfig) error {
```
alaibe: suggestion: upgradeNodeConfig => upsertNodeConfig

server/server.go
```diff
@@ -95,16 +98,27 @@ func PublicTLSCert() (string, error) {
...
 type imageHandler struct {
 	db     *sql.DB
-	logger *zap.Logger
+	logger log.Logger
 }
 
 type audioHandler struct {
 	db     *sql.DB
-	logger *zap.Logger
+	logger log.Logger
```
alaibe: why not using zap? I think go-ethereum log use zap under the hood anyway

alaibe (approved):
Nice

appdatabase/migrations/sql/1648333113_ipfs_dir.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE node_config ADD COLUMN ipfs_dir TEXT NOT NULL DEFAULT "";
+UPDATE node_config SET ipfs_dir = "./ipfs";
```
flexsurfer: is it default path ? do you use it in desktop ?

server/server.go
```diff
@@ -189,23 +203,79 @@ func (s *audioHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	content, err := s.downloader.Get(stickerHash)
+	if err != nil {
+		s.logger.Error("could not download hash", zap.Error(err))
+		return
+	}
+
+	w.Header().Set("Cache-Control", "max-age:290304000, public")
+	w.Header().Set("Expires", time.Now().AddDate(60, 0, 0).Format(http.TimeFormat))
```
flexsurfer: 60 years ?

appdatabase/migrations/sql/1648333113_ipfs_dir.up.sql
flexsurfer: @cammellos do you think we could use default path ? or we should set in cljs ?

server/server.go
```diff
@@ -245,6 +315,9 @@ func (s *Server) Start() error {
 	handler.Handle("/messages/images", &imageHandler{db: s.db, logger: s.logger})
 	handler.Handle("/messages/audio", &audioHandler{db: s.db, logger: s.logger})
 	handler.Handle("/messages/identicons", &identiconHandler{logger: s.logger})
+	handler.Handle("/messages/sticker", &stickerHandler{db: s.db, logger: s.logger, downloader: s.downloader})
+	handler.Handle("/ipfs", &ipfsHandler{logger: s.logger, downloader: s.downloader})
```
flexsurfer, line 251: how will ipfs be used?

flexsurfer:
Great job!

server/server.go
```diff
@@ -189,23 +203,79 @@ func (s *audioHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
+func (s *stickerHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	messageIDs, ok := r.URL.Query()["messageId"]
+	if !ok || len(messageIDs) == 0 {
+		s.logger.Error("no messageID")
+		return
+	}
+	messageID := messageIDs[0]
+	var stickerHash string
+	err := s.db.QueryRow(`SELECT sticker_hash FROM user_messages WHERE id = ?`, messageID).Scan(&stickerHash)
```
flexsurfer: i think there is no need to "hide" hash over messageid for stickers, we could add hash right in the sticker URL, so in that case we don't need to query it from db

protocol/common/message.go
```diff
@@ -176,12 +178,16 @@ func (m *Message) PrepareServerURLs(port int) {
 	if m.ContentType == protobuf.ChatMessage_AUDIO {
 		m.AudioLocalURL = fmt.Sprintf("https://localhost:%d/messages/audio?messageId=%s", port, m.ID)
 	}
+	if m.ContentType == protobuf.ChatMessage_STICKER {
+		m.StickerLocalURL = fmt.Sprintf("https://localhost:%d/messages?sticker?messageId=%s", port, m.ID)
```
flexsurfer: better to use hash directly here instead of messageId

appdatabase/migrations/sql/1648333113_ipfs_dir.up.sql
cammellos: If we don't need to set it, better I suppose, maybe we can get away without storing anything in the database and have a sensible default for both projects?
If we do have to set it, for now we have to do as above, since how we start the node pulls it from the database, but ideally we don't store anything in the db (we can always change that at a later stage, if we have to)

protocol/common/message.go
flexsurfer: also `messages?sticker` should be `messages/sticker`

cammellos: URL also looks a bit funny, double `?`, or am I missing something?

server/server.go
richard-ramos: http://localhost:1234/ipfs/hash=sticker_hash_goes_here

protocol/common/message.go
richard-ramos: yeah, it's a typo :D

flexsurfer:
pushed commit replacing messageid with hash

server/server.go
flexsurfer: i mean, will it be used from client somehow ?

server/server.go
richard-ramos: yes. In practice it means that the resource will never expire, and kept in browser cache

server/server.go
flexsurfer: i don't see where we can use it in client,just wondering why do wee need it

appdatabase/migrations/sql/1648333113_ipfs_dir.up.sql
richard-ramos: I can change it to a default value instead of using the DB. The reason I'm using the node_config tables is so it matches the way we store the datadir and keystore dirs. Do you prefer I use a default path instead?

cammellos: In general, I tend to favor not storing state unless necessary, it's less committal and it's easier to modify etc, but I don't have a strong opinion on the matter, so up to you

server/server.go
richard-ramos: It's being used in `stickers_market` in the preview and thumbnail attributes. I'm going to remove  /messages/ipfs since as it was pointed out in other comments, we could access the hash directly so there's no need of using the messageId

appdatabase/migrations/sql/1648333113_ipfs_dir.up.sql
richard-ramos: Will remove it then!

flexsurfer:
for some reason with latest changes `stickers_market` return value is nil

flexsurfer:
```
t=2022-03-29T08:06:30+0000 lvl=warn msg="Could not retrieve stickerpack data" packID=0 error="open /Users/andrey/Library/Developer/CoreSimulator/Devices/161B6748-1FF5-456D-BD2B-BD3E00A56C26/data/Containers/Data/Application/D98A1FF4-5681-4185-A7E3-5895E21326E0/Library/ipfs/bafybeidzeerdxo4ygldu3kje3vb5tojoapmsydhqlermb6o7jvxw5k7fhq: permission denied"

```

services/stickers/transactions.go
```diff
@@ -84,63 +85,99 @@ func (api *API) Buy(ctx context.Context, chainID uint64, txArgs transactions.Sen
...
+	return ethereum.CallMsg{
 		From:  common.Address(from),
 		To:    &sntAddress,
 		Value: big.NewInt(0),
 		Data:  data,
-	})
+	}, nil
+}
+
+func (api *API) BuyPrepareTx(ctx context.Context, chainID uint64, from types.Address, packID *bigint.BigInt) (interface{}, error) {
```
flexsurfer, line 140: please review my changes, i've added this method, because we sign transaction in status-react

flexsurfer (approved):

params/config.go
```diff
@@ -313,6 +313,8 @@ type NodeConfig struct {
 	// NetworkID sets network to use for selecting peers to connect to
 	NetworkID uint64 `json:"NetworkId" validate:"required"`
 
+	RootDataDir string `json:"-"`
+
```
Samyoul, line 317: @richard-ramos I know this is a very old PR, but do you remember why you added `RootDataDir` with a `json:"-"` tag? Was it a security reason or something else?

richard-ramos: Sadly, the reasons for this field not being exported when marshalled were lost in the mists of time

PR #2618, opened 2022-03-31
fix: Add derivation path to wallet account generation

Khushboo-dev-cpp:
Added new endpoints for 
1. Importing a wallet account with a give path 
2. Generating wallet accounts given the path and index to be generated

Important changes:


ghost:
### Pull Request Checklist

account/generator/generator.go
```diff
@@ -275,8 +275,6 @@ func (g *Generator) store(acc *Account, password string) (AccountInfo, error) {
 		}
 	}
 
-	g.Reset()
```
alaibe, line 278: Is that on purpose?

appdatabase/migrations/sql/1648826780_add_derivedfrom_accounts.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE accounts ADD COLUMN derivedfrom TEXT NOT NULL DEFAULT "";
+UPDATE accounts SET derivedfrom = "";
```
alaibe: Is it a breaking change?

multiaccounts/accounts/database.go
```diff
@@ -16,17 +16,18 @@ const (
...
+	Chat        bool           `json:"chat"`
+	Type        string         `json:"type,omitempty"`
+	Storage     string         `json:"storage,omitempty"`
+	Path        string         `json:"path,omitempty"`
+	PublicKey   types.HexBytes `json:"public-key,omitempty"`
+	Name        string         `json:"name"`
+	Emoji       string         `json:"emoji"`
+	Color       string         `json:"color"`
+	Hidden      bool           `json:"hidden"`
+	DerivedFrom string         `json:"derivedfrom,omitempty"`
```
alaibe: nit: json: derived-from to keep existing syntax

appdatabase/migrations/sql/1648826780_add_derivedfrom_accounts.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE accounts ADD COLUMN derivedfrom TEXT NOT NULL DEFAULT "";
```
alaibe: derivedfrom => derived_from

services/accounts/accounts.go
```diff
@@ -31,6 +31,12 @@ type API struct {
 	feed    *event.Feed
 }
 
+type Address struct {
```
alaibe: could we avoid: `Address.Address`

Can the struct be name with something more explicit: `AddressAndPath`

services/accounts/accounts.go
```diff
@@ -97,6 +103,50 @@ func (api *API) AddAccountWithMnemonic(
 	return api.SaveAccounts(ctx, []accounts.Account{account})
 }
 
+func (api *API) AddAccountWithMnemonicAndPath(
```
alaibe, line 81: Can u update the addAccountWithMnemonic to call this function?
That way we avoid code duplication

services/accounts/accounts.go
```diff
@@ -189,6 +239,49 @@ func (api *API) GenerateAccount(
 	return api.SaveAccounts(ctx, []accounts.Account{acc})
 }
 
+func (api *API) GenerateAccountWithDerivedPath(
```
alaibe, line 171: Same, can you call this function from generateAccount? to avoid code duplication

services/accounts/accounts.go
```diff
@@ -197,3 +290,73 @@ func (api *API) verifyPassword(password string) error {
...
+
+func (api *API) GetAddressListForPath(password string, derivedFrom string, path string, pageSize int, pageNumber int) ([]*Address, error) {
+	derivedAddresses := make([]*Address, 0)
+
+	info, err := api.manager.AccountsGenerator().LoadAccount(derivedFrom, password)
+	if err != nil {
+		return nil, err
+	}
+
+	if pageNumber <= 0 || pageSize <= 0 {
```
alaibe, line 215: there is no max value for page size?

services/accounts/accounts.go
```diff
@@ -197,3 +290,73 @@ func (api *API) verifyPassword(password string) error {
...
+		return nil, fmt.Errorf("pageSize and pageNumber should be greater than 0")
+	}
+
+	var startIndex = (pageNumber - 1) * pageSize
+	var endIndex = pageNumber * pageSize
+
+	for i := startIndex; i < endIndex; i++ {
+		derivedPath := fmt.Sprint(path, "/", i)
+
+		infos, err := api.manager.AccountsGenerator().DeriveAddresses(info.ID, []string{derivedPath})
```
alaibe: not sure of the speed of that but maybe we should use a coroutine here to make things faster?

services/accounts/accounts.go
```diff
@@ -197,3 +290,73 @@ func (api *API) verifyPassword(password string) error {
...
+		if err != nil {
+			return nil, err
+		}
+
+		address := &Address{
+			Address:     common.HexToAddress(infos[derivedPath].Address),
+			Path:        derivedPath,
+			HasActivity: false,
```
alaibe, line 233: What is this attribute? it is always false

services/accounts/accounts.go
```diff
@@ -197,3 +290,73 @@ func (api *API) verifyPassword(password string) error {
 	_, err = api.manager.VerifyAccountPassword(api.config.KeyStoreDir, address.Hex(), password)
 	return err
 }
+
+func (api *API) GetAddressListForPath(password string, derivedFrom string, path string, pageSize int, pageNumber int) ([]*Address, error) {
```
alaibe: Maybe GetDerivedAddressesForPath

services/accounts/accounts.go
alaibe: After reading the code, maybe `DerivedAddress`

services/accounts/accounts.go
```diff
@@ -197,3 +290,73 @@ func (api *API) verifyPassword(password string) error {
...
+			HasActivity: false,
+		}
+
+		derivedAddresses = append(derivedAddresses, address)
+	}
+
+	return derivedAddresses, nil
+}
+
+func (api *API) GetAddressListForMenominicWithPath(mnemonic string, path string, pageSize int, pageNumber int) ([]*Address, error) {
```
alaibe: Same suggestion:
GetDerivedAddressesForMnemonicAndPath

Also this function seems to duplicate code

account/generator/generator.go
Khushboo-dev-cpp: Yes so I need it cause, right now from the nim layer we call saveAccount and if its deleted I am not able to save it. Also I did not understand why this map needs to be cleared after every save and did not see any issue removing this.

services/accounts/accounts.go
Khushboo-dev-cpp: sure. Will create a new func for that

services/accounts/accounts.go
Khushboo-dev-cpp: I think there can be infinite number of pages. as in HD wallets. the UI can decide how many pages it wants to show?

services/accounts/accounts.go
Khushboo-dev-cpp: yes can do, its quite quick right now from what I see though

services/accounts/accounts.go
Khushboo-dev-cpp: yes, its not implemented yet. it tells if there has been an activity on that address and will be shown on the UI

Khushboo-dev-cpp:
@alaibe all done :)

alaibe (approved):

PR #2619, opened 2022-04-01
feat: get wallet api

alaibe:
Build a new endpoint for the wallet which include all necessary data to display the wallet

### DIFF
--- a/services/wallet/api.go
+++ b/services/wallet/api.go
@@ -12,12 +12,22 @@ import (
 )
 
 func NewAPI(s *Service) *API {
-	return &API{s}
+	r := NewReader(s)
+	return &API{s, r}
 }
 
 // API is class with methods available over RPC.
 type API struct {
 	s *Service
+	r *Reader
+}
+
+func (api *API) StartWallet(ctx context.Context, chainIDs []uint64) error {
+	return api.r.Start(ctx, chainIDs)
+}
+
+func (api *API) GetWallet(ctx context.Context, chainIDs []uint64) (*Wallet, error) {
+	return api.r.GetWallet(ctx, chainIDs)
 }
 
 // SetInitialBlocksRange sets initial blocks range
@@ -139,7 +149,7 @@ func (api *API) DeleteCustomTokenByChainID(ctx context.Context, chainID uint64, 
 	return err
 }
 
-func (api *API) GetSavedAddresses(ctx context.Context) ([]*SavedAddress, error) {
+func (api *API) GetSavedAddresses(ctx context.Context) ([]SavedAddress, error) {
 	log.Debug("call to get saved addresses")
 	rst, err := api.s.savedAddressesManager.GetSavedAddresses(api.s.rpcClient.UpstreamChainID)
 	log.Debug("result from database for saved addresses", "len", len(rst))
@@ -231,7 +241,7 @@ func (api *API) WatchTransactionByChainID(ctx context.Context, chainID uint64, t
 	return api.s.transactionManager.watch(ctx, transactionHash, chainClient)
 }
 
-func (api *API) GetFavourites(ctx context.Context) ([]*Favourite, error) {
+func (api *API) GetFavourites(ctx context.Context) ([]Favourite, error) {
 	log.Debug("call to get favourites")
 	rst, err := api.s.favouriteManager.GetFavourites()
 	log.Debug("result from database for favourites", "len", len(rst))
--- a/services/wallet/reader.go
+++ b/services/wallet/reader.go
@@ -1,0 +1,226 @@
+package wallet
+
+import (
+	"context"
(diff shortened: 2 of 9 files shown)


services/wallet/reader.go
```diff
@@ -0,0 +1,238 @@
...
+
+type ReaderAccount struct {
+	Account      *accounts.Account               `json:"account"`
+	Collections  map[uint64][]*OpenseaCollection `json:"collections"`
+	Tokens       map[uint64][]*ReaderToken       `json:"tokens"`
+	Transactions map[uint64][]*transfer.View     `json:"transactions"`
+}
+
+type Wallet struct {
+	Accounts            []*ReaderAccount                 `json:"Accounts"`
```
richard-ramos:
```suggestion
	Accounts            []*ReaderAccount                 `json:"accounts"`
```

Ignore this suggestion if this is already used by mobile

services/wallet/reader.go
```diff
@@ -0,0 +1,238 @@
...
+	for _, chainID := range chainIDs {
+		client, err := newOpenseaClient(chainID, r.s.openseaAPIKey)
+		if err == nil {
+			c, _ := client.fetchAllCollectionsByOwner(common.Address(account.Address))
+			collections[chainID] = c
+		}
+
+		for _, token := range visibleTokens {
+			oraclePrice := prices[token.Symbol]
+			cryptoBalance := balances[token.Address].ToInt()
```
richard-ramos, line 83: Could it be possible that the balance for a token does not exist in the map, and we run into a null exception when calling `.ToInt()`?

alaibe: That shouldn't happen, at least I don t see how, as we are fetching the balances for all addresses/all visible tokens. So should be at least 0

Khushboo-dev-cpp (approved):
Looks really really good to me! excited to see how the calls are super simplified on the BE side :)

services/wallet/api.go
```diff
@@ -223,11 +239,11 @@ func (api *API) AddFavourite(ctx context.Context, favourite Favourite) error {
 	return err
 }
 
-func (api *API) GetCryptoOnRamps(ctx context.Context) ([]CryptoOnRamp, error) {
+func (api *API) GetCryptoOnRamps(ctx context.Context) ([]*CryptoOnRamp, error) {
```
Samyoul: I'm curious, what benefit do we get from passing a slice of references over a slice of values?

services/wallet/api.go
```diff
@@ -6,18 +6,34 @@ import (
...
 	"github.com/ethereum/go-ethereum/log"
+	"github.com/status-im/status-go/multiaccounts/accounts"
 	"github.com/status-im/status-go/params"
 	"github.com/status-im/status-go/services/wallet/chain"
 	"github.com/status-im/status-go/services/wallet/transfer"
 )
 
 func NewAPI(s *Service) *API {
-	return &API{s}
+	accountDb, _ := accounts.NewDB(s.db)
```
Samyoul: Why are we ignoring the error here?

services/wallet/service.go
```diff
@@ -38,6 +39,7 @@ func NewService(db *sql.DB, rpcClient *rpc.Client, accountFeed *event.Feed, open
 
 // Service is a wallet service.
 type Service struct {
+	db                    *sql.DB
```
Samyoul, line 44: Service accepts a `*sql.DB`, but then only uses it to generate an `accounts.Database`, perhaps `Service` should accept just the `accounts.Database`

Samyoul (changes requested):

services/wallet/api.go
alaibe: Updating by using the accountsdb given by the service

services/wallet/service.go
alaibe: It doesn t only use for generating an accounts.DB but I pass specifically the accounts db being build at the layer above

services/wallet/api.go
alaibe: Isn t it better to pass the pointer rather than copy the whole CryptoOnRamp?

Samyoul: No, a slice index is a pointer effectively. A slice pointer value is much slower.

richard-ramos (approved):

services/wallet/api.go
alaibe: TIL, I will update the PR accordingly, thanks

alaibe:
@Samyoul PR has been updated

Samyoul (approved):
All looks great, thank you. Also have a little look at this article https://medium.com/@philpearl/bad-go-slices-of-pointers-ed3c06b8bb41 . It has got some benchmark examples of how a slice of pointer values impacts performance.

PR #2506, opened 2022-01-27
Local Sync Pairing

Samyoul:
## Pairing Events

This PR introduces endpoints for triggering the `PairingServer` and `PairingClient`.

This PR also introduces signals and signal events for:

- `EventConnectionError`
- `EventConnectionSuccess`
- `EventTransferError`
- `EventTransferSuccess`
- `EventProcessSuccess`
- `EventProcessError`

## Pairing API Endpoints

I've added the following API endpoints:

### GetConnectionStringForBeingBootstrapped 

`GetConnectionStringForBeingBootstrapped` starts a `server.Receiving` `server.PairingServer` then generates a `server.ConnectionParams`. Used when the device is Logged out or has no Account keys and the device has no camera to read a QR code with

**Example:** A desktop device (device without camera) receiving account data from mobile (device with camera)
`func GetConnectionStringForBeingBootstrapped(configJSON string)`

### GetConnectionStringForBootstrappingAnotherDevice

`GetConnectionStringForBootstrappingAnotherDevice` starts a `server.Sending` `server.PairingServer` then generates a `server.ConnectionParams`. Used when the device is logged in and therefore has `Account` keys and the device might not have a camera.

*Example:* A mobile or desktop device (devices that MAY have a camera but **MUST** have a screen) sending account data to a mobile (device with camera)
`func GetConnectionStringForBootstrappingAnotherDevice(configJSON string)`

### InputConnectionStringForBootstrapping

`InputConnectionStringForBootstrapping` starts a `server.PairingClient`. The given `server.ConnectionParams` string will determine the `server.Mode`:

`server.Mode` = `server.Sending` Used when the device is logged in and therefore has `Account` keys and the has a camera to read a QR code

**Example:** A mobile (device with camera) sending account data to a desktop device (device without camera)

`server.Mode` = `server.Receiving`
Used when the device is logged out or has no `Account` keys and has a camera to read a QR code

**Example:** A mobile device (device with a camera) receiving account data from a device with a screen (mobile or desktop devices)
`func InputConnectionStringForBootstrapping(cs, configJSON string)`

### PairingPayloadSourceConfig

The `configJSON` string given as the above API's input param(s) is a JSON serialised string representation of `PairingPayloadSourceConfig`. `PairingPayloadSourceConfig` represents location and access data of the pairing payload ONLY available from the application client
```go
type PairingPayloadSourceConfig struct {
	KeystorePath string `json:"keystorePath"`
	KeyUID       string `json:"keyUID"`
[...]

PR #2626, opened 2022-04-05
Get preferred network IP and refactor server package to increase reusability 

Samyoul:
### What's New?

- New `GetOutboundIP` function that gets the device's preferred network IP (even works in CI! 🚀 )
  - This is needed for:
    - Generation of TLS certs
    - Creating a TLS listener for network traffic
    - Sharing the network address of the TLS server with other devices
- Restructured TLS certs to be IP address centric, moving away from domain centric
- The TLS media server now hosts on `127.0.0.1`, as `localhost` resolves to `127.0.0.1` any way
- Made all URL generation strictly a `Server` concern
  - No longer need to export `Server.Port`, or assume the server's address outside of the `Server`
  - All consuming code can now be maintained from a single package
- Removed `Server.Port` and replaced with `Server.listener net.Listener`
  - Because the port number is always chosen by the active `net.Listener`, this change just always asks the listener for the port number. If the port number needs to change we change the listener first.
- Implemented `GenerateCertFromKey` for deterministic certs
  - Useful for creating the exact same cert from only a private key
  - Required for `RootCA` whitelisting, which we need.
- Moved server handlers into a modular and extensible format
  - Replaced old Handler structs with simple `HandlerFuncs`
  - Implemented a function `WithHandlers` which allows for multiple sequential calls
- Full TLS server e2e testing via network 👨‍🍳💋
- `Server` config to customise the server

### Note for QA
Potential regression on identicon, image, audio and stickers showing, particularly on iOS. I've tested with emulator and it all works.

---

Required for https://github.com/status-im/status-go/pull/2506

---

TODO
- I would like to remove the use of `globalCertificate *tls.Certificate = nil` and `globalPem string`
  - it seems only useful for `PublicTLSCert`, and globals makes connected functionality hard to trace.


protocol/common/message.go
```diff
@@ -164,17 +161,17 @@ type Message struct {
...
-		m.QuotedMessage.PrepareImageURL(port)
+		m.QuotedMessage.ImageLocalURL = s.MakeImageURL(m.QuotedMessage.ID)
 	}
 	if m.ContentType == protobuf.ChatMessage_IMAGE {
-		m.ImageLocalURL = fmt.Sprintf("https://localhost:%d/messages/images?messageId=%s", port, m.ID)
+		m.ImageLocalURL = s.MakeImageURL(m.ID)
 	}
 	if m.ContentType == protobuf.ChatMessage_AUDIO {
-		m.AudioLocalURL = fmt.Sprintf("https://localhost:%d/messages/audio?messageId=%s", port, m.ID)
+		m.AudioLocalURL = s.MakeAudioURL(m.ID)
```
Samyoul: Making URLs is `Server`'s concern, `Message` doesn't care.

protocol/messenger.go
```diff
@@ -390,7 +390,8 @@ func NewMessenger(
 	}
 
 	mailservers := mailserversDB.NewDB(database)
-	httpServer, err := server.NewServer(database, logger)
+	mediaServer, err := server.NewServer(database, logger)
+	mediaServer.WithMediaHandlers()
```
Samyoul: Handlers need to be attached depending on the server being inited

protocol/messenger_images.go
```diff
@@ -1,7 +1,5 @@
 package protocol
 
-import "fmt"
-
 func (m *Messenger) ImageServerURL() string {
-	return fmt.Sprintf("https://localhost:%d/messages/", m.httpServer.Port)
+	return m.httpServer.MakeImageServerURL()
```
Samyoul, line 4: URLs are the concern of `Server`, `Messenger` doesn't want to know how they were made

server/handlers.go
```diff
@@ -0,0 +1,114 @@
...
+	"go.uber.org/zap"
+
+	"github.com/status-im/status-go/protocol/identity/identicon"
+	"github.com/status-im/status-go/protocol/images"
+)
+
+type HandlerPatternMap map[string]http.HandlerFunc
+
+func handleIdenticon(logger *zap.Logger) func(w http.ResponseWriter, r *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
```
Samyoul, line 26: The three media handlers were originally creating structs that took params, these new `HandlerFunc` funcs take params and return handlers. Less overhead, more extensible, easier to implement

Samyoul:
Aww yeah! 🥳 🎉 

```
[2022-04-17T20:13:58.089Z] === RUN   Test_makeSerialNumberFromKey
[2022-04-17T20:13:58.089Z] --- PASS: Test_makeSerialNumberFromKey (0.00s)
[2022-04-17T20:13:58.089Z] === RUN   TestGetOutboundIPWithFullServerE2e
[2022-04-17T20:13:58.089Z] (string) (len=12) "135.181.1.48"
[2022-04-17T20:13:58.089Z] --- PASS: TestGetOutboundIPWithFullServerE2e (0.01s)
[2022-04-17T20:13:58.089Z] === RUN   TestServer_MakeBaseURL
[2022-04-17T20:13:58.089Z] --- PASS: TestServer_MakeBaseURL (0.00s)
[2022-04-17T20:13:58.089Z] === RUN   TestServer_MakeImageServerURL
[2022-04-17T20:13:58.089Z] --- PASS: TestServer_MakeImageServerURL (0.00s)
[2022-04-17T20:13:58.089Z] === RUN   TestServer_MakeIdenticonURL
[2022-04-17T20:13:58.089Z] --- PASS: TestServer_MakeIdenticonURL (0.00s)
[2022-04-17T20:13:58.089Z] === RUN   TestServer_MakeImageURL
[2022-04-17T20:13:58.089Z] --- PASS: TestServer_MakeImageURL (0.00s)
[2022-04-17T20:13:58.089Z] === RUN   TestServer_MakeAudioURL
[2022-04-17T20:13:58.089Z] --- PASS: TestServer_MakeAudioURL (0.00s)
[2022-04-17T20:13:58.089Z] PASS
[2022-04-17T20:13:58.089Z] ok  	github.com/status-im/status-go/server	0.028s
```

server/server.go
```diff
@@ -2,202 +2,86 @@ package server
...
 	// in case of restart, we should use the same port as the first start in order not to break existing links
-	addr := fmt.Sprintf("localhost:%d", s.Port)
+	addr := fmt.Sprintf("%s:%d", s.netIP, s.port)
 
 	listener, err := tls.Listen("tcp", addr, cfg)
 	if err != nil {
 		s.logger.Error("failed to start server, retrying", zap.Error(err))
-		s.Port = 0
+		s.port = 0 //TODO find out why the port is set to 0 here
 		err = s.Start()
```
bitgamma, line 43: because this error happens when something is already listening on that port. Setting to 0 takes another random port. This happens on iOS when the app is put in background the port is not released for a long time but the socket is closed, so you would enter an infinite loop trying to retake the previous port.

Samyoul: Thank you very much. That is helpful to know, I'll change my code to handle randomly selected port numbers rather than hardcoding a specific port.

Samyoul: @bitgamma I've change the way `Server` works, we no longer use `port` instead we rely on having an active `net.Listener`. This method is explicit that the listener should always choose the port number, we don't need to reset the port number, we just reset the listener. What do you think?

bitgamma: I am ooo (vacation) so I cannot review in details, but in general make sure that the transition from fg/bg and back on both Android and iOS is not broken by this, cause they both have quirks. Also, is there the risk that if you leave an image or a chat open and lock the screen, when you resume the images will all be broken since the port changed. The reason why I used a random port but then tried to keep it was to avoid broken links

Samyoul: Ah sorry, I didn't realise you were on holiday.

> The reason why I used a random port but then tried to keep it was to avoid broken links

Yes, this is what I've done basically. I really like the idea of the `net.Listener` choosing the port number. So instead of storing the port number on the `Server`, I've replaced that with storing the listener itself, as the listener is always the source of the port number.

Samyoul:
👍 full e2e server / client test with UDP selected IP and Listener selected port number
```
[2022-04-19T10:10:53.843Z] === RUN   TestGetOutboundIPWithFullServerE2e
[2022-04-19T10:10:53.843Z] (string) (len=26) "https://135.181.5.32:37389"
[2022-04-19T10:10:53.843Z] --- PASS: TestGetOutboundIPWithFullServerE2e (0.12s)
```

server/server.go
Samyoul: What I'll do is write some tests to check fg/bg transitioning doesn't produce any errors. My standard e2e test works fine, so that's a good staring point.

bitgamma: ok, but e2e is not going to be enough, especially on iOS sockets are silently killed in some cases. This really needs to be tested on hw

bitgamma (approved):

churik:
@Samyoul move it to `Ready for client testing` when it is time to verify it on the client-side. Also if there is any possible regression from your POV, mentioning that in the description would be highly appreciated. Thanks!

pavloburykh:
@Samyoul thanx for the PR. Please, take a look at following issue

### ISSUE 1 Images, audio and stickers are not shown both on IOS and Android

**Steps:**
1) Login/register multiaccount
2) Start a chat
3) Send image/sticker/audio message 
4) Observe the result

**Actual result:**
Image/sticker messages are not shown, audio messages can't be played neither for sender nor for receiver. Same issue with images send in timeline.

**Expected result:**
Image/sticker messages are shown, audio messages can be played. 

Tested on builds from status-react PR that was created with status-go "version": "feature/lan_connection_string"
https://github.com/status-im/status-react/pull/13391

logs from Android attached. Bug is reproducible for both IOS and Android
[geth.log](https://github.com/status-im/status-go/files/8755485/geth.log)


https://user-images.githubusercontent.com/97245802/169845163-4791c421-da3c-4d9a-809c-263165022a34.mp4

PR #2630, opened 2022-04-07
Populate colorHash and colorId

osmaczko:
required by: https://github.com/status-im/status-desktop/pull/5424

accounts displayed to users on login page should display colorHash and
avatar fallback color (aka colorId)

### DIFF
--- a/mobile/status.go
+++ b/mobile/status.go
@@ -278,6 +278,18 @@ func SaveAccountAndLogin(accountData, password, settingsJSON, configJSON, subacc
 	if err != nil {
 		return makeJSONResponse(err)
 	}
+
+	for _, acc := range subaccs {
+		if acc.Chat {
+			colorHash, _ := colorhash.GenerateFor(string(acc.PublicKey.Bytes()))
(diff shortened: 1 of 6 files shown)


mobile/status.go
```diff
@@ -277,6 +278,18 @@ func SaveAccountAndLogin(accountData, password, settingsJSON, configJSON, subacc
 	if err != nil {
 		return makeJSONResponse(err)
 	}
+
+	for _, acc := range subaccs {
+		if acc.Chat {
+			colorHash, _ := colorhash.GenerateFor(string(acc.PublicKey.Bytes()))
+			colorID, _ := identityUtils.ToColorID(string(acc.PublicKey.Bytes()))
```
richard-ramos, line 285: This is not bad per se, but i wonder if it would be a good idea to log an error in case any of these functions fail?

richard-ramos (approved):

iurimatias (approved):
tested in https://github.com/status-im/status-desktop/pull/5424#pullrequestreview-936791776

PR #2631, opened 2022-04-08
Enable history archive support for all non-admin communities

0x-r4bbit:
Previously, we've turned archive support off by default, this means,
community member nodes won't try to download torrent data when they
receive a magnetlink.

This commit ensures that communities joined in the future will have
this setting on, and also all existing communities that the node
isn't admin of will have this feature enabled.

### DIFF
--- a/protocol/messenger.go
+++ b/protocol/messenger.go
@@ -1273,18 +1273,35 @@ func (m *Messenger) Init() error {
 			logger.Warn("failed to check if community settings exist", zap.Error(err))
 			continue
 		}
-		if exists {
+
+		if !exists {
+			communitySettings := communities.CommunitySettings{
+				CommunityID:                  org.IDString(),
+				HistoryArchiveSupportEnabled: true,
+			}
+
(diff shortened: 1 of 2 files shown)


0x-r4bbit:
Only the last commit here is relevant

protocol/messenger.go
```diff
@@ -1273,19 +1273,37 @@ func (m *Messenger) Init() error {
...
+		if !exists {
+			communitySettings := communities.CommunitySettings{
+				CommunityID:                  org.IDString(),
+				HistoryArchiveSupportEnabled: true,
+			}
+
+			err = m.communitiesManager.SaveCommunitySettings(communitySettings)
+			if err != nil {
+				logger.Warn("failed to save community settings", zap.Error(err))
+				continue
```
richard-ramos: Is this `continue` necessary?  we're callng `continue` anyway as soon as we exit this block

0x-r4bbit: Nope, not necessary!

richard-ramos (approved):

PR #2632, opened 2022-04-08
fix: do not export all symbols in windows

richard-ramos:
Fixes:
```
C:\Program Files\Go\pkg\tool\windows_amd64\link.exe: running g++ failed: exit status 1
c:/programdata/chocolatey/lib/mingw/tools/install/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/11.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: error: export ordinal too large: 70071
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,8 +21,7 @@ ifeq ($(detected_OS),Darwin)
     GOBIN_SHARED_LIB_CFLAGS=CGO_ENABLED=1 GOOS=darwin GOARCH=amd64
   endif
 else ifeq ($(detected_OS),Windows)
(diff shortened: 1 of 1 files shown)


jrainville (commented):
It still works without it? The comment seems to say it was needed

richard-ramos:
It does. That comment applied to the go version we were using back then which I assume was 1.15, and looks like it was fixed here: https://github.com/golang/go/issues/43591

jrainville (approved):
Tested, it works fine

PR #2633, opened 2022-04-08
fix: prevent to import same account twice

alaibe:


richard-ramos (approved):

iurimatias (approved):
tested in https://github.com/status-im/status-desktop/pull/5444

PR #2634, opened 2022-04-08
fix: multiaccounts migrations

richard-ramos:
Migrations for accounts.sql where not being executed


iurimatias (approved):
tested in https://github.com/status-im/status-desktop/pull/5456#pullrequestreview-936926281

PR #2636, opened 2022-04-10
Update un-viewed messages counter

jo-mut:
When a message is sent while the recipient device is not active and then deleted the last message of that particular chat is updated but the un-viewed messages counter is not updated to reflect the same which results in showing the count of unseen messages as more than they actually are.

This pr checks if the received message is not seen and if the message is deleted and if true, it updates un-viewed message counter by decrementing the count

### DIFF
--- a/protocol/messenger_handler.go
+++ b/protocol/messenger_handler.go
@@ -758,6 +758,8 @@ func (m *Messenger) HandleDeleteMessage(state *ReceivedMessageState, deleteMessa
 		return errors.New("chat not found")
 	}
 
+	unViewedCounts := chat.UnviewedMessagesCount
+
 	// Check edit is valid
 	if originalMessage.From != deleteMessage.From {
 		return errors.New("invalid delete, not the right author")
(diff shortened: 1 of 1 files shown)


protocol/messenger_handler.go
```diff
@@ -789,6 +791,12 @@ func (m *Messenger) HandleDeleteMessage(state *ReceivedMessageState, deleteMessa
 		}
 	}
 
+	if !originalMessage.Seen && originalMessage.Deleted {
+		if chat.UnviewedMentionsCount >= unViewedCounts {
```
Parveshdhull, line 796: `chat.UnviewedMentionsCount >= unViewedCounts` Isn't this will be always true?

protocol/messenger_handler.go
```diff
@@ -1604,3 +1612,10 @@ func (m *Messenger) updateUnviewedCounts(chat *Chat, mentioned bool) {
 		chat.UnviewedMentionsCount++
 	}
 }
+
+func (m *Messenger) decreamentUnviewedCounts(chat *Chat, mentioned bool) {
```
Parveshdhull, line 1617: Please also rename `updateUnviewedCounts` function to `increamentUnviewedCounts` (So that they both look similar)

Parveshdhull (approved):
Hey @jo-mut, Thank you very much for the PR, other than the above-mentioned changes, great implementation.
(Please also make sure PR is QA tested before merging)

protocol/messenger_handler.go
jo-mut: cool

jo-mut:
Thank you for your review and approval

PR #2637, opened 2022-04-11
fix: generation path

alaibe:


iurimatias (approved):
tested in https://github.com/status-im/status-desktop/pull/5484

PR #2639, opened 2022-04-11
Backup deleted communities

rasom:
fix for status-im/status-react#13198

status: ready


protocol/messenger_backup.go
```diff
@@ -152,6 +152,38 @@ func (m *Messenger) BackupData(ctx context.Context) (uint64, error) {
...
+			Payload:             encodedMessage,
+			SkipEncryption:      true,
+			SendOnPersonalTopic: true,
+			MessageType:         protobuf.ApplicationMetadataMessage_BACKUP,
+		})
+		if err != nil {
+			return 0, err
+		}
+
+	}
```
Samyoul: Do you think that this could be wrapped into a function? I notice that it seems to be identical to the code at line 126.

Samyoul (approved):
Looks good, only one little comment

protocol/messenger_backup.go
rasom: sure, I'm not even sure we need this code at all :D on it

rasom: fixed

Dzhijian:
收到，谢谢。

protocol/communities/persistence.go
```diff
@@ -161,6 +149,38 @@ WHERE c.Joined OR r.state = ?`
...
+	return p.rowsToCommunities(memberIdentity, rows)
+}
+
+func (p *Persistence) DeletedCommunities(memberIdentity *ecdsa.PublicKey) (comms []*Community, err error) {
+	query := `SELECT
+c.id, c.private_key, c.description, c.joined, c.verified, c.muted,
+r.id, r.public_key, r.clock, r.ens_name, r.chat_id, r.community_id, r.state
+FROM communities_communities c
+LEFT JOIN communities_requests_to_join r ON c.id = r.community_id AND r.public_key = ?
+WHERE NOT c.Joined AND (r.community_id IS NULL or r.state != ?)`
```
osmaczko, line 174: Hey @rasom , I am trying to understand why `r.community_id IS NULL` condition is needed, do you remember by any chance? At first glance it doesn't make much sense.

CC: @kounkou

Dzhijian:
收到，谢谢。

protocol/communities/persistence.go
kounkou: It seems to me that https://github.com/rasom is not actively contributing. Should we take a decision here @osmaczko, I believe we should remove the `r.community_id IS NULL` as it doesn't make much sense.

jrainville: He's on parental leave

osmaczko: Congrats!! :baby:  And sorry for bothering.

PR #2642, opened 2022-04-12
chore: test that login still works with an account created with 0.97.3

richard-ramos:
Adds a test that can be used to determine if the current version of status-go still allows login with the status-go version used in the latest release of status-react


api/backend_test.go
```diff
@@ -678,3 +679,82 @@ func TestConvertAccount(t *testing.T) {
...
+
+func TestLoginAndMigrationsStillWorkWithExistingUsers(t *testing.T) {
+	utils.Init()
+
+	b := NewGethStatusBackend()
+
+	srcFolder := "../static/test-0.97.3-account/"
+
+	// The following passwords and DB used in this test unit are only
+	// used to determine if login process works correctly after a migration
```
Samyoul, line 713: 👍

Samyoul (approved):
Ah, this is a nice PR

PR #2646, opened 2022-04-19
uncomment sticker pack synchronization

audriu:
Uncomment sticker pack synchronisation code

Closes #12889


Samyoul:
@audriu there is another issue that relates to this, https://github.com/status-im/status-react/issues/13193 . From the issue there is something else to consider:

> An additional thing to consider is the handling of differences between the sticker pack data payloads.
>
> - Example: stickers/packs-installed may contain 3 sticker packs on device A and 4 sticker packs on device B, and both sets have 2 sticker packs in common and the other sticker packs are unique to their respective devices. What should happen when syncing this data?
> - A potential solution would be to combine the data into a merged set

PR #2651, opened 2022-04-26
fix: default derived from for status account

alaibe:
Default derived from for status account needs to be set

### DIFF
--- a/services/accounts/accounts.go
+++ b/services/accounts/accounts.go
@@ -49,7 +49,23 @@ func (api *API) SaveAccounts(ctx context.Context, accounts []accounts.Account) e
 }
 
 func (api *API) GetAccounts(ctx context.Context) ([]accounts.Account, error) {
-	return api.db.GetAccounts()
+	accounts, err := api.db.GetAccounts()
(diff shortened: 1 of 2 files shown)


services/accounts/accounts.go
```diff
@@ -49,7 +49,23 @@ func (api *API) SaveAccounts(ctx context.Context, accounts []accounts.Account) e
...
-	return api.db.GetAccounts()
+	accounts, err := api.db.GetAccounts()
+	if err != nil {
+		return nil, err
+	}
+
+	for i := range accounts {
+		account := &accounts[i]
+		if account.Wallet && account.DerivedFrom == "" {
+			address, err := api.db.GetWalletRoodAddress()
```
richard-ramos:
```suggestion
			address, err := api.db.GetWalletRootAddress()
```

typo?  `Rood` => `Root`

richard-ramos (approved):

services/accounts/accounts.go
alaibe: Will rename function declaration

PR #2652, opened 2022-04-28
fix: delete account sync keystore

alaibe:
Since we can choose the path when generating account, we can be in a situation where we re-generate an account previously deleted

We need to also clean the keystore file


anastasiyaig:
@churik @cammellos please make it happen to have an ability to test in on mobile side before merge and test it :)

churik:
@alaibe we need to make PR on status-react for this, ping me when it's ready, thanks!

flexsurfer:
https://github.com/status-im/status-react/pull/13315

richard-ramos (approved):

PR #2662, opened 2022-05-05
feat: create General channel with community creation

alaibe:
When creating a community, also create the channel

### DIFF
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -652,9 +652,24 @@ func (m *Messenger) CreateCommunity(request *requests.CreateCommunity) (*Messeng
 		return nil, err
 	}
 
+	chatResponse, err := m.CreateCommunityChat(community.ID(), &protobuf.CommunityChat{
+		Identity: &protobuf.ChatIdentity{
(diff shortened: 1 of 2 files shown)


0x-r4bbit (approved):

protocol/messenger_communities.go
```diff
@@ -652,9 +652,24 @@ func (m *Messenger) CreateCommunity(request *requests.CreateCommunity) (*Messeng
 		return nil, err
 	}
 
+	chatResponse, err := m.CreateCommunityChat(community.ID(), &protobuf.CommunityChat{
+		Identity: &protobuf.ChatIdentity{
+			DisplayName: "General",
```
richard-ramos: I think channel names are lowercase

jrainville (commented):
Looks good, but yeah, I think we only accept lowercase channel names for now. Maybe we can enforce it in status-go after discussing with mobile to make sure they block uppercase as well.

protocol/messenger_communities.go
alaibe: updated

PR #2663, opened 2022-05-05
feat: Add hasActivity param to derived addresses

Khushboo-dev-cpp:
check for has activity for derived addresses

Important changes:

Closes #

### DIFF
--- a/node/status_node_services.go
+++ b/node/status_node_services.go
@@ -443,7 +443,7 @@ func (b *StatusNode) appmetricsService() common.StatusService {
 
 func (b *StatusNode) walletService(accountsDB *accounts.Database, accountsFeed *event.Feed, openseaAPIKey string) common.StatusService {
 	if b.walletSrvc == nil {
-		b.walletSrvc = wallet.NewService(b.appDB, accountsDB, b.rpcClient, accountsFeed, openseaAPIKey)
+		b.walletSrvc = wallet.NewService(b.appDB, accountsDB, b.rpcClient, accountsFeed, openseaAPIKey, b.gethAccountManager)
 	}
 	return b.walletSrvc
 }
--- a/services/wallet/api.go
+++ b/services/wallet/api.go
@@ -2,11 +2,16 @@ package wallet
 
 import (
 	"context"
+	"fmt"
+	"math/big"
+	"strings"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/log"
+	"github.com/status-im/status-go/eth-node/types"
 	"github.com/status-im/status-go/params"
+	"github.com/status-im/status-go/services/wallet/async"
 	"github.com/status-im/status-go/services/wallet/chain"
 	"github.com/status-im/status-go/services/wallet/transfer"
 )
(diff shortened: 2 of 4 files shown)


ghost:
### Pull Request Checklist

node/status_node_services.go
```diff
@@ -441,9 +441,9 @@ func (b *StatusNode) appmetricsService() common.StatusService {
 	return b.appMetricsSrvc
 }
 
-func (b *StatusNode) walletService(accountsFeed *event.Feed, openseaAPIKey string) common.StatusService {
+func (b *StatusNode) walletService(accountsFeed *event.Feed, openseaAPIKey string, accountsDB *accounts.Database) common.StatusService {
 	if b.walletSrvc == nil {
-		b.walletSrvc = wallet.NewService(b.appDB, b.rpcClient, accountsFeed, openseaAPIKey)
+		b.walletSrvc = wallet.NewService(b.appDB, b.rpcClient, accountsFeed, openseaAPIKey, b.gethAccountManager, accountsDB)
```
alaibe: nit: appDB is already a parameter, you don t really need to add accountsDB which can be rebuild from appDB?

services/wallet/api.go
```diff
@@ -293,3 +304,74 @@ func (api *API) GetSuggestedFees(ctx context.Context, chainID uint64) (*Suggeste
...
+		alreadyExists := false
+		for _, account := range addedAccounts {
+			if types.Address(common.HexToAddress(info[derivedPath].Address)) == account.Address {
+				alreadyExists = true
+				break
+			}
+		}
+
+		var ctx context.Context
+		transactions, err := api.s.transferController.GetTransfersByAddress(ctx, api.s.rpcClient.UpstreamChainID, common.HexToAddress(info[derivedPath].Address), nil, (*hexutil.Big)(big.NewInt(1)), false)
```
alaibe: I think this loop and in particular this call should be made in a coroutine to execute all the calls in parallel. If that helps, you can see how it is done here: https://github.com/status-im/status-go/blob/develop/services/wallet/token.go#L245

node/status_node_services.go
Khushboo-dev-cpp: aaah ok let me check :)

services/wallet/api.go
Khushboo-dev-cpp: hey thanks for pointing to an example. It was just really quick when I tested it and didnt do it. Also do you mean https://github.com/status-im/status-go/blob/d884749ebf5e4edff46295c27e583425820e64ac/services/wallet/token.go#L260? cause this qpi getTokenBalance is just a normal function right?

Khushboo-dev-cpp: hey made getting each of the derived address concurrent ! @alaibe

alaibe (approved):

churik:
@Khushboo-dev-cpp any clues on what can be affected on mobile-client side? 
Thanks!

Khushboo-dev-cpp:
> @Khushboo-dev-cpp any clues on what can be affected on mobile-client side? Thanks!

hey @churik this only adds new Apis for account generation and giving the user to choose a derivation path and corresponding address. Doesnt touch any already implemented API.

qoqobolo (approved):

qoqobolo:
@Khushboo-dev-cpp thanks for your patience and waiting! PR is tested on the Mobile side.

PR #2668, opened 2022-05-10
feat: introduce `CommunityAdminSettings` in `CommunityDescription`

0x-r4bbit:
This allows to store community admin settings that are meant to be propagated
to community members (as opposed to the already existing
`CommunitySettings` which are considered local to every account).

The first setting introduced as part of this commit is one that enables
community admins to configure whether or not members of the community
are allowed to pin messages in community channels.

Prior to this commit, this was not restricted at all on the protocol
level and only enforced by clients via UI (e.g. members don't see an
option to pin messages, although they could).

This config setting now ensures that:

1. If turned off, members cannot send a pin message
2. If turned off, pin messages from members are not handled/processed

This is needed by https://github.com/status-im/status-desktop/issues/5662


protocol/messenger_handler.go
```diff
@@ -1379,7 +1379,16 @@ func (m *Messenger) matchChatEntity(chatEntity common.ChatEntity) (*Chat, error)
...
+
+		community, err := m.communitiesManager.GetByIDString(chat.CommunityID)
+		if err != nil {
+			return nil, err
+		}
+
+		isMemberAdmin := community.IsMemberAdmin(chatEntity.GetSigPubKey())
+		pinMessageAllowed := community.AllowsAllMembersToPinMessage()
+
+		if (!isMemberAdmin && !pinMessageAllowed) || (!emojiReaction && !canPost) {
```
jrainville: Should there be an additionnal check on pin? It seems like this might block people from posting normal messages as well.
Something like 
```go
(isPinMessage && !isMemberAdmin && !pinMessageAllowed)
```

Where `isPinMessage` comes from a new `case` in the `switch` above.

jrainville (commented):
Looks good, but I think on the receiving end, there might be a missing condition. I might be wrong though.

protocol/messenger_handler.go
0x-r4bbit: Ah, forgot to add that. In my head I was still inside `HandlePinMessage`...

Thanks man, will update

Samyoul (approved):
Looks good. ~~I'm presuming that in the future more parameters will be added to the `CommunityAdminSettings`?~~ Ah, I just reread the description.

0x-r4bbit:
Tests are failing but unrelated to these changes.
This has been reported here as well: https://github.com/status-im/status-go/pull/2672#issuecomment-1125766117

Also, I think https://github.com/status-im/status-desktop/issues/5729 is related.

Investigating...

0x-r4bbit:
So it seems it affects all tests that involve `JoinCommunity` APIs. I can confirm that `TestJoinCommunity` does **not** fail when run standalone

protocol/protobuf/communities.proto
```diff
@@ -44,6 +44,11 @@ message CommunityDescription {
   repeated string ban_list = 7;
   map<string,CommunityCategory> categories = 8;
   uint64 archive_magnetlink_clock = 9;
+  CommunityAdminSettings admin_settings = 10;
+}
+
+message CommunityAdminSettings {
+  bool pin_message_all_members_enabled = 1;
 }
```
richard-ramos, line 52: This is very nice!

richard-ramos (approved):

churik:
Hello!
Is there a particular reason why this PR was not going through testing?

0x-r4bbit:
@churik That one is on me!
The PR was sitting around for quite a while and there wasn't any activity anymore after getting a bunch of approvals..
But now that I think of it, I should've pinged you, correct?

churik:
@PascalPrecht 
No worries, https://notes.status.im/pipeline-status-go - is the doc with workflow for next time

PR #2669, opened 2022-05-10
notification settings added to the settings

saledjenic:


multiaccounts/settings/columns.go
```diff
@@ -197,6 +197,11 @@ var (
 		dBColumnName:   "notifications_enabled",
 		valueHandler:   BoolHandler,
 	}
+	NotificationsSettings = SettingField{
+		reactFieldName: "notifications-settings",
+		dBColumnName:   "notifications_settings",
+		valueHandler:   JSONBlobHandler,
+	}
```
richard-ramos: Ideally we should not store JSON blobs (in fact in the future we should migrate the existing jsons to separate tables)

appdatabase/migrations/sql/1652194966_add_notifications_settings_column.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE settings ADD COLUMN notifications_settings BLOB;
```
0x-r4bbit: I wonder if we should rather introduce a new table `notification_settings` instead, otherwise doing migrations on this data is going to be hard..

0x-r4bbit: Ah, @richard-ramos beat me to it: https://github.com/status-im/status-go/pull/2669#discussion_r869542156

jakubgs:
The Lint stage failures appear to be specific to `linux-01` only since that's the only host upgraded to Go `1.18.1`:
https://ci.status.im/job/status-go/job/prs/job/tests/job/PR-2669/1/execution/node/60/log/

@cammellos is working on upgrading to that version. What do you think, should we just ignore those failures for now?

EDIT: This issue has been resolved in https://github.com/status-im/status-go/pull/2670.

multiaccounts/settings/columns.go
saledjenic: @richard-ramos I agree and I am totally up to that, but I just followed what we had there. If we're talking how it should be, we also shouldn't have new column for each new settings field. How our `settings` table looks now... it has 30+ maybe 40+ columns and a single row, that's all. So that's not the way of doing things, but changing that flow is a big change and should be done in a separate ticket aligning with mobile team.

appdatabase/migrations/sql/1652194966_add_notifications_settings_column.up.sql
saledjenic: @PascalPrecht introducing new table is good way to redo (refactor) settings thing in an appropriate way. I mentioned [here](https://github.com/status-im/status-go/pull/2669#discussion_r869598466) how it looks now.

multiaccounts/settings/columns.go
alaibe: the migration has started: https://github.com/status-im/status-go/pull/2470 also as u said quite complex (and we don t have much time for it). maybe we should not just add new blob.

saledjenic: I see it as some of the following:
- accept this PR and deal with this column along with the other columns (since this is not the only blob containing json string)
- to add separate column within this table for each notification settings property we need to store, but there is another issue... we need to store exemptions as well and they refer to `1:1` and `group` chat and `community` and those notifications properties are maintained by their ids

@alaibe wdyt, your suggestion?

0x-r4bbit: Another option is to introduce a new table instead of columns here and then created dedicated APIs in go to easily access those settings.

See all `communities_settings` related APIs as an inspiration (did the same thing there as storing the data as BLOB wasn't feasible due to future migration challenges)

saledjenic: Ok, so what do you guys think? Should I do that within this PR?

cammellos: I'd say yes (not for all the fields, just for the newly introduced of course), as once it's out, it's difficult to migrate, so best to have it in the shape we want in the first go, what do you think?

saledjenic: @cammellos deal, will add these settings to a new table soon.

saledjenic:
@all please give another round of review since, as we have discussed, I moved notification related settings to a new table `notifications_settings`.

saledjenic:
Next steps we should do in regards of notifications settings are:
- add them to the synchronization so all devices may share the same notifications settings
- move notifications logic from the desktop app to `status-go` side

multiaccounts/settings_notifications/database.go
```diff
@@ -0,0 +1,245 @@
...
+		ex_personal_mentions,
+		ex_global_mentions,
+		ex_other_messages) 
+		%s;`, values)
+}
+
+// Non exemption settings
+func (ns *NotificationsSettings) GetAllowNotifications() (result bool, err error) {
+	err = ns.db.QueryRow(ns.buildSelectQuery(ColumnBoolValue), KeyAllowNotifications).Scan(&result)
+	return result, err
```
richard-ramos: One of the errors that these new functions might return is `sql.ErrNoRows`, which happen if the resultset is empty. Should that be considered an error? if not, something like this could be necessary
```go
if err != nil && err != sql.ErrNoRows {
  return false, err
}
```

multiaccounts/settings_notifications/database.go
```diff
@@ -0,0 +1,245 @@
...
+// Columns' names
+const ColumnID = "id"
+const ColumnExemption = "exemption"
+const ColumnTextValue = "text_value"
+const ColumnIntValue = "int_value"
+const ColumnBoolValue = "bool_value"
+const ColumnExMuteAllMessages = "ex_mute_all_messages"
+const ColumnExPersonalMentions = "ex_personal_mentions"
+const ColumnExGlobalMentions = "ex_global_mentions"
+const ColumnExOtherMessages = "ex_other_messages"
```
richard-ramos: Good candidate for go's "fake enums" :)
```suggestion
type Column string
const (
    ColumnID Column = "id"
    ColumnExemption = "exemption"
    ColumnTextValue = "text_value"
    ColumnIntValue = "int_value"
    ColumnBoolValue = "bool_value"
    ColumnExMuteAllMessages = "ex_mute_all_messages"
    ColumnExPersonalMentions = "ex_personal_mentions"
    ColumnExGlobalMentions = "ex_global_mentions"
    ColumnExOtherMessages = "ex_other_messages"
 )
```
Then, in functions like `buildSelectQuery(column string)`, you'd change the parameter name to `Column`. That way you guarantee that the functions will be called with the columns you intended and not with any random string

multiaccounts/settings_notifications/database.go
```diff
@@ -0,0 +1,245 @@
...
+const KeyIdentityVerificationRequests = "IdentityVerificationRequests"
+const KeySoundEnabled = "SoundEnabled"
+const KeyVolume = "Volume"
+const KeyMessagePreview = "MessagePreview"
+
+// Possible values some of ids may have
+const ValueSendAlerts = "SendAlerts"
+const ValueDeliverQuietly = "DeliverQuietly"
+const ValueTurnOff = "TurnOff"
+
```
richard-ramos: other good candidates for go "fake enums"

multiaccounts/settings_notifications/database.go
```diff
@@ -0,0 +1,245 @@
...
+type SettingType int
+
+const (
+	String SettingType = iota
+	Integer
+	Bool
+)
+
+// Columns' names
+const ColumnID = "id"
```
richard-ramos: If these constants are not meant to be exported, use lowercase

services/accounts/settings.go
```diff
@@ -41,3 +41,118 @@ func (api *SettingsAPI) NodeConfig(ctx context.Context) (*params.NodeConfig, err
...
+// Notifications Settings
+func (api *SettingsAPI) GetAllowNotifications() (bool, error) {
+	return api.db.GetAllowNotifications()
+}
+
+func (api *SettingsAPI) SetAllowNotifications(value bool) error {
+	return api.db.SetAllowNotifications(value)
+}
+
+func (api *SettingsAPI) GetOneToOneChats() (string, error) {
```
richard-ramos: This function will be available as `settings_getOneToOneChats`. Without knowing the context or implementation, it's difficult to know that this function deals with notifications. Consider either adding a prefix like "Notif" to the function name, or using a different service.  (same observation applies to other functions)

saledjenic:
@richard-ramos thanks for the review, I updated constants as you suggested.

multiaccounts/settings_notifications/database.go
saledjenic: @richard-ramos this depends a lot on the flow you're applying. I added that case you suggested, and returning now default values for each setting in case there is no appropriate row containing that setting in the table, but still in case of other sql errors returning the error, which are properly handled on the desktop app side.

richard-ramos (approved):
Nice!

multiaccounts/settings_notifications/database.go
```diff
@@ -0,0 +1,315 @@
...
+	columnBoolValue          column = "bool_value"
+	columnExMuteAllMessages  column = "ex_mute_all_messages"
+	columnExPersonalMentions column = "ex_personal_mentions"
+	columnExGlobalMentions   column = "ex_global_mentions"
+	columnExOtherMessages    column = "ex_other_messages"
+)
+
+// Static ids we use for notifications settings
+const (
+	keyAllowNotifications           = "AllowNotifications"
```
cammellos, line 31: We tend to use enums with `iota` (so integers) for these, unless there's any particular reason to use strings?

saledjenic: @cammellos primary key is string cause for exemptions it will represent either community id, or 1:1 chat id, or group chat id and all of them are strings, leaded by that I created those constants which represent primary keys for the static notification settings (those known in advance, defined by design) and found it more readable using a string  for them than integer.

Also these constants starting with "key" are not part of enum, those are just constants.

cammellos (approved):

PR #2670, opened 2022-05-11
ci: pin builds to hosts with specific Go version

jakubgs:
Otherwise Go version upgrades are a pain due to randomly hitting a host with newer version of Go compiler.


cammellos (approved):

jakubgs:
Looks like unit tests are failing: https://ci.status.im/job/status-go/job/prs/job/tests/job/PR-2670/

image

But I'm just going to ignore that, because this change couldn't possibly cause that.

PR #2664, opened 2022-05-06
fix: merge node config allows overide

alaibe:

PR #2675, opened 2022-05-16
the way of initializing networks list is updated

saledjenic:


api/geth_backend.go
```diff
@@ -363,6 +363,8 @@ func (b *GethStatusBackend) OverwriteNodeConfigValues(conf *params.NodeConfig, n
 		return nil, err
 	}
 
+	conf.Networks = n.Networks
```
saledjenic, line 365: Following [description](https://pkg.go.dev/github.com/imdario/mergo@v0.3.12#Merge) for `Merge` function, Networks slice is not updated accordingly, so it remains the same as it was in `conf`, that's why I did that manually here.

rpc/network/network.go
```diff
@@ -83,16 +84,27 @@ func (nm *Manager) Init(networks []params.Network) error {
 		return nil
 	}
 
+	var errors string
 	currentNetworks, _ := nm.Get(false)
-	if len(currentNetworks) > 0 {
```
saledjenic, line 87: This condition here will, in all cases but when you're running the app for the first time, be true. So newly added networks won't be added to the db, `networks` table ever. So in the for loop below we're checking if the network with `ChainID` exists in the table and if not, we're adding it.

But I think that we should overwrite entries always since `rpc_url` contains infura key which can be changed depends we're running the app locally or one built on CI. Do you agree?

api/geth_backend.go
alaibe: FYI, I am waiting on this PR: https://github.com/status-im/status-go/pull/2664
That should make that line redundant right?

rpc/network/network.go
alaibe: > This condition here will, in all cases but when you're running the app for the first time, be true. So newly added networks won't be added to the db, networks table ever. So in the for loop below we're checking if the network with ChainID exists in the table and if not, we're adding it.

Yes agree with that

> But I think that we should overwrite entries always since rpc_url contains infura key which can be changed depends we're running the app locally or one built on CI. Do you agree?

I agree with this statement, but we also need to take in consideration that user can update the network. We may be need a flag `changedByUser` and if true, don t do the update

alaibe (approved):

api/geth_backend.go
saledjenic: Referring to this line from the PR you mentioned `if err := mergo.Merge(conf, n, mergo.WithOverride); err != nil {`

From the description it says `WithOverride will make merge override non-empty dst attributes with non-empty src attributes values.` since `conf.Networks` is not an empty array, I think it won't do the overwrite.

richard-ramos (approved):

saledjenic:
@jakubgs please may you have a look why the tests fail here?

jakubgs:
You tell me:
```
[2022-05-26T12:35:28.622Z] panic: test timed out after 10m0s
[2022-05-26T12:35:28.622Z] 
[2022-05-26T12:35:28.622Z] goroutine 1148417 [running]:
[2022-05-26T12:35:28.622Z] testing.(*M).startAlarm.func1()
[2022-05-26T12:35:28.622Z] 	/usr/local/go/src/testing/testing.go:2029 +0x8e
[2022-05-26T12:35:28.622Z] created by time.goFunc
[2022-05-26T12:35:28.622Z] 	/usr/local/go/src/time/sleep.go:176 +0x32
[2022-05-26T12:35:28.622Z] 
[2022-05-26T12:35:28.622Z] goroutine 1 [chan receive]:
[2022-05-26T12:35:28.622Z] testing.(*T).Run(0xc000b7e000, {0x1e3e50c?, 0x529385?}, 0x20e1ec0)
[2022-05-26T12:35:28.622Z] 	/usr/local/go/src/testing/testing.go:1487 +0x37a
[2022-05-26T12:35:28.622Z] testing.runTests.func1(0xc000b789c0?)
[2022-05-26T12:35:28.622Z] 	/usr/local/go/src/testing/testing.go:1839 +0x6e
[2022-05-26T12:35:28.622Z] testing.tRunner(0xc000b7e000, 0xc0011c3cd8)
[2022-05-26T12:35:28.622Z] 	/usr/local/go/src/testing/testing.go:1439 +0x102
[2022-05-26T12:35:28.622Z] testing.runTests(0xc0007023c0?, {0x33b1000, 0x3b, 0x3b}, {0x7f7a446621d8?, 0x40?, 0x35160e0?})
[2022-05-26T12:35:28.622Z] 	/usr/local/go/src/testing/testing.go:1837 +0x457
[2022-05-26T12:35:28.622Z] testing.(*M).Run(0xc0007023c0)
[2022-05-26T12:35:28.622Z] 	/usr/local/go/src/testing/testing.go:1719 +0x5d9
[2022-05-26T12:35:28.622Z] main.main()
[2022-05-26T12:35:28.622Z] 	_testmain.go:163 +0x1aa
```

saledjenic:
@jakubgs maybe you can increase that time to 15mins?

richard-ramos:
I wonder what's going on, because tests across different PRs are from time to time panicking due to a 10m timeout

jakubgs:
> @jakubgs maybe you can increase that time to 15mins?

This is not CI related my dude. This is a matter of how the tests are implemented.

PR #2678, opened 2022-05-17
fix: ping dns4 multiaddresses

richard-ramos:
This PR fixes an issue that happened when multiaddresses contain `dns4` protocol and not `ip4`.
The fix will extract the `dns4` domain name from multiaddresses (if available), and resolve the IP from that domain before pinging the store node. If `dns4` is not available, it will use `ip4`. Otherwise fail with an err.
Also adds the `status.test` and `status.prod` waku2 fleet storenodes

### DIFF
--- a/services/mailservers/fleet.go
+++ b/services/mailservers/fleet.go
@@ -131,5 +131,60 @@ func DefaultMailservers() []Mailserver {
 			Fleet:   params.FleetWakuV2Test,
 			Version: 2,
 		},
+		Mailserver{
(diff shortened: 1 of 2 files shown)


0x-r4bbit (approved):
Nice!

alaibe (approved):

iurimatias:
hi @churik we need this somewhat urgenly, would it be ok with if you if we merge it? cc: @cammellos

cammellos:
@iurimatias thanks, just spoke to @churik and she's fine merging, thank you

PR #2682, opened 2022-05-20
refactor: Require request access for all communities 

0x-r4bbit:
This commit introduces a few changes regarding users accessing
communities:

1. While the APIs still exist, community invites should no longer be
   used, instead communities should merely be "shared".
2. Sharing a community to users allows users to "join" the community,
   which in reality makes them request access to that community.
3. This means, users have to request access to any community, even if
   the community has permissions set to `NO_MEMBERSHIP`
4. Only difference between `ON_REQUEST` and `NO_MEMBERSHIP` is that
   `ON_REQUEST` communities require manual approval of the owner/admin
   to access a community. `NO_MEMBERSHIP` communities accept
   automatically (as soon as owner/admin receives the request).

This also implies that users are no longer optimistically added to the
member list of communities, but only after they have been accepted.

This introduces a bit of a message ping-pong for users to know that
someone is now part of a community


protocol/communities_messenger_test.go
```diff
@@ -618,15 +620,13 @@ func (s *MessengerCommunitiesSuite) TestImportCommunity() {
 		if len(response.Communities()) == 0 {
 			return errors.New("community not received")
 		}
+		if !response.Communities()[0].IsAdmin() {
+			return errors.New("isn't admin despite import")
+		}
```
0x-r4bbit, line 630: I tried putting the `Joined()` check for alice in here as well, but it always makes the test fail in a weird way.

0x-r4bbit (commented):
@cammellos @Samyoul sorry to bug you with this, but I just can't seem to figure out how to make the tests pass in this one.

The issue still seems to be that the accept-to-join-responses are sent/processed asynchronously, and therefore are not part of the `response` that's being tested.

I tried putting in `time.Sleep()` calls to have some time pass but that either makes the tests time out or just fail with the same error.

This is working when testing it in actual desktop instances, just having trouble getting this simulated properly in the testsuite :(

protocol/communities_messenger_test.go
cammellos: looking into it, would you mind pushing the failed test if you have time please?

0x-r4bbit: It's already pushed.

The tests that are failing are `TestRequestAccess` and `TestRequestAccessAgain`

Let me know if I can be of any assistance and thanks for taking the time!

protocol/communities/manager.go
```diff
@@ -819,9 +840,63 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request
...
+
+	// We need to wrap `request.Community` in an `ApplicationMetadataMessage`
+	// of type `CommunityDescription` because `UpdateCommunityDescription` expects this.
+	//
+	// This is merely for marsheling/unmarsheling, hence we attaching a `Signature`
+	// is not needed.
+	metadataMessage := &protobuf.ApplicationMetadataMessage{
+		Payload: communityDescriptionBytes,
+		Type:    protobuf.ApplicationMetadataMessage_COMMUNITY_DESCRIPTION,
+	}
```
0x-r4bbit, line 908: @Samyoul Any objections here to leave the signature out?

0x-r4bbit:
BTW I'll squash the second commit into the first one once this has been reviewed, so that we don't introduce a commit that breaks .
I've split it up for now so it's easier to review, so no need to review the second commit.

protocol/communities_messenger_test.go
```diff
@@ -574,6 +574,8 @@ func (s *MessengerCommunitiesSuite) TestImportCommunity() {
 	s.Require().NotNil(response)
 	s.Require().Len(response.Communities(), 1)
 	s.Require().Len(response.CommunitiesSettings(), 1)
+	s.Require().True(response.Communities()[0].Joined())
```
saledjenic, line 582: @PascalPrecht isn't a bit dangerous accessing here with direct index without checking before that `if len(response.Communities()) > 0`?

0x-r4bbit: There's 

> 	s.Require().Len(response.Communities(), 1)

Two lines above that (L575), which should do the trick, no?

saledjenic: @PascalPrecht yes, you're right, missed that somehow even I looked around this line.

protocol/protobuf/communities.proto
```diff
@@ -83,9 +83,10 @@ message CommunityRequestToJoin {
 
 message CommunityRequestToJoinResponse {
   uint64 clock = 1;
-  CommunityDescription community = 2;
-  bool accepted = 3;
-  bytes grant = 4;
+  bytes community_id = 2;
+  CommunityDescription community = 3;
+  bool accepted = 4;
+  bytes grant = 5;
```
Samyoul: This is a breaking change. If an older client sends this message type it will be parsed incorrectly by a newer client, if an older client receives this message from a newer client it will be parsed incorrectly by the older client.

0x-r4bbit: Good catch!
This is not a problem if we introduce `community_id` as field `5` and leave the rest as is, right?

Samyoul: Yes correct

0x-r4bbit: Updated

protocol/communities/manager.go
Samyoul: This is fine

Samyoul (approved):
I've spent a bit of time going through this and this looks great.

0x-r4bbit:
@qoqobolo @churik This one is ready for testing on mobile!

elina2015:
@richard-ramos could you kindly build  https://github.com/status-im/status-desktop/pull/6082  for Desktop testing of this ticket?

0x-r4bbit:
The failing tests here seem unrelated to the changes (the PR has passed test stage before, I've only rebased it due to conflicts in generated bindata)

richard-ramos (approved):

PR #2684, opened 2022-05-20
feat: remove explicit group chat join

osmaczko:
- users added to the group don't need to explicitly join anymore
- for backward compatibility, the client emits a join event (mimic auto-join), so that legacy clients can interact with new ones
- legacy join events are ignored

Required by: [#status-im/status-desktop/5717](https://github.com/status-im/status-desktop/issues/5717)


protocol/messenger.go
```diff
@@ -1975,6 +1975,7 @@ func (m *Messenger) AddAdminsToGroupChat(ctx context.Context, chatID string, mem
 	return m.addMessagesAndChat(chat, buildSystemMessages([]v1protocol.MembershipUpdateEvent{event}, m.systemMessagesTranslations), &response)
 }
 
+// Kept only for backward compatibility, explicit join has been removed
```
cammellos: I think we can drop it, we only need to be backward compatible protocol between users, but this will ship with the application (so it's a "local" breaking change, which is ok, while we want to have backward compatibility with peers)

protocol/messenger_handler.go
```diff
@@ -187,7 +187,8 @@ func (m *Messenger) HandleMembershipUpdate(messageState *ReceivedMessageState, c
 	// Store in chats map as it might be a new one
 	messageState.AllChats.Store(chat.ID, chat)
 
-	if waitingForApproval {
+	autoJoin := wasUserAdded // explicit join has been removed, mimic auto-join for backward compatibility
```
cammellos: This is a bit tricky, because it can be used as a vector for attack, a spammer can send a lot of these requests and force the device to reply to each one, effectively DOSing the device.
I think we can get away with a slightly different behavior, we can auto-join if:

1) the chat is active (in that case the user must have already opened the chat)
2) When the user accepts the notification on the activity center

That way (if I am not mistaking):

1) If the user is invited by someone running a new build, no backward compatibility problems, as they send to everyone, messages will be saved even if you haven't explicitly joined, as per this code :+1: 

2) Old users will only send messages to this user once they join, and they will see that he has joined, only when they open the chat from the activity center, which is (for them), expected behavior

Not sure I am missing something, what do you think?

services/chat/api_group_chats.go
```diff
@@ -120,13 +121,7 @@ func (api *API) MakeAdmin(ctx context.Context, communityID types.HexBytes, chatI
 }
 
 func (api *API) ConfirmJoiningGroup(ctx context.Context, communityID types.HexBytes, chatID string) (*GroupChatResponse, error) {
```
cammellos: This can be removed as well, as per comment above

cammellos (commented):
Looks great!
Just left some points to consider.

protocol/messenger_handler.go
osmaczko: That's a good point!

> the chat is active (in that case the user must have already opened the chat)

according to this condition: `isActive := messageState.CurrentMessageState.Contact.Added || messageState.CurrentMessageState.Contact.ID == ourKey || waitingForApproval`chat is active if we were added by added contact.  As a consequence we can be spammed by added contacts, but that's acceptable, isn't it? We decide who we add to the contacts and we can remove malicious contacts at any time :thinking: 

Could we simplify the flow and only auto-join only if we are added by a contact? I know it will not cover all scenarios, e.g. when legacy user invites us before we accept contact request, but since it's kind of an edge case, it's not worth complicating the flow.

protocol/messenger.go
osmaczko: This one we can't, we use it for auto-join. I will update the comment.

protocol/messenger_handler.go
cammellos: > As a consequence we can be spammed by added contacts, but that's acceptable, isn't it? 


Yes, that's fine I think

> Could we simplify the flow and only auto-join only if we are added by a contact? 


Sure, we can start with that, maybe leave a comment explaining that not all the cases are covered (i.e invited by a non-contact), and we can either address separately or leave it as it is, we can make a decision closer to mobile release (we'd be looking into it, you don't have to), thanks!

protocol/messenger.go
cammellos: oh yes you are rightt, sorry missed that we use this internally, thanks!

cammellos (approved):

osmaczko:
Auto-join functionality (on the left legacy client, on the right new client):

https://user-images.githubusercontent.com/33099791/169777093-ffa81d2d-5112-48d2-833d-f2b5de7b73fc.mp4

osmaczko:
This PR waits for Mobile adaptation and tests. @cammellos please link mobile PR when done.

richard-ramos (approved):

PR #2686, opened 2022-05-23
feat: Simple transaction router

alaibe:
* Add a new transaction endpoint accepting the chain id as a parameter
* Add a simple router, which check amongst all networks available the one where a transaction can be routed to

### DIFF
--- a/api/geth_backend.go
+++ b/api/geth_backend.go
@@ -875,6 +875,22 @@ func (b *GethStatusBackend) SendTransaction(sendArgs transactions.SendTxArgs, pa
 	return
 }
 
+func (b *GethStatusBackend) SendTransactionWithChainID(chainID uint64, sendArgs transactions.SendTxArgs, password string) (hash types.Hash, err error) {
+	verifiedAccount, err := b.getVerifiedWalletAccount(sendArgs.From.String(), password)
+	if err != nil {
+		return hash, err
+	}
+
+	hash, err = b.transactor.SendTransactionWithChainID(chainID, sendArgs, verifiedAccount)
+	if err != nil {
+		return
+	}
+
+	go b.statusNode.RPCFiltersService().TriggerTransactionSentToUpstreamEvent(hash)
+
+	return
+}
+
 func (b *GethStatusBackend) SendTransactionWithSignature(sendArgs transactions.SendTxArgs, sig []byte) (hash types.Hash, err error) {
 	hash, err = b.transactor.SendTransactionWithSignature(sendArgs, sig)
 	if err != nil {
--- a/services/wallet/router.go
+++ b/services/wallet/router.go
@@ -1,0 +1,122 @@
+package wallet
+
+import (
+	"context"
+	"math"
+	"math/big"
+	"sync"
+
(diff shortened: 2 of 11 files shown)


api/geth_backend.go
```diff
@@ -874,6 +874,22 @@ func (b *GethStatusBackend) SendTransaction(sendArgs transactions.SendTxArgs, pa
...
+	}
+
+	hash, err = b.transactor.SendTransactionWithChainID(chainID, sendArgs, verifiedAccount)
+	if err != nil {
+		return
+	}
+
+	go b.statusNode.RPCFiltersService().TriggerTransactionSentToUpstreamEvent(hash)
+
+	return
```
saledjenic, line 891: @alaibe is this needed here?

services/wallet/router.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	var (
+		group      = async.NewAtomicGroup(ctx)
+		mu         sync.Mutex
+		candidates = make([]params.Network, 0)
+	)
+	for networkIdx := range networks {
+		network := networks[networkIdx]
+		if network.IsTest != areTestNetworksEnabled {
```
saledjenic, line 70: @alaibe is this condition what you really want to achieve? 
That condition means you're adding network in the following 2 cases:
- if it is a test network and test networks are enabled 
- if it is not a test network and test networks are not enabled 

but I think you should add it in the following case as well:
- if it is not a test network and test networks are enabled

if above is true then condition should be like:
```
if network.IsTest && !areTestNetworksEnabled {
    continue
}
```

api/geth_backend.go
alaibe: Will double check and remove if not needed

services/wallet/router.go
alaibe: > if it is not a test network and test networks are enabled

Why do think I should add this:
I don t want to make transaction on mainnet if test networks are enable.
If test networks are enable, I only want to transfer on ropsten/rinkeby...

api/geth_backend.go
richard-ramos: It's needed when you use named return parameters

services/wallet/router.go
```diff
@@ -0,0 +1,120 @@
...
+
+type Router struct {
+	s *Service
+}
+
+func (r *Router) suitableTokenExists(ctx context.Context, network *params.Network, tokens []*Token, account common.Address, amount float64, tokenSymbol string) (bool, error) {
+	for _, token := range tokens {
+		if token.Symbol != tokenSymbol {
+			continue
+		}
```
richard-ramos, line 31: What do you think of checking the token address instead of symbol?

richard-ramos: Ah no, this is supposed to be multi chain, right? so the addresses are gonna be different.

richard-ramos (approved):

services/wallet/router.go
alaibe: correct, I need to use the symbol as key

Samyoul (approved):

PR #2687, opened 2022-05-23
feat: add banner support for communities

### LINKED ISSUE #5403: Add banner support to backend (status-go)

stefandunca:
Add banner image as a special "banner" `IdentityImage` beside "thumbnail" and "large"

Changes to `images` module

- Refactor `EncodeToBestSize` as `EncodeToLimits` to accept arbitrary dimmensions to allow for custom size
- Define `DimensionLimits` for banner not to exceed 450 KB

Closes [status-desktop #5403](https://github.com/status-im/status-desktop/issues/5403)

### DIFF
--- a/images/encode.go
+++ b/images/encode.go
@@ -25,7 +25,7 @@ func renderJpeg(w io.Writer, m image.Image, config EncodeConfig) error {
 	return jpeg.Encode(w, m, o)
 }
 
-func EncodeToBestSize(bb *bytes.Buffer, img image.Image, size ResizeDimension) error {
+func EncodeToLimits(bb *bytes.Buffer, img image.Image, bounds DimensionLimits) error {
 	q := MaxJpegQuality
 	for q > MinJpegQuality-1 {
 
@@ -34,17 +34,17 @@ func EncodeToBestSize(bb *bytes.Buffer, img image.Image, size ResizeDimension) e
 			return err
 		}
 
-		if DimensionSizeLimit[size].Ideal > bb.Len() {
+		if bounds.Ideal > bb.Len() {
 			return nil
 		}
 
 		if q == MinJpegQuality {
-			if DimensionSizeLimit[size].Max > bb.Len() {
+			if bounds.Max > bb.Len() {
 				return nil
 			}
 			return fmt.Errorf(
 				"image size after processing exceeds max, expect < '%d', received < '%d'",
-				DimensionSizeLimit[size].Max,
+				bounds.Max,
 				bb.Len(),
 			)
 		}
@@ -54,6 +54,10 @@ func EncodeToBestSize(bb *bytes.Buffer, img image.Image, size ResizeDimension) e
 	}
 
 	return nil
+}
+
+func EncodeToBestSize(bb *bytes.Buffer, img image.Image, size ResizeDimension) error {
+	return EncodeToLimits(bb, img, DimensionSizeLimit[size])
 }
 
 func GetPayloadDataURI(payload []byte) (string, error) {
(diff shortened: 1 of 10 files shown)


images/encode.go
```diff
@@ -25,7 +25,7 @@ func renderJpeg(w io.Writer, m image.Image, config EncodeConfig) error {
 	return jpeg.Encode(w, m, o)
 }
 
-func EncodeToBestSize(bb *bytes.Buffer, img image.Image, size ResizeDimension) error {
+func EncodeToLimits(bb *bytes.Buffer, img image.Image, bounds DimensionLimits) error {
```
Samyoul, line 28: Nice refactor

richard-ramos (approved):

images/meta.go
```diff
@@ -58,3 +60,12 @@ type DimensionLimits struct {
 
 type ImageType uint
 type ResizeDimension uint
+
+const BannerIdentityName string = "banner"
```
Samyoul: Maybe you could move this up to line 24 to be with the other image names

images/meta.go
```diff
@@ -58,3 +60,12 @@ type DimensionLimits struct {
...
+
+func GetBannerDimension() DimensionLimits {
+	return DimensionLimits{
+		Ideal: 460800, // We want the best quality supported
```
Samyoul: Why do we want to aim for the maximum size? Shouldn't at least try to save space on the image?

protocol/requests/cropped_image.go
```diff
@@ -0,0 +1,9 @@
+package requests
+
+type CroppedImage struct {
+	ImagePath string `json:"imagePath"`
+	X         int    `json:"x"`
+	Y         int    `json:"y"`
+	Width     int    `json:"width"`
+	Height    int    `json:"height"`
+}
```
Samyoul: This type should probably live in the `images` package

Samyoul (commented):
Looks good. A few points nothing major. One thing that I don't see is any tests for the banner functionality, do you feel confident in adding these?

images/meta.go
stefandunca: Thanks for the heads up. I now realize I didn't give much thought to the importance of size to traffic and storage. I only take into consideration that
- community admins would want maximum quality
- a user won't be part of many communities and sync many logos
- the design requirement is that we are not allowed to have images bigger than 450 KB

Should we conduct an analysis similar to the logo image [here](https://github.com/status-im/status/issues/36#issuecomment-694970473) ?

Is it enough for now to use a rough estimation and test the effects? I changed it to `102400` which looks good enough in my tests

stefandunca (commented):
> ... I don't see is any tests for the banner functionality, do you feel confident in adding these?

@Samyoul, how about I add integration tests to check that banner image processing respects the imposed size limits? However, that means working with big images, which might make tests longer than a few milliseconds. Is this a problem?

Do you have anything else in mind for the test scope?

Also, I'm not familiar with image processing in go. Any hints on what should I use to generate input test images so that I don't pollute the test source code with big files?

images/meta.go
Samyoul: Yes, that sounds fine to me. Yeah bandwidth is always a big concern with the nature of our application

Samyoul:
Perhaps check the `IdentityImage` components of `ToCommunityDescription()`

stefandunca:
@Samyoul, I did all the requested changes, so please go ahead and have another look.

Also, after some tests with different compression quality, base resolution, and image complexity and talking to @John-44 I raised the ideal image quality to 300KB, which is an edge case, to target maximum quality.

I also changed the behaviour for banners not to upscale to target quality but only shrink if necessary. We can handle the upscaling in the UIs and save some space/traffic.

Samyoul (approved):
Great, looks good

PR #2688, opened 2022-05-24
Feat/community intro message

osmaczko:
needed by: https://github.com/status-im/status-desktop/issues/5746
status-desktop PR: https://github.com/status-im/status-desktop/pull/5882


protocol/protobuf/communities.proto
```diff
@@ -45,6 +45,8 @@ message CommunityDescription {
   map<string,CommunityCategory> categories = 8;
   uint64 archive_magnetlink_clock = 9;
   CommunityAdminSettings admin_settings = 10;
+  string introMessage = 11;
+  string outroMessage = 12;
```
osmaczko: I will change it to snake_case to be consistent with other fields

protocol/communities/community.go
```diff
@@ -273,6 +281,24 @@ func (o *Community) DescriptionText() string {
 	return ""
 }
 
+func (o *Community) IntroMessage() string {
+	if o != nil &&
```
0x-r4bbit, line 287: I don't think we need this check. `IntroMessage()` wouldn't be callable on `o` if `o was `nil`.

protocol/requests/create_community_request.go
```diff
@@ -19,6 +19,8 @@ var (
 type CreateCommunity struct {
 	Name                         string                               `json:"name"`
 	Description                  string                               `json:"description"`
+	IntroMessage                 string                               `json:"introMessage"`
+	OutroMessage                 string                               `json:"outroMessage"`
```
0x-r4bbit: Probably makes sense to have these `omitempty` to ensure this won't break when clients create `CreateCommunity` requests that don't include these fields

services/chat/api.go
```diff
@@ -87,6 +87,8 @@ type ChannelGroup struct {
 	Admin          bool                                     `json:"admin"`
 	Verified       bool                                     `json:"verified"`
 	Description    string                                   `json:"description"`
+	IntroMessage   string                                   `json:"introMessage"`
+	OutroMessage   string                                   `json:"outroMessage"`
```
0x-r4bbit, line 91: I'm probably missing something here, but why does this need to be part of the channel group?

0x-r4bbit (commented):
This looks good!
Just a few minor comments

protocol/communities/community.go
osmaczko: I've followed the current file convention, assumed these checks are needed in some cases, see:
```
func (o *Community) Name() string {
	if o != nil &&
		o.config != nil &&
		o.config.CommunityDescription != nil &&
		o.config.CommunityDescription.Identity != nil {
		return o.config.CommunityDescription.Identity.DisplayName
	}
	return ""
}
func (o *Community) DescriptionText() string {
	if o != nil &&
		o.config != nil &&
		o.config.CommunityDescription != nil &&
		o.config.CommunityDescription.Identity != nil {
		return o.config.CommunityDescription.Identity.Description
	}
	return ""
}
func (o *Community) MembersCount() int {
	if o != nil &&
		o.config != nil &&
		o.config.CommunityDescription != nil {
		return len(o.config.CommunityDescription.Members)
	}
	return 0
}
```

services/chat/api.go
osmaczko: As far as I understand `ChannelGroup` is an abstraction unifying personal and community channel groups. This is how the new `getChats` API propagates information about communities. There was partial adaptation done already in the status-desktop, see: https://github.com/status-im/status-desktop/pull/5182

@richard-ramos please confirm if that was the intention and if these parameters fits here

0x-r4bbit: I guess my question was rather: why does the intro/outro message have to be part of the channel/chat ?
We already have it in the community description which we receive anyways, so can't we just grab it from there if we need it inside a channel (do we need it inside a channel anyways? I thought these were meant to be shown in some modal)

osmaczko: Channel group is not channel/chat, it contains chats. In other words: `ChannelGroup ~== Community`.

```
type ChannelGroup struct {
	Type           ChannelGroupType                         `json:"channelGroupType"`
	Name           string                                   `json:"name"`
	Images         map[string]images.IdentityImage          `json:"images"`
	Color          string                                   `json:"color"`
	Chats          map[string]*Chat                         `json:"chats"`
	Categories     map[string]communities.CommunityCategory `json:"categories"`
	EnsName        string                                   `json:"ensName"`
	Admin          bool                                     `json:"admin"`
	Verified       bool                                     `json:"verified"`
	Description    string                                   `json:"description"`
	IntroMessage   string                                   `json:"introMessage"`
	OutroMessage   string                                   `json:"outroMessage"`
	Permissions    *protobuf.CommunityPermissions           `json:"permissions"`
	Members        map[string]*protobuf.CommunityMember     `json:"members"`
	CanManageUsers bool                                     `json:"canManageUsers"`
	Muted          bool                                     `json:"muted"`
}
```

0x-r4bbit: Ah, sorry right.
Forget my questions!

richard-ramos:
@osmaczko What do you think of adding a condition in
- protocol/requests/create_community_request.go `func (c *CreateCommunity) Validate() error`
- protocol/requests/edit_community.go func (u *EditCommunity) Validate() error
to limit the number of characters that a description can have?

If you add it, maybe the same can be done for the community description and name? :)

protocol/communities/community.go
cammellos: In golang you can call a function on a nil point :exploding_head: 

You can check this playground https://go.dev/play/p/x2xJIKsG2Dj , basically you can call functions on a `nil` object as long as they don't refer to themselves, so the checks above make some sense. 
In theory, the caller should always check for nils, so they should never be called, and having those checks might actually not be beneficial since you might just be hiding or delaying the issue (though we don't want to panic as that crashes the whole app, we should return an error), but that's contextual, can't remember/don't know why we do it that way, but seems like if the community is nil, we should just initiate an empty struct, so those checks looks a bit pesky.

cammellos (approved):

osmaczko:
> @osmaczko What do you think of adding a condition in
> 
> * protocol/requests/create_community_request.go `func (c *CreateCommunity) Validate() error`
> * protocol/requests/edit_community.go func (u *EditCommunity) Validate() error
>   to limit the number of characters that a description can have?
> 
> If you add it, maybe the same can be done for the community description and name? :)

Sure, done! :) Also removed duplication between these two :police_officer:

osmaczko:
Waiting for https://github.com/status-im/status-react/pull/13423

protocol/communities/community.go
osmaczko: >  but that's contextual, can't remember/don't know why we do it that way  

So I will leave it as is, for now, don't want to risk any regression.

0x-r4bbit (approved):
Solid!

PR #2690, opened 2022-05-25
improve MarkAllRead performance

flexsurfer:
improve MarkAllRead performance


Samyoul (approved):
This is a bold change, but I like it.

PR #2693, opened 2022-05-27
WIP: dev: allow option to force compiling for apple silicon

stefandunca:
Adding the optional switch to satisfy the following requirements

- The desktop nim app requires to build for x86_64
- Desktop C++ app requires native support with Qt6.3+

The default is still forcing x86_64 builds on apple silicon.

Updates https://github.com/status-im/status-desktop/issues/5676

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,9 +16,13 @@ endif
 
 ifeq ($(detected_OS),Darwin)
  GOBIN_SHARED_LIB_EXT := dylib
+  # Building on M1 is still not supported, so in the meantime we crosscompile by default to amd64
   ifeq ("$(shell sysctl -nq hw.optional.arm64)","1")
-    # Building on M1 is still not supported, so in the meantime we crosscompile to amd64
-    GOBIN_SHARED_LIB_CFLAGS=CGO_ENABLED=1 GOOS=darwin GOARCH=amd64
+    USE_ARCH := amd64
+	ifneq ($(origin FORCE_ARCH),undefined)
+		USE_ARCH := $(FORCE_ARCH)
+	endif
+    GOBIN_SHARED_LIB_CFLAGS=CGO_ENABLED=1 GOOS=darwin GOARCH=$(USE_ARCH)
   endif
 else ifeq ($(detected_OS),Windows)
  GOBIN_SHARED_LIB_CGO_LDFLAGS := CGO_LDFLAGS=""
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -16,9 +16,13 @@ endif
 
 ifeq ($(detected_OS),Darwin)
  GOBIN_SHARED_LIB_EXT := dylib
+  # Building on M1 is still not supported, so in the meantime we crosscompile by default to amd64
   ifeq ("$(shell sysctl -nq hw.optional.arm64)","1")
-    # Building on M1 is still not supported, so in the meantime we crosscompile to amd64
-    GOBIN_SHARED_LIB_CFLAGS=CGO_ENABLED=1 GOOS=darwin GOARCH=amd64
+    USE_ARCH := amd64
+	ifneq ($(origin FORCE_ARCH),undefined)
+		USE_ARCH := $(FORCE_ARCH)
```
jakubgs, line 23: Why can't you just use the `USE_ARCH` variable by itself and use lazy assignment?
```Makefile
USE_ARCH ?= amd64
```
https://www.gnu.org/software/make/manual/html_node/Flavors.html

richard-ramos:
Interesting PR!
Last time i tried to use status-go compiled for arm64, I ran into an issue I reported here https://github.com/golang/go/issues/48437 . We used go 1.17. Now we're using 1.18. Is it working now?

Makefile
stefandunca: Good idea. Didn't know about the conditional variable assignment. Thanks for the heads up @jakubgs.

stefandunca:
> Interesting PR! Last time i tried to use status-go compiled for arm64, I ran into an issue I reported here [golang/go#48437](https://github.com/golang/go/issues/48437) . We used go 1.17. Now we're using 1.18. Is it working now?

I didn't test the API yet. Thanks for the great information. I will keep it WIP until I have the chance to try it in the context of the C++ APP and get back to you with feedback.

stefandunca:
> > Interesting PR! Last time i tried to use status-go compiled for arm64, I ran into an issue I reported here [golang/go#48437](https://github.com/golang/go/issues/48437) . We used go 1.17. Now we're using 1.18. Is it working now?
> 
> I didn't test the API yet. Thanks for the great information. I will keep it WIP until I have the chance to try it in the context of the C++ APP and get back to you with feedback.

@richard-ramos I just run the call test for the `callPrivateRPC()` with `{"jsonrpc":"2.0", "method": "wakuext_startMessenger", "params":[]}` payload and got the `JSON-RPC client is unavailable`. Not sure I don't have something initialized or the issue you reported with nim bindings is not reproducible with C++ runtime.

PR #2694, opened 2022-05-27
fix: handle nil descriptions in communities for chat_getChats

richard-ramos:


borismelnik (approved):
👍

0x-r4bbit (approved):
Nice and clean.

Just wondering: At this point we should always have a community description with a chat, since we're generating a `general` chat by default, no?

In which scenarios would we still run into `nil` descriptions?

richard-ramos:
> @PascalPrecht: In which scenarios would we still run into `nil` descriptions?

If you import a community with a private key, and the mailserver you're connected to does not have the community description available, or you're not connected to a mailserver at all, a community will be created, but its description will be nil. `func New(config Config) (*Community, error)` in `protocol/communities/community.go` does not assign a default value to the description

cc: @osmaczko since he was dealing with nil descriptions in another PR too

richard-ramos:
> At this point we should always have a community description with a chat, since we're generating a `general` chat by default, no?

Yes, but this is in case you are creating the community. If you're importing an existing community using its private key, there's no guarantee that the general chat will exist (it might have been deleted)

0x-r4bbit:
Thanks for clarifying!

anastasiyaig:
https://github.com/status-im/status-desktop/pull/5918 is in testing now for this @churik

elina2015:
Tested in https://ethstatus.testrail.net/index.php?/runs/view/9347&group_by=tests:status_id&group_order=asc

PR #2698, opened 2022-05-30
improve MarkAllRead performance

flexsurfer:
improve MarkAllRead performance

500ms before, 1ms after


Samyoul (approved):
That is amazing.

richard-ramos (approved):

PR #2703, opened 2022-06-08
Wallet sync for generated accounts

vitvly:
Provides sync support for generated wallet accounts (in addition to watch-only accounts already being synced).

`[Save/Delete]AccountAndPublish` functions have been removed along with publish/subscribe machinery. This has been replaced by `[Save/Delete]Accounts` funcs in `protocol.Messenger`.


protocol/messenger.go
```diff
@@ -2799,11 +2801,6 @@ func (m *Messenger) syncWallets(accs []*accounts.Account) error {
 
 	accountMessages := make([]*protobuf.SyncWalletAccount, 0)
 	for _, acc := range accs {
-		// Only sync watch type accounts
```
cammellos, line 2829: Does this support as well imported wallets through seed phrase/private key?

cammellos (commented):
Looks good, the only thing I am not sure is whether imported accounts would work? (seems like it might not, so we probably need to limit the type of accounts we sync)

vitvly:
@cammellos you're right. Obviously they shouldn't work. I'll fix this.

vitvly:
@cammellos limited the synced account types to generated/watch only

cammellos (approved):

account/accounts.go
```diff
@@ -33,6 +33,14 @@ var (
...
 )
 
+type ErrCannotLocateKeyFile struct {
+	Msg string
+}
+
+func (e ErrCannotLocateKeyFile) Error() string {
+	return e.Msg
+}
+
```
Samyoul, line 43: Implementing the error interface is a legitimate path to achieving what you are doing here. But I thought I'd also introduce you to the concept of `Error Equality` via `Error Wrapping`. See https://trstringer.com/errors-and-error-wrapping-go/#error-equality for a nice breakdown of how it works.

I'm not asking for any changes at all here, but I know you'd appreciate this feature of Go

services/accounts/accounts.go
```diff
@@ -15,21 +15,23 @@ import (
...
 	"github.com/status-im/status-go/params"
+	"github.com/status-im/status-go/protocol"
 )
 
 const pathWalletRoot = "m/44'/60'/0'/0"
 const pathDefaultWallet = pathWalletRoot + "/0"
 
-func NewAccountsAPI(manager *account.GethManager, config *params.NodeConfig, db *accounts.Database, feed *event.Feed) *API {
-	return &API{manager, config, db, feed}
+func NewAccountsAPI(manager *account.GethManager, config *params.NodeConfig, db *accounts.Database, feed *event.Feed, messenger **protocol.Messenger) *API {
```
Samyoul, line 24: I'm curious why are you passing in a pointer to pointer `messenger **protocol.Messenger`?

Samyoul (approved):
Looks good. One observation and one question 😄

account/accounts.go
vitvly: Thanks, good point! I recall skimming it over, but need to dig deeper. Didn't completely get it the first time.

services/accounts/accounts.go
vitvly: I'm sorry for this, this definitely stinks. I should have provided some explanatory comments about it. The problem is that when the `API` is initialized from within the `Service`, `initProtocol` hasn't been invoked yet, therefore there is no `Messenger`. During testing i've realized that `Init(messenger *protocol.Messenger)` occurs later than `APIs()` is invoked, and hence accounts API is initialised with `Service.messenger=nil`.
If you can suggest something, please do. Double pointer is shameful.

PR #2709, opened 2022-06-09
feat: introduce `AddBookmark`, `RemoveBookmark` and `UpdateBookmark` APIs to messenger

0x-r4bbit:
These APIs are being introduced to address https://github.com/status-im/status-go/issues/2706 and https://github.com/status-im/status-go/issues/2704, provided
that clients will move to using these APIs instead of the currently
provided equivalent APIs in the browser service.

The `bookmarks` table is being extended with a `deleted_at` field which
can later be used for garbage collection, as "removing" a bookmark is
merely soft deletion and doesn't actually remove the data from the
database.

In addition to those APIs adding and soft deleting bookmark entries,
they also automatically perform a sync operation to ensure that
bookmarks are synced in real-time (https://github.com/status-im/status-go/issues/2704).

Closes https://github.com/status-im/status-go/issues/2706, https://github.com/status-im/status-go/issues/2704

PR #2711, opened 2022-06-09
feat: add garbage collection for soft deleted bookmarks

### LINKED ISSUE #2705: Implement bookmark garbage collection
As discussed in #2704, "deletion" of browser bookmarks merely means setting their `removed` flag to `true`, resulting in them not being actually deleted.

This will create memory pressure that needs to be accounted for. To solve this, we'll implement a simple garbage collection mechanism that is run every time the messenger bootstraps. The algorithm checks for bookmarks flagged as `removed`, then checks if their `deleted_at` field is older than 14 days (to be added). If that's the case the entry gets garbage collected. 

So to sum it up we need to:

1. Add a `deleted_at` field to `bookmarks`
2. Ensure that the field is updated when `RemoveBookmark` RPC APIs are called
3. Implement the garbage collection described above and have it run during `Messenger.Start()`

0x-r4bbit:
This introduces a simple garbage collection which checks for all soft
deleted bookmarks (`removed = 1`) which have been marked for garbage
collection more than 30 days ago from the time of bootstrapping the
messenger.

Closes #2705

### DIFF
--- a/protocol/messenger_bookmarks.go
+++ b/protocol/messenger_bookmarks.go
@@ -38,3 +38,7 @@ func (m *Messenger) UpdateBookmark(ctx context.Context, oldURL string, bookmark 
 	}
 	return m.SyncBookmark(ctx, &bookmark)
 }
+
+func (m *Messenger) GarbageCollectRemovedBookmarks() error {
+	return m.persistence.DeleteSoftRemovedBookmarks(uint64(time.Now().AddDate(0, 0, -30).Unix()))
+}
--- a/protocol/persistence.go
+++ b/protocol/persistence.go
@@ -1081,3 +1081,19 @@ func (db *sqlitePersistence) UpdateBookmark(oldURL string, bookmark browsers.Boo
 	_, err = insert.Exec(bookmark.URL, bookmark.Name, bookmark.ImageURL, bookmark.Removed, bookmark.Clock, bookmark.DeletedAt, oldURL)
 	return err
 }
+
+func (db *sqlitePersistence) DeleteSoftRemovedBookmarks(threshold uint64) error {
+	tx, err := db.db.Begin()
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err == nil {
+			err = tx.Commit()
+			return
+		}
+		_ = tx.Rollback()
+	}()
+	_, err = tx.Exec(`DELETE from bookmarks WHERE removed = 1 AND deleted_at < ?`, threshold)
+	return err
+}
--- a/services/browsers/database.go
+++ b/services/browsers/database.go
@@ -136,7 +136,7 @@ type Bookmark struct {
 	ImageURL  string `json:"imageUrl"`
 	Removed   bool   `json:"removed"`
 	Clock     uint64 `json:"-"` //used to sync
-	DeletedAt uint64 `json:"deletedAt"`
+	DeletedAt uint64 `json:"deletedAt,omitempty"`
 }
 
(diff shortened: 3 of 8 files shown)


protocol/messenger_bookmarks.go
```diff
@@ -38,3 +38,8 @@ func (m *Messenger) UpdateBookmark(ctx context.Context, oldURL string, bookmark
 	}
 	return m.SyncBookmark(ctx, &bookmark)
 }
+
+func (m *Messenger) GarbageCollectRemovedBookmarks() error {
+	threshold := uint64(time.Now().AddDate(0, 0, -14).Unix())
```
cammellos: We probably want to bump this to 30 days, same time we guarantee messages to be on mailservers

0x-r4bbit: Updated!

protocol/persistence.go
```diff
@@ -1081,3 +1081,19 @@ func (db *sqlitePersistence) UpdateBookmark(oldURL string, bookmark browsers.Boo
...
+		return err
+	}
+	defer func() {
+		if err == nil {
+			err = tx.Commit()
+			return
+		}
+		_ = tx.Rollback()
+	}()
+	_, err = tx.Exec(`DELETE from bookmarks WHERE removed = 1 AND deleted_at < ?`, threshold)
```
cammellos, line 1097: Maybe we should add an index, since this will get called on login, we don't want to slow down in case someone has many bookmarks (unlikely, but who knows)

cammellos (approved):
Looks good, I would just add an index, and maybe a test? :)

0x-r4bbit:
Yes sir, on it!

protocol/persistence.go
0x-r4bbit: Added index creation to this PR!

services/browsers/database.go
```diff
@@ -136,7 +136,7 @@ type Bookmark struct {
 	ImageURL  string `json:"imageUrl"`
 	Removed   bool   `json:"removed"`
 	Clock     uint64 `json:"-"` //used to sync
-	DeletedAt uint64 `json:"deletedAt"`
+	DeletedAt uint64 `json:"deletedAt,omitempty"`
```
0x-r4bbit, line 139: I made this `omitempty` as I don't expect `mobile` to use this property yet, so it should default to `0`.

This is required, because I need the `StoreBookmark` API of this service in tests to explicitly set a `DeletedAt` value for bookmarks. This requires `StoreBookmark` to honor that property when inserting entries into the DB.

Because `mobile` uses `StoreBookmark` from browser service, we need to ensure it doesn't break with this change (hence omitempty)

0x-r4bbit:
@churik @qoqobolo Ready for testing!

churik:
@PascalPrecht https://github.com/status-im/status-go/pull/2709 and this PR just introduce new methods which are still not used on the client-side, right?

0x-r4bbit:
@churik https://github.com/status-im/status-go/pull/2709 adds new messenger APIs that are not used in mobile or desktop (yet), correct. Desktop addresses this here https://github.com/status-im/status-desktop/pull/6043

Mobile is fine with the current APIs, but if mobile wants to participate in getting real-time synchronized bookmark updates, it has to use the new messenger APIs instead of the bookmark service APIs.

**This** PR however, doesn't introduce an API that is explicitly relevant for any client to call. It's a garbage collection that, when the messenger starts, goes through all bookmarks that are flagged as `removed` and have been removed more than 30 days ago and deletes them from the database.

Samyoul (approved):

PR #2712, opened 2022-06-10
Enable contact request settings

cammellos:
Adds settings to status-go for mutual contact request.
This will want to be enabled by default on desktop.


Samyoul (commented):
This looks good. One thing do you think that this setting should be the same across all synced devices?

cammellos:
> This looks good. One thing do you think that this setting should be the same across all synced devices?

I don't think so, it's an experimental feature/feature flag toggle, so does not have to be synchronized

Samyoul (approved):

PR #2713, opened 2022-06-10
fix old messages marked as seen

flexsurfer:
fix old messages marked as seen

### DIFF
--- a/protocol/messenger_handler.go
+++ b/protocol/messenger_handler.go
@@ -1137,7 +1137,7 @@ func (m *Messenger) HandleChatMessage(state *ReceivedMessageState) error {
 		return err // matchChatEntity returns a descriptive error message
 	}
 
-	if chat.ReadMessagesAtClockValue > state.CurrentMessageState.WhisperTimestamp {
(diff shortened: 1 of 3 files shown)


protocol/messenger_handler.go
```diff
@@ -1074,9 +1074,15 @@ func (m *Messenger) HandleChatMessage(state *ReceivedMessageState) error {
 		return err // matchChatEntity returns a descriptive error message
 	}
 
-	if chat.ReadMessagesAtClockValue > state.CurrentMessageState.WhisperTimestamp {
-		receivedMessage.Seen = true
-	}
+	// This will prevent older messages to be marked as seen, currently the UI/UX behavior is unspecified,
```
Samyoul: Perhaps add a `TODO` to flag that this needs to be address in the future?

Samyoul (approved):

protocol/messenger_handler.go
flexsurfer: it didn't work, @Samyoul pls review again, thank you

Samyoul (approved):
Looks good

PR #2715, opened 2022-06-13
fix: sync clear history in real-time

### LINKED ISSUE #2714: Clearing a chat history is not synced across paired devices in real-time
Similar to bookmarks, the action of clearing a chat history is not sync across paired devices in real-time, but only when devices are explicitly synced.

This needs fixing.

0x-r4bbit:
Prior to this commit sync clear history messags were only dispatched when
devices where synced, but not when histories where actually cleared.

Closes #2714

### DIFF
--- a/protocol/messenger.go
+++ b/protocol/messenger.go
@@ -4256,8 +4256,21 @@ func (m *Messenger) DeleteMessage(id string) error {
 	return m.persistence.DeleteMessage(id)
 }
 
-func (m *Messenger) DeleteMessagesByChatID(id string) error {
-	return m.persistence.DeleteMessagesByChatID(id)
+func (m *Messenger) DeleteMessagesByChatID(ctx context.Context, id string) error {
+	err := m.persistence.DeleteMessagesByChatID(id)
+	if err != nil {
+		return err
+	}
+
+	chat := m.Chat(id)
+	if chat != nil {
+		err := m.syncClearHistory(ctx, chat)
+		if err != nil {
+			return err
+		}
+	}
+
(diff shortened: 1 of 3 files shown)


protocol/messenger.go
```diff
@@ -5642,7 +5655,7 @@ func (m *Messenger) handleSyncClearHistory(state *ReceivedMessageState, message
 		return ErrChatNotFound
 	}
 
-	if existingChat.DeletedAtClockValue >= message.ClearedAt {
+	if existingChat.DeletedAtClockValue >= message.ClearedAt && (existingChat.DeletedAtClockValue != 0 && message.ClearedAt != 0) {
```
0x-r4bbit, line 5658: This is horrible and shouldn't be needed.
I couldn't come up with a scenario however in which `DeleteAtClockValue` and `ClearedAt` were not `0`.

@cammellos @Samyoul maybe you can provide some insights here?

It's crucial we don't return here with `nil` because then the message state response won't have information about cleared histories..

0x-r4bbit:
Turns out desktop is using the wrong RPC API to clear histories. These changes here might not be needed after all. Verifying this now..

protocol/messenger.go
0x-r4bbit: Turns out desktop was not using the correct RPC API which results in those values here being `0` when a sync clear message is received.

0x-r4bbit:
Closing this as not necessary. Desktop just needs to use `ClearHistory` instead of `DeleteMessagesByID`

PR #2716, opened 2022-06-14
Add default contact request

cammellos:
If a contact update or a legacy contact request is sent, we create a
notification for the user in the activity center so it can be replied
to.
If at a later date a new contact request is received from the same user,
this will replace it, so the proper message can be displayed.


protocol/messenger_handler.go
```diff
@@ -631,6 +669,7 @@ func (m *Messenger) HandlePinMessage(state *ReceivedMessageState, message protob
 }
 
 func (m *Messenger) HandleAcceptContactRequest(state *ReceivedMessageState, message protobuf.AcceptContactRequest) error {
+	m.logger.Info("HANDLING")
```
Samyoul: Is this a left over from development?

protocol/messenger_handler.go
```diff
@@ -738,15 +777,14 @@ func (m *Messenger) HandleContactUpdate(state *ReceivedMessageState, message pro
 		contact.LastUpdated = message.Clock
 		state.ModifiedContacts.Store(contact.ID, true)
 		state.AllContacts.Store(contact.ID, contact)
-		/* Disabling for now in order to avoid duplicated contact requests in activity center
```
Samyoul, line 741: Perhaps add a `TODO` for removing or reactivating when the right circumstances have been met?

Samyoul (approved):
Lovely PR, nice comments

protocol/messenger_handler.go
cammellos: yes :)

protocol/messenger_handler.go
cammellos: wait, this is a removal, or I misunderstand your comment? i.e, I have now removed the comment and enabled the feature

Samyoul: OH, I'm so slow! I completely misread the diff. Ignore this 🤦

richard-ramos (approved):
:100:

PR #2721, opened 2022-06-20
feat(communities): Enable Community history archive

borismelnik:
Part of: https://github.com/status-im/status-desktop/issues/6098

Enable Community history archive by default

Important changes:

### DIFF
--- a/cmd/statusd/main.go
+++ b/cmd/statusd/main.go
@@ -45,7 +45,7 @@ var (
 	ipcFile                        = flag.String("ipcfile", "", "Set IPC file path")
 	pprofEnabled                   = flag.Bool("pprof", false, "Enable runtime profiling via pprof")
 	pprofPort                      = flag.Int("pprof-port", 52525, "Port for runtime profiling via pprof")
-	communityArchiveSupportEnabled = flag.Bool("community-archives", false, "Enable community history archive support")
+	communityArchiveSupportEnabled = flag.Bool("community-archives", true, "Enable community history archive support")
 	torrentClientPort              = flag.Int("torrent-client-port", 9025, "Port for BitTorrent protocol connections")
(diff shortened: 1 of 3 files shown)


services/ext/api.go
```diff
@@ -974,14 +974,6 @@ func (api *PublicAPI) GetCommunitiesSettings() ([]communities.CommunitySettings,
...
 }
 
-func (api *PublicAPI) EnableCommunityHistoryArchiveProtocol() error {
-	return api.service.messenger.EnableCommunityHistoryArchiveProtocol()
-}
-
-func (api *PublicAPI) DisableCommunityHistoryArchiveProtocol() error {
-	return api.service.messenger.DisableCommunityHistoryArchiveProtocol()
-}
-
```
0x-r4bbit, line 984: There's no particular reason to remove these.

cmd/statusd/main.go
```diff
@@ -45,7 +45,7 @@ var (
 	ipcFile                        = flag.String("ipcfile", "", "Set IPC file path")
 	pprofEnabled                   = flag.Bool("pprof", false, "Enable runtime profiling via pprof")
 	pprofPort                      = flag.Int("pprof-port", 52525, "Port for runtime profiling via pprof")
-	communityArchiveSupportEnabled = flag.Bool("community-archives", false, "Enable community history archive support")
+	communityArchiveSupportEnabled = flag.Bool("community-archives", true, "Enable community history archive support")
```
0x-r4bbit, line 48: I don't think we should do this. Mobile should not have this enabled by default

0x-r4bbit (changes requested):
I don't think we can merge this.
Community archive protocol should be disabled on mobile (for now). So enabling the protocol needs to be done for desktop nodes specifically.

Also, I don't see why the messenger RPC APIs need to be removed

PR #2730, opened 2022-06-25
Implement identity ring

### LINKED ISSUE #13334: Implement identity ring
As a user
I want other users to display their identity ring in chat
So that I can make sure they are not impersonating someone else

### Description

In the new designs, there's an identity ring around users pictures.
This should be served as an image in status-go, similarly how we display profile pictures.
Ideally there's an endpoint that serves the profile picture already encircled in an identity ring, as a single picture.

### Designs

https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=261%3A13411

### Notes

It needs to be consistent with desktop so 

1) use the same color palette (tricky thing is that according to design it differs per theme, see https://github.com/status-im/StatusQ/blob/bb9f08155fd68a0e2f1a257ee2566099c86cecd0/src/StatusQ/Core/Theme/StatusDarkTheme.qml#L165 and https://github.com/status-im/StatusQ/blob/bb9f08155fd68a0e2f1a257ee2566099c86cecd0/src/StatusQ/Core/Theme/StatusLightTheme.qml#L163)
2) use the same initalAngleRad and the direction of colors 

https://github.com/status-im/StatusQ/blob/master/src/StatusQ/Controls/StatusIdenticonRing.qml is the desktop implementation

https://github.com/status-im/status-go/blob/d0f4a94f7502ce37adf1f08065b07262b682c4bf/protocol/identity/colorhash/colorhash.go is the status-go implementation that defines how colorshash is constructed, but no rendering

qfrank:
Closes [#13334](https://github.com/status-im/status-react/issues/13334)


qfrank:
closed since we have #2788 which contains commits from this PR

PR #2731, opened 2022-06-27
Update message preview and unread message counter

jo-mut:
This PR decreases unread message counts and unread mention counts when messages are deleted


protocol/message_persistence.go
```diff
@@ -2012,6 +2012,34 @@ func (db sqlitePersistence) GetEdits(messageID string, from string) ([]*EditMess
 	return messages, nil
 }
 
+func (db sqlitePersistence) GetUnviewedCounts(id string) (messageCounts uint, mentionCounts uint, err error) {
```
cammellos: This is going to be too slow, you remember that we discussed another method in our call?

protocol/messenger_handler.go
```diff
@@ -1928,13 +1939,21 @@ func (m *Messenger) isMessageAllowedFrom(publicKey string, chat *Chat) (bool, er
...
 
-func (m *Messenger) updateUnviewedCounts(chat *Chat, mentioned bool) {
+func (m *Messenger) increaseUnviewedCounts(chat *Chat, mentioned bool) {
 	chat.UnviewedMessagesCount++
 	if mentioned {
 		chat.UnviewedMentionsCount++
 	}
 }
 
+func (m *Messenger) decreaseUnviewedMessageCounts(chat *Chat, unViewedMessagesCount uint) {
```
cammellos: This is not quite decreasing though, it's setting, it might decrease or it might not change it in case the message was already seen (which is the correct behavior), but the naming of the function is a bit off.

cammellos (changes requested):
This is going to be too slow I think on devices that have many messages, as discussed in our call the best way would be instead to:

1)  Check whether the original message is `Seen` or not (you already have the original message)
2) If the message is `Seen` -> Decrease count, otherwise don't decrease the count

That way you don't have to query the database for all the messages, which is going to be expensive.

protocol/message_persistence.go
jo-mut: yeah. I noticed its slow. The other approach will be faster as you have quoted above

cammellos (approved):

PR #2732, opened 2022-06-28
dev: allow option to force compiling for apple silicon

stefandunca:
Adding the optional switch to satisfy the following requirements

- The desktop `nim` app requires still building for `x86_64`
- Desktop C++ app requires native support with `Qt 6`

The default is still forcing `x86_64` builds on apple silicon

Follow up on [#2693](https://github.com/status-im/status-go/pull/2693)

@richard-ramos in the meantime I ran the [test](https://github.com/status-im/status-desktop/blob/9602dada11330d77be9e8a6b3be3e677efa201ac/test/libs/StatusGoQt/test_messaging.cpp#L21) that was causing problems in the past (before go 1.8) and validated that it is working and can be safely merged.

Closes #5676

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@ SET(LIB_SHARED_FILE ${CMAKE_CURRENT_SOURCE_DIR}/build/bin/libstatus${CMAKE_SHARE
 SET(LIB_HEADER_FOLDER ${CMAKE_CURRENT_SOURCE_DIR}/build/bin/)
 
 add_custom_command(OUTPUT  ${LIB_SHARED_FILE}
-                   COMMAND make statusgo-shared-library
+                   COMMAND make FORCE_ARCH=${STATUSGO_FORCE_ARCH} statusgo-shared-library
                    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
 
 add_custom_target(statusgo_shared_target DEPENDS ${LIB_SHARED_FILE} ${CMAKE_CURRENT_SOURCE_DIR})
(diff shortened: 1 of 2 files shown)


richard-ramos (approved):

CMakeLists.txt
```diff
@@ -5,9 +5,15 @@ SET(LIB_FILE ${CMAKE_CURRENT_SOURCE_DIR}/build/bin/libstatus.a)
 SET(LIB_SHARED_FILE ${CMAKE_CURRENT_SOURCE_DIR}/build/bin/libstatus${CMAKE_SHARED_LIBRARY_SUFFIX})
 SET(LIB_HEADER_FOLDER ${CMAKE_CURRENT_SOURCE_DIR}/build/bin/)
 
-add_custom_command(OUTPUT  ${LIB_SHARED_FILE}
-                   COMMAND make statusgo-shared-library
-                   WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+if(DEFINED STATUSGO_FORCE_ARCH)
+    add_custom_command(OUTPUT  ${LIB_SHARED_FILE}
+                    COMMAND make FORCE_ARCH=${STATUSGO_FORCE_ARCH} statusgo-shared-library
```
jakubgs: What would happen if we skipped the `if` and just left `FORCE_ARCH` empty?

stefandunca: @jakubgs it seems `FORCE_ARCH ?= amd64` does the job in this case also. Thanks for the heads up

jakubgs (approved):

osmaczko (approved):

PR #2736, opened 2022-06-29
feat: Add mute category methods

borismelnik:
Add muting category methods

Important changes:

Closes #https://github.com/status-im/status-desktop/issues/5975

### DIFF
--- a/protocol/communities/community_categories.go
+++ b/protocol/communities/community_categories.go
@@ -6,6 +6,22 @@ import (
 	"github.com/status-im/status-go/protocol/protobuf"
 )
 
+func (o *Community) ChatsByCategoryID(categoryID string) []string {
+	o.mutex.Lock()
+	defer o.mutex.Unlock()
+	var chatIDs []string
+	if o == nil || o.config == nil || o.config.CommunityDescription == nil {
+		return chatIDs
+	}
+
+	for chatID, chat := range o.config.CommunityDescription.Chats {
+		if chat.CategoryId == categoryID {
+			chatIDs = append(chatIDs, chatID)
+		}
+	}
+	return chatIDs
+}
+
 func (o *Community) CreateCategory(categoryID string, categoryName string, chatIDs []string) (*CommunityChanges, error) {
 	o.mutex.Lock()
 	defer o.mutex.Unlock()
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -347,6 +347,26 @@ func (m *Messenger) SetMuted(communityID types.HexBytes, muted bool) error {
 	return m.communitiesManager.SetMuted(communityID, muted)
 }
 
+func (m *Messenger) SetMutePropertyOnChatsByCategory(communityID string, categoryID string, muted bool) error {
+	community, err := m.communitiesManager.GetByIDString(communityID)
+	if err != nil {
+		return err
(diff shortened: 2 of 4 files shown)


protocol/messenger_communities.go
```diff
@@ -347,6 +347,26 @@ func (m *Messenger) SetMuted(communityID types.HexBytes, muted bool) error {
 	return m.communitiesManager.SetMuted(communityID, muted)
 }
 
+func (m *Messenger) MuteCategory(communityID string, categoryID string, muted bool) error {
```
cammellos: Maybe I would rename this method as something like `SetMutePropertyOnChats` or something similar (can't figure out a good name :) ), `MuteCategory` gave me the impression that you want to mute the chats, but you might effectively unmuting them.

cammellos (approved):
Looks good, only a comment about naming, a test would be nice :)

protocol/messenger_communities.go
Samyoul: `categoryID` is passed in as a parameter but it isn't used in this function.

How does this function mute a category if it doesn't target a category?

borismelnik: @cammellos agreed :) will fix

borismelnik: fixed :)

0x-r4bbit (approved):
LGTM!

As @cammellos said, a 1-2 tests would be wonderful. Let me know if you need help with this!

protocol/communities/community_categories.go
```diff
@@ -6,6 +6,17 @@ import (
...
+func (o *Community) ChatsByCategoryID(categoryID string) []string {
+	var chatIDs []string
+	for chatID, chat := range o.config.CommunityDescription.Chats {
+		if chat.CategoryId == categoryID {
+			chatIDs = append(chatIDs, chatID)
+		}
+	}
+
+	return chatIDs
+}
```
richard-ramos:
```suggestion
	var chatIDs []string
	if o == nil || o.config == nil || o.config.CommunityDescription == nil {
		return chatIDs 
	}
	
	for chatID, chat := range o.config.CommunityDescription.Chats {
		if chat.CategoryId == categoryID {
			chatIDs = append(chatIDs, chatID)
		}
	}

	return chatIDs
}
```

protocol/communities/community_categories.go
```diff
@@ -6,6 +6,17 @@ import (
 	"github.com/status-im/status-go/protocol/protobuf"
 )
 
+func (o *Community) ChatsByCategoryID(categoryID string) []string {
```
richard-ramos, line 9:
```suggestion
func (o *Community) ChatsByCategoryID(categoryID string) []string {
	o.mutex.Lock()
	defer o.mutex.Unlock()
```

anastasiyaig:
@borismelnik tests failed, please take a look

protocol/communities/community_categories.go
borismelnik: done :)

protocol/communities/community_categories.go
borismelnik: done :)

0x-r4bbit:
@borismelnik The tests still seems to be failing on CI

richard-ramos (approved):

PR #2742, opened 2022-07-05
feat: Add logic to re caculate route for sending a transaction based on networks disabled by the user

Khushboo-dev-cpp:
Add logic to re calculate route for sending a transaction based on networks disabled by the user

### DIFF
--- a/services/wallet/api.go
+++ b/services/wallet/api.go
@@ -323,9 +323,9 @@ func (api *API) GetTransactionEstimatedTime(ctx context.Context, chainID uint64,
 	return api.s.feesManager.transactionEstimatedTime(ctx, chainID, maxFeePerGas), nil
 }
 
-func (api *API) GetSuggestedRoutes(ctx context.Context, account common.Address, amount float64, tokenSymbol string) (*SuggestedRoutes, error) {
+func (api *API) GetSuggestedRoutes(ctx context.Context, account common.Address, amount float64, tokenSymbol string, disabledChainIDs []uint64) (*SuggestedRoutes, error) {
 	log.Debug("call to GetSuggestedRoutes")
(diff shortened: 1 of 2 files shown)


services/wallet/api.go
```diff
@@ -318,9 +318,9 @@ func (api *API) GetSuggestedFees(ctx context.Context, chainID uint64) (*Suggeste
 	return api.s.feesManager.suggestedFees(ctx, chainID)
 }
 
-func (api *API) GetSuggestedRoutes(ctx context.Context, account common.Address, amount float64, tokenSymbol string) (*SuggestedRoutes, error) {
+func (api *API) GetSuggestedRoutes(ctx context.Context, account common.Address, amount float64, tokenSymbol string, disabledNetworks []string) (*SuggestedRoutes, error) {
```
alaibe: could you maybe keep same naming + type as in other methods:
```disabledNetworks []string```
become:
```
disabledChainIDs []uint64
```

**Review by alaibe — APPROVED** [review]
Approving to unblock, but comment about disabledChainIDs is still valid

File: services/wallet/router.go [hunk 2]
```diff
@@ -71,6 +72,23 @@ func (r *Router) suggestedRoutes(ctx context.Context, account common.Address, am
 			continue
 		}
 
+		networkFound := false
+		for _, disabledNetwork := range disabledNetworks {
+			networkFound = false
+			uintVar, err := strconv.ParseUint(disabledNetwork, 10, 64)
```
alaibe: why not passing int directly?

Khushboo-dev-cpp:
resolved the review comments @alaibe :)

anastasiyaig (approved):

PR #2743, opened 2022-07-06
Reduce max keep in tests

cammellos:
TestMaxKeep is one of the slowest tests, this reduces the amount of messages so that it should take faster.
I am not sure it's the reason of the timeouts (I don't think so, there's probably something else), but it should speed things up.


ghost:
### Pull Request Checklist

0x-r4bbit (approved):

richard-ramos (approved):

PR #2752, opened 2022-07-13
feat: introduce messenger APIs to extract discord channels

0x-r4bbit:
As part of the new Discord <-> Status Community Import functionality,
we're adding an API that extracts all discord categories and channels
from a previously exported discord export file.

These APIs can be used in clients to show the user what categories and
channels will be imported later on.

There are two APIs:

1. `Messenger.ExtractDiscordCategoriesAndChannels(filesToimport
   []string) (*MessengerResponse, map[string]*discord.ImportError)`

   This takes a list of exported discord export (JSON) files (typically one per
   channel), reads them, and extracts the categories and channels into
   dedicated data structures (`[]DiscordChannel` and `[]DiscordCategory`)

   It also returns the oldest message timestamp found in all extracted
   channels.

   The API is synchronous and returns the extracted data as
   a `*MessengerResponse`. This allows to make the API available
   status-go's RPC interface.

   The error case is a `map[string]*discord.ImportError` where each key
   is a file path of a JSON file that we tried to extract data from, and
   the value a `discord.ImportError` which holds an error message and an
   error code, allowing for distinguishing between "critical" errors and
   "non-critical" errors.

2. `Messenger.RequestExtractDiscordCategoriesAndChannels(filesToImport
   []string)`

   This is the asynchronous counterpart to
   `ExtractDiscordCategoriesAndChannels`. The reason this API has been
   added is because discord servers can have a lot of message and
   channel data, which causes `ExtractDiscordCategoriesAndChannels` to
   block the thread for too long, making apps potentially feel like they
   are stuck.

   This API runs inside a go routine, eventually calls
   `ExtractDiscordCategoriesAndChannels`, and then emits a newly
   introduced `DiscordCategoriesAndChannelsExtractedSignal` that clients
   can react to.

   Failure of extraction has to be determined by the
   `discord.ImportErrors` emitted by the signal.

**A note about exported discord history files**

We expect users to export their discord histories via the
[DiscordChatExporter](https://github.com/Tyrrrz/DiscordChatExporter/wiki/GUI%2C-CLI-and-Formats-explained#exportguild)
tool. The tool allows to export the data in different formats, such as
JSON, HTML and CSV.

We expect users to have their data exported as JSON.


protocol/discord_importer/types.go
```diff
@@ -0,0 +1,46 @@
...
+
+type DiscordCategory struct {
+	ID   string `json:"id"`
+	Name string `json:"name"`
+}
+
+type DiscordExportedChannel struct {
+	Channel  DiscordChannel   `json:"channel"`
+	Messages []DiscordMessage `json:"messages"`
+}
```
0x-r4bbit: Chances are, these types will be extended with additional properties as soon as I start working on converting discord messages to Waku messages.

protocol/messenger_communities.go
```diff
@@ -1770,3 +1774,63 @@ func (m *Messenger) SyncCommunitySettings(ctx context.Context, settings *communi
...
+		if err != nil {
+			m.logger.Error("failed to extract discord channel and categories", zap.Error(err))
+			m.config.messengerSignalsHandler.ExtractDiscordCategoriesAndChannelsFailed()
+			return
+		}
+		m.config.messengerSignalsHandler.DiscordCategoriesAndChannelsExtracted(response.DiscordCategories, response.DiscordChannels, response.DiscordOldestMessageTimestamp)
+	}()
+}
+
+func (m *Messenger) ExtractDiscordChannelsAndCategories(filesToImport []string) (*MessengerResponse, error) {
```
0x-r4bbit: This function I'll probably move into `protocol/discord_importer`

protocol/messenger_communities.go
```diff
@@ -1770,3 +1774,63 @@ func (m *Messenger) SyncCommunitySettings(ctx context.Context, settings *communi
...
+
+	response := &MessengerResponse{}
```
0x-r4bbit: Theoretically, this function doesn't **have** to return a `MessengerResponse`. It just happens to be the case because I started out with a synchronous RPC API and this was the fastest way to get data across the wire.

Eventually I've added an asynchronous option, because it might take too long to do this synchronously.
So we could decide to drop the synchronous RPC API altogether, which means we could remove the newly introduced `DiscordCategories`, `DiscordChannels` and `DiscordOldestMessageTimestamp` properties from `MessengerResponse`.

@cammellos @Samyoul let me know what you think!

cammellos: In general, it's simpler if API's are synchronous, and if you need, you wrap them in an asynchronous layer through the client (that's what we do in status-react) or provide an `Async` endpoint, so I would use async endpoint sparingly

0x-r4bbit: > you wrap them in an asynchronous layer through the client (that's what we do in status-react) 

I agree... The async wrapper story in desktop is unfortunately far from ideal, so it'd be great if we could have that `Async` endpoint in status-go.

Would also rename it from `Request*` to `*Async`.

0x-r4bbit:
Added a test for `ExtractDiscordCategoriesAndChannels` API

protocol/messenger_communities.go
```diff
@@ -1773,3 +1777,82 @@ func (m *Messenger) SyncCommunitySettings(ctx context.Context, settings *communi
...
+		discordChannels = append(discordChannels, &discordExportedData)
+
+		if len(discordExportedData.Messages) > 0 && (oldestMessageTime == 0 || int(discordExportedData.Messages[0].Timestamp) <= oldestMessageTime) {
+			// Exported discord channel data already comes with `messages` being
+			// sorted, starting with the oldest, so we can safely rely on the first
+			// message
+			oldestMessageTime = int(discordExportedData.Messages[0].Timestamp)
+		}
+	}
+	return discordCategories, discordChannels, oldestMessageTime, nil
```
0x-r4bbit: I've extracted this logic into a more generic function because we need it in the next iteration where we perform the actual import (which also receives the files to import from and needs to extract the data again as well [although it might just be a subset])

protocol/protobuf/chat_message.proto
```diff
@@ -54,6 +54,22 @@ message DeleteMessage {
...
+  DiscordMessageAuthor author = 6;
+}
+
+message DiscordMessageAuthor {
+  string id = 1;
+  string name = 2;
+  string discriminator = 3;
+  string nickname = 4;
+  string avatar_url = 5;
+}
```
Samyoul: When are these protobufs used for protobuf related things?

0x-r4bbit: Already answered offline, but for transparency's sake I'll echo my response here:

As part of this PR they indeed only serve as structs. However, later on, when a new `ChatMessage` type for `DiscordMessage` is introduce, this data **has** to live in a protobuf because that's the type we rely on to save messages in the database.

I started out in fact with just simple structs but then had to move them in protobuf to be able to store data.

0x-r4bbit:
Pinging @qoqobolo this as well is ready for mobile testing! :)

richard-ramos (approved):
:100:

0x-r4bbit:
Holding back on merging this one. I'll probably revise it a little to account for what is required by latest designs.

Samyoul (approved):
Thanks for your answers, this all looks great. Another nice PR.

protocol/discord/types.go
```diff
@@ -0,0 +1,54 @@
...
+	// Critical errors are the ones that prevent the imported community
+	// from functioning properly. For example, if the creation of the community
+	// or its categories and channels fails, this is a critical error.
+	//
+	// Non-critical errors are the ones that would not prevent the imported
+	// community from functioning. For example, if the channel data to be imported
+	// has no messages, or is not parsable.
+	Code    uint   `json:"code"`
+	Message string `json:"message"`
+}
```
0x-r4bbit, line 55: @Samyoul This now introduces this new `ImportError` type that we've discussed offline.
Anything you would change here?

protocol/messenger_communities.go
```diff
@@ -1778,3 +1782,107 @@ func (m *Messenger) SyncCommunitySettings(ctx context.Context, settings *communi
 	chat.LastClockValue = clock
 	return m.saveChat(chat)
 }
+
+func (m *Messenger) ExtractDiscordDataFromImportFiles(filesToImport []string) (*discord.ExtractedData, map[string]*discord.ImportError) {
```
0x-r4bbit, line 1786: @Samyoul Notice that I've changed the signature to return a `map[string]*discord.ImportError`

This is so that we know which `filePath` caused which errors (and can be quickly looked-up `O(1)` instead of iterating over a list of errors in the client.

protocol/messenger_communities.go
```diff
@@ -1778,3 +1782,107 @@ func (m *Messenger) SyncCommunitySettings(ctx context.Context, settings *communi
...
+				// sorted, starting with the oldest, so we can safely rely on the first
+				// message
+				extractedData.OldestMessageTimestamp = int(msgTime.Unix())
+			}
+		}
+	}
+	return extractedData, errors
+}
+
+func (m *Messenger) ExtractDiscordChannelsAndCategories(filesToImport []string) (*MessengerResponse, map[string]*discord.ImportError) {
```
0x-r4bbit, line 1860: Because the underlying `ExtractDiscordDataFromImportFiles` API reutrns a `map[string]*discord.ImportError`, this API will do so as well, which is then also bubbles up to the RPC API.

As discussed offline, this works and should be fine. If you have any more thoughts or preferences here in terms of changes, please let me know

protocol/messenger_communities.go
```diff
@@ -1778,3 +1782,107 @@ func (m *Messenger) SyncCommunitySettings(ctx context.Context, settings *communi
...
+
+func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []string) {
+	go func() {
+		response, errors := m.ExtractDiscordChannelsAndCategories(filesToImport)
+		m.config.messengerSignalsHandler.DiscordCategoriesAndChannelsExtracted(
+			response.DiscordCategories,
+			response.DiscordChannels,
+			int64(response.DiscordOldestMessageTimestamp),
+			errors)
+	}()
```
0x-r4bbit, line 1887: Another change here:

Previously, this would emit s `ExtractDiscordCategoriesAndChannelsFailed` signal in case there was an error.
Now there can be `0` or `n` errors and it's semantically okay if only a subset of import files cause an error.

So instead of emitting the failure signal, we now only emit the `DiscordCategoriesAndChannelsExtracted` signal with data that can include `0` or `n` errors. It's up to the client to decide how much of a failure that is :D

0x-r4bbit:
@qoqobolo sorry to ping you again here. I've done some changes to this PR (also with new tests) and while it's not super invasive, I still think we should run this through your battery of tests once more.

protocol/messenger_communities.go
```diff
@@ -1778,3 +1782,107 @@ func (m *Messenger) SyncCommunitySettings(ctx context.Context, settings *communi
...
+
+	extractedData := &discord.ExtractedData{
+		Categories:             map[string]*discord.Category{},
+		ExportedData:           make([]*discord.ExportedData, 0),
+		OldestMessageTimestamp: 0,
+	}
+
+	errors := map[string]*discord.ImportError{}
+
+	for _, fileToImport := range filesToImport {
```
0x-r4bbit, line 1796: Oh and one thought here:

I think we can optimize this even further and spin up goroutines for each file here.
By running this concurrently we should see an increase in speed.

Would do this in another iteration though. I've run this the way it is with many discord channel files and it only took a few seconds at max already.

qoqobolo:
@PascalPrecht sure, no problem!
On it.

qoqobolo:
@PascalPrecht no issues detected by e2e.

Not sure if it needs to be re-reviewed / re-approved, so not moving it to the `Merge` column for now.
Feel free to do this when PR is ready.

protocol/messenger_communities.go
Samyoul: Yes, this is a great idea and this should probably be done in another PR. Concurrency can introduce a bunch of issues, so this is fine for now.

Samyoul:
@PascalPrecht I've gone through the changes and they all look fine to me. The only potential issue, but not on my side, is a potential inconsistency with other API methods returning only a single error as a 2nd return parameter, but I think this is fine. We pass multiple inputs via the API and receive multiple outputs, that's fine to me. we want the importing to be fast and robust.

0x-r4bbit:
Thanks @Samyoul for taking the time and giving your approval. In the absolute worst case we can change this to an `[]*discord.ImportError` later on. But having a `map[string]*discordImportError` makes it easier and faster to locate relevant errors later int he clients.

I'll merge this once CI is green

PR #2754, opened 2022-07-15
Feat/color in groupchat

endulab:
Changes in group chat:

- add color information to protobuf file
- propagate chat color (with name) during chat creation
- add possibility to change chat color

Issue #5928


services/chat/api_group_chats.go
```diff
@@ -129,6 +129,16 @@ func (api *API) RenameChat(ctx context.Context, communityID types.HexBytes, chat
 	})
 }
 
+func (api *API) EditChat(ctx context.Context, communityID types.HexBytes, chatID string, name string, color string, image string) (*GroupChatResponse, error) {
```
jrainville: This function can be moved to `chat/api.go`, since even though it only supports group chats for now, the goal is for it to support others too (probably)

services/chat/api_group_chats.go
```diff
@@ -129,6 +129,16 @@ func (api *API) RenameChat(ctx context.Context, communityID types.HexBytes, chat
...
+	if len(communityID) != 0 {
+		return nil, ErrCommunitiesNotSupported
+	}
+
+	return api.execAndGetGroupChatResponse(func() (*protocol.MessengerResponse, error) {
```
jrainville: What happens if you call this function with a public chat ID or a 1-1 chat ID? Should we add guards?

jrainville (commented):
Good job. Just two comments.

Also, don't forget to update the status-go board so QAs can test it https://github.com/status-im/status-go/projects/18#card-83919418

protocol/messenger.go
```diff
@@ -1804,6 +1804,87 @@ func (m *Messenger) ChangeGroupChatName(ctx context.Context, chatID string, name
 	return m.addMessagesAndChat(chat, buildSystemMessages([]v1protocol.MembershipUpdateEvent{event}, m.systemMessagesTranslations), &response)
 }
 
+func (m *Messenger) EditGroupChat(ctx context.Context, chatID string, name string, color string, image string) (*MessengerResponse, error) {
+	logger := m.logger.With(zap.String("site", "EditGroupChat"))
+	logger.Info("Editing group chat details", zap.String("chatID", chatID), zap.String("name", name), zap.String("color", color))
```
0x-r4bbit, line 1822: I believe, to stay consistent, we want all `logger` outputs to be lowercase

protocol/messenger.go
```diff
@@ -1804,6 +1804,87 @@ func (m *Messenger) ChangeGroupChatName(ctx context.Context, chatID string, name
...
+		event := v1protocol.NewColorChangedEvent(color, clock)
+		event.ChatID = chat.ID
+		err = signAndProcessEvent(m, &event)
+		if err != nil {
+			return nil, err
+		}
+		events = append(events, event)
+	}
+
+	// TODO add image change
```
0x-r4bbit, line 1872: This still a work in progress?

0x-r4bbit (commented):
Overall this LGTM!
Just some minor comments and I think @jrainville has some good points as well.

Also tests seem to be failing.

0x-r4bbit:
@endulab Generating the bindata files has to be part of the feature commit otherwise it won't compile

endulab:
> @endulab Generating the bindata files has to be part of the feature commit otherwise it won't compile

Ok, I will squash these commits.

protocol/messenger.go
endulab: I followed the convention.

services/chat/api_group_chats.go
endulab: Moved

services/chat/api_group_chats.go
endulab: I added guards and allow operation only for group chats.

protocol/messenger.go
endulab: Yes, the next phase is to add chat icon to group chat.

0x-r4bbit (approved):

0x-r4bbit:
@churik This is ready for mobile-client testing :)

endulab:
@PascalPrecht @jrainville please re-review.

qoqobolo:
PR is tested on the mobile side, no issues found.
Not drugging it to the `Merge` column as it still needs re-review. 
Please, feel free to move it once it's ready to be merged or ping us if retesting is needed after review.

0x-r4bbit:
@endulab One tiny thing missing, we'll have to update the version number. It's a feature, so the minor number needs to be increased by one

endulab:
> @endulab One tiny thing missing, we'll have to update the version number. It's a feature, so the minor number needs to be increased by one

@PascalPrecht done

jrainville (approved):

PR #2755, opened 2022-07-15
Dont send notification to our own devices

cammellos:
We make sure that we are not sending notification to our paired devices.


Samyoul (approved):
Looks good.

0x-r4bbit (approved):
Nice and clean.

Should we add a test that ensures that the own device does indeed not get a notification?

PR #2767, opened 2022-07-21
feat: extend `user_messages` table with discord message fields

### LINKED ISSUE #2760: Introduce new `ChatMessage` content type for `DiscordMessage`
Imported discord messages will be stored as `common.Message` and therefore a new `ChatMessage` content type needs to be introduced with a payload that looks something like this:

```
message DiscordMessage {
  uint64 timestamp = 1;
  uint64 timestamp_edited = 2;
  string content = 3;
  DiscordMessageAuthor author = 4;
}

message DiscordMessageAuthor {
  string id = 1;
  string name = 2;
  string nickname = 3;
  string avatar_url = 4;
}
```

One thing to note here is that part of that payload will be stored in a dedicated table for discord message authors (https://github.com/status-im/status-go/issues/2759), so part of the task is also to adjust existing persistence APIs to account for joining the correct data when requesting messages from the DB.

0x-r4bbit:
This extends the `user_messages` table with fields to store discord
message data. This contains:

- Discord message ID
- Discord message type
- Discord message timestamp
- Discord message timestamp_edited
- Discord message content
- Discord message author_id

Furthermore, this commit updates the messenger persistence APIs to take
the new fields into account, as well as LEFT JOIN'ing
`discord_message_authors` when user_messages with discord data are
retrieved.

To store application messages with discord data, the newly introduced
`ChatMessage` type `DiscordMessage` can be used.

The upcoming functionality to import discord communities into status
will make use of storing `ChatMessage`'s that hold `DiscordMessage`'s.

**Tasks**:
- [ ] Add tests that confirm `DiscordMessage` payload is properly stored an reconstructed

Closes #2760


protocol/protobuf/chat_message.proto
```diff
@@ -99,6 +99,7 @@ message ChatMessage {
     ImageMessage image = 10;
     AudioMessage audio = 11;
     bytes community = 12;
+    DiscordMessage discord_message = 99;
```
0x-r4bbit, line 102: @cammellos went with an arbitrary `99` here. Lemme know if this should change

protocol/migrations/sqlite/1658309863_add_discord_message_fields_to_user_messages.up.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE user_messages ADD COLUMN discord_message_id TEXT DEFAULT "";
+ALTER TABLE user_messages ADD COLUMN discord_message_type VARCHAR DEFAULT "";
+ALTER TABLE user_messages ADD COLUMN discord_message_timestamp INT;
+ALTER TABLE user_messages ADD COLUMN discord_message_timestamp_edited INT;
+ALTER TABLE user_messages ADD COLUMN discord_message_content TEXT DEFAULT "";
+ALTER TABLE user_messages ADD COLUMN discord_message_author_id TEXT DEFAULT "";
```
0x-r4bbit, line 6: This is analog to how we store other payload times.
Could be worth created an additional table for `discord_messages` and only store a `discord_message_id` here (since not every user_message will have a discord_message).

Lemme know what you think

richard-ramos: Would moving these columns to a separate table and doing a left join against such table result in a slow query to retrieve messages?

The reason I ask is because it seems to me that adding new columns to `user_messages` table for each messaging platform we want to support doesn't look elegant. If in the future we wanted to support importing messages from matrix / slack / teams / etc (just to give an example), that would mean adding 6 new columns for each platform

0x-r4bbit: > Would moving these columns to a separate table and doing a left join against such table result in a slow query to retrieve messages?

Something to measure, but definitely would prefer that!

0x-r4bbit: This is now done in https://github.com/status-im/status-go/pull/2784

0x-r4bbit:
Closing this in favour of https://github.com/status-im/status-go/pull/2784

PR #2768, opened 2022-07-21
feat: introduce API to import discord community

0x-r4bbit:
**THIS IS A WIP**

Here's what it does so far:

1. It adds a new `ImportDiscordCommunity` RPC API that takes `filesToImport []string` and a `CreateCommunityRequest`
2. It creates a community
3. It creates community categories based on extracted discord categories
4. It creates community channels based on extracted discord channels
5. It converts extracted `DiscordMessage`'s to `ChatMessage`'s and stores them in the DB

What still needs to be done:

- [ ] Importing large amounts of data sometimes loses channels (Needs investigation)
- [ ] Provide an async version as importing large data takes long
- [ ] Introduce signals so clients know progress of import
- [ ] Figure out how to handle attachements/embedds/assets (chances are this has to be done in the client)


protocol/messenger_communities.go
```diff
@@ -1869,3 +1869,149 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+
+	for _, exportedChannel := range exportedData {
+
+		communityChat := &protobuf.CommunityChat{
+			Permissions: &protobuf.CommunityPermissions{
+				Access: protobuf.CommunityPermissions_NO_MEMBERSHIP,
+			},
+			Identity: &protobuf.ChatIdentity{
+				DisplayName: exportedChannel.Channel.Name,
+				Emoji:       "😎",
```
0x-r4bbit: @John-44 We need to decide what the default emoji of an imported discord channel is going to be

protocol/messenger_communities.go
```diff
@@ -1869,3 +1869,149 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+		response.AddCommunities(r.Communities())
+
+		channelId := ""
+
+		// This looks like we keep overriding the `channelId` but in practice,
+		// this iterates only once
+		addedChatChange := r.CommunityChanges[0]
+		for chatId, _ := range addedChatChange.ChatsAdded {
+			channelId = chatId
+		}
```
0x-r4bbit, line 1945: I assume there's no better way to get that **one** value from the map, without knowing its key?

protocol/messenger_communities.go
```diff
@@ -1869,3 +1869,149 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+
+		for _, discordMessage := range exportedChannel.Messages {
+			exists, err := m.persistence.HasDiscordMessageAuthor(discordMessage.Author.GetId())
+			if err != nil {
+				return nil, err
+			}
+			if !exists {
+				err := m.persistence.SaveDiscordMessageAuthor(discordMessage.Author)
```
0x-r4bbit, line 1953: Need to check if `discordMessage.Timestamp` is >= `ImportDiscordCommunity.From`

protocol/messenger_communities.go
```diff
@@ -1869,3 +1869,149 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+
+			// Creating a `common.RawMessage` to get a `messageID`
+			rawMessage := &common.RawMessage{
+				Payload:        payload,
+				MessageType:    protobuf.ApplicationMetadataMessage_CHAT_MESSAGE,
+				Sender:         m.identity,
+				SkipEncryption: true,
+			}
+
+			messageID, err := m.sender.GetMessageID(rawMessage)
```
0x-r4bbit, line 1972: This looks like a lot of ceremony to get a message id for `common.Message`. 
Is there a leaner way to do this?

protocol/messenger_communities.go
```diff
@@ -1869,3 +1869,149 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+			if err != nil {
+				return nil, err
+			}
+
+			layout := "2006-01-02T15:04:05+00:00"
+			timestamp, err := time.Parse(layout, discordMessage.Timestamp)
```
0x-r4bbit, line 1978: This is a funny one. The discord message timestamp come with the `layout` assigned above.
I'm trying to parse it to `time` so I can get a proper unix timestamp.

If this looks like too much ceremony to get the result, please tell me if there's a better way :D

protocol/messenger_communities.go
```diff
@@ -1869,3 +1869,149 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+			if err != nil {
+				return nil, err
+			}
+			discordMessage.Timestamp = fmt.Sprintf("%d", timestamp.Unix())
```
0x-r4bbit, line 1982: Oh and here I'm changing it to actually become that unix timestamp as string. Makes it easier to deal with later.

protocol/messenger_communities.go
```diff
@@ -1869,3 +1869,149 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+				ChatMessage: protobuf.ChatMessage{
+					MessageType: protobuf.MessageType_COMMUNITY_CHAT,
+					ContentType: protobuf.ChatMessage_DISCORD_MESSAGE,
+					Clock:       m.getTimesource().GetCurrentTime(),
+					ChatId:      channelId,
+					Payload: &protobuf.ChatMessage_DiscordMessage{
+						DiscordMessage: discordMessage,
+					},
+				},
+			}
```
0x-r4bbit, line 2017: Here, I was wondering if there are any fields in `common.Message` that are super mandatory but not set here.

PR #2775, opened 2022-07-27
fix: apt -> apt-get in install_deps.sh

yqrashawn:
there's a default /usr/bin/apt cli in macos

https://discussions.apple.com/thread/252845493

### DIFF
--- a/_assets/scripts/install_deps.sh
+++ b/_assets/scripts/install_deps.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
-if [[ -x $(command -v apt) ]]; then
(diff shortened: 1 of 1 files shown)


jakubgs (approved):
Had no idea `apt` existed on MacOS, and no idea what it's for, but this is an okay fix.

yqrashawn:
> Had no idea `apt` existed on MacOS, and no idea what it's for, but this is an okay fix.

It's probably a new command added in recent macOS. I'm on 12.3.1.

@jakubgs I don't have the permission to check those checkboxes above. Can you help me with that? https://github.com/status-im/status-go/pull/2775#issuecomment-1196451280

PR #2786, opened 2022-08-05
Feat/bio social links

### LINKED ISSUE #2782: Add bio, social links to contacts and profile
The client should be able to share/receive social links and contact bio (description).

- [x] extend ChatIdentity protobuf with social links
- [x] extend Contact with social links and bio
- [x] extend Settings with social links and bio
- [x] propagate social links and bio through ContactCodeAdvertisement
- [x] handle ChatIdentity message with social links and bio
- [x] expose new endpoints for setting social links and bio

osmaczko:
Closes: https://github.com/status-im/status-go/issues/2782
Needed by: https://github.com/status-im/status-desktop/issues/5984

- [x] extend ChatIdentity protobuf with social links
- [x] extend Contact with social links and bio
- [x] extend Settings with social links and bio
- [x] propagate social links and bio through ContactCodeAdvertisement
- [x] handle ChatIdentity message with social links and bio
- [x] expose new endpoints for setting social links and bio

Additionally: 
- [x] fixed HandleChatIdentity by ignoring messages with older than already handled clock values (except for images, cause old messages may bring different image types)


multiaccounts/settings_social_links/database.go
```diff
@@ -0,0 +1,84 @@
...
+	// remove everything except static links
+	_, err = tx.Exec(`DELETE from social_links_settings WHERE link_text != ? AND link_text != ? AND link_text != ? AND link_text != ? AND link_text != ? AND link_text != ?`,
+		identity.TwitterID, identity.PersonalSiteID, identity.GithubID, identity.YoutubeID, identity.DiscordID, identity.TelegramID)
+
+	for _, link := range *links {
+		stmt, err := tx.Prepare("INSERT INTO social_links_settings (link_text, link_url) VALUES (?, ?)")
+		if err != nil {
+			return err
+		}
+		defer stmt.Close()
```
richard-ramos: Since the same SQL is used always, the prepared statement will be more efficient if moved outside the loop. Also defers within a loop might cause unintended problems!

```suggestion
	stmt, err := tx.Prepare("INSERT INTO social_links_settings (link_text, link_url) VALUES (?, ?)")
	if err != nil {
		return err
	}
	defer stmt.Close()
	for _, link := range *links {
```

multiaccounts/settings_social_links/database.go
```diff
@@ -0,0 +1,84 @@
...
+			link.URL = url.String
+		}
+		links = append(links, link)
+	}
+
+	return links, nil
+}
+
+// Be careful, it removes every row except static links (__twitter, etc.) before insertion
+func (s *SocialLinksSettings) SetSocialLinks(links *identity.SocialLinks) (err error) {
```
richard-ramos, line 51: Consider doing a nil verification, to avoid someone mistakenly calling this with a nil object and having the code fail in the `for` loop in line 68
```
if links == nil {
     return errors.New(".......")
}
```

protocol/messenger_handler.go
```diff
@@ -1870,18 +1860,45 @@ func (m *Messenger) HandleChatIdentity(state *ReceivedMessageState, ci protobuf.
...
 			if contact.Images == nil {
 				contact.Images = make(map[string]images.IdentityImage)
 			}
 			contact.Images[imageType] = images.IdentityImage{Name: imageType, Payload: image.Payload}
 
 		}
+		contactModified = true
+	}
+
+	if clockChanged {
```
richard-ramos, line 1880: Nice!

protocol/messenger_identity.go
```diff
@@ -70,3 +71,53 @@ func (m *Messenger) SetDisplayName(displayName string) error {
...
+	currentBio, err := m.settings.Bio()
+	if err != nil {
+		return err
+	}
+
+	if currentBio == bio {
+		return nil // Do nothing
+	}
+
+	// TODO: add validation
```
richard-ramos: Is this still pending?

protocol/messenger_identity.go
```diff
@@ -70,3 +71,53 @@ func (m *Messenger) SetDisplayName(displayName string) error {
...
+	currentSocialLinks, err := m.settings.GetSocialLinks()
+	if err != nil {
+		return err
+	}
+
+	if currentSocialLinks.Equals(*socialLinks) {
+		return nil // Do nothing
+	}
+
+	// TODO: add validation
```
richard-ramos: Is this still pending?

protocol/persistence.go
```diff
@@ -678,9 +741,27 @@ func (db sqlitePersistence) SaveContactChatIdentity(contactID string, chatIdenti
...
+			}
+			defer stmt.Close()
+			_, err = stmt.Exec(
+				contactID,
+				link.Text,
+				link.Url,
+			)
+			if err != nil {
+				return clockUpdated, imagesUpdated, err
+			}
```
richard-ramos:
```suggestion
	if clockUpdated && chatIdentity.SocialLinks != nil {
		stmt, err := tx.Prepare(`INSERT INTO chat_identity_social_links (chat_id, link_text, link_url) VALUES (?, ?, ?)`)
		if err != nil {
			return clockUpdated, imagesUpdated, err
		}
		defer stmt.Close()
		for _, link := range chatIdentity.SocialLinks {
			_, err = stmt.Exec(
				contactID,
				link.Text,
				link.Url,
			)
			if err != nil {
				return clockUpdated, imagesUpdated, err
			}
```

services/accounts/settings.go
```diff
@@ -7,16 +7,19 @@ import (
 	"github.com/status-im/status-go/multiaccounts/settings"
 	"github.com/status-im/status-go/nodecfg"
 	"github.com/status-im/status-go/params"
+	"github.com/status-im/status-go/protocol"
+	"github.com/status-im/status-go/protocol/identity"
 )
 
-func NewSettingsAPI(db *accounts.Database, config *params.NodeConfig) *SettingsAPI {
-	return &SettingsAPI{db, config}
+func NewSettingsAPI(messenger **protocol.Messenger, db *accounts.Database, config *params.NodeConfig) *SettingsAPI {
```
richard-ramos, line 14: `messenger` is already a pointer. Why is a pointer to a pointer necessary?

richard-ramos (changes requested):

multiaccounts/settings_social_links/database.go
```diff
@@ -0,0 +1,84 @@
...
+		// don't shadow original error
+		_ = tx.Rollback()
+	}()
+
+	// remove everything except static links
+	_, err = tx.Exec(`DELETE from social_links_settings WHERE link_text != ? AND link_text != ? AND link_text != ? AND link_text != ? AND link_text != ? AND link_text != ?`,
+		identity.TwitterID, identity.PersonalSiteID, identity.GithubID, identity.YoutubeID, identity.DiscordID, identity.TelegramID)
+
+	for _, link := range *links {
+		stmt, err := tx.Prepare("INSERT INTO social_links_settings (link_text, link_url) VALUES (?, ?)")
```
0x-r4bbit: Shouldn't the preparation happen outside of the loop?

protocol/identity/social_links.go
```diff
@@ -0,0 +1,75 @@
...
+
+func NewSocialLinks(links []*protobuf.SocialLink) *SocialLinks {
+	res := SocialLinks{}
+	for _, link := range links {
+		res = append(res, SocialLink{Text: link.Text, URL: link.Url})
+	}
+	return &res
+}
+
+func (s *SocialLinks) TransformToProtobuf() []*protobuf.SocialLink {
```
0x-r4bbit: Would just call it `ToProtobuf` but obviously that's a matter of taste..

protocol/messenger_identity.go
osmaczko: Not anymore :P

protocol/messenger_identity.go
osmaczko: Not anymore :P

services/accounts/settings.go
osmaczko: Because apparently messenger points to nil just after service creation. If it was not pointer to pointer, later init would be not reflected in settings API, resulting in a null pointer dereference and crash (I was there).

The same convention is used there: https://github.com/status-im/status-go/blob/d0e0deac952aaf10d2deafc406fe62e8b552eb2e/services/accounts/accounts.go#L24

richard-ramos: Ah! i have run into that issue too. If possible please add a comment on why is the pointer pointer necessary.
Another alternative to the pointer pointer could be to call a `Init` method in here https://github.com/status-im/status-go/blob/develop/api/geth_backend.go#L1258-L1265 .

osmaczko:
waits for https://github.com/status-im/status-mobile/pull/13795

0x-r4bbit:
@osmaczko please ensure https://github.com/status-im/status-go/pull/2786/commits/1f108030ac63e843e7447beaf9d79aee28c09476 is done in the commits where it's needed instead of adding a separate commit for fixing lint issue. Otherwise we'll introduce commits that won't pass CI

richard-ramos (approved):

0x-r4bbit (approved):

osmaczko:
Tested on desktop, no status-go related issues were found.

PR #2788, opened 2022-08-07
Use image server for all kind of images and integrate identity rings

### LINKED ISSUE #13670: Use image server for all kind of images and integrate identity rings
### Description

Currently the http server in status-go https://github.com/status-im/status-go/blob/develop/server/server.go is used only for identicon and image messages.
We should also use it for profile pictures. 

There's a branch out that also handles this case, which includes work from @qfrank:
https://github.com/status-im/status-mobile/tree/feature/identity-images-http-server
https://github.com/status-im/status-go/tree/feature/13334

This branch has a few changes:

- Adds an identity ring to the image
- Adds a signal when the port of the http server changes
- Serves user profile pictures

Some of the work integrating the branch has already been done here:

The important things to note when finishing integrating the branch:

- If the port of the server changes, the URL should of the image should be updated and re-fetched
- Status-go should not serve user images if the user has their settings set to not display profile pictures, only identicons should be server
- If this setting changes, pictures should be reloaded
- Dark/light theme should be likely passed as a parameter, so that the identity ring can be correctly set
- Identity ring should be toggled, as we don't want to show it in the old UI, only on the new UI

qfrank:
Closes [#13670](https://github.com/status-im/status-mobile/issues/13670)


qoqobolo:
Hello @qfrank, thanks for your work on this!
Is this ready to be reviewed? If so, please move PR to the `Review` column on the board and send requests to reviewers. 
We'll start testing the related status-mobile PR https://github.com/status-im/status-mobile/pull/13771 after getting approval on the go-side and moving it to the `Ready for client testing` column.
Thank you!

qfrank:
> 

hi @qoqobolo , it's ready for review :)

images/decode.go
```diff
@@ -33,6 +36,15 @@ func Decode(fileName string) (image.Image, error) {
...
 
+func DecodeFromB64Uri(b64Uri string) ([]byte, error) {
+	a := strings.Split(b64Uri, ";")
+	if len(a) != 2 || len(a[1]) <= 7 {
+		return nil, errors.New(fmt.Sprintf("b64Uri can not be parsed. b64Uri: %s", b64Uri))
+	}
+	imageB64String := strings.Split(b64Uri, ";")[1][7:]
+	return base64.StdEncoding.DecodeString(imageB64String)
+}
+
```
Samyoul: Would you be able to introduce a test for this function? 🙂

mobile/status.go
```diff
@@ -8,6 +8,9 @@ import (
 	"os"
 	"unsafe"
 
+	"github.com/status-im/status-go/extkeys"
+	"github.com/status-im/status-go/protocol/identity/ring"
```
Samyoul: The import convention (and zealously enforced by the linter 😆) in this repo is to group imports alphabetically as follows:

- standard lib
- 3rd party libs
- core libs
   - ethereum/go-ethereum
   - status-im/status-go

server/handlers.go
```diff
@@ -1,11 +1,19 @@
...
+	"image"
 	"io/ioutil"
 	"net/http"
+	"net/url"
+	"strconv"
 	"time"
 
+	"github.com/status-im/status-go/multiaccounts"
+	"github.com/status-im/status-go/protocol/identity/colorhash"
+	"github.com/status-im/status-go/protocol/identity/ring"
```
Samyoul: Same as https://github.com/status-im/status-go/pull/2788/files#r943285171

server/handlers.go
```diff
@@ -24,13 +32,169 @@ const (
...
+		w.Header().Set("Cache-Control", "no-store")
+
+		_, err = w.Write(payload)
+		if err != nil {
+			logger.Error("failed to write image", zap.Error(err))
+		}
+	}
+}
+
+func enableRing(params url.Values) bool {
```
Samyoul: `enableRing` reads like a command but returns a bool, perhaps `ringEnabled` communicates the question this function answers.

server/handlers.go
```diff
@@ -24,13 +32,169 @@ const (
...
+	addRings, ok := params["addRing"]
+	addRing := false
+	if ok && len(addRings) == 1 && addRings[0] == "1" {
+		addRing = true
+	}
+	return addRing
```
Samyoul: This could potentially be simplified to:

```suggestion
	return ok && len(addRings) == 1 && addRings[0] == "1"
```

mobile/status.go
qfrank: emm.. i thought `make lint-fix` would fix this, i was wrong, thanks for pointing out!

images/decode.go
qfrank: this function can be removed now, we don't need it anymore, since we don't want to transfer images through the bridge anymore

qfrank:
Hi @Samyoul , ready to review again :)

server/server.go
```diff
@@ -48,6 +49,7 @@ func (s *Server) listenAndServe() {
 	}
 
 	s.port = listener.Addr().(*net.TCPAddr).Port
+	signal.SendMediaServerStarted(s.port)
```
Samyoul: This is triggered inside the `Server.listenAndServe` function, this will trigger for all types of `Server` including the `PairingServer`

qfrank: Hi @Samyoul , thanks for pointing out, just made a small fix.

Samyoul (approved):
Nice PR, thank you for addressing my points so fast

PR #2792, opened 2022-08-08
feat: obtain ENSUsername address from registry

richard-ramos:
Instead of having hard coded a list of addresses per chain for the ens usernames registy, this change retrieves the address of stateofus.eth by querying the ens registry.

Also, the ens usernames on goerli expire in 2min instead of 1y


alaibe (approved):

services/ens/api.go
```diff
@@ -54,10 +58,18 @@ type API struct {
 	accountsManager *account.GethManager
 	rpcFiltersSrvc  *rpcfilters.Service
 	config          *params.NodeConfig
+
+	addrPerChain      map[uint64]common.Address
+	addrPerChainMutex sync.Mutex
+	quit              chan struct{}
```
alaibe: would ` quit chan int` works? seems simpler

anastasiyaig:
actually can't test it without https://github.com/status-im/status-desktop/issues/6921 being fixed, so will wait for @alaibe to fix the issue first and then will retest

services/ens/api.go
richard-ramos: It would work. I went for `chan struct{}` because it is a common pattern to use for signal channel, due to it not allocating memory for a value.

anastasiyaig (approved):

churik:
@richard-ramos as VERSION is bumped in `develop`, would you mind rebasing it?
We need that for making PR on mobile side otherwise our script failed with conflict.

churik:
@richard-ramos 
hello!
Builds on the mobile side are stuck on `TESTS`  [stage](https://ci.status.im/job/status-mobile/job/prs/job/android/view/change-requests/job/PR-13833/3/console):
```
14:50:54  INFO [08-22|12:50:52.334] Messenger failed to process sent messages error="Can't get raw message with id 0x936e981db9d5710ef3071d17e1dca772f2480c31f68725f403cb0b8bf80ae0ec: sql: no rows in result set"
14:50:55  ERROR[08-22|12:50:55.323] could not retrieve message               error="record not found"
```

therefore it is blocked for testing on our end.
Would you mind to take a look?

richard-ramos:
@churik taking a look, but I'm actually not sure what is that error about as these changes do not deal with messages at all, but with registering / releasing ens usernames

churik:
Comment from Jakub for context:
```
seems to me like the status-go changes cause some kind of loop or something
```
Another PR on 0.105.1 is built successfully

richard-ramos:
@churik, i added some extra logs and code for parts that might freeze. Please try again with latest commit. Hopefully it will fix the issue. Thank you

churik:
Thank you, fixed!
Tested on status-mobile on Goerli network, however, the `Release username` process is not implemented on the client side, I'll create a follow-up issue on this.

PR #2795, opened 2022-08-10
change ipfs gateway and add Mode build param

jrainville:
The Infura IPFS gateway was deprecated so all stickers are broken.

here is a hotfix , that should be temporary, where I replace the gateway with the cloudflare one for now.

I tested it in Desktop, it works, but it's super slow and also has a limit, so it's very unreliable.


richard-ramos (approved):

flexsurfer:
hey @jrainville @richard-ramos i've created new IPFS project on infura, so now we have a dedicated subdomain, we should use it, but first we need to move pins i guess https://blog.infura.io/post/migrate-your-files-to-infuras-new-ipfs-service-in-3-easy-steps#how-to-pin-ipfs-data-already-hosted-by-infura

flexsurfer:
probably we should use it only for production and keep cloudflare for develop

ipfs/ipfs.go
```diff
@@ -18,7 +18,7 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 )
 
-const infuraAPIURL = "https://ipfs.infura.io:5001/api/v0/cat?arg="
+const gatewayAPIURL = "https://cloudflare-ipfs.com/ipfs/"
```
richard-ramos:
```suggestion
const gatewayAPIURL = "https://ipfs.status.im/ipfs/"
```

osmaczko (approved):

richard-ramos (approved):

jrainville:
Thanks for the reviews and test. Sadly for you, you'll have to do it again. I improved this PR significantly.

I added the `mode` build param. That means that we can now pass if the builds `DEV` or `PROD` (or more if we want).

I coded the passing of that param in desktop, sending it from mobile should be similar.

Anyway, when we are in dev, we use the status gateway, in prod we use the infura one.

I default to dev when `mode` is undefined, let me know if you want the opposite.

churik:
@jrainville  as we need it on status-react side for release, can you please also add VERSION bump in your PR? 
Thanks!

jakubgs (changes requested):
Are we sure this is how we want to do it? Wouldn't it make more sense to do it right the first time and put it in the Node config so it can be changed via RPC by user?

services/ens/api.go
```diff
@@ -484,7 +487,11 @@ func (api *API) ResourceURL(ctx context.Context, chainID uint64, username string
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to obtain base36 representation")
 		}
-		host := str + ".ipfs.infura-ipfs.io/"
+		gatewayAPIURL := devGatewayAPIURL
+		if params.Mode == "PROD" {
```
cammellos: Best to `const` this somewhere

Makefile
```diff
@@ -41,10 +41,12 @@ AUTHOR ?= $(shell git config user.email || echo $$USER)
 ENABLE_METRICS ?= true
 BUILD_FLAGS ?= $(shell echo "-ldflags='\
 	-X github.com/status-im/status-go/params.Version=$(RELEASE_TAG:v%=%) \
+	-X github.com/status-im/status-go/params.Mode=$(mode) \
```
cammellos: I tend to prefer a specific parameter, rather than a `mode` for the whole app, i.e
params.IPFSGatewayMode, or rather just pass the URL:
params.IPFSGatewayURL, but maybe @jakubgs has a preference for one or the other, I don't have a strong opinion on it, though catchall mode makes is less explicit imo.

jakubgs:
Adding weird `mode` type arguments is just confusing. We should be instead setting this like we set the OpenSea API URL:
https://github.com/status-im/status-go/blob/fea37ff5b46ff84153591d327f4672b7d3c45915/params/config.go#L511-L514
Which is part of the `NodeConfig`:
https://github.com/status-im/status-go/blob/fea37ff5b46ff84153591d327f4672b7d3c45915/params/config.go#L454-L455
Which in mobile we normally pass via RPC when creating accounts or when changing settings:
```clojure
(def login-node-config
  {:WalletConfig (cond-> {:Enabled true}
                   (not= config/opensea-api-key "")
                   (assoc :OpenseaAPIKey config/opensea-api-key))})
```
https://github.com/status-im/status-mobile/blob/8ec2f23203dd1643205a8ff022ff4e4530b7e905/src/status_im/node/core.cljs#L63-L66
Which is passed to the node via the `Statusgo.saveAccountAndLogin` call in `saveAccountAndLogin`:
```java
    public void saveAccountAndLogin(final String multiaccountData, final String password, final String settings, final String config, final String accountsData) {
        try {
            Log.d(TAG, "saveAccountAndLogin");
            String finalConfig = prepareDirAndUpdateConfig(config, this.getKeyUID(multiaccountData));
            String result = Statusgo.saveAccountAndLogin(multiaccountData, password, settings, finalConfig, accountsData);
```
https://github.com/status-im/status-mobile/blob/526835fcf062acc35b2582a9c69af10d9ce3a67e/modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java#L323-L327

This way the setting is not passed at build time, which generates more build variations, but also allows for more easily changing such a setting by the user, if they want to specify they own URL.

jakubgs:
There is also `LoginWithConfig` function which can pass the `NodeConfig` for an already created account:
```nim
    let response = status_go.loginWithConfig($payload, hashedPassword, nodeCfgObj)
```
https://github.com/status-im/status-desktop/blob/70a1381e8606b975c829c9e63382033a23dd45ac/src/backend/accounts.nim#L254

cammellos:
> Are we sure this is how we want to do it? Wouldn't it make more sense to do it right the first time and put it in the Node config so it can be changed via RPC by user?

I think unless we allow the user to actually set it in this PR (and in the UI), it would be safer not to store it in the database, since that would be sort of a field that we would not really use, but we'd still have to migrate or it might be susceptible to changes Essentially we don't have to commit for now, while creating a migration would be fairly committal, it's easy to move from this approach to storing to the db, while it's a bit more complicated to say move from storing to not storing.
But I am honestly fine with any approach.

jakubgs:
We do the same thing for Infura RPC URL:
https://github.com/status-im/status-go/blob/ad326fa290c73a6ca5c9c07ddf607b8007111f12/params/config.go#L503-L506
It's also part of config.

jakubgs:
I understand how migrations are a pain, and that this is more of a hotfix, so I'm not saying we HAVE TO use the config.

But you should be aware that we'll have to use it EVENTUALLY if we want users to be able to change this setting.

EDIT: And I agree with @cammellos that `Mode` is a bad name, far too generic.

Makefile
jakubgs: I agree that `Mode` is far too generic of a name for a paremeter.

jakubgs: I think if we have to do it at build time `params.IPFSGatewayURL` makes more sense as it's more explicit on what it does an makes it easier for people to find and customize if they want to. `Mode` can be literally anything.

jrainville:
I agree that `Mode` is a bad name. 

So, I think the consensus for now is to just rename the params to something better like suggested and then, if needed, we can move to a node config approach?

cammellos:
sounds good

Makefile
jrainville: I used this Mode way because I thought maybe in the future, we'd have more per build mode configurations, so that's why I left it kinda generic.

jrainville:
Update done

Makefile
```diff
@@ -41,10 +41,12 @@ AUTHOR ?= $(shell git config user.email || echo $$USER)
 ENABLE_METRICS ?= true
 BUILD_FLAGS ?= $(shell echo "-ldflags='\
 	-X github.com/status-im/status-go/params.Version=$(RELEASE_TAG:v%=%) \
+	-X github.com/status-im/status-go/params.IPFSGatewayURL=$(IPFSGatewayURL) \
```
jakubgs: Call it `IPFS_GATEWAY_URL` please. And also you can set a default in the `Makefile` using the same `?=` lazy assignment so it can be overridden by environment variables when calling `make`.

jrainville: Good idea. Updated

jakubgs (approved):
Approved on the basis that this will be eventually migrated to the node config so it can be configurable by the user at runtime.

Makefile
```diff
@@ -7,6 +7,7 @@ RELEASE_DIR := /tmp/release-$(RELEASE_TAG)
 PRE_RELEASE := "1"
 RELEASE_TYPE := $(shell if [ $(PRE_RELEASE) = "0" ] ; then echo release; else echo pre-release ; fi)
 GOLANGCI_BINARY=golangci-lint
+IPFS_GATEWAY_URL ?= https://ipfs.status.im/
```
flexsurfer, line 10: is it only for dev ? where will "infura" be set for prod?

jakubgs: In the desktop/mobile repos I would assume.

flexsurfer: ah, right, i see https://github.com/status-im/status-desktop/pull/6965/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R191

richard-ramos (approved):

flexsurfer:
hey @jrainville for some reason `stickers_market` doesn't return anything in mobile with this PR

flexsurfer:
oh thats because i need to add parameter in mobile makefile ?

jrainville:
@flexsurfer with the latest build, it should default to the status Gateway if none is provided thanks to this line https://github.com/status-im/status-go/pull/2795/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R10

I'm not sure why it didn't work for you. I can test again later to see if the default value works.

Otherwise, passing the param from mobile would also work.

jrainville:
@flexsurfer I just tested by removing the param from the Status-desktop Makefile and the fault was set correctly for me.

Make sure status-go rebuilds correctly and you use the right commit maybe?

Also, the loading is a bit slower with the status gateway (dev) than the infura one, so maybe you just need to wait a little?

flexsurfer:
hey @jrainville the problem is not in loading, the problem is that `stickers_market` returns an empty value

flexsurfer:
just tested with makefile value, same, empty result

flexsurfer:
just wondering why there is no error but empty result

flexsurfer:
`stickers_market response {"jsonrpc":"2.0","id":1,"result":null}`

richard-ramos:
Very weird. I thought it was the IPFS gateway that was causing trouble, but this is something else. Is this happening with mainnet? or is it in a different chain?

flexsurfer:
mainnet

flexsurfer:
btw i used this PR for testing https://github.com/status-im/status-mobile/pull/13850

flexsurfer:
can it be something with hashes ?

jrainville:
> can it be something with hashes ?

It,s possible. The new commit hash is `e8da61d5566397ed8bd574eef52823f6df2c1846`. I amend the previous commit each time I do a change or rebase, that's why the hash changes each time.

flexsurfer:
i mean ipfs hashes

jrainville:
I don't think so. It works perfectly on Desktop. I even created an e2e test for it

flexsurfer:
```
t=2022-08-26T08:42:46+0000 lvl=warn msg="Could not retrieve stickerpack data" packID=3 error="Get \"QmbYzUKEPM3X2Yi7dXZVbNcbUz9nMJ4M6DNtpCC5kKbjh6\": unsupported protocol scheme \"\""
t=2022-08-26T08:42:47+0000 lvl=warn msg="Could not retrieve stickerpack data" packID=4 error="Get \"Qmcehb2dboemjLXDqLZ2V6Vf8w1Z1MVje9QSErERCMMUYs\": unsupported protocol scheme \"\""
t=2022-08-26T08:42:47+0000 lvl=warn msg="Could not retrieve stickerpack data" packID=5 error="Get \"QmeR9LuF8b4QG5HpsmN3gJCXbFiWy5Zki6DDYdcNSuqawn\": unsupported protocol scheme \"\""
t=2022-08-26T08:42:47+0000 lvl=warn msg="Could not retrieve stickerpack data" packID=6 error="Get \"QmWWB4D8e5iGszLe4peGuNf6pqMef4hr233EKRHq9pnbQu\": unsupported protocol scheme \"\""
t=2022-08-26T08:42:48+0000 lvl=warn msg="Could not retrieve stickerpack data" packID=7 error="Get \"QmdMRLZzE6hkdUWCZkULEpzNi5hhKQXp4fk4zxiEHV1dPz\": unsupported protocol scheme \"\""
t=2022-08-26T08:42:48+0000 lvl=warn msg="Could not retrieve stickerpack data" packID=8 error="Get \"QmWAhePgrMqKcmBMBX6rVtA1LiWQ8YgJcQ7WHbB2N4u6HD\": unsupported protocol scheme \"\""
t=2022-08-26T08:42:48+0000 lvl=warn msg="Could not retrieve stickerpack data" packID=9 error="Get \"QmUB1dD3syoU6pgF92CvvRDSgpBuS5JFKFxjFMrqAQmYLz\": unsupported protocol scheme \"\""
t=2022-08-26T08:42:49+0000 lvl=warn msg="Could not retrieve stickerpack data" packID=10 error="Get \"QmPg9DD1JaPDm5mi8XG8AAYTYtuxSseAvESQZr45V66NPk\": unsupported protocol scheme \"\""
t=2022-08-26T08:42:49+0000 lvl=warn msg="Could not retrieve stickerpack data" packID=11 error="Get \"QmRXF9LKeE7vEJhxpSRCMf9CDv7UdB1saczNTfQXYQTqU7\": unsupported protocol scheme \"\""
t=2022-08-26T08:42:49+0000 lvl=warn msg="Could not retrieve stickerpack data" packID=12 error="Get \"QmPF2C8DZCw6QonaLC9iggYrrq3JwuxKdHwU6FFeeVqhVd\": unsupported protocol scheme \"\""
t=2022-08-26T08:42:50+0000 lvl=warn msg="Could not retrieve stickerpack data" packID=13 error="Get \"QmWjtD7YjTVBM2WKSoZB6e6R8M2KMHamQy2em9Sa65jB1E\": unsupported protocol scheme \"\""
t=2022-08-26T08:42:50+0000 lvl=warn msg="Could not retrieve stickerpack data" packID=14 error="Get \"QmUKQ4HM5Qu7v1hoQSY9Q9nAPBBTZNnAFWvNvhxfaTNmvo\": unsupported protocol scheme \"\""
t=2022-08-26T08:42:50+0000 lvl=warn msg="Could not retrieve stickerpack data" packID=15 error="Get \"QmSbA2as2ShK3V5cLLj6ujrUjsy9zwSmyQhocXNab3DRQP\": unsupported protocol scheme \"\""
t=2022-08-26T08:42:50+0000 lvl=info msg="Messenger failed to process sent messages" error="Can't get raw message with id 0x1a9b1bd726530f8ede066da56a7746e442c96e78234472922de8693eed2bfe58: sql: no rows in result set"
t=2022-08-26T08:42:51+0000 lvl=warn msg="Could not retrieve stickerpack data" packID=16 error="Get \"QmZ6geozxVJR5iTBW5Uqm3UcUK7GVkcKBfoy7LvSYqSjSF\": unsupported protocol scheme \"\""
t=2022-08-26T08:42:51+0000 lvl=warn msg="STICKER PACKS :: DONE fetching "
t=2022-08-26T08:42:51+0000 lvl=warn msg="Could not retrieve stickerpack data" packID=17 error="Get \"QmZGFJBpby5QFe5Ga7e4G7nPKDkGsCTvkz7oEhoKWnPp3L\": unsupported protocol scheme \"\""
t=2022-08-26T08:42:51+0000 lvl=warn msg="Could not retrieve stickerpack data" packID=18 error="Get \"QmWP9WD7jaVxncNQp7LJfQmPzCpeAgVRV2rAKkd3M6Xh9Z\": unsupported protocol scheme \"\""
t=2022-08-26T08:42:52+0000 lvl=warn msg="Could not retrieve stickerpack data" packID=19 error="Get \"QmfEQGLCz7UJqhJ2wxqDMjc7FePAGPBJrwqZ3WYKTmww49\": unsupported protocol scheme \"\""
t=2022-08-26T08:42:52+0000 lvl=warn msg="STICKER PACKS :: PACK - " LOG15_ERROR= LOG15_ERROR="Normalized odd number of arguments by adding nil"
t=2022-08-26T08:42:52+0000 lvl=warn msg="STICKER PACKS :: PACK - " LOG15_ERROR= LOG15_ERROR="Normalized odd number of arguments by adding nil"
t=2022-08-26T08:42:52+0000 lvl=warn msg="STICKER PACKS :: PACK - " LOG15_ERROR= LOG15_ERROR="Normalized odd number of arguments by adding nil"
t=2022-08-26T08:42:52+0000 lvl=warn msg="STICKER PACKS :: ERR " LOG15_ERROR= LOG15_ERROR="Normalized odd number of arguments by adding nil"
t=2022-08-26T08:42:52+0000 lvl=warn msg="STICKER PACKS :: ERR " LOG15_ERROR= LOG15_ERROR="Normalized odd number of arguments by adding nil"
t=2022-08-26T08:42:52+0000 lvl=warn msg="STICKER PACKS :: " LOG15_ERROR= LOG15_ERROR="Normalized odd number of arguments by adding nil"
```

flexsurfer:
`log.Warn("new request", "url", params.IPFS_GATEWAY_URL+cid)` prints only `cid` so the issue in `IPFS_GATEWAY_URL`

@jakubgs could you help please

params/version.go
```diff
@@ -6,3 +6,6 @@ var Version string
 
 // GitCommit is a commit hash.
 var GitCommit string
+
+// IPFS_GATEWAY_URL is the Gateway URL to use for IPFS
+var IPFS_GATEWAY_URL string
```
jakubgs: Why is this param names using the bash env variable style while other params here are pascal-case?

Shouldn't it be something like `IpfsGatewayUrl`?

jakubgs:
@flexsurfer indeed, the issue was that in Nix builds in Mobile we don't use the Makefile and we explicitly set the variable:
```nix
{
  # Params to be set at build time, important for About section and metrics
  goBuildParams = {
    GitCommit = source.rev;
    Version = source.cleanVersion;
    # FIXME: This should be moved to status-go config.
    IPFS_GATEWAY_URL = getConfig "status-go.ipfs-gateway-url" "https://ipfs.status.im/";
  };

  # These are necessary for status-go to show correct version
  paramsLdFlags = attrValues (mapAttrs (name: value:
    "-X github.com/status-im/status-go/params.${name}=${value}"
  ) goBuildParams);
}
```
https://github.com/status-im/status-mobile/blob/aa9dd139/nix/status-go/default.nix#L16-L27

That should work, though that variable name style should be fixed...

flexsurfer:
@jrainville @richard-ramos also would be better to return an error instead of empty result in this case

Makefile
```diff
@@ -41,10 +42,12 @@ AUTHOR ?= $(shell git config user.email || echo $$USER)
 ENABLE_METRICS ?= true
 BUILD_FLAGS ?= $(shell echo "-ldflags='\
 	-X github.com/status-im/status-go/params.Version=$(RELEASE_TAG:v%=%) \
+	-X github.com/status-im/status-go/params.IPFS_GATEWAY_URL=$(IPFS_GATEWAY_URL) \
```
jakubgs: Purely an aesthetic nitpick, but putting it between `Version` and `GitCommit` is kinda weird.

jakubgs: I will take the liberty of fixing this before merging.

params/version.go
jakubgs: I will take the liberty of fixing this before merging.

jakubgs: Even the linter is on my side:
```
params/version.go:11:5: don't use ALL_CAPS in Go names; use CamelCase (golint)
var IPFS_GATEWAY_URL string
    ^
```

jakubgs:
@richard-ramos said he will add the error on empty URL in a separate PR.

jakubgs:
Actually, the fix is simple, so I applied it myself:
```diff
diff --git a/services/stickers/api.go b/services/stickers/api.go
index 759495bd7..db879d167 100644
--- a/services/stickers/api.go
+++ b/services/stickers/api.go
@@ -285,6 +285,7 @@ func (api *API) fetchStickerPacks(chainID uint64, resultChan chan<- *StickerPack
                        stickerPack, err := api.fetchPackData(stickerType, packID, true)
                        if err != nil {
                                log.Warn("Could not retrieve stickerpack data", "packID", packID, "error", err)
+                               errChan <- err
                                return
                        }
```
Thanks @richard-ramos.

PR #2796, opened 2022-08-11
feat: make kdf iterations configurable when creating a multiaccount

richard-ramos:
In `SaveAccount` and `SaveAccountAndLogin`, the `account` json (the first parameter) can have an optional `kdfIterations` attribute which should contain the number of PBKDF2 iterations. If this attribute is not specified, the default value `3200` is used (which is what we currently use for encrypting the database. 

On desktop we should try using the default value of SQLCipher: `256000` and see the performance, and if it's not adequate, tune it until an acceptable value is reached. (i.e. [NIST](https://pages.nist.gov/800-63-3/sp800-63b.html) recommends at least 10000 iterations).

**NOTE**: This change only affects the number of KDF iterations used when creating a new database. I tried changing existing databases but ended up with `out of memory` errors. Perhaps the DB must be exported unencrypted, and then reencrypted with the new encryption settings?

**NOTE for mobile testing**: the following scenarios must be tested:
- Login with an existing account
- Login after changing the password
- Exporting a database
- New account creation
- Creating an account with seed phrase
- Creating an account with keycard


jrainville (approved):
Looks good to me

qoqobolo:
Hi @richard-ramos ! Can you rebase your PRs, please? As VERSION was bumped to 0.105.1, most PRs have conflicts now.

richard-ramos:
Rebased!

stefandunca (approved):

PR #2797, opened 2022-08-12
feat: Handle image in group chat

endulab:
Add image_payload column to chats table.
Add Base64Image to chat struct.
Add ImageChange event to propagate image.
Change EditChat API - use CroppedImage.
Process and crop image in EditGroupChat.

Technical details:
The image is kept in database as raw bytes.
The image is propagated by events as raw bytes also.
Chat structure keeps image in base64 format "data:image/{type};base64,{data}"

Required by https://github.com/status-im/status-desktop/issues/6466


protocol/persistence.go
```diff
@@ -136,13 +137,21 @@ func (db sqlitePersistence) saveChat(tx *sql.Tx, chat Chat) error {
...
+	stmt, err := tx.Prepare(`INSERT INTO chats(id, name, color, emoji, active, type, timestamp,  deleted_at_clock_value, unviewed_message_count, unviewed_mentions_count, last_clock_value, last_message, members, membership_updates, muted, invitation_admin, profile, community_id, joined, synced_from, synced_to, description, highlight, read_messages_at_clock_value, received_invitation_admin, image_payload)
+	    VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?,?, ?,?,?,?,?,?,?,?,?,?,?,?)`)
 	if err != nil {
 		return err
 	}
 	defer stmt.Close()
 
+	var imagePayload []byte
+	if len(chat.Base64Image) > 0 {
+		imagePayload, err = userimage.GetPayloadFromURI(chat.Base64Image)
```
jrainville, line 150: I'm curious, how does the other image saves work? I see you added this new function to get the Payload, but maybe we should use the same way others did, since they do a resize as well?

jrainville (commented):
Looks good. I just had a question about reusing the code to resize the images that profile image does. Basically, are you already reusing that? I'm too lazy to check myself, so I just ask directly haha.

protocol/persistence.go
richard-ramos: Could `chat.imagePayload` be populated with the image `[]byte` before calling this function? that way this function shouldnt have to call `userimage.GetPayloadFromURI` but just insert the imagePayload value directly (Something +- similar is done when saving messages containing images. We deal with the []bytes directly)

richard-ramos (approved):

protocol/persistence.go
endulab: In case of images in messages, there are 3 columns: payload, file type, base64. I added only base64 column because this is the data which is only used. Payload and filetype seemed to be redundant. However, if I add them (payload and optionally file type) I can remove GetPayloadFromURI. I do not have a strong opinion about it. If GetPayloadFromURI is too performace-expensive I can add another field.

cammellos (approved):

endulab:
@qoqobolo can you test the mobile part ? The functionality is handled only in desktop. There should be no regressions.

qoqobolo:
Heya @endulab, sure, will launch the tests.

Also, just in case remind (not sure if everyone has seen [this message](https://discord.com/channels/624205794384281629/862717376915308544/1011576643040583731)) that now you can ping us using these keywords `@status-im/mobile-qa`  and `@status-im/desktop-qa`, so we can take PR to work faster. But personal pings are absolutely fine too :)

protocol/v1/membership_update_message_test.go
```diff
@@ -223,6 +232,20 @@ func TestGroupValidateEvent(t *testing.T) {
...
 		},
+		{
+			Name:   "image-changed allowed because from is admin",
+			From:   "0xabc",
+			Group:  createGroup([]string{"0xabc"}, nil),
+			Event:  NewImageChangedEvent([]byte{1, 2, 3}, 0),
+			Result: true,
+		},
+		{
+			Name:   "image-changed not allowed for non-admins",
```
osmaczko, line 243: That's wrong. Every member should be able to change the name, image, and color of the group. This comment is just to raise awareness, no action is required. I am already resolving it in a separate issue.

PR #2801, opened 2022-08-17
Issue 13813 -- Use messenger ns for browser CRUD

ibrkhalil:
Use the messenger NS for CRUD operations on browsers
Closes #13813


ibrkhalil:
@cammellos 
I removed tests from the `api_tests.go` namespace, should they go to `messenger_test.go`?

0x-r4bbit:
Definitely need to reintroduce tests otherwise it feels super uncomfortable to land these changes :D

ibrkhalil:
> Definitely need to reintroduce tests otherwise it feels super uncomfortable to land these changes :D
@PascalPrecht
Can you please check?

richard-ramos (approved):
LGTM, although I don't understand the purpose of moving the browser API to messenger.

ibrkhalil:
> LGTM, although I don't understand the purpose of moving the browser API to messenger.

@cammellos Told me that we need to move the whole messenger logic inside the messenger namespace I think

0x-r4bbit:
Only thing left: I think we don't need 7 commits for this. If you could squash them into a single one that moves the APIs around, that'd be great (also assuming that tests will be green of course).

@richard-ramos  The move in itself probably doesn't add any value. However, if we plan to perform `sync` operations as part of these APIs they have to live in `Messenger`.

0x-r4bbit:
@ibrkhalil You usually never need to merge `develop` into any of your branches. What you rather do is rebase your commits on top of `develop`.

Also, next time, please squash your commits. Commits like:
Screenshot from 2022-08-24 16-15-27
Are not very meaningful. Plus we now have a bunch of commits in the history that don't even pass the linter

ibrkhalil:
> @ibrkhalil You usually never need to merge `develop` into any of your branches. What you rather do is rebase your commits on top of `develop`.
> 
> Also, next time, please squash your commits. Commits like: Screenshot from 2022-08-24 16-15-27 Are not very meaningful. Plus we now have a bunch of commits in the history that don't even pass the linter

Oh, Apologies for that
Do you want me to revert it and fix it ?

0x-r4bbit:
The commits have already landed. I'd avoid force pushing into it.
Maybe just keep an eye on it in the future re: commit hygiene.

PR #2809, opened 2022-08-19
added function which creates an account from the provided mnemonic

saledjenic:
Added function creates an account from the provided mnemonic

### DIFF
--- a/account/generator/generator.go
+++ b/account/generator/generator.go
@@ -100,6 +100,23 @@ func (g *Generator) ImportJSONKey(json string, password string) (IdentifiedAccou
 	return acc.ToIdentifiedAccountInfo(id), nil
 }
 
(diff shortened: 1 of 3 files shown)


richard-ramos (approved):
I think it's okay but would like for @cammellos to give his blessings to this PR :)

0x-r4bbit:
Can we have a test for this?

Samyoul (approved):
This PR is fine, as long as all you want to do is get only the master key generated from the mnemonic without doing anything else with the key.

Basically `ImportMnemonic` but no storing in memory

saledjenic:
> Can we have a test for this?

@PascalPrecht test is added.

0x-r4bbit (approved):

PR #2810, opened 2022-08-19
chore: hardcode optimism testnet for retrieving communities portal

richard-ramos:
It does not require mobile testing as it is only used on Desktop

### DIFF
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -210,25 +210,9 @@ func (m *Messenger) JoinedCommunities() ([]*communities.Community, error) {
 }
 
 func (m *Messenger) CuratedCommunities() (*communities.KnownCommunitiesResponse, error) {
-	/*testNetworksEnabled, err := m.settings.TestNetworksEnabled()
-	if err != nil {
-		return nil, err
-	}
+	// Revert code to https://github.com/status-im/status-go/blob/e6a3f63ec7f2fa691878ed35f921413dc8acfc66/protocol/messenger_communities.go#L211-L226 once the curated communities contract is deployed to mainnet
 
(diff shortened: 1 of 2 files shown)


protocol/messenger_communities.go
```diff
@@ -208,7 +210,7 @@ func (m *Messenger) JoinedCommunities() ([]*communities.Community, error) {
 }
 
 func (m *Messenger) CuratedCommunities() (*communities.KnownCommunitiesResponse, error) {
-	testNetworksEnabled, err := m.settings.TestNetworksEnabled()
+	/*testNetworksEnabled, err := m.settings.TestNetworksEnabled()
```
0x-r4bbit: Shall we remove this then?

protocol/messenger_communities.go
```diff
@@ -222,6 +224,34 @@ func (m *Messenger) CuratedCommunities() (*communities.KnownCommunitiesResponse,
 	if err != nil {
 		return nil, err
 	}
+	*/
+
+	// --- Delete this code once we have the curated communities contract in mainnet
+	// Since we are bypassing tesnet being enabled or not, we gotta obtain the networks from the node config
+	chainID := uint64(69)
```
0x-r4bbit: Oh , this optimisms chain ID? Maybe add a comment for that?

0x-r4bbit (approved):

qoqobolo:
> It does not require mobile testing as it is only used on Desktop

Got it. Will mark it as `tested` for better visibility

protocol/messenger_communities.go
richard-ramos: yeah, will remove it and add a comment containing the commit that contained the code before the change so it's easier to revert without having to browse the commit history

PR #2813, opened 2022-08-23
distribute information about leaving community

### LINKED ISSUE #2812: Distribute information about leaving community
# Problem

Currently leaving the community is only local to the user. The user should advertise it is no longer part of the community so that community's member list can be updated accordingly.

# Implementation

- Introduce `RequestToLeave` (similar to `RequestToJoin`) and distribute it in `LeaveCommunity`
- Handle `RequestToLeave` as admin and update the members list

# Notes

All changes to the community will be reflected only when admin processes community requests, that is: `RequestToJoin` and `RequestToLeave`, which implies the admin has to be online in order to see members list update by others.

## Considerations
Implementation should take into account the order of requests arrival and compare clocks before any further actions. Example scenarios:

--admin is offline--
a) t(1): user A leaves the community
b) t(2): user A sends request to join (aka rejoin)
--admin is online--

scenario A - message a) arrives first
1) remove the user from the community
2) add the user back to the community

scenario B - message b) arrives first
1) add the user to the community even though it is already on the member list (aka positive request to join response)
2) ignore user removal, because t(2) > t(1)

osmaczko:
closes: https://github.com/status-im/status-go/issues/2812

- [feat: introduce and distribute RequestToLeave community](https://github.com/status-im/status-go/commit/f0196b257eaf712994de527a4e346533bd0fd225)
- [chore: add tests for leaving and rejoining community](https://github.com/status-im/status-go/commit/41617ee4395a538764b8f1a39bddb37190968ded)


osmaczko:
witing for: https://github.com/status-im/status-mobile/pull/13847

richard-ramos (approved):

protocol/communities/manager.go
```diff
@@ -876,7 +871,11 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request
 		return nil, err
 	}
 
-	if community.AcceptRequestToJoinAutomatically() {
+	// If user is already a member, then accept request automatically
+	// It may happen when member removes itself from community and then tries to rejoin
+	// More specifically, CommunityRequestToLeave may be delivered later than CommunityRequestToJoin, or not delivered at all
+	acceptAutomatically := community.AcceptRequestToJoinAutomatically() || community.HasMember(signer)
+	if acceptAutomatically {
```
0x-r4bbit, line 878: Just to raise awareness, this will trigger a `AcceptRequestToJoinCommunity` signal, which might not be wanted, given that with this condition, the requester is already a member.

0x-r4bbit (approved):
Great work!

protocol/communities/manager.go
osmaczko: Thanks, I am aware of that. As stated in the comment above, this logic is expected. If the user is a member and asks to join the community, the response should be positive, because we can't know if `RequestToLeave` wasn't sent before that.

See "Considerations" section from the issue description: https://github.com/status-im/status-go/issues/2812

0x-r4bbit: Correct. Just wanted to make sure we double check in clients (Desktop) that `AcceptRequestToJoin` won't cause funny results.

Feel free to ignore this :)

osmaczko: What you say is hard to reproduce, because once you join the community the UI doesn't give you the ability to join again, unless you leave. The basic flow (join, leave, join) was tested both on Desktop and Mobile. Hope there is no any edge case I've missed, but if there is, it will be fixed as a separate issue I guess.

PR #2815, opened 2022-08-24
fix(messenger): don't remove inactive chats when syncing

0x-r4bbit:
This commit ensures we're relying on `chat.DeletedAtClockValue` instead
of `chat.Active` to know whether or not we need to remove the chat from
paired devices.

Because we were relying on `Active != true`, we ended up with a serious
bug that would result in deactivating all chats on paired devices.

The reason the chats would disappear on paired devices is because, when
setting up a new device by importing a seedphrase, chances are this
device will receive `HandleBackUp` signals (which originate from other
devices with the same account that backed up contacts etc).

When backups are handled, we create chats for every contact that's part
of the backup signal. Those chats are set to `Active = false` because
the signal handling shouldn't cause those chats to show up in the UI.

However, because those are set to `Active = false`, the next time the
user tries to sync from this devices, all those chats are considered as
"removed", hence sending "chat removed" signals when syncing (which then
causes those chats to disappear on all paired devices.

We need to rely on `DeletedAtClockValue` to know whether a chat was
indeed removed and only then emit such a signal.

### DIFF
--- a/protocol/messenger.go
+++ b/protocol/messenger.go
@@ -2865,7 +2865,7 @@ func (m *Messenger) SyncDevices(ctx context.Context, ensName, photoPath string) 
 			}
 		}
 
-		if (isPublicChat || chat.OneToOne() || chat.PrivateGroupChat()) && !chat.Active {
+		if (isPublicChat || chat.OneToOne() || chat.PrivateGroupChat()) && !chat.Active && chat.DeletedAtClockValue > 0 {
 			pending, err := m.persistence.HasPendingNotificationsForChat(chat.ID)
 			if err != nil {
 				return false
--- a/go.mod
+++ b/go.mod
@@ -259,6 +259,7 @@ require (
 	golang.org/x/time v0.0.0-20220224211638-0e9765cccd65 // indirect
 	golang.org/x/tools v0.1.11 // indirect
 	golang.org/x/xerrors v0.0.0-20220609144429-65e65417b02f // indirect
+	gopkg.in/go-playground/assert.v1 v1.2.1 // indirect
 	gopkg.in/natefinch/npipe.v2 v2.0.0-20160621034901-c1b8fa8bdcce // indirect
 	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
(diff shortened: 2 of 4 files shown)


richard-ramos (approved):
:medal_military: :100:

cammellos (approved):

Samyoul (approved):

protocol/messenger.go
```diff
@@ -2846,7 +2846,7 @@ func (m *Messenger) SyncDevices(ctx context.Context, ensName, photoPath string)
 			}
 		}
 
-		if (isPublicChat || chat.OneToOne() || chat.PrivateGroupChat()) && !chat.Active {
+		if (isPublicChat || chat.OneToOne() || chat.PrivateGroupChat()) && !chat.Active && chat.DeletedAtClockValue > 0 {
```
0x-r4bbit, line 2868: @cammellos I've extended the condition here instead of replacing it (`!Active && DeletedAtClockValye > 0`) to be sure.
Also, the `Clock` used in `SyncChatRemoved` message, I've reverted to being the `clock` it was before (instead of the `DeletedAtClockValue`).

Reason being, one of the tests is failing and I'm no longer sure if it's really right to replace the clock.

Will push another commit to demo it

protocol/messenger.go
```diff
@@ -3167,7 +3167,7 @@ func (m *Messenger) syncChatRemoving(ctx context.Context, id string) error {
 	clock, chat := m.getLastClockWithRelatedChat()
 
 	syncMessage := &protobuf.SyncChatRemoved{
-		Clock: clock,
+		Clock: chat.DeletedAtClockValue,
```
0x-r4bbit: This one causes a test to fail 

```golang
func (s *MessengerInstallationSuite) TestSyncInstallation() {
```

So not sure if it's really correct to use this clock here

protocol/messenger.go
0x-r4bbit: See here: https://github.com/status-im/status-go/pull/2815/files#r953943414

cammellos: I see why, it's because we don't set that field on public chats, only on one to ones, which is a bit annoying. We can address in a separate issue I suppose

PR #2821, opened 2022-08-25
feat: allow passing variables to docker-compose and fix var name

richard-ramos:


_assets/compose/bootnode/Makefile
```diff
@@ -72,7 +72,7 @@ keys:
 
 start: keys/nodekey keys/nodeaddr
 	@echo " * $(GRN)Starting '$(CONTAINER_NAME)' container...$(RST)"
-	docker-compose up -d
+	docker-compose $(COMPOSE_START_FLAGS) up -d
```
jakubgs: If anythin it should be `COMPOSE_UP_FLAGS` since they are for `docker-compose up` call.

jakubgs (approved):
Looks good, but I'd mention the variable in the `README`. After the container ones.

jakubgs (changes requested):
You didn't update the bootnode readme.

jakubgs (approved):

PR #2826, opened 2022-08-26
feat(messenger): introduce rudimentary discord import tool

0x-r4bbit:
**NOTE: I've put up 2 commits here, the second one is just for `make generate` and will be squashed after review is done. Only the first one needs to be reviewed.**

This commit aims to implement the first parts of the discord import
tool. It's not complete as there are more things that need to be handle
by it, but as of this commit, it introduces the basic functionality
that:

1. Let's clients request a discord import
2. Creates a community + its categories/channels derived from import
   data
3. Converts discord messages to users messages and saves them to the DB
4. Downloads discord message author avatar assets and stores them as
   base64 encoded payloads
5. Initializes the imported community by setting up filters and
   notifications
6. During all of the above, it sends progress updates to clients so they
   can show users meaningful UIs


appdatabase/migrationsprevnodecfg/bindata.go
```diff
@@ -137,7 +137,7 @@ func _0001_appDownSql() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "0001_app.down.sql", size: 356, mode: os.FileMode(0644), modTime: time.Unix(1644328579, 0)}
+	info := bindataFileInfo{name: "0001_app.down.sql", size: 356, mode: os.FileMode(0664), modTime: time.Unix(1660816022, 0)}
 	a := &asset{bytes: bytes, info: info, digest: [32]uint8{0xb5, 0x25, 0xa0, 0xf8, 0x7d, 0x2d, 0xd, 0xcf, 0x18, 0xe4, 0x73, 0xc3, 0x95, 0xf5, 0x24, 0x20, 0xa9, 0xe6, 0x9e, 0x1d, 0x93, 0xe5, 0xc5, 0xad, 0x93, 0x8f, 0x5e, 0x40, 0xb5, 0x30, 0xaa, 0x25}}
```
Macias131987, line 141: Fernando macias

Macias131987 (approved):
Fernando macias Saldana

protocol/discord/assets.go
```diff
@@ -0,0 +1,32 @@
...
+	encBuf := make([]byte, maxEncLen)
+	e64.Encode(encBuf, payload)
+
+	mime, err := images.GetMimeType(payload)
+	if err != nil {
+		return nil, nil, "", err
+	}
+	imageBase64 := fmt.Sprintf("data:image/%s;base64,%s", mime, encBuf)
+	return payload, &imageType, imageBase64, nil
+}
```
0x-r4bbit, line 32: This function will be used to download discord message author avatar assets.
Right now it returns there actual image payload + the type + the base64 string, but as of now I'm only storing the base64 string (don't think we need both, payload + base64).

However I've noticed we store payload + base64 for message images as well. Shall I do the same for discord author images?

protocol/discord/types.go
```diff
@@ -1,17 +1,24 @@
...
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/status-im/status-go/protocol/protobuf"
 )
 
-type ErrorType uint
+type ErrorCodeType uint
```
0x-r4bbit, line 10: Renamed this so we can have `WarningType` and `ErrorType` as enum below.
Reason I want those for the enum below is so we can have `Error()` and Warning` as function types.

protocol/discord/types.go
```diff
@@ -57,10 +92,102 @@ type ImportError struct {
...
+		Code:    ErrorType,
+	}
+}
+
+func Warning(message string) *ImportError {
+	return &ImportError{
+		Message: message,
+		Code:    WarningType,
+	}
+}
```
0x-r4bbit, line 115: Here are `Error()` and `Warning()`

protocol/messenger_communities.go
```diff
@@ -1832,29 +1833,20 @@ func (m *Messenger) ExtractDiscordDataFromImportFiles(filesToImport []string) (*
 		filePath := strings.Replace(fileToImport, "file://", "", -1)
 		bytes, err := os.ReadFile(filePath)
 		if err != nil {
-			errors[fileToImport] = &discord.ImportError{
-				Code:    2,
-				Message: err.Error(),
-			}
+			errors[fileToImport] = discord.Error(err.Error())
```
0x-r4bbit, line 1838: Here's how `Error()` and Warning()` are used as convenience APIs

protocol/messenger_communities.go
```diff
@@ -1921,3 +1914,389 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+			if totalMessageCount == 0 {
+				importError.Message = discord.ErrNoMessageData.Error()
+			}
+			importProgress.AddTaskError(discord.CommunityCreationTask, importError)
+			importProgress.StopTask(discord.CommunityCreationTask)
+			m.publishImportProgress(importProgress)
+			return
+		}
+
+		importProgress.UpdateTaskProgress(discord.CommunityCreationTask, 0.5)
```
0x-r4bbit, line 1954: Setting the progress to `0.5` is somewhat arbitrary here. I split up the `CommunityCreationTask` into a few parts and extracting the data accounts for `50%` of the work.

Happy to change this to something more reasonable.

protocol/messenger_communities.go
```diff
@@ -1921,3 +1914,389 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+		if err != nil {
+			m.cleanUpFailedImport(discordCommunity)
+			importProgress.AddTaskError(discord.CommunityCreationTask, discord.Error(err.Error()))
+			importProgress.StopTask(discord.CommunityCreationTask)
+			m.publishImportProgress(importProgress)
+			return
+		}
+
+		importProgress.CommunityID = discordCommunity.IDString()
+		importProgress.UpdateTaskProgress(discord.CommunityCreationTask, 0.75)
```
0x-r4bbit, line 1985: Same thing here. `0.75` is arbitrary. Happy to change this. (but also: in reality this happens so fast, that these updates are barely noticable in the UI)

protocol/messenger_communities.go
```diff
@@ -1921,3 +1914,389 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+			}
+			discordCommunity = communityWithCategories
+
+			// This looks like we keep overriding the same field but there's
+			// only one `CategoriesAdded` change at this point.
+			for _, addedCategory := range changes.CategoriesAdded {
+				processedCategoriesIds[category.ID] = addedCategory.CategoryId
+			}
+			m.publishImportProgress(importProgress)
+		}
```
0x-r4bbit, line 2022: Notice how the categories are created sequentially. Theoretically, we can do this in parallel, but judging by what I've experienced using real data, this is rather fast. Anyways, if anyone thinks it's worth it to pursue I'm happy to do that in a follow up PR. Right now I don't see the need.

protocol/messenger_communities.go
```diff
@@ -1921,3 +1914,389 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+
+		processedChannelIds := make(map[string]string, 0)
+		processedMessages := make([]*common.Message, 0)
+		var discordMessagesToSave []*protobuf.DiscordMessage
+		var chatsToSave []*Chat
+		authorAvatarUrls := map[string]string{}
+		// messageAssetsToDownload := map[string][]*protobuf.DiscordMessageAttachment{}
```
0x-r4bbit: Whoops, left over. Need to remove this.

protocol/messenger_communities.go
```diff
@@ -1921,3 +1914,389 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+							DiscordMessage: discordMessage,
+						},
+					},
+				}
+				discordMessagesToSave = append(discordMessagesToSave, discordMessage)
+				processedMessages = append(processedMessages, messageToSave)
+			}
+			importProgress.UpdateTaskProgress(discord.ImportMessagesTask, float32(len(processedMessages))/float32(totalMessageCount)*0.90)
+			m.publishImportProgress(importProgress)
+		}
```
0x-r4bbit, line 2167: Processing the channels however, I think we can definitely do in parallel. Something I'd like to explore as optimization once the core functionality has landed.

protocol/messenger_communities.go
```diff
@@ -1921,3 +1914,389 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+						ChatId:      processedChannelIds[channel.Channel.ID],
+						Payload: &protobuf.ChatMessage_DiscordMessage{
+							DiscordMessage: discordMessage,
+						},
+					},
+				}
+				discordMessagesToSave = append(discordMessagesToSave, discordMessage)
+				processedMessages = append(processedMessages, messageToSave)
+			}
+			importProgress.UpdateTaskProgress(discord.ImportMessagesTask, float32(len(processedMessages))/float32(totalMessageCount)*0.90)
```
0x-r4bbit, line 2165: Oh here, I'm multiplying by `0.9` so the progress bar goes to 90%. Reason being, we still need to save the processed messages which will take time. So we don't want it to be 100% yet.

protocol/messenger_communities.go
```diff
@@ -1921,3 +1914,389 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+
+		err = m.persistence.SaveDiscordMessages(discordMessagesToSave)
+		if err != nil {
+			m.cleanUpFailedImport(discordCommunity)
+			importProgress.AddTaskError(discord.ImportMessagesTask, discord.Error(err.Error()))
+			importProgress.StopTask(discord.ImportMessagesTask)
+			m.publishImportProgress(importProgress)
+			return
+		}
+		importProgress.UpdateTaskProgress(discord.ImportMessagesTask, 0.95)
```
0x-r4bbit, line 2177: Same here. Arbitrary value, as we don't get updates in between save calls.

protocol/messenger_communities.go
```diff
@@ -1921,3 +1914,389 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+		filters, err := m.transport.InitPublicFilters(filterChatIds)
+		if err != nil {
+			m.cleanUpFailedImport(discordCommunity)
+			importProgress.AddTaskError(discord.InitCommunityTask, discord.Error(err.Error()))
+			importProgress.StopTask(discord.InitCommunityTask)
+			m.publishImportProgress(importProgress)
+			return
+		}
+
+		importProgress.UpdateTaskProgress(discord.InitCommunityTask, 0.5)
```
0x-r4bbit, line 2265: Also arbitrary, and the following ones for `InitCommunityTask` as well

0x-r4bbit:
Closing this as it's stale

PR #2829, opened 2022-08-30
refactor(messenger): emit discord import updates via interval

0x-r4bbit:
This changes publishing progress updates of the discord import tool to
happen in an interval instead of synchronously whenever an emission was
done.

The benefit of this is that we reduce the amount of signals being sent
which also improves signal handling in clients as they have to handle
less load.

### DIFF
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -1919,6 +1919,11 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
 func (m *Messenger) RequestImportDiscordCommunity(request *requests.ImportDiscordCommunity) {
 	go func() {
 
+		progressUpdates := make(chan *discord.ImportProgress)
+		done := make(chan struct{})
+		cancel := make(chan string)
+		m.startPublishImportProgressInterval(progressUpdates, cancel, done)
+
 		importProgress := &discord.ImportProgress{}
 		importProgress.Init([]discord.ImportTask{
 			discord.CommunityCreationTask,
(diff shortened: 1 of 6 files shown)


protocol/messenger_communities.go
```diff
@@ -2306,3 +2312,31 @@ func (m *Messenger) cleanUpFailedImport(community *communities.Community) {
 func (m *Messenger) publishImportProgress(progress *discord.ImportProgress) {
 	m.config.messengerSignalsHandler.DiscordCommunityImportProgress(progress)
 }
+
+func (m *Messenger) startPublishImportProgressInterval(c chan *discord.ImportProgress, done chan struct{}) {
+
+	var currentProgress *discord.ImportProgress
+
+	go func() {
+		ticker := time.NewTicker(1 * time.Second)
```
0x-r4bbit, line 2423: I think 1 second is reasonable. If anybody thinks this should be bigger/smaller, let me know

protocol/messenger_communities.go
```diff
@@ -2306,3 +2312,31 @@ func (m *Messenger) cleanUpFailedImport(community *communities.Community) {
...
+				}
+			case progressUpdate := <-c:
+				currentProgress = progressUpdate
+			case <-done:
+				if currentProgress != nil {
+					m.publishImportProgress(currentProgress)
+				}
+				return
+			case <-m.quit:
+				return
```
0x-r4bbit, line 2447: Need to do cleanup work as part of `quit`

protocol/messenger_communities.go
```diff
@@ -2306,3 +2312,31 @@ func (m *Messenger) cleanUpFailedImport(community *communities.Community) {
...
+				if currentProgress != nil {
+					m.publishImportProgress(currentProgress)
+				}
+			case progressUpdate := <-c:
+				currentProgress = progressUpdate
+			case <-done:
+				if currentProgress != nil {
+					m.publishImportProgress(currentProgress)
+				}
+				return
```
0x-r4bbit, line 2438: Need to handle `cancel` case here as well

PR #2831, opened 2022-08-31
Add activity center read/unread endpoints

cammellos:
Add activity center read/unread endpoints to accommodate new designs.

This is a breaking change, since we accommodate for multiple requests from a user.
It will remove previous requests, but trusted status won't be changed, we could try to make it backward compatible, let me know if that's a concern (not a concern for mobile).

cc @richard-ramos


richard-ramos:
It should be okay on desktop, but just in case, this PR will be tracked also on this issue: https://github.com/status-im/status-desktop/issues/7373
Thank you!

protocol/activity_center_persistence.go
```diff
@@ -301,6 +330,13 @@ func (db sqlitePersistence) buildActivityCenterQuery(tx *sql.Tx, cursor string,
 
 	}
 
+	switch read {
+	case activityCenterQueryParamsReadRead:
+		readWhere = "AND a.read = 1"
+	case activityCenterQueryParamsReadUnread:
+		readWhere = "AND NOT(a.read)"
+	}
+
```
Samyoul, line 339: Your types don't have a predefined unknown type, so you may need a default case for uninitialised `activityCenterQueryParamsRead` vars.

Samyoul (approved):
Nice work. Looks nice and robust.

PR #2833, opened 2022-09-01
feat: batch balance

alaibe:
Batch request to get balance, the chunk of token is of size 100.
This should improve 2 things:
1 - the number of call we make to infura should be reduced due to the batching
2 - the latency of this call should be reduced too due to less network call

### DIFF
--- a/contracts/contracts.go
+++ b/contracts/contracts.go
@@ -4,6 +4,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/status-im/status-go/contracts/directory"
+	"github.com/status-im/status-go/contracts/ethscan"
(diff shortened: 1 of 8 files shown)


richard-ramos (approved):

rasom:
> 1 - the number of call we make to infura should be reduced due to the batching

unfortunately that's not how it works

> All requests within the array are counted as individual requests against the daily request limit.  For example, if the array contains 30 requests, then those 30 requests count against the daily request limit.

https://docs.infura.io/infura/networks/ethereum/how-to/make-batch-requests

alaibe:
@rasom what do u mean that s not how it works?
I am calling only one contract which query many balances and I am not using the infura batch endpoint

rasom:
@alaibe sorry, I didn't get that from PR's description

rasom (approved):

PR #2836, opened 2022-09-02
Implemented more comprehensive file compression to handle large files

Samyoul:
## What's Changed?

I've added a new function `CompressToFileLimits` takes an image.Image and analyses the pixel dimensions, if the longest side is greater than the `longSideMax` image.Image will be resized, before compression begins. 
Next the `image.Image` is repeatedly encoded and resized until the data fits within the given `FileSizeLimits`. There is no limit on the number of times the cycle is performed, the `image.Image` is reduced to 95% of its size at the end of every round the file size exceeds the given limits.

Fixes https://github.com/status-im/status-mobile/issues/13916

PR #2837, opened 2022-09-02
Implemented more comprehensive file compression to handle large files

### LINKED ISSUE #13916: Sometimes images of HEIF and HEIC formats are not sent from iOS
# Bug Report

## Problem 
The error occurs when sending a HEIF and HEIC image in chats and timeline. 

Screenshot 2022-09-01 at 16 16 47

The issue is not reproduced with all images of these formats. The size is also presumably bigger than 3 MB (but not always)
Image info with which the issue can be reproduced:
Screenshot 2022-09-01 at 16 18 28

Here you can download two examples of such images:
https://drive.google.com/file/d/1TFqRxga659Qc-V6qyOvX-2aQ8RZRYtsL/view?usp=sharing
https://drive.google.com/file/d/1rzIPVCbRYwH_cKjNE5YqecUwlkyJ9BNh/view?usp=sharing

For large JPEG images (e.g. 11 MB) it works fine - you can send it from iPhone.

### Reproduction

1. Open 1-1 chat or timeline
2. Send an image taken on iPhone

### Additional Information

- Status version: RC 1.20
- Operating System: iOS

Samyoul:
## What's Changed?

I've added a new function `CompressToFileLimits` takes an image.Image and analyses the pixel dimensions, if the longest side is greater than the `longSideMax` image.Image will be resized, before compression begins. 
Next the `image.Image` is repeatedly encoded and resized until the data fits within the given `FileSizeLimits`. There is no limit on the number of times the cycle is performed, the `image.Image` is reduced to 95% of its size at the end of every round the file size exceeds the given limits.

Fixes https://github.com/status-im/status-mobile/issues/13916

---

🤦 https://github.com/status-im/status-go/pull/2836#issuecomment-1235560084

### DIFF
--- a/images/encode.go
+++ b/images/encode.go
@@ -29,6 +29,15 @@ func renderJpeg(w io.Writer, m image.Image, config EncodeConfig) error {
 	return jpeg.Encode(w, m, o)
 }
 
+type FileSizeError struct {
+	expected int
+	received int
+}
+
+func (e *FileSizeError) Error() string {
+	return fmt.Sprintf("image size after processing exceeds max, expected < '%d', received < '%d'", e.expected, e.received)
+}
+
 func EncodeToLimits(bb *bytes.Buffer, img image.Image, bounds FileSizeLimits) error {
 	q := MaxJpegQuality
 	for q > MinJpegQuality-1 {
@@ -46,11 +55,7 @@ func EncodeToLimits(bb *bytes.Buffer, img image.Image, bounds FileSizeLimits) er
 			if bounds.Max > bb.Len() {
 				return nil
 			}
-			return fmt.Errorf(
-				"image size after processing exceeds max, expect < '%d', received < '%d'",
-				bounds.Max,
-				bb.Len(),
-			)
+			return &FileSizeError{expected: bounds.Max, received: bb.Len()}
 		}
 
 		bb.Reset()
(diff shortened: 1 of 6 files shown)


Samyoul:
@qoqobolo This seems to be fixing the issue and is doing so in a reasonable time, even for pretty large files. Although the tests were only done on my laptop, but it seems to work regardless of the image used.

qoqobolo:
Thank you @Samyoul! I will create a `status-mobile` PR as soon as this one is reviewed to test the fix on my devices.

images/encode.go
```diff
@@ -58,6 +60,40 @@ func EncodeToLimits(bb *bytes.Buffer, img image.Image, bounds DimensionLimits) e
...
+	}
+
+	for {
+		err := EncodeToLimits(bb, img, bounds)
+		if err == nil {
+			return nil
+		}
+		if err.Error()[:50] != "image size after processing exceeds max, expect < " {
+			return err
+		}
```
richard-ramos: Maybe something like [this](https://stackoverflow.com/a/68654658) could be implemented to avoid having to compare the error string?

richard-ramos (approved):

qoqobolo:
@Samyoul the issue is fixed, thank you!

Tested:
- sending the images from the issue in 1-1 chats, group chats, and timeline
- sending jpeg file >50MB
- setting those images as profile pictures
- save/reply/share options

PR can be merged and cherry-picked.

PR #2838, opened 2022-09-05
feat: add and distribute `chatIdentity.FirstMessageTimestamp`

osmaczko:
`FirstMessageTimestamp` enables members of the community to determine if
there are any messages they can fetch on the community channel(chat).

`FirstMessageTimestamp` is advertised by admin for each community chat
through `CommunityDescription`. It assumes admin is online frequently
enough to capture the first channel message.

For existing communities admin determines first message timestamp by
finding oldest chat message in its local database.

task: status-im/status-desktop#6731


protocol/chat.go
```diff
@@ -35,6 +35,11 @@ const (
 	ChatTypeCommunityChat
 )
 
+const (
+	FirstMessageTimestampUndefined = 0
+	FirstMessageTimestampNoMessage = 1
```
0x-r4bbit, line 40: I wonder if this is really needed. I'm probably missing something here.. Can't we work with only `0` and assume that no messages have been saved (yet)? So it's either `0` or `> 0` ?

0x-r4bbit (approved):
This LGTM just not sure if the `FirstMessageTimestampNoMessage` is really necessary

richard-ramos (approved):
:100:

protocol/chat.go
osmaczko: Is is needed, because of backward compatibility. It is to distinguish if the admin actually propagated the first message timestamp or not.
If there is an existing community chat that has messages to fetch and the admin hasn't logged in yet, then members would not be able to fetch these messages if there is only `0` state (undefined != no message).

PR #2839, opened 2022-09-05
feat: `key_uid` column added to the `accounts` table

saledjenic:
Each account must know about `key-uid` of the keyapair it belongs to. We need that info when we're signing a transaction in order to know if an appropriate card is inserted or not.


0x-r4bbit (approved):
This LGTM.
I'd prefer if we used `KeyUID` as opposed to `KeyUid` throughout, but not blocker.

richard-ramos (approved):

Samyoul (approved):

PR #2845, opened 2022-09-06
fix: lint

richard-ramos:


eth-node/types/topic.go
```diff
@@ -95,9 +95,5 @@ func StringToTopic(s string) (t TopicType) {
 }
 
 func TopicTypeToByteArray(t TopicType) []byte {
-	topic := make([]byte, 4)
-	for i, b := range t {
-		topic[i] = b
-	}
-	return topic
+	return t[:]
```
Samyoul: Would this be more faithful to the original implementation?

```suggestion
	return t[:4]
```

Samyoul (approved):

PR #2847, opened 2022-09-07
feat: Added new BaseFee param inside transactions

Khushboo-dev-cpp:

Important changes:

Closes #


services/wallet/transfer/commands.go
```diff
@@ -148,10 +149,11 @@ type controlCommand struct {
 	feed               *event.Feed
 	errorsCount        int
 	nonArchivalRPCNode bool
+	rpcClient          *rpc.Client
```
alaibe: I think you don't need to pass rpcClient all the way down, you already should have access to the chainClient in the downloader which is a wrapper around the rpc Client.

services/wallet/transfer/downloader.go
```diff
@@ -162,6 +170,17 @@ func (d *ETHDownloader) getTransfersInBlock(ctx context.Context, blk *types.Bloc
...
+				var feeHistory FeeHistory
+				err = rpcClient.Call(&feeHistory, d.chainClient.ChainID, "eth_feeHistory", "0x1", (*hexutil.Big)(blk.Number()), nil)
+				if err != nil {
+					return nil, err
+				}
+
+				var baseGasFee string = ""
+				if len(feeHistory.BaseFeePerGas) > 0 {
+					baseGasFee = feeHistory.BaseFeePerGas[0]
+				}
```
alaibe: Could you extract this block in the chainClient that will avoid to duplicate it later, wdyt?

appdatabase/migrations/sql/1662589108_add_base_fee_transaction.up.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE transfers ADD COLUMN base_gas_fee TEXT NOT NULL DEFAULT "";
+UPDATE transfers SET base_gas_fee = "";
+
+ALTER TABLE pending_transactions ADD COLUMN base_gas_fee TEXT NOT NULL DEFAULT "";
+UPDATE pending_transactions SET base_gas_fee = "";
```
alaibe: You are not updating pending_transactions right?

alaibe (commented):
I think what you did will works but the downloader contains everything already, no need to pass down another type.

appdatabase/migrations/sql/1662589108_add_base_fee_transaction.up.sql
Khushboo-dev-cpp: damn yes, I will remove that.

services/wallet/transfer/downloader.go
Khushboo-dev-cpp: I am not sure I follow @alaibe

services/wallet/transfer/commands.go
Khushboo-dev-cpp: ok I am trying to use the chainClient now

services/wallet/transfer/downloader.go
Khushboo-dev-cpp: ok I got it, you are suggesting we move this piece if code over to the chain client?

alaibe (approved):

PR #2848, opened 2022-09-08
feat(token): enable changing token address from the node config

alaibe:
Allow to override address of token.

This is used when running test with ganache

### DIFF
--- a/services/wallet/service.go
+++ b/services/wallet/service.go
@@ -30,7 +30,7 @@ func NewService(
 	cryptoOnRampManager := NewCryptoOnRampManager(&CryptoOnRampOptions{
 		dataSourceType: DataSourceStatic,
 	})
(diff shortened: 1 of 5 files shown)


services/wallet/service.go
```diff
@@ -31,6 +31,7 @@ func NewService(
 		dataSourceType: DataSourceStatic,
 	})
 	tokenManager := &TokenManager{db: db, RPCClient: rpcClient, networkManager: rpcClient.NetworkManager}
+	tokenManager.OverrideTokens()
```
Samyoul: I wonder if `OverrideTokens()` should be an internally called function of a new function called `NewTokenManager`. It seems that this function should always be called by default on a new `TokenManager`

alaibe: You are  correct, doing NewTokenManager makes more sense, I will change

Samyoul (approved):
Looks great. Thank you.

iurimatias (approved):

PR #2853, opened 2022-09-09
Audio Progressive Compression

Samyoul:
First steps to improving audio transport issues, see https://github.com/status-im/status-mobile/issues/13914

Unfortunately there is a bug that I don't really have want to spend more time resolving, it is really taking too much time to resolve. Basically hangs for some encodings params, I don't know why, but it hangs indefinitely and that is a big problem and if I try to interrupt the process I get a segmentation violation crash. And recover isn't seeming to work. Also it would be better if we can decode AAC and then re-encode to AAC, so this whole PR is a massive wash.

## HINT FOR FUTURE DEVELOPERS OF THE FUTURE >>>

One thing to notice, decreasing the `SamplingRate` actually will increase the end file size relative to an encoding with a higher `SamplingRate`, So start with the original `SamplingRate` run through the different `ByteRates` and then increase the `SamplingRate`.


Samyoul:
Closing, see the PR description for details.

PR #2857, opened 2022-09-14
fix: workaround to wrongly renamed migration file for keypairs table

stefandunca:
Restore a wrongly renamed migration file done in the merge favourite [PR #2844](https://github.com/status-im/status-go/pull/2844)
  - The wrong renaming https://github.com/status-im/status-go/pull/2844/files#diff-35a2a92571286f19a4534533cb97e7c5d27c77e0580b3064f280eee8421966f2

The renaming might block startup for clients who already run the newly renamed migration and beak the onld one


alaibe:
As it is already the latest migration on develop, I am not sure to understand how it will solve any issue?

AFAIK, the migration run if there is a new one after the last one, if I am correct, this new migration will never run, correct?

cammellos:
> he migration run if there is a new one after the last one, if I am correct, this ne

The issue is for clients that were on the removed migration, say that you have migrations 1,2,3

We release 2
Client runs 2
We delete/rename 2 to 3

If clients install that version, they will crash, because it looks for the migration 2 on start, so you will get a file not found exception.

Newer clients will migrate to 3 directly and have no issues.

alaibe:
@cammellos gotcha, thanks for clarifying

alaibe (approved):

cammellos (approved):

PR #2861, opened 2022-09-15
fix(communities): ensure existing archives are seeded on startup

0x-r4bbit:


protocol/communities/manager.go
```diff
@@ -2096,6 +2096,13 @@ func (m *Manager) LoadHistoryArchiveIndexFromFile(communityID types.HexBytes) (*
 	return wakuMessageArchiveIndexProto, nil
 }
 
+func (m *Manager) TorrentFileExists(communityID string) bool {
+	if _, err := os.Stat(m.torrentFile(communityID)); err == nil {
+		return true
+	}
+	return false
```
```suggestion
	_, err := os.Stat(m.torrentFile(communityID))
	return err == nil
```

osmaczko (approved):

PR #2862, opened 2022-09-15
fix: remove ourselves from members when leaving community

osmaczko:
Otherwise clients see invalid members count in invitation bubbles.

### DIFF
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -702,16 +702,11 @@ func (o *Community) RemoveUserFromChat(pk *ecdsa.PublicKey, chatID string) (*pro
 	return o.config.CommunityDescription, nil
 }
 
-func (o *Community) RemoveUserFromOrg(pk *ecdsa.PublicKey) (*protobuf.CommunityDescription, error) {
-	o.mutex.Lock()
-	defer o.mutex.Unlock()
-
-	if o.config.PrivateKey == nil {
-		return nil, ErrNotAdmin
-	}
+func (o *Community) removeMemberFromOrg(pk *ecdsa.PublicKey) {
 	if !o.hasMember(pk) {
-		return o.config.CommunityDescription, nil
+		return
 	}
+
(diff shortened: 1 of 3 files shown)


protocol/communities/community.go
```diff
@@ -723,7 +718,23 @@ func (o *Community) RemoveUserFromOrg(pk *ecdsa.PublicKey) (*protobuf.CommunityD
 	}
 
 	o.increaseClock()
+}
+
+func (o *Community) RemoveOurselvesFromOrg(pk *ecdsa.PublicKey) {
+	o.mutex.Lock()
+	defer o.mutex.Unlock()
+	o.removeMemberFromOrg(pk)
+}
```
0x-r4bbit, line 727: Why exactly is this called `RemoveOurselvesFromOrg` ?
This seems a pretty generic function. In fact it seems to be just an exposed API for `removeMemberFromOrg`.

osmaczko: If you compare `RemoveOurselvesFromOrg` with `RemoveUserFromOrg` you will see they only differ with one check. If we remove ourselves from org we don't need to check if we are admins.

The other option was to have one function with an additional boolean parameter:
`RemoveUserFromOrg(pk *ecdsa.PublicKey, requireAdmin bool)`

Since boolean arguments are bad for code readability (see example below), I decided to go with the first one.

```
func (a *Api) LeaveCommunity() {
    community.RemoveOurselvesFromOrg(m.identity.pubKey)
}
func (a *Api) RemoveUserFromOrg(pubKey *PubKey) {
    community.RemoveUserFromOrg(pubKey)
}
```
vs
```
func (a *Api) LeaveCommunity() {
    community.RemoveUserFromOrg(m.identity.pubKey, false)
}
func (a *Api) RemoveUserFromOrg(pubKey *PubKey) {
    community.RemoveUserFromOrg(pubKey, true)
}
```

0x-r4bbit (approved):

PR #2863, opened 2022-09-15
Fix bug with tracking

cammellos:
If it's not creating a `newMessage` we should not be tracking it.


vitvly (approved):

vitvly:
@cammellos sorry, just wondering - how did the bug manifest itself?

cammellos:
> @cammellos sorry, just wondering - how did the bug manifest itself?

You could not invite people to a community, but I introduced it with my last commit :)

protocol/common/message_sender.go
```diff
@@ -275,6 +275,7 @@ func (s *MessageSender) sendCommunity(
 				return nil, err
 			}
 		}
+                return nil, nil
```
Samyoul: 💡 Ah!

Samyoul (approved):

PR #2865, opened 2022-09-16
feat: sync saved addresses

stefandunca:
### Required by: https://github.com/status-im/status-desktop/issues/7229

Considerations:
- Clients need to switch to Messenger (`ext`) Saved Addresses APIs to trigger sync for each change (add/update/delete)
- Old `wallet` APIs are still available but don't trigger a sync. Plan to be retired after clients move to Messenger APIs

Main changes:
- [x] Extend saved addresses DB with sync info: removed flag and update clock
- [x] Create a custom protobuf message payload to sync saved addresses
- [x] Add AddressManager instance to messenger and specific APIs that trigger sync
- [x] Send new sync messages (add and delete)
- [x] Process sync messages
- [x] Cleanup saved addresses on each start of messenger
- [x] Notify clients of saved addresses
- [x] Tests

Open questions:
- [ ] Is there a way to do incremental sync on a manual trigger?
- [ ] `haveSameElements` is duplicated between sync and saved address tests, where is the best place to put helper functions between modules?


protocol/messenger.go
```diff
@@ -2865,10 +2894,11 @@ type ReceivedMessageState struct {
...
 	// Timesource is a time source for clock values/timestamps.
-	Timesource              common.TimeSource
-	AllBookmarks            map[string]*browsers.Bookmark
-	AllVerificationRequests []*verification.Request
-	AllTrustStatus          map[string]verification.TrustStatus
+	Timesource                 common.TimeSource
+	AllBookmarks               map[string]*browsers.Bookmark
+	AllVerificationRequests    []*verification.Request
+	AllTrustStatus             map[string]verification.TrustStatus
+	SavedAddressesChangesCount int
```
cammellos: Could you elaborate what is this used for?

services/wallet/api.go
```diff
@@ -176,14 +176,14 @@ func (api *API) AddSavedAddress(ctx context.Context, sa SavedAddress) error {
 	if sa.ChainID == 0 {
 		sa.ChainID = api.s.rpcClient.UpstreamChainID
 	}
-	err := api.s.savedAddressesManager.AddSavedAddress(sa)
+	_, err := api.s.savedAddressesManager.AddOrUpdateSavedAddress(sa)
```
cammellos: `UpsertSavedAddress` maybe? a bit less wordy

services/wallet/saved_addresses.go
```diff
@@ -15,41 +19,169 @@ type SavedAddress struct {
...
+			return false, err
+		}
+
+		return dbUpdateClock <= updateClock, nil
+	}
+	return true, nil
+}
+
+func (sam *SavedAddressesManager) AddSavedAddressIfNewerUpdate(sa SavedAddress, syncClock int64, updateClock int64) (insertedOrUpdated bool, err error) {
+	newer, err := sam.checkIfNewerChange(sa.ChainID, sa.Address, updateClock)
```
cammellos: I would have a the two things in a transaction, you can check if newer and update if so in a single go

services/wallet/saved_addresses.go
```diff
@@ -15,41 +19,169 @@ type SavedAddress struct {
...
+	updateClock := time.Now().Unix()
+	err = sam.addRawSavedAddress(sa, SavedAddressMeta{false, syncClockCreatedEditedHere, sql.NullInt64{Int64: updateClock, Valid: true}})
+	if err != nil {
+		return 0, err
+	}
+	return updateClock, nil
+}
+
+func (sam *SavedAddressesManager) checkIfNewerChange(ChainID uint64, Address common.Address, updateClock int64) (newer bool, err error) {
+	rows, err := sam.db.Query("SELECT update_clock FROM saved_addresses WHERE network_id = ? AND address = ?", ChainID, Address)
```
cammellos: If you are querying a single row you can use `QueryRow`

services/wallet/saved_addresses.go
```diff
@@ -15,41 +19,169 @@ type SavedAddress struct {
...
+	err = sam.addRawSavedAddress(sa, SavedAddressMeta{false, sql.NullInt64{Int64: syncClock, Valid: true}, sql.NullInt64{Int64: updateClock, Valid: true}})
+	if err != nil {
+		return false, err
+	}
+
+	return true, err
+}
+
+func (sam *SavedAddressesManager) DeleteSavedAddressIfNewerUpdate(chainID uint64, address common.Address, syncClock int64, updateClock int64) (deleted bool, err error) {
+	newer, err := sam.checkIfNewerChange(chainID, address, updateClock)
```
cammellos: Similarly, best to use a transaction here

services/wallet/saved_addresses.go
```diff
@@ -15,41 +19,169 @@ type SavedAddress struct {
...
+	updateClock := time.Now().Unix()
+	_, err = insert.Exec(chainID, address, updateClock)
+	if err != nil {
+		return 0, err
+	}
+	return updateClock, nil
+}
+
+func (sam *SavedAddressesManager) DeleteSoftRemovedSavedAddresses(threshold int64) error {
+	tx, err := sam.db.Begin()
```
cammellos: No need to use a transaction here, unless you are doing multiple operation, but  this is not it seems

services/wallet/saved_addresses.go
```diff
@@ -15,41 +19,169 @@ type SavedAddress struct {
...
 
-	var rst []SavedAddress
+// Retrieve raw data based on SELECT query using rawQueryColumnsOrder
+func getRawSavedAddressesFromDBRows(rows *sql.Rows) ([]SavedAddress, []SavedAddressMeta, error) {
+	var addresses []SavedAddress
+	var metas []SavedAddressMeta
 	for rows.Next() {
 		sa := SavedAddress{}
-		err := rows.Scan(&sa.Address, &sa.Name, &sa.Favourite, &sa.ChainID)
+		sam := SavedAddressMeta{}
```
cammellos: Any reason we are splitting the two in `SavedAddress` and `SavedAddressMeta`? 
it seems like it's adding complexity (we need to iterate over both and use it together), rather then having a single struct, and I'd have expected the data to be kept separate if in two different structs, so it could be counter-intuitive

protocol/messenger_saved_address.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+	}
+	return
+}
+
+func (m *Messenger) handleSyncSavedAddress(state *ReceivedMessageState, syncMessage protobuf.SyncSavedAddress) (err error) {
+	address := gethcommon.BytesToAddress(syncMessage.Address)
+	if syncMessage.Removed {
+		_, err = m.savedAddressesManager.DeleteSavedAddressIfNewerUpdate(syncMessage.ChainId,
+			address, int64(syncMessage.SyncClock), syncMessage.UpdateClock)
```
cammellos: I would not cast to `int64`, it's probably ok, but best to keep uint64 and compare uint64, any reason to use `int64`? (the only one I can think of is the fact that we have `nullint64` and not `nullUint64` , but if that's the case, you should convert as close to the db as possible, no earlier)

services/wallet/api.go
stefandunca: Will do.

P.S. I personally prefer a more human-readable approach to saving a few letters.

protocol/messenger.go
stefandunca: Notifies clients about sync changes so they can update the saved addresses view

services/wallet/saved_addresses.go
stefandunca: > Any reason we are splitting the two in SavedAddress and SavedAddressMeta?

I didn't want to pollute public APIs with internal metadata, allow messing with sync/soft-delete functionality and have this separation of concerns.
I also searched for some kind of inheritance in go and try to use anonymous structures but didn't work. 

@cammellos how about having a higher struct that is composed of both? Or do you prefer to have it simplified and expose the internal data in the API?

> and I'd have expected the data to be kept separate if in two different structs, so it could be counter-intuitive

That was my first approach, to have a sync DB and build a more generalized sync framework and extend it to other structures, then realise that this will be a bigger change and premature given that I'm new to the code.

protocol/messenger_saved_address.go
stefandunca: > ... any reason to use `int64`? (the only one I can think of is the fact that we have `nullint64` and not `nullUint64`

Exactly

> but if that's the case, you should convert as close to the db as possible, no earlier)

I was thinking the opposite, better to have the limitations exposed higher the call chain, closer to requirements, where we have enough information to know if we will be affected instead of hiding it deeper where we miss domain knowledge.

protocol/messenger.go
cammellos: I think how we generally do it is similarly to the rest of messenger response, we actually pass the new/changed objects, so we keep the behavior consistent

services/wallet/api.go
cammellos: Me too :), I find Upsert more readable since I am more familiar with it as commonly used for API/db operations, but either are fine by me

protocol/messenger_saved_address.go
cammellos: I think it's best to keep uint64 everywhere, and we should not be using nullInt sql in the metadata struct, as that's more of a database concern, in the rest of the codebase I think for the most part with use the `sql.Null` just to pull from the db in temporary variable, but then we use native go types for our structs, there's a few examples I believe in the various persistence.go files

Another note is that the approach you used leads to some subtle bugs, for example in some code in the PR, when you convert a uint64 to a int64 and use them, you need to be careful about the value being wrapped around:

```
 dbUpdateClock <= updateClock
```

If both wraps around, it works correctly, but if only ones does, it would not be what you expect

https://go.dev/play/p/V49B-mtjcJ1

So we should honor the original type before doing any operation, and the easiest and less prone to error way is to only use nullint64 to pull/store from the db, but immediately convert back to the original type in my opinion

services/wallet/saved_addresses.go
cammellos: > I didn't want to pollute public APIs with internal metadata, allow messing with sync/soft-delete functionality and have this separation of concerns.

It depends on what you mean, in general, you can use lower case fields, and those won't be accessible outside the namespace (you can see it as `private` fields essentially), they won't show in the JSON etc
That's probably common, @Samyoul is the in-house expert on what's the best way
I'd generally would not have 2 separate objects, but I'd be ok with either something like:
```
struct A {
 metadata metadatatype
 FieldA
 FieldB
}

struct metadatatype {
  metadataFileldA
  metadataFieldB
}
```
or even
```
struct A {

 FieldA
 FieldB
  metadataFileldA
  metadataFieldB
}

```

depending on usage and complexity of metadatatype

protocol/messenger_sync_saved_addresses_test.go
```diff
@@ -0,0 +1,248 @@
...
+	"github.com/status-im/status-go/eth-node/crypto"
+	"github.com/status-im/status-go/protocol/encryption/multidevice"
+	"github.com/status-im/status-go/protocol/tt"
+	"github.com/status-im/status-go/services/wallet"
+	"github.com/status-im/status-go/waku"
+
+	"github.com/stretchr/testify/suite"
+	"go.uber.org/zap"
+
+	"github.com/status-im/status-go/eth-node/types"
```
Samyoul: Group imports like this:

- standard lib
- 3rd party libs
- go-ethereum & status-go libs

```suggestion
	"github.com/stretchr/testify/suite"
	"go.uber.org/zap"
	
	"github.com/ethereum/go-ethereum/common"
	gethbridge "github.com/status-im/status-go/eth-node/bridge/geth"
	"github.com/status-im/status-go/eth-node/crypto"
	"github.com/status-im/status-go/eth-node/types"
	"github.com/status-im/status-go/protocol/encryption/multidevice"
	"github.com/status-im/status-go/protocol/tt"
	"github.com/status-im/status-go/services/wallet"
	"github.com/status-im/status-go/waku"
```

protocol/messenger_saved_address.go
```diff
@@ -0,0 +1,112 @@
...
+import (
+	"context"
+	"time"
+
+	gethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/golang/protobuf/proto"
+	"github.com/status-im/status-go/protocol/common"
+	"github.com/status-im/status-go/protocol/protobuf"
+	"github.com/status-im/status-go/services/wallet"
+)
```
Samyoul:
```suggestion
	"github.com/golang/protobuf/proto"

	gethcommon "github.com/ethereum/go-ethereum/common"
	"github.com/status-im/status-go/protocol/common"
	"github.com/status-im/status-go/protocol/protobuf"
	"github.com/status-im/status-go/services/wallet"
)
```

protocol/messenger_test.go
```diff
@@ -2387,6 +2387,7 @@ func (t *testTimeSource) GetCurrentTime() uint64 {
 	return uint64(time.Now().Unix())
 }
 
+// Wait until the condition is true or the timeout is reached.
 func WaitOnMessengerResponse(m *Messenger, condition func(*MessengerResponse) bool, errorMessage string) (*MessengerResponse, error) {
```
Samyoul: See the convention for documenting your go code https://go.dev/blog/godoc

```suggestion
// WaitOnMessengerResponse Wait until the condition is true or the timeout is reached.
func WaitOnMessengerResponse(m *Messenger, condition func(*MessengerResponse) bool, errorMessage string) (*MessengerResponse, error) {
```

services/wallet/saved_addresses.go
Samyoul: You could use type embedding like @cammellos suggests, this is the closest thing to inheritance that go has.

```go
type metadatatype struct {
  metadataFileldA string
  metadataFieldB bool
}

type A struct {
 metadatatype
 FieldA int
 FieldB int
}
```

If you are concerned about polluting the API see the docs on this https://pkg.go.dev/encoding/json, unexported fields do not get marshalled into json. See example https://go.dev/play/p/NzXXXTwkRqb

If the fields need to be exported for whatever reason we can deal with that too:

```go
// Field is ignored by this package.
Field int `json:"-"
```

```go
type Metadatatype struct {
	MetadataFileldA string `json:"-"`
	MetadataFieldB  bool   `json:"-"`
}

type A struct {
	Metadatatype
	FieldA int
	FieldB int
}
```

See example here https://go.dev/play/p/564m4R-U5us , play around delete the json tags and see the result of marshalling.

stefandunca: Thanks, both for the great description.

protocol/messenger_saved_address.go
stefandunca: In the end, me relying on the client to mind the limitations and not reach overflowing and wrapping around looks like a more error-prone solution. Thanks

protocol/messenger.go
stefandunca: Done.

> ... so we keep the behavior consistent

In desktop, we reload all the saved addresses when a change is reported hence we don't use the incremental data. Using the messenger updates by merging with the presentation layer model looked to me like a premature optimization at this point.

I updated the implementation to follow the other models and added the ChainId and Address on a delete without the rest of the data, please let me know if this will work for mobile.

services/wallet/saved_addresses.go
```diff
@@ -2,54 +2,214 @@ package wallet
...
+	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 )
 
+var SyncClockCreatedEditedHere uint64 = 0
+
+type SavedAddressMeta struct {
+	Removed     bool
+	SyncClock   uint64 // clock of the last sync
```
cammellos: What is `SyncClock` used for?
I can see being set etc, but I can't see any usage, do you know where it's used?

stefandunca: That is part of the plan to do an incremental manual sync. I realized late that we don't store a sync clock for paired devices and I can't use it for now.

I think will be great to have it for the future and send only new changes to paired devices based on the last sync timestamp. This requires additional work that doesn't help the current feature. Should I remove it completely and bring it back later on?

stefandunca: This is similar to other data types like bookmarks that have a `delete_at` and `clock`

cammellos: It would be a bit more complex as you don't have only 2 devices, you might have 3, or you might remove one and add one etc, so you would have to keep track of the state of the sync for each device

We can revisit later I think, if necessary, for now it's best to remove it I think

cammellos: > This is similar to other data types like bookmarks that have a delete_at and clock

Ah that would be a different concern, not sure about bookmarks, also seems like it's not used, I'd avoid unnecessary fields unless there's an actual usage, since `clock` should signal already the last update and it's generally all we need in most cases

stefandunca: > It would be a bit more complex as you don't have only 2 devices, you might have 3, or you might remove one and add one etc, so you would have to keep track of the state of the sync for each device
> 
I see we keep a Lamport time stamp that should be consistent between each device and might be what I was searching for. As you mentioned this requires investing some time and I'm not sure if the bandwidth saved brings worthy benefits at the present moment, will remove it for now.

protocol/messenger_saved_address.go
```diff
@@ -0,0 +1,117 @@
...
+	})
+	if err != nil {
+		return err
+	}
+
+	chat.LastClockValue = clock
+	return m.saveChat(chat)
+}
+
+func (m *Messenger) SyncNewSavedAddress(ctx context.Context, savedAddress *wallet.SavedAddress, updateClock uint64) error {
```
cammellos: do these need to be public?

protocol/messenger_saved_address.go
```diff
@@ -0,0 +1,117 @@
...
+	return m.dispatchSyncSavedAddress(ctx, protobuf.SyncSavedAddress{
+		Address:     savedAddress.Address.Bytes(),
+		Name:        savedAddress.Name,
+		Favourite:   savedAddress.Favourite,
+		ChainId:     savedAddress.ChainID,
+		UpdateClock: updateClock,
+	})
+}
+
+func (m *Messenger) SyncDeletedSavedAddress(ctx context.Context, chainID uint64, address gethcommon.Address, updateClock uint64) error {
```
cammellos: same as above, public?

protocol/messenger_sync_saved_addresses_test.go
```diff
@@ -0,0 +1,251 @@
...
+	s.Require().NoError(err)
+	_, err = savedAddressesManager.UpdateMetadataAndUpsertSavedAddress(sa2)
+	s.Require().NoError(err)
+
+	// Trigger's a sync between devices
+	err = s.main.SyncDevices(context.Background(), "ens-name", "profile-image")
+	s.Require().NoError(err)
+
+	// Wait and check that saved addresses are synced
+	_, err = WaitOnMessengerResponse(
```
cammellos, line 149: You can also test response to make sure saved addresses are returned

services/wallet/saved_addresses.go
```diff
@@ -2,54 +2,195 @@ package wallet
...
+}
+
 type SavedAddress struct {
 	Address common.Address `json:"address"`
 	// TODO: Add Emoji and Networks
 	// Emoji    string         `json:"emoji"`
 	Name      string `json:"name"`
 	Favourite bool   `json:"favourite"`
 	ChainID   uint64 `json:"chainId"`
+	SavedAddressMeta
```
cammellos: Do you want to expose this through the API?
If that's the case, you might want to set `json:"savedAddressMeta"`, otherwise I'd make it private (`savedAddressMeta`), if it needs to be public, then you can make sure is not encoded in `JSON` "json:"-"`, I think it needs to be one of the 3 :)

cammellos (approved):
Looks good, the only thing I would address is the `JSON` point, the rest it's minor things, great work, thank you!

PR #2866, opened 2022-09-16
feat: delete for me

yqrashawn:
### Summary
Implement the delete for me feature.
`deleted_for_me` will be synced between paired devices using `DeleteForMeMessage`

Feature request issue: https://github.com/status-im/status-mobile/issues/13873
Related status-mobile PR https://github.com/status-im/status-mobile/pull/14007

### Important changes:
1. add `deleted_for_me` column in `user_messages` table
2. add new `DeleteForMeMessage` message and `protobuf.ApplicationMetadataMessage_SYNC_DELETE_FOR_ME_MESSAGE`
3. add new `DeleteMessageForMeAndSync` to `Messenger`
4. undo delete for me (TBD)

**Note:**
**First time writing Golang code**
**First time working on status-go logic**


yqrashawn:
Should we exclude `.pb.go` files in `.codeclimate.yml` as well?

protocol/message_persistence.go
```diff
@@ -612,6 +621,49 @@ func (db sqlitePersistence) MessageByChatID(chatID string, currCursor string, li
 	return result, newCursor, nil
 }
 
+func (db sqlitePersistence) LatestMessageByChatID(chatID string) ([]*common.Message, string, error) {
+	args := []interface{}{chatID}
+	where := fmt.Sprintf(`
+            WHERE
+                NOT(m1.hide) AND NOT(m1.deleted_for_me) AND m1.local_chat_id = ?
```
cammellos: Wondering if it's worth creating an index for this query, as it might be slow in chats with many messages

protocol/messenger_handler.go
```diff
@@ -1180,9 +1180,60 @@ func (m *Messenger) HandleDeleteMessage(state *ReceivedMessageState, deleteMessa
...
+		}
+	}
+
+	chat, ok := m.allChats.Load(originalMessage.LocalChatID)
+	if !ok {
+		return errors.New("chat not found")
+	}
+
+	// Update message and return it
+	originalMessage.DeletedForMe = true
```
cammellos, line 1211: I think original message my be nil in here?
That would happen if the sync delete message for me arrives before the actual message, for example:

A1 is online
B1 is online
B2 is offline

1) User A1 posts a message
2) User B1 receives the message, deletes the message for them
3) User B2 goes online, receives first the `DeleteMessageForMe`, after the actual message 

In this case I think the code above might crash (MessageByID returns `nil`, and it's not in the response).

I think in theory we should store the `DeleteMessageForMe` instruction if the message is not present, and then check if there are any pending when we receive a message (we do the same for edits & deletes).

What do you think? We could also do that in a separate PR

protocol/protobuf/chat_message.proto
```diff
@@ -54,6 +54,19 @@ message DeleteMessage {
...
 }
 
+message DeleteForMeMessage {
+  uint64 clock = 1;
+
+  string chat_id = 2;
+  string message_id = 3;
+
+  // Grant for community delete messages
+  bytes grant = 4;
```
cammellos: No need of grant, also do we need anything else other than `message_id`? `clock` is probably helpful, though the operation can't be reverted, so I don't think is necessary, the rest of the fields are useful in normal deletes since you want to check permissions (i.e can user x delete message y in chat z), but because it's only for paired devices, here I think we don't need to check any permission. Though not 100% sure, but I think we can get away with only message_id (and maybe clock)

cammellos (commented):
Great work! 
I have left a few comments, but overall looks like it's basically there apart from an edge case

protocol/message_persistence.go
yqrashawn: Shall we include  deleted for me message as latest message and just show deleted for me as latest message?

cammellos: good question, probably we should ask ux, @pedro-et maybe has an opinion

yqrashawn: I've confirmed with @pedro-et there's no latest message in new design anymore.

protocol/protobuf/chat_message.proto
yqrashawn: Fixed with 
```go
message DeleteForMeMessage {
  uint64 clock = 1;
  string message_id = 2;
  bool undo = 3;
}
```
`undo` is for the undo feature

protocol/messenger_handler.go
yqrashawn: Fixed with `checkForDeleteForMes`

protocol/protobuf/chat_message.proto
cammellos: I think `undo` it should be only a local action, basically we should not dispatch the message until the `undo` has timed out, for example:
1) User deletes message
2) The client will update the UI, and show the undo button
3) If the undo button is not clicked, it will then make a query to status-go and actually delete the message

The issue is that if the app crashes/ is closed, the action won't be performed, but I think that's preferrable to send `undo` over the wire (unless you can always undo, but as I understand it's only for a limited period, i.e 3/5 seconds),
what do you think?

yqrashawn: Considering the delete for everyone feature, undo action should definitely guard by time. And we better keep these two undo consistent. 
I'll change the code.

yqrashawn: Fixed

yqrashawn:
ready for review @cammellos

cammellos (approved):

yqrashawn:
add another commit exlucde `*.pb.go` files generated by protoc for codeclimate check

0x-r4bbit (approved):
This looks solid to me!
Before this is merged, make sure that your migration script is indeed the latest one (has highest timestamp prefix)

PR #2869, opened 2022-09-20
feat: add `DiscordMessageAttachment` types and APIs

0x-r4bbit:
This adds a new `DiscordMessageAttachment` type which is part of `DiscordMessage`. Along with that type, there's also a new database table for `discord_message_attachments` and corresponding persistence APIs.

This commit also changes how chat messages are retrieved. Here's why:

`DiscordMessage` can have multiple `DiscordMessageAttachment`. A chat message can have a `DiscordMessage`.

Because we're `LEFT JOIN`'ing the discord message attachments into the chat messages, there's a possibility of multiple rows per message.

Hence, this commit ensures we collect queried discord message attachments on chat messages.


protocol/message_persistence.go
```diff
@@ -2207,3 +2181,161 @@ func (db sqlitePersistence) SetContactRequestState(id string, state common.Conta
...
+type ByClock struct{ Messages }
+type ByClockPinned struct{ PinnedMessages }
+
+func (m ByClock) Less(i, j int) bool {
+	return m.Messages[i].Clock < m.Messages[j].Clock
+}
+
+func (m ByClockPinned) Less(i, j int) bool {
+	return m.PinnedMessages[i].Message.Clock < m.PinnedMessages[j].Message.Clock
+}
```
0x-r4bbit, line 2256: Both `ByClock `and `ByClockPinned` look a bit over engineered to my eyes.
I assume there's a nicer way to get sorting capabilities of `[]*someStructType` in go?

protocol/message_persistence.go
```diff
@@ -2207,3 +2181,161 @@ func (db sqlitePersistence) SetContactRequestState(id string, state common.Conta
...
+			msg.Payload = getUpdatedChatMessagePayload(discordMessage, message.GetDiscordMessage())
+			messageIdx[message.ID] = msg
+		}
+	}
+
+	var messages []*common.Message
+	for _, message := range messageIdx {
+		messages = append(messages, message)
+	}
+	sort.Sort(sort.Reverse(ByClock{messages}))
```
0x-r4bbit: This is a bit unfortunate, we have to sort `messages` here because they come from `messageIdx` which is a map (hence, no sorting guaranteed).

See the comment in the code (L:2239) on why this is.

@Samyoul this can probably completely be removed if we go with attachments-index-to-content-type approach we've discussed (just didn't get to implementing this yet)

protocol/message_persistence.go
```diff
@@ -2207,3 +2181,161 @@ func (db sqlitePersistence) SetContactRequestState(id string, state common.Conta
...
+
+	return messages, cursors, nil
+}
```
0x-r4bbit, line 2325: `getMessagesAndCursorsFromScanRows()` is almost identical to `getMessagesFromScanRows()`, however where it's used a `cursor` parameter in the query is required.

I could probably merge the two and add a `cursor: bool` flag or something.

protocol/message_persistence.go
```diff
@@ -2207,3 +2181,161 @@ func (db sqlitePersistence) SetContactRequestState(id string, state common.Conta
...
+	}
+
+	var messages []*common.PinnedMessage
+	for _, message := range messageIdx {
+		messages = append(messages, message)
+	}
+	sort.Sort(sort.Reverse(ByClockPinned{messages}))
+
+	return messages, cursors, nil
+}
```
0x-r4bbit, line 2363: Very similar here: `getPinnedMessagesAndCursorsFromScanRows()` just operates on a different type (`common.PinnedMessage` instead of `common.Message`)

Any way to make this nicer in go?

protocol/message_persistence.go
0x-r4bbit: > this can probably completely be removed if we go with attachments-index-to-content-type approach we've discussed (just didn't get to implementing this yet)

Although... we still need the `contentType` so probably not an option... happy to discuss offline as well

protocol/message_persistence.go
Samyoul: You are basically doing the best you can here, because you are choosing which property of `Message` to sort by. It is annoying, but for structs you have multiple fields that a sorting algorithm has no way to default to. There is a way we can make this code a bit more reusable.

```go
type HasClocks interface {
    GetClock(i int) int
}

func (m *Messages) GetClock(i int) int {
    return m.Messages[i].Clock
}

func (m *PinnedMessages) GetClock(i int) int {
    return m.PinnedMessages[i].Message.Clock
}

func SortByClock(msgs HasClocks) {
    sort.Slice(msgs, func(i, j int) bool {
        return msgs.GetClock(j) < msgs.GetClock(i)
    })
}
```

Notice in the `sort.Slice` func I compare `j < i` this will allow for reverse sorting by default.
Also you no longer need to implement the `sort.Interface`

0x-r4bbit: Oh this is much nicer, thank you! Will update

protocol/message_persistence.go
```diff
@@ -2207,3 +2181,161 @@ func (db sqlitePersistence) SetContactRequestState(id string, state common.Conta
...
+			msg = &message
+		} else if discordMessage := msg.GetDiscordMessage(); discordMessage != nil {
+			msg.Payload = getUpdatedChatMessagePayload(discordMessage, message.GetDiscordMessage())
+		}
+	}
+	return msg, nil
+}
+
+type Messages []*common.Message
+type PinnedMessages []*common.PinnedMessage
```
Samyoul: I would also move these declarations into the `common` package so that they can benefit from the some importability that the base structs have.

protocol/message_persistence.go
0x-r4bbit: Alright so, I gave this a try but couldn't get it to compile.
It seems it is required that we move the `Messages` and `PinnedMessages` types into a struct and implement `HasClocks` there, because go doesn't recognize `HasClocks` being implemented on these types otherwise.
Once moved into a struct, the sort.Slice doesn't seem to work anymore.

So for now I've rolled back to what I had before, renamed `ByClock` and `ByClockPinned` to `MessagesWrapper` and `PinnedMessagesWrapper` and moved them to `common`.

0x-r4bbit: As discussed offline, I made a mistake with the type annotations.
This is no implemented.

0x-r4bbit:
@Samyoul updated and also moved the `PinnedMessages` and `Messages` definitions to `common`

Samyoul (approved):
Hmm, this is weird I approved this PR yesterday on the Github app, but I can clearly see it isn't approved here. 🤔

This all looks great, thank you.

0x-r4bbit:
Thank you for taking the time!

PR #2870, opened 2022-09-20
feat(MediaServer): add support for discord author and attachment URLs

0x-r4bbit:


protocol/messenger.go
```diff
@@ -4164,6 +4165,35 @@ func (m *Messenger) prepareMessage(msg *common.Message, s *server.MediaServer) {
...
+				hasPayload, err := m.persistence.HasDiscordMessageAttachmentPayload(attachment.Id)
+				if err != nil {
+					m.logger.Error("failed to check if message attachment exist", zap.Error(err))
+					continue
+				}
+				if hasPayload {
+					localURL := s.MakeDiscordAttachmentURL(dm.Id, attachment.Id)
+					dm.Attachments[idx].LocalUrl = localURL
+				}
+			}
```
Samyoul, line 4191: If the attachment is a type of image do something, else ignore the attachment.

I presume in the future this can be added to

Samyoul (approved):
Looks great, thank you

protocol/messenger.go
0x-r4bbit: I assume with "ignore the attachment" you mean removing it from the payload entirely?

PR #2871, opened 2022-09-20
chore: `UpdateKeycardUID` function exposed

saledjenic:


osmaczko (approved):
lgtm

multiaccounts/keypairs/database.go
```diff
@@ -161,31 +162,32 @@ func (kp *KeyPairs) SetKeycardName(kcUID string, kpName string) (err error) {
...
-func (kp *KeyPairs) updateKeycardLocked(kcUID string, locked bool) (err error) {
-	update, err := kp.db.Prepare(`
-		UPDATE 
-			keypairs 
-		SET 
-			keycard_locked = ?
-		WHERE 
-			keycard_uid = ?
-	`)
+func (kp *KeyPairs) updateQuery(kcUID string, field string, value interface{}) (err error) {
```
stefandunca: Just a note. I consider showing the intention is a better naming strategy than showing the implementation; something like the following might give a hint on what the function is supposed to do `updateValueForUid`, `updateKeypairsForUid`. My first thought was that the function only generates the query but I see it is also executing it.

stefandunca (approved):

PR #2873, opened 2022-09-21
Setting owner roles during community import

endulab:
Set owner roles during import community.


protocol/communities/manager.go
```diff
@@ -1200,7 +1200,7 @@ func (m *Manager) InviteUsersToCommunity(communityID types.HexBytes, pks []*ecds
 	return m.inviteUsersToCommunity(community, pks)
 }
 
-func (m *Manager) AddMemberToCommunity(communityID types.HexBytes, pk *ecdsa.PublicKey) (*Community, error) {
+func (m *Manager) AddMemberAdminToCommunity(communityID types.HexBytes, pk *ecdsa.PublicKey) (*Community, error) {
```
0x-r4bbit: Don't we still want to keep the `AddMemberToCommunity()` API around?

protocol/communities/manager.go
```diff
@@ -1209,7 +1209,7 @@ func (m *Manager) AddMemberToCommunity(communityID types.HexBytes, pk *ecdsa.Pub
 		return nil, ErrOrgNotFound
 	}
 
-	err = community.AddMember(pk)
+	err = community.AddMember(pk, []protobuf.CommunityMember_Roles{protobuf.CommunityMember_ROLE_ALL})
```
0x-r4bbit, line 1214: Could be wrong but should admins really get `ROLE_ALL`? Thought `ROLE_MANAGE_USERS` would be more appropriate.

0x-r4bbit (commented):
Just a few comments, Otherwise LGTM

protocol/communities/manager.go
endulab: It is used only in ImportCommunity(). If it was used in a couple of places also for adding regular members I would leave it. Now, I think that maybe Owner (instead of Admin) will be more correct word, so probably I will rename it.

protocol/communities/manager.go
endulab: Every owner has ROLE_ALL.

0x-r4bbit: Admin != owner but I think you pointed that out here ` Now, I think that maybe Owner (instead of Admin) will be more correct word, so probably I will rename it.`

endulab: Yeah, I've just changed the name to AddMemberOwnerToCommunity

0x-r4bbit (approved):

protocol/communities/manager.go
```diff
@@ -1200,7 +1200,7 @@ func (m *Manager) InviteUsersToCommunity(communityID types.HexBytes, pks []*ecds
 	return m.inviteUsersToCommunity(community, pks)
 }
 
-func (m *Manager) AddMemberToCommunity(communityID types.HexBytes, pk *ecdsa.PublicKey) (*Community, error) {
+func (m *Manager) AddMemberOwnerToCommunity(communityID types.HexBytes, pk *ecdsa.PublicKey) (*Community, error) {
```
osmaczko, line 1205: This name sounds a little strange
```suggestion
func (m *Manager) AddOwnerToCommunity(communityID types.HexBytes, pk *ecdsa.PublicKey) (*Community, error) {
```

osmaczko (approved):

cammellos (approved):

PR #2876, opened 2022-09-21
Send all encryption keys

cammellos:
Changes the way we propagate keys.
Instead of sending a single key, we send a batch of keys, so that new users will be able to catch up using older keys.

This commit also makes sure that out of order messages are correctly decrypted, this happens when you for example receive a message that has a key that does not exist and you receive it later.


0x-r4bbit:
@cammellos have there been any changes other than the commits being squashed after I've rebased this?

vitvly (approved):

cammellos:
@PascalPrecht some, but not meaningful, linting etc

0x-r4bbit (approved):

PR #2879, opened 2022-09-23
fix(RequestCommunityInfoFromMailserver): Add useDatabase flag to read…

### LINKED ISSUE #2878: RequestCommunityInfoFromMailserver should not read db during community import
# Problem

During community import there is a request to mailserver to fetch recent community description.
However, RequestCommunityInfoFromMailserver first looks for community in database and returns an empty community.

# Implementation

Add flag to RequestCommunityInfoFromMailserver in order to do not check database.

endulab:
Add useDatabase flag to read/not read community info before requesting mailserver.

Fix #2878

### DIFF
--- a/services/ext/api.go
+++ b/services/ext/api.go
@@ -1032,7 +1032,7 @@ func (api *PublicAPI) EnsVerified(pk, ensName string) error {
 }
 
 func (api *PublicAPI) RequestCommunityInfoFromMailserver(communityID string) (*communities.Community, error) {
-	return api.service.messenger.RequestCommunityInfoFromMailserver(communityID)
+	return api.service.messenger.RequestCommunityInfoFromMailserver(communityID, true)
 }
 
 func (api *PublicAPI) RequestCommunityInfoFromMailserverAsync(communityID string) error {
--- a/cmd/ping-community/main.go
+++ b/cmd/ping-community/main.go
@@ -145,7 +145,7 @@ func main() {
 
 	messenger := wakuextservice.Messenger()
 
-	community, err := messenger.RequestCommunityInfoFromMailserver(*communityID)
+	community, err := messenger.RequestCommunityInfoFromMailserver(*communityID, true)
 	if err != nil {
 
 		logger.Error("community error", "error", err)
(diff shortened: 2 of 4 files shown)


cammellos (approved):

richard-ramos (approved):

0x-r4bbit (approved):

services/ext/api.go
```diff
@@ -1032,7 +1032,7 @@ func (api *PublicAPI) EnsVerified(pk, ensName string) error {
 }
 
 func (api *PublicAPI) RequestCommunityInfoFromMailserver(communityID string) (*communities.Community, error) {
-	return api.service.messenger.RequestCommunityInfoFromMailserver(communityID)
+	return api.service.messenger.RequestCommunityInfoFromMailserver(communityID, true)
```
siddarthkay, line 1035: Wouldn't passing `true` here make sure that community info is always checked from local db instead of mailserver?

My assumption is that if a community is created by another device then we would not have the community info in our device local db ? unless we fetch from mailserver? @cammellos @endulab 

Should we perhaps add a parameter in the function to accept this flag and if no flag is provided we then default to true?

siddarthkay: I would love to know your thoughts before I make this change in my upcoming PR :)

endulab: @siddarthkay 
If we get deeper into the code into communities/persistence.go GetByID() we will see that in case of ErrNoRows nil, nil (community, error) is returned. This means that if useDatabase is true, then db will be checked first. But when there is no such a community in db, then requestCommunityInfoFromMailserver will be executed.
Is it a good approach ?
Can you explain how an additional parameter will help here ?

siddarthkay: Hi @endulab  I had not dug deeper into the code and had no idea that `requestCommunityInfoFromMailserver` is executed when community is not found in db.
In this case additional parameter is not needed.

Thanks for explanation and for clarification :)

siddarthkay: btw having `requestCommunityInfoFromMailserver` and `RequestCommunityInfoFromMailserver` in the same file is really confusing  😅

PR #2884, opened 2022-09-29
feat: introduce discord import tool

0x-r4bbit:
Here it is.


ghost:
### Pull Request Checklist

protocol/discord/types.go
```diff
@@ -57,10 +113,150 @@ type ImportError struct {
...
+
+func (progress *ImportProgress) AddTaskError(task ImportTask, err *ImportError) {
+	progress.m.Lock()
+	for i, t := range progress.Tasks {
+		if t.Type == task.String() {
+			errorsAndWarningsCount := progress.Tasks[i].ErrorsCount + progress.Tasks[i].WarningsCount
+			if errorsAndWarningsCount < MaxTaskErrorItemsCount {
+				errors := progress.Tasks[i].Errors
+				progress.Tasks[i].Errors = append(errors, err)
+			}
```
0x-r4bbit, line 195: While I've built the import tool, I ran into perf issues the bigger the `errors` list of an individual task gets.
This is because, when for example the "download assets" task keeps timing out when requesting asset payloads, there's many errors added to the task (this can be hundreds or even thousands).

When the progress is then published to clients, they have a lot of data to parse in a single signal.
And since so many errors/issues aren't really usable to a user, we decided to max cap visible errors per task to 3

If there's more than 3 errors per task we only increase the counter and let clients/UIs decide how to handle this.

protocol/message_validator.go
```diff
@@ -233,8 +233,10 @@ func ValidateReceivedChatMessage(message *protobuf.ChatMessage, whisperTimestamp
 		return errors.New("timestamp can't be 0")
 	}
 
-	if err := ValidateText(message.Text); err != nil {
-		return err
+	if message.ContentType != protobuf.ChatMessage_DISCORD_MESSAGE {
+		if err := ValidateText(message.Text); err != nil {
+			return err
+		}
```
0x-r4bbit, line 239: Discord messages won't have a `Text` in their `ChatMessage` (but is later retrieved from the discord message payload in client UIs)

protocol/messenger.go
```diff
@@ -3009,6 +3012,101 @@ func (m *Messenger) outputToCSV(timestamp uint32, messageID types.HexBytes, from
 	}
 }
 
+func (m *Messenger) handleImportedMessages(messagesToHandle map[transport.Filter][]*types.Message) error {
```
0x-r4bbit, line 3048: @cammellos this is related to our discussion offline re: using a third party ID to handle imported messages differently (and not passing them to `handleRetrievedMessages()`.

Notice that the function signature of `handleImportedMessages()` is the same.
Also, the functionality is really just a trimmed down version of what `handleRetrievedMessages()` does, plus, it sets the `messageID = `msg.TransportMessage.ThirdPartyID`

We could consider using `handleRetrievedMessage()` instead and simply override `messageID` with `ThirdPartyID` if it exists. Not entirely sure if adding `handleImportMessages()` adds way more value then otherwise.

protocol/messenger_backup.go
```diff
@@ -131,37 +131,38 @@ func (m *Messenger) BackupData(ctx context.Context) (uint64, error) {
 
 	cs := append(joinedCs, deletedCs...)
 	for _, c := range cs {
+		_, beingImported := m.importingCommunities[c.IDString()]
+		if !beingImported {
+			settings, err := m.communitiesManager.GetCommunitySettingsByID(c.ID())
+			if err != nil {
+				return 0, err
+			}
```
0x-r4bbit, line 139: It's important we don't perform backups for communities that are being imported as they otherwise could show up too early in the UI when an import is still ongoing (we don't want to show communities that haven't finished importing yet)

protocol/messenger_communities.go
```diff
@@ -1947,3 +1953,802 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+		if m.DiscordImportMarkedAsCancelled(communityID) {
+			importProgress.StopTask(discord.DownloadAssetsTask)
+			progressUpdates <- importProgress
+			cancel <- communityID
+			return
+		}
+
+		for idxRange := range gopart.Partition(len(messageAttachmentsToDownload), 20) {
+			attachments := messageAttachmentsToDownload[idxRange.Low:idxRange.High]
+			wg.Add(1)
```
0x-r4bbit, line 2593: @Samyoul I've looked into worker pools (finally) and realized that it's very similar to what I'm already doing.
However, I'm happy to play around with this in a second iteration and do some profiling to see whether this improves performance.

Also notice here that I've hardcoded the partitions (for message attachments) to 20 lists.
This number worked well for real data, but might not work well for other data.

protocol/messenger_communities.go
```diff
@@ -1947,3 +1953,802 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+				return
+			case communityID := <-cancel:
+				if currentProgress != nil {
+					m.publishImportProgress(currentProgress)
+				}
+				m.cleanUpImport(communityID)
+				m.config.messengerSignalsHandler.DiscordCommunityImportCancelled(communityID)
+				return
+			case <-m.quit:
+				return
```
0x-r4bbit, line 2917: Should cancel ongoing imports and cleanup here

protocol/messenger_communities.go
```diff
@@ -1947,3 +1953,802 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+			importProgress.AddTaskError(discord.DownloadAssetsTask, discord.Error(err.Error()))
+			importProgress.Stop()
+			progressUpdates <- importProgress
+			return
+		}
+
+		importProgress.UpdateTaskProgress(discord.DownloadAssetsTask, 1)
+		progressUpdates <- importProgress
+
+		err = m.publishOrg(discordCommunity)
```
0x-r4bbit, line 2679: So this publishes the community description and it's also what causes the community to show up Desktop's UI.
This works nice as long as there's mailservers available. In face of a connection error, the community description doesn't reach its own client.

We don't actually need to make the client aware of this community via dispatching messages, but should probably do something similar to MessengerResponse handling.

protocol/messenger_handler.go
```diff
@@ -937,7 +938,26 @@ func (m *Messenger) HandleHistoryArchiveMagnetlinkMessage(state *ReceivedMessage
...
+					}
+				}
+
+				err = m.handleImportedMessages(importedMessages)
+				if err != nil {
+					log.Println("failed to handle imported messages", err)
+					m.logger.Debug("failed to write history archive messages to database", zap.Error(err))
+				}
+
+				response, err := m.handleRetrievedMessages(otherMessages, false)
```
0x-r4bbit, line 972: @cammellos here's where we decide which code path we take for imported messages.

protocol/migrations/sqlite/1664452901_add_discord_messages_and_authors_indices.up.sql
```diff
@@ -0,0 +1,4 @@
+CREATE INDEX idx_dm_author_id_dm_id ON discord_messages (id, author_id);
+CREATE INDEX idx_response_to_source_discord_message_id ON user_messages (response_to, source, discord_message_id);
+CREATE INDEX idx_pinned_response_to_source_discord_message_id ON pin_messages (message_id, pinned);
```
0x-r4bbit: I tried adding these to improve query performance.
Turns out with discord community with 200000+ messages, message querying can get slower.

protocol/communities/manager.go
```diff
@@ -337,6 +337,15 @@ func (m *Manager) CreateCommunity(request *requests.CreateCommunity, publish boo
 	return community, nil
 }
 
+func (m *Manager) DeleteCommunity(id types.HexBytes) error {
+	err := m.persistence.DeleteCommunity(id)
+	if err != nil {
+		return err
+	}
+	err = m.persistence.DeleteCommunitySettings(id)
+	return err
```
richard-ramos:
```suggestion
	return m.persistence.DeleteCommunitySettings(id)
```

protocol/discord/assets.go
```diff
@@ -0,0 +1,35 @@
...
+	imgs, err := images.GenerateIdentityImagesFromURL(url)
+	if err != nil {
+		return nil, err
+	}
+	payload := imgs[0].Payload
+	return payload, nil
+}
+
+func DownloadAsset(url string) ([]byte, string, error) {
+	client := http.Client{}
```
richard-ramos: Should the httpClient have a timeout?
```suggestion
	client := http.Client{Timeout: time.Minute} // 1 min

```

protocol/discord/types.go
```diff
@@ -57,10 +113,150 @@ type ImportError struct {
...
+			if err.Code > NoErrorType {
+				progress.Tasks[i].WarningsCount++
+			}
+		}
+	}
+	if err.Code > WarningType {
+		progress.ErrorsCount++
+		return
+	}
```
richard-ramos, line 207: You can end up here with `progress.m` forever locked. To avoid that, replace line 186 by
```
progress.m.Lock()
defer progress.m.Unlock()
```
and remove `progress.m.Unlock()` from line 209

du82:
Whats the chances that Discord hides the token from the Inspect tool just because of Status?

protocol/messenger_communities.go
```diff
@@ -26,10 +29,14 @@ import (
 	"github.com/status-im/status-go/protocol/protobuf"
 	"github.com/status-im/status-go/protocol/requests"
 	"github.com/status-im/status-go/protocol/transport"
+	v1protocol "github.com/status-im/status-go/protocol/v1"
 )
 
 // 7 days interval
-var messageArchiveInterval = 7 * 24 * time.Hour
+var messageArchivePartition = 7 * 24 * time.Hour
+var messageArchiveInterval = 2 * time.Minute
```
0x-r4bbit: This still needs to change

protocol/discord/types.go
```diff
@@ -1,26 +1,80 @@
...
-	MessageTypeReply   MessageType = "Reply"
+	MessageTypeDefault       MessageType = "Default"
+	MessageTypeReply         MessageType = "Reply"
+	MessageTypeChannelPinned MessageType = "ChannelPinnedMessage"
+)
+
+type ImportTask uint
+
+const (
+	CommunityCreationTask ImportTask = iota
```
Samyoul: Unless your first enum is and `Unknown` type eg `UnknownTask`, it is best practice to declare the first line like below:
```suggestion
	CommunityCreationTask ImportTask = iota + 1
```

This is because by default an uninitialised uint will have a `0` value, and therefore risk a collision with your first enum

protocol/discord/types.go
```diff
@@ -1,26 +1,80 @@
...
+	case InitCommunityTask:
+		return "import.initializeCommunity"
+	}
+	return "unknown"
+}
+
+type ImportTaskState uint
+
+const (
+	TaskStateInitialized ImportTaskState = iota
```
Samyoul, line 64: Unitialised uint collision issue

Samyoul (approved):
Great, lots hard work in this PR

protocol/discord/types.go
```diff
@@ -57,10 +113,151 @@ type ImportError struct {
...
+			State:         TaskStateInitialized.String(),
+		})
+	}
+	progress.ErrorsCount = 0
+	progress.WarningsCount = 0
+	progress.Stopped = false
+}
+
+func (progress *ImportProgress) Stop() {
+	progress.Stopped = true
```
stefandunca, line 182: Why not guard this also? Also, it doesn't seem to be used anywhere.

protocol/discord/types.go
```diff
@@ -57,10 +113,151 @@ type ImportError struct {
...
+		progress.ErrorsCount++
+		return
+	}
+	if err.Code > NoErrorType {
+		progress.WarningsCount++
+	}
+}
+
+func (progress *ImportProgress) StopTask(task ImportTask) {
+	for i, t := range progress.Tasks {
```
stefandunca, line 216: I'm wondering why not guard access to tasks here

protocol/message_persistence.go
```diff
@@ -1354,6 +1356,36 @@ func (db sqlitePersistence) SetHideOnMessage(id string) error {
 	return err
 }
 
+func (db sqlitePersistence) DeleteMessagesByCommunityID(id string) error {
+	return db.deleteMessagesByCommunityID(id, nil)
+}
+
+func (db sqlitePersistence) deleteMessagesByCommunityID(id string, tx *sql.Tx) (err error) {
+	if tx == nil {
```
stefandunca: Why passing transaction from outside? Is it recycled/batched somewhere?

stefandunca (approved):
Great job. See some questions from my side.

protocol/discord/types.go
0x-r4bbit: It's used when a task is stopped :)

0x-r4bbit: Might actually guard this

protocol/discord/types.go
0x-r4bbit: Should be guarding here too

protocol/message_persistence.go
0x-r4bbit: This is a leftover. Good catch!

PR #2887, opened 2022-10-04
FIX - Check and make keystore if required on receiving device

Samyoul:
## What's Changed

I've added some checks to `PairingPayloadRepository.store()` to ensure the following:

- `keyStorePath` is not empty
- That if the last directory in the `keyStorePath` is `"keystore"` the application should retrieve the `KeyUID` from the received `multiaccount`

## Why Change?

Any device calling `PairingPayloadRepository.store()` can not know its `KeyUID` and therefore does not know the full directory path to store its keys.

I've allowed for the application to be given the option to simply store the keys in any directory by default, and only attempt to build the standard keystore dir when the last directory in the `keyStorePath` is `"keystore"`.


0x-r4bbit (approved):
Changes LGTM

siddarthkay (approved):

PR #2889, opened 2022-10-05
Connection string identifier

Samyoul:
## What's Changed?

I've added a string identifier to the device pairing `ConnectionString`.

## Why Change?

We need a simple way to identify if a string is a `ConnectionString`


siddarthkay (approved):
lovely!

0x-r4bbit (approved):

PR #2895, opened 2022-10-06
feat(communities): add archiveLogger for stdout logs

0x-r4bbit:
The default logger writes to `geth.log`, which makes debugging the archive protocol pretty hard.

This adds an additional logger that logs to stdout, while keeping the default logger intact for production.

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -43,6 +43,7 @@ type Manager struct {
 	ensVerifier                  *ens.Verifier
 	identity                     *ecdsa.PublicKey
 	logger                       *zap.Logger
+	stdoutLogger                 *zap.Logger
 	transport                    *transport.Transport
 	quit                         chan struct{}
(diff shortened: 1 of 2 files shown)


osmaczko (approved):
Simple but leads to duplication. Wouldn't it be possible to have only one logger that is configured to print both to `geth.log` and stdout?

If it is crucial to keep the original logger intact then maybe you could create a delegate that calls `logger` and `stdoutLogger` behind the scenes.

You can ignore the above, at the end of the day, it does not have a big impact on the code readability/maintainability and seems to be more flexible.

0x-r4bbit:
> Wouldn't it be possible to have only one logger that is configured to print both to geth.log and stdout?
That was my original aim, but I couldn't figure out how.

Simplest move forward was to add another logger that always logs to stdout regardless.
Happy to put things into a generic `log` function which will do both loggers for these cases.

PR #2897, opened 2022-10-07
feat(communities): retry downloading archive data

0x-r4bbit:
When fetching torrent info after receiving a magnet link, it can happen that the request times out.

We want to retry downloading the data again at least once more before giving up

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -36,6 +36,8 @@ var defaultAnnounceList = [][]string{
 }
 var pieceLength = 100 * 1024
 
+var ErrTorrentTimedout = errors.New("torrent has timed out")
+
 type Manager struct {
 	persistence                  *Persistence
 	ensSubscription              chan []*ens.VerificationRecord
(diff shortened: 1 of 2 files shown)


protocol/communities/manager.go
```diff
@@ -64,8 +67,14 @@ func NewManager(identity *ecdsa.PublicKey, db *sql.DB, logger *zap.Logger, verif
 		}
 	}
 
+	stdoutLogger, err := zap.NewDevelopment()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create archive logger")
```
endulab: maybe archive -> stdout ?

protocol/communities/manager.go
```diff
@@ -88,6 +97,11 @@ func NewManager(identity *ecdsa.PublicKey, db *sql.DB, logger *zap.Logger, verif
 	return manager, nil
 }
 
+func (m *Manager) LogStdout(msg string, fields ...zap.Field) {
```
endulab: I think that the name of the func is misleading because it logs to both std and logger.
Maybe LogAndStdout will be slightly better ?

endulab (approved):

protocol/communities/manager.go
0x-r4bbit: Actually, that commit has landed already last week. Sorry. Need to rebase this one

PR #2899, opened 2022-10-07
fix: community encryption was not being marshalled

richard-ramos:


0x-r4bbit (approved):
Good catch!

0x-r4bbit:
Restarted unit tests on CI

PR #2904, opened 2022-10-12
Server Port Race Condition Fix

Samyoul:
## What's Changed?

I've introduced `portManager` which is responsible for maintaining segregated access to the `port` field via the use of `rwLock` `sync.RWMutex` and `mustRead` `sync.Mutex`
- `rwLock` establishes a standard read write mutex that allows consecutive reads and exclusive writes
- `mustRead` forces `MustGetPort()` to wait until port has a none 0 value

`mustRead` guards `MustGetPort` and only allows access to the underlying `port` field once the unlocked. This causes an dependent function to wait until the port number is available.

## Why Change?

Before this PR we had a number of issues:

- Calling `PairingServer.MakeConnectionParams()` has a race condition that intermittently returns an error because the `Server` port is not yet set
- Calling `Server.MakeBaseURL()` and all dependent functions have a race condition that intermittently returns an invalid URL because the port has not yet been set.
- `afterPortChanged` `func(port int)` used to be only called once in the server's whole life cycle, meaning updated port numbers would not trigger.

---

**NOTE FOR CODE REVIEWERS**

This PR has heavy use of mutex and although I've took great care to implement this fix / feature as sensibly as I saw fit, there may very well be things I've overlooked. Don't be fooled by the simplicity of the PR.

---

**NOTE FOR QA**

This PR has the potential to cause unexpected hanging on getting

- audio
- identicons
- images
- ipfs
- account Images (profile picture)
- contact Images (contact profile pictures)


server/ports.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"fmt"
+	"sync"
+)
+
+// portManager is responsible for maintaining segregated access to the port field
+// via the use of rwLock sync.RWMutex and mustRead sync.Mutex
+// rwLock establishes a standard read write mutex that allows consecutive reads and exclusive writes
+// mustRead forces MustGetPort to wait until port has a none 0 value
+type portManger struct {
```
0x-r4bbit, line 11: Should this be `PortManager` instead? Or is that so that it's not becoming a public API

server/ports.go
```diff
@@ -0,0 +1,81 @@
...
+	rwLock           *sync.RWMutex
+	mustRead         *sync.Mutex
+}
+
+// newPortManager returns a newly initialised portManager with a pre-Locked portManger.mustRead sync.Mutex
+func newPortManager(afterPortChanged func(int)) portManger {
+	pm := portManger{
+		afterPortChanged: afterPortChanged,
+		rwLock:           new(sync.RWMutex),
+		mustRead:         new(sync.Mutex),
```
0x-r4bbit: Out of curiosity: doing `new(SomeType)` is just another way of doing `&SomeType{}` ?

server/ports.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	pm.mustRead.Lock()
```
0x-r4bbit: Locking it right away first looked off to me, but reading your comments below, it seems like a good trick to block execution until a port has been set

0x-r4bbit (approved):

server/ports.go
Samyoul: Yes, it ideally needs to be an internal struct and functions. It could be exported but I can't think of any other part of the application that would use this specific set of functionality.

Also I wanted the `portManager` to exist to be the only place that port handling occurred, so it was clear that only `portManager` should do port related activity.

But I 100% do not want any dependent code trying to interact with the internal workings of `portManager` because this will definitely lead to deadlocks and infinite hangs.

server/ports.go
Samyoul: Yes, that's correct. It is kind of the "correct" way to do it, but it is generally fine to declare a struct literal pointer. The main advantage of declaring a with `new` is that it can work for any literal pointer type, see this example of its usefulness https://go.dev/play/p/5OoeMXnJsb-

server/ports.go
Samyoul: So, this is a little hacky, mutex doesn't perfectly fit this context, another type of sync struct is a `waitgroup`, but a `waitgroup` applies a cumulative number of deltas to wait, and `ResetPort` forces an additional `Add()` call, but `ResetPort` can be called `n` times before `MustGetPort` is called. This would add `n` wait deltas to the group, which can't be removed with a single `Done()` call.

Another approach could be preventing `ResetPort` from executing if there is a lock, but I don't know how to do this without using another mutex to guard the function.

I also considered resetting the whole `portManager`, but I've realised that there is no way to reset a mutex already blocking without crashing any routine waiting to acquire the lock.

pavloburykh:
Hi @Samyoul, thank you for the PR !

I've performed regression testing on mobile here https://github.com/status-im/status-mobile/pull/14167. Please, take a look at the following issues:

### ISSUE 1 Share/save image buttons stop responding after image was once shared/saved

**Steps:**

1. Send some image in any kind of chat
2. Open the sent image 
3. Tap save or share button
4. Try to save or share image (you can try this action on any other image in chat) once again
5. Observe the result

**Actual result:** 
tapping on save button results in closing image but the image is not saved. 
tapping on share image doesn not result in opening bottom sheet with sending options


### ISSUE 2 Endless loader after opening image in chat

**Steps:**

1. Send 2 or more images in any kind of chat
2. Open any of sent image 
3. Tap save or share button
4. Close the image
5. Try to open other images
6. Observe the result

**Actual result:** Endless loader, image is not opened


Both issues are reproducible on Android. On IOS it is not reproducible every time, though I also faced the described behaviour on IOS too. 

Video and logs for both ISSUE 1 and ISSUE 2 are attached below 

https://user-images.githubusercontent.com/97245802/196207591-041fa595-43a9-470c-96c9-f4a2d76bb0bb.mp4

[geth.log](https://github.com/status-im/status-mobile/files/9801631/geth.log)
[Status.log](https://github.com/status-im/status-mobile/files/9801632/Status.log)

Samyoul:
A little update:

So what seems to be happening, I think, is that sharing or saving puts the application temporally into the background, which triggers the `MediaServer` to restart, which may or may not regenerate a new port number which may or may not be updated in the UI. I've moved when a signal is triggered but it triggers only when the port number updates, this is expected behaviour, but the UI may or may not be responding to this.

Another potential line of inquiry is that he UI may have a really aggressive timeout on localhost requests.

richard-ramos (approved):

PR #2906, opened 2022-10-13
fix(MessengerResponse): ensure `savedAddresses` response as `omitempty`

0x-r4bbit:
Otherwise the json response will be `null` instead of an empty array
(which broke desktop): https://github.com/status-im/status-desktop/pull/7888


stefandunca (approved):
@PascalPrecht thanks for the great finding

caybro (approved):

PR #2907, opened 2022-10-14
feat(communities): add support for archive encryption

0x-r4bbit:
Only the **last** commit of this PR handles encryption.

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -23,6 +23,7 @@ import (
 	"github.com/status-im/status-go/eth-node/types"
 	"github.com/status-im/status-go/params"
 	"github.com/status-im/status-go/protocol/common"
+	"github.com/status-im/status-go/protocol/encryption"
 	"github.com/status-im/status-go/protocol/ens"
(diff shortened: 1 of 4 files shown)


ghost:
### Pull Request Checklist

0x-r4bbit:
This needs #2876 to land first

protocol/communities/manager.go
```diff
@@ -2053,8 +2079,29 @@ func (m *Manager) ExtractMessagesFromHistoryArchives(communityID types.HexBytes,
 
 		err := proto.Unmarshal(data, archive)
```
cammellos, line 2080: Are you sure this fails for encrypted archives?
If they are both protobuf structs encoded, but different types, generally it doesn't fail, but will be an empty struct with some extra fields

0x-r4bbit: Yea so this is interesting.
I've noticed that, when it's decrypted on the community owner side, it indeed doesn't error out. But on the community member side it will error.

Will verify this once more, but this is why I ended up handling it this way.

Samyoul (approved):

PR #2908, opened 2022-10-19
feat: add API endpoint to fetch token balance history for graph plotting purpose

stefandunca:
Add API endpoint to fetch account balance history

Add functionality to sample and retrieve balance history and cache it in memory for the current transfer controller.

The end of the balance history is snapped at twice per day to avoid having to query the blockchain again for each fetching within 12 hours interval
  
The functionality will be extended with DB caching, API call batching, "smarter" cache hitting and syncing between devices

Updates [#7662](https://github.com/status-im/status-desktop/issues/7662)

### DIFF
--- a/services/wallet/api.go
+++ b/services/wallet/api.go
@@ -63,10 +63,22 @@ func (api *API) CheckRecentHistoryForChainIDs(ctx context.Context, chainIDs []ui
 	return api.s.transferController.CheckRecentHistory(chainIDs, addresses)
 }
 
+func hexBigToBN(hexBig *hexutil.Big) *big.Int {
+	var bN *big.Int
+	if hexBig != nil {
+		bN = hexBig.ToInt()
+	}
+	return bN
+}
+
 // GetTransfersByAddress returns transfers for a single address
 func (api *API) GetTransfersByAddress(ctx context.Context, address common.Address, toBlock, limit *hexutil.Big, fetchMore bool) ([]transfer.View, error) {
 	log.Debug("[WalletAPI:: GetTransfersByAddress] get transfers for an address", "address", address)
-	return api.s.transferController.GetTransfersByAddress(ctx, api.s.rpcClient.UpstreamChainID, address, toBlock, limit, fetchMore)
+	var intLimit = int64(1)
+	if limit != nil {
+		intLimit = limit.ToInt().Int64()
+	}
+	return api.s.transferController.GetTransfersByAddress(ctx, api.s.rpcClient.UpstreamChainID, address, hexBigToBN(toBlock), intLimit, fetchMore)
 }
 
 // LoadTransferByHash loads transfer to the database
@@ -77,7 +89,7 @@ func (api *API) LoadTransferByHash(ctx context.Context, address common.Address, 
 
 func (api *API) GetTransfersByAddressAndChainID(ctx context.Context, chainID uint64, address common.Address, toBlock, limit *hexutil.Big, fetchMore bool) ([]transfer.View, error) {
 	log.Debug("[WalletAPI:: GetTransfersByAddressAndChainIDs] get transfers for an address", "address", address)
-	return api.s.transferController.GetTransfersByAddress(ctx, chainID, address, toBlock, limit, fetchMore)
+	return api.s.transferController.GetTransfersByAddress(ctx, chainID, address, hexBigToBN(toBlock), limit.ToInt().Int64(), fetchMore)
 }
 
 func (api *API) GetCachedBalances(ctx context.Context, addresses []common.Address) ([]transfer.LastKnownBlockView, error) {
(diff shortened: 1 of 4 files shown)


services/wallet/api.go
```diff
@@ -63,10 +63,18 @@ func (api *API) CheckRecentHistoryForChainIDs(ctx context.Context, chainIDs []ui
...
+		bN = hexBig.ToInt()
+	}
+	return bN
+}
+
 // GetTransfersByAddress returns transfers for a single address
 func (api *API) GetTransfersByAddress(ctx context.Context, address common.Address, toBlock, limit *hexutil.Big, fetchMore bool) ([]transfer.View, error) {
 	log.Debug("[WalletAPI:: GetTransfersByAddress] get transfers for an address", "address", address)
-	return api.s.transferController.GetTransfersByAddress(ctx, api.s.rpcClient.UpstreamChainID, address, toBlock, limit, fetchMore)
+	return api.s.transferController.GetTransfersByAddress(ctx, api.s.rpcClient.UpstreamChainID, address, hexBigToBN(toBlock), limit.ToInt().Int64(), fetchMore)
```
alaibe: limit can be nil too as this is a pointer.
Also is `.Int64()` valid? I'd agree it is weird to have this as a big int due, logically hard to make sense but better verify

services/wallet/api.go
```diff
@@ -105,6 +113,15 @@ func (api *API) GetTokensBalancesForChainIDs(ctx context.Context, chainIDs []uin
 	return api.s.tokenManager.GetBalances(ctx, clients, accounts, addresses)
 }
 
+// GetBalanceHistoryOnChain retrieves native token. Will be extended later to support token balance history
+func (api *API) GetBalanceHistoryOnChain(ctx context.Context, address common.Address, secondsToNow int64, sampleCount int) ([]transfer.BalanceState, error) {
```
alaibe: nit GetBalanceHistoryOnChain => GetBalanceHistory

If you want to support none native token you need to add the symbol/address of the token

services/wallet/api.go
```diff
@@ -105,6 +113,15 @@ func (api *API) GetTokensBalancesForChainIDs(ctx context.Context, chainIDs []uin
...
+	transactionPoints, err := api.s.transferController.GetBalanceHistoryOnChain(ctx, api.s.rpcClient.UpstreamChainID, address, secondsToNow, sampleCount)
```
alaibe: `api.s.rpcClient.UpstreamChainID` you should pass the chain id as we don't use that on desktop, this attribute is only for mobile

services/wallet/api.go
```diff
@@ -105,6 +113,15 @@ func (api *API) GetTokensBalancesForChainIDs(ctx context.Context, chainIDs []uin
...
+	if err != nil {
+		return make([]transfer.BalanceState, 0), err
+	}
+	return transactionPoints, err
```
alaibe: nit: not need to handle the error right, you can directly do:

`return api.s.transferController.GetBalanceHistoryOnChain(ctx, api.s.rpcClient.UpstreamChainID, address, secondsToNow, sampleCount)`

services/wallet/transfer/controller.go
```diff
@@ -298,3 +293,79 @@ func (c *Controller) GetCachedBalances(ctx context.Context, chainID uint64, addr
...
+
+type BalanceState struct {
+	Value     big.Int `json:"value"`
+	Timestamp uint64  `json:"time"`
+}
+
+var BalanceHistorySecondsToNowFromInitialBlock int64 = -1
+
+// GetBalanceHistoryOnChain expect a time precision of +/- block time (12s)
+func (c *Controller) GetBalanceHistoryOnChain(ctx context.Context, chainID uint64, address common.Address, secondsToNow int64, sampleCount int) ([]BalanceState, error) {
```
alaibe: I am curious of what is the duration for this function?

Also instead of accepting everything, wouldn't it be better if we had a enum for sampleCount and secondsToNow

e.g: secondsToNot: Last6Month and sampleCount: 182 (once a day)

services/wallet/transfer/controller.go
```diff
@@ -298,3 +293,79 @@ func (c *Controller) GetCachedBalances(ctx context.Context, chainID uint64, addr
...
+	} else {
+		firstBlockNo.Sub(lastBlockNumber, historyBlockCount)
+	}
+
+	blocksInStep := historyBlockCount.Int64() / int64(sampleCount)
+	stepDuration := blocksInStep * blockDuration
+
+	points := make([]BalanceState, 0)
+
+	balanceCache := newBalanceCache()
```
alaibe: I might be wrong but this doesn't save anything into the db, only memory, and this is quite intensive in terms of infura call, due to our limitation, could we not save the result in db so that we do it once per block

services/wallet/transfer/controller.go
stefandunca: > I am curious of what is the duration for this function?

Few seconds, depending on the sample count.

I have on my list to look into how can we batch multiple block requests in one call. I see that the underlying RPC protocol supports it, not sure if it is possible with our go-wrapper on top of the RPC.

This batching optimization should speed up at least one order of magnitude and create only a few requests with tens of blocks in each and be limited by the node performance.

services/wallet/transfer/controller.go
stefandunca: I agree. However, due to out-of-time DB caching won't make it in this PR but the functionality can be already tested. Will come back with a follow-up task for that.

alaibe (approved):

PR #2915, opened 2022-10-21
feat(communities): Cancel request to join community

borismelnik:
Introduce `CommunityCancelRequestToJoin` request.

Important changes:


protocol/messenger_handler.go
```diff
@@ -958,6 +958,21 @@ func (m *Messenger) HandleHistoryArchiveMagnetlinkMessage(state *ReceivedMessage
...
+		return errors.New("invalid community id")
+	}
+
+	requestToJoin, err := m.communitiesManager.HandleCommunityCancelRequestToJoin(signer, &cancelRequestToJoinProto)
+	if err != nil {
+		return err
+	}
+
+	state.Response.RequestsToJoinCommunity = append(state.Response.RequestsToJoinCommunity, requestToJoin)
+	fmt.Println("CHECK REQUESTS", state.Response.RequestsToJoinCommunity, requestToJoin)
```
cammellos: extra log :)

cammellos (commented):
Looks good so far, could you please adding tests to cover the new functionality please?
thank you!

protocol/messenger_handler.go
borismelnik: oh... sorry :)

borismelnik: fixed :)

0x-r4bbit (approved):
This is a nice PR

protocol/communities/manager.go
```diff
@@ -841,10 +841,43 @@ func (m *Manager) markRequestToJoin(pk *ecdsa.PublicKey, community *Community) e
...
+	if err := m.markRequestToJoinAsCanceled(pk, community); err != nil {
+		return nil, nil, err
+	}
+
+	err = m.persistence.SaveCommunity(community)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	m.publish(&Subscription{Community: community})
```
osmaczko: I don't think it should be here, requests to join are not published, besides that only community owners should publish community descriptions. The one who calls `CancelRequestToJoin` can't be a community owner, right?

0x-r4bbit: Actually, this is a very good catch

borismelnik: fixed :) thank you :)

jakubgs:
Seems to me like your weird result of `make generate` is related to your local version of `protoc`.

I undid the protobuf changes in your branch locally and the ran `make generate` but for me only timestamps changed.

jakubgs:
So it appears the issue was having multiple versions of `go-bindata` installed, where the newer one came from Homebrew.

This is annoying, because we explicitly lock `go-bindata` at `3.13.0` in the `Makefile` to avoid this issue:
https://github.com/status-im/status-go/blob/a399c04346f676854245bd9e2e5606e53a07bf23/Makefile#L253-L255
Maybe we need some additional check for this in the `Makefile`.

osmaczko (approved):

protocol/communities/manager.go
```diff
@@ -1316,6 +1374,11 @@ func (m *Manager) DeclinedRequestsToJoinForCommunity(id types.HexBytes) ([]*Requ
 	return m.persistence.DeclinedRequestsToJoinForCommunity(id)
 }
 
+func (m *Manager) CanceledRequestsToJoinForCommunity(id types.HexBytes) ([]*RequestToJoin, error) {
+	m.logger.Info("fetching canceled invitations", zap.String("community-id", id.String()))
```
cammellos, line 1378: Debug maybe?

cammellos (approved):

PR #2916, opened 2022-10-21
Fixes needed to make login with the Keycard work

saledjenic:


alaibe (approved):

services/accounts/accounts.go
```diff
@@ -221,7 +301,7 @@ func (api *API) verifyPassword(password string) error {
 	return err
 }
 
-func (api *API) addAccountWithMnemonic(
+func (api *API) addAccountWithMnemonicPasswordVerified(
```
Samyoul, line 308: Does the name `addAccountWithMnemonicPasswordVerified` mean that the password has already been verified?

If so, perhaps you could add some go docs to explain the function. Something like:

```suggestion
// addAccountWithMnemonicPasswordVerified adds an accounts.Account derived from the given Mnemonic ... (some info about the usage of the password)
func (api *API) addAccountWithMnemonicPasswordVerified(
```

Samyoul (approved):
I don't see any problem with the implementation looks good. My only question is make the function understandable for future developers, and adding some go docs to the function would greatly help with that.

services/accounts/accounts.go
saledjenic: @Samyoul thanks, yes, will do that.

PR #2925, opened 2022-10-26
feat(communities): Send signal when request to join was declined

borismelnik:
Add declined requests to message response.

Important changes:

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -929,13 +929,23 @@ func (m *Manager) GetRequestToJoin(ID types.HexBytes) (*RequestToJoin, error) {
 	return m.persistence.GetRequestToJoin(ID)
 }
 
-func (m *Manager) DeclineRequestToJoin(request *requests.DeclineRequestToJoinCommunity) error {
+func (m *Manager) DeclineRequestToJoin(request *requests.DeclineRequestToJoinCommunity) (*Community, error) {
 	dbRequest, err := m.persistence.GetRequestToJoin(request.ID)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	return m.persistence.SetRequestToJoinState(dbRequest.PublicKey, dbRequest.CommunityID, RequestToJoinStateDeclined)
+	err = m.persistence.SetRequestToJoinState(dbRequest.PublicKey, dbRequest.CommunityID, RequestToJoinStateDeclined)
+	if err != nil {
+		return nil, err
+	}
+
+	community, err := m.GetByID(dbRequest.CommunityID)
+	if err != nil {
+		return nil, err
+	}
+
+	return community, nil
 }
 
 func (m *Manager) HandleCommunityCancelRequestToJoin(signer *ecdsa.PublicKey, request *protobuf.CommunityCancelRequestToJoin) (*RequestToJoin, error) {
@@ -1362,6 +1372,9 @@ func (m *Manager) CanceledRequestsToJoinForUser(pk *ecdsa.PublicKey) ([]*Request
 
 func (m *Manager) PendingRequestsToJoinForUser(pk *ecdsa.PublicKey) ([]*RequestToJoin, error) {
 	return m.persistence.PendingRequestsToJoinForUser(common.PubkeyToHex(pk))
+}
+func (m *Manager) DeclinedRequestsToJoinForUser(pk *ecdsa.PublicKey) ([]*RequestToJoin, error) {
+	return m.persistence.DeclinedRequestsToJoinForUser(common.PubkeyToHex(pk))
 }
 
 func (m *Manager) PendingRequestsToJoinForCommunity(id types.HexBytes) ([]*RequestToJoin, error) {
--- a/services/ext/api.go
+++ b/services/ext/api.go
@@ -477,9 +477,14 @@ func (api *PublicAPI) MyPendingRequestsToJoin() ([]*communities.RequestToJoin, e
 	return api.service.messenger.MyPendingRequestsToJoin()
 }
 
-// MyCanceledRequestsToJoin returns the pending requests for the logged in user
+// MyCanceledRequestsToJoin returns the canceled requests for the logged in user
 func (api *PublicAPI) MyCanceledRequestsToJoin() ([]*communities.RequestToJoin, error) {
 	return api.service.messenger.MyCanceledRequestsToJoin()
+}
+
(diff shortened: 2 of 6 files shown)


ghost:
### Pull Request Checklist

protocol/communities/manager.go
```diff
@@ -887,13 +887,28 @@ func (m *Manager) GetRequestToJoin(ID types.HexBytes) (*RequestToJoin, error) {
...
-		return err
+		return nil, err
+	}
+
+	err = m.persistence.SetRequestToJoinState(dbRequest.PublicKey, dbRequest.CommunityID, RequestToJoinStateDeclined)
+	if err != nil {
+		return nil, err
+	}
+
+	community, err := m.GetByID(dbRequest.CommunityID)
```
cammellos, line 943: Why are you saving the community? I can't see any changes to it

services/ext/api.go
```diff
@@ -477,6 +477,11 @@ func (api *PublicAPI) MyPendingRequestsToJoin() ([]*communities.RequestToJoin, e
 	return api.service.messenger.MyPendingRequestsToJoin()
 }
 
+// MyDeclinedRequestsToJoin returns the pending requests for the logged in user
```
cammellos: pending?

cammellos (changes requested):
@borismelnik not sure it's a good idea to explicitly decline requests from users.
The reason is that they are used as a spam vector, so dismiss them should be effortless for the most part for the community owner.
We had the same problem in the past with private group chats, where a spammer would send thousands of requests, and the user would be answering them, generating traffic and being a slow operation in general.
This is consistent with what we do with contact requests, no notification on the other end that their request has been declined, I think it would be safer until better spam protection is in place.
cc @John-44

services/ext/api.go
borismelnik: sorry :)

borismelnik:
> @borismelnik not sure it's a good idea to explicitly decline requests from users.
> The reason is that they are used as a spam vector, so dismiss them should be effortless for the most part for the community owner.
> We had the same problem in the past with private group chats, where a spammer would send thousands of requests, and the user would be answering them, generating traffic and being a slow operation in general.
> This is consistent with what we do with contact requests, no notification on the other end that their request has been declined, I think it would be safer until better spam protection is in place.
> cc @John-44

Repeating requests is blocking from our side. If it make sense, I can add a code to blocking request when it was declined. Wdyt? @cammellos @John-44

cammellos:
> > @borismelnik not sure it's a good idea to explicitly decline requests from users.
> > The reason is that they are used as a spam vector, so dismiss them should be effortless for the most part for the community owner.
> > We had the same problem in the past with private group chats, where a spammer would send thousands of requests, and the user would be answering them, generating traffic and being a slow operation in general.
> > This is consistent with what we do with contact requests, no notification on the other end that their request has been declined, I think it would be safer until better spam protection is in place.
> > cc @John-44
> 
> Repeating requests is blocking from our side. If it make sense, I can add a code to blocking request when it was declined. Wdyt? @cammellos @John-44

Not sure I understand fully, but the easiest vector of attack for a spammer is to sent 100s of requests, each from a different account, so blocking the user would not help in that case, if that's what you were suggesting, but I might have misunderstood.

borismelnik:
> > > @borismelnik not sure it's a good idea to explicitly decline requests from users.
> > > The reason is that they are used as a spam vector, so dismiss them should be effortless for the most part for the community owner.
> > > We had the same problem in the past with private group chats, where a spammer would send thousands of requests, and the user would be answering them, generating traffic and being a slow operation in general.
> > > This is consistent with what we do with contact requests, no notification on the other end that their request has been declined, I think it would be safer until better spam protection is in place.
> > > cc @John-44
> > 
> > 
> > Repeating requests is blocking from our side. If it make sense, I can add a code to blocking request when it was declined. Wdyt? @cammellos @John-44
> 
> Not sure I understand fully, but the easiest vector of attack for a spammer is to sent 100s of requests, each from a different account, so blocking the user would not help in that case, if that's what you were suggesting, but I might have misunderstood.

Yes, you got it correctly :) but anyway, the spammer is able to send 100 requests even if he didn't get notification.

jrainville:
@cammellos I'd like to clarify some points here.

Right now, spammers could create accounts and request to access a community. The admin can only ignore them and it's annoying.

If we merge this PR, what changes? One more broadcast of the denied request, right?

Was there a conversation offline? I heard that maybe we wanted to wait for the captcha before merging this?

@borismelnik what is the advantage of merging this vs doing what @cammellos is suggesting; ie just dismissing the request?
Do we really want the person requesting access to know they have been refused? Maybe that last question is more for designers.

cammellos:
> If we merge this PR, what changes? One more broadcast of the denied request, right?

Correct, and also the admin need to make way more effort in terms of resources to deny the request (transferring data over the wire is way more costly than just making an entry in the database).

I think this is the kind of stuff that doesn't matter for the most part, most of the times, honest people will not be denied access to a community (spammers though will for sure), so it's just useful in a very small subset of cases, and costly/useless in more common scenarios.

The way we handle this, is that a user can re-request access after 1 day or so (client side check), if the admin has already a pending request, it will just be ignored, otherwise it will be shown again. That works fine, I had a conversation with John about this and he was ok with this solution.

jrainville:
Ok makes sense. Thanks for the explanation.

Do you recommend we just drop this PR/feature or wait for a captcha, @cammellos ?

cammellos:
@jrainville I think this is not needed, yes, captcha we also need to have a conversation, we could schedule a meeting to discuss it if you'd like?

jrainville:
> we could schedule a meeting to discuss it if you'd like?

jrainville:
> we could schedule a meeting to discuss it if you'd like?

@cammellos yeah. Do you want John to be there to discuss the requirements? Or you want it more for the software part?

cammellos:
> > we could schedule a meeting to discuss it if you'd like?
> 
> @cammellos yeah. Do you want John to be there to discuss the requirements? Or you want it more for the software part?

No need for John, it's more on the technical side, since we don't know if hcaptcha will do more good than harm

jrainville:
@cammellos ok perfect. You can schedule the meeting whenever you have time. Any time after 9am EST is good with me.
You can invite whoever you feel appropriate. Add Boris too, he'll probably be the one doing it on our side since he did all the new community join flow.

PR #2926, opened 2022-10-26
fix: wakuv2 mailserver cycle ticker

richard-ramos:


0x-r4bbit (approved):

0x-r4bbit:
Ah, this doesn't pass the linter.

PR #2934, opened 2022-10-28
Reintroduce api endpoint

cammellos:
mistakenly removed


0x-r4bbit (approved):
Tested this. Works well again. Thanks for the quick fix!

jrainville (approved):

PR #2946, opened 2022-11-04
fix(communities): improve error message about duplicated chat names

0x-r4bbit:
We don't allow multiple channels with the same name in communities. Discord allows for multiple channels with the same name (living in different categories), so this is an error case in our import tool.

This commit improves the user facing error message of this scenario.

### DIFF
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/ecdsa"
 	"encoding/json"
+	_errors "errors"
 	"fmt"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


richard-ramos (approved):

protocol/messenger_communities.go
```diff
@@ -2300,7 +2300,11 @@ func (m *Messenger) RequestImportDiscordCommunity(request *requests.ImportDiscor
 			communityWithChats, changes, err := m.communitiesManager.CreateChat(discordCommunity.ID(), communityChat, false)
 			if err != nil {
 				m.cleanUpImport(communityID)
-				importProgress.AddTaskError(discord.ChannelsCreationTask, discord.Error(err.Error()))
+				errmsg := err.Error()
+				if err == communities.ErrInvalidCommunityDescriptionDuplicatedName {
```
richard-ramos: This although is valid, it's not a recommended practice:

```suggestion
				if errors.Is(err, communities.ErrInvalidCommunityDescriptionDuplicatedName) {
```

PR #2947, opened 2022-11-04
feat(communities): add community images to import progress signal

0x-r4bbit:
This is needed so clients can render the selected community image while the import is running.

### DIFF
--- a/protocol/discord/types.go
+++ b/protocol/discord/types.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/status-im/status-go/images"
 	"github.com/status-im/status-go/protocol/protobuf"
 )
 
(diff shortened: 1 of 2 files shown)


protocol/discord/types.go
```diff
@@ -149,14 +150,15 @@ type ImportTaskProgress struct {
...
 type ImportProgress struct {
-	CommunityID   string                `json:"communityId,omitempty"`
-	CommunityName string                `json:"communityName"`
-	Tasks         []*ImportTaskProgress `json:"tasks"`
-	Progress      float32               `json:"progress"`
-	ErrorsCount   uint                  `json:"errorsCount"`
-	WarningsCount uint                  `json:"warningsCount"`
-	Stopped       bool                  `json:"stopped"`
-	m             sync.Mutex
+	CommunityImages map[string]images.IdentityImage `json:"communityImages"`
```
jrainville: This is not alphabetically ordered, so it'd make more sense to but the Images after the Name or something (let ID be first)

jrainville (approved):
Looks good

protocol/discord/types.go
0x-r4bbit: Sure, fair enough!

PR #2948, opened 2022-11-04
test(waku2): log all envelope IDs for tracing message loss

### LINKED ISSUE #2939: wakuv2: increase connection timeout to 30s
Under slow network conditions, the timeout configured in status-go might be too low to connect to a wakuv2 peer

richard-ramos:
The following go build -tags need to be added: `gowaku_skip_migrations` ,  
and if the build is done for ios:`nowatchdog gowaku_skip_migrations`.

I added these tags already in the Makefile, including mobile targets, but dunno if status-mobile is using this makefile or doing the build via nix, so tagging @jakubgs and @cammellos to help me figure out this :pray: 

Also fixes #2939


cammellos (approved):

jakubgs:
The mobile build does not use the `Makefile`:
https://github.com/status-im/status-mobile/blob/develop/nix/status-go/mobile/build.nix

PR #2950, opened 2022-11-07
Sync community encryption keys

cammellos:
Community keys needs to be synced with other devices, since only the admin propagates them, but if you add another device, admin won't be there to babysit, so you have to sync the keys yourself.


0x-r4bbit (approved):

protocol/encryption/persistence.go
```diff
@@ -871,7 +871,40 @@ func (s *sqlitePersistence) SaveHashRatchetKey(
...
+
+	err = stmt.QueryRow(groupID, keyID).Scan(&result)
+	if err != nil && err != sql.ErrNoRows {
+		return err
+	} else if err == nil {
+		// already in the database, nothing to do
+		return nil
+	}
+
+	stmt, err = tx.Prepare(`INSERT OR IGNORE INTO hash_ratchet_encryption(group_id, key_id, key)
```
cammellos: remove this

PR #2953, opened 2022-11-08
fix(edit-message): Add check ContentType for request and original message

borismelnik:
When user edit message, status-go can replace content type of message to another. That issue was in desktop app which never send content type of edited message and content type of original message was changed to `unknown`

Add check of types in request and original message.

Important changes:

### DIFF
--- a/protocol/messenger_edit_message_test.go
+++ b/protocol/messenger_edit_message_test.go
@@ -94,8 +94,9 @@ func (s *MessengerEditMessageSuite) TestEditMessage() {
 
 	editedText := "edited text"
 	editedMessage := &requests.EditMessage{
-		ID:   messageID,
-		Text: editedText,
+		ID:          messageID,
+		Text:        editedText,
+		ContentType: protobuf.ChatMessage_TEXT_PLAIN,
 	}
 
 	sendResponse, err = theirMessenger.EditMessage(context.Background(), editedMessage)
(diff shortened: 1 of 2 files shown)


Samyoul (approved):

0x-r4bbit:
So... this is an interesting scenario. As @caybro pointed out in https://github.com/status-im/status-desktop/pull/8168#issuecomment-1308428006 this essentially locks you into a certain message type, preventing you from, say, sending an emoji message then editing this to be a plain text message.

What would be the implications of determining the message content type in the client when editing and then send the new content type along?

cammellos:
> So... this is an interesting scenario. As @caybro pointed out in [status-im/status-desktop#8168 (comment)](https://github.com/status-im/status-desktop/pull/8168#issuecomment-1308428006) this essentially locks you into a certain message type, preventing you from, say, sending an emoji message then editing this to be a plain text message.
> 
> What would be the implications of determining the message content type in the client when editing and then send the new content type along?

more unknowns and likely more risk of bugs, this one was a good catch, I think locking to being unable to change content type is the best thing to do for now, at least for mvp

0x-r4bbit:
Alright, fair enough

0x-r4bbit (approved):

borismelnik:
> So... this is an interesting scenario. As @caybro pointed out in [status-im/status-desktop#8168 (comment)](https://github.com/status-im/status-desktop/pull/8168#issuecomment-1308428006) this essentially locks you into a certain message type, preventing you from, say, sending an emoji message then editing this to be a plain text message.
> 
> What would be the implications of determining the message content type in the client when editing and then send the new content type along?

I though about 2 scenario:
1. Just checking if content type property in request contains expected value (text or emoji)
2.  Check is request content type is equal to original message type (@Samyoul message was recognized by me as best solution)

Anyway, we not able to change message type from message editor, and message with just emoji also can be a text message.

richard-ramos (approved):

PR #2954, opened 2022-11-08
fix: limit store queries to 10 content topics

### LINKED ISSUE #2930: wakuv2: limit queries to 10 content topics per history request
Might be solved in go-waku with https://github.com/status-im/go-waku/issues/354

richard-ramos:
Fixes #2930 

cc: @LNSD

### DIFF
--- a/protocol/messenger_mailserver.go
+++ b/protocol/messenger_mailserver.go
@@ -21,10 +21,14 @@ import (
 
 // tolerance is how many seconds of potentially out-of-order messages we want to fetch
 var tolerance uint32 = 60
+
 var mailserverRequestTimeout = 10 * time.Second
 var oneMonthInSeconds uint32 = 31 * 24 * 60 * 60
 var mailserverMaxTries uint = 2
 var mailserverMaxFailedRequests uint = 2
+
+// maxTopicsPerRequest sets the batch size to limit the number of topics per store query
+var maxTopicsPerRequest int = 10
 
 var ErrNoFiltersForChat = errors.New("no filter registered for given chat")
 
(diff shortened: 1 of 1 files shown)


protocol/messenger_mailserver.go
```diff
@@ -21,11 +21,15 @@ import (
...
 // tolerance is how many seconds of potentially out-of-order messages we want to fetch
 var tolerance uint32 = 60
+
 var mailserverRequestTimeout = 10 * time.Second
 var oneMonthInSeconds uint32 = 31 * 24 * 60 * 60
 var mailserverMaxTries uint = 2
 var mailserverMaxFailedRequests uint = 2
 
+// maxTopicsPerRequest sets the batch size to limit the number of topics per store query
+var maxTopicsPerRequest int = 20
```
LNSD: We were thinking of limiting it to 10, which is a lot. And recommend providing 2/3 as much (for the current SQLite implementation).

More or less, the reasoning is:
 
> A "broader" query, with many content topics (strings), implies many more comparisons during the query execution. This translates into slower queries.

LNSD (approved):
Only one concern with the limit; the rest, more or less, LGTM ✅ 

-----

At this moment, the only error that the nwaku implementation can return is: `INVALID_CURSOR`. This forces us to temporarily use this error in the  `nwaku` node when any query is invalid. So, it should be interpreted as _"Bad request"_.

Some changes in the waku protocol wire format are required for better error reporting. I will do this work as part of the following `nwaku` release (v0.14.0).

protocol/messenger_mailserver.go
LNSD: I just noticed that this PR's title states 10 topics, but you set the limit to 20 😜

richard-ramos: :man_facepalming:

alaibe (approved):

PR #2962, opened 2022-11-16
feat: cache balance history in DB

stefandunca:
Implement fetching of balance history from the database if it exists, otherwise, retrieves it using RPC calls. Every balance is stored in DB as an entry.

Also fetch regular balance history for all enabled networks, accounts and tokens. Fetching for a specific configuration will pause the full update which will resume after the specific configuration is fetched.

The granularity of data points returned is defined by the constant increment step defined in `timeIntervalToHoursPerStep` for each interval. The `timeIntervalToHoursPerStep` values have a common divisor to have a cache hit between requests.

To ensure cache hit, the requested blocks are "snapped" to specific time points defined by the `timeIntervalToSnapUnit` enumeration

The cached data is stored in `balance_history` table.
The "calibration" data, special case, used for bloc average timestamp is stored with address `0x0` and balance is not fetched/used.

Uniqueness constrained is enforced by the `balance_history_identify_entry` UNIQUE index.

Optimal DB fetching is ensured by the `balance_history_filter_entries` index

Tests are implemented using a mock of `BlockInfoSource` interface used to intercept the RPC calls.
`getBalanceHistoryFromBlocksSource` contains main functionality independent of the `chainClientSource`, `BlockInfoSource`'s concrete implementation.

Notes:

- the timestamp for a block balance is approximate and based on a very rough heuristic: blocks in between data points have a constant timestamp (average of the first and last block in the one-year rage)
- this is a complete refactoring of the logic used to fetch due to the nature of the heuristic chosen to ensure cache hit independent of the time interval if the granularity matches the cache
- moved balance history implementation from the `transaction.Controller` to `transaction.BalanceHistory` struct

Updates status-desktop #8226


ghost:
### Pull Request Checklist

appdatabase/migrations/sql/1671451982_add_balance_history.up.sql
```diff
@@ -0,0 +1,15 @@
+-- One entry per chain_id, address, currency, block
+-- bitset is used to select time interval granularity. The least significant bit has the coarsest granularity
+-- balance is a serialized big.Int
+CREATE TABLE IF NOT EXISTS balance_history (
+    chain_id UNSIGNED BIGINT NOT NULL,
+    address VARCHAR NOT NULL,
+    currency VARCHAR NOT NULL,
```
alaibe, line 7: I believe we should always store the value in USD, and convert only if needed.
WDYT?

appdatabase/migrations/sql/1671451982_add_balance_history.up.sql
```diff
@@ -0,0 +1,15 @@
...
+    block BIGINT NOT NULL,
+    timestamp INT NOT NULL,
+    bitset INT NOT NULL,
+    balance BLOB
+);
```
alaibe, line 12: Shall we also add token_address ?

appdatabase/migrations/sql/1671451982_add_balance_history.up.sql
```diff
@@ -0,0 +1,15 @@
...
+-- bitset is used to select time interval granularity. The least significant bit has the coarsest granularity
+-- balance is a serialized big.Int
+CREATE TABLE IF NOT EXISTS balance_history (
+    chain_id UNSIGNED BIGINT NOT NULL,
+    address VARCHAR NOT NULL,
+    currency VARCHAR NOT NULL,
+    block BIGINT NOT NULL,
+    timestamp INT NOT NULL,
+    bitset INT NOT NULL,
+    balance BLOB
```
alaibe, line 11: Why a blob? Can we store the hex value?

protocol/protobuf/chat_message.pb.go
```diff
@@ -161,7 +161,6 @@ func (m *StickerMessage) GetPack() int32 {
 type ImageMessage struct {
 	Payload              []byte    `protobuf:"bytes,1,opt,name=payload,proto3" json:"payload,omitempty"`
 	Type                 ImageType `protobuf:"varint,2,opt,name=type,proto3,enum=protobuf.ImageType" json:"type,omitempty"`
-	AlbumID   			 string    `protobuf:"bytes,3,opt,name=album_id,json=albumId,proto3" json:"album_id,omitempty"`
```
alaibe, line 164: I am surprise to see modification here?

services/wallet/transfer/balance_history.go
```diff
@@ -0,0 +1,728 @@
...
+func (src *chainClientSource) Currency() string {
+	return src.currency
+}
+
+func (src *chainClientSource) TimeNow() int64 {
+	return time.Now().UTC().Unix()
+}
+
+func (bh *BalanceHistory) StartBalanceHistory(rpcClient *rpc.Client, networkManager *network.Manager, tokenManager *token.Manager) {
+	go func() {
```
alaibe, line 494: We need to be able to stop it when the wallet stop. you can look at how to do it in the StartWallet from the API.
Also shall this is not be triggered when we start the wallet?

services/wallet/transfer/balance_history.go
```diff
@@ -0,0 +1,728 @@
...
+const (
+	BalanceHistory7Hours BalanceHistoryTimeInterval = iota + 1
+	BalanceHistory1Month
+	BalanceHistory6Months
+	BalanceHistory1Year
+	BalanceHistoryAllTime
+)
+
+var secondsInTimeInterval = map[BalanceHistoryTimeInterval]int64{
+	BalanceHistory7Hours:  7 * 24 * 60 * 60,
```
alaibe, line 89: Why not using Duration?

services/wallet/transfer/balance_history.go
```diff
@@ -0,0 +1,728 @@
...
+	BalanceHistory1Month:  30 * 24 * 60 * 60,
+	BalanceHistory6Months: 6 * 30 * 24 * 60 * 60,
+	BalanceHistory1Year:   365 * 24 * 60 * 60,
+	BalanceHistoryAllTime: 50 * 365 * 24 * 60 * 60,
+}
+
+const secondsInHour = 60 * 60
```
alaibe, line 96: Why not using duration?

services/wallet/transfer/balance_history.go
```diff
@@ -0,0 +1,728 @@
...
+		entry.chainID = identify.chainID
+		result = append(result, entry)
+		currentEntry++
+	}
+	return result, nil
+}
+
+func (bh *BalanceHistory) getDBBalanceEntriesByTimeIntervalAndSortedAsc(identify *bhIdentity, startingAtBlock *big.Int, filter *bhFilter, maxEntries int) ([]*balanceHistoryDBEntry, error) {
+	return bh.getDBBalanceEntriesTimeSorted(identify, startingAtBlock, filter, maxEntries, true)
+}
```
alaibe, line 716: More like a personal opinion: those functions are useless, they bring very little to no value, but just confuse me.

If parameters are complex we can create a struct to pass to the function

If it is too much, in this context, you could create 2 type, for the orders: Asc and Desc

services/wallet/transfer/balance_history.go
```diff
@@ -0,0 +1,728 @@
...
+	if startingAtBlock == nil {
+		startingAtBlock = big.NewInt(0)
+	}
+	// We are interested in order by timestamp, but we request by block number that correlates to the order of timestamp and it is indexed
+	var queryStr string
+	if asc {
+		queryStr = "SELECT block, timestamp, balance FROM balance_history WHERE chain_id = ? AND address = ? AND currency = ? AND block >= ? AND timestamp BETWEEN ? AND ? AND (bitset & ?) > 0 ORDER BY block ASC LIMIT ?"
+	} else {
+		queryStr = "SELECT block, timestamp, balance FROM balance_history WHERE chain_id = ? AND address = ? AND currency = ? AND block >= ? AND timestamp BETWEEN ? AND ? AND (bitset & ?) > 0 ORDER BY block DESC LIMIT ?"
+	}
```
alaibe, line 686: no need to rewrite the whole string, using `fmt.Sprint` function should do it

services/wallet/transfer/balance_history.go
```diff
@@ -0,0 +1,728 @@
...
+		return nil, 0, err
+	}
+	res.block.Set(block)
+	return
+}
+
+type bhIdentity struct {
+	chainID  uint64
+	address  common.Address
+	currency string
```
alaibe, line 660: I think it is a mistake to store the currency, this will be simple multiplication to convert to the right one, storing to a default one seems much simpler

services/wallet/transfer/balance_history.go
```diff
@@ -0,0 +1,728 @@
+package transfer
```
alaibe, line 1: This whole balance history deserves its own package to me, it is quite complex feature. Then we could split into multiple files for readbility: e.g: db.go

alaibe (commented):
Maybe I am totally missing something but the logic to ensure caching seems very complex. Wouldn't it be simpler to:
1) Assume block length is always 12s.
2) Store in cache the block number and the time
3) if we need to go back in time 2 days, because last cache was 2 days ago, we calculate 2 point in time for the 2 days. and we fetch the data by adding constant block number

to continue on the example:
last record in cache is:
block number is 100
date is 6th jan at 8:00 UTC

block interval  is 86400 / 15 => so fetch 5760 + 100 and 5760 *2+100

stefandunca:
> Maybe I am totally missing something but the logic to ensure caching seems very complex. Wouldn't it be simpler to:

I agree with you and also thought of why something I estimated to be so simple ended up so complicated. It is a combination of the following factors:
- We don't have a centralized environment to optimize filtering and to cache of balance change events as they happen.
- The requirements with which I started were assuming that I could rely on the transaction history, then evolved that the user can fetch at any time with varying time intervals for any account that can have internal transactions done outside our system.
- My shallow experience with blockchain development and taking the requirements (like 12s fix block interval target literally)

With the newly learned lessons and new long-time caching and flexible sampling requirements, I can see how to simplify the logic, but that will mean a complete rewrite of the fetch and caching, which already works. I would like to see how it goes in the field and prioritize the simplified logic rewrite if we have issues with the current logic. Let's discuss this in the next meeting.

> 1. Assume block length is always 12s.

That is how I started based on the information that the target block time is 12 seconds. However, I had substantial errors that accumulated to generate offsets of days and weeks of missing data for some intervals.
The errors seemed too much to me. After more investigations, I discovered that the block time ranges from 12s to 15s and varies significantly from block to block depending on demand and other factors.

Therefore, I added the calibration part that fetches a few more block timestamps for each blockchain to compensate for errors that significantly improved the results. This complicated the code, but I added tests to support it.

> 2. Store in cache the block number and the time

I store this info along with the identity part (blockchain, currency and address) and a bitset marker to keep the entry associated with the period for quicker fetching.

> 3. if we need to go back in time 2 days, because last cache was 2 days ago, we calculate 2 point in time for the 2 days. and we fetch the data by adding constant block number
> to continue on the example: last record in cache is: block number is 100 date is 6th jan at 8:00 UTC
> 
> block interval is 86400 / 15 => so fetch 5760 + 100 and 5760 *2+100

You described how the implementation was initially done, but I was still missing cache hits due to errors introduced by variations in the block length combined with going back in time. Hence to stabilize the estimations, I had to start fetching from the first block and go up to have a reproducible pattern.
These changes were still insufficient because the start block was varying; hence the errors in the subsequent blocks were propagating to accumulate and have offsets between the blocks. I improved on this by adding the time snapping.

appdatabase/migrations/sql/1671451982_add_balance_history.up.sql
stefandunca: I use this to identify the currency for which the balance was cached (ISO), which, along with chain id and address, is part of the unique balance identity. This is the way to get the ground truth, and any other value reinterpretation will add errors and complexity that may propagate and generate side effects if we want to do anything other transformations.

Your example will work only if we want to show also the history of the currency converted to USD. That means showing the USD value at the time and not accounting for inflation. We want to display the historical amount with the current conversion (adjusted for inflation).

Thanks for the heads up; I check this point of view with the design.

appdatabase/migrations/sql/1671451982_add_balance_history.up.sql
stefandunca: Good question, thanks for the heads up. My understanding is that the short currency is standardized and enough. The `token_address` for the contract is retrieved from our token store. Now I wonder if this can change. Can it be that the contract address for an ERC20 token can change without changing the short currency? Will ask around and look into.

appdatabase/migrations/sql/1671451982_add_balance_history.up.sql
stefandunca: This is more efficient and simple, and I have seen it done in other parts of the code for `big.Int`s. I can't see how this will improve other than human readability, which I didn't consider.

protocol/protobuf/chat_message.pb.go
stefandunca: I'm also surprised to see it; I didn't check what was generated by the `make generate`. Will look closer and clarify this.

services/wallet/transfer/balance_history.go
stefandunca: The task is cancelable and is done by the `PriorityBasedAsyncQueue` when a request for an intermediate balance report comes in, not to run the short update with the complete update simultaneously. However, I don't stop it if the wallet service stops.

I decided this after inquiring if we ever stop the wallet after the service is started and learning that we don't.

Now I'm wondering what happens when we try to close the app tough. I will look into it to be sure we gracefully stop the current update process.

services/wallet/transfer/balance_history.go
stefandunca: I wasn't considering this at that time. However, all the latest improvements are using duration. I will also improve on this with the latest improvements PR (WIP) if the type transition doesn't incur any unnecessary conversions.

services/wallet/transfer/balance_history.go
stefandunca: I will fix this in the current improvements PR to have fewer changes to rebase on.

services/wallet/transfer/balance_history.go
stefandunca: Actually, readability was the reason why I added them. Good to know it didn't work out, and a type alias is a better approach. I will prefer to fix it in the follow-up improvements PR not to rebase all these changes on the ERC20 PR also.

services/wallet/transfer/balance_history.go
stefandunca: This is how it was done initially, but I got warnings from our security check-in CI that using formatting introduces security risks and should be avoided.

stefandunca: Not sure how I can do a secure string formatting and be sure the formated values are correctly escaped.

services/wallet/transfer/balance_history.go
stefandunca: If I do this, I also have to store the conversion rate to have the value in another currency or shown at the current conversion rate, which looks unnecessary. But I will have a better view of this matter after discussing with the design if we need to show the balance history with the current conversion rate or keep the historical one.

services/wallet/transfer/balance_history.go
stefandunca: I agre but will like to have this done in a separate follow up PR.

appdatabase/migrations/sql/1671451982_add_balance_history.up.sql
alaibe: Good point, regarding the difference between various currency. The trade off is that it will cost more request, I would imagine that showing the value according to current currency values is what makes more sense.

But yes better check with John

appdatabase/migrations/sql/1671451982_add_balance_history.up.sql
alaibe: The only cons of this is that it is not possible to query against this value correct?

services/wallet/transfer/balance_history.go
alaibe: Yes this is better to ensure that the wallet can be stopped, this is how we it in other part, also we need to consider mobile which I believe stop the wallet

services/wallet/transfer/balance_history.go
alaibe: we use `fmt.Sprintf ` in many other places, i think we should not duplicate that string

services/wallet/transfer/balance_history.go
alaibe: agree it all depends of: `we need to show the balance history with the current conversion rate or keep the historical one`

services/wallet/transfer/balance_history.go
alaibe: this makes it super hard to review, please consider this

alaibe:
image

FYI this is why I am saying we can consider block are produced every 12s, this is very stable now.

I still don't understand how you will miss cache hit if you start from latest cache?

appdatabase/migrations/sql/1671451982_add_balance_history.up.sql
stefandunca: The query should work with byte comparison (I can't see any case where it is required for balance, and we don't index it, though)

stefandunca:
> ...
> FYI this is why I am saying we can consider block are produced every 12s, this is very stable now.

From which chain is the graph? It is the stable part after transitioning to PoS? Ca we rely on the fact that all blockchains we support will have this stability? In my tests, I haven't seen a stable one.

> I still don't understand how you will miss cache hit if you start from latest cache?

When computing the next block with a fixed block length, we deviate (miss the cache) with one block as soon as the error sum of all the blocks you step over accumulates close to one block length (e.g. 12s and depends on the rounding factor)

For example, in the 1-year time interval, with 12s block length at 100 samples, there will be 26298 blocks in one stride. That means the error of each block should be under ~0.00045 seconds or to vary up and down in a stable manner to compensate for this, which is improbable.

stefandunca:
I will simplify the algorithm as we discussed. Given the recent change in user requirements, the precision and flexibility offered by the current implementation might not be worth the maintenance burden.

stefandunca:
Closing in favour of the simplification refactoring: https://github.com/status-im/status-go/pull/3071

PR #2963, opened 2022-11-16
chore: fix vscode test run

osmaczko:
To run tests locally `-tags "gowaku_skip_migrations"` is needed, otherwise, there is a circular dependency resulting in multiple definitions error.


alaibe (approved):

stefandunca (approved):
Thanks for the finding and fix

PR #2964, opened 2022-11-17
fix: typo in message handler which I found quite annoying

0x-r4bbit:


osmaczko (approved):
My brain struggled to see the difference :D

PR #2967, opened 2022-11-17
Added device name set on Messenger.Start()

Samyoul:
## What's Changed?

Added functionality to automatically get the device name using `os.Hostname()`, this functionality populates the installation table row if there is no name or only a default name.

Example : **Samuels MacBook Pro f4b2e**

## Why Change?

We need this functionality so that we can automatically show a default name of the user's device during pairing.

### DIFF
--- a/protocol/messenger_saved_address.go
+++ b/protocol/messenger_saved_address.go
@@ -2,13 +2,16 @@ package protocol
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/golang/protobuf/proto"
 
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/status-im/status-go/protocol/common"
+	"github.com/status-im/status-go/protocol/encryption/multidevice"
 	"github.com/status-im/status-go/protocol/protobuf"
+	"github.com/status-im/status-go/server"
 	"github.com/status-im/status-go/services/wallet"
 )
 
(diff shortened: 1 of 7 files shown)


protocol/messenger_saved_address.go
```diff
@@ -32,6 +35,45 @@ func (m *Messenger) garbageCollectRemovedSavedAddresses() error {
...
+	} else {
+		imd = ourInstallation.InstallationMetadata
+	}
+
+	// If the name is already set, don't do anything
+	// TODO check the full working mechanics of this
+	if len(imd.Name) > randomDeviceIDLen {
+		return nil
+	}
+
```
Samyoul: @cammellos Do you know if a default value is set? In the current UI we show a 5 character random string, but I don't see where that comes from.

Is this the last/first 5 chars of the installation ID?

If so I can remove all the random string appending in the `InstallationMetadata.Name`

protocol/messenger_saved_address.go
```diff
@@ -32,6 +35,45 @@ func (m *Messenger) garbageCollectRemovedSavedAddresses() error {
 	return m.savedAddressesManager.DeleteSoftRemovedSavedAddresses(uint64(time.Now().AddDate(0, 0, -30).Unix()))
 }
 
+func (m *Messenger) setInstallationHostname() error {
```
stefandunca: @Samyoul is this file a good place for installation hostname? Won't `messenger_identity.go` or `messenger_config.go` be more suitable?

P.S. I now realize this file's name should have been `messenger_saved_addresses.go`

stefandunca (approved):

protocol/messenger_saved_address.go
cammellos: > Is this the last/first 5 chars of the installation ID?

yes, you can remove the random thing from here, it's take from installation id

protocol/messenger_saved_address.go
Samyoul: Thank you, that is a good point, I will move this function `messenger_identity.go`

PR #2968, opened 2022-11-18
Disable abridged memebership update

cammellos:
Ports one commit that was only applied to release, where we addressed the changes to explicit join request, and after disables the abridged behavior since the optimization it's not applicable after the changes in rules of group chat.
This will make sending message slower in a group chat for now, but we need to re-think the way we want to optimize this, but that would be a separate PR.


protocol/messenger.go
```diff
@@ -5799,7 +5799,10 @@ func (m *Messenger) encodeChatEntity(chat *Chat, message common.ChatEntity) ([]b
 				return nil, err
 			}
 
-			encodedMessage, err = m.sender.EncodeAbridgedMembershipUpdate(group, message)
+			// NOTE(cammellos): Disabling for now since the optimiziation is not
+			// applicable anymore after we changed group rules to allow
+			// anyone to change group details
+			encodedMessage, err = m.sender.EncodeMembershipUpdate(group, message)
```
osmaczko, line 5805: :+1:

osmaczko (approved):

PR #2970, opened 2022-11-21
fix: disable gossipsub peer exchange

richard-ramos:


alaibe (approved):

jm-clius (approved):
LGTM! Out of interest's sake, wouldn't it be possible to just disable in the `cfg` (although I'm speaking as a complete status-go noob)?

richard-ramos:
> LGTM! Out of interest's sake, wouldn't it be possible to just disable in the `cfg` (although I'm speaking as a complete status-go noob)?

It is possible. Would you prefer that I keep the option available, but by default keep it disabled?

jm-clius:
> It is possible. Would you prefer that I keep the option available, but by default keep it disabled?

I don't see us enabling this feature in Status-environments in future, so I don't think it's necessary. Was just curious.

PR #2973, opened 2022-11-23
fix(communities): properly delete remove chats when handling community

0x-r4bbit:
updates

When channels where removed from communities by the community owner, community members would not remove the chats in their own databases.

See https://github.com/status-im/status-desktop/issues/8000 for more info.

This commit ensures that nodes delete removed chats from their local chats table and also deregister the corresponding transport filters.


jrainville (approved):
LGTM

PR #2976, opened 2022-11-23
backup profile data to and fetch them from waku

saledjenic:
Changes applied here introduce backing up profile data (display name and identity
images) to and fetch them from waku. Information about those data is sent
as a separate signal to a client via `sync.from.waku.profile` signal.

New signal `sync.from.waku.progress` is introduced which will be used to notify a client
about the progress of fetching data from waku.


api/geth_backend.go
```diff
@@ -704,6 +704,16 @@ func (b *GethStatusBackend) saveAccountsAndSettings(settings settings.Settings,
 	if err != nil {
 		return err
 	}
+
+	// In case of setting up new account either way (creating new, importing seed phrase, keycard account...) we should not
+	// back up any data after login, as it was the case before, that's the reason why we're setting last backup time to the time
+	// when an account was created.
```
0x-r4bbit, line 710: Might be worth adding **why** we should not perform a backup after login

protocol/messenger_handler.go
```diff
@@ -406,20 +412,114 @@ func (m *Messenger) handleCommandMessage(state *ReceivedMessageState, message *c
...
+	} else {
+		// since both images (large and thumbnail) are always stored in the same time, we're free to use either of those two clocks for comparison
+		lastImageStoredClock := dbImages[0].Clock
+		syncWithBackedUpImages = lastImageStoredClock < backupTime
+	}
+
+	if syncWithBackedUpImages {
+		if len(message.Pictures) == 0 {
+			err = m.multiAccounts.DeleteIdentityImage(message.KeyUid)
+			response.AddImages(nil, err == nil)
```
cammellos: This looks a bit sketchy, what is it trying to do exactly?

protocol/messenger_backup.go
```diff
@@ -78,6 +78,118 @@ func (m *Messenger) startBackupLoop() {
...
+	profileToBackup, err := m.backupProfile(ctx, clock)
+	if err != nil {
+		return 0, err
+	}
+
+	// Update contacts messages encode and dispatch
+	for i, d := range contactsToBackup {
+		d.ContactsDetails = &protobuf.BackupDataDetails{
+			DataNumber:  uint32(i + 1),
+			TotalNumber: uint32(len(contactsToBackup)),
```
0x-r4bbit: So this looks a bit unintuitive. `contactsToBackup` is a single message that can actually contain multiple contacts.
In other words: `TotalNumber` might be `1` and `DateNumber` might be `1` as well, but the amount of contacts to backup is actually #n 

Not sure if it causes a problem here, just want to raise this to be sure..

protocol/messenger_handler.go
```diff
@@ -406,20 +412,114 @@ func (m *Messenger) handleCommandMessage(state *ReceivedMessageState, message *c
...
+					Width:        int(pic.Width),
+					Height:       int(pic.Height),
+					FileSize:     int(pic.FileSize),
+					ResizeTarget: int(pic.ResizeTarget),
+					Clock:        pic.Clock,
+				}
+				idImages[i] = img
+			}
+			err = m.multiAccounts.StoreIdentityImages(message.KeyUid, idImages, false)
+			response.AddImages(idImages, err == nil)
```
cammellos: Similarly here, if there's any error but we want to continue, probably best to leave comments explaining why and what we are trying to achieve

protocol/messenger_handler.go
```diff
@@ -406,20 +412,114 @@ func (m *Messenger) handleCommandMessage(state *ReceivedMessageState, message *c
...
 	}
+	err = m.handleSyncSettings(state, message.Settings)
+	if err != nil {
+		errors = append(errors, err)
+	}
+
+	if m.config.messengerSignalsHandler != nil {
+		response := wakusync.SyncFromWakuResponse{}
+
+		response.AddWakuSyncDetails(SyncWakuSectionKeyProfile, message.ProfileDetails)
```
cammellos: This does not take into account different backups, you might receive multiple and from different devices

protocol/protobuf/pairing.proto
```diff
@@ -255,3 +267,16 @@ message SyncContactRequestDecision {
...
+message SyncProfile {
+  string key_uid = 1;
+  string display_name = 2;
+  repeated SyncProfilePicture pictures = 3;
+}
+
+/* `SyncProfileSettings` is used to sync user preferences which are stored locally in the `settings` table 
+ * but for example we do not need to sync `display_name` which is one of the fields of `settings` table cause 
+ * we will already have it via `SyncProfile` */
+message SyncProfileSettings {
```
cammellos: Do we need an empty struct, we have already SyncSettings?

protocol/messenger_handler.go
```diff
@@ -33,6 +34,11 @@ const (
 	transactionRequestDeclinedMessage           = "Transaction request declined"
 	requestAddressForTransactionAcceptedMessage = "Request address for transaction accepted"
 	requestAddressForTransactionDeclinedMessage = "Request address for transaction declined"
+
+	SyncWakuSectionKeyProfile     = "profile"
+	SyncWakuSectionKeyContacts    = "contacts"
+	SyncWakuSectionKeyCommunities = "communities"
+	SyncWakuSectionKeySettings    = "settings"
```
0x-r4bbit: Instead of defining the section keys here and later use them to hydrate `WakuSyncDetails`, how would you feel about encoding the possible sections in the struct itself?

So instead of having:

```golang
type SyncFromWakuResponse struct {
	SyncingProgressDetails map[string]protobuf.BackupDataDetails // key represents the data/section backup details refer to
	Profile                *SyncProfile
}
```

We'd have something like

```golang
type SyncingProgressDetails struct {
  Profile *protobuf.BackupDataDetails `json:"profile,omitempty"`
  Contacts *protobuf.BackupDataDetails  `json:"contacts,omitempty"`
  Communities *protobuf.BackupDataDetails  `json:"communities,omitempty"`
  Settings *protobuf.BackupDataDetails  `json:"settings,omitempty"`
}

type SyncFromWakuResponse struct {
	SyncingProgressDetails * SyncingProgressDetails
	Profile                *SyncProfile
}
```

That way it's explicit what data can be expected in a progress signal.
This would also allow to add a `TotalProgress` value in case this is interesting for clients.

I've done a similar thing for the discord import (look for `ImportProgress` if you need inspiration)

The downside is obviously that one can't add more backup details without changing the type as well.

protocol/protobuf/pairing.proto
```diff
@@ -255,3 +267,16 @@ message SyncContactRequestDecision {
     DECLINED = 1;
   }
 }
+
+/* `SyncProfile` is used to describe profile of logged in user */
+message SyncProfile {
+  string key_uid = 1;
+  string display_name = 2;
```
cammellos, line 278: Display name I believe could be added to `SyncSetting`, for consistency, instead of creating a new protobuf.
Images are a bit trickier, but I defer to @Samyoul

protocol/messenger_handler.go
```diff
@@ -406,20 +412,114 @@ func (m *Messenger) handleCommandMessage(state *ReceivedMessageState, message *c
...
+	contact, ok := state.AllContacts.Load(pubKey)
+	if !ok {
+		return errors.New("cannot find logged in profile among all known contacts")
+	}
+
+	response := wakusync.SyncFromWakuResponse{
+		Profile: &wakusync.SyncProfile{},
+	}
+
+	if contact.LastUpdated < backupTime {
```
cammellos: BackupTime is not the clock, since that's only the clock of the time you sent a backup, but it might not reflect the latest user's update. You need to have a clock for each piece that syncs independently.

cammellos (changes requested):
Left some comments, thanks for the PR.

Could you please add some tests for the new functionalities?

Another thing is that it does not seem to take into account the fact that multiple backups can be fetched (from the same device, or from different devices), I think that should be taken care of/figured out.

There's also an argument to have more logic in status-go keeping track of the backup that is being currently fetched/processed, rather than just pushing everything to the client (i.e keep the count in status-go, and publish updates, rather than publishing only counts).

protocol/protobuf/pairing.proto
```diff
@@ -255,3 +267,16 @@ message SyncContactRequestDecision {
...
+  repeated SyncProfilePicture pictures = 3;
+}
+
+/* `SyncProfileSettings` is used to sync user preferences which are stored locally in the `settings` table 
+ * but for example we do not need to sync `display_name` which is one of the fields of `settings` table cause 
+ * we will already have it via `SyncProfile` */
+message SyncProfileSettings {
+}
```
0x-r4bbit: Any reason this is empty?

protocol/messenger_handler.go
saledjenic: There are 2 properties describing each change (`AddImages`, `AddDisplayName` they all expect result of real action did the real go well or not)... cause client needs to know about the action is it completed or not, in some cases we may want to notify user about what's really happened.

For example in case of setting new identity images, we're sending a list of images as `images` property and boolean flag as `imagesStored` property, then on the client side we always check `if imagesStored` and then do the things, if not we may notify user what went wrong and offer him to store images again.

So a short answer is `err == nil` notifies client if the action went well or not.

protocol/messenger_handler.go
saledjenic: explained above

protocol/messenger_handler.go
saledjenic: Your question makes me think that it's quite impossible then to have progress update in place as it's planned by UI design, cause we cannot be sure 100% ever that we're receiving the latest backed up messages. That further implies that in the middle of receiving the last known backed up messages we receive a newer one which will reset the progress as their progress details may be completely different.

So what could be a solution for that if we want to keep progress for fetching messages from waku in the UI?

protocol/protobuf/pairing.proto
saledjenic: That empty struct is more like predisposition for further changes, that's why it's empty. I wanted to do the all parts which need to be sync in the same way.

protocol/protobuf/pairing.proto
saledjenic: Idea was to create a new buffer which refers to profile, to combine display name and identity images.

protocol/messenger_handler.go
saledjenic: `Clock` value of `protobuf.Backup` message is actually `clock` value got from this call `clock, chat := m.getLastClockWithRelatedChat()` is that wrong?

saledjenic:
> Could you please add some tests for the new functionalities?

@cammellos yes, will add them, had that on my mind and forgot at the end.

> Another thing is that it does not seem to take into account the fact that multiple backups can be fetched (from the same device, or from different devices), I think that should be taken care of/figured out.

Yes, we need to think about it, commented [here](https://github.com/status-im/status-go/pull/2976#discussion_r1036276576)

> There's also an argument to have more logic in status-go keeping track of the backup that is being currently fetched/processed, rather than just pushing everything to the client (i.e keep the count in status-go, and publish updates, rather than publishing only counts).

I agree that logic should be kept in status go, but because of the UI design requirements information about the progress is sent via `sync.from.waku.progress` signal to the client. At this moment I don't see the need to keep those details in statusgo, if we need them ever we can easily add them.

protocol/protobuf/pairing.proto
saledjenic: answered [here](https://github.com/status-im/status-go/pull/2976#discussion_r1036277728)

protocol/messenger_backup.go
saledjenic: This mechanism packs messages pretty well and accurate.

Please check `backupContacts` in context of `BackupData` function, so we're preparing backup messages in a way that each of them contains up to `BackupContactsPerBatch` contacts. Now for example if there are 25 contacts and we're packing let's say 10 contacts per message, we will have 3 backup messages, 2 of them with 10 contacts and one with 5 contacts and backup messages will look in that case like this:
```
// Message 1
protobuf.Backup {
   Contacts: []*SyncInstallationContactV2 // <-----contains 10 contacts
  BackupDataDetails {
    DataNumber:  1,
    TotalNumber:  3
}

// Message 2
protobuf.Backup {
   Contacts: []*SyncInstallationContactV2 // <-----contains 10 contacts
  BackupDataDetails {
    DataNumber:  2,
    TotalNumber:  3
}

// Message 3
protobuf.Backup {
   Contacts: []*SyncInstallationContactV2 // <-----contains 5 contacts
  BackupDataDetails {
    DataNumber:  3,
    TotalNumber:  3
}
```

protocol/messenger_handler.go
saledjenic: As you've said, it can be that way as well... what I had on my mind is that it will be a lot of sections/details for syncing, like more than 30 for sure, maybe more than 40 and having a map `map[string]protobuf.BackupDataDetails` looked more attractive to me, where key represents the section/detail we need to send a progress description for and value describes how many messages for that section we need to receive in total and which one current message refers to.

To summarize... this map `map[string]protobuf.BackupDataDetails` is a part of each backup message and it describes how many messages of which type we have backed up and which one is the received backup message in order.

0x-r4bbit: That's fair.

Can take this one offline but: how do you get to 30 or 40 things to be synced? I'd assume most of the stuff we're interested in lives already in profile, contacts, communities and settings

saledjenic: > I'd assume most of the stuff we're interested in lives already in profile, contacts, communities and settings

Most likely you're referring to syncing among devices, that's true a lot of things are already in place for that syncing.

But in terms of backing up data to waku and fetching/handling them we had only contacts and communities, nothing else. I didn't updated that functionality just reorganize a code a bit so it fits new needs we have and add profile details (display name and identity images) to backup.

Here in this PR https://github.com/status-im/status-desktop/pull/8433 you can see how the changes applied here fit our need. 

When I said we will need to sync 30 - 40 things (will see when the list becomes final), yes we will have that much maybe even more but we will group them and display progress per group, for example as a part of profile syncing group will be some keycard related details, wallet details and so on.

But the point is that each backed up message carries complete information or part of information (depends if the thing we backed up is backed up as a single message or through more messages) and also the details of all backed up messages (how many of them we should receive for each thing and which one is the one being received in order).

cammellos: I think I am with @PascalPrecht on this one, basically we are talking about essentially the same data-struct, as a matter of fact it yields the same JSON, but one is using dynamic keys, while the other is using the struct fields.

A dynamic map use case is when you have dynamic keys, but here there's a set of well-defined keys.

The cost of change is pretty much identical, since you either add a new `const` or add a new field. Setting the map is probably easier, since you could have a generic function.

Though using fields seems more idiomatic and explicit to just use a struct, so I don't have to lookup somewhere else to understand what's in this struct, @Samyoul is the one that has likely the best eye for what's most idiomatic, but to me doesn't feel there's much advantage by using a dynamic map that's worth straying from what's more idiomatic

protocol/messenger_handler.go
cammellos: Yes, what you are interested here is the time the picture/display name (or any piece of data we sync) has changed, not the time a backup is published.
So we need to keep track of that and pass it along to make a comparison.

The example where this would break is:

1) User on device A1 changes picture at T1, goes offline
2) User on device A2 changes picture at T2, publishes a backup with backup time B1
3) User on device A1 goes online, backup triggers with backup time B2, before the message from A2 with T2 is received (therefore including picture from T1 but backup time B2)
4) User fetches a backup on device A3 (or event A2), it should have eventually picture at T2, but it will have picture from T1 since we are using backup time.

protocol/protobuf/pairing.proto
cammellos: Though I don't get this, why not use `SyncSetting` which we have already instead of adding a new struct? 
I get that you'd like to have SyncProfile to collate similar infos (though I think we should re-use `SyncSetting` since it's there if we can, I believe display name is fine, but not sure about images @Samyoul ?)

If we want to do things the same way, we should reuse what we have for everything or extend it, instead of creating a new one, but maybe I misunderstand the purpose

protocol/messenger_handler.go
saledjenic: @cammellos great example, thanks, explains everything, will update the code accordingly.

protocol/wakusync/response.go
```diff
@@ -0,0 +1,25 @@
...
+
+	"github.com/status-im/status-go/protocol/protobuf"
+)
+
+type SyncFromWakuResponse struct {
+	SyncingProgressDetails map[string]protobuf.BackupDataDetails // key represents the data/section backup details refer to
+	Profile                *SyncProfile
+}
+
+func (sfwr *SyncFromWakuResponse) MarshalJSON() ([]byte, error) {
```
cammellos: So,
I think the reason we are doing this marshaling etc, is because we want to convert the protobuf to our own struct:

```
type SyncSectionDetails struct {
	DataNumber  uint32 `json:"dataNumber,omitempty"`
	TotalNumber uint32 `json:"totalNumber,omitempty"`
}
```
```
message BackupDataDetails {
  uint32 data_number = 1;
  uint32 total_number = 2;
}
```

But otherwise the two are equivalent.
I think that's fine, we should not have protobuf moving around, although we do that sometimes but likely we should not.

If we are doing so though, the best place to do the conversion is not at marshal time, can we not just store it as `SyncSectionDetails` when we add it? That way we don't have to override the MarshalJSON function.
Also why two separate files? it seems like this could be a single one

protocol/messenger_handler.go
cammellos: This is a very uncustomary  way to handle errors, I think it's best to understand why something would fail, if it's expected, we should handle it, if it's not expected, we should fail fast, this seems like it's really swallowing the error and continuing, leaving it up to the client to handle something that's not able to handle.
Are there any expected errors happening here? or you don't know how it could fail?

protocol/messenger_handler.go
cammellos: > Your question makes me think that it's quite impossible then to have progress update in place as it's planned by UI design, cause we cannot be sure 100% ever that we're receiving the latest backed up messages. 

Yes, that's correct, we can't be 100% sure, and as you pointed out, we might have to restart the process, or you might have multiple backups from multiple devices, both of them relevant, coming at different times etc.

Now, what we do about it, it's a discussion to be had, we either don't show a detailed progress (i.e only a generic spinner, which suits better our technology), or if we want to show a detailed progress, we accept the fact it will not be reliable in some cases, it will restart in others etc. Of course there might be other technical solutions, but the fact that we have permissionless non-consistent decentralized storage makes these things a nightmare to work with.

This would have been caught early if we had a tech feasibility analysis, but what we do about it at this stage is up to us.

cc @John-44 @xAlisher

protocol/messenger_handler.go
saledjenic: > Are there any expected errors happening here? or you don't know how it could fail?

At this point errors which may happen are only those returned from `db.Exec` call. 

Now for example if error happened for setting display name action we will send a signal containing `{"CustomName", false}` if there is no errors we send `{"CustomName", true}`, when client receives such signal, it will firstly check the value and if action was performed with success or not and do the things accordingly. For example if all went well it will update display where is needed, if didn't go well it will display an error to a user or offer him to try to store `"CustomName"` again. That was my idea when I was working on this, always to have an option.

cammellos: I think in this case there's nothing we can really do, if the database is not responsive, we might want to bubble up an error to the user, but we should stop immediately any operation, there's no easy way to gracefully handle this

protocol/wakusync/response.go
saledjenic: @cammellos I wanted not to keep everything in a single file, but have more shorter files, that way I find it easier for maintaining later. What was my idea, so the structure is like this:
```
protocol/wakusync/
    - profile_response.go
    - progress_response.go
    - response.go
    - settings_response.go // will be added later
    - communities_response.go // will be added later
    - ...
```
latter we will add here `settings_response.go`, `communities_response.go` and so on when we have the full list what else need to be backed up.

Now `response.go` is the main file there, and describes the response which will be sent via signal and it looks like this:
```golang
type SyncFromWakuResponse struct {
	SyncingProgressDetails map[string]protobuf.BackupDataDetails
	Profile                *SyncProfile
	Settings               *SyncSettings // will be added later
	Communities            *SyncCommunities // will be added later
}
```
That way I wanted to introduce a bit of consistency in responses, that all signals which are sending data being fetched from waku will have the same form, for example if we're sending data that profile info is fetched that struct will have only `Profile` which carrying info what is changed and `SyncingProgressDetails` describing the current progress of fetching from waku.

Responses would look like:
- when we receive profile 
```
SyncFromWakuResponse { 
    Profile: SyncProfile, 
    SyncingProgressDetails:  map[string]BackupDataDetails 
}
```
- when we receive settings 
```
SyncFromWakuResponse { 
    Settings: SyncSettings, 
    SyncingProgressDetails:  map[string]BackupDataDetails 
}
```
and so on, but that was my general idea, to have clean state.

saledjenic:
@all please review this once again, I applied changes following your comments above.
What's done:
- new 3 signals introduced: `waku.fetching.backup.progress`, `waku.backedup.profile`, `waku.backedup.settings` and will be received on the client side carrying only necessary information
- backing up/fetching contacts and communities were already there, so I just reorganized them a bit to fit the new backing up mechanizam and wrapped them into `protobuf.Backup`, since everything we want to backup will go through `BackupData` function
- `protobuf.SyncSetting` is used for backing up settings, this is the same struct, reused, which we're using for device syncing (I removed previously introduced `SyncProfileSettings`)
- once messenger is started, we're waiting for 30 seconds hoping that at least one of the last backed up messages will be received that we could possibly know which messages we want to work with and discard older backed up messages and after 30 seconds we start processing them and notifying client about that
- test added

Here is a PR if somebody wants to check how is the corresponding part done in the desktop app (though the UI there is not the final, but functionality will be more or less the same) https://github.com/status-im/status-desktop/pull/8433

appdatabase/migrations/sql/1670249678_display_name_to_settings_sync_clock_table.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE settings_sync_clock ADD COLUMN display_name INTEGER NOT NULL DEFAULT 0;
```
richard-ramos, line 1: I'm not sure if this is necessary, but i have seen it done in other migrations
```suggestion
ALTER TABLE settings_sync_clock ADD COLUMN display_name INTEGER NOT NULL DEFAULT 0;
UPDATE settings_sync_clock SET display_name = 0;
```

api/geth_backend.go
```diff
@@ -704,6 +704,16 @@ func (b *GethStatusBackend) saveAccountsAndSettings(settings settings.Settings,
...
+	now := time.Now().Unix()
```
richard-ramos, line 711: Should we use ntp time here instead of time.Now()? if so, do check https://github.com/status-im/status-go/pull/3005/ on how to use it

protocol/messenger_backup_handler.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	bh.messagesToProceed = append(bh.messagesToProceed, message)
+}
+
+func (m *Messenger) startWaitingForTheLatestBackedupMessageLoop() {
+	ticker := time.NewTicker(timeToPostponeBackedUpMessagesHandling)
+	m.backedupMessagesHandler.postponeTasksWaitGroup.Add(1)
+	go func() {
+		defer m.backedupMessagesHandler.postponeTasksWaitGroup.Done()
```
richard-ramos: `ticker` should probably be part of the `go func()`

richard-ramos: Also, probably makes sense to do something like:
```go
ticker := time.NewTicker(timeToPostponeBackedUpMessagesHandling)
defer ticker.Stop()
```

because inside the case of <-ticker.C there is a `return`, meaning that the ticker would be left dangling since it's only stopped on `<-m.quit`

protocol/messenger_backup_handler.go
```diff
@@ -0,0 +1,205 @@
...
+		for {
+			select {
+			case <-ticker.C:
+				if !m.backedupMessagesHandler.postponeHandling {
+					return
```
richard-ramos: Should it be `return` or `continue`?  the `return` will kill the goroutine.

richard-ramos (approved):
Looks good, just left some comments

appdatabase/migrations/sql/1670249678_display_name_to_settings_sync_clock_table.up.sql
saledjenic: According to documentation it's not needed, you can check here: https://www.sqlite.org/lang_altertable.html
What should happen is:
```
sqlite> CREATE TABLE test (col1 INTEGER NOT NULL);
sqlite> INSERT INTO test VALUES(1);
sqlite> SELECT * FROM test;
1
sqlite> ALTER TABLE test ADD COLUMN new-col INTEGER DEFAULT 100 NOT NULL;
sqlite> SELECT * FROM test;
1|100
```

api/geth_backend.go
saledjenic: What we're doing here is, just setting the time when a new user is created, that way we're avoiding backing up data for a fresh newly created user. Why I am storing this current time here, cause [this (check it)](https://github.com/status-im/status-go/blob/49b3c573f6683eb708632bb4b9f597121e6c2a7d/protocol/messenger_backup.go#L59-L60) is the place where we're using it and  comparing.

protocol/messenger_backup_handler.go
saledjenic: Good point, I have to call `ticker.Stop()` before return definitely, cause according to documentation `Stop the ticker to release associated resources`, so need to update it like this:
```
if !m.backedupMessagesHandler.postponeHandling {
        ticker.Stop() // <--need to add this
	return
}
```

protocol/messenger_backup_handler.go
saledjenic: Yes, it should be return, cause that go routine is run only once when user logs in the app following one of two onboarding flows for old Status' user.

saledjenic:
@richard-ramos thanks a lot for review.

protocol/messenger_backup_handler.go
cammellos: `defer ticker.Stop()` is a bit more idiomatic, and a bit safer, since you don't have to explicitly stop it before each return

protocol/messenger_backup.go
```diff
@@ -220,3 +299,54 @@ func (m *Messenger) syncBackupContact(ctx context.Context, contact *Contact) *pr
...
+	if err != nil {
+		return nil, err
+	}
+
+	displayNameClock, err := m.settings.GetSettingLastSynced(settings.DisplayName)
+	if err != nil {
+		return nil, err
+	}
+	if displayNameClock == 0 {
+		displayNameClock = clock
```
cammellos: This I am not sure, why are we setting the displayNameClock to a non-zero value?

protocol/messenger_backup_handler.go
```diff
@@ -0,0 +1,205 @@
...
+	"github.com/status-im/status-go/images"
+	"github.com/status-im/status-go/multiaccounts/settings"
+	"github.com/status-im/status-go/protocol/protobuf"
+	"github.com/status-im/status-go/protocol/wakusync"
+)
+
+const (
+	// timeToPostponeBackedUpMessagesHandling - the idea is to wait for 30 secs in loading state with the hope that at least
+	// one message from the set of last backed up messages will arrive, that we can know which set of messages to work with
+	timeToPostponeBackedUpMessagesHandling = 30 * time.Second
```
cammellos: How do you know which one is the right one? Does this run everytime the user login? i.e we always delay processing of back ups even if it's not the first login?

I don't quite understand how we handle multiple backups (from different devices for example) or older backups from the same device, is that something that the code takes into consideration?

protocol/messenger_backup_handler.go
```diff
@@ -0,0 +1,205 @@
...
+		return err
+	}
+
+	response := wakusync.WakuBackedUpDataResponse{
+		Profile: &wakusync.BackedUpProfile{},
+	}
+
+	if dbDisplayNameClock < message.DisplayNameClock {
+		err := m.SetDisplayName(message.DisplayName, false)
+		response.AddDisplayName(message.DisplayName, err == nil)
```
cammellos: we want to stop immediately at errors, better to fail fast, unless you expect an error, or you don't care about it, this does not seem to be the case

protocol/messenger_backup_handler.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	syncWithBackedUpImages := false
+	dbImages, err := m.multiAccounts.GetIdentityImages(message.KeyUid)
+	if err != nil {
```
cammellos, line 87: do you know why it would fail here and why we would continue if an error is returned?

protocol/messenger_backup_handler.go
```diff
@@ -0,0 +1,205 @@
...
+	} else {
+		// since both images (large and thumbnail) are always stored in the same time, we're free to use either of those two clocks for comparison
+		lastImageStoredClock := dbImages[0].Clock
+		syncWithBackedUpImages = lastImageStoredClock < backupTime
+	}
+
+	if syncWithBackedUpImages {
+		if len(message.Pictures) == 0 {
+			err = m.multiAccounts.DeleteIdentityImage(message.KeyUid)
+			response.AddImages(nil, err == nil)
```
cammellos: Similar as above, I think we should stop executing if an unexpected error is returned

cammellos (commented):
Left some comments, instinctively, delaying the processing for 30 seconds does not seem to be a robust solution, as well because it's mostly a client/UI concern that we are carrying over to the backend (cache and release the updates in the client for example).

Generally, I still believe that we are trying to do something that it's fundamentally not inline with the underlying technology, so it's going to be always an uphill battle

protocol/messenger_backup.go
saledjenic: When user is recovering an account the `CreateSettings` function is called where we're initializing `settings` table only with the init values for that account, but `settings_sync_clock` is not set, so all fields there default to 0. I didn't want to initialize `settings_sync_clock` in that case (even it should be by my opinion) cause I was not sure about all use cases for syncing among devices, and that was the reason why I set it here. The display name should have other than 0 value in the moment we're backing it up if we want to cove the following scenario:
- user creates an account and display name which was entered during onboarding flow is stored as it is to `settings` table, at this moment display name time in `settings_sync_clock` is set to 0
- user don't change his display name after that
- after 8 hrs backup is triggered, without this check backed up display name time would be 0
- now if user tries to recover an account using the same seed phrase display name won't be set cause we're setting it only if its time is greater than the last sync time stored in db (which is also 0)

protocol/messenger_backup_handler.go
saledjenic: An idea is to wait for 30 seconds and proceed with applying most recent **known** backup we receive within the timeframe of 30 seconds. That means newer backup messages may arrive later and they will be applied later, but older backup messages than those known as the newest during that time frame will be applied firstly.

Yes, aggregating backup messages is happening for each login for the period of 30 seconds.
We don't need to know which device a backup is coming from if it's the most recent one since all messages backed up  in a single go have the same time.

Example If device 1 backed up its state at time t1, marking that time as D1T1, now following that logic if we have backups done in this order `D1T1` < `D2T1` < `D1T2` < `D3T1` and let's say that for each of these backups 5 messages were backed up.

Now when user is recovering an account at time `T` which is now (means `D3T1` < `T`) some or all of messages backed up from any devices `D1`, `D2`, `D3` may or may not arrive within the time of 30 seconds after app start, but how we handle them is:
- timer of 30 secs started
- first arrives msg `#4` from `D2T1` we add it to queue but don't process
- then arrives msg `#2` from `D2T1` we add it to queue but don't process 
- then arrives msg `#3` from `D1T1` since it has less time that stored messages we just skip it
- then arrives msg `#4` from `D1T1` we skip it
- then arrives msg `#1` from `D1T2` we remove all messages from queue since they have less time than the most recent one and add that one
- then arrives msg `#2` from `D1T1` we skip it
- then arrives msg `#3` from `D1T2` we add it
- and now time's up, 30 seconds passed

what do we do now, we process messages from the queue even 0 messages from `D3T1` arrived, when other messages arrives later, they will be processed immediately and data they are carrying will be applied if the time for that data is greater then the time of last change.

protocol/messenger_backup_handler.go
saledjenic: What was my logic, if backup is made of multiple messages, where any of them may refer to more than a single thing, for example a single message may carry information about up to 20 contacts, profile message carries info about display name + identity images, why would we get back immediately instead of trying to apply as much as we can from the backup?

protocol/messenger_backup_handler.go
saledjenic: At this point errors which may happen are only those returned from `db.Query` or `rows.Scan` call and I agree if that's the case there is nothing we can really do to recover db, but error may happen also if there are no rows in the db, for example if user hasn't set profile image.

cammellos: Ok, then we should still return an error and stop execution, unless is a `sql.ErrNoRows`, in which case we continue (there are some examples of this already in the codebase)

protocol/messenger_backup_handler.go
cammellos: I see why you'd want to do that, but it's better to throw red flags at this stage, we need to know why things don't work at early stages of development, otherwise we risk sweeping errors under the carpet, https://en.wikipedia.org/wiki/Fail-fast at this stage is better (at the message-level, we don't want to completely stop processing messages), since if anything throws an error here, it would be completely unexpected, but maybe there's a good reason

protocol/messenger_backup.go
cammellos: >  I didn't want to initialize settings_sync_clock in that case (even it should be by my opinion) cause I was not sure about all use cases for syncing among devices

I think you are right here, this should be the way, `CreateSettings` should set the sync clock value of any field that's user set, that would be the root cause of the issue, since the above code might override a user display_name, since it might have changed it on a different device before this code is triggered, but before the backup is received.

We can address  it in a separate PR if you'd like, but I think we should fix the root cause, since the code above I believe might introduce a race condition/bug.
If you decide that we should address it in a separate PR, would you mind creating an issue describing the issue so we can track it please?

protocol/messenger_backup_handler.go
cammellos: I see, thanks for the explanation.

One thing to note is that you can't discard backups from different devices, since they might not have converged yet if the user hasn't paired, and if you do that the risk is that they will not. 
You can safely discard though backups from the same device with different timestamps, keeping the one with the oldest timestamp as you are already doing.

Regarding the 30s, I don't think that's acceptable to delay processing, as it would impact every login and that's just not worth it, moreover is a purely client side concern and can be addressed in the client if the client wishes to do so (I still think the feature is just not worth the effort and we should just show a loading screen, this is not a solvable problem).

It's also not really addressing completely the issue since it might be that connecting to a store node is slow for example, the user might be offline, etc. Solutions like this were tried in the past and never really worked, so I don't think it's worth the complexity.

saledjenic: @cammellos I applied all other suggestions you gave me, you can check them in [this commit](https://github.com/status-im/status-go/pull/2976/commits/4ed705ae2d59c51a8a379bd5f6d1c2e9b865a6c3).

> Regarding the 30s, I don't think that's acceptable to delay processing, as it would impact every login and that's just not worth it, moreover is a purely client side concern and can be addressed in the client if the client wishes to do so (I still think the feature is just not worth the effort and we should just show a loading screen, this is not a solvable problem).

It's done in separate thread and it doesn't affect the app since it will only postpone a processing of messages of `protobuf.Backup` type, other messages will be processed as they are coming. After 30 seconds that everything gets destroyed and that's it, back to normal handling. 

It shouldn't be a big impact in terms of performances, but I agree we should not trigger/run anything if that is not applicable in that context. In general, we could do some improvements and actually discard all messages of `protobuf.Backup` type always, unless account was created following restore account via seed phrase/keycard flow. If you agree we can do that in another PR?

Final question... do I need to do something? Or we can leave that for later?
It was John's idea to wait for 30 secs and proceed with the messages which are know at that moment (that also doesn't stop us from applying other messages if they appear later) and to do that on the `stauts-go` side cause the same logic should be for both mobile and desktop apps.

protocol/messenger_backup.go
saledjenic: Handled here in this PR.

protocol/messenger_backup_handler.go
```diff
@@ -172,6 +184,9 @@ func (m *Messenger) handleBackedUpProfile(message *protobuf.BackedUpProfile, bac
 			}
 			err = m.multiAccounts.StoreIdentityImages(message.KeyUid, idImages, false)
 			response.AddImages(idImages, err == nil)
```
cammellos: I would move this line after error checking, no need to add to response if an error is thrown, better to just fail immediately, similar above we should do the same

multiaccounts/settings/database.go
```diff
@@ -144,6 +144,38 @@ INSERT INTO settings (
 		return err
 	}
 
+	// It's enough to set sync clock to any value greater than 0, but not to the current time, cause `CreateSettings` function
+	// is being called for every newly created account (in which case we're good to set current time for the last sync time),
+	// but it's also called for recover account flow, where we need to set data stored on waku (previously backed up data) because
+	// of this we cannot set last sync time to the current time. Also we cannot leave it to 0, cause if account is created, then
```
cammellos: Asked you a question about this, I think I might need more clarification on the `1` case, the first case is clear, the second one maybe we could word it a bit better, but not sure I fully understand

saledjenic:
@cammellos as we discussed in private chat I applied those changes, please have a look and let me know if something else needs to be changed? Aggregating messages is now again a client responsibility.

appdatabase/migrations/sql/1670249678_display_name_to_settings_sync_clock_table.up.sql
qfrank: Oh, i thought we won't use this method to solve the clock issue as discussed before😅 that's why my [pr](https://github.com/status-im/status-go/pull/2989) still in draft state @saledjenic

cammellos (approved):
There's some places where we should use a NTP synced source, I would address that in this PR, or at least add a comment, rest looks good, thank you!

multiaccounts/settings/database.go
```diff
@@ -144,6 +145,14 @@ INSERT INTO settings (
 		return err
 	}
 
+	if s.DisplayName != "" {
+		now := time.Now().Unix()
+		err = db.SetSettingLastSynced(DisplayName, uint64(now))
```
qfrank, line 150: could u explain why we need call `db.SetSettingLastSynced` here, as i saw there'r already some places take care of settings_sync_clock ? @saledjenic thanks

saledjenic: @qfrank yes, sure, will try to explain it now, in general it covers a special case.

Starting from an empty state (empty local data folder, no database) regardless which flow you follow to create an account (is it create a new account, create an account using seed phrase or recover account or whatever) initial settings will be inserted into `settings` table using `CreateSettings` where client client sent them via settings json object. That means any value we want to track a sync time of will be set in the settings table without any proper time set in the `settings_sync_clock`. Since other values have a default value will continue explanation on display name, cause empty string `""` is the default name for it. Display name is set by user during the onboarding flow, and if we leave the code without this check then `0` will be set for display name in `settings_sync_clock` table.

So when we perform a backup, the time of last change of display name will be `0`. Then if user recovers an account using the same seed phrase fetched display name won't be applied cause its time in backed up message will be `0` which is the same time as for display name stored in `settings_sync_clock` for an empty display name (cause during recover account flow there is no option to set display name, but it's expected that it will be fetched form waku) so user gets landed in the app without set display name in that case. To overcome that it's enough to set initial time to anything greater than `0`, initially I set it to `1` and that works, but @cammellos suggested to use current time here.

My opinion is that we need to set last sync time for all fields we want to track a last sync time of, not only for display name. That should be set to current time or last clock in case of all but the recover onboarding flows. But why it's not done that way, cause that firstly need to be discussed and secondly I was trying to apply this feature with as few possible changes as possible.

PR #2985, opened 2022-11-29
Setup autorelay

cammellos:
This commit setups AutoRelay and Hole punching.

What we observed using this code:
1) With a static address for auto relay, a hole punching session was setup successfully between two peers behind NAT and no UPnP.
This happened without asking the peers to connect directly, so the peer must have discovered themselves somehow.

I haven't replicated yet using the dynamic PeerSource, but I will try in the meantime, though it should work just the same as the same nodes are discovered.


Menduist:
LGTM, but can't comment on the specifics

richard-ramos (approved):
:100:

PR #2988, opened 2022-11-29
chore(waku2): bump go-waku

richard-ramos:


alaibe (approved):

rasom (approved):

vendor/github.com/waku-org/go-waku/waku/v2/discv5/discover.go
```diff
@@ -345,39 +377,21 @@ func (d *DiscoveryV5) FindNodes(ctx context.Context, topic string, opts ...disco
...
 
-	const maxLimit = 600
 	limit := options.Limit
-	if limit == 0 || limit > maxLimit {
-		limit = maxLimit
+	if limit == 0 || limit > MaxPeersToDiscover {
+		limit = MaxPeersToDiscover
+	}
+
+	if limit > MaxPeersToDiscover {
```
jrainville: This condition will never be met, since we change `limit` on line 382 in the case it's too high. Unless I'm missing something?

jrainville (commented):
Looks good, I think there is just one useless line

vendor/github.com/waku-org/go-waku/waku/v2/discv5/discover.go
richard-ramos: you're right!

jrainville (approved):

PR #2989, opened 2022-11-30
synchronize display name

### LINKED ISSUE #14395: Synchronize & backup profile picture and display name
Display name & profile pictures should be backed up and synchronize across devices:

### Scenario 1

1) Create account on device A
2) Set display name on account A
3) Set profile picture on account A
4) Pair device B
5) Sync all from device A to device B

Expected:
- Profile picture and display name is shown on device B

### Scenario 2 

1) Create account on device A
2) Set display name on account A
3) Set profile picture on account A
4) Unistall the app
5) Restore on the same device
Expected:
- Profile picture and display name is shown after relogin and fetching backup

### Note

Some of these features might be already implemented by desktop, please check the code and check with them to make sure no one is working on it.

qfrank:
this PR implemented the Scenario 1 and this [PR](https://github.com/status-im/status-go/pull/2976) implemented Scenario 2 for thie [issue](https://github.com/status-im/status-mobile/issues/14395), this is just a draft ATM, it need improvement as considering better solution for the clock of `display_name` when sync

Closes [#14395](https://github.com/status-im/status-mobile/issues/14395)

### DIFF
--- a/api/utils.go
+++ b/api/utils.go
@@ -22,7 +22,9 @@ func RunAsync(f func() error) <-chan error {
 
 // HashMessage calculates the hash of a message to be safely signed by the keycard
 // The hash is calulcated as
-//   keccak256("\x19Ethereum Signed Message:\n"${message length}${message}).
+//
+//	keccak256("\x19Ethereum Signed Message:\n"${message length}${message}).
+//
 // This gives context to the signed message and prevents signing of transactions.
 func HashMessage(message string) ([]byte, error) {
 	buf := bytes.NewBufferString("\x19Ethereum Signed Message:\n")
--- a/eth-node/crypto/crypto.go
+++ b/eth-node/crypto/crypto.go
@@ -202,7 +202,8 @@ func generateSecureRandomData(length int) ([]byte, error) {
 // safely used to calculate a signature from.
 //
 // The hash is calulcated as
-//   keccak256("\x19Ethereum Signed Message:\n"${message length}${message}).
+//
+//	keccak256("\x19Ethereum Signed Message:\n"${message length}${message}).
 //
 // This gives context to the signed message and prevents signing of transactions.
 func TextHash(data []byte) []byte {
@@ -214,7 +215,8 @@ func TextHash(data []byte) []byte {
 // safely used to calculate a signature from.
 //
 // The hash is calulcated as
-//   keccak256("\x19Ethereum Signed Message:\n"${message length}${message}).
+//
+//	keccak256("\x19Ethereum Signed Message:\n"${message length}${message}).
 //
 // This gives context to the signed message and prevents signing of transactions.
 func TextAndHash(data []byte) ([]byte, string) {
--- a/eth-node/crypto/gethcrypto.go
+++ b/eth-node/crypto/gethcrypto.go
@@ -37,7 +37,7 @@ import (
 	"github.com/status-im/status-go/eth-node/types"
 )
 
-//SignatureLength indicates the byte length required to carry a signature with recovery id.
+// SignatureLength indicates the byte length required to carry a signature with recovery id.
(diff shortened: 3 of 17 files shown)


qfrank:
closed as this [pr](https://github.com/status-im/status-go/pull/2976) already contains most of the changes

protocol/messenger_sync_settings.go
```diff
@@ -112,6 +111,14 @@ func (m *Messenger) extractSyncSetting(syncSetting *protobuf.SyncSetting) (*sett
 		value = json.RawMessage(v)
 	}
 
+	if syncSetting.GetType() == protobuf.SyncSetting_DISPLAY_NAME {
+		m.account.Name = value.(string)
+		err = m.multiAccounts.SaveAccount(*m.account)
```
saledjenic: @qfrank if the only need for this is to update account's display name in case when we're fetching data from waku, then it's redundant cause it is already handled [here](https://github.com/status-im/status-go/blob/5005b7b3cf45fed801257060fd68e19c4422e317/protocol/messenger_backup_handler.go#L77)

qfrank: emm..but [this](https://github.com/status-im/status-go/blob/5005b7b3cf45fed801257060fd68e19c4422e317/protocol/messenger_backup_handler.go#L77) only happens for back up, what if it's coming from #syncDevices? @saledjenic

saledjenic: Yes, that's true, that's why I said "if the only need for this is..." what do you think of moving that part to the place where you need it, [here](https://github.com/status-im/status-go/blob/5005b7b3cf45fed801257060fd68e19c4422e317/protocol/messenger.go#L3711-L3716), after that block or do the handling in a new fuction. 
Cause an intention of `extractSyncSetting` func is to convert `protobuf.SyncSetting` to `settings.SyncSettingField` that's why I wouldn't set account form that function.
What do you think about that?

qfrank: Oh, good point, sorry i forgot u renamed `handleSyncSetting` to `extractSyncSetting`, but i think it would be better not rename it so that we can keep the same code style as before as u can c that there'r lots of handleXXX within handleRetrievedMessages @saledjenic

qfrank: and we should invoke something like messageState.Response.AddSettings() to modify Response(@cammellos told me before in my old prs before), and it would be good within handleXXX, sorry I did not review your pr in time 😥 WDYT?

saledjenic: What I wanted initially is to have clear separation among handling device syncing and applying backed up data, like having different handling functions which at the end share the same functions like that `extractSyncSetting` to handle the job they need to do. 

I know that those 2 things are almost the same (just one is happening live with paired devices and another when needed), but I like clear separation like having a top level function for handling backed up messages (like this one [here](https://github.com/status-im/status-go/blob/5005b7b3cf45fed801257060fd68e19c4422e317/protocol/messenger_backup_handler.go#L21-L59) where we will add new things we need to handle) and having top level function for handling device syncing messages and they may call the same functions after that.

saledjenic: > and we should invoke something like messageState.Response.AddSettings() to modify Response(@cammellos told me before in my old prs before), and it would be good within handleXXX, sorry I did not review your pr in time 😥 WDYT?

This showed up while I was writing my previous message and didn't see it before.

So I wouldn't mix signals for notifying client about syncing among devices and applying fetched data from waku.

That's why I introduced new signal type and sending signals only in case if data from waku are applied (means if there are newer than those stored in the db). Also according to UI design we have there is a fetching data progress signal which supports it.

saledjenic (approved):

qfrank:
thanks for reviewing! @saledjenic

PR #2991, opened 2022-11-30
fix map iteration order

### LINKED ISSUE #2990: Map iteration order
# Problem
After submitted my draft [PR](https://github.com/status-im/status-go/pull/2989#issuecomment-1331855903) , i got error `jenkins/prs/tests — This commit cannot be built` , [error detail is here](https://ci.status.im/blue/rest/organizations/jenkins/pipelines/status-go/pipelines/prs/pipelines/tests/branches/PR-2989/runs/1/nodes/75/steps/82/log/?start=0)

the test `TestLatestMessageByChatID`  fails only some times not every time, the cause is map iteration in function `getMessagesAndCursorsFromScanRows`(protocol/message_persistence.go:2400), we can't rely on iterating over a map in a specific order, the iteration order of maps is also unstable, meaning that it won't necessarily be the same every time you run the program(maps are really messy!) , although we called `SortByClock` in `getMessagesAndCursorsFromScanRows`, what if message.Clock have the same value?

qfrank:
2 solutions to fix #2990  
the first one showed as what this PR does.
the second one: modify test function `TestLatestMessageByChatID`, make sure message.Clock have the right value,
but i prefer the first one :) as we won't fall into the trap easily when writing test code 🤔

Closes #2990

### DIFF
--- a/protocol/message_persistence.go
+++ b/protocol/message_persistence.go
@@ -2360,6 +2360,7 @@ func SortByClock(msgs HasClocks) {
 
 func getMessagesFromScanRows(db sqlitePersistence, rows *sql.Rows, withCursor bool) ([]*common.Message, error) {
 	messageIdx := make(map[string]*common.Message, 0)
+	var messages common.Messages
 	for rows.Next() {
 		// There's a possibility of multiple rows per message if the
 		// message has a discordMessage and the discordMessage has multiple
@@ -2382,16 +2383,12 @@ func getMessagesFromScanRows(db sqlitePersistence, rows *sql.Rows, withCursor bo
 
 		if msg, ok := messageIdx[message.ID]; !ok {
 			messageIdx[message.ID] = &message
+			messages = append(messages, &message)
 		} else if discordMessage := msg.GetDiscordMessage(); discordMessage != nil {
 			msg.Payload = getUpdatedChatMessagePayload(discordMessage, message.GetDiscordMessage())
-			messageIdx[message.ID] = msg
 		}
 	}
 
-	var messages common.Messages
-	for _, message := range messageIdx {
-		messages = append(messages, message)
-	}
 	SortByClock(messages)
 
 	return messages, nil
@@ -2400,8 +2397,8 @@ func getMessagesFromScanRows(db sqlitePersistence, rows *sql.Rows, withCursor bo
 func getMessagesAndCursorsFromScanRows(db sqlitePersistence, rows *sql.Rows) ([]*common.Message, []string, error) {
 
 	var cursors []string
+	var messages common.Messages
 	messageIdx := make(map[string]*common.Message, 0)
-
 	for rows.Next() {
 		// There's a possibility of multiple rows per message if the
 		// message has a discordMessage and the discordMessage has multiple
@@ -2420,16 +2417,12 @@ func getMessagesAndCursorsFromScanRows(db sqlitePersistence, rows *sql.Rows) ([]
 		if msg, ok := messageIdx[message.ID]; !ok {
 			messageIdx[message.ID] = &message
 			cursors = append(cursors, cursor)
+			messages = append(messages, &message)
 		} else if discordMessage := msg.GetDiscordMessage(); discordMessage != nil {
 			msg.Payload = getUpdatedChatMessagePayload(discordMessage, message.GetDiscordMessage())
-			messageIdx[message.ID] = msg
 		}
 	}
 
-	var messages common.Messages
-	for _, message := range messageIdx {
-		messages = append(messages, message)
-	}
 	SortByClock(messages)
 
 	return messages, cursors, nil
@@ -2438,6 +2431,7 @@ func getMessagesAndCursorsFromScanRows(db sqlitePersistence, rows *sql.Rows) ([]
 func getPinnedMessagesAndCursorsFromScanRows(db sqlitePersistence, rows *sql.Rows) ([]*common.PinnedMessage, []string, error) {
 
 	var cursors []string
+	var messages common.PinnedMessages
 	messageIdx := make(map[string]*common.PinnedMessage, 0)
 
 	for rows.Next() {
@@ -2457,17 +2451,13 @@ func getPinnedMessagesAndCursorsFromScanRows(db sqlitePersistence, rows *sql.Row
 				PinnedBy: pinnedBy,
 			}
 			messageIdx[message.ID] = pinnedMessage
+			messages = append(messages, pinnedMessage)
 			cursors = append(cursors, cursor)
 		} else if discordMessage := msg.Message.GetDiscordMessage(); discordMessage != nil {
 			msg.Message.Payload = getUpdatedChatMessagePayload(discordMessage, message.GetDiscordMessage())
-			messageIdx[message.ID] = msg
 		}
 	}
 
-	var messages common.PinnedMessages
-	for _, message := range messageIdx {
-		messages = append(messages, message)
-	}
 	SortByClock(messages)
 
 	return messages, cursors, nil
--- a/contracts/ethscan/address.go
+++ b/contracts/ethscan/address.go
@@ -9,9 +9,9 @@ import (
 var errorNotAvailableOnChainID = errors.New("not available for chainID")
 
 var contractAddressByChainID = map[uint64]common.Address{
-	1: common.HexToAddress("0x08A8fDBddc160A7d5b957256b903dCAb1aE512C5"), // mainnet
-	5: common.HexToAddress("0x08A8fDBddc160A7d5b957256b903dCAb1aE512C5"), // goerli
-	420: common.HexToAddress("0xf532c75239fa61b66d31e73f44300c46da41aadd"), // goerli optimism
+	1:      common.HexToAddress("0x08A8fDBddc160A7d5b957256b903dCAb1aE512C5"), // mainnet
+	5:      common.HexToAddress("0x08A8fDBddc160A7d5b957256b903dCAb1aE512C5"), // goerli
+	420:    common.HexToAddress("0xf532c75239fa61b66d31e73f44300c46da41aadd"), // goerli optimism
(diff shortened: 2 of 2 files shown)


protocol/message_persistence.go
```diff
@@ -2420,15 +2419,16 @@ func getMessagesAndCursorsFromScanRows(db sqlitePersistence, rows *sql.Rows) ([]
...
 			msg.Payload = getUpdatedChatMessagePayload(discordMessage, message.GetDiscordMessage())
 			messageIdx[message.ID] = msg
 		}
 	}
 
 	var messages common.Messages
-	for _, message := range messageIdx {
-		messages = append(messages, message)
+	for _, messageID := range messageIDs {
+		messages = append(messages, messageIdx[messageID])
```
0x-r4bbit: Can you do the same for `getPinnedMessagesAndCursorsFromScanRows` please?

0x-r4bbit: And also `getMessagesFromScanRows`

0x-r4bbit (changes requested):

protocol/message_persistence.go
qfrank: Sounds good! This will make my PR colorful 😉

protocol/message_persistence.go
```diff
@@ -2420,15 +2419,16 @@ func getMessagesAndCursorsFromScanRows(db sqlitePersistence, rows *sql.Rows) ([]
 		if msg, ok := messageIdx[message.ID]; !ok {
 			messageIdx[message.ID] = &message
 			cursors = append(cursors, cursor)
+			messageIDs = append(messageIDs, message.ID)
```
Samyoul: Could we append `message` directly to `var messages common.Messages`? That way the conversion loop is no longer required

Samyoul: Something like below:

```go
	var cursors []string
	var messages common.Messages

	messageIdx := make(map[string]*common.Message, 0)

	for rows.Next() {
		// There's a possibility of multiple rows per message if the
		// message has a discordMessage and the discordMessage has multiple
		// attachments
		//
		// Hence, we make sure we're aggregating all attachments on a single
		// common.Message
		var (
			message common.Message
			cursor  string
		)
		if err := db.tableUserMessagesScanAllFields(rows, &message, &cursor); err != nil {
			return nil, nil, err
		}
		if msg, ok := messageIdx[message.ID]; !ok {
			messageIdx[message.ID] = &message
			cursors = append(cursors, cursor)
			message = append(messages, message)
		} else if discordMessage := msg.GetDiscordMessage(); discordMessage != nil {
			msg.Payload = getUpdatedChatMessagePayload(discordMessage, message.GetDiscordMessage())
			messageIdx[message.ID] = msg
		}
	}
	SortByClock(messages)

	return messages, cursors, nil

```

0x-r4bbit: I think we can if the update to `messageIdx[message.ID]` in the `else if` case will also update the same item in `messages` (via reference)

protocol/message_persistence.go
qfrank: also getMessagesAndCursorsFromScanRows? found by searching SortByClock

protocol/message_persistence.go
```diff
@@ -2382,16 +2383,13 @@ func getMessagesFromScanRows(db sqlitePersistence, rows *sql.Rows, withCursor bo
 
 		if msg, ok := messageIdx[message.ID]; !ok {
 			messageIdx[message.ID] = &message
+			messages = append(messages, &message)
 		} else if discordMessage := msg.GetDiscordMessage(); discordMessage != nil {
 			msg.Payload = getUpdatedChatMessagePayload(discordMessage, message.GetDiscordMessage())
 			messageIdx[message.ID] = msg
```
Samyoul: The current implementation doesn't update a `message` in `messages`. It overwrites the pointer in `messageIdx[message.ID]`. But what we need to do is update the value of the pointer and not the pointer itself.

I believe the below code would fix that problem

```suggestion
			*messageIdx[message.ID] = *msg
```

qfrank: emm.. I think current implementation will update a message in messages? also i think we don't even need the line?
```
messageIdx[message.ID] = msg
```

0x-r4bbit: We need that line because otherwise we'll never update the message's payload in case there's multiple rows for that message. That's what I tried pointing out here: https://github.com/status-im/status-go/pull/2991#discussion_r1036038644


Let's play this through and imagine there's 2 rows for this message, so we'll pass through this snippet here twice:
```golang
		if msg, ok := messageIdx[message.ID]; !ok {
			messageIdx[message.ID] = &message
			messages = append(messages, &message)
		} else if discordMessage := msg.GetDiscordMessage(); discordMessage != nil {
			msg.Payload = getUpdatedChatMessagePayload(discordMessage, message.GetDiscordMessage())
			messageIdx[message.ID] = msg
	        }
```

1. In the first round, `messageIdx[message.iD]` doesn't exist, so we end up in that first `if` block and store the message in the map (we store it there, so we can quickly access it in the second row, to update its payload). We then add our `message` to the list of messages to be returned.
2. In the second round, the message now does exists in `messageIdx` so we endup in the `else if` block. We extract the message paylaod from the second row and update the `msg` in our `messageIdx` such that it includes that payload.(This is the part we can not leave out, otherwise we'll return incomplete message data. It's crucial that the `msg` that lives in `messages` gets that same update.

As per @Samyoul 's comment, if we do

```golang
messageIdx[message.ID] = msg
```

In we'll overwrite the pointer in `messagesIdx` for `msg`, which is the same that lives in `messages`!
We want instead update the **value** of the pointer hence:

```golang
messageIdx[message.ID] = *msg
```

This will ensure the message in `messages` will receive that update of the second round.

Hope this makes sense..

Samyoul: > emm.. I think current implementation will update a message in messages? also i think we don't even need the line?

@qfrank, I apologise I see that you are correct. `msg` is already a pointer to the active `common.Messenger`

**Sample Code**

```go
type discordMsg struct {
	Data []byte
}
type msge struct {
	ID         string
	Payload    []byte
	DiscordMsg *discordMsg
}

func (m *msge) GetDiscordMessage() *discordMsg {
	return m.DiscordMsg
}

func getUpdatedChatMessagePayload(original, attachment *discordMsg) []byte {
	original.Data = append(original.Data, attachment.Data...)
	return original.Data
}

func Test(t *testing.T) {
	var messages []*msge
	messageIdx := make(map[string]*msge, 0)

	data := []msge{
		{ID: "123"},
		{ID: "234"},
		{ID: "345"},
		{ID: "456", DiscordMsg: &discordMsg{Data: []byte{0xff, 0xfe, 0xfd}}},
		{ID: "456", DiscordMsg: &discordMsg{Data: []byte{0xfc, 0xfb, 0xfa}}},
		{ID: "456", DiscordMsg: &discordMsg{Data: []byte{0xf9, 0xf8, 0xf7}}},
	}

	for i := range data {
		if msg, ok := messageIdx[data[i].ID]; !ok {
			messageIdx[data[i].ID] = &data[i]
			messages = append(messages, &data[i])
		} else if discordMessage := msg.GetDiscordMessage(); discordMessage != nil {
			msg.Payload = getUpdatedChatMessagePayload(discordMessage, data[i].GetDiscordMessage())
		}
	}

	spew.Dump(messageIdx, messages)
}
```

Output
```
(map[string]*server.msge) (len=4) {
 (string) (len=3) "345": (*server.msge)(0xc0005202a0)({
  ID: (string) (len=3) "345",
  Payload: ([]uint8) <nil>,
  DiscordMsg: (*server.discordMsg)(<nil>)
 }),
 (string) (len=3) "456": (*server.msge)(0xc0005202d0)({
  ID: (string) (len=3) "456",
  Payload: ([]uint8) (len=9 cap=16) {
   00000000  ff fe fd fc fb fa f9 f8  f7                       |.........|
  },
  DiscordMsg: (*server.discordMsg)(0xc00050ef18)({
   Data: ([]uint8) (len=9 cap=16) {
    00000000  ff fe fd fc fb fa f9 f8  f7                       |.........|
   }
  })
 }),
 (string) (len=3) "123": (*server.msge)(0xc000520240)({
  ID: (string) (len=3) "123",
  Payload: ([]uint8) <nil>,
  DiscordMsg: (*server.discordMsg)(<nil>)
 }),
 (string) (len=3) "234": (*server.msge)(0xc000520270)({
  ID: (string) (len=3) "234",
  Payload: ([]uint8) <nil>,
  DiscordMsg: (*server.discordMsg)(<nil>)
 })
}
([]*server.msge) (len=4 cap=4) {
 (*server.msge)(0xc000520240)({
  ID: (string) (len=3) "123",
  Payload: ([]uint8) <nil>,
  DiscordMsg: (*server.discordMsg)(<nil>)
 }),
 (*server.msge)(0xc000520270)({
  ID: (string) (len=3) "234",
  Payload: ([]uint8) <nil>,
  DiscordMsg: (*server.discordMsg)(<nil>)
 }),
 (*server.msge)(0xc0005202a0)({
  ID: (string) (len=3) "345",
  Payload: ([]uint8) <nil>,
  DiscordMsg: (*server.discordMsg)(<nil>)
 }),
 (*server.msge)(0xc0005202d0)({
  ID: (string) (len=3) "456",
  Payload: ([]uint8) (len=9 cap=16) {
   00000000  ff fe fd fc fb fa f9 f8  f7                       |.........|
  },
  DiscordMsg: (*server.discordMsg)(0xc00050ef18)({
   Data: ([]uint8) (len=9 cap=16) {
    00000000  ff fe fd fc fb fa f9 f8  f7                       |.........|
   }
  })
 })
}
```

Samyoul (approved):
Really nice change

0x-r4bbit (approved):

protocol/message_persistence.go
0x-r4bbit: Sorry for the late update here. Yes @qfrank and I have discussed this offline as well.
Thanks @Samyoul for the detailed reproducible!

PR #2993, opened 2022-12-01
Feat/wallet ticker

alaibe:
Add a timer of 10min which send event that client should reload data


services/wallet/reader.go
```diff
@@ -89,13 +94,22 @@ func getTokenAddresses(tokens []*token.Token) []common.Address {
...
 
-func (r *Reader) Start(ctx context.Context, chainIDs []uint64) error {
-	accounts, err := r.s.accountsDB.GetAccounts()
-	if err != nil {
-		return err
-	}
-
-	return r.s.transferController.CheckRecentHistory(chainIDs, getAddresses(accounts))
+func (r *Reader) Start() error {
+	ticker := time.NewTicker(10 * time.Minute)
```
richard-ramos:
```suggestion
	ticker := time.NewTicker(10 * time.Minute)
	defer ticker.Stop()
```

richard-ramos: Also, probably makes sense to move the ticker inside the goroutine

services/wallet/reader.go
```diff
@@ -89,13 +94,22 @@ func getTokenAddresses(tokens []*token.Token) []common.Address {
...
+	ctx := context.Background()
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
```
richard-ramos: Is ctx.Done() called when `Logout()` is called? if not, this goroutine will run forever

services/wallet/reader.go
alaibe: indeed, moved to inside the coroutine

services/wallet/reader.go
alaibe: updated

richard-ramos (approved):

jrainville (approved):
LGTM

PR #2995, opened 2022-12-01
feat: add signal for starting history archive downloads

0x-r4bbit:

PR #2998, opened 2022-12-02
feat: add signal indicating that history messages are being imported

0x-r4bbit:
In order to give clients more insights about archive messages being processed, we're adding this additional signal that informs clients when the import of downloaded history archive messages has started.


0x-r4bbit:
I've recently introduced https://github.com/status-im/status-go/pull/2995 which adds a similar signal for when the download of archives started.

In Desktop we use this signal to show a banner to reduce chances users will close the app while archive messages are being processed. @John-44 requested that we show the banner only for when messages are indeed imported into the database, not necessarily already when the torrent data is being downloaded.

To make this work we either have to move the signal introduced in #2995 to a different place, or introduce a new signal that is emitted at the correct time. I went with introducing an additional signal because it would be semantically incorrect to signal that message download has started, when in reality the download has already finished.

In desktop we'll then use this new signal to show the banner.

endulab (approved):

PR #2999, opened 2022-12-02
feat(communities): Add admin node status update

borismelnik:
For [this issue](https://github.com/status-im/status-desktop/issues/8417) we need a mechanism which broadcast that admin node is online.


borismelnik:
Guys, it will be great to discuss this solution :)

protocol/messenger_status_updates.go
```diff
@@ -91,6 +91,41 @@ func (m *Messenger) sendUserStatus(ctx context.Context, status UserStatus) error
 	return nil
 }
 
+func (m *Messenger) sendNodeStatus(ctx context.Context) {
+	communities, err := m.Communities()
```
jrainville, line 95: Maybe `Joined` would be better here

jrainville (commented):
Code looks good. Just 1 small comment.

I wonder if a ping and ACK wouldn't be less spammy for the network. By that I mean, when you observe a community, you send a ping to the owner and if the owner is online, he sends an ACK so the person wanting to join knows he can instantly join.

The downside would be if a big community has a lot new joiners at the same time, that would be a lot of pings, but we can make the ACK general (like what you did here) so that the owner responds once and everyone online can understand.

At least, when there is no person wanting to join, the owner node wouldn't have to broadcast and also wouldn't have to expose their status when not needed.

What do you think? cc @cammellos @richard-ramos

cammellos:
> Code looks good. Just 1 small comment.
> 
> I wonder if a ping and ACK wouldn't be less spammy for the network. By that I mean, when you observe a community, you send a ping to the owner and if the owner is online, he sends an ACK so the person wanting to join knows he can instantly join.
> 
> The downside would be if a big community has a lot new joiners at the same time, that would be a lot of pings, but we can make the ACK general (like what you did here) so that the owner responds once and everyone online can understand.
> 
> At least, when there is no person wanting to join, the owner node wouldn't have to broadcast and also wouldn't have to expose their status when not needed.
> 
> What do you think? cc @cammellos @richard-ramos

I think we should push back with these kind of features, it's not really worth making these changes just to show the user "Join community" vs "Request to join", it's a minor UX improvement and it's going to degrade other parts of the user experience (more data on store nodes, more traffic in the network, slower fetching of history, slower dispatching of messages), which will have a real impact over the usage of the app.

This is completely in control of the admin of the community, if they want to give their users the best user experience, they just have to be online, and if they are and they have automatic approval, the user will be allowed in the community immediately anyway.

Not your fault, this is something we discussed recently, unfortunately designs are loaded with features that haven't been vetted from the tech team before going ahead, but since that seems to be the case, we should probably all try to be vigilant to spot these early in the process.

cc @pedro-et @John-44

protocol/messenger.go
```diff
@@ -3433,6 +3434,17 @@ func (m *Messenger) handleRetrievedMessages(chatWithMessages map[transport.Filte
 							continue
 						}
 
+					case protobuf.NodeStatusUpdate:
+						p := msg.ParsedMessage.Interface().(protobuf.NodeStatusUpdate)
+						m.outputToCSV(msg.TransportMessage.Timestamp, msg.ID, senderID, filter.Topic, filter.ChatID, msg.Type, p)
+						logger.Debug("Handling NodeStatusUpdate", zap.Any("message", p))
+						err = m.HandleNodeStatusUpdate(messageState, p)
```
0x-r4bbit, line 3441: This doesn't seem to take into account that this message has not originated from the community owner.
Even if this is just a "Hey I'm online" ping, theoretically any node can send these for any community. 

So if we want to keep this, we should probably check whether we're interested in this particular node status update in the first place, before we process it.

0x-r4bbit:
I'm with @cammellos here on this one.

We're broadcasting a node status update every 5 minutes just so we have a signal to change a label in the UI.
Probably not worth it if we could just always default to "Request access" labels.

0x-r4bbit:
> you send a ping to the owner and if the owner is online, he sends an ACK so the person wanting to join knows he can instantly join.

@jrainville this sounds like a lightweight version of what request-to-join -> accept request-to-join already does.
I don't think this effort is worth it given that we just want to change the semantics of a label..

0x-r4bbit:
Another thing to keep in mind, which I've mentioned in https://github.com/status-im/status-desktop/issues/8417#issuecomment-1337137698, even if we do implement something like this, there's still no guarantee that the owner node will be online by the time the soon-to-be-member hits a "join community" button.

richard-ramos (commented):
Please do not include migrations files that were not affected by this PR (like thie bindata.go inside appdatabase/migrations, appdatabase/migrationsprevnodecfg/, etc)

richard-ramos:
> Please do not include migrations files that were not affected by this PR (like thie bindata.go inside appdatabase/migrations, appdatabase/migrationsprevnodecfg/, etc)

Oops, I just saw that this PR is ancient. Is it still necessary?

jrainville:
No longer needed. Issue was closed

PR #3002, opened 2022-12-05
chore: bump go-waku to disconnect peers if keep alive loop wasnt executed recently

### LINKED ISSUE #2996: waku2: periods of no connectivity are not identified correctly?
Before suspending my machine, i had ~15 peers and was receiving messages successfully.

The next morning, after waking up my machine, status-desktop appeared to function normally (i could send and receive messages), but @jrainville told me about some messages in Status community's #messenger channel that I did not receive and were sent during the time my machine was suspended.

No new requests were done to the storenodes and the disconnected banner did not appear when I woke up my machine. This makes me think that go-waku / status-go did not identify correctly that there was a period of time where the machine was off, and did not notice that weren't any peers connected (go-waku does take some time to know when a peer is not connected before removing it from the list)

cc: @cammellos @rasom @iurimatias 
Also cc: @jm-clius @LNSD @fryorcraken since it might be something interesting to look at for chat2

richard-ramos:
Fixes #2996

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -80,7 +80,7 @@ require github.com/fogleman/gg v1.3.0
 require (
 	github.com/gorilla/sessions v1.2.1
 	github.com/meirf/gopart v0.0.0-20180520194036-37e9492a85a8
-	github.com/waku-org/go-waku v0.2.3-0.20221129201435-94d20ab12276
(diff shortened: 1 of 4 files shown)


jrainville (approved):
LGTM

PR #3006, opened 2022-12-06
fix(community): don't delete chat and messages when leaving community

jrainville:
Desktop issue: https://github.com/status-im/status-desktop/issues/7512

This creates a smoother experience for users when they leave a community since they can see the exact same messages they had before without having to rely on the mailserver.

To make it work on the desktop or mobile, side you need to make sure that you re-fetch the messages for the channels when re-joining (or re-spectating). This might require to reset the cursor.

### DIFF
--- a/protocol/communities_messenger_test.go
+++ b/protocol/communities_messenger_test.go
@@ -1746,12 +1746,32 @@ func (s *MessengerCommunitiesSuite) TestLeaveAndRejoinCommunity() {
 	s.Require().NoError(err)
 	s.Require().Equal(2, joinedCommunities[0].MembersCount())
 
+	chats, err := s.alice.persistence.Chats()
+	s.Require().NoError(err)
+	var numberInactiveChats = 0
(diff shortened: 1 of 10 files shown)


cammellos (changes requested):
I think this one might leave the chats active, so that you'd be still fetching messages etc (on next login for example).
We have `deactivateChat` that sets the chat to inactive, and that could be used, also I think we should add a few tests to validate the behavior.

jrainville:
Good point @cammellos . I added the call to `deactivateChat`. I had to add a new argument so that we can select if we want the history to be cleared. 
I'll add the tests tomorrow morning.

jrainville:
@cammellos ready for review.
I updated the `TestLeaveAndRejoinCommunity` to check the number of active chats after leave and then after rejoin

richard-ramos (approved):

cammellos (approved):

PR #3011, opened 2022-12-09
Remove group members

### LINKED ISSUE #3009: An endpoint for removing a list of users from a group
We currently have a method `RemoveMemberFromGroupChat` which removes a single member from a group chat. We need to have an endpoint that removes a list of users from the group chat.

OmarBasem:
Fixes #3009

This PR adds a method to remove a list of members from a group at once.
Needed by: https://github.com/status-im/status-mobile/pull/14494


0x-r4bbit (approved):
This LGTM however there are some details/APIs in this PR that I'm not familiar with, so I'll defer to the other reviewers to double check the correctness here.

Also, bonus points if there was a test for this new API

cammellos (changes requested):
Great job @OmarBasem! looks good, only thing, could you please add some tests for the new functionality?

osmaczko (changes requested):
The logic of the new endpoint is almost the same as for `RemoveMemberFromGroupChat`. Could you please remove the duplication by making `RemoveMemberFromGroupChat` use `RemoveMembersFromGroupChat` ?

```go
func (m *Messenger) RemoveMemberFromGroupChat(ctx context.Context, chatID string, member string) (*MessengerResponse, error) {
    return m.RemoveMembersFromGroupChat(ctx, chatID, []string{member})
}
```

OmarBasem:
> The logic of the new endpoint is almost the same as for `RemoveMemberFromGroupChat`. Could you please remove the duplication by making `RemoveMemberFromGroupChat` use `RemoveMembersFromGroupChat` ?
> 
> ```go
> func (m *Messenger) RemoveMemberFromGroupChat(ctx context.Context, chatID string, member string) (*MessengerResponse, error) {
>     return m.RemoveMembersFromGroupChat(ctx, chatID, []string{member})
> }
> ```

Sure! Just to confirm, do you mean `RemoveMemberFromGroupChat` to use `RemoveMembersFromGroupChat` or the opposite?

OmarBasem:
> Great job @OmarBasem! looks good, only thing, could you please add some tests for the new functionality?

Sure, I will!

osmaczko (approved):
LGTM :)

cammellos (approved):
It'd be good to have a test that removes multiple members, but it's ok I guess, also don't forget to bump the version (this is breaking change, so you can bump the `y`)
Great work!

PR #3012, opened 2022-12-09
Better logs and more robust download handling for community archives

0x-r4bbit:
There's descriptions in each commit but in a nutshell:

- Adding better logs to history archive handling for better debugging
- Better handling of multiple magnetlinks that are being processed at roughly the same time (proper interruption of download and data import with resumability)
- Tracking of which archives were successfully imported (not just downloaded)

PR #3013, opened 2022-12-09
feat(communities): add magnetlink to RequestToJoinResponse

0x-r4bbit:
This is done so that when member join a community by being accepted by the community owner, they also receive the most up-to-date magnetlink along with it.


protocol/messenger_handler.go
```diff
@@ -1146,8 +1150,22 @@ func (m *Messenger) HandleCommunityRequestToJoinResponse(state *ReceivedMessageS
...
+			community := response.Communities()[0]
+			state.Response.AddCommunity(community)
+			state.Response.AddCommunitySettings(communitySettings)
+
+			magnetlink := requestToJoinResponseProto.MagnetUri
+			if m.torrentClientReady() && communitySettings != nil && communitySettings.HistoryArchiveSupportEnabled && magnetlink != "" {
+				m.downloadHistoryArchiveTasksWaitGroup.Add(1)
+				go func() {
+					defer m.downloadHistoryArchiveTasksWaitGroup.Done()
+					m.downloadAndImportHistoryArchives(community.ID(), magnetlink)
```
cammellos: What happens if we receive the message twice? does it start two separate tasks?

cammellos (commented):
Looks good, I have checked only though the magnet link part and not the refactoring, I left a question

qfrank:
Thanks for inviting to review, i'll do it today @0x-r4bbit

protocol/messenger_handler.go
```diff
@@ -927,109 +927,113 @@ func (m *Messenger) HandleHistoryArchiveMagnetlinkMessage(state *ReceivedMessage
...
+
+importMessageArchivesLoop:
+	for {
+		select {
+		case <-cancel:
+			m.communitiesManager.LogStdout("interrupted importing history archive messages")
+			return
+		case <-importTicker.C:
+
+			archiveIDsToImport, err := m.communitiesManager.GetMessageArchiveIDsToImport(id)
```
qfrank, line 995: is there a possible that 2 goroutines  reach here at almost the same time?

qfrank (commented):
i left only one question for u :)

protocol/messenger_handler.go
0x-r4bbit: Right, so generally we keep track of any history archive tasks per community such that, in needed, we can cancel them.

The way we do this is by keeping a list of channels that can emit a `cancel` signal. For example, you can see it being retrieved here, a few lines above: https://github.com/status-im/status-go/pull/3013/files#diff-59d196f7f72de56c64ba2c559fcbb51762c418a22d2b83c59662b772a09b3c0aR961

Here's roughly how it works:

1. When we receive a magnetlink, we and there's already a `historyArchiveDownloadTask` for this community, we'll take it and emit a cancel signal on it.  Then, we create a new one for the currently processed magnet link messages
2. The `cancel` channel is used to:
2.1. Cancelling the download of the community archive index
2.2. Cancelling the download of community archive data
2.3 Cancelling the import of download archive data into the database

So at any stage, when there's multiple magnetlink messages for the same community coming in, every message will first mark any previous download/import task for that community as cancelled before proceeding.

So there should only ever be one download/import task in progress.

protocol/messenger_handler.go
0x-r4bbit: So here the same mechanism kicks in as mentioned in: https://github.com/status-im/status-go/pull/3013/files#r1045551410

essentially `downloadAndImportHistoryArchives()` will first cancel any ongoing download/importing tasks for this community, then start its own.

protocol/messenger_handler.go
qfrank: Suppose we have two goroutines A and B, A executes to line 973 (message_handler.go), then B executes to line 2138 (manager.go) `close(stop)`, B won't block after executing close and won't wait for A to finish, then it is possible that B and A both enter the same area(line 973)? @0x-r4bbit

0x-r4bbit: Ah I see.

So yea, I think in theory it's possible. In practice I'd say it's unlikely though. Assuming the scenario you've described above, even though B won't block, it would have to be very fast (downloading archive data) to reach this line before A hits `cancel`

But I see where you're coming from. To be 100% sure, we'd have to block `B` until we know for sure that `A` has finished cancelling.

0x-r4bbit: I've made the necessary changes in #3012 
This PR has been updated accordingly 

TLDR:

It checks if there's an ongoing download task via and cancels it using its attached cancel channel. It then piggy bags on the same mechanism where we wait for cancellation to finish until we continue

qfrank: nice job🫶

0x-r4bbit:
@qfrank would you mind approving this PR then?

qfrank (approved):

qfrank:
> @qfrank would you mind approving this PR then?

NP 🙌

PR #3014, opened 2022-12-09
restart discv5

cammellos:
When discovery fails to be seeded with bootstrap/fallback nodes, it
never recovers.

This commit changes the behavior so that status-go retries fetching
bootnodes, and restarts discovery when that happens.

It also sets the default DNS to cloudflare, not great, but as a quick fix for mobile.

I have made some changes to vendor/ one PR is out, https://github.com/waku-org/go-discover/pull/3
but the other one (waku-org/go-waku), relies on the updated go-discover, but you can look at vendor/ for the changes, they are pretty trivial.


rasom (approved):

richard-ramos (approved):
:100:
I'll test it on Desktop

PR #3019, opened 2022-12-12
debug: add more logs to determine why community requests are missing

richard-ramos:


jrainville (approved):

protocol/transport/filters_manager.go
```diff
@@ -158,6 +158,8 @@ func (f *FiltersManager) InitCommunityFilters(pks []*ecdsa.PrivateKey) ([]*Filte
 
 		f.filters[filterID] = filter
 
+		f.logger.Debug("registering filter for", zap.String("chatID", filterID), zap.String("topic", rawFilter.Topic.String()))
```
jrainville: We could add a similar logs below when we add other filters

richard-ramos: Sounds good. I'll add them.

PR #3021, opened 2022-12-13
Images Album

OmarBasem:
This PR adds `AlbumID` field to messages in order to be able to group images sent together like a collage.


stefandunca (approved):

cammellos (approved):

0x-r4bbit:
@OmarBasem we have users reporting:

```
ERR 2022-12-14 16:13:22.100+01:00 rpc response error                         topics="rpc" tid=52873 file=core.nim:29 err="\nstatus-go error [methodName:chat_getChats, code:-32000, message:sql: Scan error on column index 18, name \"album_id\": converting NULL to string is unsupported ]\n"
```

Are we show this works with databases that have messages which don't have an `album_id` ?

0x-r4bbit:
I think we need to ensure that this works with existing databases that don't have an `AlbumID` by wrapping `album_id` in 


```
COALESCE(album_id, "")
```

0x-r4bbit:
Created https://github.com/status-im/status-go/pull/3023 accordingly

PR #3023, opened 2022-12-14
fix: ensure queries coalesce `album_id` value when scanning

0x-r4bbit:
Otherwise databases with existing messages that don't have an `album_id` will run into and `converting NULL to string is unsupported` error.

See: https://github.com/status-im/status-go/pull/3021#issuecomment-1351623814


caybro (approved):
Tested it locally, works (w/o it, status-desktop goes straight crashing on startup)

OmarBasem (approved):
Thank you for this fix!

PR #3025, opened 2022-12-14
fix: fix special messages not being received because not encrypted

jrainville:
Fixes this desktop issue https://github.com/status-im/status-desktop/issues/8689

The special messages like: delete, edit, pin and reaction were **all** not being received, because they were not encrypted, so they were not using the right topic.
Debugging this was a pain in the ass, but @richard-ramos and I finally figured it out.

Let me know if more message types need to be encrypted.

Also, is there any change that the pending community requests might be caused by something similar, let me know, I'll check it out.

### DIFF
--- a/protocol/common/message_sender.go
+++ b/protocol/common/message_sender.go
@@ -232,6 +232,14 @@ func (s *MessageSender) getMessageID(rawMessage *RawMessage) (types.HexBytes, er
 	return messageID, nil
 }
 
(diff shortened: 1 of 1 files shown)


cammellos (approved):
Nice catch, probably we should make this function a bit less dangerous and be more explicit when we want to send a message to the community admin/owner vs send a message to anyone in the community, to avoid similar issues in the future, but that's for a different PR.
Great work!

richard-ramos (approved):

osmaczko (approved):

PR #3026, opened 2022-12-15
Recognize `@everyone` as mention in messages

0x-r4bbit:
This "id" or "tag" will be interpeted/interpolated as `@everyone` by
clients.

For this change to work

https://github.com/status-im/markdown/pull/5

is needed.

### DIFF
--- a/protocol/common/message.go
+++ b/protocol/common/message.go
@@ -70,6 +70,8 @@ const (
 	ContactVerificationStateCanceled
 )
 
+const everyoneMentionTag = "0x00001"
+
 type CommandParameters struct {
 	// ID is the ID of the initial message
 	ID string `json:"id"`
@@ -470,7 +472,7 @@ func (v *MentionsAndLinksVisitor) Visit(node ast.Node, entering bool) ast.WalkSt
 	switch n := node.(type) {
 	case *ast.Mention:
 		mention := string(n.Literal)
-		if mention == v.identity {
+		if mention == v.identity || mention == everyoneMentionTag {
 			v.mentioned = true
 		}
 		v.mentions = append(v.mentions, mention)
--- a/go.mod
+++ b/go.mod
@@ -51,7 +51,7 @@ require (
 	github.com/prometheus/client_golang v1.13.0
 	github.com/russolsen/transit v0.0.0-20180705123435-0794b4c4505a
 	github.com/status-im/doubleratchet v3.0.0+incompatible
-	github.com/status-im/markdown v0.0.0-20220622180305-7ee4aa8bbc3f
+	github.com/status-im/markdown v0.0.0-20221220095528-8f1babe09d1e
 	github.com/status-im/migrate/v4 v4.6.2-status.2
 	github.com/status-im/rendezvous v1.3.6
 	github.com/status-im/status-go/extkeys v1.1.2
--- a/go.sum
+++ b/go.sum
@@ -1965,8 +1965,8 @@ github.com/status-im/gomoji v1.1.3-0.20220213022530-e5ac4a8732d4/go.mod h1:hmpnZ
 github.com/status-im/keycard-go v0.0.0-20190316090335-8537d3370df4/go.mod h1:RZLeN1LMWmRsyYjvAu+I6Dm9QmlDaIIt+Y+4Kd7Tp+Q=
 github.com/status-im/keycard-go v0.0.0-20200402102358-957c09536969 h1:Oo2KZNP70KE0+IUJSidPj/BFS/RXNHmKIJOdckzml2E=
 github.com/status-im/keycard-go v0.0.0-20200402102358-957c09536969/go.mod h1:RZLeN1LMWmRsyYjvAu+I6Dm9QmlDaIIt+Y+4Kd7Tp+Q=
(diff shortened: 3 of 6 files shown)


richard-ramos (dismissed):

protocol/common/message.go
```diff
@@ -504,7 +529,7 @@ func (m *Message) PrepareContent(identity string) error {
 	// Leave it set if already set, as sometimes we might run this without
 	// an identity
 	if !m.Mentioned {
-		m.Mentioned = visitor.mentioned
+		m.Mentioned = visitor.mentioned || containsMentionOf(m.Text, EveryoneMentionLiteral)
```
jrainville: Maybe there is a way to put `@everyone` in the checks done for `runMentionsAndLinksVisitor` but I have no idea how it works so it can be done later.

jrainville (approved):

protocol/common/message.go
```diff
@@ -70,6 +70,8 @@ const (
 	ContactVerificationStateCanceled
 )
 
+const EveryoneMentionLiteral = "@everyone"
```
cammellos: You can use lowercase if not exported

cammellos (changes requested):

protocol/common/message.go
```diff
@@ -481,6 +483,29 @@ func (v *MentionsAndLinksVisitor) Visit(node ast.Node, entering bool) ast.WalkSt
...
+		idx := strings.Index(literal, mention)
+
+		if strings.HasPrefix(literal, mention) && string(literal[mentionLen]) == " " ||
+			strings.HasSuffix(literal, mention) && idx > 0 && string(literal[idx-1]) == " " {
+			return true
+		}
+
+		if literalLen >= mentionLen+2 && ((idx > 0 && string(literal[idx-1]) == " ") && (string(literal[idx+mentionLen]) == " ")) {
+			return true
+		}
```
jrainville: Maybe a regex would be better here for the optional spaces

cammellos:
We don't do parsing in status-go (at least we should not), we have our markdown library and that's the place where we want to do this for performance reasons, this is also scanning the string multiple times (`Contains`, `IndexOf`, all of this can be easily done in one pass of the string), and also no tests.

Definitely something that should have been flagged @richard-ramos @jonathanbarker please be more careful next time before approving

protocol/common/message.go
cammellos: we should not use regexes for parsing messages as much as possible, as they tend to lead to performance issue (and possible dos), the markdown library we use tries (I believe we do), to pass only once through the string.

richard-ramos:
Oops! good catch @cammellos!  removing my approval immediatly :-)

jrainville:
Sorry about that, I didn't know about the no parsing in status-go.
Also, sorry to JB for being pinged for no reason :joy:

protocol/common/message.go
0x-r4bbit: Ah, that's what I did initially. The problem then is however, that when a node of the AST is visited, we don't know it's surroundings. These will end up in the `default` case (they aren't `ast.Mention` or `ast.Link`).

So the `literal` to inspect will be "@everyone" even though the message might be  [backtick]@everyone[backtick] at which point we don't want this to become recgonized as mention.

The options are to

1. either extend the markdown parser to know what an `@everyone` mention is and that it's to interpret as a mention or
2. Simply check the literal on the entire text (instead of AST)

I went with 2) as it seem simpler

protocol/common/message.go
0x-r4bbit: Will change this

protocol/common/message.go
0x-r4bbit: Well, now that I read https://github.com/status-im/status-go/pull/3026#issuecomment-1353427503 I guess I'll have to make it work with 1)  haha

0x-r4bbit:
@cammellos thanks for pointing this out. Will move this into markdown's parser

cammellos:
> Sorry about that, I didn't know about the no parsing in status-go. Also, sorry to JB for being pinged for no reason joy

No worries :)

>  @cammellos thanks for pointing this out. Will move this into markdown's parser


Thanks,

We kind of need to decided as well how to handle these tags, since we might have custom ones etc, potentially we could even have `@{id}` for example, where everyone is just `@0` or `@0x001` say, and then the might be contextual to the place you receive in, say we add groups eventually (in community `y`, `@0x002` will refer to group 2, which is created by the admin, which translates to `Desktop` for example), or something like that, not sure what's the best

0x-r4bbit:
@cammellos what you're describing there are some very interesting ideas. I didn't think about it that way.

It probably makes sense to introduce "namespaces" for different things. For example:

1. `@0x000 - `@0x010`
For system related mentions (like `@everyone`, `@here`, `@channel` etc)

2. `@0x100` - `0x1xx`
For groups. Although this would only allow for 100 groups (which is probably enough), we could make it more explicit and have `@0xg000 - @0xgxx...`

We should nail these down before implementing them. At least the system related namespace

0x-r4bbit:
@cammellos added support for recognizing `@0xXXXXX` as per out discussion in: https://github.com/status-im/markdown/pull/5

Will update this PR accordingly

jrainville (approved):

0x-r4bbit:
This is green, just not compiling on macOS because something's wrong with CI's macOS

cammellos (approved):

PR #3027, opened 2022-12-15
fix: fix disappearing group chats by setting active to true by default

### LINKED ISSUE #8773: Group chat disapeared but notificaitons still came in
# Bug Report

## Description

Very weird bug that happened to me and Rick. 

I opened the app and everything looked normal, but then I started seeing notifications come in for new messages (both OS and ephemeral depending on the focus state of the app), but **none** of my channel or communities showed the badge for new messages. I checked my muted channels just in case, and it wasn't that either.

Turns out, one of my private group chats just disappeared from the list. I could not see it at all, but  I could still receive the messages and notifications. So the signals from status-go worked.

This means the issue is either in Nim or QML.

Closing and reopening didn't fix it.

We could fix it by kicking me out of the group and re-adding me. Then I could see the group and even the old messages. 

## Steps to reproduce

No idea. It happened seemingly randomly

### Additional Information

- Status desktop version: 0.8.1
- Operating System: Ubuntu 20

jrainville:
This should fix https://github.com/status-im/status-desktop/issues/8773 where group chats sometimes disappeared, or when re-importing from seed, they didn't appear, as they were not set Active.

I reproduced the issue after an import and I saw that `Active` was set to false for the chat in the DB.

I tracked in the code that this would be the most probable place where Active could be set to `false` by error.

Since we no longer have group chats with invitation and approvals, I think we can set it to `true` by default. I also removed `waitingForApproval` as it seems like legacy code that is no longer needed.
Let me know if I should add it back for some sort of backwards compatibility (does mobile still use the old way?)

### DIFF
--- a/protocol/messenger_handler.go
+++ b/protocol/messenger_handler.go
@@ -69,43 +69,15 @@ func (m *Messenger) HandleMembershipUpdate(messageState *ReceivedMessageState, c
 		return ErrMessageNotAllowed
 	}
 
-	//if chat.InvitationAdmin exists means we are waiting for invitation request approvement, and in that case
-	//we need to create a new chat instance like we don't have a chat and just use a regular invitation flow
-	waitingForApproval := chat != nil && len(chat.InvitationAdmin) > 0
 	ourKey := contactIDFromPublicKey(&m.identity.PublicKey)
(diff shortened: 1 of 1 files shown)


jrainville:
Also, do let me know if I missed another place that could be the culprit for that bug

cammellos:
@jrainville mmh, not sure if that coudl cause the bug, I would want to be more confident (ideally a failing tests ) before merging and removing the feature

PR #3030, opened 2022-12-18
fix(communities): Ignore requests to join from banned user

borismelnik:
In fact of this [issue](https://github.com/status-im/status-desktop/issues/8761) we should check if user isn't in ban list

Important changes:


jrainville (approved):
Nice job. 
Do we also have the "delay" where you can only request every X minutes/hours? If not, maybe you can add it here or in another PR? :)

0x-r4bbit (approved):
This change LGTM.

I like the idea @jrainville mentioned, to throttle requests, but that needs a bit more work as Status would have to keep track of when access to some community was requested last time.

I personally wouldn't mind if we get that in a separate PR

borismelnik:
> Nice job.
> Do we also have the "delay" where you can only request every X minutes/hours? If not, maybe you can add it here or in another PR? :)

yes, I will fix some bug with requests to join and can add delay there :)

jrainville (approved):

jrainville:
You could up the VERSION too

borismelnik:
> You could up the VERSION too

fixed

jrainville (approved):

jrainville (approved):
Nice test

cammellos (approved):

Samyoul (approved):
image

Code to test ratio 👨‍🍳 🤌 💋

borismelnik:
@cammellos @Samyoul I found my test flaky. The problem here:
```
func (m *Manager) RequestToJoin(requester *ecdsa.PublicKey, request *requests.RequestToJoinCommunity) (*Community, *RequestToJoin, error) {
	community, err := m.persistence.GetByID(&m.identity.PublicKey, request.CommunityID)
	if err != nil {
		return nil, nil, err
	}

	// We don't allow requesting access if already joined
	if community.Joined() {
		return nil, nil, ErrAlreadyJoined
	}
```
I got `panic` on `community.Joined()` cos the community doesn't exists, but `GetByID` return no error. Is it okay? or its a good idea to add some checks for `community`?

0x-r4bbit:
> @cammellos @Samyoul I found my test flaky. The problem here:
> 
> ```
> func (m *Manager) RequestToJoin(requester *ecdsa.PublicKey, request *requests.RequestToJoinCommunity) (*Community, *RequestToJoin, error) {
> 	community, err := m.persistence.GetByID(&m.identity.PublicKey, request.CommunityID)
> 	if err != nil {
> 		return nil, nil, err
> 	}
> 
> 	// We don't allow requesting access if already joined
> 	if community.Joined() {
> 		return nil, nil, ErrAlreadyJoined
> 	}
> ```
> 
> I got `panic` on `community.Joined()` cos the community doesn't exists, but `GetByID` return no error. Is it okay? or its a good idea to add some checks for `community`?

@borismelnik I think we should return if we don't have a `community` but I wonder in what scenario this would happen.
Typically `RequestToJoin` results from a client that already has the `community` info. 

This means this error must have happened in a scenario where only the community ID was known without the description being retrieved yet.

If this happens regularly/is reproducible we should probably figure out why that is. 
Imagine users want to request access to a community but the client keeps failing because the community info isn't available. 
That sounds off to me.

borismelnik:
> > @cammellos @Samyoul I found my test flaky. The problem here:
> > ```
> > func (m *Manager) RequestToJoin(requester *ecdsa.PublicKey, request *requests.RequestToJoinCommunity) (*Community, *RequestToJoin, error) {
> > 	community, err := m.persistence.GetByID(&m.identity.PublicKey, request.CommunityID)
> > 	if err != nil {
> > 		return nil, nil, err
> > 	}
> > 
> > 	// We don't allow requesting access if already joined
> > 	if community.Joined() {
> > 		return nil, nil, ErrAlreadyJoined
> > 	}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I got `panic` on `community.Joined()` cos the community doesn't exists, but `GetByID` return no error. Is it okay? or its a good idea to add some checks for `community`?
> 
> @borismelnik I think we should return if we don't have a `community` but I wonder in what scenario this would happen. Typically `RequestToJoin` results from a client that already has the `community` info.
> 
> This means this error must have happened in a scenario where only the community ID was known without the description being retrieved yet.
> 
> If this happens regularly/is reproducible we should probably figure out why that is. Imagine users want to request access to a community but the client keeps failing because the community info isn't available. That sounds off to me.

Yeah I think its mostly in my tests cos I don't get the community on `Alice` side.

protocol/communities_messenger_test.go
```diff
@@ -3242,3 +3242,78 @@ func (s *MessengerCommunitiesSuite) TestCommunityTokensMetadata() {
...
+	s.Require().NoError(err)
+	s.Require().NotNil(response)
+	s.Require().Len(response.Communities(), 1)
+
+	community = response.Communities()[0]
+	s.Require().False(community.HasMember(&s.alice.identity.PublicKey))
+	s.Require().True(community.IsBanned(&s.alice.identity.PublicKey))
+
+	// Retrieve request to join
+	err = tt.RetryWithBackOff(func() error {
```
cammellos: This is the issue, you need to make sure the community is retrieved, so within the backoff function, you will have to check `response`, you should be use `WaitOnMessengerResponse` and have your condition inside the function

cammellos:
@borismelnik @0x-r4bbit is right, the issue is on how the test is written https://github.com/status-im/status-go/pull/3030/files#r1182371579

protocol/communities_messenger_test.go
borismelnik: oh okay

Samyoul:
@borismelnik, in this case the user is not aware that the community exists, which would not be a likely scenario. You need to update your test so that the users get a details of the community before attempting to join.

---

Edit: Oh, 😆 looks like I was last to reply on this.

borismelnik:
@cammellos @Samyoul guys pls review my test again I hope I got it right :)

protocol/communities_messenger_test.go
```diff
@@ -3242,3 +3242,81 @@ func (s *MessengerCommunitiesSuite) TestCommunityTokensMetadata() {
...
+		Name:        "status",
+		Color:       "#ffffff",
+		Description: "status community description",
+	}
+
+	// Create an community chat
+	response, err := s.bob.CreateCommunity(description, true)
+	s.Require().NoError(err)
+	s.Require().NotNil(response)
+	s.Require().Len(response.Communities(), 1)
```
Samyoul, line 3258: If you want a thorough test I'd recommend not relying of the response, as this doesn't necessarily prove that the community was created, only that `Messenger` acknowledges that it processed a community.

The more robust thing to do would be to call a function like `m.communitiesManager.All()` to check that the community is in the db.

From that point you can check the number of communities and inspect the desired community like normal.

cammellos: (both should be tested, one is testing the RPC API, the other one as @Samyoul correctly points out, is the side effect)

Samyoul: @borismelnik another thing that helps me when writing these tests is to write comments first then fill in the functionality.

```go
// declare community
// store and publish community
// check community was stored
// check community has the correct details
// check other users received community
// check etc..
```

PR #3035, opened 2022-12-21
feat: make replies act as mentions

jrainville:
Fixes this Desktop issue https://github.com/status-im/status-desktop/issues/8324

When there is a reply to one of your messages, set `Mentioned` to `true` so that a mention marker appears on the channel and the section.

My initial idea was to put that logic in `PrepareContent`, but the message sadly doesn't have access to `QuotedMessage` at that point, so I just check afterwards. In a way, it's lighter, because we only check once and if `Mentioned` is  `false` at that point, since if you were already mentioned by ID, there is no need to check the reply; you can't be double mentioned.

I tested on Desktop all wanted scenarios
- on signal
- on sent during the app closed and then fetched
- on signal, not marking as seen, closing and reopening 

Let me know if there are other places we should do that check, but it seems like the only needed place.

~~I'll add a test in a few minutes~~ Done


jrainville:
I added a reply test file. Let me know if I need to add it to a list or it gets automatically ran by the CI wince it ends by `test`

protocol/messenger_handler.go
```diff
@@ -1523,6 +1523,19 @@ func (m *Messenger) HandleChatMessage(state *ReceivedMessageState) error {
 	if err != nil {
 		return fmt.Errorf("failed to prepare message content: %v", err)
 	}
+
+	// If the user was not mentionned, we still check if there is a reply as they count as mentions as well
+	if !receivedMessage.Mentioned && receivedMessage.ResponseTo != "" {
+		repliedTo, err := m.persistence.MessageByID(receivedMessage.ResponseTo)
+		if err != nil {
+			logger.Warn("failed to get quoted message", zap.Error(err))
```
0x-r4bbit: I think we can fail fast here and return `err`

0x-r4bbit (approved):
This LGTM and thanks for providing tests.
Only thing I'd like to see is to fail fast

protocol/messenger_handler.go
jrainville: Yeah, I wasn't sure if I should. That error only causes the message to not have the `Mentioned` set correctly to `true`, so I didn't want to kill the whole message handling for that. Let me know what you think

0x-r4bbit: Figured that! I'm good with this as well

jrainville:
So, I just dug around some more and I found out that the QuotedMessage does get populated later in the cycle of `handleRetrievedMessages`, but `updateUnviewedCounts` is only done in `HandleChatMessage`, so is it ok if we sometimes do `MessageByID` twice for the same QuotedMessage or should I try to switch around some order of calls to only call it once?

protocol/messenger_handler.go
```diff
@@ -1523,6 +1523,19 @@ func (m *Messenger) HandleChatMessage(state *ReceivedMessageState) error {
...
+		} else {
+			if repliedTo.From == common.PubkeyToHex(&m.identity.PublicKey) {
+				receivedMessage.Mentioned = true
```
igor-sirotin: Hmmm. I don't think https://github.com/status-im/status-desktop/issues/8324 was about this behavior. 

Having this done, all replies will be highlighted in UI as if there is a `@mention` for current user inside. But that is not what we want. All we wanted is to have those messages counted in the "counter badge" in the channels list.

igor-sirotin (changes requested):
The code's good for sure.
But I have doubts about the logic itself.

protocol/messenger_handler.go
jrainville: The counter uses the `unviewedMentionsCount`, which comes directly from status-go's Chat object and uses `Mentioned` to count the number of mentions. So setting a reply as `Mentioned=true` makes the counter go up by 1.

> Having this done, all replies will be highlighted in UI as if there is a @mention for current user inside

You might be right. The AC might also think that a message is both a mention and a reply at the same time in that case. I'll test it out.

jrainville: Ok, you were right, it makes the reply count as a mention in the UI, so it's not exactly what we want. Though my change made changing the counter work very easily without changes on the front-end (Nim or QML), it broke some small parts.

I'll update this PR then to add a `Replied` property and update the `unviewedMentionsCount` to take those into account as well.

Nice find @igor-sirotin

jrainville: @igor-sirotin I updated. There is now a `Replied` property on messages. We update the count the same as with `Mentioned`, but it doesn't have `Mentioned=true`, this way, the Activity center and message formatting is as before. Only the badge changed.

jrainville:
@richard-ramos @cammellos can you give a review when you have time, thank you :)

protocol/messenger_handler.go
```diff
@@ -1517,6 +1517,19 @@ func (m *Messenger) HandleChatMessage(state *ReceivedMessageState) error {
 	if err != nil {
 		return fmt.Errorf("failed to prepare message content: %v", err)
 	}
+
+	// If the user was not mentionned, we still check if there is a reply as they count as mentions as well
```
cammellos: typo, `mentioned`

protocol/messenger_handler.go
cammellos: we should fail fast on any error that's not expected.
If `MessageByID` returns some known errors (i.e `sql.ErrNoRows` because for example the message are arriving out of order), we should handle it explicitly, but if the error is unexpected, better to return the error and stop processing.

cammellos (commented):
Left a couple of comments, it looks good for the most part, one thing, we changed many queries, could you run an `EXPLAIN QUERY PLAN` on those that have changed and check with the indices that are either still covered, or that those are up to date?

jrainville:
It's the first time I use `EXPLAIN QUERY PLAN`, so I'm not sure if I did it right, but here is the result for the queries that changed (that are not just the basic `SELECT`s)

Query in `deleteMessagesByChatIDAndClockValueLessThanOrEqual`
```sql
sqlite> EXPLAIN QUERY PLAN UPDATE chats SET unviewed_message_count = (SELECT COUNT(1) FROM user_messages WHERE local_chat_id = ? AND seen = 0), unviewed_mentions_count = (SELECT COUNT(1) FROM user_messages WHERE local_chat_id = ? AND seen = 0 AND (mentioned OR replied)), highlight = 0 WHERE id = ?;
0|0|0|SEARCH TABLE chats USING INDEX sqlite_autoindex_chats_1 (id=?)
0|0|0|EXECUTE SCALAR SUBQUERY 0
0|0|0|SEARCH TABLE user_messages USING COVERING INDEX seen_local_chat_id_idx (local_chat_id=? AND seen=?)
0|0|0|EXECUTE SCALAR SUBQUERY 1
1|0|0|SEARCH TABLE user_messages USING INDEX local_chat_id_seen_mentioned_clock_value_idx (local_chat_id=? AND seen=?)
```

---
Queries for `MarkAllRead`
```sql
sqlite> EXPLAIN QUERY PLAN UPDATE user_messages SET seen = 1 WHERE local_chat_id = ? AND seen = 0 AND clock_value <= ? AND not(mentioned) AND not(replied);
0|0|0|SEARCH TABLE user_messages USING INDEX local_chat_id_seen_mentioned_clock_value_idx (local_chat_id=? AND seen=?)

sqlite> EXPLAIN QUERY PLAN UPDATE user_messages SET seen = 1 WHERE local_chat_id = ? AND seen = 0 AND clock_value <= ? AND (mentioned OR replied);
0|0|0|SEARCH TABLE user_messages USING INDEX local_chat_id_seen_mentioned_clock_value_idx (local_chat_id=? AND seen=?)

sqlite> EXPLAIN QUERY PLAN UPDATE chats SET unviewed_message_count = (SELECT COUNT(1) FROM user_messages WHERE local_chat_id = ? AND seen = 0), unviewed_mentions_count = (SELECT COUNT(1) FROM user_messages WHERE local_chat_id = ? AND seen = 0 AND (mentioned or replied)), highlight = 0 WHERE id = ?;
0|0|0|SEARCH TABLE chats USING INDEX sqlite_autoindex_chats_1 (id=?)
0|0|0|EXECUTE SCALAR SUBQUERY 0
0|0|0|SEARCH TABLE user_messages USING COVERING INDEX seen_local_chat_id_idx (local_chat_id=? AND seen=?)
0|0|0|EXECUTE SCALAR SUBQUERY 1
1|0|0|SEARCH TABLE user_messages USING INDEX local_chat_id_seen_mentioned_clock_value_idx (local_chat_id=? AND seen=?)
```

---
Queries for `MarkMessagesSeen`
```sql
sqlite> EXPLAIN QUERY PLAN UPDATE user_messages SET seen = 1 WHERE NOT(seen) AND (mentioned OR replied) AND id IN ("0x");
0|0|0|SEARCH TABLE user_messages USING INDEX sqlite_autoindex_user_messages_1 (id=?)

sqlite> EXPLAIN QUERY PLAN UPDATE user_messages SET seen = 1 WHERE NOT(seen) AND NOT(mentioned) AND NOT(replied) AND id IN ("0x");
0|0|0|SEARCH TABLE user_messages USING INDEX sqlite_autoindex_user_messages_1 (id=?)

sqlite> EXPLAIN QUERY PLAN UPDATE chats SET unviewed_message_count = (SELECT COUNT(1) FROM user_messages WHERE local_chat_id = ? AND seen = 0), unviewed_mentions_count = (SELECT COUNT(1) FROM user_messages WHERE local_chat_id = ? AND seen = 0 AND (mentioned OR replied)), highlight = 0 WHERE id = ?;
0|0|0|SEARCH TABLE chats USING INDEX sqlite_autoindex_chats_1 (id=?)
0|0|0|EXECUTE SCALAR SUBQUERY 0
0|0|0|SEARCH TABLE user_messages USING COVERING INDEX seen_local_chat_id_idx (local_chat_id=? AND seen=?)
0|0|0|EXECUTE SCALAR SUBQUERY 1
1|0|0|SEARCH TABLE user_messages USING INDEX local_chat_id_seen_mentioned_clock_value_idx (local_chat_id=? AND seen=?)
```

--- 
Query for `BlockContact`
```sql
sqlite> EXPLAIN QUERY PLAN UPDATE chats SET unviewed_message_count = (SELECT COUNT(1) FROM user_messages WHERE seen = 0 AND local_chat_id = chats.id), unviewed_mentions_count = (SELECT COUNT(1) FROM user_messages WHERE seen = 0 AND local_chat_id = chats.id AND (mentioned OR replied));
0|0|0|SCAN TABLE chats
0|0|0|EXECUTE CORRELATED SCALAR SUBQUERY 0
0|0|0|SEARCH TABLE user_messages USING COVERING INDEX seen_local_chat_id_idx (local_chat_id=? AND seen=?)
0|0|0|EXECUTE CORRELATED SCALAR SUBQUERY 1
1|0|0|SEARCH TABLE user_messages USING INDEX local_chat_id_seen_mentioned_clock_value_idx (local_chat_id=? AND seen=?)
```

@cammellos let me know if what I did is ok and if I need to do anything to improve indexes

igor-sirotin (approved):
Nice, thanks for the update 🙂

jrainville:
@0x-r4bbit @cammellos @richard-ramos can you please (re)-review when you have the time :slightly_smiling_face:

cammellos (approved):
Thanks for the `EXPLAIN QUERY`, looks most of the stuff is covered.

Looks good to me!

richard-ramos (approved):
Very cool!
If possible, modify the PR to not include migrations / protobuffer files that shouldn't be affected by this PR. `protocol/migrations/migrations.go` is the only one that should be included.

richard-ramos (approved):

PR #3037, opened 2022-12-27
fix TestGroupChatMembersRemoval sometimes failed

### LINKED ISSUE #3036: TestGroupChatMembersRemoval fails sometimes
# Problem

TestGroupChatMembersRemoval sometimes succeeds and sometimes fails, u can run it in your local machine many times to find out, we need this unit test pass so that our PR for status-go won't get blocked `sometimes`

qfrank:
Closes #3036

### DIFF
--- a/protocol/messenger_handler.go
+++ b/protocol/messenger_handler.go
@@ -121,8 +121,10 @@ func (m *Messenger) HandleMembershipUpdate(messageState *ReceivedMessageState, c
 		wasUserAdded = true
 		newChat := CreateGroupChat(messageState.Timesource)
 		// We set group chat inactive and create a notification instead
-		// unless is coming from us or a contact or were waiting for approval
-		newChat.Active = isActive
+		// unless is coming from us or a contact or were waiting for approval.
+		// Also, as message MEMBER_JOINED may come from member(not creator, not our contact)
+		// reach earlier than CHAT_CREATED from creator, we need check if creator is our contact
+		newChat.Active = isActive || m.checkIfCreatorIsOurContact(group)
 		newChat.ReceivedInvitationAdmin = senderID
 		chat = &newChat
 
@@ -211,6 +213,16 @@ func (m *Messenger) HandleMembershipUpdate(messageState *ReceivedMessageState, c
 	}
 
 	return nil
+}
+
+func (m *Messenger) checkIfCreatorIsOurContact(group *v1protocol.Group) bool {
+	creator, err := group.Creator()
+	if err == nil {
+		contact, _ := m.allContacts.Load(creator)
+		return contact != nil && contact.Added && contact.HasAddedUs
+	}
+	m.logger.Warn("failed to get creator from group", zap.String("group name", group.Name()), zap.String("group chat id", group.ChatID()), zap.Error(err))
+	return false
 }
 
 func (m *Messenger) createMessageNotification(chat *Chat, messageState *ReceivedMessageState) {
--- a/protocol/v1/membership_update_message.go
+++ b/protocol/v1/membership_update_message.go
@@ -479,7 +479,7 @@ func (g Group) LastClockValue() uint64 {
 	return g.events[len(g.events)-1].ClockValue
 }
 
-func (g Group) creator() (string, error) {
+func (g Group) Creator() (string, error) {
 	if len(g.events) == 0 {
 		return "", errors.New("no events in the group")
 	}
@@ -491,7 +491,7 @@ func (g Group) creator() (string, error) {
 }
 
 func (g Group) validateChatID(chatID string) bool {
-	creator, err := g.creator()
+	creator, err := g.Creator()
 	if err != nil || creator == "" {
 		return false
 	}
--- a/protocol/v1/membership_update_message_test.go
+++ b/protocol/v1/membership_update_message_test.go
@@ -67,7 +67,7 @@ func TestGroupCreator(t *testing.T) {
 	require.NoError(t, err)
 	g, err := NewGroupWithCreator("abc", "#fa6565", 20, key)
 	require.NoError(t, err)
-	creator, err := g.creator()
+	creator, err := g.Creator()
 	require.NoError(t, err)
 	require.Equal(t, publicKeyToString(&key.PublicKey), creator)
 }


protocol/messenger_group_chat_test.go
```diff
@@ -264,6 +264,12 @@ func (s *MessengerGroupChatSuite) TestGroupChatMembersRemoval() {
 	s.makeMutualContacts(admin, memberA)
 	s.makeMutualContacts(admin, memberB)
 	s.makeMutualContacts(admin, memberC)
+	// if we expect r.Chats()[0].Active == true when #verifyGroupChatCreated
+	// then we need make admin/memberA/memberB/memberC as mutual contacts, not only just with admin
+	// since memberB may receive protobuf.MembershipUpdateMessage from memberA rather than admin
+	s.makeMutualContacts(memberA, memberB)
+	s.makeMutualContacts(memberA, memberC)
+	s.makeMutualContacts(memberB, memberC)
```
OmarBasem: I do not understand why they should be mutual. We can have some user create a group, then add 3 other users from his contacts who do not know each other to the group, right?

qfrank: > I do not understand why they should be mutual. We can have some user create a group, then add 3 other users from his contacts who do not know each other to the group, right?

right, but then we may got inactive chat, we can c the logical [here](https://github.com/status-im/status-go/blob/5005b7b3cf45fed801257060fd68e19c4422e317/protocol/messenger_handler.go#L76) , messageState.CurrentMessageState.Contact.Added will be false, contact means sender(may not be the group creator) who send protobuf.MembershipUpdateMessage

OmarBasem: Alright, LGTM but it's better if someone else looks at this also. So what would/should actually happen if someone creates a group then adds some users who are not mutual with each other, then leaves the group?

OmarBasem (approved):

protocol/messenger_group_chat_test.go
qfrank: I'm trying to find a better solution to fix this test, since we have 
```golang
s.makeMutualContacts(admin, memberA)
s.makeMutualContacts(admin, memberB)
s.makeMutualContacts(admin, memberC)
```
and admin sent  protobuf.MembershipUpdateMessage to each member already.
We can't ensure the order of message received, but we should figure out a better way to handle this situation

osmaczko: Thanks @qfrank for approaching this problem.

> We can't ensure the order of message received, but we should figure out a better way to handle this situation

I agree. The solution of making all members mutual is just masking the issue. Also, it makes the intention of the test unclear, as mentioned [here](https://github.com/status-im/status-go/pull/3037#discussion_r1057481203).

The problem should be solved in `verifyGroupChatCreated`. It should handle "out of order messages issue". Theoretically, it should work fine just now, because it uses `ExponentialBackOff` behind the scenes, meaning it will retry the operation until it does not return an error. In other words, it will wait for the response where the given condition is true (and it should be true when receiver eventually gets the message from the admin).
```
	response, err := WaitOnMessengerResponse(
		member,
			return len(r.Chats()) == 1 && r.Chats()[0].Active == expectedChatActive
		},
		"chat invitation not received",
	)
```

The root cause is either:
a) `ExponentialBackOff` failure, we can verify that by inspecting if the receiver got the message from the admin
b) bug in group chat logic, when we first get the `MembershipUpdateMessage` from member rather than admin, resulting in chat.Active == false

jrainville: I discussed this issue with Andrea before, because I discovered the same thing when investigating this desktop issue https://github.com/status-im/status-desktop/issues/8773 (you might be able to see the discussion in the status-go channel, it's from Decmber 16th)

Basically, this is the way to fix the test for now. 
Like you rightly pointed out @qfrank , if the membership update comes from a non-mutual contact, that channel will be made inactive.
That is because we wanted to negate the spam. Back in the day, group chats could be created by anyone and with anyone, so you'd get spammed with group chats and you had to manually reject them. Now, you don't receive them from non-mutual contacts to negate that. 
It does create the annoying scenario above when if the person that adds you, or like in the issue I linked, if you re-import your account and the signal comes in from a non-mutual contact (from a normal chat message for example), then the chat will be marked as inactive.

The solution we'll do in desktop is show that invite in the AC only. I think mobile does something similar already. We're just waiting for designs.

I hope it was clear and makes sense.
One way to improve this test further would be to voluntarily test the negative case where if the signal comes from a non-mutual contact, then we assert that `Active` will be `false`.

jrainville (approved):
Bonus points if you create the scenario where receiving the MembershipUpdate from a non-mutual contact sets the chat as `Active=false`

qfrank:
> Bonus points if you create the scenario where receiving the MembershipUpdate from a non-mutual contact sets the chat as `Active=false`

Emm..Looks like there is already one TestGroupChatCreation @jrainville

qfrank:
Thank you for your enthusiastic comments! I've updated my fix, pls take a further look, thanks! @jrainville @OmarBasem @osmaczko

osmaczko (approved):
Very nice :+1:

protocol/messenger_handler.go
```diff
@@ -213,6 +215,15 @@ func (m *Messenger) HandleMembershipUpdate(messageState *ReceivedMessageState, c
 	return nil
 }
 
+func (m *Messenger) checkIfCreatorIsOurContact(group *v1protocol.Group) bool {
+	creator, err := group.Creator()
+	if err == nil {
+		contact, _ := m.allContacts.Load(creator)
```
jrainville, line 221: You need to check the error here as well in case

jrainville (commented):
That looks like a good solution. Let's just check with @cammellos if it's legit, he might see a flaw that I don't see.
Otherwise, it seems like it might solve my problem as well and it makes sense in my books.

protocol/messenger_handler.go
qfrank: just return false when get err from group.Creator() was what i expected, i've added a logger for this error, passing this error up to HandleMembershipUpdate may will make things complicated, WDYT of the latest commit? @jrainville

jrainville: Ah, I see that you check if `contact` is `nil`. That's fine then.

jrainville (approved):

protocol/messenger_handler.go
```diff
@@ -213,6 +215,16 @@ func (m *Messenger) HandleMembershipUpdate(messageState *ReceivedMessageState, c
...
+		return contact != nil && contact.Added && contact.HasAddedUs
```
cammellos, line 222: I think you should be able to use `ContactRequestStateMutual` in `ContactRequestState`, it should be the same, and we will be moving away from `Added` `HasAddedUs`

cammellos (approved):
nice find!

protocol/messenger_handler.go
qfrank: well, i think we can't use ContactRequestState here, i've tested with:
```
if contact.Added && contact.HasAddedUs && contact.ContactRequestState != ContactRequestStateMutual {
	println("bomb!")
}
```
we will c "bomb!" @cammellos

qfrank: will fix the inconsistent in a separate PR

PR #3041, opened 2022-12-29
feat/currency rounding

dlipicar:
Needed for issue #8640

- Added library iso4217, used to get standard information about fiat currencies (mainly the number of decimal places to display).
- Added api functions for iso4217 data
- Added "Peg" information to tokenStore and token structure. Tells if a certain crypto currency is pegged to a fiat currency.
- Use numbers instead of display strings for token market values. Reorganized data in token struct.
- Adapted API to fetch prices in multiple currencies at the same time.
- Added currency "usd" every time we fetch prices.


services/wallet/router.go
```diff
@@ -399,10 +399,14 @@ func (r *Router) suggestedRoutes(
 		return nil, err
 	}
 
-	prices, err := fetchCryptoComparePrices([]string{"ETH", tokenSymbol}, "USD")
+	pricesMap, err := fetchCryptoComparePrices([]string{"ETH", tokenSymbol}, []string{"USD"})
```
saledjenic, line 402: @dlipicar don't we need to calculate a suggested root in the user's selected currency?

saledjenic (approved):

services/wallet/router.go
dlipicar: Hmm. I didn't really change anything here, just changed the string argument "USD" to a vector containing that single argument, because the method signature changed.
What you say makes sense to me. This might be a question for @Khushboo-dev-cpp or @alaibe

saledjenic: @dlipicar I know that approach was already here, but since I see it for the first time I just wanted to clarify that.

services/wallet/reader.go
```diff
@@ -32,29 +36,34 @@ type Reader struct {
...
+	Name                    string                       `json:"name"`
+	Symbol                  string                       `json:"symbol"`
+	Color                   string                       `json:"color"`
+	Decimals                uint                         `json:"decimals"`
+	BalancesPerChain        map[uint64]ChainBalance      `json:"balancesPerChain"`
+	Description             string                       `json:"description"`
+	AssetWebsiteURL         string                       `json:"assetWebsiteUrl"`
+	BuiltOn                 string                       `json:"builtOn"`
+	MarketValuesPerCurrency map[string]TokenMarketValues `json:"marketValuesPerCurrency"`
+	Peg                     string                       `json:"peg"`
```
stefandunca: This name sounds like a fit for a boolean value. Won't PegSymbol make it more straightforward?

services/wallet/reader.go
```diff
@@ -32,29 +36,34 @@ type Reader struct {
 	cancel       context.CancelFunc
 }
 
+type TokenMarketValues struct {
```
stefandunca, line 39: Nice extracting this.

services/wallet/api.go
```diff
@@ -312,14 +315,14 @@ func (api *API) GetEthereumChains(ctx context.Context, onlyEnabled bool) ([]*par
 	return api.s.rpcClient.NetworkManager.Get(onlyEnabled)
 }
 
-func (api *API) FetchPrices(ctx context.Context, symbols []string, currency string) (map[string]float64, error) {
+func (api *API) FetchPrices(ctx context.Context, symbols []string, currencies []string) (map[string]map[string]float64, error) {
```
stefandunca, line 318: I am trying to understand why this parameter changed from a single currency to multiple currencies. In which case do we need results in various currencies?

stefandunca (approved):
I left some questions hoping for clarifications.

services/wallet/reader.go
dlipicar: Did some extra reading to see what names are used in the literature.
The "boolean" state is commonly referred to as "pegged". (USDT is pegged to the US Dollar)
"Peg" seems to commonly refer to a value + a currency. (USDT has a peg of 1 USD).
I don't think there is one, but I guess there could be a peg with a value different than 1 (pegged to 100 USD?), so I'm guessing your suggestion makes sense and we should call this "PegSymbol". Will do the change.

services/wallet/api.go
dlipicar: - We always need the price in USD to calculate how many decimals we display for a certain token, regardless of the currency chosen by the user in the settings. For this reason, the underlying function call (`fetchCryptoComparePrices`) now takes multiple currencies, so all the prices can be retrieved in a single CryptoCompare request.
- I had the option of keeping `FetchPrices` with a single currency (and pass a vector with that single currency to `fetchCryptoComparePrices`) or extending the flexibility of multiple currencies to it as well. I went with the second option since it could eventually help condense several API calls into one.
- John mentioned in a call that in the future the user might want to select multiple user currencies instead of a single one, and they should be able to quickly switch between them (for example, in high-inflation countries people normally "think" in their local currency and some more stable one -like USD- at the same time). In that case, we'll want to always fetch values for all chosen currencies and not have to re-fetch (with its corresponding delay) when the user decides to switch between them. I did this change to "future-proof" the Go API a little bit.

Does this clarify things a little bit? Does this decision make sense now?

services/wallet/reader.go
dlipicar: Fixed

services/wallet/router.go
dlipicar: Discussed with Anthony and Khushboo, this will be changed to use the user's currency

services/wallet/api.go
stefandunca: Thanks, @dlipicar, for the excellent clarifications, and I support your decision.

PR #3042, opened 2022-12-29
Convert Keys from Desktop to Mobile format

siddarthkay:
In desktop when a community is created the share url looks like : 

`https://join.status.im/c/zQ3shTAten2v9CwyQD1Kc7VXAqNPDcHZAMsfbLHCZEx6nFqk9`  where 
`zQ3shTAten2v9CwyQD1Kc7VXAqNPDcHZAMsfbLHCZEx6nFqk9`
is the serialised key which desktop uses.

In mobile when a community is created the share url looks like : 
`https://join.status.im/c/0x025596a7ff87da36860a84b0908191ce60a504afc94aac93c1abd774f182967ce6` where
`0x025596a7ff87da36860a84b0908191ce60a504afc94aac93c1abd774f182967ce6`
is the non serialised key (but compressed) key which mobile uses.

The goal of this PR is to give mobile the ability to go from 
`zQ3shTAten2v9CwyQD1Kc7VXAqNPDcHZAMsfbLHCZEx6nFqk9` to `0x025596a7ff87da36860a84b0908191ce60a504afc94aac93c1abd774f182967ce6`

### DIFF
--- a/mobile/response_test.go
+++ b/mobile/response_test.go
@@ -25,3 +25,10 @@ func TestPrepareJSONResponseErrorWithError(t *testing.T) {
 	data := prepareJSONResponse("0x123", errors.New("some error"))
 	require.Contains(t, data, `{"error":{"message":"some error"}}`)
 }
+
+func TestDeserializeAndCompressKeyApi(t *testing.T) {
+	desktopKey := "zQ3shTAten2v9CwyQD1Kc7VXAqNPDcHZAMsfbLHCZEx6nFqk9"
+	mobileKeyExpected := "0x025596a7ff87da36860a84b0908191ce60a504afc94aac93c1abd774f182967ce6"
+	mobileKeyConverted := DeserializeAndCompressKey(desktopKey)
+	require.Equal(t, mobileKeyConverted, mobileKeyExpected)
+}
--- a/mobile/status.go
+++ b/mobile/status.go
@@ -1088,3 +1088,9 @@ func ToChecksumAddress(address string) string {
 	}
 	return address
 }
+
+func DeserializeAndCompressKey(DesktopKey string) string {
+	deserialisedKey := MultiformatDeserializePublicKey(DesktopKey, "f")
+	sanitisedKey := "0x" + deserialisedKey[5:]
+	return CompressPublicKey(sanitisedKey)
+}


mobile/response_test.go
```diff
@@ -25,3 +25,15 @@ func TestPrepareJSONResponseErrorWithError(t *testing.T) {
 	data := prepareJSONResponse("0x123", errors.New("some error"))
 	require.Contains(t, data, `{"error":{"message":"some error"}}`)
 }
+
+func TestCompressAnyKey(t *testing.T) {
+	key := CompressPublicKey("0x045596a7ff87da36860a84b0908191ce60a504afc94aac93c1abd774f182967ce694f1bf2d8773cd59f4dd0863e951f9b7f7351c5516291a0fceb73f8c392a0e88")
```
qfrank: Hi @siddarthkay , thanks for inviting! Can we generate some random community ids once running test rather than fixed value here so that makes the test looks more convincing?

qfrank (changes requested):

mobile/status.go
```diff
@@ -1088,3 +1088,9 @@ func ToChecksumAddress(address string) string {
 	}
 	return address
 }
+
+func ConvertDesktopKeyToMobileKey(DesktopKey string) string {
```
briansztamfater: I am not sure if we should stick with the concepts of "desktop key" or "mobile key". Are actually "desktop keys" and "mobile keys" part of the Status domain?
In my understanding, both clients are currently using different key formats but in the end all clients should use the same one, so if this is true, the function name could create confusion. Maybe using a more generic naming as `DeserializeAndCompressKey` is more accurate, which in the end is what this function is doing, so is self-descriptive. What do you think?

siddarthkay: Hi @briansztamfater sounds good I will rename this to `DeserializeAndCompressKey`

mobile/response_test.go
siddarthkay: Hi @qfrank : Yes sounds good!

siddarthkay: @qfrank : `TestComressAnyKey` was actually not needed and put there so that I could see whats the output..

qfrank: emm.. i still think it's needed as it will also let ppl more easier to know how to transfer from desktop key into mobile key, does this make sense? @siddarthkay

siddarthkay: > emm.. i still think it's needed as it will also let ppl more easier to know how to transfer from desktop key into mobile key, does this make sense? @siddarthkay

Oh for that I have added `TestDeserializeAndCompressKeyApi`

qfrank (approved):

PR #3044, opened 2023-01-03
fix: set nameserver via config

richard-ramos:
Having `1.1.1.1` hardcoded ended up generating this error on @caybro machine:
```
t=2023-01-03T12:53:40+0100 lvl=warn msg="dns discovery error " error="lookup prod.nodes.status.im on 127.0.0.53:53: cannot marshal DNS message"
```
Which seems to indicate some compatibility failure between the machine and the nameserver.

Since I can't easily identify whether status-go is running on mobile or in desktop, I extracted the hardcoded nameserver, and added a `Nameserver` attribute to `WakuV2Config`, which should be provided in mobile to temporarily set the nameserver to `1.1.1.1`.

cc: @jrainville @iurimatias as it might be a good idea to include this in 0.9.0


caybro (approved):
An obvious +1 from me as it fixes the problem here.

It seems out the reason for this is that newer versions of go (`go version go1.19.3 linux/amd64` on my machine) have strict(er) rules for validating DNS replies, see https://go-review.googlesource.com/c/net/+/100055

qfrank (approved):
LGTM 🫶

PR #3049, opened 2023-01-05
Persist switcher cards

Parveshdhull:
status-react PR: https://github.com/status-im/status-mobile/pull/14701

status: ready


1672910445_switcher_card.up.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE switcher_cards (
+  id TEXT PRIMARY KEY ON CONFLICT REPLACE,
+  type INT NOT NULL DEFAULT 0,
+  clock INT NOT NULL,
+  extra TEXT DEFAULT ""
```
Parveshdhull: In the case of community cards, the id will correspond to the community id even when any channel is opened (only one card per community). So we need an extra field to store channel-id to open the corresponding channel on card press.
And similarly, for wallets, with only one card & one id. And `extra` field will be used to store which screen in that account.

Parveshdhull: Maybe extra field is not self explanatory, open for suggestions

protocol/persistence.go
```diff
@@ -1134,6 +1134,58 @@ func (db sqlitePersistence) StatusUpdates() (statusUpdates []UserStatus, err err
 	return
 }
 
+func (db sqlitePersistence) DeleteSwitcherCard(id string) error {
+	_, err := db.db.Exec("DELETE from switcher_cards WHERE id = ?", id)
+	return err
+}
+
+func (db sqlitePersistence) AddSwitcherCard(switcherCard SwitcherCard) error {
```
cammellos: Because of `ON CONFLICT REPLACE`, I would call this endpoint `Upsert` (and the APIs etc), as it better reflects what it does (`insert or update` essentially)

cammellos (approved):

1672910445_switcher_card.up.sql
ilmotta: Maybe `metadata` could be a better name? It's just as generic, but at least to me it implies that its *text* could be a serialized data structure.

Edit: Added a related comment in the Status Mobile PR https://github.com/status-im/status-mobile/pull/14701/files#r1062327608

Parveshdhull: We are only going to store the id of a screen which card going to open in the extra field. So to avoid confusion and keeping fields flat I am thinking of renaming `id` to `cardId` and `extra` to `screenId`. wdyt?

ilmotta: I just commented on renaming to `metadata` because I got the impression we don't know what we're going to store in the future, but I prefer your suggestion a lot more :)

Parveshdhull: thank you

PR #3050, opened 2023-01-05
feat: nonce management in multi tx

alaibe:
Extract the nonce manager so that it is available for transaction being made outside the transaction via contract call

Create an approval transaction before sending bridge transaction

### DIFF
--- a/transactions/nonce.go
+++ b/transactions/nonce.go
@@ -1,0 +1,67 @@
+package transactions
+
+import (
+	"context"
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/status-im/status-go/eth-node/types"
+)
+
+type Nonce struct {
+	addrLock   *AddrLocker
+	localNonce map[uint64]*sync.Map
+}
+
+func NewNonce() *Nonce {
+	return &Nonce{
+		addrLock:   &AddrLocker{},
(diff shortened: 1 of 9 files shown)


richard-ramos (approved):

transactions/nonce.go
```diff
@@ -0,0 +1,71 @@
...
+		localNonce: make(map[uint64]*sync.Map),
+	}
+}
+
+func (n *Nonce) Next(rpcWrapper *rpcWrapper, from types.Address) (uint64, func(inc bool, nonce uint64)) {
+	n.addrLock.LockAddr(from)
+	current := n.GetCurrent(rpcWrapper, from)
+	unlock := func(inc bool, nonce uint64) {
+		if inc {
+			fmt.Println(n.localNonce[rpcWrapper.chainID])
```
stefandunca: Leftovers?

transactions/nonce.go
```diff
@@ -0,0 +1,71 @@
...
+			if _, ok := n.localNonce[rpcWrapper.chainID]; !ok {
+				n.localNonce[rpcWrapper.chainID] = &sync.Map{}
+			}
+
+			fmt.Println(n.localNonce[rpcWrapper.chainID])
```
stefandunca: Same ^?

stefandunca (approved):

PR #3052, opened 2023-01-05
chore: resend contact requests to pending contact during backup

0x-r4bbit:
This is done as a result of https://github.com/status-im/status-desktop/issues/8702

Essentially, there's cases where outgoing contact requests are backed-up but not their responses.
Then, if the account in question is re-imported, the backup will recover contacts as still outgoing pending requests (even though, requests might have been accepted already).

So what we do in such cases (where `added: true` and `hasAddedUs: false`), is resending contact requests to those accounts and hope that they'll send an ACK back.

### DIFF
--- a/protocol/messenger_handler.go
+++ b/protocol/messenger_handler.go
@@ -517,6 +517,22 @@ func (m *Messenger) HandleSyncInstallationContact(state *ReceivedMessageState, m
 		state.AllChats.Store(chat.ID, chat)
 	}
 
+	if contact.Added && !contact.HasAddedUs {
+		// This means we have an outgoing contact request that is pending.
+		// There might have been a response to this request already
+		// which wasn't backed up, so we're resending the contact request
+		// and hope for a response.
+
+		request := &requests.SendContactRequest{
+			ID:      types.Hex2Bytes(contact.ID),
+			Message: "Resending pending contact request",
+		}
+		_, err := m.SendContactRequest(context.Background(), request)
+		if err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


protocol/messenger_handler.go
```diff
@@ -1575,16 +1591,33 @@ func (m *Messenger) HandleChatMessage(state *ReceivedMessageState) error {
...
+			ensName, err := m.settings.ENSName()
+			if err != nil {
+				return err
+			}
 
+			displayName, err := m.settings.DisplayName()
+			if err != nil {
+				return err
+			}
+			_, err = m.sendContactUpdate(context.Background(), currentContact.ID, displayName, ensName, "")
```
0x-r4bbit, line 1606: @cammellos in https://github.com/status-im/status-desktop/issues/8702#issuecomment-1369793368 you've mentioned that contacts should send an ACK response which would add one byte.

I"m not sure where those byte would be added, as we would just send a contact update (which is interpreted as the request being accepted).

If this is not at all what you were looking for, please let me know. I just took a first stab on this now.

protocol/messenger_handler.go
```diff
@@ -517,6 +517,22 @@ func (m *Messenger) HandleSyncInstallationContact(state *ReceivedMessageState, m
...
 
+	if contact.Added && !contact.HasAddedUs {
+		// This means we have an outgoing contact request that is pending.
+		// There might have been a response to this request already
+		// which wasn't backed up, so we're resending the contact request
+		// and hope for a response.
+
+		request := &requests.SendContactRequest{
+			ID:      types.Hex2Bytes(contact.ID),
+			Message: "Resending pending contact request",
```
jrainville, line 528: I think maybe we'd need a new type of message/contact request for this particular case, because if the other person did **not** approve us, maybe because they hadn't seen it before, then they will see this new message, which is not very personal, so they might reject.

jrainville (commented):
Nice one. It was a good idea to resend something to fix the pending requests.

See my comment below for my suggestion of creating a new type.

protocol/messenger_handler.go
cammellos: @0x-r4bbit I meant something a bit different, the most economical way to do it, would be to piggy back a field on whether you have added the other to contact to the datasync ACK ( https://github.com/vacp2p/mvds/blob/d6b27d3843ffae66806cc969596af342c6e7a839/protobuf/sync.proto#L7 ) for example. (That could literally be a single bit 0/1 whether we have the pk of the user in contacts)

Modifying datasync it's a bit of a pain but I guess worth considering, the solution here is probably on the heavier side, since we are increasing considerably the number of messages, while modifying datasync would not change the number of messages in the network, just make them slightly large, probably we should have a discussion about potential approaches, as there could be better solutions

protocol/messenger_handler.go
0x-r4bbit: I guess this PR needs more discussion as per https://github.com/status-im/status-go/pull/3052#discussion_r1066295420

Put it on backlog now (because it's an edge case really), so I can focus on more urgent things

0x-r4bbit:
Closing this for now as this needs further exploration as per https://github.com/status-im/status-go/pull/3052#discussion_r1066295420

PR #3057, opened 2023-01-06
chore: balance history improv

stefandunca:
### Updates [#8909](https://github.com/status-im/status-desktop/issues/8909)

Changes:
- don't fail fetching if one of the requests fails (it might be due to the node not having access to the specific chain)
- split calibration in two stages: coarse update interval and fine update the interval for intervals other than all time
- ends fetching when block not found to account for estimation errors due to high variation in block time
- fetch block timestamp and use it in reported data instead of the estimated one
- adapt tests to account for changes
- minor improvements in implementation and tests


stefandunca:
Closing in favour of the simplification refactoring: https://github.com/status-im/status-go/pull/3071

PR #3061, opened 2023-01-10
Image width height

### LINKED ISSUE #3060: Add imageWidth and imageHeight fields
Currently, an image's dimensions are calculated after loading the image. But there are several use cases where an image's dimensions need to be known in advance:

1. When building a collage/album of 3 images, either one image will be displayed horizontally/landscape and the other 2 below it, or one image will be displayed vertically/portrait and the other 2 beside it. The photo's dimensions need to be known to know which style to use
2. When using a shared element transition animation, where a small image is animated to full-screen mode
3. When using a loading placeholder, and making it occupy only as much space as the image

Probably there are more use cases

OmarBasem:
Fixes #3060

This PR adds `imageWidth` and `imageHeight` fields to messages


protocol/message_persistence.go
```diff
@@ -116,6 +118,8 @@ func (db sqlitePersistence) tableUserMessagesAllFieldsJoin() string {
 		m1.image_payload,
 		m1.image_type,
 		COALESCE(m1.album_id, ""),
+		COALESCE(m1.image_width, ""),
+		COALESCE(m1.image_height, ""),
```
0x-r4bbit: I think this needs to be coalesced to `0` since it's `INTEGER`

protocol/message_persistence.go
```diff
@@ -333,6 +341,11 @@ func (db sqlitePersistence) tableUserMessagesScanAllFields(row scanner, message
 		message.AlbumID = albumID.String
 	}
 
+	if imageWidth.Valid {
+		message.ImageWidth = uint32(imageWidth.Int32)
+		message.ImageHeight = uint32(imageHeight.Int32)
```
0x-r4bbit, line 352: Do we assume height to be valid if width is valid?

0x-r4bbit (changes requested):

protocol/message_persistence.go
OmarBasem: That's right! Fixed it.

protocol/message_persistence.go
OmarBasem: I was thinking about this. Having width without height isn't useful. But I have added a check for height also anyways.

qfrank (approved):
LGTM!

0x-r4bbit (approved):
Thanks for the update @OmarBasem 
This LGTM now.

Before merging please make sure to:

1. Update your migration script's timestamp if necessary so it's indeed the latest one
2. Squash your "update" commits so we have a single one that introduces this feature

OmarBasem:
Thanks @0x-r4bbit.
I am not sure why the iOS build is failing now?

PR #3062, opened 2023-01-10
keycard api extension

saledjenic:
status-go api extended with the following endpoints:
- `RemoveMigratedAccountsForKeycard` - removes an account by address for a keycard with uid from `keypairs` table
- `DeleteAccountForMigratedKeypair` - deletes an account by address from `accounts` table

### DIFF
--- a/multiaccounts/accounts/database_test.go
+++ b/multiaccounts/accounts/database_test.go
@@ -201,7 +201,7 @@ func TestKeypairs(t *testing.T) {
 		KeycardUID:        "00000000000000000000000000000001",
 		KeycardName:       "Card01",
 		KeycardLocked:     false,
-		AccountsAddresses: []types.Address{{0x01}, {0x02}, {0x03}},
+		AccountsAddresses: []types.Address{{0x01}, {0x02}, {0x03}, {0x04}},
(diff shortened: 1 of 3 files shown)


multiaccounts/accounts/database_test.go
```diff
@@ -302,6 +302,17 @@ func TestKeypairs(t *testing.T) {
 	}
 	require.Equal(t, true, locked)
 
+	// Test detleting accounts (addresses) for a certain keycard
```
stefandunca, line 305: Why not another test function?

stefandunca (approved):
Good job. I have a question inline.

alaibe (approved):

multiaccounts/accounts/database_test.go
saledjenic: @stefandunca it could be separate test for each function from `TestKeypairs` test. But added it here cause `TestKeypairs` tests all functionalities we have for keypairs, that's why I added it here.

stefandunca: Thanks for the explanations. It is usually easier to debug and maintain when you have separate unit-tests.

PR #3066, opened 2023-01-10
feat: Added ens_usernames database table

igor-sirotin:
Required by https://github.com/status-im/status-desktop/issues/3234

* Add `ens_usernames` table to save ENS usernames with other information, e.g. `chainId`
* Automatically add/remove values from `ens_usernames` table
* Migrate current data from `settings.usernames`

NOTE: I implemented the migration manually in go.
I first did it with SQL, but `json_each` function seem to be not supported by our go lib.
Please, let me know if this is a wrong way/place to do this.

```sql
CREATE TABLE IF NOT EXISTS ens_usernames (
	"username" TEXT NOT NULL,
	"chain_id" UNSIGNED BIGINT DEFAULT 1,
);

INSERT INTO ens_usernames (username) 
	SELECT json_each.value FROM settings, json_each(usernames);
```


appdatabase/database.go
```diff
@@ -56,6 +57,13 @@ func InitializeDB(path, password string, kdfIterationsNumber int) (*sql.DB, erro
 	if err != nil {
 		return nil, err
 	}
+
+	// Migrate `settings.usernames` here, because current SQL implementaion doesn't support `json_each`
+	err = MigrateEnsUsernames(db)
+	if err != nil {
+		return nil, err
+	}
+
```
richard-ramos: I think this should be moved to be executed before the `err = migrations.Migrate(db)` in line 56,  otherwise it will not be possible to execute migrations over the ens_usernames table in case we need to

services/ens/database.go
```diff
@@ -0,0 +1,57 @@
...
+		return result, err
+	}
+
+	defer rows.Close()
+
+	for rows.Next() {
+		var ensUsername EnsUsernameDetails
+		err = rows.Scan(&ensUsername.Username, &ensUsername.ChainId)
+		if err != nil {
+			return result, err
```
richard-ramos:
```suggestion
			return nil, err
```
To avoid returning partial results

appdatabase/database.go
```diff
@@ -107,3 +115,71 @@ func GetDBFilename(db *sql.DB) (string, error) {
...
+		usernames = append(usernames, list...)
+	}
+
+	for _, username := range usernames {
+		_, err := db.Exec(`INSERT INTO ens_usernames (username) VALUES (?)`, username)
+		if err != nil {
+			log.Error("Migrating ens usernames: failed to insert username into new database")
+			return err
+		}
+	}
```
richard-ramos: What should happen if one of the inserts fail? is it okay if the table contains incomplete results? if no, consider using a transaction and rolling back on error

appdatabase/database.go
```diff
@@ -107,3 +115,71 @@ func GetDBFilename(db *sql.DB) (string, error) {
...
+	for rows.Next() {
+		var usernamesJson string
+		err := rows.Scan(&usernamesJson)
+
+		if err != nil {
+			return err
+		}
+
+		var list []string
+		json.Unmarshal([]byte(usernamesJson), &list)
```
richard-ramos:
```suggestion
		err := json.Unmarshal([]byte(usernamesJson), &list)
		if err != nil {
		   return err 
		}
```

appdatabase/database.go
```diff
@@ -107,3 +115,71 @@ func GetDBFilename(db *sql.DB) (string, error) {
...
+	// -- 2. Move current `settings.usernames` to the new table
+	/*
+		INSERT INTO ens_usernames (username)
+			SELECT json_each.value FROM settings, json_each(usernames);
+	*/
+
+	rows, err := db.Query(`SELECT usernames FROM settings`)
+
+	if err != nil {
+		log.Error("Migrating ens usernames: failed to query 'settings.usernames'")
```
richard-ramos:
```suggestion
		log.Error("Migrating ens usernames: failed to query 'settings.usernames'", "err", err.Error())
```

appdatabase/database.go
```diff
@@ -107,3 +115,71 @@ func GetDBFilename(db *sql.DB) (string, error) {
...
+	}
+
+	// -- 1. Create new ens_usernames table
+
+	_, err = db.Exec(`CREATE TABLE IF NOT EXISTS ens_usernames (
+		"username" TEXT NOT NULL,
+		"chain_id" UNSIGNED BIGINT DEFAULT 1);`)
+
+	if err != nil {
+		log.Error("Migrating ens usernames: failed to create table")
```
richard-ramos:
```suggestion
		log.Error("Migrating ens usernames: failed to create table", "err", err.Error())
```

appdatabase/database.go
igor-sirotin: Good point.
I would say that it's ok for the table to be incomplete.
But I've changed the logic a bit, so that we keep the cycle going to try to insert all names. 
I also added printing witch username failed to insert.

wdyt?

appdatabase/database.go
```diff
@@ -107,3 +115,74 @@ func GetDBFilename(db *sql.DB) (string, error) {
...
+
+	if err != nil && err != sql.ErrNoRows {
+		return err
+	}
+
+	if tableExists {
+		return nil
+	}
+
+	// -- 1. Create new ens_usernames table
```
alaibe, line 135: could we have the structure migration in a normal migration file and only the data migration in go?

services/ens/api.go
```diff
@@ -75,6 +83,18 @@ func (api *API) Stop() {
 	})
 }
 
+func (api *API) GetEnsUsernames(ctx context.Context) ([]*EnsUsernameDetails, error) {
```
alaibe: As you are adding new API, is it going to work for mobile?

alaibe: After reading the code, maybe this should not be public?

services/ens/api.go
```diff
@@ -399,6 +420,7 @@ func (api *API) Register(ctx context.Context, chainID uint64, txArgs transaction
 	}
 
 	go api.rpcFiltersSrvc.TriggerTransactionSentToUpstreamEvent(types.Hash(tx.Hash()))
+	api.Add(ctx, chainID, fullDomainName(username))
```
alaibe: Is there no other function where we attach the username? I think setPubKey too

appdatabase/database.go
igor-sirotin: Well, sure we could.
But I'd prefer having it all in just one place. Is there any benefit of splitting this?

services/ens/api.go
igor-sirotin: Yeap, there's no problem for mobile team to switch to new API.
At the same time, the old API is kept working.

And probably a dummy question, cause I'm still new to status-go, but what exactly do you mean by "not public"?

services/ens/api.go
igor-sirotin: I'm not sure what does `setPubKey` function do. But I doesn't seem that we were saving some names from there.

Previously in desktop app we were saving it to `settings` with `saveNewEnsUsername`.
So I tracked all places where it was used and replaced them with saving to new table.

And there was just 2 places for this:
* `connectOwnedUsername`
* `ensTransactionConfirmed`

Both are now processed for the new table.

richard-ramos (approved):

appdatabase/database.go
alaibe: The benefit will be to not have a "special" migration for structure and have all the migration grouped together.

The data can probably be remove in few years while the structure change are permanent

services/ens/api.go
alaibe: Sorry by not public I mean, the function name being lowercase, that would make it private: https://yourbasic.org/golang/public-private/

services/ens/api.go
alaibe: SetPubKey is being used in this context: https://github.com/status-im/status-desktop/blob/master/src/app_service/service/ens/async_tasks.nim#L30

The wallet address owns the ens but the public key is attached to a different chat key.

I believe we also need to do it in set pub key:
```
self.confirmTransaction($PendingTransactionTypeDto.SetPubKey, receivedData.data, receivedData.transactionHash)
```

services/ens/api.go
igor-sirotin: @alaibe, I wasn't aware about that syntax, thanks 👍 

This function is called from the app with RPC, will it work if it's private?
I just followed main functions here like `Register`, `Release`, `SetPubKey` - they're public.

services/ens/api.go
igor-sirotin: Ok, I see it now. Didn't find it at first, because we weren't saving it before 🙂 
Nice catch, thank you!

alaibe (approved):

services/ens/api.go
alaibe: Nope indeed, if called from RPC they need to be capitalized

PR #3071, opened 2023-01-11
feat: retrieve balance history for tokens and cache it to DB

stefandunca:
### Refactor https://github.com/status-im/status-go/pull/2962 to simplify balance history fetching

Extends wallet module with the `history` package containing the following components:

#### `BalanceDB` (`balance_db.go`)

- Keeps track of balance information (token count, block, block timestamp) for a token identity (chain, address, currency)
- The cached data is stored in `balance_history` table.
- Uniqueness constrained is enforced by the `balance_history_identify_entry` UNIQUE index.
- The `balance_history_filter_entries` index ensures optimal DB fetching

#### `Balance` (`balance.go`)

- Provides two stages:
    - Fetch of balance history using RPC calls (`Balance.update` function)
    - Retrieving of cached balance data from the DB it exists (`Balance.get` function)
- Fetching and retrieving of data is done for specific time intervals defined by TimeInterval "enumeration"
- The update process is done for a token identity by the `Balance.Update` function
- The granularity of data points returned is defined by the constant increment step defined in `timeIntervalToStride` for each time interval.
- The `blocksStride` values have a common divisor to have cache hit between time intervals.

#### `Service` (`service.go`)

- Main APIs
    - `StartBalanceHistory`: regularly updates balance history for all enabled networks, available accounts and provided tokens.
    - `GetBalanceHistory`: retrieves cached token count for a token identity (chain, address, currency) for multiple chains
    - `UpdateVisibleTokens`: will set the list of tokens to have historical balance fetched. This is a simplification to limit tokens to a small list that makes sense

#### Fetch balance history for ECR20 tokens

- Add token.Manager.GetTokenBalanceAt to fetch the balance of a specific block number of ECR20.
- Add tokenChainClientSource concrete implementation of DataSource to fetch the balance of ECR20 tokens.
- Choose the correct DataSource implementation based on the token "is native" property.

#### Tests

Tests are implemented using a mock of the `DataSource` interface used
to intercept the RPC calls.

#### Notes:

- the timestamp used for retrieving block balance is constant
- simplification tradeoffs with negligible user side effects
  - can't fetch at any time; the user will have to wait for the update to reach and fetch the balance data for a specific token identity to be available for plotting retrieval, which is minor inconvenient, and the time to wait should improve with batching optimization if required
    - the reason is the missing arbitrary snapping implementation, I still found a way to keep the cache hit
  - degraded timestamp fetch precision, from a few seconds for most of the intervals to arbitrary based on the blockchain's block time variation, which should not be noticeable in the graph overview
  - missing snapping close to year, month, and day borders

Updates status-desktop: [#8175](https://github.com/status-im/status-desktop/issues/8175)
Closes status-destkop: [#8226](https://github.com/status-im/status-desktop/issues/8226), [#8862](https://github.com/status-im/status-desktop/issues/8862), [#8909](https://github.com/status-im/status-desktop/issues/8909)


services/wallet/history/balance_db.go
```diff
@@ -0,0 +1,178 @@
...
+	return &BalanceDB{
+		db: sqlDb,
+	}
+}
+
+// dBEntry represents a single row in the balance_history table.
+type entry struct {
+	chainID   uint64
+	address   common.Address
+	currency  string
```
alaibe: Do we still store the currency?

services/wallet/history/service.go
```diff
@@ -0,0 +1,235 @@
...
+	})
+}
+
+func (s *Service) StartBalanceHistory() {
+	go func() {
+		s.serviceContext, s.cancelFn = context.WithCancel(context.Background())
+		timer := time.NewTimer(balanceHistoryUpdateInterval)
+
+		update := func() {
+			err := s.updateBalanceHistoryForAllEnabledNetworks(s.serviceContext)
```
alaibe, line 84: do u need to pass the parameter serviceContext?

services/wallet/history/service.go
```diff
@@ -0,0 +1,235 @@
...
+
+	tokens, err := s.tokenManager.GetVisible(networkIds)
+	if err != nil {
+		return err
+	}
+
+	for _, address := range addresses {
+		s.triggerEvent(EventBalanceHistoryUpdateStarted, address, "")
+		for chainID, tokens := range tokens {
+			for _, token := range tokens {
```
alaibe: This will build the whole history for "every" token we have in store, am I correct?
How many infura token?

services/wallet/history/service.go
```diff
@@ -0,0 +1,235 @@
...
+	if err != nil {
+		return err
+	}
+
+	networkIds := make([]uint64, 0)
+	for _, network := range networks {
+		networkIds = append(networkIds, network.ChainID)
+	}
+
+	tokens, err := s.tokenManager.GetVisible(networkIds)
```
alaibe: GetVisible is not being use by desktop wallet anymore, we show token based on balance not on visibility set.

Your watcher needs to be based on the balance being currently positive for a token

services/wallet/history/service.go
```diff
@@ -0,0 +1,235 @@
...
+	return time.Now().UTC().Unix()
+}
+
+type tokenChainClientSource struct {
+	chainClientSource
+	TokenManager   *token.Manager
+	NetworkManager *network.Manager
+}
+
+func (src *tokenChainClientSource) BalanceAt(ctx context.Context, account common.Address, blockNumber *big.Int) (*big.Int, error) {
```
alaibe, line 174: When fetching balances, there is a contract to fetch balances for multiple token, do u know if that works with block number?

services/wallet/history/balance.go
```diff
@@ -0,0 +1,301 @@
...
+
+func (b *Balance) update(ctx context.Context, source DataSource, address common.Address, endingTimestamp int64, timeInterval TimeInterval) error {
+	startTimestamp := int64(0)
+	fetchTimestamp := int64(0)
+	if timeInterval != BalanceHistoryAllTime {
+		// Ensure we always get the complete range by fetching the next block also
+		startTimestamp = endingTimestamp - int64(timeIntervalDuration[timeInterval].Seconds())
+		fetchTimestamp = startTimestamp - int64(strideDuration(timeInterval).Seconds())
+	}
+	firstCached, _, err := b.db.filter(&assetIdentity{source.ChainID(), address, source.Currency()}, nil, &balanceFilter{fetchTimestamp, maxAllRangeTimestamp, expandFlag(timeIntervalToBitsetFilter[timeInterval])}, 1, asc)
```
alaibe: why not using getFirst ?

services/wallet/history/balance.go
```diff
@@ -0,0 +1,301 @@
...
+		return err
+	}
+
+	var oldestCached *big.Int
+	var oldestTimestamp int64
+	var newestCached *big.Int
+	if len(firstCached) > 0 {
+		oldestCached = new(big.Int).Set(firstCached[0].block)
+		oldestTimestamp = firstCached[0].timestamp
+		lastCached, _, err := b.db.filter(&assetIdentity{source.ChainID(), address, source.Currency()}, nil, &balanceFilter{minAllRangeTimestamp, maxAllRangeTimestamp, expandFlag(timeIntervalToBitsetFilter[timeInterval])}, 1, desc)
```
alaibe: Don't u also have a getLast

services/wallet/history/balance.go
```diff
@@ -0,0 +1,301 @@
...
+		if err != nil {
+			return err
+		}
+
+		oldestCached = new(big.Int).Set(mostRecentBlock)
+		oldestTimestamp = int64(mostRecentDataPoint.Timestamp)
+		newestCached = new(big.Int).Set(mostRecentBlock)
+	}
+
+	if oldestTimestamp > startTimestamp {
```
alaibe, line 211: why do we need to cache backward if the cache values are already fullfilled?

services/wallet/history/balance_db.go
stefandunca: Yes, this is part of the balance identity, so it is necessary to identify the balance for which token (native or ERC20). Another option will be to have a native flag and token contract, but I think the short ISO code covers the same more simply, right?

services/wallet/history/service.go
stefandunca: That is used for cancelling the ongoing long update process when we stop the wallet.

services/wallet/history/service.go
stefandunca: These are coming from `GetVisible`, and I relied on this as the ground truth for the visible tokens. However, based on your other comment, I have to change and source them from the "balance, not zero" store.

services/wallet/history/service.go
stefandunca: I will change this, thanks for the heads up.

services/wallet/history/service.go
stefandunca: I'm not aware of the multiple-balance contract. I will have a look. For ERC20 individual tokens, it works, it has already been tested with the previous implementation. Is this what you asked?

services/wallet/history/balance.go
stefandunca: This is for a specific token identity only. `getFirst` is a general query for the first cached on this chain. I will add an alias.

services/wallet/history/balance.go
stefandunca: I will add an alias to simplify the code.

services/wallet/history/balance.go
stefandunca: This is to cover the first fetch when we don't have a historical point to start from in order to cache forward. The previous implementation was finding the start point and only cache forward, but estimating a start point is difficult.

services/wallet/history/balance_db.go
alaibe: just to be sure to understand, what would be the value of this?

services/wallet/history/service.go
alaibe: this contract works for ERC20 and for native chain balance. But here clearly the interest is for erc20

services/wallet/history/service.go
alaibe: it is a bit weird to pass s."something" to a function member but i understand why

services/wallet/history/balance_db.go
stefandunca: This is the `token.symbol` from [AssetsDetailView](https://github.com/status-im/status-desktop/blob/e0c1a8cdfef7e1d9bfafed409acb718c71ad5b17/ui/imports/shared/views/AssetsDetailView.qml#L128). e.g. `ETH`, `SNT`

services/wallet/api.go
```diff
@@ -119,9 +125,27 @@ func (api *API) GetTokensBalancesForChainIDs(ctx context.Context, chainIDs []uin
...
-	return api.s.transferController.GetBalanceHistory(ctx, chainID, address, timeInterval)
+func (api *API) StartBalanceHistory(ctx context.Context) error {
+	api.s.history.StartBalanceHistory()
+	return nil
+}
+
+func (api *API) StopBalanceHistory(ctx context.Context) error {
+	api.s.history.Stop()
+	return nil
+}
```
alaibe, line 140: Do we need dedicated Start/Stop for the history, cannot we use the global Start/Stop Wallet?

services/wallet/history/balance_db.go
alaibe: Gotcha, this is what confused me, as when I read currency, I imagine USD, maybe rename it `tokenSymbol` ? this is what we use elsewhere

services/wallet/history/balance_db.go
```diff
@@ -0,0 +1,178 @@
...
+
+const (
+	asc  sortDirection = 0
+	desc sortDirection = 1
+)
+
+type assetIdentity struct {
+	ChainID  uint64
+	Address  common.Address
+	Currency string
```
alaibe: nit: TokenSymbol

services/wallet/history/balance_db.go
```diff
@@ -0,0 +1,178 @@
...
+	db *sql.DB
+}
+
+func NewBalanceDB(sqlDb *sql.DB) *BalanceDB {
+	return &BalanceDB{
+		db: sqlDb,
+	}
+}
+
+// dBEntry represents a single row in the balance_history table.
```
alaibe: nit: I think docstring is wrong, it should just be entry correct?

services/wallet/history/balance_db.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	rows, err := b.db.Query(queryStr, identity.ChainID, identity.Address, identity.Currency, (*bigint.SQLBigInt)(startingAtBlock), filter.minTimestamp, filter.maxTimestamp, filter.bitset, maxEntries)
+	if err != nil {
+		return nil, nil, err
+	}
+	defer rows.Close()
+
+	currentEntry := 0
+	result := make([]*entry, 0)
+	for rows.Next() && currentEntry < maxEntries {
```
alaibe: why not just `for rows.Next() {` as you already have the limit in the query?

services/wallet/history/service.go
```diff
@@ -0,0 +1,339 @@
...
+	eventFeed      *event.Feed
+	rpcClient      *statusrpc.Client
+	networkManager *network.Manager
+	tokenManager   *token.Manager
+	serviceContext context.Context
+	cancelFn       context.CancelFunc
+
+	timer          *time.Timer
+	visibleSymbols []string
+	mutex          sync.Mutex // Protects access to visibleSymbols
```
alaibe: nit: it seems the "good" practice is to reuse variable name + mutext e.g `visibleSymbolsMutex`

For consistency maybe name visibleSymbols: visibleTokenSymbols

services/wallet/history/service.go
```diff
@@ -0,0 +1,339 @@
...
+		db:             db,
+		eventFeed:      eventFeed,
+		rpcClient:      rpcClient,
+		networkManager: rpcClient.NetworkManager,
+		tokenManager:   tokenManager,
+	}
+}
+
+func (s *Service) Stop() {
+	s.cancelFn()
```
alaibe: can it be nil?

services/wallet/history/service.go
```diff
@@ -0,0 +1,339 @@
...
+}
+
+func (s *Service) triggerEvent(eventType walletevent.EventType, account statustypes.Address, message string) {
+	s.eventFeed.Send(walletevent.Event{
+		Type: eventType,
+		Accounts: []common.Address{
+			common.Address(account),
```
alaibe, line 72: isn't account always empty?

services/wallet/history/service.go
```diff
@@ -0,0 +1,339 @@
...
+			resetTimer(s.timer, balanceHistoryUpdateInterval)
+
+			if update() {
+				return
+			}
+		}
+	}()
+}
+
+func resetTimer(timer *time.Timer, interval time.Duration) {
```
alaibe: why not making it a function of (s *Service), the service knows about both

services/wallet/history/service.go
```diff
@@ -0,0 +1,339 @@
...
+	if network == nil {
+		return nil, errors.New("network not found")
+	}
+	token := src.TokenManager.FindToken(network, src.currency)
+	if token == nil {
+		return nil, errors.New("token not found")
+	}
+	balance, err := src.TokenManager.GetTokenBalanceAt(ctx, src.chainClient, account, token.Address, blockNumber)
+	if err != nil {
+		if err.Error() == "no contract code at given address" {
```
alaibe, line 188: Could we not reuse that to indicate that we should not continue to go back in time?

services/wallet/api.go
stefandunca: I don't want to start for mobile. However, it should be merged when it is implemented from their side.

services/wallet/history/service.go
stefandunca: No, only in case of cancellation or critical error. Otherwise, it will contain the account address when starting and finishing the fetching for the specific address. Plan to use it in UI when visualizing an account's graph to update when data is available.

services/wallet/history/service.go
stefandunca: I added `firstUnavailableBlockNo` to mark the oldest known `blockNumber` without contract and avoid fetching older blocks later. Otherwise, it is needed to mark history cached. No need for a specific state, a balance of `0` works best.

services/wallet/api.go
alaibe: Mobile do not use the StartWallet function yet, it can be merged no?

alaibe (approved):
Very nice one

services/wallet/api.go
stefandunca: Then indeed. However, I would like to postpone it until the entire work is finished and integrated into the app.

PR #3075, opened 2023-01-11
fix(reply-test): fix reply test because msgs could come out of order

### LINKED ISSUE #3070: TestReceiveReply failed sometimes
# Problem

when running [TestReceiveReply](https://github.com/qfrank/status-go/blob/0135ad942784cd66940d96cd2a3e53920bdb14e6/protocol/messenger_reply_test.go#L63) multi times in local(say 20 times, u can set -test.count=20), it may fail or success.
we may also check this fail on [jenkins](https://ci.status.im/blue/rest/organizations/jenkins/pipelines/status-go/pipelines/prs/pipelines/tests/branches/PR-3069/runs/2/nodes/34/steps/39/log/?start=0)

jrainville:
Fixes #3070

The test was failing sometimes because sometimes messages come out of order.

Is it maybe the way I send the messages that cause this out of order problem, like wrong clock or something, or it's normal?

### DIFF
--- a/protocol/messenger_reply_test.go
+++ b/protocol/messenger_reply_test.go
@@ -117,11 +117,17 @@ func (s *MessengerReplySuite) TestReceiveReply() {
 	)
 	s.Require().NoError(err)
 
-	s.Require().Len(response.Messages(), 2)
+	messages := response.Messages()
+	s.Require().Len(messages, 2)
 	//* Verify that the reply reponds to the original message
-	s.Require().True(response.Messages()[0].ResponseTo == messageID)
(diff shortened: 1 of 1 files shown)


cammellos:
> Fixes #3070
> 
> The test was failing sometimes because sometimes messages come out of order.
> 
> Is it maybe the way I send the messages that cause this out of order problem, like wrong clock or something, or it's normal?

messages are stored in a map, map don't have a default order (also messages might come out of order, in general), that's likely why it's failing

cammellos (approved):

PR #3076, opened 2023-01-11
feat(QuotedMessage): pass Deleted to QuotedMessage

jrainville:
This is to let the front end determine a new message if the quoted message is deleted instead of just unknown

I undid the change to the SELECT I did previously, and I determine in the Go code instead whether the quoted message data should be passed.
I could have done the condition in SQL instead, but I don't think the performance gain would be worth the ugly SQL code it would have needed

Based on top of https://github.com/status-im/status-go/pull/3075 just to have the fix


cammellos (approved):

0x-r4bbit (approved):
Nice changes

PR #3077, opened 2023-01-12
feat: add deleted by xxx support

yqrashawn:
add `DeletedBy` in Message, DeletedMessage, RemovedMessage only when DeletedBy != message.From
related mobile PR https://github.com/status-im/status-mobile/pull/14768

**Note: this might affect desktop, since there's addtional `DeletedBy` in messages**


0x-r4bbit (approved):
The changes LGTM but I think it makes sense to do these changes in a single commit.

yqrashawn:
> The changes LGTM but I think it makes sense to do these changes in a single commit.

Thanks for the review. I'll squash and merge this as single commit when manual QA finished on the mobile side.

The commit made that way so that it's easier to handle rebase conflicts produced by proto and migration (just drop the commit with generated file, rebase and re generate those files).

yqrashawn:
Added another commit [`2c28cc1` (#3077)](https://github.com/status-im/status-go/pull/3077/commits/2c28cc1cb74586d2b4258c8ad051084df67a2a24)

This makes the logic applies to group chat as well. Related comment https://github.com/status-im/status-mobile/pull/14768#issuecomment-1386398263

PR #3078, opened 2023-01-12
fix: split waku2 queries in 1d batches

richard-ramos:
This should reduce some lock contention in the storenodes

### DIFF
--- a/protocol/messenger_mailserver.go
+++ b/protocol/messenger_mailserver.go
@@ -573,26 +573,33 @@ func (m *Messenger) processMailserverBatch(batch MailserverBatch) error {
 
 		topicsForIteration := batch.Topics[i:j]
 
-		cursor, storeCursor, err := m.transport.SendMessagesRequestForTopics(ctx, mailserverID, batch.From, batch.To, nil, nil, topicsForIteration, true)
-		if err != nil {
-			logger.Error("failed to send request", zap.Error(err))
-			return err
-		}
+		for fromSeconds := batch.From; fromSeconds < batch.To; fromSeconds += 86400 { // Limit queries to 1 day (86400s) per batch
+			toSeconds := fromSeconds + 86400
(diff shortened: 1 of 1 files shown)


cammellos:
@richard-ramos it would be good to make sure this is tested in nim-waku, just wondering what happens if there's many messages with the same timestamps for example etc (i.e if there are many messages at the batch boundary, how does nim-waku behaves? does it skip messages, does it duplicate messages etc), do you know if that's something we looked at?

alaibe (approved):

LNSD:
Could you provide some context about which is the problem you are trying to work around with this "1-day batches" strategy?

> This should reduce some lock contention in the storenodes

What do you mean by "lock contention"? 🤔

richard-ramos:
> Could you provide some context about which is the problem you are trying to work around with this "1-day batches" strategy?
> 
> > This should reduce some lock contention in the storenodes
> 
> What do you mean by "lock contention"? thinking

I was reading about [WAL](https://www.sqlite.org/wal.html) and [locking](https://www.sqlite.org/lockingv3.html) and saw that while WAL mode lets you have multiple readers, while a single writer writes to the wal file, there will be periodic checkpoints when the writer will need to do an exclusive lock on the DB to commit the wal file back into the database. If i understood correctly, while this happens, no new readers will be allowed, and the writer will wait until existing readers finish retrieving results

Using a smaller range of time should reduce the execution time of these queries (as well as probably reducing the memory usage as it would not have to work with 30d of data (either by using the indexes or scanning the table depending on how the execution plan decided to do the query) allowing the writer to commit the wal file faster. 

It's probably a good idea to check the execution plan in sqlite to know if this change is worth it, so in the meantime this PR should not be merged...

richard-ramos:
I converted the PR to draft to avoid merging it by mistake :)

PR #3079, opened 2023-01-12
Feature/8640 unified currency rounding

dlipicar:
Needed for issue https://github.com/status-im/status-desktop/issues/8640, PR https://github.com/status-im/status-desktop/pull/8911

-Added library iso4217, used to get standard information about fiat currencies (mainly the number of decimal places to display).
-Added api functions for iso4217 data
-Added "Peg" information to tokenStore and token structure. Tells if a certain crypto currency is pegged to a fiat currency.
-Use numbers instead of display strings for token market values. Reorganized data in token struct.
-Adapted API to fetch prices in multiple currencies at the same time.
-Added currency "usd" every time we fetch prices.
-Added DAI on Goerli to tokenstore


alaibe (approved):

services/wallet/cryptocompare.go
```diff
@@ -94,13 +94,14 @@ func chunkSymbols(symbols []string) [][]string {
...
+func fetchCryptoComparePrices(symbols []string, currencies []string) (map[string]map[string]float64, error) {
 	chunks := chunkSymbols(symbols)
-	result := make(map[string]float64)
+	result := make(map[string]map[string]float64)
+	realCurrencies := renameSymbols(currencies)
 	for _, smbls := range chunks {
 		realSymbols := renameSymbols(smbls)
 		httpClient := http.Client{Timeout: time.Minute}
-		url := fmt.Sprintf("%s/data/pricemulti?fsyms=%s&tsyms=%s&extraParams=Status.im", cryptocompareURL, strings.Join(realSymbols, ","), currency)
+		url := fmt.Sprintf("%s/data/pricemulti?fsyms=%s&tsyms=%s&extraParams=Status.im", cryptocompareURL, strings.Join(realSymbols, ","), strings.Join(realCurrencies, ","))
```
stefandunca, line 104: This looks like a job for `net/url` package. Not sure if any of the query parameters are unsafe and might benefit from escaping. Do any come from third-party input?

stefandunca (approved):
Please have a look at my question inline.

services/wallet/cryptocompare.go
dlipicar: Hmm I've found references to a "custom Token", which I'm guessing is some token the user can manually enter if it's not in our `tokenstore`, so it might be possible they can add whatever string to the `realSymbols` list. No idea if that feature's working, if custom tokens are somehow sanitized/checked before getting accepted or how much damage they can do with that.

dlipicar: @stefandunca wanna create an issue to investigate this?

stefandunca: Here it is https://github.com/status-im/status-desktop/issues/9099

PR #3081, opened 2023-01-12
feat(collectibles): Mint collectibles (ERC-721):

### LINKED ISSUE #3051: Collectibles - smart contract and service - phase 1
Generating go API for smart contract with abigen.
Creating a service for smart contract which will use the generated API.
Testing deployment with a desktop application.

In the begining testing smart contract and Goerli network will be used.
ERC-721 smart contract will be used when ready.

endulab:
Add 1st version of ERC721 smart contract and go API.
Add collectibles service.

Issue #3051


endulab:
This is a draft which will be evolving. We will upgrade smart contract to ERC-721 soon.

richard-ramos (approved):
Good work! 
I have some questions:

1. Is this contract also part of another status-im repo? or is it only available in status-go?

2. If i understood correctly, this contract is going to be deployed by the user when they invoke `collectibles_deploy`. How much gas does it take to deploy this contract? The reason on why I ask this is because I wonder if in a future PR we could make it cheaper for the user to create multiple instances of this contract by using a factory pattern in solidity (It would ideally require having people from the status community deploy a factory and base contract in the chains we have available, and then we'd later have in status-go a map of the factory addresses in all chains (which we would use to create new instances), and in case we connect to a chain where these contracts are not deployed, the user would deploy the contract normally (even if it's more expensive).

endulab:
> Good work! I have some questions:
> 
> 1. Is this contract also part of another status-im repo? or is it only available in status-go?

It comes from https://github.com/status-im/communities-contracts.

> 2. If i understood correctly, this contract is going to be deployed by the user when they invoke `collectibles_deploy`. How much gas does it take to deploy this contract? The reason on why I ask this is because I wonder if in a future PR we could make it cheaper for the user to create multiple instances of this contract by using a factory pattern in solidity (It would ideally require having people from the status community deploy a factory and base contract in the chains we have available, and then we'd later have in status-go a map of the factory addresses in all chains (which we would use to create new instances), and in case we connect to a chain where these contracts are not deployed, the user would deploy the contract normally (even if it's more expensive).

Not user, rather community owner. You can create collectibles for your community and then you can transfer to your members in order to give some permissions later.
The cost of the last contract I deployed was 13.60671987 Gwei.
@gravityblast what do you think about the factory idea ?

endulab:
@richard-ramos I asked @gravityblast about the factory idea. Unfortunately, it is not possible in our case.
It is the user who have to deploy the contract.

jrainville (approved):
LGTM

PR #3085, opened 2023-01-13
fix: add discord message data to quoted messages

0x-r4bbit:
Because `QuotedMessage` doesn't include imported message data, some of the author information in imported messages is lost in frontends.

This commit adds a `discordMessage` (soon replaced by `importedMessage`) to `Quotedmessage`, although only hydrated with a subset of data, namely author display name and avatar URL, as those are the only ones needed by front-end atm.

### DIFF
--- a/protocol/message_persistence.go
+++ b/protocol/message_persistence.go
@@ -25,6 +25,10 @@ LEFT JOIN discord_message_authors dm_author
 ON        dm.author_id = dm_author.id
 LEFT JOIN discord_message_attachments dm_attachment
 ON        dm.id = dm_attachment.discord_message_id
+LEFT JOIN discord_messages m2_dm
+ON        m2.discord_message_id = m2_dm.id
+LEFT JOIN discord_message_authors m2_dm_author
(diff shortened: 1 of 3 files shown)


protocol/message_persistence.go
```diff
@@ -1019,10 +1038,20 @@ func (db sqlitePersistence) PinnedMessageByChatIDs(chatIDs []string, currCursor
...
+
+			 LEFT JOIN 
+							discord_messages m2_dm
+			 ON        
+			 m2.discord_message_id = m2_dm.id
+
+				LEFT JOIN 
+							discord_message_authors m2_dm_author
+			 ON        
+			 m2_dm.author_id = m2_dm_author.id
```
jrainville, line 1054: The indentations look crazy on Github. Not sure if spaces and tabs might have been mixed?

jrainville (approved):
Good job

protocol/message_persistence.go
0x-r4bbit: I know, unfortunately they aren't honored by the lint-fix tool. I see if I can get them aligned.

PR #3086, opened 2023-01-13
refactor(communities): improve archive handling memory pressure

0x-r4bbit:
Instead of loading the entire torrent file into memory when trying to extrract active messages, we now only read the chunks that are necessary to decode any individual archive and then process extracted messages in chunks.

This doesn't introduce a max cap of allowed memory yet, since the chunk size depends entirely on the size of the archive, but this will be done soon.

### DIFF
--- a/protocol/messenger_handler.go
+++ b/protocol/messenger_handler.go
@@ -1027,36 +1027,15 @@ importMessageArchivesLoop:
 			// wait for all archives to be processed first
 			downloadedArchiveID := archiveIDsToImport[0]
 
-			messagesToHandle, err := m.communitiesManager.ExtractMessagesFromHistoryArchives(id, []string{downloadedArchiveID})
+			archiveMessages, err := m.communitiesManager.ExtractMessagesFromHistoryArchive(id, downloadedArchiveID)
 			if err != nil {
 				m.communitiesManager.LogStdout("failed to extract history archive messages", zap.Error(err))
 				continue
 			}
 
-			importedMessages := make(map[transport.Filter][]*types.Message, 0)
-			otherMessages := make(map[transport.Filter][]*types.Message, 0)
-
-			for filter, messages := range messagesToHandle {
-				for _, message := range messages {
-					if message.ThirdPartyID != "" {
-						importedMessages[filter] = append(importedMessages[filter], message)
-					} else {
-						otherMessages[filter] = append(otherMessages[filter], message)
-					}
-				}
-			}
-
-			m.config.messengerSignalsHandler.ImportingHistoryArchiveMessages(types.EncodeHex(id))
-
-			err = m.handleImportedMessages(importedMessages)
+			response, err := m.handleArchiveMessages(archiveMessages, id)
 			if err != nil {
-				m.communitiesManager.LogStdout("failed to handle imported messages", zap.Error(err))
-				continue
-			}
-
-			response, err := m.handleRetrievedMessages(otherMessages, false)
-			if err != nil {
-				m.communitiesManager.LogStdout("failed to write history archive messages to database", zap.Error(err))
+				m.communitiesManager.LogStdout("failed to handle archive messages", zap.Error(err))
 				continue
 			}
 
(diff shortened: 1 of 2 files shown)


0x-r4bbit:
@cammellos I took a first stab at this based on our discussion. For now I tried the most simple solution (using callbacks when processing chunks).

The changes are very limited, as expected and discussed.
I'm putting this up here for early review, but will need to test this with real world data.

cammellos:
@0x-r4bbit looks about right, next step is to check how it behaves with large files, best thing is to compare the two with the same and check memory usage!

0x-r4bbit:
@cammellos tested this today with the Status community and this improves things a lot.
Memory is stable, CPU usage as well.

If you approve we can merge this and give our users a better time haha

protocol/messenger_handler.go
```diff
@@ -1027,58 +1027,75 @@ importMessageArchivesLoop:
 			// wait for all archives to be processed first
 			downloadedArchiveID := archiveIDsToImport[0]
```
Samyoul, line 1028: I notice that this gets the first archiveID, which is then passed into `[]string{downloadedArchiveID}` to satisfy the function signature of `ExtractMessagesFromHistoryArchives`.

`ExtractMessagesFromHistoryArchives` is only called in one place, so I wonder if this function should expect `archiveIDs []string` and rather expect a `archiveID string`?

protocol/messenger_handler.go
```diff
@@ -1027,58 +1027,75 @@ importMessageArchivesLoop:
...
 
-			messagesToHandle, err := m.communitiesManager.ExtractMessagesFromHistoryArchives(id, []string{downloadedArchiveID})
-			if err != nil {
-				m.communitiesManager.LogStdout("failed to extract history archive messages", zap.Error(err))
-				continue
-			}
-
-			importedMessages := make(map[transport.Filter][]*types.Message, 0)
-			otherMessages := make(map[transport.Filter][]*types.Message, 0)
+			m.communitiesManager.ExtractMessagesFromHistoryArchives(id, []string{downloadedArchiveID}, func(archiveMessages []*protobuf.WakuMessage, err error) {
```
Samyoul: This is a very large inline function, and as far as I can see the only parameter it consumes from outside the function scope is `downloadedArchiveID`, which could be passed as a parameter.

Perhaps this logic should be made an explicit function of `Messenger`, you can still pass it in the same way.

`m.communitiesManager.ExtractMessagesFromHistoryArchives(id, []string{downloadedArchiveID}, m.yourPostProcessingFunction`)

Or alternatively return `archiveMessages []*protobuf.WakuMessage, e error` from `ExtractMessagesFromHistoryArchives ` and then pass those vars to the new function. I can see that the cause of this requirement in your code is because you require structs and functionality from outside of the scope of `communitiesManager`.

Returning the result of the new function will keep a clean separation of concerns and will be more maintainable. What do you think?

protocol/messenger_handler.go
0x-r4bbit: This has historically changed over time indeed.
Originally `ExtractMessagesFromHistoryArchives` took a list of archive IDs and would then extract all messages for every archive matching a given ID.

This also means however that it would extract **all** messages for the given archive IDs, which means there's more messages to process/import/save. 

Because we need to wait for this operation to finish before the task can be cancelled, in case it has been cancelled, we've reduced this to only ever process one archive at a time such that, if the process is marked as cancelled, we can stop the process midflight and don't have to wait for all messages to be processed.

The only reason I haven't changed the method signature back then is because I tend to avoid breaking changes.
But you're right that we can probably change this by now.

protocol/messenger_handler.go
0x-r4bbit: > Or alternatively return archiveMessages []*protobuf.WakuMessage, e error from ExtractMessagesFromHistoryArchives  and then pass those vars to the new function. I can see that the cause of this requirement in your code is because you require structs and functionality from outside of the scope of communitiesManager.

This was actually the case before, but since we're now processing message archives in chunks, instead of loading all archive messages into memory that we're interested in, we have to provide a callback (or use some channel gymnastics) to let the function process a subset of messages before it loads the next chunk.

> Perhaps this logic should be made an explicit function of Messenger, you can still pass it in the same way.

Very happy to make this change. In fact I started out like this and decided to keep it inline because there was no other place where this functionality is used.

Will update this!

0x-r4bbit: > This was actually the case before, but since we're now processing message archives in chunks, instead of loading all archive messages into memory that we're interested in, we have to provide a callback (or use some channel gymnastics) to let the function process a subset of messages before it loads the next chunk.

Actually, now that we change this to only expect a single archive ID anyways, we don't need the `done` callback anymore, so what you're saying makes sense, we can now go back to returning `messages, err`

Thanks for pointing that out!

0x-r4bbit:
@Samyoul updated this according to your comments.
Much better like this and thank you for your valuable feedback

Samyoul (approved):
This all looks great @0x-r4bbit. Great work.

Do you plan on checking the archive size and chunking archive greater than some limit, like 64, 128 or 256mb? Just in case an archive file is 1GB or greater that.

0x-r4bbit:
@Samyoul I've put up a PR for that here: https://github.com/status-im/status-go/pull/3095

It's not exactly operating on the overall size of an archive, but gives a rough estimate. Maybe you can have a look there?

PR #3087, opened 2023-01-13
Upgrade linter to 1.50.0

cammellos:
Nothing of note, apart some pesky `if o == nil` in `communities/community.go`, those probably we should address and have the caller do the nil checking, rather than the callee.
I have disabled a few that looked like the linter was getting confused ("import "C") and some that I am not sure what it wanted me to do and they were in the tests (MustCompile).


jrainville (approved):
LGTM

PR #3089, opened 2023-01-14
chore: build library using vendor folder

richard-ramos:
I noticed that when we execute `make statusgo-library` or `make statusgo-shared-library` go would build status-go using the source code from $GOHOME instead the source from vendor/ folder

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -151,6 +151,7 @@ statusgo-library: ##@cross-compile Build status-go as static library for current
 	go run cmd/library/*.go > $(GOBIN)/statusgo-lib/main.go
 	@echo "Building static library..."
 	go build \
+		-mod vendor \
 		-tags '$(BUILD_TAGS)' \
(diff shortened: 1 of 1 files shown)


cammellos (approved):

jakubgs (approved):
Good catch.

osmaczko (commented):
> By default, if the [go version](https://go.dev/ref/mod#go-mod-file-go) in go.mod is 1.14 or higher and a vendor directory is present, the go command acts as if -mod=vendor were used. Otherwise, the go command acts as if -mod=readonly were used.

https://go.dev/ref/mod#build-commands

Just wondering why we need to be explicit here :thinking:

richard-ramos:
Indeed. I tried on a VM and I dont need to set the `-mod vendor` option. I'll close the PR in the meantime to avoid merging it, and will figure out what's going on on my machine.
Thank you @osmaczko

PR #3092, opened 2023-01-16
waku2 - restart discv5 when peers are low and use newer go-waku code to deal with discovery

richard-ramos:
In https://github.com/waku-org/go-waku/pull/424 a peer connector object is introduced that will try to keep the node connected up to N peers. Each time you discover a peer it's pushed into a buffered channel which is consummed by the peer connector to attempt to connect to it (up to 15 concurrent connection attempts might happen).

The reason why it's done this way instead of using go-libp2p-pubsub discovery interface is because it uses the function [EnoughPeers](https://github.com/libp2p/go-libp2p-pubsub/blob/master/discovery.go#L139-L145) to determine if it has to find new peers or not, and the criteria it uses is to check if we have less than `D_Low` peers connected (and `D_Low` = 4).

In addition to that, it restarts the discv5 service when status-go is disconnected and reconnects, because go-ethereum discv5 will not reconnect to bootnodes as described here: https://github.com/ethereum/go-ethereum/issues/25203


rasom (approved):

wakuv2/waku.go
```diff
@@ -346,6 +348,20 @@ func New(nodeKey string, fleet string, cfg *Config, logger *zap.Logger, appDB *s
...
+					isConnected = false
+					waku.node.DiscV5().Stop()
+				} else if latestConnStatus.IsOnline && !isConnected {
+					isConnected = true
+					if !waku.node.DiscV5().IsStarted() {
+						err := waku.node.DiscV5().Start(ctx)
+						if err != nil {
+							waku.logger.Error("Could not start DiscV5", zap.Error(err))
+						}
+					}
```
rasom: I think I makes sense to add debug logs in those branches just in case

PR #3095, opened 2023-01-16
refactor(Communities): ensure generated archives don't exceed max cap

0x-r4bbit:
This introduces an addition constraint to archive generation, in which the payload + signature size of all partitioned message that go into an archive should not exceed a certain threshold.

This is to ensure that archives won't get too big when they are later read into memory.

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -37,6 +37,8 @@ var defaultAnnounceList = [][]string{
 }
 var pieceLength = 100 * 1024
 
+const maxArchiveSizeInBytes = 30000000
+
 var ErrTorrentTimedout = errors.New("torrent has timed out")
 
 type Manager struct {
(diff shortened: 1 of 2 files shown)


protocol/communities/manager.go
```diff
@@ -36,6 +36,7 @@ var defaultAnnounceList = [][]string{
 	{"udp://tracker.openbittorrent.com:6969/announce"},
 }
 var pieceLength = 100 * 1024
+const maxArchiveSizeInBytes = 30000000
```
0x-r4bbit, line 40: Based on some measurements, one of the biggest message partition size was ~20 MB 
(size = all messages len(payload) + len(sig), so I added 10 MB for slack as there's more in a message, plus there's also the archive metadata etc)

You think this makes sense @cammellos or should I adjust this?

saledjenic (approved):

protocol/communities/manager.go
```diff
@@ -1894,58 +1896,81 @@ func (m *Manager) CreateHistoryArchiveTorrent(communityID types.HexBytes, msgs [
...
-			return archiveIDs, err
+		// Not only do we partition messages, we also chunk them
+		// roughly by size, such that each chunk will not exceed a given
+		// size and archive data doesn't get too big
+		messageChunks := make([][]types.Message, 0)
+		currentChunkSize := 0
+		currentChunk := make([]types.Message, 0)
+		for i, msg := range messages {
+			msgSize := len(msg.Payload) + len(msg.Sig)
+			if currentChunkSize+msgSize > maxArchiveSizeInBytes && len(currentChunk) > 0 {
```
stefandunca: I see that this `len(currentChunk) > 0` check makes sense only in the first iteration, when the first message is already over `maxArchiveSizeInBytes`, in which case we violate the constraint and add it anyway.

Isn't going over `maxArchiveSizeInBytes` if we include only one message an issue?

In the second iteration onwards, there will always be something in the `currentChunk`.

protocol/communities/manager.go
```diff
@@ -1894,58 +1896,81 @@ func (m *Manager) CreateHistoryArchiveTorrent(communityID types.HexBytes, msgs [
...
+				messageChunks = append(messageChunks, currentChunk)
+				currentChunk = make([]types.Message, 0)
+				currentChunkSize = 0
+			}
+			currentChunk = append(currentChunk, msg)
+			currentChunkSize = currentChunkSize + msgSize
+			if i == len(messages)-1 {
+				messageChunks = append(messageChunks, currentChunk)
```
stefandunca: It looks to me that if you move `messageChunks = append(messageChunks, currentChunk)` after the loop, this check `if i == len(messages)-1 {` is not needed; what do you think?

stefandunca (approved):
Nice change, please see my questions inline.

protocol/communities/manager.go
0x-r4bbit: Good catch, updated!

protocol/communities/manager.go
0x-r4bbit: After you've pointed that out, I decided to drop this altogether and simply drop messages that are already bigger than our max cap.
Please have another look!

PR #3097, opened 2023-01-17
feat: api to return service up or down

alaibe:
Check connection to service

### DIFF
--- a/services/wallet/chain/client.go
+++ b/services/wallet/chain/client.go
@@ -3,6 +3,7 @@ package chain
 import (
 	"context"
 	"math/big"
+	"sync"
 
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/common"
@@ -13,10 +14,14 @@ import (
 	"github.com/status-im/status-go/services/rpcstats"
 )
 
+var ChainClientInstances = make(map[uint64]*Client)
+
 type Client struct {
-	eth       *ethclient.Client
-	ChainID   uint64
-	rpcClient *rpc.Client
+	eth             *ethclient.Client
+	ChainID         uint64
+	rpcClient       *rpc.Client
+	IsConnected     bool
+	IsConnectedLock sync.RWMutex
 }
 
 type FeeHistory struct {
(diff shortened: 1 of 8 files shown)


services/wallet/chain/client.go
```diff
@@ -46,63 +55,93 @@ func NewClients(rpc *rpc.Client, chainIDs []uint64) (res []*Client, err error) {
...
+
 func (cc *Client) ToBigInt() *big.Int {
 	return big.NewInt(int64(cc.ChainID))
 }
 
 func (cc *Client) HeaderByHash(ctx context.Context, hash common.Hash) (*types.Header, error) {
 	rpcstats.CountCall("eth_getBlockByHash")
-	return cc.eth.HeaderByHash(ctx, hash)
+	resp, err := cc.eth.HeaderByHash(ctx, hash)
+	defer cc.toggleIsConnected(err)
```
dlipicar, line 77: Hmm I'd like avoiding repetition by moving this to a single location like you did in `doQuery`. I'd need to do some reading to know how to do this in go, but I'm guessing passing a lambda instead of a URL. Not sure it ends up being tidier, though.

services/wallet/reader.go
```diff
@@ -105,6 +107,21 @@ func (r *Reader) Start() error {
...
+	go func() {
+		ticker := time.NewTicker(time.Minute)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				r.walletFeed.Send(walletevent.Event{
+					Type: EventWalletTickCheckConnected,
```
dlipicar, line 120: Where is this event handled? Cannot find another reference to it

alaibe: It is in status-desktop

dlipicar: ohh sorry, missed the other PR

dlipicar (approved):

services/wallet/chain/client.go
alaibe: I might be missing something but I don't see how i could avoid to call the `defer` here, to me it is similar than countCall being on all function. Happy to change to you suggestion, if you can help me how to

services/wallet/chain/client.go
```diff
@@ -13,22 +13,31 @@ import (
...
+var ChainClientInstances = make(map[uint64]*Client)
+
 type Client struct {
-	eth       *ethclient.Client
-	ChainID   uint64
-	rpcClient *rpc.Client
+	eth         *ethclient.Client
+	ChainID     uint64
+	rpcClient   *rpc.Client
+	IsConnected bool
```
richard-ramos: Is it possible for a client to be called by goroutines? if so, it might be a good idea to convert the client into a mutex and do  `cc.Lock()` / `cc.RLock()` wherever you set or access the `IsConnected` attribute

```suggestion
        sync.RWMutex

	eth         *ethclient.Client
	ChainID     uint64
	rpcClient   *rpc.Client
	IsConnected bool
```

richard-ramos (approved):

richard-ramos (approved):

PR #3107, opened 2023-01-18
feat(wallet): add price cache

dlipicar:
In this PR, we add persistent caching for token prices.

For every token symbol + currency we store an entry on the database with the latest fetched price.

A new table `price_cache` was added to the DB.

Price fetching/getting is now routed through `PriceManager` instead of direct calls to the `CryptoCompare` API.

A unit test was added and the new feature was tested in status-desktop.


appdatabase/migrations/sql/1674056187_add_price_cache.up.sql
```diff
@@ -0,0 +1,8 @@
+-- One entry per token, currency
+CREATE TABLE IF NOT EXISTS price_cache (
+    token VARCHAR NOT NULL,
+    currency VARCHAR NOT NULL,
+    price VARCHAR NOT NULL
```
stefandunca: Wondering why the choice to store it as string? Won't `REAL` be a better fit?

services/wallet/price.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func getCachedPricesFromDBRows(rows *sql.Rows) (PricesPerTokenAndCurrency, error) {
+	prices := make(PricesPerTokenAndCurrency)
+	for rows.Next() {
+		var token string
+		var currency string
+		var priceStr string
+		if err := rows.Scan(&token, &currency, &priceStr); err != nil {
+			return prices, err
```
stefandunca: Wondering why choosing to return valid `prices` instance instead of `nil`? What is the value of the partial results in case of an error? Also usually, in case of exceptional errors, the call results are discarded and should need documenting if otherwise.

services/wallet/price.go
```diff
@@ -0,0 +1,82 @@
...
+		}
+
+		_, present := prices[token]
+		if !present {
+			prices[token] = map[string]float64{}
+		}
+
+		price, err := strconv.ParseFloat(priceStr, 64)
```
stefandunca: If the data type of the price column is `REAL`, there is no need for this conversion; `Scan` will do it instead.

services/wallet/price_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/status-im/status-go/appdatabase"
+	"github.com/status-im/status-go/sqlite"
+)
+
+func setupTestPriceDB(t *testing.T) (*PriceManager, func()) {
+	tmpfile, err := ioutil.TempFile("", "wallet-price-tests-")
```
stefandunca: DB in memory should be enough and faster in memory for this type of test. See an example [here](https://github.com/status-im/status-go/blob/da73e8161273fafe248c539d2f6b332c872cde87/services/wallet/history/balance_db_test.go#L17)

services/wallet/price_test.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+}
+
+func TestPrice(t *testing.T) {
+	manager, stop := setupTestPriceDB(t)
+	defer stop()
+
+	rst, err := manager.GetCachedPrices()
+	require.NoError(t, err)
+	require.Empty(t, rst)
```
stefandunca, line 25: Reminder, if the query fails, this will be `nil`.

services/wallet/price.go
```diff
@@ -0,0 +1,82 @@
...
+		(token,	currency, price)
+		VALUES
+		(?, ?, ?)`)
+	if err != nil {
+		return err
+	}
+
+	for token, pricesPerCurrency := range prices {
+		for currency, price := range pricesPerCurrency {
+			_, err = insert.Exec(token, currency, fmt.Sprintf("%f", price))
```
saledjenic, line 55: @dlipicar for multiple actions in one go like this one we should use begin/commit transaction. Could you update it?

stefandunca (approved):
Nicely done. Please see my notes inline.

saledjenic (changes requested):
@dlipicar it's enough to commit only `appdatabase/migrations/bindata.go` and `appdatabase/migrations/sql/1674056187_add_price_cache.up.sql` you don't need other automatically generated files.

@dlipicar please update the version.

services/wallet/price_test.go
saledjenic: That's why we have line 30, if it fails we won't get to line 31, right?

appdatabase/migrations/sql/1674056187_add_price_cache.up.sql
dlipicar: Ugh, yeah... I was reading MySQL docs (https://www.w3schools.com/sql/sql_datatypes.asp) and it seemed one needed to specify some parameters when dealing with floats, so a string seemed more flexible. 
It seems SQLite is simpler in that regard, I'll make the change. Thanks!

services/wallet/price_test.go
dlipicar: Thanks! I was just following the SavedAddresses example

services/wallet/price.go
dlipicar: I'll look that up. Sorry, don't have much experience with SQL

dlipicar:
@stefandunca @saledjenic I think I applied all your suggestions
- Upload only necessary migration files
- Return nil in case of an error
- Use DB in memory for test
- Updated VERSION
- Queue all SQL write transactions before executing them
- Use floating point field for price in DB

services/wallet/api.go
```diff
@@ -317,7 +317,12 @@ func (api *API) GetEthereumChains(ctx context.Context, onlyEnabled bool) ([]*par
...
 func (api *API) FetchPrices(ctx context.Context, symbols []string, currencies []string) (map[string]map[string]float64, error) {
 	log.Debug("call to FetchPrices")
-	return fetchCryptoComparePrices(symbols, currencies)
+	return api.s.priceManager.FetchPrices(symbols, currencies)
+}
+
+func (api *API) GetCachedPrices(ctx context.Context) (map[string]map[string]float64, error) {
+	log.Debug("call to GetCachedPrices")
+	return api.s.priceManager.GetCachedPrices()
 }
```
alaibe, line 326: How does a client knows when he has to call fetch prices or get cached prices.

Could we not make it transparent for the client?

services/wallet/api.go
```diff
@@ -317,7 +317,12 @@ func (api *API) GetEthereumChains(ctx context.Context, onlyEnabled bool) ([]*par
 
 func (api *API) FetchPrices(ctx context.Context, symbols []string, currencies []string) (map[string]map[string]float64, error) {
 	log.Debug("call to FetchPrices")
-	return fetchCryptoComparePrices(symbols, currencies)
```
alaibe, line 320: shall your code not replace the cryptocompare API?

Also there is things being done like Token rewriting that needs to be added here to avoid problems, like STT<->SNT

services/wallet/api.go
dlipicar: Hmm I guess it depends on whether you care or not about the values being the current ones. Whether you need fresh prices or old ones is up to the client to decide.

In this particular case, the cache is used to have "some" price instead of none/having to fetch it during startup.

We can probably be smarter about caching in status-go so we don't need to have the timed cache on the Nim side (perhaps status-go should return every price with a timestamp so the client knows if they need to update it or not?), but that's a different problem than what this PR is trying to solve.

services/wallet/api.go
dlipicar: I tried to kind of build the `PriceManager` so that it would some day become a full wrapper for Cryptocompare (in the future we might want to replace it or have other providers as backup). When we decide to do that, it'd be smart to move some of the things cryptocompare.go does to the PriceManager, I just didn't think it was within the scope of this PR. Here I was just trying to make sure all price fetches went through `PriceManager` so I would have the chance to cache the results.

saledjenic (approved):

PR #3111, opened 2023-01-19
Resume import of message history archives upon bootstrap

0x-r4bbit:
This adds the functionality that history archives continue to be imported in case the import has been interrupted the last time the app/client was running.

This typically happens when users don't wait for an ongoing import to finish, which sometimes can take a while. Users then close the app/kill the client which leaves the database in a state where there's downloaded archives that haven't been fully imported.

Prior to this change, the node will have to wait until it receives a new magnetlink that it hasn't seen before, until it processes imports again. This can take several days.

Now, it will check on startup if there are any archives left to be imported and resumes the import from there.

### DIFF
--- a/protocol/messenger.go
+++ b/protocol/messenger.go
@@ -720,6 +720,20 @@ func (m *Messenger) Start() (*MessengerResponse, error) {
 		}
 	}
 
+	joinedCommunities, err := m.communitiesManager.Joined()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, joinedCommunity := range joinedCommunities {
+		// resume importing message history archives in case
+		// imports have been interrupted previously
+		err := m.resumeHistoryArchivesImport(joinedCommunity.ID())
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	if m.httpServer != nil {
 		err = m.httpServer.Start()
 		if err != nil {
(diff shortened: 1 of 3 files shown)


ghost:
### Pull Request Checklist

jrainville (approved):
LGTM

protocol/messenger.go
```diff
@@ -720,6 +720,20 @@ func (m *Messenger) Start() (*MessengerResponse, error) {
...
 
+	joinedCommunities, err := m.communitiesManager.Joined()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, joinedCommunity := range joinedCommunities {
+		// resume importing message history archives in case
+		// imports have been interrupted previously
+		err := m.resumeHistoryArchivesImport(joinedCommunity.ID())
```
richard-ramos, line 731: If there's an error with a single community, you might end up some communities being processed and some not.
Is that problematic? Could the error handling instead of returning the error, log it and continue with the next community?

richard-ramos (approved):

protocol/messenger.go
0x-r4bbit: Was considering that as well! And then I had a @cammellos in my head saying something like "generally we want to fail fast".

Happy to change that though if there's strong opinions! :D

PR #3117, opened 2023-01-21
Add compressed key to multiaccount/messages/contacts

cammellos:
Adds compressed keys to contacts, messages, multiaccount.

Adds also a BuildContact endpoint and a SerializeLegacyKey binding for convenience.


multiaccounts/settings/structs.go
```diff
@@ -187,3 +188,21 @@ type Settings struct {
 	GifAPIKey                      string                        `json:"gifs/api-key"`
 	TestNetworksEnabled            bool                          `json:"test-networks-enabled?,omitempty"`
 }
+
+func (s Settings) MarshalJSON() ([]byte, error) {
+	type Alias Settings
+	item := struct {
+		Alias
+		CompressedKey string `json:"compressed-key"`
```
ilmotta, line 196: Shouldn't status-go always use camel case for JSON keys?

cammellos: It should :), for historical reasons though `settings` it's using kebab case, so I kept it consistent (if you look above there's the whole struct), we should change it, but probably requires some coordinated effort with desktop, and maybe best to do in a separate pr, what do you think?

ilmotta: I see now, renaming the keys would be a breaking change for Desktop. I agree, better do it in a separate PR.

ilmotta (approved):

richard-ramos (approved):

PR #3120, opened 2023-01-21
Support outgoing contact requests

alwx:
This PR is required to implement outgoing contact requests and to display the status of contact requests to end users.
See more details here: https://www.figma.com/file/eDfxTa9IoaCMUy5cLTp0ys/Shell-for-Mobile?node-id=4537%3A515620&t=YeDEcoVkqUsZNjIW-0

It also refactors the API and removes the redundant methods (for example, there were two methods to dismiss/cancel the incoming contact request, one of them was not even used).

All these changes  required for the following issues:
- https://github.com/status-im/status-mobile/issues/13963
- https://github.com/status-im/status-mobile/issues/13973
- https://github.com/status-im/status-mobile/issues/13972

Here is the PR that depends on it:
https://github.com/status-im/status-mobile/pull/14853

# What's done?

1. The new outgoing contact request notification introduced.
2. Some minor refactoring of existing methods — in short, it was a bit of a mess, and some of the functions were not even used, for example there was `RejectContactRequest` and `DismissContactRequest` and one of them was not used anywhere.


cammellos:
@alwx I have just noticed this PR, we probably need to discuss it a bit, seems like some of the features are already present, is that possible or I am misunderstanding?

alwx:
@cammellos no, you're misunderstanding — the new flow (activity center) introduces the indication of a pending outgoing contact request (which means it's saved for a user who sent the request + status are updated + it all needs to be stored & synced). That part didn't exist before because there was no indication of a pending outgoing contact request. I will update the description later and will move it all to "Ready to review" stage a bit later today once I check all the bits and pieces.

More details here: https://www.figma.com/file/eDfxTa9IoaCMUy5cLTp0ys/Shell-for-Mobile?node-id=4537%3A515620&t=YeDEcoVkqUsZNjIW-0

cammellos:
> @cammellos no, you're misunderstanding — the new flow (activity center) introduces the indication of a pending outgoing contact request (which means it's saved for a user who sent the request + status are updated + it all needs to be stored & synced). That part didn't exist before because there was no indication of a pending outgoing contact request. I will update the description later and will move it all to "Ready to review" stage a bit later today once I check all the bits and pieces.
> 
> More details here: https://www.figma.com/file/eDfxTa9IoaCMUy5cLTp0ys/Shell-for-Mobile?node-id=4537%3A515620&t=YeDEcoVkqUsZNjIW-0

Thanks @alwx 

so, yes, there's no concept of pending notifications, that needs adding.

Though your changes have added a few more things that I believe are already present:

```
protobuf.ContactRequestDecision
```

is already present and used I believe:

```
message SyncContactRequestDecision {
  uint64 clock = 1;
  string requestId = 2;
  DecisionStatus decision_status = 3;

  enum DecisionStatus {
    ACCEPTED = 0;
    DECLINED = 1;
  }
}
```
in `protocol/protobuf/pairing.proto`

Similarly I believe outgoing contact requests are already dispatched to paired devices, though there might be some issues I suppose (we should cover that we a test to verify), but we have them sent out as normal chat messages for backward compatibility, so clients should already be receiving them.
Moreover dismissing a contact request should not notify the other end for spam/security reasons etc.

Probably best if we go through it, as I think there's a bit of duplicate work, though impressive work :)

alwx:
`SyncContactRequestDecision` is used to sync the decision between paired devices of one user. The newly added `ContactRequestDecision` is used to notify a user who initiated the operation to add another user about whether it was accepted or denied (it's needed to update the pending outgoing notification). 

I wanted to re-use `SyncContactRequestDecision` for this purpose as well but quickly figured that 1) it has no `contactId` attached to it because it's basically used to distribute the information between different devices of one user; 2) it might be very confusing to use the same object for different type of stuff (might potentially lead to some ifs and complicated logic in the future.

cammellos:
> SyncContactRequestDecision is used to sync the decision between paired devices of one user. The newly added ContactRequestDecision is used to notify a user who initiated the operation to add another user about whether it was accepted or denied (it's needed to update the pending outgoing notification).

This is already present in the protocol:
`AcceptContactRequest` -> Notifies the other end that a contact request has been accepted
&
`RetractContactRequest` -> Reset the contact request flow

As mentioned, we don't notify the other end of Denied contact requests, this has been discussed with design/product for privacy/spam reasons, so that won't be needed, apologies if it was still in the designs, it should have been updated if it was there.

> I wanted to re-use SyncContactRequestDecision for this purpose as well but quickly figured that 1) it has no contactId attached to it because it's basically used to distribute the information between different devices of one user;

Yes, we don't re-use protobufs for syncing vs actions.

For achieving what's in the issue, I don't think we need to make any change to the protocol as it currently stand, the feature is only to cancel outgoing contact requests and create a notification for them (which we don't currently)

alwx:
If we don't need to notify that the request was denied then it's easier and we indeed don't need to introduce `ContactRequestDecision` — all the other changes are still valid though and getting rid of the protobuf changes is very easy in that case.

And an open question is still what to do with the pending contact request notification if it was denied by the other user — should it just remain unchanged? It changes in the design, and in order to actually change it we need to know the request status.

alwx:
@cammellos ready!

protocol/messenger_contacts.go
```diff
@@ -45,157 +45,182 @@ func (m *Messenger) AcceptContactRequest(ctx context.Context, request *requests.
 	return response, nil
```
alwx, line 47: Changes in this file are actually not too big as it might seem — better to open it all in editor to investigate.

protocol/messenger_contact_requests_test.go
```diff
@@ -1029,7 +1029,7 @@ func (s *MessengerContactRequestSuite) TestReceiveMultipleLegacy() {
 
 	// Remove contact
 
-	_, err = s.m.RetractContactRequest(&requests.RetractContactRequest{ContactID: types.Hex2Bytes(contactID)})
+	_, err = s.m.RetractContactRequest(&requests.RetractContactRequest{ID: types.Hex2Bytes(contactID)})
```
alwx, line 1041: This is done mainly to make it uniform with other similar methods.

MishkaRogachev:
On the desktop side current behaviour is:

1) _User1_ sends contact request to _User2_. _User2_ gets the notification with pending contact request. There is no notification for _User1_ currently, we should add notification for sender's side with cancel CTA.

2) _User1_ cancels sent contact request to _User2_.  We remove notification with canceled contact request for _User2_ 

3) _User2_ accepts the contact request from _User1_. Both users gets the notification with `Accepted` state.

4) _User2_ declines the contact request from _User1_. _User2_ gets the notification with `Declined` state. _User1_ does not get notified about declination.

Screenshot 2023-01-25 at 13 47 14

protocol/messenger_contact_requests_test.go
```diff
@@ -242,7 +242,7 @@ func (s *MessengerContactRequestSuite) TestReceiveAndDismissContactRequest() {
 	s.Require().Equal(contactRequests[0].ContactRequestState, common.ContactRequestStatePending)
 
 	// Dismiss contact request, receiver side
-	resp, err = theirMessenger.DismissContactRequest(context.Background(), &requests.DismissContactRequest{ID: types.Hex2Bytes(contactRequests[0].ID)})
+	resp, err = theirMessenger.DeclineContactRequest(context.Background(), &requests.DeclineContactRequest{ID: types.Hex2Bytes(contactRequests[0].ID)})
```
cammellos, line 273: We used `Dismiss` intentionally, because we thought it better conveyed the fact that we would not notify the other end, vs `Decline`, not really against the change though if you feel strongly about it

alwx:
@cammellos updated & ready!

protocol/messenger_contacts.go
```diff
@@ -45,157 +45,182 @@ func (m *Messenger) AcceptContactRequest(ctx context.Context, request *requests.
...
 	}
 
 	return response, nil
 }
 
-// NOTE: This sets HasAddedUs to false, so next time we receive a contact request it will be reset to true
-func (m *Messenger) RejectContactRequest(ctx context.Context, request *requests.RejectContactRequest) (*MessengerResponse, error) {
-	err := request.Validate()
+func (m *Messenger) cancelOutgoingContactRequest(ctx context.Context, ID string) (*MessengerResponse, error) {
+	m.logger.Info("cancelOutgoingContactRequest")
```
cammellos: Do we need this log? I'd either remove it, or add more info if necessary

cammellos (approved):
Looks good! left some minor comments, great work!
Don't forget to bump `VERSION`

protocol/messenger_handler.go
```diff
@@ -320,29 +319,42 @@ func (m *Messenger) createContactRequestNotification(contact *Contact, messageSt
...
 	}
 
+	if contactRequest != nil && contactRequest.ContactRequestState == common.ContactRequestStateAccepted {
+		// Pull one from the db if there
+		notification, err := m.persistence.GetActivityCenterNotificationByID(types.FromHex(contactRequest.ID))
+		if err != nil {
+			return err
+		}
+		if notification != nil {
+			err := m.persistence.UpdateActivityCenterNotificationMessage(notification.ID, contactRequest)
```
MishkaRogachev: FYI: In [this](https://github.com/status-im/status-go/pull/3126) PR i'm going to replace `UpdateActivityCenterNotificationMessage` calls with `SaveActivityCenterNotification `

protocol/messenger_handler.go
```diff
@@ -712,13 +725,12 @@ func (m *Messenger) HandleAcceptContactRequest(state *ReceivedMessageState, mess
 		return nil
 	}
 
-	// TODO: Handle missing contact request message
-	request, err := m.persistence.MessageByID(message.Id)
+	request, err := m.persistence.MessageByID(defaultContactRequestID(message.Id))
```
cammellos: This does not look right actually, are you sure it's what you want to do?

`defaultContactRequestID` is generally calculated over the contact.id, not the contact request id

alwx: `message.Id` basically contains a contact ID at this point (does it need to be renamed in this case? or should the ID property actually be a request ID instead?

alwx: (from my perspective, it's better and easier to be understood when `AcceptContactRequest` contains `ContactID` instead of just `ID`)

cammellos: mmh, I don't think it contains a `contact.ID`.

There are 2 kinds of contact requests:
1) Explicitly sent by the user, this one has a contact request id == to the message id sent
2) Default contact requests, these were added for backward compatibility / cases where the user's device data is lost. Those have an id derived by the contact.id (`defaultContactRequestID(contact.id))`.

But when you accept it, as far as I can tell, you always use the ID you received, and I don't believe we ever derive the id from `message.Id`, so I'd expect the line you added:

```
m.persistence.MessageByID(defaultContactRequestID(message.Id))
```
To never return any message, if I am not mistaking

alwx: You're a bit mistaken — the `protobuf.AcceptContactRequest` was created with contactID. I changed it in the last commit though to make it a bit more clear.

cammellos: are you sure?

https://github.com/status-im/status-go/blob/5d80c5cc427cd5979a60122bd725e1380a5cfa55/protocol/messenger_contacts.go#L240

https://github.com/status-im/status-go/blob/5d80c5cc427cd5979a60122bd725e1380a5cfa55/protocol/messenger_contacts.go#L909

both cases where it's created it seems to be using the id of the contact request (which makes sense), where did you see that it was using the contact id?

alwx: But it all has been changed in this PR — now it sends back the current contact ID just because that's what actually needs to be used in order to update the local pending notification request for the user that requested the contact to be added.
It does work perfectly.

cammellos: did you revert the change?
it would have broken compatibility between clients btw, so wasn't quite working perfectly ;)

alwx: yes, it's now `m.persistence.MessageByID(message.Id)` as it was before

protocol/messenger_handler.go
```diff
@@ -262,7 +262,27 @@ func (m *Messenger) PendingNotificationContactRequest(contactID string) (*Activi
...
+		// Pull one from the db if there
+		notification, err := m.persistence.GetActivityCenterNotificationByID(types.FromHex(contactRequest.ID))
+		if err != nil {
+			return err
+		}
+
+		if notification != nil {
+			notification.Message = contactRequest
+			notification.Read = false
+			notification.Accepted = true
```
MishkaRogachev, line 276: According new AC logic it should always sync contactRequest  and notification state
```
notification.Accepted = contactRequest.ContactRequestState == common.ContactRequestStateAccepted
notification.Dismissed = contactRequest.ContactRequestState == common.ContactRequestStateDismissed
```

alwx:
Regarding number 4: I can indeed make it tappable but I said it's out of scope because of this comment: https://github.com/status-im/status-mobile/pull/14853/files#r1086651896

alwx:
This issue was pending for a while — needed to work on other issues instead. Now coming back to it — rebasing and making sure tests work.

PR #3121, opened 2023-01-22
Fix image fields

### LINKED ISSUE #3110: imageWidth, imageHeight, albumID fields missing in the receiver's message
These 2 PRs https://github.com/status-im/status-go/pull/3061, https://github.com/status-im/status-go/pull/3021 adds `imageWidth`, `imageHeight` and `albumID` fields to the `Message`. These fields are currently included in the response that the sender gets after sending a message, but are missing in the message that the receiver gets.

OmarBasem:
Fixes: https://github.com/status-im/status-go/issues/3110

This PR adds missing fields (imageWidth, imageHeight, albumID) in the delivered message.

Status: ready


protocol/common/message.go
```diff
@@ -244,7 +244,7 @@ func (m *Message) MarshalJSON() ([]byte, error) {
 		EnsName                  string                           `json:"ensName"`
 		DisplayName              string                           `json:"displayName"`
 		Image                    string                           `json:"image,omitempty"`
-		AlbumID                  string                           `json:"albumId,omitempty"`
+		ImageAlbumID             string                           `json:"imageAlbumId,omitempty"`
```
0x-r4bbit: This looks like a breaking change. Just raising awareness that this could break clients that don't expect `imageAlbumId` yet

protocol/message_persistence.go
```diff
@@ -68,7 +68,7 @@ func (db sqlitePersistence) tableUserMessagesAllFields() string {
 		sticker_hash,
 		image_payload,
 		image_type,
-		album_id,
+		image_album_id,
```
0x-r4bbit: This is a breaking change as well.
We need to make sure we have a migration in place that moves previous `album_id`s to `image_album_id`s

protocol/migrations/sqlite/1670921937_add_album_id.up.sql
```diff
@@ -1 +0,0 @@
-ALTER TABLE user_messages ADD COLUMN album_id VARCHAR;
```
0x-r4bbit: I don't think we should remove any existing migrations. Instead we need a migration that moves album_id s over to the new column after it has been created

0x-r4bbit (changes requested):
Changes LGTM so far, but the migrations need to be adjusted IMO

protocol/migrations/sqlite/1670921937_add_album_id.up.sql
cammellos: Yes, this will break the app on some devices, migrations should never be removed

protocol/message_persistence.go
cammellos: @OmarBasem  if this is only a name change, I think we should just keep the old one in the database field, leave a comment if you wish, but migrations can be expensive and error prone, and still, we would have a dangling field since columns cannot be removed. Not worth the effort/risk in my opinion.

OmarBasem: @cammellos it is not just a name change. There is an albumId field on the Message struct and the ChatMessage struct (protobuf). Initially, both had the same name albumId, but then it was not working. I think they need to have different names, otherwise they would clash and the field would be nil (I think that's why it was not being delievered). It did not work until I changed the name on the Message struct to be ImageAlbumId.

cammellos: > @cammellos it is not just a name change. There is an albumId field on the Message struct and the ChatMessage struct (protobuf). Initially, both had the same name albumId, but then it was not working. I think they need to have different names, otherwise they would clash and the field would be nil (I think that's why it was not being delievered). It did not work until I changed the name on the Message struct to be ImageAlbumId.

That's a separate issue, we are only talking about the field in the database here, `image_album_id`, and that doesn't need to match the field in the structure.

In your case for example, is perfectly fine keeping `album_id` as a field in the database, but deserialize it into a field named `ImageAlbumID`

OmarBasem: Okay got it!

OmarBasem:
> Changes LGTM so far, but the migrations need to be adjusted IMO

Reverted the database migrations

OmarBasem:
I am wondering at what point does the snake-cased fields read from the database becomes camel case? Is it inside the marshal / unmarshal methods?

0x-r4bbit:
> I am wondering at what point does the snake-cased fields read from the database becomes camel case? Is it inside the marshal / unmarshal methods?

Correct, notice how you specify the camelCase names in the struct annotations.

```go
type Foo struct {
  Bar  string `json:"bar,omitempty"` // you can replace `bar` with anything really
}

OmarBasem:
> > I am wondering at what point does the snake-cased fields read from the database becomes camel case? Is it inside the marshal / unmarshal methods?
> 
> Correct, notice how you specify the camelCase names in the struct annotations.
> 
> ```go
> type Foo struct {
>   Bar  string `json:"bar,omitempty"` // you can replace `bar` with anything really
> }
> ```

@0x-r4bbit oh okay, thanks! I think all is fixed now.

richard-ramos (commented):
This PR LGTM, but ideally part of the commit history should be squashed and have its description reworded with a proper description instead of  the text `updates`, and it shouldn't include files that were not really modified by this PR  like some of the migrations.go / bindata.go / *.pb.go files, which while they were affected by `make generate`m the changes done by the command are not really meaningful (some timestamps updates and spaces automatically removed from comments)

OmarBasem:
> This PR LGTM, but ideally part of the commit history should be squashed and have its description reworded with a proper description instead of the text `updates`, and it shouldn't include files that were not really modified by this PR like some of the migrations.go / bindata.go / *.pb.go files, which while they were affected by `make generate`m the changes done by the command are not really meaningful (some timestamps updates and spaces automatically removed from comments)

Hi @richard-ramos, apologies for the bad commit history. Squashed the commits, and will make sure to have a proper commit message before merging. 

I did a change to a proto file, so I ran `make generate` to update the file. Is there a way to have `make generate` run for a specific file?

0x-r4bbit:
> I did a change to a proto file, so I ran make generate to update the file. Is there a way to have make generate run for a specific file?

Sadly no. You'll have to ensure you only put the changes necessary under version control.
I typically test this by only committing the bindata changes I think were needed, then resetting the repo to remove all other changes created by `make generate` and then re-run an tests on a clean database

OmarBasem:
@0x-r4bbit I think all is fixed now

0x-r4bbit (approved):

0x-r4bbit:
Well done @OmarBasem

PR #3122, opened 2023-01-23
chore: allow GetMembers for the communty by communityID only

mprakhov:
Make GetMembers API return community members if only communityID provided (without chatID)

Required for:
- https://github.com/status-im/status-desktop/issues/9090
- https://github.com/status-im/status-desktop/pull/9230

### DIFF
--- a/services/chat/api.go
+++ b/services/chat/api.go
@@ -19,6 +19,7 @@ import (
 
 var (
 	ErrChatNotFound            = errors.New("can't find chat")
+	ErrCommunityNotFound       = errors.New("can't find community")
 	ErrCommunitiesNotSupported = errors.New("communities are not supported")
 	ErrChatTypeNotSupported    = errors.New("chat type not supported")
 )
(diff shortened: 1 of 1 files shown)


services/chat/api.go
```diff
@@ -408,14 +435,11 @@ func (api *API) getChatAndCommunity(pubKey string, communityID types.HexBytes, c
...
-		if err != nil {
-			return nil, nil, err
-		}
+		var err error
+		community, err = api.getCommunityByID(messengerChat.CommunityID)
 
-		community, err = api.s.messenger.GetCommunityByID(communityID)
 		if err != nil {
-			return nil, nil, err
+			return nil, nil, ErrCommunityNotFound
```
jrainville: I think we should only return `ErrCommunityNotFound` if we know for sure that was the issue. I think that's what @cammellos would say :laughing:

jrainville (commented):
Looks good. I just think there might be something with the error return

services/chat/api.go
mprakhov: Refactored to return `ErrCommunityNotFound` only if we know the community was not found. For other cases return an occurred error

P.S. Hmmm... maybe we should return this error during the DB request (from protocol/communities/persistence.go)... but from another point of view, no data is not an error from persistence.go side

jrainville (approved):

PR #3126, opened 2023-01-24
Mark as read acivated notifications with CTA

MishkaRogachev:
For: https://github.com/status-im/status-desktop/pull/9226

- [x] Accepting & dismissing contact requests
- [x] Accepting & rejecting community requests
- [x] Accepting & rejecting verification requests


MishkaRogachev:
@cammellos @ilmotta there are a few questions i would like to ask:
1) I find some functions  from `activity_center_persistance.go` confusing: `UpdateActivityCenterNotificationMessage`, 
`UpdateActivityCenterNotificationFields` and `UpdateActivityCenterNotificationContactVerificationStatus`. They are not used any more because we are updating whole notification object.  Can i remove them?
2) This PR fix behaviour of new notifications. But we still have old notifications with finished CTA and read = false. Should i create the migration to update them?
3) I've added correct `Read` flag for contact, community membership and verification requests notifications. Should i also add `Accepted` and `Dismissed` flags?

protocol/messenger_contact_verification.go
```diff
@@ -357,10 +355,12 @@ func (m *Messenger) AcceptContactVerificationRequest(ctx context.Context, id str
 		message := notification.Message
 		message.ContactVerificationState = common.ContactVerificationStateAccepted
 		notification.ReplyMessage = replyMessage
+		notification.Read = true
 
-		err := m.persistence.UpdateActivityCenterNotificationFields(notification.ID, message, replyMessage, verification.RequestStatusACCEPTED)
-		if err != nil {
-			return nil, err
+		saveErr := m.persistence.SaveActivityCenterNotification(notification)
```
cammellos: we generally just call it `err`, since you check it, and then either is `nil` and everyone is happy, or you just return it

ilmotta:
Hey @MishkaRogachev:

> 1. I find some functions  from `activity_center_persistance.go` confusing: `UpdateActivityCenterNotificationMessage`, `UpdateActivityCenterNotificationFields` and `UpdateActivityCenterNotificationContactVerificationStatus`. They are not used any more because we are updating whole notification object.  Can i remove them?

I thought our golangci-lint rules would catch unused functions. I did a checkout in your branch to test this, but `golangci-lint` reported nothing. I couldn't find any usages, so I guess we should/can delete them.

>This PR fix behaviour of new notifications. But we still have old notifications with finished CTA and read = false. Should i create the migration to update them?

I would say so, but I'm not 100% sure.

>I've added correct Read flag for contact, community membership and verification requests notifications. Should i also add Accepted and Dismissed flags?

I'm not sure I understand what do you mean by "add Accepted and Dismissed flags".

MishkaRogachev:
@cammellos @ilmotta UPD:
1. Fixed
2. Old notifications can be marked as read with `MarkAllActivityCenterNotificationsRead`, no extra need for migration
3. We have `Accepted`  and `Dismissed` falgs in notification object. But those field are not used (yet). My suggestion is to fill them in same place i'm setting read flag in this PR

protocol/messenger_contacts.go
```diff
@@ -170,11 +170,14 @@ func (m *Messenger) dismissContactRequest(requestID string, syncing bool) (*Mess
...
-		err := m.persistence.UpdateActivityCenterNotificationMessage(notification.ID, contactRequest)
-		if err != nil {
-			return nil, err
-		}
 		notification.Message = contactRequest
+		notification.Read = true
+
+		saveErr := m.persistence.SaveActivityCenterNotification(notification)
+		if saveErr != nil {
+			m.logger.Warn("failed to save notification", zap.Error(saveErr))
```
ilmotta: I'm wondering if we should log as `Error` instead of `Warn`. It seems to me the errors that can happen inside `SaveActivityCenterNotification` should rarely happen, like marshaling a message, but when they do it's because something weird is going on, so better report as error. Wdyt?

protocol/messenger_handler.go
```diff
@@ -351,6 +352,7 @@ func (m *Messenger) createContactRequestNotification(contact *Contact, messageSt
 		Author:    messageState.CurrentMessageState.Contact.ID,
 		Timestamp: messageState.CurrentMessageState.WhisperTimestamp,
 		ChatID:    contact.ID,
+		Read:      contactRequest.ContactRequestState == common.ContactRequestStateAccepted || contactRequest.ContactRequestState == common.ContactRequestStateDismissed,
```
ilmotta, line 355: >We have Accepted and Dismissed falgs in notification object. But those field are not used (yet). My suggestion is to fill them in same place i'm setting read flag in this PR

I think so :+1: For instance, a couple lines above and we see the notification might be dismissed in the DB, but then here we build a new struct instance that doesn't have the `Dismissed` field. To be consistent with what's in the DB, I think we should also define the other fields.

ilmotta (approved):

protocol/messenger_handler.go
MishkaRogachev: Working on it. And one thing is not clear for me:

When we accept or decline contact request we should mark linked notification as Read and Accepted?

Or for accepting contact request we should mark notification as Read and Accepted and for declining contact request we should mark notification as Read and Dismissed?

For me second way is closer, but we have filtering based on Dismissed flag in activity_center_persistance

ilmotta: Just for the record, @MishkaRogachev and I discussed outside Github what exactly are these dismissed/accepted fields, because up until this point, both Desktop and Mobile apps relied exclusively on the notification states. For example, to know if a contact request has been `accepted`, clients use the `contactRequestState` in the RPC response. https://github.com/status-im/status-go/blob/develop/protocol/common/message.go#L58

protocol/messenger_contacts.go
MishkaRogachev: fixed

protocol/messenger_handler.go
MishkaRogachev: I've added dismissed and accepted with latest commit

cammellos (approved):

protocol/messenger_contact_verification.go
MishkaRogachev: fixed

PR #3127, opened 2023-01-24
fix: don't return an error on empty rows

richard-ramos:


saledjenic (approved):

0x-r4bbit (approved):
omg finally :D
Thank you @richard-ramos

PR #3128, opened 2023-01-24
Bug/add contact request local clock

cammellos:
This PR changes the way we handle contact requests:

`added` field, is moved to `contactRequestLocalState`
`hasAddedUs` field is moved to `contactRequestRemoteState`

We added clocks for both fields, since before we had only one, but it doesn't work for syncing contacts, since it needs its own clock, otherwise we can't really understand whether we are up to date with any of the two fields, leading to potentially inconsistent state between devices that can't be recovered from.

This is a non breaking change with regards of clients, since the external json hasn't changed (`added`, `hasAddedUs`, `contactRequestState` are still present).

I have removed an unused function as well.

It also fixes a couple of issues with notifications, though there's still some work to do there. 
I have also started encapsulating logic with the `Contact` struct, as it was starting to get a bit hairy around.

I have preserved for now the commit history for clarity, apologies for the large PR.


jrainville (approved):
Huge work. LGTM

protocol/contact_test.go
```diff
@@ -0,0 +1,544 @@
...
+			expectedHasAddedUs:  true,
+			expectedMutual:      true,
+		},
+	}
+
+	for testNum, tc := range tests {
+		contact := tc.Contact()
+
+		contact.ContactRequestSent(clock)
+		validateContactTest(t, contact, tc, testNum+1)
```
ilmotta, line 125: [Nitpick] Zero-based index is more obvious to me to look up which test failed.

protocol/contact_test.go
```diff
@@ -0,0 +1,544 @@
...
+			actualRemoteState:   ContactRequestStateReceived,
+			expectedLocalState:  ContactRequestStateSent,
+			expectedRemoteState: ContactRequestStateReceived,
+			expectedAdded:       true,
+			expectedHasAddedUs:  true,
+			expectedMutual:      true,
+		},
+	}
+
+	for testNum, tc := range tests {
```
ilmotta, line 191: Just for fun you could go a bit further to run a slice of tests given it's always the same steps.

```go
func runTests(t *testing.T, tests []contactTest, action func(*Contact)) {
	for index, tc := range tests {
		contact := tc.Contact()
		action(&contact)
		validateContactTest(t, contact, tc, index+1)
	}
}

// Then in each test
runTests(t, tests, func(c *Contact) {
	c.RetractContactRequest(clock)
})
```

protocol/persistence.go
```diff
@@ -561,21 +561,21 @@ func (db sqlitePersistence) Contacts() ([]*Contact, error) {
...
+			nickname                  sql.NullString
+			contactRequestLocalState  sql.NullInt64
+			contactRequestLocalClock  sql.NullInt64
+			contactRequestRemoteState sql.NullInt64
+			contactRequestRemoteClock sql.NullInt64
+			displayName               sql.NullString
+			imageType                 sql.NullString
+			ensName                   sql.NullString
+			ensVerified               sql.NullBool
+			blocked                   sql.NullBool
```
ilmotta, line 574: Just curious: why do we allow boolean fields to be null instead of always making sure its written as true/false?

protocol/migrations/sqlite/1674210659_add_contact_request_local_clock.up.sql
```diff
@@ -0,0 +1,16 @@
+
+ALTER TABLE contacts ADD COLUMN contact_request_local_clock INT;
```
ilmotta, line 2: @cammellos, what's your opinion for this particular case, about having a separate table for tracking the state of contact requests instead of adding more fields to the `contacts` table? I'm not suggesting a change, just wondering if some parts of the `contacts` table should be normalized.

protocol/messenger_handler.go
```diff
@@ -1671,6 +1717,7 @@ func (m *Messenger) HandleChatMessage(state *ReceivedMessageState) error {
 		}
 	}
 
+	// TODO(cammellos): This is unclear when the contact is ourselves
```
ilmotta: @cammellos, in the mobile repository we set some guidelines for adding `TODO` comments to help us better track the work left to be done. In summary, over there we prefer to create Github issues because they tend to help collaborators describe what needs to be done in more detail, whereas `TODO` comments tend to be superficial, like this one.

What's the preference in `status-go`?

ilmotta (approved):
Great work! It's much clear now, especially due to the unit tests. My comments are more like questions than anything :rocket:

protocol/contact_test.go
cammellos: That's true, I started using 0 index, then changed to be 1 based :)

protocol/persistence.go
cammellos: For bools, I think it's probably ok to not have null fields, I think sqlite handles this values fine and they will always be defaulted to some non-null valu.

For strings for example, they would be `nil` in the database on migrations when adding a column (these are all columns that have been added after the initial table was created), and we don't want to set those fields on a migration, since it might take a bit longer to migrate on some potentially large collections (as `contacts` might be), which leads to longer login time after an upgrade.

Therefore if we are not writing or defaulting these fields, we use nullvalues for those.

protocol/migrations/sqlite/1674210659_add_contact_request_local_clock.up.sql
cammellos: Not sure, operationally is a bit more complex since you have to deal with JOINs etc, so I tend to prefer to add to the `contacts` table, but I am not really an expert on sql database design, I don't have a strong opinion on the matter myself.

protocol/messenger_handler.go
cammellos: no preference :) probably issues are better I guess, I can move it to an issue

richard-ramos (approved):
LGTM, the only change I'd make is to not include generated files that were not really affected by this change. (maybe there's a way to get bindata and protoc to not update the timestamp of the files?)

protocol/migrations/sqlite/1674210659_add_contact_request_local_clock.up.sql
ilmotta: Cool, no strong preference myself, just leaning a bit more on normalization. It's helpful for avoiding duplicate data across tables for example, but I don't even know if this is a problem in `status-go`, probably not. There are other benefits that could outweigh the increased complexity in queries. I'm no expert either. Thanks Andrea

protocol/persistence.go
ilmotta: Thanks, quite informative

PR #3130, opened 2023-01-24
Fix activation of deleted chat on old contact update

rasom:
needed for  https://github.com/status-im/status-mobile/issues/14669

status: ready


jrainville (approved):
Nice find

protocol/messenger_handler.go
```diff
@@ -837,7 +837,7 @@ func (m *Messenger) HandleContactUpdate(state *ReceivedMessageState, message pro
 		chat.LastClockValue = message.Clock
 	}
 
-	if contact.ContactRequestState == ContactRequestStateMutual {
+	if contact.ContactRequestState == ContactRequestStateMutual && chat.DeletedAtClockValue < message.Clock {
```
ilmotta, line 840: It would be nice to update test(s) to cover this particular condition.

ilmotta (approved):

protocol/messenger_handler.go
rasom: Yep, it would be. Sadly though there are no tests to update, and I'm not into writing new ones for this particular if clause.

PR #3131, opened 2023-01-25
allow building shared library on M1 Macs

jakubgs:
Seems like this wasn't supported in the past, but should work now.
```
 % make statusgo-shared-library GOBIN_SHARED_LIB_CFLAGS='GOOS=darwin GOARCH=arm64'
...
Shared library built:
-rw-r--r--  1 admin  admin  81850034 Jan 25 14:57 /Users/admin/status-go/build/bin/libstatus.dylib
-rw-r--r--  1 admin  admin      6658 Jan 25 14:57 /Users/admin/status-go/build/bin/libstatus.h
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,14 +17,10 @@ endif
 
 ifeq ($(detected_OS),Darwin)
  GOBIN_SHARED_LIB_EXT := dylib
-  # Building on M1 is still not supported, so in the meantime we crosscompile by default to amd64
-  ifeq ("$(shell sysctl -nq hw.optional.arm64)","1")
-    FORCE_ARCH ?= amd64
-    GOBIN_SHARED_LIB_CFLAGS=CGO_ENABLED=1 GOOS=darwin GOARCH=$(FORCE_ARCH)
-  endif
+ GOBIN_SHARED_LIB_CFLAGS := CGO_ENABLED=1 GOOS=darwin
 else ifeq ($(detected_OS),Windows)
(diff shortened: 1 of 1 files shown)


ghost:
### Pull Request Checklist

flexsurfer (approved):

richard-ramos (approved):

jakubgs:
Looks like all builds for desktop work fine:

* https://github.com/status-im/status-desktop/pull/9291

So I'm merging this.

stefandunca:
> Looks like all builds for desktop work fine:

@jakubgs the CI build passes because we don't have a Jenkins node building on apple silicon. This PR breaks builds of `status-desktop` on M1 machines. Could you please revert it until we move to Qt5.15 and have the option to build natively for M1? Thanks

jakubgs:
How does it break it exactly? What Go version are you on, are you running in a rosetta shell? What's the error?

stefandunca:
> How does it break it exactly? What Go version are you on, are you running in a rosetta shell? What's the error?

It is a linking error not a runtime: "Undefined symbols for architecture `x86_64`".

The reason why do we need to cross-compile on apple silicon architecture is because Qt14.x doesn't provides arm64 binaries for the platform. Hence we force the entire app to build for x86 and all the dependencies have to be on x86 to link properly.

Only the C++ POC builds and links with status-go on arm64 on apple silicon. Hence without this change you can always force an arm64 by calling `FORCE_ARCH=arm64 make statusgo-shared-library`

If you insist to keep the change we have to move the check for apple silicon platform into the make file of the desktop app, which might be a proper way.

PR #3134, opened 2023-01-26
refactor(import-tool): process import data in chunks

0x-r4bbit:
This commit refactors the discord import tool such that, instead of loading all data to be imported into memory at once, it will now perform the import on a per file basis.

This improves the memory pressure for the node performing the import and seems to increase its performance as well.

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -504,7 +504,7 @@ func (m *Manager) ImportCommunity(key *ecdsa.PrivateKey) (*Community, error) {
 	return community, nil
 }
 
-func (m *Manager) CreateChat(communityID types.HexBytes, chat *protobuf.CommunityChat, publish bool) (*Community, *CommunityChanges, error) {
+func (m *Manager) CreateChat(communityID types.HexBytes, chat *protobuf.CommunityChat, publish bool, thirdPartyID string) (*Community, *CommunityChanges, error) {
 	community, err := m.GetByID(communityID)
 	if err != nil {
 		return nil, nil, err
@@ -513,6 +513,10 @@ func (m *Manager) CreateChat(communityID types.HexBytes, chat *protobuf.Communit
 		return nil, nil, ErrOrgNotFound
 	}
 	chatID := uuid.New().String()
+	if thirdPartyID != "" {
+		chatID = chatID + thirdPartyID
+	}
+
 	changes, err := community.CreateChat(chatID, chat)
 	if err != nil {
 		return nil, nil, err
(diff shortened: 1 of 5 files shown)


protocol/communities/manager.go
```diff
@@ -598,7 +602,11 @@ func (m *Manager) CreateCategory(request *requests.CreateCommunityCategory, publ
 	if community == nil {
 		return nil, nil, ErrOrgNotFound
 	}
+
 	categoryID := uuid.New().String()
+	if request.ThirdPartyID != "" {
+		categoryID = categoryID + request.ThirdPartyID
+	}
```
0x-r4bbit, line 609: @cammellos as we now process the import in multiple chunks/files, there's a possibility that we'll run into files that reference an already existing chat or category.

In order to check whether or not a given category/or channel already exists, we attach the discord category/chat ID to ours and then enable ourselves to check if any existing category/chat has a certain suffix.

Chat and category IDs are otherwise random IDs, so it's hard to check if we already have a chat/category otherwise (name alone isn't enough)

protocol/messenger_communities.go
```diff
@@ -2437,129 +2414,178 @@ func (m *Messenger) RequestImportDiscordCommunity(request *requests.ImportDiscor
...
-			return
-		}
+				categories := discordCommunity.Categories()
+				exists := false
+				for catID := range categories {
+					if strings.HasSuffix(catID, category.ID) {
+						exists = true
+						break
+					}
+				}
```
0x-r4bbit, line 2458: @cammellos here's how we check if a given category was already created during the import

protocol/messenger_communities.go
```diff
@@ -2437,129 +2414,178 @@ func (m *Messenger) RequestImportDiscordCommunity(request *requests.ImportDiscor
...
+			channel := exportData.ExportedData[0]
+			chatIDs := discordCommunity.ChatIDs()
+
+			exists := false
+			for _, chatID := range chatIDs {
+				if strings.HasSuffix(chatID, channel.Channel.ID) {
+					exists = true
+					break
+				}
+			}
```
0x-r4bbit, line 2515: @cammellos here's how we check for any chat that has already been created during the import

protocol/messenger_communities.go
```diff
@@ -2732,220 +2721,296 @@ func (m *Messenger) RequestImportDiscordCommunity(request *requests.ImportDiscor
...
+						}
+
+						assetCounter.Increase()
+						progressValue := calculateProgress(i+1, totalImportChunkCount, 0.25+(float32(assetCounter.Value())/float32(totalAssetsCount))*0.25)
+						importProgress.UpdateTaskProgress(discord.DownloadAssetsTask, progressValue)
+						progressUpdates <- importProgress
+					}
+				}(attachments)
+			}
+			wg.Wait()
```
0x-r4bbit, line 2919: While this looks like a new thing, it's just been moved around as we now put the import process into a loop (on a per file basis)

0x-r4bbit (commented):
@cammellos @richard-ramos this is a somewhat bigger PR, but really only because I've moved a big portion of the import logic into a loop, so that we can perform imports on a per file basis (instead of loading everything into memory at once)

I've added comments to the parts that you might want to double-check

richard-ramos (approved):

PR #3136, opened 2023-01-26
Fix incorrect pagination of notifications

rasom:
needed for https://github.com/status-im/status-mobile/issues/14173

So basically we have `HexBytes` in `id` field and to avoid some weird encoding/decoding issues it might be better to convert it to hex number before concatenation. 

status: ready


ilmotta (approved):
It seems there's an inverse relationship between the weirdness of a bug and the amount of characters to fix it :rocket:

PR #3144, opened 2023-01-30
feat: convert history balance tokens to fiat

### LINKED ISSUE #9279: Fetch historical currency and convert token count to default FIAT
Considerations
- [ ] Start balance history update process on WalletStart; [see comment](https://github.com/status-im/status-go/pull/3071#discussion_r1082221663).

stefandunca:
Add `history.exchange` sub-package with the following responsibilities.

- Fetch and cache daily exchange rates for tokens - partial update if missing data starting from yesterday (going backward)
- Fetches the price of the token after merging entries for the selected chains
- Implements all-time fetching particular case

`history. service` package changes

- Return `ValuePoint` instead of `DataPoint` - contains the value in fiat as float64 instead
- Convert merged values from tokens balance (wei) to fiat

Other changes

- Move Cryptocompare implementation from `wallet` to `thirdparty` package to avoid recursive import
- Rename `DataPoint.Value` to `DataPoint.Balance`
- Move start/stop balance history to startWallet/stopWallet
- Don't merge entries for one chain and preserve original data

Closes [#9279](https://github.com/status-im/status-desktop/issues/9279)

### DIFF
--- a/services/wallet/history/exchange.go
+++ b/services/wallet/history/exchange.go
@@ -1,0 +1,176 @@
+package history
+
+import (
+	"errors"
+	"sync"
+	"time"
+
+	"github.com/status-im/status-go/services/wallet/thirdparty"
+)
+
+type tokenType = string
+type currencyType = string
+type yearType = int
+
+type allTimeEntry struct {
+	value          float32
+	startTimestamp int64
+	endTimestamp   int64
+}
+
+// Exchange caches conversion rates in memory on a daily basis
+type Exchange struct {
+	// year map keeps a list of values with days as index in the slice for the corresponding year (key) starting from the first to the last available
+	cache map[tokenType]map[currencyType]map[yearType][]float32
+	// special case for all time information
+	allTimeCache map[tokenType]map[currencyType][]allTimeEntry
+	fetchMutex   sync.Mutex
+
+	cryptoCompare *thirdparty.CryptoCompare
+}
+
+func NewExchange(cryptoCompare *thirdparty.CryptoCompare) *Exchange {
+	return &Exchange{
+		cache:         make(map[tokenType]map[currencyType]map[yearType][]float32),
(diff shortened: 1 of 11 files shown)


services/wallet/history/exchange.go
```diff
@@ -0,0 +1,174 @@
...
+
+	daysToFetch := extendDaysSliceForYear(yearsMap, endOfPrevYearTime)
+
+	curYearTime := time.Date(currentTime.Year(), currentTime.Month(), currentTime.Day(), 0, 0, 0, 0, time.UTC)
+	daysToFetch += extendDaysSliceForYear(yearsMap, curYearTime)
+	if daysToFetch == 0 {
+		return nil
+	}
+
+	onlineSource := thirdparty.NewCryptoCompare()
```
alaibe: could you inject crypocompare, rather than building a new one, it is not stateless and help to know if the api is up or down

services/wallet/history/exchange.go
```diff
@@ -0,0 +1,174 @@
...
+		t := time.Unix(res[i].Timestamp, 0).UTC()
+		yearDayIndex := t.YearDay() - 1
+		yearValues, found := yearsMap[t.Year()]
+		if found && yearDayIndex < len(yearValues) {
+			yearValues[yearDayIndex] = float32(res[i].Value)
+		}
+	}
+
+	// Fetch all time
+	allTime, err := onlineSource.FetchDailyMarketValues(token, currency, 1, true, 30)
```
alaibe: is comment correct? the 1 correspond to only one day?

stefandunca: This is a particular call, the `1` i got from the token value implementation. The next `true` value corresponds to "all time", which is actually fetching from 2010 onwards (hence ignoring the 0 values later on). The 30 is to aggregate every 30 days (maximum available)

alaibe (approved):

PR #3147, opened 2023-01-30
Fix for syncing settings (empty field is not getting added)

### LINKED ISSUE #14891: null is not an object (evaluating 'n.name') when syncing desktop app with mobile on initial sync
# Bug Report

## Problem 
After initial syncing (using flow Sync setting > devices > Sync all) certain data causes the error `null is not an object (evaluating 'n.name') `
Couldn't reproduce with a fresh multi-account with a couple of contacts, so not sure what exact data may be the reason for the failure

#### Expected behavior
no error

#### Actual behavior
IMAGE 2023-01-25 16:43:20
Issue re-appears from time to time on mobile device

### Reproduction

1) Mobile: restore multi account that you're using on status desktop, go to sync settings > devices > advertise device
2) Desktop: go to settings > devices, enable sync with mobile
3) Desktop: tap on Sync all

### Additional Information

- Status version: nightly 25/01/23
- Desktop version: 0.9.0 rc7
- Logs: 
[status_logs.zip](https://github.com/status-im/status-mobile/files/10500912/status_logs.zip)

alwx:
Fixes sync by adding a small check to make sure empty field is not getting added to `messageState.Response.Settings`. 

One of the common reasons of it being added is actually hidden in `extractSyncSetting` function in `messsenger_sync_settings.go` — it might return `nil, nil` when the new clock value is older than the current clock value.

Closes https://github.com/status-im/status-mobile/issues/14891


cammellos (approved):
bump minor version, otherwise looks great!

qfrank (approved):
👍

PR #3148, opened 2023-01-31
feat: Add seen/unseen activity center setting

MishkaRogachev:
This is my proposition for storing the state of bell icon (seen/unseen). I'll add migrations after we decide this approach has the right to life


protocol/activity_center_persistence.go
```diff
@@ -135,6 +135,11 @@ func (db sqlitePersistence) SaveActivityCenterNotification(notification *Activit
 		notification.Accepted,
 		notification.Dismissed,
 	)
+
+	if err == nil {
+		err = db.MarkAsUnseenActivityCenterNotifications()
```
ilmotta: Is this second update in the DB running in a separate transaction? It would be better for consistency if we could do all the updates in a single transaction.

services/ext/api.go
```diff
@@ -1086,6 +1086,14 @@ func (api *PublicAPI) UnreadAndAcceptedActivityCenterNotificationsCount() (uint6
 	return api.service.messenger.UnreadAndAcceptedActivityCenterNotificationsCount()
 }
 
+func (api *PublicAPI) HasUnseenActivityCenterNotifications() (bool, error) {
```
ilmotta, line 1085: I'm thinking if it wouldn't be better to also publish a signal when the `seen` state changes. Otherwise clients need to be tightly coupled with the backend. If at any point the backend decides to change the `seen` state in other flows/places, all clients will need to be synchronized to also do the same. State synchronization is hard, so maybe it'd be better if we could reduce the chance for such bugs in the first place.

protocol/activity_center_persistence.go
```diff
@@ -857,3 +862,19 @@ func (db sqlitePersistence) RemoveAllContactRequestActivityCenterNotifications(c
 	`, chatID, ActivityCenterNotificationTypeContactRequest)
 	return err
 }
+
+func (db sqlitePersistence) HasUnseenActivityCenterNotifications() (bool, error) {
+	var hasUnseen bool
+	err := db.db.QueryRow(`SELECT has_unseen FROM activity_center_settings`).Scan(&hasUnseen)
```
ilmotta: The table name `activity_center_settings` is confusing to me, but maybe it's just me. The word `settings` in Status reminds me of *personal application settings*. Here the *seen* state is not really a choice of settings, but rather a visibility state. I don't have any suggestion now.

protocol/activity_center_persistence.go
```diff
@@ -857,3 +863,19 @@ func (db sqlitePersistence) RemoveAllContactRequestActivityCenterNotifications(c
...
+	return hasUnseen, err
+}
+
+func (db sqlitePersistence) MarkAsSeenActivityCenterNotifications() error {
+	_, err := db.db.Exec(`UPDATE activity_center_settings SET has_unseen = 0`)
```
ilmotta: I find it easier to comprehend code without negation in the words, so instead of `unseen`, use just `seen`. It helps avoid double-negation as well, like `!seen` is easier for most people to comprehend than `!unseen`. Maybe not the best example here, but it's a good practice to follow nonetheless.

protocol/activity_center_persistence.go
MishkaRogachev: Yep, i'll improve it

protocol/activity_center_persistence.go
MishkaRogachev: What about `activity_center` or `activity_center_states`?

ilmotta: :+1: I like the explicitness of `activity_center_states`, very clear purpose.

protocol/activity_center_persistence.go
MishkaRogachev: Sure, ill fix it

services/ext/api.go
ilmotta: @MishkaRogachev, @cammellos, what are your thoughts on my previous comment? Downsides/upsides? I generally prefer if we could write clients that don't need to constantly pull data from `status-go`, except for an initial data load.

MishkaRogachev: @ilmotta I temporary switched for an urgent issue in status-desktop and i need to finish some experiment so that I can form an opinion. Initially I've planed to use existing signal about new notification and call `HasUnseenActivityCenterNotifications` only when activity center popup is closed

ilmotta: Totally understandable :mechanical_arm:

MishkaRogachev: @ilmotta i've tested it [here](https://github.com/status-im/status-desktop/pull/9479) and i think there is no need for an additional signal, at least for desktop. I call `unreadActivityCenterNotificationsCount` and `hasUnseenActivityCenterNotifications` on start and when new notifications arrive. When i close activity center popup i also call `markActivityCenterNotificationRead` and `hasUnseenActivityCenterNotifications` to double check state saved.

cammellos: so, the way I would naturally code it on status-react is:

1) Have and endpoint that returns the has-seen state (as per PR)
2) Everytime a notification arrives as a signal, the client will check 
```
is the user on the activity center?
 no -> mark as seen state to false in the client state 
 yes -> don't update ui, mark as seen to true
```

The backend cannot fully drive this piece of state since it's contextual to where the user is.

There's some complications though, for example notifications can also be removed from the network, in that case, `hasSeen` should disappear. 
the case for example would be something like (there's many more though):

1) User A receives a contact request from user B, is not on the activity center, has_seen is true
2) User B retracts the contact request, the notification is removed from user A device. has_seen should be false.

In this case the frontend will need a signal to check whether this has changed (or of course poll, but we tend not to poll in status-react).

Whether we want this to be a separate piece of state (MessengerResponse.ActivityCenterHasSeen say), or a derivation of notifications state (not sure that works, seems harder), it's something we could discuss, but my 2p is that at some point, we will need a signal (or polling)

MishkaRogachev: `User B retracts the contact request, the notification is removed from user A device. has_seen should be false.`
In this case notification gets updated with `read == false` and it should work in same way as regular notification. I'll check it tomorrow

protocol/messenger_response.go
```diff
@@ -461,6 +465,14 @@ func (r *MessengerResponse) ActivityCenterNotifications() []*ActivityCenterNotif
 	return ns
 }
 
+func (r *MessengerResponse) SetActivityCenterState(activityCenterState *ActivityCenterState) {
+	r.activityCenterState = activityCenterState
+}
+
+func (r *MessengerResponse) ActivityCenterState() *ActivityCenterState {
```
ilmotta, line 528: I think it's more idiomatic in Go to not write getters & setters like this, unless they actually perform more computation. Wouldn't it be better to make `activityCenterState` public?

MishkaRogachev: I've made it same way as other properties of MessengerResponse, most of them use getters & setters

ilmotta: Ouch, okay @MishkaRogachev, let's follow the trend for now.

protocol/activity_center.go
```diff
@@ -63,12 +63,16 @@ type ActivityCenterPaginationResponse struct {
 	Notifications []*ActivityCenterNotification `json:"notifications"`
 }
 
+type ActivityCenterState struct {
+	Unread  uint64 `json:"unread"`
```
ilmotta: @MishkaRogachev, just a heads up, this `Unread` count won't be used by mobile at all, because we need to track the count of notifications by type, given that every tab in the Activity Center needs to display its own *unread indicator*.

As of now, the solution on mobile has been to call the endpoint `UnreadAndAcceptedActivityCenterNotificationsCount` for each type, which accepts a notification type as a filter. It's not ideal, but gets the job done. It'd be even better to get a single endpoint on status-go providing counts for multiple notification types so we can reduce the overhead of doing multiple RPC calls.

protocol/messenger_handler.go
```diff
@@ -293,7 +293,7 @@ func (m *Messenger) createContactRequestNotification(contact *Contact, messageSt
 				// Otherwise, it means we have already passed it to the client, so we add it with a `dismissed` flag
 				// so it can clean up
 				if !found {
-					messageState.Response.AddActivityCenterNotification(notification)
+					messageState.Response.AddActivityCenterNotification(notification) // TODO: does we need here to update state?
```
ilmotta: I would leave out TODO comment because it's just an open question :man_shrugging: In case you want to keep it, it's good practice to add your username to the TODO comment, so it's easier for other collaborators to know who they can talk to.

services/ext/api.go
ilmotta: So just to confirm @MishkaRogachev, you think we won't need signals at all? Or do you think a follow up PR might be necessary to implement that?

protocol/activity_center.go
MishkaRogachev: @ilmotta @cammellos 
(Total) unread count and seen i use for this bell icon. Isn't there need to show same in mobile client?
Screenshot 2023-02-15 at 12 35 58

About `UnreadAndAcceptedActivityCenterNotificationsCount` i also can add it for each activity category

protocol/messenger_handler.go
MishkaRogachev: This todo for reviewers) I'm not sure we need to update state here because i don't see saving updated notification to the database

services/ext/api.go
MishkaRogachev: @ilmotta `ActivityCenterState` is added to response. For sync endpoints those responses are delivered as return value from rpc, but for async they delivered via singal. @cammellos does i get it right?

protocol/activity_center.go
ilmotta: @MishkaRogachev, there's certainly a need, but the value is **derived** (summed up) from a different state that's tracking the unread count of different types.

protocol/messenger_handler.go
ilmotta: Cool, I didn't realize that because the comment doesn't say it's just for reviewers. I see a lot of people (including myself) doing this differently @MishkaRogachev. If there's any question, comment, etc, they can be added as a Github comment on the line of the code, Github then creates an open thread that other reviewers can comment on. This is better IMO because you don't need to remember to clean up code comments before merging.

ilmotta (approved):

cammellos (approved):

PR #3133, opened 2023-01-26
Convert a keycard account to a regular account

saledjenic:
This PR is created to support a conversion of a keycard account to a regular account.

What's done:
- `DeleteKeypair` endpoint added
- convert a regular to a keycard account process updated in a way that previous implementation was not deleting all keystore files for the keypir being migrated
- `ConvertToRegularAccount` endpoint added
- `TestConvertAccount` tests the entire process of converting a regular account to a keycard account and then converting that keycard account back to a regular account

PR #3151, opened 2023-01-31
fix: `DeleteAccount` and `AddMigratedKeyPair` declaration change

saledjenic:
Both functions `DeleteAccount` and `AddMigratedKeyPair` require password to be provided in order to delete account from the keystore properly (removing account from the cache and deleting corresponding local keystore file).

Password parameter can be also an empty string, since there are cases when an account is not added to the keystore (in case of keycard account), so we have nothing to delete.

@rasom @qfrank this will be a breaking change for mobile, this is continuation to the changes done here: https://github.com/status-im/status-go/pull/3133

### DIFF
--- a/services/accounts/accounts.go
+++ b/services/accounts/accounts.go
@@ -72,23 +72,35 @@ func (api *API) GetAccounts(ctx context.Context) ([]*accounts.Account, error) {
 	return accounts, nil
 }
 
-func (api *API) DeleteAccount(ctx context.Context, address types.Address) error {
-	acc, err := api.db.GetAccountByAddress(address)
-	if err != nil {
-		return err
-	}
-	if acc.Type != accounts.AccountTypeWatch {
-		err = api.manager.DeleteAccount(address, "")
-		var e *account.ErrCannotLocateKeyFile
-		if err != nil && !errors.As(err, &e) {
+func (api *API) DeleteAccount(ctx context.Context, address types.Address, password string) error {
+	if len(password) > 0 {
+		acc, err := api.db.GetAccountByAddress(address)
+		if err != nil {
 			return err
+		}
+		if acc.Type != accounts.AccountTypeWatch {
+			err = api.manager.DeleteAccount(address, password)
+			var e *account.ErrCannotLocateKeyFile
+			if err != nil && !errors.As(err, &e) {
+				return err
+			}
+
+			allAccountsOfKeypairWithKeyUID, err := api.db.GetAccountsByKeyUID(acc.KeyUID)
+			if err != nil {
+				return err
+			}
+
+			lastAcccountOfKeypairWithTheSameKey := len(allAccountsOfKeypairWithKeyUID) == 1
+			if lastAcccountOfKeypairWithTheSameKey {
+				err = api.manager.DeleteAccount(types.Address(common.HexToAddress(acc.DerivedFrom)), password)
+				var e *account.ErrCannotLocateKeyFile
+				if err != nil && !errors.As(err, &e) {
+					return err
+				}
+			}
 		}
 	}
 
-	return (*api.messenger).DeleteAccount(address)
-}
-
-func (api *API) DeleteAccountForMigratedKeypair(ctx context.Context, address types.Address) error {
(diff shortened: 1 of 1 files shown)


richard-ramos (approved):
For me it looks good, but let's wait for @rasom @cammellos @qfrank review in case this PR breaks the mobile app

services/accounts/accounts.go
```diff
@@ -429,7 +441,7 @@ func (api *API) VerifyPassword(password string) bool {
 	return err == nil
 }
 
-func (api *API) AddMigratedKeyPair(ctx context.Context, kcUID string, kpName string, keyUID string, accountAddresses []string, keyStoreDir string) error {
+func (api *API) AddMigratedKeyPair(ctx context.Context, kcUID string, kpName string, keyUID string, accountAddresses []string, password string) error {
```
rasom, line 444: This method is not used on mobile so no problem here.

services/accounts/accounts.go
```diff
@@ -72,23 +72,35 @@ func (api *API) GetAccounts(ctx context.Context) ([]*accounts.Account, error) {
...
-func (api *API) DeleteAccount(ctx context.Context, address types.Address) error {
-	acc, err := api.db.GetAccountByAddress(address)
-	if err != nil {
-		return err
-	}
-	if acc.Type != accounts.AccountTypeWatch {
-		err = api.manager.DeleteAccount(address, "")
-		var e *account.ErrCannotLocateKeyFile
-		if err != nil && !errors.As(err, &e) {
+func (api *API) DeleteAccount(ctx context.Context, address types.Address, password string) error {
```
rasom, line 75: Just for the context: why the account can not be deleted without the password?

services/accounts/accounts.go
```diff
@@ -441,10 +453,12 @@ func (api *API) AddMigratedKeyPair(ctx context.Context, kcUID string, kpName str
...
 
 	// Once we migrate a keypair, corresponding keystore files need to be deleted.
-	for _, addr := range addresses {
-		// This action deletes an account from the keystore, no need to check for error in this context here, cause if this
-		// action fails from whichever reason the account is still successfully migrated since keystore won't be used any more.
-		_ = api.manager.DeleteAccount(addr, "")
+	if len(password) > 0 {
+		for _, addr := range addresses {
+			// This action deletes an account from the keystore, no need to check for error in this context here, cause if this
+			// action fails from whichever reason the account is still successfully migrated since keystore won't be used any more.
```
rasom: I know that this comment is not added specifically in this PR, but I still don't fully understand why it is safe to ignore any error here. Considering we want to remove file, and we do not because of some error, it might be a problem from security pov to not handle this error (or at least to not notify user about it) even if everything was migrated as expected. Am I missing something?

services/accounts/accounts.go
saledjenic: @rasom cause that's the way how delete account from the keystore function works, check it here: 
https://github.com/status-im/status-go/blob/develop/vendor/github.com/ethereum/go-ethereum/accounts/keystore/keystore.go#L244-L266

that function is called from here:
https://github.com/status-im/status-go/blob/develop/account/accounts.go#L609-L611

Without provided password we're not able to delete neither keystore file for that account nor keystore cache.

services/accounts/accounts.go
saledjenic: Great! ;)

services/accounts/accounts.go
saledjenic: @rasom yes, I agree. That originates from the old `DeleteAccount` function which was not removing cache properly and I was thinking in that time about the case if somebody manually removes the keystore from the file system that we could just simply ignore that and complete the action with success. But I agree, will change that.

saledjenic:
@rasom thanks for the review, updated.

services/accounts/accounts.go
rasom: ok, I se, thanks

rasom (approved):
It is safe to merge now. UI is not implemented yet on mobile side and it doesn't break CI or anything.

PR #3145, opened 2023-01-30
fix: fixes issues on mobile app side after updating `ConvertToKeycardAccount`

saledjenic:
An update of `ConvertToKeycardAccount` function caused some issues on the mobile app side. This commit fixes that trying to satisfy both sides, desktop and mobile:
- `ConvertToKeycardAccount` function is now the same as it was
- updated `ConvertToKeycardAccount` function (which introduced some changes for the desktop app needs) is not renamed to `ConvertToKeycardAccountDesktop`
- `DeleteAccount` function is now the same as it was
- updated `DeleteAccount` function is renamed to `DeleteAccountNew`

PR #3155, opened 2023-02-01
Adjust ConvertToKeycardAccount to make it compatible with mobile version

rasom:
https://github.com/status-im/status-go/pull/3145 this PR can be closed 

Two changes here:
- if for some reason account's key can not be deleted the error returned instead of being ignored
- in case if the key can not be decrypted with a new password we attempt to do this with an old one. Currently circumstances at which migration to keycard happens might differ on mobile and desktop versions, so that on mobile there is no guarantee that a user has logged into any account before this operation. Which means configs might be empty and this keys will not be re encrypted on db password change. 

status: ready


ghost:
### Pull Request Checklist

api/geth_backend.go
```diff
@@ -564,6 +564,19 @@ func (b *GethStatusBackend) ChangeDatabasePassword(keyUID string, password strin
 	return nil
 }
 
+func (b *GethStatusBackend) tryToDeleteAccount(address types.Address, password, newPassword string) error {
```
saledjenic, line 567: @rasom once you know about the final requests for mobile app, if the only way to remove convert an account to a keycard account is from within the app (after user logs in) we should remove this and just use `b.accountManager.DeleteAccount` where is needed.

saledjenic (approved):
@rasom I checked this against the desktop app, it works fine.

api/geth_backend.go
rasom: yep I understand this, thanks

PR #3156, opened 2023-02-01
Fix broken migrations

cammellos:
migrations was noop, it should be fixed now


caybro (commented):
does not fix the issue for me on status-desktop/master

jrainville (approved):
I tested both from before the faulty migration to this fix and by applying the faulty migration and both worked with this.

protocol/migrations/sqlite/1674210659_add_contact_request_local_clock.up.sql
```diff
@@ -3,13 +3,16 @@ ALTER TABLE contacts ADD COLUMN contact_request_local_clock INT;
 ALTER TABLE contacts ADD COLUMN contact_request_remote_clock INT;
 ALTER TABLE contacts ADD COLUMN contact_request_remote_state INT;
 
+-- Broken migration, leaving for posterity and eternal embarrassment
```
richard-ramos, line 6: :medal_military:

richard-ramos (approved):

caybro (approved):

caybro:
My bad, after a forced rebuild seems to work fine, thanks for the quick fix!

cammellos:
> My bad, after a forced rebuild seems to work fine, thanks for the quick fix!

Thank you for testing, and apologies for the mix up

PR #3159, opened 2023-02-03
fix: data/log dir looks weired after local pair sync

qfrank:
after local pair sync, it(data/log dir) looks strange on the receiver side, e.g. the path of log file maybe like:
`/data/data/im.status.ethereum.debug/no_backup/Users/frank/Library/Developer/CoreSimulator/Devices/8A50E96A-1A3C-4047-9351-DC077FC017C4/data/Containers/Data/Application/FF06F89D-522F-408D-90E9-44976E753F25/Library/geth.log`

this PR contains a temprary fix for it, we may need pass the missed info from client finally.

### DIFF
--- a/server/pairing/raw_message_handler.go
+++ b/server/pairing/raw_message_handler.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"path/filepath"
+	"strings"
 
 	"github.com/google/uuid"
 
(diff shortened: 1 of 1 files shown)


cammellos (approved):
As discussed, this is ok for the time being, but eventually we need to have this info passed from the client, to use the proper data dir

PR #3160, opened 2023-02-03
Feat/community permissions

borismelnik:
This `draft` for discussion technical decision in permissions

- I add a comment with planning data base tables and persistence code
- Add and update requests, protobuf and structures for permissions

Will add the commits today with api methods.

It will be great if you guys have much time to check and give me your opinion.

Important changes:

Closes #


appdatabase/migrations/sql/1674570455_communities_permissions.up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE IF NOT EXISTS communities_permissions (
```
jrainville: I think community singular is more grammatically accurate and it sounds better
```suggestion
CREATE TABLE IF NOT EXISTS community_permissions (
```

appdatabase/migrations/sql/1674570455_communities_permissions.up.sql
```diff
@@ -0,0 +1,12 @@
...
+    community_id TEXT NOT NULL,
+    permission_id VARCHAR NOT NULL PRIMARY KEY,
+    hide BOOLEAN NOT NULL DEFAULT FALSE,
+    is_allowed_to INT NOT NULL DEFAULT 0,
+    holds_tokens BOOLEAN NOT NULL DEFAULT FALSE 
+);
+
+CREATE TABLE IF NOT EXISTS communities_permissions_chats {
```
jrainville: I wonder if the chat's permissions can be merged in the communities table. It would simplify queries and speed it up too. Especially since both the whole community and simple channels can have permissions. Maybe just need a `type` to differentiate the two

appdatabase/migrations/sql/1674570455_communities_permissions.up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE IF NOT EXISTS communities_permissions (
+    community_id TEXT NOT NULL,
+    permission_id VARCHAR NOT NULL PRIMARY KEY,
+    hide BOOLEAN NOT NULL DEFAULT FALSE,
+    is_allowed_to INT NOT NULL DEFAULT 0,
+    holds_tokens BOOLEAN NOT NULL DEFAULT FALSE
```
jrainville: I think you're missing a field for the token address.

appdatabase/migrations/sql/1674570455_communities_permissions.up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE IF NOT EXISTS communities_permissions (
+    community_id TEXT NOT NULL,
+    permission_id VARCHAR NOT NULL PRIMARY KEY,
+    hide BOOLEAN NOT NULL DEFAULT FALSE,
```
jrainville: What does hide do?

jrainville (commented):
Nice start. I had questions about the tables and the fields

appdatabase/migrations/sql/1674570455_communities_permissions.up.sql
endulab: I think this is used to indicate that the permission is visible or not. I'd prefer to have "visible" or "public" instead of "hide".

protocol/protobuf/communities.proto
```diff
@@ -50,6 +50,7 @@ message CommunityDescription {
   string outro_message = 12;
   bool encrypted = 13;
   repeated string tags = 14;
+  map<string, CommunityPermission> community_permissions = 15;
```
endulab, line 53: Hmm, We will have now CommunityPermissions and CommunityPermission. Is it possible to merge them some day ? If not maybe we should think about better naming for CommunityPermissions (older one).

appdatabase/migrations/sql/1674570455_communities_permissions.up.sql
borismelnik: yes I changed it to `hidden` everywhere but not here :( are you think `visible` more correct?

protocol/protobuf/communities.proto
borismelnik: yes I forgot to ask that question here... 
@jrainville @endulab @cammellos we already have `Permissions` in Community, but it in fact its `Access`. Should I change it here or leave it as it is?

appdatabase/migrations/sql/1674570455_communities_permissions.up.sql
borismelnik: We have 2 conditions:
1. If `is_allowed_to` is `become_member` or `become_admin` we should ignore chats
2. if `is_allowed_to` is another value - we must have to get chat_ids from this table.

I add a new table for it, cos storing permission_ids in `chats` table as json isn't comfortable for using.

appdatabase/migrations/sql/1674570455_communities_permissions.up.sql
borismelnik: I didn't do it cos we not expect it in MVP and I don't know actually how we plan to store it. In fact of we can assign several token addresses with special amount as required for permission it should be another table.

borismelnik: So I add `holds_tokens` just in case for future and its saves state of switch comment in ui (Who holds?)

appdatabase/migrations/sql/1674570455_communities_permissions.up.sql
borismelnik: 👍

appdatabase/migrations/sql/1674570455_communities_permissions.up.sql
endulab: Hmmm, CommunityPermission should contain information how many tokens and of what type should you have. What does "holds_tokens" do ?

borismelnik: maybe you right... I will change it to list of assets

borismelnik:
@jrainville @endulab I update PR, fix your comments and remove `holds_tokens` cos its not in MVP and need to think about it. Also, I not include update method cos we still not agreed with tables.

appdatabase/migrations/sql/1674570455_communities_permissions.up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS community_permissions (
+    community_id TEXT NOT NULL,
+    permission_id VARCHAR NOT NULL PRIMARY KEY,
+    permission_private BOOLEAN NOT NULL DEFAULT FALSE,
+    is_allowed_to INT NOT NULL DEFAULT 0
```
jrainville, line 5: You're missing the condition in this. For example, we have `is_allowed_to` that can be `admin`, but we don't have how people can become that.

There is "holds token", "holds collectible", maybe others. We need to add this.

I think starting with just holding >1 token of something is the simplest starting point, but it can be a collectible since that's what Michal started on for minting

endulab: We have 3 types of "hold": Assets, Collectibles and Ens. So, in case of Assets and Collectibles permission will have "amount" field, and in case of Ens, there will be string.

jrainville:
Closing since it's been done by Pascal instead

PR #3161, opened 2023-02-03
Add community request to response when a response handled

cammellos:
Adds the updated community request when a response is received.

In general, any time a piece of state is updated in the backend, that
should be propagated to the client through signals.

In this case, when a request was accepted, the client wasn't notified,
requiring them to re-fetch the accepted requests and causing
inconsistent state between status-go and client.


0x-r4bbit (approved):
LGTM!
Would be cool if you could add some comment on why this change was done?

cammellos:
> LGTM! Would be cool if you could add some comment on why this change was done?

sure, you mean in the PR description or you'd like me to do it in the code?

0x-r4bbit:
Ah, just a comment here is enough. I see what the PR is doing, but was just wondering why. I put such reasoning typically in the commit message, just to ensure people don't run into open questions.

But a comment here is enough.

cammellos:
@0x-r4bbit good point! updated git commit & pr description

richard-ramos (approved):

PR #3163, opened 2023-02-04
Update AC notification on EditMessage

rasom:
needed for https://github.com/status-im/status-mobile/issues/14962

status: ready


protocol/common/message.go
```diff
@@ -214,6 +214,8 @@ type Message struct {
 	ContactVerificationState ContactVerificationState `json:"contactVerificationState,omitempty"`
 
 	DiscordMessage *protobuf.DiscordMessage `json:"discordMessage,omitempty"`
+
+	JustEdited bool
```
0x-r4bbit: I assume `Edited` as property would clash with an already existing one? Or is there any other reason for this particular naming?

0x-r4bbit (approved):
Changes LGTM, just a question about the naming

protocol/common/message.go
rasom: The idea is that there is no point in storing this field in the database thus if I name it `Edited` it will always be `false` on fetching message from the database, although in reality it might have been actually edited already and be in conflict with `EditedAt`. That potentially might be confusing. But it seems that `JustEdited` name for a short living flag might be confusing as well... :) Let me know if you have some better options

0x-r4bbit: I see! I guess it makes sense to at least add a comment to that property then with the clarifications you've just made.

protocol/messenger_handler.go
```diff
@@ -1387,8 +1387,19 @@ func (m *Messenger) HandleEditMessage(response *MessengerResponse, editMessage E
...
 	}
-	response.AddMessage(originalMessage)
-	response.AddChat(chat)
+	responseTo, err := m.persistence.MessageByID(originalMessage.ResponseTo)
+
+	if err != nil && err != common.ErrRecordNotFound {
+		return err
+	}
+
+	originalMessage.JustEdited = true
```
cammellos: I think it would be cleaner if we don't set the field and we have just two different functions:

Something like:

```
updateActivityNotificationCenterMessage(...)
```
and
```
addNew...
```
and then factor out the shared logic inside


or, if you really want to have a single function, then
```
upsertActivityNotification(....., editedMessage bool)
```
instead of modifying message

protocol/common/message.go
rasom: Added the comment, thanks!

protocol/messenger_handler.go
rasom: well.. unfortunately I agree

protocol/messenger.go
```diff
@@ -2940,6 +2940,30 @@ func (r *ReceivedMessageState) addNewMessageNotification(publicKey ecdsa.PublicK
 	return nil
 }
 
+// updateActivityCenterNotification updates AC notification if it exists and hasn't been read yet
+func (r *ReceivedMessageState) updateActivityCenterNotification(publicKey ecdsa.PublicKey, m *Messenger, message *common.Message, responseTo *common.Message) error {
```
ilmotta: I have a tendency to prefer to separate the CRUD operations into separate functions if possible, so that the functions are more flexible and can end up being used in different contexts.

In this concrete case, the function says it updates the entity, but it also reads it first from the db. Instead, we could separate the read from the update, i.e. the function `updateActivityCenterNotification` would receive an instance of `ActivityCenterNotification`.

There's only a marginal gain for this PR, so take my suggestion with a grain of salt, but applying this practice throughout the system could be quite beneficial.

ilmotta (approved):

protocol/messenger.go
rasom: now it says that it updates existing entry, I consider the case closed :D

protocol/messenger_handler.go
rasom: @cammellos done, thanks

cammellos (approved):

PR #3165, opened 2023-02-06
Handle a few edge cases in contact requests

cammellos:
This PR handles a few edge cases in contact request state management, you can check the tests for a more detailed description.
Removes some unused protobufs fields as well.


protocol/contact.go
```diff
@@ -383,3 +403,56 @@ func (c *Contact) MarshalJSON() ([]byte, error) {
 
 	return json.Marshal(item)
 }
+
+// ContactRequestPropagatedStateReceived handles the propagation of state from
+// the other end.
+func (c *Contact) ContactRequestPropagatedStateReceived(state *protobuf.ContactRequestPropagatedState) ContactRequestProcessingResponse {
```
ilmotta, line 409: :warning: Not necessarily a suggestion for this PR, but more like something we could improve later on.

It's probably just me, but I find many of the functions related to contact request flows to be less clear than they could be because they don't tell what they do, instead, their names describe the state of some entity or process. Examples: `c.ContactRequestPropagatedState`, `c.ContactRequestPropagatedStateReceived`, `c.ContactRequestSent`, etc. This kinds of forces me to always read their implementation and constantly translate in my mind what they do.

In the example below, I already know we're in the propagated state due to the conditional, but the name of the function just reiterates the state, not what it does. Upon further inspection I realized what it does, but only because I read it line by line. Naming is hard, one suggestion: `syncContactRequestClocksAndState`.

```go
if message.ContactRequestPropagatedState != nil {
	result := contact.ContactRequestPropagatedStateReceived(message.ContactRequestPropagatedState)
```

Another example:

```go
case ContactRequestStateReceived:
	c.ContactRequestReceived(remoteClock)
```

protocol/contact.go
```diff
@@ -383,3 +403,56 @@ func (c *Contact) MarshalJSON() ([]byte, error) {
...
+	}
+
+	// If they expect our state to be more up-to-date, we only
+	// trust it if the state is set to None, in this case we can trust
+	// it, since a retraction can be initiated by both parties
+	if expectedLocalClock > c.ContactRequestLocalClock && c.ContactRequestLocalState != ContactRequestStateDismissed && expectedLocalState == ContactRequestStateNone {
+		response.processed = true
+		c.ContactRequestLocalClock = expectedLocalClock
+		c.ContactRequestLocalState = ContactRequestStateNone
+		// We set they remote state, as this was an implicit retraction
```
ilmotta: Could you give an example of what you mean by *implicit retraction*? It might not be clear to other people as well.

jrainville:
```suggestion
		// We set the remote state, as this was an implicit retraction
```

protocol/contact_test.go
```diff
@@ -559,3 +560,152 @@ func TestMarshalContactJSON(t *testing.T) {
...
+	c.ContactRequestPropagatedStateReceived(
+		&protobuf.ContactRequestPropagatedState{
+			RemoteState: uint64(ContactRequestStateNone),
+			RemoteClock: 2,
+			LocalState:  uint64(ContactRequestStateNone),
+			LocalClock:  1,
+		},
+	)
+
+	require.False(t, c.added())
```
ilmotta, line 637: The assertions in `TestContactContactRequestPropagatedStateReceivedOutOfDateLocalStateOnTheirSide` verify the result returned by `ContactRequestPropagatedStateReceived`, but these ones only check the mutations in `c`. I think these tests would give stronger guarantees if they also asserted the full returned result from `ContactRequestPropagatedStateReceived`, as well as the mutations on `c`.

protocol/messenger_contact_requests_test.go
```diff
@@ -1276,3 +1276,398 @@ func (s *MessengerContactRequestSuite) TestPairedDevicesRemoveContact() {
...
+	// Wait for the message to reach its destination
+	_, err = WaitOnMessengerResponse(
+		bob,
+		func(r *MessengerResponse) bool {
+			return len(r.ActivityCenterNotifications()) > 0
+		},
+		"no messages",
+	)
+	s.Require().NoError(err)
+
```
jrainville, line 1581: Did you forget here to test Bob's side?

jrainville (approved):
LGTM

protocol/messenger_contact_requests_test.go
```diff
@@ -1276,3 +1276,398 @@ func (s *MessengerContactRequestSuite) TestPairedDevicesRemoveContact() {
...
+	s.Require().NotNil(bobFromAlice)
+	s.Require().NotNil(aliceFromBob)
+
+	// We can't simulate out-of-order messages easily, so we need to do
+	// things manually here
+
+	result := aliceFromBob.ContactRequestPropagatedStateReceived(bobFromAlice.ContactRequestPropagatedState())
+	s.Require().True(result.newContactRequestReceived)
+
```
ilmotta, line 1672: I see a handful of these extra newlines at the end/beginning of function bodies. Have we ever looked at using https://github.com/mvdan/gofumpt? It's capable of auto formatting more stuff and it seems to be a drop-in replacement for gofmt. This tool can, among other *more useful* things, remove extraneous newlines at the end of function bodies, so all in all it could be a win.

protocol/messenger_handler.go
```diff
@@ -869,6 +877,29 @@ func (m *Messenger) HandleContactUpdate(state *ReceivedMessageState, message pro
 
 	logger.Info("Handling contact update")
 
+	if message.ContactRequestPropagatedState != nil {
+		result := contact.ContactRequestPropagatedStateReceived(message.ContactRequestPropagatedState)
+		if result.sendBackState {
+			// This is a bit dangerous, since it might trigger a ping-pong of contact updates
```
ilmotta, line 886: How about we track this problem as an issue and remove the comment? It seems rather important.

protocol/contact_test.go
```diff
@@ -559,3 +560,152 @@ func TestMarshalContactJSON(t *testing.T) {
...
+		},
+	)
+
+	require.False(t, result.sendBackState)
+
+}
+
+func TestContactContactRequestPropagatedStateReceivedOutOfDateLocalStateOnOurSide(t *testing.T) {
+	// We receive a message with expected contact request state == none
+	// and clock > our clock. We consider this a retraction, unless we are in the dismissed state, since that should be only changed by a trusted device
```
ilmotta: These 2 comment lines are a `gq` away from looking pretty ;)

ilmotta (approved):
Exemplary work! I especially enjoyed reading all the unit tests and the detailed explanations in the comments. Left a couple comments, no blockers to me.

protocol/contact.go
```diff
@@ -22,6 +23,9 @@ const (
 	ContactRequestStateNone ContactRequestState = iota
 	ContactRequestStateMutual
 	ContactRequestStateSent
+	// Received is a confusing state, we should use
+	// sent for both, since they are now stored in different
+	// states
 	ContactRequestStateReceived
```
qfrank, line 29: this comment confuse me now(no confusion before this comment), could u elebrate it, thank you 🙂 @cammellos

qfrank: oh, i c, `states` u mean here r ContactRequestRemoteState/ContactRequestLocalState

qfrank (approved):

protocol/messenger_contact_requests_test.go
cammellos: No, we are happy with just checking that they receive a notificaton, which is tested otherwise `err` would be nil

protocol/contact.go
cammellos: So, it means that we don't know whether the contact request was retracted expliciltly, for example by receiving a `ContactRequestRetractedMessage`, but we noticed that their state has changed so that effectively the contact request flow was restarted, this could have been done on their end, and we haven't (yet) received the retraction message, or on our end from a paired device, and we have not received that message.
I'll update the comment to make it clearer

protocol/messenger_contact_requests_test.go
cammellos: I'll look into it, thanks for the suggestion!

protocol/messenger_handler.go
cammellos: I am not sure how to address this, but I can add an issue to at least have a look to make sure that we don't create a storm of contact requests :)

protocol/contact.go
ilmotta: Thank you for the clarification :bulb:

protocol/contact_test.go
cammellos: I'll fortify the tests in a follow up PR, if that's ok, I have yet one fix to add separately :)

protocol/messenger_handler.go
ilmotta: I kind of like your idea of using an exponential backoff to decide if/when to send back the state. It sounds reasonable and not too difficult to implement, but maybe it could cause side-effects I'm not aware of.

cammellos: I think that's useful to avoid spamming and we will implement it, but might hide the issue at this stage, say that the two states never converge, and keep messaging each other, then it would be hard to see from logs, while now it would be easier since we would see a lot of requests being exchanged, though 100% something we want to implement

ilmotta: You're right, the matter deserves more analysis.

PR #3167, opened 2023-02-06
Remove implicit marking messages as seen based on clock

MishkaRogachev:
Need for https://github.com/status-im/status-desktop/pull/9432

This code forced unseen messages coming out of order (by clock) become seen

Screenshot 2023-02-06 at 17 22 30


jrainville (commented):
I think it makes sense to not mark as seen for new received messages, I just wonder if it will be annoying if we ever receive very old messages. Maybe we need a cut off time? If message is older than, let's say, 8 hours, we mark as seen?

MishkaRogachev:
@jrainville @cammellos @osmaczko the requirements need to be clarified from the UX

PR #3168, opened 2023-02-06
fix: request token balance for overridden tokens

stefandunca:
## Updates [#9091](https://github.com/status-im/status-desktop/issues/9091)

Fix: balance for tokens overrides are not requested in case of fallback when Etherscan bulk fetch is not working.

### DIFF
--- a/services/wallet/token/token.go
+++ b/services/wallet/token/token.go
@@ -44,6 +44,7 @@ func (t *Token) IsNative() bool {
 	return t.Address == nativeChainAddress
 }
 
+// Manager is used for accessing token store. It changes the token store based on overridden tokens
 type Manager struct {
 	db             *sql.DB
 	RPCClient      *rpc.Client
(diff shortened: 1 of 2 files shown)


alaibe (commented):
What I am curious of is how is it possible that we try to fetch balance for the non overidden address. As the addresses all comes from the store they should also be using this address and not the real one?

stefandunca:
> What I am curious of is how is it possible that we try to fetch balance for the non overidden address. As the addresses all comes from the store they should also be using this address and not the real one?

Good question, I thought that we use a copy of the store, however, following your inquiry I realize we modify the global `tokenMap` and indeed the values don't change (logging before and after). See the updated code.

I extracted the code and added a test which passes with test values. Now I'm following this lead and look into golang global memory particularities. Thanks for the heads up.

stefandunca:
Found the root cause and removed the workaround. We were replacing the address but not the lookup address. The fix was to add a new entry with the overridden address and remove the original one. @alaibe please have another look.

alaibe (approved):

PR #3169, opened 2023-02-07
fix(LinksPreview): Validate image links in the links preview request

alexjba:
Motivated by: [9084](https://github.com/status-im/status-desktop/pull/9084)

While we don't request preview data from Tenor because the url itself will contain the GIF image, we still need to validate the url in the preview request. This is done using a HEAD request where we expect the response status code to be 200 OK. At the same time we will add the content type to the preview response.

Companion PR [9428](https://github.com/status-im/status-desktop/pull/9428)


stefandunca (approved):
Great finding and fix, thanks.

caybro (approved):

jrainville (approved):

mprakhov (approved):

alexandraB99 (approved):

PR #3172, opened 2023-02-08
Improve QR Handler and add Tests

siddarthkay:
In this PR : 

- Moved out functions from `qrcode/qrcode.go` package to images package  at `images/manipulation.go` and `images/decode.go` because few of these functions were general purpose and image related functions.
- Added `server/qrops.go` which now contains just the QR generation logic and abstracted it out of `handlers.go`
- Added `server/qrops_test.go` which explicitly tests QR generation capabilities
- Added a test in `server/server_test.go` which verifies the bytes received after calling the QRGeneration endpoint at media server.
- Added test data in `images/test_data.go` which now contains bytes for a dummy logo to test logo on QR functionality.
- Added `_assets/tests/QRWithLogo.png` and `_assets/tests/defaultQR.png` so that we can compare their bytes with bytes generated by tests.

Why I did these things : 
- In some scenarios ( strange and unknown ) I was unable to get a QR image on status-mobile side and when i decided to debug that issue on status-go side it was almost impossible to figure out ( which was my fault since i wrote the initial version).
- Adding tests and making the logic more modular allowed me to debug properly.


siddarthkay:
also thanks @qfrank  for helping with the http request to media server from status-go side

server/qrops.go
```diff
@@ -0,0 +1,273 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("encoding PNG failed: %v", err)
+	}
+	return resultBytes, nil
+}
+
+func GetLogoImage(multiaccountsDB *multiaccounts.Database, keyUID string, imageName string) ([]byte, error) {
+	var (
+		imageFolderBasePath = "../_assets/tests/"
+		logoFileStaticPath  = imageFolderBasePath + "status.png"
```
qfrank: Curious if this path will work when status-go integerated into mobile? if not, i suggest using `go-bindata`. I thought the resources under _assets/tests/ will only be used for test purpose?

server/qrops.go
```diff
@@ -0,0 +1,273 @@
...
+		correctionLevel = qrcode.ErrorCorrectionLow
+	case "2":
+		correctionLevel = qrcode.ErrorCorrectionMedium
+	case "3":
+		correctionLevel = qrcode.ErrorCorrectionQuart
+	case "4":
+		correctionLevel = qrcode.ErrorCorrectionHighest
+	}
+
+	buf := NewWriterCloserByteBuffer()
```
qfrank: i'd prefer move code (line 209) to line 220 so that `buf` is closing its usage

server/qrops.go
```diff
@@ -0,0 +1,273 @@
...
+		LogoBytes           []byte
+	)
+
+	identityImageObjectFromDB, err := multiaccountsDB.GetIdentityImage(keyUID, imageName)
+
+	if err != nil {
+		return nil, err
+	}
+
+	staticLogoFileBytes, _ := images.ImageToBytes(logoFileStaticPath)
```
richard-ramos: Should the error here be handled or logged?

server/qrops.go
siddarthkay: Oh yes i intended to update this later but I missed out and thanks for the suggestion @qfrank . I will refactor this to use `go-bindata`

server/qrops.go
siddarthkay: makes sense, will do :)

server/qrops.go
siddarthkay: hmmmm, yes it wouldn't hurt to log it, I'll refactor, thanks for looking at this @richard-ramos

server/qrops.go
```diff
@@ -0,0 +1,273 @@
...
+
+		payload = images.SuperimposeLogoOnQRImage(payload, logo)
+	}
+
+	if size > 0 {
+
+		if err != nil {
+			logger.Error("could not convert string to int for size param ", zap.Error(err))
+			return nil
+		}
```
Samyoul: You seem to be evaluating an error that isn't created.

server/qrops.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	return imageToBeResized
+}
+
+func generateQRBytes(params url.Values, logger *zap.Logger, multiaccountsDB *multiaccounts.Database) []byte {
+
+	qrURLBase64Decoded := getQrURL(params, logger)
+	superImposeLogoOnQR := getAllowProfileImage(params, logger)
+	level := getLevel(params, logger)
```
Samyoul: Should this function return a `qrcode. ErrorCorrection` type? rather than parsing the output externally, and the string variable `level` isn't used after parsing.

server/qrops.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+func getKeyUID(params url.Values, allowProfileImages bool, logger *zap.Logger) string {
+	keyUID, ok := params["keyUid"]
+
+	// if allowProfileImage was also not passed
+	// in that case we don't care and can return an empty string back
+	if !allowProfileImages {
+		return ""
+	}
```
Samyoul: If the condition is already known by the consuming code, should this function be responsible for returning an empty value?

server/qrops.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+func getImageName(params url.Values, allowProfileImages bool, logger *zap.Logger) string {
+	imageName, ok := params["imageName"]
+
+	// if allowProfileImage was also not passed
+	// in that case we don't care and can return an empty string back
+	if !allowProfileImages {
+		return ""
+	}
```
Samyoul: Same as above https://github.com/status-im/status-go/pull/3172/files#r1106496837

server/qrops.go
```diff
@@ -0,0 +1,273 @@
...
+	switch level {
+	case "1":
+		correctionLevel = qrcode.ErrorCorrectionLow
+	case "2":
+		correctionLevel = qrcode.ErrorCorrectionMedium
+	case "3":
+		correctionLevel = qrcode.ErrorCorrectionQuart
+	case "4":
+		correctionLevel = qrcode.ErrorCorrectionHighest
+	}
```
Samyoul: This code is basically building a configuration. Perhaps these values could be encapsulated in a struct.

Additionally each of these functions consume the same variables, `params` and `logger`. Is it possible that this functionality could belong to a single struct that manages holds these variables?

Something like:

```go
type qrConfig struct {
	Data            string
	WithLogo        bool
	CorrectionLevel qrcode.ErrorCorrectionLevel
	KeyUID          string
	ImageName       string
	Size            uint
}
``` 

Then your consuming code can handle the struct fields and you get default empty values for free (`""`, `0`, etc)

server/qrops.go
```diff
@@ -0,0 +1,273 @@
...
+	imageToBeResized, err := strconv.Atoi(size[0])
+
+	if err != nil {
+		logger.Error("[qrops-error] could not convert string to int for size param ", zap.Error(err))
+		return 0
+	}
+
+	return imageToBeResized
+}
```
Samyoul: Perhaps add a check to see if the value is `< 0`? Would a negative value ever be desirable here?

server/qrops.go
```diff
@@ -0,0 +1,273 @@
...
+
+	if !ok || len(qrURL) == 0 {
+		logger.Error("[qrops-error] no qr url provided")
+	}
+
+	decodedURL, err := base64.StdEncoding.DecodeString(qrURL[0])
+
+	if err != nil {
+		logger.Error("[qrops-error] unable to decoding string from base64", zap.Error(err))
+	}
```
Samyoul: This seems like the kind of error that should stop the current flow and warn the user / consuming code about.

server/handlers.go
```diff
@@ -40,25 +36,6 @@ const (
 
 type HandlerPatternMap map[string]http.HandlerFunc
 
-type QROptions struct {
-	URL                  string `json:"url"`
-	ErrorCorrectionLevel string `json:"errorCorrectionLevel"`
-	Capacity             string `json:"capacity"`
-	AllowProfileImage    bool   `json:"withLogo"`
-}
```
Samyoul, line 48: Ah, I see that something close to what I mention in this comment https://github.com/status-im/status-go/pull/3172/files#r1106508086 used to exist.

qrcode/qrcode.go
```diff
@@ -1,182 +0,0 @@
-package qrcode
```
Samyoul, line 1: Ah this is nice, moving code directly related to its respective domain, into that domain / package. It is also a good way of preventing circle dependencies.

images/manipulation.go
```diff
@@ -84,3 +108,107 @@ func CropCenter(img image.Image) (image.Image, error) {
...
+	if err = png.Encode(&newImgBytes, newImg); err != nil {
+		return nil, err
+	}
+	return newImgBytes.Bytes(), nil
+}
+
+func SuperimposeLogoOnQRImage(imageBytes []byte, qrFilepath []byte) []byte {
+	// Read the two images from bytes
+	img1, _, _ := image.Decode(bytes.NewReader(imageBytes))
+	img2, _, _ := image.Decode(bytes.NewReader(qrFilepath))
```
Samyoul: If I'm not mistaken `image.Decode()` returns an error as its 3rd return value. These errors should be handled

images/manipulation.go
```diff
@@ -84,3 +108,107 @@ func CropCenter(img image.Image) (image.Image, error) {
...
+	img2Bounds := img2.Bounds()
+	// Calculate the x and y coordinates to center the second image
+	x := (img1.Bounds().Dx() - img2Bounds.Dx()) / 2
+	y := (img1.Bounds().Dy() - img2Bounds.Dy()) / 2
+	// Draw the second image on top of the first image at the calculated coordinates
+	draw.Draw(result, img2Bounds.Add(image.Pt(x, y)), img2, image.ZP, draw.Over)
+	// Encode the final image to a desired format
+	var b bytes.Buffer
+	if err := png.Encode(&b, result); err != nil {
+		fmt.Println(err)
```
Samyoul: This error needs to be handled, `fmt.Println(err)` shouldn't be in production unless you are writing CLI code.

server/qrops.go
siddarthkay: oh lol, somehow that got left in by mistake, will fix :)

server/qrops.go
siddarthkay: now that I'm more confident, I shall attempt this.

server/qrops.go
siddarthkay: hmm I was facing a stupid type error when I had tried doing that, I will try again 👍🏻

server/qrops.go
siddarthkay: negative sizes are definitely a no no, will add that validation!

server/qrops.go
siddarthkay: hmm yeah you are right, I could shift responsibility to the code calling this method to validate if `allowProfileImages` was `false`

server/qrops.go
siddarthkay: hmmm, now that I think about it, this could be made better.

images/manipulation.go
siddarthkay: will do 🫡

images/manipulation.go
siddarthkay: ah my bad, will fix

server/qrops.go
siddarthkay: fixed here : https://github.com/status-im/status-go/pull/3172/commits/7f680f39c7dfe1047b7ef85f844b12ae9a35effb
Although I'm not sure if we should have 2 files that are go-bindata'ed 
I will read more into the subject

server/qrops.go
siddarthkay: fixed here : https://github.com/status-im/status-go/pull/3172/commits/18c912167a62bf83f79354af52da32c640fb4111

server/qrops.go
siddarthkay: fixed here : https://github.com/status-im/status-go/pull/3172/commits/b5b8339943e3287dea723e4ce2b4f95686c6f88f

server/qrops.go
siddarthkay: fixed here : https://github.com/status-im/status-go/pull/3172/commits/e802e70586126844a5c6bbe242368217f222d8f0

server/qrops.go
siddarthkay: fixed here. :https://github.com/status-im/status-go/pull/3172/commits/25967f099add2691e53d5558ed8e661d1ab097c4

server/qrops.go
siddarthkay: fixed here : https://github.com/status-im/status-go/pull/3172/commits/25967f099add2691e53d5558ed8e661d1ab097c4

server/qrops.go
siddarthkay: fixed here : https://github.com/status-im/status-go/pull/3172/commits/3f45a500e394591b67c7dbedc0a07ac3b0ae1623

server/qrops.go
siddarthkay: fixed here : https://github.com/status-im/status-go/pull/3172/commits/2539fa1e2fe00a719f5c3dfdcfd097693916525a

images/manipulation.go
siddarthkay: fixed here : https://github.com/status-im/status-go/pull/3172/commits/35fd121cfa2729b0ab9c85f9e482408047609fd6

images/manipulation.go
siddarthkay: fixed here : https://github.com/status-im/status-go/pull/3172/commits/35fd121cfa2729b0ab9c85f9e482408047609fd6

server/qrops.go
```diff
@@ -0,0 +1,281 @@
...
+
+	return config
+}
+
+func (q *QRConfig) SetQrURL() {
+	qrURL, ok := q.Params["url"]
+	logger := q.Logger
+
+	if !ok || len(qrURL) == 0 {
+		logger.Error("[qrops-error] no qr url provided")
```
qfrank: if it's not that serious, maybe using logger.Warn or something else would be more suitable?

server/server_test.go
```diff
@@ -29,11 +40,14 @@ func (s *ServerURLSuite) SetupTest() {
...
-	s.server = &MediaServer{Server: Server{
-		hostname:   DefaultIP.String(),
-		portManger: newPortManager(s.Logger, nil),
-	}}
-	err := s.server.SetPort(1337)
+	mediaServer, err := NewMediaServer(nil, nil, nil)
+	s.Require().NoError(err)
+
+	s.server = mediaServer
+	err = s.server.SetPort(1337)
```
qfrank: there are > 1 places using 1337, define it as const or variable would be better

server/qrops.go
siddarthkay: Hmm, the `url` is a required parameter and hence this is a critical error.
Nothing will work if `url` is not provided or if an empty string is passed for this parameter from client side.

Samyoul: If this is a critical error you should probably return an error at this point to inform the consuming code.

server/qrops.go
```diff
@@ -28,177 +28,201 @@ func NewWriterCloserByteBuffer() *WriterCloserByteBuffer {
...
+	config.Logger = logger
+	config.SetQrURL()
+	config.SetAllowProfileImage()
+	config.SetErrorCorrectionLevel()
+	config.SetSize()
+
+	if config.WithLogo {
+		config.SetKeyUID()
+		config.SetImageName()
 	}
```
Samyoul: If you intend on keeping these functions purely for initialising a `QRConfig`, perhaps make them unexported, we don't want other code calling these if they are only for internal usage.

server/qrops.go
```diff
@@ -28,177 +28,201 @@ func NewWriterCloserByteBuffer() *WriterCloserByteBuffer {
...
+type QRConfig struct {
+	DecodedQRURL    string
+	WithLogo        bool
+	CorrectionLevel qrcode.EncodeOption
+	KeyUID          string
+	ImageName       string
+	Size            int
+	Params          url.Values
+	Logger          *zap.Logger
 }
```
Samyoul: Nice refactor. I like how clean and segregated everything is.

server/qrops.go
```diff
@@ -28,177 +28,201 @@ func NewWriterCloserByteBuffer() *WriterCloserByteBuffer {
...
-	}
-	return size / 5
+	return config
 }
 
-func getQrURL(params url.Values, logger *zap.Logger) string {
-	qrURL, ok := params["url"]
+func (q *QRConfig) SetQrURL() {
+	qrURL, ok := q.Params["url"]
+	logger := q.Logger
```
Samyoul: It isn't necessary to pass this pointer to a new variable.

You can just call `q.Logger.Error(...)`

server/qrops.go
```diff
@@ -0,0 +1,281 @@
...
+
+type QRConfig struct {
+	DecodedQRURL    string
+	WithLogo        bool
+	CorrectionLevel qrcode.EncodeOption
+	KeyUID          string
+	ImageName       string
+	Size            int
+	Params          url.Values
+	Logger          *zap.Logger
```
Samyoul: It is very rare that you would need to expose a logger to any external code. I'd suggest making this unexported.

```suggestion
	logger          *zap.Logger
```

server/qrops.go
siddarthkay: fixed here : https://github.com/status-im/status-go/pull/3172/commits/b55243dbf691338c82ca6c09e00042f7c8decfe1

server/qrops.go
siddarthkay: got rid of this in : https://github.com/status-im/status-go/pull/3172/commits/b55243dbf691338c82ca6c09e00042f7c8decfe1

server/qrops.go
siddarthkay: got rid of this in : https://github.com/status-im/status-go/pull/3172/commits/b55243dbf691338c82ca6c09e00042f7c8decfe1

server/qrops.go
siddarthkay: fixed here : https://github.com/status-im/status-go/pull/3172/commits/0143994402892fc3f9125727cc3ef017ccba0397

server/server_test.go
siddarthkay: fixed here : https://github.com/status-im/status-go/pull/3172/commits/dbdcd08f2c1efb8ea0506efa429e7082407483ff

server/server.go
```diff
@@ -21,6 +21,9 @@ type Server struct {
 	portManger
 }
 
+const customPortForTests = 1337
+const defaultPortForTests = 80
```
qfrank: it would be better that const relate to test moved to xxx_test.go 🙂

siddarthkay: good suggestion, moved here : https://github.com/status-im/status-go/pull/3172/commits/6839f458b4b3a71c3d5bf7e61e3d80c44131d4fd

qfrank (approved):

images/manipulation.go
```diff
@@ -1,16 +1,42 @@
 package images
 
 import (
+	"bytes"
 	"fmt"
+
```
Samyoul: Maybe a little IDE quirk adding a new line here

server/qrops.go
```diff
@@ -0,0 +1,291 @@
...
+	switch levelInt {
+	case 1:
+		q.CorrectionLevel = qrcode.WithErrorCorrectionLevel(qrcode.ErrorCorrectionLow)
+	case 2:
+		q.CorrectionLevel = qrcode.WithErrorCorrectionLevel(qrcode.ErrorCorrectionMedium)
+	case 3:
+		q.CorrectionLevel = qrcode.WithErrorCorrectionLevel(qrcode.ErrorCorrectionQuart)
+	case 4:
+		q.CorrectionLevel = qrcode.WithErrorCorrectionLevel(qrcode.ErrorCorrectionHighest)
+	}
```
Samyoul: Perhaps add a default to this switch incase you get a result that is not 1 - 4

```suggestion
	switch levelInt {
	case 1:
		q.CorrectionLevel = qrcode.WithErrorCorrectionLevel(qrcode.ErrorCorrectionLow)
	case 2:
		q.CorrectionLevel = qrcode.WithErrorCorrectionLevel(qrcode.ErrorCorrectionMedium)
	case 3:
		q.CorrectionLevel = qrcode.WithErrorCorrectionLevel(qrcode.ErrorCorrectionQuart)
	case 4:
		q.CorrectionLevel = qrcode.WithErrorCorrectionLevel(qrcode.ErrorCorrectionHighest)
	default:
		q.CorrectionLevel = qrcode.WithErrorCorrectionLevel(qrcode.ErrorCorrectionMedium)
	}
```

server/qrops.go
```diff
@@ -0,0 +1,291 @@
...
+
+func (q *QRConfig) setSize() error {
+	size, ok := q.Params["size"]
+
+	if ok {
+		imageToBeResized, err := strconv.Atoi(size[0])
+
+		if err != nil || imageToBeResized <= 0 {
+			return err
+		}
```
Samyoul: If the err is nil, but `imageToBeResized` is 0 or less, this would return a nil error. But you should return a specific error should be returned for this kind of issue.

server/qrops.go
```diff
@@ -0,0 +1,291 @@
...
+
+	q.KeyUID = keyUID[0]
+	return nil
+}
+
+func (q *QRConfig) setImageName() {
+	imageName, ok := q.Params["imageName"]
+	//if the imageName was not passed, we default to "thumbnail"
+	if !ok || len(imageName) == 0 {
+		q.ImageName = "thumbnail"
```
Samyoul: Perhaps you could reference the `image` package constant `image.SmallDimName`?

If you use the larger dimension does it improve the quality of the output QR code image?

server/qrops.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+func NewQRConfig(params url.Values, logger *zap.Logger) *QRConfig {
+	config := &QRConfig{}
+	config.Params = params
+	err := config.setQrURL()
+
+	if err != nil {
+		logger.Error("[qrops-error] error in setting QRURL", zap.Error(err))
+	}
```
Samyoul: If there is a fail on setting the QR URL, should subsequent processes execute? Or should the process terminate at this point?

server/qrops.go
```diff
@@ -0,0 +1,291 @@
...
+
+	config.setAllowProfileImage()
+	config.setErrorCorrectionLevel()
+	err = config.setSize()
+
+	if err != nil {
+		logger.Error("[qrops-error] could not convert string to int for size param ", zap.Error(err))
+	}
```
Samyoul: If there is a fail with the size param could the application proceed with the rest of this process? Or would this cause fails later in the flow? Is this an error that should return and halt the function?

server/qrops.go
```diff
@@ -0,0 +1,291 @@
...
+
+	if config.WithLogo {
+		err = config.setKeyUID()
+
+		if err != nil {
+			logger.Error(err.Error())
+		}
```
Samyoul: If a Key UID is required, what would affect would having no key UID have on later code execution?

Samyoul (commented):
Great work, I've added a few points with regards to behaviour around errors

server/qrops.go
siddarthkay: hmm you're right, large image provides better quality. thanks for looking this over

server/qrops.go
siddarthkay: Its better to terminate here itself because if there is a fail in setting URL other parts of the code would also fail and it would be cumbersome to go through the stack trace to figure out root cause. 
Good catch @Samyoul !

siddarthkay: fixed here : https://github.com/status-im/status-go/pull/3172/commits/8d1a34bcda651166d09ea2672711e0b662200d54

server/qrops.go
siddarthkay: fixed as a side effect of : https://github.com/status-im/status-go/pull/3172/commits/8d1a34bcda651166d09ea2672711e0b662200d54

server/qrops.go
siddarthkay: fixed as a side effect of : https://github.com/status-im/status-go/commit/8d1a34bcda651166d09ea2672711e0b662200d54

server/qrops.go
siddarthkay: fixed here : https://github.com/status-im/status-go/pull/3172/commits/78945a6ebe561d76075cc25e1ee8d68a8cc9beb5

server/qrops.go
siddarthkay: fixed here : https://github.com/status-im/status-go/pull/3172/commits/8c97f5aeffb5d9804cfa144fbb3f6c315c1ca32d

server/qrops.go
siddarthkay: fixed here : https://github.com/status-im/status-go/pull/3172/commits/14e8bb82545b5208326ad97e1227174acc19d645

images/manipulation.go
siddarthkay: fixed here : https://github.com/status-im/status-go/pull/3172/commits/a640c686e60535e219088499717f664c8b338a6e

Samyoul (approved):
Super, good job. Thank you.

PR #3174, opened 2023-02-09
take care of kdf when local pairing

qfrank:
when doing local pairing, use status-desktop as sender and mobile as receiver, use mobile to scan sync code, after done local pairing, it will create app-db file with KDFIterations 0, but `KDFIterations` in db on mobile is 3200, so it's inconsistent which shouldn't be, and when we login again, status-mobile won't pass KDFIterations stored in DB(3200) to status-go, we should use 3200(get from db) if status-mobile didn't pass this parameter when login

### DIFF
--- a/api/geth_backend.go
+++ b/api/geth_backend.go
@@ -183,7 +183,7 @@ func (b *GethStatusBackend) SaveAccount(account multiaccounts.Account) error {
 	if b.multiaccountsDB == nil {
 		return errors.New("accounts db wasn't initialized")
 	}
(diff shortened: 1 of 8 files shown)


qfrank:
BTW, i've noticed there was a [PR Decrease number of kdf iterations in pfs database](https://github.com/status-im/status-go/pull/1343) to improve performance relate to kdf, `KDFIterations` synced from desktop is 256000, not sure how much impact will take to the performance for mobile ATM. @cammellos @Samyoul

PR #3176, opened 2023-02-09
Init keystore fix

cammellos:
We delegate to services to init the keystore file, but there's a race condition that can be triggered, whereby the keystore file is not initialized before login, and that results in the user not being able to login.

This commit changes the services so that the keystore is initialized before login, I have left also the code in services just to make sure nothing relies on it, but it could probably be removed. Initializing keystore twice seems like a noop if already initialized.

### DIFF
--- a/api/geth_backend.go
+++ b/api/geth_backend.go
@@ -929,10 +929,18 @@ func (b *GethStatusBackend) startNode(config *params.NodeConfig) (err error) {
 		return err
 	}
 
+	if b.accountManager.GetManager() == nil {
+		err = b.accountManager.InitKeystore(config.KeyStoreDir)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


richard-ramos (approved):

richard-ramos:
Seems like this race condition happens very rarely? i've never seen it happening on desktop

cammellos:
> Seems like this race condition happens very rarely? i've never seen it happening on desktop

fairly rare, but it happens to one of our QA, not all of them, and they have to login/logout/login/logout/login/logout, so fairly difficult to replicate, it might also be due to the fact that mobile are low powered devices and takes longer to login etc

though once a release is out, I am sure will happen more often :)

PR #3179, opened 2023-02-10
support passing kdf and node config from receiver when local pairing

### LINKED ISSUE #3166: Support passing parameters relate to node configuration when doing local pair sync
When doing local pair, we need some information to implement login automatically on the receiver side. e.g. RootDataDir/LogFile/LogDir,  current implementation calculate these based on `keystorePath` which is a temporary solution.
We should make [InputConnectionStringForBootstrapping](https://github.com/status-im/status-go/blob/fa77bba3033aaa5374dae66c295df16f942a2e27/mobile/status.go#L1006)  / [GetConnectionStringForBeingBootstrapped](https://github.com/status-im/status-go/blob/fa77bba3033aaa5374dae66c295df16f942a2e27/mobile/status.go#L965) support passing relate parameters.

Relate [PR and comment](https://github.com/status-im/status-go/pull/3159#pullrequestreview-1284870399)

qfrank:
Closes #3166 

this PR contains break changes, since it won't support sync node config from sender, so we need wait frontend to finish corresponding changes. then we can merge this PR.

Status: ready to review.


cammellos (approved):
looks good to me, it would be good to have @Samyoul eyes on it

server/pairing/raw_message_handler.go
```diff
@@ -5,18 +5,15 @@ import (
...
 	"github.com/status-im/status-go/multiaccounts/settings"
-
-	"github.com/golang/protobuf/proto"
+	"github.com/status-im/status-go/params"
 
 	"github.com/status-im/status-go/api"
 	"github.com/status-im/status-go/multiaccounts"
 	"github.com/status-im/status-go/multiaccounts/accounts"
-	"github.com/status-im/status-go/params"
 	"github.com/status-im/status-go/protocol/protobuf"
```
Samyoul: This seems a strange import order. Could the order look more like:

```go
	"github.com/status-im/status-go/api"
	"github.com/status-im/status-go/multiaccounts/settings"
```

This way all the project imports are together

server/pairing/payload_manager.go
```diff
@@ -9,6 +9,8 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/status-im/status-go/params"
+
 	"github.com/status-im/status-go/api"
```
Samyoul: Hmm, I wonder if there is something wrong with the linter.

The pattern of imports that it normally enforces is:

Group together in alphabetical

- standard library
- 3rd party librarys
- `ethereum/go-ethereum` & `status-im/status-go`

server/pairing/raw_message_handler.go
qfrank: emm.. maybe something better u'd like
```
	"github.com/google/uuid"

	"github.com/status-im/status-go/api"
	"github.com/status-im/status-go/multiaccounts/settings"
```

server/pairing/payload_manager.go
qfrank: > I wonder if there is something wrong with the linter

it must be.
let me check all the files, any other issue for the review? 🙂

server/pairing/payload_manager.go
```diff
@@ -22,7 +24,10 @@ import (
 )
 
 var (
-	ErrKeyFileAlreadyExists = errors.New("key file already exists")
+	ErrKeyFileAlreadyExists       = errors.New("key file already exists")
+	ErrKeyUIDEmptyAsSender        = errors.New("keyUID must be provided as sender")
+	ErrNodeConfigNilAsReceiver    = errors.New("node config must be provided as receiver")
+	ErrPayloadSourceConfigBothSet = errors.New("payloadSourceSenderConfig and payloadSourceReceiverConfig cannot be both set")
```
Samyoul, line 28: This is an interesting error. The fact that it warns it is possible to received both Sender and Receiver types is a sign that perhaps the config should be refactored. I'll add some notes below:

server/pairing/payload_manager.go
```diff
@@ -49,18 +54,71 @@ type PayloadManager interface {
...
+
+type PayloadSourceReceiverConfig struct {
+	KDFIterations int `json:"kdfIterations"`
+	NodeConfig    *params.NodeConfig
+	// this field already exists within params.NodeConfig, but it doesn't support json marshalling, so we need to duplicate it here
+	RootDataDir string
+	// corresponding to field current_network from table settings, so that we can override current network from sender
+	SettingCurrentNetwork string
+}
+
```
Samyoul, line 68: Rather than these 2 structs being embedded in `PayloadSourceConfig`, perhaps we should inverse the relationship.

A `PayloadSourceSenderConfig` should have an embedded `PayloadSourceConfig` because it is a kind of `PayloadSourceConfig`
A `PayloadSourceReceiverConfig ` should have an embedded `PayloadSourceConfig` because it is a kind of `PayloadSourceConfig`

But a `PayloadSourceConfig` is not a kind of `PayloadSourceSenderConfig` or `PayloadSourceReceiverConfig`, does it make sense to embed either?

Samyoul: Hmm, I notice that even with the changes in this PR that `PayloadSourceConfig` will only be left with one field `KeystorePath` which is used in two different ways:

- With keyUID
- Without keyUID

So perhaps the time of `PayloadSourceConfig` is over now.

Samyoul: Hmmm. If / when we merge my latest PR https://github.com/status-im/status-go/pull/3192/files#diff-737158e6a8fd20bd9226198c0c4ff00ae4d5c129c455a790194de3503b02bf4b my new field makes no sense in the context of `PayloadSourceConfig`. Because this change is not related to the payload source, it is specifically related to the `PairingServer` configuration and not `Sender` / `Receiver` configuration.

```go
	Timeout uint `json:"timeout"`
```

We should probably consider redesigning the configuration schema, because we are starting to grow it into more complex use cases.

qfrank: > Rather than these 2 structs being embedded in `PayloadSourceConfig`, perhaps we should inverse the relationship.
> 
> A `PayloadSourceSenderConfig` should have an embedded `PayloadSourceConfig` because it is a kind of `PayloadSourceConfig` A `PayloadSourceReceiverConfig ` should have an embedded `PayloadSourceConfig` because it is a kind of `PayloadSourceConfig`
> 
> But a `PayloadSourceConfig` is not a kind of `PayloadSourceSenderConfig` or `PayloadSourceReceiverConfig`, does it make sense to embed either?

FYI, I added these 2 types of config(sender/receiver) because i saw a problem that the param `configJSON` is not very clear for our relate API(GetConnectionStringForBeingBootstrapped/InputConnectionStringForBootstrapping etc.) in status.go

qfrank: > Hmmm. If / when we merge my latest PR https://github.com/status-im/status-go/pull/3192/files#diff-737158e6a8fd20bd9226198c0c4ff00ae4d5c129c455a790194de3503b02bf4b my new field makes no sense in the context of `PayloadSourceConfig`. Because this change is not related to the payload source, it is specifically related to the `PairingServer` configuration and not `Sender` / `Receiver` configuration.
> 
> ```go
> 	// Timeout the number of milliseconds after which the pairing server will automatically terminate
> 	Timeout uint `json:"timeout"`
> ```
> 
> We should probably consider redesigning the configuration schema, because we are starting to grow it into more complex use cases.

agree, we need a refactor

Samyoul: Hmm . We could have `PairingConfig` which is a global level config, with sub configs. But then I'll be a hypocrite and completely contradict my whole point I'm making here https://github.com/status-im/status-go/pull/3179/files#r1107080212 🤦 , maybe we can collaborate on this.

qfrank: for the time being, can we leave the refactor in a separate PR? 🤔

Samyoul: Yes, that's fine I'll raise an issue on this point

qfrank: thank you

Samyoul (approved):
Thank you Frank, this looks great. I've created the issue for the configuration complexity.

PR #3183, opened 2023-02-13
Feat/8599 update saved addresses

IvanBelyakoff:
Update saved addresses DB to store ENS, isTest flag and chains
prefixes. Changed primary keys and API methods.

Fixes #8599

Important changes:
- [x] Removed ChainID from saved addresses DB as not needed and not really used
- [x] Decision made to not support same addresses in DB with different names for multiple reasons:

1. Using name + is_test fields as PRIMARY KEY will not work in case we want to change the name of the saved address - a new records is created.
2. Adding a new GUID column as PRIMARY KEY makes DB migration more complex since it must be NOT NULL and old database does not have that column, so need extra efforts to fill that column and pass GUID from UI for new records.
3. Not really an important feature

- [x] Replaced a tmp DB file with in memory DB to speed up testing
- [x] Fixed tests and added new ones 

Closes #8599


ghost:
### Pull Request Checklist

services/wallet/saved_addresses.go
```diff
@@ -15,16 +15,18 @@ type savedAddressMeta struct {
...
 type SavedAddress struct {
 	Address common.Address `json:"address"`
-	// TODO: Add Emoji and Networks
-	// Emoji    string         `json:"emoji"`
+	// TODO: Add Emoji
+	// Emoji	string		   `json:"emoji"`
 	Name      string `json:"name"`
 	Favourite bool   `json:"favourite"`
-	ChainID   uint64 `json:"chainId"`
+	Networks  string `json:"chains"` // used with address only, not with ENSName
```
alaibe: nit, why naming this Networks, instead of ChainIDs, this is what we use in wallet?

services/ext/api.go
```diff
@@ -914,19 +914,11 @@ func (api *PublicAPI) DestroyWalletConnectSession(ctx context.Context, PeerID st
...
 	return api.service.messenger.UpsertSavedAddress(ctx, sa)
 }
 
-func (api *PublicAPI) DeleteSavedAddress(ctx context.Context, chainID uint64, address ethcommon.Address) error {
-	if chainID == 0 {
-		chainID = api.service.rpcClient.UpstreamChainID
-	}
-
-	return api.service.messenger.DeleteSavedAddress(ctx, chainID, address)
+func (api *PublicAPI) DeleteSavedAddress(ctx context.Context, address ethcommon.Address, ens string, isTest bool) error {
```
alaibe, line 924: isTest can be inferred from the settings, so that u don't need to expose this in the API?

services/wallet/api.go
```diff
@@ -214,24 +214,21 @@ func (api *API) DeleteCustomTokenByChainID(ctx context.Context, chainID uint64,
...
-	if sa.ChainID == 0 {
-		sa.ChainID = api.s.rpcClient.UpstreamChainID
-	}
 	_, err := api.s.savedAddressesManager.UpdateMetadataAndUpsertSavedAddress(sa)
 	log.Debug("result from database for create or edit saved address", "err", err)
 	return err
 }
 
-func (api *API) DeleteSavedAddress(ctx context.Context, address common.Address) error {
+func (api *API) DeleteSavedAddress(ctx context.Context, address common.Address, ens string, isTest bool) error {
```
alaibe, line 217: isTest can be inferred right?

appdatabase/migrations/sql/1676279805_add_chains_ens_istest_to_saved_addresses.up.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE IF NOT EXISTS saved_addresses_new (
+  address VARCHAR NOT NULL,
+  name TEXT NOT NULL,
+  favourite BOOLEAN NOT NULL DEFAULT FALSE,
+  removed BOOLEAN NOT NULL DEFAULT FALSE,
+  update_clock INT NOT NULL DEFAULT FALSE,
+  chains VARCHAR DEFAULT "",
```
alaibe: nit: rename to chainIDs ?

services/wallet/saved_addresses.go
IvanBelyakoff: Initially I did not remove ChainID, so I thought maybe it is not good to have ChainID and ChainIDs, ans since in the code - for example in database ChainId was called network_id, hence the name.
But the question is the same as below - the field stores chains short names as entered by user: "eth:opt:".
What do you think, ChainShortNames is good enough? If call it ChainIDs, I would expect 1 for Eth, 42161 for Arbitrum

services/ext/api.go
IvanBelyakoff: I was thinking about it, but I thought maybe it is more reasonable to not add an extra dependency on other service at this low level unless necessary to avoid tighter coupling and have it as a param from higher level instead. So that was more like an architectural decision.
Do you think it is better to have retrieve the value from `multiaccounts` service?

appdatabase/migrations/sql/1676279805_add_chains_ens_istest_to_saved_addresses.up.sql
IvanBelyakoff: what about `chain_ids` - not to mix camel and snake cases in the table. Is it good?

IvanBelyakoff: Btw, they are prefixes as this is a string. So the values are for example. "eth:opt:"
that is why I did not use "ids" word. Or you think it is still good?

services/wallet/saved_addresses.go
alaibe: yes for: ChainShortNames

appdatabase/migrations/sql/1676279805_add_chains_ens_istest_to_saved_addresses.up.sql
alaibe: as u suggested before: chain_short_names makes sense

services/wallet/api.go
alaibe: maybe not in the API but savedAddressesManager should have access to the multiaccounts

services/ext/api.go
IvanBelyakoff: @alaibe there is one case I discovered when adding more sync tests, why I decided to drop the fix to get isTest flag from multiaccounts service. Here it is:
- First device has test mode enabled
- Second device has normal mode enabled
- when syncing saved addresses happens, isTest flag is taken from accounts service on both devices and they are different, so syncing fails in case of deletion of saved address. When it is passed along with other DB primary keys, it works as expected.
Please let me know if you agree to keep isTest param in parameters list

stefandunca (approved):
Great work!

alaibe (approved):

appdatabase/migrations/sql/1676279805_add_chains_ens_istest_to_saved_addresses.up.sql
```diff
@@ -0,0 +1,17 @@
...
+  update_clock INT NOT NULL DEFAULT FALSE,
+  chain_short_names VARCHAR DEFAULT "",
+  ens_name VARCHAR DEFAULT "",
+  is_test BOOLEAN DEFAULT FALSE,
+  PRIMARY KEY (address, ens_name, is_test)
+) WITHOUT ROWID;
+
+INSERT INTO saved_addresses_new (address, name, favourite, removed, update_clock)
+   SELECT address, name, favourite, removed, update_clock FROM saved_addresses;
+ALTER TABLE saved_addresses RENAME TO saved_addresses_old;
```
saledjenic: @IvanBelyakoff you don't need this line, you could do `DROP TABLE saved_addresses;` here and remove line 17.

protocol/protobuf/pairing.proto
```diff
@@ -223,9 +223,11 @@ message SyncSavedAddress {
   bytes  address = 1;
   string name = 2;
   bool   favourite = 3;
-  uint64 chain_id = 4;
```
saledjenic, line 231: @IvanBelyakoff in such cases it's better to use `uint64 chain_id = 4 [deprecated = true];` or `reserved 4;` then deleting it, cause that will prevent users from using it.

protocol/protobuf/pairing.proto
```diff
@@ -223,9 +223,11 @@ message SyncSavedAddress {
...
   bool   removed = 5;
   uint64 update_clock = 7;
```
saledjenic, line 233: @IvanBelyakoff maybe we could use just `clock`, just for the sake of consistency, to stick to the way how it's done for other messages.

services/wallet/saved_addresses.go
```diff
@@ -54,8 +56,13 @@ func getSavedAddressesFromDBRows(rows *sql.Rows) ([]SavedAddress, error) {
 	return addresses, nil
 }
 
-func (sam *SavedAddressesManager) GetSavedAddressesForChainID(chainID uint64) ([]SavedAddress, error) {
-	rows, err := sam.db.Query(fmt.Sprintf("SELECT %s FROM saved_addresses WHERE network_id = ? AND removed != 1", rawQueryColumnsOrder), chainID)
+func (sam *SavedAddressesManager) getSavedAddresses(condition string) ([]SavedAddress, error) {
+	var whereCondition string
+	if condition != "" {
+		whereCondition = "WHERE " + condition
```
saledjenic:
```suggestion
whereCondition = fmt.Sprintf("WHERE %s", condition)
```

services/wallet/saved_addresses_test.go
```diff
@@ -14,49 +13,50 @@ import (
...
+	ensMember int = iota
+	isTestMember
+	addressMember
+)
+
 func setupTestSavedAddressesDB(t *testing.T) (*SavedAddressesManager, func()) {
-	tmpfile, err := ioutil.TempFile("", "wallet-saved_addresses-tests-")
-	require.NoError(t, err)
-	db, err := appdatabase.InitializeDB(tmpfile.Name(), "wallet-saved_addresses-tests", sqlite.ReducedKDFIterationsNumber)
+	db, err := appdatabase.InitializeDB(sqlite.InMemoryPath, "wallet-saved_addresses-tests", sqlite.ReducedKDFIterationsNumber)
```
saledjenic, line 23: @IvanBelyakoff why did you change this?

Not sure, but maybe this could work as well:
```
func setupTestSavedAddressesDB(t *testing.T) (* SavedAddressesManager, func()) {
	db, stop, err := appdatabase.SetupTestSQLDB("wallet-saved_addresses-tests")
	if err != nil {
		require.NoError(t, stop())
	}
	require.NoError(t, err)

        return &SavedAddressesManager{db}, func() {
		require.NoError(t, stop())
	}
}
```

appdatabase/migrations/sql/1676279805_add_chains_ens_istest_to_saved_addresses.up.sql
IvanBelyakoff: Thank you for the feedback. That is just normal precautions when you want to overwrite something - make a backup, overwrite, and if that goes well - remove the backup.
In this case we might want to skip the backup part. I can do that

services/wallet/saved_addresses_test.go
IvanBelyakoff: I did that because file creation and working with DB on hard drive takes a lot of time and I have found that there is already an option to use in-memory DB. So I think for test cases we don't need a DB on hard drive. And my tests show 10 times performance gain (5 sec for wallet tests -> 0.5).

IvanBelyakoff: SetupTestSQLDB - creates a file as well, which is slow. Do we really need to create a DB on hard drive for test suites?

protocol/protobuf/pairing.proto
IvanBelyakoff: you mean `clock` instead of `update_clock` ? I think I will have to add something like `create_clock` soon, to be able to sort records by creation date (requirement from Ben was that a new saved address must popup at the top of the list, and update_clock will not work for this purpose, as it will change the order of the items in the list for every single change on any record -like clicking `favourite` button). So I think just single "clock" will not be enough anyways

appdatabase/migrations/sql/1676279805_add_chains_ens_istest_to_saved_addresses.up.sql
saledjenic: In general you don't have to do anything, but just to clarify:

- what's done:
from `saved_addresses` to `saved_addresses_new`
rename `saved_addresses` to `saved_addresses_old`
rename `saved_addresses_new` to  `saved_addresses`
delete `saved_addresses_old`

- what I suggested:
from `saved_addresses` to `saved_addresses_new`
delete `saved_addresses`
rename `saved_addresses_new` to  `saved_addresses`

protocol/protobuf/pairing.proto
saledjenic: If that's the request than ok. Will we have to select among sort criteria, like `newest first`/`most recently updated first`? If there are both options, then ok, but if there is no options I guess it should be sort by `most recently updated first`.

saledjenic (approved):

protocol/protobuf/pairing.proto
stefandunca: If I remember correctly, for other implementations I had as an example, the clock is coming from the messenger (lamport clock maybe?). In the case of the saved address, this is the user's wall clock from when the entry was updated independent of sync or messenger. It is used to solve editing conflicts.

services/wallet/saved_addresses_test.go
stefandunca: Great thinking @IvanBelyakoff. Usually, the file I/O is much slower than the random memory access, which should provide considerable gains in the development environment and a good speed up in the CI. Most of the tests, should not require file-based storage (for sure, the ones that don't deal with user folder management).

Let's bring this matter to the `status-go` channel and find out if anybody notices any issue with bringing this improvement to all tests (`SetupTestSQLDB ` and general setup entries)

PR #3184, opened 2023-02-13
LocalPairing minor fixes. `ValidateConnectionString` method.

igor-sirotin:
Required by https://github.com/status-im/status-desktop/pull/9482

- Added `ValidateConnectionString` public method
- Added check if connection string is longer then 2 (to read `connectionStringID`)
- Other minor improvements and comments


mobile/status.go
```diff
@@ -1003,7 +1003,7 @@ func GetConnectionStringForBootstrappingAnotherDevice(configJSON string) string
 //
 // Example: A mobile device (device with a camera) receiving account data from
 // a device with a screen (mobile or desktop devices)
-func InputConnectionStringForBootstrapping(cs, configJSON string) string {
+func InputConnectionStringForBootstrapping(cs string, configJSON string) string {
 	if configJSON == "" {
```
qfrank: can we not change this, i didn't see this is a problem ..

server/pairing/client.go
```diff
@@ -267,7 +267,7 @@ func (c *Client) getChallenge() error {
 	return err
 }
 
-func StartUpPairingClient(backend *api.GethStatusBackend, cs, configJSON string) error {
+func StartUpPairingClient(backend *api.GethStatusBackend, cs string, configJSON string) error {
```
qfrank: similar to above

mobile/status.go
igor-sirotin: Well, I just thought i's a typo. Was it intended to be no-type? If so, why?

qfrank: when you see something like `cs, configJSON string` in golang, it means cs has the same type(string) as configJSON, so it's not no-type here 🙂

richard-ramos (approved):

qfrank (approved):

PR #3186, opened 2023-02-13
Fix duplicated notifications on receiving contact update

cammellos:
This commit fixes an issue where when accepting a contact request the other end would display an extra notification.

It also changes WaitOnResponse to collect results. This should make tests less flaky, since sometimes messages are processed in different batches.
Now we need to be though exact on what we expect from the response (i.e use == instead of >, otherwise the same behavior applies) This uncovered a couple of issues with messenger.Merge, so I have moved the struct to use a map based collection instead of an array.

the function `createcontactrequestnotification` badly needs refactoring, there's many paths that uses it and it's too many options, but for now, I am covering the behavior from the outside, so it can be safely refactored.


ghost:
### Pull Request Checklist

MishkaRogachev (approved):

qfrank (approved):

cammellos:
ops, merged to early, I'll revert if it fails to test, apologies!

ilmotta (commented):
LGTM

PR #3189, opened 2023-02-14
PR to fix media server sleep issue 

### LINKED ISSUE #15050: The images are not shown after the app came back from the background
**Steps to reproduce:**
1. Open 1-1 chat
2. Send images
3. Send the app into background mode
4. Send device into sleeping mode
5. WAIT AT LEAST 3-4 MINUTES
6. Go back to the app -> open the chat
7. Send images

**Actual result:**
Old and newly sent Images are not shown

https://user-images.githubusercontent.com/52490791/213688616-3c4d035e-df3b-4359-8d97-ab955fe2ed78.mp4

**Expected result:**
Images are still shown

**ENV:** iPhone 11 Pro max, v15
**Logs:** 
[Status.log](https://github.com/status-im/status-mobile/files/10465950/Status.log)
[geth.log](https://github.com/status-im/status-mobile/files/10465952/geth.log)

siddarthkay:
fixes https://github.com/status-im/status-mobile/issues/15050
status : ready


api/geth_backend.go
```diff
@@ -1289,32 +1289,34 @@ func (b *GethStatusBackend) ConnectionChange(typ string, expensive bool) {
...
+		messenger = b.statusNode.WakuExtService().Messenger()
+	}
 
-			if messenger != nil {
-				if s == appStateForeground {
-					messenger.ToForeground()
-				} else {
-					messenger.ToBackground()
-				}
+	if b.statusNode != nil {
```
cammellos: The nil checking needs to be before you use it,  you are checking it after you use it, you can use a guard pattern at the beginning:
Top of the function:
```

if b.statusNode == nil {
   return...
}
```
and remove the nil checking, similarly with messenger, just after you think you set it

```
if messenger == nil {
log.Warn("blah")
return ...
}
if s == appStateForeground  {
m.ToF.
} else {
m.ToBack
}
```

cammellos (changes requested):
Nil checking is a bit off, otherwise looks good, don't forget to bump version!

siddarthkay:
> don't forget to bump version!

I almost forgot about that, thanks for the reminder!

api/geth_backend.go
siddarthkay: Addressed here : https://github.com/status-im/status-go/pull/3189/commits/07205b79a2cb16c95e1742e230b928770103dca6

cammellos (approved):

PR #3191, opened 2023-02-14
feat: implement currency formatting calculation

dlipicar:
This PR implements Currency formatting calculation + caching in status-go.
It's done as part of https://github.com/status-im/status-desktop/issues/9453, in an effort to make formatting more lightweight and reliable. Moreover, it can now share logic with the mobile app.

Fiat currency formats are now calculated once when the service starts. Token formats are calculated when requested by the client. A signal is emitted every 1h for the client to request re-calculation of the token formats based on the last fetched prices and update its local cache.

List of changes:
- Moved `peg` from the tokenStore table to a separate one. Tidier, less error-prone and needed for when we start using the standardized token lists.
- Moved `price` functionality to a separate module.
- Removed use of the `price` db. Now a simple table is kept in memory.
- Added some functions to make use of the price cache in status-go. Some are used now, some will probably come handy for future changes.
- Replaced library `github.com/rmg/iso4217` with `github.com/ladydascalie/currency`, which exports the full list of available fiat currencies.
- Added `currency` module in charge of calculating and caching currency formatting information
- Removed some functions for the API which were used to calculate the CurrencyFormat on the client side, since that's no longer needed.

There's a couple of unit tests, and I tested this with the desktop client.


services/wallet/api.go
```diff
@@ -567,18 +560,12 @@ func (api *API) CreateMultiTransaction(ctx context.Context, multiTransaction *Mu
...
-	code, minor := iso4217.ByName(strings.ToUpper(name))
-
-	if code == 0 {
-		return code, errors.New("Unknown currency: " + name)
-	}
-
-	return minor, nil
+func (api *API) FetchAllCurrencyFormats() (currency.FormatPerSymbol, error) {
+	log.Debug("call to FetchAllCurrencyFormats")
+	return api.s.currency.FetchAllCurrencyFormats()
```
alaibe, line 559: Why cannot we have only one endpoint?

dlipicar: `GetCachedCurrencyFormats` is a quick, non-blocking operation so we use that during initialization, to ensure the client can properly represent currency amounts as strings even if there's no internet connection or whatever.

`FetchAllCurrencyFormats` might cause token prices to be fetched if they're not already cached. We don't know how long this will take, so we want to make it an async call.

services/wallet/currency/currency.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func calculateTokenDisplayDecimals(price float64) uint {
+	var displayDecimals float64 = 0.0
+
+	if price > 0 {
+		const lowerCurrencyResolution = 0.1
+		const higherCurrencyResolution = 0.01
+		lowerDecimalsBound := math.Max(0.0, math.Log10(price)-math.Log10(lowerCurrencyResolution))
+		upperDecimalsBound := math.Max(0.0, math.Log10(price)-math.Log10(higherCurrencyResolution))
```
stefandunca: I'm not sure I understand the logic behind this, can you please clarify? Also why not use +1 and +2 instead of `-math.Log10(...)`?

services/wallet/price/price.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	pm.updatePriceCache(result)
+
+	return result, nil
+}
+
+func (pm *Manager) getCachedPrices(symbols []string, currencies []string) DataPerTokenAndCurrency {
```
stefandunca: Using the same function name as `GetCachedPrices` is confusing. How about calling the other `getAllCachedPrices` or this one `getCachedPricesFor`

services/wallet/token/tokenstore.go
```diff
@@ -2,25 +2,44 @@ package token
 
 import "github.com/ethereum/go-ethereum/common"
 
+var tokenPeg = map[string]string{
```
stefandunca, line 5: Nice optimizaiton 👍🏼

vendor/github.com/ladydascalie/currency/.gitignore
```diff
@@ -0,0 +1 @@
+.idea/
```
stefandunca, line 1: Why not add it to the global `.gitignore`? It looks more practical.

stefandunca (approved):
Great work!
Please see some questions inline

saledjenic (approved):

vendor/github.com/ladydascalie/currency/.gitignore
dlipicar: Hmm this file was part of the dependency repo, not something I manually created, so I'd better leave it as-is

services/wallet/price/price.go
dlipicar: True, will do!

services/wallet/currency/currency.go
dlipicar: > I'm not sure I understand the logic behind this, can you please clarify?

This is supposed to reflect the specification (https://github.com/status-im/status-desktop/issues/8640).
You want as many decimal places as needed so that the last decimal place is worth between 1 and 10 USD cents.

If you've got `N` decimals available, the value of your last decimal place (your resolution) is `10ˆ(-N)`
Example: 2 decimals gives you a resolution of 0.01, 3 decimals gives you 0.001

That's in token units, so how much is your last decimal place worth in USD? If `P` is your price in USD, your resolution is worth `P * 10ˆ(-N)`

You want your resolution to be worth `Z` USD, then:
`Z = P * 10ˆ(-N)`
`log10(Z) = log10(P * 10ˆ(-N))`
`log10(Z) = log10(P) + log10(10ˆ(-N))`
`log10(Z) = log10(P) - N`
`N = log10(P) - log10(Z)`

> Also why not use +1 and +2 instead of -math.Log10(...)?

0.1 and 0.01 (amount in USD) are the limits set by the specification, which I guess could change if John changes his mind in the future. That's why I decided to leave the calculation in place instead of using the pre-calculated results of those `Log` of those values. I'm gonna assume the compiler is smart enough to do that optimization for me.

services/wallet/price/price.go
dlipicar: Done

services/wallet/currency/currency.go
stefandunca: Thanks, @dlipicar, for the great clarifications.

For configuration values, like `lowerCurrencyResolution` and `higherCurrencyResolution`, I, personally, like having them at the top of the code with comments for easy finding not deep down in the code, not sure if this is the case.

vendor/github.com/ladydascalie/currency/.gitignore
stefandunca: I missed that its part of the external library 🤦‍♂️

services/wallet/currency/currency.go
dlipicar: You're right... Changed the location and gave it a clearer name

PR #3192, opened 2023-02-14
Added timeout functionality to Servers

Samyoul:
## What's Changed?

I've added timeout functionality to the Server types. Currently it is only possible to timeout a `PairingServer`.

This is configurable only via:

```go
type PayloadSourceConfig struct {
	// required
	KeystorePath string `json:"keystorePath"`
	// following 2 fields r optional.
	// optional cases:
	// 1. server mode is Receiving and server side doesn't contain this info
	// 2. server mode is Sending and client side doesn't contain this info
	// they are required in other cases
	KeyUID   string `json:"keyUID"`
	Password string `json:"password"`

	Timeout uint `json:"timeout"`
}
```

With the new Pairing config field `timeout`, consuming code can set a pairing server to terminate after n milliseconds.

## Why Make the Change?

Sometimes we want the Servers to only live for a specific amount of time. Setting a timeout will allow consuming code to startup a server and not need to manage termination calls.

---

## Follow Up

Add functionality that ensure the TLS cert expires in sync with the timeout.

### DIFF
--- a/server/timeout.go
+++ b/server/timeout.go
@@ -3,8 +3,6 @@ package server
 import (
 	"sync"
 	"time"
-
-	"go.uber.org/zap"
 )
 
 // timeoutManager represents a discrete encapsulation of timeout functionality.
@@ -20,15 +18,12 @@ type timeoutManager struct {
 	// exitQueue handles the cancel signal channels that circumvent timeout operations and prevent the
 	// execution of any `terminate` func()
 	exitQueue *exitQueueManager
-
-	logger *zap.Logger
 }
 
 // newTimeoutManager returns a fully qualified and initialised timeoutManager
-func newTimeoutManager(logger *zap.Logger) *timeoutManager {
+func newTimeoutManager() *timeoutManager {
 	return &timeoutManager{
-		exitQueue: &exitQueueManager{queue: []chan struct{}{}, logger: logger},
-		logger:    logger,
+		exitQueue: &exitQueueManager{queue: []chan struct{}{}},
 	}
 }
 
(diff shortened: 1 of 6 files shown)


server/timeout.go
```diff
@@ -0,0 +1,92 @@
...
+
+// add handles new exit channels adding them to the exit queue
+func (e *exitQueueManager) add(exit chan struct{}) {
+	e.queueLock.Lock()
+	defer e.queueLock.Unlock()
+
+	e.queue = append(e.queue, exit)
+}
+
+// empty sends a signal to evert exit channel in the queue and then resets the queue
```
siddarthkay: maybe a typo here
evert exit -> revert exit

Samyoul: Oh it should say "every"

server/timeout_test.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+		to, err := rand.Int(rand.Reader, big.NewInt(11))
+		if err != nil {
+			t.Error(err)
+		}
+
+		tm.SetTimeout(uint(to.Int64() * 10))
+
+		if b.Int64() == 1 {
+			tm.StartTimeout(t.FailNow)
```
qfrank, line 44: this PR looks good except this test, especially this for loop, I was wondering if there is a possible that it will take up > `uint(to.Int64() * 10)` so that this test may fail 🙂

Samyoul: The minimum timeout will be 10 milliseconds which is a long time for a function like this to run. However, you have highlighted an important point 💡.

If the timeout is 10 milliseconds and is the final call in the loop, it is possible that the `t.FailNow` could be called before line 52 is executed. So to prevent this I will add a `tm.StopTimeout()` between line 47 and 48.

The purpose of this loop is to flood the timeout manager with multiple random calls. If the process blocks or hangs anywhere `FailNow` should be triggered.

I could increase the timeout multiplier from `10` to `20` milliseconds, meaning that the loop has 20 - 200 milliseconds before a `FailNow`. That is a **lot** of time for this sequence to execute, if it fails under this circumstance something has blocked, deadlocked or the machine running the test is really really slow.

What do you think @qfrank ?

qfrank: sounds good 🫶

Samyoul: I played around with this and it seems that at least on my machine 2 milliseconds is safe, 1 millisecond is not safe. I played around with the below code.

```go
if b.Int64() == 1 {
	tm.StartTimeout(t.FailNow)
	time.Sleep(time.Duration(tm.timeout-2) * time.Millisecond)
} else {
	tm.StopTimeout()
}
```

Another thing I noticed without the sleep calls, the entire test takes ~20 milliseconds to run, with the sleeps it takes ~2000 milliseconds. So that suggests to me that unless a machine is more than 100 times slower than my machine there won't be problems.

But I do take the point that this test in strict terms is flaky and non-deterministic.

qfrank (approved):

PR #3194, opened 2023-02-14
Sync display names

cammellos:
Turned out we never synced display names :)

It was a bit difficult to catch since many of them were re-fetched by the network.
This commit adds syncing of display names (photos are too expensive to sync).

We check `lastUpdated` to honor the remote user decision about what's the last, but we don't set it, since it's used to track multiple fields (image & display name), and that would cause some issues, I left an explanation on the issue.

The fix would be to either add syncing of images (but that would be expensive, at least for backups for example), or to split the two clocks (always a safe option).

This is to be included in release, ideally by desktop as well cc @iurimatias so I'd avoid any large change


protocol/messenger_handler.go
```diff
@@ -499,6 +499,30 @@ func (m *Messenger) HandleSyncInstallationContact(state *ReceivedMessageState, m
...
+	// We don't set `LastUpdated`, since that would cause some issues
+	// as `LastUpdated` tracks both display name & picture.
+	// The case where it would break is as follow:
+	// 1) User A pairs A1 with device A2
+	// 2) User B publishes display name and picture with LastUpdated = 3
+	// 3) Device A1 receives display name & picture with LastUpdated = 3
+	// 4) Device A1 syncs with A2 (which has not received the previous message)
+	// 5) Device A2 saves Display name and sets LastUpdated = 3,
+	//    note that picture has not been set as it's not synced
+	// 6) Device A2 receives the message from 2. Because LastUpdated is 3
```
ilmotta: I read half a dozen times to really get the problem you're describing. Thank you, very useful comment :100: 

- Tiny typo on `6)`, "Because LastUpdated is 3" is duplicated.
- On step 3, it would've been easier to me if it was written: `3) Device A1 receives message from 2.`
- On step 4, it would've been a bit easier to follow with `4) Device A1 syncs with A2 (which has not received message from 3).`
- Nitpick: add the missing periods to sentences.

ilmotta (approved):

richard-ramos (approved):
Nice finding and solution!

anastasiyaig:
@iurimatias can we please include this into 0.10.0?

PR #3195, opened 2023-02-14
chore(keycard): keypair improvements

saledjenic:
Due to easier maintaining in future the following is done:
- keypairs table removed
- keycard table added, storing only keycards/keypairs
- keycard_accounts table added, storing only accounts migrated to a keycard

Migration is done keeping the current keycard state accurate (no keycard records will be lost).


appdatabase/migrations/sql/1676368933_keypairs_to_keycards.up.sql
```diff
@@ -0,0 +1,20 @@
...
+    FOREIGN KEY(keycard_uid) REFERENCES keycards(keycard_uid) 
+      ON UPDATE CASCADE 
+      ON DELETE CASCADE    
+);
+
+INSERT INTO keycards select distinct keycard_uid, keycard_name, keycard_locked, key_uid from keypairs;
+
+INSERT INTO keycards_accounts select keycard_uid, account_address from keypairs;
+
+DROP TABLE keypairs;
```
alaibe, line 20: does this need a sync with mobile team?

alaibe (approved):

appdatabase/migrations/sql/1676368933_keypairs_to_keycards.up.sql
saledjenic: I think no, they are not using this yet. Anyhow I've already posted a link to this PR in #status-go channel. 
@qfrank @rasom could you review/confirm?

qfrank: I'm not familiar with this, maybe @rasom knows better, but if u mean if mobile used api `GetMigratedKeyPairByKeyUID`, i'd say no since i just checked.

multiaccounts/keypairs/database.go
```diff
@@ -142,41 +150,114 @@ func (kp *KeyPairs) AddMigratedKeyPair(kcUID string, kpName string, KeyUID strin
 		_ = tx.Rollback()
 	}()
 
-	insert, err = tx.Prepare(`
+	var tmpKeyUID string
+	err = tx.QueryRow(`SELECT keycard_uid FROM keycards WHERE keycard_uid = ?`, kcUID).Scan(&tmpKeyUID)
+	if err != nil {
```
stefandunca, line 155: AFAIK this is true all the time.

stefandunca (approved):

PR #3197, opened 2023-02-15
feat(wallet): keep and propagate multi-transactions relation after transaction is complete

stefandunca:
## Updates [#7663](https://github.com/status-im/status-desktop/issues/7663)

Changes:
- Store multi-transaction ID into transactions DB
  - Add `TransactionManager.GetPendingEntry` API to get the ID and store it in `transfers`
- Delete pending transactions after downloading the first success status
- Add `multiTransactionID` to `transfer.View` to provide the ID to the presentation layer
- Move `TransactionManager` from `wallet` to `transfer` package

### DIFF
--- a/services/wallet/transfer/commands.go
+++ b/services/wallet/transfer/commands.go
@@ -2,6 +2,7 @@ package transfer
 
 import (
 	"context"
+	"database/sql"
 	"math/big"
 	"strings"
 	"time"
@@ -168,10 +169,11 @@ type controlCommand struct {
 	feed               *event.Feed
 	errorsCount        int
 	nonArchivalRPCNode bool
+	transactionManager *TransactionManager
 }
 
 func (c *controlCommand) LoadTransfers(ctx context.Context, downloader *ETHDownloader, limit int) (map[common.Address][]Transfer, error) {
-	return loadTransfers(ctx, c.accounts, c.block, c.db, c.chainClient, limit, make(map[common.Address][]*big.Int))
+	return loadTransfers(ctx, c.accounts, c.block, c.db, c.chainClient, limit, make(map[common.Address][]*big.Int), c.transactionManager)
 }
 
 func (c *controlCommand) Run(parent context.Context) error {
@@ -262,6 +264,7 @@ func (c *controlCommand) Run(parent context.Context) error {
 		signer:      types.NewLondonSigner(c.chainClient.ToBigInt()),
 		db:          c.db,
 	}
+
 	_, err = c.LoadTransfers(parent, downloader, 40)
 	if err != nil {
 		if c.NewError(err) {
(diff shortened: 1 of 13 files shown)


ghost:
### Pull Request Checklist

services/wallet/transfer/commands.go
```diff
@@ -355,10 +359,36 @@ func (c *transfersCommand) Run(ctx context.Context) (err error) {
...
+				if err == sql.ErrNoRows {
+					log.Warn("Pending transaction not found for", c.chainClient.ChainID, transfer.ID)
+				} else {
+					return err
+				}
+			} else {
+				transfer.MultiTransactionID = entry.MultiTransactionID
+				if transfer.Receipt != nil && transfer.Receipt.Status == types.ReceiptStatusSuccessful {
+					// TODO: Nim logic was deleting pending previously, should we notify UI about it?
+					err := c.transactionManager.DeletePending(c.chainClient.ChainID, transfer.ID)
```
alaibe, line 377: Did u check with mobile team that this won't break anything on their side?

alaibe (approved):
All good for me, just make sure mobile team also update their logic if needed

services/wallet/transfer/transaction.go
```diff
@@ -67,22 +78,28 @@ const (
...
+	From               common.Address         `json:"from"`
+	To                 common.Address         `json:"to"`
+	Data               string                 `json:"data"`
+	Symbol             string                 `json:"symbol"`
+	GasPrice           bigint.BigInt          `json:"gasPrice"`
+	GasLimit           bigint.BigInt          `json:"gasLimit"`
+	Type               PendingTrxType         `json:"type"`
+	AdditionalData     string                 `json:"additionalData"`
+	ChainID            uint64                 `json:"network_id"`
+	MultiTransactionID MultiTransactionIDType `json:"multi_transaction_id"`
```
dlipicar, line 93: Should we use `MultiTransactionIDType` for `MultiTransactionResult.ID` as well?

services/wallet/transfer/transaction.go
```diff
@@ -187,12 +198,47 @@ func (tm *TransactionManager) getPendingByAddress(chainIDs []uint64, address com
 	return transactions, nil
 }
 
-func (tm *TransactionManager) addPending(transaction PendingTransaction) error {
+// getPendingForChainAndHash returns sql.ErrNoRows if no pending transaction is found for the given identity
```
dlipicar: Looks like the method name should be updated in the comment

dlipicar (approved):

services/wallet/transfer/commands.go
stefandunca: @rasom please have a look at how will this line change impact the mobile side. You should have a `DeletePending` statusgo RPC call on your side, which should be redundant and fail after this change

services/wallet/transfer/transaction.go
stefandunca: I left it out for now because it is used in the public API; not sure if it doesn't require custom json parsing

services/wallet/transfer/commands.go
rasom: We haven't started re-implementation of the wallet UI yet, so as long as you don't change signatures in `mobile/status.go` (`status-go`'s non RPC API stays backward compatible) it is unlikely it will have any impact on mobile side at the moment.

rasom (approved):

services/wallet/transfer/commands.go
stefandunca: Thanks for the update

PR #3201, opened 2023-02-16
Support soft deletion for activity center notifications

ilmotta:
### Summary

This PR adds a new column named `deleted` to the table `activity_center_notifications`. 

I have heavily tested this branch in the mobile app. If interested, see the mobile PR https://github.com/status-im/status-mobile/pull/15106 for a lot more details of the feature.

#### Why?

Relying on the `dismissed` column for soft deletion is no longer viable because the mobile & desktop clients should display dismissed notifications (sometimes), hence the need for a new column to truly represent soft deletion.

#### Notes

The bulk of the work in this PR is in just two files `activity_center_persistence.go` and its test file.

- [x] Created a separate test file `activity_center_persistence_test.go`. All AC tests were extracted from `persistence_test.go`.
- [x] Added tests for the majority of functions I touched, in order to avoid regressions and document what they do.
- [x] Added a new endpoint named `DeleteActivityCenterNotifications`, so that the user can (soft) delete notifications.
- [x] Updated *unread count endpoint* to also take in consideration dismissed notifications (for the moment, used only by status-mobile).

**I expect no breaking changes from this PR.**, but @MishkaRogachev, if you could do a quick smoke test using this branch that'd very helpful.


rasom (approved):

protocol/migrations/sqlite/1676315400_add_deleted_column_to_activity_center_notifications_table.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE activity_center_notifications ADD COLUMN deleted BOOLEAN NOT NULL DEFAULT FALSE;
```
MishkaRogachev: Does `DEFAULT` constraint update existing columns? E.g do we need migration setting `deleted=false` for existing notification?

protocol/messenger_activity_center.go
```diff
@@ -235,6 +235,10 @@ func (m *Messenger) DismissActivityCenterNotifications(ctx context.Context, ids
 	return nil, err
 }
 
+func (m *Messenger) DeleteActivityCenterNotifications(ctx context.Context, ids []types.HexBytes, sync bool) error {
```
MishkaRogachev, line 273: Should we return response here saying what notifications were deleted? Or we expecting client well delete notifications on the ui by itself?

MishkaRogachev (approved):
Looks good! Just a few question to ensure understanding.
and i also added a ticket to desktop for supporting this feature https://github.com/status-im/status-desktop/issues/9568

protocol/migrations/sqlite/1676315400_add_deleted_column_to_activity_center_notifications_table.up.sql
ilmotta: In this case no @MishkaRogachev. If it was an existing column, then yes, using the `DEFAULT` constraint would not suffice, and a migration would be required.

protocol/messenger_activity_center.go
ilmotta: Good question. I went with the current "pattern" of defining an endpoint that's meant to be called in a *fire and forget* style by the clients. At least in status-mobile, a lot of calls to status-go are always assumed to work and the UI is optimistically updated and if an error happens it's only logged or sometimes even ignored :man_shrugging: 

This is something we should definitely improve at some point. But for the way things currently are, I think it's enough to only return nil or an error.

yqrashawn (approved):

protocol/activity_center_persistence.go
```diff
@@ -446,23 +468,34 @@ func (db sqlitePersistence) GetNotReadActivityCenterNotificationIds() ([][]byte,
...
+		SELECT a.id
+		FROM activity_center_notifications a
+		WHERE NOT a.deleted AND NOT a.dismissed AND NOT a.accepted
+		`)
 }
 
 func (db sqlitePersistence) HasPendingNotificationsForChat(chatID string) (bool, error) {
-	rows, err := db.db.Query("SELECT 1 FROM activity_center_notifications a WHERE a.chat_id = ? AND NOT a.dismissed AND NOT a.accepted", chatID)
+	rows, err := db.db.Query(`
+		SELECT 1 FROM activity_center_notifications a
```
cammellos, line 498: Maybe we want to check indexes on this query? (if they need to be modified/added etc)

protocol/activity_center_persistence.go
```diff
@@ -610,11 +644,39 @@ func (db sqlitePersistence) DismissAllActivityCenterNotifications() error {
 }
 
 func (db sqlitePersistence) DismissAllActivityCenterNotificationsFromUser(userPublicKey string) error {
-	_, err := db.db.Exec(`UPDATE activity_center_notifications SET read = 1, dismissed = 1 WHERE NOT dismissed AND NOT accepted AND author = ?`, userPublicKey)
+	_, err := db.db.Exec(`
+		UPDATE activity_center_notifications
+		SET read = 1, dismissed = 1
+		WHERE NOT deleted
```
cammellos: Similarly, we could check indexes here, just to make sure

cammellos (commented):
Looks good! Only a comment about indexes, maybe we want to check what's the state of things is

protocol/activity_center_persistence.go
ilmotta: Thanks for bringing this up. I did a rough investigation on how/if indexes are used for some queries related to the `activity_center_notifications` table.

SQLite is doing a full table scan when finding pending notifications. Once I add the following index, then it reports what we want:

```sql
CREATE INDEX activity_center_chat_id_deleted_dismissed_accepted_idx
  ON activity_center_notifications(chat_id, deleted, dismissed, accepted);

EXPLAIN QUERY PLAN
SELECT * FROM activity_center_notifications
WHERE chat_id = 'test-chat'
  AND NOT deleted
  AND NOT dismissed
  AND NOT accepted;

-- Reports:
-- SEARCH activity_center_notifications USING INDEX activity_center_chat_id_deleted_dismissed_accepted_idx (chat_id=?)
```

Currently, there are only two indexes, and they don't cover the query above, so I'm going to add the index to the migration.

- `activity_center_dimissed_accepted` (yes, there's a typo)
- `activity_center_read`
- The automatic index for the primary key

protocol/activity_center_persistence.go
ilmotta: Similar results from the other thread @cammellos. Currently there's no index covering the query (even in `develop` branch).

Running the following statements solves the problem.

```sql
CREATE INDEX activity_center_deleted_dismissed_accepted_author_idx
  ON activity_center_notifications(deleted, dismissed, accepted, author);

EXPLAIN QUERY PLAN
UPDATE activity_center_notifications
  SET read = 1, dismissed = 1
  WHERE NOT deleted AND NOT dismissed AND NOT accepted AND author = 'somebody';

-- Result:
-- SCAN activity_center_notifications USING COVERING INDEX activity_center_deleted_dismissed_accepted_author_idx
```

protocol/migrations/sqlite/1676315400_add_deleted_column_to_activity_center_notifications_table.up.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE activity_center_notifications
+  ADD COLUMN deleted BOOLEAN NOT NULL DEFAULT FALSE;
+
+CREATE INDEX activity_center_chat_id_deleted_dismissed_accepted_idx
```
ilmotta: These are the new indexes @cammellos. I've checked a couple queries in the `activity_center_persistence.go` file, and SQLite reported they were using indexes as expected. Is this good to go in your opinion? Anything you wish could be better?

protocol/activity_center_persistence.go
ilmotta: @cammellos, what do you think about the solution? Can I resolve this thread and the other ones about indexes or do you have other concerns related to that?

cammellos (approved):

PR #3202, opened 2023-02-17
support local pairing after logged in as receiver

qfrank:
## Summary Changes
- supported use case:
  1) user creates account on a1
  2) user restores with seed phrase on a2
  3) user scan qr code on a2, from a1 and goes through the pairing process
  4) expected behavior is that a1 syncs with a2 and viceversa, and they are now paired
- fixed some incorrect handling for logging.
- exchange installation data when local pairing
- support local pairing when account(if password is still the same on both side) already exist on the receiver side

[relate issue](https://github.com/status-im/status-mobile/issues/15140)


server/pairing/payload_manager.go
```diff
@@ -545,16 +551,28 @@ func (apr *AccountPayloadRepository) loadKeys(keyStorePath string) error {
 }
 
 func (apr *AccountPayloadRepository) StoreToSource() error {
+	keyUID := apr.multiaccount.KeyUID
+	if apr.loggedInKeyUID != "" && apr.loggedInKeyUID != keyUID {
+		return ErrLoggedInKeyUIDConflict
```
qfrank, line 570: not sure if this is what we want @Samyoul

server/pairing/payload_manager.go
```diff
@@ -125,6 +126,8 @@ func unmarshalPayloadSourceConfig(configJSON string, successCallback payloadSour
 type AccountPayloadManagerConfig struct {
 	DB *multiaccounts.Database
 	*PayloadSourceConfig
+	// only used for the receiver side
+	LoggedInKeyUID string
```
Samyoul, line 133: Thank you, this will be helpful for the config refactor 🙏

Samyoul:
@qfrank I wonder if I am misunderstanding something.

The issue you reference (https://github.com/status-im/status-mobile/issues/15140) details:

---
### In scope

- "Sign in" flow
- Sign in by syncing flow
- Scan QR
- Syncing devices screen
- Device sync complete screen
- Enable notifications screen

...

### Definitions

"Sign in" means restoring your account on a different device, and enable pairing between the two. For example I have an account on desktop, and I can "sign-in" on mobile, meaning I don't have the account on mobile, but I restore it from desktop, using the local pairing flow.

---

If I am not mistaken this issue, is describing local syncing, local pairing and then login from a single QR code. Effectively the same process we have in place now, except we are missing, if I'm not mistaken, the ability to exchange installation data between the two devices. Which is what currently happens when we initiate manual pairing, each device broadcasts via waku the below data:

```protobuf
message PairInstallation {
  uint64 clock = 1;
  string installation_id = 2;
  string device_type = 3;
  string name = 4;
}
```

Then each device independently authorises the sending of sync data to a channel readable by the device with the given `installation_id`.

This PR seems to implement something different, allowing a `local sync` between 2 devices that are already share the same private keys, but have a divergent database state. Which I'm sure is a feature we would want to use.

My question is, is this PR resolving the issue detailed above, or another issue?

qfrank:
> This PR seems to implement something different, allowing a local sync between 2 devices that are already share the same private keys, but have a divergent database state. Which I'm sure is a feature we would want to use.

Your feelings are right,  the steps within description of this PR comes from @cammellos and i think i ignored "and they are now paired", let me add the rest implementation. thank you for pointing it out 🙂 @Samyoul

qfrank:
ready to review 🙂 @Samyoul @cammellos @richard-ramos @siddarthkay

multiaccounts/database.go
```diff
@@ -208,7 +208,7 @@ func (db *Database) GetAccounts() (rst []Account, err error) {
 }
 
 func (db *Database) GetAccount(keyUID string) (*Account, error) {
-	rows, err := db.db.Query("SELECT  a.name, a.loginTimestamp, a.identicon, a.colorHash, a.colorId, a.keycardPairing, a.keyUid, ii.key_uid, ii.name, ii.image_payload, ii.width, ii.height, ii.file_size, ii.resize_target, ii.clock FROM accounts AS a LEFT JOIN identity_images AS ii ON ii.key_uid = a.keyUid WHERE a.keyUid = ? ORDER BY loginTimestamp DESC", keyUID)
+	rows, err := db.db.Query("SELECT  a.name, a.loginTimestamp, a.identicon, a.colorHash, a.colorId, a.keycardPairing, a.keyUid, a.kdfIterations, ii.key_uid, ii.name, ii.image_payload, ii.width, ii.height, ii.file_size, ii.resize_target, ii.clock FROM accounts AS a LEFT JOIN identity_images AS ii ON ii.key_uid = a.keyUid WHERE a.keyUid = ? ORDER BY loginTimestamp DESC", keyUID)
```
Samyoul, line 211: Would it make sense for this change to apply to `GetAccounts` also?

qfrank: emm, i think no need to change GetAccounts since it already  select kdfIterations before this PR

protocol/messenger.go
```diff
@@ -750,18 +750,21 @@ func (m *Messenger) Start() (*MessengerResponse, error) {
...
-
 	return response, nil
 }
 
 func (m *Messenger) IdentityPublicKey() *ecdsa.PublicKey {
 	return &m.identity.PublicKey
 }
 
+func (m *Messenger) IdentityPublicKeyInHex() string {
+	return types.EncodeHex(crypto.FromECDSAPub(m.IdentityPublicKey()))
```
cammellos: I believe there's already a function for this, `myHexIdentity`, we should probably use a single one (make it public as well if you re-use myHexIdentity, if needed outside this namesapce)

cammellos (commented):
Looks ok to me, just left a minor comment, if @Samyoul is happy we can move forward

cammellos (approved):

multiaccounts/database.go
Samyoul: 🤦 Ah, I see it. Sorry. I just presumed that both would be changed at the same time. Your change is actually the rebalancing change.

protocol/messenger.go
```diff
@@ -2499,7 +2507,7 @@ func (m *Messenger) getLastClockWithRelatedChat() (uint64, *Chat) {
 }
 
 // SendPairInstallation sends a pair installation message
-func (m *Messenger) SendPairInstallation(ctx context.Context) (*MessengerResponse, error) {
+func (m *Messenger) SendPairInstallation(ctx context.Context, rawMessageHandler RawMessageHandler) (*MessengerResponse, error) {
```
Samyoul, line 2516: This isn't something that should to be changed in this PR, but perhaps it could be in the scope for discussion.

First point I want to make is that I really like the way you capture the messages from `Messenger`. It is a nice solution, it reaches into `Messenger` grabs message data we want and then the `pairing` processes can take over from this point.

One thing I notice is that a lot of the syncing methods now have a nilable parameter, that seems to be nil in the majority of cases. Perhaps we could simplify the method logic and update `Messenger`'s state with a "RawMessageHandler" register.

The consuming code would look like something below:

```go
m.SetRawMessageHandler(rawMessageHandler)
m.SyncDevices()
m.SendPairInstallation(context.TODO)
m.ResetRawMessageHandler() // Defaults back to `dispatchPairInstallationMessage()`
```

Nothing to change now. But implementing this change would simplify the calling of all the functions that have an optional `RawMessageHandler `, future developers don't need to know what a `RawMessageHandler` is unless they want to change "broadcasting" behaviour.

Samyoul: I'll make a separate issue for this, the topic can be discussed there.

server/pairing/client.go
```diff
@@ -131,14 +188,50 @@ func (c *Client) sendSyncDeviceData() error {
...
 		signal.SendLocalPairingEvent(Event{Type: EventTransferError, Error: err.Error(), Action: ActionSyncDevice})
-		return fmt.Errorf("status not ok, received '%s'", resp.Status)
+		return err
 	}
 
 	signal.SendLocalPairingEvent(Event{Type: EventTransferSuccess, Action: ActionSyncDevice})
 	return nil
 }
 
+func (c *Client) receiveInstallationData() error {
```
Samyoul, line 200: These functions seem to only send one side of the installation data. Forgive me if I've misunderstood your work, but it seems that only `receiving` device gets the installation data of its counterpart.

Where as the 2 devices should exchange their installation data.

For example a `receiving client` would call `receiveInstallationData()` then `POST` it's installation data and receive the `sending server`'s installation data in the http response.

In the case of a `sending client` it would call `sendInstallationData()` then `POST` it's installation data and receive the `sending server`'s installation data in the http response.

Ah, so actually, you only need one client and one server function for this behaviour, as both can be exchanged in a single HTTPS request/response . And the process flow for both should be identical.

@qfrank Does that sound right to you?

qfrank: My point was: `protobuf.PairInstallation` is also a kind of RawMessage so we can handle it within [function HandleSyncRawMessages](https://github.com/status-im/status-go/blob/6cbeedd880fd76a9e13a2b6ae45303029d06e347/protocol/messenger_sync_raw_messages.go#L22) for receiver, but what about sender? how can it get installation data from receiver? the solution was endpoints `receiveInstallation/sendInstallation` was created to resolve this. So it does  implemented exchanging installation data 
in this PR, otherwise [relate test 1](https://github.com/status-im/status-go/blob/6cbeedd880fd76a9e13a2b6ae45303029d06e347/server/pairing/sync_device_test.go#L280-L281) and [relate test 2](https://github.com/status-im/status-go/blob/6cbeedd880fd76a9e13a2b6ae45303029d06e347/server/pairing/sync_device_test.go#L194-L195) would fail. There was a [comment](https://github.com/status-im/status-go/blob/6cbeedd880fd76a9e13a2b6ae45303029d06e347/server/pairing/client.go#L128-L129) to explain `PairInstallation`. 
Above explaination sounds good? 🙂@Samyoul

Samyoul: Ah, No I didn't have any objection with specific server endpoints to send / receive the installation data. I didn't realise that the installation data of the sender was sent in a previous call. I thought that this PR introduced all installation data exchange, rather than equalising the exchange of installation data.

Your tests make this point very clear. Thank you for your patience 😆.

Samyoul (approved):
Great work as always, thank you so much for this PR.

PR #3204, opened 2023-02-17
feat: add fallback rpc client

alaibe:
* Unify query to chain via one unique client
* Add a fallback client
* Use circuit breaker


appdatabase/migrations/sql/1676536757_add_fallback_rpc_to_networks.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE networks ADD COLUMN fallback_url VARCHAR NOT NULL DEFAULT "";
```
richard-ramos: In other migrations I saw that after adding a column, somethign like this is executed immediatly:
```
UPDATE networks SET fallback_url = "";
```

richard-ramos: If possible, please remove migrationsprevnodecfg/bindata.go from this commit as it wasnt really affected by the PR

services/wallet/service.go
```diff
@@ -156,9 +155,9 @@ func (s *Service) IsStarted() bool {
 
 func (s *Service) CheckConnected(ctx context.Context) *ConnectedResult {
 	infura := make(map[uint64]bool)
-	for chainID, client := range chain.ChainClientInstances {
-		infura[chainID] = client.IsConnected
-	}
+	// for chainID, client := range chain.ChainClientInstances {
+	// 	infura[chainID] = client.IsConnected
+	// }
```
richard-ramos: Is this not needed anymore?

rpc/chain/client.go
```diff
@@ -0,0 +1,678 @@
...
+	}
+}
+
+func (c *ClientWithFallback) Close() {
+	c.main.Close()
+	c.fallback.Close()
+}
+
+func (c *ClientWithFallback) makeCallNoReturn(main func() error, fallback func() error) error {
+	output := make(chan bool, 1)
```
richard-ramos:
```suggestion
	output := make(chan struct{}, 1)
```

rpc/chain/client.go
```diff
@@ -0,0 +1,678 @@
...
+	errChan := hystrix.Go(fmt.Sprintf("ethClient_%d", c.ChainID), func() error {
+		err := main()
+		if err != nil {
+			return err
+		}
+		output <- true
```
richard-ramos:
```suggestion
		output <- struct{}{}
```

rpc/chain/client.go
```diff
@@ -0,0 +1,678 @@
...
+	}, func(err error) error {
+		if c.fallback == nil {
+			return err
+		}
+
+		err = fallback()
+		if err != nil {
+			return err
+		}
+		output <- true
```
richard-ramos:
```suggestion
		output <- struct{}{}
```

rpc/chain/client.go
richard-ramos: If the output value is not used, then a signal channel can be used (these don't use memory at all and are cooler ;) )

appdatabase/migrations/sql/1676536757_add_fallback_rpc_to_networks.up.sql
alaibe: Yes to both, done

services/wallet/service.go
alaibe: Added back with new logic

rpc/chain/client.go
```diff
@@ -0,0 +1,691 @@
...
+		SleepWindow:           300000,
+		ErrorPercentThreshold: 25,
+	})
+
+	return &ClientWithFallback{
+		ChainID:     chainID,
+		main:        ethclient.NewClient(main),
+		fallback:    nil,
+		mainRpc:     main,
+		fallbackRpc: main,
```
stefandunca: Doesn't this mean it will retry two times? I'm wondering if this is intentional and if there is no problem retrying using the same client.

rpc/chain/client.go
```diff
@@ -0,0 +1,691 @@
...
+func NewClient(main, fallback *rpc.Client, chainID uint64) *ClientWithFallback {
+	hystrix.ConfigureCommand(fmt.Sprintf("ethClient_%d", chainID), hystrix.CommandConfig{
+		Timeout:               10000,
+		MaxConcurrentRequests: 100,
+		SleepWindow:           300000,
+		ErrorPercentThreshold: 25,
+	})
+
+	var fallbackEthClient *ethclient.Client
+	if fallback != nil {
```
stefandunca, line 64: I'm wondering why allow a `nil` fallback, don't we have `NewSimpleClient` for that case?

rpc/chain/client.go
```diff
@@ -0,0 +1,691 @@
...
+		fallback:    fallbackEthClient,
+		mainRpc:     main,
+		fallbackRpc: fallback,
+		IsConnected: true,
+	}
+}
+
+func (c *ClientWithFallback) Close() {
+	c.main.Close()
+	c.fallback.Close()
```
stefandunca: This is allowed to be `nil` with `NewClient` constructor. I also see a double close in `NewSimpleClient` use case.

stefandunca (approved):
Nice implementation; looking forward to seeing it in action. I left some questions inline.

rpc/chain/client.go
alaibe: that s a problem, I should set it to nil, indeed, good catch

rpc/chain/client.go
alaibe: good catch, will set it to nil

rpc/chain/client.go
alaibe: Once again, good catch :) i will check for nil

PR #3206, opened 2023-02-20
feat: Add `ActivityCenterNotificationsCountBy` endpoint

MishkaRogachev:
For https://github.com/status-im/status-desktop/pull/9530

Add `ActivityCenterNotificationsCountBy` endpoint  for counting ac notifications by different conditions (activity type, read and accepted)


ghost:
### Pull Request Checklist

ilmotta:
@MishkaRogachev, as you mentioned to me in private, this new endpoint would have the potential to replace many others. That's great!

But I still think the endpoint won't solve an important problem, which is that clients need to know the count of unread notifications for each tab (a logical group of 1+ types), as well as the total. Right now, as you know, the client needs to make N requests to `status-go`, one for each tab.

What would be ideal IMO is to evolve this endpoint to return a struct containing the counts for each type of notification sent in the request and maybe even the sum, so the client gets everything they need in a single round-trip.

But if we want to proceed with this endpoint as it is, perhaps we should name it `ActivityCenterNotificationsTotalCount` or something. I'm not a fan of the suffix `By` at the end because it implies to me a group/map like response. Also `By` doesn't say *by what*, so it's sort of meh? :man_shrugging:

MishkaRogachev:
> @MishkaRogachev, as you mentioned to me in private, this new endpoint would have the potential to replace many others. That's great!
> 
> But I still think the endpoint won't solve an important problem, which is that clients need to know the count of unread notifications for each tab (a logical group of 1+ types), as well as the total. Right now, as you know, the client needs to make N requests to `status-go`, one for each tab.
> 
> What would be ideal IMO is to evolve this endpoint to return a struct containing the counts for each type of notification sent in the request and maybe even the sum, so the client gets everything they need in a single round-trip.
> 
> But if we want to proceed with this endpoint as it is, perhaps we should name it `ActivityCenterNotificationsTotalCount` or something. I'm not a fan of the suffix `By` at the end because it implies to me a group/map like response. Also `By` doesn't say _by what_, so it's sort of meh? 🤷‍♂️

Created separated [issue](https://github.com/status-im/status-go/issues/3232) for that. Planning to stat it after this one is merged.

ilmotta (approved):

protocol/activity_center_persistence.go
```diff
@@ -800,43 +818,21 @@ func (db sqlitePersistence) MarkActivityCenterNotificationsUnread(ids []types.He
...
 }
 
 func (db sqlitePersistence) UnreadAndAcceptedActivityCenterNotificationsCount(activityTypes []ActivityCenterType) (uint64, error) {
 	var count uint64
-	err := db.buildActivityCenterNotificationsCountQuery(true, activityTypes).Scan(&count)
+	err := db.buildActivityCenterNotificationsCountQuery(false, ActivityCenterQueryParamsReadUnread, activityTypes).Scan(&count)
+	return count, err
+}
+
+func (db sqlitePersistence) ActivityCenterNotificationsCountBy(activityTypes []ActivityCenterType, readType ActivityCenterQueryParamsRead, accepted bool) (uint64, error) {
```
ilmotta, line 833: It be better if we got a test for this public method. If in a hurry, maybe do it a follow-up PR?

MishkaRogachev: yep, we going to replace this method durning endpoints refactor

richard-ramos (approved):

PR #3207, opened 2023-02-20
feat: add `CommunityTokenPermission` to community description

0x-r4bbit:
This adds token permission capabilities to `CommunityDescription` such that token permissions can be created, updated and deleted.


protocol/protobuf/communities.proto
```diff
@@ -36,6 +36,37 @@ message CommunityPermissions {
...
+  repeated uint64 token_ids = 5;
+  string ens_pattern = 6;
+}
+
+message CommunityTokenPermission {
+
+  enum Type {
+    UNKNOWN_TOKEN_PERMISSION = 0;
+    BECOME_ADMIN = 1;
+    BECOME_MEMBER = 2;
```
0x-r4bbit, line 55: This will have two more types, but not planned for MVP

protocol/protobuf/communities.proto
```diff
@@ -36,6 +36,37 @@ message CommunityPermissions {
...
+  }
+
+  string id = 1;
+  Type type = 2;
+  repeated TokenCriteria token_criteria = 3;
+  repeated string chat_ids = 4;
```
0x-r4bbit, line 61: `chat_ids` are also ignored for MVP

protocol/communities/community.go
```diff
@@ -1390,6 +1399,80 @@ func (o *Community) Categories() map[string]*protobuf.CommunityCategory {
...
+
+	if o.config.PrivateKey == nil {
+		return nil, ErrNotAdmin
+	}
+
+	if o.config.CommunityDescription.TokenPermissions == nil {
+		o.config.CommunityDescription.TokenPermissions = make(map[string]*protobuf.CommunityTokenPermission)
+	}
+
+	if _, ok := o.config.CommunityDescription.TokenPermissions[permission.Id]; ok {
```
jrainville: nitpick, but maybe `exists` would be a better name than `ok`

protocol/communities/community.go
```diff
@@ -1390,6 +1399,80 @@ func (o *Community) Categories() map[string]*protobuf.CommunityCategory {
 	return response
 }
 
+func (o *Community) AddTokenPermission(permission *protobuf.CommunityTokenPermission) (*CommunityChanges, error) {
+	o.mutex.Lock()
+	defer o.mutex.Unlock()
+
+	if o.config.PrivateKey == nil {
```
jrainville, line 1406: Is this the best way to check if you're admin? I wonder if when we have owners and admins, does only the owner have the privatekey? Can the admin add permissions anyway? If no, then disregard this comment.

protocol/requests/create_community_token_permission_request.go
```diff
@@ -0,0 +1,56 @@
...
+	"errors"
+	"strconv"
+
+	"github.com/status-im/status-go/eth-node/types"
+	"github.com/status-im/status-go/protocol/protobuf"
+)
+
+var (
+	ErrCreateCommunityTokenPermissionInvalidCommunityID    = errors.New("create community token permission needs a valid community id")
+	ErrCreateCommunityTokenPermissionTooManyTokenCriteria  = errors.New("too many token criertia")
```
jrainville:
```suggestion
	ErrCreateCommunityTokenPermissionTooManyTokenCriteria  = errors.New("too many token criteria")
```

protocol/requests/create_community_token_permission_request.go
```diff
@@ -0,0 +1,56 @@
...
+	Type          protobuf.CommunityTokenPermission_Type `json:"type"`
+	TokenCriteria []*protobuf.TokenCriteria              `json:"tokenCriteria"`
+}
+
+func (p *CreateCommunityTokenPermission) Validate() error {
+	if len(p.CommunityID) == 0 {
+		return ErrCreateCommunityTokenPermissionInvalidCommunityID
+	}
+
+	if len(p.TokenCriteria) > 5 {
```
jrainville: Should we move this to a constant somewhere so we can modify it only in one place in case we need this type of check somewhere else?

jrainville (commented):
Great job! LGTM, just 4 small comments. Only the question about admins is "important"

protocol/communities/community.go
0x-r4bbit: This is a good point. As per design, it looks like admins can create a subset of permissions as well.
But I see a problem here. 

Let's assume you don't need to own the private key here to make changes to the community description, but merely have the admin role. We can build it so that if you don't have the role, we'll exit here, but nothing can prevent you from actually changing the source, so that you **can** change the community description.

It then boils down to whether or not the updated community description was signed by the community (its private key), otherwise everyone could just make changes to a community.

So I guess, we then have to check what happens if:

1. We remove this check and we try to publish the community description without the private key being available
2. If there's anything in place that allows for 1), then we should check whether any community related messages are either signed by the community **or** by someone who has admin roles.

protocol/requests/create_community_token_permission_request.go
0x-r4bbit: Yes, I'll make it a constant

protocol/communities/community.go
0x-r4bbit: Okay, so I did some digging and it turns out, basically any changes made that haven't been signed by the community itself are being ignored:

https://github.com/status-im/status-go/blob/d0cc036d486092262382d45967db11f19d66641e/protocol/communities/community.go#L958-L960

So I believe we need to extend the protocol to make this work for admins.
Essentially something like: 

- If a community description message is being sent and the private key of that community is not available, check if the current account is a) a member and b) has admin roles. If a) and b) is the case, then the message will be signed by the admin's account
- If we receive a community description and the message wasn't signed by the community itself, check if it was signed by an admin member

However, as I'm writing this out I realized, this is probably not a good path forward either.
If we implement this, then this will enable "admins" to do anything they want really, because access to changing community description is an all or nothing game. So they themselves could make others admins etc.

Just thinking out loud here, but this smells like our protocol won't easily support this.

@cammellos summoning you for you additional valuable thoughts on this.

0x-r4bbit: Discussed this with @cammellos offline and right now there's no solution for this in place and needs further exploratoin/research.

So I'll continue assuming that only owners can create permissions for now until we have a solution for this.

0x-r4bbit:
@jrainville this is ready for re-review

jrainville (approved):
Nice work

PR #3211, opened 2023-02-21
feat(keycard): keycard details are being synced among devices

saledjenic:
Keycard details are being synced among devices:
- sync all keycards state
- sync every keycard change

Entire work is splitted across few independent commits.

How it fits the desktop app can be seen in the video in this pr https://github.com/status-im/status-desktop/pull/9600


protocol/protobuf/pairing.proto
```diff
@@ -298,3 +298,34 @@ message SyncRawMessage {
...
+    LOCKED = 4;
+    UNLOCKED = 5;
+    UID_UPDATED = 6;
+    NAME_CHANGED = 7;
+  }
+}
+
+message SyncAllKeycards {
+  repeated SyncKeycard keycards = 1;
+  uint64 syncing_clock = 2;
```
richard-ramos: This is me being nitpicky, so comment could be ignored, but could the `*_clock` attributes be named `clock`, similar to how it's done in other .proto files?

multiaccounts/keypairs/database.go
```diff
@@ -14,6 +20,48 @@ type KeyPair struct {
...
+	for _, addr := range kc.Addresses {
+		kp.AccountsAddresses = append(kp.AccountsAddresses, types.BytesToAddress(addr))
+	}
+}
+
+func removeElementAtIndex[T any](s []T, index int) []T {
+	if index < 0 || index >= len(s) {
+		panic("keycard: index out of the range")
+	}
+	return append(s[:index], s[index+1:]...)
```
richard-ramos, line 62: nice!

multiaccounts/keypairs/database.go
```diff
@@ -2,10 +2,16 @@ package keypairs
...
 	"fmt"
 	"strings"
 
 	"github.com/status-im/status-go/eth-node/types"
+	"github.com/status-im/status-go/protocol/protobuf"
+)
+
+const (
+	dbTransactionIsNil = "keycard: database transaction is nil"
 )
```
richard-ramos:
```suggestion
var dbTransactionIsNil = errors.New("keycard: database transaction is nil")
```

richard-ramos (approved):
Cool PR :crown:

saledjenic:
@richard-ramos thanks for the review. I addressed all your comments.

alaibe (approved):

rasom (approved):

protocol/messenger.go
```diff
@@ -2348,6 +2348,12 @@ func (m *Messenger) SyncDevices(ctx context.Context, ensName, photoPath string,
 			return err
 		}
 	}
+
+	err = m.syncAllKeycards(ctx, rawMessageHandler)
+	if err != nil {
```
qfrank, line 2353: Hi @saledjenic , can you elebrate why only syncAllKeycards which corresponding protobuf.SyncAllKeycards within SyncDevices? we don't take care of protobuf.SyncKeycardAction within SyncDevices?

saledjenic: @qfrank yes, sure... this is how UI looks in the desktop app:

`SyncKeycard` message is used to send details about each keycard change to enabled paired devices so they can update the their state for changed keycard.

But if two devices are using the same seed phrase and they are not paired, or pairing was disabled at the moment when some change were made on one of devices, user is able to sync the enabled paired devices clicking on `Sync All Devices` button (image below), in that case we're sending `SyncAllKeycards` message which contains the list of `SyncKeycard` messages that the paired devices can update their states based on that list. Cause in that case the device on which `Sync All Devices` button was clicked becomes source of truth.

qq

qfrank: sorry, my question maybe not clear, i mean why we don't process syncing/dispatch protobuf.SyncKeycardAction within SyncDevices so that another device will receive message  which's type is protobuf.SyncKeycardAction? I knew syncAllKeycards will dispatch message protobuf.SyncKeycard and i have no question about it :) @saledjenic

saledjenic: @qfrank we do that on each keycard action.

Please check the `dispatchKeycardActivity` function here https://github.com/status-im/status-go/blob/66373d4e0201796222026ce263948c0069be3807/protocol/messenger_keycard.go#L87-L113

Check this [video](https://drive.google.com/file/d/1FnOUlld0iSZPrDz-eGqyrbzfnlXguPrr/view?usp=share_link) you will see that each action is dispatched and changes are received and applied on other device.

protocol/messenger_sync_raw_messages.go
```diff
@@ -188,6 +188,28 @@ func (m *Messenger) HandleSyncRawMessages(rawMessages []*protobuf.RawMessage) er
...
+			err := proto.Unmarshal(rawMessage.GetPayload(), &message)
+			if err != nil {
+				return err
+			}
+			err = m.handleSyncKeycards(state, message)
+			if err != nil {
+				m.logger.Error("failed to handleSyncKeycards when HandleSyncRawMessages", zap.Error(err))
+				continue
+			}
+		case protobuf.ApplicationMetadataMessage_SYNC_KEYCARD_ACTION:
```
qfrank: if this type of message(ApplicationMetadataMessage_SYNC_KEYCARD_ACTION) doesn't come from SyncDevices, then i think we don't need process it here, since method HandleSyncRawMessages only invoked during local pairing @saledjenic

qfrank (changes requested):

protocol/messenger_sync_raw_messages.go
saledjenic: @qfrank you're perfectly right, thanks for the heads-up, updated that.

qfrank (approved):

PR #3215, opened 2023-02-21
Use omitempty for displayName

### LINKED ISSUE #3205: Investigate why displayName field is empty in settings
For some reason `displayName` field in settings object we receive on mobile is an empty string as opposed to `nil` value like e.g. `preferredName`, so for now a hack is required https://github.com/status-im/status-mobile/pull/15122.

vitvly:
This will help to better address https://github.com/status-im/status-mobile/issues/15120 that required an ugly fix https://github.com/status-im/status-mobile/pull/15122

Suggested by @cammellos https://github.com/status-im/status-go/issues/3205#issuecomment-1434925403

Fixes #3205


flexsurfer:
display-name is not used anymore

PR #3219, opened 2023-02-21
chore: add println and logs to check what's being stored in the enr, and preemptively delete the multiaddr field

richard-ramos:
Since this https://github.com/status-im/status-desktop/issues/9589 is not easy to reproduce, I'm adding here some extra logs and also printing information in the terminal (that should be removed once the cause of the bug is found and fixed). Hopefully it will help capture the bug.


stefandunca (approved):
Thanks for the swift change. Hope it helps to nail down this issue.

PR #3220, opened 2023-02-22
use 0.0.0.0 address for RPC port

jakubgs:
Simpler than making flags. Which seem to be ignored. As far as I can tell the `config` created in `main` is not used at all.

This is just a temporary hack-fix, and this should be refactored in the future.

Also, I'm not sure why that file is full of spaces instead of tabs


ghost:
### Pull Request Checklist

cammellos (approved):

PR #3222, opened 2023-02-22
Add primary & secondary names

cammellos:
adds primary secondary name, and a bunch of other fields that are used in mobile frontend.

### DIFF
--- a/protocol/contact.go
+++ b/protocol/contact.go
@@ -40,18 +40,6 @@ type ContactDeviceInfo struct {
 	FCMToken string `json:"fcmToken"`
 }
 
-func (c *Contact) CanonicalName() string {
-	if c.LocalNickname != "" {
-		return c.LocalNickname
-	}
-
-	if c.ENSVerified {
-		return c.EnsName
-	}
-
-	return c.Alias
-}
-
 func (c *Contact) CanonicalImage(profilePicturesVisibility settings.ProfilePicturesVisibilityType) string {
 	if profilePicturesVisibility == settings.ProfilePicturesVisibilityNone || (profilePicturesVisibility == settings.ProfilePicturesVisibilityContactsOnly && !c.added()) {
 		return c.Identicon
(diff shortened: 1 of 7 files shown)


protocol/contact.go
```diff
@@ -191,10 +179,45 @@ func (c *Contact) mutual() bool {
...
+func (c *Contact) PrimaryName() string {
+	return c.names()[0]
+}
+
+func (c *Contact) SecondaryName() string {
+	names := c.names()
+	if len(names) > 1 {
+		return names[1]
+	}
+	return ""
```
flexsurfer, line 222: can we return nil ?

flexsurfer (approved):

protocol/contact.go
```diff
@@ -191,10 +179,45 @@ func (c *Contact) mutual() bool {
...
 }
 
+func (c *Contact) names() []string {
+	var names []string
+
+	if c.LocalNickname != "" {
+		names = append(names, c.LocalNickname)
+	}
+
+	if c.ENSVerified && len(c.EnsName) != 0 {
```
flexsurfer, line 197: hey @cammellos sorry i didn't review properly, we should have secondary name only if we have nickname, if we don't have it, secondary name should be empty, so there is no need in names array, we should set only two fields, 
```clojure
(defn assoc-names
  [{:keys [nickname display-name ens-name ens-verified alias public-key] :as contact}]
  (let [username (cond ens-verified
                       ens-name
                       (not (string/blank? display-name))
                       display-name
                       :else
                       (or alias (gfycat/generate-gfy public-key)))]
    (if (not (string/blank? nickname))
      (assoc contact :first-name nickname :second-name username)
      (assoc contact :first-name username))))
```

flexsurfer (changes requested):

protocol/contact.go
cammellos: oh I see, I'll change that then

MishkaRogachev (approved):
LGTM!

flexsurfer (approved):

alwx (approved):

PR #3223, opened 2023-02-22
chore(wallet): hardcode specific average block duration for L2

### LINKED ISSUE #9582: Benchmark request count on the usual L1 and L2 and see how much they deviate from the ideal/expected based on the hardcoded 12s block length
Noticed on L2 a lot of requests. At a quick investigation, I found that L2 doesn't follow the block length of the main chain but has different strategies to generate block numbers.

@alaibe also reported not finishing fetching for Arbitrum/Optimistic

- [x] A plan: have a look at how much they deviate from the hardcoded block count and if hardcoding a different stride for these specific blockchains might work fine
- [ ] B plan: reimplement block length estimation (similar to the initial implementation while keeping the logic simple)

stefandunca:
### Closes [#9582](https://github.com/status-im/status-desktop/issues/9582)

The L2 networks have their block number and average block duration, so we need to hardcode to keep the logic simplified.

### DIFF
--- a/services/wallet/history/balance.go
+++ b/services/wallet/history/balance.go
@@ -16,18 +16,24 @@ type Balance struct {
 	db *BalanceDB
 }
 
-type blocksStride = int
-
 const (
-	blockTime = time.Duration(12) * time.Second
-	aDay      = time.Duration(24) * time.Hour
+	defaultChains = uint64(0)
+	aDay          = time.Duration(24) * time.Hour
 )
+
+var averageBlockDurationForChain = map[uint64]time.Duration{
+	defaultChains: time.Duration(12000) * time.Millisecond,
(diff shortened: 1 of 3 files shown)


alaibe (approved):

alaibe:
have u notice improvement?

stefandunca:
> have u notice improvement?

Yes, I posted results in the reported issue https://github.com/status-im/status-desktop/issues/9582#issuecomment-1441457060

1683 -> 57
1434 -> 48
96 -> 5

But also identify issues:
- Short time span generate only few entries which don't look very nice on the graph: https://github.com/status-im/status-desktop/issues/9624
- The block count per second varies very much; the average block length moved from all time `0.864` to `0.316` in recent weeks

It looks to me that bringing back the estimation of the block length, even if not as precise, is the way to go and future-proof.

PR #3226, opened 2023-02-23
use random tcp port for wakuv2 instead of using fixed value 60000

qfrank:
If client pass a port value(not 0) for wakuv2 config, it will use it, otherwise it will use a random port instead. using random port can avoid port conflict issue, e.g. when you want run 2 ios simulators on local machine at the same time.

[relate PR](https://github.com/status-im/status-go/pull/3210)

### DIFF
--- a/node/geth_node.go
+++ b/node/geth_node.go
@@ -3,7 +3,6 @@ package node
 import (
 	"errors"
 	"fmt"
-	"net"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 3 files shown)


richard-ramos:
go-waku IIRC does support setting a 0 and it will attempt to use a random open TCP port. Could be interesting to try to remove the logic that sets the default port to 60000, and just use 0 directly

richard-ramos:
Yup, look at this:
https://github.com/waku-org/go-waku/blob/master/waku/v2/node/wakunode2.go#L150-L155

*I think* that if you set 0 here,
```go
var DefaultConfig = Config{
	MaxMessageSize:    common.DefaultMaxMessageSize,
	Host:              "0.0.0.0",
	Port:              0, <-----------------------------------------------
	KeepAliveInterval: 10, // second
	DiscoveryLimit:    40,
	MinPeersForRelay:  1, // TODO: determine correct value with Vac team
	AutoUpdate:        false,
}
```

and remove 
```go
	if cfg.Port == 0 {
		cfg.Port = DefaultConfig.Port
	}
```

it should find a port automagically

qfrank:
> go-waku IIRC does support setting a 0 and it will attempt to use a random open TCP port. Could be interesting to try to remove the logic that sets the default port to 60000, and just use 0 directly



> go-waku IIRC does support setting a 0 and it will attempt to use a random open TCP port. Could be interesting to try to remove the logic that sets the default port to 60000, and just use 0 directly

you remind me that there was also no need to change geth_node.go 👍 @richard-ramos

richard-ramos (approved):

PR #3233, opened 2023-02-24
feat: add coingecko fallback

alaibe:
* Create an interface for our price provider
* Add coingecko client
Next: add the fallback strategy


services/wallet/service.go
```diff
@@ -175,19 +174,19 @@ func (s *Service) CheckConnected(ctx context.Context) *ConnectedResult {
...
+		openseaClients[chainID] = Connection{
 			Up:            client.IsConnected,
 			LastCheckedAt: client.LastCheckedAt,
 		}
 	}
 	return &ConnectedResult{
 		Infura:  infura,
-		Opensea: opensea,
+		Opensea: openseaClients,
 		CryptoCompare: Connection{
```
dlipicar: Do we want to know the state per provider or per "category"? If it's the second, we might want to re-name the fields to something like "Blockchain", "Collectibles" and "Market" (and some indication of which provider we're actually using?)

services/wallet/thirdparty/types.go
```diff
@@ -0,0 +1,37 @@
...
+	AssetLaunchDate      string  `json:"AssetLaunchDate"`
+	AssetWhitepaperURL   string  `json:"AssetWhitepaperUrl"`
+	AssetWebsiteURL      string  `json:"AssetWebsiteUrl"`
+	BuiltOn              string  `json:"BuiltOn"`
+	SmartContractAddress string  `json:"SmartContractAddress"`
+}
+
+type MarketDataProvider interface {
+	FetchPrices(symbols []string, currencies []string) (map[string]map[string]float64, error)
+	HistoricalDailyPrices(symbol string, currency string, limit int, allData bool, aggregate int) ([]HistoricalPrice, error)
```
dlipicar: I like having `Fetch` at the beginning of method names that will involve an RPC call, to be consistent with the others

services/wallet/thirdparty/utils/symbols.go
```diff
@@ -0,0 +1,37 @@
...
+func GetRealSymbol(symbol string) string {
+	if val, ok := renameMapping[strings.ToUpper(symbol)]; ok {
+		return val
+	}
+	return strings.ToUpper(symbol)
+}
+
+func ChunkSymbols(symbols []string) [][]string {
+	var chunks [][]string
+	chunkSize := 20
```
dlipicar, line 25: Does this chunk size come from a provider limitation or to limit how much data we download in one go? Anyway, it's something we probably want to define somewhere else and pass to this method as a parameter

services/wallet/thirdparty/coingecko/client.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	var tokens []GeckoToken
+	err = json.Unmarshal(body, &tokens)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, token := range tokens {
+		if id, ok := coinGeckoMapping[strings.ToUpper(token.Symbol)]; ok {
```
stefandunca, line 123: I see we add tokens that are not in the `coinGeckoMapping`. I wonder why we do this filtering.

stefandunca (approved):
Excellent abstraction

services/wallet/thirdparty/coingecko/client.go
alaibe: This one is a bit annoying, the Symbol are not unique in coingekco, e.g SNT has 2 possible ids, this is why I do some force mapping. Also some token in our store are not in coingecko which is another issue

services/wallet/thirdparty/coingecko/client.go
```diff
@@ -0,0 +1,298 @@
...
+		return c.tokens, nil
+	}
+
+	url := fmt.Sprintf("%scoins/list", baseURL)
+	resp, err := c.DoQuery(url)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
```
richard-ramos, line 110: Should we handle error codes here? https://support.coingecko.com/hc/en-us/articles/6472757474457-How-can-I-differentiate-between-the-status-codes-I-am-receiving-and-what-do-they-mean-

services/wallet/thirdparty/coingecko/client.go
```diff
@@ -0,0 +1,298 @@
...
+	if err != nil {
+		return nil, err
+	}
+	url := fmt.Sprintf("%ssimple/price?ids=%s&vs_currencies=%s", baseURL, strings.Join(ids, ","), strings.Join(currencies, ","))
+	resp, err := c.DoQuery(url)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
```
richard-ramos, line 168: See prev comment

services/wallet/thirdparty/coingecko/client.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+	return result, nil
+}
+
+func (c *Client) FetchTokenMarketValues(symbols []string, currency string) (map[string]thirdparty.TokenMarketValues, error) {
+	ids, err := c.mapSymbolsToIds(symbols)
+	if err != nil {
+		return nil, err
+	}
+	url := fmt.Sprintf("%scoins/markets?ids=%s&vs_currency=%s&order=market_cap_desc&per_page=250&page=1&sparkline=false&price_change_percentage=1h,24h", baseURL, strings.Join(ids, ","), currency)
```
richard-ramos: This is just me being annoying so feel free to ignore the comment, but you can make url building more elegant with [url.Values](https://pkg.go.dev/net/url#Values):
```go
queryParams := url.Values{
    "id":  {string.Join(ids, ",")},
    "vs_currency": {currency},
    "order": {"market_cap_desc"}
     etc etc etc
}
url := "baseURL" + "coins/markets" + queryParams.Encode() 
```

richard-ramos (approved):

services/wallet/thirdparty/coingecko/client.go
alaibe: updated

services/wallet/thirdparty/coingecko/client.go
alaibe: I didn't thought about handling error as it will be part of the circuit. If error => go to next.

If we manage error we will have to unify errors across all the "provider". It would be nice to do but I don't yet know what to do with the errors :thinking:

PR #3235, opened 2023-02-24
Compute emojiHash on-the-fly

### LINKED ISSUE #3162: add emoji-hash to `multiaccount` struct
# Problem
Emoji Hash is required for each profile, so its better if its also a part of the `multiaccount` struct.

##Note : 
We should keep the existing EmojiHash API as it is because Desktop Team is using it.

vitvly:
Due to https://github.com/golang/go/issues/49030 it's not possible to embed a generic type parameter. If it were possible, then `GetPkeyData` could be a generic function returning marshaled JSON, this could look better and could save some LOC.

Another issue is that Settings serialisation uses kebab-case. Not sure if it's ok to let client know about it (`compressed-key`->`compressedKey`).

Fixes #3162


account/json/utils.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/status-im/status-go/api/multiformat"
+	"github.com/status-im/status-go/protocol/identity/emojihash"
+)
+
+type PublicKeyData struct {
+	CompressedKey string   `json:"compressedKey"`
+	EmojiHash     []string `json:"emojiHash"`
+}
+
+func GetPkeyData(publicKey string) (*PublicKeyData, error) {
```
ilmotta: @siphiuel, `Pkey` is a *word* used to mean either a public or private key (at least in some circles), as we can see from this example https://docs.rs/openssl/latest/openssl/pkey/index.html.

But this function is returning only public key data. Wouldn't it be better to give a more strict name to avoid confusion? E.g. rename the function to `GetPublicKeyData` or shorter `GetPubKeyData`?

And while doing so, we could reduce the confusion if the variables are also renamed, for example:

```go
// From this
pkeyData, err := accountJson.GetPkeyData(s.PublicKey)

// To this
pubKeyData, err := accountJson.GetPubKeyData(s.PublicKey)

File: multiaccounts/settings/columns_test.go [hunk 2]
```diff
@@ -52,7 +52,8 @@ func TestJSONEncoding(t *testing.T) {
 	encoded, err := json.Marshal(settings)
 	require.NoError(t, err)
 
-	require.True(t, strings.Contains(string(encoded), "\"compressed-key\":\"zQ3shudJrBctPznsRLvbsCtvZFTdi3b34uzYDuqE9Wq9m9T1C\""))
+	require.True(t, strings.Contains(string(encoded), "\"compressedKey\":\"zQ3shudJrBctPznsRLvbsCtvZFTdi3b34uzYDuqE9Wq9m9T1C\""))
```
ilmotta, line 55: >Another issue is that Settings serialisation uses kebab-case. Not sure if it's ok to let client know about it (compressed-key->compressedKey).

As you said yourself, renaming can be breaking change, but honestly I don't know where this is being used on the mobile/desktop clients. I know that the work to revamp settings on mobile hasn't started yet, so it might be okay to rename. Not sure about desktop though.

But given renames in payloads are potentially breaking changes for consumers, I guess we should bump the minor version in `VERSION`.

protocol/contact.go
```diff
@@ -401,34 +401,27 @@ func (c *Contact) processSyncContactRequestState(remoteState ContactRequestState
...
 		ContactRequestState ContactRequestState `json:"contactRequestState"`
 		HasAddedUs          bool                `json:"hasAddedUs"`
 		Mutual              bool                `json:"mutual"`
-
-		Active        bool   `json:"active"`
-		PrimaryName   string `json:"primaryName"`
-		SecondaryName string `json:"secondaryName,omitempty"`
-	}{
-		Alias: (*Alias)(c),
+		Active              bool                `json:"active"`
```
ilmotta, line 410: Since you're removing the `Active` field, it might be a good idea to double-check if it's not breaking the clients. In `status-mobile` I found this https://github.com/status-im/status-mobile/blob/develop/src/status_im2/contexts/contacts/events.cljs#L22

If it's a breaking change then it might be a good idea to create an issue in `status-mobile` and/or `status-desktop` so somebody (or you) can quickly work on the adaptations, otherwise most likely someone will upgrade `status-go` in the client and introduce a regression. I wonder if we're using semversion appropriately on the clients... :speak_no_evil:

vitvly:
Thanks a lot for great points. Will rename.

Regarding “Active” field - i did not remove it, just removed superfluous assignment.

ilmotta:
> Regarding “Active” field - i did not remove it, just removed superfluous assignment.

Oh that's correct, missed that, my bad @siphiuel

multiaccounts/settings/columns_test.go
Samyoul: If a setting has a column in the settings database table it will have a kebab-case `reactFieldName` associated with it, that's the convention for `Settings`. This is a strange case because `compressed-key` isn't a field of the `Settings` struct or a column of the settings db table.

However we do want to purge all the kebab-case from `status-go` at some point.

See this discussion https://github.com/status-im/status-go/pull/3117#discussion_r1083276062

protocol/contact.go
```diff
@@ -401,34 +401,27 @@ func (c *Contact) processSyncContactRequestState(remoteState ContactRequestState
...
 	}
-	item.CompressedKey = compressedKey
 
 	item.Added = c.added()
 	item.HasAddedUs = c.hasAddedUs()
 	item.Mutual = c.mutual()
 	item.Active = c.active()
 	item.PrimaryName = c.PrimaryName()
 	item.SecondaryName = c.SecondaryName()
-	item.Active = c.active()
```
Samyoul, line 431: The `Active` field seems to be dropped here. I don't see where the field is set from the target `Contact`.

multiaccounts/settings/columns_test.go
ilmotta: Thanks for sharing. Didn't know about that translation mechanism. So eventually we want to get rid of `reactFieldName`, all kebab cases and just stick to camelCase?

Samyoul:
@siphiuel I think I may have a solution for you:

So you were discussing on Monday that you would like to have a generic embed so that you can extend any struct with the same code. In the call I didn't properly understand the relationship between the structs you wanted to embed, but after looking at your code I can see that the `reflect` package will give you the generic goodness you want.

Add the following function to your `account/utils/json.go` file

```go

func ExtendStructWithPubKeyData(publicKey string, item any) (any, error) {
	pkd, err := GetPubKeyData(publicKey)
	if err != nil {
		return nil, err
	}

	typ := reflect.StructOf([]reflect.StructField{
		{
			Name:      "Item",
			Anonymous: true,
			Type:      reflect.TypeOf(item),
		},
		{
			Name:      "Pkd",
			Anonymous: true,
			Type:      reflect.TypeOf(pkd),
		},
	})

	v := reflect.New(typ).Elem()
	v.Field(0).Set(reflect.ValueOf(item))
	v.Field(1).Set(reflect.ValueOf(pkd))
	s := v.Addr().Interface()

	return s, nil
}
```

Then run this test in a new file `account/utils/json_test.go`

```go
func Test(t *testing.T) {
	key := "0x04deaafa03e3a646e54a36ec3f6968c1d3686847d88420f00c0ab6ee517ee1893398fca28aacd2af74f2654738c21d10bad3d88dc64201ebe0de5cf1e313970d3d"

	type Goat struct {
		Beard bool
		Horns bool
		Smell uint
	}

	testItem := struct {
		Name string
		Age  uint
		Goat Goat
	}{
		Name: "Test Name",
		Age:  100,
		Goat: Goat{
			Beard: true,
			Horns: true,
			Smell: 9001,
		},
	}

	ext, err := ExtendStructWithPubKeyData(key, testItem)
	spew.Dump(err)

	json, err := json.MarshalIndent(ext, "", "\t")
	spew.Dump(err)
	fmt.Printf("%s\n", json)
}
```

This gives an output of the following (at least on my machine):

```text
=== RUN   Test
(interface {}) <nil>
(interface {}) <nil>
{
	"Name": "Test Name",
	"Age": 100,
	"Goat": {
		"Beard": true,
		"Horns": true,
		"Smell": 9001
	},
	"compressedKey": "zQ3shudJrBctPznsRLvbsCtvZFTdi3b34uzYDuqE9Wq9m9T1C",
	"emojiHash": [
		"🔩",
		"🎊",
		"👨‍👩‍👧‍👧",
		"👋🏻",
		"🙇🏾",
		"➡️",
		"🧖🏼",
		"🧙‍♂️",
		"👹",
		"👍🏽",
		"👩🏾‍💼",
		"✌🏾",
		"🧑‍🤝‍🧑",
		"🛖"
	]
}
--- PASS: Test (0.00s)
PASS
```

This function can be used like below:

```go
func (a AccountInfo) MarshalJSON() ([]byte, error) {
	ext, err := accountJson.ExtendStructWithPubKeyData(a.PrivateKey, a)
	if err != nil {
		return nil, err
	}

	return json.Marshal(ext)
}
```

```go
func (s Settings) MarshalJSON() ([]byte, error) {
	// We need this typedef in order to overcome stack overflow
	// when marshaling JSON
	type Alias Settings
	ext, err := accountJson.ExtendStructWithPubKeyData(s.PublicKey, Alias(s))
	if err != nil {
		return nil, err
	}
	return json.Marshal(ext)
}
```

```go
	if item.From != "" {
		ext, err := accountJson.ExtendStructWithPubKeyData(item.From, item)
		if err != nil {
			return nil, err
		}

		return json.Marshal(ext)
	}
```

```go

	ext, err := accountJson.ExtendStructWithPubKeyData(item.ID, item)
	if err != nil {
		return nil, err
	}

	return json.Marshal(ext)
}
```

Samyoul:
@siphiuel Just did a quick test with pointers and embedding a scala type

```go
func Test(t *testing.T) {
	key := "0x04deaafa03e3a646e54a36ec3f6968c1d3686847d88420f00c0ab6ee517ee1893398fca28aacd2af74f2654738c21d10bad3d88dc64201ebe0de5cf1e313970d3d"

	type Goat struct {
		Beard bool
		Horns bool
		Smell uint
	}

	tests := []any{
		struct {
			Name string
			Age  uint
			Goat Goat
		}{
			Name: "Test Name",
			Age:  100,
			Goat: Goat{
				Beard: true,
				Horns: true,
				Smell: 9001,
			},
		},
		&struct {
			Name string
			Age  uint
			Goat *Goat
		}{
			Name: "Test Name",
			Age:  100,
			Goat: &Goat{
				Beard: true,
				Horns: true,
				Smell: 9001,
			},
		},
		&struct {
			Uncle  string
			Height uint
			Goats  []Goat
		}{
			Uncle:  "An Uncle's Kind of Name",
			Height: 1337,
			Goats: []Goat{
				{
					Beard: true,
					Horns: true,
					Smell: 9001,
				},
				{
					Beard: false,
					Horns: false,
					Smell: 9002,
				},
				{
					Beard: true,
					Horns: false,
					Smell: 3,
				},
			},
		},
		"hello",
	}

	for _, ti := range tests {
		ext, err := ExtendStructWithPubKeyData(key, ti)
		require.NoError(t, err)

		jb, err := json.MarshalIndent(ext, "", "\t")
		require.NoError(t, err)
		fmt.Printf("%s\n", jb)
	}
}
```

Output

```text
=== RUN   Test
{
	"Name": "Test Name",
	"Age": 100,
	"Goat": {
		"Beard": true,
		"Horns": true,
		"Smell": 9001
	},
	"compressedKey": "zQ3shudJrBctPznsRLvbsCtvZFTdi3b34uzYDuqE9Wq9m9T1C",
	"emojiHash": [
		"🔩",
		"🎊",
		"👨‍👩‍👧‍👧",
		"👋🏻",
		"🙇🏾",
		"➡️",
		"🧖🏼",
		"🧙‍♂️",
		"👹",
		"👍🏽",
		"👩🏾‍💼",
		"✌🏾",
		"🧑‍🤝‍🧑",
		"🛖"
	]
}
{
	"Name": "Test Name",
	"Age": 100,
	"Goat": {
		"Beard": true,
		"Horns": true,
		"Smell": 9001
	},
	"compressedKey": "zQ3shudJrBctPznsRLvbsCtvZFTdi3b34uzYDuqE9Wq9m9T1C",
	"emojiHash": [
		"🔩",
		"🎊",
		"👨‍👩‍👧‍👧",
		"👋🏻",
		"🙇🏾",
		"➡️",
		"🧖🏼",
		"🧙‍♂️",
		"👹",
		"👍🏽",
		"👩🏾‍💼",
		"✌🏾",
		"🧑‍🤝‍🧑",
		"🛖"
	]
}
{
	"Uncle": "An Uncle's Kind of Name",
	"Height": 1337,
	"Goats": [
		{
			"Beard": true,
			"Horns": true,
			"Smell": 9001
		},
		{
			"Beard": false,
			"Horns": false,
			"Smell": 9002
		},
		{
			"Beard": true,
			"Horns": false,
			"Smell": 3
		}
	],
	"compressedKey": "zQ3shudJrBctPznsRLvbsCtvZFTdi3b34uzYDuqE9Wq9m9T1C",
	"emojiHash": [
		"🔩",
		"🎊",
		"👨‍👩‍👧‍👧",
		"👋🏻",
		"🙇🏾",
		"➡️",
		"🧖🏼",
		"🧙‍♂️",
		"👹",
		"👍🏽",
		"👩🏾‍💼",
		"✌🏾",
		"🧑‍🤝‍🧑",
		"🛖"
	]
}
{
	"Item": "hello",
	"compressedKey": "zQ3shudJrBctPznsRLvbsCtvZFTdi3b34uzYDuqE9Wq9m9T1C",
	"emojiHash": [
		"🔩",
		"🎊",
		"👨‍👩‍👧‍👧",
		"👋🏻",
		"🙇🏾",
		"➡️",
		"🧖🏼",
		"🧙‍♂️",
		"👹",
		"👍🏽",
		"👩🏾‍💼",
		"✌🏾",
		"🧑‍🤝‍🧑",
		"🛖"
	]
}
--- PASS: Test (0.00s)
PASS
```

protocol/contact.go
Samyoul: Ah, `Active` is set twice 🤦

multiaccounts/settings/columns_test.go
Samyoul: Yes, ideally. Convert everything to camelCase.

vitvly:
@Samyoul @ilmotta thanks a lot for your suggestions and review, can you please have a look at the PR again?

ilmotta (approved):
Looks good @siphiuel :)

From my personal perspective, I sort of liked the original implementation you did without reflection. Although using reflection is possible and an interesting approach, in the philosophy of Go, I'm inclined for the more explicit, bare bones solution, even though it requires more LOC. I'm not 100% positive, but I think reflection also has more runtime overhead and we lose some compile time guarantees.

But I digress, this is just a biased opinion.

vitvly:
@ilmotta thanks. I'm sorry, didn't provide more context in previous comment. We discusses the reflection approach with @Samyoul , and compared pros and cons vs the original version. Seems that runtime overhead is negligible, given that `json.Marshal()` uses reflection anyway, and implementation complexity is not that bad if only present at lower level + commented appropriately. So that's how i got convinced, even though originally i also preferred the original approach.

ilmotta:
> @ilmotta thanks. I'm sorry, didn't provide more context in previous comment. We discusses the reflection approach with @Samyoul , and compared pros and cons vs the original version. Seems that runtime overhead is negligible, given that `json.Marshal()` uses reflection anyway, and implementation complexity is not that bad if only present at lower level + commented appropriately. So that's how i got convinced, even though originally i also preferred the original approach.

I guess you just convinced me too :wink: Good arguments, thanks for sharing.

Samyoul (approved):
Looks good.

I think that either approach would be fine. This approach could be made more useful by generalising the function more.

What you're solving here is producing a json object with all immediate struct fields promoted to the parent level.

So not for this PR, but potentially in the future you could expand this to a function that takes `n` structs and embeds them all into a holding struct for the sole purpose of being JSON Marshalled. Although, we may want to question why we'd want to create such a struct for marshalling, as although the runtime generation is dynamic the consuming code would still need some level of static fields in order to correctly interpret the data.

PR #3236, opened 2023-02-27
feat: backup keycards to and fetch them from waku

saledjenic:
Changes applied here introduce backing up keycards data to and fetch them from waku.
Information about received keycards data is sent via `sync.from.waku.keycards` signal.


alaibe (approved):

protocol/messenger_backup_test.go
```diff
@@ -634,3 +634,48 @@ func (s *MessengerBackupSuite) TestBackupCommunities() {
...
+
+	// Backup
+	_, err = bob1.BackupData(context.Background())
+	s.Require().NoError(err)
+
+	// Wait for the message to reach its destination
+	_, err = WaitOnMessengerResponse(
+		bob2,
+		func(r *MessengerResponse) bool {
+			_, err := bob2.RetrieveAll()
```
qfrank: feel strange here since it already called `RetrieveAll` within `WaitOnMessengerResponse` 🤔 , in general, we return the condition that indicate we received the relate data we want

protocol/messenger_backup_test.go
```diff
@@ -634,3 +634,48 @@ func (s *MessengerBackupSuite) TestBackupCommunities() {
...
+				return false
+			}
+
+			return true
+		},
+		"keycards data not backed up",
+	)
+	s.Require().NoError(err)
+
+	s.Require().NoError(err)
```
qfrank: duplicate line

saledjenic: removed

protocol/messenger_backup_test.go
saledjenic: I agree, but this is very steril test environment, don't expect issues, also this is how we're doing the same in other tests here.

qfrank: if you need "Wait for the message to reach its destination", some code like following should be better or you can check other things relate with variable r
```
WaitOnMessengerResponse(
		bob2,
		func(r *MessengerResponse) bool { return len(r.Messages()) > 0 },
		"no messages",
	)
```
@saledjenic

saledjenic: @qfrank it doesn't work in that case, test fails.

saledjenic: But it works without this line `_, err := bob2.RetrieveAll()`, this works:
```golang
_, err = WaitOnMessengerResponse(
		bob2,
		func(r *MessengerResponse) bool { return true },
		"no messages",
	)
```

saledjenic: @qfrank I found a way to make the tests more reliable, please check it now.

qfrank (approved):

richard-ramos (approved):

PR #3238, opened 2023-02-28
Feat/transaction history for collectibles

dlipicar:
Part of [#8811](https://github.com/status-im/status-desktop/issues/8811)

This is composed of 2 commits.
In the first one, erc20 transfers are differentiated from erc721 ones, and the tokenID is retrieved for the second ones.
In the second one, we allow fetching OpenSea assets by Contract Address + tokenID. Additionally, http error handling and a retry mechanism (with wait between retries) is introduced for OpenSea, plus some minor fixes.

### DIFF
--- a/services/wallet/thirdparty/opensea/client.go
+++ b/services/wallet/thirdparty/opensea/client.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"strconv"
 	"strings"
 	"sync"
@@ -13,10 +14,15 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
+
+	"github.com/status-im/status-go/services/wallet/bigint"
 )
 
 const AssetLimit = 50
 const CollectionLimit = 300
+
+const RequestRetryMaxCount = 1
+const RequestWaitTime = 300 * time.Millisecond
 
 var OpenseaClientInstances = make(map[uint64]*Client)
 
@@ -27,6 +33,11 @@ var BaseURLs = map[uint64]string{
 }
 
 type TraitValue string
+
+type NFTUniqueID struct {
+	ContractAddress common.Address `json:"contractAddress"`
+	TokenID         bigint.BigInt  `json:"tokenID"`
+}
 
 func (st *TraitValue) UnmarshalJSON(b []byte) error {
 	var item interface{}
@@ -47,11 +58,9 @@ func (st *TraitValue) UnmarshalJSON(b []byte) error {
 }
 
 type AssetContainer struct {
-	Assets []Asset `json:"assets"`
-}
-
-type AssetCollection struct {
-	Name string `json:"name"`
+	Assets         []Asset `json:"assets"`
+	NextCursor     string  `json:"next"`
+	PreviousCursor string  `json:"previous"`
 }
 
 type Contract struct {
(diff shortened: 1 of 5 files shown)


services/wallet/transfer/view.go
```diff
@@ -109,6 +123,32 @@ func parseLog(ethlog *types.Log) (from, to common.Address, amount *big.Int) {
...
 	}
-	amount = new(big.Int).SetBytes(ethlog.Data)
+	amount.SetBytes(ethlog.Data)
+
+	return
+}
+
+func parseErc721Log(ethlog *types.Log) (from, to common.Address, tokenId *big.Int) {
+	tokenId = new(big.Int)
+	if len(ethlog.Topics) < 4 {
```
saledjenic, line 151: @dlipicar 4 or 3?

services/wallet/thirdparty/opensea/client.go
```diff
@@ -167,28 +189,53 @@ func (o *Client) FetchAllCollectionsByOwner(owner common.Address) ([]Collection,
...
-	o.LastCheckedAt = time.Now().Unix()
+
+	queryParams["limit"] = []string{strconv.Itoa(AssetLimit)}
+	var currentCursor string
 	for {
-		url := fmt.Sprintf("%s/assets?owner=%s&collection=%s&offset=%d&limit=%d", o.url, owner, collectionSlug, offset, AssetLimit)
+		if len(currentCursor) > 0 {
+			queryParams["cursor"] = []string{currentCursor}
+		} else {
+			queryParams["cursor"] = []string{`""`}
```
saledjenic: @dlipicar is this what you really want? To set an array with a single element which is an empty string to `queryParams["cursor"]`? I thought maybe it's better to proceed without else block here, but if we need to have this (because of final url), then instead of `if` condition a single line `queryParams["cursor"] = []string{currentCursor}` is enough.

saledjenic (approved):

stefandunca (approved):

services/wallet/transfer/view.go
dlipicar: It's 4 elements for erc721 transfers (signature, from, to, tokenId)

services/wallet/thirdparty/opensea/client.go
dlipicar: If I set it to an empty string, the request will look something like:
`... &cursor=&limit=50& ...` which OpenSea doesn't like. So my options would be to remove the parameter from the `queryParams ` map, or do what I did (which OpenSea accepts). The first option seems cleaner now that I think about it... I'll make the change.

dlipicar: Done, parameter is only added after first query, if next cursor is empty we simply exit the loop since there are no more collectibles to fetch

services/wallet/thirdparty/opensea/client.go
```diff
@@ -131,33 +143,45 @@ func NewOpenseaClient(chainID uint64, apiKey string) (*Client, error) {
 	client := &http.Client{
 		Timeout: time.Second * 5,
 	}
+	rateLimiter := rate.NewLimiter(rate.Every(1*time.Second), 1) // 1 request per second
```
alaibe: that is quite high, is that what the doc says??

services/wallet/transfer/controller.go
```diff
@@ -192,6 +192,30 @@ func (c *Controller) LoadTransferByHash(ctx context.Context, rpcClient *rpc.Clie
...
+		}
+	}
+	return tx.Type
+}
+
+func transfersToTransferViews(transfers []Transfer) []View {
+	for idx, tx := range transfers {
+		transfers[idx].Type = getFixedTransferType(tx)
+	}
+	return castToTransferViews(transfers)
```
alaibe: That function is quite weird, we have a case function, this is where we should put the logic to of the type

services/wallet/transfer/downloader.go
```diff
@@ -203,7 +208,7 @@ func (d *ETHDownloader) getTransfersInBlock(ctx context.Context, blk *types.Bloc
 
 // NewERC20TransfersDownloader returns new instance.
 func NewERC20TransfersDownloader(client *chain.ClientWithFallback, accounts []common.Address, signer types.Signer) *ERC20TransfersDownloader {
-	signature := crypto.Keccak256Hash([]byte(erc20TransferEventSignature))
+	signature := crypto.Keccak256Hash([]byte(erc20_721TransferEventSignature))
 	return &ERC20TransfersDownloader{
```
alaibe, line 217: Should it also be named: ERC20_721TransfersDownloader

services/wallet/thirdparty/opensea/client.go
dlipicar: There's nothing about it on the docs, and it seems to depend on which endpoint you use (some are limited, some are not). Searched a little bit, someone from OpenSea posted a message stating it's supposed to be limited to 4 requests per second, reduced the wait time to 300ms to be safe. Also reworked this a little bit to be smarter about error handling. We'll now try to make the request right away, and retry once if we get HTTP error `StatusTooManyRequests`. Tested it, it seems to work fine. Does it look better to you?

services/wallet/transfer/controller.go
dlipicar: Do you mean the `case` statement in `CastToTransferView` from `view.go`? So, my reasoning was:

-) Ideally this would be fixed at the `downloader.go` level (properly set the type to the `Transfer` object, and fix the ones we've got stored in the DB).
-) Next best option is to go one level up, which is `controller.go` (take the transfer list from wherever they come and fixes the type)
-) `view.go` seems to be "too close to the top layer" for being in charge of this.

alaibe: I mean a cast function *

alaibe (approved):

PR #3240, opened 2023-02-28
feat(CommunityTokenMetadata): Keep token metadata in CommunityDescrip…

### LINKED ISSUE #9545: Keep smart community token metadata (icon, description) in waku
Add setting icon in UI and Nim logic.
Create .proto file for metadata propagation. It will be send with some global topic.
Smart contract's uri will contain the topic.
.proto will have to be resend because it will be destaryed after 2 weeks.

Epic https://github.com/status-im/status-desktop/issues/8694

endulab:
Keep token metadata in CommunityDescription.

Community tokens has some metadata (image, description) which must be kept in waku(CommunityDescription),
because they are not part of a smart contract.

Add community_token_metadata.proto.
Add []CommunityTokenMetadata to CommunityDescription.

Issue #9545


ghost:
### Pull Request Checklist

protocol/communities/manager.go
```diff
@@ -2571,6 +2571,32 @@ func (m *Manager) GetCommunityTokens(communityID string, chainID int) ([]*Commun
...
+		return err
+	}
+	if community == nil {
+		return ErrOrgNotFound
+	}
+
+	tokenMetadata := &protobuf.CommunityTokenMetadata{
+		ContractAddresses: map[uint64]string{uint64(token.ChainID): token.Address},
+		Description:       token.Description,
+		Image:             token.Base64Image,
```
0x-r4bbit, line 2716: For now at least we will not support minting ERC20 token, but probably in the future (?) /cc @John-44 
Maybe it makes sense to add a `type` here so we can check if it's an ERC721 or an ERC20
I know that we can check it onchain as well by asking the contract, but that would save one roundrip to the chain to query that data.

Same goes for `symbol` as well.

protocol/protobuf/community_token_metadata.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+
+option go_package = "./;protobuf";
+package protobuf;
+
+message CommunityTokenMetadata {
+  map<uint64, string> contract_addresses = 1;
+  string description = 2;
+  string image = 3;
+}
```
0x-r4bbit: Maybe we can just put it in `communities.proto` ?

0x-r4bbit (approved):
Generally it looks good to me.
Left some comments, but approving

protocol/communities/manager.go
jrainville: I think it makes sense to add the type if we have a different view/workflow depending on the type. The symbol is probably important too, more than the description even

jrainville (commented):
Looks good, just adding to Pascal's comment

protocol/communities/manager.go
John-44: @0x-r4bbit @jrainville we won't support minting ERC20 tokens in 0.11 but we may start work on adding support for this in either 0.12 or 0.13 e.g. there is a good likelihood we will end up adding support for minting ERC20 tokens before MVP.

endulab: @0x-r4bbit @jrainville I added symbol and type.

protocol/protobuf/community_token_metadata.proto
endulab: I decided to keep it separately because it is bigger now and just in case we want to use it as a separated messages one day. Though, I am not sure these reasons are convincing :)

protocol/protobuf/community_token_metadata.proto
```diff
@@ -0,0 +1,17 @@
+syntax = "proto3";
+
+option go_package = "./;protobuf";
+package protobuf;
+
+message CommunityTokenMetadata {
+  enum TokenType {
+    ERC721 = 0;
+    ERC20 = 1;
+  }
```
0x-r4bbit: If we move this `message` definition to `communities.proto` then we can probably reuse a single enum for different things.

For example, here you can see a similar enum I've introduced: https://github.com/status-im/status-go/pull/3207/files#diff-b8c0db7add756912e199c0c950bd8a41a7b6c71ca47687022bf6e3ed5865be8fR40-R45

richard-ramos (approved):

0x-r4bbit:
@endulab this doesn't pass the linter atm

PR #3245, opened 2023-03-01
chore(accounts): `type` column from `accounts` table updated

saledjenic:
`type` column is set for all rows to appropriate value. Before this change accounts which were generated from the keypair created importing seed phrase had `generated` value for the `type`.

According to above, a function for generating an account sets the `type` based on the passed derive from address.


appdatabase/migrations/sql/1677681143_accounts_table_type_column_update.up.sql
```diff
@@ -0,0 +1 @@
+UPDATE accounts SET type = "seed" WHERE derived_from IN (SELECT derived_from FROM accounts WHERE type = "seed");
```
qfrank: this sql seems very very dangerous.. I ran it on my local machine, guess what happened? all rows affected.. and all of its type updated to "seed", before this, there were different types: seed/key/generated/""(empty), do you think it's normal?  🤔

saledjenic: @qfrank thanks for the review.

Exactly that was my intention, to update `type` field for all accounts generated from the same seed phrase. Without that line we cannot be backward compatible.

> this sql seems very very dangerous.. 

I am wondering what's that dangerous, the only possible values for `type` column are `generated`, `seed`, `key`, `watch` and empty string (but the empty string is special case, refers only to chat address and default wallet address). 

Having that in mind if you read the query carefully you will see that only the `type` column of accounts which have `seed` as a value for the `type` column will be updated to `seed` value, what actually was my intention.

saledjenic: > there were different types: seed/key/generated/""(empty)

I am not sure how it's possible that all your accounts are affected when we're checking if `derived_from` is in the set returned by this query `SELECT derived_from FROM accounts WHERE type = "seed"`

That makes me think that your `derived_from` column was empty for all your accounts before you run migration script, what further means that you're using a very very old account. Is that the case maybe?

qfrank: > all your accounts are affected 

that's what i mean dagnerous.
subquery SELECT derived_from FROM accounts WHERE type = "seed" will return set("","0x..") which contains empty string, that's why all my accounts were affected. we should take care of empty string, right? my accounts are not very very old ... 
image

saledjenic: Seems either that your Status account is very old or mobile app somehow adds account without set `derived_from`.

> we should take care of empty string, right?

You must not have at all any account for the `type` value set to `seed` or `generated` with an empty `derived_from`. Cause that is an unallowed, an incorrect, `accounts` table state. But seems that your `accounts` table had exactly that (before you run the migration script), an empty string set for `derived_from` column.


The only accounts from the `accounts` table which may have an empty `derived_from` value are:
- Status chat account, that's an account with `m/43'/60'/1581'/0'/0` derivation path
- accounts added via importing private key, those with value `key` for `type` column

alaibe: @saledjenic derived_from might not be used today by the mobile app as mobile don't need to derive new account yet

saledjenic: @alaibe even they don't use it, they have to set it when user is creating new account via mobile app and then use the same account in the desktop app, cause otherwise keypairs on the desktop app side will be broken. Cause all accounts with the same `derived_from` value belong to the single keypair.

saledjenic: @alaibe @qfrank but thinking now, I am not sure how we can set `derived_from` field for already created accounts?

We need to have it set for:
- Status default wallet account (path `m/44'/60'/0'/0/0`), profile keypair
- all other derivations from the profile keypair
- for account created by importing seed phrase, imported keypair
- for all derivations from the imported keypair

qfrank: > @alaibe even they don't use it, they have to set it when user is creating new account via mobile app and then use the same account in the desktop app, cause otherwise keypairs on the desktop app side will be broken. Cause all accounts with the same `derived_from` value belong to the single keypair.

my quick reply: even mobile don't use it, we should consider earlier user for status mobile, right? 🙂

saledjenic: > even mobile don't use it, we should consider earlier user for status mobile, right? 

I am not sure when exactly in the past `derived_from` was introduce, but definitely from that point onwards adding accounts via mobile and desktop app divergates. All in all at the moment we need to find a way how to overcome that and as I see handling most likely should be on the mobile app side.
  
This is definitely an obstacle we have to discuss with @John-44

saledjenic: @qfrank we (wallet team) had a discussion with John and he just said that you can choose one of the following two solutions to overcome this:
1. make imported accounts with empty `derived_from` field locked and ask users to provide a seed phrase for such accounts if they want to use them
2. make imported accounts with empty `derived_from` field locked and ask users to re-add such accounts (remove and add them again)

Also I guess the following one should be good as well (haven't discussed with John):

3. once user is logged in if he has imported accounts with empty `derived_from` you can run a wizard for each of accounts asking user to provide seed phrase or remove account if he doesn't have a seed phrase.

But till then would it work for you (without any drawbacks) if I would update the query with `derived_from != ""` to look like this `SELECT derived_from FROM accounts WHERE type = "seed" AND derived_from != ""`, cause I would like to have this merged, need this for other things I have to work on?

qfrank: add `derived_from != ""` seems safe to mobile db, but i want hear something from @cammellos 🙂

cammellos: @saledjenic @alaibe this needs to be addressed by whoever created the issue and introduced the breaking change, we can't have breaking changes not handled and let someone else having to deal with the consequences, this should be agreed before hand we make a breaking change.

The solution proposed are adding considerable development work and should have been discussed before making a breaking change. cc @John-44 

Is there any alternative solution? What's the difference between `derived_from` and `account_type`?

alaibe: @cammellos Hey, I wouldn't say it is a breaking change. As nothing is broken from both apps. It is only that in mobile wallet the derived from is not being used yet.

The derived from is to allow us to derive more than one account from a seed phrase. Which is not yet a feature implemented in mobile AFAIK.

Alternative solution would be create edge cases and limit the possible feature on such account. However that would create additional maintenance work for desktop and mobile and this would be an edge case we have to deal with

saledjenic: > What's the difference between derived_from and account_type?

@cammellos `account_type` is there to differ accounts, is an account added by:
- deriving from a profile keypair (master key a default Status wallet account is derived from), in this case `account_type` is set to `generated`
- importing seed phrase or deriving from master key added by importing seed phrase, in this case `account_type` is set to `seed`
- importing private key, in this case `account_type` is set to `key`
- adding an account address, watch only account is added, in this case `account_type` is set to `watch` 

`derived_from` is there that we can know which accounts are derived from the same master key.

cammellos: @saledjenic @alaibe thanks for the explanation.

I checked the code on mobile, and it seems we do set the type to `seed` when creating from a seed phrase, so that should not really apply to us ( I haven't really run the code, just checked it), but maybe I am missing something, would you mind explaining why this change is necessary and when we created accounts with the wrong type?

John-44: This will be needed by the new Mobile wallet designs as well, we should aim for the new mobile wallet to be functionally identical with the desktop wallet

saledjenic: @cammellos trying to explain below.

Just to be on the same page, since mobile app hasn't introduced keypairs yet, let's define a "keypair" as a term.
A single keypair is defined by:
- **name**, which needs to be unique across the app (not added yet on the desktop app side, will be in the next step) 
- **accounts**, one or more, but all of them have to be derived from the same master key

<ins>For example</ins>:
When user imports a seed phrase into the wallet, a new account `Account1` is generated from that seed phrase and its type is set to `seed`. This added `Account1` belongs to a new keypair named `MyKeypair1`. 

> it seems we do set the type to seed when creating from a seed phrase

What you have in mobile app is adding a single account from the seed phrase, setting the `type` to `seed` for that account, that's ok, we have that feature as well. But now what we have in the desktop app, and I guess you don't have in mobile app, is an ability to let user creates another account for a keypair (the same master key) he already added earlier.

<ins>Continuation to the example above</ins>:
After adding `Account2` to the keypair `MyKeypair1` that keypair will have 2 accounts `Account1` and `Account2` and both of those accounts will have the same `derived_from` address and both of them will have the same `type` set to `seed`.

That's how it should be, but in the desktop app, `type` in this case was set to `generated` (that's fixed by this PR).

> would you mind explaining why this change is necessary and when we created accounts with the wrong type?

This change is needed cause it fixes an issue on the desktop app, dating from who knows when, by updating `type` for all accounts generated from a keypair which is created by importing the seed phrase.

qfrank: > but in the desktop app, type in this case was set to generated (that's fixed by this PR).

Hi @saledjenic , could you explain why in this case it should not set to generated? according to your example, i think type for Account1 should be `seed`, type for Account2 should be `generated` if the procedure of generating is the same as original type `generated`(i'm not sure how you generate Account2 exactly, original `generated` account invoked [function](https://github.com/status-im/status-go/blob/7c7c3a1f1335ea166b7839217d9ec37b74ad0191/account/generator/generator.go#L171)), otherwise maybe a new type introduced would be better?

saledjenic: Let's see on an example (hope it will be better):
| address | type | name | derived_from|
| -- | -- | -- | -- |
| 0x11 |   | Status account | 0x1111 |
| 0x12 |  generated  | Derived1 | 0x1111 |
| 0x13 |  generated  | Derived2 | 0x1111 |
| 0x21 |  seed  | SImported1 | 0x2222 |
| 0x22 |  generated  | SImpoGen11 | 0x2222 |
| 0x23 |  generated  | SImpoGen12 | 0x2222 |
| 0x41 |  key  | KImpo1 |  |

In that table we can identify 3 keypairs:
- Status account (`0x11`, `0x12`, `0x13`) <--- we know this is the default (profile keypair)
- SImported1 (`0x21`, `0x22`, `0x23`) <--- we know this is a keypair imported via seed phrase
- KImported1 (`0x31`) <--- we know this is a keypair imported via private key

So the table above represents the current state, without changes done in this PR. 
Addresses `0x21`, `0x22` and `0x23` should be the same, since after removing for example address `0x21` user is able to generate it again until at least one of accounts connected to the same keypair exists, so that's one reason why `seed` should be set for the `type` for account generated from the imported seed phrase (all accounts derived from the same master key are functionally the same), another is that the desktop app is using it in several places to know which icon to display (cause icon for default, seed phrase imported and private key imported keypair is different).

But one bigger issue is that none of accounts have `derived_from` field set, that further means if user creates accounts using mobile app and then sync them with the desktop app all keypairs will be broken, cause we differ them by `derived_from` value. And that's the main reason why, as @qfrank said [here](https://github.com/status-im/status-go/pull/3245#discussion_r1122536029), type for all accounts were set to `seed`.

My proposal was (which will work for you without any drawbacks):
- to update the query with `derived_from != ""` to look like this `SELECT derived_from FROM accounts WHERE type = "seed" AND derived_from != ""`
- ignore accounts with empty `derived_from` from syncing

With these 2 changes we're good in the desktop app and you won't see any drawbacks in mobile app. But once you need to introduce keypairs you will have to ask users to provide seed phrase or re-add already added accounts cause there is no other way to set `derived_from` properly for already added accounts.

saledjenic: Are you ok with the changes I proposed in my previous comment, so I can do them here and overcome this in that way?

saledjenic: I am currently working on updating add account flow in the desktop app (according to the latest UI changes for that) and within those changes I will add new column `keypair_name` column to `accounts` table (that's a new request).

qfrank: what about introducing new type? let's say it's `keypair-generated`, i feel the way you create account 0x22/0x23 is not the same as the old way create `generated`, but you didn't explain it in your reply

address | type | name | derived_from
-- | -- | -- | --
0x11 |   | Status account | 0x1111
0x12 | generated | Derived1 | empty
0x13 | generated | Derived2 | empty
0x21 | seed | SImported1 | 0x2222
0x22 | keypair-generated | SImpoGen11 | 0x2222
0x23 | keypair-generated | SImpoGen12 | 0x2222

qfrank: > Are you ok with the changes I proposed in my previous comment, so I can do them here and overcome this in that way?

It's okay if @cammellos says it's okay, i think my okay doesn't count in this situation 😅

cammellos: > @cammellos trying to explain below.
> 
> Just to be on the same page, since mobile app hasn't introduced keypairs yet, let's define a "keypair" as a term. A single keypair is defined by:
> 
> * **name**, which needs to be unique across the app (not added yet on the desktop app side, will be in the next step)
> * **accounts**, one or more, but all of them have to be derived from the same master key
> 
> For example: When user imports a seed phrase into the wallet, a new account `Account1` is generated from that seed phrase and its type is set to `seed`. This added `Account1` belongs to a new keypair named `MyKeypair1`.
> 
> > it seems we do set the type to seed when creating from a seed phrase
> 
> What you have in mobile app is adding a single account from the seed phrase, setting the `type` to `seed` for that account, that's ok, we have that feature as well. But now what we have in the desktop app, and I guess you don't have in mobile app, is an ability to let user creates another account for a keypair (the same master key) he already added earlier.
> 
> Continuation to the example above: After adding `Account2` to the keypair `MyKeypair1` that keypair will have 2 accounts `Account1` and `Account2` and both of those accounts will have the same `derived_from` address and both of them will have the same `type` set to `seed`.
> 
> That's how it should be, but in the desktop app, `type` in this case was set to `generated` (that's fixed by this PR).
> 
> > would you mind explaining why this change is necessary and when we created accounts with the wrong type?
> 
> This change is needed cause it fixes an issue on the desktop app, dating from who knows when, by updating `type` for all accounts generated from a keypair which is created by importing the seed phrase.

Thanks for the explanation
with regard to that, no migration I believe is needed on mobile, since we have not allowed users to derive a wallet from an imported seed yet, so there won't be any account that has the bug you mentioned (which is only present on desktop, is that correct?)
If so, the migration after the changes asked by @qfrank should work fine (i.e as long as we are migrating account that have a `derived_from` set, it should not impact existing mobile users, and their database is consistent since they don't have any of these new type of accounts as we never allowed them to generate any)

Is that a correct representation of the state of things or I am missing something?

cammellos: > It's okay if @cammellos says it's okay, i think my okay doesn't count in this situation

You ok counts :), I am not going to make a strong call on this issue, this is a discussion between peers so my opinion is worth just as much as anyone else's in this case, so please do continue the discussion so we can reach an agreement on what you think it's the best way forward.

 I see a point in having a separate type for this accounts, but it's also true that you can distinguish them by looking at two fields as @saledjenic is suggesting, I am ambivalent on this one, and I think it's more contextual on how we query them/use them, etc, but either seems to be a fine approach by me.

saledjenic: > what about introducing new type? let's say it's keypair-generated

I don't see need for new type. Why would we introduce a new type if accounts `0x21`, `0x22` and `0x23` are functionally wise the same?

>  i feel the way you create account 0x22/0x23 is not the same as the old way create generated

Not sure what old way you're referring to? Before my changes done here we were using [this function](https://github.com/status-im/status-go/blob/1b24342a75d99b8e2601d7639295899ea75b6fb3/services/accounts/accounts.go#L290) for adding account generated from already imported seed phrase and it sets the `type` column to `seed`. In this PR we're still using the same call, but I change the `generateAccountPasswordVerified` that it sets `type` correctly now.

saledjenic: @cammellos yes, that's correct, after changes I am going to add no migration will be needed for mobile users.

Will post here and @qfrank may give another look and test if mobile app users are somehow affected by the changes.

qfrank: note: mobile won't take care of field `derived_from`(so it will be empty) when import account from seed phrase

qfrank: > Before my changes done here we were using [this function](https://github.com/status-im/status-go/blob/1b24342a75d99b8e2601d7639295899ea75b6fb3/services/accounts/accounts.go#L290) for adding account generated from ...

make sense, so old way i refer is the same as the link you mentioned, so new type is not needed.. 

> Addresses 0x21, 0x22 and 0x23 should be the same, since after removing for example address 0x21 user is able to generate it again until at least one of accounts connected to the same keypair exists, so that's one reason why seed should be set for the type for account generated from the imported seed phrase

i got some questions for above statement. 
1. Addresses 0x21, 0x22 and 0x23 should be the same? they are not the same in local mobile test
image
2. we need set type to 'seed' for 0x22/0x23 because mobile may create `generated` account which have empty value for derived_from seems not a good solution to me, it looks like trying to give `generated` account different meaning @saledjenic

qfrank: > My proposal was (which will work for you without any drawbacks):

> 1. to update the query with derived_from != "" to look like this SELECT derived_from FROM accounts WHERE type = "seed" AND derived_from != ""
> 2. ignore accounts with empty derived_from from syncing

3. what if we only apply the second proposal? i mean we don't need the migration appdatabase/migrations/sql/1677681143_accounts_table_type_column_update.up.sql

alaibe: 1 - Addresses 0x21, 0x22 and 0x23 should be the same? they are not the same in local mobile test
How did u create addresses for this case in mobile, I thought u didn't support yet generating more than one account from a seed phrase?

2- If i understand correctly we are trying to limit the scope of the value "generated" to only being "generated" as they were before no?

saledjenic: @qfrank I've just committed the the update, please check if this PR anyhow affects the mobile app users?

It contains `derived_from != ""` for the query and added a comment to the `HandleSyncWalletAccount` function where we need to do some handling later (will do that in another PR when the time comes, once I finished my current work and get back to syncing).

qfrank: > 1 - Addresses 0x21, 0x22 and 0x23 should be the same? they are not the same in local mobile test How did u create addresses for this case in mobile, I thought u didn't support yet generating more than one account from a seed phrase?
> 
> 2- If i understand correctly we are trying to limit the scope of the value "generated" to only being "generated" as they were before no?

Thank you for your reply, make sense @alaibe , before your reply i got confused with default status account/account imported from seed phrase, my bad sorry.

protocol/messenger_handler.go
```diff
@@ -2633,6 +2633,23 @@ func (m *Messenger) HandleSyncWalletAccount(state *ReceivedMessageState, message
 
 	var accs []*accounts.Account
 	for _, message := range message.Accounts {
+
+		// We need to ignore accounts in the desktop app with empty `DerivedFrom`, otherwise that will break keypairs.
+		nodeConfig, err := m.settings.GetNodeConfig()
+		if err != nil {
+			return err
+		}
+		// The only way we can deffer if `status-go` is used by desktop app or not is `Name` field from node config (at least
```
qfrank: incorrect word `deffer`? maybe it's differ?

alaibe (approved):
Looks good to me. Let's merge once @qfrank also give a go from mobile

appdatabase/migrations/sql/1677681143_accounts_table_type_column_update.up.sql
saledjenic: > what if we only apply the second proposal? i mean we don't need the migration appdatabase/migrations/sql/1677681143_accounts_table_type_column_update.up.sql

- accounts `0x21`, `0x22` and `0x23` are functionally wise the same, no point in having some of them with `type` set to `seed` and some of them to `generated`. An account added along with importing a seed phrase or new accounts generated later from already imported seed phrase should have `type` set to `seed`, it's more intuitive and cleaner that way. 
- the desktop app already uses `type` in several places, meaning that without updating `type` to `seed` for accounts generated from imported seed phrase we will have some keypairs broken if the account created in the moment of importing a seed phrase gets deleted (since only it has `seed` for the `type`).

protocol/messenger_handler.go
saledjenic: @qfrank  yes, it's a typo, should be "differ", but it's a comment for me for later which will be removed later. Will update it to avoid confusion. 😉

qfrank (approved):

qfrank:
LGTM, thank you for detailed explaination 🙂

PR #3247, opened 2023-03-01
chore: find free port for torrent service if set as 0

richard-ramos:
@0x-r4bbit should I create a migration to set the ports to 0 too?


0x-r4bbit (approved):
Nice @richard-ramos !
I don't think we need a migration, because it'd still be unexpected if this change would affect existing configs IMO

Issue #3241, opened 2023-02-28
Local Pairing - Change `Messenger` sync behaviour based on state


Samyoul:
I notice is that a lot of the syncing methods now have a nilable parameter, that seems to be nil in the majority of cases. Perhaps we could simplify the method logic and update `Messenger`'s state with a "RawMessageHandler" register.

The consuming code would look like something below:

```go
m.SetRawMessageHandler(rawMessageHandler)
m.SyncDevices()
m.SendPairInstallation(context.TODO)
m.ResetRawMessageHandler() // Defaults back to `dispatchPairInstallationMessage()`
```

Implementing this change would simplify the calling of all the functions that have an optional `RawMessageHandler `, future developers don't need to know what a `RawMessageHandler` is unless they want to change "broadcasting" behaviour.

---

Additionally we have precedent for altering `Messenger` behaviour via an update to state with the use of the `localPairing` field, this is used as below:

```go
func (m *Messenger) hasPairedDevices() bool {
	logger := m.logger.Named("hasPairedDevices")

	if m.localPairing {
		return true
	}

	...
}
```

Consider the relationship between these two factors, `localPairing` and `RawMessageHandler`. Is there a link between them that would affect the implementation?

Perhaps if `localPairing` is true, a non-default `RawMessageHandler` should be present?

## Details

https://github.com/status-im/status-go/pull/3202#discussion_r1118755199

PR #3248, opened 2023-03-02
Improved Local Pairing Separation of Concerns

### LINKED ISSUE #3198: Upgrade of `PayloadSourceConfig` into a more robust config schema
## Details:

With the introduction of the below PRs we are including very specific context configurations of the pairing process. There are 4 contexts to consider:

- The device is a `Sender` *(bootstrapping another device)*
- The device is a `Receiver` *(being bootstrapped by another device)*
- The device should run a `PairingServer` *(the device exposing a tls server on the LAN)*
- The device should run a `PairingClient` *(the device connecting to a tls server on the LAN)*

A device maybe a :
- `SenderServer`
- `SenderClient`
- `ReceiverClient`
- `ReceiverServer`

The application needs a much more robust config to ensure the maintainability of the code.

## Possible Solution:

One potential solution to this issue is to create a new type of config and use a component pattern to batch specific expected configs together:

```go
type PairingSenderConfig struct {...}
type PairingReceiverConfig struct {...}
type PairingServerConfig struct {...}
type PairingClientConfig struct{...}

type PairingSenderServerConfig struct {
    PairingSenderConfig
    PairingServerConfig
}

type PairingSenderClientConfig struct {
    PairingSenderConfig
    PairingClientConfig
}

type PairingReceiverClientConfig struct {
    PairingReceiverConfig
    PairingClientConfig
}

type PairingReceiverServerConfig struct {
    PairingReceiverConfig
    PairingServerConfig
}
```

This means that changes specific to each config component only apply to the respective consuming configs, we are free to extend these elements, we keep a strict separation of concerns, we are free to add additional config components.

## More information:

See for more details :
- https://github.com/status-im/status-go/pull/3179/files#r1107080212
- https://github.com/status-im/status-go/pull/3179
- https://github.com/status-im/status-go/pull/3192
- https://github.com/status-im/status-go/pull/3202 (introduces additional config fields to be considered)

Samyoul:
# What's Changed?

Created a stronger separation of concerns between 4 key elements of the local pairing architecture:

**Dataflow:**
- `Sender`
- `Receiver`

**Mode of Transport:**
- `Server`
- `Client`

## Dataflow Interfaces

To facilitate this change the following interfaces were introduced:

`PayloadManager` was replaced with:
- `PayloadMounterReceiver`
  - `PayloadMounter`
  - `PayloadReceiver`

`PayloadRepository` was defined as a distinct interface type
- `PayloadRepository`
  - `PayloadLoader`
  - `PayloadStorer`

`PayloadEncryptor` was introduced to replace `PayloadEncryptionManager`, this change simplifies the use of the encryptor and removes a lot of the type embedding that created ambiguity over the payload state. This change also removes the concept of receive and send data types, there is now only a single payload.

### `PayloadMounter`

Is responsible for mounting data into memory for processing, encrypting and ultimately made ready for transport. The interface consists of 2 base methods and an embedded `PayloadLocker` interface (see below):

```go
type PayloadMounter interface {
	PayloadLocker

	// Mount Loads the payload into the PayloadManager's state
	Mount() error

	// ToSend returns an outbound safe (encrypted) payload
	ToSend() []byte
}
```

A `PayloadMounter` will usually have one or many internal `PayloadLoader`(s) to interface with the datastore and retrieve the data from the store.

### `PayloadReceiver`

Is responsible for accepting data from inbound transport, decrypting, processing and ultimately storing the data. The interface consists of 2 base methods and an embedded `PayloadLocker` interface (see below):

```go
type PayloadReceiver interface {
	PayloadLocker

	// Receive accepts data from an inbound source into the PayloadReceiver's state
	Receive(data []byte) error

	// Received returns a decrypted and parsed payload from an inbound source
	Received() []byte
}
```

A `PayloadReceiver` will usually have one or many internal `PayloadStorer`(s) to interface with the datastore and store the data to the store.

### `PayloadLoader`

Is responsible for loading data from a hardcoded data store. The interface consists of 1 base method:

```go
type PayloadLoader interface {
	LoadFromSource() error
}
```

A `PayloadLoader` will usually be an internal field of a `PayloadMounter`.

### `PayloadStorer`

Is responsible for storing data to a hardcoded data store. The interface consists of 1 base method:

```go
type PayloadStorer interface {
	StoreToSource() error
}
```

A `PayloadStorer` will usually be an internal field of a `PayloadReceiver`.

### `PayloadLocker`

Represents a struct capable of locking its payload from any further access, useful for when data should only be made available once. The interface consists of 1 base method:

```go
type PayloadLocker interface {
	// LockPayload prevents future excess to outbound safe and received data
	LockPayload()
}
```

## Dataflow Implementations

To implement interfaces in the spirit of separating mounting and loading from receiving and storing the following implementation were introduced:

**AccountPayload**
- `AccountPayloadManager`
  - `AccountPayloadMounter`
  - `AccountPayloadReceiver`
- `AccountPayloadRepository`
  - `AccountPayloadLoader`
  - `AccountPayloadStorer`

**RawMessagePayload**
- `RawMessagePayloadManager`
  - `RawMessagePayloadMounter`
  - `RawMessagePayloadReceiver`
- `RawMessagePayloadRepository`
  - `RawMessagePayloadLoader`
  - `RawMessagePayloadStorer`

**InstallationPayload**
- `InstallationPayloadManager`
  - `InstallationPayloadMounter`
  - `InstallationPayloadReceiver`
- `InstallationPayloadRepository`
  - `InstallationPayloadLoader`
  - `InstallationPayloadStorer`

## Mode of Transport

The dataflow separation is cemented via the splitting of `Server` and `Client` types in to the following:

The concepts of `PairingServer` and `PairingClient` are replaced as follows:

- `PairingServer`
  - `BaseServer`
  - `SenderServer`
  - `ReceiverServer`
- `PairingClient`
  - `BaseClient`
  - `SenderClient`
  - `ReceiverClient`

### `Base` types

`Base` types contain all the common logic for the instantiation of all `Client` and `Server` types respectively.

### `Sender` types

`Sender` types instantiate `PayloadMounter`s and facilitate the transport of the mounted data.

### `Receiver` types

`Receiver` types instantiate `PayloadReceiver`s and facilitate the connection and acceptance of data to be stored.

## Base Configuration

I've introduced 4 base configuration types:

- `SenderConfig`
- `ReceiverConfig`
- `ServerConfig`
- `ClientConfig`

Detailed below:

### `SenderConfig`

Go struct:

```go
type SenderConfig struct {
	// SenderConfig.KeystorePath must end with keyUID
	KeystorePath string `json:"keystorePath"`
	// DeviceType SendPairInstallation need this information
	DeviceType string `json:"deviceType"`

	KeyUID   string `json:"keyUID"`
	Password string `json:"password"`

	DB *multiaccounts.Database `json:"-"`
}
```

JSON struct

```json
{
	"keystorePath" : ""
	"deviceType" : ""
	"keyUID" : ""
	"password" : ""
}
```

### `ReceiverConfig`

Go struct:

```go
type ReceiverConfig struct {
	NodeConfig *params.NodeConfig `json:"nodeConfig"`

	// ReceiverConfig.KeystorePath must not end with keyUID (because keyUID is not known yet)
	KeystorePath string `json:"keystorePath"`
	// DeviceType SendPairInstallation need this information
	DeviceType    string `json:"deviceType"`
	KDFIterations int    `json:"kdfIterations"`
	// SettingCurrentNetwork corresponding to field current_network from table settings, so that we can override current network from sender
	SettingCurrentNetwork string `json:"settingCurrentNetwork"`

	DB             *multiaccounts.Database `json:"-"`
	LoggedInKeyUID string                  `json:"-"`
}
```

JSON struct

```json
{
	"keystorePath" : ""
	"deviceType" : ""
	"nodeConfig" : {...}
	"kdfIterations" : 0
	"settingCurrentNetwork" : ""
}
```

### `ServerConfig`

Go struct:

```go
type ServerConfig struct {
	Timeout uint `json:"timeout"`

	// Connection fields, not json (un)marshalled
	// Required for the server, but MUST NOT come from client

	PK       *ecdsa.PublicKey `json:"-"`
	EK       []byte           `json:"-"`
	Cert     *tls.Certificate `json:"-"`
	Hostname string           `json:"-"`
	Mode     Mode             `json:"-"`
}
```

JSON struct

```json
{
	"timeout" : 0
}
```

### `ClientConfig`

Go struct:

```go
type ClientConfig struct{}
```

JSON struct

```json
{}
```

## API Endpoint Configuration

Each of the above 4 base configuration types form part of the 4 types of transport dataflow combinations:

### `SenderServerConfig`

Endpoint on `status-go`:

```go
func GetConnectionStringForBootstrappingAnotherDevice(configJSON string) string
```

Go struct:

```go
type SenderServerConfig struct {
	Sender *SenderConfig
	Server *ServerConfig
}
```

JSON struct:

```json
{
  "sender" : {
    "keystorePath" : ""
    "deviceType" : ""
    "keyUID" : ""
    "password" : ""
  },
  "server" : {
    "timeout" : 0
  }
}
```

### `SenderClientConfig`

Endpoint on `status-go`:

```go
func InputConnectionStringForBootstrappingAnotherDevice(cs, configJSON string) string
```

Go struct:

```go
type SenderClientConfig struct {
	Sender *SenderConfig
	Client *ClientConfig
}
```

JSON struct:

```json
{
  "sender" : {
    "keystorePath" : ""
    "deviceType" : ""
    "keyUID" : ""
    "password" : ""
  },
  "client" : {}
}
```

### `ReceiverClientConfig`

Endpoint on `status-go`:

```go
func InputConnectionStringForBootstrapping(cs, configJSON string) string
```

Go struct:

```go
type ReceiverClientConfig struct {
	Receiver *ReceiverConfig
	Client   *ClientConfig
}
```

JSON struct

```json
{
  "receiver" : {
    "keystorePath" : "",
    "deviceType" : "",
    "nodeConfig" : {...},
    "kdfIterations" : 0,
    "settingCurrentNetwork" : ""
  },
  "client" : {}
}
```

### `ReceiverServerConfig`

Endpoint on `status-go`:

```go
func GetConnectionStringForBeingBootstrapped(configJSON string) string
```

Go struct:

```go
type ReceiverServerConfig struct {
	Receiver *ReceiverConfig
	Server   *ServerConfig
}
```

JSON struct

```json
{
  "receiver" : {
    "keystorePath" : "",
    "deviceType" : "",
    "nodeConfig" : {...},
    "kdfIterations" : 0,
    "settingCurrentNetwork" : ""
  },
  "server" : {
    "timeout" : 0
  }
}
```

### ⚠️ `NodeConfig` ⚠️

A **Note** about `NodeConfig`, this field is identical to the `NodeConfig` used during the application boot.
Also **Note** this PR makes a small but important change to `NodeConfig`, it allows for the JSON (un)marshalling of the `RootDataDir` field. This is a **MANDATORY** field for receiving data.

## Additional Changes

I've added some other little changes to this PR.

- Added `http.Error`s to the places in the server handlers where the code terminates due to errors. This ensures the client will be given a non `200` response.
- Moved a few constants to be together in a declaration group

# Why Make The Change?

To make the code simpler and the config less complex. I've got OCD and somewhat opinionated about code structure.

# To Dos

- [x] [Make an Issue for each `TODO` item](https://github.com/status-im/status-go/pull/3248#discussion_r1138503617) then make a PR that adds links to the code `TODO` comments to the respective issues.
- [x] https://github.com/status-im/status-go/issues/3299
  - **PR:** https://github.com/status-im/status-go/pull/3290
- [x] https://github.com/status-im/status-go/issues/3300
  - **PR:** https://github.com/status-im/status-go/pull/3322
- [x] https://github.com/status-im/status-go/issues/3301
  - **PR:** https://github.com/status-im/status-go/pull/3323
- [x] https://github.com/status-im/status-go/issues/3302
- [x] https://github.com/status-im/status-go/issues/3303
- [ ] https://github.com/status-im/status-go/issues/3304
- [ ] https://github.com/status-im/status-go/issues/3305
- [ ] https://github.com/status-im/status-go/issues/3306
- [ ] https://github.com/status-im/status-go/issues/3241

---

Resolves #3198


server/pairing/connection.go
```diff
@@ -178,7 +180,7 @@ func (cp *ConnectionParams) validateAESKey() error {
 
 func (cp *ConnectionParams) validateServerMode() error {
 	switch cp.serverMode {
-	case Receiving, Sending:
+	case 0, Receiving, Sending:
```
Samyoul, line 184: The refactored code doesn't use `ServerMode` anymore, the only place it is used is in the generation and parsing of `ConnectionString`s. So we want to ignore any 0 value `ServerModes` because it doesn't matter.

I've also added a `TODO` to remove `ServerMode` entirely from the process flow, it is no longer needed.

server/pairing/client.go
```diff
@@ -232,169 +238,236 @@ func (c *Client) receiveInstallationData() error {
...
-	if err != nil {
+
+	if resp.StatusCode != http.StatusOK {
+		err = fmt.Errorf("[client] status not okay when sending installation data, status: %s", resp.Status)
+		signal.SendLocalPairingEvent(Event{Type: EventTransferError, Error: err.Error(), Action: ActionPairingInstallation})
 		return err
 	}
-	return c.PairInstallation()
+
+	signal.SendLocalPairingEvent(Event{Type: EventTransferSuccess, Action: ActionPairingInstallation})
```
ilmotta, line 431: Just a question. I see under the hood `SendLocalPairingEvent` uses `signals.go#send`, which can fail due to marshalling, but the error is not returned (only logged). Is this because of the fallback to the C implementation? What do you think is the reason the `send` function doesn't return an error?

server/pairing/handlers.go
```diff
@@ -29,145 +28,172 @@ const (
...
 			return
 		}
 		signal.SendLocalPairingEvent(Event{Type: EventTransferSuccess, Action: ActionPairingAccount})
 
-		err = ps.PayloadManager.Receive(payload)
+		err = pr.Receive(payload)
 		if err != nil {
 			signal.SendLocalPairingEvent(Event{Type: EventProcessError, Error: err.Error(), Action: ActionPairingAccount})
-			logger.Error("ps.PayloadManager.Receive(payload)", zap.Error(err))
+			logger.Error("pr.Receive(payload)", zap.Error(err))
```
ilmotta: Perhaps in the logs we can be more explicit about naming and avoid acronyms with the intent of eliminating this translation overhead (e.g. pr -> PayloadReceiver). These short variable names are understandable when contextualized, but maybe in the logs, where many things are being logged concurrently, being more explicit is a bonus. So for an `Error` log, instead of `pr.Receive(payload)`, maybe `payloadReceiver.Receive(payload)`.

server/pairing/handlers.go
```diff
@@ -29,145 +28,172 @@ const (
...
+// Challenge middleware and handling
+
+func middlewareChallenge(hs HandlerServer, next http.Handler) http.HandlerFunc {
+	logger := hs.GetLogger()
 	return func(w http.ResponseWriter, r *http.Request) {
-		s, err := ps.cookieStore.Get(r, sessionChallenge)
+		s, err := hs.GetCookieStore().Get(r, sessionChallenge)
 		if err != nil {
-			logger.Error("ps.cookieStore.Get(r, pairingStoreChallenge)", zap.Error(err))
+			logger.Error("hs.GetCookieStore().Get(r, pairingStoreChallenge)", zap.Error(err))
```
ilmotta: Maybe the error message is a bit outdated? Should it be `logger.Error("hs.GetCookieStore().Get(r, sessionChallenge)", zap.Error(err))`?

server/pairing/handlers.go
```diff
@@ -185,9 +211,9 @@ func challengeMiddleware(ps *Server, next http.Handler) http.HandlerFunc {
 			return
 		}
 
-		c, err := common.Decrypt(base58.Decode(pc), ps.ek)
+		c, err := hs.DecryptPlain(base58.Decode(pc))
 		if err != nil {
-			logger.Error("c, err := common.Decrypt(rc, ps.ek)", zap.Error(err))
+			logger.Error("c, err := hs.DecryptPlain(rc, hs.ek)", zap.Error(err))
```
ilmotta: Again me about log messages ;) What do we gain by adding these Go idioms to the log message, like `c, err := ...`? It would be more understandable if it was a `logger.Debug` call because it would be intended mostly for development diagnostics.

server/pairing/payload_receiver.go
```diff
@@ -0,0 +1,383 @@
...
+
+	// Receive accepts data from an inbound source into the PayloadReceiver's state
+	Receive(data []byte) error
+
+	// Received returns a decrypted and parsed payload from an inbound source
+	Received() []byte
+}
+
+type PayloadStorer interface {
+	StoreToSource() error
```
ilmotta: As I read the code I felt the `ToSource` and `FromSource` were a bit redundant, and that `PayloadStorer.Store` and `PayloadLoader.Load` would be just as clear. Just a personal choice though :)

server/pairing/payload_receiver.go
```diff
@@ -0,0 +1,383 @@
...
+	// If lastDir == "keystore" we presume we need to create the rest of the keystore path
+	// else we presume the provided keystore is valid
+	if lastDir == "keystore" {
+		if aps.multiaccount == nil || aps.multiaccount.KeyUID == "" {
+			return fmt.Errorf("no known Key UID")
+		}
+		keyStorePath = filepath.Join(keyStorePath, aps.multiaccount.KeyUID)
+		_, err := os.Stat(keyStorePath)
+		if os.IsNotExist(err) {
+			err := os.MkdirAll(keyStorePath, 0777)
```
ilmotta: Do we really need to give full permission to the `keystore` path or can we lower the flags a little bit?

server/pairing/payload_receiver.go
```diff
@@ -0,0 +1,383 @@
...
+		}
+	}
+
+	for name, data := range aps.keys {
+		accountKey := new(keystore.EncryptedKeyJSONV3)
+		if err := json.Unmarshal(data, &accountKey); err != nil {
+			return fmt.Errorf("failed to read key file: %s", err)
+		}
+
+		if len(accountKey.Address) != 40 {
```
ilmotta: This validation is identical in `payload_mounter.go#loadKeys`, so it may be worth an extraction to a common place.

server/pairing/payload_receiver.go
```diff
@@ -0,0 +1,383 @@
...
+func (aps *AccountPayloadStorer) storeKeys(keyStorePath string) error {
+	if keyStorePath == "" {
+		return fmt.Errorf("keyStorePath can not be empty")
+	}
+
+	_, lastDir := filepath.Split(keyStorePath)
+
+	// If lastDir == "keystore" we presume we need to create the rest of the keystore path
+	// else we presume the provided keystore is valid
+	if lastDir == "keystore" {
```
ilmotta: It'd be nice if we had a constant for this magic string, like something similar to what `sync_device_test.go` does to at least give this important string a name.

server/pairing/payload_receiver.go
```diff
@@ -0,0 +1,383 @@
...
+		} else if err != nil {
+			return err
+		} else {
+			return ErrKeyFileAlreadyExists
+		}
+	}
+
+	for name, data := range aps.keys {
+		accountKey := new(keystore.EncryptedKeyJSONV3)
+		if err := json.Unmarshal(data, &accountKey); err != nil {
```
ilmotta: We could reduce the chance of leaving dangling files due to an error or failure in the middle of the loop by separating the "pure" from the side-effectful part (file system).

So we can first loop, unmarshal and validate data, and only then loop on the keys again and write to disk.

Should we *attempt* to revert file system changes if there's an error while processing one of the keys and others were successful?

server/pairing/payload_mounter.go
```diff
@@ -0,0 +1,339 @@
...
+		if len(accountKey.Address) != 40 {
+			return fmt.Errorf("account key address has invalid length '%s'", accountKey.Address)
+		}
+
+		apl.keys[fileInfo.Name()] = rawKeyFile
+
+		return nil
+	}
+
+	err := filepath.Walk(keyStorePath, fileWalker)
```
ilmotta: Most definitely not a problem in this context, but I felt like sharing, from the Go doc says:

> Walk is less efficient than WalkDir, introduced in Go 1.16, which avoids calling os.Lstat on every visited file or directory. 
>
> -- https://cs.opensource.google/go/go/+/go1.20.2:src/path/filepath/path.go;l=566

Should we just use `WalkDir` in general? The doc only says the upside, not any downside, so I'm left wondering if it's always better. Reference for the the other `WalkDirFunc` https://cs.opensource.google/go/go/+/go1.20.2:src/io/fs/walk.go;l=69

server/pairing/payload_mounter.go
```diff
@@ -0,0 +1,339 @@
...
+
+func (r *RawMessagePayloadMounter) ToSend() []byte {
+	return r.encryptor.getEncrypted()
+}
+
+func (r *RawMessagePayloadMounter) LockPayload() {
+	r.encryptor.lockPayload()
+}
+
+func (r *RawMessagePayloadMounter) ResetPayload() {
```
ilmotta: Super light suggestion, we could use the nice naming `LockPayload` + `UnlockPayload` instead of `LockPayload` + `ResetPayload`. Or change `LockPayload` to `SetupPayload` + `ResetPayload`.

ilmotta (approved):
It was a pleasure to review this PR @Samyoul, although *relatively huge* :hear_no_evil: I definitely see many improvements to the clarity of the pairing processes.

In case it helps anybody, I better understood the motivation and changes after you shared with me this documentation you wrote some time ago https://notes.status.im/4VbZ0PlKR0GBE4d_A-qG_A#. It nicely complements the excellent PR description.

Since this is my first exposure to the lower levels of *pairing*, I feel like I'm missing the forest for some trees, so my approval should be taken lightly. Nonetheless, I left some minor comments.

server/pairing/client.go
Samyoul: Ah, you've really been digging into this code.

Ultimately signals are treated as outside the main critical path, so if a signal fails that failure should not interrupt the current process flow.

Additionally `json.Marshal()` generally only fails if you make a mistake in the code, and only if you are trying to marshal abstract things. There are 2 possible errors from this package `UnsupportedTypeError` and `UnsupportedValueError`, this means you've passed in something too abstract to represent as a JSON object.

You can see in [the source code for the go `json` package](https://github.com/golang/go/blob/master/src/encoding/json/encode.go#L426-L428) that if a type doesn't have a specific handler an `unsupportedTypeEncoder` will be returned.

Example:

```go
j, err := json.Marshal(make(chan struct{}))
if _, ok := err.(*json.UnsupportedTypeError); ok {
	fmt.Println("json error : json.UnsupportedTypeError")
}
fmt.Println("marshalled json : '" + string(j) + "'")

j, err = json.Marshal(func() {})
if _, ok := err.(*json.UnsupportedTypeError); ok {
	fmt.Println("json error : json.UnsupportedTypeError")
}
fmt.Println("marshalled json : '" + string(j) + "'")

j, err = json.Marshal(complex128(1))
if _, ok := err.(*json.UnsupportedTypeError); ok {
	fmt.Println("json error : json.UnsupportedTypeError")
}
fmt.Println("marshalled json : '" + string(j) + "'")

j, err = json.Marshal(math.Inf(1))
if _, ok := err.(*json.UnsupportedValueError); ok {
	fmt.Println("json error : json.UnsupportedValueError")
}
fmt.Println("marshalled json : '" + string(j) + "'")
```

Output:

```text
json error : json.UnsupportedTypeError
marshalled json : ''
json error : json.UnsupportedTypeError
marshalled json : ''
json error : json.UnsupportedTypeError
marshalled json : ''
json error : json.UnsupportedValueError
marshalled json : ''

Program exited.
```

See : https://go.dev/play/p/1dC31lpuB07

So you really would be trying to break the marshaller if you get an error from `json.Marshal()`. But ordinarily the error would be handled, as the error would constitute a critical failure.

With a signal failure, the worst that happens is that the signal is not sent, which although inconvenient isn't a processes halting

ilmotta: Now it all makes sense to me, we're good after all :) Thank you for the detailed explanation!

server/pairing/handlers.go
Samyoul: That's reasonable.

The reason the text is "pr.Receive(payload)", is because the code that generates the potential error is:

```go
err = pr.Receive(payload)
```

It just makes it easier to find if you search the codebase for the literal string. But I've had a think about what you say and I think that I will do one better. I'll change the text as you suggest, but I'll also pass in the input `payload` as an additional field to be logged.

server/pairing/handlers.go
Samyoul: Yes it is outdated and I will also apply the change discussed https://github.com/status-im/status-go/pull/3248#discussion_r1135840711

server/pairing/handlers.go
Samyoul: https://github.com/status-im/status-go/pull/3248#discussion_r1135840711

server/pairing/payload_receiver.go
Samyoul: Yes, I agree. Ah clarity of another perspective! Your suggestion is preferrable, I'm sad I didn't implement it myself. 😆

server/pairing/handlers.go
ilmotta: > The reason the text is "pr.Receive(payload)", is because the code that generates the potential error is

I thought it was the case, and grepping for the error is an important reason as you said.

> but I'll also pass in the input payload as an additional field to be logged.

:+1:

server/pairing/payload_receiver.go
Samyoul: That is reasonable, the codebase uses a combination of `0700`, `0755` and `0777` and the const `os.ModePerm` which resolves to `0777`.

I agree that it is better practice to use lower permissions particularly in the desktop environment, I'll lower the flags to `0700`

server/pairing/payload_receiver.go
ilmotta: I'm always afraid of getting into the weeds of naming, it's hard to see the fine line between personal opinion vs something marginally better. The risk of being pedantic is too high sometimes!

server/pairing/payload_receiver.go
Samyoul: Yep. I agree:

```go
	accountKey := new(keystore.EncryptedKeyJSONV3)
	if err := json.Unmarshal(rawKeyFile, &accountKey); err != nil {
		return fmt.Errorf("failed to read key file: %s", err)
	}

	if len(accountKey.Address) != 40 {
		return fmt.Errorf("account key address has invalid length '%s'", accountKey.Address)
	}
```

This should probably live in the `keystore` package, so I'll do that.

server/pairing/payload_receiver.go
Samyoul: Good catch.

server/pairing/payload_receiver.go
Samyoul: Yes. We can do file validation first then data store in a second loop.

If one of the keys fails, in either case, something quite bad has happened. So yes I agree the cleanest solution would be to attempt to revert the process. Depending on the error the application may no longer have io access, but that's probably why you've written "*attempt*". IF the clean up fails we just need to abort the process and start cascading errors.

server/pairing/payload_mounter.go
Samyoul: Hmm, this is a good point. `WalkDir` was introduced recently I've not even considered its usage. Oh I just did a Google search and it was released 2 years ago 😆. `WalkDir` seems to be superior. I'll update it.

Buuuut, now I have a question! This comment you've made is an unusually insightful suggestion and question and requires a large degree of context, how were you able to give such feedback?

server/pairing/payload_receiver.go
ilmotta: Yes, I agree with you it's definitely an edge case.

I was also thinking if the whole pairing process isn't idempotent (it is right?), so the user can just initiate the process again and the files would be recreated/replaced in case an exceptional error happened. As long as the user get the proper feedback from clients that could be more than good enough.

>  IF the clean up fails we just need to abort the process and start cascading errors.

Exactly @Samyoul, we can just try to revert on a best effort basis.

server/pairing/payload_mounter.go
Samyoul: We never unlock. Locking is permanent. Once locked we never want the code to access the payload again. But in tests we still want to check the values of the `EncryptionPayload` to ensure nothing has broken internally, so we don't simply delete the payload values.

`ResetPayload` is not the reversal of `LockPayload`, they are not relate in function. The main purpose of `ResetPayload` is to swap out the `EncryptionPayload` and remove any reference to the underlying values. A consequence of this is that the lock value is reset, but that is not the whole intent of the function. The main intent of the function is allow reuse of an encryptor without polluting the process with old references.

**However**, I've just reviewed `ResetPayload` and my refactor has made it obsolete, so I will remove `ResetPayload` from all implementations.

server/pairing/payload_mounter.go
ilmotta: > how were you able to give such feedback?

Oh that's a great question. I'm trying to learn Go for more than just closing issues, I'm assuming I know nothing, even though we all know programming languages have a lot of overlap. I really like that many Go libraries are reasonably well documented, they point out interesting points, not just *what* functions do. But in the end, curiosity? I find it incredibly satisfying to read API docs and jumping to source every now and then. Ah the beauty of open-source!

server/pairing/payload_mounter.go
ilmotta: Oh, I linked these pairs of actions way too much. It's much clearer now to me.

server/pairing/certs.go
```diff
@@ -18,6 +18,8 @@ import (
 	"github.com/status-im/status-go/signal"
 )
 
+// TODO Reconcile duplicate function here and in server/certs.go
```
qfrank, line 21: it would be better if we can link the issue instead of pure text description for TODO

server/pairing/client.go
```diff
@@ -71,117 +78,116 @@ func NewPairingClient(backend *api.GethStatusBackend, c *ConnectionParams, confi
...
+
+	bc, err := NewBaseClient(c)
+	if err != nil {
+		return nil, err
 	}
-}
 
-func (c *Client) sendInstallationData() error {
-	err := c.installationPayloadManager.Mount()
+	am, rmm, imr, err := NewPayloadMounters(logger, pe, backend, config.Sender)
```
qfrank: can we rename config.Sender to config.SenderConfig .. 🤔

server/pairing/client.go
```diff
@@ -232,169 +238,236 @@ func (c *Client) receiveInstallationData() error {
...
-		return err
+		return nil, err
 	}
-	signal.SendLocalPairingEvent(Event{Type: EventProcessSuccess, Action: ActionSyncDevice})
-	return nil
+
+	logger := logutils.ZapLogger().Named("ReceiverClient")
+	pe := NewPayloadEncryptor(c.aesKey)
+
+	ar, rmr, imr, err := NewPayloadReceivers(logger, pe, backend, config.Receiver)
```
qfrank: similar to above, config.ReceiverConfig 🤔

server/pairing/config.go
```diff
@@ -0,0 +1,100 @@
...
+	EK       []byte           `json:"-"`
+	Cert     *tls.Certificate `json:"-"`
+	Hostname string           `json:"-"`
+	Mode     Mode             `json:"-"`
+}
+
+type ClientConfig struct{}
+
+type SenderServerConfig struct {
+	Sender *SenderConfig `json:"sender"`
```
qfrank: can we rename Sender to SenderConfig ? `Sender` looks confused to me 🙂

server/pairing/config.go
```diff
@@ -0,0 +1,100 @@
...
+	Server *ServerConfig `json:"server"`
```
qfrank: can we rename Server to ServerConfig ?

server/pairing/server.go
```diff
@@ -115,106 +94,216 @@ func (s *Server) MakeConnectionParams() (*ConnectionParams, error) {
...
+		accountReceiver:      ar,
+		rawMessageReceiver:   rmr,
+		installationReceiver: imr,
+	}, nil
+}
 
-		AccountPayloadManagerConfig: accountPayloadManagerConfig,
+func (s *ReceiverServer) startReceivingData() error {
+	s.SetHandlers(server.HandlerPatternMap{
+		pairingChallenge:         handlePairingChallenge(s),
```
qfrank, line 267: `pairingChallenge` seems not being used by client in this case?

qfrank (approved):
Super super cool PR! Nice work!! Thank you @Samyoul 🫶 I left some comments(not big issue), the answer is up to you, this PR seems a break change one, changes to frontends also needed?

server/pairing/client.go
Samyoul: You know I thought about this while I was doing the PR, but now looking back at the code I agree with you.

It makes more sense to name the fields `*Config`, I will make these changes.

server/pairing/certs.go
Samyoul: Yes, I can do that. I've only noted these `TODO`s in this PR, but I can make issues for all the `TODO`s and then I will add links in the github issues for each of the `TODO`s.

Some of the `TODO`s are debatable, they may or may not be needed, they were just some things that I noticed, but perhaps with better context they may not be desirable or too much work.

I will make the issues, link them in the description of this PR, then I will make a new PR and include all the links in the new PR. Would that be acceptable to you @qfrank ?

server/pairing/server.go
Samyoul: Ah, that is true, I may have deleted the middleware. I have made a mistake here, you're original implementation was correct, because we are asking for data from the `ReceiverServer`.

I'm currently upgrading the Challenge lifecycle and management in this PR https://github.com/status-im/status-go/pull/3290. @qfrank would it be acceptable to address the issue you've risen here in PR #3290?

The worst case scenario if this went to production right now is that a malicious actor would be able to ask the `ReceiverServer` for an encrypted payload containing the installation meta data. Which is not useful to an attacker even if they could decrypt that data.

Samyoul:
> Super super cool PR! Nice work!! Thank you @Samyoul 🫶 I left some comments(not big issue), the answer is up to you, this PR seems a break change one, changes to frontends also needed?

Thank you very much @qfrank, I'm happy that you are happy 😆. I will address all the points you've made.

Yes you are correct, we will need to change the incoming configs from both Desktop and Mobile.

server/pairing/server.go
Samyoul: I've added an issue for this and it directly mentions our discussion on this issue https://github.com/status-im/status-go/issues/3299

server/pairing/certs.go
Samyoul: I've added all new issues now. All the TODOs mentioned in this PR's description are now replaced with links to issues. I will add the links to the code next.

server/pairing/client.go
igor-sirotin, line 1: This is quite a big file now. Maybe it would make sense to separate it into:
- BaseClient (client_base.go)
- SenderClient (client_sender.go)
- ReceiverClient (client_receiver.go)

I don't like putting `client_` in front, but it will make them visually group in the list of files.
Or maybe we could create a `client` subdirectory for those.

server/pairing/connection.go
```diff
@@ -69,6 +69,8 @@ func (cp *ConnectionParams) ToString() string {
 	ek := base58.Encode(cp.aesKey)
 	m := base58.Encode(new(big.Int).SetInt64(int64(cp.serverMode)).Bytes())
 
+	// TODO remove server mode from the connection string, rely on specific function calls rather than algorithmic orchestration
```
igor-sirotin, line 72: 👍

server/pairing/server.go
igor-sirotin, line 1: Same here, maybe divide into several files?

igor-sirotin (approved):
Nice PR with a very good description 👍

igor-sirotin:
As @qfrank already pointed out, we will need to make corresponding changes to Desktop and Mobile.
Maybe we could make a intermediate branch where all those changes are already reviewed/tested? Then we could use that branch to make required changes in desktop/mobile? 
Or any other way of preventing those changes to suddenly break both clients 🙂

server/pairing/client.go
Samyoul: Thank you for your review @igor-sirotin, I appreciate it.

May I respectfully disagree? The Go best practices document covers the topic of file length here https://google.github.io/styleguide/go/best-practices.html#package-size.

> There is no “one type, one file” convention as in some other languages. As a rule of thumb, files should be focused enough that a maintainer can tell which file contains something, and the files should be small enough that it will be easy to find once there. The standard library often splits large packages to several source files, grouping related code by file. **The source for [package bytes](https://go.dev/src/bytes/) is a good example.**

Looking at the [`bytes` source code](https://go.dev/src/bytes/bytes.go) you can see that the package contains many different but related types and that the file length is about 1300 lines. Does this seem reasonable to you?

qfrank:
> As @qfrank already pointed out, we will need to make corresponding changes to Desktop and Mobile. Maybe we could make a intermediate branch where all those changes are already reviewed/tested? Then we could use that branch to make required changes in desktop/mobile? Or any other way of preventing those changes to suddenly break both clients 🙂

we can merge this PR util the changes to mobile/desktop have been done and test get passed, i can help with the changes for mobile, or are you(@Samyoul ) going to do it?  @igor-sirotin

Samyoul:
> we can merge this PR util the changes to mobile/desktop have been done and test get passed, i can do the changes for mobile, or are you(@Samyoul ) going to do it? @igor-sirotin

@qfrank I am happy for you to do the changes to mobile. Thank you 🙏 .

To be clear, we will not merge this PR before the changes to mobile/desktop have been done and test. Is that correct?

qfrank:
> > we can merge this PR util the changes to mobile/desktop have been done and test get passed, i can do the changes for mobile, or are you(@Samyoul ) going to do it? @igor-sirotin
> 
> @qfrank I am happy for you to do the changes to mobile. Thank you 🙏 .
> 
> To be clear, we will not merge this PR before the changes to mobile/desktop have been done and test. Is that correct?

yes, i think so, or you have better solution to avoid introducing break changes into develop branch  ..

Samyoul:
@qfrank I do not have a better solution. This is the only way I think we can avoid introducing a breaking change.

server/pairing/config.go
```diff
@@ -0,0 +1,100 @@
...
+	EK       []byte           `json:"-"`
+	Cert     *tls.Certificate `json:"-"`
+	Hostname string           `json:"-"`
+	Mode     Mode             `json:"-"`
+}
+
+type ClientConfig struct{}
+
+type SenderServerConfig struct {
+	SenderConfig *SenderConfig `json:"sender"`
```
qfrank: can we rename the json field as well 🤔 @Samyoul

Samyoul:
@igor-sirotin do you know if `status-desktop` has a matching PR for the changes made in this PR, like this one https://github.com/status-im/status-mobile/pull/15412 ?

igor-sirotin:
@Samyoul, nope, there's no work yet have been done. I was polishing some things about pairing, will start upgrading to these changes tomorrow 👌🏻

igor-sirotin:
@Samyoul @qfrank 
Here's desktop update: https://github.com/status-im/status-desktop/pull/9992

Tested with mobile too, it works 👍

PR #3251, opened 2023-03-02
fix: reduce log level for closing conn messages

### LINKED ISSUE #3249: Reduce verbosity of waku v2 logs for libp2p connections
Right now when running `status-go` with Waku v2 enabled we are generating a ridiculoud amount of logs like this:
```
INFO [03-02|13:13:36.183] closing conn         peer=16Uiu2HAkyjvXPmymR5eRnvxCufRGZdfRrgjME6bmn3Xo6aprE1eo
```
Here's logs from 1.5 days:
```
admin@node-01.sm-eu-zur1.spiff.app-test:~ % ls -lh /var/log/docker/spiff-workflow-wakunode/docker.log
-rw-r--r-- 1 syslog syslog 13G Mar  2 13:13 /var/log/docker/spiff-workflow-wakunode/docker.log

admin@node-01.sm-eu-zur1.spiff.app-test:~ % grep 'closing conn' /var/log/docker/spiff-workflow-wakunode/docker.log | wc -l
57729782
```
That's just not reasonable. Why would such a log be `INFO` level?
https://github.com/status-im/status-go/blob/e1ea1ccab5c6cf2be63dcda844e412436aa2484a/vendor/github.com/libp2p/go-libp2p/p2p/net/connmgr/connmgr.go#L374-L381

This is forcing me to use `ERROR` log level for all `status-go` nodes using Waku v2 because this volume is absurd.
I'd appreciate if we could make this `DEBUG`, or even better, `TRACE` level log.

richard-ramos:
Fixes #3249

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,8 @@ replace github.com/nfnt/resize => github.com/status-im/resize v0.0.0-20201215164
 
 replace github.com/forPelevin/gomoji => github.com/status-im/gomoji v1.1.3-0.20220213022530-e5ac4a8732d4
 
+replace github.com/libp2p/go-libp2p => github.com/status-im/go-libp2p v0.25.1-debug-closing-conn
+
 require (
 	github.com/anacrolix/torrent v1.41.0
 	github.com/beevik/ntp v0.3.0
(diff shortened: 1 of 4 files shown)


richard-ramos:
Created this PR: https://github.com/libp2p/go-libp2p/pull/2165  . Hopefully it will be accepted so we dont have to maintain a fork of go-libp2p

jakubgs (approved):
Thanks!

PR #3258, opened 2023-03-05
[15128] Add mute for a specific amount of time

ibrkhalil:
This PR adds ability to mute a chat for a specific time by calculating the difference between the muting action timestamp and the current timestamp
Had to create a V2 endpoint to not break the status-desktop implementation
Closes #14556 on status-mobile

- [x] note to self : NEED TO MAKE SURE IT'S BACKWARDS COMPATIBLE


protocol/chat.go
```diff
@@ -40,6 +40,23 @@ const (
 	FirstMessageTimestampNoMessage = 1
 )
 
+const (
+	MuteFor15MinsDuration = 900
```
cammellos: It'd be nice to have them in order of duration

protocol/common/message.go
```diff
@@ -49,6 +49,12 @@ type QuotedMessage struct {
 
 type CommandState int
 
+type MuteParams struct {
```
cammellos: There's a namespace for API parameters `protocol/requests`, we should move this there

protocol/messenger.go
```diff
@@ -4716,19 +4716,26 @@ func (m *Messenger) MuteChat(chatID string) error {
...
 	if chat.OneToOne() {
-		contact, _ = m.allContacts.Load(chatID)
+		contact, _ = m.allContacts.Load(MuteParams.ChatID)
 	}
 
-	return m.muteChat(chat, contact)
+	return m.muteChat(chat, contact, MuteParams.MutedAt, MuteParams.MutedType)
+}
+
+func (m *Messenger) MuteChatV2(chatID string, mutedAt int64, muteType int) error {
```
cammellos: You can use a single map parameter and add it to `protocol/requests`

protocol/messenger_communities.go
```diff
@@ -467,7 +467,7 @@ func (m *Messenger) SetMutePropertyOnChatsByCategory(communityID string, categor
 
 	for _, chatID := range community.ChatsByCategoryID(categoryID) {
 		if muted {
-			err = m.MuteChat(communityID + chatID)
+			err = m.MuteChat(&common.MuteParams{ChatID: communityID + chatID, MutedAt: 0, MutedType: 5})
```
cammellos: `MutedType` should not use `5` but the const you defined

protocol/messenger_handler.go
```diff
@@ -583,7 +583,7 @@ func (m *Messenger) HandleSyncInstallationContact(state *ReceivedMessageState, m
 		}
 		if chat != nil && message.Muted != chat.Muted {
 			if message.Muted {
-				err := m.muteChat(chat, contact)
+				err := m.muteChat(chat, contact, 0, 5)
```
cammellos: no magic numbers

services/ext/api.go
```diff
@@ -302,8 +302,11 @@ func (api *PublicAPI) UnmuteCommunityCategory(communityID string, categoryID str
 	return api.service.messenger.SetMutePropertyOnChatsByCategory(communityID, categoryID, false)
 }
 
+func (api *PublicAPI) MuteChatV2(parent context.Context, chatID string, MutedAt int64, MuteType int) error {
```
cammellos: we can use a struct for parameters under `protocol/requests`

protocol/chat.go
```diff
@@ -40,6 +40,23 @@ const (
...
+	MuteFor1HrsDuration   = 3600
+	MuteFor8HrsDuration   = 28800
+	MuteFor1WeekDuration  = 604800
+	MuteFor1MinDuration   = 60
+)
+
+const (
+	MuteFor15MinsType = 1
```
cammellos: You can use `iota`  https://yourbasic.org/golang/iota/ , starting at 1, any specific reason why you want `MuteTill1MinType` as `99`?

protocol/messenger_chats.go
```diff
@@ -104,6 +107,56 @@ func (m *Messenger) ChatsPreview() []*ChatPreview {
 				}
 			}
 
+			switch chatPreview.MuteType {
```
cammellos: This bit I am not sure, I think we need to discuss a bit the strategy on what we want to do and how we want to handle muted timing

cammellos (commented):
@ibrkhalil great job overall!
I think the skeleton of the PR is correct, I left some comments, but we need to discuss probably how we want to handle muting, as of now it looks like `Muted` is still used for the actual functionality https://github.com/status-im/status-go/blob/f60716412259aece88d899d09fc71de4c4ba5d2e/protocol/messenger.go#L2962

Also the client won't be notified when a chat "unmutes", so probably best to have a quick chat on how to implement this.

protocol/chat.go
ibrkhalil: Because this was just a testing value to help QA test the functionality .. Will be removed once tested and approved

protocol/messenger.go
ibrkhalil: Well, We definitely can .. What do you prefer to be sent over the wire from status-mobile to status-go .. maps or values ?

ibrkhalil: And if it's a map .. Will I send it as a javascript object or a CLJ map?

ilmotta: I, for one, much prefer maps in `status-go` endpoints because they're open and extensible. Endpoints with positional arguments can't evolve their signature without causing breaking changes on clients.

>And if it's a map .. Will I send it as a javascript object or a CLJ map?

In Mobile you just send a CLJ map in the `json-rpc/call` effect and it'll be correctly serialized for status-go.

protocol/messenger_chats.go
ibrkhalil: Implemented the goroutine

protocol/messenger.go
```diff
@@ -1337,6 +1338,50 @@ func (m *Messenger) watchExpiredMessages() {
...
+		for {
+			chats, _ := m.persistence.Chats()
+			select {
+			case <-time.After(3 * time.Second): // Poll every 3 seconds
+				for _, chat := range chats {
+					switch chat.MuteType {
+					case MuteFor15MinsType:
+						if time.Now().Unix()-chat.MutedAt > MuteFor15MinsDuration && chat.MuteType != 0 {
+							_ = m.persistence.UnmuteChat(chat.ID)
+							m.config.messengerSignalsHandler.ChatUnmutedEvent(chat.ID)
```
cammellos: Here I would just get all the chats and signal once with a `MessengerResponse`, that way you don't have to handle a new signal in the frontend

protocol/messenger.go
```diff
@@ -1337,6 +1338,50 @@ func (m *Messenger) watchExpiredMessages() {
...
+						}
+					case MuteFor1WeekType:
+						if time.Now().Unix()-chat.MutedAt > MuteFor1WeekDuration && chat.MuteType != 0 {
+							_ = m.persistence.UnmuteChat(chat.ID)
+							m.config.messengerSignalsHandler.ChatUnmutedEvent(chat.ID)
+						}
+					case MuteTill1MinType:
+						if time.Now().Unix()-chat.MutedAt > MuteFor1MinDuration && chat.MuteType != 0 {
+							_ = m.persistence.UnmuteChat(chat.ID)
```
cammellos: You have to check errors

protocol/messenger.go
```diff
@@ -1337,6 +1338,50 @@ func (m *Messenger) watchExpiredMessages() {
...
 }
 
+// watchUnmutedChats regularly checks for chats that should be unmuted
+func (m *Messenger) watchUnmutedChats() {
+	m.logger.Debug("watching unmuted chats")
+	go func() {
+		for {
+			chats, _ := m.persistence.Chats()
+			select {
+			case <-time.After(3 * time.Second): // Poll every 3 seconds
```
ilmotta: Isn't this loop too energy hungry to run every 3s? Especially since it talks to the database to fetch chats. I'm worried about battery consumption for mobile devices.

cammellos: once every 3 seconds is not going to have any noticeable effect on battery life etc, but it should not poll chats from the db, we actually keep them in memory, so you can reach into messenger (`allChats` and loop)

ibrkhalil: If we're going to keep chats in db, Aren't they not going to be updated when the polling happens again ?
And therefore will always be returned

cammellos: we keep chats in the db, you need to save them, but we keep them in memory as well, and we use references, so modifying them in memory will have the effect you want

ilmotta: >once every 3 seconds is not going to have any noticeable effect on battery life etc

I can't say much, never measured this, but I take your word for now ;)

>we actually keep them in memory, so you can reach into messenger (allChats and loop)

But what if the list of chats change, how could fetching from memory see the most up-to-date view of the data? Do you mean that `allChats` is already kept updated by some other goroutine outside this periodic one?

cammellos: > But what if the list of chats change

It works very much the same as status-mobile, basically we keep all chats in memory, if you change anything, it will be reflected every else, and we just flush to disk every time something changes, but maybe I misunderstand the concern, did I answer the right question :) ?

> I can't say much, never measured this, but I take your word for now ;)

Basically if you think about it, generally we would worry about stuff that happens < 50ms, or busy loops etc, but this is just 3 seconds of inactivity, which is a very long time, almost negligible in computer times, and then some activity which is moderate work, but nothing crazy. There's a cost in creating a new goroutine, but they are fairly cheap as they are green threads, so this is most likely fine. (as a comparison think about a network connection or polling mechanism, that would have to be activated much more often and it would require much more work, but still it's something generally apps handle ok).

ilmotta: > It works very much the same as status-mobile, basically we keep all chats in memory, if you change anything, it will be reflected every else, and we just flush to disk every time something changes, but maybe I misunderstand the concern, did I answer the right question :) ?

You did and thank you :)

> Basically if you think about it, generally we would worry about stuff that happens < 50ms

I agree, it should be fine, especially in isolation, the cost might be negligible. I was thinking more in terms of total cost, since the mobile env is quite constrained, so everything adds up. I was also wondering if the mobile OS won't be able to put the process to sleep less frequently, which might affect battery consumption, but I have no clue if that's really a thing to consider, likely not.

It should be fine as you said :+1:, especially because we can take the DB load out of the loop.

protocol/chat.go
```diff
@@ -89,6 +106,13 @@ type Chat struct {
 	// push notifications for this chat
 	Muted bool `json:"muted"`
 
+	// Time in which chat was muted
+	MuteTill int64 `json:"mutedTill,omitempty"`
```
cammellos: I would name the json and the field the same, now they are: `MuteTill` and `mutedTill`, so I would pick one or the other (`mutedUntil`, or `mutedTill` sounds better)

protocol/chat.go
```diff
@@ -89,6 +106,13 @@ type Chat struct {
...
+
+	// chat mute type, Can be from 15 mins to 7 days .. And can be
+	// till turned back on
+	MuteType int `json:"muteType,omitempty"`
```
cammellos: I believe we discussed we can remove this

protocol/chat.go
```diff
@@ -40,6 +40,23 @@ const (
...
+const (
+	MuteFor1MinDuration   = 60
+	MuteFor15MinsDuration = 900
+	MuteFor1HrsDuration   = 3600
+	MuteFor8HrsDuration   = 28800
+	MuteFor1WeekDuration  = 604800
+)
+
+const (
+	MuteFor15MinsType = 1
```
cammellos: You can also use a different type for this, to have a bit more type safety

protocol/messenger.go
```diff
@@ -1337,6 +1339,32 @@ func (m *Messenger) watchExpiredMessages() {
...
 }
 
+// watchUnmutedChats regularly checks for chats that should be unmuted
+func (m *Messenger) watchUnmutedChats() {
+	m.logger.Debug("watching unmuted chats")
+	go func() {
+		for {
+			select {
+			case <-time.After(3 * time.Second): // Poll every 3 seconds
+				chats, _ := m.persistence.Chats()
```
cammellos: this should use in memory chats `m.allChats.Range...`, there's a few examples in the codebase

protocol/messenger.go
```diff
@@ -1337,6 +1339,32 @@ func (m *Messenger) watchExpiredMessages() {
...
+				for _, chat := range chats {
+					if time.Now().Unix() > chat.MuteTill && chat.MuteTill != 0 && chat.Muted {
+						_ = m.persistence.UnmuteChat(chat.ID)
+						response := &MessengerResponse{}
```
cammellos: You want to have the response outside the `chats` loop, inside the chat loop you do `response.AddChat(chat)`
and after the loop you dispatch the signal.

So instead of sending 1 signal per chat, you send a single singnal.
Also you want to check if there's anything to dispatch if you do that, since we don't want to send empty signals

protocol/messenger.go
```diff
@@ -4692,54 +4720,86 @@ func (m *Messenger) MarkAllReadInCommunity(communityID string) ([]string, error)
 
 // MuteChat signals to the messenger that we don't want to be notified
 // on new messages from this chat
-func (m *Messenger) MuteChat(chatID string) error {
-	chat, ok := m.allChats.Load(chatID)
+func (m *Messenger) MuteChat(MuteParams *requests.MuteParams) (int64, error) {
```
cammellos: The convention is not to use `MuteParams` as a name, we use the name of the call for the most part,
so the type should be `*requests.MuteChat`

protocol/messenger.go
```diff
@@ -4692,54 +4720,86 @@ func (m *Messenger) MarkAllReadInCommunity(communityID string) ([]string, error)
...
+	chat, ok := m.allChats.Load(MuteParams.ChatID)
```
cammellos: Variable names should be lower case, so `muteParams`. I would call it `request` here, also don't forget to `validate` before using it.

protocol/messenger.go
```diff
@@ -4692,54 +4720,86 @@ func (m *Messenger) MarkAllReadInCommunity(communityID string) ([]string, error)
...
+	case MuteFor1HrsType:
+		MuteParams.MutedTill = time.Now().Unix() + MuteFor1HrsDuration
+	case MuteFor8HrsType:
+		MuteParams.MutedTill = time.Now().Unix() + MuteFor8HrsDuration
+	case MuteFor1WeekType:
+		MuteParams.MutedTill = time.Now().Unix() + MuteFor1WeekDuration
+	default:
+		MuteParams.MutedTill = 0
 	}
+	saveMuteTillErr := m.saveChat(chat)
```
cammellos: calling it `err` is generally more conventional, we don't use new variables for err

protocol/messenger.go
```diff
@@ -4692,54 +4720,86 @@ func (m *Messenger) MarkAllReadInCommunity(communityID string) ([]string, error)
...
 
 	var contact *Contact
 	if chat.OneToOne() {
-		contact, _ = m.allContacts.Load(chatID)
+		contact, _ = m.allContacts.Load(MuteParams.ChatID)
+	}
+
+	switch MuteParams.MutedType {
+	case MuteTill1MinType:
+		MuteParams.MutedTill = time.Now().Unix() + MuteFor1MinDuration
```
cammellos: not sure you need `MutedTill` in the `MuteParams` structure, I would just create a local variable for this, but I might be missing something

protocol/messenger_mute_test.go
```diff
@@ -107,3 +111,57 @@ func (s *MessengerMuteSuite) TestSetMute() {
...
+	if actualChat.OneToOne() {
+		contact, _ = s.m.allContacts.Load(chatID)
+	}
+
+	_, error := s.m.muteChat(actualChat, contact, int64(mockTime))
+	s.NoError(error)
+	// Mock Routine
+	for _, chat := range allChats {
+		if time.Now().Unix() > chat.MuteTill && chat.MuteTill != 0 && chat.Muted {
+			log.Print("hey", time.Now().Unix(), chat.MuteTill)
```
cammellos: extra log

protocol/requests/mute_chat_request.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"errors"
+)
+
+var ErrInvalidMuteChatParams = errors.New("mute-chat: invalid params")
+
+type MuteParams struct {
+	ChatID    string
+	MutedTill int64
```
cammellos: Do you need this? it seems it should not be set by the client

protocol/requests/mute_chat_request.go
```diff
@@ -0,0 +1,21 @@
+package requests
+
+import (
+	"errors"
+)
+
+var ErrInvalidMuteChatParams = errors.New("mute-chat: invalid params")
+
+type MuteParams struct {
```
cammellos: The name of this struct should be `MuteChat`, we use the name of the endpoint for the request

protocol/requests/mute_chat_request.go
```diff
@@ -0,0 +1,21 @@
...
+	ChatID    string
+	MutedTill int64
+	MutedType int
```
cammellos: You will have to validate as well `MutedType` in the validate fuction

protocol/chat.go
ibrkhalil: Hmmm, Do you suggest any specific type ?

cammellos: You create one, for enums is customary to have their own type, for example:

```
type ActivityCenterType int

const (
	ActivityCenterNotificationNoType ActivityCenterType = iota
	ActivityCenterNotificationTypeNewOneToOne
	ActivityCenterNotificationTypeNewPrivateGroupChat
	ActivityCenterNotificationTypeMention
	ActivityCenterNotificationTypeReply
	ActivityCenterNotificationTypeContactRequest
	ActivityCenterNotificationTypeCommunityInvitation
	ActivityCenterNotificationTypeCommunityRequest
	ActivityCenterNotificationTypeCommunityMembershipRequest
	ActivityCenterNotificationTypeCommunityKicked
	ActivityCenterNotificationTypeContactVerification
)

```

Note that the first type, is the "unknown" type, since if a value is not in the enum, it will default to that one in some cases.

protocol/messenger.go
```diff
@@ -1337,6 +1339,31 @@ func (m *Messenger) watchExpiredMessages() {
...
+func (m *Messenger) watchUnmutedChats() {
+	m.logger.Debug("watching unmuted chats")
+	go func() {
+		for {
+			select {
+			case <-time.After(3 * time.Second): // Poll every 3 seconds
+				response := &MessengerResponse{}
+				m.allChats.Range(func(chatID string, c *Chat) bool {
+					if time.Now().Unix() > c.MuteTill && c.MuteTill != 0 && c.Muted {
+						_ = m.persistence.UnmuteChat(c.ID)
```
cammellos: You need to handle the error here, I would stop the loop and return

protocol/messenger.go
```diff
@@ -4692,54 +4719,88 @@ func (m *Messenger) MarkAllReadInCommunity(communityID string) ([]string, error)
...
+		return 0, Err
 	}
 
-	return m.muteChat(chat, contact)
+	return m.muteChat(chat, contact, MuteTill)
 }
 
-func (m *Messenger) muteChat(chat *Chat, contact *Contact) error {
-	err := m.persistence.MuteChat(chat.ID)
+func (m *Messenger) MuteChatV2(MuteParams *requests.MuteChat) (int64, error) {
```
cammellos: `MuteParams` should be `muteParams`

protocol/chat.go
```diff
@@ -40,6 +41,28 @@ const (
...
+const (
+	MuteFor15MinsType requests.MutingType = iota + 1
+	MuteFor1HrsType
+	MuteFor8HrsType
+	MuteFor1WeekType
+	MuteTillUnmuted
+	MuteTill1MinType
+)
+
+type MuteChat struct {
```
cammellos: Are you still using this one? or it can be removed?

protocol/messenger.go
ibrkhalil: I returned `false`

protocol/messenger.go
ibrkhalil: Camel case done

protocol/chat.go
ibrkhalil: Sorry forgot to remove it .. Removed

protocol/chat.go
```diff
@@ -40,6 +41,23 @@ const (
...
+	MuteFor1MinDuration   = 60
+	MuteFor15MinsDuration = 900
+	MuteFor1HrsDuration   = 3600
+	MuteFor8HrsDuration   = 28800
+	MuteFor1WeekDuration  = 604800
+)
+
+const (
+	MuteFor15MinsType requests.MutingType = iota + 1
+	MuteFor1HrsType
```
ilmotta: Adding `Type` at the end seems superfluous, since any type is a type ;)

protocol/chat.go
```diff
@@ -40,6 +41,23 @@ const (
 	FirstMessageTimestampNoMessage = 1
 )
 
+const (
+	MuteFor1MinDuration   = 60
+	MuteFor15MinsDuration = 900
+	MuteFor1HrsDuration   = 3600
+	MuteFor8HrsDuration   = 28800
+	MuteFor1WeekDuration  = 604800
```
ilmotta: Perhaps a better idea is to store these constants as durations, instead of integers. This way we can manipulate time more easily, like so:

```go
const (
    MuteFor1MinDuration   = time.Minute
    MuteFor15MinsDuration = 15 * time.Minute
    MuteFor1HrsDuration   = time.Hour
    MuteFor8HrsDuration   = 8 * time.Hour
    MuteFor1WeekDuration  = 7 * 24 * time.Hour
)

// time.Now().Add(MuteFor15MinsDuration).Unix()
```

Please, see my other comment in the function `MuteChat` for a better example.

protocol/messenger.go
```diff
@@ -4692,54 +4722,88 @@ func (m *Messenger) MarkAllReadInCommunity(communityID string) ([]string, error)
...
 	if chat.OneToOne() {
-		contact, _ = m.allContacts.Load(chatID)
+		contact, _ = m.allContacts.Load(request.ChatID)
+	}
+
+	MuteTill := int64(0)
+
+	switch request.MutedType {
+	case MuteTill1MinType:
+		MuteTill = time.Now().Unix() + MuteFor1MinDuration
```
ilmotta: Disclaimer: my recommendation here may be controversial for some.

If we store the durations as `time.Duration` instead of integers, then we can more safely manipulate time and the type system becomes more useful. Also, when we convert to a Unix timestamp, we lose information from a `time.Time` value. Function signatures become more meaningful, using `time.Time` instead of `int64`.

The general idea is that we can keep passing `time.Time` around and transform to a Unix timestamp only at the edges of the system (to store in the DB, when storing in structs for JSON marshaling, etc).

Memory consumption should be fine, and would be a good tradeoff for the added maintainability IMO.

Example:

```go
const (
    MuteFor1MinDuration   = time.Minute
    MuteFor15MinsDuration = 15 * time.Minute
    // ...
)

switch request.MutedType {
case MuteTill1MinType:
    MuteTill = time.Now().Add(MuteFor1MinDuration)
case MuteFor15MinsType:
    MuteTill = time.Now().Add(MuteFor15MinsDuration)
// ...
default:
    MuteTill = time.Unix(0, 0)
}
```

protocol/messenger.go
```diff
@@ -4692,54 +4722,88 @@ func (m *Messenger) MarkAllReadInCommunity(communityID string) ([]string, error)
...
+	case MuteFor1HrsType:
+		MuteTill = time.Now().Unix() + MuteFor1HrsDuration
+	case MuteFor8HrsType:
+		MuteTill = time.Now().Unix() + MuteFor8HrsDuration
+	case MuteFor1WeekType:
+		MuteTill = time.Now().Unix() + MuteFor1WeekDuration
+	default:
+		MuteTill = 0
 	}
+	Err := m.saveChat(chat)
```
ilmotta: The error variable should be named `err` lowercase.

protocol/requests/mute_chat_request.go
```diff
@@ -0,0 +1,27 @@
...
+
+type MutingType int
+
+type MuteChat struct {
+	ChatID    string
+	MutedType MutingType
+}
+
+func (a *MuteChat) Validate() error {
+
```
ilmotta: Extra line.

protocol/messenger_mute_test.go
```diff
@@ -107,3 +110,56 @@ func (s *MessengerMuteSuite) TestSetMute() {
...
+	}
+
+	var contact *Contact
+	if actualChat.OneToOne() {
+		contact, _ = s.m.allContacts.Load(chatID)
+	}
+
+	_, error := s.m.muteChat(actualChat, contact, int64(mockTime))
+	s.NoError(error)
+	// Mock Routine
```
ilmotta, line 154: This test could be improved. I was thinking about testing `watchUnmutedChats` directly. The test would basically call `watchUnmutedChats` to start polling for chats and another go routine would add dummy chats that should be muted or not, depending on the test case, and with `time.Sleep`s controlled by the test.

In order to avoid time bombs in tests, it'd be ideal if we could do time travel and control what `time.Now()` means. This is a common problem in every test suite and programming language.

Here in Go I found this [gomonkey](https://github.com/agiledragon/gomonkey) library, but maybe that's a bazooka... Anyway, it'd be good if we had a way to better control time in tests, since status-go deals a lot with such things. The simpler way is to be able to pass the current time as a parameter, but I've found so many instances where this idea can't be cleanly applied that it's good to have a second solution.

I understand this might be too much for this PR, but the idea has been shared :wink:

protocol/requests/mute_chat_request.go
```diff
@@ -0,0 +1,27 @@
+package requests
+
+import (
+	"errors"
+)
+
+var ErrInvalidMuteChatParams = errors.New("mute-chat: invalid params")
+
+type MutingType int
```
ilmotta: Again, maybe the suffix `Type` is not the best.

protocol/chat.go
ibrkhalil: Done .. Actually this is easier to grasp .. But I had to cast their values to integers to get numeric values like 60 instead of something like `1m00`

protocol/messenger.go
ibrkhalil: Oh I get it now .. We're trying to not treat time as Unix integers but as time objects .. Trying

protocol/messenger_mute_test.go
ibrkhalil: I feel like it's a bazooka for such a simple task too .. But it's something to keep in mind when writing tests

protocol/chat.go
```diff
@@ -40,6 +42,23 @@ const (
...
+	MuteFor1MinDuration   = time.Minute / time.Second
+	MuteFor15MinsDuration = 15 * time.Minute / time.Second
+	MuteFor1HrsDuration   = time.Hour / time.Second
+	MuteFor8HrsDuration   = 8 * time.Hour / time.Second
+	MuteFor1WeekDuration  = 7 * 24 * time.Hour / time.Second
+)
+
+const (
+	MuteFor15Mins requests.MutingVariation = iota + 1
+	MuteFor1Hrs
```
ilmotta: Nitpick on my part, but the abbreviated unit for hour is just `h` and `1 hour` is on the singular form. So the constant could be `MuteFor1H` (weird to me), or `MuteFor1Hour` (the word is longer, but it's also following the same name used by `time.Hour`).

It doesn't matter, just sharing this tiny little thing.

protocol/messenger.go
```diff
@@ -4692,54 +4722,88 @@ func (m *Messenger) MarkAllReadInCommunity(communityID string) ([]string, error)
...
 		}
 
 		// Create a one to one chat and set active to false
-		chat = CreateOneToOneChat(chatID, publicKey, m.getTimesource())
+		chat = CreateOneToOneChat(request.ChatID, publicKey, m.getTimesource())
 		chat.Active = false
 		err = m.initChatSyncFields(chat)
 		if err != nil {
-			return err
+			return time.Unix(0, 0), err
```
ilmotta: Even though using `time` instances is better in many ways, the fact that the zero value needs to redefined over and over is "uglier". If you're using `time.Unix(0, 0)` a lot of times, it may be worth extracting to a constant named `epoch`.

> The Unix [epoch](https://en.wikipedia.org/wiki/Epoch_(computing)) is the time 00:00:00 UTC on 1 January 1970.
>
> -- https://en.wikipedia.org/wiki/Unix_time#cite_note-single-unix-spec-4.16-3

ilmotta (approved):
@ibrkhalil, great work and thank you for taking the time to work on the feedback I gave.

Please, remember to also change the `VERSION` file in this PR, since you'll need to git tag after merging.

protocol/messenger.go
ibrkhalil: I used `time.Time{}`

protocol/messenger.go
```diff
@@ -4692,54 +4724,88 @@ func (m *Messenger) MarkAllReadInCommunity(communityID string) ([]string, error)
...
 	if chat.OneToOne() {
-		contact, _ = m.allContacts.Load(chatID)
+		contact, _ = m.allContacts.Load(request.ChatID)
+	}
+
+	MuteTill := time.Time{}
+
+	switch request.MutedType {
+	case MuteTill1Min:
+		MuteTill = time.Now().Add(time.Duration(MuteFor1MinDuration))
```
ilmotta: @ibrkhalil, why do we need to cast to a duration if the constant is already a `time.Duration`?

protocol/chat.go
```diff
@@ -40,6 +42,23 @@ const (
 	FirstMessageTimestampNoMessage = 1
 )
 
+const (
+	MuteFor1MinDuration   = time.Minute
+	MuteFor15MinsDuration = 15 * time.Minute
+	MuteFor1HrsDuration   = time.Hour
+	MuteFor8HrsDuration   = 8 * time.Hour / time.Second
```
ilmotta: I think one of the selling points of using something like `time.Hour` is to help us delay the decision about the time precision to the consumer of these constants, i.e. the places that need to do the math in seconds, microseconds, etc. But if you want to store the duration in seconds, I think this should be reflected in the name of the constants (e.g.  `MuteFor1MinDurationInSeconds`), since a duration by default has a nanosecond precision in Go.

protocol/messenger.go
ibrkhalil: I can't remember exactly why I did it, let me test and reply

ibrkhalil: Removed

protocol/chat.go
ibrkhalil: Got this warning from our linter
```protocol/chat.go:46:2: var MuteFor1MinDurationInSecs is of type time.Duration; don't use unit-specific suffix "Secs" (golint)``` after trying seconds

protocol/messenger.go
ibrkhalil: Turned out to be unnecessary .. Removed

protocol/chat.go
ilmotta: Interesting, thanks for sharing @ibrkhalil. I respectfully disagree with the linter ;) Looks like we should follow the linter suggestion then.

ibrkhalil:
@vkjr @ilmotta 
Please take a look and check if these changes c7eb0918230b3a4485a571f230f9ebe70b68693c are correct or not, I was request to not send AC notifications if chat is muted

vkjr:
> @vkjr @ilmotta Please take a look and check if these changes [c7eb091](https://github.com/status-im/status-go/commit/c7eb0918230b3a4485a571f230f9ebe70b68693c) are correct or not, I was request to not send AC notifications if chat is muted

I think we don't need to build a notification structure at all in this case. There is a function call `showMentionOrReplyActivityCenterNotification()` above your changes, I believe you can add a check for muted chat there. It already lists a bunch of checks when notification is not needed.  Wdyt?

ibrkhalil:
> > @vkjr @ilmotta Please take a look and check if these changes [c7eb091](https://github.com/status-im/status-go/commit/c7eb0918230b3a4485a571f230f9ebe70b68693c) are correct or not, I was request to not send AC notifications if chat is muted
> 
> I think we don't need to build a notification structure at all in this case. There is a function call `showMentionOrReplyActivityCenterNotification()` above your changes, I believe you can add a check for muted chat there. It already lists a bunch of checks when notification is not needed. Wdyt?

Totally agree, Thanks for the wonderful suggestion! :)

vkjr:
@ibrkhalil, just to clarify, I mean putting the check inside `showMentionOrReplyActivityCenterNotification()`

ibrkhalil:
> @ibrkhalil, just to clarify, I mean putting the check inside `showMentionOrReplyActivityCenterNotification()`

Done, Thanks you!

PR #3259, opened 2023-03-06
feat: allow group admin to delete others message

### LINKED ISSUE #15182: The message non-admin user's message is not removed in group chat if 'delete for everyone' option is used
#### Steps to reproduce:
1. Create a group chat as User_A (admin)
2. Invite User_B
3. Send the message as User_B
4. Remove the message as User_A using 'delete for everyone' option
5. Check the message as User_B
6. Reopen chat as User_A

#### Actual behavior
The message is shown

https://user-images.githubusercontent.com/52490791/221178636-7636a6a7-afff-413c-a583-e00fa7ad0919.mp4

#### Expected behavior
The message is removed for both users

### Additional Information

- Status version: nightly 24 Feb 2023

yqrashawn:
-------

Allow group admin to delete members' message

Check DeleteMessage.From against group admins in private and public group chat

Closes https://github.com/status-im/status-mobile/issues/15182


protocol/messenger_messages.go
```diff
@@ -125,6 +125,18 @@ func (m *Messenger) DeleteMessageAndSend(ctx context.Context, messageID string)
 			}
 		}
 
+		if message.MessageType == protobuf.MessageType_PRIVATE_GROUP || message.MessageType == protobuf.MessageType_PUBLIC_GROUP {
+			group, err := newProtocolGroupFromChat(chat)
+			if err != nil {
+				return nil, err
+			}
+			admins := group.Admins()
+			canDeleteMessageForEveryone = stringSliceContains(admins, common.PubkeyToHex(&m.identity.PublicKey))
```
ilmotta: @yqrashawn, given the body of the `if` is basically the same here as in `messenger_handler.go`, what do you think of creating a single function and parameterize the second argument to `stringSliceContains` (the only difference)?

It looks like a true duplication, so whenever someone changes A will necessarily need to change B, so that's why I'm suggesting this refactoring.

protocol/messenger_handler.go
```diff
@@ -1541,6 +1541,19 @@ func (m *Messenger) HandleDeleteMessage(state *ReceivedMessageState, deleteMessa
 				return ErrInvalidDeletePermission
 			}
 		}
+
+		if chat.ChatType == ChatTypePrivateGroupChat || chat.ChatType == ChatTypePublic {
```
ilmotta: Would be good to get some test coverage for the changes in `HandleDeleteMessage`.

jrainville: Public chats don't have admins so you can remove the `ChatTypePublic` condition.

Also, you can it an `else if` with the condition above, since it cannot be a community chat and a group chat.

protocol/messenger_messages.go
```diff
@@ -125,6 +125,18 @@ func (m *Messenger) DeleteMessageAndSend(ctx context.Context, messageID string)
 			}
 		}
 
+		if message.MessageType == protobuf.MessageType_PRIVATE_GROUP || message.MessageType == protobuf.MessageType_PUBLIC_GROUP {
```
jrainville: Same comment as above.

protocol/messenger_messages.go
jrainville: Agreed. Same thing for the community block above, so the whole validation of `can delete` can be extracted.

yqrashawn:
Thanks for your review @ilmotta @jrainville 
All comments are fixed.

ilmotta (approved):

PR #3264, opened 2023-03-07
Update wakuext_buildContact

rasom:
needed for https://github.com/status-im/status-mobile/issues/15152 

status: ready


cammellos (approved):
Looks good, we could also change
`BuildContact` and take a `protocol/request` that might have an optional field for ENS name, I think that's a bit cleaner, but either is ok.
Would you mind adding a test for this please?
Thank you!

protocol/messenger_contacts.go
```diff
@@ -1035,3 +1035,17 @@ func (m *Messenger) BuildContact(contactID string) (*Contact, error) {
 	}
 	return contact, nil
 }
+
+func (m *Messenger) BuildContactWithENSName(contactID, ENSName string) (*Contact, error) {
+	contact, ok := m.allContacts.Load(contactID)
+	if !ok {
+		var err error
+		contact, err = buildContactFromPkString(contactID)
+		contact.ENSVerified = true
```
ilmotta: Sorry to ask, but I couldn't understand if/when we verified the ENS because here I only see the flag being set to true. It looks like `buildContactFromPkString` doesn't do any kind of ENS lookup. So what do we mean by `ENSVerified`?

cammellos: This is the case where the user has added a user through ENS in the client,
say when you add a contact you type `cammellos.eth`, the name is resolved to a pk, and then you want to add the user.
In this case we don't need to re-resolve the ens, since the client has already verified it.
Correct me if I am wrong @rasom

ilmotta: Thanks @cammellos. Now that you explained, I think it makes sense to not re-check. Normally, in centralized solutions where the backend is connected to any number of clients, it would normally re-validate anything done by clients, but since status-go + client is a completely different beast, it seems like re-checking would be a waste of resources.

rasom: > In this case we don't need to re-resolve the ens, since the client has already verified it.

yep that's it

rasom:
@cammellos 
> BuildContact and take a protocol/request that might have an optional field for ENS name

how do we communicate those breaking changes with desktop team? In case they are using this method

ilmotta (approved):

rasom:
@cammellos one tiny test added

cammellos (approved):

PR #3265, opened 2023-03-08
feat: add saved address created at

alaibe:


services/wallet/saved_addresses.go
```diff
@@ -82,8 +83,18 @@ func (sam *SavedAddressesManager) GetRawSavedAddresses() ([]SavedAddress, error)
 }
 
 func (sam *SavedAddressesManager) upsertSavedAddress(sa SavedAddress, tx *sql.Tx) error {
-	sqlStatement := "INSERT OR REPLACE INTO saved_addresses (address, name, favourite, removed, update_clock, chain_short_names, ens_name, is_test) VALUES (?, ?, ?, ?, ?, ?, ?, ?)"
-	var err error
+	savedAddresses, err := sam.GetSavedAddresses()
```
IvanBelyakoff: you could use a helper `getSavedAddresses(fmt.Sprintf("address = %s and ens_name =  %s and is_test = %t",  sa.Address, sa.ENSName, sa.IsTest))` to do SELECT on DB side and avoid manual iteration loop

IvanBelyakoff (approved):

services/wallet/saved_addresses.go
saledjenic: @alaibe if the transaction `tx` has already started (not nil), you will get stuck here. 

Would be great if you could get saved addresses within transaction (otherwise new one may be added/removed) and also would be good to remove `else` block from condition in the line `99` and do insert or replace under transaction only, cause that insertion now depends on `CreatedAt`.

appdatabase/migrations/sql/1678275064_add_created_at_to_saved_addresses.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE saved_addresses ADD COLUMN created_at INT DEFAULT 0;
```
stefandunca: How about using NULL to mark that a value was not set? Then in select, we can use `coalesce` to substitute `NULL` with a default value of `0`

appdatabase/migrations/sql/1678275064_add_created_at_to_saved_addresses.up.sql
```diff
@@ -0,0 +1,2 @@
...
+UPDATE saved_addresses SET created_at = 0;
```
stefandunca: This should not be needed if we have a default value, right?

services/wallet/saved_addresses.go
```diff
@@ -82,8 +83,18 @@ func (sam *SavedAddressesManager) GetRawSavedAddresses() ([]SavedAddress, error)
...
+	if err != nil {
+		return err
+	}
+	sa.CreatedAt = time.Now().Unix()
+	for _, savedAddress := range savedAddresses {
+		if savedAddress.Address == sa.Address && savedAddress.IsTest == sa.IsTest && savedAddress.ENSName == sa.ENSName {
+			sa.CreatedAt = savedAddress.CreatedAt
```
stefandunca, line 114: ~~This approach will also update the `created_at` value for updated addresses. Is this intended? Then the column name should be `updated_at`.~~

~~As others noted, this should be done in SQL.~~
~~The simplest way to do it, for the update case done here, is to change the query with:~~
```sql
INSERT OR REPLACE INTO saved_addresses (address, name, favourite, removed, update_clock, chain_short_names, ens_name, is_test, updated_at) VALUES (?, ?, ?, ?, ?, ?, ?, ?, strftime('%s', 'now'))
```
~~See https://www.sqlite.org/lang_datefunc.html~~

stefandunca (changes requested):
Not sure why

appdatabase/migrations/sql/1678275064_add_created_at_to_saved_addresses.up.sql
saledjenic: @stefandunca I think that the request was to sort the addresses by `created_at` time, not by the time when they were updated last time.

services/wallet/saved_addresses.go
IvanBelyakoff: @stefandunca As far as I see, while all the values will be updated for the row, because this is an `insert or update` called, the actualy value for `created_at` will stay the same, as it is copied from select: `sa.CreatedAt = savedAddress.CreatedAt`
But `created_at` is aimed to be timestamp of creation, to be able to sort the addressess - new at the top.

appdatabase/migrations/sql/1678275064_add_created_at_to_saved_addresses.up.sql
IvanBelyakoff: @stefandunca Could you please explain why we would need to substitute NULL with 0 in select? The result will be 0 at the end, right? And 0 in database for timestamp actually means that the value has not been set.

stefandunca: > @stefandunca I think that the request was to sort the addresses by created_at time, not by the time when they were updated last time.

Thanks for the clarification. The implementation looked to me like it is updating the `created_at` timestamp on upsert (update also not only creation)

> @stefandunca Could you please explain why we would need to substitute NULL with 0 in select? The result will be 0 at the end, right? And 0 in database for timestamp actually means that the value has not been set.

It looked to me that `NULL` might say better that the field was not set (before feature). I agree that 0 is also as good and simpler.

services/wallet/saved_addresses.go
stefandunca: Sorry for the confusion. I meant to say that the entire implementation in this function should be replaced with the SQL query change I mentioned and still get the same result.

appdatabase/migrations/sql/1678275064_add_created_at_to_saved_addresses.up.sql
alaibe: I believe by default the new value is not set to old record

services/wallet/saved_addresses.go
stefandunca: I missed the identity condition up there, thanks @alaibe and @IvanBelyakoff, for clarifying it.

stefandunca (approved):

saledjenic (approved):
@alaibe approved, but please don't forget to update `VERSION` file.

PR #3269, opened 2023-03-08
Add collapsed community categories

cammellos:
adds collapsed commuity categories, to keep track of which categories the user has collapsed in the UI.

I have refactored a bit the messenger test suite, it's just a start, I can finish it off if we are happy with it.


protocol/persistence_collapsed_community_categories.go
```diff
@@ -0,0 +1,38 @@
...
+			&c.CategoryID,
+			&c.Collapsed,
+		)
+		if err != nil {
+			return nil, err
+		}
+		categories = append(categories, c)
+	}
+
+	return categories, nil
```
ilmotta, line 42: For good measure, after looping it's recommended to replace this line with:

```go
if err = rows.Err(); err != nil {
        return categories, err
}
```

protocol/requests/toggle_collapsed_community_category.go
```diff
@@ -0,0 +1,26 @@
+package requests
+
+import (
+	"errors"
+)
+
+var ErrToggleCollapsedCommunityCategoryInvalidCommunityID = errors.New("create-community-category: invalid community id")
+var ErrToggleCollapsedCommunityCategoryInvalidName = errors.New("create-community-category: invalid category name")
```
ilmotta: The error says `create-community-category: ...`, shouldn't it be something like `toggle-collapsed-community-category: invalid category name`? Same thing for the error above it.

services/ext/api.go
```diff
@@ -1264,6 +1264,14 @@ func (api *PublicAPI) UpdateCommunityTokenState(contractAddress string, deploySt
 	return api.service.messenger.UpdateCommunityTokenState(contractAddress, deployState)
 }
 
+func (api *PublicAPI) ToggleCollapsedCommunityCategory(request *requests.ToggleCollapsedCommunityCategory) error {
+	return api.service.messenger.ToggleCollapsedCommunityCategory(request)
+}
+
+func (api *PublicAPI) CollapsedCommunityCategories() ([]protocol.CollapsedCommunityCategory, error) {
+	return api.service.messenger.CollapsedCommunityCategories()
```
ilmotta, line 1272: When I first read `CollapsedCommunityCategories` I thought the endpoint was returning only the records where `Collapsed` is true. Maybe more correct, but also longer, is to refer to these entities as *collapsible*, not *collapsed*, since they can be either one of the two collapsed or non-collapsed states. If you think this is a valid suggestion, then consider renaming the table and all other variables/functions too so that there's no confusion about the state of the record and the name of the entity.

ilmotta (commented):
LGTM @cammellos, I made only minor suggestions.

Next time, could you please attach an issue related to the PR? I think I understand the requirements from the PR description alone, but maybe I could give you a better review if I understood the client's needs in more detail. Who knows, maybe the client had future requirements for collapsible categories + subcategories (just a fake example), so knowing in advance could help me review the PR.

protocol/requests/toggle_collapsed_community_category.go
cammellos: whoops, copy pasted too quickly :)

services/ext/api.go
cammellos: Actually, in theory we don't need to store non-collapsed categories, so you are making a very good point, we could delete them from the database if `collapsed` is `false`, I will change that, and then the endpoint name makes sense, what do you think?

protocol/persistence_collapsed_community_categories.go
cammellos: oh nice, didn't know that, thank you!

cammellos:
> LGTM @cammellos, I made only minor suggestions.
> 
> Next time, could you please attach an issue related to the PR? I think I understand the requirements from the PR description alone, but maybe I could give you a better review if I understood the client's needs in more detail. Who knows, maybe the client had future requirements for collapsible categories + subcategories (just a fake example), so knowing in advance could help me review the PR.

That's true, I will do that, apologies and thanks for taking the time to review the PR!

services/ext/api.go
ilmotta: You're completely right, I missed that we don't need to store them.

protocol/migrations/sqlite/1678292329_add_collapsed_categories.up.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE collapsed_community_categories (
+  community_id VARCHAR NOT NULL,
+  category_id VARCHAR NOT NULL,
+  collapsed INTEGER NOT NULL,
+  UNIQUE(community_id, category_id) ON CONFLICT REPLACE
+);
```
Samyoul: Is the intent of this implementation to prevent pulling message data about categories that the user has collapsed?

Is this the reason why there is a separate table and not just a `collapsed` bool on the categories?

Oh. It is because community categories are not persisted in the DB. Hmmm, that's a bit of an confusing dynamic

Samyoul (approved):
Looks great. Just a little question.

protocol/migrations/sqlite/1678292329_add_collapsed_categories.up.sql
cammellos: The intent is just to record which category in a community the user has collapsed https://github.com/status-im/status-mobile/issues/15290

It's just a UI thing, we show them as closed (i.e not showing the list of chats within a category), or open (we list the chats in this category), similarly to discord, but other than that, it has no impact.

Regarding storage, categories are stored with the community description, and the community description is a field that is set/overriden by the admin of the community (it should not contain any user-set data), so I think it's best to store this separately, but not sure if I addressed your concern, did I understand the question correctly?

Samyoul: Yeah, categories live in the community description, user data related to the display of the categories live in the user DB.

You will always need to read the comm description to check for changes. Although presumably messages record what category they belong to. It is fine.

Samyoul: Concern addressed.

ilmotta (approved):

PR #3271, opened 2023-03-10
Fetch details of user when building a contact

alwx:
Required for https://github.com/status-im/status-mobile/pull/15070
Adds fetching of details when building a contact.

### DIFF
--- a/protocol/messenger_contacts.go
+++ b/protocol/messenger_contacts.go
@@ -1033,5 +1033,20 @@ func (m *Messenger) BuildContact(contactID string) (*Contact, error) {
 			return nil, err
 		}
 	}
+
+	// Schedule sync filter to fetch information about the contact
+	publicKey, err := contact.PublicKey()
+	if err != nil {
+		return nil, err
+	}
+	filter, err := m.transport.JoinPrivate(publicKey)
+	if err != nil {
+		return nil, err
+	}
+	_, err = m.scheduleSyncFilters([]*transport.Filter{filter})
+	if err != nil {
+		return nil, err
+	}
+
 	return contact, nil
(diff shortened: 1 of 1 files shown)


rasom (approved):

cammellos (commented):
It would be good to test this behavior, we could create two contacts, have one publish something on their contact code, and then retrieve this info using build contact from the other side.

Would you mind adding a test for this please?

Thanks!

alwx:
@cammellos I'm struggling a bit with that — the problem is it requires a working mailserver and it means a bunch of changes to the test suite — I tried to make it quick but spend most of the day doing that, actually. How about I create a follow up story for the test? Just because the clojure code needs to be merged asap.

cammellos:
> @cammellos I'm struggling a bit with that — the problem is it requires a working mailserver and it means a bunch of changes to the test suite — I tried to make it quick but spend most of the day doing that, actually. How about I create a follow up story for the test? Just because the clojure code needs to be merged asap.

Of course, no problems, if you want we can have a quick session and we can look at it, I think we might be able to get away without using a mailserver, but we can have a deeper look at it together if you'd like

alwx:
@cammellos sure, we can do that. Just ping me in Status when you're ready — I will push the skeleton of the test I wrote yesterday in the meantime.

cammellos:
> @cammellos sure, we can do that. Just ping me in Status when you're ready — I will push the skeleton of the test I wrote yesterday in the meantime.

cool, in the meantime, feel free to merge when tested/ready, thank you!

alwx:
https://github.com/status-im/status-go/issues/3278

PR #3272, opened 2023-03-10
Add methods to sign and recover messages/signatures to AccountManager

0x-r4bbit:
Also, make AccountManager a dependency of `Messenger`. This is needed for community token permissions as we'll need a way to access wallet accounts and sign messages when sending requests to join a community.

The APIs have been mostly taken from `GethStatusBackend` and `personal` service.

### DIFF
--- a/account/accounts.go
+++ b/account/accounts.go
@@ -1,6 +1,7 @@
 package account
 
 import (
+	"context"
 	"crypto/ecdsa"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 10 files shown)


account/accounts.go
```diff
@@ -44,11 +51,24 @@ func (e ErrCannotLocateKeyFile) Error() string {
 
 var zeroAddress = types.Address{}
 
+type SignParams struct {
+	Data     interface{} `json:"data"`
+	Address  string      `json:"account"`
+	Password string      `json:"password"`
```
MishkaRogachev, line 57: just curious is this a real password or just a password hash?

account/accounts.go
```diff
@@ -612,3 +632,96 @@ func (m *Manager) ReEncryptKeyStoreDir(keyDirPath, oldPass, newPass string) erro
 func (m *Manager) DeleteAccount(address types.Address, password string) error {
 	return m.keystore.Delete(types.Account{Address: address}, password)
 }
+
+func (m *Manager) GetVerifiedWalletAccount(db *accounts.Database, address, password string) (*SelectedExtKey, error) {
```
MishkaRogachev, line 636: address types.Address

account/accounts.go
```diff
@@ -612,3 +632,96 @@ func (m *Manager) ReEncryptKeyStoreDir(keyDirPath, oldPass, newPass string) erro
...
+		return nil, err
+	}
+
+	return &SelectedExtKey{
+		Address:    key.Address,
+		AccountKey: key,
+	}, nil
+}
+
+func (m *Manager) generatePartialAccountKey(db *accounts.Database, address string, password string) (*types.Key, error) {
```
MishkaRogachev, line 664: types.Address?

MishkaRogachev (approved):
Looks good for me!

account/accounts.go
0x-r4bbit: This is a password hash

account/accounts.go
0x-r4bbit: As you see this takes a string. It's exactly the same API that we have in wallet. I copied it over so we have it avaiable in messenger. unfortunately we can't easily depend on wallet in messenger due to how stuff is orchestrated at the moment. I'd rather not touch/change these functions

account/accounts.go
0x-r4bbit: Same here as https://github.com/status-im/status-go/pull/3272/files#r1138423666

PR #3273, opened 2023-03-10
feat(Wallet): implement token list from Uniswap

IvanBelyakoff:
Fixes #8807

Fetched token list from Uniswap, validated against schema, and merged with default token list.

Closes #8807


ghost:
### Pull Request Checklist

services/wallet/token/uniswap_tokenstore.go
```diff
@@ -0,0 +1,59 @@
...
+
+const uniswapTokensURL = "https://gateway.ipfs.io/ipns/tokens.uniswap.org" // nolint:gosec
+const tokenListSchemaURL = "https://uniswap.org/tokenlist.schema.json"     // nolint:gosec
+
+func NewUniswapStore() *uniswapStore {
+	return &uniswapStore{client: &http.Client{Timeout: time.Minute}, tokensFetched: false}
+}
+
+func (ts *uniswapStore) doQuery(url string) (*http.Response, error) {
+	resp, err := ts.client.Get(url)
```
saledjenic: @IvanBelyakoff you could just do `return ts.client.Get(url)` without the part below.

services/wallet/token/tokenstore.go
```diff
@@ -23,1639 +38,1688 @@ func GetTokenPegSymbol(symbol string) string {
...
+				Color:    "#f8f8f8",
+				Decimals: 6,
+				ChainID:  421613,
+			},
+		},
+		tokensFetched: false,
+	}
+}
+
+func (ts *DefaultStore) GetTokens() ([]*Token, error) {
```
saledjenic, line 1718: @IvanBelyakoff why do you return `error` if it's always nil?

services/wallet/token/token.go
```diff
@@ -107,6 +114,31 @@ func (tm *Manager) inStore(address common.Address, chainID uint64) bool {
...
+
+	return true
+}
+
+func (tm *Manager) fetchTokens() []*Token {
+	result := make([]*Token, 0)
+	for _, store := range tm.stores {
+		tokens, err := store.GetTokens()
+		if err != nil {
+			log.Error("can't fetch tokens from store: %s", err)
```
saledjenic, line 132: @IvanBelyakoff you're not fetching them, the list of tokens is hardcoded, according to my previous comment you don't need this `if err != nil {` check.

IvanBelyakoff: @saledjenic tokens are not fetched only for DefaultStore, where they are hardcoded. But for UniswapStore they are fetched for the first time and it is possible to get error here. Since I access the stores via interface, I check for errors

services/wallet/token/uniswap_tokenstore.go
IvanBelyakoff: Good catch, indeed!

services/wallet/token/tokenstore.go
IvanBelyakoff: Because DefaultStore implements `Store` interface. `UniswapStore` implements it as well, both are accessed from `Token` via interface methods. Anthony said we most likely will add more token lists later

saledjenic (approved):

PR #3274, opened 2023-03-13
fix(wallet/coingecko): Fix coingecko crash on race writing to tokens map

IvanBelyakoff:
Fix crash in coingecko client on concurrent writes to token map

Crash happened in `getTokens()` method which is called from multiple places via coroutines.

```
fatal error: concurrent map writes

goroutine 1391 [running]:
github.com/status-im/status-go/services/wallet/thirdparty/coingecko.(*Client).getTokens(0x1c002727470)
    /home/blade/work/projects/Status/status-desktop/vendor/status-go/services/wallet/thirdparty/coingecko/client.go:128 +0x3e5
github.com/status-im/status-go/services/wallet/thirdparty/coingecko.(*Client).mapSymbolsToIds(0x7ffa911fe0a0?, {0x1c006d4f800, 0x172, 0x1c005079f50?})
    /home/blade/work/projects/Status/status-desktop/vendor/status-go/services/wallet/thirdparty/coingecko/client.go:135 +0x3e
github.com/status-im/status-go/services/wallet/thirdparty/coingecko.(*Client).FetchPrices(0x1c002727470, {0x1c006d4f800?, 0x172, 0x172?}, {0x1c006cef6a0?, 0x1, 0x1})
    /home/blade/work/projects/Status/status-desktop/vendor/status-go/services/wallet/thirdparty/coingecko/client.go:158 +0x94
github.com/status-im/status-go/services/wallet/market.(*Manager).FetchPrices.func2()
    /home/blade/work/projects/Status/status-desktop/vendor/status-go/services/wallet/market/market.go:162 +0x3e
github.com/status-im/status-go/services/wallet/market.(*Manager).makeCall.func2({0x7ffa912f9860?, 0x1c00064ea60?})
    /home/blade/work/projects/Status/status-desktop/vendor/status-go/services/wallet/market/market.go:61 +0x4d
github.com/afex/hystrix-go/hystrix.Go.func2({0x7ffa8fb03696?, 0x7ffa90eee720?}, {0x7ffa912f9860?, 0x1c00064ea60?})
    /home/blade/work/projects/Status/status-desktop/vendor/status-go/vendor/github.com/afex/hystrix-go/hystrix/hystrix.go:62 +0x2d
github.com/afex/hystrix-go/hystrix.(*command).tryFallback(0x1c006d567e0, {0x7ffa91308c78?, 0x1c0000da038?}, {0x7ffa912f9860?, 0x1c00064ea60?})
    /home/blade/work/projects/Status/status-desktop/vendor/status-go/vendor/github.com/afex/hystrix-go/hystrix/hystrix.go:290 +0x6a
github.com/afex/hystrix-go/hystrix.(*command).errorWithFallback(0x1c006d567e0, {0x7ffa91308c78, 0x1c0000da038}, {0x7ffa912f9860?, 0x1c00064ea60})
    /home/blade/work/projects/Status/status-desktop/vendor/status-go/vendor/github.com/afex/hystrix-go/hystrix/hystrix.go:278 +0x305
github.com/afex/hystrix-go/hystrix.GoC.func4.2()
    /home/blade/work/projects/Status/status-desktop/vendor/status-go/vendor/github.com/afex/hystrix-go/hystrix/hystrix.go:190 +0x85
sync.(*Once).doSlow(0x1c000732f98?, 0x1c000732ee4?)
    /usr/local/go/src/sync/once.go:74 +0xc2
sync.(*Once).Do(...)
    /usr/local/go/src/sync/once.go:65
github.com/afex/hystrix-go/hystrix.GoC.func4()
    /home/blade/work/projects/Status/status-desktop/vendor/status-go/vendor/github.com/afex/hystrix-go/hystrix/hystrix.go:188 +0x1d8
created by github.com/afex/hystrix-go/hystrix.GoC
    /home/blade/work/projects/Status/status-desktop/vendor/status-go/vendor/github.com/afex/hystrix-go/hystrix/hystrix.go:173 +0x4d8

```

```
goroutine 1930 [runnable]:
strings.ToUpper({0x1c0016bc870, 0x3})
    /usr/local/go/src/strings/strings.go:559 +0x25a
github.com/status-im/status-go/services/wallet/thirdparty/coingecko.(*Client).getTokens(0x1c002727470)
    /home/blade/work/projects/Status/status-desktop/vendor/status-go/services/wallet/thirdparty/coingecko/client.go:123 +0x35b
github.com/status-im/status-go/services/wallet/thirdparty/coingecko.(*Client).FetchTokenDetails(0x1c002727470?, {0x1c007288000, 0x172, 0x200?})
    /home/blade/work/projects/Status/status-desktop/vendor/status-go/services/wallet/thirdparty/coingecko/client.go:196 +0x3e
github.com/status-im/status-go/services/wallet/market.(*Manager).FetchTokenDetails.func2()
    /home/blade/work/projects/Status/status-desktop/vendor/status-go/services/wallet/market/market.go:133 +0x32
github.com/status-im/status-go/services/wallet/market.(*Manager).makeCall.func2({0x7ffa912f9860?, 0x1c004e1ba70?})
    /home/blade/work/projects/Status/status-desktop/vendor/status-go/services/wallet/market/market.go:61 +0x4d
github.com/afex/hystrix-go/hystrix.Go.func2({0x7ffa8fb03696?, 0x7ffa90eee720?}, {0x7ffa912f9860?, 0x1c004e1ba70?})
    /home/blade/work/projects/Status/status-desktop/vendor/status-go/vendor/github.com/afex/hystrix-go/hystrix/hystrix.go:62 +0x2d
github.com/afex/hystrix-go/hystrix.(*command).tryFallback(0x1c0066b1f80, {0x7ffa91308c78?, 0x1c0000da038?}, {0x7ffa912f9860?, 0x1c004e1ba70?})
    /home/blade/work/projects/Status/status-desktop/vendor/status-go/vendor/github.com/afex/hystrix-go/hystrix/hystrix.go:290 +0x6a
github.com/afex/hystrix-go/hystrix.(*command).errorWithFallback(0x1c0066b1f80, {0x7ffa91308c78, 0x1c0000da038}, {0x7ffa912f9860?, 0x1c004e1ba70})
    /home/blade/work/projects/Status/status-desktop/vendor/status-go/vendor/github.com/afex/hystrix-go/hystrix/hystrix.go:278 +0x305
github.com/afex/hystrix-go/hystrix.GoC.func4.2()
    /home/blade/work/projects/Status/status-desktop/vendor/status-go/vendor/github.com/afex/hystrix-go/hystrix/hystrix.go:190 +0x85
sync.(*Once).doSlow(0x1c0007b4f98?, 0x1c0007b4ee4?)
    /usr/local/go/src/sync/once.go:74 +0xc2

```

Refactored coingecko client to make it testable.
Implemented a test to reproduce the crash, crash fix is confirmed with test.

### DIFF
--- a/services/wallet/thirdparty/coingecko/client.go
+++ b/services/wallet/thirdparty/coingecko/client.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/status-im/status-go/services/wallet/thirdparty"
@@ -79,12 +80,14 @@ type GeckoToken struct {
 }
 
 type Client struct {
-	client *http.Client
-	tokens map[string]GeckoToken
+	client           *http.Client
+	tokens           map[string]GeckoToken
+	tokensURL        string
+	fetchTokensMutex sync.Mutex
 }
 
 func NewClient() *Client {
-	return &Client{client: &http.Client{Timeout: time.Minute}, tokens: make(map[string]GeckoToken)}
+	return &Client{client: &http.Client{Timeout: time.Minute}, tokens: make(map[string]GeckoToken), tokensURL: fmt.Sprintf("%scoins/list", baseURL)}
 }
 
 func (c *Client) DoQuery(url string) (*http.Response, error) {
@@ -96,13 +99,27 @@ func (c *Client) DoQuery(url string) (*http.Response, error) {
 	return resp, nil
 }
 
+func mapTokensToSymbols(tokens []GeckoToken, tokenMap map[string]GeckoToken) {
+	for _, token := range tokens {
+		if id, ok := coinGeckoMapping[strings.ToUpper(token.Symbol)]; ok {
+			if id != token.ID {
+				continue
+			}
+		}
+		tokenMap[strings.ToUpper(token.Symbol)] = token
+	}
+}
+
 func (c *Client) getTokens() (map[string]GeckoToken, error) {
+
+	c.fetchTokensMutex.Lock()
+	defer c.fetchTokensMutex.Unlock()
+
 	if len(c.tokens) > 0 {
 		return c.tokens, nil
 	}
 
-	url := fmt.Sprintf("%scoins/list", baseURL)
-	resp, err := c.DoQuery(url)
+	resp, err := c.DoQuery(c.tokensURL)
 	if err != nil {
 		return nil, err
 	}
@@ -119,14 +136,7 @@ func (c *Client) getTokens() (map[string]GeckoToken, error) {
 		return nil, err
 	}
 
-	for _, token := range tokens {
-		if id, ok := coinGeckoMapping[strings.ToUpper(token.Symbol)]; ok {
-			if id != token.ID {
-				continue
-			}
-		}
-		c.tokens[strings.ToUpper(token.Symbol)] = token
-	}
+	mapTokensToSymbols(tokens, c.tokens)
 
 	return c.tokens, nil
 }
--- a/services/wallet/thirdparty/coingecko/client_test.go
+++ b/services/wallet/thirdparty/coingecko/client_test.go
@@ -1,0 +1,82 @@
+package coingecko
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"reflect"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func setupTest(t *testing.T, response []byte) (*httptest.Server, func()) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(200)
+		_, err := w.Write(response)
+		if err != nil {
+			return
+		}
+	}))
+
+	return srv, func() {
+		srv.Close()
+	}
+}
+
+func TestGetTokensSuccess(t *testing.T) {
+	expected := []GeckoToken{
+		{
+			ID:     "ethereum",
+			Symbol: "eth",
+			Name:   "Ethereum",
+		},
+		{
+			ID:     "status",
+			Symbol: "snt",
+			Name:   "Status",
+		},
+	}
+
+	expectedMap := map[string]GeckoToken{
+		"ETH": {
+			ID:     "ethereum",
+			Symbol: "eth",
+			Name:   "Ethereum",
+		},
+		"SNT": {
+			ID:     "status",
+			Symbol: "snt",
+			Name:   "Status",
+		},
+	}
+	response, _ := json.Marshal(expected)
+
+	srv, stop := setupTest(t, response)
+	defer stop()
+
+	geckoClient := &Client{
+		client:    srv.Client(),
+		tokens:    make(map[string]GeckoToken),
+		tokensURL: srv.URL,
+	}
+
+	tokenMap, err := geckoClient.getTokens()
+	require.NoError(t, err)
+	require.True(t, reflect.DeepEqual(expectedMap, tokenMap))
+}
+
+func TestGetTokensFailure(t *testing.T) {
+	resp := []byte{}
+	srv, stop := setupTest(t, resp)
+	defer stop()
+
+	geckoClient := &Client{
+		client:    srv.Client(),
+		tokens:    make(map[string]GeckoToken),
+		tokensURL: srv.URL,
+	}
+
+	_, err := geckoClient.getTokens()
+	require.Error(t, err)
+}


services/wallet/thirdparty/coingecko/client.go
```diff
@@ -96,13 +101,35 @@ func (c *Client) DoQuery(url string) (*http.Response, error) {
...
+		}
+		tokenMap[strings.ToUpper(token.Symbol)] = token
+	}
+}
+
 func (c *Client) getTokens() (map[string]GeckoToken, error) {
+	fetching := (&c.isTokenFetchPending).Load()
+
+	if fetching {
+		c.fetchTokensMutex.Lock()
```
saledjenic: @IvanBelyakoff I am not sure that I understand your intention here? 
I mean... you have locking here and in the line 127 again? Is that what you wanted?

IvanBelyakoff: Yes, the idea is we have to:
1) know if fetch has already been called and in this case we don't want to return from current thread, we just want to wait until the previous fetch called from another thread will finish. So I add one lock for this case - all subsequent calls will block here and when fetching is done, and mutex will be unlocked, they will just check if  tokens list is not empty, and most likely return the tokens without fetching.
2) Block all subsequent calls to `getTokens` at line 119 - while fetching is done - this is why I lock the mutex at line 127. I also need to handle the worst case, when tokens length is 0 after fetch (due to some error, internet down, etc) and then multiple threads will  proceed and call `doQuery` simultaneously and then will try to concurrently update the map, so this lock at 127 will protect from it, but won't prevent re-fetching in the worst case scenario, but that will be safe in my understanding

IvanBelyakoff: Not sure if I exaplined it well. The second lock (line 127) is needed to block subsequent calls at line 119, and also prevent concurrent fetching and update of the map in the worst case

dlipicar: hmm also not liking how the mutex is managed.

services/wallet/thirdparty/coingecko/client.go
```diff
@@ -96,13 +101,35 @@ func (c *Client) DoQuery(url string) (*http.Response, error) {
...
+		defer c.fetchTokensMutex.Unlock()
+	}
+
 	if len(c.tokens) > 0 {
 		return c.tokens, nil
 	}
 
-	url := fmt.Sprintf("%scoins/list", baseURL)
-	resp, err := c.DoQuery(url)
+	c.fetchTokensMutex.Lock()
```
dlipicar: There's something I don't understand here. If you enter the `fetching` condition above, AFAIK the `defer`ed `Unlock()` will be called at the end of the function. If for whatever reason `len(c.tokens) == 0` and you call `Lock()` here again, won't that be a deadlock?

services/wallet/thirdparty/coingecko/client.go
dlipicar: I don't think you gain much by using the atomic + the mutex, and it makes the logic more complex. In any case, it sounds like premature optimization.
In the end, it seems you're treating both resources (the tokens map and the remote endpoint) as a single entity you want to protect from concurrent access, so simply locking the mutex at the top should do the trick IMO.

services/wallet/thirdparty/coingecko/client_test.go
```diff
@@ -0,0 +1,98 @@
+package coingecko
```
dlipicar, line 1: 👏 👏 👏

services/wallet/thirdparty/coingecko/client.go
IvanBelyakoff: Ok, @saledjenic @dlipicar I believe you are right. Thank you for the feedback. I fixed that

dlipicar (approved):

services/wallet/thirdparty/coingecko/client_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func setupTest(t *testing.T, delay time.Duration, response []byte) (*httptest.Server, func()) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		time.Sleep(delay * time.Second)
```
stefandunca: I think `delaySecs` should be `int` if you want to use it this way. The naming means a  count of seconds, but I see here `time.Duration`, which is unit-of-time agnostic.

The best is to rename `delaySecs` to `delay` and keep the `time.Duration`, pass it as such and later on, where you use it, pass `1 * time.Second`

services/wallet/thirdparty/coingecko/client_test.go
```diff
@@ -0,0 +1,98 @@
...
+		},
+		"SNT": {
+			ID:     "status",
+			Symbol: "snt",
+			Name:   "Status",
+		},
+	}
+	response, _ := json.Marshal(expected)
+
+	var delay time.Duration = 1
```
stefandunca: It looks to me that `delaysSecs` is one nanosecond. See here https://pkg.go.dev/time#Duration

If you want one second, you have to fix the `setupTest` as I mentioned above and have this changed to
```go
delaySecs := time.Duration(1)*time.Second
```

stefandunca (approved):

services/wallet/thirdparty/coingecko/client_test.go
IvanBelyakoff: @stefandunca good point. Fixed

services/wallet/thirdparty/coingecko/client_test.go
IvanBelyakoff: you are right. thanx! fixed

IvanBelyakoff:
After conversation with @stefandunca it was decided to drop race-related lines from the PR tests, because that is not related to unit testing but to dev testing, that the issue has been fixed, but no need to run race test on CI every time. So the tests just test `getTokens` success and failure, without race condition

stefandunca (approved):

dlipicar (approved):

PR #3275, opened 2023-03-13
Add spiff workflow code

cammellos:
This commit adds spiff worflow code, it's a node that needs some extra requirements.

The code is a bit rough, but until requirements are finalized, I don't think it's worth to polish up the code.

It also adds support for both compressed keys/non compressed in some endpoints, ideally we do the same for the rest so at some point we can clean swap them.


ilmotta (approved):
I'm so biased to approve this PR. My Emacs says a big thanks too. Thanks @cammellos

image

ilmotta:
> It also adds support for both compressed keys/non compressed in some endpoints, ideally we do the same for the rest so at some point we can clean swap them.

Curious, is the idea to eventually not even return public keys in the endpoints' responses? Effectively keep them almost like an implementation detail and have only compressed keys ever exposed to clients?

cammellos:
> 

Yes, ideally ids are treated as ids and not public keys, so that we can just push out what we would like, and the client just uses what we give them, so there's no special meaning attached to them

PR #3276, opened 2023-03-13
feat: dynamic social links

caybro:
- completely replace social links on save
- respect the order of items and also the URL when comparing

Needed for https://github.com/status-im/status-desktop/issues/9777

Rationale: for MVP, we'll want the user to be able to add several links of the same type, and adjust/preserve their order by drag'n'drop

### DIFF
--- a/multiaccounts/settings_social_links/database.go
+++ b/multiaccounts/settings_social_links/database.go
@@ -47,7 +47,7 @@ func (s *SocialLinksSettings) GetSocialLinks() (identity.SocialLinks, error) {
 	return links, nil
 }
 
-// Be careful, it removes every row except static links (__twitter, etc.) before insertion
+// Be careful, it removes every row before insertion
 func (s *SocialLinksSettings) SetSocialLinks(links *identity.SocialLinks) (err error) {
 	if links == nil {
 		return errors.New("can't set social links, nil object provided")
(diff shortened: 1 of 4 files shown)


ghost:
### Pull Request Checklist

osmaczko (changes requested):
Static links were introduced to have one source of truth for all clients in terms of `link_text` values of supported socials. Each user's database already contains [preloaded](https://github.com/status-im/status-go/blob/develop/appdatabase/migrations/sql/1660134042_add_social_links_settings_table.up.sql) settings placeholders.

To make your change complementary you would need to introduce a migration that removes these from database.

Perhaps it would be more convenient to filter out empty static social links on the client side by utilizing SortFilterProxy, wouldn't it?

caybro:
> Static links were introduced to have one source of truth for all clients in terms of `link_text` values of supported socials. Each user's database already contains [preloaded](https://github.com/status-im/status-go/blob/develop/appdatabase/migrations/sql/1660134042_add_social_links_settings_table.up.sql) settings placeholders.
> 
> To make your change complementary you would need to introduce a migration that removes these from database.

It's not needed, it happens already: when the links get saved (triggered by the frontend), the empty values are removed (see https://github.com/status-im/status-go/pull/3276/files#diff-65320c66040ca2e9a00dd6f9d497c2984cfce150c29368b9e277d4960531eb0bR70)

> Perhaps it would be more convenient to filter out empty static social links on the client side by utilizing SortFilterProxy, wouldn't it?

Yes, existing empty ones are hidden in QML (not via SortFilterProxyModel as that one wouldn't work with the drag'n'drop)

osmaczko:
> It's not needed, it happens already: when the links get saved (triggered by the frontend), the empty values are removed (see https://github.com/status-im/status-go/pull/3276/files#diff-65320c66040ca2e9a00dd6f9d497c2984cfce150c29368b9e277d4960531eb0bR70)

What will happen if a new account is created and no links are saved yet?

> Yes, existing empty ones are hidden in QML (not via SortFilterProxyModel as that one wouldn't work with the drag'n'drop)

Hm.. I don't get a point of this PR then :thinking: Why do you need to remove empty static links from the database if they are not shown on the client side?

caybro:
> > It's not needed, it happens already: when the links get saved (triggered by the frontend), the empty values are removed (see https://github.com/status-im/status-go/pull/3276/files#diff-65320c66040ca2e9a00dd6f9d497c2984cfce150c29368b9e277d4960531eb0bR70)
> 
> What will happen if a new account is created and no links are saved yet?
 
No links should be created or displayed, just this empty placeholder:
image

> Hm.. I don't get a point of this PR

The other purpose of this PR is to fully save the links sent from the frontend

osmaczko (approved):
Now I got it, the order of static links does matter :facepalm:  Looks good, just one test to fix :crossed_fingers:

PR #3277, opened 2023-03-13
fix(activity_center): Add remove activity center notifications array to message response

borismelnik:
in fact of this [issue](https://github.com/status-im/status-desktop/issues/9751) I add an array to provide event for deleting notifications from activity center

Important changes:

Closes #

### DIFF
--- a/protocol/messenger_delete_message_for_everyone_test.go
+++ b/protocol/messenger_delete_message_for_everyone_test.go
@@ -107,11 +107,15 @@ func (s *MessengerDeleteMessageForEveryoneSuite) TestDeleteMessageForEveryone() 
 	deleteMessageResponse, err := s.moderator.DeleteMessageAndSend(ctx, message.ID)
 	s.Require().NoError(err)
 
-	_, err = WaitOnMessengerResponse(s.bob, func(response *MessengerResponse) bool {
+	response, err = WaitOnMessengerResponse(s.bob, func(response *MessengerResponse) bool {
 		return len(response.RemovedMessages()) > 0
 	}, "removed messages not received")
 	s.Require().Equal(deleteMessageResponse.RemovedMessages()[0].DeletedBy, contactIDFromPublicKey(s.moderator.IdentityPublicKey()))
+
 	s.Require().NoError(err)
+	s.Require().Len(response.ActivityCenterNotifications(), 1)
+	s.Require().True(response.ActivityCenterNotifications()[0].Deleted)
+
 	message, err = s.bob.MessageByID(message.ID)
 	s.Require().NoError(err)
 	s.Require().True(message.Deleted)
--- a/protocol/messenger_handler.go
+++ b/protocol/messenger_handler.go
@@ -1583,6 +1583,10 @@ func (m *Messenger) HandleDeleteMessage(state *ReceivedMessageState, deleteMessa
 	state.Response.AddRemovedMessage(&RemovedMessage{MessageID: messageID, ChatID: chat.ID, DeletedBy: deleteMessage.DeleteMessage.DeletedBy})
 	state.Response.AddChat(chat)
 	state.Response.AddNotification(DeletedMessageNotification(messageID, chat))
+	state.Response.AddActivityCenterNotification(&ActivityCenterNotification{
+		ID:      types.FromHex(messageID),
+		Deleted: true,
+	})
 
 	return nil
 }


cammellos:
@ilmotta @MishkaRogachev do we use `deleted` or `dismissed` for this already? or this calls for a different behavior?

ilmotta:
> @ilmotta @MishkaRogachev do we use `deleted` or `dismissed` for this already? or this calls for a different behavior?

@cammellos: I'm not sure I understand the question exactly. What do you mean by `this` in the sentence?

Looking at the original issue https://github.com/status-im/status-desktop/issues/9751, it looks like Mobile probably has the same bug, but I've never checked and no QA reported yet.

protocol/messenger_response.go
```diff
@@ -489,6 +502,20 @@ func (r *MessengerResponse) AddRemovedChat(chatID string) {
 	r.removedChats[chatID] = true
 }
 
+func (r *MessengerResponse) AddRemovedActivityCenterNotifications(notifications []string) {
+	for _, n := range notifications {
+		r.AddRemoveActivityCenterNotification(n)
+	}
+}
+
+func (r *MessengerResponse) AddRemoveActivityCenterNotification(notificationID string) {
```
ilmotta: Tiny typo, should probably be `AddRemovedActivityCenterNotification` (missing `d`)

cammellos:
> > @ilmotta @MishkaRogachev do we use `deleted` or `dismissed` for this already? or this calls for a different behavior?
> 
> @cammellos: I'm not sure I understand the question exactly. What do you mean by `this` in the sentence?
> 
> Looking at the original issue [status-im/status-desktop#9751](https://github.com/status-im/status-desktop/issues/9751), it looks like Mobile probably has the same bug, but I've never checked and no QA reported yet.

Basically, looking at the code in status react, if a notification has a `deleted` flag, it's removed from the frontend.
This PR introduces a similar behavior, to remove/delete notifications from the frontend, should we have 2 different ways or just a single one? (though they are slightly different operations here, since "deleted" is coming from the user, while this comes from the backend)

cammellos:
> > > @ilmotta @MishkaRogachev do we use `deleted` or `dismissed` for this already? or this calls for a different behavior?
> > 
> > 
> > @cammellos: I'm not sure I understand the question exactly. What do you mean by `this` in the sentence?
> > Looking at the original issue [status-im/status-desktop#9751](https://github.com/status-im/status-desktop/issues/9751), it looks like Mobile probably has the same bug, but I've never checked and no QA reported yet.
> 
> Basically, looking at the code in status react, if a notification has a `deleted` flag, it's removed from the frontend. This PR introduces a similar behavior, to remove/delete notifications from the frontend, should we have 2 different ways or just a single one? (though they are slightly different operations here, since "deleted" is coming from the user, while this comes from the backend)

Looking at the code, this calls:
```
{
	var tx *sql.Tx
	var err error

	tx, err = db.db.BeginTx(context.Background(), &sql.TxOptions{})
	if err != nil {
		return err
	}
	defer func() {
		if err == nil {
			err = tx.Commit()
			return
		}
		// don't shadow original error
		_ = tx.Rollback()
	}()

	params := activityCenterQueryParams{
		chatID: chatID,
	}

	_, notifications, err := db.buildActivityCenterQuery(tx, params)

	if err != nil {
		return err
	}

	var ids []types.HexBytes

		if notification.LastMessage != nil && notification.LastMessage.ID == messageID {
			ids = append(ids, notification.ID)
		}

		if notification.Message != nil && notification.Message.ID == messageID {
			ids = append(ids, notification.ID)
		}
	}

	if len(ids) > 0 {
		idsArgs := make([]interface{}, 0, len(ids))
		for _, id := range ids {
			idsArgs = append(idsArgs, id)
		}

		inVector := strings.Repeat("?, ", len(ids)-1) + "?"
		query := "UPDATE activity_center_notifications SET read = 1, dismissed = 1, deleted = 1 WHERE id IN (" + inVector + ")" // nolint: gosec
		_, err = tx.Exec(query, idsArgs...)
		return err
	}

	return nil
}
```
Which is soft-deleting them, so looks like just setting `deleted`/`dismissed` to true on the notification message, should be enough to communicate to the frontend to remove the notification from the activity center (and should require, correct me if I am wrong, no changes on mobile)

ilmotta:
> > > @ilmotta @MishkaRogachev do we use `deleted` or `dismissed` for this already? or this calls for a different behavior?
> > 
> > 
> > @cammellos: I'm not sure I understand the question exactly. What do you mean by `this` in the sentence?
> > Looking at the original issue [status-im/status-desktop#9751](https://github.com/status-im/status-desktop/issues/9751), it looks like Mobile probably has the same bug, but I've never checked and no QA reported yet.
> 
> Basically, looking at the code in status react, if a notification has a `deleted` flag, it's removed from the frontend. This PR introduces a similar behavior, to remove/delete notifications from the frontend, should we have 2 different ways or just a single one? (though they are slightly different operations here, since "deleted" is coming from the user, while this comes from the backend)

Okay, I see now. I agree with you, we should ideally have one reliable solution for all clients.

And that's correct, any notification with the `deleted` flag as true is removed from the app state in `status-mobile`.

I've never checked if the problem reported by https://github.com/status-im/status-desktop/issues/9751 actually happens in Mobile. I can double-check.

cammellos:
> > > > @ilmotta @MishkaRogachev do we use `deleted` or `dismissed` for this already? or this calls for a different behavior?
> > > 
> > > 
> > > @cammellos: I'm not sure I understand the question exactly. What do you mean by `this` in the sentence?
> > > Looking at the original issue [status-im/status-desktop#9751](https://github.com/status-im/status-desktop/issues/9751), it looks like Mobile probably has the same bug, but I've never checked and no QA reported yet.
> > 
> > 
> > Basically, looking at the code in status react, if a notification has a `deleted` flag, it's removed from the frontend. This PR introduces a similar behavior, to remove/delete notifications from the frontend, should we have 2 different ways or just a single one? (though they are slightly different operations here, since "deleted" is coming from the user, while this comes from the backend)
> 
> Okay, I see now. I agree with you, we should ideally have one reliable solution for all clients.
> 
> And that's correct, any notification with the `deleted` flag as true is removed from the app state in `status-mobile`.
> 
> I've never checked if the problem reported by [status-im/status-desktop#9751](https://github.com/status-im/status-desktop/issues/9751) actually happens in Mobile. I can double-check.

It definitely happens, but basically this code could just return the notifications with the `deleted` flag set, instead of creating a new entry in `MessengerResponse`, basically here:

https://github.com/status-im/status-go/pull/3277/files#diff-59d196f7f72de56c64ba2c559fcbb51762c418a22d2b83c59662b772a09b3c0aR1586

We could do something like:

```
notification = ActivityCenterNotification{id: id, Deleted: true, Dismissed: true} // or we could pull it from db
response.AddActivityCenterNotification(notification)
```

and that should be enough to fix the issue I think

ilmotta:
> > > > > @ilmotta @MishkaRogachev do we use `deleted` or `dismissed` for this already? or this calls for a different behavior?
> > > > 
> > > > 
> > > > @cammellos: I'm not sure I understand the question exactly. What do you mean by `this` in the sentence?
> > > > Looking at the original issue [status-im/status-desktop#9751](https://github.com/status-im/status-desktop/issues/9751), it looks like Mobile probably has the same bug, but I've never checked and no QA reported yet.
> > > 
> > > 
> > > Basically, looking at the code in status react, if a notification has a `deleted` flag, it's removed from the frontend. This PR introduces a similar behavior, to remove/delete notifications from the frontend, should we have 2 different ways or just a single one? (though they are slightly different operations here, since "deleted" is coming from the user, while this comes from the backend)
> > 
> > 
> > Okay, I see now. I agree with you, we should ideally have one reliable solution for all clients.
> > And that's correct, any notification with the `deleted` flag as true is removed from the app state in `status-mobile`.
> > I've never checked if the problem reported by [status-im/status-desktop#9751](https://github.com/status-im/status-desktop/issues/9751) actually happens in Mobile. I can double-check.
> 
> It definitely happens, but basically this code could just return the notifications with the `deleted` flag set, instead of creating a new entry in `MessengerResponse`, basically here:
> 
> https://github.com/status-im/status-go/pull/3277/files#diff-59d196f7f72de56c64ba2c559fcbb51762c418a22d2b83c59662b772a09b3c0aR1586
> 
> We could do something like:
> 
> ```
> notification = ActivityCenterNotification{id: id, Deleted: true, Dismissed: true} // or we could pull it from db
> response.AddActivityCenterNotification(notification)
> ```
> 
> and that should be enough to fix the issue I think

:+1: Absolutely. Any notification that arrives with the `deleted` flag set will cause the reconciliation to remove it from the app state.

cammellos (commented):
@borismelnik so after the discussion with @ilmotta , I think the best implementation is to just return the notification in `ActivityCenterNotification` with `Deleted` set to true.

I think you can just return an empty notification with only `ID` and `Deleted` set, since the client should just remove it or not add it.

Please @ilmotta / @MishkaRogachev check if that's ok with you

Thanks!

ilmotta:
> @borismelnik so after the discussion with @ilmotta , I think the best implementation is to just return the notification in `ActivityCenterNotification` with `Deleted` set to true.
> 
> I think you can just return an empty notification with only `ID` and `Deleted` set, since the client should just remove it or not add it.
> 
> Please @ilmotta / @MishkaRogachev check if that's ok with you
> 
> Thanks!

Thanks @cammellos. Looks like it should work without any modification on `status-mobile` as you said.

@borismelnik, once you have implemented all the changes, let me know and I can test the mobile app pointing to this branch.

MishkaRogachev:
> @borismelnik so after the discussion with @ilmotta , I think the best implementation is to just return the notification in `ActivityCenterNotification` with `Deleted` set to true.
> 
> I think you can just return an empty notification with only `ID` and `Deleted` set, since the client should just remove it or not add it.
> 
> Please @ilmotta / @MishkaRogachev check if that's ok with you
> 
> Thanks!

From my perspective solution returning empty notification with ID & Deleted flag should work good, but it requires some extra logic on the ui side (on desktop).

From other side, have same logic for removing messages and notifications make sense

cammellos:
> > @borismelnik so after the discussion with @ilmotta , I think the best implementation is to just return the notification in `ActivityCenterNotification` with `Deleted` set to true.
> > I think you can just return an empty notification with only `ID` and `Deleted` set, since the client should just remove it or not add it.
> > Please @ilmotta / @MishkaRogachev check if that's ok with you
> > Thanks!
> 
> From my perspective solution returning empty notification with ID & Deleted flag should work good, but it requires some extra logic on the ui side (on desktop).
> 
> From other side, have same logic for removing messages and notifications make sense

Cool, thanks for confirming @MishkaRogachev , this is where we want to have the same behavior between clients, so it's good that we align!

borismelnik:
@cammellos @ilmotta @MishkaRogachev 
I updated the PR and hope I understand you correctly :)

cammellos (commented):
Looks good, could you add/extend a test please covering this case?

Thank you!

MishkaRogachev (approved):
LGTM

ilmotta:
@borismelnik, LGTM. Tested on mobile:

1.  Create group chat with users `A` and `B` as mutual contacts.
2. `A` mentions `B`.
3. `B` sees a mention notification.
4. `A` deletes the message for everyone.
5. `B` sees the notification disappear and the unread counters are updated accordingly.

ilmotta (approved):

cammellos (approved):

PR #3287, opened 2023-03-14
feat(images): Add albumId to images album and allow to receive image messages without a text

borismelnik:
The reason of this changes was that [issue](https://github.com/status-im/status-desktop/issues/6374).
This changes includes:
- Adding `AlbumId` to messages which sends like images album
- Skip text validation for image messages with empty `Text` field

This may require to skip usage of `<p>placeholder</p>` text from mobile app

Important changes:

Closes #

### DIFF
--- a/protocol/message_validator.go
+++ b/protocol/message_validator.go
@@ -233,7 +233,7 @@ func ValidateReceivedChatMessage(message *protobuf.ChatMessage, whisperTimestamp
 		return errors.New("timestamp can't be 0")
 	}
 
-	if message.ContentType != protobuf.ChatMessage_DISCORD_MESSAGE {
+	if message.ContentType != protobuf.ChatMessage_DISCORD_MESSAGE && (message.ContentType != protobuf.ChatMessage_IMAGE || message.Text != "") {
 		if err := ValidateText(message.Text); err != nil {
 			return err
 		}
--- a/protocol/messenger.go
+++ b/protocol/messenger.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/davecgh/go-spew/spew"
 	"github.com/golang/protobuf/proto"
+	"github.com/google/uuid"
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
 
@@ -1968,7 +1969,18 @@ func (m *Messenger) SendChatMessage(ctx context.Context, message *common.Message
 func (m *Messenger) SendChatMessages(ctx context.Context, messages []*common.Message) (*MessengerResponse, error) {
 	var response MessengerResponse
 
+	generatedAlbumID, err := uuid.NewRandom()
+	if err != nil {
+		return nil, err
+	}
+
 	for _, message := range messages {
+		if message.ContentType == protobuf.ChatMessage_IMAGE && len(messages) > 1 {
+			err = message.SetAlbumID(generatedAlbumID.String())
+			if err != nil {
+				return nil, err
+			}
+		}
 		messageResponse, err := m.SendChatMessage(ctx, message)
 		if err != nil {
 			return nil, err
@@ -1996,6 +2008,7 @@ func (m *Messenger) sendChatMessage(ctx context.Context, message *common.Message
 		if err != nil {
 			return nil, err
 		}
+
 	} else if len(message.CommunityID) != 0 {
 		community, err := m.communitiesManager.GetByIDString(message.CommunityID)
 		if err != nil {
--- a/protocol/common/message.go
+++ b/protocol/common/message.go
@@ -698,3 +698,14 @@ func (m *Message) LoadImage() error {
 
 	return nil
 }
+
+func (m *Message) SetAlbumID(albumID string) error {
+	imageMessage := m.GetImage()
+	if imageMessage == nil {
+		return errors.New("Image is empty")
+	}
+	imageMessage.AlbumId = albumID
+	m.Payload = &protobuf.ChatMessage_Image{Image: imageMessage}
+
+	return nil
+}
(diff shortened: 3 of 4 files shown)


protocol/messenger.go
```diff
@@ -1939,7 +1940,17 @@ func (m *Messenger) SendChatMessage(ctx context.Context, message *common.Message
...
 	var response MessengerResponse
 
+	generatedAlbumId := uuid.New().String()
 	for _, message := range messages {
+		if message.GetContentType() != protobuf.ChatMessage_IMAGE {
+			generatedAlbumId = ""
+		}
+	}
+
+	for _, message := range messages {
```
jrainville, line 1977: Can you do only one `for`? You can just check the `ContentType` to decide if you apply the album ID

protocol/message_validator.go
```diff
@@ -233,7 +233,7 @@ func ValidateReceivedChatMessage(message *protobuf.ChatMessage, whisperTimestamp
 		return errors.New("timestamp can't be 0")
 	}
 
-	if message.ContentType != protobuf.ChatMessage_DISCORD_MESSAGE {
+	if message.ContentType != protobuf.ChatMessage_DISCORD_MESSAGE && message.ContentType != protobuf.ChatMessage_IMAGE {
```
jrainville: I think we still want to validate the text, for the max length for example. Maybe we just need another argument in the validate function called `acceptEmptyMessage bool` that we set to `true` if the type is `IMAGE`.

jrainville (commented):
Small questions/suggestions. 
Otherwise, albumId wasn't applied before? Was it missing?

OmarBasem:
Can you please elaborate why are these changes needed? The image struct already has an albumId field, and we can send the `albumId` from the client side to `status-go` and it will be set.

cammellos:
> Can you please elaborate why are these changes needed? The image struct already has an albumId field, and we can send the `albumId` from the client side to `status-go` and it will be set.

I think it's better to keep as much logic in status-go as possible, so this is a change in this direction, so that the logic can be shared/reused by desktop as well.

@OmarBasem do you think there's a reason why we should keep generation of uuid on the client or you think it's ok to move it in status-go?

OmarBasem:
> I think it's better to keep as much logic in status-go as possible, so this is a change in this direction, so that the logic can be shared/reused by desktop as well.

@cammellos oh okay, that would be better ofc 👍

protocol/message_validator.go
borismelnik: No needs to validate text for `ImageMessage` cos user can send just image without a text. Bc of it, we added emoji for sending images if no input text. If we leave validation as is we should add `extra` text and skip it every time in UI part.

jrainville: Yeah, but if they do send text, we need to validate, otherwise, they can send a super long message and spam

protocol/messenger.go
borismelnik: this `for` loop was before me, git marked it as my changes :) but I will do it in previous one.

protocol/message_validator.go
borismelnik: oh yes you right... I will skip validation if Image just for empty text. Good catch.

protocol/messenger.go
```diff
@@ -1939,7 +1940,12 @@ func (m *Messenger) SendChatMessage(ctx context.Context, message *common.Message
 func (m *Messenger) SendChatMessages(ctx context.Context, messages []*common.Message) (*MessengerResponse, error) {
 	var response MessengerResponse
 
+	generatedAlbumId := uuid.New().String()
```
ilmotta: Note that `New()` panics if a UUID can't be created, whereas `NewRandom()` simply returns an error. It should be extremely rare to panic, but anyway.

Reference: https://github.com/google/UUID/blob/v1.0.0/version4.go#L13

ilmotta (approved):

protocol/message_validator.go
```diff
@@ -233,7 +233,13 @@ func ValidateReceivedChatMessage(message *protobuf.ChatMessage, whisperTimestamp
...
+		if err := ValidateText(message.Text); err != nil {
+			return err
+		}
+	}
+
+	if message.ContentType == protobuf.ChatMessage_IMAGE && message.Text != "" {
```
jrainville: It would be cleaner to add this condition to the previous like :
```go
if message.ContentType != protobuf.ChatMessage_DISCORD_MESSAGE || (message.ContentType == protobuf.ChatMessage_IMAGE && message.Text != "") {
```

jrainville (commented):
Looks good, just a code cleaning suggestion

jrainville (commented):
Looks good, just a code cleaning suggestion

protocol/message_validator.go
borismelnik: I'm thought about it so long... but I think you wrong :)

`||` that means `condition==true OR condition==true`

so if `message.ContentType == protobuf.ChatMessage_IMAGE && message.Text == ""` is true means  `message.ContentType != protobuf.ChatMessage_DISCORD_MESSAGE` is true  as well :) which means `true || false`. 

Am I right?

jrainville: Right, I messed up the condition, my bad. It would be 
```go
if message.ContentType != protobuf.ChatMessage_DISCORD_MESSAGE && (message.ContentType != protobuf.ChatMessage_IMAGE || message.Text != "") {
```

jrainville (approved):

ilmotta (approved):

cammellos (approved):
It'd be good to have a test covering the behavior

PR #3288, opened 2023-03-15
chore: Enable arm64 build on MacOs

alexjba:
Remove the forced amd64 compilation flag on arm64 MacOs.

We don need this anymore. Qt5.15 can be compiled on arm64.
The architecture will be configured by status-desktop Makefile.


stefandunca (approved):
Finally gone! :)

PR #3290, opened 2023-03-15
Local Pairing: Challenge Lifecycle Upgrade

### LINKED ISSUE #3299: Upgrade Local Pairing challenge process for multiple calls
Upgrade Local Pairing challenge process for multiple calls.

When the challenge process was first introduced there was only one endpoint to protect against unauthorised client access. **Note** that challenges are only one of many security considerations implemented to protect user data transmitted via LAN. The local pairing process is now more complex and requires multiple endpoints, it is prudent to update the challenge life cycle to manage dynamic challenges and multiple challenge processes. A single challenge should not be guarding each endpoint.

- [x] Limit session to only one client
- [x] Change challenge after every successful challenge response.
- [x] Resolve discussed point with @qfrank ,  https://github.com/status-im/status-go/pull/3248#discussion_r1138606205
  - Challenge should apply to getting installation data from `ReceiverServer` during `SenderClient` process flow.

I'm currently working on this here https://github.com/status-im/status-go/pull/3290

See https://github.com/status-im/status-go/pull/3248 where originally raised.

Samyoul:
## What's Changed?

- Migrated all challenge related functionality into `ChallengeGiver` and `ChallengeTaker` structs
- Integrated `ChallengeGiver` functionality into `Server` types
- Integrated `ChallengeTaker` functionality into `Client` types
- Simplified the challenge endpoint and middleware to execute `ChallengeGiver` functions
- Added back challenge functionality to sending installation data from the `ReceiverServer`
- Added IP lock to the`ChallengeGiver`, this prevents any other device on the network from acquiring the challenge session
- Updated challenge lifecycle to generate a new challenge after every successfully completed challenge
- Updated tests to confirm the new functionality

## Why Make the Change?

To improve the separation of concerns, the code now has a single type that handles all challenge giving and a single type that handles all challenge taking functionality. Improving maintainability and removing state management from structs that do not require access to specific states.

To improve connection security. Challenge changing after every success aligns the challenge system with best practices for multi call challenge protection [SASL](https://en.wikipedia.org/wiki/Simple_Authentication_and_Security_Layer).

Locking the session to the first device that requests a challenge prevents multiple clients on the LAN to attempt to access the payload data. If an attacker claims the lock first, they have a single attempt to succeed a challenge. A single failure blocks that session from any additional attempts.

## TODO

- [ ] Implement a bidirectional challenge system to prevent malicious actors from injecting their own keys in to a `ReceiverServer`. Unlikely as the attacker would also need to know the shared secret AES key

---

closes https://github.com/status-im/status-go/issues/3299

### DIFF
--- a/server/pairing/challenge_management.go
+++ b/server/pairing/challenge_management.go
@@ -5,6 +5,7 @@ import (
 	"crypto/rand"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 
 	"github.com/btcsuite/btcutil/base58"
@@ -48,6 +49,7 @@ type ChallengeGiver struct {
 	cookieStore *sessions.CookieStore
 	encryptor   *PayloadEncryptor
 	logger      *zap.Logger
+	authedIP    net.IP
 }
 
 func NewChallengeGiver(e *PayloadEncryptor, logger *zap.Logger) (*ChallengeGiver, error) {
@@ -61,6 +63,50 @@ func NewChallengeGiver(e *PayloadEncryptor, logger *zap.Logger) (*ChallengeGiver
 		encryptor:   e.Renew(),
 		logger:      logger,
 	}, nil
+}
+
+func (cg *ChallengeGiver) getIP(r *http.Request) (net.IP, *ChallengeError) {
+	h, _, err := net.SplitHostPort(r.RemoteAddr)
+	if err != nil {
+		cg.logger.Error("getIP: h, _, err := net.SplitHostPort(r.RemoteAddr)", zap.Error(err), zap.String("r.RemoteAddr", r.RemoteAddr))
+		return nil, &ChallengeError{"error", http.StatusInternalServerError}
+	}
+	return net.ParseIP(h), nil
+}
+
+func (cg *ChallengeGiver) registerClientIP(r *http.Request) *ChallengeError {
+	hIP, err := cg.getIP(r)
+	if err != nil {
+		return err
+	}
+	cg.authedIP = hIP
+	return nil
+}
+
+func (cg *ChallengeGiver) validateClientIP(r *http.Request) *ChallengeError {
+	// If we haven't registered yet register the IP
+	if cg.authedIP == nil || len(cg.authedIP) == 0 {
+		err := cg.registerClientIP(r)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Then compare the current req RemoteIP with the authed IP
+	hIP, err := cg.getIP(r)
+	if err != nil {
+		return err
+	}
+
+	if !cg.authedIP.Equal(hIP) {
+		cg.logger.Error(
+			"request RemoteAddr does not match authedIP: expected '%s', received '%s'",
+			zap.String("expected", cg.authedIP.String()),
+			zap.String("received", hIP.String()),
+		)
+		return &ChallengeError{"forbidden", http.StatusForbidden}
+	}
+	return nil
 }
 
 func (cg *ChallengeGiver) getSession(r *http.Request) (*sessions.Session, *ChallengeError) {
@@ -102,6 +148,11 @@ func (cg *ChallengeGiver) block(s *sessions.Session, w http.ResponseWriter, r *h
 }
 
 func (cg *ChallengeGiver) checkChallengeResponse(w http.ResponseWriter, r *http.Request) *ChallengeError {
+	ce := cg.validateClientIP(r)
+	if ce != nil {
+		return ce
+	}
+
 	s, ce := cg.getSession(r)
 	if ce != nil {
 		return ce
@@ -136,12 +187,17 @@ func (cg *ChallengeGiver) checkChallengeResponse(w http.ResponseWriter, r *http.
 		return cg.block(s, w, r)
 	}
 
-	// If every is ok, regenerate the challenge
+	// If every is ok, generate a new challenge for the next req
 	_, ce = cg.generateNewChallenge(s, w, r)
 	return ce
 }
 
 func (cg *ChallengeGiver) getChallenge(w http.ResponseWriter, r *http.Request) ([]byte, *ChallengeError) {
+	err := cg.validateClientIP(r)
+	if err != nil {
+		return nil, err
+	}
+
 	s, err := cg.getSession(r)
 	if err != nil {
 		return nil, err
--- a/server/pairing/challenge_management.go
+++ b/server/pairing/challenge_management.go
@@ -1,0 +1,144 @@
+package pairing
+
+import (
+	"bytes"
+	"crypto/rand"
+	"fmt"
+	"net/http"
+
+	"github.com/btcsuite/btcutil/base58"
+	"github.com/gorilla/sessions"
+	"go.uber.org/zap"
+)
+
+type ChallengeError struct {
+	Text     string
+	HttpCode int
+}
+
+func (ce *ChallengeError) Error() string {
+	return fmt.Sprintf("%s : %d", ce.Text, ce.HttpCode)
+}
+
+func makeCookieStore() (*sessions.CookieStore, error) {
+	auth := make([]byte, 64)
+	_, err := rand.Read(auth)
+	if err != nil {
+		return nil, err
+	}
+
+	enc := make([]byte, 32)
+	_, err = rand.Read(enc)
+	if err != nil {
+		return nil, err
+	}
+
+	return sessions.NewCookieStore(auth, enc), nil
+}
+
+type ChallengeGiver struct {
+	cookieStore *sessions.CookieStore
+	encryptor   *PayloadEncryptor
+	logger      *zap.Logger
+}
+
+func NewChallengeGiver(e *PayloadEncryptor, logger *zap.Logger) (*ChallengeGiver, error) {
+	cs, err := makeCookieStore()
+	if err != nil {
+		return nil, err
+	}
+
+	return &ChallengeGiver{
+		cookieStore: cs,
+		encryptor:   e,
+		logger:      logger,
+	}, nil
+}
+
+func (cg *ChallengeGiver) handleChallengeResponse(w http.ResponseWriter, r *http.Request) *ChallengeError {
+	s, err := cg.cookieStore.Get(r, sessionChallenge)
+	if err != nil {
+		cg.logger.Error("hs.GetCookieStore().Get(r, sessionChallenge)", zap.Error(err))
+		return &ChallengeError{"error", http.StatusInternalServerError}
+	}
+
+	blocked, ok := s.Values[sessionBlocked].(bool)
+	if ok && blocked {
+		return &ChallengeError{"forbidden", http.StatusForbidden}
+	}
+
+	// If the request header doesn't include a challenge don't punish the client, just throw a 403
+	pc := r.Header.Get(sessionChallenge)
+	if pc == "" {
+		return &ChallengeError{"forbidden", http.StatusForbidden}
+	}
+
+	c, err := cg.encryptor.decryptPlain(base58.Decode(pc))
+	if err != nil {
+		cg.logger.Error("c, err := hs.DecryptPlain(rc, hs.ek)", zap.Error(err))
+		return &ChallengeError{"error", http.StatusInternalServerError}
+	}
+
+	// If the challenge is not in the session store don't punish the client, just throw a 403
+	challenge, ok := s.Values[sessionChallenge].([]byte)
+	if !ok {
+		return &ChallengeError{"forbidden", http.StatusForbidden}
+	}
+
+	// Only if we have both a challenge in the session store and in the request header
+	// do we entertain blocking the client. Because then we know someone is trying to be sneaky.
+	if !bytes.Equal(c, challenge) {
+		s.Values[sessionBlocked] = true
+		err = s.Save(r, w)
+		if err != nil {
+			cg.logger.Error("err = s.Save(r, w)", zap.Error(err))
+		}
+
+		return &ChallengeError{"forbidden", http.StatusForbidden}
+	}
+	return nil
+}
+
+func (cg *ChallengeGiver) challenge(w http.ResponseWriter, r *http.Request) *ChallengeError {
+	s, err := cg.cookieStore.Get(r, sessionChallenge)
+	if err != nil {
+		cg.logger.Error("hs.GetCookieStore().Get(r, sessionChallenge)", zap.Error(err))
+		return &ChallengeError{"error", http.StatusInternalServerError}
+	}
+
+	challenge, ok := s.Values[sessionChallenge].([]byte)
+	if !ok {
+		challenge = make([]byte, 64)
+		_, err = rand.Read(challenge)
+		if err != nil {
+			cg.logger.Error("_, err = rand.Read(challenge)", zap.Error(err))
+			return &ChallengeError{"error", http.StatusInternalServerError}
+		}
+
+		s.Values[sessionChallenge] = challenge
+		err = s.Save(r, w)
+		if err != nil {
+			cg.logger.Error("err = s.Save(r, w)", zap.Error(err))
+			return &ChallengeError{"error", http.StatusInternalServerError}
+		}
+	}
+
+	w.Header().Set("Content-Type", "application/octet-stream")
+	_, err = w.Write(challenge)
+	if err != nil {
+		cg.logger.Error("_, err = w.Write(challenge)", zap.Error(err))
+	}
+	return nil
+}
+
+func (s *BaseServer) GetCookieStore() *sessions.CookieStore {
+	return s.cookieStore
+}
+
+func (s *BaseServer) DecryptPlain(data []byte) ([]byte, error) {
+	return s.encryptor.decryptPlain(data)
+}
+
+type ChallengeTaker struct {
+	encryptor *PayloadEncryptor
+}
--- a/server/pairing/challenge_management.go
+++ b/server/pairing/challenge_management.go
@@ -20,11 +20,11 @@ const (
 
 type ChallengeError struct {
 	Text     string
-	HttpCode int
+	HTTPCode int
 }
 
 func (ce *ChallengeError) Error() string {
-	return fmt.Sprintf("%s : %d", ce.Text, ce.HttpCode)
+	return fmt.Sprintf("%s : %d", ce.Text, ce.HTTPCode)
 }
 
 func makeCookieStore() (*sessions.CookieStore, error) {
@@ -142,14 +142,17 @@ func (cg *ChallengeGiver) checkChallengeResponse(w http.ResponseWriter, r *http.
 }
 
 func (cg *ChallengeGiver) getChallenge(w http.ResponseWriter, r *http.Request) ([]byte, *ChallengeError) {
-	s, ce := cg.getSession(r)
-	if ce != nil {
-		return nil, ce
+	s, err := cg.getSession(r)
+	if err != nil {
+		return nil, err
 	}
 
 	challenge, ok := s.Values[sessionChallenge].([]byte)
 	if !ok {
-		challenge, ce = cg.generateNewChallenge(s, w, r)
+		challenge, err = cg.generateNewChallenge(s, w, r)
+		if err != nil {
+			return nil, err
+		}
 	}
 	return challenge, nil
 }
--- a/server/pairing/challenge_management.go
+++ b/server/pairing/challenge_management.go
@@ -58,7 +58,7 @@ func NewChallengeGiver(e *PayloadEncryptor, logger *zap.Logger) (*ChallengeGiver
 func (cg *ChallengeGiver) handleChallengeResponse(w http.ResponseWriter, r *http.Request) *ChallengeError {
 	s, err := cg.cookieStore.Get(r, sessionChallenge)
 	if err != nil {
-		cg.logger.Error("hs.GetCookieStore().Get(r, sessionChallenge)", zap.Error(err))
+		cg.logger.Error("handleChallengeResponse: cg.cookieStore.Get(r, sessionChallenge)", zap.Error(err), zap.String("sessionChallenge", sessionChallenge))
 		return &ChallengeError{"error", http.StatusInternalServerError}
 	}
 
@@ -75,7 +75,7 @@ func (cg *ChallengeGiver) handleChallengeResponse(w http.ResponseWriter, r *http
 
 	c, err := cg.encryptor.decryptPlain(base58.Decode(pc))
 	if err != nil {
-		cg.logger.Error("c, err := hs.DecryptPlain(rc, hs.ek)", zap.Error(err))
+		cg.logger.Error("handleChallengeResponse: cg.encryptor.decryptPlain(base58.Decode(pc))", zap.Error(err), zap.String("pc", pc))
 		return &ChallengeError{"error", http.StatusInternalServerError}
 	}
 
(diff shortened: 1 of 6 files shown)


server/pairing/challenge_management.go
```diff
@@ -0,0 +1,247 @@
...
+type ChallengeError struct {
+	Text     string
+	HTTPCode int
+}
+
+func (ce *ChallengeError) Error() string {
+	return fmt.Sprintf("%s : %d", ce.Text, ce.HTTPCode)
+}
+
+func makeCookieStore() (*sessions.CookieStore, error) {
```
ilmotta, line 31: The `gorilla` project seems to be archived. Do you know if we have any plans of replacing any of its libraries, like `sessions`?

server/pairing/challenge_management.go
```diff
@@ -0,0 +1,247 @@
...
+	h, _, err := net.SplitHostPort(r.RemoteAddr)
+	if err != nil {
+		cg.logger.Error("getIP: h, _, err := net.SplitHostPort(r.RemoteAddr)", zap.Error(err), zap.String("r.RemoteAddr", r.RemoteAddr))
+		return nil, &ChallengeError{"error", http.StatusInternalServerError}
+	}
+	return net.ParseIP(h), nil
+}
+
+func (cg *ChallengeGiver) registerClientIP(r *http.Request) *ChallengeError {
+	hIP, err := cg.getIP(r)
```
ilmotta: What does `hIP` stands for? Could it be just `ip`?

server/pairing/challenge_management.go
```diff
@@ -0,0 +1,247 @@
...
+	s, err := cg.cookieStore.Get(r, sessionChallenge)
+	if err != nil {
+		cg.logger.Error("checkChallengeResponse: cg.cookieStore.Get(r, sessionChallenge)", zap.Error(err), zap.String("sessionChallenge", sessionChallenge))
+		return nil, &ChallengeError{"error", http.StatusInternalServerError}
+	}
+	return s, nil
+}
+
+func (cg *ChallengeGiver) generateNewChallenge(s *sessions.Session, w http.ResponseWriter, r *http.Request) ([]byte, *ChallengeError) {
+	challenge := make([]byte, 64)
```
ilmotta, line 122: We could give a name to 64, since it's used twice in this file and seems to be important. Do you know why we chose 64 random bytes as the challenge length?

server/pairing/challenge_management.go
```diff
@@ -0,0 +1,247 @@
...
+	if err != nil {
+		cg.logger.Error("regenerateNewChallenge: err = s.Save(r, w)", zap.Error(err))
+		return nil, &ChallengeError{"error", http.StatusInternalServerError}
+	}
+
+	return challenge, nil
+}
+
+func (cg *ChallengeGiver) block(s *sessions.Session, w http.ResponseWriter, r *http.Request) *ChallengeError {
+	s.Values[sessionBlocked] = true
```
ilmotta, line 140: In case the session can't be saved, shouldn't we revert the change to the `s.Values` map since it's being mutated? Or maybe since the change wasn't persisted it's fine?

server/pairing/challenge_management.go
```diff
@@ -0,0 +1,247 @@
...
+		return ce
+	}
+
+	blocked, ok := s.Values[sessionBlocked].(bool)
+	if ok && blocked {
+		return &ChallengeError{"forbidden", http.StatusForbidden}
+	}
+
+	// If the request header doesn't include a challenge don't punish the client, just throw a 403
+	pc := r.Header.Get(sessionChallenge)
```
ilmotta: I had to do a bit of research to understand what `pc` means. Is it `previous challenge`? Maybe it's just me, but otherwise we could be more explicit.

server/pairing/challenge_management.go
```diff
@@ -0,0 +1,247 @@
...
+	if err != nil {
+		cg.logger.Error("block: err = s.Save(r, w)", zap.Error(err))
+		return &ChallengeError{"error", http.StatusInternalServerError}
+	}
+
+	return &ChallengeError{"forbidden", http.StatusForbidden}
+}
+
+func (cg *ChallengeGiver) checkChallengeResponse(w http.ResponseWriter, r *http.Request) *ChallengeError {
+	ce := cg.validateClientIP(r)
```
ilmotta: For the size of this function, I found `ce` hard to grasp and I had to jump back to the declaration more than once. Maybe `challengeErr` is longer, but clearer. Or `cErr` to follow the common usage of the word `err` (although the character `c` is not very helpful).

Another case, what does `c` mean as the return value of `decryptPlain`? Maybe it's just my ignorance of common acronyms in security-oriented code :smile: 

Do you know if we're more or less trying to follow the naming guidelines from this: https://google.github.io/styleguide/go/decisions#variable-names

server/pairing/challenge_management.go
```diff
@@ -0,0 +1,247 @@
...
+
+	// If the request header doesn't include a challenge don't punish the client, just throw a 403
+	pc := r.Header.Get(sessionChallenge)
+	if pc == "" {
+		return &ChallengeError{"forbidden", http.StatusForbidden}
+	}
+
+	c, err := cg.encryptor.decryptPlain(base58.Decode(pc))
+	if err != nil {
+		cg.logger.Error("checkChallengeResponse: cg.encryptor.decryptPlain(base58.Decode(pc))", zap.Error(err), zap.String("pc", pc))
```
ilmotta: @Samyoul, It almost feels like we're trying to build a stack trace of sorts because the error messages are basically tracking the source code 1:1. Doesn't `err` contain enough information that we can grasp what happened from the logs? By default `zap` annotate all logs with the functions' file name and line number. If the logs do contain such annotations, at least to me it seems like we can reduce the verbosity of the log messages, because maintaining these logs also have a cost, especially since they're very detailed (copy from source).

Do we have any kind of guideline on `status-go` about how we prefer to log errors and the rationale?

ilmotta (approved):
Another batch of nice refactors! :fire:

server/pairing/challenge_management.go
Samyoul: This is a sad time, `gorilla` has / is the standard for all kinds of web tools for years. I have looked at different implementations and `gorilla` is still the best. But obviously we can't keep using an archived project forever. I'll raise an issue for this and we can investigate the feasibility. I feel like this is too much work for this PR though.

server/pairing/challenge_management.go
Samyoul: Oh, yeah, that is not clear. A lot of this functionality was in one function, and then I refactored my refactor and he var names moved with the section of code that they were associated with.

So `hIP` means `hostIP` because the IP address is derived from the `r.RemoteAddr` which returns a `host:port` string. We need to split the host and port then parse the host value. That's the background 😆 

But `hostIP` only makes sense in the internal context of `getIP()`, so your suggestion is a good one. Also a good sign is that when rereading this code, I had forgotten myself what `hIP` meant 🙃

server/pairing/challenge_management.go
ilmotta: Sorry I if wasn't clear, but I think it's totally fine to use `sessions` in this PR or even in future ones. I was just curious to know if you've thought about how/when to replace a subset of it in `status-go`, such as `sessions`. Thanks @Samyoul

ilmotta: It's sad really. I tried to research alternatives for Go, but gorilla libs still kept showing up.

server/pairing/challenge_management.go
Samyoul: Ah, the two `64` values represent two different things.

The first one that generates the `auth` var is the higher value of the recommended length for a session authentication key:

> // It is recommended to use an authentication key with 32 or 64 bytes.

The second usage, the usage you mention here, can be any length at all. The challenge requires AES encryption using a 256 bit / 32 byte AES key, which is the upper value specified in the AES specification. The challenge length could be any number really as the security of the challenge response is anchored in the strength of AES encryption. I chose 64 bytes because 2^6 is more "square" than 2^5 (2^6 == 8^2), 2^4 is a little small and 2^8 is crazy too big (2^7 is big too but less square than 2^8). 2^8 is a nice number though, 2^8 == 4^4 == 16^2

server/pairing/challenge_management.go
Samyoul: Yes, it is safe to assume that a failure on `Save` means that the data has not persisted:

```go
// Save adds a single session to the response.
func (s *CookieStore) Save(r *http.Request, w http.ResponseWriter,
	session *Session) error {
	encoded, err := securecookie.EncodeMulti(session.Name(), session.Values,
		s.Codecs...)
	if err != nil {
		return err
	}
	http.SetCookie(w, NewCookie(session.Name(), encoded, session.Options))
	return nil
}
```

Notice in the underlying function that errors only come from a failure to encode the session values, meaning there is no partial save.

server/pairing/challenge_management.go
Samyoul: I do try to follow the guidelines 😆 , I regularly sight them in my own reviews of other PRs.

`cErr` seems better, the only reason that I don't use the `err` var here is because `cErr` is an explicit `error` type `ChallengeError` and this value conflicts with function calls that do return only the untyped `error` interface.

The reason I don't return the `error` interface from these functions is because the consuming code uses the properties of `ChallengeError` to create `http.Error()`s .

However this seems like a place where a change to the whole strategy makes sense. I'll still return `ChallengeError`s within the function, but all functions will return the `error` interface. Consuming code will type check the error and if it is a `ChallengeError` process a `http.Error()`, if not throw a `http.StatusInternalServerError`.

Then we don't need `cErr` we will only have `err`

Samyoul: `c` means "challenge", I'll rename it to `dc` == `decrypted challenge`

server/pairing/challenge_management.go
ilmotta: 2^8 is prettier for sure! Thanks for the detailed explanation :)

server/pairing/challenge_management.go
Samyoul: Ah, `pc` is a hangover from a different implementation, I called it `presented challenge`. I think I'll rename it to `c` as the code makes it clear we are getting a challenge.

server/pairing/challenge_management.go
ilmotta: Indeed, no partial save. Thanks for pointing out the source code.

server/pairing/challenge_management.go
ilmotta: > Then we don't need cErr we will only have err

Good idea!

> c means "challenge"

Ah okay, I guess what confused me is that a couple lines below we have a variable with a "long" name `challenge`, so I was thinking, if this is `challenge`, then what's `c`?

server/pairing/challenge_management.go
Samyoul: This is a bit of my own creation. `zap` logs *can* show the full stacktrace along with a lot of other very helpful metadata, however the logger isn't always configured to show this helpful information. We have a lot of different logger instantiations and most put the logger in the least verbose mode. Copying the source into the log message is my way to try to even this inconsistency.

Ultimately I could put a descriptive message in the text field, but I would only be using it to describe in every day speech what the source code does, which is human readable. Also I'd try to make the comment as unique as possible so that I could grep it easily.

Better logging is something that `status-go` needs, we should be taking full advantage of `zap`'s functionality so I propose that:

- I change logger messages in this PR to something human readable
- I create an issue to plan and track upgrading logging generally across all `status-go` to use more featureful logging.
- We along with anyone else that cares formalises logging guidelines

Does this seem reasonable to you @ilmotta?

ilmotta: Sounds like a most excellent plan @Samyoul.

Given the way we log will soon/eventually improve based on the plan you outlined above, I'm totally fine if you keep the logs the way they are in this PR and any other ones coming after :) Until the new logging guidelines are brought to life it might be better to stick to what we've been doing, even if there are multiple conventions.

I was just not on the same page as you in terms of the current logging situation in `status-go`, but now I understand better your rationale for logging the way this PR does.

Thank you!

server/pairing/challenge_management.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func NewChallengeTaker(e *PayloadEncryptor) *ChallengeTaker {
+	return &ChallengeTaker{
+		encryptor: e.Renew(),
+	}
+}
+
+func (ct *ChallengeTaker) SetChallenge(resp *http.Response) error {
+	challenge, err := ioutil.ReadAll(resp.Body)
```
qfrank, line 229: emm..`ioutil.ReadAll(resp.Body)` makes me a little confused since we already returned the challenge via http header `Set-Cookie` right?

server/pairing/client.go
```diff
@@ -210,6 +192,12 @@ func (c *SenderClient) receiveInstallationData() error {
 		return err
 	}
 
+	err = c.challengeTaker.DoChallenge(req)
```
qfrank, line 195: maybe we can hide the detail of doing challenge in `BaseClient` so that we only need focus on which endpoint we need interact and what real `business data` we need exchange? 🤔

server/pairing/challenge_management.go
Samyoul: Perhaps I need to change the name of this function `SetChallenge`, or better explain the context. `SetChallenge` takes a `http.Response` and sets the internal field `ct.serverChallenge`.

The function doesn't set the `http.Request` challenge, `ChallengeTaker` sets its own internal value from a `http.Response`. I can add a comment on this function to better explain it. Does that make sense @qfrank?

qfrank: sorry, maybe my previous comment is not clear. i mean we can get challenge by parsing from http header e.g. `resp.Header.Get("Set-Cookie")`, we don't need read http response body since challenge already being included in http header ? @Samyoul

qfrank: there is already a place used for challenge to return, now http body also contains it, this(> 1 places) makes me confused

Samyoul: Hmm @qfrank , I'm confused now because we don't use `Set-Cookie` anywhere in the code. Well at least I can't see it. We've always took the challenge from the response body, because the value in the cookie is encrypted with a random key.

---

I did a little bit of research on this point just to be certain, I added some code to catch the cookie value and the challenge:

```go
func (c *BaseClient) getChallenge() error {
	c.baseAddress.Path = pairingChallenge
	resp, err := c.Get(c.baseAddress.String())
	if err != nil {
		return err
	}

	if resp.StatusCode != http.StatusOK {
		return fmt.Errorf("[client] status not ok when getting challenge, received '%s'", resp.Status)
	}

	// START New code
	cs := resp.Cookies()
	for _, c := range cs {
		fmt.Printf("  %s=%s\n", c.Name, c.Value)
		decodedBytes, err := base64.StdEncoding.DecodeString(c.Value)
		if err != nil {
			fmt.Println("Error decoding string:", err)
		}
		spew.Dump(decodedBytes)
	}

	spew.Dump(resp.Header.Get("Set-Cookie"))
	// END New code

	return c.challengeTaker.SetChallenge(resp)
}
```

```go
func (ct *ChallengeTaker) SetChallenge(resp *http.Response) error {
	challenge, err := ioutil.ReadAll(resp.Body)
	if err != nil {
		return err
	}
	spew.Dump("SetChallenge", challenge)
	ct.serverChallenge = challenge
	return nil
}
```

The output was:

```text
challenge=MTY3OTU4NjIyN3x1cHJrVDVRUmlwMHEzbFdTc3YtR24xYUoxcmpWbWJPVVkxekVHVG9IR0pVX3B0Q3lDUzdwWklsRVZMbWlyaGZkckpRSUVuRGRPdUtKNW1nRUtGRzMwSm1QREo3ZzVfZlhJdDlCVy13eHpIMG1ucWxhcENfeWJDdkVCRE56RXI3cWJtSEZ1eGpkcEQyT3BweENaWnhheTZkYWtHZWRoeHRqZFNJZExHamVybnpCekQ1MnzjeaCVUm-W5lfjbbWCT-3ynIv9BlRcECqcDomzx9mJQA==
Error decoding string: illegal base64 data at input byte 258
([]uint8) (len=192 cap=222) {
 00000000  31 36 37 39 35 38 36 32  32 37 7c 75 70 72 6b 54  |1679586227|uprkT|
 00000010  35 51 52 69 70 30 71 33  6c 57 53 73 76 2d 47 6e  |5QRip0q3lWSsv-Gn|
 00000020  31 61 4a 31 72 6a 56 6d  62 4f 55 59 31 7a 45 47  |1aJ1rjVmbOUY1zEG|
 00000030  54 6f 48 47 4a 55 5f 70  74 43 79 43 53 37 70 5a  |ToHGJU_ptCyCS7pZ|
 00000040  49 6c 45 56 4c 6d 69 72  68 66 64 72 4a 51 49 45  |IlEVLmirhfdrJQIE|
 00000050  6e 44 64 4f 75 4b 4a 35  6d 67 45 4b 46 47 33 30  |nDdOuKJ5mgEKFG30|
 00000060  4a 6d 50 44 4a 37 67 35  5f 66 58 49 74 39 42 57  |JmPDJ7g5_fXIt9BW|
 00000070  2d 77 78 7a 48 30 6d 6e  71 6c 61 70 43 5f 79 62  |-wxzH0mnqlapC_yb|
 00000080  43 76 45 42 44 4e 7a 45  72 37 71 62 6d 48 46 75  |CvEBDNzEr7qbmHFu|
 00000090  78 6a 64 70 44 32 4f 70  70 78 43 5a 5a 78 61 79  |xjdpD2OppxCZZxay|
 000000a0  36 64 61 6b 47 65 64 68  78 74 6a 64 53 49 64 4c  |6dakGedhxtjdSIdL|
 000000b0  47 6a 65 72 6e 7a 42 7a  44 35 32 7c e3 79 a0 95  |GjernzBzD52|.y..|
}
(string) (len=380) "challenge=MTY3OTU4NjIyN3x1cHJrVDVRUmlwMHEzbFdTc3YtR24xYUoxcmpWbWJPVVkxekVHVG9IR0pVX3B0Q3lDUzdwWklsRVZMbWlyaGZkckpRSUVuRGRPdUtKNW1nRUtGRzMwSm1QREo3ZzVfZlhJdDlCVy13eHpIMG1ucWxhcENfeWJDdkVCRE56RXI3cWJtSEZ1eGpkcEQyT3BweENaWnhheTZkYWtHZWRoeHRqZFNJZExHamVybnpCekQ1MnzjeaCVUm-W5lfjbbWCT-3ynIv9BlRcECqcDomzx9mJQA==; Path=/; Expires=Sat, 22 Apr 2023 15:43:47 GMT; Max-Age=2592000"
(string) (len=12) "SetChallenge"
([]uint8) (len=64 cap=512) {
 00000000  b8 ce 95 33 b3 e1 e5 dc  0e 70 17 5d 8f 43 f0 ed  |...3.....p.].C..|
 00000010  f0 28 c2 36 14 3b 65 74  9d c2 4e 9b fc 05 6e ef  |.(.6.;et..N...n.|
 00000020  c1 e0 55 26 57 af b9 98  82 b3 aa bd b4 2d 59 12  |..U&W........-Y.|
 00000030  3a 5e b4 94 03 ec 4c 64  a4 54 b4 f8 8f 2b 2b 01  |:^....Ld.T...++.|
}
```

Notice that the output values are very different and not client side readable? Does this make sense @qfrank or am I missing something?

qfrank: > We've always took the challenge from the response body, because the value in the cookie is encrypted with a random key.

oh make sense, my bad 🤦‍♂️, i forgot it's encrypted. thank you for your detailed explanation 

> we don't use Set-Cookie anywhere in the code

we used it indirectly, that's one part how session works

server/pairing/client.go
Samyoul: I like this idea I've been trying to figure out how we could do this. The difficulty I am having is how do we pass the new request to the base client before the request is instantiated by the receiving / sending client?

What sort of approach did you have in mind?

qfrank: My initial idea is to add a method called DoChallengeRequest to the SenderClient to wrap the Client.Do method. The rough code is as follows:
```
func (c *SenderClient) DoChallengeRequest(req *Request) (*Response, error) {
    err = c.challengeTaker.DoChallenge(req)
    if err != nil {
        signal.SendLocalPairingEvent(Event{Type: EventTransferError, Error: err.Error(), Action: ActionPairingInstallation})
        return nil, err
    }
    c.Do(req) ...
}
```

is this a good start?

Samyoul: Link to issue: https://github.com/status-im/status-go/issues/3343

Ah, thank you @qfrank for your suggestion. I think that this change is getting into the area of generalising `Client` logic rather than updating the Challenge management.

However I did initially work with this idea during the development of #3248 . It is cool to know that we both broadly agree on this idea.

I think that we could generalise so that the `BaseClient` has 3 main functions `sendData`, `receiveData` and `exchangeData`.

I'll define `exchangeData` later, I need to think about how that could work.

**`sendData`**

```go
func (c *BaseClient) sendData(m PayloadMounter, action Action, path, description string) error {
	err := m.Mount()
	if err != nil {
		return err
	}

	c.baseAddress.Path = path
	req, err := http.NewRequest(http.MethodPost, c.baseAddress.String(), bytes.NewBuffer(m.ToSend()))
	if err != nil {
		return err
	}
	req.Header.Set("Content-Type", "application/octet-stream")

	err = c.challengeTaker.DoChallenge(req)
	if err != nil {
		signal.SendLocalPairingEvent(Event{Type: EventTransferError, Error: err.Error(), Action: action})
		return err
	}

	resp, err := c.Do(req)
	if err != nil {
		signal.SendLocalPairingEvent(Event{Type: EventTransferError, Error: err.Error(), Action: action})
		return err
	}

	if resp.StatusCode != http.StatusOK {
		err = fmt.Errorf("[client] status not okay when sending %s, status: %s", description, resp.Status)
		signal.SendLocalPairingEvent(Event{Type: EventTransferError, Error: err.Error(), Action: action})
		return err
	}

	signal.SendLocalPairingEvent(Event{Type: EventTransferSuccess, Action: action})
	return nil
}
```

**`receiveData`**

```go
func (c *BaseClient) receiveData(r PayloadReceiver, path string, action Action, description string) error {
	c.baseAddress.Path = path
	req, err := http.NewRequest(http.MethodGet, c.baseAddress.String(), nil)
	if err != nil {
		return err
	}

	err = c.challengeTaker.DoChallenge(req)
	if err != nil {
		signal.SendLocalPairingEvent(Event{Type: EventTransferError, Error: err.Error(), Action: action})
		return err
	}

	resp, err := c.Do(req)
	if err != nil {
		signal.SendLocalPairingEvent(Event{Type: EventTransferError, Error: err.Error(), Action: action})
		return err
	}

	if resp.StatusCode != http.StatusOK {
		err = fmt.Errorf("[client] status not ok when receiving %s, received '%s'", description, resp.Status)
		signal.SendLocalPairingEvent(Event{Type: EventTransferError, Error: err.Error(), Action: action})
		return err
	}

	payload, err := io.ReadAll(resp.Body)
	if err != nil {
		signal.SendLocalPairingEvent(Event{Type: EventTransferError, Error: err.Error(), Action: action})
		return err
	}
	signal.SendLocalPairingEvent(Event{Type: EventTransferSuccess, Action: action})

	err = r.Receive(payload)
	if err != nil {
		signal.SendLocalPairingEvent(Event{Type: EventProcessError, Error: err.Error(), Action: action})
		return err
	}
	signal.SendLocalPairingEvent(Event{Type: EventProcessSuccess, Action: action})
	return nil
}
```

Change all the `Sender` and `Receiver` funcs like below:

```go
func (c *SenderClient) sendAccountData() error {
	return c.sendData(c.accountMounter, ActionPairingAccount, pairingReceiveAccount, "account data")
}
```

What I'll do is add an issue from this comment and implement the full solutions in a dedicated PR.

PR #3292, opened 2023-03-15
Change getChats to getChannelGroups and add getChannelGroupByID

jrainville:
Needed for https://github.com/status-im/status-desktop/issues/9858

Changes `getChats` to `getChannelGroups`, since they don't just get chats, but also everything relating to channel groups.

I also simplified it so that it only gets the channel group info and not chats or categories. That way, we can create the channel group buttons (navigation).

Then, when a section is activated, we call the other new function called `getChannelGroupByID` to get all the info about that channel group (chats, categories, members).

I added a condition to `toApiChat` to only get the chat info, because we do not need the pinned chats or the chat members, and those calls take a long time.

### DIFF
--- a/protocol/chat.go
+++ b/protocol/chat.go
@@ -228,6 +228,10 @@ func (c *Chat) PrivateGroupChat() bool {
 	return c.ChatType == ChatTypePrivateGroupChat
 }
 
+func (c *Chat) IsActivePersonalChat() bool {
+	return c.Active && (c.OneToOne() || c.PrivateGroupChat() && c.Public()) && c.CommunityID == ""
+}
+
 func (c *Chat) CommunityChatID() string {
 	if c.ChatType != ChatTypeCommunityChat {
 		return c.ID
--- a/services/chat/api.go
+++ b/services/chat/api.go
@@ -101,6 +101,8 @@ type ChannelGroup struct {
 	BanList                 []string                                 `json:"banList"`
 	Encrypted               bool                                     `json:"encrypted"`
 	CommunityTokensMetadata []*protobuf.CommunityTokenMetadata       `json:"communityTokensMetadata"`
+	UnviewedMessagesCount   int                                      `json:"unviewedMessagesCount"`
+	UnviewedMentionsCount   int                                      `json:"unviewedMentionsCount"`
 }
 
 func NewAPI(service *Service) *API {
@@ -124,7 +126,7 @@ func unique(communities []*communities.Community) (result []*communities.Communi
 	return result
 }
 
-func (api *API) GetChats(ctx context.Context) (map[string]ChannelGroup, error) {
+func (api *API) GetChannelGroups(ctx context.Context) (map[string]ChannelGroup, error) {
 	joinedCommunities, err := api.s.messenger.JoinedCommunities()
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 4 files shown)


protocol/chat.go
```diff
@@ -228,6 +228,10 @@ func (c *Chat) PrivateGroupChat() bool {
 	return c.ChatType == ChatTypePrivateGroupChat
 }
 
+func (c *Chat) IsActivePersonnalChat() bool {
```
0x-r4bbit:
```suggestion
func (c *Chat) IsActivePersonalChat() bool {
```

services/chat/api.go
```diff
@@ -192,32 +208,138 @@ func (api *API) GetChats(ctx context.Context) (map[string]ChannelGroup, error) {
...
-				Position: int(cat.Position),
-			}
+		result[community.IDString()] = chGrp
+	}
+
+	return result, nil
+}
+
+func (api *API) GetChatsByChannelGroupId(ctx context.Context, channelGroupId string) (*ChannelGroup, error) {
+	pubKey := types.EncodeHex(crypto.FromECDSAPub(api.s.messenger.IdentityPublicKey()))
```
0x-r4bbit, line 226:
```suggestion
	pubKey := types.EncodeHex(api.s.messenger.IdentityPublicKeyCompressed()))
```
I believe we can use `IdentityPublicKeyCompressed()` here

services/chat/api.go
```diff
@@ -192,32 +208,138 @@ func (api *API) GetChats(ctx context.Context) (map[string]ChannelGroup, error) {
...
+
+	joinedCommunities, err := api.s.messenger.JoinedCommunities()
+	if err != nil {
+		return nil, err
+	}
+
+	found := false
+	var community *communities.Community
+	for _, comm := range joinedCommunities {
+		if comm.ID().String() == channelGroupId {
```
0x-r4bbit:
```suggestion
		if comm.IDString() == channelGroupId {
```

services/chat/api.go
```diff
@@ -192,32 +208,138 @@ func (api *API) GetChats(ctx context.Context) (map[string]ChannelGroup, error) {
...
+	}
+
+	if !found {
+		spectatedCommunities, err := api.s.messenger.SpectatedCommunities()
+		if err != nil {
+			return nil, err
+		}
+
+		for _, comm := range spectatedCommunities {
+			if comm.ID().String() == channelGroupId {
```
0x-r4bbit:
```suggestion
			if comm.IDString() == channelGroupId {
```

0x-r4bbit (changes requested):

services/chat/api.go
jrainville: It didn't work :shrug:

0x-r4bbit (approved):

services/chat/api.go
```diff
@@ -192,32 +208,138 @@ func (api *API) GetChats(ctx context.Context) (map[string]ChannelGroup, error) {
...
-				Name:     cat.Name,
-				Position: int(cat.Position),
-			}
+		result[community.IDString()] = chGrp
+	}
+
+	return result, nil
+}
+
+func (api *API) GetChatsByChannelGroupID(ctx context.Context, channelGroupID string) (*ChannelGroup, error) {
```
osmaczko, line 225: I noticed that the function can return chats with an empty members list, which goes against the 'rule of least surprise' and may be confusing for clients who rely on this property without anticipating such cases. To address this, we could consider introducing a new type, such as 'ChatInfo', that contains only essential fields. It may be a good idea to implement it in the future to ensure a more explicit and consistent representation of the data.

osmaczko (approved):

PR #3294, opened 2023-03-16
feat: Added `setInstallationName` method

igor-sirotin:
Added `SetInstallationName` name.
Previous `SetInstallationMetadata` requires the client to know and copy all other metadata, which was not quite convenient.


protocol/messenger.go
```diff
@@ -1660,6 +1660,16 @@ func (m *Messenger) SetInstallationMetadata(id string, data *multidevice.Install
...
 }
 
+func (m *Messenger) SetInstallationName(id string, name string) error {
+	installation, ok := m.allInstallations.Load(id)
+	if !ok {
+		return errors.New("no installation found")
+	}
+
+	installation.InstallationMetadata.Name = name
+	return m.encryptor.SetInstallationName(&m.identity.PublicKey, id, name)
```
qfrank: we should prefer use m.IdentityPublicKey() instead of access field directly

igor-sirotin: Well I just copied from `setInstallationMetadata` 😄 
https://github.com/status-im/status-go/blob/93cee626f06132e58eddc013b9adb229c6e95f8e/protocol/messenger.go#L1656

But no prob, will fix both 👌

igor-sirotin:
@qfrank fixed, please checkout :)

qfrank (approved):

Samyoul (approved):
I'm quite surprised this didn't already exist

PR #3295, opened 2023-03-16
feat: add verified wallet accounts to community requests

0x-r4bbit:
This commit extends the `CommunityRequestToJoin` with `RevealedAddresses` which represent wallet addresses and signatures provided by the sender, to proof a community owner ownership of those wallet addresses.

**Note: This only works with keystore files maanged by status-go**

At high level, the follwing happens:

1. User instructs Status to send a request to join to a community. By adding a password hash to the instruction, Status will try to unlock the users keystore and verify each wallet account.
2. For every verified wallet account, a signature is created for the following payload, using each wallet's private key

   ``` keccak256(chatkey + communityID + requestToJoinID) ``` A map of walletAddress->signature is then attached to the community request to join, which will be sent to the community owner
3. The owner node receives the request, and if the community requires users to hold tokens to become a member, it will check and verify whether the given wallet addresses are indeed owned by the sender. If any signature provided by the request cannot be recovered, the request is immediately declined by the owner.
4. The verified addresses are then added to the owner node's database such that, once the request should be accepted, the addresses can be used to check on chain whether they own the necessary funds to fulfill the community's permissions

The checking of required funds is **not** part of this commit. It will be added in a follow-up commit.


Samyoul (approved):
Cool. Looks all good to me. The only issue is the growth of the `NewManager` func signature. But I've left a comment about in the new PR.

protocol/communities/persistence.go
```diff
@@ -322,6 +322,59 @@ func (p *Persistence) SaveRequestToJoin(request *RequestToJoin) (err error) {
...
+	defer func() {
+		if err == nil {
+			err = tx.Commit()
+			return
+		}
+		// don't shadow original error
+		_ = tx.Rollback()
+	}()
+
+	query := `INSERT OR REPLACE INTO communities_requests_to_join_revealed_addresses (request_id, address, signature) VALUES (?, ?, ?)`
```
jrainville: I don't think saving the signature is useful. Once we have confirmed that the addresses belong to the account, I think we can just save the addresses? Maybe I forgot a use case though.

jrainville (commented):
Good job. I just had a question whether to save the signature or not.

protocol/communities/persistence.go
0x-r4bbit: Actually that's true.
Originally I kept it to re-validate given addresses when memberships have to be accepted manually.
But I don't think there's a reason to do so anymore. revealed addresses only end up in the database if they have been successfully recovered from the signature in the first place.

Will remove that.

protocol/communities/manager.go
```diff
@@ -81,7 +83,19 @@ func (t *HistoryArchiveDownloadTask) Cancel() {
...
+	accountsManager *account.GethManager
+}
+
+type ManagerOption func(*managerOptions)
+
+func WithAccountManager(accountsManager *account.GethManager) ManagerOption {
+	return func(opts *managerOptions) {
+		opts.accountsManager = accountsManager
+	}
+}
```
0x-r4bbit, line 96: @Samyoul FYI introduced `managerOptions` here. This will extend in the follow-up PRs

PR #3297, opened 2023-03-16
Add CreateAccount for a simpler endpoint for creating accounts

cammellos:
This is still a bit of a work in progress, it's only used by the new UI for mobile, but I would like to get it in so clients can still playing with it, but more is needed to make it fully functional.

Currently, status clients have to do a lot of work to create an account (derive addresses, store them, pass node config etc).

The aim of this PR is to simplify account creation so that clients have to pass the bare minimum of parameters.

It will be a work in progress for a bit, but I would like to merge as I go along not to create a too large PR.

This should have no impact on existing clients, since it's a brand new endpoint.


api/defaults.go
```diff
@@ -0,0 +1,819 @@
...
+	c, err := rand.Int(rand.Reader, length)
+	if err != nil {
+		return "", err
+	}
+
+	return dictionary[a.Int64()] + " " + dictionary[b.Int64()] + " " + dictionary[c.Int64()], nil
+
+}
+
+var dictionary = []string{
```
ilmotta, line 199: We can use `...` to set the length of the array.

api/geth_backend.go
```diff
@@ -692,6 +696,88 @@ func (b *GethStatusBackend) ConvertToKeycardAccount(account multiaccounts.Accoun
...
+	b.UpdateRootDataDir(request.BackupDisabledDataDir)
+	err := b.OpenAccounts()
+	if err != nil {
+		b.log.Error("failed open accounts", err)
+		return err
+	}
+
+	generator := b.accountManager.AccountsGenerator()
+
+	generatedAccountInfos, err := generator.Generate(12, 1, "")
```
J-Son89: what do these numbers mean? perhaps a variable for context would be good? 🤔

api/geth_backend.go
```diff
@@ -692,6 +696,88 @@ func (b *GethStatusBackend) ConvertToKeycardAccount(account multiaccounts.Accoun
...
+
+	walletDerivedAccount := derivedAddresses[pathDefaultWallet]
+	walletAccount := &accounts.Account{
+		PublicKey: types.Hex2Bytes(walletDerivedAccount.PublicKey),
+		KeyUID:    info.KeyUID,
+		Address:   types.HexToAddress(walletDerivedAccount.Address),
+		Color:     "",
+		Wallet:    true,
+		Path:      pathDefaultWallet,
+		Name:      "Ethereum account",
```
J-Son89: is there no constants for strings like these? 🤔

cammellos: I can const at the top of the file

api/geth_backend.go
cammellos: indeed :)

cammellos:
@J-Son89 addressed feedback, thank you!

J-Son89 (approved):

ilmotta (commented):
Sorry I couldn't review in time @cammellos.

If possible, I usually prefer to separate configuration from code, particularly these types of settings. The default settings could be loaded from a separate non-Go file. We could read the settings from a TOML file, or something decent that allows comments (not JSON). This is quite useful for users that want to customize `status-go` without messing with the code, or want to play around with different settings persisted to different configuration files, etc.

Just a question: Obviously the dictionary words were carefully chosen, but who did and how? As a possible improvement, should we also let the dictionary be configurable? Do you see value in that?

api/defaults.go
cammellos: > Sorry I couldn't review in time @cammellos.
> 
> If possible, I usually prefer to separate configuration from code, particularly these types of settings. The default settings could be loaded from a separate non-Go file. We could read the settings from a TOML file, or something decent that allows comments (not JSON). This is quite useful for users that want to customize `status-go` without messing with the code, or want to play around with different settings persisted to different configuration files, etc.
> 
> Just a question: Obviously the dictionary words were carefully chosen, but who did and how? As a possible improvement, should we also let the dictionary be configurable? Do you see value in that?

@ilmotta we can do for sure,
this is very rough, I want to get it working and make sure I am capturing only the essential values that the frontend needs to pass to the backend for proper functioning (ideally the least amount of configurable options as possible).

I am happy to keep the config in some other formats, the only nice thing about code is that it's type safe and if we change the internal config we will have to update the config.

One example was in the frontend, we were passing `StatusAccountsService: {Enabled: true}` as JSON, but that wasn't used anymore (or at least I could not find any reference in status-go).

So ideally something like you suggested, but with the added property that it would throw an error on extra keys?
I can look into something like that, but maybe @Samyoul / yourself have some suggestions 

> Just a question: Obviously the dictionary words were carefully chosen, but who did and how? As a possible improvement, should we also let the dictionary be configurable? Do you see value in that?

Copied it from the frontend, that's pretty much what I know about it :) I don't think there's much value into making it configurable, the signing phrase is used to prevent spoofing in the frontend, so most people will not want to customize it, but of course we could

ilmotta: I agree Andrea, my comment about making the default settings configurable would be just a nice to have, to be done later on once you're satisfied with the improvements you're trying to bring.

>So ideally something like you suggested, but with the added property that it would throw an error on extra keys?

100%. I think the coercion from a config file into Go code should be as strict as possible and fail fast with good error messages. Maps should be closed, this prevents typos for example or silent settings passed from clients that do nothing.

> I am happy to keep the config in some other formats, the only nice thing about code is that it's type safe and if we change the internal config we will have to update the config.

Yes, we should not lose any type safety because of settings loaded from the environment (files, env vars). We would coerce, validate and store them in well defined structs.

Good point, there's a balance between configurability vs code config. It can be inconvenient if it's "too configurable".

>Copied it from the frontend, that's pretty much what I know about it :) I don't think there's much value into making it configurable, the signing phrase is used to prevent spoofing in the frontend, so most people will not want to customize it, but of course we could

Thanks. Sounds good to me :)

Samyoul: @ilmotta / @cammellos Sorry I'm even more late to this discussion. Can I ask are we wanting to overwrite select fields of a default configuration with a some input configuration?

Like this https://go.dev/play/p/fSRVAnr2F7K ?

```go
type Cthulhu struct {
    AeonsSlumbering      uint
    SummoningIncantation string
    Awoken               bool
    PlanesOccupied       int
    Sacrifices           []string
}

func TestJSONMarshalling(t *testing.T) {
    defaultCthulhu := &Cthulhu{
        AeonsSlumbering:      6,
        SummoningIncantation: "ph'nglui mglw'nafh Cthulhu R'lyeh wgah'nagl fhtagn",
        Awoken:               false,
        PlanesOccupied:       -3,
        Sacrifices:           []string{"Sam", "Andrea", "Vitaliy"},
    }

    incomingJSON := `{"Awoken":true,"Sacrifices":["Our Dignity"]}`

    err := json.Unmarshal([]byte(incomingJSON), defaultCthulhu)
    if err != nil {
        fmt.Println(err.Error())
    }
    fmt.Printf("%+v\n", defaultCthulhu)
}
```

```text
=== RUN   TestJSONMarshalling
&{AeonsSlumbering:6 SummoningIncantation:ph'nglui mglw'nafh Cthulhu R'lyeh wgah'nagl fhtagn Awoken:true PlanesOccupied:-3 Sacrifices:[Our Dignity]}
--- PASS: TestJSONMarshalling (0.00s)
PASS
```

ilmotta: Yes @Samyoul, something like that was in my mind (although with something like TOML that supports comments). I understand increasing configurability via external mechanisms is sometimes hard to justify, since the hard coded alternative works well.

In the end, I don't know if separating configuration from code will bring enough value versus the effort + convenience of just changing the code. It's just an idea to be considered in future iterations.

I found that sometimes an external configuration forced me to create systems that made less assumptions, since the config state could be more easily changed by consumers (devs, services, etc). Configurability is usually brought up in discussions related to 12-factor apps or when there's an intention of applying some of the principles, but perhaps this is a weak argument for status-go, right?

cammellos: I think external config is fine, for overriding the JSON, in theory works ok, and that was our initial idea with @Samyoul , but then you look at what we need to pass, and it's all deeply nested JSON, so it looks like we are bleeding out a lot of implementation details to the client.
For example, say the client wants to override the fleet for waku2.
The will have to pass JSON that looks like:

`{NodeConfig: {WakuV2Config: {Enabled: true, Fleet: "...."}}, RootDirectory: "..."}`

and even if we make sure that only relevant parts are overriden, so it doesn't have to pass a full map if only overriding a single value, I can't think that's better than a flat structure:

`{:WakuV2Fleet : "status.prod", :rootDirectory ...}`

So we are not exposing the structure of our internal config to the client, but still allow to customize some (not all, that's the drawback I guess) fields

ilmotta: >So we are not exposing the structure of our internal config to the client, but still allow to customize some (not all, that's the drawback I guess) fields

Good point, we can create an interface to clients that's more stable than our internal representation.

My only gripe with JSON is that it is kind of a bad media for human consumption, and IMO should be generally relegated for system-to-system communication. Something like a git config file (INI) or [TOML](https://toml.io/en/) would be nice, since the file can be nicely documented for humans.

Samyoul: So I don't have an objection to TOML as a input format, except that all other inputs are done via JSON. However we could create an endpoint that is agnostic to the input format until the point of inspection, then will throw an error if the input is not JSON or TOML.

---

> I can't think that's better than a flat structure:
>
>`{:WakuV2Fleet : "status.prod", :rootDirectory ...}`

Ah we can build a marshaller for this functionality. Provided that the fields present in the input match tags or field names of the go struct. I can make a mock up of this.

Samyoul: Ok, I've done some noodling. The process was slightly more work than I first thought, but I think I have a solutions:

So the first thing that we need to do is inflate a flat list, presuming the input is a flat list ...

```go
func inflateFlat(list map[string]any) map[string]any {
	nm := make(map[string]any)
	for k, v := range list {
		ks := strings.Split(k, ".")
		cm := nm
		for i := 0; i < len(ks)-1; i++ {
			if _, ok := cm[ks[i]]; !ok {
				cm[ks[i]] = make(map[string]interface{})
			}
			cm = cm[ks[i]].(map[string]interface{})
		}
		cm[ks[len(ks)-1]] = v
	}

	return nm
}
```

When ran within the following code:

```go
func TestInflateFlat(t *testing.T) {
	flatJSON := []byte(`{
"Awoken":true,
"Sacrifices":["Our Dignity"],
"SummoningRitual.Type":"Disco dance-off",
"Order.Name":"The Political Party You Don't Like",
"SummoningRitual.Order.Sign":"A blue orb"}`)
	holder := make(map[string]any)

	err := json.Unmarshal(flatJSON, &holder)
	if err != nil {
		fmt.Printf("Err: %s\n", err.Error())
	}
	ih := inflateFlat(holder)
	fmt.Printf("Result: %#v\n", ih)
}
```

We get the following output

```text
=== RUN   TestInflateFlat
Result: map[string]interface {}{
  "Awoken":true,
  "Order":map[string]interface {}{
    "Name":"The Political Party You Don't Like"
  },
  "Sacrifices":[]interface {}{
    "Our Dignity"
  },
  "SummoningRitual":map[string]interface {}{
    "Order":map[string]interface {}{
      "Sign":"A blue orb"
    },
    "Type":"Disco dance-off"
  }
}
--- PASS: TestInflateFlat (0.00s)
PASS
```

So you can see we inflate this code, but the next step is to map the inflated map on to our struct.

Samyoul:
So inflating was something I did second, because I originally did the following, and attempted to inflate during this function, but it was getting to complex. So the below code only does mapping of map fields to struct fields:

Note this function ignores any field that is not a struct, which I think is right, we don't want to update specific elements of a map or array/slice.

```go
func walkThroughUpdateStructRecursive(config map[string]any, val reflect.Value) {
	switch val.Kind() {
	case reflect.Struct:
		for i := 0; i < val.NumField(); i++ {
			field := val.Field(i)
			name := val.Type().Field(i).Name
			fmt.Printf("Field name: %s\n", name)

			for n, v := range config {
				ns := strings.Split(n, ".")
				if ns[0] == name {
					fmt.Printf("config name matches: %v\n", v)
					if len(ns) == 1 {
						field.Set(reflect.ValueOf(v))
					}
					if len(ns) > 1 {
						nc := config
						nc[ns[0]] = config[n]
						delete(nc, n)
					}
				}
			}

			walkThroughUpdateStructRecursive(config, field)
		}
	}
}
```

Then we can tie the walkThrough func and the inflate func together:

```go

func FlatUpdate(input []byte, v any) error {
	m := make(map[string]any)
	err := json.Unmarshal(input, &m)
	if err != nil {
		return err
	}
	spew.Dump(m)

	im := inflateFlat(m)
	spew.Dump(im)

	walkThroughStructFieldsRecursive(im, reflect.ValueOf(v).Elem())

	return nil
}
```

Then we can tie in the `FlatUpdate` func with a test similar to the one I shared here https://github.com/status-im/status-go/pull/3297#discussion_r1140148802

```go
func TestJSONMarshalling(t *testing.T) {
	defaultCthulhu := &Cthulhu{
		AeonsSlumbering:      6,
		SummoningIncantation: "ph'nglui mglw'nafh Cthulhu R'lyeh wgah'nagl fhtagn",
		Awoken:               false,
		PlanesOccupied:       -3,
		Sacrifices:           []string{"Sam", "Andrea", "Vitaliy"},
		SummoningRitual: Ritual{
			Name:        "Awakening of Great Old One",
			Type:        "sacrifice",
			Ingredients: []string{"goat", "migo wing", "shoggoth ooze"},
			Order: Order{
				Name: "Esoteric Order of Dagon",
				Sign: "Golden Tiara",
			},
		},
	}

	incomingJSON := []byte(`{"Awoken":true,"Sacrifices":["Our Dignity"]}`)

	err := json.Unmarshal(incomingJSON, defaultCthulhu)
	if err != nil {
		fmt.Println(err.Error())
	}
	fmt.Printf("%+v\n", defaultCthulhu)

	flatJSON := []byte(`{"SummoningRitual.Type":"Disco dance-off",}`)

	err = FlatUpdate(flatJSON, defaultCthulhu)
	if err != nil {
		fmt.Println(err.Error())
	}
	fmt.Printf("%+v\n", defaultCthulhu)
}
```

⚠️ hmm, there is a bug : 🐞 ⚠️ . Give me a second. I need to typecast the field update types.

ilmotta:
Very interesting approach @Samyoul. The dot syntax reminds of Java's property files. Just curious, have you seen this inflating approach in other projects?

In case you're interested, I found this [example TOML](https://github.com/BurntSushi/toml/blob/master/_example/example.toml) and the accompanying Go code is relatively clean https://github.com/BurntSushi/toml/blob/master/_example/example.go. The example is from the library https://github.com/BurntSushi/toml

cammellos:
I am biased and you know already my opinion :), but allowing the user to override any value is not worth the complexity in my opinion, and we'd be still bleeding out implementation details (i.e the shape of NodeConfig), I think we should simplify the external API, and provide a set of well defined configuration options to clients, as opposed of what we do currently which is similar, albeit you have to specify the whole thing.
That way we are free to simplify internally, but keep the same API interface from the outside (at that point, does not matter whether we want to store internally the config as TOML, JSON etc)

Samyoul:
Ah @cammellos , but we don't bleed out the implementation details. For example:

```json
"Order.Name":"The Political Party You Don't Like"
```

This updates the nested Order struct

Samyoul:
But, if you have preference, then do that. I was under the impression that you were open to the idea of an overridable config that doesn't expose the structure of the entire config. The solution I'm presenting does that.

You can list different deeply nested structs in a simple flat list without needing to know their position in the nesting scheme.

cammellos:
Correct, which bleeds out that `Order` is a struct, that has a `Name` field, as opposed to provide an option to clients, say:

```
{OrderName: "order-name"}
```

and then we parse it, and we can store it internally in:

```
{"Order": {"Name": "x"}}
```
Or, if we decide to change it, in something different:

```
{"InternalConfig": {"OrderName": y}}
```

While the method above, ties internal representation of the config to the external API.

> I was under the impression that you were open to the idea of an overridable config that doesn't expose the structure of the entire config. 

I was, you are correct, I thought it was good stepping stone, but after working on it a bit, the config structure is not really rational in status-go, it's a lot of historic baggage and I think providing a cleaner/tailored API it's more valuable, since already destkop/mobile have slight different behavior because they pass different configs and I think that's the most likely issue that we'll be facing, and it helps us to know what clients can specify, to better handle defaults and migrations.

Apologies if I wasn't more explicit about the change of mind, this was the comment where I mentioned it https://github.com/status-im/status-go/pull/3297#discussion_r1140319002 , " for overriding the JSON, in theory works ok, and that was our initial idea with @Samyoul , but ...", sorry I should have been clearer on it

ilmotta:
> Correct, which bleeds out that Order is a struct, that has a Name field, as opposed to provide an option to clients, say:

IMO, as long as the exposed interface is stable, flattening or not the structure is less important. There are countless tools that allow the user to group settings (like git), which may or may not "bleed" the internal structures (we never know). The point is that the user should never know if the configuration is nested by design or because it's tied to the implementation. We only know it's bleeding because we've read the code.

But as you said @cammellos https://github.com/status-im/status-go/pull/3297#issuecomment-1476063025 and I totally agree, perhaps we should first find a stable interface between clients, and only then make them more configurable via external files.

Samyoul:
Ok, that's completely fine. You have context over what you are working on and you know what you want.

I did read that comment and I took the following from it:

> , I can't think that's better than a flat structure:

So I made a solution that can handle a flat structure that can take `n` fields and update values of deeply nested fields, that the user does not need to know the position of.

The above example shows nested structs, `Order` is nested 2 levels in, but it could be 20 levels in and the code would still update the `Name` field of `Order`.

If you object to the `.` syntax, that can be adapted, `Order.Name` could be `OrderName` and we check the existence of a field called `OrderName` or `Order.Name`.

```go
	defaultCthulhu := &Cthulhu{
		AeonsSlumbering:      6,
		SummoningIncantation: "ph'nglui mglw'nafh Cthulhu R'lyeh wgah'nagl fhtagn",
		Awoken:               false,
		PlanesOccupied:       -3,
		Sacrifices:           []string{"Sam", "Andrea", "Vitaliy"},
		SummoningRitual: Ritual{
			Name:        "Awakening of Great Old One",
			Type:        "sacrifice",
			Ingredients: []string{"goat", "migo wing", "shoggoth ooze"},
			Order: Order{
				Name: "Esoteric Order of Dagon",
				Sign: "Golden Tiara",
				Helicopter: Helicopter {
					Spaceship: Spaceship {
						Name: "whizzy woo",
						Planet: Planet{
							Name: "Wolf 358 C",
							Civilisation: Civilisation {
								FoundFire: true,
							},
						},
					},
					Goat: "baaah", 
				},
			},
		},
	}
```

The following input , would update the structs as needed. Notice that `CivilisationFoundFire` lives very deeply nested, but its position is not required only that it exists:

```json
{
  "CivilisationFoundFire":false,
  "HelicopterGoat": "mooo",
  "Awoken": true,
  "OrderName": "Status IM"
}
```

This point:

> While the method above, ties internal representation of the config to the external API.

I think I may not have explained myself very well. The code doesn't care where the fields are or what the internal structure is the config input can be completely flexible and is basically uncoupled from the internal structure.

PR #3308, opened 2023-03-17
Check token funds when handling community requests to join

0x-r4bbit:
This adds checks to `HandleCommunityRequestToJoin` and `AcceptRequestToJoinCommunity` that ensure a given user's revealed wallet addresses own the token funds required by a community.

When community has token permissions of type `BECOME_MEMBER`, the following happens when the owner receives a request:

1. Upon verifying provided wallet addresses by the requester, the owner node accumulates all token funds related to the given wallets that match the token criteria in the configured permissions
2. If the requester does not meet the necessary requirements, the request to join will be declined. If the requester does have the funds, he'll either be automatically accepted to the community, or enters the next stage where an owner needs to manually accept the request.
3. The the community does not automatically accept users, then the funds check will happen again, when the owner tries to manually accept the request. If the necessary funds do not exist at this stage, the request will be declined
4. Upon accepting, whether automatically or manually, the owner adds the requester's wallet addresses to the `CommunityDescription`, such that they can be retrieved later when doing periodic checks or when permissions have changed.


protocol/protobuf/communities.proto
```diff
@@ -88,6 +93,7 @@ message CommunityDescription {
   repeated string tags = 14;
   map<string, CommunityTokenPermission> token_permissions = 15;
   repeated CommunityTokenMetadata community_tokens_metadata = 16;
+  map<string, MemberWallet> member_wallets = 17;
```
0x-r4bbit: Realized that we don't necessarily need a new `MemberWallets` field, could also leverage `CommunityMember` instead. Will update this.

protocol/communities/manager.go
```diff
@@ -83,7 +90,7 @@ func (t *HistoryArchiveDownloadTask) Cancel() {
 	close(t.CancelChan)
 }
 
-func NewManager(identity *ecdsa.PrivateKey, db *sql.DB, encryptor *encryption.Protocol, logger *zap.Logger, verifier *ens.Verifier, accountsManager *account.GethManager, transport *transport.Transport, torrentConfig *params.TorrentConfig) (*Manager, error) {
+func NewManager(identity *ecdsa.PrivateKey, db *sql.DB, encryptor *encryption.Protocol, logger *zap.Logger, verifier *ens.Verifier, accountsManager *account.GethManager, tokenManager *token.Manager, transport *transport.Transport, torrentConfig *params.TorrentConfig) (*Manager, error) {
```
Samyoul: I notice that this function has a very long signature. I wonder if this is a sign that the function is growing too complex or the method of interfacing it needs to be simplified.

See Go style guide on this subject https://google.github.io/styleguide/go/best-practices.html#function-argument-lists . One potential solution is to use an Option Structure see https://google.github.io/styleguide/go/best-practices.html#option-structure.

In fact `NewMessenger` already uses a variadic parameter for Option structs.

protocol/messenger.go
```diff
@@ -241,6 +243,7 @@ func NewMessenger(
 	installationID string,
 	peerStore *mailservers.PeerStore,
 	accountsManager *account.GethManager,
+	rpcClient *rpc.Client,
 	opts ...Option,
```
Samyoul: Here is the use of `Option` structs in `NewMessenger`, this prevents the bloating of the function signature and allows for a more configurable struct with `Option`al parameters.

protocol/messenger_sync_settings_test.go
```diff
@@ -114,6 +114,7 @@ func (s *MessengerSyncSettingsSuite) newMessengerWithOptions(shh types.Waku, pri
 		uuid.New().String(),
 		nil,
 		nil,
+		nil,
 		options...,
```
Samyoul: Just a little point. When you start to see lots of `nil`s when calling a "New" function, this means you should rethink the function signature, as there may be unnecessary complexity or you require a different initialisation pattern.

Samyoul:
@0x-r4bbit good work on this PR, I'm sorry I missed the last PR this PR is built on top of. I hope that my feedback on this PR seems reasonable and actionable. If you have any questions or you simply disagree 😆 just let me know.

0x-r4bbit:
Thank you @Samyoul !
I'll see if I can simplify the signature. Appreciate your feedback a lot here

protocol/communities/manager.go
0x-r4bbit: Applied this pattern in the previous PR and updated this one accordingly as well

0x-r4bbit:
@Samyoul updated this PR according to your comments!

endulab (approved):

PR #3309, opened 2023-03-17
rewrite mention in go

### LINKED ISSUE #15220: Move mentions logic to status-go
Currently all mentions related logic in react, suggestions list, replacement in the text-input, and in the messages, we should move this to status-go

current implementation is here https://github.com/status-im/status-mobile/blob/develop/src/status_im/chat/models/mentions.cljs

qfrank:
The main purpose of this PR is to rewrite the mention functionality, which was originally implemented in ClojureScript, using Golang. Currently, no optimizations have been made. If performance issues are discovered during client testing based on the current rewrite, I will submit a separate PR for optimization.

close [#15220](https://github.com/status-im/status-mobile/issues/15220)

status: ready.


server/server_test.go
```diff
@@ -162,10 +162,11 @@ func (s *ServerURLSuite) TestQRCodeGeneration() {
...
 
 	_ = rootCAs.AppendCertsFromPEM(certPem)
+	// #nosec G402
 	tr := &http.Transport{
 		TLSClientConfig: &tls.Config{
 			MinVersion:         tls.VersionTLS12,
-			InsecureSkipVerify: true, //nolint:all // MUST BE FALSE
+			InsecureSkipVerify: true,
 			RootCAs:            rootCAs,
 		},
```
Samyoul: Does this PR need a rebase because the changes you made to this PR https://github.com/status-im/status-go/pull/3332 don't seem to be present.

Samyoul (approved):
Looks great

server/server_test.go
qfrank: > Does this PR need a rebase because the changes you made to this PR #3332 don't seem to be present.

yeah, not rebased before your comment, just rebased. 🙂

protocol/messenger_mention.go
```diff
@@ -0,0 +1,1115 @@
...
+}
+
+type MentionableUser struct {
+	*Contact
+
+	primaryName       string
+	secondaryName     string
+	searchablePhrases []string
+
+	Key          string // used by :key-fn in react
```
ilmotta: The comment is too specific for React, but status-go doesn't care about the frontend framework in use. Maybe just saying `Key` is a unique identifier of a mentionable user is enough, or renaming to `ID` would be enough.

protocol/messenger_mention.go
```diff
@@ -0,0 +1,1115 @@
...
+	contactJSONMap["primaryName"] = c.PrimaryName()
+	contactJSONMap["secondaryName"] = c.SecondaryName()
+	contactJSONMap["ensVerified"] = c.ENSVerified
+	contactJSONMap["added"] = c.added()
+	contactJSONMap["displayName"] = c.GetDisplayName()
+	contactJSONMap["key"] = c.Key
+	contactJSONMap["match"] = c.Match
+	contactJSONMap["searchedText"] = c.SearchedText
+	contactJSONMap["id"] = c.ID
+	compressedKey, err := multiformat.SerializeLegacyKey(c.ID)
```
ilmotta, line 72: I think the code could be improved a little bit if we defined a named struct or if we used an inline struct (with/without JSON tags). We don't need to give up on type safety by using `interface{}`. We can also move the compressed key line to the top, so we can fail fast.

protocol/messenger_mention.go
```diff
@@ -0,0 +1,1115 @@
...
+	MentionEnd   int
+	PreviousText string // searched text
+	NewText      string // the matched username
+	Start        int    // position after the @
+	End          int    // position of the end of newText
+}
+
+type ChatMentionContext struct {
+	ChatID string
+	//[:chat/inputs-with-mentions chat-id]
```
ilmotta: This comment doesn't make sense for somebody who doesn't know Clojure and particularly the re-frame app db state in `status-mobile`. I think we should remove them or update them to something agnostic of the frontend framework.

protocol/messenger_mention.go
```diff
@@ -0,0 +1,1115 @@
...
+	m.clearSuggestions(chatID)
+}
+
+func (m *MentionManager) CheckSelection(chatID, text string, start int, end int) *ChatMentionContext {
+	ctx := m.getChatMentionContext(chatID)
+	m.checkSelection(chatID, text, start, end, ctx.MentionSuggestions)
+	return ctx
+}
+
+func (m *MentionManager) checkSelection(chatID, text string, start int, end int, mentionableUsers map[string]*MentionableUser) {
```
ilmotta: I have trouble understanding the meaning of the word `check` in `checkSelection` and in many other functions that use this word in this PR (`checkIdxForMentions`, `checkStyleTag`, etc). Actually, I have to confess I don't understand them at all just by their names. I see the original names came from the Clojure code, but *check* to me is sort of related to validation, but that's not what none of these functions are doing.

protocol/messenger_mention.go
```diff
@@ -0,0 +1,1115 @@
...
+	}
+	return nil
+}
+
+type AtIndexEntry struct {
+	From    int
+	To      int
+	Checked bool
+
+	Mentioned bool // corresponding :mention?
```
ilmotta: These 2 comments seem superfluous and related to the Clojure code. What does this comment mean to somebody working in the Desktop client?

ilmotta (approved):
It's a ton of work in here @qfrank, happy to see so much stuff getting out of the mobile client. I left only minor comments.

I'm wondering if the implementation isn't too difficult to grasp for the Desktop team to use in the future. The endpoints have no explanation/documentation and seem to be tied to what the mobile client does nowadays. Since we want to avoid client specific code in status-go, I'm curious to know if the endpoints available in this PR will serve the needs of the Desktop client, because they should, at least for the most part. Do you know anything about this or have you checked with the Desktop team how they envision mentions could/should be added to `status-go`?

Call me crazy, but before opening this PR I thought I would see some kind of parser that would transform the input text to data structures containing mentions, etc and that the code would always walk and manipulate the data structures, not manipulate indexes and strings as it does in many places. Then the parsed structure would be encoded back to clients, so they can efficiently decorate certain parts with their own styles, colors, etc. But all of this would require a rewrite from scratch and you wouldn't be able to reuse much of the existing Clojure code to translate to Go.

I know, delusional thoughts of mine! Maybe an interesting idea for another day ;) Maybe we should one day unify how any valid chat input text is represented structurally, for any Status client (even a hypothetical terminal client), like an efficient two-way representation of `chat input text <-> data`.

We're getting there with this PR. Great work o/

qfrank:
> have you checked with the Desktop team how they envision mentions could/should be added to status-go

thank you for your patient review(it's complex 😅), i didn't checked yet, my thought was: this is our first rewrite for mention, i just want to make sure it works as before in status-mobile. for your concern, i thought it should be in the next/future iteration. @ilmotta

protocol/messenger_mention.go
qfrank: i asked chatgpt4 to give a better name, and it said `checkIdxForMentions`/`checkStyleTag`/`checkEntry` looks good 😂
> The name of the function checkXXX already conveys its purpose 

checkSelection -> handleSelectionChange agree?

ilmotta:
> > have you checked with the Desktop team how they envision mentions could/should be added to status-go
> 
> thank you for your patient review(it's complex sweat_smile), i didn't checked yet, my thought was: this is our first rewrite for mention, i just want to make sure it works as before in status-mobile. for your concern, i thought it should be in the next/future iteration. @ilmotta

I fully agree, first make things work! :rocket:

protocol/messenger_mention.go
qfrank: removed

protocol/messenger_mention.go
qfrank: removed

protocol/messenger_mention.go
qfrank: fixed

protocol/messenger_mention.go
qfrank: fixed

protocol/messenger_mention.go
ilmotta: huahauah :rofl: Perhaps if ChatGPT knew both the mobile and status-go codebases and could be given a detailed Github issue it would figure out better names for us.

I guess I only have a general feedback @qfrank. If giving the name to a function is difficult, then maybe the current name is fine or the function is doing too many things. Your suggestion seems clear to me :)

For the others it's a case by case thing. `CheckMentions` main goal is to replace @ signs with mentionable users, so the name could be `ReplaceWithMentionableUsers`, or `ReplaceMentions`. In this case *replace* seems more clear to me than *check*. `RecheckAtIdxs` does things differently, and it could be just `UpdateAtIdxs` or `RecalculateMentionIdxs`, etc.

Check can be a synonym to verify, assure, ascertain, and all of them give me a vibe of a pure operation, whereas *replace*, *update*, *recalculate*, etc seem a bit more precise about the mutability nature of these functions.

But don't worry too much, these might just be my own personal bias.

PR #3310, opened 2023-03-17
Connection state changes now correctly propagates to messenger

vkjr:
status-mobile has an [issue](https://github.com/status-im/status-mobile/issues/15087) with messages marked as "sent" even when they are sent while the internet connection is off.

Turned out that `ConnectionChanged()` wasn't invoked correctly

[Related status-mobile PR](https://github.com/status-im/status-mobile/pull/15148)

### DIFF
--- a/node/get_status_node.go
+++ b/node/get_status_node.go
@@ -615,11 +615,13 @@ func (n *StatusNode) PeerCount() int {
 }
 
 func (n *StatusNode) ConnectionChanged(state connection.State) {
-	if n.wakuExtSrvc == nil {
-		return
+	if n.wakuExtSrvc != nil {
+		n.wakuExtSrvc.ConnectionChanged(state)
 	}
 
-	n.wakuExtSrvc.ConnectionChanged(state)
+	if n.wakuV2ExtSrvc != nil {
+		n.wakuV2ExtSrvc.ConnectionChanged(state)
+	}
 }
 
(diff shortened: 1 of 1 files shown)


ghost:
### Pull Request Checklist

node/get_status_node.go
```diff
@@ -615,11 +615,13 @@ func (n *StatusNode) PeerCount() int {
...
-	if n.wakuExtSrvc == nil {
-		return
+	if n.wakuExtSrvc != nil {
+		n.wakuExtSrvc.ConnectionChanged(state)
 	}
 
-	n.wakuExtSrvc.ConnectionChanged(state)
+	if n.wakuV2ExtSrvc != nil {
+		n.wakuV2ExtSrvc.ConnectionChanged(state)
+	}
```
Samyoul, line 624: Should we default to wakuV2? By putting that check first?

cammellos: They can't be both active, so I think the order is not relevant

vkjr: Since it is not `else-if`, it should work in any case, I believe. Do you think there is any benefit in changing the order?

Samyoul: Ah, if they can't both be active then the order is not relevant.

Samyoul (approved):

siddarthkay (approved):

PR #3311, opened 2023-03-17
Prevent receving messages from removed contact in 1-1 chat

rasom:
needed https://github.com/status-im/status-mobile/issues/15149

That fixes the issue although I'm not 100% sure that we want to have this behaviour.

status: ready


cammellos:
@rasom I'd hold on for this one, until things have settled, seems like for now we want to err on the side of allowing users to communicate instead of preventing communications

rasom:
@cammellos understood, closing it

PR #3312, opened 2023-03-17
Support display name/images/password

cammellos:
Minor changes to support password/display-name/image when creating an account.
I have added a couple of endpoint to collate more login logic in single calls.


api/geth_backend.go
```diff
@@ -776,6 +793,17 @@ func (b *GethStatusBackend) CreateAccountAndLogin(request *requests.CreateAccoun
 		b.log.Error("start node", err)
 		return err
 	}
+
+	iis, err := images.GenerateIdentityImages(request.ImagePath, 0, 0, 1000, 1000)
```
J-Son89, line 798: what's iis?

J-Son89 (approved):

api/geth_backend.go
cammellos: images, also I probably need to put those consts somewhere

ilmotta: @J-Son89, whether we like it or not, in Go it's idiomatic to use very short names. When looked in isolation, unfortunately these short names can be cryptic and I find necessary to sometimes (eye) jump to the declaration right side to see the actual function name because the variable name is sometimes useless.

There's a very specific method to madness though: https://google.github.io/styleguide/go/decisions#variable-names

ilmotta (approved):

api/geth_backend.go
```diff
@@ -153,6 +154,16 @@ func (b *GethStatusBackend) GetMultiaccountDB() *multiaccounts.Database {
 	return b.multiaccountsDB
 }
 
+func (b *GethStatusBackend) InitializeAccounts(rootDirectory string) error {
+	b.UpdateRootDataDir(rootDirectory)
+	manager := b.AccountManager()
+	keystoreDir := filepath.Join(rootDirectory, "keystore")
```
ilmotta: I remember a recent PR from @Samyoul also had this "keystore" magic string. Could we extract it into a single place?

api/geth_backend.go
J-Son89: Thanks for the clarification @ilmotta. I would still suggest imgs is a better short alternative in this instance. 
Ultimately I have no strong opinions of how the code in status go looks/ is written as it's all completely new to me :)

api/geth_backend.go
Samyoul: We do have `common` packages that are used for these cases. At the moment my #3248 is still in testing on the client sides

api/geth_backend.go
Samyoul: @cammellos these dimensions `0, 0, 1000, 1000`, seem larger than we normally use, which is fine. FYI we do have dimension consts:

https://github.com/status-im/status-go/blob/45b287370a57b5b9aa374a1c628190c53a41722c/images/meta.go#L13-L26

Also you may want to consider the quality to file size calculations we do with the existing dimensions if you are introducing a larger image size type:

https://github.com/status-im/status-go/blob/45b287370a57b5b9aa374a1c628190c53a41722c/images/meta.go#L34-L43

cammellos: mmh, I will keep the ones used in the frontend for now just to make sure I am not breaking anything, but I will use the const once code has settled, thanks!

PR #3317, opened 2023-03-20
feat: check member permission funds periodically

0x-r4bbit:
This adds a periodic member permission check for every admin community such that member funds are checked every hour.

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -49,27 +49,30 @@ var pieceLength = 100 * 1024
 
 const maxArchiveSizeInBytes = 30000000
 
+var memberPermissionsCheckInterval = 1 * time.Hour
+
 var ErrTorrentTimedout = errors.New("torrent has timed out")
 
 type Manager struct {
-	persistence                  *Persistence
-	encryptor                    *encryption.Protocol
-	ensSubscription              chan []*ens.VerificationRecord
-	subscriptions                []chan *Subscription
-	ensVerifier                  *ens.Verifier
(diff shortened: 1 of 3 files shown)


protocol/communities/manager.go
```diff
@@ -47,27 +47,30 @@ var pieceLength = 100 * 1024
 
 const maxArchiveSizeInBytes = 30000000
 
+var memberPermissionsCheckInterval = 1 * time.Hour
```
0x-r4bbit, line 52: Wonder if we should go with a different interval here

Samyoul: When implementing these kind of things you just pull a figure out of the air, it can be any reasonable number.

In this case, the functionality (If I'm reading it correctly) checks if a member's list contains members that have permission to be members, if not those members are removed.

So you are basically you are running a purge hourly. Is it possible to purge anytime the membership list is updated? Is that possible?

0x-r4bbit: Thank you for sharing your thoughts on this Sam.
So the check is already happening when the membership list is updated (on handle request to join).

However, because the state of any member's wallet can change at any given time really, we have to regularly check on-chain if the all the required funds are still in each member's wallet.

In other words, with this we'd query various chains every hour.

Samyoul: Ah, that is a very fair point. I was assuming permissions were soul bound but they may not be.

0x-r4bbit: Humans don't have souls.

Samyoul (approved):

PR #3318, opened 2023-03-20
feat: event on error

alaibe:
Use event to sent signal as connection error

DO NOT MERGE until desktop is ready

### DIFF
--- a/rpc/chain/client.go
+++ b/rpc/chain/client.go
@@ -34,10 +34,9 @@ type ClientWithFallback struct {
 
 	WalletNotifier func(chainId uint64, message string)
 
-	IsConnected             bool
-	consecutiveFailureCount int
-	IsConnectedLock         sync.RWMutex
-	LastCheckedAt           int64
+	IsConnected     bool
+	IsConnectedLock sync.RWMutex
+	LastCheckedAt   int64
 }
 
 var vmErrors = []error{
@@ -71,14 +70,13 @@ func NewSimpleClient(main *rpc.Client, chainID uint64) *ClientWithFallback {
 	})
 
 	return &ClientWithFallback{
-		ChainID:                 chainID,
-		main:                    ethclient.NewClient(main),
-		fallback:                nil,
-		mainRPC:                 main,
-		fallbackRPC:             nil,
-		IsConnected:             true,
-		consecutiveFailureCount: 0,
-		LastCheckedAt:           time.Now().Unix(),
+		ChainID:       chainID,
+		main:          ethclient.NewClient(main),
+		fallback:      nil,
+		mainRPC:       main,
+		fallbackRPC:   nil,
+		IsConnected:   true,
+		LastCheckedAt: time.Now().Unix(),
 	}
 }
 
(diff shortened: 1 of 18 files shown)


ghost:
### Pull Request Checklist

services/wallet/thirdparty/cryptocompare/client.go
```diff
@@ -51,7 +51,7 @@ func (c *Client) DoQuery(url string) (*http.Response, error) {
 }
 
 func (c *Client) FetchPrices(symbols []string, currencies []string) (map[string]map[string]float64, error) {
-	chunks := utils.ChunkSymbols(symbols)
+	chunks := utils.ChunkSymbols(symbols, 60)
```
Khushboo-dev-cpp, line 54: why is this 60? just for me to understand

Khushboo-dev-cpp (approved):
LGTM :)

services/wallet/thirdparty/cryptocompare/client.go
alaibe: there is a limit on the api regarding the number of character, and I counted 5 character per symbol: 5*60 => 300

services/wallet/reader.go
```diff
@@ -210,13 +197,17 @@ func (r *Reader) GetWalletToken(ctx context.Context, addresses []common.Address)
...
-		clients, err := r.rpcClient.EthClients(chainIDs)
-		if err != nil {
-			return err
-		}
 		balances, err = r.tokenManager.GetBalancesByChain(ctx, clients, addresses, tokenAddresses)
 		if err != nil {
+			for _, client := range clients {
+				client.SetIsConnected(false)
+			}
+			log.Info("tokenManager.GetBalancesByChain err", err)
```
dlipicar, line 210: Do we want to keep this?

rpc/chain/client.go
```diff
@@ -118,6 +122,28 @@ func isVMError(err error) bool {
...
+				c.WalletNotifier(c.ChainID, "down")
+			}
+			c.IsConnected = false
+		}
+
+	} else {
+		if !c.IsConnected {
+			c.IsConnected = true
+			if c.WalletNotifier != nil {
+				c.WalletNotifier(c.ChainID, "up")
```
dlipicar, line 141: These "up" and "down" strings are defined in several different places. Could we have constants defined in a single spot?

services/wallet/reader.go
dlipicar: (And have it with log level `Info`)

services/wallet/thirdparty/cryptocompare/client.go
dlipicar: A constant defined at the top would be clearer

services/wallet/reader.go
alaibe: yeah, similar to other places, we can tweak later

rpc/chain/client.go
alaibe: I ll update with constants

PR #3321, opened 2023-03-21
Add restore account endpoint

cammellos:
Add restore account endpoint, reusing most of the logic from create account.


protocol/requests/create_account.go
```diff
@@ -32,6 +32,10 @@ type CreateAccount struct {
 }
 
 func (c *CreateAccount) Validate() error {
+	return ValidateAccountCreationRequest(*c)
+}
+
+func ValidateAccountCreationRequest(c CreateAccount) error {
 	// TODO(cammellos): Add proper validation for password/displayname/etc
```
J-Son89, line 39: does this TODO have an issue alongside it?

J-Son89 (approved):

protocol/requests/create_account.go
cammellos: No, but I can create one

protocol/requests/restore_account.go
```diff
@@ -0,0 +1,20 @@
...
+
+var ErrRestoreAccountInvalidMnemonic = errors.New("restore-account: invalid mnemonic")
+
+type RestoreAccount struct {
+	Mnemonic string `json:"mnemonic"`
+	CreateAccount
+}
+
+func (c *RestoreAccount) Validate() error {
+	if len(c.Mnemonic) == 0 {
```
ilmotta, line 15: It'd be good if we could have some tests for `restore_account#Validate` and `create_account#Validate`, maybe not in this PR if it's more urgent to merge the fix.

ilmotta (approved):

PR #3322, opened 2023-03-21
Local Pairing: Reconcile A Tiny Number of Duplicate Funcs

### LINKED ISSUE #3300: Reconcile duplicate functions in `server/pairing/certs.go` and in `server/certs.go`
Reconcile duplicate functions in [server/pairing/certs.go](https://github.com/status-im/status-go/pull/3248/files#diff-882b3a343179b11ddab5a62baaca80207f6136b1fa5c731578ab3bc1d88e6020) and in server/certs.go. There are a number of functions that are identical between these two file and it would make the `server` packages DRYer and less confusing over which function to use in what context.

Pay attention to avoid introducing import loops.

Samyoul:
## What's Changed?

Removed a few duplicate functions, really minor stuff actually. When I first spotted this I thought it was going to be a lot more work, but it wasn't any work at all.

## Why Change?

Because duplicate functions lead to divergence of implementation but convergence of functionality, which make finding bugs harder.

If you are curious about the `servertest` package see https://pkg.go.dev/net/http/httptest as an example for how the stn lib handles test util packages

Closes https://github.com/status-im/status-go/issues/3300

### DIFF
--- a/server/certs.go
+++ b/server/certs.go
@@ -4,7 +4,6 @@ import (
 	"crypto/ecdsa"
 	"crypto/elliptic"
 	"crypto/rand"
-	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
 	"crypto/x509/pkix"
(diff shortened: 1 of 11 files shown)


qfrank (approved):
interesting solution 👍

PR #3324, opened 2023-03-21
feat: fetch NFT metadata from Communities

### LINKED ISSUE #9941: Support fetching Community collectibles
Collectibles used for community access do not follow the standard `tokenURI` metadata format. That means OpenSea will return empty metadata and won't provide the link.
In such cases, we need to get the `tokenURI` by interacting with the contract (see https://docs.infura.io/infura/tutorials/ethereum/retrieve-and-display-erc-721-and-erc-1155-tokens). Prefer Pokt, use Infura as fallback.
Our `tokenURI` will contain some sort of identifier for the corresponding community. We must then fetch the metadata and the image from Waku. Gonna need some help with this last part @0x-r4bbit @endulab

dlipicar:
Closes https://github.com/status-im/status-desktop/issues/9941

In this PR, we attempt to fetch NFT metadata from a Community when:
- OpenSea does not return any metadata
- The NFT contract implements `IERC721Metadata`
- The `tokenURI` returns a string with format `compressedCommunityID/tokenID`

A wrapper for collectible fetching was added, which for the moment simply fetches from OpenSea.

Tested with a Community Token minted to my account (`0xa1E277eA6b97eFfc5b61B3BF5dE03F438981247E`)

image

image

### DIFF
--- a/node/status_node_services.go
+++ b/node/status_node_services.go
@@ -473,10 +473,17 @@ func (b *StatusNode) appmetricsService() common.StatusService {
 
 func (b *StatusNode) walletService(accountsDB *accounts.Database, accountsFeed *event.Feed, openseaAPIKey string) common.StatusService {
 	if b.walletSrvc == nil {
+		var extService *ext.Service
+		if b.WakuV2ExtService() != nil {
+			extService = b.WakuV2ExtService().Service
+		} else if b.WakuExtService() != nil {
+			extService = b.WakuExtService().Service
+		}
 		b.walletSrvc = wallet.NewService(
 			b.appDB, accountsDB, b.rpcClient, accountsFeed, openseaAPIKey, b.gethAccountManager, b.transactor, b.config,
 			b.ensService(),
 			b.stickersService(accountsDB),
+			extService,
 		)
 	}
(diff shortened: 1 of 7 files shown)


node/status_node_services.go
```diff
@@ -473,10 +473,17 @@ func (b *StatusNode) appmetricsService() common.StatusService {
 
 func (b *StatusNode) walletService(accountsDB *accounts.Database, accountsFeed *event.Feed, openseaAPIKey string) common.StatusService {
 	if b.walletSrvc == nil {
+		var extService *ext.Service
+		if b.WakuV2ExtService() != nil {
```
endulab, line 477: You check WakuV2ExtService twice

dlipicar: 🤦

dlipicar: BTW is this a proper approach? Not sure (no idea) what the deal is with WakuV1 and WakuV2

stefandunca: We don't use WakuV1 anymore, so I expect the latter check not to help.

stefandunca (approved):

endulab (approved):

PR #3331, opened 2023-03-22
Add customization color

cammellos:
Adds customization color and rename british spelling to american spelling :gb: , I hope you will find it in your :heart: to forgive me @Samyoul , we went for strings after a discussion with @ilmotta , but neither of us has a strong opinion.


ilmotta (approved):
@Samyoul, as @cammellos said, I have no strong opinion too, but I've read about arguments against iota when the values are exposed to consumers, which made sense to me. `iota` has many use cases where it's useful and valid, the deeper question is, when should we use iota in `status-go`, or in Go in general? Would love to hear your thoughts on the matter.

I'll try to explain some of my thoughts:

Most of the usages of iota in status-go are to generate values from 0/1 to N to simulate an "enum". But once we expose those values to clients, it means status-go needs to be very careful in how the *enum* evolves. Removing a constant requires adding `_`, which is somewhat easy to miss. If we do miss, then status-go would happily and silently expose the incorrect values to clients with potential bad effects to users. There's also a more problematic case where the values could be stored incorrectly in the database. Although these problems should be rare or never happen, it's something to consider. Unit tests can greatly reduce the chance of these bugs.

There are other arguments, some related to reduced type safety in some cases. I found the most interesting discussions on Reddit, since the internet of good quality blog posts seem to be surely dying.

At least for the mobile client, there's also a good advantage to strings (or keywords) in that when we inspect the application state, it's evident what the enumeration value means because it's not just an integer. To get this same benefit the client would need to do the transformation back from ints to strings, thus sort of duplicating the enumeration from `status-go`.

This quote is interesting:

> Don’t use iota for defining constants where its values are explicitly defined (elsewhere).
For example, when implementing parts of a specification and the specification says
which values are assigned to which constants, you should explicitly write the constant
values. Use iota for “internal” purposes only. That is, where the constants are referred
to by name rather than by value. That way you can optimally enjoy iota by inserting
new constants at any moment in time / location in the list without the risk of breaking
everything.
>
> -- Danny van Heumen

Samyoul:
## Defilement of the one true iNgRiSh

`Color` is a re-latinisation of the Anglo-Norman word `Colour`, it is a relatively modern change by American linguists. The root word for `Colour` is in fact `Color` from Latin. I will forgive this, but just note that this spelling is only used in a single country, the rest of the English speaking world and nations use `Colour`.

However I propose a change for clarity `ColorHash` -> `HueMash` and completely de-latinise the word . 😆 

---

## Iota vs String



There is a common and std go tool that allows you to have both `iota` and `string` consts, the `stringer`.

See the go docs on it https://pkg.go.dev/golang.org/x/tools/cmd/stringer

A nice blog about the go `stringer` tool : https://arjunmahishi.medium.com/golang-stringer-ad01db65e306

Example :

Your code could look like below:

```go
//go:generate stringer -type CustomizationColor -trimprefix=CustomizationColor

type CustomizationColor int

const (
	CustomizationColorPrimary   CustomizationColor = iota + 1
	CustomizationColorPurple                       
	CustomizationColorIndigo                       
	CustomizationColorTurquoise                    
	CustomizationColorBlue                         
	CustomizationColorGreen                        
	CustomizationColorYellow                       
	CustomizationColorOrange                       
	CustomizationColorRed                          
	CustomizationColorPink                         
	CustomizationColorBrown                        
	CustomizationColorSky                          
	CustomizationColorArmy                         
	CustomizationColorMagenta                      
	CustomizationColorCopper                       
	CustomizationColorCamel                        
	CustomizationColorYin                          
	CustomizationColorYang                         
	CustomizationColorBeige                        
)
```

After make `make generate` there will be another file:

```go
// Code generated by "stringer -type CustomizationColor -trimprefix=CustomizationColor"; DO NOT EDIT.

package protocol

import "strconv"

func _() {
	// An "invalid array index" compiler error signifies that the constant values have changed.
	// Re-run the stringer command to generate them again.
	var x [1]struct{}
	_ = x[CustomizationColorPrimary-1]
	_ = x[CustomizationColorPurple-2]
	_ = x[CustomizationColorIndigo-3]
	_ = x[CustomizationColorTurquoise-4]
	_ = x[CustomizationColorBlue-5]
	_ = x[CustomizationColorGreen-6]
	_ = x[CustomizationColorYellow-7]
	_ = x[CustomizationColorOrange-8]
	_ = x[CustomizationColorRed-9]
	_ = x[CustomizationColorPink-10]
	_ = x[CustomizationColorBrown-11]
	_ = x[CustomizationColorSky-12]
	_ = x[CustomizationColorArmy-13]
	_ = x[CustomizationColorMagenta-14]
	_ = x[CustomizationColorCopper-15]
	_ = x[CustomizationColorCamel-16]
	_ = x[CustomizationColorYin-17]
	_ = x[CustomizationColorYang-18]
	_ = x[CustomizationColorBeige-19]
}

const _CustomizationColor_name = "PrimaryPurpleIndigoTurquoiseBlueGreenYellowOrangeRedPinkBrownSkyArmyMagentaCopperCamelYinYangBeige"

var _CustomizationColor_index = [...]uint8{0, 7, 13, 19, 28, 32, 37, 43, 49, 52, 56, 61, 64, 68, 75, 81, 86, 89, 93, 98}

func (i CustomizationColor) String() string {
	i -= 1
	if i < 0 || i >= CustomizationColor(len(_CustomizationColor_index)-1) {
		return "CustomizationColor(" + strconv.FormatInt(int64(i+1), 10) + ")"
	}
	return _CustomizationColor_name[_CustomizationColor_index[i]:_CustomizationColor_index[i+1]]
}
```

Then you can treat any `CustomizationColor` type like a string

```go
fmt.Printf("%s", CustomizationColorGreen)
```

Output:

```text
Green
```

See it run for your self https://go.dev/play/p/nOODLzWrggH

Samyoul:
Granted my suggestion doesn't entirely / directly address some of your most impactful concerns:

> Removing a constant requires adding _, which is somewhat easy to miss. If we do miss, then status-go would happily and silently expose the incorrect values to clients with potential bad effects to users. There's also a more problematic case where the values could be stored incorrectly in the database. 

I think the only thing that is missing would be the reversal, some kind of functionality that could turn a `stringer`ed enum back into an enum.

There is a such a tool, though not std : https://github.com/alvaroloes/enumer

qfrank (approved):
LGTM 🫶

ilmotta:
> Granted my suggestion doesn't entirely / directly address some of your most impactful concerns:
> 
> > Removing a constant requires adding _, which is somewhat easy to miss. If we do miss, then status-go would happily and silently expose the incorrect values to clients with potential bad effects to users. There's also a more problematic case where the values could be stored incorrectly in the database.
> 
> I think the only thing that is missing would be the reversal, some kind of functionality that could turn a `stringer`ed enum back into an enum.
> 
> There is a such a tool, though not std : https://github.com/alvaroloes/enumer

I didn't know about any of those libs @Samyoul, thanks a lot for the code snippet and playground :heart_decoration: 

Although certainly useful, I find `stringer` to be a bit too much when compared to simply typing the values of the constants, which is very explicit (a plus to me) and requires no learning on the dev part about how stringer will generate code. The risk of typos when manually typing enumeration values as the blog post comments is real, but that's true of every constant we type, so :man_shrugging: 

Maybe `stringer` and even `enumer` can be more useful with more complicated enumerations? It's interesting that `enumer` generates something that resembles Java enum's, e.g. to get all values.

But really, I'd be happy with any solution, we just have to be careful with `iota`s. Thanks for the tips @Samyoul

multiaccounts/database.go
```diff
@@ -12,30 +12,55 @@ import (
 	"github.com/status-im/status-go/sqlite"
 )
 
-type ColourHash [][2]int
+type ColorHash [][2]int
+
+type CustomizationColor string
+
+const (
+	CustomizationColorPrimary   = "primary"
```
ilmotta: @cammellos, I think we're missing the type in the constants, like:

```go
const (
	CustomizationColorPrimary CustomizationColor = "primary"
	CustomizationColorPurple  CustomizationColor = "purple"
)
```

PR #3332, opened 2023-03-23
fix: App crashes on syncing QR of the account

### LINKED ISSUE #15446: App crashes on syncing QR of the account that has previously been restored numerous times on different devices before.
cc @qfrank 

For reproduction use this seed phrase that belongs to account which has been restored lot of times on different devices: section involve corn song gold verb topic nurse morning dream snack very

**Steps:** 

1. Restore given above account on Device A
2. Generate sync QR on Device A
3. Scan sync QR on Device B
4. Observe the result

**Actual result:** Device B crashes after scanning QR. It is possible to login account on Device B manually, data is synced **BUT** devices are not paired. 

https://user-images.githubusercontent.com/97245802/226872792-9a30219f-f110-43e7-a394-259a42b92c23.mp4

### Additional Information

- Status version: nightly 

qfrank:
before invoke [HandleContactUpdate](https://github.com/status-im/status-go/blob/7c8a9aa1b53b064d148a84bc2dd669dc5ed500a7/protocol/messenger_sync_raw_messages.go#L55) , CurrentMessageState should be inited.

Closes [issue](https://github.com/status-im/status-mobile/issues/15446)


ghost:
### Pull Request Checklist

server/server_test.go
```diff
@@ -161,10 +161,11 @@ func (s *ServerURLSuite) TestQRCodeGeneration() {
...
 
 	_ = rootCAs.AppendCertsFromPEM(certPem)
+	//nolint: gosec
 	tr := &http.Transport{
 		TLSClientConfig: &tls.Config{
 			MinVersion:         tls.VersionTLS12,
-			InsecureSkipVerify: true, //nolint:all // MUST BE FALSE
+			InsecureSkipVerify: true,
 			RootCAs:            rootCAs,
 		},
```
Samyoul: I know you didn't introduce this, but would you be ok removing `InsecureSkipVerify: true,`? 😆 it is hilarious that that the comment says `MUST BE FALSE`, but the implementation sets the value to `true` and `nolint:all` to stop the linter complaining 😭 .

qfrank: sure

Samyoul (approved):

Samyoul:
https://github.com/status-im/status-go/blob/a8b0067e9a1ea356b0090e68e0402dda594d2808/protocol/messenger_sync_raw_messages.go#L48-L50

This seems a little bit hacky, but I can't think of better way to do this.

We are using this `" "` as a way to define an exceptional case, but what is the exception? We don't need to identify it in this PR.

If we knew the reason why it is ok to pass a `" "` in this case, but not in other cases, then we may be able to add some other condition that makes the code self describe what is happening.

PR #3337, opened 2023-03-23
feat: add `verify-ens` cmd utility

osmaczko:
Build&Run: `make verify-ens && ./build/bin/verify-ens`

A very simple example of how to use ens verifier.

### DIFF
--- a/cmd/verify-ens/main.go
+++ b/cmd/verify-ens/main.go
@@ -1,0 +1,39 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/status-im/status-go/eth-node/bridge/geth/ens"
+	enstypes "github.com/status-im/status-go/eth-node/types/ens"
+	"go.uber.org/zap"
+)
+
+const infuraToken = "putYourTokenHere"
+const verifyENSContractAddress = "0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e"
+
+var verifyENSURL = fmt.Sprintf("https://mainnet.infura.io/v3/%s", infuraToken)
+
+func main() {
+	logger := zap.NewExample()
+	defer logger.Sync()
+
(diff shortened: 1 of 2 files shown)


osmaczko:
CC: @felicio

cmd/verify-ens/main.go
felicio, line 1: The original test file that I've shared with you could be ran and stepped through while debugging just fine. And was enough for my use case.

I'm not sure if turning it into this utility is useful for anyone else. Maybe so if instead it just returned a pk for name via CLI param. But I'm not convinced about that either, so leaving that up to you.

Thanks anyway.

cmd/verify-ens/main.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"fmt"
+
+	"github.com/status-im/status-go/eth-node/bridge/geth/ens"
+	enstypes "github.com/status-im/status-go/eth-node/types/ens"
+	"go.uber.org/zap"
+)
+
+const infuraToken = "putYourTokenHere"
+const verifyENSContractAddress = "0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e"
```
felicio, line 12: Not really used.

osmaczko: WDYM? It is used in line 27

cmd/verify-ens/main.go
osmaczko: Sorry, I didn't notice the test file you shared with me earlier. It seems that you had edited the previous message, which is why I missed it. I agree that the utility is not useful, the changes I made were solely for illustrative purposes and were not meant to be merged.

cmd/verify-ens/main.go
felicio: I meant further down the stack. I didn't know that when looking up the value and later sharing it with you.

PR #3338, opened 2023-03-24
Use correct timesource in waku-2

cammellos:
Wakuv2 was not using the timesource provided, as we saw that if the device was out of sync it would send messages with the wrong timestamp, resulting in those being dropped.

The issue is fixed, I have also removed the ability to set a timesource since it's not used, and it defaults to a ntp synchronized timesource if none is provided.


ilmotta (approved):
> Wakuv2 was not using the timesource provided, as we saw that if the device was out of sync it would send messages with the wrong timestamp, resulting in those being dropped.

Since you say wakuv2 is not using the timesource provided, does this mean the actual bug is on wakuv2?

cammellos:
> > Wakuv2 was not using the timesource provided, as we saw that if the device was out of sync it would send messages with the wrong timestamp, resulting in those being dropped.
> 
> Since you say wakuv2 is not using the timesource provided, does this mean the actual bug is on wakuv2?

It's not at the protocol level, at the protocol level waku is agnostic about the source of the timestamp, is the responsibility of the client to do that, and in `waku/waku.go` we were not using an NTP synced timestamp, using local time of the device.

So it was an implementation issue, rather than an issue with wakuv2 itself.

vitvly (approved):

jrainville (approved):

Samyoul (approved):
Net reduction in code

image

PR #3345, opened 2023-03-24
fix(unread_count): Skip extra count of new messages for album of images

borismelnik:
Check if we already have one of the messages from album and skip increasing unread count

Important changes:

Closes #


protocol/message_persistence.go
```diff
@@ -552,6 +552,16 @@ func (db sqlitePersistence) messageByID(tx *sql.Tx, id string) (*common.Message,
 	return getMessageFromScanRows(db, rows)
 }
 
+func (db sqlitePersistence) messageByAlbumID(albumID string) ([]*common.Message, error) {
+	query := db.buildMessagesQuery("WHERE m1.album_id = ?")
```
cammellos: this will need an index, also I think we should query by local_chat_id and album_id, since album_id it's in theory not unique, and this will allow in theory someone to control whether their messages are going to be marked as seen/unseen in a different chat

protocol/messenger_handler.go
```diff
@@ -1785,7 +1785,35 @@ func (m *Messenger) HandleChatMessage(state *ReceivedMessageState) error {
...
+			image := receivedMessage.GetImage()
+
+			if image != nil && image.AlbumId != "" {
+				messages, err := m.persistence.MessagesByAlbumID(image.AlbumId)
+				if err != nil {
+					return err
+				}
+
+				// Skip unviewed counts increasing for other messages from album if we have it in db
+				skipUnviewedForAlbums = len(messages) > 0
```
cammellos: if you are skipping you can shortcut already

protocol/message_persistence.go
```diff
@@ -552,6 +552,16 @@ func (db sqlitePersistence) messageByID(tx *sql.Tx, id string) (*common.Message,
 	return getMessageFromScanRows(db, rows)
 }
 
+func (db sqlitePersistence) messageByAlbumID(albumID string) ([]*common.Message, error) {
```
ilmotta: Just a suggestion for naming: `albumMessages`, since we're getting a slice of messages filtered for a single album. The name `XByY` is often times used to represent *maps of X by Y*, or to say *X indexed by Y*. The same naming suggestion could then be used in the public function `MessagesByAlbumID` -> `AlbumMessages`.

protocol/messenger_send_images_album_test.go
```diff
@@ -136,3 +136,53 @@ func (s *MessengerSendImagesAlbumSuite) TestAlbumImageMessagesSend() {
...
+		"no messages",
+	)
+
+	s.Require().NoError(err)
+	s.Require().Len(response.Chats(), 1)
+	s.Require().Len(response.Messages(), messageCount)
+
+	for _, message := range response.Messages() {
+		image := message.GetImage()
+		s.Require().NotNil(image)
```
ilmotta: When asserting inside loops, a good practice I try to follow is to add a message that uniquely identifies the element under iteration, so that if the assertion fails, we can get a better idea of which element failed (not just the line number). Notice you can use format strings, like the ones supported by `fmt.Sprintf`.

```go
for _, message := range response.Messages() {
  image := message.GetImage()
  s.Require().NotNil(image, "Message.ID=%s", message.ID)
  s.Require().NotEmpty(image.AlbumId, "Message.ID=%s", message.ID)
}

//messenger_send_images_album_test.go:184:
//      Error:        Expected value not to be nil.
//      Test:         TestMessengerSendImagesAlbumSuite/TestAlbumImageMessagesWithMentionSend
//      Messages:     Message.ID=0x4bea63133548e365f1b654f8a8bb4cfeb86ea3768e389810422c74da95b44181
```

protocol/migrations/sqlite/000001_init.up.db.sql
```diff
@@ -62,6 +62,7 @@ CREATE TABLE IF NOT EXISTS user_messages (
     command_state INT
 );
 
+CREATE INDEX idx_album_id on user_messages(local_chat_id, album_id);
```
ilmotta: It's very common to add the table name to the index name, since it helps to understand which index is applied to which table.

But the more important question I have is, should we change an existing migration to add an index? Shouldn't we create a new migration file? cc: @cammellos

ilmotta (approved):

protocol/messenger_send_images_album_test.go
```diff
@@ -132,7 +132,57 @@ func (s *MessengerSendImagesAlbumSuite) TestAlbumImageMessagesSend() {
...
 		image := message.GetImage()
-		s.Require().NotNil(image)
-		s.Require().NotEmpty(image.AlbumId)
+		s.Require().NotNil(image, "Message.ID=%s", message.ID)
+		s.Require().NotEmpty(image.AlbumId, "Message.ID=%s", message.ID)
 	}
 }
+
+func (s *MessengerSendImagesAlbumSuite) TestAlbumImageMessagesWithMentionSend() {
+	theirMessenger := s.newMessenger()
```
ilmotta, line 153: @borismelnik, just a reminder to update the `VERSION` file before merging.

protocol/migrations/sqlite/000001_init.up.db.sql
jrainville: Yes, we need a new migration file. The migrations that have already been applied are skipped, so this index would never apply to people just updating their version.

ilmotta: Thanks @jrainville :)

borismelnik:
@ilmotta @cammellos @jrainville pls add a new review :)

jrainville (approved):
LGTM

ilmotta (approved):

cammellos (approved):

PR #3346, opened 2023-03-24
fix(encryptor): correctly send an error when there is no encryption key

jrainville:
It would crash otherwise, leaving no message


jrainville:
I found this crash when importing my account using the seed phrase. The community was fetched, but writing in it would make it crash. I'm not yet sure what to do in the client with that. The community is pretty much dead without the key

0x-r4bbit (approved):

caybro:
Yup, this worked for me, Status CC restored correctly now

caybro (approved):

PR #3347, opened 2023-03-26
Add album count key to messages

### LINKED ISSUE #15383: Photo album reply doesn't display the number of photos in the album
# Bug Report

## Problem 
Currently, when replying to a photo album, we only display the word "Image" in the quoted message in the composer. We should display the number of photos in the album instead. 

#### Expected behavior

`3 photos` when replying to the album with three images

Screenshot 2023-03-16 at 11 56 43

Figma: https://www.figma.com/file/PPWkgOYlZZDxZv5SDGsZVV/Posts-%26-Attachments-for-Mobile?node-id=2839-278601&t=56KGyLmXnk2cbJ5X-4

#### Actual behavior

Screenshot 2023-03-16 at 11 57 53

### Reproduction

1. Send a photo album
2. Long press on the message > `Reply`

### Additional Information

- Status version: nightly 16/03

ibrkhalil:
This commit adds album count key to messages with multiple photos
Closes [#15383](https://github.com/status-im/status-mobile/issues/15383) on status-mobile


protocol/messenger.go
```diff
@@ -1974,8 +1974,10 @@ func (m *Messenger) SendChatMessages(ctx context.Context, messages []*common.Mes
 		return nil, err
 	}
 
+	messagesLength := len(messages)
 	for _, message := range messages {
 		if message.ContentType == protobuf.ChatMessage_IMAGE && len(messages) > 1 {
+			message.AlbumCount = messagesLength
```
ilmotta: I don't understand this part @ibrkhalil, could you shed some light?

Let's assume we have 2 messages, but only one has the content type `ChatMessage_IMAGE`, then the message of type `ChatMessage_IMAGE` will have its `AlbumCount` field set to 2, but shouldn't it be 1?

I think we should aggregate the count `messagesLength` with the filtering logic, or did I get something wrong?

ilmotta (commented):
It'd be good to get some unit tests in this PR @ibrkhalil, or update the existing ones.

protocol/messenger.go
vkjr: As far as I can see, status-mobile always sends images in a single set of messages, where every message is image-related, so this logic probably won't cause problems now. But I would agree that relying on that is fragile. 

https://github.com/status-im/status-mobile/blob/develop/src/status_im/chat/models/input.cljs#L188

protocol/common/message.go
```diff
@@ -170,6 +170,8 @@ type Message struct {
 	AudioPath string `json:"audioPath,omitempty"`
 	// ImageLocalURL is the local url of the image
 	ImageLocalURL string `json:"imageLocalUrl,omitempty"`
+	// AlbumCount for a collage of images
+	AlbumCount int `json:"albumCount,omitempty"`
```
vkjr: Wdyt about the name `AlbumSize`? Sounds a bit clearer to me. Also, I'd adjust the comment line to be "AlbumCount is a number of images in a collage"

protocol/messenger.go
ilmotta: I didn't know that, thanks @vkjr. Still, as you said, I think status-go shouldn't assume this much.

protocol/common/message.go
ibrkhalil: Personnally .. I prefer `count` I think it refers to integer .. Will edit the comment though

ibrkhalil:
> It'd be good to get some unit tests in this PR @ibrkhalil, or update the existing ones.

I extended the test currently in place to check if the albumCount property is set correctly
07a57e7932da24c8d8f40cfe257727df702c61b4

protocol/common/message.go
vkjr: It sounds a bit of misleading to me. "Album count" can be incorrectly understood as the number of albums. Maybe `AlbumPhotosCount`?

ilmotta: That's exactly how I felt about the word `count`. I had to re-read many times the name `AlbumCount` until I was sure it wasn't a count of albums, but the count of messages of type *image*.

ibrkhalil: Well if you put it that way, then I guess it's a bad name
Okie dokie
Renaming

ibrkhalil: Done

protocol/messenger_send_images_album_test.go
```diff
@@ -116,6 +116,12 @@ func (s *MessengerSendImagesAlbumSuite) TestAlbumImageMessagesSend() {
 	s.NoError(err)
 	response, err := s.m.SendChatMessages(context.Background(), album)
 	s.NoError(err)
+
```
vkjr, line 119: This test covers the case when all messages passed to `m.SendChatMessages` are of image type. Currently `status-mobile` uses this function exactly this way. But I believe it would be safer to suppose that behavior on `status-mobile` can change. And in the case when among sent messages there will be few of Image type and few of other types. That doesn't contradict the name of the function `SendChatMessages`.  In that case we will get an incorrect behaviour with counting `AlbumImagesCount`. I would propose changing the `SendChatMessages` implementation and counting the number of image messages as a first step, than sending all the messages and adding `AlbumImagesCount` property only to those of image type. Wdyt?

ibrkhalil: I totally agree .. That'd make it more future-proof .. Thanks for the suggestion Volodymyr

ibrkhalil: Done .. Kindly take another look @vkjr

vkjr (approved):

protocol/messenger_send_images_album_test.go
vkjr: Thanks!

PR #3348, opened 2023-03-27
Check ownership of collectibles in community permissions

0x-r4bbit:
This adds an additional check for collectibles when community permissions are validated.

Specifically this uses opensea to request all NFTs given an owner wallet and a list of contract addresses (collectibles).

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -37,6 +37,7 @@ import (
 	"github.com/status-im/status-go/protocol/protobuf"
 	"github.com/status-im/status-go/protocol/requests"
 	"github.com/status-im/status-go/protocol/transport"
+	"github.com/status-im/status-go/services/wallet/thirdparty/opensea"
 	"github.com/status-im/status-go/services/wallet/token"
 	"github.com/status-im/status-go/signal"
 )
@@ -72,6 +73,7 @@ type Manager struct {
 	quit                           chan struct{}
 	torrentConfig                  *params.TorrentConfig
 	torrentClient                  *torrent.Client
+	walletConfig                   *params.WalletConfig
 	historyArchiveTasksWaitGroup   sync.WaitGroup
 	historyArchiveTasks            map[string]chan struct{}
 	periodicMemberPermissionsTasks map[string]chan struct{}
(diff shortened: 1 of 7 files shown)


protocol/communities/manager.go
```diff
@@ -1433,25 +1511,91 @@ func (m *Manager) checkPermissionToJoin(permissions []*protobuf.CommunityTokenPe
...
+		for contractAddress := range erc721Tokens {
+			contractAddresses = append(contractAddresses, gethcommon.HexToAddress(contractAddress))
+		}
+
+		if _, exists := ownedERC721Tokens[chainID]; !exists {
+			ownedERC721Tokens[chainID] = make(map[string][]opensea.Asset)
+		}
+
+		for _, owner := range walletAddresses {
+			client, err := opensea.NewOpenseaClient(chainID, m.walletConfig.OpenseaAPIKey)
```
jrainville: Do we need a new client for each wallet address?

jrainville (commented):
Good job. Just a small question

protocol/communities/manager.go
0x-r4bbit: Yea this has been a bit of a back and forth because of the request limit error we got from opensea, but eventually this has been changed such that we'll always get the same client instance per chain ID, no matter how often you call `NewOpenseaClient`.

In other words, it's kinda like a singleton at this point so it doesn't make a difference. But I can change it since it (rightfully) causes questions.

0x-r4bbit: Updated this

jrainville (approved):

services/wallet/thirdparty/opensea/client.go
```diff
@@ -22,18 +21,27 @@ import (
...
+	switch chainID {
+	case 1:
+		return "https://api.opensea.io/api/v1", nil
+	case 4:
+		return "https://rinkeby-api.opensea.io/api/v1", nil
+	case 5:
+		return "https://testnets-api.opensea.io/api/v1", nil
+	}
 
-var BaseURLs = map[uint64]string{
```
alaibe, line 30: The variable could have just made private but fine by me, a function works too

alaibe (approved):

PR #3358, opened 2023-03-28
Add a few more fields that are to be set by the client and fix logs directory

cammellos:
We were missing a few defaults, and the log directory was set to the wrong path.


rasom (approved):

osmaczko (approved):

api/defaults.go
```diff
@@ -28,38 +28,42 @@ var paths = []string{pathWalletRoot, pathEIP1581, pathDefaultChat, pathDefaultWa
 func defaultSettings(generatedAccountInfo generator.GeneratedAccountInfo, derivedAddresses map[string]generator.AccountInfo, mnemonic *string) (*settings.Settings, error) {
 	chatKeyString := derivedAddresses[pathDefaultChat].PublicKey
 
-	settings := &settings.Settings{}
```
Samyoul, line 31: Nice to see others removing vars that shadow package names.

Samyoul (approved):

PR #3359, opened 2023-03-28
feat: avoid vm error to retry

alaibe:


dlipicar (approved):
Looks good. Tested, works fine in case of reverted execution.

PR #3366, opened 2023-03-30
feat: collectibles cache

dlipicar:
NFT metadata should not normally change, so we can save us several Opensea requests by caching it.

When requesting NFTs by their unique ID, we'll first look in the cache. If not found, we'll fetch them.
The cache gets updated when the assets are retrieved through the ownership-related methods.

### DIFF
--- a/services/wallet/collectibles/collectibles.go
+++ b/services/wallet/collectibles/collectibles.go
@@ -3,6 +3,7 @@ package collectibles
 import (
 	"context"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
@@ -19,6 +20,8 @@ type Manager struct {
 	rpcClient        *rpc.Client
 	metadataProvider thirdparty.NFTMetadataProvider
 	openseaAPIKey    string
+	nftCache         map[uint64]map[string]opensea.Asset
+	nftCacheLock     sync.RWMutex
 }
 
 func NewManager(rpcClient *rpc.Client, metadataProvider thirdparty.NFTMetadataProvider, openseaAPIKey string) *Manager {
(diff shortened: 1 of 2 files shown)


IvanBelyakoff:
@dlipicar it is great idea to cache this data. What worries me is bloating RAM usage. In my opinion we should cache data in RAM only if it is used super often. I'd go with caching it in DB, since I don't think this is used that often that we should keep this in RAM. Wdyt?
I'd listen to other opinions as well

stefandunca:
> ... I don't think this is used that often that we should keep this in RAM. Wdyt? I'd listen to other opinions as well

I agree that caching to DB on the status-go side looks like a good idea for saving requests and in-memory caching looks good for the short term only. Wondering what is the eviction policy for this cache. What happens with users that scroll through accounts that have thousands of collectables given that our app is meant to be kept alive always between restarts?

stefandunca (approved):

dlipicar:
Yeah, this started as a POC mainly to see if I could improve app responsiveness when the transaction list gets refreshed (which requires NFT metadata in case of ER721 transactions). The current approach does work, but the most proper one would be to store this data in a DB. That will require a bit more work though, since collectibles are a relatively complex struct and I need to define it in a way that it's provider-agnostic instead of exclusive to OpenSea.
@IvanBelyakoff  @stefandunca Would you be OK if moving the cache to a DB is left as a (near) future improvement?

IvanBelyakoff:
Yes, it is fine with me.  @dlipicar Could you please create an issue for that?

IvanBelyakoff (approved):

dlipicar:
> Yes, it is fine with me. @dlipicar Could you please create an issue for that?

Done https://github.com/status-im/status-desktop/issues/10107

saledjenic (approved):

PR #3369, opened 2023-04-03
fix(DeployCollectiblesEstimate): Add DeployCollectiblesEstimate function

### LINKED ISSUE #10035: Improve Sign transaction backend in minting
Make service calls asynchronous.
Format values in QML.

endulab:
Issue #10035


borismelnik (approved):

borismelnik:
@endulab is it always constant? Maybe good to add new constant for it?

services/collectibles/api.go
```diff
@@ -105,6 +105,11 @@ func (api *API) Deploy(ctx context.Context, chainID uint64, deploymentParameters
 	return DeploymentDetails{address.Hex(), tx.Hash().Hex()}, nil
 }
 
+// Returns gas units + 10%
+func (api *API) DeployCollectiblesEstimate(ctx context.Context) (uint64, error) {
+	return 3702411, nil
```
jrainville, line 110: Is it normal that it is hardcoded?

endulab: @alaibe is it possible to use estimateGas() to compute contract deployment cost ?

alaibe: U can try to use the estimageGas function, however verify that the numbers are correct, i had very weird behaviour with bridge contract

endulab: @jrainville yes, in case of deployment it is a common way that we check how many gas units was required during testing deployments and we use that number. We should keep it as it is and remember that the number will change when the smart contract changes (different bytecode). We can experiment with estimateGas() later.

jrainville (approved):

PR #3373, opened 2023-04-04
UDP network checking

Samyoul:
## What's Changed

This PR adds functionality to the Local Pairing process flow that allows pairing devices to perform preflight network checks to attempt to discover a pairing peer on the network.

This feature provides proof positive that there is another peer in the network that claims to be capable of pairing with the active device.

**⚠️ NOTE: ⚠️**
- Failure to find a UDP peer is **NOT** proof that Local Pairing is not possible.
  - UDP packets may be blocked by the OS, firewall, network etc.

### Process Flow:

- A UDP beacon multicasts on a known port (`9999`) to notifying the network of the multicaster device's pairing device data
- Simultaneously the UDP beacon listens on a known port (`9999`) for any pairing peer data.
- If device is found, the payload is checked, the signature is verified against *the* shared private key*.
- If verification succeeds the application will emit a signal containing the pairing peer's data.
- A 5 second countdown begins after which the beacon will shutdown.
  - The 5 second window give the counterpart peer reasonable time to discover the discoveree peer, if it has not already found the peer.

*The key doesn't need to be secret because its purpose is to give confidence that the peer is another status installation, we don't *need* proof, as this feature is designed to give an indication that the another device is accessible on the network.

### API Access:

This PR adds and exposes a new API endpoint `StartSearchForLocalPairingPeers`, this method will return nil or error.

`StartSearchForLocalPairingPeers` starts a UDP multicast beacon that both listens for and broadcasts to LAN peers on discovery the beacon will emit a signal with the details of the discovered peer.

Currently, beacons are configured to search for 2 minutes pinging the network every 500 ms;
  - If no peer discovery is made before this time elapses the operation will terminate.
  - If a peer is discovered the `pairing.PeerNotifier` will terminate operation after 5 seconds, giving the peer
    reasonable time to discover this device.

Peer details are represented by a `json.Marshal` `peers.LocalPairingPeerHello`

### New Signal:

This PR introduces a new signal with the following details:

```go
Event{
    Type:   EventPeerDiscovered,
    Action: ActionPeerDiscovery,
    Data:   peers.LocalPairingPeerHello,
}
```

`EventPeerDiscovered` is declared as follows:

```go
EventPeerDiscovered    EventType = "peer-discovered"
```

`ActionPeerDiscovery` resolves to an int with value `5`

`peers.LocalPairingPeerHello` will be marshalled with the below structure:

```go
struct {
	PeerID     []byte // base64 encoded
	DeviceName string
	DeviceType string
	Address    string
}
```

### UDP Payload

The UDP beacon will emit a payload in addition to its network address, this payload is a marshalled protobuf defined as below:

```protobuf
message LocalPairingPeerHello {
  bytes peer_id = 1;
  string device_name = 2;
  string device_type = 3;
  bytes signature = 4;
}
```

The `signature` field is performed with the hash of the other fields, declared as below:

```go
sha256.Sum256(append(h.PeerId, []byte(h.DeviceName+h.DeviceType)...))
```

Which effectively concatenates the data as bytes in `peer_id` + `device_name` + `device_type` order.

## Why Change?

- As a user I would like to receive some kind of feedback that Local Pairing should be possible.
- As a tester I want to be sure the devices should be capable of communicating via the network.
- In the future we may want to have confidence that the devices are able to "see" each other on the network, as a mechanism of guiding the user pairing flow.
  - Example: If we want to offer multiple pairing mechanisms


ghost:
### Pull Request Checklist

server/pairing/peers/udp_notifier.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return err
+	}
+
+	settings, err := un.MakeUDPP2PSettings(deviceName, deviceType)
+	if err != nil {
+		return err
+	}
+
+	settings.Delay = 500 * time.Millisecond
```
ilmotta: It's probably a good idea to parameterize the 500ms and 2min values, example names: *ping delay* and *search time limit*.

mobile/status.go
```diff
@@ -1023,6 +1023,21 @@ func GenerateImages(filepath string, aX, aY, bX, bY int) string {
...
+// StartSearchForLocalPairingPeers starts a UDP multicast beacon that both listens for and broadcasts to LAN peers
+// on discovery the beacon will emit a signal with the details of the discovered peer.
+//
+// Currently, beacons are configured to search for 2 minutes pinging the network every 500 ms;
+//   - If no peer discovery is made before this time elapses the operation will terminate.
+//   - If a peer is discovered the pairing.PeerNotifier will terminate operation after 5 seconds, giving the peer
+//     reasonable time to discover this device.
+//
+// Peer details are represented by a json.Marshal peers.LocalPairingPeerHello
+func StartSearchForLocalPairingPeers() string {
```
ilmotta, line 1035: @Samyoul, do you think clients might need to sometimes terminate the search before waiting for 2min? Just curious, would it be possible to implement something in the future to preemptively terminate the search? For example, a client might want or need to restart the search.

Also, what if some client finds 2min too long or 500ms pings excessively short, and so on. I was wondering if it would be a good idea to change `StartSearchForLocalPairingPeers` to accept some arguments, but if they're not present allow `Search` to use sensible defaults (the values you already hardcoded seem good).

We probably don't need my suggestion (YAGNI right?), but anyway, I always think of status-go as something that can (and probably will) be used in the future to build clients differently than the beautiful couple desktop :heart_decoration: mobile, especially once we "succeed" in the market and the community starts to experiment with their own ideas.

ilmotta (approved):
No major comments. Always a pleasure to review your PRs.

mobile/status.go
Samyoul: Thank you @ilmotta for your feedback, you make some good points.

> do you think clients might need to sometimes terminate the search before waiting for 2min? Just curious, would it be possible to implement something in the future to preemptively terminate the search? For example, a client might want or need to restart the search.

I actually struggled with this consideration myself, and I ultimately do not disagree with you. It is technically possible, the only route to achieving this would be to attach something linked to the peer search functionality to the `GethStatusBackend`. This is because the search functionality occurs in a sub routine which is inaccessible once fired, communication can only be done via `chan`. The lower level UDP functionality actually does use a "stop chan", which we use to terminate the search after a peer is discovered.

Attaching the stop `chan` to the backend would be the simplest way to allow for ad hoc termination, but it may pollute the concerns of the `GethStatusBackend`, see how all of the concerns of the of this struct are quite low level.

```go
// GethStatusBackend implements the Status.im service over go-ethereum
type GethStatusBackend struct {
	mu sync.Mutex
	// rootDataDir is the same for all networks.
	rootDataDir string
	appDB       *sql.DB
	config      *params.NodeConfig

	statusNode           *node.StatusNode
	personalAPI          *personal.PublicAPI
	multiaccountsDB      *multiaccounts.Database
	account              *multiaccounts.Account
	accountManager       *account.GethManager
	transactor           *transactions.Transactor
	connectionState      connection.State
	appState             appState
	selectedAccountKeyID string
	log                  log.Logger
	allowAllRPC          bool // used only for tests, disables api method restrictions

}
```

---

> Also, what if some client finds 2min too long or 500ms pings excessively short, and so on. I was wondering if it would be a good idea to change StartSearchForLocalPairingPeers to accept some arguments, but if they're not present allow Search to use sensible defaults (the values you already hardcoded seem good).

Yes we can make this configurable. I just pulled these numbers out of the air anyway 😆 .

server/pairing/peers/private_key.go
```diff
@@ -0,0 +1,24 @@
...
+
+var (
+	// pk Yes this is an actual ECDSA **private** key committed to a public repository visible to anyone.
+	// DO NOT use this key for anything other than signing udp "hellos". The key's value is in giving other Status
+	// installations CONFIDENCE, NOT proof, that the sender of the UDP pings is another Status device.
+	// We do not rely on UDP message information to orchestrate connections or swap secrets. The use case is purely
+	// to make preflight checks which ADVISE the application and the user.
+	//
+	// A signature is more robust and flexible than an application identifier, and serves the same role as an ID, while
+	// securing the payload against tampering.
```
richard-ramos, line 17: Thank you for this comment! :pray:

richard-ramos (approved):
LGTM! :100:

The only thing i'd change is to not include migrations / protobuffer changes that were not really affected by the PR (makes it easier to see the history of migration files to find which PRs affected them)

mobile/status.go
ilmotta: > Attaching the stop chan to the backend would be the simplest way to allow for ad hoc termination, but it may pollute the concerns of the GethStatusBackend

It's good to know it's technically viable without too much trouble. Since there's no clear need from clients nowadays, it's reasonable to me to delay implementing this capability. Thanks for the explanation and code example.

> Yes we can make this configurable

Nice! :bicyclist:

qfrank (approved):
Cool PR 👍

PR #3374, opened 2023-04-04
fix(AddTokenOwners): Move AddTokenOwners from MintTo

### LINKED ISSUE #10161: Track minting transaction
Token holders should not be added to database after airdrop.
First we need to track minting transaction and add to database when transaction is success.

endulab:
Issue #10161

The initial thought was to incorporate transaction system in collectibles service and track airdrop transaction
inside collectibles service. After some discussions it turned out that it is not possible with current design.
The tracking logic is tied to Nim side and we need to do this just like in case of other actions:
1. Send MinTo request to blockchain
2. Track transaction in Nim
3. If transaction is successful update database.

In this pr I moved database function from MinTo into a separated API call.


services/collectibles/api.go
```diff
@@ -165,3 +165,15 @@ func (api *API) ContractOwner(ctx context.Context, chainID uint64, contractAddre
 	}
 	return owner.String(), nil
 }
+
+func (api *API) AddTokenOwners(ctx context.Context, chainID uint64, contractAddress string, walletAddresses []string, amount int) error {
+	if len(walletAddresses) == 0 {
+		return errors.New("wallet addresses list is empty")
+	}
+	if amount <= 0 {
+		return errors.New("amount is <= 0")
```
jrainville, line 187: Since you use those error message twice, can you add them as constants at the top.

jrainville (approved):

mprakhov (approved):
LGTM

services/collectibles/api.go
endulab: Done

borismelnik (approved):

PR #3375, opened 2023-04-04
Update album_images_count column on receiver side

ibrkhalil:
Bug fix for a feature I recently added


protocol/common/message.go
```diff
@@ -171,7 +171,7 @@ type Message struct {
 	// ImageLocalURL is the local url of the image
 	ImageLocalURL string `json:"imageLocalUrl,omitempty"`
 	// AlbumImagesCount is the number of images in a collage
-	AlbumImagesCount int `json:"albumImagesCount,omitempty"`
+	AlbumImagesCount int32 `json:"albumImagesCount,omitempty"`
```
cammellos: That should be a uint, unless you want to use it as an actual int

protocol/messenger.go
```diff
@@ -1979,7 +1979,7 @@ func (m *Messenger) SendChatMessages(ctx context.Context, messages []*common.Mes
 		return nil, err
 	}
 
-	ImagesCount := 0
+	ImagesCount := int32(0)
```
cammellos: lowercase letter for variables

protocol/common/message.go
```diff
@@ -703,12 +703,13 @@ func (m *Message) LoadImage() error {
 	return nil
 }
 
-func (m *Message) SetAlbumID(albumID string) error {
+func (m *Message) SetAlbumID(albumID string, ImagesCount uint32) error {
```
borismelnik: I'm not sure, but for me it will be better if:
```
func (m *Message) SetCountAndAlbumID(imagesCount uint32, albumID string) error

**Comment by borismelnik** [comment]
I think the better place to add it here:
```
type ImageMessage struct {
```

where `AlbumID` as well. This property making a sense just for `ImageMessage`.

File: protocol/common/message.go [hunk 4]
```diff
@@ -703,12 +703,13 @@ func (m *Message) LoadImage() error {
...
 	imageMessage := m.GetImage()
 	if imageMessage == nil {
 		return errors.New("Image is empty")
 	}
 	imageMessage.AlbumId = albumID
+	imageMessage.AlbumImagesCount = ImagesCount
```
borismelnik: Here you have `ImageMessage` instance which doesn't have `AlbumImagesCount`.  Is it should be `m.AlbumImagesCount`?

ibrkhalil: Shouldn't it be defaulted to `0` automatically ?

borismelnik:
@cammellos  I tested it on desktop - it works.

protocol/messenger_send_images_album_test.go
```diff
@@ -118,14 +118,15 @@ func (s *MessengerSendImagesAlbumSuite) TestAlbumImageMessagesSend() {
 	s.NoError(err)
 
 	// Check that album count was the number of the images sent
-	imagesCount := 0
+	imagesCount := uint32(0)
```
cammellos, line 121: `var imagesCount uint32` is probably more idiomatic

cammellos (approved):

borismelnik (approved):

PR #3380, opened 2023-04-05
fix(image-album): make sure to delete all images part of an album

### LINKED ISSUE #10171: Images Album: Can't delete album correctly
# Bug Report

## Description

When user trying to delete the message with album of images, only first images will be deleted.

## Steps to reproduce

1. Send album of images (2-3 images) to someone.
2. Delete the message
3. Reload the app

#### Expected behavior

Message with album was deleted

#### Actual behavior

Just first message was deleted

### Additional Information

- Status desktop version: `0.11.0 RC`
- Operating System: `all`

jrainville:
Needed for https://github.com/status-im/status-desktop/issues/10171

When we delete a message, if it is an Image message, we make sure to delete all the messages part of the album.

Same thing when we get the signal to delete a message, we remove all of the messages in the album.

I might need do a check if we receive later an image from the album and it was already deleted, I need to check if that's handled.
I'll also add a test


borismelnik (approved):
LGTM :) but will be great to add some test for it

protocol/messenger_handler.go
```diff
@@ -1574,21 +1574,44 @@ func (m *Messenger) HandleDeleteMessage(state *ReceivedMessageState, deleteMessa
...
+			messagesToDelete = append(messagesToDelete, messageInAlbum)
+		}
+	} else {
+		messagesToDelete = append(messagesToDelete, originalMessage)
+	}
+
 	// Update message and return it
-	originalMessage.Deleted = true
-	originalMessage.DeletedBy = deleteMessage.DeleteMessage.DeletedBy
+	for _, messageToDelete := range messagesToDelete {
```
ilmotta, line 1591: @jrainville, do you know if the action of deleting messages is happening inside a transaction? If not, then although rare, it would be better if we could save all messages in one go (query) or do one by one and rollback in case one of them fail before the loop finishes.

protocol/messenger_messages.go
```diff
@@ -157,6 +157,21 @@ func (m *Messenger) DeleteMessageAndSend(ctx context.Context, messageID string)
 		return nil, ErrInvalidDeleteTypeAuthor
 	}
 
+	var messagesToDelete []*common.Message
```
ilmotta: This piece of code that builds up the messages to be deleted is identical or very similar to what's in `messenger_handler.go` (I haven't checked thoroughly). I think we should at least extract it into a common function.

protocol/messenger_messages.go
```diff
@@ -184,11 +199,16 @@ func (m *Messenger) DeleteMessageAndSend(ctx context.Context, messageID string)
 		return nil, err
 	}
 
-	message.Deleted = true
-	message.DeletedBy = deletedBy
-	err = m.persistence.SaveMessages([]*common.Message{message})
-	if err != nil {
-		return nil, err
+	response := &MessengerResponse{}
+	for _, messageToDelete := range messagesToDelete {
```
ilmotta, line 193: As @borismelnik mentioned already, I think we should cover the new code with a bit of testing in this PR, or in a follow-up PR if this one needs to be merged asap.

Edit: Sorry, I didn't see you mentioned in the PR description about adding tests.

jrainville:
Thanks for the reviews. As stated in the description, I'm working on adding tests right now. I'll push them in another commit this afternoon.

jrainville:
I added a test for deleting an album. I'll add another one later in case not all messages arrived when we got a deletion.
I also need to address @ilmotta 's comment.
I'll do it tomorrow, there is no rush to merge this. I need to tackle an issue for our release in the meantime

protocol/messenger_handler.go
jrainville: I don't think it happens in a transaction.
As for rolling back, I'm not sure if it's necessary. Since the goal is to delete, if one of the messages fails, it's ok if the some of them are deleted IMO.
What would happen on the front is that one of the images would disappear, the user could try to delete again for the remaining ones.

jrainville:
Alright, I addressed the comments and implemented the tests.
We don't support DeleteForMe on Desktop yet, so I wasn't able to test that code in the client itself, so if someone from mobile could check that I didn't break it (it's pretty much the same as Delete, so it should work), it would be great. Thanks.

protocol/messenger_handler.go
ilmotta: Thanks for taking a look @jrainville. It would be a bit better if we could be more consistent by default, but as you said and I agree, in the rare case that something happens the user could just try again, so it's fine.

ilmotta:
> Alright, I addressed the comments and implemented the tests. We don't support DeleteForMe on Desktop yet, so I wasn't able to test that code in the client itself, so if someone from mobile could check that I didn't break it (it's pretty much the same as Delete, so it should work), it would be great. Thanks.

@jrainville, I can test this next Monday. Today is a bit tight because tomorrow I'm off. Would that be okay?

ilmotta (approved):

protocol/messenger_test.go
```diff
@@ -2319,6 +2321,87 @@ func (s *MessengerSuite) TestResendExpiredEmojis() {
...
+		Width:   1200,
+		Height:  1000,
+	}
+	message.Payload = &protobuf.ChatMessage_Image{Image: &image}
+
+	return message, nil
+}
+
+/*
+func testSendImages(s *suite.Suite, messageCount int, ourChat *Chat, theirMessenger *Chat) []*common.Message {
```
ilmotta: Do we really need to keep this commented test?

jrainville:
> I can test this next Monday. Today is a bit tight because tomorrow I'm off. Would that be okay?

yeah that's good. I won't pull this into our release in the end. It's not a crucial fix and it touches more code than I thought it would.

ilmotta:
@jrainville I just tried to test this branch, but I'm getting an error while using the chat because the endpoint `ChatMentionCheckMentions` is not available. Could you rebase so I can proceed with the tests in the mobile app?

jrainville:
@ilmotta done

ilmotta:
> @ilmotta done

@jrainville Tested on mobile and it behaves as in `develop`. Double-checked Delete for me, and all images are deleted from the album :rocket:

PR #3382, opened 2023-04-06
fix: add supported chains to OpenSea

dlipicar:
Opensea supports several chains (both mainnet and testnet: https://support.opensea.io/hc/en-us/articles/4404027708051-Which-blockchains-are-compatible-with-OpenSea-). Add the ones we use in the app, since, although we normally pass the Ethereum chain when fetching owned collectibles, the specific one is used when fetching NFT metadata for transactions.

### DIFF
--- a/services/wallet/thirdparty/opensea/client.go
+++ b/services/wallet/thirdparty/opensea/client.go
@@ -35,15 +35,15 @@ const ChainIDRequiringAPIKey = 1
 
 func getbaseURL(chainID uint64) (string, error) {
 	switch chainID {
-	case 1:
+	case 1, 10, 42161:
 		return "https://api.opensea.io/api/v1", nil
 	case 4:
 		return "https://rinkeby-api.opensea.io/api/v1", nil
-	case 5:
+	case 5, 420, 421613:
 		return "https://testnets-api.opensea.io/api/v1", nil
 	}
 
-	return "", fmt.Errorf("chainID not supported")
+	return "", fmt.Errorf("chainID not supported: %d", chainID)
 }
 
 var OpenseaClientInstances = make(map[uint64]*Client)


IvanBelyakoff (approved):

services/wallet/thirdparty/opensea/client.go
```diff
@@ -35,15 +35,15 @@ const ChainIDRequiringAPIKey = 1
 
 func getbaseURL(chainID uint64) (string, error) {
 	switch chainID {
-	case 1:
+	case 1, 10, 42161:
```
alaibe, line 38: Opensea don't support arbitrum/optimism AFAIK why would u add it here?

dlipicar: It does, according to the docs: https://support.opensea.io/hc/en-us/articles/4404027708051-Which-blockchains-are-compatible-with-OpenSea-

alaibe: It doesn't state anything for goerli optimism and goerli arbitrum?
also does that mean that nfts are going to be aggregated under one api?

dlipicar: They do list them: 
image

Yeap, there's no way to tell OpenSea "give me the NFTs for this specific chain" when we do a request, your only choice there is to use the mainnet or the testnet endpoint. You'll get assets from all chains on the same response, and you can filter by looking at `"chain_identifier": "ethereum"` in each asset from the response (we're currently not doing that).

The app currently uses only Ethereum (chainID 1) for Mainnet mode and Goerli (chainID 5) for Testnet mode when fetching owned NFTs, but we're actually getting collectibles from all other chains as well. This change only became needed after I added collectibles to the Transaction list, since I'm using the ChainID from the transaction in the backend call to get the metadata.

stefandunca (approved):

PR #3385, opened 2023-04-10
backup/sync if seed phrase backed up

qfrank:
This PR mainly implements the synchronization of the mnemonic_removed in the settings


ghost:
### Pull Request Checklist

saledjenic (approved):
@qfrank looks good, just don't forget to update `VERSION` file before merging.

Samyoul (commented):
Thank you for this PR @qfrank , ordinarily syncing a field like this should be done exactly as you've implemented this code.

However with the seed phrase we should be exceptionally careful. May I suggest a different approach?

We need to avoid ANY chance that the mnemonic is sent via waku. Which I know this PR has a check for, but the chance / risk that this functionality breaks with another change by another developer is too high.

Perhaps we could add a new column called `mnemonic_present` or `mnemonic_removed` or something like that, make the field a `boolean` and then only ever sync that boolean value. On the recipient side we need to remove the mnemonic if the new field is true / false.

Also perhaps we could put something in place to ensure that the value of `mnemonic` is never synced without a very very deliberate change. MAybe for the moment just a comment that says `// ⚠️🚨 DO NOT EVER SYNC! 🚨⚠️` 😆

Does that make sense, does that seem reasonable to you?

qfrank:
> Thank you for this PR @qfrank , ordinarily syncing a field like this should be done exactly as you've implemented this code.
> 
> However with the seed phrase we should be exceptionally careful. May I suggest a different approach?
> 
> We need to avoid ANY chance that the mnemonic is sent via waku. Which I know this PR has a check for, but the chance / risk that this functionality breaks with another change by another developer is too high.
> 
> Perhaps we could add a new column called `mnemonic_present` or `mnemonic_removed` or something like that, make the field a `boolean` and then only ever sync that boolean value. On the recipient side we need to remove the mnemonic if the new field is true / false.
> 
> Also perhaps we could put something in place to ensure that the value of `mnemonic` is never synced without a very very deliberate change. MAybe for the moment just a comment that says `// ⚠️🚨 DO NOT EVER SYNC! 🚨⚠️` 😆
> 
> Does that make sense, does that seem reasonable to you?

yeah, make sense👍! That‘s why i'm waiting for your review 😉 In fact, I had previously considered the method of adding a new bool column, which is also safer, but before that, I was more inclined towards a simple implementation without considering the safety (I mean future risks) as much as you did. @Samyoul

multiaccounts/settings/columns.go
```diff
@@ -185,9 +189,22 @@ var (
...
 	Mnemonic = SettingField{
-		reactFieldName: "mnemonic",
-		dBColumnName:   "mnemonic",
+		reactFieldName: DBColumnMnemonic,
+		dBColumnName:   DBColumnMnemonic,
 	}
+
+	MnemonicBackedUp = SettingField{
+		reactFieldName: "mnemonic-backed-up?",
+		dBColumnName:   "mnemonic_backed_up",
```
Samyoul: The term `MnemonicBackedUp` seems to communicate that mnemonic data *can* be backed up. But what we really are checking is if `mnemonic` is empty. This language could cause confusion with other developers who may believe that the mnemonic value should be backed up.

Perhaps change this field name to something like `MnemonicRemoved`

multiaccounts/settings/database.go
```diff
@@ -242,6 +252,10 @@ func (db *Database) SaveSettingField(sf SettingField, value interface{}) error {
 	return db.parseSaveAndSyncSetting(sf, value)
 }
 
+func (db *Database) EmptyMnemonic() error {
+	return db.saveSetting(Mnemonic, nil)
+}
+
```
Samyoul: Perhaps change to `DeleteMnemonic` just so that the terminology is aligned with CRUD conventions.

Samyoul (approved):
Obviously the points about naming need addressing, but this looks great to go after that has been resolved. Really good job.

multiaccounts/settings/database.go
qfrank: fixed

multiaccounts/settings/columns.go
qfrank: fixed

qfrank:
> Obviously the points about naming need addressing, but this looks great to go after that has been resolved. Really good job.

All fixed. Thank you for your review, may i get your approval now ? @Samyoul 😉

Samyoul:
@qfrank You've always had it. 😆 

https://github.com/status-im/status-go/pull/3385#pullrequestreview-1380910397
Screenshot 2023-04-12 at 11 32 19

multiaccounts/settings/database.go
```diff
@@ -287,7 +301,7 @@ func (db *Database) SetSettingLastSynced(setting SettingField, clock uint64) err
...
...ide_home_tooltip, installation_id, key_uid, keycard_instance_uid, keycard_paired_on, keycard_pairing, last_updated, latest_derived_path, link_preview_request_enabled, link_previews_enabled_sites, log_level, mnemonic, mnemonic_removed, name, networks, notifications_enabled, push_notifications_server_enabled, push_notifications_from_contacts_only, remote_push_notifications_enabled, send_push_notifications, push_notifications_block_mentions, photo_path, pinned_mailservers, preferred_name, preview_privacy, public_key, remember_syncing_choice, signing_phrase, stickers_packs_installed, stickers_packs_pending, stickers_recent_stickers, syncing_on_mobile_network, default_sync_period, use_mailservers, messages_from_contacts_only, usernames, appearance, profile_pictures_show_to, profile_pictures_visibility, wallet_root_address, wallet_set_up_passed, wallet_visible_tokens, waku_bloom_filter_mode, webview_allow_permission_requests, current_user_status, send_status_updates, gif_recents, gif_favorites, opensea_enabled, last_backup, backup_enabled, telemetry_server_url, auto_message_enabled, gif_api_key, test_networks_enabled, mutual_contact_enabled FROM settings WHERE synthetic_id = 'id'").Scan(
```
saledjenic, line 304: @qfrank could we just do `err := db.db.QueryRow("SELECT * FROM settings WHERE synthetic_id = 'id'")` instead?

saledjenic (commented):
sorry for being late :)

multiaccounts/settings/database.go
qfrank: > @qfrank could we just do `err := db.db.QueryRow("SELECT * FROM settings WHERE synthetic_id = 'id'")` instead?

LGTM if we take care of Scan 👍 but this PR already get merged 🥹

saledjenic: Yeah I know, but just saying.

qfrank: if we need add a new column to settings in future PR, maybe it's a good time to do the change

PR #3387, opened 2023-04-11
Community request to join changes for pending and declined states

smohamedjavid:
### Summary

- If the request to join a community is ignored or unattended by the Admin/Community Manager for more than seven (7) days, the request is removed from both Admin and Requester databases
- The community "Join request" notification from Admin/Community Manager's Activity Center is deleted after seven (7) days
- If the request is declined by the Admin/Community Manager, giving the Sender/Requester ability to request again after seven (7) days

related https://github.com/status-im/status-mobile/issues/15322

Mobile PR: https://github.com/status-im/status-mobile/pull/15627


protocol/communities/request_to_join.go
```diff
@@ -57,3 +62,18 @@ func (r *RequestToJoin) InitFromSyncProtobuf(proto *protobuf.SyncCommunityReques
...
+	requestToJoinClock, err := strconv.ParseInt(fmt.Sprint(clock), 10, 64)
+	if err != nil {
+		return 0, err
+	}
+
+	// Adding 7 days to request
+	// requestTimeOutClock := uint64(time.Unix(requestToJoinClock, 0).AddDate(0, 0, 7).Unix())
+
+	// For testing on client - setting 5 minutes as timeout
+	requestTimeOutClock := uint64(time.Unix(requestToJoinClock, 0).Add(time.Minute * 5).Unix())
```
jrainville: You forgot to switch the comments

protocol/messenger_communities.go
```diff
@@ -992,6 +991,73 @@ func (m *Messenger) leaveCommunity(communityID types.HexBytes) (*MessengerRespon
...
+	response := &MessengerResponse{}
+
+	for _, requestToJoin := range pendingRequestsToJoin {
+
+		requestTimeOutClock, err := communities.AddTimeoutToRequestToJoinClock(requestToJoin.Clock)
+		if err != nil {
+			return nil, err
+		}
+
+		timeNow := uint64(time.Now().Unix())
```
jrainville: It doesn't matter much, but `timeNow` can be declared above the loop.

protocol/messenger.go
```diff
@@ -1397,6 +1398,24 @@ func (m *Messenger) watchIdentityImageChanges() {
 	}()
 }
 
+func (m *Messenger) watchPendingCommunityRequestToJoin() {
+	m.logger.Debug("watching community request to join")
+
+	go func() {
+		for {
+			select {
+			case <-time.After(time.Minute):
```
jrainville: Does it do it every minute? maybe we can increase it, since it's not critical to do it that often?

jrainville (commented):
Good job. Just minor comments

protocol/communities/request_to_join.go
smohamedjavid: The Mobile PR is open, and we need this 5-minute timeout (requested by the QA team) until the testing is completed.

I will remove this 5-minute timeout and will replace/uncomment the required seven days timeout once the testing is completed.

protocol/messenger.go
smohamedjavid: Yes, It will check every minute. I and @cammellos initially agreed to do it every minute because if the admin opens the mobile/desktop near the request timeout, a shorter polling time ensures the request is deleted.

jrainville: Is there a way to do it kind of like a `do/while`, where it does it the first time immediately (or even better after a short delay), and then does it every, let's say, 10 minutes?
This way, we ensure to delete the requests at open, but also don't do a check on all of them each minute?
Maybe it's not that big of a deal either TBH. It's just that we have/had performance issues, so limiting DB calls as much as possible is good.

smohamedjavid: I agree we should limit the DB calls. I guess 10 minutes should be fine, given that the timeout is seven days.

We expose the `CheckAndDeletePendingRequestToJoinCommunity` method in `PublicAPI` so the client can call it at the successful login. This would clear any pending requests which are expired/timed out.

I will update the polling to 10 minutes 👍

protocol/messenger_communities.go
```diff
@@ -992,6 +991,71 @@ func (m *Messenger) leaveCommunity(communityID types.HexBytes) (*MessengerRespon
...
+
+			// Delete activity centre notification for community admin
+			notification, err := m.persistence.GetActivityCenterNotificationByID(requestToJoin.ID)
+			if err != nil {
+				m.logger.Error("failed to fetch pending request to join", zap.Error(err))
+				return nil, err
+			}
+
+			if notification != nil && !notification.Deleted && notification.Type == ActivityCenterNotificationTypeCommunityMembershipRequest {
+				err = m.persistence.DeleteActivityCenterNotification(types.FromHex(requestToJoin.ID.String()))
```
ilmotta: I have the impression we should be running these DB operations inside a transaction. In which state is the system left if any error happens after `DeletePendingRequestToJoin` has succeeded but `DeleteActivityCenterNotification` failed?

I think each processed `requestToJoin` in the loop should either succeed or not, and if any failure happens the changes in the DB should be rolled back and the loop can attempt to process the next `requestToJoin` in a clean slate. Or maybe better, short circuit everything in case of failure and rollback any changes to the DB, for all instances in `pendingRequestsToJoin`.

ilmotta (approved):
Looks great @smohamedjavid. I left a comment for what I think is important in terms of reliability, but it's not a blocker.

protocol/messenger_communities.go
smohamedjavid: I agree it's not a best practice to execute multiple DB operations without a transaction. I would love to run all DB operations inside a single transaction for this feature but `protocol/communities/persistence` and `protocol/activity_center_persistence` are in a different package. 

It's creating a circular dependency if we try to execute DB operations in a single transaction. Either way, updating the AC table from the same communities package (under persistence) or moving the AC related DB operation to AC go file with receiving transaction pointer. GO has a low tolerance towards circular dependency and will produce a compile-time error.

We need to refactor/move the code to prevent this and have all DB queries running in a single transaction.

I suggest we need to look into this in a separate issue.

ilmotta: Thanks for attempting this improvement @smohamedjavid.

> I suggest we need to look into this in a separate issue.

I agree with you, it's too much for this PR, so let's break down the work into smaller chunks.

MishkaRogachev (approved):
Looks god from my perspective

siddarthkay (approved):

PR #3388, opened 2023-04-12
fix mobile mention issue #15616

qfrank:
[relate mobile issue](https://github.com/status-im/status-mobile/issues/15616)

status: ready.


ghost:
### Pull Request Checklist

Samyoul (approved):

ilmotta (approved):

qfrank:
added function `diffText` in this PR so that we don't have to depend on event function `on-text-input` on the frontend anymore, [`on-text-input`](https://github.com/status-im/status-react/blob/8af7011eb4b01925fac7f59eb8c5592309759d43/src/status_im/ui2/screens/chat/composer/input.cljs#L149) and [`on-change`](https://github.com/status-im/status-react/blob/8af7011eb4b01925fac7f59eb8c5592309759d43/src/status_im/ui2/screens/chat/composer/input.cljs#L121) have different behaviors on ios and android,  e.g. when user delete chars quickly, on-text-input and on-change won't be called at the same times. 
for `on-text-input`, it also behaves different between ios and android when user do the same input order, e.g. user type `Abc` in the input text field, we may get different result for 
```
{:new-text .., :previous-text .., :start .., :end ..}
```
`diffText` simplified the complex of mention, all we need to know is what user previously input and later

PR #3392, opened 2023-04-13
Update to FilterV2

### LINKED ISSUE #3279: waku2: use filterv2
Use improved filter implementation replacing the following pieces of code:
- https://github.com/status-im/status-go/blob/develop/wakuv2/waku.go#L290
- https://github.com/status-im/status-go/blob/develop/wakuv2/waku.go#L741-L748
- https://github.com/status-im/status-go/blob/develop/wakuv2/waku.go#L1069-L1071

See https://github.com/waku-org/go-waku/blob/master/waku/v2/protocol/filterv2/filter_test.go#L62 for usage

cc: @cammellos

vitvly:
Fixes #3279 

This PR adds support for multiple filter peer connections (upper limit set by `MinPeersForFilter`).

`WakuFilterLightnode.IsSubscriptionAlive` is used in `Waku.runFilterMsgLoop` in order to try to maintain a minimum specified number of active filter peer connections.

### DIFF
--- a/wakuv2/waku.go
+++ b/wakuv2/waku.go
@@ -28,6 +28,7 @@ import (
 	"math"
 	"net"
 	"runtime"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -55,7 +56,7 @@ import (
 
 	"github.com/waku-org/go-waku/waku/v2/dnsdisc"
 	"github.com/waku-org/go-waku/waku/v2/protocol"
-	"github.com/waku-org/go-waku/waku/v2/protocol/legacy_filter"
+	"github.com/waku-org/go-waku/waku/v2/protocol/filter"
 	"github.com/waku-org/go-waku/waku/v2/protocol/peer_exchange"
 	"github.com/waku-org/go-waku/waku/v2/protocol/relay"
 
@@ -80,6 +81,7 @@ const bootnodesMaxRetries = 7
 type settings struct {
 	LightClient         bool   // Indicates if the node is a light client
 	MinPeersForRelay    int    // Indicates the minimum number of peers required for using Relay Protocol
+	MinPeersForFilter   int    // Indicates the minimum number of peers required for using Filter Protocol
 	MaxMsgSize          uint32 // Maximal message length allowed by the waku node
 	EnableConfirmations bool   // Enable sending message confirmations
 	PeerExchange        bool   // Enable peer exchange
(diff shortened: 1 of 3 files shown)


ghost:
### Pull Request Checklist

richard-ramos:
Looks good!
btw. FilterV2 includes also a 'ping' functionality which can be used to determine if a subscription is active or not. Do you think we should run this ping periodically against the nodes we're subscribed to?

vitvly:
@richard-ramos yes, good! I see, `IsSubscriptionAlive` there is.

vitvly:
@richard-ramos @cammellos updated the PR with latest changes

wakuv2/waku.go
```diff
@@ -632,41 +639,66 @@ func (w *Waku) runFilterMsgLoop() {
...
+							_ = envelopeErrors
+							_ = err
+						}
+					default:
+						// Use an empty default so that reading from each sub.C is non-blocking
+					}
+				}
 			}
 		}
 	}
```
richard-ramos: I think it would be more performant to remove this `default` and do the filter subscription reading by creating a separate goroutine for each filter subscribtion created (in line 668). Something +- like this:

```go
go func(sub *FilterSubscription){
    for {
          select {
              case <-w.quit:
                   return
              case env, ok := <-sub.C:
                   if ok {
                                  envelopeErrors, err := w.OnNewEnvelopes(env, common.RelayedMessageType)
                                  // TODO: should these be handled?
                                  _  = envelopeErrors
                                  _  = err
                            }
          }
    }

}(subDetails)
```
because then you wouldnt have to worry about reading the channel being blockng and you'd have independent 'workers' reading envelopes from the subscriptions.

richard-ramos (commented):
Cool PR!

wakuv2/waku.go
vitvly: Yes, thanks, agree, done.

vitvly: Will add a test for a case when `MinPeersForFilter == 2`. This should rather be renamed to `MaxPeersForFilter` it seems.

vitvly: Added a test for filter re-subscriptions.

vitvly:
@richard-ramos fixed the test, some tests were flaky, so required a re-run.

richard-ramos (approved):

PR #3393, opened 2023-04-14
feat(edit-message): Edit image messages

borismelnik:
Adds possibility to change text for image messages

Important changes:

### DIFF
--- a/protocol/messenger.go
+++ b/protocol/messenger.go
@@ -6293,20 +6293,21 @@ func (m *Messenger) GetMentionsManager() *MentionManager {
 	return m.mentionsManager
 }
 
-func (m *Messenger) getMessagesToDelete(message *common.Message, chatID string) ([]*common.Message, error) {
-	var messagesToDelete []*common.Message
+func (m *Messenger) getConnectedMessages(message *common.Message, chatID string) ([]*common.Message, error) {
+	var connectedMessages []*common.Message
 	// In case of Image messages, we need to delete all the images in the album
 	if message.ContentType == protobuf.ChatMessage_IMAGE {
 		image := message.GetImage()
-		messagesInTheAlbum, err := m.persistence.albumMessages(chatID, image.GetAlbumId())
-		if err != nil {
-			return nil, err
+		if image != nil && image.AlbumId != "" {
+			messagesInTheAlbum, err := m.persistence.albumMessages(chatID, image.GetAlbumId())
+			if err != nil {
+				return nil, err
+			}
+			connectedMessages = append(connectedMessages, messagesInTheAlbum...)
+			return connectedMessages, nil
 		}
-		messagesToDelete = append(messagesToDelete, messagesInTheAlbum...)
-	} else {
-		messagesToDelete = append(messagesToDelete, message)
 	}
-	return messagesToDelete, nil
+	return append(connectedMessages, message), nil
 }
 
 func (m *Messenger) withChatClock(callback func(string, uint64) error) error {
--- a/protocol/messenger_messages.go
+++ b/protocol/messenger_messages.go
@@ -40,88 +40,17 @@ func (m *Messenger) EditMessage(ctx context.Context, request *requests.EditMessa
 	// A valid added chat is required.
 	chat, ok := m.allChats.Load(message.ChatId)
 	if !ok {
-		return nil, errors.New("Chat not found")
+		return nil, ErrChatNotFound
 	}
 
-	clock, _ := chat.NextClockAndTimestamp(m.getTimesource())
-
-	editMessage := &EditMessage{}
-
-	editMessage.Text = request.Text
-	editMessage.ContentType = request.ContentType
-	editMessage.ChatId = message.ChatId
-	editMessage.MessageId = request.ID.String()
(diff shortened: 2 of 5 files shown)


ilmotta (commented):
The implementation looks good, just missing the accompanying tests ;)

protocol/messenger_messages.go
```diff
@@ -84,12 +84,84 @@ func (m *Messenger) EditMessage(ctx context.Context, request *requests.EditMessa
...
+	messages, err := m.persistence.AlbumMessages(chatID, albumID)
+	if err != nil {
+		return nil, err
+	}
+	var result []*common.Message
+	for _, message := range messages {
+		// A valid added chat is required.
+		chat, ok := m.allChats.Load(message.ChatId)
+		if !ok {
+			return nil, errors.New("Chat not found")
```
jrainville: Don't we have a constant for this error? If not, you can create it :smiley:

protocol/messenger_messages.go
```diff
@@ -84,12 +84,84 @@ func (m *Messenger) EditMessage(ctx context.Context, request *requests.EditMessa
 	}
```
jrainville: I can't comment higher than this, but basically, what is above here is also being done in `editMessagesFromAlbum` (disatching and updating LastMessage). So it might be worth to put the code above in the `else` of the condition below for `ChatMessage_IMAGE`

jrainville (commented):
Does the handler already handle correctly the edits on image messages? I'm asking because you didn't modify that part.
What I mean is, when we receive the signal fro editing an image message in another instance, does it work?

borismelnik:
> Does the handler already handle correctly the edits on image messages? I'm asking because you didn't modify that part.
> What I mean is, when we receive the signal fro editing an image message in another instance, does it work?

Actually yes, I checked it. We just send a several `editMessage` cos we have to change the text for every message in album

borismelnik:
> The implementation looks good, just missing the accompanying tests ;)

oh... I did it, just not pushed it :D

ilmotta (approved):

protocol/messenger_messages.go
borismelnik: missed this comment... what do you mean?

jrainville: If you check the code above, lines 56 to 84, it's basically the same thing that you do in your function  
`editMessagesFromAlbum`.

I initially suggested to put the code above inside an `else` below `if message.ContentType == protobuf.ChatMessage_IMAGE `, but I have a better idea.

We can merge the code you did in `editMessagesFromAlbum` and move it to this function.
Basically, we'll do something very similar to what I did for the DeleteMessage for images.
You can see the code here: https://github.com/status-im/status-go/blob/develop/protocol/messenger_handler.go#L1587-L1624

First, I get an array of messages, if it's an image message, I get all images part of the album, otherwise I just put the one message in the array.

Then I perform the delete actions in a loop, similarly to what you did in `editMessagesFromAlbum`.

If you want, you can probably reuse `getMessagesToDelete`, maybe just find another name like `getAllMessagesAssociatedWithThisMessage`.

You can probably improve that function too. It doesn't do the check you did here ` if image != nil && image.AlbumId != ""`

And that's it.

The goal is just to reduce the amount of duplicated code, since both your function and the code above does the same thing basically.

borismelnik: I think I got it :)

protocol/messenger.go
```diff
@@ -6259,16 +6259,18 @@ func (m *Messenger) GetMentionsManager() *MentionManager {
...
-		messagesInTheAlbum, err := m.persistence.albumMessages(chatID, image.GetAlbumId())
-		if err != nil {
-			return nil, err
+		if image != nil && image.AlbumId != "" {
+			messagesInTheAlbum, err := m.persistence.albumMessages(chatID, image.GetAlbumId())
+			if err != nil {
+				return nil, err
+			}
+			messagesToDelete = append(messagesToDelete, messagesInTheAlbum...)
 		}
```
jrainville, line 6308: I think we need an `else` here to append the original `message`. So if we have no album, we just send the simple message so that it can be edited or deleted.

If you want to reduce duplication, you can do something like this instead:

```go
func (m *Messenger) getConnectedMessages(message *common.Message, chatID string) ([]*common.Message, error) {
	var messagesToDelete []*common.Message
	// In case of Image messages, we need to delete all the images in the album
	if message.ContentType == protobuf.ChatMessage_IMAGE {
		image := message.GetImage()
                if image != nil && image.AlbumId != "" {
			messagesInTheAlbum, err := m.persistence.albumMessages(chatID, image.GetAlbumId())
			if err != nil {
				return nil, err
			}
			messagesToDelete = append(messagesToDelete, messagesInTheAlbum...)
                        return messagesToDelete
		}
         }
         return append(messagesToDelete, message), nil
}
```

Also, rename `messagesToDelete` to `connectedMessages` :sweat_smile:

jrainville (changes requested):
Yup that's exactly what I meant. Good job.

I'm requesting changes, because I found an issue in the code I suggested. See inline for details.

jrainville (approved):

PR #3400, opened 2023-04-17
feat: add api to get list of owners from a given nft contract

dlipicar:
Fixes #10290

Adds an API to retrieve the list of addresses and balances of NFT owners from a certain contract address. This is going to be used by community owners for NFT-gated communities.

Alchemy provides this endpoint but their free plan does not allow access to more than 5 chains per account. Since a different API key is used for each chain, a map of keys-per-chain is required.
Infura has currently support for Eth (Mainnet and Goerli) and Arbitrum (Mainnet). Not only the token is required for this endpoint, the token secret is needed as well.

### DIFF
--- a/node/status_node_services.go
+++ b/node/status_node_services.go
@@ -56,8 +56,6 @@ var (
 	// ErrRPCClientUnavailable is returned if an RPC client can't be retrieved.
 	// This is a normal situation when a node is stopped.
 	ErrRPCClientUnavailable = errors.New("JSON-RPC client is unavailable")
-	// OpenseaKeyFromEnv passed from on env var during build time
-	OpenseaKeyFromEnv string
 )
 
 func (b *StatusNode) initServices(config *params.NodeConfig, mediaServer *server.MediaServer) error {
@@ -132,11 +130,7 @@ func (b *StatusNode) initServices(config *params.NodeConfig, mediaServer *server
 	}
 
 	if config.WalletConfig.Enabled {
-		openseaKey := config.WalletConfig.OpenseaAPIKey
-		if len(openseaKey) == 0 {
-			openseaKey = OpenseaKeyFromEnv
-		}
-		walletService := b.walletService(accDB, accountsFeed, openseaKey, config.WalletConfig.AlchemyAPIKeys, config.WalletConfig.InfuraAPIKey, config.WalletConfig.InfuraAPIKeySecret)
+		walletService := b.walletService(accDB, accountsFeed)
 		services = append(services, walletService)
 	}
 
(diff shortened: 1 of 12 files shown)


ghost:
### Pull Request Checklist

services/wallet/thirdparty/alchemy/client.go
```diff
@@ -0,0 +1,152 @@
...
+	return err == nil
+}
+
+func (o *Client) FetchNFTOwnersByContractAddress(chainID uint64, contractAddress common.Address) (*thirdparty.NFTContractOwnership, error) {
+	queryParams := url.Values{
+		"contractAddress":   {contractAddress.String()},
+		"withTokenBalances": {"true"},
+	}
+
+	url, err := getBaseURL(chainID, o.apiKeys[chainID])
```
IvanBelyakoff: Just a suggestion,  imo `getBaseUrl` does not need to know API keys, since it should only return the base part. After that you still modify that url to get the complete url, so another function like `buildUrl(baseUrl, queryString, <optional API keys>)` could do the job.

IvanBelyakoff (approved):
Overall LGTM. Just I'd split some long methods into multiple to make them more cohesive and more testable

services/wallet/thirdparty/alchemy/client.go
```diff
@@ -0,0 +1,152 @@
...
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/status-im/status-go/services/wallet/bigint"
+	"github.com/status-im/status-go/services/wallet/thirdparty"
+)
+
+func getBaseURL(chainID uint64, apiKey string) (string, error) {
+	switch chainID {
+	case 1, 5, 11155111:
```
endulab: I think we need some global const ids for chains.

node/status_node_services.go
```diff
@@ -136,7 +136,7 @@ func (b *StatusNode) initServices(config *params.NodeConfig, mediaServer *server
 		if len(openseaKey) == 0 {
 			openseaKey = OpenseaKeyFromEnv
 		}
-		walletService := b.walletService(accDB, accountsFeed, openseaKey)
+		walletService := b.walletService(accDB, accountsFeed, openseaKey, config.WalletConfig.AlchemyAPIKeys, config.WalletConfig.InfuraAPIKey, config.WalletConfig.InfuraAPIKeySecret)
```
endulab: Maybe it's better to pass config.WalletConfig ?

endulab (approved):

node/status_node_services.go
dlipicar: Yeap... The `OpenseaKeyFromEnv` thing was needed before the Mobile client passed the key through the config, but it's doing it now so I'll remove that and pass the struct to make it cleaner

services/wallet/thirdparty/alchemy/client.go
dlipicar: Super-agree. What's the right place for that in status-go? Never ventured too far from the wallet service

node/status_node_services.go
dlipicar: Done

services/wallet/thirdparty/alchemy/client.go
dlipicar: Done

services/wallet/thirdparty/alchemy/client.go
dlipicar: Split the function so that `getBaseURL` has a single responsibility

saledjenic (approved):

PR #3403, opened 2023-04-17
fix: set default value in sync period column to 9d

richard-ramos:
In https://github.com/status-im/status-go/pull/3391 only existing accounts are modified. With this PR new accounts will have the default sync period set to 9d as well. Since the sqlite/sqlcipher version we use does not support modifying columns, i recreate the settings table with the new default value for the sync period column


ghost:
### Pull Request Checklist

jrainville (commented):
Why does the other migration you did not work for new accounts and this one does?

richard-ramos:
> Why does the other migration you did not work for new accounts and this one does?

Because when a new account is created, the SQL that inserts a row in `settings` table does not set a value for the sync period, so it defaults to the value from the table schema: `default_sync_period INTEGER DEFAULT 86400`.

jrainville (approved):
> Because when a new account is created, the SQL that inserts a row in settings table does not set a value for the sync period, so it defaults to the value from the table schema: default_sync_period INTEGER DEFAULT 86400.

Ah ok i get it. So creating a new table and copying in the old one makes sure the default value is 9 days instead.

PR #3407, opened 2023-04-18
feat:support backup/sync social settings

qfrank:
RT


protocol/messenger_backup_handler.go
```diff
@@ -136,6 +136,13 @@ func (m *Messenger) handleBackedUpProfile(message *protobuf.BackedUpProfile, bac
 		}
 	}
 
+	for _, s := range message.SocialLinkSettings {
+		err = m.handleSyncSocialLinkSetting(state, *s)
```
saledjenic: @qfrank why don't you use `WakuBackedUpDataResponse` but adding these details to `ReceivedMessageState`?

saledjenic: I mean if you are backing it up as a part of `protobuf.BackedUpProfile`, why don't you handle it that way as well and keep the separation between handling data coming from backed up messages against data coming from syncing?

qfrank: actually, i was struggled with this before, i agree you now 😂 @saledjenic thank you

multiaccounts/settings_social_links/database.go
```diff
@@ -87,3 +115,31 @@ func (s *SocialLinksSettings) SetSocialLinks(links *identity.SocialLinks) (err e
 
 	return nil
 }
+
+func (s *SocialLinksSettings) UpdateSocialLinkFromSync(link *identity.SocialLink) error {
```
osmaczko, line 119: If possible, could you please cover this with tests in `multiaccounts/settings_social_links/database_test.go`?

protocol/messenger.go
```diff
@@ -6260,3 +6274,66 @@ func (m *Messenger) getMessagesToDelete(message *common.Message, chatID string)
...
+	clock, chat := m.getLastClockWithRelatedChat()
+	err := callback(chat.ID, clock)
+	if err != nil {
+		return err
+	}
+	chat.LastClockValue = clock
+	return m.saveChat(chat)
+}
+
+func (m *Messenger) syncSocialSettings(ctx context.Context, rawMessageHandler RawMessageHandler) error {
```
osmaczko: It was not clear to me from the code itself how the messages are dispatched, please consider this small readability improvement.
```suggestion
func (m *Messenger) syncSocialSettings(ctx context.Context, rawMessageDispatcher RawMessageHandler) error {
```

osmaczko (approved):
Looks good :+1: 

(as long as Sale's suggestions are applied)

protocol/messenger_backup_handler.go
qfrank: fixed @saledjenic

multiaccounts/settings_social_links/database.go
qfrank: because there is already a [test](https://github.com/status-im/status-go/blob/e72314c2ec5293c15dfdf594d65881d622ad1352/protocol/messenger_backup_test.go#L143) which will call `UpdateSocialLinkFromSync` internally, so i ignored it, is it okay? @osmaczko

protocol/messenger.go
qfrank: fixed @osmaczko

protocol/messenger_backup_handler.go
```diff
@@ -67,7 +69,7 @@ func (m *Messenger) HandleBackup(state *ReceivedMessageState, message protobuf.B
 	return errors
 }
 
-func (m *Messenger) handleBackedUpProfile(message *protobuf.BackedUpProfile, backupTime uint64) error {
+func (m *Messenger) handleBackedUpProfile(state *ReceivedMessageState, message *protobuf.BackedUpProfile, backupTime uint64) error {
```
saledjenic: We don't need `state *ReceivedMessageState`

protocol/messenger_backup_handler.go
```diff
@@ -132,9 +134,20 @@ func (m *Messenger) handleBackedUpProfile(message *protobuf.BackedUpProfile, bac
...
+	var links identity.SocialLinks
+	for _, s := range message.SocialLinkSettings {
+		err = m.handleSyncSocialLinkSetting(*s, func(link *identity.SocialLink) {
+			links = append(links, *link)
+		})
+		if err != nil {
+			return err
 		}
 	}
+	response.SetSocialLinks(links)
```
saledjenic, line 146: Add only `if len(links) > 0` and set `contentSet` in that case.

qfrank: i don't understand why we need `if len(links) > 0` , could you explain more?

saledjenic: If there are no set images and there are no set links, why would we send a signal then?

saledjenic: ahh I see that DisplayName is always set, ok, then we can remove `contentSet` completely. It was needed for previous implementation.

qfrank: oh, i didn't see the variable `contentSet` 🤦‍♂️

qfrank: emm... looks like `contentSet` is always true after line 87 ?

saledjenic: yes, that's why we can remove it (was needed before)

qfrank: fixed ❤️

saledjenic (approved):

PR #3408, opened 2023-04-18
feat: use 24h batches for retrieving message history

richard-ramos:


Samyoul (approved):
Nice PR. Is it possible to test this functionality, just to ensure that a future developer doesn't break it and we don't notice?

richard-ramos:
I asked here about how to use the mailserver cycle in tests. https://github.com/status-im/status-go/pull/3420#issuecomment-1525711554 I believe once we're able to do that, it should be easy to add a test unit to this PR

PR #3415, opened 2023-04-20
feat:support backup/sync ens name

qfrank:
RT


saledjenic (approved):

igor-sirotin:
FYI: I've previously introduced a new `ens_usernames` table, which replaces `settings.usernames`: https://github.com/status-im/status-go/pull/3066

The old `settings.usernames` still works and maintained by status-go, but desktop uses the new table now.

cc @cammellos

qfrank:
> FYI: I've previously introduced a new `ens_usernames` table, which replaces `settings.usernames`: #3066
> 
> The old `settings.usernames` still works and maintained by status-go, but desktop uses the new table now.
> 
> cc @cammellos

interesting, i have something to say, do we really need a table to store ens names? it won't have much records in this table i think, and `settings.usernames` is a BLOB type represented with JSON, we can extend it dynamicly, and table `settings` is very convinient to implement sync/backup now @cammellos @igor-sirotin @saledjenic @Samyoul @richard-ramos

cammellos:
> > FYI: I've previously introduced a new `ens_usernames` table, which replaces `settings.usernames`: #3066
> > The old `settings.usernames` still works and maintained by status-go, but desktop uses the new table now.
> > cc @cammellos
> 
> interesting, i have something to say, do we really need a table to store ens names? it won't have much records in this table i think, and `settings.usernames` is a BLOB type represented with JSON, we can extend it dynamicly, and table `settings` is very convinient to implement sync/backup now @cammellos @igor-sirotin @saledjenic @Samyoul @richard-ramos

We tend to prefer tables over JSON blobs, migrating json blobs requires user level migrations and it's very opaque and difficult to extend, so unless there's strong reason to use JSON and a single field, we prefer separate tables

qfrank:
@cammellos emm.. so now we need support sync/backup table `ens_usernames` as well?

saledjenic:
> interesting, i have something to say, do we really need a table to store ens names?

There are pros and cons for both approaches. I personally prefer having a table and a single column for each data, that gives a granularity, a possibility to write highly customized queries with different views of the database and prepare data as you need on the db level with high speed and without additional coding.

> so now we need support sync/backup table ens_usernames as well?

We definitely need it, but since mobile app is still using `settings.usernames` I thought you wanted to sync only that and leave to somebody from the desktop team to handle `ens_usernames`.

qfrank:
> We definitely need it, but since mobile app is still using settings.usernames I thought you wanted to sync only that and leave to somebody from the desktop team to handle ens_usernames.

i'm going to do it in this PR. @saledjenic  cc @igor-sirotin

igor-sirotin:
Thanks @qfrank ❤️. I wasn't able to catch up previous week.
You can leave it for me as well, but I think you will implement it much faster, as you've been around it for awhile already

services/ens/database.go
```diff
@@ -8,21 +8,29 @@ type Database struct {
 	db *sql.DB
 }
 
-type UsernameDetails struct {
+type UsernameDetail struct {
```
igor-sirotin, line 11: Surely this is a minor change, but does it make sense? It's not just 1 detail inside 🤔

protocol/protobuf/sync_settings.proto
```diff
@@ -31,6 +31,7 @@ message SyncSetting {
     DISPLAY_NAME = 13;
     BIO = 14;
     MNEMONIC_REMOVED = 15;
+    USERNAMES = 16;
```
igor-sirotin: Maybe `ENS_USERNAMES` would be better here?

igor-sirotin (approved):
LGTM, thank you! 👍

protocol/messenger.go
```diff
@@ -2902,6 +2907,78 @@ func (m *Messenger) SyncBookmark(ctx context.Context, bookmark *browsers.Bookmar
...
+
+	rawMessage := common.RawMessage{
+		LocalChatID:         contactIDFromPublicKey(&m.identity.PublicKey),
+		Payload:             encodedMessage,
+		MessageType:         protobuf.ApplicationMetadataMessage_SYNC_ENS_USERNAME_DETAIL,
+		ResendAutomatically: true,
+	}
+
+	_, err = rawMessageHandler(ctx, rawMessage)
+	if err != nil {
```
saledjenic: @qfrank minor thing, you could just `return err` here.

services/ens/database.go
```diff
@@ -8,21 +8,29 @@ type Database struct {
...
 	Username string `json:"username"`
 	ChainID  uint64 `json:"chainId"`
+	Clock    int64  `json:"clock"`
```
saledjenic: `uint64` maybe?

saledjenic (approved):

services/ens/database.go
qfrank: make sense when something like `var usernameDetails []*UsernameDetail` comes to me

> It's not just 1 detail inside

if you mean username|chainId are details, IMO, username/chainId could belongs to UsernameDetail, each of them is part of UsernameDetail. @igor-sirotin

PR #3416, opened 2023-04-20
feat: sync and backup wallet accounts

saledjenic:
Changes applied here introduce:
- improvements to sync wallet accounts among devices (including all account types)
- backing up wallet accounts to and fetch them from waku (an information about received wallet accounts is sent via `waku.backedup.wallet-account` signal to a client)


protocol/messenger.go
```diff
@@ -2485,49 +2488,66 @@ func (m *Messenger) DeleteAccount(address types.Address) error {
...
 	accountMessages := make([]*protobuf.SyncWalletAccount, 0)
 	for _, acc := range accs {
-		if acc.Type != accounts.AccountTypeWatch &&
-			acc.Type != accounts.AccountTypeGenerated {
+		if acc.Chat {
 			continue
 		}
+
+		// Once mobile app supports seed phrase and private key imported accounts we should remove the following `if` block
+		if isMobileApp {
```
cammellos: Could you elaborate why we need this?

saledjenic: There are 2 contexts where `isMobileApp` is used:
- when the app is going to dispatch a wallet accounts message and 
- when the app is going to handle a received wallet accounts message (but not sure how mobile app sets the account's type for all possible accounts' types and that's why I added this `if` here to keep the code as it was in case of mobile app).

My intention was to check `isMobileApp` to avoid any potential issues in mobile app and keep the code as it was before cause as far as I know mobile app is handling differently some account types (was discussed [here](https://github.com/status-im/status-go/pull/3245#discussion_r1122536029)) and not sure that it supports derivation from imported accounts and that it has import private key feature.

saledjenic: By checking `isMobileApp` the following sync is done:
- `mobile` to `mobile`, mobile app sends default wallet acc, generated acc from the default wallet account and wo accounts, other mobile app handles them
- `mobile` to `desktop`, mobile app sends default wallet acc, generated acc from the default wallet account and wo accounts, desktop app handles them cause it's able to handle any account type
- `desktop` to `mobile`, desktop app sends all accounts, mobile app handles only default wallet acc, generated acc from the default wallet account and wo accounts
- `desktop` to `desktop`, one desktop app sends all accounts, another desktop app handles them all cause it's able to handle any account type

saledjenic: Also you can see that `LatestDerivedPath` is set and used only in case of mobile app, the desktop app doesn't use it. Once the mobile app aligns with the desktop app in terms of wallet accounts we will need just to remove those `if`s (that's why I added comments next to them).

protocol/messenger_handler.go
```diff
@@ -2762,55 +2767,90 @@ func (m *Messenger) updateUnviewedCounts(chat *Chat, mentionedOrReplied bool) {
...
+	// Once mobile app supports seed phrase and private key imported accounts we should remove the following `if` block
+	if isMobileApp {
+		accType := accounts.AccountType(message.Type)
+		if accType != accounts.AccountTypeWatch &&
+			accType != accounts.AccountTypeGenerated {
+			return nil, ErrWalletAccountNotSupportedForMobileApp
+		}
+	}
+
+	if dbAccount != nil && message.Clock <= dbAccount.Clock {
```
qfrank, line 2794: Could you explain `message.Clock`? `message.Clock` is generated by m.getLastClockWithRelatedChat(), so looks like it's always greater than the clock in db, what if dbAccount.Clock get updated by user on device B, then we received the backup message from device A? assuming we have the same account on device A/B @saledjenic

protocol/messenger.go
```diff
@@ -77,6 +78,8 @@ const (
 
 	publicChat  chatContext = "public-chat"
 	privateChat chatContext = "private-chat"
+
+	isMobileApp = runtime.GOOS == "android" || runtime.GOOS == "ios"
```
qfrank: not sure if this will work, have you tested? 🙂

protocol/messenger_handler.go
saledjenic: @qfrank if you look more detailed you will see that the code here, technically, was not changed at all. It's exactly the same as it was before, what I did actually is that I extracted one part from the old [HandleSyncWalletAccount](https://github.com/status-im/status-go/blob/1fe4a898b60694077c6df6fef352c71bb9b7d8eb/protocol/messenger_handler.go#L2765) function into `handleSyncWalletAccount` that I can reuse that function in `handleBackedUpWalletAccount` function. That's all.

> message.Clock is generated by m.getLastClockWithRelatedChat()

This is true.

> so looks like it's always greater than the clock in db

This is not entirely true. Cause there are 3 use cases for this part. 
1. `recover an account, applying previously backed up data` - since data were backed up in some point in the past after which user might update an account, in that case clock from the backed up message is older then the clock stored in the db.   Elaboration: since backed up messages are applied during onboarding (when user recovers an account) they should always be applied (cause at that moment the db is empty) but there is always a possibility that some messages come later, in meanwhile user may do some local changes and a message coming from the backup, related to the same account should be discarded in that case.
2. `recover an account - syncing with another device` - since this action is available during onboarding (when user recovers an account) and that only if devices are connected to the same local network then messages should always be applied (cause at that moment the db is empty)
3. `sync among paired devices` - in this case devices which are paired do not need to be close to each other, they communicate over network, so there is no guarantee that a change made on device A will be immediately visible on the device B, that means if user make a change on the device be in meantime after it has been made on the device A, when device B receives a change from the device A it should discard it, cause it has clock value less than value stored on the device B.

Hope that above clarifies possible confusion.

protocol/messenger.go
saledjenic: No, I haven't tested it, but don't see the reason why it should not work? (hope this is a temporary here, till mobile app aligns with the desktop app 🙂)
If you have any better idea to achieve the same let me know and I am ready to change it?

stefandunca: How about having this flag named after its final purpose (e.g. enableWalletAccountSync) and be set from the API by the mobile?

stefandunca (approved):

protocol/messenger.go
cammellos: I would get rid of `isMobile`, we don't currently use wallet at the moment, and it's better address any potential issue now, instead of later, when we forgot all about the code ;) 
I think it should just work fine, since wallet is not currently used, so as long as it's not crashing etc, we should be good, what do you think?

alaibe (approved):

protocol/messenger.go
```diff
@@ -2485,49 +2488,66 @@ func (m *Messenger) DeleteAccount(address types.Address) error {
...
+
+		// Once mobile app supports seed phrase and private key imported accounts we should remove the following `if` block
+		if isMobileApp {
+			if acc.Type != accounts.AccountTypeWatch &&
+				acc.Type != accounts.AccountTypeGenerated {
+				continue
+			}
+		}
 		var accountClock uint64
 		if acc.Clock == 0 {
```
cammellos: I am not sure we need this, if it was never modified, a clock of `0` should be fine (that's generally what we do, we only set a clock when things changes, if it's a default thing, the clock should be 0)
I think this is the same point as @qfrank 

Basically:

1) If something is defaulted, the clock should be set to 0
2) If the user makes an explicit action, the clock should be set to the time that decision was made

The code above, does not follow the rules above, since it looks like something might be defaulted (i.e has a clock of 0), but then the clock is set **at sync time**, which is not a pattern we generally use, and looks risky, since it might override an actual decision of the user on a different device.
Though I might be missing context.

protocol/messenger.go
saledjenic: @cammellos to me it sounds ok, I will remove that now and if somebody from mobile team could test if that doesn't break mobile app would be perfect.

protocol/messenger_handler.go
qfrank: thank you for detailed explaination ❤️ but i still have the same question as @cammellos [mentioned](https://github.com/status-im/status-go/pull/3416#discussion_r1173435288) , his comment might looks more clear:
> it might override an actual decision of the user on a different device.

protocol/messenger.go
saledjenic: If we change it here (when messages are dispatched), then we need to change it on the handling side [here](https://github.com/status-im/status-go/blob/1fe4a898b60694077c6df6fef352c71bb9b7d8eb/protocol/messenger_handler.go#L2779) (if it remains 0 it won't be applied with that condition then).

@cammellos that's the code which was already here, not written by me and I didn't want to change it cause I still do think that we don't need to change it and the reasons:
- we do perform a backup every 8hrs and if we would change that it means when user recovers Status account we will overwrite each wallet account as many times as many backed up messages are received instead just discard older messages and apply only newer ones
- hypothetically :) - if user has 2 devices (with the same seed phrase) which are not synced, device A has wallet account named "AccA", and device B has the same wallet account named "AccB" and each device does backup every 8hrs. Then device A is lost, and user decide to delete account on device B and recover it. Then while recovering if device B firstly receives backed up message from device B and set the wallet account name to "AccB", then receives backed up message from device A and overwrite that account cause both have clock set to 0, the end result will be that wallet account on device B has name "AccA", which is not what user would expect.

protocol/messenger.go
saledjenic: @cammellos thinking just now, what about this part [here](https://github.com/status-im/status-go/blob/76813cef3a992e75452d6b5fc2fccd0e53280e1b/protocol/messenger_handler.go#L2864-L2886)?
`LatestDerivedPath` from `settings` is used only in mobile app, the desktop app doesn't use it.

saledjenic: May I remove that and expect that mobile app will align with the desktop app soon?

cammellos: Yes, I think we should do that, since wallet in mobile is not used, we'll have to align with desktop before it's usable again

protocol/messenger.go
cammellos: > If we change it here (when messages are dispatched), then we need to change it on the handling side [here](https://github.com/status-im/status-go/blob/1fe4a898b60694077c6df6fef352c71bb9b7d8eb/protocol/messenger_handler.go#L2779) (if it remains 0 it won't be applied with that condition then).

Just to make sure we are on the same page, I am not arguing for changing the clock on dispatch, the opposite, clock should never changed when dispatching sync messages, it should only changed in response of a user decision. (i.e we should always use `acc.Clock`, never chat.nextClock, if clock is 0, there's an argument for not syncing that piece of data, but that's dependent on what we are syncing)

The code above though, sets the clock to `chat.nextClock` if the account.Clock is == 0, which means it is setting the clock at dispatch time, but that's something we generally don't do as I can't think of a case where that's necessary (but maybe I am missing something).

Are we on the same page on this point? (from your reply it looked to me that you think I am arguing for changing clock value at dispatch time)

saledjenic: > which means it is setting the clock at dispatch time

We don't set it to db for an account, but we set that `chat.nextClock` as `Clock` value only to dispatched message (one we're backing up) for that account.

How I understood you is that you're saying that we **_don't need_** to set `Clock` of `protobuf.SyncWalletAccount` to `chat.nextClock`, in case the db clock value is 0, at the moment when we're dispatching a message, but use `Clock` value which is set in the db for the account we're building a `protobuf.SyncWalletAccount` message.

Now what I am saying is, if we would set always a `Clock` value of `protobuf.SyncWalletAccount` to clock value from the `db`, then we would have many `protobuf.SyncWalletAccount` messages with the same `Clock` value which is `0`. That means at the moment user is doing an account recovering the app will overwrite an account in the db for each received backed up message, cause they will all have `0` value for `Clock`. (Points 1 and 2 from my previous comment).

cammellos: Ok, a question, when would an account have a clock value of 0?

saledjenic: As I see in the desktop app accounts have always `0` value for clock untill one of the following is true:
- the app is recovered from backed up messages
- the app is synced with another app (happens when user is recovering Status account from other device)
- the app is paired with another app and account is updated on that other app

cammellos: if we update the account locally, it should set the clock value right?

the cases you mentioned are only in response to a network event, but `clock` must be set locally at some point right?

saledjenic: > the cases you mentioned are only in response to a network event, but clock must be set locally at some point right?

That's not the case at the moment and has never been in the past as I see 🙂. When an account is updated locally the desktop app sets only `updated_at`, but that's because of [this line](https://github.com/status-im/status-go/blob/0eff61c57ae8fa1e80a3032e1ba8ada583598747/multiaccounts/accounts/database.go#L406)

saledjenic: Though doesn't status-go needs to care about that, not a client?

cammellos: mmh, looks like then there's a bit of an issue :) 

Anytime the user makes a decision about a piece of data, it should update the clock, I am not sure about the lifecycle of accounts, but if we can update them (i.e change name/emojii), we should set the clock to the time it has been updated, so that changes are correctly propagated to clients

cammellos: > Though doesn't status-go needs to care about that, not a client?

correct, status-go should set the clock, `saveAccounts` though is a bit of an annoying endpoint, since we don't really know what has changed, it would depend on the usage I think

saledjenic: I fully agree with all you've said in the last 2 messages and going to update that now.

Then I guess it does make sense to set `Clock` for newly created accounts as well? And then we can remove the discutable check if db clock is `0` or not, where we started this discussion from.

protocol/messenger.go
saledjenic: `isMobileApp` is completely removed in [this commit](https://github.com/status-im/status-go/pull/3416/commits/6d0b1b04d050e8f7fae1426397153257fb91486f)

protocol/messenger.go
cammellos: > Then I guess it does make sense to set Clock for newly created accounts as well? And then we can remove the discutable check if db clock is 0 or not, where we started this discussion from.

It would generally make sense, yes, unless it's a default account generated without user input (i.e no need to backup/sync)

saledjenic: @cammellos related changes are applied in [this commit](https://github.com/status-im/status-go/pull/3416/commits/dfa3cbbac0aed4ce2aee7e3779b3baeb6e11274a).

- setting clock when adding/updating accounts
- syncing/backing up accounts with the time they actually were updated instead the time when dispatching is done as it was before

protocol/messenger.go
```diff
@@ -2461,6 +2461,10 @@ func (m *Messenger) SyncDevices(ctx context.Context, ensName, photoPath string,
 }
 
 func (m *Messenger) SaveAccounts(accs []*accounts.Account) error {
+	clock, _ := m.getLastClockWithRelatedChat()
+	for _, acc := range accs {
```
cammellos: Unfortunately this might not be correct, say that you updated a single field on a single account, then each account will have their clock modified, and they will overwrite potential changes on other accounts.

Basically this assumes that Accounts (all of them) are a single entity and synchronize together, which is a bit risky. 

How do we update accounts, do you have a link to desktop?

saledjenic: The only place where this `(m *Messenger) SaveAccounts` function is called from is from the `(api *API) SaveAccounts` that further means it is used by client only.

Even `SaveAccounts` receives an array of accounts (which dates from who knows when and who knows why) that array in 100% cases has only a single element. User is not able to make changes over 2 or more accounts simultaneously. Even if that's possible that also means all those accounts are changed at the same time, so all of them need to have the same `Clock` in that case.

> Basically this assumes that Accounts (all of them) are a single entity and synchronize together, which is a bit risky.

This part doesn't have anything with the synchronization or backing up accounts. Also handling accounts received from synchronization or backed up messages do not use this function. It's only used when client interacts with an account(s) and sets the clock which is set to db and used to notify other paired devices when the change was made over that account and used at the moment of backing up accounts.

> How do we update accounts, do you have a link to desktop?

The desktop app updates each individual account separately, also the desktop app doesn't set any Clock value either it's a new account creation or an account update action.

cammellos: > The only place where this (m *Messenger) SaveAccounts function is called from is from the (api *API) SaveAccounts that further means it is used by client only.

Ok, so it's not saving all the existing accounts at the same time, only those that have changed, makes sense.

> This part doesn't have anything with the synchronization or backing up accounts

Maybe we misunderstand each other, but clock is exclusively used for synchronization/backup when it comes to wallet, and we sync just a few lines below, so looks like is very much related to synchronization this code.
But my concern has been addressed, I assumed `SaveAccounts` will always to take the full account list (changed/unchanged), but if it's only used for changed accounts, the code above is correct.

saledjenic:
Since there are changes in last 3 commits which may affect mobile app would be good to check it before merging this.

So removing `isMobileApp` the code is now as it should be once mobile app implements features the desktop app has, also `Clock` is explicitelly set on the status-go side for added/updated accounts and `latest_derived_path` is removed from settings as there is no point to have it there if we removed code which is using it.

protocol/messenger.go
saledjenic: Yes, we're always storing only changed field/s for a single account when the change is made. Finally on the same page 🙂

qfrank: >  we're always storing only changed field/s for a single account 

emm.. it would be better to change `func (m *Messenger) SaveAccounts(accs []*accounts.Account) error` to `func (m *Messenger) SaveAccount(acc *accounts.Account) error` although we checked we're always storing only single account. We can't gurantuee future developer won't make mistake 🙂 @saledjenic

qfrank:
I'm glad that we finally made a agreement on `clock` 🫶 
I left last one minor [comment](https://github.com/status-im/status-go/pull/3416#discussion_r1173763084) @saledjenic

protocol/messenger.go
saledjenic: @qfrank I updated the `SaveAccounts` endpoint, please chek [this commit](https://github.com/status-im/status-go/pull/3416/commits/07575eebd1999d5defb82fb29a26b730005009af).

qfrank (approved):
LGTM 👍

PR #3417, opened 2023-04-20
Added SetCommunityEncryption

### LINKED ISSUE #9942: Community encryption should automatically switch on/off depending on community permissions
**This is a feature for the 0.12 release**

-- Change request --

1. remove the "Encrypt community" toggle from the "Create community" and "Edit community" screens

2. If the community is an open community (e.g. there are no token permissions to become a member at the community level) then community encryption should automatically be Off

3. If a community is a closed community (e.g. there are token permissions to become a member at the community level) then community encryption should automatically be On

4. If an open community becomes a closed community (e.g. if a token permission to become a member at the community level is added) than community encryption should automatically be switched On

5. If a closed community becomes an open community (e.g. if all token permissions to become a member at the community level are deleted) then community encryption should automatically be switched Off 

cc'ing @iurimatias @jrainville

Samyoul:
## What's Changed?

I've added a new function to `Messenger` `SetCommunityEncryption(communityID string, useEncryption bool)`.

`SetCommunityEncryption` takes a communityID string and an encryption state, then finds the community and encrypts / decrypts the community. Community is republished along with any keys if needed.

I've added a new function to `communities.Manager` `UpdateCommunity(c *Community)`.

`UpdateCommunity` takes a Community persists it and republishes it. The clock is incremented meaning even a no change update will be republished by the admin, and parsed by the member.

### ⚠️ Note:

- This function cannot decrypt previously encrypted messages, and it cannot encrypt previous unencrypted messages.
- This functionality has some known race conditions:
  - community description is processed by members before the receiving the key exchange messages
  - members maybe sending encrypted messages after the community description is updated and a new member joins

## Why Change?

Sometimes you want to make a community public, sometimes you want to me a community private

closes https://github.com/status-im/status-desktop/issues/9942


jrainville (commented):
LGTM. Waiting on removal from draft to give :+1:

Samyoul:
@cammellos the functionality for this feature seems to work. But I do not have tests that check if messages sent to the updated community are encrypted or not. This is not very satisfactory as a subset of breaking changes to this functionality would include scenarios where the community has an encryption flag set to a value, but the behaviour does not reflect the community encryption state.

Buuut, I don't know how to test this. Do you (or anyone else) know how to test that the correct keys were received and are being used on the `Messenger` level?

cammellos:
@Samyoul best way I can think of, is to check the database/expose seq number to check if it was received/decrypted etc, unfortunately with encryption it's completely invisible to the user, so we need to dig into db etc

igor-sirotin:
@Samyoul is this something we still need?

igor-sirotin:
Closing as currently out of scope

PR #3419, opened 2023-04-24
feat(Wallet): add activity filter basic API to extract metadata

stefandunca:
### Updates status-desktop [#10366](https://github.com/status-im/status-desktop/issues/10366)

status-desktop PR [#10407](https://github.com/status-im/status-desktop/pull/10407)

Add the possibility of retrieving the metadata of wallet activity based
on the given filter criteria.

The current implementation relies that after fetching the metadata, the user will follow up with more requests for details. However, after experimenting, I'm considering extracting all required information for the summary viewing while filtering. This way there will be no need for another batch requests for transfers, multi-transactions and pending transactions to show the summary. Only when the user wants to see the details for one will specifically request it.

For this prototype, the filter criteria are limited to the following:
- time
- type
- addresses

Limitations
- No complete send/receive status. It relies on the passed addresses filter. It will be extended with 
- Didn't index the filter columns so the queries are slow, works for testing though

Major changes:
- Add the filter definition to be used in propagating the filter
  information
- Add GetActivityEntries API to return the list of activity entries
  for the given addresses/chainIDs by a view in the complete list
- GetTransfersForIdentities to batch retrieve further details of the
  transfers
- GetPendingTransactionsForIdentities to batch retrieve further details
  of the pending transactions
- Added a new package testutils for tests.
- Added tests

Closes status-desktop [#10633](https://github.com/status-im/status-desktop/issues/10633)

### DIFF
--- a/services/wallet/activity/activity.go
+++ b/services/wallet/activity/activity.go
@@ -1,0 +1,299 @@
+package activity
+
+import (
+	"database/sql"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/status-im/status-go/services/wallet/transfer"
+)
+
+type PayloadType = int
+
+const (
+	MultiTransactionPT PayloadType = iota + 1
+	SimpleTransactionPT
+	PendingTransactionPT
+)
+
+type Entry struct {
+	// TODO: rename in payloadType
+	transactionType PayloadType
+	transaction     *transfer.TransactionIdentity
+	id              transfer.MultiTransactionIDType
+	timestamp       int64
+	activityType    Type
+}
+
+type jsonSerializationTemplate struct {
+	TransactionType PayloadType                     `json:"transactionType"`
+	Transaction     *transfer.TransactionIdentity   `json:"transaction"`
+	ID              transfer.MultiTransactionIDType `json:"id"`
+	Timestamp       int64                           `json:"timestamp"`
+	ActivityType    Type                            `json:"activityType"`
+}
+
+func (e *Entry) MarshalJSON() ([]byte, error) {
+	return json.Marshal(jsonSerializationTemplate{
+		TransactionType: e.transactionType,
+		Transaction:     e.transaction,
+		ID:              e.id,
+		Timestamp:       e.timestamp,
+		ActivityType:    e.activityType,
+	})
+}
+
+func (e *Entry) UnmarshalJSON(data []byte) error {
+	aux := jsonSerializationTemplate{}
+
+	if err := json.Unmarshal(data, &aux); err != nil {
+		return err
+	}
+
+	e.transactionType = aux.TransactionType
+	e.transaction = aux.Transaction
+	e.id = aux.ID
+	e.timestamp = aux.Timestamp
+	e.activityType = aux.ActivityType
+	return nil
+}
+
+func NewActivityEntryWithTransaction(transactionType PayloadType, transaction *transfer.TransactionIdentity, timestamp int64, activityType Type) Entry {
+	if transactionType != SimpleTransactionPT && transactionType != PendingTransactionPT {
+		panic("invalid transaction type")
+	}
+
+	return Entry{
+		transactionType: transactionType,
+		transaction:     transaction,
+		id:              0,
+		timestamp:       timestamp,
+		activityType:    activityType,
+	}
+}
+
+func NewActivityEntryWithMultiTransaction(id transfer.MultiTransactionIDType, timestamp int64, activityType Type) Entry {
+	return Entry{
+		transactionType: MultiTransactionPT,
+		id:              id,
+		timestamp:       timestamp,
+		activityType:    activityType,
+	}
+}
+
+func (e *Entry) TransactionType() PayloadType {
+	return e.transactionType
+}
+
+func multiTransactionTypeToActivityType(mtType transfer.MultiTransactionType) Type {
+	if mtType == transfer.MultiTransactionSend {
+		return SendAT
+	} else if mtType == transfer.MultiTransactionSwap {
+		return SwapAT
+	} else if mtType == transfer.MultiTransactionBridge {
+		return BridgeAT
+	}
+	panic("unknown multi transaction type")
+}
+
+func typesContain(slice []Type, item Type) bool {
+	for _, a := range slice {
+		if a == item {
+			return true
+		}
+	}
+	return false
+}
+
+func joinMTTypes(types []transfer.MultiTransactionType) string {
+	var sb strings.Builder
+	for i, val := range types {
+		if i > 0 {
+			sb.WriteString(",")
+		}
+		sb.WriteString(strconv.Itoa(int(val)))
+	}
+
+	return sb.String()
+}
+
+func joinAddresses(addresses []common.Address) string {
+	var sb strings.Builder
+	for i, address := range addresses {
+		if i == 0 {
+			sb.WriteString("('")
+		} else {
+			sb.WriteString("'),('")
+		}
+		sb.WriteString(strings.ToUpper(hex.EncodeToString(address[:])))
+	}
+	sb.WriteString("')")
+
+	return sb.String()
+}
+
+func activityTypesToMultiTransactionTypes(trTypes []Type) []transfer.MultiTransactionType {
+	mtTypes := make([]transfer.MultiTransactionType, 0, len(trTypes))
+	for _, t := range trTypes {
+		var mtType transfer.MultiTransactionType
+		if t == SendAT {
+			mtType = transfer.MultiTransactionSend
+		} else if t == SwapAT {
+			mtType = transfer.MultiTransactionSwap
+		} else if t == BridgeAT {
+			mtType = transfer.MultiTransactionBridge
+		} else {
+			continue
+		}
+		mtTypes = append(mtTypes, mtType)
+	}
+	return mtTypes
+}
+
+// TODO: extend with SEND/RECEIVE for transfers and pending_transactions
+// TODO: clarify if we include sender and receiver in pending_transactions as we do for transfers
+// TODO optimization: consider implementing nullable []byte instead of using strings for addresses
+// Query includes duplicates, will return multiple rows for the same transaction
+const queryFormatString = `
+	WITH filter_conditions AS (
+		SELECT
+			? AS startFilterDisabled,
+			? AS startTimestamp,
+			? AS endFilterDisabled,
+			? AS endTimestamp,
+
+			? AS filterActivityTypeAll,
+			? AS filterActivityTypeSend,
+			? AS filterActivityTypeReceive,
+
+			? AS filterAllAddresses
+		),
+		filter_addresses(address) AS (
+			VALUES %s
+		)
+	SELECT
+		transfers.hash AS transfer_hash,
+		NULL AS pending_hash,
+		transfers.network_id AS network_id,
+		0 AS multi_tx_id,
+		transfers.timestamp AS timestamp,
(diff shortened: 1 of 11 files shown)


ghost:
### Pull Request Checklist

vendor/github.com/ethereum/go-ethereum/common/types.go
```diff
@@ -414,7 +414,7 @@ func (ma *MixedcaseAddress) String() string {
 	if ma.ValidChecksum() {
 		return fmt.Sprintf("%s [chksum ok]", ma.original)
 	}
-	return fmt.Sprintf("%s [chksum INVALID]", ma.original)
+return fmt.Sprintf("%s [chksum INVALID]", ma.original)
```
alaibe: need revert

services/wallet/activity/activity.go
```diff
@@ -0,0 +1,313 @@
+package activity
```
alaibe, line 1: If I understand correctly, this package is the query side and the transfer package is the collector?

services/wallet/api.go
```diff
@@ -230,6 +237,20 @@ func (api *API) GetPendingTransactionsByChainIDs(ctx context.Context, chainIDs [
 	return rst, err
 }
 
+func (api *API) GetPendingTransactionsForIdentities(ctx context.Context, identities []transfer.TransactionIdentity) (result []*transfer.PendingTransaction, err error) {
```
alaibe, line 240: Why not mixing with the rest? Out of curiosity can we just have GetActivityEntries?

alaibe (approved):

services/wallet/activity/activity.go
stefandunca: Yes, this is only for processing activity history for all transactions. Basically will reflect most of the UX requirements and take the heavy lifting from Nim and QML. What do you think? Any concerns?

services/wallet/api.go
stefandunca: I wanted to keep the separation of concerns at the current level. The `GetActivityEntries` should only provide the metadata based on filter criteria so that there is no need for merging the results on the Nim side.

Now I'm thinking even further with this and looking into `GetActivityEntries` serving all the needs for the activity summary. These details APIs should only be called when a user views the transaction details (with a sub-transactions list).

Does it make sense? Do you see some inconveniences or risks?

From a performance point of view seems optimal; most of the important details we have them because we need to take them in account for the filter when we query. Extending with more columns and joining other tables sounds like a small overhead.

services/wallet/activity/activity.go
```diff
@@ -0,0 +1,299 @@
...
+	e.transaction = aux.Transaction
+	e.id = aux.ID
+	e.timestamp = aux.Timestamp
+	e.activityType = aux.ActivityType
+	return nil
+}
+
+func NewActivityEntryWithTransaction(transactionType PayloadType, transaction *transfer.TransactionIdentity, timestamp int64, activityType Type) Entry {
+	if transactionType != SimpleTransactionPT && transactionType != PendingTransactionPT {
+		panic("invalid transaction type")
```
IvanBelyakoff, line 68: I think we should use assert instead of panic to catch such issues during dev cycle, but have only errors in production do avoid the whole app abortion if we can't process some transaction

services/wallet/activity/activity.go
```diff
@@ -0,0 +1,299 @@
...
+
+func multiTransactionTypeToActivityType(mtType transfer.MultiTransactionType) Type {
+	if mtType == transfer.MultiTransactionSend {
+		return SendAT
+	} else if mtType == transfer.MultiTransactionSwap {
+		return SwapAT
+	} else if mtType == transfer.MultiTransactionBridge {
+		return BridgeAT
+	}
+	panic("unknown multi transaction type")
```
IvanBelyakoff, line 101: return some InvalidType instead?

services/wallet/activity/activity.go
```diff
@@ -0,0 +1,299 @@
...
+			entry = NewActivityEntryWithTransaction(SimpleTransactionPT, &transfer.TransactionIdentity{ChainID: uint64(chainID.Int64), Hash: common.BytesToHash(transferHash), Address: thisAddress}, timestamp, activityType)
+		} else if pendingHash != nil && chainID.Valid {
+			var activityType Type = SendAT
+			entry = NewActivityEntryWithTransaction(PendingTransactionPT, &transfer.TransactionIdentity{ChainID: uint64(chainID.Int64), Hash: common.BytesToHash(pendingHash)}, timestamp, activityType)
+		} else if multiTxID.Valid {
+			activityType := multiTransactionTypeToActivityType(transfer.MultiTransactionType(dbActivityType.Byte))
+			entry = NewActivityEntryWithMultiTransaction(transfer.MultiTransactionIDType(multiTxID.Int64),
+				timestamp, activityType)
+		} else {
+			return nil, errors.New("invalid row data")
```
IvanBelyakoff, line 289: Do we have somewhere a wrapper around such errors to get some context of it if it happens, like method signature, package, etc? So that if the error occurs we don't rely on the error message only to be unique across the lib.
I normally add some params int the message to see what query/request exactly failed.

services/wallet/activity/activity.go
```diff
@@ -0,0 +1,299 @@
...
+		NULL AS owner_address
+	FROM multi_transactions, filter_conditions
+	WHERE ((startFilterDisabled OR timestamp >= startTimestamp) AND (endFilterDisabled OR timestamp <= endTimestamp))
+		AND (filterActivityTypeAll OR (multi_transactions.type IN (%s)))
+		AND (filterAllAddresses OR (HEX(multi_transactions.from_address) IN filter_addresses) OR (HEX(multi_transactions.to_address) IN filter_addresses))
+
+	ORDER BY timestamp DESC
+	LIMIT ? OFFSET ?`
+
+func GetActivityEntries(db *sql.DB, addresses []common.Address, chainIDs []uint64, filter Filter, offset int, limit int) ([]Entry, error) {
```
IvanBelyakoff, line 227: do we have some const or enum for NoLimit for limit? how do we pass no limit for `limit` parameter?

services/wallet/activity/activity_test.go
```diff
@@ -0,0 +1,406 @@
...
+}
+
+func TestGetActivityEntriesAll(t *testing.T) {
+	db, close := setupTestActivityDB(t)
+	defer close()
+
+	td := fillTestData(t, db)
+
+	var filter Filter
+	entries, err := GetActivityEntries(db, []common.Address{}, []uint64{}, filter, 0, 10)
```
IvanBelyakoff, line 79: Do we need a limit here? The test name suggests fetching all entries

services/wallet/activity/filter.go
```diff
@@ -0,0 +1,46 @@
+package activity
+
+import "github.com/ethereum/go-ethereum/common"
+
+type Period struct {
+	// 0 means no limit
```
IvanBelyakoff, line 6: having "no limit" declared as const or enum looks like a better way to go with it. I think there are cases when we receive rpc requests and still have to maintain (not sure if is is doable to avoid this?) consistency between a sender and receiver of such consts, but at least for using on status-go side it might be better

services/wallet/api.go
IvanBelyakoff: So this one will be remove in the upcoming PRs?

services/wallet/transfer/database.go
```diff
@@ -273,6 +274,22 @@ func (db *Database) GetTransfers(chainID uint64, start, end *big.Int) (rst []Tra
 	return query.Scan(rows)
 }
 
+func (db *Database) GetTransfersForIdentities(ctx context.Context, identities []TransactionIdentity) (rst []Transfer, err error) {
```
IvanBelyakoff, line 277: Could you please clarify, in which case it will be used? From what I see this function will be called when we know transfers hashes and want to get full information about these transfers.

IvanBelyakoff (commented):
Great job, @stefandunca! I know this has been merged, still wanted to leave some feedback and ask some questions.

services/wallet/activity/activity.go
stefandunca: I'm also used to testing debug with `assert`s in the development cycle and release with disabled asserts in production. However, I didn't see this used with status-go, so I understand that for developer errors, the best way is to `panic` in Go.

Thanks for bringing up this topic, let's address it in our team discussions.

services/wallet/activity/activity.go
stefandunca: I'm not a fan of adding logical branches to address developer errors. This will increase the branching further with each use, and I don't see a good reason in this case. For now, I'm thinking of adding a comment, in the follow-up [PR](https://github.com/status-im/status-go/pull/3478), to the enum to remind the developer to update the translation functions.

services/wallet/activity/activity.go
stefandunca: Great idea to add one. We don't use for now so I didn't add but as soon as it is needed can be added like this cases in the follow up PR: https://github.com/status-im/status-go/blob/bde1482d140e1c4c02cfc9087ea8c629f2702ebc/services/wallet/activity/filter.go#L24

In general, I never expect to use this function without a limit, so that case is not addressed.

Thanks for the heads up, I will add a comment to the function to remind the developer about this.

services/wallet/activity/activity.go
stefandunca: > Do we have somewhere a wrapper around such errors to get some context of it if it happens, like method signature, package, etc

Not really, because it is not handled anywhere.

> I normally add some params int the message to see what query/request exactly failed.

For the case it is handled in the future, the idiomatic way for golang is to define an error value that is exported from the package and return that one. This way, it can be checked by comparison and the case specifically addressed. That will be added as needed. Until then we rely on logs from the app to log this error for debugging purpuses.

Also if required to be handled outside the RPC API (Nim) then a method as you suggested to map the golang errors to enums can be implemented.

services/wallet/activity/filter.go
stefandunca: We don't yet have a way of having common definitions between status-go and clients. I also added this comment in regards to this in the status-desktop PR: https://github.com/status-im/status-desktop/blob/136b2019f4ccef8804b81c2692bf4ed6e324f34b/src/backend/activity.nim#L11

Until then, we relay that the developer checks the documentation and these comments.

For this work package  rely that this knowledge is preserved by adding the factory function 
 function that uses `Option` to convert to 0, as you can see here: https://github.com/status-im/status-desktop/blob/136b2019f4ccef8804b81c2692bf4ed6e324f34b/src/backend/activity.nim#L39
 
Thanks for the heads up, I will make it more explicit in the follow up PR.

services/wallet/api.go
stefandunca: > So this one will be remove in the upcoming PRs?

Yes, or converted to extract only one entry. Or used like this for the case of one entry and multiple after analysing the requirements for pending transactions details view.

services/wallet/transfer/database.go
stefandunca: Filtering works by returning identities for items that match the filter, in this case, TrasactionIdentity; in the case of multi-transaction multi-transaction ID, in the future collectibles will probably be CollectibleIndentities. Hence we need this kind of functionality to retrieve the details when the user selects to see a detailed view.

stefandunca (commented):
Thanks for the great feedback. The discussed actionable items will be implemented in the follow-up PR: https://github.com/status-im/status-go/pull/3478 and the weekly developer discussions.

services/wallet/api.go
stefandunca: > So this one will be remove in the upcoming PRs?

Yes, or converted to extract only one entry. Or used like this for the case of one entry and multiple after analysing the requirements for pending transactions details view.

PR #3420, opened 2023-04-24
refactor: improve mailserver batch handling

richard-ramos:
Previously, if the function `processMailserverBatch` failed, the error handler on https://github.com/status-im/status-go/blob/ddfe16f56fa9791e5a98ab211aacc91660e487e4/protocol/messenger_mailserver.go#L588-L593 just returned and the function itself did not notice that an error had happened, because this operation was done on a separate goroutine.

With this PR, an error channel is introduced in https://github.com/status-im/status-go/blob/f81a9e6021c0f30908aa7bfc98b30a31f930ae34/protocol/messenger_mailserver.go#L613 which is used to return the error in https://github.com/status-im/status-go/blob/f81a9e6021c0f30908aa7bfc98b30a31f930ae34/protocol/messenger_mailserver.go#L695-L713 so a failing query might be retried automatically later by the mailserver cycle.

In addition to this change, I noticed that sometimes not all the batches were processed entirely, due to a race condition happening while using the work channel. I ended up introducing another work channel that will receive all the queries that use a cursor https://github.com/status-im/status-go/blob/f81a9e6021c0f30908aa7bfc98b30a31f930ae34/protocol/messenger_mailserver.go#L620-L623 and have these later being sent to the work channel. 

Finally proper cleanup of all goroutines is added in here https://github.com/status-im/status-go/blob/f81a9e6021c0f30908aa7bfc98b30a31f930ae34/protocol/messenger_mailserver.go#L716-L739 so they're killed once all work is processed or an error happens


jrainville (approved):
LGTM. It's worth testing a bunch to see if everything behave the same/better than before

Samyoul (commented):
@richard-ramos would you mind adding a description to this PR, just so other reviewers have a clear view on what this functionality should do?

Also would it be possible to add tests to validate the work you've done. Thank you : 😊

richard-ramos:
@Samyoul I updated the description with the purpose of the PR and main highlights.

However I'm not sure how to create a test to validate this work, as it deals with the mailserver cycle and would also require starting up a storenode :thinking: Perhaps it should be left as an exercise for the reader? 


image

alaibe (approved):

richard-ramos:
@Samyoul based on our conversation of today, I simplified the code and also added a semaphore that would let up to 3 workers to perform a query concurrently

Samyoul (approved):
Thank you very much. Looks great, and I'm looking forward to seeing the ongoing work on this functionality.

Just 2 minor questions. How do we know this functionality works as intended? How could a future developer know if they've broken this functionality before it went to production?

cammellos:
@Samyoul @richard-ramos we should have e2e running ideally against a live store node, there some work already done in some branch, we can discuss further if you'd like.
This way we could run and validate these functionalities, it would be expensive and it requires network access, but at least it provides some confidence that nwaku/go works ok.

richard-ramos:
@cammellos , we could launch a storenode on demand via go-waku (it's currently 1:1 parity with nwaku), and that way do not have to use the fleet nodes, But the part i'm not sure how to do is get the mailserver cycle to work in tests. If possible, point me to a PR/branch you mentioned so I could copy how this has been done.

richard-ramos:
@cammellos, I've rebased this PR to include https://github.com/status-im/status-go/pull/3437 which tests `processMailserverBatch` function

PR #3421, opened 2023-04-24
fix(mentions): deleting or editing a mention should remove the mention

jrainville:
Needed for https://github.com/status-im/status-desktop/issues/10335

If you delete or edit out a mention (or reply for a deletion), it decrements the chat's unread mention count.


ilmotta (approved):
LGTM

Just double-checking @jrainville, I've noticed `updateUnviewedCounts` is called by `HandleDeleteMessage` and `HandleEditMessage`, which you did fix in this PR. Have you checked if the functions `HandleChatMessage` and `handleCommandMessage` don't need to be fixed as well? Because they too call `updateUnviewedCounts`.

cammellos (commented):
@jrainville would you mind adding a test for the two cases please?

Thanks!

jrainville:
>  I've noticed updateUnviewedCounts is called by HandleDeleteMessage and HandleEditMessage, which you did fix in this PR. Have you checked if the functions HandleChatMessage and handleCommandMessage don't need to be fixed as well? Because they too call updateUnviewedCounts.

Good question. In those two other cases, there is nothing to be fixed, because it's the start of the mentions, if it makes sense.

The Edit and Deletes needed to be fixed because those modify existing messages, so the mention could be added or removed and we didn't handle it completely.

jrainville:
> @jrainville would you mind adding a test for the two cases please?
> 
> Thanks!

Done. I added one commit per test. We can squash them if needed at the end

protocol/messenger_delete_message_test.go
```diff
@@ -391,3 +391,62 @@ func (s *MessengerDeleteMessageSuite) TestDeleteImageMessageFirstThenMessage() {
...
+	inputMessage.Text = "text with a mention @" + common.PubkeyToHex(&s.privateKey.PublicKey)
+	sendResponse, err := theirMessenger.SendChatMessage(context.Background(), inputMessage)
+	s.NoError(err)
+	s.Require().Len(sendResponse.Messages(), 1)
+
+	messageID := sendResponse.Messages()[0].ID
+
+	response, err := WaitOnMessengerResponse(
+		s.m,
+		func(r *MessengerResponse) bool { return len(r.messages) > 0 },
```
cammellos: If you expect a single message, `len(r.messages) == 1`, a bit clearer

cammellos (approved):
Nice fix!

PR #3425, opened 2023-04-25
[Fix] Admin should remove the user from community upon user's request to leave

smohamedjavid:
Needed for https://github.com/status-im/status-mobile/issues/15705

### Problem (described in the bug)

An existing member of a community leaves the community and requests again. Then, cancels the request.  
Makes a request again. Now the requester joins the community without the approval of the admin.

### Actual Problem

When a user leaves the community. The user is not removed from the member's list (as the admin has the right to do it) but marked as **not** joined the community. Until the `RequestToLeave` message is received by the admin, the user is not removed from the community.

If the admin node is offline, the user would be under the false impression that they have left the community.

### Solution

Introducing the `RequestToLeaveCommunity` method (to be called by the client), where the user will send just the request to leave without modifying any community data/description.

In the existing handling of the request to leave, the admin will automatically remove the user from the community when the admin is online. This will remove the user in a **clean state**. 

We need to inform the user that they will be removed from the community when the admin node is online.

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -1233,7 +1233,6 @@ func (m *Manager) CancelRequestToJoin(request *requests.CancelRequestToJoinCommu
 		return nil, nil, err
 	}
 
-	community.RemoveOurselvesFromOrg(pk)
 	err = m.persistence.SaveCommunity(community)
 	if err != nil {
 		return nil, nil, err
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -1233,6 +1233,7 @@ func (m *Manager) CancelRequestToJoin(request *requests.CancelRequestToJoinCommu
 		return nil, nil, err
 	}
 
+	community.RemoveOurselvesFromOrg(pk)
 	err = m.persistence.SaveCommunity(community)
 	if err != nil {
 		return nil, nil, err
--- a/protocol/communities_messenger_test.go
+++ b/protocol/communities_messenger_test.go
@@ -2237,8 +2237,6 @@ func (s *MessengerCommunitiesSuite) TestRequestAccessAgain() {
 	community = response.Communities()[0]
 	s.Require().False(community.HasMember(&s.alice.identity.PublicKey))
 
-	s.logger.Info("->ALICE------")
-
 	// Alice should then be removed
 	err = tt.RetryWithBackOff(func() error {
 		response, err = s.alice.RetrieveAll()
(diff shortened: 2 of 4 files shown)


ghost:
### Pull Request Checklist

cammellos:
@smohamedjavid not sure exactly if this works, are we not just removing ourselves locally? 
if so it will likely be overridden next time we receive an updated community description, resulting in the same behavior (I guess we unsubscribe, though people might just share a community description).

Maybe @osmaczko knows, when someone leaves the community, does the admin removes them as well?

borismelnik (approved):
LGTM 👍 
Is it should be requested by the user?

smohamedjavid:
> @smohamedjavid not sure exactly if this works, are we not just removing ourselves locally? if so it will likely be overridden next time we receive an updated community description, resulting in the same behavior (I guess we unsubscribe, though people might just share a community description).

@cammellos - You are right. We should not remove ourselves.

> Maybe @osmaczko knows, when someone leaves the community, does the admin removes them as well?

Yes, the admin removes the user automatically when the `RequestToLeave` message is received on the admin node.

<hr />

The bug is when a user leaves the community. The user is **not** removed from the member's list (as the admin has the right to do it) but marked as not joined the community.

So, trying a different solution by introducing the `RequestToLeaveCommunity` method (to be called by the client), where the user will send just the request to leave without modifying any community data/description. 

In the existing handling of the request to leave, the admin will remove the user from the community automatically when the admin is online. This will remove the user in a clean state.  The user needs to be informed that they will be removed from the community when the admin node is online.

smohamedjavid:
> LGTM 👍 Is it should be requested by the user?

Yes @borismelnik

ilmotta (approved):
Well thought out and tested :) LGTM

osmaczko:
Sorry to be late in the discussion.

> Introducing the `RequestToLeaveCommunity` method (to be called by the client), where the user will send just the request to leave without modifying any community data/description.

Does this mean that if the admin never becomes online again, we will still receive messages, notifications, etc. from this community?

smohamedjavid:
> Does this mean that if the admin never becomes online again, we will still receive messages, notifications, etc. from this community?

I hate to say this. Yes, The user will still receive messages, notifications, etc. from the community until the admin is online.

osmaczko:
> > Does this mean that if the admin never becomes online again, we will still receive messages, notifications, etc. from this community?
> 
> I hate to say this. Yes, The user will still receive messages, notifications, etc. from the community until the admin is online.

Isn't it a no-go then? :sweat:

smohamedjavid:
Closing this PR as we need to discuss with the design team a proper UX for leaving a community if the admin is offline.

osmaczko:
> Closing this PR as we need to discuss with the design team a proper UX for leaving a community if the admin is offline.

:+1: In my opinion, it is essential to provide users with the option to depart from abandoned communities.

PR #3426, opened 2023-04-25
feat: extend `Chat` with `FirstUnviewedMessage`

osmaczko:
iterates: status-im/status-desktop#10302

Extend `Chat` structure with `FirstUnviewedMessage` to reduce the overhead associated with repeated database queries when switching channels in status-desktop.


ghost:
### Pull Request Checklist

protocol/chat.go
```diff
@@ -370,6 +371,17 @@ func (c *Chat) UpdateFromMessage(message *common.Message, timesource common.Time
...
 	}
+	// Update FirstUnviewedMessage
+	if !message.Seen && !message.Deleted && !message.DeletedForMe {
+		if c.FirstUnviewedMessage != nil {
+			if message.Clock < c.FirstUnviewedMessage.Clock {
+				c.FirstUnviewedMessage = message
+			}
+		} else {
+			c.FirstUnviewedMessage = message
+		}
```
jrainville, line 382:
```suggestion
		if c.FirstUnviewedMessage == nil || message.Clock < c.FirstUnviewedMessage.Clock {
			c.FirstUnviewedMessage = message
		} 
```

protocol/chat_persistence_unviewed_counts.go
```diff
@@ -0,0 +1,136 @@
...
+				WHERE local_chat_id = ?
+				AND NOT(seen) AND %s AND (mentioned OR replied)),
+			first_unviewed_message_id =
+				(SELECT id
+				FROM user_messages m1
+				WHERE local_chat_id = ?
+				AND NOT(seen) AND %s
+				ORDER BY clock_value ASC
+				LIMIT 1),
+			highlight = 0
```
jrainville, line 85: The top of the query is the same as above I think, so maybe you can extract it to a constant or function

protocol/migrations/sqlite/1682070624_add_first_unviewed_message_id.up.sql
```diff
@@ -0,0 +1,17 @@
...
+        WHERE
+            local_chat_id = chats.id
+            AND NOT(seen)
+            AND NOT(hide)
+            AND NOT(deleted)
+            AND NOT(deleted_for_me)
+        ORDER BY
+            clock_value ASC
+        LIMIT 1
+    );
```
jrainville, line 17: I assume this works also for channels that have 0 messages, but I'm asking just in case to know if you tested.

jrainville (commented):
Nice job. This seems way cleaner than what we did before.

Just small comments

cammellos:
@osmaczko just to get a bit more context, could you elaborate on how exactly you use the first unseen message?

Meaning, how does generally desktop queries for it before and after this PR?

Thanks!

osmaczko:
> @osmaczko just to get a bit more context, could you elaborate on how exactly you use the first unseen message?
> 
> Meaning, how does generally desktop queries for it before and after this PR?
> 
> Thanks!

@cammellos of course.

Simplifying, before this PR, status-desktop was explicitly querying first unseen message each time the channel was made active, using this endpoint: https://github.com/status-im/status-go/blob/213dc463bcc1ae449bc78b6cf7c598c4f68830c5/services/ext/api.go#L626

This approach was highly inefficient for managing channels with a large number of messages, leading to [lengthy channel-switching delays.](https://user-images.githubusercontent.com/33099791/232008391-bfd0a7e2-3023-481a-bd1c-d5bd87462914.mp4)

After this PR, the first unseen message is part of the `Chat` structure and is updated as needed by status-go. This eliminates the need for status-desktop to explicitly fetch and wait for the first unseen message, as the data is readily available.

protocol/chat_persistence_unviewed_counts.go
```diff
@@ -0,0 +1,136 @@
...
+				err = tx.Commit()
+			}
+		}()
+	}
+
+	return fn(tx)
+}
+
+func (db sqlitePersistence) updateAllChatsUnviewedCounts(tx *sql.Tx) error {
+	return db.executeTx(tx, func(tx *sql.Tx) error {
```
ilmotta, line 32: Hey @osmaczko, I think the `UPDATE` statement is atomic by nature, so there's no need to wrap it in a transaction. The same is true for other functions in this file.

So individually, calling these functions in their own transactions should be unnecessary and I think we can remove the `db.executeTx` surrounding them. Of course, in other files consuming these functions, it may be necessary if there are multiple sequential queries. Some functions like `BlockContact` (which uses `updateAllChatsUnviewedCounts`) already run in a transaction, but I haven't looked others.

osmaczko:
@jrainville @ilmotta @cammellos thanks for reviews.

Closing for now in favor of: https://github.com/status-im/status-go/pull/3442

@cammellos convinced me that denormalized fields are difficult to maintain and prone to bugs. Moreover, [it turns out](https://github.com/status-im/status-desktop/issues/10282#issuecomment-1526081182) that the first unseen message retrieval is not the cause of slow channel switching on Desktop.

PS: @cammellos thanks for support in investigation.

PR #3427, opened 2023-04-25
[Fix] Added Community Kicked notification to response

smohamedjavid:
Needed for https://github.com/status-im/status-mobile/pull/15734#issuecomment-1521929578

### What changed

This PR adds the Activity Center `Community Kicked` notification type to the response. So, the user's Activity Center gets updated instantly without user interaction.


ilmotta (approved):
An accompanying unit test would be good @smohamedjavid, but I have no further comments.

protocol/messenger.go
```diff
@@ -4349,11 +4349,17 @@ func (m *Messenger) handleRetrievedMessages(chatWithMessages map[transport.Filte
...
 					}
 
 					err = m.addActivityCenterNotification(response, notification)
 					if err != nil {
-						m.logger.Error("failed to save notification", zap.Error(err))
+						logger.Error("failed to save notification", zap.Error(err))
+						continue
+					}
+
+					if err := messageState.Response.Merge(response); err != nil {
```
MishkaRogachev, line 4359: Ah, i missed here we are using `messageState.Response`. Thanks for fixing!

MishkaRogachev (approved):

smohamedjavid:
> An accompanying unit test would be good @smohamedjavid, but I have no further comments.

Added 👍

protocol/messenger.go
smohamedjavid: No worries 😄

PR #3428, opened 2023-04-25
feat: add module containing centralized exchanges info

dlipicar:
Add util for fetching list of centralized exchanges and their known addresses from [CoinCarp](https://www.coincarp.com/). The list can be updated and amended in the future. The util generates a go module for accessing that list from status-go.

This is part of https://github.com/status-im/status-desktop/issues/10251


ghost:
### Pull Request Checklist

exchanges/cmd/exchanges.txt
```diff
@@ -0,0 +1,89 @@
+package exchanges
```
alaibe, line 1: Is that normal? to have a txt files?

exchanges/exchanges.json
```diff
@@ -0,0 +1,1227 @@
+[
```
alaibe, line 1: How did u get this list?

exchanges/cmd/exchanges.txt
dlipicar: I've seen a couple modules (for example, the one we use for the iso4217 fiat currency list: https://github.com/ladydascalie/currency) follow this approach for converting data downloaded from some external source to golang code. This file is the "template" used for the automatic conversion (https://pkg.go.dev/text/template)

exchanges/exchanges.json
dlipicar: Auto-generated by the util inside `cmd`

alaibe (approved):

PR #3432, opened 2023-04-27
feat(wallet): refactor transfers to allow multiple transfer fetching

IvanBelyakoff:
strategies.

On mobile devices we might not want to load all transfers history to avoid high internet traffic consumption or to save some space. On desktop we will load all transactions in background.

- Moved current fetching transfers implementation to OnDemandFetchStrategy
- Implemented QueuedAtomicGroup to cache outgoing commands above a threshold to avoid exceeding provider's RPS limit
- Refactored some type names, methods for better clarity:
  - `LastKnownBlock` name was misused
  - `Block` renamed to `BlockDAO` for clarity and avoiding confusion with geth `Block` type
- Cleanup - removed unused parameters and types' fields
- Changed tests to use in memory DB for speedup
- Added more logging and comments

Updates #10246

### DIFF
--- a/services/wallet/api.go
+++ b/services/wallet/api.go
@@ -97,11 +97,13 @@ func (api *API) GetTransfersByAddressAndChainID(ctx context.Context, chainID uin
 	return api.s.transferController.GetTransfersByAddress(ctx, chainID, address, hexBigToBN(toBlock), limit.ToInt().Int64(), fetchMore)
 }
 
-func (api *API) GetCachedBalances(ctx context.Context, addresses []common.Address) ([]transfer.LastKnownBlockView, error) {
+// Deprecated: GetCachedBalances is deprecated. Use GetTokensBalances instead
+func (api *API) GetCachedBalances(ctx context.Context, addresses []common.Address) ([]transfer.BlockView, error) {
 	return api.s.transferController.GetCachedBalances(ctx, api.s.rpcClient.UpstreamChainID, addresses)
 }
 
-func (api *API) GetCachedBalancesbyChainID(ctx context.Context, chainID uint64, addresses []common.Address) ([]transfer.LastKnownBlockView, error) {
+// Deprecated: GetCachedBalances is deprecated. Use GetTokensBalancesForChainIDs instead
+func (api *API) GetCachedBalancesbyChainID(ctx context.Context, chainID uint64, addresses []common.Address) ([]transfer.BlockView, error) {
 	return api.s.transferController.GetCachedBalances(ctx, chainID, addresses)
 }
 
--- a/services/wallet/async/async.go
+++ b/services/wallet/async/async.go
@@ -8,6 +8,10 @@ import (
 
 type Command func(context.Context) error
 
+type Commander interface {
+	Command() Command
+}
+
 // FiniteCommand terminates when error is nil.
 type FiniteCommand struct {
 	Interval time.Duration
@@ -93,13 +97,16 @@ func (g *Group) WaitAsync() <-chan struct{} {
 
 func NewAtomicGroup(parent context.Context) *AtomicGroup {
 	ctx, cancel := context.WithCancel(parent)
-	return &AtomicGroup{ctx: ctx, cancel: cancel}
+	ag := &AtomicGroup{ctx: ctx, cancel: cancel}
+	ag.done = ag.onFinish
+	return ag
 }
 
 // AtomicGroup terminates as soon as first goroutine terminates..
 type AtomicGroup struct {
 	ctx    context.Context
 	cancel func()
+	done   func()
 	wg     sync.WaitGroup
 
 	mu    sync.Mutex
@@ -110,7 +117,7 @@ type AtomicGroup struct {
 func (d *AtomicGroup) Add(cmd Command) {
 	d.wg.Add(1)
 	go func() {
-		defer d.wg.Done()
+		defer d.done()
 		err := cmd(d.ctx)
 		d.mu.Lock()
 		defer d.mu.Unlock()
@@ -154,4 +161,61 @@ func (d *AtomicGroup) Error() error {
 
 func (d *AtomicGroup) Stop() {
 	d.cancel()
+}
+
+func (d *AtomicGroup) onFinish() {
+	d.wg.Done()
+}
+
+func NewQueuedAtomicGroup(parent context.Context, limit uint32) *QueuedAtomicGroup {
+	qag := &QueuedAtomicGroup{NewAtomicGroup(parent), limit, 0, []Command{}, sync.Mutex{}}
+	baseDoneFunc := qag.done // save original done function
+	qag.AtomicGroup.done = func() {
+		baseDoneFunc()
+		qag.onFinish()
+	}
+	return qag
+}
+
+type QueuedAtomicGroup struct {
+	*AtomicGroup
+	limit       uint32
+	count       uint32
+	pendingCmds []Command
+	mu          sync.Mutex
+}
+
+func (d *QueuedAtomicGroup) Add(cmd Command) {
+
+	d.mu.Lock()
+	if d.limit > 0 && d.count >= d.limit {
+		d.pendingCmds = append(d.pendingCmds, cmd)
+		d.mu.Unlock()
+		return
+	}
+
+	d.mu.Unlock()
+	d.run(cmd)
+}
+
+func (d *QueuedAtomicGroup) run(cmd Command) {
+	d.mu.Lock()
+	d.count++
+	d.mu.Unlock()
+	d.AtomicGroup.Add(cmd)
+}
+
+func (d *QueuedAtomicGroup) onFinish() {
+	d.mu.Lock()
+	d.count--
+
+	if d.count < d.limit && len(d.pendingCmds) > 0 {
+		cmd := d.pendingCmds[0]
+		d.pendingCmds = d.pendingCmds[1:]
+		d.mu.Unlock()
+		d.run(cmd)
+		return
+	}
+
+	d.mu.Unlock()
 }
--- a/services/wallet/transfer/commands.go
+++ b/services/wallet/transfer/commands.go
@@ -45,7 +45,6 @@ var (
 )
 
 type ethHistoricalCommand struct {
-	db           *Database
 	eth          Downloader
 	address      common.Address
 	chainClient  *chain.ClientWithFallback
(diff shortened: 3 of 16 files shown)


jakubgs:
This PR needs to be rebased, the builds can't run because of old labels:
```
There are no nodes with the label 'linux&&x86_64&&go-1.18'
```

services/wallet/transfer/concurrent.go
```diff
@@ -86,11 +86,15 @@ type Downloader interface {
...
-func checkRanges(parent context.Context, client BalanceReader, cache BalanceCache, downloader Downloader, account common.Address, ranges [][]*big.Int) ([][]*big.Int, []*DBHeader, error) {
+// Returns new block ranges that contain transfers and found block headers that contain transfers.
+func checkRanges(parent context.Context, client BalanceReader, cache BalanceCache, downloader Downloader,
+	account common.Address, ranges [][]*big.Int) ([][]*big.Int, []*DBHeader, error) {
+
 	ctx, cancel := context.WithTimeout(parent, 30*time.Second)
 	defer cancel()
 
-	c := NewConcurrentDownloader(ctx)
+	var noThreadlimit int64 = 0
```
dlipicar: const? better if defined once at the top

services/wallet/transfer/concurrent_test.go
```diff
@@ -16,7 +16,8 @@ import (
 )
 
 func TestConcurrentErrorInterrupts(t *testing.T) {
-	concurrent := NewConcurrentDownloader(context.Background())
+    var noThreadLimit int64 = 0
```
dlipicar: if defined in a common place, use that instead

services/wallet/transfer/concurrent_test.go
```diff
@@ -36,7 +37,8 @@ func TestConcurrentErrorInterrupts(t *testing.T) {
 }
 
 func TestConcurrentCollectsTransfers(t *testing.T) {
-	concurrent := NewConcurrentDownloader(context.Background())
+    var noThreadLimit int64 = 0
```
dlipicar: if defined in a common place, use that instead

services/wallet/transfer/downloader.go
```diff
@@ -368,7 +370,8 @@ func (d *ERC20TransfersDownloader) transfersFromLogs(parent context.Context, log
 }
 
 func (d *ERC20TransfersDownloader) blocksFromLogs(parent context.Context, logs []types.Log, address common.Address) ([]*DBHeader, error) {
-	concurrent := NewConcurrentDownloader(parent)
+	var noThradLimit int64 = 0
```
dlipicar: ditto

services/wallet/transfer/downloader.go
```diff
@@ -344,7 +345,8 @@ func (d *ERC20TransfersDownloader) transferFromLog(parent context.Context, ethlo
 }
 
 func (d *ERC20TransfersDownloader) transfersFromLogs(parent context.Context, logs []types.Log, address common.Address) ([]Transfer, error) {
-	concurrent := NewConcurrentDownloader(parent)
+	var noThradLimit int64 = 0
```
dlipicar: ditto

services/wallet/transfer/concurrent.go
IvanBelyakoff: Great catch, @dlipicar ! Fixed

services/wallet/transfer/reactor.go
```diff
@@ -60,35 +85,216 @@ func (r *Reactor) newControlCommand(chainClient *chain.ClientWithFallback, accou
...
+
+		from, err := findFirstRange(ctx, address, block, chainClient)
+		if err != nil {
+			if nonArchivalNodeError(err) {
+				if s.feed != nil {
+					s.feed.Send(walletevent.Event{
+						Type: EventNonArchivalNodeDetected,
+					})
+				}
+				from = big.NewInt(0).Sub(block, big.NewInt(100))
```
dlipicar: This `100` represents a chunk/page size, right? Not sure if it will be reused in the new strategy, either way a const could be defined for clarity.

services/wallet/async/async.go
```diff
@@ -155,3 +160,57 @@ func (d *AtomicGroup) Error() error {
...
+
+type QueuedAtomicGroup struct {
+	*AtomicGroup
+	limit       int64
+	count       int64
+	pendingCmds []Command
+}
+
+func (d *QueuedAtomicGroup) Add(cmd Command) {
+	counter := atomic.LoadInt64(&d.count)
```
dlipicar: Hmm it seems to me that we're making the counter thread-safe by using atomic ops, but not the list of commands, which might cause trouble (for example, two threads adding commands while `counter >= d.limit` will try to append to the list at the same time).

services/wallet/async/async.go
```diff
@@ -155,3 +160,57 @@ func (d *AtomicGroup) Error() error {
 func (d *AtomicGroup) Stop() {
 	d.cancel()
 }
+
+func NewQueuedAtomicGroup(parent context.Context, limit int64) *QueuedAtomicGroup {
+	return &QueuedAtomicGroup{NewAtomicGroup(parent), limit, 0, []Command{}}
+}
+
+type QueuedAtomicGroup struct {
```
dlipicar, line 180: I see some duplicate code between `AtomicGroup` and `QueuedAtomicGroup`.
I think some code in `AtomicGroup` can be split into functions that can then be called by `QueuedAtomicGroup` to avoid this. For example, it should be possible to let `AtomicGroup` handle the `WaitGroup` entirely instead of accessing it directly from `QueuedAtomicGroup`.

rasom:
@IvanBelyakoff
 
> On mobile devices we might not want to load all transfers history to avoid high internet traffic consumption or to save some space. 

One thing to mention here because it still might be (or not) relevant to desktop: we were limited in number of RPC requests we are allowed to execute. Traffic was also a problem (and not so much space), but as we relied on third party services and archival requests are usually expensive, the number of requests we were allowed to execute has never been infinite.

services/wallet/async/async.go
IvanBelyakoff: Good question. Not sure if adding commands to the list at the same time will cause any troubles. But in general it is better to protect when we have reading and writing, so coupled with your other comment, I think I will rework a little this type for less code duplication and more thread safety

services/wallet/transfer/commands.go
```diff
@@ -461,10 +449,11 @@ func (c *findAndCheckBlockRangeCommand) Command() async.Command {
 
 func (c *findAndCheckBlockRangeCommand) Run(parent context.Context) (err error) {
 	log.Debug("start findAndCHeckBlockRangeCommand")
+
 	newFromByAddress, ethHeadersByAddress, err := c.fastIndex(parent, c.balanceCache, c.fromByAddress, c.toByAddress)
 	if err != nil {
 		c.error = err
-		return nil
+		return err // In case c.noLimit is true, hystrix "max concurrency" may be reached and we will not be able to index ETH transfers
```
alaibe: we can increase the concurrency if needed

services/wallet/transfer/sequential_fetch_strategy.go
```diff
@@ -0,0 +1,90 @@
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if s.group != nil {
+		return errAlreadyRunning
+	}
+	s.group = async.NewGroup(context.Background())
+
+	// Run the command on interval (5 seconds)
+	for _, chainClient := range s.chainClients {
```
alaibe, line 60: does it mean we load all transfer for all chain with interval between chain at 5s?

alaibe (approved):

services/wallet/api.go
```diff
@@ -97,11 +97,13 @@ func (api *API) GetTransfersByAddressAndChainID(ctx context.Context, chainID uin
 	return api.s.transferController.GetTransfersByAddress(ctx, chainID, address, hexBigToBN(toBlock), limit.ToInt().Int64(), fetchMore)
 }
 
-func (api *API) GetCachedBalances(ctx context.Context, addresses []common.Address) ([]transfer.LastKnownBlockView, error) {
+// Deprecated: GetCachedBalances is deprecated. Use GetTokensBalances instead
```
stefandunca, line 100: These APIs should be removed if they are not required by mobile.

services/wallet/transfer/reactor.go
stefandunca: I am wondering where we check that this subtraction won't generate invalid block numbers.

services/wallet/api.go
rasom: We still have it the code, although I'm not 100% it is used. Better to keep here for the moment.

services/wallet/transfer/reactor.go
IvanBelyakoff: @stefandunca good question. From what I see, the error will happen here: https://github.com/status-im/status-go/blob/5d607ae541acc7dd21fec957d048a132af7e3d63/services/wallet/transfer/concurrent.go#L111
which will finally be propagated up in the current file at line 185. It might make sense to do a sanity check for non-negative block numbers here to avoid further calls.

IvanBelyakoff:
> @IvanBelyakoff
> 
> > On mobile devices we might not want to load all transfers history to avoid high internet traffic consumption or to save some space.
> 
> One thing to mention here because it still might be (or not) relevant to desktop: we were limited in number of RPC requests we are allowed to execute. Traffic was also a problem (and not so much space), but as we relied on third party services and archival requests are usually expensive, the number of requests we were allowed to execute has never been infinite.

Thanx @rasom for the heads up. We still care about a number of requests. Not sure if we care that much about the limits at the moment. If we load all the transactions for users immediately, that will definitely impact our limits. We might want to discuss this once again

services/wallet/transfer/sequential_fetch_strategy.go
IvanBelyakoff: I think I have this comment as some leftover when I was gifuring out how things work. I think we will have infinite command here to check for transfers with interval ~15 secs between checks for each chain,  not between chains. One command will be responsible for loading old transfers, another - for loading new transfers. This interval will be used for checking new transfers

services/wallet/transfer/reactor.go
IvanBelyakoff: > This `100` represents a chunk/page size, right? Not sure if it will be reused in the new strategy, either way a const could be defined for clarity.

done

IvanBelyakoff: > This `100` represents a chunk/page size, right? Not sure if it will be reused in the new strategy, either way a const could be defined for clarity.

done

dlipicar (approved):

PR #3437, opened 2023-04-27
test: processMailserverBatch

richard-ramos:
In this PR I created a mock for the transport `SendMessagesRequestForTopics` function, so it simulates doing a request to retrieve storenode results. In here the test does not attempt to validate the actual storenode response but in the correct handling of the batch process, cursors and handling errors in this function.

This PR should be merged before:
- https://github.com/status-im/status-go/pull/3420
- https://github.com/status-im/status-go/pull/3408


Samyoul (approved):
Thank you for this beautiful work

PR #3439, opened 2023-04-27
refactor(chat/api): re-add chats in the getChannelGroups api

jrainville:
Needed for https://github.com/status-im/status-desktop/issues/10076

The refactor I did last time was nice, but caused a couple issues with not having the all the chats in cache at the start.

The impact on performance to revert is not that bad, since we do not hit the DB without getting the pinned messages.

### DIFF
--- a/services/chat/api.go
+++ b/services/chat/api.go
@@ -126,7 +126,7 @@ func unique(communities []*communities.Community) (result []*communities.Communi
 	return result
 }
 
-func (api *API) GetChannelGroups(ctx context.Context) (map[string]ChannelGroup, error) {
+func (api *API) getChannelGroups(ctx context.Context, channelGroupID string) (map[string]ChannelGroup, error) {
 	joinedCommunities, err := api.s.messenger.JoinedCommunities()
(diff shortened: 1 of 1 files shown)


osmaczko (approved):

jrainville:
Re-asking reviews because I added a new commit

services/chat/api.go
```diff
@@ -215,133 +232,46 @@ func (api *API) GetChannelGroups(ctx context.Context) (map[string]ChannelGroup,
...
-			result.Chats[c.ID] = c
-		}
+func (api *API) GetChannelGroupById(ctx context.Context, channelGroupId string) (map[string]ChannelGroup, error) {
+	channelGroups, err := api.getChannelGroups(ctx, channelGroupId)
+	if err != nil {
+		return nil, err
 	}
-
-	return result, nil
+	return channelGroups, err
```
osmaczko: `return api.getChannelGroups(ctx, channelGroupId)`

osmaczko (approved):

PR #3443, opened 2023-04-28
Extend wallet config for mobile / add retrieve balances loop

cammellos:

1) Extend create/import account endpoint to get wallet config, some of
   which has been moved to the backend
2) Set up a loop for retrieving balances every 10 minutes, caching the
   balances
3) Return information about which checks are not passing when trying to
   join a token gated community
4) Add tests to the token gated communities
5) Fixes an issue with addresses not matching when checking for
   permissions
6) Fixes an issue with requesting to join a community, if the password was wrong, it would show the request as pending, but it would fail to send a request

The move to the wallet as a background task is not yet complete, I need to publish a signal, and most likely I will disable it before merging for now, as it's currently not used by desktop/mobile, but the PR was getting to big.


protocol/migrations/sqlite/1683725608_add_token_balances.up.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS token_balances (
+  user_address VARCHAR NOT NULL,
+  token_name VARCHAR NOT NULL,
+  token_symbol VARCHAR NOT NULL,
+  token_address VARCHAR NOT NULL,
+  token_color VARCHAR NOT NULL DEFAULT "",
+  token_decimals INT NOT NULL,
+  token_description VARCHAR NOT NULL DEFAULT "",
+  token_url VARCHAR NOT NULL DEFAULT "",
+  balance INT NOT NULL,
```
alaibe: Is INT enought?

services/wallet/api.go
```diff
@@ -44,6 +44,10 @@ func (api *API) GetWalletToken(ctx context.Context, addresses []common.Address)
 	return api.reader.GetWalletToken(ctx, addresses)
 }
 
+func (api *API) GetCachedWalletTokens(ctx context.Context) (map[common.Address][]Token, error) {
```
alaibe: That can be quite confusing that the market data won't be send, I would try to find a specific name for this. Or add it

protocol/migrations/sqlite/1683725608_add_token_balances.up.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS token_balances (
```
alaibe: Out of curiosity why not using appdatatabase migration?

protocol/messenger_wallet.go
```diff
@@ -0,0 +1,174 @@
...
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/status-im/status-go/eth-node/types"
+	"github.com/status-im/status-go/multiaccounts/accounts"
+	"github.com/status-im/status-go/protocol/common"
+	"github.com/status-im/status-go/protocol/protobuf"
+)
+
+var checkBalancesInterval = time.Minute * 10
+
+func (m *Messenger) retrieveWalletBalances() error {
```
alaibe, line 19: Why cannot this be in wallet service? I don't think it uses anything from the messenger?
We also have a "Start" function in the wallet (which only emit a tick)

protocol/communities/manager.go
```diff
@@ -81,6 +84,21 @@ type Manager struct {
...
+}
+
+type openseaClientBuilder interface {
+	NewOpenseaClient(uint64, string, *event.Feed) (openseaClient, error)
+}
+
+type defaultOpenseaBuilder struct {
+}
+
+func (b *defaultOpenseaBuilder) NewOpenseaClient(chainID uint64, apiKey string, feed *event.Feed) (openseaClient, error) {
```
alaibe, line 98: I don't think the opensea client is suppose to be exposed. There is a collectible package which I think is the one that we could leverage.

alaibe (commented):
Lot of nice changes here, a few questions from me belows

protocol/migrations/sqlite/1683725608_add_token_balances.up.sql
cammellos: SQLite as far as I understand does not make difference between INT/BIGINT etc, they are all stored in the same way.
That would give us 2^63-1 as the maximum value ( 9,223,372,036,854,775,807 ), do you think that's enough?
Otherwise it would be a varchar, what do you think?

services/wallet/api.go
cammellos: I would add it in a separate PR, for now probably best to change the name then (`GetCachedWalletTokensWithoutMarketData` ? a bit verbose but very explicit :) )

protocol/migrations/sqlite/1683725608_add_token_balances.up.sql
cammellos: I think they effectively get merged in the same db, probably we should actually have everything in a single directory, since now it's not very clear (I believe they are separate only for historical reasons, since `protocol` was a separate project)
I can move them there if you prefer (I don't have a strong opinion)

protocol/messenger_wallet.go
cammellos: I think eventually this will have to check some data from messenger, for example, you might want to be notified when you are able to access a given community, or the opposite, similarly we will want to show notifications to the user etc, so I would think that we might want the loop to be in messenger as it will do all the coordination, what do you think?

protocol/communities/manager.go
cammellos: Basically the code in `communities/manager.go` was already using `OpenseaClient` directly, it needed 

`FetchAllAssetsByOwnerAndContractAddress(owner gethcommon.Address, contractAddresses []gethcommon.Address, cursor string, limit int) (*opensea.AssetContainer, error)`

I made it an interface in order to swap out the implementation for testing (code wasn't tested and it's really hard to modify).

I am happy with anything that allows us to mock/stub in order to test (I generally don't like this kind of testing, but in this case the benefits are more than the drawbacks), where do you think it should go/which one we should be using?

protocol/communities/manager.go
```diff
@@ -1501,26 +1584,55 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request
...
+		for _, criteria := range p.Criteria {
+			if !criteria {
+				c.Satisfied = false
+				break
+			}
+		}
+	}
+}
+
+func (m *Manager) checkPermissionToJoin(permissions []*protobuf.CommunityTokenPermission, walletAddresses []gethcommon.Address, shortcircuit bool) (*CheckPermissionToJoinResponse, error) {
```
qfrank, line 1616: could you explain `shortcircuit` or add corresponding param description? looks like it will simplify permission check, make it easier to pass checking, what's the background? thank you

protocol/communities/manager_test.go
```diff
@@ -52,6 +59,10 @@ func (s *ManagerSuite) SetupTest() {
 	s.manager = m
 }
 
+func intToBig(n int64) *hexutil.Big {
+	return (*hexutil.Big)(big.NewInt(n))
+}
+
 func (s *ManagerSuite) getHistoryTaksCount() int {
```
qfrank: getHistoryTaskCount , i know it's not you, would you mind ...

protocol/messenger_wallet.go
```diff
@@ -0,0 +1,174 @@
...
+	if m.walletAPI == nil {
+		m.logger.Warn("wallet api not enabled")
+	}
+	accounts, err := m.settings.GetAccounts()
+	if err != nil {
+		return err
+	}
+
+	if len(accounts) == 0 {
+		m.logger.Info("no accounts to sync")
```
qfrank: maybe "no accounts to sync wallet balance" sounds better?

protocol/messenger_wallet.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	var ethAccounts []ethcommon.Address
+
+	for _, acc := range accounts {
+		m.logger.Info("syncing account", zap.String("account", acc.Address.Hex()))
```
qfrank: "syncing account" seems not very clear, would you mind make it clear?

protocol/messenger_wallet.go
```diff
@@ -0,0 +1,174 @@
...
+	if m.walletAPI == nil {
+		m.logger.Warn("wallet service not enabled")
+		return
+	}
+	go func() {
+		for {
+			select {
+			case <-time.After(checkBalancesInterval):
+
+				m.logger.Info("running wallet balances")
```
qfrank: what about "retrieving wallet balances" ?

protocol/requests/create_account.go
```diff
@@ -12,27 +12,43 @@ var ErrCreateAccountInvalidBackupDisabledDataDir = errors.New("create-account: i
...
 
 type CreateAccount struct {
-	DisplayName        string `json:"displayName"`
-	Password           string `json:"password"`
-	ImagePath          string `json:"imagePath"`
-	CustomizationColor string `json:"customizationColor"`
 	// RootKeystoreDir is the directory where keys are stored
 	RootKeystoreDir string `json:"rootKeystoreDir"`
 	// BackupDisabledDataDir is the directory where backup is disabled
 	BackupDisabledDataDir string `json:"backupDisabledDataDir"`
```
qfrank, line 16: not relate to your PR, but can we rename it to `RootDataDir` ? It seems they are equal and I prefer `RootDataDir`(the base dir(absolute path) application used) since it looks more clear

protocol/requests/create_account.go
```diff
@@ -12,27 +12,43 @@ var ErrCreateAccountInvalidBackupDisabledDataDir = errors.New("create-account: i
 var ErrCreateAccountInvalidLogFilePath = errors.New("create-account: invalid log file path")
 
 type CreateAccount struct {
-	DisplayName        string `json:"displayName"`
-	Password           string `json:"password"`
-	ImagePath          string `json:"imagePath"`
-	CustomizationColor string `json:"customizationColor"`
 	// RootKeystoreDir is the directory where keys are stored
 	RootKeystoreDir string `json:"rootKeystoreDir"`
```
qfrank: not relate to your PR, can we remove this field now? because we calculate it in the backend.

protocol/communities/manager.go
cammellos: I will, but the idea is that you want to stop as soon as you know the answer (similarly how boolean operators do https://en.wikipedia.org/wiki/Short-circuit_evaluation ), since if you don't need detailed answer (i.e which checks pass and which ones don't), it's wasteful to check all the wallet tokens

protocol/requests/create_account.go
cammellos: is that already merged or it's in a different pr?

qfrank: > then rootKeystoreDir is ignored on status-go side and hardcoded "keystore" is used https://github.com/status-im/status-go/blob/9151aa7f042ca0fe3bf9f0c3201a122ee452926d/api/geth_backend.go#L738 

referenced from discord, i thought you saw the message 🙂

> is that already merged or it's in a different pr?

i mean can we remove this field in this PR or new PR?

cammellos: I'd do it in a different PR, this one is already quite a few changes :)

cammellos:
@qfrank addressed feedback

@alaibe I answered some of your questions, would you mind having a look again please?

protocol/migrations/sqlite/1683725608_add_token_balances.up.sql
alaibe: I might be wrong but if for example you own 100 SNT, as SNT use 18 decimals, it means you will store
100000000000000000000 which is already bigger than > 9,223,372,036,854,775,807

If you don't store the decimals then maybe it shouldn't be int?

services/wallet/api.go
alaibe: Find by me, I don't have anything better

protocol/migrations/sqlite/1683725608_add_token_balances.up.sql
alaibe: neither do I, was just asking as i put all my migrations in appdatabase ;)

protocol/communities/manager.go
alaibe: In "theory" we should already have an interface for it located here: https://github.com/status-im/status-go/blob/d55adaf3a486246fa7f266cf0d864cad853db179/services/wallet/thirdparty/types.go

You can see how it is done for the market provider. But it looks like the nft alternative is not "final". If it was used before then be it

protocol/messenger_wallet.go
alaibe: Regarding the "show notifications" we do have event being sent from status-go to mobile from the wallet. But I guess you mean  notification via a "notification center" ?

It feels weird to me to have this here but if we can only send notification via the messenger then it is the blocker?
Shall we make notification a "global" thing?

alaibe (approved):
@cammellos approved for me, no reason to not let that one go even if discussion are still open

qfrank (approved):
LGTM🫶

protocol/migrations/sqlite/1683725608_add_token_balances.up.sql
cammellos: probably best to store it as a varchar then

protocol/messenger_wallet.go
cammellos: Messenger should know about notifications, if it doesn't then we would get into issues when you have to sync them between devices, since the flow would be:

Network -> SyncNotification -> Messenger -> ActivityCenter

So messenger is essentially always involved (it's the orchestrator), at least that's the way I see it, which should avoid circular dependencies (you can break down packages further and remove logic from messenger, that makes sense, but the top level object I think should be messenger)

protocol/communities/manager.go
cammellos: Ok, I will address this in a separate PR If it's ok, I need to add a few tests in any case, thank you!

PR #3444, opened 2023-05-01
fix(wallet): don't delete pending transactions for multi-transaction

stefandunca:
This is done in Nim after the watched pending transaction changes state.

Note: tried to delete all transactions, not only multi-transaction, as soon as the state is downloaded but it didn't work: noticed crashes with minted transactions after restart.

Updates status-desktop [#10474](https://github.com/status-im/status-desktop/issues/10474)

### DIFF
--- a/services/wallet/transfer/commands.go
+++ b/services/wallet/transfer/commands.go
@@ -362,24 +362,13 @@ func (c *transfersCommand) Run(ctx context.Context) (err error) {
 	// Update MultiTransactionID from pending entry
 	for index := range allTransfers {
 		transfer := &allTransfers[index]
-		if transfer.MultiTransactionID == NoMultiTransactionID {
-			entry, err := c.transactionManager.GetPendingEntry(c.chainClient.ChainID, transfer.ID)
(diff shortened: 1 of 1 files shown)


services/wallet/transfer/commands.go
```diff
@@ -368,24 +367,11 @@ func (c *transfersCommand) Run(ctx context.Context) (err error) {
...
-					// TODO: Nim logic was deleting pending previously, should we notify UI about it?
-					err := c.transactionManager.DeletePending(c.chainClient.ChainID, transfer.ID)
-					if err != nil {
-						return err
-					}
-				}
-			}
+
+		entry, err := c.transactionManager.GetPendingEntry(c.chainClient.ChainID, transfer.ID)
+		if err == nil {
```
endulab, line 366: And what if err is not nil ? Don't we need any log or return this err ?

stefandunca: Great catch. Fixed it!

endulab (approved):

PR #3454, opened 2023-05-04
Feat/remove accounts update

saledjenic:
This PR introduces a removing account feature without passing a password. That's possible now since `keystore` in `go-ethereum` is updated.


alaibe (approved):

qfrank (approved):

saledjenic:
@qfrank I had to update `ConvertToKeycardAccount`, added `keycardUID` parameter there. Please check if that affects somehow mobile app and if you could apply necessary changes to align with this?

qfrank:
> @qfrank I had to update `ConvertToKeycardAccount`, added `keycardUID` parameter there. Please check if that affects somehow mobile app and if you could apply necessary changes to align with this?

just checked , mobile called `ConvertToKeycardAccount` , but looks like mobile doesn't have such thing `keycardUID ` yet? Probably, some UI relate work is needed @cammellos

saledjenic:
> > @qfrank I had to update `ConvertToKeycardAccount`, added `keycardUID` parameter there. Please check if that affects somehow mobile app and if you could apply necessary changes to align with this?
> 
> just checked , mobile called `ConvertToKeycardAccount` , but looks like mobile doesn't have such thing `keycardUID ` yet? Probably, some UI relate work is needed @cammellos

@qfrank would it work for you if I add the check like `if keycardUID != ""` and do the things necessary for the desktop app under `then` block, but you (mobile app) will need in that case to pass an empty string for `keycardUID` param cause the `ConvertToKeycardAccount` endpoint declaration has been changed? Is that ok?

saledjenic:
@qfrank I've added this check: https://github.com/status-im/status-go/blob/67e05a7111a6181ba31a257ad23bc913b8d8ccd4/api/geth_backend.go#L638-L640

The only update we need to do for the mobile app is passing an empty string to `ConvertToKeycardAccount` for `keycardUID`, could you do that after I merge this?

qfrank:
> but you (mobile app) will need in that case to pass an empty string for keycardUID param cause the ConvertToKeycardAccount endpoint declaration has been changed? Is that ok?

okay @saledjenic

PR #3455, opened 2023-05-04
Community encryption should automatically be enabled/disabled depending on community permissions (closed/open)

mprakhov:
This is a refactored version of https://github.com/status-im/status-go/pull/3417 PR as we do not need a separate UI for encryption and decryption, we always can know about it on a stage of creation/edition/removing TokenPermission

**Description:**
I've added a new function to Messenger UpdateCommunityEncryption(communityID [byte]).

UpdateCommunityEncryption takes a communityID and an encryption state, then finds the community and encrypts / decrypts the community based on TokenPermission. Community is republished along with any keys if needed.

I've added a new function to communities.Manager UpdateCommunity(c *Community).

UpdateCommunity takes a Community persists it and republishes it. The clock is incremented meaning even a no change update will be republished by the admin, and parsed by the member.

⚠️ Note:
This function cannot decrypt previously encrypted messages, and it cannot encrypt previous unencrypted messages.
This functionality has some known race conditions:
community description is processed by members before the receiving the key exchange messages
members maybe sending encrypted messages after the community description is updated and a new member joins

**Why Change?**
Community must be automatically encrypted/decrypted based on TokenPermission

Closes #
[#9942](https://github.com/status-im/status-desktop/issues/9942)

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -3334,3 +3334,17 @@ func (m *Manager) SetCommunityActiveMembersCount(communityID string, activeMembe
 
 	return nil
 }
+
+// UpdateCommunity takes a Community persists it and republishes it.
+// The clock is incremented meaning even a no change update will be republished by the admin, and parsed by the member.
+func (m *Manager) UpdateCommunity(c *Community) error {
+	c.increaseClock()
+
+	err := m.persistence.SaveCommunity(c)
+	if err != nil {
+		return err
+	}
+
+	m.publish(&Subscription{Community: c})
+	return nil
+}
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/ecdsa"
 	"encoding/json"
+	"errors"
 	_errors "errors"
 	"fmt"
 	"os"
@@ -1238,15 +1239,6 @@ func (m *Messenger) CreateCommunity(request *requests.CreateCommunity, createDef
 		response.AddChat(chatResponse.Chats()[0])
 	}
 
-	if request.Encrypted {
-		// Init hash ratchet for community
-		_, err = m.encryptor.GenerateHashRatchetKey(community.ID())
-
-		if err != nil {
-			return nil, err
-		}
-	}
-
 	response.AddCommunity(community)
 	response.AddCommunitySettings(&communitySettings)
 	err = m.syncCommunity(context.Background(), community, m.dispatchMessage)
@@ -1271,8 +1263,16 @@ func (m *Messenger) CreateCommunityTokenPermission(request *requests.CreateCommu
 		return nil, err
 	}
 
-	response := &MessengerResponse{}
-	response.AddCommunity(community)
+	response, err := m.UpdateCommunityEncryption(community)
+	if err != nil {
+		return nil, err
+	}
+
+	if response == nil {
+		response = &MessengerResponse{}
+		response.AddCommunity(community)
+	}
+
 	response.CommunityChanges = []*communities.CommunityChanges{changes}
 
 	return response, nil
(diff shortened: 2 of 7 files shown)


protocol/communities/manager.go
```diff
@@ -3334,3 +3334,17 @@ func (m *Manager) SetCommunityActiveMembersCount(communityID string, activeMembe
 
 	return nil
 }
+
+// UpdateCommunity takes a Community persists it and republishes it.
+// The clock is incremented meaning even a no change update will be republished by the admin, and parsed by the member.
+func (m *Manager) UpdateCommunity(c *Community) error {
+	c.increaseClock()
```
borismelnik, line 3341: maybe it should be below error processing?

borismelnik (approved):

protocol/messenger_communities.go
```diff
@@ -3536,3 +3560,54 @@ func (m *Messenger) AddCommunityToken(token *communities.CommunityToken) (*commu
...
 }
+
+// UpdateCommunityEncryption takes a communityID bytes, then finds the community and
+// encrypts / decrypts the community base on TokenPermission. Community is republished along with any keys if needed.
+//
+// Note: This function cannot decrypt previously encrypted messages, and it cannot encrypt previous unencrypted messages.
+// This functionality introduces some race conditions:
+//   - community description is processed by members before the receiving the key exchange messages
+//   - members maybe sending encrypted messages after the community description is updated and a new member joins
+func (m *Messenger) UpdateCommunityEncryption(communityID []byte) (*MessengerResponse, error) {
```
borismelnik: Maybe need to add the test for it :)

protocol/messenger_communities.go
```diff
@@ -3536,3 +3560,54 @@ func (m *Messenger) AddCommunityToken(token *communities.CommunityToken) (*commu
 func (m *Messenger) UpdateCommunityTokenState(contractAddress string, deployState communities.DeployState) error {
 	return m.communitiesManager.UpdateCommunityTokenState(contractAddress, deployState)
 }
+
+// UpdateCommunityEncryption takes a communityID bytes, then finds the community and
+// encrypts / decrypts the community base on TokenPermission. Community is republished along with any keys if needed.
```
```suggestion
// encrypts / decrypts the community based on TokenPermission. Community is republished along with any keys if needed.
```

protocol/messenger_communities.go
```diff
@@ -1288,8 +1296,16 @@ func (m *Messenger) EditCommunityTokenPermission(request *requests.EditCommunity
 		return nil, err
 	}
 
-	response := &MessengerResponse{}
-	response.AddCommunity(community)
+	response, err := m.UpdateCommunityEncryption(community.ID())
```
jrainville: I think it would be faster if you passed the whole community here instead of just the ID. No need to hit the DB.

jrainville (commented):
Good job. Just one suggestion

protocol/messenger_communities.go
mprakhov: fixed

protocol/messenger_communities.go
```diff
@@ -1238,15 +1239,6 @@ func (m *Messenger) CreateCommunity(request *requests.CreateCommunity, createDef
...
 	}
 
-	if request.Encrypted {
-		// Init hash ratchet for community
-		_, err = m.encryptor.GenerateHashRatchetKey(community.ID())
-
-		if err != nil {
-			return nil, err
-		}
-	}
```
0x-r4bbit, line 1248: Just want to make sure: this means, we can never create a community and have it encrypted right away. It will always be unencrypted by default.

This is desired, yes?

protocol/messenger_communities.go
```diff
@@ -3536,3 +3539,52 @@ func (m *Messenger) AddCommunityToken(token *communities.CommunityToken) (*commu
...
+			return nil, err
+		}
+
+		err = m.SendKeyExchangeMessage(community.ID(), community.GetMemberPubkeys(), common.KeyExMsgReuse)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// 🧙 There is no magic that decrypts a community, we just need to tell everyone to not use encryption
```
0x-r4bbit, line 3577: This comment is a little bit confusing (however this is mainly due to my lack of knowledge).
When and how do we "tell everyone to not use encryption" ?

I believe nodes check on `community.Encrypted()` themselves and then act accordingly no?

protocol/messenger_communities.go
mprakhov: yes

protocol/messenger_communities.go
mprakhov: From what I know - yes, they receive community updates with the community.Encrypted() == false and act accordingly.
Again, I just refactored the implementation not to use API, but to set encryption automatically.

jrainville (approved):

PR #3460, opened 2023-05-07
Return consistent version of muteTill

ibrkhalil:
muteTill was sometimes returned with milliseconds and sometimes without, This was problematic


protocol/messenger.go
```diff
@@ -4989,7 +4989,13 @@ func (m *Messenger) MuteChat(request *requests.MuteChat) (time.Time, error) {
 		return time.Time{}, err
 	}
 
-	return m.muteChat(chat, contact, MuteTill)
+	muteTillTimeRemoveMS, err := time.Parse("2006-01-02T15:04:05Z", MuteTill.Format("2006-01-02T15:04:05Z"))
```
ilmotta: The suffix convention is usually `Ms`, since millisecond is a single word. `MS` looks like Microsoft ;)

ilmotta (approved):

protocol/messenger.go
ibrkhalil: Okay, Thanks Icaro!

PR #3462, opened 2023-05-08
sync message for `delete for me` should not be sent to someone else

qfrank:
- sync message for `delete for me` should not be sent to someone else. 
- small fix when invoking `getConnectedMessages`


protocol/messenger_handler.go
```diff
@@ -1702,7 +1702,7 @@ func (m *Messenger) HandleDeleteForMeMessage(state *ReceivedMessageState, delete
 		return errors.New("chat not found")
```
qfrank, line 1711: hi @jrainville , could you pls explain more the TODO you add at line 1684 ? thank you

qfrank: generally , we prefer to add an issue link with TODO 🙂

jrainville: Oops, that's a leftover from my previous PR on this chunk of code. 
I added that comment to remember that I needed to change the code in this function, which I did.
I just forgot to remove it. Also, no one saw it in the review :thinking: 

Can you delete it for me please? :slightly_smiling_face:

protocol/messenger_handler.go
```diff
@@ -1703,7 +1703,7 @@ func (m *Messenger) HandleDeleteForMeMessage(state *ReceivedMessageState, delete
...
 	}
 
-	messagesToDelete, err := m.getConnectedMessages(originalMessage, deleteForMeMessage.LocalChatID)
+	messagesToDelete, err := m.getConnectedMessages(originalMessage, deleteForMeMessage.ChatId)
```
jrainville, line 1714: Was `LocalChatID` not right? Wouldn't this make it not work correctly in one to one chats?

jrainville (commented):
Looks good, just one question

protocol/messenger_delete_message_for_me_test.go
osmaczko, line 1: As far as I understand, the case "sync message for delete for me should not be sent to someone else" is not tested properly here :thinking: 

`otherMessenger.retrieveAll()` is never called, it does not even has a chance to retrieve `SYNC_DELETE_FOR_ME_MESSAGE` message.

protocol/messenger_messages.go
```diff
@@ -287,27 +287,29 @@ func (m *Messenger) DeleteMessageForMeAndSync(ctx context.Context, chatID string
...
 
 	if m.hasPairedDevices() {
-		clock, _ := chat.NextClockAndTimestamp(m.getTimesource())
-
-		deletedForMeMessage := &DeleteForMeMessage{}
+		err = m.withChatClock(func(chatID string, clock uint64) error {
+			deletedForMeMessage := &protobuf.DeleteForMeMessage{
+				MessageId: messageID,
+				Clock:     clock,
+				ChatId:    message.ChatId,
```
osmaczko: Not sure I understand why this fixes the issue, isn't the chat id already specified in `common.RawMessage` (line 304)?

protocol/messenger_handler.go
qfrank: sure😉

protocol/messenger_handler.go
qfrank: just update assign ChatId to message.LocalChatID, i think the old LocalChatID can be removed? because i can't find where we assigned it? @cammellos

protocol/messenger_delete_message_for_me_test.go
qfrank: fixed

protocol/messenger_messages.go
qfrank: line 304 won't be transfered to protocol.DeleteForMeMessage(HandleDeleteForMeMessage used).LocalChatID

osmaczko: Sorry, I still don't see how it fixes the "sync message for delete for me should not be sent to someone else", as to whom the message is sent is declared in `common.RawMesssage` or am I missing something?

osmaczko: Oh, now I see it :eyes: original `chatID` argument is shadowed by the one from the `m.withChatClock` callback which is effectively chat id of ourselves. Do I get it right?

(I am surprised linter does not check against it)

osmaczko (approved):

osmaczko (approved):

jrainville (approved):

protocol/messenger_messages.go
qfrank: > Do I get it right?

you are right 😉

PR #3465, opened 2023-05-08
fix(logrotation): fix issue where all logs are kept in archive

jrainville:
Part of https://github.com/status-im/status-desktop/issues/3610

There is a problem/weird design decision with Lumberjack (the log rotator lib we use) where if `MaxBackups` is set to 0, all archived log files will be kept.

I tried setting `MaxAge` to 28 and it still kept every file.

See docs here https://pkg.go.dev/gopkg.in/natefinch/lumberjack.v2@v2.0.0#readme-cleaning-up-old-log-files

The solution I added was to make sure `MaxBackups` is set to 1 at a minimum.


logutils/logrotation.go
```diff
@@ -20,6 +20,12 @@ type FileOptions struct {
 
 // FileHandlerWithRotation instantiates log.Handler with a configured rotation
 func FileHandlerWithRotation(opts FileOptions, format log.Format) log.Handler {
+	maxBackups := opts.MaxBackups
+	if maxBackups == 0 {
+		// Setting MaxBackups to 0 causes all log files to be kept. Even setting MaxAge to > 0 doesn't fix it
+		// Docs: https://pkg.go.dev/gopkg.in/natefinch/lumberjack.v2@v2.0.0#readme-cleaning-up-old-log-files
+		maxBackups = 1
+	}
```
osmaczko: I believe it would not be advisable to override this parameter at this level of low abstraction. Instead, it should be propagated through the higher layers using the configuration:
https://github.com/status-im/status-go/blob/31144ed5a38589fd129945b634f54e800b39645a/params/config.go#L409-L410

osmaczko (changes requested):

logutils/logrotation.go
jrainville: I also just realized I never actually used `maxBackups` :facepalm:

jrainville: updated

osmaczko: Hm.. you still override it ~~here~~ in low abstraction :thinking:

Isn't 0 an library specific for infinite?

jrainville: Yeah. What do you recommend? I also override it in the node_configuration in the desktop side.
I can create a migration to update the DB value of the setting, but it felt like more work for a very small fix.

osmaczko: >  I also override it in the node_configuration in the desktop side.

Exactly that. Are changes here required then?

jrainville: > Exactly that. Are changes here required then?

The changes here are like a fail proof. To make sure someone doesn't set 0 thinking it does no archives. There is the side effect that if someone **really** wants to keep all archives and knows the library, they might set it to 0 and then be surprised, but I doubt that would happen. Also, it's a bit crazy to want to keep all archives. A dev would set it to a high number like 10 instead IMO.

It also takes care of previous accounts, since `node_configuration` settings are only taken into account for new accounts.

osmaczko: I agree, without prior knowledge of the documentation, most developers would intuitively interpret a value of '0' to signify 'no logging' rather than 'log all' within the context of node configuration.

~~I would personally go with:~~
```diff
diff --git a/logutils/override.go b/logutils/override.go
index 7505bca20..0d980ae6b 100644
--- a/logutils/override.go
+++ b/logutils/override.go
@@ -51,7 +51,7 @@ func OverrideRootLog(enabled bool, levelStr string, fileOpts FileOptions, termin
     var (
         handler log.Handler
     )
-    if fileOpts.Filename != "" {
+    if fileOpts.Filename != "" && fileOpts.MaxBackups > 0 {
         handler = FileHandlerWithRotation(fileOpts, log.LogfmtFormat())
     } else {
         handler = log.StreamHandler(os.Stderr, log.TerminalFormat(terminal))

```

~~instead of tempering with the parameter value itself.~~

The approach I wanted to suggest wouldn't work as expected because it would not create a file in case of `MaxBackups == 0`.

osmaczko (approved):

logutils/logrotation.go
jrainville: That would mean that all old accounts would lose the file log. I guess it kinda makes sense, since 0 means no archive :thinking: 

Also, does `log.StreamHandler` just print everything in the console? That could be crazy lol.

I did add a setting in Desktop to modify the MaxBackups value, so we can edit it in our case. It might be a bit confusing for those not aware though.

I'll try it out.

jrainville: > Also, does log.StreamHandler just print everything in the console?

Yup, that's what it does. In Debug mode, it goes wild lol.

borismelnik (approved):

PR #3467, opened 2023-05-09
Add Messenger.ReverseResolveENS() API

0x-r4bbit:
This is so clients can do ad-hoc checks on chain if wallet addresses own any particular ENS.

### DIFF
--- a/eth-node/bridge/geth/ens/ens.go
+++ b/eth-node/bridge/geth/ens/ens.go
@@ -11,6 +11,7 @@ import (
 	ens "github.com/wealdtech/go-ens/v3"
 	"go.uber.org/zap"
 
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethclient"
 
 	"github.com/status-im/status-go/eth-node/crypto"
@@ -28,6 +29,17 @@ type Verifier struct {
 // NewVerifier returns a Verifier attached to the specified logger
 func NewVerifier(logger *zap.Logger) *Verifier {
 	return &Verifier{logger: logger}
+}
+
+func (m *Verifier) ReverseResolve(address common.Address, rpcEndpoint string) (string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), contractQueryTimeout)
+	defer cancel()
+
+	ethClient, err := ethclient.DialContext(ctx, rpcEndpoint)
+	if err != nil {
+		return "", err
+	}
+	return ens.ReverseResolve(ethClient, address)
 }
 
 func (m *Verifier) verifyENSName(ensInfo enstypes.ENSDetails, ethclient *ethclient.Client) enstypes.ENSResponse {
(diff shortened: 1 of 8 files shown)


0x-r4bbit:
There's an argument to be made that one could rely on `ens_getEnsUsernames` instead, but that RPC will always return the stored ENS from the DB and it doesn't look like it's updated regularly (meaning, unless a client takes care of checking ownership on-chain, nobody is going to remove entries if there's no ownership anymore).

Hence I've decided to introduce this API here which always checks on-chain.

Also just noticed we have `ens_GetName()` which uses `ReverseResolve()` under the hood but will throw, while `ReverseResolveENS()` does not and always returns a list of names (or an empty list).

@cammellos lemme know what you think about this, it does indeed feel like this wants to be changed/adjusted.

cammellos:
@0x-r4bbit I don't mind the endpoint, but how are you going to use it in desktop? It's a fairly low level endpoint, so it might be an indication that some logic is in desktop that should be in status-go instead (might be that it wants to be in desktop/clients, but just to make sure)

0x-r4bbit:
>  don't mind the endpoint, but how are you going to use it in desktop?

In desktop we call it with all owned wallet addresses of the currently logged-in user.
Once we receive the list of owned ENS names, we can check them against token permissions (eg has `.stateofus.eth` suffix or match equal xyz)

In desktop we updated the wallet token balances every 10 mins (I think), at the same time we'll check token permissions, so this will inherently check ENS names every 10 mins as well

cammellos:
> > don't mind the endpoint, but how are you going to use it in desktop?
> 
> In desktop we call it with all owned wallet addresses of the currently logged-in user. Once we receive the list of owned ENS names, we can check them against token permissions (eg has `.stateofus.eth` suffix or match equal xyz)
> 
> In desktop we updated the wallet token balances every 10 mins (I think), at the same time we'll check token permissions, so this will inherently check ENS names every 10 mins as well

this looks like logic it should be in status-go, since it's something any client would probably want to do (for sure mobile), and use signals/etc to notify the client of a change, unless there's a clear benefit in keeping it in the client

0x-r4bbit:
@cammellos That's fair enough. We could leverage `hasPermission()` in remove all local checks in clients entirely..
I'm going to look into it.

cammellos:
> @cammellos That's fair enough. We could leverage `hasPermission()` in remove all local checks in clients entirely.. I'm going to look into it.

I started moving some of the balance checking code to status-go here https://github.com/status-im/status-go/tree/feature/wallet-config-mobile

0x-r4bbit:
I just realized that simply relying on hasPermission() won't do it because at least in desktop, we need to know exactly which of the token requirements are met/not met.

So even if we check for all permissions once in status-go, we might still need to check for, say, ens names individually (which we need such an API for, unless we rely on already connected ENS exclusively)

iurimatias (approved):

0x-r4bbit:
I'm going to close this PR.
The original idea was that clients like desktop and mobile can use this to figure out whether a given ENS token permission criteria was satisfied or not.
We will instead rely purely on all checks done in `CheckPermissionsToJoinCommunity()` so clients don't have to individually check tokens/collectibles and ENS.

PR #3471, opened 2023-05-10
URL unfurling (initial implementation)

ilmotta:
### Summary

This PR is the initial implementation for the new _URL unfurling_ requirements. The most important one is that only the message sender will pay the _privacy cost_ for unfurling and extracting metadata from websites. Once the message is sent, the unfurled data will be stored at the protocol level and receivers will just profit and happily decode the metadata to render it.

There's a lot to share here because this is the first PR and I accumulated notes throughout development.

Further development of this _URL unfurling_ capability will be mostly guided by issues created on clients. For the moment in `status-mobile`: https://github.com/status-im/status-mobile/labels/url-preview

- https://github.com/status-im/status-mobile/issues/15918
- https://github.com/status-im/status-mobile/issues/15917
- https://github.com/status-im/status-mobile/issues/15910
- https://github.com/status-im/status-mobile/issues/15909
- https://github.com/status-im/status-mobile/issues/15908
- https://github.com/status-im/status-mobile/issues/15906
- https://github.com/status-im/status-mobile/issues/15905

### Demo

I have used this branch to implement the mobile experience you see in the video below. The mobile PR will be created once I'm confident (based on your reviews) that the solution is good enough, but ~**it will exist behind a dev-only feature flag**, so that's why this status-go PR should be 100% backwards compatible.~ Edit: The feature will be available in mobile.

[link-previews.webm](https://github.com/status-im/status-go/assets/46027/bff0290e-c532-4802-9ebc-30d28ab41e7b)

### Terminology

In the code, I've tried to stick to the word "unfurl URL" to really mean the process of extracting metadata from a website, sort of lower level. I use "link preview" to mean a higher level structure which is enriched by unfurled data. "link preview" is also how designers refer to it.

### User flows

Please, expand the **Details** section below to see detailed diagrams about user flows.

<details>
<summary><b>:point_down: See details</b></summary>

1. Carol needs to see link previews while typing in the chat input field. Notice from the diagram nothing is persisted and that status-go endpoints are essentially _stateless_.

```
#+begin_src plantuml :results verbatim
  Client->>Server: Call wakuext_getTextURLs
  Server-->>Client: Normalized URLs
  Client->>Client: Render cached unfurled URLs
  Client->>Server: Unfurl non-cached URLs.\nCall wakuext_unfurlURLs
  Server->>Website: Fetch metadata
  Website-->>Server: Metadata (thumbnail URL, title, etc)
  Server->>Website: Fetch thumbnail
  Server->>Website: Fetch favicon
  Website-->>Server: Favicon bytes
  Website-->>Server: Thumbnail bytes
  Server->>Server: Decode & process images
  Server-->>Client: Unfurled data (thumbnail data URI, etc)
#+end_src
```

```
     ,------.                                 ,------.                             ,-------.
     |Client|                                 |Server|                             |Website|
     `--+---'                                 `--+---'                             `---+---'
        |        Call wakuext_getTextURLs        |                                     |
        | --------------------------------------->                                     |
        |                                        |                                     |
        |             Normalized URLs            |                                     |
        | <- - - - - - - - - - - - - - - - - - - -                                     |
        |                                        |                                     |
        |----.                                   |                                     |
        |    | Render cached unfurled URLs       |                                     |
        |<---'                                   |                                     |
        |                                        |                                     |
        |         Unfurl non-cached URLs.        |                                     |
        |         Call wakuext_unfurlURLs        |                                     |
        | --------------------------------------->                                     |
        |                                        |                                     |
        |                                        |            Fetch metadata           |
        |                                        | ------------------------------------>
        |                                        |                                     |
        |                                        | Metadata (thumbnail URL, title, etc)|
        |                                        | <- - - - - - - - - - - - - - - - - -
        |                                        |                                     |
        |                                        |           Fetch thumbnail           |
        |                                        | ------------------------------------>
        |                                        |                                     |
        |                                        |            Fetch favicon            |
        |                                        | ------------------------------------>
        |                                        |                                     |
        |                                        |            Favicon bytes            |
        |                                        | <- - - - - - - - - - - - - - - - - -
        |                                        |                                     |
        |                                        |           Thumbnail bytes           |
        |                                        | <- - - - - - - - - - - - - - - - - -
        |                                        |                                     |
        |                                        |----.                                |
        |                                        |    | Decode & process images        |
        |                                        |<---'                                |
        |                                        |                                     |
        | Unfurled data (thumbnail data URI, etc)|                                     |
        | <- - - - - - - - - - - - - - - - - - - -                                     |
     ,--+---.                                 ,--+---.                             ,---+---.
     |Client|                                 |Server|                             |Website|
     `------'                                 `------'                             `-------'
```

2. Carol _sends_ the text message with link previews in the RPC request `wakuext_sendChatMessages`. status-go assumes the link previews are good because it can't and shouldn't attempt to re-unfurl them.

```
#+begin_src plantuml :results verbatim
  Client->>Server: Call wakuext_sendChatMessages
  Server->>Server: Transform link previews to\nbe proto-marshalled
  Server->DB: Write link previews serialized as JSON
  Server-->>Client: Updated message response
#+end_src
```

```
     ,------.                       ,------.                                  ,--.
     |Client|                       |Server|                                  |DB|
     `--+---'                       `--+---'                                  `+-'
        | Call wakuext_sendChatMessages|                                       |
        | ----------------------------->                                       |
        |                              |                                       |
        |                              |----.                                  |
        |                              |    | Transform link previews to       |
        |                              |<---' be proto-marshalled              |
        |                              |                                       |
        |                              |                                       |
        |                              | Write link previews serialized as JSON|
        |                              | -------------------------------------->
        |                              |                                       |
        |   Updated message response   |                                       |
        | <- - - - - - - - - - - - - - -                                       |
     ,--+---.                       ,--+---.                                  ,+-.
     |Client|                       |Server|                                  |DB|
     `------'                       `------'                                  `--'
```

3. The message was sent over waku and persisted locally in Carol's device. She should now see the link previews in the chat history. There can be many link previews shared by other chat members, therefore it is important to serve the assets via the media server to avoid overloading the ReactNative bridge with lots of big JSON payloads containing base64 encoded data URIs (maybe this concern is meaningless for desktop). When a client is rendering messages with link previews, they will have the field `linkPreviews`, and the thumbnail URL will point to the local media server.

```
 #+begin_src plantuml :results verbatim
   Client->>Server: GET /link-preview/thumbnail (media server)
   Server->>DB: Read from user_messages.unfurled_links
   Server->Server: Unmarshal JSON
   Server-->>Client: HTTP Content-Type: image/jpeg/etc
 #+end_src
```

```
     ,------.                                    ,------.                                  ,--.
     |Client|                                    |Server|                                  |DB|
     `--+---'                                    `--+---'                                  `+-'
        | GET /link-preview/thumbnail (media server)|                                       |
        | ------------------------------------------>                                       |
        |                                           |                                       |
        |                                           | Read from user_messages.unfurled_links|
        |                                           | -------------------------------------->
        |                                           |                                       |
        |                                           |----.                                  |
        |                                           |    | Unmarshal JSON                   |
        |                                           |<---'                                  |
        |                                           |                                       |
        |     HTTP Content-Type: image/jpeg/etc     |                                       |
        | <- - - - - - - - - - - - - - - - - - - - -                                        |
     ,--+---.                                    ,--+---.                                  ,+-.
     |Client|                                    |Server|                                  |DB|
     `------'                                    `------'                                  `--'
```

</details>

### Some limitations of the current implementation

The following points will become separate issues in status-go that I'll work on over the next couple weeks. In no order of importance:

- Improve how multiple links are fetched; retries on failure and testing how unfurling behaves around the timeout limits (deterministically, not by making real HTTP calls as I did). https://github.com/status-im/status-go/issues/3498
- Unfurl favicons and store them in the protobuf too.
- For this PR, I added unfurling support only for websites with [OpenGraph](https://ogp.me/) meta tags. Other _unfurlers_ will be implemented on demand. The next one will probably be for [oEmbed](https://oembed.com/), the protocol supported by YouTube, for example.
- Resize and/or compress thumbnails (and favicons). Often times, thumbnails are huge for the purposes of link previews. There is already support for compressing JPEGs in status-go, but I prefer to work with compression in a separate PR because I'd like to also solve the problem for PNGs (probably convert them to JPEGs, plus compress them). This would be a safe choice for thumbnails, favicons not so much because transparency is desirable.
- Editing messages is not yet supported.
- I haven't coded any artificial limit on the number of previews or on the size of the thumbnail payload. This will be done in a separate issue. I have heard the ideal solution may be to split messages into smaller chunks of ~125 KiB because of libp2p, but that might be too complicated at this stage of the product (?).
- Link preview deletion.
- ~Add support for the sender (only) to build up a dynamic allowlist.~ No allowlist will be necessary.
- For the moment, OpenGraph metadata is extracted by requesting data for the English language (and fallback to whatever is available). In the future, we'll want to unfurl by respecting the user's local device language. Some websites, like GoDaddy, are already localized based on the device's IP, but many aren't.
- The website's description text should be limited by a certain number of characters, especially because it's outside our control. Exactly how much has not been decided yet, so it'll be done separately.
- URL normalization can be tricky, so I implemented only the basics to help with caching. For example, the url https://status.im and HTTPS://status.im are considered identical. Also, a URL is considered valid for unfurling if its TLD exists according to `publicsuffix.EffectiveTLDPlusOne`. This was essential, otherwise the default Go `url.Parse` approach would consider many invalid URLs valid, and thus the server would waste resources trying to unfurl the unfurleable.

### Other requirements

- If the message is edited, the link previews should reflect the edited text, not the original one. This has been aligned with the design team as well.
- If the website's thumbnail or the favicon can't be fetched, just ignore them. The only mandatory piece of metadata is the website's title and URL.
- Link previews in clients should be generated in near real-time, that is, as the user types, previews are updated. In mobile this performs very well, and it's what other clients like WhatsApp, Telegram, and Facebook do.

### Decisions

Here are the important decisions I have made for this PR.

- While the user typing in the input field, the client is constantly (debounced) asking status-go to parse the text and extract normalized URLs and then the client checks if they're already in its in-memory cache. If they are, no RPC call is made. I chose this approach to achieve the best possible performance in mobile and avoid the whole RPC overhead, since the chat experience is already **not smooth enough**. The mobile client uses URLs as cache keys in a hashmap, i.e. if the key is present, it means the preview is readily available (naive, but good enough for now). This decision also gave me more flexibility to find the best UX at this stage of the feature.

- Due to the requirement that users should be able to see independent *loading indicators* for each link preview, when status-go can't unfurl a URL, it doesn't return it in the response.

- As an initial implementation, I added the `BLOB` column `unfurled_links` to the `user_messages` table. The preview data is then serialized as JSON before being stored in this column. I felt that creating a separate table and the related code for this initial PR would be inconvenient. Is that reasonable to you? Once things stabilize I can create a proper table if we want to avoid this kind of solution with serialized columns.

### Don't we already have code to unfurl URLs?

Yes, in `protocol/urls/`, but I found it to be far from what we needed, so I created a new package in `protocols/linkpreview`. Once the new implementation stabilizes in the upcoming status-go PRs, I'll remove the old one. Nevertheless, the old implementation most certainly helped during development, since there are similarities.

~Currently, this PR should have no production impact, because the implementation in status-mobile is behind a feature flag, and status-desktop hasn't started the work yet. So we have time to gradually improve the implementation in follow-up PRs.~ Edit: the feature will be included in the next mobile release due to priority changes.


ghost:
### Pull Request Checklist

protocol/chat_test.go
```diff
@@ -129,7 +129,7 @@ func (s *ChatTestSuite) TestSerializeJSON() {
 
 	message.From = "0x04deaafa03e3a646e54a36ec3f6968c1d3686847d88420f00c0ab6ee517ee1893398fca28aacd2af74f2654738c21d10bad3d88dc64201ebe0de5cf1e313970d3d"
 	message.Clock = 1
-	message.Text = "`some markdown text`"
+	message.Text = "`some markdown text` https://status.im"
```
cammellos, line 132: Is this change relevant? we just want to check it doesn't choke on links?

protocol/linkpreview/linkpreview.go
```diff
@@ -0,0 +1,321 @@
...
+		}
+	}
+
+	return urls
+}
+
+// UnfurlURLs assumes clients pass URLs verbatim that were validated and
+// processed by GetURLs.
+func UnfurlURLs(urls []string) ([]common.LinkPreview, error) {
+	logger, err := zap.NewDevelopment()
```
cammellos: generally is passed around, so it logs in the right place, not a big deal though

cammellos:
@ilmotta amazing work, awesome quality!

> As an initial implementation, I added the BLOB column unfurled_links to
the user_messages table. The preview data is then serialized as JSON before
being stored in this column. I felt that creating a separate table and the
related code for this initial PR would be inconvenient. Is that reasonable to
you? Once things stabilize I can create a proper table if we want to avoid
this kind of solution with serialized columns.

Moving to a table would be a breaking change most likely, since we don't want to support both, and migrating would be painful.
I think storing them as a BLOB in this case is probably ok, it can't be migrated easily though, so if you expect changes in the structure, that's going to be annoying, I am a bit ambivalent on this one, so probably someone else might have a stronger opinion? @Samyoul?

> Add support for the sender (only) to build up a dynamic allowlist.

I believe that's not necessary, we agreed with design/john that it would only be a on/off toggle

 > Resize and/or compress thumbnails (and favicons).

This could be important, depending on the size of images, but of course better in a separate PR

>  I have heard
the ideal solution may be to split messages into smaller chunks of 

It could be, but probably is best to limit to a safe number, but only after we implement compression.


Again, amazing work!

cammellos (approved):

ilmotta:
> It could be, but probably is best to limit to a safe number, but only after we implement compression.

Good to know because just limiting is so much simpler also. I talked to John about this, but he hold the opinion that it would be better to avoid hardcoding a limit. We eventually agreed on the number 5. It's open for discussion how we want to limit. And I agree with you, an artificial limit is a safe choice for now.

> generally is passed around, so it logs in the right place, not a big deal though

I didn't quite understand @cammellos. In this part I had to instantiate a zap logger and I passed it around. Should I do it differently?

> I believe that's not necessary, we agreed with design/john that it would only be a on/off toggle

Cool, I got that wrong. So much simpler.

> Moving to a table would be a breaking change most likely, since we don't want to support both, and migrating would be painful.

Yes, it would be painful. I saw a lot of serialized stuff in the user_messages table and I thought it be okayish since the implementation wouldn't be used yet (behind a feature flag). The broken window effect... I think the structure can accrete, for example when we need to store favicons. I was thinking I would just throw away the data in the migration since no user would be using the new unfurling yet. But now that you mentioned, I don't know, I could use a table for sure to avoid all this.

- Edit: I created issue https://github.com/status-im/status-go/issues/3499

> @ilmotta amazing work, awesome quality!

Thank you Andrea!

protocol/linkpreview/linkpreview.go
ilmotta: I didn't quite understand @cammellos. In this part I had to instantiate a zap logger and I passed it around. Should I do it differently?

protocol/chat_test.go
ilmotta: Not much, I guess it's one of those changes I did right at the beginning and it survived. Let me remove it.

ilmotta: Interestingly, I found a bug during development in the markdown parser. I still need to report it.

Extracted from the tests I wrote:

```go
// There is a bug in the code that builds the AST from markdown text,
// because it removes the closing parenthesis, which means it won't be
// possible to unfurl this URL.
{args: "https://en.wikipedia.org/wiki/Status_message_(instant_messaging)", expected: []string{"https://en.wikipedia.org/wiki/Status_message_(instant_messaging"}},
```

ilmotta:
@alexjba, @Samyoul, @caybro, @siphiuel, the implementation is mostly done and under scrutiny by QAs in `status-mobile` and soon I'll be able to merge both PRs (client & server).

I would prefer to merge this status-go PR with at least two approvals (but I only got one). Could one of you take a look at it please? Thanks a lot!

vitvly (approved):
A lesson in PR-making

protocol/linkpreview/linkpreview.go
ilmotta: Leaving the comment for reference, as @cammellos explained to me. It is important to get the logger instance from `api.service.messenger`, that way, logs are written to `geth.log`, otherwise they go to stdout only.

Done https://github.com/status-im/status-go/pull/3471/commits/a4039c6c0a1467a4425f419cd4321c358b9ef99c

PR #3473, opened 2023-05-11
feat: DecodeFromURL support self signed cert

yqrashawn:
-------

update the http client used by `DecodeFromURL` when `generateTLSCert`

so that we can call `StoreIdentityImageFromURL` with media server URL


yqrashawn:
close this PR, no longer need this

PR #3474, opened 2023-05-11
Community Rekey loop basics

### LINKED ISSUE #3404: Rotate community encryption keys periodically
We should rotate encryption keys periodically, say once a week, or when it hits a large enough sequence number (10k?) in order to make sure that the latest could be retrieved from a mailserver, and that users won't have to hash too many times to decode a message.

It's also a good opportunity to optimize rekeying.

Currently, we send a message to each user, that's wasteful in large enough communities. 
An alternative, is to send a protobuf in the form (roughly):

```
message RekeyMessage {
  string key_id = 1;
  map<string><bytes> encrypted_keys;
}
```

Where key_id is the id of the new key

while the map is indexed by the first 4 bytes (could be more/less) of the public key of the user (to save on space, and avoid too many collisions), and the value is an array of keys, encrypted with the pk of the user(s). 
The array could be either a separate protobuf, or to save space, and provided encryption is constant size, the concatenation of the encrypted keys.

A user receiving this message, would take the first 4 bytes of their key, lookup into the map to see if there's anything, if there's anything, it would split it in chunk of size `n`, where `n` is the size of the encrypted key, and then it would trial/decrypt the key.

Other methods/solutions are possible of course, feel free to discuss any and/or improvements/changes.

Samyoul:
## What's Changed?

Introduced the concept of `RekeyedAt` on the `Community` level, allowing for the tracking and managing of `Community` rekeying events.

Introduced the concept of `RekeyInterval` on the `communities.Manager` level, allowing for custom interval setting. Currently limited only to testing environment but can be expanded to be config based in addition.

Introduces a refactor of `UpdateCommunityEncryption` to removed external concerns from `Community` encryption management as the role of this function should be to update a `Community`'s encryption state based **ONLY** on a boolean, this allows all calling code to set their conditions independent of other state concerns.

Adds an explicit `RekeyCommunity` function.

Adds a new `Messenger` loop `startCommunityRekeyLoop()`, using a default ticker value of 5 minutes, and a default rekey interval of 1 hour. 

Exposed `GetKeyIDsForGroup()` at the `MessageSender` level. Required for testing, in order to prove new keys existed and therefore the rekeying process was successfully firing, a inspection function has been exposed.

## Why Change?

Partially resolves https://github.com/status-im/status-go/issues/3404 and provides the framework for periodic `Community` rekeying.


protocol/messenger_communities.go
```diff
@@ -1286,7 +1286,8 @@ func (m *Messenger) CreateCommunityTokenPermission(request *requests.CreateCommu
 		return nil, err
 	}
 
-	response, err := m.UpdateCommunityEncryption(community)
+	becomeMemberPermissions := community.TokenPermissionsByType(protobuf.CommunityTokenPermission_BECOME_MEMBER)
+	response, err := m.UpdateCommunityEncryption(community, len(becomeMemberPermissions) > 0)
```
Samyoul: I've refactored this function's calls and signature as the previous implementation polluted the concept of updating a `Community`'s encryption state with other conditionals. This refactor moves the conditionals to outside the function.

protocol/messenger_communities.go
```diff
@@ -3567,28 +3589,25 @@ func (m *Messenger) UpdateCommunityTokenState(contractAddress string, deployStat
...
 //   - members maybe sending encrypted messages after the community description is updated and a new member joins
-func (m *Messenger) UpdateCommunityEncryption(community *communities.Community) (*MessengerResponse, error) {
+func (m *Messenger) UpdateCommunityEncryption(community *communities.Community, useEncryption bool) (*MessengerResponse, error) {
 	if community == nil {
 		return nil, errors.New("community is nil")
 	}
 
-	becomeMemberPermissions := community.TokenPermissionsByType(protobuf.CommunityTokenPermission_BECOME_MEMBER)
-	isEncrypted := len(becomeMemberPermissions) > 0
-
```
Samyoul, line 4079: As noted above I've removed external concerns from `Community` encryption management as the role of this function should be to update a `Community`'s encryption state based ONLY on a boolean, this allows all calling code to set their conditions independent of other state concerns.

protocol/messenger_communities.go
```diff
@@ -3635,3 +3654,64 @@ func (m *Messenger) CheckPermissionsToJoinCommunity(request *requests.CheckPermi
...
+	}
+
+	// Get and loop over all communities in persistence
+	cs, err := m.Communities()
+	if err != nil {
+		logger.Error("error getting communities", zap.Error(err))
+		return
+	}
+	for _, c := range cs {
+		if c.IsAdmin() && c.Encrypted() && c.RekeyedAt().Add(rki).Before(time.Now()) {
```
vitvly: Probably does not make much sense, but: would it be ok to move the boolean checks to SQL query itself, instead of using a generic `m.Communities()` call?

vitvly (approved):

protocol/messenger_communities.go
Samyoul: I see what you mean, that would be a much more efficient and clean.

But unfortunately a SQL query for this information is not possible for 2 reasons:

- I'm refactoring this PR to remove the concept of `rekeyedAt` and instead rely on the keyID timestampID which is an encryptor concern. (although technically we could query across multiple tables that exist in the same DB I don't know how messy that would get.)
- The encrypted state of community is stored in the description which is marshalled into bytes in the DB so isn't searchable.

vitvly: Ahhh, of course, clear now.

Samyoul:
Currently blocked by https://github.com/status-im/status-go/issues/3675

protocol/messenger_communities.go
```diff
@@ -4205,3 +4216,88 @@ func chunkAttachmentsByByteSize(slice []*protobuf.DiscordMessageAttachment, maxF
...
+			logger.Error("error getting current keyTimestampID for community", zap.Error(err), zap.Binary("community ID", c.ID()))
+			continue
+		}
+
+		// TODO add functionality to encryptor.go that compares the timestamps and returns a bool
+		// c.RekeyedAt().Add(rki).Before(time.Now())
+		// keyTimestampID + rki < time.Now()
+		deleteThisFunction(rki, keyTimestampID)
+
+		if c.IsAdmin() && c.Encrypted() { // && c.RekeyedAt().Add(rki).Before(time.Now())
```
```suggestion
		if c.IsControlNode() && c.Encrypted() { // && c.RekeyedAt().Add(rki).Before(time.Now())
```

Only control node can rekey.

protocol/messenger_communities.go
```diff
@@ -4205,3 +4216,88 @@ func chunkAttachmentsByByteSize(slice []*protobuf.DiscordMessageAttachment, maxF
...
+		keyTimestampID, err := m.GetCurrentKeyForGroup(c.ID())
+		if err != nil {
+			logger.Error("error getting current keyTimestampID for community", zap.Error(err), zap.Binary("community ID", c.ID()))
+			continue
+		}
+
+		// TODO add functionality to encryptor.go that compares the timestamps and returns a bool
+		// c.RekeyedAt().Add(rki).Before(time.Now())
+		// keyTimestampID + rki < time.Now()
+		deleteThisFunction(rki, keyTimestampID)
```
osmaczko: :thinking:

osmaczko (approved):

protocol/messenger_communities.go
Samyoul: Ah, a new concept that I didn't have before. Ok I'll change that

protocol/messenger_communities.go
Samyoul: Haha, I suppose I should `deleteThatFunction`

PR #3475, opened 2023-05-11
`keypair` struct, db table introduction

saledjenic:
Recently we introduced a new term "keypair" which binds together all accounts derived from the same master key. 

An addition is that a single keypair may be migrated to a single keycard and also it is identified and displayed in many UI screens in the app. Because of its role it deserves to be better identified on the db level.

This PR is about keypair syncing (covering devices syncing, sync on user's activity, recovering from other device and recovering from waku).


saledjenic:
Mobile app needs to align with changes done here before we merge this PR.

saledjenic:
@cammellos @qfrank kindly reminder if you could review this PR and/or add someone from mobile team who will review this and potentially apply necessary changes to the mobile app.

qfrank:
> @cammellos @qfrank kindly reminder if you could review this PR and/or add someone from mobile team who will review this and potentially apply necessary changes to the mobile app.

Sorry, got a long meeting yesterday, continue reviewing now

multiaccounts/accounts/database.go
```diff
@@ -382,87 +590,225 @@ func (db *Database) SaveAccounts(accounts []*Account) (err error) {
...
+	// If keypair is being saved, not updated, then it must be at least one account and all accounts must have the same key uid.
+	dbKeypair, err := db.getKeypairByKeyUID(tx, keypair.KeyUID)
+	if err != nil && err != ErrDbKeypairNotFound {
+		return err
 	}
+	if dbKeypair == nil {
+		validAccounts := len(keypair.Accounts) > 0
+		for _, acc := range keypair.Accounts {
+			if acc.KeyUID == "" || acc.KeyUID != keypair.KeyUID {
+				validAccounts = false
```
qfrank: we can return Err directly from here?

api/backend_test.go
```diff
@@ -839,17 +884,22 @@ func TestConvertAccount(t *testing.T) {
...
+	}
+
+	profileKeypair.Accounts = append(profileKeypair.Accounts, &accounts.Account{
+		Address:   types.HexToAddress(chatAddress),
+		KeyUID:    profileKeypair.KeyUID,
+		Type:      accounts.AccountTypeGenerated,
+		PublicKey: types.Hex2Bytes(accountInfo.PublicKey),
+		Path:      pathEIP1581Chat,
+		Wallet:    false,
+		Chat:      true,
```
qfrank, line 901: Chat: true?

api/backend_test.go
```diff
@@ -1072,7 +1121,7 @@ func login(t *testing.T, conf *params.NodeConfig) {
 func TestLoginAndMigrationsStillWorkWithExistingUsers(t *testing.T) {
 	utils.Init()
 
-	srcFolder := "../static/test-0.97.3-account/"
+	srcFolder := "../static/test-0.132.0-account/"
```
qfrank, line 1124: LoginAndMigrations not WorkWithExistingUsers anymore? why changing?

api/geth_backend.go
```diff
@@ -619,27 +618,31 @@ func (b *GethStatusBackend) ConvertToKeycardAccount(account multiaccounts.Accoun
...
 	}
 
-	err = accountDB.SaveSettingField(settings.KeycardPairedOn, s.KeycardPairedOn)
+	err = accountDB.SaveSettingField(settings.KeycardInstanceUID, s.KeycardInstanceUID)
 	if err != nil {
 		return err
 	}
 
-	err = accountDB.SaveSettingField(settings.KeycardPairing, s.KeycardPairing)
+	err = accountDB.SaveSettingField(settings.KeycardPairedOn, s.KeycardPairedOn)
```
qfrank, line 635: can you add some comment on settings.KeycardPairedOn? i've no idea what's this field meaning.. thank you

api/geth_backend.go
```diff
@@ -619,27 +618,31 @@ func (b *GethStatusBackend) ConvertToKeycardAccount(account multiaccounts.Accoun
...
 	if err != nil {
 		return err
 	}
 
-	err = accountDB.SaveSettingField(settings.Mnemonic, nil)
+	err = accountDB.SaveSettingField(settings.KeycardPairing, s.KeycardPairing)
```
qfrank, line 640: can you add some comment on settings. KeycardPairing? same as above.. thank you

api/geth_backend.go
```diff
@@ -1058,6 +1061,7 @@ func (b *GethStatusBackend) StartNodeWithAccountAndInitialConfig(
 	return b.StartNodeWithAccount(account, password, nodecfg)
 }
 
+// TODO: change in `saveAccountsAndSettings` function param `subaccs []*accounts.Account` parameter to `profileKeypair *accounts.Keypair` parameter
```
qfrank, line 1064: won't include this change in this PR?

appdatabase/migrations/sql/1683627613_accounts_and_keycards_improvements.up.sql
```diff
@@ -0,0 +1,94 @@
...
+-- "fully"     // an account is fully operable if it is not a keycard account and there is a keystore file for it
+
+-- Adding new tables `keypairs` and `keypairs_accounts`
+CREATE TABLE IF NOT EXISTS keypairs (
+  key_uid VARCHAR PRIMARY KEY NOT NULL CHECK (length(trim(key_uid)) > 0),
+  name VARCHAR NOT NULL DEFAULT "",
+  type VARCHAR NOT NULL DEFAULT "",
+  derived_from VARCHAR NOT NULL DEFAULT "",
+  last_used_derivation_index INT NOT NULL DEFAULT 0,
+  synced_from VARCHAR NOT NULL DEFAULT "", -- keeps an info which device this keypair is added from
```
qfrank, line 13: can you add example value on this field?

appdatabase/migrations/sql/1683627613_accounts_and_keycards_improvements.up.sql
```diff
@@ -0,0 +1,94 @@
+-- Possible `operable` column values:
+-- "no"        // an account is non operable it is not a keycard account and there is no keystore file for it and no keystore file for the address it is derived from
+-- "partially" // an account is partially operable if it is not a keycard account and there is created keystore file for the address it is derived from
+-- "fully"     // an account is fully operable if it is not a keycard account and there is a keystore file for it
+
+-- Adding new tables `keypairs` and `keypairs_accounts`
+CREATE TABLE IF NOT EXISTS keypairs (
```
qfrank, line 7: after we added table keypairs, when user login, there should be/must exist only 1 record which's type is `KeypairTypeProfile` and we can have many records other than `KeypairTypeProfile`, correct?

appdatabase/migrations/sql/1683627613_accounts_and_keycards_improvements.up.sql
```diff
@@ -0,0 +1,94 @@
...
+  key_uid VARCHAR,
+  pubkey VARCHAR,
+  path VARCHAR NOT NULL DEFAULT "",
+  name VARCHAR NOT NULL DEFAULT "",
+  color VARCHAR NOT NULL DEFAULT "",
+  emoji VARCHAR NOT NULL DEFAULT "",
+  wallet BOOL NOT NULL DEFAULT FALSE,
+  chat BOOL NOT NULL DEFAULT FALSE,
+  hidden BOOL NOT NULL DEFAULT FALSE,
+  operable VARCHAR NOT NULL DEFAULT "no", -- describes an account's operability (read an explanation at the top of this file)
```
qfrank, line 28: can we split `operable` into 3 separate conditional fields so that code will looks more readable? 
is_keycard_account bool 
exist_keystore_file bool
exist_keystore_file_derived_from bool

appdatabase/migrations/sql/1683627613_accounts_and_keycards_improvements.up.sql
```diff
@@ -0,0 +1,94 @@
...
+  clock INT NOT NULL DEFAULT 0,
+  FOREIGN KEY(key_uid) REFERENCES keypairs(key_uid) 
+    ON DELETE CASCADE
+);
+
+-- Fulfilling the tables
+INSERT INTO keypairs 
+  SELECT key_uid, keypair_name, "profile", derived_from, last_used_derivation_index, "", clock 
+  FROM accounts 
+  WHERE type != "watch" AND type != "seed" AND type != "key"
```
qfrank, line 40: i'm not sure, i thought condition would be:  WHERE type = "" or type = "seed" or type = "key" ? maybe i misunderstood something...

appdatabase/migrations/sql/1683627613_accounts_and_keycards_improvements.up.sql
```diff
@@ -0,0 +1,94 @@
...
+  GROUP BY key_uid;
+
+INSERT INTO keypairs 
+  SELECT key_uid, keypair_name, type, derived_from, last_used_derivation_index, "", clock 
+  FROM accounts 
+  WHERE type != "watch" AND type != "" AND type != "generated"
```
qfrank, line 46: oh, just saw another fulfilling, so maybe these 2 inserts could combine into one and use my above condition?

qfrank (commented):
reviewed part of this PR, i'll continue later .. when creating PR, i think it's good to be as small as possible so that it could be easily reviewed, i saw this PR trying to:
- introduce a new term "keypair"
- sync
- backup
so we may could split it into 3 smaller ones... 🙂

multiaccounts/accounts/database.go
saledjenic: Updated, added `break` as it suits more here.

api/backend_test.go
saledjenic: Yes, it refers to chat account, check the path.

api/backend_test.go
saledjenic: Yes, we cannot use that old Status folder any more.

In that version `0.97.3` key_uid was not a part of `accounts_table` and we're unable to do a migration properly with new migration sql script added to this PR. There is no way to get a `key_uid` from anywhere for already added accounts.

api/geth_backend.go
saledjenic: I haven't touched this, don't know why it's rendered like this by github diff. The only change I did here is that I added this line `keypair, err := accountDB.GetKeypairByKeyUID(account.KeyUID)` instead of the old line `relatedAccounts, err := accountDB.GetAccountsByKeyUID(account.KeyUID)`. The rest remained as it was before.

saledjenic: Explained above.

api/geth_backend.go
saledjenic: Explained above.

api/geth_backend.go
saledjenic: Isn't already too much for this PR? Thought to add this to another one, cause that change will require more changes here and as well more changes on the client side.

appdatabase/migrations/sql/1683627613_accounts_and_keycards_improvements.up.sql
saledjenic: If you have a paired device and new keypair gets synced from it, we need to add a label saying which device that keypair was synced from.

appdatabase/migrations/sql/1683627613_accounts_and_keycards_improvements.up.sql
saledjenic: Yes, there will be only a single keypair marke as "profile" and there may be 0 to N keypairs of type "seed" or "key".

appdatabase/migrations/sql/1683627613_accounts_and_keycards_improvements.up.sql
saledjenic: I agree that this column in the db can store anything (since it's varchar), but on the code side there is `AccountOperable` type defined which has 3 predefined values and that type is used in the code, which is descriptive enough for usage. Adding those 3 columns I see as redundant and unnecessary thing in this case.

appdatabase/migrations/sql/1683627613_accounts_and_keycards_improvements.up.sql
saledjenic: This one inserts all accounts referring to a profile keypair.

appdatabase/migrations/sql/1683627613_accounts_and_keycards_improvements.up.sql
saledjenic: This one inserts all accounts referring to a seed or private key keypairs.

saledjenic:
> i  think it's good to be as small as possible so that it could be easily reviewed

I agree, that's why I tried to split the work in 3 commits, but I am not sure how I could made a smaller PRs cause accounts are used everywhere and the whole app relies on them, they are used in many places in code in almost all tests and so on. If you look closer real changes are in 2 db related files and 2 messenger files, other changes are caused by those changes made in those few files.

Thanks for the review!

api/backend_test.go
qfrank: does it mean we introduced a break change in this PR ?

appdatabase/migrations/sql/1683627613_accounts_and_keycards_improvements.up.sql
saledjenic: > emm.. wouldn't it be much clear to use WHERE type = "" or type = "generated" or type = "seed" or type = "key"  instead of not condition?

No we cannot. In the old `accounts` table, accounts which were referring to "profile" had type either `""` (empty) in case of default wallet account and chat account or `generated` in case of accounts generated from the profile (master) keypair. In new table profile keypair `type` is set to `profile` and that's why we have the first insertion like this ` SELECT key_uid, keypair_name, "profile", derived_from,...` so `"profile"` is something we don't have where to get it from.

In second insertion, since we use `seed` for seed imported keypairs and `key` for private key imported keypairs, which are the same values as "old" account's type, we are able to use this sql `SELECT key_uid, keypair_name, type, derived_from,...` cause it refers to correct type values.

Hence that's where the conditions in both cases come from.

qfrank: the records for type="seed" and type="key" shouldn't be treated as 2 separate keypairs? because they may have different master key, note: you used group by key_uid here, so type="seed" or type="key" will combined into one record, is this you pretend?

api/backend_test.go
saledjenic: Well, is it a break change or not... I would say it is not, since it doesn't break anything. It refers to locally stored Status folder which is used for tests. At that point when we were using `0.97.3` version, `key_uid` column was not part of `accounts` table as well as many other columns. Check the `accounts` table at the location `/status-go/static/test-0.97.3-account/0x7c46c8f6f059ab72d524f2a6d356904db30bb0392636172ab3929a6bd2220f84.db` and you will see how it looked at that time. Was no chance to do a proper migration without having `key_uid` for each account.

qfrank: sorry, when i said it's a break change, i mean it would be a break change for earlier mobile user?

appdatabase/migrations/sql/1683627613_accounts_and_keycards_improvements.up.sql
saledjenic: > the records for type="seed" and type="key" shouldn't be treated as 2 separate keypairs?

How do you mean "shouldn't"?
They are different and they refer to the way how a user imported new keypair, via seed phrase or via private key. 

If you have seed phrase and private key for the same master key, meaning that they both generate the same `keyUid`, you will be able to add that keypair only once either via seed phrase or private key (or you can add it via seed phrase then delete it and add the same keypair via private key), but you definitely cannot add the same master key twice, once via seed phrase once via private key.

api/backend_test.go
saledjenic: > sorry, when i said it's a break change, i mean it would be a break change for earlier mobile user?

Yes, that's true. So if user has a very old local `Status` folder and hasn't run the app since that version (which was present in May 2022) this will break. Though that should be very unlikely, but even if there are such users we shouldn't worry too much about them we're still under the development, that will be more dangerous after mvp.

appdatabase/migrations/sql/1683627613_accounts_and_keycards_improvements.up.sql
qfrank: let's consider following user cases:
after logined status mobile app, user add a wallet account(import from seed) A using seed phrase "a b c..." , then add a wallet account(import from seed) B using seed phrase "d e f ...", then add wallet account(import from pk) C using private key "ghi", then add wallet account(import from pk) D using private key "jkl", what/how many records we would expect in table keypairs ?

saledjenic: So after user logged in:
- that means there is already a `profile` keypair added to `keypairs` table (this is first keypair in `keypairs` table)
- `user add a wallet account(import from seed) A using seed phrase "a b c..."` this is second keypair in `keypairs` table (but the first one with type `seed`) and will have `keyUid` = `abc123`
- `then add a wallet account(import from seed) B using seed phrase "d e f ..."` this is third keypair in `keypairs` table (but the second one with type `seed`) and will have `keyUid` = `def123`
- `then add wallet account(import from pk) C using private key "ghi"` this will be fourth keypair in `keypairs` table (but the first one with type `key`) and will have `keyUid` = `ghi123`
- `then add wallet account(import from pk) D using private key "jkl"` this will be fifth keypair in `keypairs` table (but the second one with type `key`) and will have `keyUid` = `jkl123`

> what/how many records we would expect for table keypairs ?

`keypairs` table will have 5 rows

qfrank: I see, thank you for your explain, surprise for me 😂 . I thought we still use the same keyUid for different wallet account as used to be ... and one keyUid meant one login account

saledjenic: All accounts derived from the same seed phrase share the same `keyUid`.

multiaccounts/accounts/database.go
```diff
@@ -18,44 +17,82 @@ import (
...
+	zeroAddress                         = "0x0000000000000000000000000000000000000000"
+	SyncedFromBackup                    = "backup"     // means a account is coming from backed up data
+	SyncedFromRecoveringFromOtherDevice = "recovering" // means a account is coming from another device when user is reocovering Status account
 )
 
+var (
+	errDbTransactionIsNil             = errors.New("accounts: database transaction is nil")
+	ErrDbKeypairNotFound              = errors.New("accounts: keypair is not found")
+	ErrDbAccountNotFound              = errors.New("accounts: account is not found")
+	ErrKeypairWithoutKeyUIDOrAccounts = errors.New("cannot store keypair without accounts or without set the same key uid for its accounts")
```
qfrank: split it into 2 Error type sounds more reasonable for me 🙂

protocol/messenger.go
```diff
@@ -2469,11 +2465,33 @@ func (m *Messenger) SyncDevices(ctx context.Context, ensName, photoPath string,
...
 
-	err = m.syncWallets(accounts, rawMessageHandler)
+	keypairs, err := m.settings.GetKeypairs()
+	if err != nil {
+		return err
+	}
+
+	for _, kp := range keypairs {
+		if recoveringFromAnotherDevice {
+			kp.SyncedFrom = accounts.SyncedFromRecoveringFromOtherDevice
```
qfrank: what about renaming SyncedFromRecoveringFromOtherDevice to SyncedFromLocalPair ?

protocol/messenger_config.go
```diff
@@ -57,8 +57,9 @@ type MessengerSignalsHandler interface {
 	SendWakuFetchingBackupProgress(response *wakusync.WakuBackedUpDataResponse)
 	SendWakuBackedUpProfile(response *wakusync.WakuBackedUpDataResponse)
 	SendWakuBackedUpSettings(response *wakusync.WakuBackedUpDataResponse)
-	SendWakuBackedUpWalletAccount(response *wakusync.WakuBackedUpDataResponse)
+	SendWakuBackedUpKeypair(response *wakusync.WakuBackedUpDataResponse)
 	SendWakuBackedUpKeycards(response *wakusync.WakuBackedUpDataResponse)
+	SendWakuBackedUpWoAccount(response *wakusync.WakuBackedUpDataResponse)
```
qfrank: Can we rename `SendWakuBackedUpWoAccount` to `SendWakuBackedUpWatchAccount` ? When i first saw `Wo`  i start guessing what's this ?

server/pairing/payload_management.go
```diff
@@ -130,8 +130,8 @@ func (rmm *RawMessagePayloadMarshaller) MarshalProtobuf() (data []byte, err erro
 	syncRawMessage := new(protobuf.SyncRawMessage)
 
 	syncRawMessage.RawMessages = rmm.payload.rawMessages
-	if len(rmm.payload.subAccounts) > 0 {
-		syncRawMessage.SubAccountsJsonBytes, err = json.Marshal(rmm.payload.subAccounts)
+	if rmm.payload.profileKeypair != nil && len(rmm.payload.profileKeypair.KeyUID) > 0 {
+		syncRawMessage.SubAccountsJsonBytes, err = json.Marshal(rmm.payload.profileKeypair)
```
qfrank, line 134: rename SubAccountsJsonBytes to ProfileKeypairJsonBytes ?

signal/events_sync_from_waku.go
```diff
@@ -31,10 +34,14 @@ func SendWakuBackedUpSettings(obj json.Marshaler) {
...
-	send(EventWakuBackedUpWalletAccount, obj)
+func SendWakuBackedUpKeypair(obj json.Marshaler) {
+	send(EventWakuBackedUpKeypair, obj)
 }
 
 func SendWakuBackedUpKeycards(obj json.Marshaler) {
 	send(EventWakuBackedUpKeycards, obj)
 }
+
+func SendWakuBackedUpWoAccount(obj json.Marshaler) {
```
qfrank: pls rename `SendWakuBackedUpWoAccount` to `SendWakuBackedUpWatchAccount` 🙏

saledjenic:
@qfrank all your comments are addressed and PR rebased.

saledjenic:
@cammellos do you prefer mobile team to apply necessary changes before or after this PR merge?

server/pairing/payload_management.go
saledjenic: Answered in this https://github.com/status-im/status-go/pull/3475#discussion_r1198659377, prefer to handle that in a new PR, too much for this one cause that change will require many other changes, as well as client change.

appdatabase/migrations/sql/1683627613_accounts_and_keycards_improvements.up.sql
IvanBelyakoff: Agree with @qfrank actually. Imho, I'd prefer to have multiple flags/properties as columns in DB which clearly describe the state of accounts, as it would be more readable, flexible, extendable and you wont have to hardcode definitions and implementation of operable in comment. I think a variable operable will tell other decs better of the intention, which can change in future. Otherwise, it might be that you will have to introduce more values for operable like "partially-but-not-fully" operable or forget to update the comment what it means after you make a change. But what you wont forget to do is to change the code. A dev, looking at the columns, could identify better what the problem is with specific account.

IvanBelyakoff: > AccountOperable

I don't think this naming is good if you have more than 2 values - true and false, because it becomes not descriptive enough. So it basically could be either a bitmask or enum-like depending on the needs

multiaccounts/accounts/database.go
```diff
@@ -18,44 +17,83 @@ import (
...
 const (
 	AccountTypeGenerated AccountType = "generated"
 	AccountTypeKey       AccountType = "key"
 	AccountTypeSeed      AccountType = "seed"
 	AccountTypeWatch     AccountType = "watch"
 )
 
+const (
+	AccountNonOperable       AccountOperable = "no"        // an account is non operable it is not a keycard account and there is no keystore file for it and no keystore file for the address it is derived from
+	AccountPartiallyOperable AccountOperable = "partially" // an account is partially operable if it is not a keycard account and there is created keystore file for the address it is derived from
```
IvanBelyakoff, line 93: I will repeat myself, that "partially" operable account does not describe what receiving side can do with it. So a bitmask and renaming "AccountOperable" to "AccountFlags" or similar could do a better job.

multiaccounts/accounts/database.go
```diff
@@ -144,236 +195,394 @@ func (db *Database) Close() error {
...
-		return err
-	}
-	defer func() {
-		if err == nil {
-			err = tx.Commit()
-			return
-		}
-		_ = tx.Rollback()
-	}()
+func (db *Database) processKeyparis(rows *sql.Rows) ([]*Keypair, error) {
```
IvanBelyakoff: typo?

protocol/messenger_backup_handler.go
```diff
@@ -196,43 +199,48 @@ func (m *Messenger) handleBackedUpSettings(message *protobuf.SyncSetting) error
...
-	acc, err := m.handleSyncWalletAccount(message)
+	acc, err := m.handleSyncWalletAccount(message, accounts.SyncedFromBackup)
 	if err != nil {
 		return err
 	}
 
 	if m.config.messengerSignalsHandler != nil {
 		response := wakusync.WakuBackedUpDataResponse{
-			WalletAccount: acc,
+			WOnlyAccount: acc,
```
IvanBelyakoff: Not sure if you named it, but if possible, I'd avoid such abbreviations as `WOnly` for WatchOnly, for structs and method names.

IvanBelyakoff (approved):
Great job Sale! It would be nice if you addressed the issues I mentioned.
Though I don't want to block you, as they are minor

alaibe (approved):
Approving, I didn't find anything else than QFrank or Ivan Found. So assuming you are going to deal with those, all good!

saledjenic:
@IvanBelyakoff thanks for the review. I addressed your last 2 comments. 

As of account operability, it can be said that it is more about what a developer prefers more.

In this case in each moment an account can be in only one of `fully`, `partially` or `no` operable state, No way to have an account at the same time in 2 or more states and also no chance that any new state may appear in future then those 3. That's the reason why I opted for storing the value in a single column.

> I will repeat myself, that "partially" operable account does not describe what receiving side can do with it

If it's more clear we can rename them to have keystore-file as a keyword in option names, but keycard accounts doesn't use keystore file, that's why I used operable, cause it describes its operability within the app, is account fully operable, partially operable or not operable at all.

> So a bitmask and renaming "AccountOperable" to "AccountFlags" or similar could do a better job.

Bitmask is something I would use if an account can be at to or more state at the same time, in that case bitmast has its purpose.

qfrank (approved):

IvanBelyakoff:
> keystore

It is up to you of course. Just the operability must be determined by code, not by comments. you store the raw data and then determine whether this is operable/non-operable state when you load it. If in future you decide (well, saying that it will never happen is not that reliable) that operable state must have a different meaning, all you do - you fix the code, but you dont have to fix the values in DB, migrate it or fix the comments, how you determine operability. So bitmask here is not about being in multiple states at the same time, but having multiple problems, why this acc is not operable. Anyone can open the DB and see, what is set and what is not for the account. Because `partially` operable says nothing about what you really can do with the acc and why it is not operable.
I am not insisting on fixing it.
Overall great work!

saledjenic:
@cammellos @qfrank I logged na issue for mobile team: https://github.com/status-im/status-go/issues/3531 (if you already have it in some other repo, feel free to close it).

PR #3476, opened 2023-05-11
Feat/sync local deleted message

qfrank:
sync `user_messages_deleted_for_mes` when `local pair` or `pair all devices` or user delete message locally.

status: ready.


protocol/message_persistence.go
```diff
@@ -2383,40 +2383,58 @@ func (db sqlitePersistence) GetDeletes(messageID string, from string) ([]*Delete
...
 		if err != nil {
 			return nil, err
 		}
 		messages = append(messages, d)
+	}
+	return messages, nil
+}
 
+func (db sqlitePersistence) GetDeleteForMes() ([]*protobuf.DeleteForMeMessage, error) {
+	rows, err := db.db.Query(`SELECT clock, message_id FROM user_messages_deleted_for_mes`)
```
ilmotta, line 2425: Maybe there's a linter somewhere that could check for us missing `Close()` calls, and the missing check for errors after iterating over `rows`.

But since we don't have this, we should remember to add `defer rows.Close()`.

After iterating with `rows.Next()`, it's important to also check for errors. Here's the example https://go.dev/play/p/MZeisUY9S6V and doc https://pkg.go.dev/database/sql#example-Rows

protocol/message_persistence.go
```diff
@@ -2383,40 +2383,58 @@ func (db sqlitePersistence) GetDeletes(messageID string, from string) ([]*Delete
...
+		err := rows.Scan(&d.Clock, &d.MessageId)
 		if err != nil {
 			return nil, err
 		}
 		messages = append(messages, d)
+	}
+	return messages, nil
+}
 
+func (db sqlitePersistence) GetDeleteForMes() ([]*protobuf.DeleteForMeMessage, error) {
```
ilmotta: I was going to suggest using the suffix `Message`, like what we use in `applyDeleteForMeMessage` or `HandleDeleteForMeMessage`, etc to be more consistent. But in fact, I just learned that `mes` can actually be considered a valid plural form of `me`.

protocol/message_persistence.go
```diff
@@ -2383,40 +2383,58 @@ func (db sqlitePersistence) GetDeletes(messageID string, from string) ([]*Delete
...
+	_, err := db.db.Exec(`INSERT OR REPLACE INTO user_messages_deleted_for_mes (clock, message_id) 
+    SELECT ?,? WHERE NOT EXISTS (SELECT 1 FROM user_messages_deleted_for_mes WHERE id = ? AND clock >= ?)`,
+		deleteForMeMessage.Clock, deleteForMeMessage.MessageId, deleteForMeMessage.MessageId, deleteForMeMessage.Clock)
 	return err
 }
 
-func (db sqlitePersistence) GetDeleteForMes(messageID string, from string) ([]*DeleteForMeMessage, error) {
-	rows, err := db.db.Query(`SELECT clock, message_id, source, id FROM user_messages_deleted_for_mes WHERE message_id = ? AND source = ? ORDER BY CLOCK DESC`, messageID, from)
+func (db sqlitePersistence) GetDeleteForMesByMessageID(messageID string) ([]*protobuf.DeleteForMeMessage, error) {
+	rows, err := db.db.Query(`SELECT clock, message_id FROM user_messages_deleted_for_mes WHERE message_id = ?`, messageID)
```
ilmotta, line 2406: Same as my other comment, ideally we should close `rows` and check for errors after iterating.

ilmotta (approved):

ilmotta:
@qfrank, would it be possible to cover the code changes with some tests?

protocol/message_persistence.go
qfrank: damn, i was too trust the old code and ignored this 😅 👍 , you know when the first guy did wrong, the latter... like a funny game...

qfrank:
> @qfrank, would it be possible to cover the code changes with some tests?

yes, thank you for your review! 🫶

protocol/message_persistence.go
```diff
@@ -2381,42 +2381,61 @@ func (db sqlitePersistence) GetDeletes(messageID string, from string) ([]*Delete
...
-
 	}
 	return messages, nil
 }
 
-func (db sqlitePersistence) SaveDeleteForMe(deleteForMeMessage DeleteForMeMessage) error {
-	_, err := db.db.Exec(`INSERT INTO user_messages_deleted_for_mes (clock, message_id, source, id) VALUES(?,?,?,?)`, deleteForMeMessage.Clock, deleteForMeMessage.MessageId, deleteForMeMessage.From, deleteForMeMessage.ID)
+func (db sqlitePersistence) SaveOrUpdateDeleteForMeMessage(deleteForMeMessage *protobuf.DeleteForMeMessage) error {
+	_, err := db.db.Exec(`INSERT OR REPLACE INTO user_messages_deleted_for_mes (clock, message_id) 
+    SELECT ?,? WHERE NOT EXISTS (SELECT 1 FROM user_messages_deleted_for_mes WHERE message_id = ? AND clock >= ?)`,
```
saledjenic, line 2400: @qfrank have you checked this query? Looking into it I am not sure that it works.

protocol/messenger.go
```diff
@@ -6415,6 +6413,33 @@ func (m *Messenger) withChatClock(callback func(string, uint64) error) error {
 	return m.saveChat(chat)
 }
 
+func (m *Messenger) syncDeleteForMeMessage(ctx context.Context, rawMessageDispatcher RawMessageHandler) error {
+	deleteForMes, err := m.persistence.GetDeleteForMeMessages()
+	if err != nil {
+		return err
+	}
+
+	return m.withChatClock(func(chatID string, clock uint64) error {
```
saledjenic: @qfrank do we really need `clock` since it is not used anywhere?

protocol/message_persistence.go
qfrank: > have you checked this query

yes

protocol/messenger.go
qfrank: just replaced clock with _

saledjenic (approved):

protocol/message_persistence.go
ilmotta: Exactly @qfrank! And at least to me there are so many linter stacked in Go, that I once thought they would warn us about these things, but apparently they don't/can't.

PR #3477, opened 2023-05-11
fix(messenger_handler): fix checkForDeletes logic with images

jrainville:
Merges in a release branch for desktop

Fixes https://github.com/status-im/status-desktop/issues/10627

I messed up when I did the code to checkForDeletes for image albums.

The logic was all tangled up and we were trying to delete even if nothing was a delete. So, well, I fixed it.

### DIFF
--- a/protocol/messenger_delete_message_test.go
+++ b/protocol/messenger_delete_message_test.go
@@ -329,11 +329,12 @@ func (s *MessengerDeleteMessageSuite) TestDeleteImageMessageFirstThenMessage() {
 	s.Require().NoError(err)
 	messageID1 := "message-id1"
 	messageID2 := "message-id2"
+	albumID := "album-id1"
 
 	messageCount := 2
 	var album []*common.Message
 	for i := 0; i < messageCount; i++ {
-		image, err := buildImageWithoutAlbumIDMessage(*ourChat)
+		image, err := buildImageWithAlbumIDMessage(*ourChat, albumID)
 		image.Clock = 1
 		s.NoError(err)
 		album = append(album, image)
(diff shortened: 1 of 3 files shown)


ilmotta (approved):
Fix LGTM. I'm only missing seeing updated unit tests to cover the correct behavior.

igor-sirotin (approved):

jrainville:
> Fix LGTM. I'm only missing seeing updated unit tests to cover the correct behavior.

I tried and I can't reproduce the bug in a test :grimacing: 

The issue is such a weird edge case that if I revert the fix and try to make the test fail, it still passes :confused: 

I think I need to send an image with no Album, then delete it, but also have the delete arrive before the image, then try to send another image without an album and that will make it fail to process the new image.
So, I tried doing that in the test and it still passes, so I'm not sure. 

At least we know that the fixed code works in tests and in an in-app scenario :shrug:

ilmotta:
> > Fix LGTM. I'm only missing seeing updated unit tests to cover the correct behavior.
> 
> I tried and I can't reproduce the bug in a test grimacing
> 
> The issue is such a weird edge case that if I revert the fix and try to make the test fail, it still passes confused
> 
> I think I need to send an image with no Album, then delete it, but also have the delete arrive before the image, then try to send another image without an album and that will make it fail to process the new image. So, I tried doing that in the test and it still passes, so I'm not sure.
> 
> At least we know that the fixed code works in tests and in an in-app scenario shrug

Ouch, you really did try, so there's not much to do :shrug: Thanks for the detailed context

jrainville:
Re-asking for review because it turns out the old test was flawed and didn't even have a valid albumID.

So basically, the issue we faced in Desktop might have been caught by the test if the test actually did it's job correctly

ilmotta (approved):

borismelnik (approved):

PR #3478, opened 2023-05-11
feat(wallet): complete filter api

stefandunca:
### Closes status-desktop [#10634](https://github.com/status-im/status-desktop/issues/10634)

It uses the current data only and doesn't extend with new types or
include new features in activity sources DBs.

Major changes:
- Partially filter by chain IDs
- Partially filter by Status if it is the case
- Partially filter by token types
- Filter by counterparty addresses
- Use accounts to detect transaction origin and type (TO/FROM). Use accounts table when all accounts filter is requested

### DIFF
--- a/services/wallet/activity/activity.go
+++ b/services/wallet/activity/activity.go
@@ -9,12 +9,19 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/ethereum/go-ethereum/common"
+	eth "github.com/ethereum/go-ethereum/common"
+
+	"github.com/status-im/status-go/services/wallet/common"
 	"github.com/status-im/status-go/services/wallet/transfer"
+
+	"golang.org/x/exp/constraints"
 )
 
 type PayloadType = int
 
+// Beware if adding/removing please check if affected and update the functions below
+// - NewActivityEntryWithTransaction
+// - multiTransactionTypeToActivityType
 const (
 	MultiTransactionPT PayloadType = iota + 1
 	SimpleTransactionPT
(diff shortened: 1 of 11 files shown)


services/wallet/activity/activity.go
```diff
@@ -28,6 +32,8 @@ type Entry struct {
 	id              transfer.MultiTransactionIDType
```
IvanBelyakoff: I see that this PR closes the issue, do you plan to have more to fix TODOs like above for payloadType or is it a good time to do it in this PR?

services/wallet/activity/activity.go
```diff
@@ -215,41 +326,87 @@ const queryFormatString = `
...
 
 	mtTypes := activityTypesToMultiTransactionTypes(filter.Types)
-	joinedMTTypes := joinMTTypes(mtTypes)
+	joinedMTTypes := joinItems(mtTypes, func(t transfer.MultiTransactionType) string {
+		return strconv.Itoa(int(t))
+	})
 
-	queryString := fmt.Sprintf(queryFormatString, joinedAddresses, joinedMTTypes)
+	queryString := fmt.Sprintf(queryFormatString, involvedAddresses, toAddresses, assets, networks,
+		joinedMTTypes)
```
IvanBelyakoff, line 408: Since query here is quite complex, as a future improvement, it would be nice to have implementation like in `services/wallet/transfer/query.go`

IvanBelyakoff (approved):
Great job! LGTM

services/wallet/activity/activity.go
stefandunca: I'll be sure to fix it here. Thanks, I forgot that one.

services/wallet/activity/activity.go
stefandunca: I agree. However, I will first consider a unification and refactoring of the involved DBs as proposed here: https://github.com/status-im/status-desktop/issues/10366

PR #3480, opened 2023-05-12
feat(wallet): Implemented SequentialFetchStrategy for transfers.

IvanBelyakoff:
By default OnDemandFetchStategy is still used.

I attempted to keep all the default behavior for OnDemandFetchStrategy to avoid breaking it while
the new scheduler is not fully tested and polished.
A separate `findBlocksCommand` has been created which load blocks from nodes per account basis. Benefits:

- downloading history separately for all accounts and avoid blocking the next iteration for account which has less transfers by fetching blocks for an account with more transfers waited in a single wait group.
- fetching transfers may be prioritized for own accounts compared to the watched ones
- simplifies code

Fetching blocks is done by chunks (currently constant 10 000 blocks), but a possible improvement is to dynamically detect a proper chunk size to reduce a number of calls.

A separate table is used for blocks info for SequentialFetchStrategy to avoid messing with the old method using `block_ranges` table, as having two different implementations updating the same table and keeping it logically consistent is troublesome.

Not covered: pending transaction and fetching new transactions. Tests TBD.

Updates #10246

### DIFF
--- a/services/wallet/transfer/block_ranges_sequential_dao.go
+++ b/services/wallet/transfer/block_ranges_sequential_dao.go
@@ -1,0 +1,146 @@
+package transfer
+
+import (
+	"database/sql"
+	"fmt"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/status-im/status-go/services/wallet/bigint"
+)
+
+const (
+	firstBlockColumn = "blk_first"
+	lastBlockColumn  = "blk_last"
+	startBlockColumn = "blk_start"
+)
+
+type BlockRangeSequentialDAO struct {
+	db *sql.DB
+}
+
+type BlockRange struct {
+	Start      *big.Int // Block of first transfer
+	FirstKnown *big.Int // Oldest scanned block
+	LastKnown  *big.Int // Last scanned block
+}
+
+func NewBlockRange() *BlockRange {
+	return &BlockRange{Start: &big.Int{}, FirstKnown: &big.Int{}, LastKnown: &big.Int{}}
+}
+
+func (b *BlockRangeSequentialDAO) getBlockRange(chainID uint64, address common.Address) (blockRange *BlockRange, err error) {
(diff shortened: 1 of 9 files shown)


appdatabase/migrations/sql/1683457503_blocks_ranges_sequential_scheduler.up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS blocks_ranges_sequential (
+    network_id UNSIGNED BIGINT NOT NULL,
+    address VARCHAR NOT NULL,
+    blk_start BIGINT,
```
stefandunca: Wondering when is the `blk_start` value `NULL` case encountered.

services/wallet/transfer/block_ranges_sequential_dao.go
```diff
@@ -0,0 +1,146 @@
...
+			return nil, err
+		}
+
+		return blockRange, nil
+	}
+
+	return nil, nil
+}
+
+// TODO call it when account is removed
```
stefandunca, line 57: Do we have any follow-up task to link with this TODO and ensure we don't leave dead data in the table?

stefandunca (approved):
The code looks good. Some notes are inline.

appdatabase/migrations/sql/1683457503_blocks_ranges_sequential_scheduler.up.sql
IvanBelyakoff: The approach is to start transfers checking from the latest block until we find a block range where balance is 0, it did not change and nonce is 0, which means a beginning of activity for this account, then we set the beginning of that range as `start_blk` and we will not check further back. `start_blk` will be NULL when we have not found it yet, so when scheduler starts, it checks if `blk_first` (the earliest checked block) > `start_blk` and if so, starts further backwards fetching

services/wallet/transfer/block_ranges_sequential_dao.go
IvanBelyakoff: good point. I was going to work on it after this PR, I will create a task that covers it today

rasom (approved):

PR #3484, opened 2023-05-15
fix: Suggested list of mentions does not appear initially when '@' is typed in the editing message composer

qfrank:
Closes mobile issue [#15890](https://github.com/status-im/status-mobile/issues/15890)


ibrkhalil (approved):
Great work, Thanks @qfrank

ilmotta (approved):

PR #3485, opened 2023-05-15
show mention notification/AC once for a message with multiple images

ibrkhalil:
When a community user send a mention with images, The receiver got as many AC notifications as the number of sent images.
This PR uses the `albumId` as the unique id for the notification record.

Important changes:
- [ ] Need to test the same issue on status-desktop too.

Closes #[15384](https://github.com/status-im/status-mobile/issues/15834) on status-mobile


protocol/messenger.go
```diff
@@ -3231,6 +3231,39 @@ func (r *ReceivedMessageState) addNewActivityCenterNotification(publicKey ecdsa.
 	return nil
 }
 
+func (r *ReceivedMessageState) addNewActivityCenterNotificationV2(publicKey ecdsa.PublicKey, m *Messenger, message *common.Message, responseTo *common.Message, albumId string) error {
```
ilmotta: @ibrkhalil, could you explain the idea behind this V2 naming? Why can't we change the original function and have only the new behavior?

protocol/messenger.go
```diff
@@ -4595,19 +4628,26 @@ func (m *Messenger) saveDataAndPrepareResponse(messageState *ReceivedMessageStat
 
 	for _, message := range messageState.Response.messages {
 		if _, ok := newMessagesIds[message.ID]; ok {
+
 			message.New = true
 
+			if message.ContentType == protobuf.ChatMessage_IMAGE {
```
ilmotta: Just a reminder to add/update tests.

protocol/messenger.go
ibrkhalil: You're totally right, I am going to extend the original function to use the correct var instead of useless overloading.

ibrkhalil: You're totally right, I am going to extend the original function to use the correct var instead of useless overloading.

protocol/messenger.go
ibrkhalil: Completely forgot, Thanks for the reminder!

protocol/messenger.go
ibrkhalil: Used a different way to do the same thing, Kindly take a look.

protocol/messenger.go
ibrkhalil: Tried writing tests, But I faced a problem that the message content type when sent to a community has a type of `protobuf.ChatMessage_COMMUNITY` and doesn't allow it to be converted as a message as I presume, Because calling `GetImage` on it returns nil even though it has a message.

cammellos: mmh, pretty sure you can write a test for it, you don't even have to use a community for testing this feature, you can test it with a one to one chat

ibrkhalil: Will try it with normal chat but I wanted to test with a community chat because the original issue reported was in a community, Thanks @cammellos.

Gonna check how it acts on nightly first.

protocol/messenger.go
```diff
@@ -3208,10 +3208,20 @@ func (r *ReceivedMessageState) addNewActivityCenterNotification(publicKey ecdsa.
 		return fmt.Errorf("chat ID '%s' not present", message.LocalChatID)
 	}
 
+	// Use albumId as notificationId to prevent multiple notifications
+	// for same message with multiple images
+	var idToUse string
+	
+	if message.GetImage() != nil {
+		idToUse = message.GetImage().GetAlbumId()
```
ilmotta, line 3149: @ibrkhalil, I'm thinking now that the notification ID is linked to the album ID, what if all images in the message are removed and there is no album ID anymore? I believe this is possible, I'm just not totally sure.

Hey @borismelnik, do you know if what I mentioned above is a valid concern?

borismelnik: @ilmotta I think usage `AlbumId` as notification ID is a good idea, cos the same concept we use for messages (MessageID). But need to process deletion as well. The same I did for deleted message, I think need to check it, if deleted, message was with AlbumId and was not last then need to skip sending empty notification :)

borismelnik: @ibrkhalil but if last message in Album was deleted, need to send empty notification with correct id (in this context its an AlbumID)

borismelnik: [there is my PR for deleting notification](https://github.com/status-im/status-go/pull/3277) may helps :)

ibrkhalil: On Status-mobile, We disable editing messages with images. [ref](https://github.com/status-im/status-mobile/blob/47a8d20940f2f9d374207a22a57c75e9f16f95ba/src/status_im2/contexts/chat/messages/drawers/view.cljs#L31-L37)
So when we enable it later this might be an issue
Just tested Deletion, Deleting an image with mentions deletes the notification associated with the mention.

protocol/messenger.go
ibrkhalil: Tested mentioning a user with images on a 1-on-1 chat, And only got an unviewed-messages indicator on the home screen above the chat where the message happened.

protocol/messenger.go
ilmotta: @borismelnik, thanks for the insights. I guess then we should process deletion as you say to play on the safe side, even if now it works in mobile. Wdyt @ibrkhalil?

ibrkhalil: Do you mean cascading the deletion from DB the notification deletion with the message deletion?

ilmotta: @ibrkhalil I was worried that since the album ID depends on the existence of images, that if all images are removed, then the notification would have an ID, but one that doesn't refer to a valid album. We just need to guarantee that deleting the album would delete the notification I guess?

pavloburykh:
@status-im/desktop-qa kindly check current changes on desktop as they may cause regression. Thank you.

anastasiyaig:
@MishkaRogachev since it is changing contact requests , please review

MishkaRogachev (approved):
It's okay from my point of view, just question from Icaro should be resolved

protocol/messenger.go
ibrkhalil: > @ibrkhalil I was worried that since the album ID depends on the existence of images, that if all images are removed, then the notification would have an ID, but one that doesn't refer to a valid album. We just need to guarantee that deleting the album would delete the notification I guess?

Tested deleting the album and the notification vanished

ilmotta (approved):

PR #3491, opened 2023-05-15
fix(unviewed): fix edge case where unviewed count would be -1

jrainville:
Fixes https://github.com/status-im/status-desktop/issues/10649

I'm not sure exactly how it's possible, but somehow, some people got -1 as `UnviewedMessagesCount` in the DB. It's supposed to be a uint, so I don't get how it's possible, but I added guards to make sure it doesn't happen again.

### DIFF
--- a/protocol/message_persistence.go
+++ b/protocol/message_persistence.go
@@ -1726,14 +1726,8 @@ func (db sqlitePersistence) MarkAllRead(chatID string, clock uint64) (int64, int
 
 	_, err = tx.Exec(
 		`UPDATE chats
-		   SET unviewed_message_count =
-		   (SELECT COUNT(1)
-		   FROM user_messages
-		   WHERE local_chat_id = ? AND seen = 0),
-		   unviewed_mentions_count =
-		   (SELECT COUNT(1)
-		   FROM user_messages
-		   WHERE local_chat_id = ? AND seen = 0 AND (mentioned or replied)),
+		   SET unviewed_message_count = 0,
+		   unviewed_mentions_count = 0,
                    highlight = 0
 		WHERE id = ?`, chatID, chatID, chatID)
 
--- a/protocol/messenger.go
+++ b/protocol/messenger.go
@@ -4859,7 +4859,7 @@ func (m *Messenger) markAllRead(chatID string, clock uint64, shouldBeSynced bool
 		return errors.New("chat not found")
 	}
 
-	seen, mentioned, err := m.persistence.MarkAllRead(chatID, clock)
+	_, _, err := m.persistence.MarkAllRead(chatID, clock)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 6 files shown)


protocol/message_persistence.go
```diff
@@ -1726,14 +1726,8 @@ func (db sqlitePersistence) MarkAllRead(chatID string, clock uint64) (int64, int
...
 	_, err = tx.Exec(
 		`UPDATE chats
-		   SET unviewed_message_count =
-		   (SELECT COUNT(1)
-		   FROM user_messages
-		   WHERE local_chat_id = ? AND seen = 0),
-		   unviewed_mentions_count =
-		   (SELECT COUNT(1)
-		   FROM user_messages
-		   WHERE local_chat_id = ? AND seen = 0 AND (mentioned or replied)),
```
jrainville, line 1736: That whole block of code was pretty useless, since we're setting `seen=1` to all messages, so of course it would return COUNT -> 0.

Let me know if there is something i'm missing, but the intended behaviour is to set all messages as read, so the `unviewed_count` should be set to 0 by default.

We can save some time that way.

protocol/messenger.go
```diff
@@ -4874,17 +4874,8 @@ func (m *Messenger) markAllRead(chatID string, clock uint64, shouldBeSynced bool
...
-		chat.UnviewedMessagesCount = 0
-	}
-
-	if chat.UnviewedMentionsCount >= uint(mentioned) {
-		chat.UnviewedMentionsCount -= uint(mentioned)
-	} else {
-		chat.UnviewedMentionsCount = 0
-	}
+	chat.UnviewedMessagesCount = 0
+	chat.UnviewedMentionsCount = 0
```
jrainville, line 4878: Again, the unviewed count is supposed to be 0 whatever whatever the previous count was.

protocol/persistence.go
```diff
@@ -498,6 +501,18 @@ func (db sqlitePersistence) Chat(chatID string) (*Chat, error) {
...
+		// Set UnviewedCounts and make sure they are above 0
+		// Since Chat's UnviewedMessagesCount is uint and the SQL column is INT, it can create a discrepancy
+		if unviewedMessagesCount < 0 {
+			unviewedMessagesCount = 0
+		}
+		if unviewedMentionsCount < 0 {
+			unviewedMentionsCount = 0
+		}
+		chat.UnviewedMessagesCount = uint(unviewedMessagesCount)
+		chat.UnviewedMentionsCount = uint(unviewedMentionsCount)
```
jrainville, line 514: In theory, with the new validations and guards I added, those lines wouldn't be necessary, but better safe than sorry.

It also unblocks the poor people who's accounts were borked because of the -1 in the DB

protocol/message_persistence.go
0x-r4bbit: This looks about right

protocol/persistence.go
0x-r4bbit: The checks above should indeed not be necessary, because judging by the original issue, SCAN would've failed already at that point and we wouldn't even make it into this code path.

0x-r4bbit (approved):
Changes LGTM but we can prolly remove some checks

protocol/messenger_handler.go
```diff
@@ -1632,8 +1632,10 @@ func (m *Messenger) HandleDeleteMessage(state *ReceivedMessageState, deleteMessa
 		// Reduce chat mention count and unread count if unread
 		if !messageToDelete.Seen && !unreadCountDecreased {
 			unreadCountDecreased = true
-			chat.UnviewedMessagesCount--
-			if messageToDelete.Mentioned || messageToDelete.Replied {
+			if chat.UnviewedMessagesCount > 0 {
+				chat.UnviewedMessagesCount--
```
ilmotta, line 1636: Even though I don't understand how the counter became negative, this change here is an important practice nonetheless, because if the uint is subtracted from zero it underflows to `18446744073709551615`, so a guard expression should always be used.

ilmotta (approved):

protocol/messenger_handler.go
ilmotta: Could it be that the value underflowed because it was decremented by 1, but in SQLite the column was an INT and then was represented as a negative number? Suspicious.

jrainville: Yeah i don't get it. Anyway, like you said, guarding against the underflow is still important.

PR #3494, opened 2023-05-16
Add chainIds to revealed accounts in memberhsip requests

### LINKED ISSUE #10481: Optimise nb queries made by owner node when checking member tokens
As an owner node of a community that has token permissions, every membership request and every membership needs to be (regularly) checked to ensure they still own the necessary funds to be part of the community.

This includes checking ERC20 token balances and ERC721 token ownership on all supported chains.
At the moment we always check every supported chain always.

One way to improve this would be to have members tell the owner on which chains they'll find which funds. 
Based on that information, the owner node can then specifically query a subset of chains to check funds.

To make this work a few things will be needed:

1. We need to change the protocol to support chain information (e.g. requirement is 200 USDC; chainID 1 => 20 USDC, chainID 1337 => 120 USDC)
2. We need to figure out how we decide which chains will be used for that information, if the necessary funds live on multiple chains
3. We also need to decide what the "split" going to be in case the funds live on multiple chains.

re: 3): here's what I mean:

Say the requirement to request access to a community is 2000 SNT.
A user owns a wallet with 2000 SNT on Ethereum and 5000 SNT Arbitrum.

Now there are various configurations that would fulfill the requirement

=> ETH: 2000 SNT
=> ARB: 2000 SNT
=> ETH: 1000 SNT, ARB: 1000 SNT
... // all other combinations that make 2000 SNT across supported chains

^ In such a scenario, how do we decide which chain info will be attached to the membership request (and later preserved in the membership data) ?

/cc @John-44 if you have thoughts on this, let me know

0x-r4bbit:

- Adds a migration that adds chainids to `communities_request_to_join_revealed_addresses`
- Removes `RevealedAddress` in favor of `RevealedAccount` which is now a struct that contains the revealed address, as well as the signature and a list of chain IDs on which to check for user funds
- Changes the logic of sending requests to join a community, such that after creating address signatures, the user node will also check which of the addresses has funds on which networks for the community's token permissions, and add the chainds to the `RevealedAccount`
- Updates `checkPermissionToJoin()` such that only relevant chainids are used when checking user's funds. Chain IDs are retrieved from `RevealedAccount`s and matched against token permission criteria chain IDs


protocol/communities/manager.go
```diff
@@ -1501,25 +1487,90 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request
...
+
+	// if there are no chain IDs attached to any of the revealed accounts
+	// we aren't able to check balances on selected networks
+	if len(erc20ChainIDsMap) > 0 && len(revealedChainIDsForERC20) == 0 {
+		return false, nil
+	}
+
+	ownedERC20TokenBalances := make(map[string]*big.Float)
+
+	if len(revealedChainIDsForERC20) > 0 {
```
jrainville: Should we add a condition that we only get the balances if there is an ERC20 requirement (`len(erc20TokenRequirements) > 0` I guess?)

protocol/communities/manager.go
```diff
@@ -1501,25 +1487,90 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request
...
+
+		// calculate accumulated balances from chain balances
+		ownedERC20TokenBalances = m.GetAccumulatedTokenBalances(balancesByChain, erc20TokenRequirements)
+		if err != nil {
+			return false, err
+		}
+	}
+
+	ownedERC721Tokens := make(CollectiblesByChain)
+	if len(revealedChainIDsForERC721) > 0 {
```
jrainville: Same thing here but for erc721

protocol/communities/manager.go
```diff
@@ -1501,25 +1487,90 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request
...
+	}
+
+	revealedChainIDsForERC20 := getRevealedChainIDs(revealedAccounts, erc20ChainIDsMap)
+	revealedChainIDsForERC721 := getRevealedChainIDs(revealedAccounts, erc721ChainIDsMap)
+
+	// if there are no chain IDs attached to any of the revealed accounts
+	// we aren't able to check balances on selected networks
+	if len(erc20ChainIDsMap) > 0 && len(revealedChainIDsForERC20) == 0 {
+		return false, nil
+	}
```
jrainville, line 1685: Should we do the same check for erc721?

jrainville (commented):
Great work. I add 3 comments inline about small possible improements

protocol/communities/manager.go
0x-r4bbit: We already do that implicitly because `revealedChainIDsForERC20` have to exist in `erc20TokenRequirements`.
That's what we ensure in `getRevealedChainIDs()`.

If there's no chainIDs in the required permissions for ERC20 tokens, then `revealedChainIDsForERC20` will be empty

protocol/communities/manager.go
0x-r4bbit: Same applies here.

protocol/communities/manager.go
0x-r4bbit: Actually we might even **remove** this check here.

Here's why:

The chain IDs we're checking on here are based on **all** token criteria from **all** permissions.
This means, it's perfectly fine to have criteria that need erc20 token balances, but not have any erc20 balances in an account **if** there's other criteria that, say, require ERC721 or ENS tokens which do exist in any of the user's accounts.

Remember that only **one** permission's criteria has to be met to be granted access.

So if you don't have any ERC20 tokens, even though there's a criteria in any of the permissions, no big deal, cause you might fulfill the criteria of a different permission.

Hence, we actually shouldn't return here.

jrainville (approved):

0x-r4bbit:
I'm working on refactoring this.
After discussing this with @cammellos we'll be piggy backing `checkPermissionsToJoin()` when sending a request to join.

protocol/communities/manager.go
```diff
@@ -1313,7 +1317,13 @@ func (m *Manager) CheckPermissionToJoin(id []byte, addresses []gethcommon.Addres
...
 	becomeMemberPermissions := community.TokenPermissionsByType(protobuf.CommunityTokenPermission_BECOME_MEMBER)
 
-	hasPermission, err := m.checkPermissionToJoin(becomeMemberPermissions, addresses, false)
+	allChainIDs, err := m.tokenManager.GetAllChainIDs()
+	if err != nil {
+		return nil, err
+	}
+
+	accountsAndChainIDs := combineAddressesAndChainIDs(addresses, allChainIDs)
+	hasPermission, err := m.checkPermissionToJoin(becomeMemberPermissions, accountsAndChainIDs, false)
```
0x-r4bbit, line 1326: @cammellos here you can see the usage of `checkPermissionToJoin()` which now receives an accounts/chainID combination as discussed

protocol/communities/manager.go
```diff
@@ -1543,14 +1548,11 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request
...
+
 		// provided wallet addresses seem to be legit, so let's check
 		// if the necessary token permission funds exist
-		verifiedAddresses := make([]gethcommon.Address, 0)
-		for walletAddress := range request.RevealedAddresses {
-			verifiedAddresses = append(verifiedAddresses, gethcommon.HexToAddress(walletAddress))
-		}
-
-		permissionResponse, err := m.checkPermissionToJoin(becomeMemberPermissions, verifiedAddresses, true)
+		permissionResponse, err := m.checkPermissionToJoin(becomeMemberPermissions, accountsAndChainIDs, true)
```
0x-r4bbit, line 1555: @cammellos  Here's another usage where the accounts/chainIDs combination is received via the request-to-join

protocol/communities/manager.go
```diff
@@ -1585,58 +1587,124 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request
...
 			if !criteria {
-				c.Satisfied = false
+				satisfied = false
 				break
 			}
 		}
+		if satisfied {
+			c.Satisfied = true
+		}
 	}
```
0x-r4bbit, line 1622: The above here turned out to be a logical bug.
As per spec, only one permission needs to be satisfied for someone to be able to join.
This fixes the logic such that this holds true.

Prior to this change, `Satisfied` would be set to `false` if just one token criteria from any permission wasn't.

protocol/communities/manager.go
```diff
@@ -1793,50 +1919,23 @@ func (m *Manager) getOwnedERC721Tokens(walletAddresses []gethcommon.Address, tok
...
-				ownedERC721Tokens[chainID][asset.Contract.Address] = append(ownedERC721Tokens[chainID][asset.Contract.Address], asset)
+			if len(assets.Assets) == 0 {
+				continue
 			}
-		}
-	}
-	return ownedERC721Tokens, nil
-}
-
-func (m *Manager) getAccumulatedTokenBalances(accounts []gethcommon.Address, tokenRequirements map[uint64]map[string]*protobuf.TokenCriteria) (map[string]*big.Float, error) {
```
0x-r4bbit, line 1807: This function is no longer used as we now calculate accumulated balances on the fly

0x-r4bbit:
@cammellos  This needs some tests, but the basic idea is there

mprakhov (approved):

jrainville (approved):
Nice work!

PR #3495, opened 2023-05-16
fix: create a CR on contact sync with received CR state

MishkaRogachev:
For https://github.com/status-im/status-desktop/issues/10032

Important changes:
- [x] Create incoming CR message and notification on received state durning a contact synchronisation
- [x] Create outgoing CR message and notification on sent state durning a contact synchronisation

### DIFF
--- a/protocol/messenger_contact_requests_test.go
+++ b/protocol/messenger_contact_requests_test.go
@@ -1150,9 +1150,9 @@ func (s *MessengerContactRequestSuite) TestBobRestoresIncomingContactRequestFrom
 }
 
 // The scenario tested is as follow:
-// 1) Alice sends a contact request to Alice
+// 1) Alice sends a contact request to Bob
 // 2) Bob receives CR from Alice
-// 3) Alice resets his device
+// 3) Alice resets her device
 // 4) Alice restores Bob's contact from backup, CR is created
 // 5) Bob accepts contact request
 // 6) Alice get notified properly
--- a/protocol/messenger_contact_requests_test.go
+++ b/protocol/messenger_contact_requests_test.go
@@ -158,9 +158,9 @@ func (s *MessengerContactRequestSuite) receiveContactRequest(messageText string,
 	return contactRequest
 }
 
-func (s *MessengerContactRequestSuite) acceptContactRequest(contactRequest *common.Message, theirMessenger *Messenger) {
+func (s *MessengerContactRequestSuite) acceptContactRequest(contactRequest *common.Message, sender *Messenger, receiver *Messenger) {
 	// Accept contact request, receiver side
-	resp, err := theirMessenger.AcceptContactRequest(context.Background(), &requests.AcceptContactRequest{ID: types.Hex2Bytes(contactRequest.ID)})
+	resp, err := receiver.AcceptContactRequest(context.Background(), &requests.AcceptContactRequest{ID: types.Hex2Bytes(contactRequest.ID)})
 	s.Require().NoError(err)
 
 	// Make sure the message is updated
@@ -187,16 +187,16 @@ func (s *MessengerContactRequestSuite) acceptContactRequest(contactRequest *comm
 	s.Require().True(resp.Chats()[1].Active)
 
 	// Make sure the sender is added to our contacts
-	contacts := theirMessenger.AddedContacts()
+	contacts := receiver.AddedContacts()
 	s.Require().Len(contacts, 1)
 
 	// Make sure we consider them a mutual contact, receiver side
-	mutualContacts := theirMessenger.MutualContacts()
+	mutualContacts := receiver.MutualContacts()
 	s.Require().Len(mutualContacts, 1)
 
 	// Wait for the message to reach its destination
 	resp, err = WaitOnMessengerResponse(
-		s.m,
+		sender,
 		func(r *MessengerResponse) bool {
 			return len(r.Contacts) == 1 && len(r.Messages()) == 1 && len(r.ActivityCenterNotifications()) == 1
 		},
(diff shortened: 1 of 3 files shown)


protocol/messenger_handler.go
```diff
@@ -410,10 +410,89 @@ func (m *Messenger) handleCommandMessage(state *ReceivedMessageState, message *c
...
+	if contactRequestID != "" {
+		return nil
+	}
+
+	if chat == nil {
+		return fmt.Errorf("no chat restored during the contact synchronisation, contact.ID = %s", contact.ID)
+	}
+
+	clock, timestamp := chat.NextClockAndTimestamp(m.transport)
+	contactRequest, err := m.generateContactRequest(clock, timestamp, contact, "Please add me to your contacts", false)
```
ilmotta: I found 4 usages of this magical sentence `Please add me to your contacts` in production code. Since this function is also duplicating this string, it would be a good opportunity to extract them all into a constant.

protocol/messenger_handler.go
```diff
@@ -410,10 +410,89 @@ func (m *Messenger) handleCommandMessage(state *ReceivedMessageState, message *c
...
+	return nil
+}
+
+func (m *Messenger) syncSentContactRequest(contact *Contact, state *ReceivedMessageState, chat *Chat) error {
+	contactRequestID, err := m.persistence.LatestPendingContactRequestIDForContact(contact.ID)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("------------> syncSentContactRequest", contactRequestID)
```
ilmotta: Probably a leftover.

protocol/messenger_handler.go
```diff
@@ -410,10 +410,89 @@ func (m *Messenger) handleCommandMessage(state *ReceivedMessageState, message *c
...
+	// create AC notification
+	err = m.createIncomingContactRequestNotification(contact, state, contactRequest, true)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (m *Messenger) syncSentContactRequest(contact *Contact, state *ReceivedMessageState, chat *Chat) error {
```
ilmotta: This and other `syncReceivedContactRequest` function are quite similar.

As far as I understood, the first part of their work is identical, except for the boolean flag passed to `m.generateContactRequest`. On the second part they differ on the call to generate a notification. At least on the first part I think it's worth an extraction refactor and parameterizing the boolean flag.

protocol/messenger_handler.go
```diff
@@ -410,10 +410,89 @@ func (m *Messenger) handleCommandMessage(state *ReceivedMessageState, message *c
...
+	contactRequest.ID = defaultContactRequestID(contact.ID)
+
+	// save this message
+	state.Response.AddMessage(contactRequest)
+	err = m.persistence.SaveMessages([]*common.Message{contactRequest})
+	if err != nil {
+		return err
+	}
+
+	// create AC notification
```
ilmotta: At least to me, this code is very understandable without the comments, same for the comment above (`save this message`).

protocol/messenger_handler.go
```diff
@@ -410,10 +410,89 @@ func (m *Messenger) handleCommandMessage(state *ReceivedMessageState, message *c
...
+		return err
+	}
+
+	fmt.Println("------------> syncSentContactRequest", contactRequestID)
+
+	if contactRequestID != "" {
+		return nil
+	}
+
+	if chat == nil {
```
ilmotta: Slightly better would be to move the `chat == nil` check at the top of the function, so we can skip the (potentially) unnecessary call to `LatestPendingContactRequestIDForContact`.

protocol/messenger_contact_requests_test.go
```diff
@@ -1052,3 +1074,151 @@ func (s *MessengerContactRequestSuite) TestReceiveAcceptAndRetractContactRequest
...
+	contacts := bob2.AddedContacts()
+	s.Require().Len(contacts, 1)
+
+	// Make sure we consider them a mutual contact, receiver side
+	mutualContacts := bob2.MutualContacts()
+	s.Require().Len(mutualContacts, 1)
+}
+
+// The scenario tested is as follow:
+// 1) Alice sends a contact request to Alice
```
ilmotta: Typos:

- `Alice sends a contact request to Bob*`
- `Alice resets her* device`

protocol/messenger_handler.go
caybro: Ideally in a way that it could be translated since it's a user visible string ;)

ilmotta: Absolutely @caybro. Ideally, I think the text used to generate the contact request message should always come from the client, so it's the client's responsibility to translate and respect the user's preferences. To me, the endpoint `SendContactRequest` should even validate the presence of the message, and if it's not there not default to anything and err, since English doesn't make any sense for a lot of users.

But of course, not something I'm suggesting for this PR.

protocol/messenger_handler.go
MishkaRogachev: Kk, i replaced this one `syncContactRequestForInstallationContact`

protocol/messenger_handler.go
MishkaRogachev: fixed

protocol/messenger_handler.go
MishkaRogachev: I hope in future we should restore CR message with valid ID and Message, and default contact request text would be used only for backward compatibility. In desktop code we can provide transition  with QT_TR_NOOP for external strings

MishkaRogachev: fixed

protocol/messenger_handler.go
ilmotta: Looks a lot cleaner. Thanks @MishkaRogachev

ilmotta (approved):

cammellos (approved):

Issue #3429, opened 2023-04-26
Generate Users Image with their initials and Identicon ring if no profile picture is set.


J-Son89:
In the new mobile designs the user avatar has the users initials showing in the case where they have not set a profile picture. 
To simplify the logic on mobile side it is easier if the users picture comes from status-Go and the mobile client only has to render the image (With identicon included into it)

An example of these designs is here: 
https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Design-System-for-Mobile?node-id=115-6787&t=SXo5k6P6jpoq6IrS-0

<img width="59" alt="Screenshot 2023-04-26 at 10 17 25" src="https://user-images.githubusercontent.com/22799766/234530504-b896107a-9166-447c-8a7c-5a1721e6989f.png">

This should also work for the desktop client if they wish to start using this approach too

J-Son89:
discussed with @cammellos about this approach 👍

yqrashawn:
hi @jo-mut I'm working on generating the ring background with media server

PR #3500, opened 2023-05-17
feat: media server and colorhash related change

yqrashawn:
related to first subissue in https://github.com/status-im/status-mobile/issues/15788
which is no ident ring when app is locked

also related to https://github.com/status-im/status-go/issues/3429 to serve initials before login

1. add backend.StartMediaServerWithoutDB to start media server before login
it starts the media server with no appdb and ipfs downloader, do nothing if node already started
1. start the media server when calling `OpenAccounts`, so that we can access media server before login
1. add enrichMultiaccountInfo to add `ColorHash` and `ColorID` to multiaccount before `SaveAccount`, so that multiaccounts always have them defined (currently they are set to default values )
1. update `handleAccountImages`so that it don't require pubkey(no access before login) if ColorHash is defined in multiaccount
1. add `handleRequestDBMissing`, `handleRequestDownloaderMissing` to guard other media server handler when appdb, downloader is missing

recommend to review commit by commit


qfrank (approved):
sorry for the delay, i got something prior yesterday. 
LGTM!

oh, don't forget bump version

PR #3507, opened 2023-05-18
feat: enable multiple db connections with sqlcipher

### LINKED ISSUE #10767: [Performance] enable multiple db connections with sqlcipher

osmaczko:
closes: https://github.com/status-im/status-desktop/issues/10767
part of: status-im/status-desktop#10558

### DIFF
--- a/sqlite/sqlite.go
+++ b/sqlite/sqlite.go
@@ -2,11 +2,13 @@ package sqlite
 
 import (
 	"database/sql"
+	"database/sql/driver"
 	"errors"
 	"fmt"
 	"os"
+	"runtime"
 
-	_ "github.com/mutecomm/go-sqlcipher" // We require go sqlcipher that overrides default implementation
+	sqlcipher "github.com/mutecomm/go-sqlcipher" // We require go sqlcipher that overrides default implementation
 
 	"github.com/status-im/status-go/protocol/sqlite"
 )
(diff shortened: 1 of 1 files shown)


sqlite/sqlite.go
```diff
@@ -81,23 +82,33 @@ func openDB(path string, key string, kdfIterationsNumber int) (*sql.DB, error) {
...
+
+		if kdfIterationsNumber <= 0 {
+			kdfIterationsNumber = sqlite.ReducedKDFIterationsNumber
+		}
+
+		if _, err = conn.Exec(fmt.Sprintf("PRAGMA kdf_iter = '%d'", kdfIterationsNumber), []driver.Value{}); err != nil {
+			return err
+		}
+
+		return nil
```
alexjba: `WAL` mode is not set. Is it on purpose?

sqlite/sqlite.go
```diff
@@ -81,23 +82,33 @@ func openDB(path string, key string, kdfIterationsNumber int) (*sql.DB, error) {
...
-	}
-
-	if _, err = db.Exec(fmt.Sprintf("PRAGMA kdf_iter = '%d'", kdfIterationsNumber)); err != nil {
-		return nil, err
+	// Workaround: casting the driver type breaks the database/sql abstraction
+	// and may lead to compatibility issues in the future.
+	// This method is used because the 'database/sql' package does not expose 'ConnectHook',
+	// thereby making it impossible to individually configure each connection.
+	// Consequently, the connections from the pool can't be properly decrypted, making them unusable.
+	sqlcipherDriver, ok := db.Driver().(*sqlcipher.SQLiteDriver)
```
alexjba: There's one other way of doing it without breaking the abstraction:

```
        driverName := fmt.Sprintf("sqlcipher_with_extensions-%d", len(sql.Drivers()))
	sql.Register(driverName , &sqlcipher.SQLiteDriver{
		ConnectHook: func(conn *sqlcipher.SQLiteConn) error {
			///Running init queries
		},
	})

	db, err := sql.Open(driverName, path)
```

sqlite/sqlite.go
alexjba, line 1: With the current configuration on a test project, when trying a few writes at the same time I get `database is locked` on 10-20% of them. The test is configured to write 1Kb of data per task.

Might need some tweaking in order to get the async writing in a better shape.
Other projects are using the busy_timeout to fix this, but I think we need to make sure every write we do doesn't take more than the timeout value.

https://github.com/mattn/go-sqlite3/issues/274

caybro (approved):

sqlite/sqlite.go
alexjba: In my tests I found the best performance when using a limited amount of connections and the same number of max idle connections. The peak performance I see on M1 Macs is when using:
```
	db.SetMaxOpenConns(10)
	db.SetMaxIdleConns(10)
```
Not sure it's a coincidence to be the same number as CPU count.

Leaving the max open connection to unlimited will open a new connection when it's needed, and close the idle connections. This has the potential of paying the cost of opening the connection on lots of queries.

The difference I have is (In the app the difference might be smaller because we're probably not throttling the DB at this magnitude, but it gives an idea on how the connection pool works):

32 queries/second with unlimited connections
60662 queries/second with 10 max connections, 10 max idle connections

alexjba (commented):
Can't wait to see it in the app 🚀
Might need some tweaking in order to improve the speed and write safety.

sqlite/sqlite.go
caybro: > In my tests I found the best performance when using a limited amount of connections and the same number of max idle connections. The peak performance I see on M1 Macs is when using:
> 
> ```
> 	db.SetMaxOpenConns(10)
> 	db.SetMaxIdleConns(10)
> ```
> 
> Not sure it's a coincidence to be the same number as CPU count.

Yeah this smells like it's supposed to ideally use the number of HW threads :)

caybro: Something like this should be used: https://go.dev/play/p/IFnlTmTD6Q

osmaczko: Smart! I will use it :+1:

sqlite/sqlite.go
osmaczko: Good catch! I thought setting it once would be sufficient, as once you set PRAGMA journal_mode=WAL;, the setting is stored in the database file itself. I have learned that even though switching the mode to WAL introduces a persistent change to the database, each new connection must still "opt-in" to using the WAL mode. If a connection does not do this, it defaults to the DELETE journal mode. While this connection can still read from and write to the database, it will not take advantage of the benefits of WAL.

sqlite/sqlite.go
caybro: I've seen the same error _before_ these changes a couple of times too, with the discord import when the app was under heavy load.

alexjba:
Closing [#10723](https://github.com/status-im/status-desktop/issues/10723)

sqlite/sqlite.go
osmaczko: Updated

sqlite/sqlite.go
osmaczko: Updated

sqlite/sqlite.go
osmaczko: I've added a workaround (`PRAGMA busy_timeout=60000`) to mitigate the "database is locked" errors during concurrent write operations.

This approach retains behavior similar to our current setup with one 'always blocking' connection, which waits as long as needed for the connection to release lock. I believe setting `busy_timeout` to 60s (aka "infinite" wait) essentially approximates the same behavior.

sqlite/sqlite.go
osmaczko: I've done some benchmarking and it seems for my machine the number of `maxIdleConns` doesn't really matter:

image

@alexjba could you please give it a try on M1? Benchmarking code: https://github.com/osmaczko/status-go-db-perf/

osmaczko: I set `maxOpenConns` to `nproc` as suggested and left `maxIdleConns` as default = 2 (see comment above).

alexjba (approved):
Nice! 🚀

cammellos (approved):

PR #3508, opened 2023-05-19
fix(desktop/wallet): fix bugs in balance_cache, crash on nil in findBlocksCommand

IvanBelyakoff:
Fix bug in balance_cache - balances and nonces were stored in cache by pointers, which caused false cache hits in loop
because pointers with same address were created for different block numbers. Now cache uses block numbers of uint64 as key, which can overflow but it is not a problem since we use this cache for values comparison, not as user data.
Fix crash on nil pointer in findBlocksCommand.
Remove some unused code.
Protect nonceRanges with mutex while reading.


IvanBelyakoff:
Was closed automatically after branch renaming

PR #3509, opened 2023-05-19
fix(desktop/wallet): fix bug in balance_cache - balances and nonces

IvanBelyakoff:
were stored in cache by pointers, which caused falsy cache hits in loop because pointers with same address were created for different block numbers. Now cache uses block numbers of uint64 as key, which can overflow but it is not a problem since we use this cache for values comparison, not as user data.
Fix crash on nil pointer in log.
Remove some unused code.
Protect nonceRanges with mutex while reading.

### DIFF
--- a/services/wallet/transfer/balance_cache.go
+++ b/services/wallet/transfer/balance_cache.go
@@ -17,8 +17,8 @@ type nonceRange struct {
 
 type balanceCache struct {
 	// balances maps an address to a map of a block number and the balance of this particular address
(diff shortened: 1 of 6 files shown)


rasom (approved):

saledjenic (approved):

stefandunca (approved):
Great catch. I don't expect block numbers over 2^31 anytime soon :). Even the fact that we use big.Int might be related to the blockchain go APIs returning big ints.

As an optimization might make sense for the use a block number type and keep that as int64 until really required to have bigger storage.

PR #3511, opened 2023-05-19
fix(import): fix HandleImport not saving the chat

jrainville:
Since HandleImport doesn't save the chat, it doesn't update the unviewedMessagesCount, so we might not correctly tell if a new message is present in the chat and community if the import is processed before the store node fetches

I created a test that validates the behavior. Before the fix, the test was not passing, and it does with the fix.

I'll test this out with my release account to confirm if it really fixes the problem and remove from draft if it does.


jrainville:
Closing in favour of https://github.com/status-im/status-go/pull/3539

PR #3512, opened 2023-05-21
feat(wallet) extract and use transaction status for activity filtering

stefandunca:
### Update status-desktop [#10746](https://github.com/status-im/status-desktop/issues/10746)

- [x] extract and migrate existing status from receipt to the activity [#3540](https://github.com/status-im/status-go/pull/3540)
- [x] extend filter to include status extraction
- [x] extend filter to aggregate data from complex activity entries
- [x] expose status in API

### DIFF
--- a/services/wallet/activity/activity.go
+++ b/services/wallet/activity/activity.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	eth "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/log"
 
 	"github.com/status-im/status-go/services/wallet/common"
 	"github.com/status-im/status-go/services/wallet/transfer"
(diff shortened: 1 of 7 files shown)


services/wallet/activity/activity.go
```diff
@@ -198,6 +199,9 @@ const (
 	//
 	// UNION ALL is used to avoid the overhead of DISTINCT given that we don't expect to have duplicate entries outside
 	// the sender and receiver addresses being in the list which is handled separately
+	//
+	// Only status FailedAS, PendingAS and CompleteAS are returned. FinalizedAS is for bundled transactions
```
dlipicar: not sure if I'm mixing concepts, didn't the "Finalized" state mean the block for this transaction is [finalized](https://docs.symbol.dev/concepts/block.html#finalization)? In that case, it wouldn't matter if its a single transaction or a transaction bundle.

dlipicar (approved):
Just havethat question about what we call a "Finalized" activity. If the definition used here is the one we want, the changes look fine

services/wallet/activity/activity.go
stefandunca: You are right. I removed the bundle reference.

PR #3513, opened 2023-05-22
feat: render initials avatar using media server

### LINKED ISSUE #3429: Generate Users Image with their initials and Identicon ring if no profile picture is set.

yqrashawn:
1. [x] add method to extract initials from name
1. [x] add method to generate initials images
1. [x] ~get bg color from `ColorID`~ passed from mobile
1. [x] new media server handler for initials
1. [ ] unit test for generating initials image
1. [x] fn to draw online status on provided image
1. [x] handler for initials/custom pic placeholder (before key generated)
1. [x] crop contact avatar

Closes #3429

### Summary

This PR adds the ability to render initials avatars and placeholder avatars (used in onboarding when creating profile) in status-go
Also refactors `server/handlers.go`, separate params parsing and following logic


server/handlers.go
```diff
@@ -52,80 +56,502 @@ func handleRequestDownloaderMissing(logger *zap.Logger) http.HandlerFunc {
...
+		logger.Error("failed to write image", zap.Error(err))
+	}
+}
+
+func handleAccountInitials(multiaccountsDB *multiaccounts.Database, logger *zap.Logger) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		params := r.URL.Query()
+		parsed := ParseParams(logger, params)
+
+		if parsed.FontFile == "" {
```
siddarthkay, line 538: I'm curious how we are passing a FontFile from the client to status-go?
Are we passing the path to it?

siddarthkay (commented):
Huge PR to review, but I will take time out and complete the pending review today.
Thanks for your contribution @yqrashawn

server/handlers.go
yqrashawn: Yes, the absolute path to the font file. mobile use react-native-fs to get the path.

siddarthkay: Thanks for the clarification.

yqrashawn: mobile code at https://github.com/status-im/status-mobile/pull/16193/files#diff-ddfc24fc1622b950e08d907033895039aef6924760bf4e7670dce02dea1c6aeaR16

server/handlers.go
```diff
@@ -52,80 +56,502 @@ func handleRequestDownloaderMissing(logger *zap.Logger) http.HandlerFunc {
...
+		}
+		parsed.InitialsLength = amountInitials
+	}
+
+	fontFiles := params["fontFile"]
+	if len(fontFiles) != 0 {
+		if _, err := os.Stat(fontFiles[0]); err == nil {
+			parsed.FontFile = fontFiles[0]
+		} else {
+			logger.Error("ParseParams: font file not exit", zap.String("FontFile", fontFiles[0]))
```
qfrank, line 143: I think we shouldn't hide the original error when log, https://stackoverflow.com/questions/12518876/how-to-check-if-a-file-exists-in-go

server/handlers.go
```diff
@@ -52,80 +56,502 @@ func handleRequestDownloaderMissing(logger *zap.Logger) http.HandlerFunc {
...
+
+	if err != nil {
+		logger.Error("handleAccountInitialsImpl: failed to generate initials image.", zap.String("keyUid", parsed.KeyUID), zap.String("name", account.Name), zap.Error(err))
+		return
+	}
+
+	if parsed.Ring {
+		if accColorHash == nil {
+			if parsed.PublicKey == "" {
+				logger.Error("handleAccountInitialsImpl: no public key or color hash, can't draw ring", zap.String("keyUid", parsed.KeyUID), zap.Error(err))
```
qfrank: maybe change 
> no public key or color hash

to no public key

qfrank (approved):
LGTM, just left minor comments, maybe @Samyoul have something to say about global `ParsedParams` 😅, that's why he did refactor for global pair configuration object with [specify ones](https://github.com/status-im/status-go/blob/f05ef09079f6972dc41edf0e4358215f2ed29397/server/pairing/config.go#L11). I personally prefer different param type rather than global since there could exist a situation that we don't always need parse all the params.  It's up to you. Thanks for inviting ❤️

images/status_indicator.go
```diff
@@ -0,0 +1,56 @@
...
+	"bytes"
+	"image"
+	"image/color"
+	"image/png"
+	"math"
+
+	"github.com/fogleman/gg"
+)
+
+func AddStatusIndicatorToImage(inputImage []byte, innerColor color.Color, indicatorSize, indicatorBorder int) ([]byte, error) {
```
flexsurfer: so even an indicator in status-go? how it will be updated? seems like its better to have it in react no?

yqrashawn: > seems like its better to have it in react no?

That's what I planned for in the first place.
Tried my best to remember why I chose to do that in status-go.
 Found out, in the figma design, the indicator border shold be transparent so that the color of it shold be whatever beneath it. Although, I just checked, it's not the case in old user-avatar component

server/handlers.go
yqrashawn: fixed

server/handlers.go
yqrashawn: fixed, thanks for the info

siddarthkay (approved):
Great Work @yqrashawn !
Looks good to me.

yqrashawn:
@qfrank Thanks for the review.
Different param types should be better. But when I tried to separate the `ParsedParams` into different types, I found the code became less readable, which I believe mainly because I'm not good at golang.
What I tried is to define params type for each handler and separate the parse of each query params into shorter functions like `ringEnabled`, then define `ParseParam` function for each handler. 
The refactor is not finished yet, but I already found it's less readable. I'll dm you for help.

images/initials.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/fogleman/gg"
+	"golang.org/x/image/font"
+	"golang.org/x/image/font/opentype"
+)
+
+type RGBA struct {
+	R, G, B, A float64
+}
+
+var parsedFont *opentype.Font = nil
```
Samyoul, line 19: If you are relying on a font global perhaps the font should be loaded as part of the package `init()` function.

There seems to be a lot of work-around work to ensure a dynamic parameter `fontFile` isn't parsed more than once, and there is a presumption in the code that only one fontFile will be used in the application lifecycle.

Perhaps instead of passing the font file into `status-go` as a parameter, add the file as a binary asset? Then you can reference the font bytes directly without the need for file-paths.

Samyoul: I can help you with this if you need any assistance.

images/initials.go
```diff
@@ -0,0 +1,84 @@
...
+		if name != "" {
+			initials.WriteString(strings.ToUpper(name[0:1]))
+		}
+	}
+	return initials.String()
+}
+
+// uppercaseRatio is <height of any upper case> / dc.FontHeight() (line height)
+// 0.60386123 for Inter-UI-Medium.otf
+func GenerateInitialsImage(initials string, bgColor, fontColor color.Color, fontFile string, size int, fontSize float64, uppercaseRatio float64) ([]byte, error) {
```
Samyoul: Just a little note here. The golang convention is to start all documentation with the name of the element. See example below:

```suggestion
// GenerateInitialsImage uppercaseRatio is <height of any upper case> / dc.FontHeight() (line height)
// 0.60386123 for Inter-UI-Medium.otf
func GenerateInitialsImage(initials string, bgColor, fontColor color.Color, fontFile string, size int, fontSize float64, uppercaseRatio float64) ([]byte, error) {
```

images/initials_test.go
```diff
@@ -0,0 +1,27 @@
+package images
```
Samyoul, line 1: Would it be possible to write a test that outputs the image bytes and compares against a predefined byte slice?

server/handlers.go
```diff
@@ -52,80 +57,507 @@ func handleRequestDownloaderMissing(logger *zap.Logger) http.HandlerFunc {
 	}
 }
 
-func handleAccountImages(multiaccountsDB *multiaccounts.Database, logger *zap.Logger) http.HandlerFunc {
-	return func(w http.ResponseWriter, r *http.Request) {
-		params := r.URL.Query()
+type ParsedParams struct {
```
Samyoul: Perhaps rename this to `ImageParams` or `ProfileImageParams` or `PlaceHolderImageParams`

server/handlers.go
```diff
@@ -52,80 +57,507 @@ func handleRequestDownloaderMissing(logger *zap.Logger) http.HandlerFunc {
...
+
+	Hash     string
+	Download bool
+}
 
-		keyUids, ok := params["keyUid"]
-		if !ok || len(keyUids) == 0 {
-			logger.Error("no keyUid")
-			return
+func ParseParams(logger *zap.Logger, params url.Values) ParsedParams {
```
Samyoul: Perhaps change this function name to `ParseImageParams`? Just something that describes what is being parsed.

server/handlers.go
```diff
@@ -52,80 +57,507 @@ func handleRequestDownloaderMissing(logger *zap.Logger) http.HandlerFunc {
...
+		if accColorHash == nil {
+			if parsed.PublicKey == "" {
+				logger.Error("handleAccountImagesImpl: no public key for color hash", zap.String("keyUid", parsed.KeyUID))
+				return
+			}
 
+			accColorHash, err = colorhash.GenerateFor(parsed.PublicKey)
 			if err != nil {
-				logger.Error("handleAccountImages: failed to GetAccount .", zap.String("keyUid", keyUids[0]), zap.Error(err))
+				logger.Error("handleAccountImagesImpl: could not generate color hash", zap.String("keyUid", parsed.KeyUID))
```
Samyoul: Here we should probably not drop the logging of the actual error.

server/handlers.go
```diff
@@ -52,80 +57,507 @@ func handleRequestDownloaderMissing(logger *zap.Logger) http.HandlerFunc {
...
+		if err != nil {
+			logger.Error("ParseParams: invalid fontSize", zap.String("FontSize", fontSizeStr[0]))
+			return parsed
+		}
+		parsed.FontSize = fontSize
+	}
+
+	colors := params["color"]
+	if len(colors) != 0 {
+		color, err := images.ParseColor(colors[0])
```
Samyoul: The `color` parameter collides with the `color` package name, while not a compile error it can lead to unexpected behaviour if you use the package in the same context as this variable.

server/handlers.go
```diff
@@ -52,80 +57,507 @@ func handleRequestDownloaderMissing(logger *zap.Logger) http.HandlerFunc {
...
+		parsed.Hash = hash[0]
+	}
+
+	_, download := params["download"]
+	parsed.Download = download
+
+	return parsed
+}
+
+func handleAccountImagesImpl(multiaccountsDB *multiaccounts.Database, logger *zap.Logger, w http.ResponseWriter, r *http.Request, parsed ParsedParams) {
```
Samyoul: `r *http.Request` is an unused parameter.

server/handlers.go
```diff
@@ -52,80 +57,507 @@ func handleRequestDownloaderMissing(logger *zap.Logger) http.HandlerFunc {
...
+	w.Header().Set("Content-Type", mime)
+	w.Header().Set("Cache-Control", "no-store")
+
+	_, err = w.Write(payload)
+	if err != nil {
+		logger.Error("handleAccountImagesImpl: failed to write image", zap.Error(err))
+	}
+}
+
+func handleAccountImagesPlaceholder(logger *zap.Logger, w http.ResponseWriter, r *http.Request, parsed ParsedParams) {
```
Samyoul: `r *http.Request` declared but not used

server/handlers.go
```diff
@@ -52,80 +57,507 @@ func handleRequestDownloaderMissing(logger *zap.Logger) http.HandlerFunc {
...
+
+		if parsed.KeyUID == "" {
+			handleAccountImagesPlaceholder(logger, w, r, parsed)
+		} else {
+			handleAccountImagesImpl(multiaccountsDB, logger, w, r, parsed)
+		}
+	}
+}
 
+func handleAccountInitialsImpl(multiaccountsDB *multiaccounts.Database, logger *zap.Logger, w http.ResponseWriter, r *http.Request, parsed ParsedParams) {
```
Samyoul: `r *http.Request` declared but not used

server/handlers.go
```diff
@@ -52,80 +57,507 @@ func handleRequestDownloaderMissing(logger *zap.Logger) http.HandlerFunc {
...
+	w.Header().Set("Content-Type", mime)
+	w.Header().Set("Cache-Control", "no-store")
+
+	_, err = w.Write(payload)
+	if err != nil {
+		logger.Error("failed to write image", zap.Error(err))
+	}
+}
+
+func handleAccountInitialsPlaceholder(logger *zap.Logger, w http.ResponseWriter, r *http.Request, parsed ParsedParams) {
```
Samyoul: `r *http.Request` declared but not used

server/handlers.go
```diff
@@ -220,27 +672,27 @@ func getTheme(params url.Values, logger *zap.Logger) ring.Theme {
...
+		parsed := ParseParams(logger, params)
+
+		if parsed.PublicKey == "" {
 			logger.Error("no publicKey")
 			return
 		}
-		pk := pks[0]
-		image, err := identicon.Generate(pk)
+
+		image, err := identicon.Generate(parsed.PublicKey)
```
Samyoul: `image` variable collides with the `image` package name. Same situation as the `color` package / variable name collision.

Samyoul: There is actually a chance that I myself have introduced some of these 😆

Samyoul (approved):
@yqrashawn Thank you for your cool work making initial avatar images, in the main this looks good, I've just added a few points that you may find useful to consider. 😄

yqrashawn:
@Samyoul thanks for the review, I'll fix them. I'm currently fixing QA issues on the mobile PR, will be back here once those issues fixed.

images/initials.go
yqrashawn: That would be great. There's also a hack on the mobile side to move the file to a location on android on startup in order to get the path that can be accessed by status-go.
@cammellos we talked about this and you prefer to pass the font file path from mobile, any inputs?

images/initials_test.go
yqrashawn: Going to merge this PR now, I'll try to add the test later

PR #3514, opened 2023-05-22
feat: return only enabled networks from the reader

alaibe:


services/wallet/reader.go
```diff
@@ -154,13 +155,25 @@ func (r *Reader) Stop() {
 }
 
 func (r *Reader) GetWalletToken(ctx context.Context, addresses []common.Address) (map[common.Address][]Token, error) {
+	areTestNetworksEnabled, err := r.accountsDB.GetTestNetworksEnabled()
```
stefandunca, line 158: Moving this responsibility of handling test or non-test networks for tokens only partially in status-go and others relying on the client (Nim) to provide the same information about active networks might cause inconsistencies. It will be good to move all the handling of active networks to status-go, and all active network dependent operations use the same source of truth.

stefandunca (approved):

services/wallet/reader.go
alaibe: Agree using, status-go only is better! I move it as soon as i see it

IvanBelyakoff (approved):

PR #3515, opened 2023-05-22
waku2: use protected topics for communities

richard-ramos:
With this PR, communities will use a pubsub topic that is protected with a private key. This key is saved into a table, and status-go on login and on community creation / join should add a signature validation to the protected pubsub topic. Community messages are also sent signed with this key.

For ease of testing the following rpc methods are created:
 - wakuext_storePubsubTopicKey(topic string, privKey string)
 - wakuext_subscribeToPubsubTopic

Current version hardcodes `/waku/2/status-signed-test-1` pubsubtopic for communities, but this should be abstracted away so people can create communities that are automatically mapped to existing gossipsub topics.

**Notes for myself**:
- [x] Discuss with @alrevuelta and @siphiuel if Filter will also benefit from this pubsub topic protection (I added some TODOs
- [x] Creating / Joining a community should automatically add a signed topic validator to the pubsub topic (might make sense to expose a function in go-waku to know if a pubsub topic is already protected or not)
- [x] Since a content topic might be used in different pubsub topics, I added this attribute to status-go's `Filter`. I have to review how are these created, to assign them a pubsub topic


appdatabase/migrations/sql/1686048342_pubsubtopic_key.up.sql
```diff
@@ -0,0 +1,24 @@
+CREATE TABLE IF NOT EXISTS pubsubtopic_key (
```
Samyoul: Perhaps this could be named `pubsubtopic_signing_key`? Just to make clear the purpose of the key.

eth-node/bridge/geth/waku.go
```diff
@@ -63,6 +63,16 @@ func (w *gethWakuWrapper) AddStorePeer(address string) (peer.ID, error) {
 	return "", errors.New("not available in WakuV1")
 }
 
+// SubscribeToPubsubTopic function only added for compatibility with waku V2
+func (w *gethWakuWrapper) SubscribeToPubsubTopic(topic string, optPublicKey *ecdsa.PublicKey) error {
+	// not available in WakuV1
+	return nil
+}
```
Samyoul: Do you think also return an error here? Just for consistency, as `StorePubsubTopicKey` returns an error.

eth-node/bridge/geth/waku.go
```diff
@@ -231,7 +241,7 @@ func (w *gethWakuWrapper) SendMessagesRequest(peerID []byte, r types.MessagesReq
 		Limit:  r.Limit,
 		Cursor: r.Cursor,
 		Bloom:  r.Bloom,
-		Topics: r.Topics,
+		Topics: r.ContentTopics,
```
Samyoul, line 244: Does Waku V1 support `ContentTopics`?

protocol/common/message_sender.go
```diff
@@ -504,10 +505,11 @@ func (s *MessageSender) EncodeAbridgedMembershipUpdate(
...
 	newMessage := &types.NewMessage{
-		TTL:       whisperTTL,
-		Payload:   wrappedMessage,
-		PowTarget: calculatePoW(wrappedMessage),
-		PowTime:   whisperPoWTime,
+		TTL:         whisperTTL,
+		Payload:     wrappedMessage,
+		PowTarget:   calculatePoW(wrappedMessage),
+		PowTime:     whisperPoWTime,
+		PubsubTopic: rawMessage.PubsubTopic,
```
Samyoul, line 531: Can `PubsubTopic`s be a nil / empty value?

protocol/messenger_mailserver.go
```diff
@@ -346,7 +347,7 @@ func (m *Messenger) syncFiltersFrom(filters []*transport.Filter, lastRequest uin
 
 	topicsData := make(map[string]mailservers.MailserverTopic)
 	for _, topic := range topicInfo {
-		topicsData[topic.Topic] = topic
+		topicsData[topic.PubsubTopic+topic.ContentTopic] = topic
```
Samyoul: That is interesting. Forgive me for suggesting this but should there be some kind of delimiter between the `topic.PubsubTopic` and the `topic.ContentTopic`?

protocol/transport/filter.go
```diff
@@ -15,8 +15,10 @@ type Filter struct {
 	// Identity is the public key of the other recipient for non-public filters.
 	// It's encoded using encoding/hex.
 	Identity string `json:"identity"`
-	// Topic is the whisper topic
-	Topic types.TopicType `json:"topic"`
+	// PubsubTopic is the waku2 pubsub topic
+	PubsubTopic string `json:"pubsubTopic"`
+	// ContentTopic is the waku topic
+	ContentTopic types.TopicType `json:"topic"`
```
Samyoul, line 21: What is the chance of a community becoming partitioned / separated if users use an older client that uses any version of `status-go` before this change is introduced?

services/mailservers/api_test.go
```diff
@@ -65,9 +66,9 @@ func TestTopic(t *testing.T) {
 	defer close()
 	topicA := "0x61000000"
 	topicD := "0x64000000"
-	topic1 := MailserverTopic{Topic: topicA, LastRequest: 1}
-	topic2 := MailserverTopic{Topic: "0x6200000", LastRequest: 2}
-	topic3 := MailserverTopic{Topic: "0x6300000", LastRequest: 3}
+	topic1 := MailserverTopic{ContentTopic: topicA, LastRequest: 1}
+	topic2 := MailserverTopic{ContentTopic: "0x6200000", LastRequest: 2}
+	topic3 := MailserverTopic{ContentTopic: "0x6300000", LastRequest: 3}
```
Samyoul: Is it possible to introduce testing that covers the new functionality introduced here? Example queries on the `pubsubTopic` and `contentTopic`.

wakuv2/api_test.go
```diff
@@ -23,6 +23,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/waku-org/go-waku/waku/v2/protocol/relay"
+
 	"github.com/status-im/status-go/wakuv2/common"
```
Samyoul, line 28: Not a point for this PR, but I wonder if we should consider the `waku-org` organisation repos as 1st party packages in the same way we treat `ethereum` org packages. Not important, but just a thought.

Samyoul (approved):
Wow, that was a lot to review 😆 ! Thank you very much for doing all this amazing work 🎉, going through this PR I saw nothing that was objectionable and only really had questions.

protocol/common/message_sender.go
Samyoul: I'm just wondering what the impact of 2 clients using an older version of `status-go` and one using a version including this change would have. Are they just totally incompatible?

eth-node/bridge/geth/waku.go
richard-ramos: Yeah i think that makes sense!

eth-node/bridge/geth/waku.go
richard-ramos: WakuV2 content topics and WakuV1 topics represent the same concept (a 4 bytes value)

protocol/messenger_mailserver.go
richard-ramos: Good idea, let's add a delimiter!

protocol/transport/filter.go
richard-ramos: Hm. Sadly the usage of shards introduces a breaking change in status-go. Older status-go client will not be able to communicate with a community that has been moved to a new shard. Maybe we should brainstorm strategies to mitigate this issue (for example, not making the sharding functionality available for at least 2 versions to give chance for users to migrate to newer status-go,  or introducing a deprecation notice in the clients to force them to upgrade it?)

protocol/common/message_sender.go
richard-ramos: A nil/empty pubsub topic will automagically become the DefaultWakuTopic in go-waku.  

Clients will be incompatible the moment a community uses a different shard. If a community continues to use the default waku topic, older and newer clients will keep being able to communicate with each other

services/mailservers/api_test.go
richard-ramos: I will try!

wakuv2/api_test.go
richard-ramos: Please expand on this idea :thinking:  ,   I did not know that we had an special treatment for ethereum-org packages except for a go-ethereum fork

protocol/common/message_sender.go
Samyoul: Thank you for your patience with me on this

protocol/transport/filter.go
Samyoul: > not making the sharding functionality available for at least 2 versions to give chance for users to migrate to newer status-go, or introducing a deprecation notice in the clients to force them to upgrade it?

That is probably the most realistic approach, it is what we do for other major breaking changes.

wakuv2/api_test.go
Samyoul: Ah this is nothing fancy. Our linter is configure to enforce the grouping of packages in the following way:

- `std lib`
- 3rd party packages
- `go-ethereum` and `status-go` packages

To me `waku-org` packages are not strictly 3rd party although clearly not "1st party". But if we treat `ethereum` org packages as something special, your work in `waku-org` really should have the same treatment as a kind of integral package. The rules could be:

- `std lib`
- 3rd party packages
- `go-ethereum`, `status-go` and `waku-org` packages

richard-ramos: Oh I see! good idea!

appdatabase/migrations/sql/1691173700_pubsubtopic_key.up.sql
```diff
@@ -0,0 +1,24 @@
+CREATE TABLE IF NOT EXISTS pubsubtopic_signing_key (
```
cammellos: Is this ok to make the primary key the topic? we can't have multiple keys for topic?

eth-node/types/mailserver.go
```diff
@@ -28,10 +28,11 @@ type MessagesRequest struct {
 	StoreCursor *StoreRequestCursor `json:"storeCursor"`
 	// Bloom is a filter to match requested messages.
 	Bloom []byte `json:"bloom"`
-
-	// Topics is a list of topics. A returned message should
+	// PubsubTopic is the gossipsub topic on which the message was broadcasted
+	PubsubTopic string
```
cammellos: do we need `json` here?

eth-node/types/mailserver.go
```diff
@@ -28,10 +28,11 @@ type MessagesRequest struct {
...
+	// ContentTopics is a list of topics. A returned message should
 	// belong to one of the topics from the list.
-	Topics [][]byte `json:"topics"`
+	ContentTopics [][]byte `json:"topics"`
```
cammellos: should we rename the json as well?

protocol/common/message_sender.go
```diff
@@ -791,6 +794,7 @@ func (s *MessageSender) handleErrDeviceNotFound(ctx context.Context, publicKey *
 }
 
 func (s *MessageSender) wrapMessageV1(rawMessage *RawMessage) ([]byte, error) {
+	fmt.Println("wrapMessageV1: pubsubTopic: ", rawMessage.PubsubTopic, " message type", rawMessage.MessageType.String())
```
cammellos: you want to use the logger here

protocol/communities_key_distributor.go
```diff
@@ -103,14 +103,15 @@ func (ckd *CommunitiesKeyDistributorImpl) distributeKey(communityID, hashRatchet
...
+func (ckd *CommunitiesKeyDistributorImpl) sendKeyExchangeMessage(community *communities.Community, hashRatchetGroupID []byte, pubkeys []*ecdsa.PublicKey, msgType common.CommKeyExMsgType) error {
 	rawMessage := common.RawMessage{
 		SkipProtocolLayer:     false,
-		CommunityID:           communityID,
+		CommunityID:           community.ID(),
 		CommunityKeyExMsgType: msgType,
 		Recipients:            pubkeys,
 		MessageType:           protobuf.ApplicationMetadataMessage_CHAT_MESSAGE,
 		HashRatchetGroupID:    hashRatchetGroupID,
+		PubsubTopic:           community.PubsubTopic(), // TODO: confirm if it should be sent in community pubsub topic
```
cammellos, line 114: I believe it will be ignored for now, as it's sent as 1-to-1 message, though there's some further optimization that we discussed that will change that.

protocol/messenger_communities.go
```diff
@@ -99,6 +99,7 @@ func (m *Messenger) publishCommunityEvents(msg *communities.CommunityEventsMessa
 		// we don't want to wrap in an encryption layer message
 		SkipProtocolLayer: true,
 		MessageType:       protobuf.ApplicationMetadataMessage_COMMUNITY_EVENTS_MESSAGE,
+		PubsubTopic:       transport.GetPubsubTopic(msg.CommunityID), // TODO: confirm if it should be sent in community pubsub topic
```
cammellos, line 103: I believe so cc @osmaczko @mprakhov

protocol/messenger_communities.go
```diff
@@ -128,6 +129,7 @@ func (m *Messenger) publishCommunityEventsRejected(community *communities.Commun
 		// we don't want to wrap in an encryption layer message
 		SkipProtocolLayer: true,
 		MessageType:       protobuf.ApplicationMetadataMessage_COMMUNITY_EVENTS_MESSAGE_REJECTED,
+		PubsubTopic:       community.PubsubTopic(), // TODO: confirm if it should be sent in community pubsub topic
```
cammellos, line 133: I believe so cc @osmaczko @mprakhov

protocol/messenger_communities.go
```diff
@@ -659,6 +666,16 @@ func (m *Messenger) JoinCommunity(ctx context.Context, communityID types.HexByte
...
 }
 
+func (m *Messenger) subscribeToCommunityShard(communityID []byte) error {
+	// TODO: store private key and topic
+	// TODO: determine pubsub topic and public key for community
+	// TODO: this should probably be moved completely to transport once pubsub topic logic is implemented
+
+	pubsubTopic := transport.GetPubsubTopic(communityID)
+	var communityPubKey *ecdsa.PublicKey
+	return m.transport.SubscribeToPubsubTopic(pubsubTopic, communityPubKey)
```
cammellos: I guess it's not implemented, but I would just pass `nil` here, so it looks a bit less like an error

protocol/messenger_communities.go
```diff
@@ -1005,6 +1030,7 @@ func (m *Messenger) RequestToJoinCommunity(request *requests.RequestToJoinCommun
 		CommunityID:       community.ID(),
 		SkipProtocolLayer: true,
 		MessageType:       protobuf.ApplicationMetadataMessage_COMMUNITY_REQUEST_TO_JOIN,
+		PubsubTopic:       community.PubsubTopic(), // TODO: confirm if it should be sent in community pubsub topic
```
cammellos, line 1031: If we assume the owner has moved (which should be a safe assumption), it should I believe, admins should also move to, but I defer to @mprakhov @osmaczko

wakuv2/persistence/signed_messages.go
```diff
@@ -0,0 +1,112 @@
...
+	return crypto.ToECDSA(privKeyBytes)
+}
+
+type ProtectedTopic struct {
+	PubKey *ecdsa.PublicKey
+	Topic  string
+}
+
+func (p *ProtectedTopicsStore) ProtectedTopics() ([]ProtectedTopic, error) {
+	rows, err := p.db.Query("SELECT pub_key, topic FROM pubsubtopic_signing_key")
```
cammellos, line 86: I don't see the closing rows statement, `defer rows.Close()` after checking for errors

cammellos (commented):
@richard-ramos great work, looks good to me, just waiting for a couple of confirmations.

This is a non-breaking change (as long as a commnunity does not enable it that is), is that correct?

richard-ramos:
> @cammellos: This is a non-breaking change (as long as a commnunity does not enable it that is), is that correct?

Indeed. It will be a breaking change the moment the community uses a different shard

appdatabase/migrations/sql/1691173700_pubsubtopic_key.up.sql
richard-ramos: This is currently not supported by nwaku, but if in the future it's needed, we can create a migration to alter the primary key definition

eth-node/types/mailserver.go
richard-ramos: Oops!

eth-node/types/mailserver.go
richard-ramos: This i'm actually not sure? Is this json used on mobile somewhere? If not, we can keep it with this name and assume that a json field `topic` always refers to content topics

protocol/common/message_sender.go
richard-ramos: oops!

eth-node/types/mailserver.go
cammellos: not used in mobile, so it's the same for us

richard-ramos:
I'm merging this PR to avoid rebasing it again :(  
I'll keep the TODO comments for the pubsub topic on which messages must be sent. After dogfooding I'll create a PR to remove these comments once confirmed they're being sent in the right pubsub topic.
cc: @mprakhov @osmaczko

PR #3519, opened 2023-05-23
Add system message for mutual contact state updates

MishkaRogachev:
For https://github.com/status-im/status-desktop/issues/10395

Changes:
 - [x] Add new protobuf message for mutual contact state updates
 - [x] Send contact state update message when becoming a mutual contacts
 - [x] Send contact state update message when removing


protocol/messenger_contact_requests_test.go
```diff
@@ -61,17 +61,34 @@ func (s *MessengerContactRequestSuite) newMessenger(shh types.Waku) *Messenger {
 	return messenger
 }
 
+func (s *MessengerContactRequestSuite) takeOutMessageByContentType(messages []*common.Message, contentType protobuf.ChatMessage_ContentType) *common.Message {
```
cammellos: `findFirstByContentType` / `findFirstWithContentType`?

protocol/messenger_contacts.go
```diff
@@ -16,6 +16,44 @@ import (
...
+	MutualStateUpdateTypeAdded
+	MutualStateUpdateTypeRemoved
+)
+
+func (m *Messenger) prepareMutualStateUpdateMessage(chatID string, updateType MutualStateUpdateType, clock uint64, timestamp uint64) *common.Message {
+	var text string
+	// TODO: provide translations & solve naming
+	switch updateType {
+	case MutualStateUpdateTypeSent:
+		text = "%1 sent %2 a Contact Request"
```
cammellos: you probably want to use `@0x....` so that is seen as a mention/changes as the username changes

protocol/messenger_contacts.go
```diff
@@ -235,6 +273,20 @@ func (m *Messenger) updateAcceptedContactRequest(response *MessengerResponse, co
 	response.AddMessage(contactRequest)
 	response.AddContact(contact)
 
+	// Send mutual state update message
+	timestamp := m.getTimesource().GetCurrentTime()
+	updateMessage := m.prepareMutualStateUpdateMessage(contact.ID, MutualStateUpdateTypeAdded, clock, timestamp)
+	ctx := context.Background()
+	updateResponse, err := m.sendChatMessage(ctx, updateMessage)
```
cammellos: mmh,looks like you are sending a system message?

MishkaRogachev: fixed, now i just emitting them locally on both sides

protocol/messenger_contact_requests_test.go
MishkaRogachev: fixed

protocol/messenger_contacts.go
MishkaRogachev: fixed

protocol/messenger_contacts.go
```diff
@@ -16,6 +16,55 @@ import (
...
+
+	var to string
+	var from string
+	if outgoing {
+		to = contactID
+		from = m.myHexIdentity()
+
+		switch updateType {
+		case MutualStateUpdateTypeSent:
+			text = "You sent a contact request to @" + to
```
ilmotta, line 31: A broader question, maybe @cammellos can clarify. Why does the server need to be concerned with the presentation layer, I mean the text? Aren't clients better equipped to do this job?

protocol/messenger_contacts.go
```diff
@@ -16,6 +16,55 @@ import (
...
+	} else {
+		to = m.myHexIdentity()
+		from = contactID
+
+		switch updateType {
+		case MutualStateUpdateTypeSent:
+			text = "@" + from + " sent you a contact request"
+		case MutualStateUpdateTypeAdded:
+			text = "@" + from + " added you as a contact"
+		case MutualStateUpdateTypeRemoved:
```
ilmotta, line 48: Since Go doesn't have pattern matching like Rust which does exhaustive checks, if a new `MutualStateUpdateType` is added, the compiler won't complain. This is by design, I'm not saying Rust is better. I'm wondering if it wouldn't be more robust to generate an error if no match is found, otherwise `text` will be empty, or at least have a default (non-empty) `text`. Maybe this is not the best example, since it's not a big deal. Anyways, wdyt?

ilmotta (approved):

protocol/messenger_contacts.go
cammellos: Yes, that's correct, we had a discussion with @MishkaRogachev and the agreement is that it's better to do that on the client side. 
Having said that, is a bit more complex and we already have a pattern that generates the message text here in status-go (it was done before when we had a command line client), and for the sake of simplicity, we agreed that's ok for now, since translations are not in scope for mvp (once we do translations, we'd want to change this behavior).
Do you think that's ok, or you'd rather address in this PR?

ilmotta: > since translations are not in scope for mvp

You mean translations are out of scope for status-go during the MVP? Because in the clients we already handle this quite decently.

> Do you think that's ok, or you'd rather address in this PR?

I see, if things are in place already, it's harder to justify the effort for the better alternative. Maybe this is the kind of decision that could be public, like lightweight decision records. Or we could also just create an issue related to UI concerns in status-go, or more specifically non-translated text. Well, thanks for the context @cammellos

cammellos: > You mean translations are out of scope for status-go during the MVP? Because in the clients we already handle this quite decently.

Yes, we have everything in places since we had that from before, but we won't be supporting it yet


Makes sense to have decisions records and/or creating issues, we had that in status-mobile before, maybe it's something that would be handy for status-go as well

ilmotta: Do you have a template at hand to share? One that you liked and/or worked for other repos in Status? I ask this because there are so many ways to write decision records, and some models can be too bureaucratic/tiring to write.

protocol/messenger_contacts.go
MishkaRogachev: I like the rust approach better too. Fixed :)

protocol/messenger_contacts.go
cammellos: sorry for the late reply, we used this in the past https://github.com/status-im/status-mobile/blob/develop/doc/decisions/0001-record-decisions.md

protocol/messenger_handler.go
```diff
@@ -939,7 +973,7 @@ func (m *Messenger) HandleAcceptContactRequest(state *ReceivedMessageState, mess
 	return nil
 }
 
-func (m *Messenger) handleRetractContactRequest(contact *Contact, message protobuf.RetractContactRequest) error {
+func (m *Messenger) handleRetractContactRequest(messageState *ReceivedMessageState, contact *Contact, message protobuf.RetractContactRequest) error {
```
cammellos: I would avoid passing the whole `messageState`, makes things a bit harder to test, and I think eventually we want to get rid of it completely, looks like all you need is `Response`, I would just pass that, I think that's a bit nicer, but up to you

cammellos (approved):
Great work @MishkaRogachev !

protocol/messenger_handler.go
MishkaRogachev: fixed)

PR #3520, opened 2023-05-23
Add new transfers fetching for sequential strategy

### LINKED ISSUE #10246: Scheduler to fetch historical transaction
New fetching strategy must fetch all transfers for all added accounts in background from latest block backwards

Tasks:

- [x] https://github.com/status-im/status-go/pull/3432
- [x] https://github.com/status-im/status-go/pull/3480
- [x] https://github.com/status-im/status-go/pull/3509
- [x] https://github.com/status-im/status-go/pull/3520
- [x] https://github.com/status-im/status-go/pull/3551
- [x] https://github.com/status-im/status-go/pull/3575
- [x] https://github.com/status-im/status-desktop/pull/10932

IvanBelyakoff:
Added fetching new transfers for SequentialFetchStrategy.

There is a bit of refactoring, so it is hard to review the changes, but they boil down to:

-  running one more command for fetching new transfers with a different range: [LastKnownBlock - Head]
-  updating fetched block ranges in DB:
    - update Start (beginning of history for account) block if found
    - update FirstKnown (oldest fetched) block if an older one is fetched
    - update LastKnown (newest fetched) block if a newer one is fetched

Closes [#10246](https://github.com/status-im/status-desktop/issues/10246)

### DIFF
--- a/services/wallet/transfer/block_ranges_sequential_dao.go
+++ b/services/wallet/transfer/block_ranges_sequential_dao.go
@@ -2,18 +2,11 @@ package transfer
 
 import (
 	"database/sql"
-	"fmt"
 	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/status-im/status-go/services/wallet/bigint"
-)
-
-const (
-	firstBlockColumn = "blk_first"
-	lastBlockColumn  = "blk_last"
-	startBlockColumn = "blk_start"
 )
 
 type BlockRangeSequentialDAO struct {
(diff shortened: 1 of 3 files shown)


services/wallet/transfer/block_ranges_sequential_dao.go
```diff
@@ -58,85 +51,84 @@ func (b *BlockRangeSequentialDAO) getBlockRange(chainID uint64, address common.A
...
-func (b *BlockRangeSequentialDAO) upsertRange(chainID uint64, account common.Address,
-	start *big.Int, first *big.Int, last *big.Int) (err error) {
+		// Overwrite last known block if there was not any or if new one is newer
+		if (blockRange.LastKnown == nil && newBlockRange.LastKnown != nil) ||
+			(blockRange.LastKnown != nil && newBlockRange.LastKnown != nil && blockRange.LastKnown.Cmp(newBlockRange.LastKnown) < 0) {
+			blockRange.LastKnown = newBlockRange.LastKnown
+		}
 
-	log.Info("upsert blocks range", "account", account, "network id", chainID, "start", start, "first", first, "last", last)
+		update, err := b.db.Prepare(`UPDATE blocks_ranges_sequential
```
stefandunca: Can't we simplify these two separate update and insert operations into one using `INSERT OR REPLACE INTO blocks_ranges_sequential ...` given that the composite primary key will take care of the WHERE clause? Don't remember if our sqlite supports it though.

services/wallet/transfer/commands_sequential.go
```diff
@@ -469,18 +464,79 @@ func loadTransfersLoop(ctx context.Context, accounts []common.Address, blockDAO
...
 				break
 			}
 		}
 	}
 
 	return nil
 }
+
+func getHeadBlockNumber(parent context.Context, chainClient *chain.ClientWithFallback) (*big.Int, error) {
+	ctx, cancel := context.WithTimeout(parent, 3*time.Second)
```
stefandunca, line 483: `3` seconds timeout with a noisy internet connection seems not much. I'm wondering what are the drawbacks of one request failing often due to this 3 seconds timeout. What are the drawbacks if we set this to a 15-20 given that we expect to run these requests on a mobile network.?

stefandunca (approved):

services/wallet/transfer/commands_sequential.go
IvanBelyakoff: Good question @stefandunca ! Indeed, 3 seconds could be not enough for some cases, though current implementation AFAIK is targeted for desktops. In this case the whole `loadBlocksAndTransfersCommand` will be retried in 13 seconds, which seems OK, because if we just try to re-fetch head within `getHeadBlockNumber` method  or increase a timeout for it to 15-20 secs, it will give basically the same result.

services/wallet/transfer/block_ranges_sequential_dao.go
IvanBelyakoff: @stefandunca Thank you for the suggestion, indeed, this a simpler approach I was not aware of, I fixed that.

services/wallet/transfer/commands_sequential.go
```diff
@@ -43,43 +41,12 @@ func (c *findBlocksCommand) Command() async.Command {
...
-
-		to = new(big.Int).Set(head.Number) // deep copy
-	} else {
-		to.Sub(to, big.NewInt(1))
-	}
-
-	var from = big.NewInt(0)
-	if to.Cmp(rangeSize) > 0 {
+	from, to := new(big.Int).Set(c.fromBlockNumber), new(big.Int).Set(c.toBlockNumber)
+	if to.Cmp(from) > DefaultNodeBlockChunkSize {
```
dlipicar:
```
// Cmp compares x and y and returns:
//
//	-1 if x <  y
//	 0 if x == y
//	+1 if x >  y
```

Don't you want to do a `Sub` between `from` and `to`, and then a `Cmp` with `rangeSize`?

services/wallet/transfer/commands_sequential.go
```diff
@@ -90,152 +57,117 @@ func (c *findBlocksCommand) Run(parent context.Context) (err error) {
...
+			"nonce", c.balanceCache.ReadCachedNonce(c.account, to))
 
-		log.Info("findBlocksCommand.Run()", "headers len", len(headers), "resFromBlock", c.resFromBlock.Number)
-		err = c.blockDAO.updateFirstBlock(c.chainClient.ChainID, c.account, c.resFromBlock.Number)
+		err = c.db.SaveBlocks(c.chainClient.ChainID, c.account, headers)
 		if err != nil {
 			c.error = err
-			log.Error("findBlocksCommand failed to update first block", "error", err)
-			break
+			// return err
```
dlipicar, line 68: hmm could you explain what we want to achieve with this? Why is it handled differently to the other errors?

services/wallet/transfer/commands_sequential.go
```diff
@@ -90,152 +57,117 @@ func (c *findBlocksCommand) Run(parent context.Context) (err error) {
...
-	// or when newFromBlock is different from fromBlock, but if I check for
-	// ERC20 transfers only when there are ETH transfers, I will miss ERC20 transfers
+	return nil
+}
+
+func (c *findBlocksCommand) checkRange(parent context.Context, from *big.Int, to *big.Int) (
+	foundHeaders []*DBHeader, err error) {
 
-	// if len(ethHeaders) > 0 || newFromBlock.Number.Cmp(fromBlock.Number) != 0 { // there is transaction history for this account
+	fromBlock := &Block{Number: from}
```
dlipicar, line 108: hmm not sure if this changed, but `fastIndex` used more than just the block number from the `Block` struct

dlipicar:
```
func (c *ethHistoricalCommand) Run(ctx context.Context) (err error) {
	log.Info("eth historical downloader start", "address", c.address, "from", c.from.Number, "to", c.to, "noLimit", c.noLimit)

	start := time.Now()
	if c.from.Number != nil && c.from.Balance != nil {
		c.balanceCache.addBalanceToCache(c.address, c.from.Number, c.from.Balance)
	}
	if c.from.Number != nil && c.from.Nonce != nil {
		c.balanceCache.addNonceToCache(c.address, c.from.Number, c.from.Nonce)
	}
	from, headers, startBlock, err := findBlocksWithEthTransfers(ctx, c.chainClient,
		c.balanceCache, c.eth, c.address, c.from.Number, c.to, c.
```

services/wallet/transfer/commands_sequential.go
IvanBelyakoff: Oops, absolutely! I've done it the right way a number of times, don't know how I missed this one. Great catch! We definitely need tests

services/wallet/transfer/commands_sequential.go
```diff
@@ -90,152 +57,117 @@ func (c *findBlocksCommand) Run(parent context.Context) (err error) {
...
-
-		log.Info("saving headers", "len", len(uniqHeaders), "lastBlockNumber", to,
-			"balance", c.balanceCache.ReadCachedBalance(c.account, to),
-			"nonce", c.balanceCache.ReadCachedNonce(c.account, to))
-
-		err = c.db.SaveBlocks(c.chainClient.ChainID, c.account, uniqHeaders)
+	// There should be transfers when either when we have found headers
+	// or newFromBlock is different from fromBlock
+	if len(ethHeaders) > 0 || newFromBlock.Number.Cmp(fromBlock.Number) != 0 {
+		erc20Headers, err := c.fastIndexErc20(parent, newFromBlock.Number, to)
```
dlipicar, line 122: Hmm I think I remember there being a good reason for calling `fastIndexErc20` regardless of the result of `fastIndex` like we currently do (that is, there was a chance of missing some blocks with transactions under some specific condition). Did you check with Roman?

services/wallet/transfer/commands_sequential.go
IvanBelyakoff: I think I overlooked it and overall error handling needs revising, which I planned, but have not done yet.
Basically from my point of view there are:

- network related issues that should cause exiting the command loop execution, because they are somewhat unrecoverable (run out of requests limits, there is no sense to do more requests), so we return nil (and might do something more to prevent further requests and an error banner on UI side)
- network related issues that should cause retry attempts and we should return err in this case for the finite command to retry
- other issues (like DB errors) when it is hard to say if we should proceed, because it could be related to a specific DB request and then some part of the functionality could not function properly or could be a common DB error, and then we should panic. I don't know the best way to handle them, so I think if saving blocks did not work, it is OK to return `nil` and not retry this command, though the whole process of checking the block range will be repeated by a caller - `loadBlocksAndTransfersCommand`

services/wallet/transfer/commands_sequential.go
IvanBelyakoff: I will have a thorough look at it if I need to fix that, thanx

services/wallet/transfer/commands_sequential.go
IvanBelyakoff: @rasom Roman, could you please have a look, do you remember any reason to proceed with ERC20 blocks if no balance and nonce change is detected for ETH for a block range?
@dlipicar for me for the account I tested, there is no difference between transfers found for both cases, and I don't known a reason why I should not stop further ERC20 check, so I decided to save some calls.

services/wallet/transfer/commands_sequential.go
IvanBelyakoff: @dlipicar When we start checking a block range, we pass it to `ethHistoricalCommand` and at this point we don't know yet the balance and nonce, otherwise why would we need to check it once again? So I think we should just remove the code above for adding balance and nonce to cache, as I think it does not make any sense. Balance and nonce for the block will be added to cache as a result of `BalanceAt` and `NonceAt` in `checkRangesWithStartBlock` at concurrent.go:
```
			lb, err := cache.BalanceAt(ctx, client, account, from)
			if err != nil {
				return err
			}
			hb, err := cache.BalanceAt(ctx, client, account, to)
			if err != nil {
				return err
			}
			if lb.Cmp(hb) == 0 {
				log.Debug("balances are equal", "from", from, "to", to, "lb", lb, "hb", hb)

				hn, err := cache.NonceAt(ctx, client, account, to)
```

IvanBelyakoff: @dlipicar For `OnDemandFetchStrategy` there is only one case when `from` block has balance and nonce set - when it is taken from `block_ranges` table. So for a new checked block range, one side of the range (block) is taken from DB, which saves some calls, but at the same time, I think sometimes transfers are loaded twice for this block (I faced this issue). I don't take one side of the range from DB, so it never hits the cache, but I don't load the transfers for that block twice.

services/wallet/transfer/commands_sequential.go
rasom: > I remember there being a good reason for calling fastIndexErc20

The reason is that we can't detect some ERC20 transfers by checking balances/nonces because they do not change those. Specifically incoming ERC20 but also some more exotic cases.

IvanBelyakoff: I believe @rasom means incoming 0-value transfers. I discussed it with @alaibe and for now the decision is to avoid searching for such transfers, because they don't make sense for users.
Having them displayed in the wallet can also cause phishing attacks, when 3rd party can transfer some 0-value token on behalf of the user (since this is 0-value, it is allowed) to a phishing address, which can mislead a user to further transfer their token to the same address but with real value, if that phishing 0-value transfer appears in wallet history.
Unless we have a good reason to allow this, it is disabled for now

rasom: @IvanBelyakoff if I receive 1000 USDT (for instance) this transaction will not be detected by checking balance/nonce (because it doesn't affect ETH balance and is not sent by me). **There is no good reason to not include this transaction.** Please don't do this.

IvanBelyakoff: @rasom Thank you for clarification. You are correct, I definitely missed your point, when you said that ERC20 transfers don't change ETH balance/nonce. My bad. I will fix it in the follow-commit.

PR #3521, opened 2023-05-23
chore: Add ColorId for Wallet accounts that can be shared with mobile

Khushboo-dev-cpp:

Important changes:

Closes #


saledjenic:
@Khushboo-dev-cpp is it possible to postpone merging of this PR till this one gets merged https://github.com/status-im/status-go/pull/3475 it introduces a new concept for accounts and will be easier to implement these changes over changes made there then vice versa.

Khushboo-dev-cpp:
> #3475

I could wait, is that PR close to merging? cause my UI PR has changes in over 70 files and I will get a lot of conflicts very soon :D

saledjenic:
As you can see it’s approved and can be merged any moment, I’m just waiting for the signal from cammellos cause he needs to launch a new mobile release and then gives me a green light for merging.

There is also a related desktop app PR for my change you can check it here: https://github.com/status-im/status-desktop/pull/10724

saledjenic (approved):

jakubgs:
These tests are taking an awful long time:

image

Maybe we should add a timeout.

multiaccounts/accounts/database.go
```diff
@@ -43,20 +44,21 @@ type Keypair struct {
...
+	Address   types.Address             `json:"address"`
+	KeyUID    string                    `json:"key-uid"`
+	Wallet    bool                      `json:"wallet"`
+	Chat      bool                      `json:"chat"`
+	Type      AccountType               `json:"type,omitempty"`
+	Path      string                    `json:"path,omitempty"`
+	PublicKey types.HexBytes            `json:"public-key,omitempty"`
+	Name      string                    `json:"name"`
+	Emoji     string                    `json:"emoji"`
+	Color     string                    `json:"color"`
```
saledjenic: @Khushboo-dev-cpp you don't need this here?

appdatabase/migrations/sql/1685109694_add_customisation_color_id.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE keypairs_accounts ADD COLUMN colorId VARCHAR NOT NULL DEFAULT "";
```
saledjenic: @Khushboo-dev-cpp please use `color_id` instead `colorId`.

appdatabase/migrations/sql/1685109694_add_customisation_color_id.up.sql
```diff
@@ -0,0 +1,2 @@
...
+UPDATE keypairs_accounts SET colorId = "";
```
saledjenic: @Khushboo-dev-cpp map old colors to new color ids in this migration script, just to be a backward compatible.

saledjenic (changes requested):
@Khushboo-dev-cpp you could proceed with `color` without renaming to `color_id`, but both are ok. But in any case you need to have migration script to map old colors to new ones.

protocol/protobuf/pairing.pb.go
```diff
@@ -5,8 +5,11 @@ package protobuf
```
saledjenic, line 5: @Khushboo-dev-cpp you have to commit `pairing.proto` file as well.

protocol/messenger_handler.go
```diff
@@ -2880,7 +2880,7 @@ func mapSyncAccountToAccount(message *protobuf.SyncAccount, accountOperability a
 		PublicKey: types.HexBytes(message.PublicKey),
 		Path:      message.Path,
 		Name:      message.Name,
-		Color:     message.Color,
```
saledjenic, line 2879: @Khushboo-dev-cpp this change here will break wallet accounts' color for all Status users who are recovering from waku (color for backed up account is under field `Color`, but we're trying to get it from field `ColorID`, also even if the name is set we need a mapper to map old colors to a new values). 

In general I am not sure do we need to be a backward compatible at this stage (before mvp) if we need, then we need that handler as well.

saledjenic (changes requested):
@Khushboo-dev-cpp I don't see migration script which will migrate old colors to a new values. If we don't provide such script color of all accounts will be broken. 

In general I am not sure do we need to be a backward compatible at this stage (before mvp) but if we need, then we need migration script for that.

protocol/messenger_handler.go
alaibe: Shall we just have a default color? if not provided?

saledjenic: > Shall we just have a default color? if not provided?

If we need to be a backward compatible at this point we would need to map "old" `color` value to a "new" `colorID`, but since we are still in the development phase I wouldn't add the code for handling that and make it more robust, at least not before mvp, especially having in mind that a very few people will recover from waku before mvp. 

What I want to say, since `colorID` contains now "old" `color` value, we will endup most likely with an empty value for `colorID`, so if we want to add a default color then better to add an appropriate mapping instead.

multiaccounts/accounts/database.go
```diff
@@ -383,7 +384,7 @@ func (db *Database) getKeypairs(tx *sql.Tx, keyUID string) ([]*Keypair, error) {
 			ka.pubkey,
 			ka.path,
 			ka.name,
-			ka.color,
+                       ka.color,
```
saledjenic: weird formatting

multiaccounts/accounts/database.go
```diff
@@ -464,7 +465,7 @@ func (db *Database) getAccounts(tx *sql.Tx, address types.Address) ([]*Account,
...
```
saledjenic: weird formatting

multiaccounts/accounts/database.go
```diff
@@ -708,7 +709,7 @@ func (db *Database) saveOrUpdateAccounts(tx *sql.Tx, accounts []*Account) (err e
 				keypairs_accounts
 			SET
 				name = ?,
-				color = ?,
+                               color = ?,
```
saledjenic: weird formatting

saledjenic (approved):
Just revert formatting, the rest looks good by my side.

We could even go without changing `color` to `colorID`, but ok, both ways are ok.

api/geth_backend.go
```diff
@@ -29,6 +29,7 @@ import (
 	"github.com/status-im/status-go/logutils"
 	"github.com/status-im/status-go/multiaccounts"
 	"github.com/status-im/status-go/multiaccounts/accounts"
+	common "github.com/status-im/status-go/multiaccounts/common"
```
Samyoul: Hey @Khushboo-dev-cpp thanks for this work, just so you know 😄 the package alias here is redundant as Go references by the final package in a path.

So `"github.com/status-im/status-go/multiaccounts/common"` would always be `common`

---

Edit: This applies to the other files, but I won't graffiti your PR with comments like this.

multiaccounts/accounts/test_helper.go
```diff
@@ -154,7 +155,7 @@ func GetSeedImportedKeypair1ForTest() *Keypair {
 		PublicKey: types.Hex2Bytes("0x000000022"),
 		Name:      "Seed Impo 1 Acc 2",
 		Emoji:     "emoji-2",
-		Color:     "blue",
+		ColorID:   common.CustomizationColorPrimary,
```
Samyoul, line 158: These are nice

api/geth_backend.go
Khushboo-dev-cpp: Done :)

Samyoul (approved):
If I'm going to moan about something in a PR, the least I can do is approve the PR after the moanable item has been fixed.

mprakhov:
@Khushboo-dev-cpp getting an error during the build

```
# github.com/status-im/status-go/multiaccounts
multiaccounts/database.go:53:3: unknown field ColorID in struct literal of type protobuf.MultiAccount
multiaccounts/database.go:83:17: ma.ColorID undefined (type *protobuf.MultiAccount has no field or method ColorID)
make[1]: *** [statusgo-shared-library] Error 1
make: *** [vendor/status-go/build/bin/libstatus.dylib] Error 2
```

How is it possible that it passed all checks and was merged?

saledjenic:
@mprakhov I faced the same issue today after executing `make generate`. Seems that `pairing.pb.go` was updated manually, that's how it passed the tests. All in all fixed in this PR: https://github.com/status-im/status-go/pull/3572 and it's in the `develop` branch now, just rebase and should be fine.

PR #3522, opened 2023-05-24
Detect UniswapV2/V3 swaps

dlipicar:
Part of #10251

Implements detection of UniswapV2 (https://docs.uniswap.org/contracts/v2/reference/smart-contracts/pair#swap) and UniswapV3 (https://docs.uniswap.org/contracts/v3/reference/core/interfaces/pool/IUniswapV3PoolEvents#swap) swap events 

First, a brief summary of how transaction downloading currently works.

For clarity, I'll use the following terms:
- Block: Ethereum block with a unique hash and number
- Transaction (Tx): Ethereum transaction with a unique hash, contained in a block
- Subtransaction (SubTx): Event of interest that happened as part of a Transaction (ERC20/ERC721 token transfer, Swap event, etc). They have a unique hash formed with the TxHash and the Log Event index used to generate it.
- Multitransaction (MultiTx): Concept used by us to represent "activities" that may span one or many Txs/SubTxs and contain their metadata. Examples of this are a Bridge action (involves transactions across multiple chains) or a Swap action (involves multiple SubTxs). Simple Send/Receive activities are special, simplified cases of a Multitransaction.

For a given account and a given block range:
1) Get a list of blocks that contain transfers (for this, `ETHDownloader` checks account balance/nonce changes, `ERC20Downloader` gets a list of ERC20/ERC721 `Transfer` event logs from/to the given address).
2) Insert these blocks to the `blocks` table (using struct `DBHeader`), marked as "not loaded". Insert the detected ERC20/ERC721 transfers to the `transfers` table, marked as "not loaded".
For every "not loaded" block:
1) For each "not loaded" ERC20/ERC721 transfer contained in it, fetch all needed additional data.
2) Look for transactions with `from` or `to` address matching the given address. Add them to the previous list.
3) Process MultiTxs as needed
4) Add/update `transfers` table with this list of transactions (marked as "loaded"). Also mark block as "loaded".

This logic needed to be changed a bit, since we're now not only interested in ERC20/ERC721 transfers but other kinds of SubTxs as well. We now do:

For a given account and a given block range:
1) Get a list of blocks that contain transfers (for this, `ETHDownloader` checks account balance/nonce changes, `ERC20Downloader` gets a list of ERC20/ERC721 `Transfer` event logs from/to the given address).
2) Insert these blocks to the `blocks` table (using struct `DBHeader`), marked as "not loaded". Insert one detected ERC20/ERC721 transfer per transaction hash (representing a "Transaction to be loaded") to the `transfers` table, marked as "not loaded".
For every "not loaded" block:
1) For each "not loaded" transaction contained in it, get the list of SubTxs of interest.
2) Look for transactions with `from` or `to` address matching the given address. Add them to the previous list.
3) Process MultiTxs as needed (in case a Swap subTx is found for a given Tx, generate a MultiTx linking all subTxs in that Tx)
3) Add/update `transfers` table with this list of transactions (marked as "loaded"). Also mark block as "loaded".

-o-

A bug was found and fixed in this PR:
- `ETHDownloader` used the (partial) header hash as the block hash, which didn't match the real block hash. `ERC20Downloader` gets the block hash from the transaction receipt, which is correct. For this reason, we had two entries in the blocks table with same Block Number but different hashes. `ETHDownloader` now obtains the real block hash from the first transaction (cheap query).

There was no need to modify the ERC20Downloader log filters since every swap transaction contains one or many `Transfer` events from/to the user's account.

In this PR, the following was done in separate commits:
1) Fix block hash bug
2) Add UniswapV2 and UniswapV3 contracts
3) Add util function to identify an ERC20 token by its contract address
4) Implement UniswapV2  identification
4a) `uniswapV2`  events are identified from the log list
4b) Split "subTransaction" detection in two stages: `ERC20Downloader` adds a single entry to the DB per TxHash. The full list of subTxs for each transaction is then obtained by the `ETHDownloader`.
4c) Transactions which contain a `uniswapV2` event generate a MultiTransaction, which is set to all its subTxs. For this, tokens and amounts are parsed from the event and obtained from the UniswapV2 `Pair` contract.
4d) Cleaned up a bunch of unused code and moved some functions to separate files.
5) Implement UniswapV3 identification
5a) `uniswapV3`  events are identified from the log list
5b) Transactions which contain a `uniswapV3` event generate a MultiTransaction, which is set to all its subTxs. For this, tokens and amounts are parsed from the event and obtained from the UniswapV3 `Pool` contract.

-o-

The way we store data in the DB is not optimal... Lots of duplicate data, since every entry in `transfers` holds the whole Transaction information on top of the "subtransaction". If we decouple what we store in that table and what we show to the user (using multitransactions for example), we could simply store the transaction data once and have a linked table with the logs.


rpc/chain/rpc.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (s *senderFromServer) Sender(tx *types.Transaction) (common.Address, error) {
+	if s.addr == (common.Address{}) {
+		return common.Address{}, errNotCached
+	}
+	return s.addr, nil
+}
+
+func (s *senderFromServer) ChainID() *big.Int {
+	panic("can't sign with senderFromServer")
```
alaibe, line 101: panic? I am not sure it is good to do any panic as we don't recover from it and that would stop the process right?

rpc/chain/rpc.go
```diff
@@ -0,0 +1,104 @@
...
+func (s *senderFromServer) Hash(tx *types.Transaction) common.Hash {
+	panic("can't sign with senderFromServer")
+}
+func (s *senderFromServer) SignatureValues(tx *types.Transaction, sig []byte) (R, S, V *big.Int, err error) {
+	panic("can't sign with senderFromServer")
+}
+
+func callFullTransactionByBlockNumberAndIndex(ctx context.Context, rpc *rpc.Client, number *big.Int, index uint) (*FullTransaction, error) {
+	var json *FullTransaction
+	err := rpc.CallContext(ctx, &json, "eth_getTransactionByBlockNumberAndIndex", toBlockNumArg(number), hexutil.Uint64(index))
```
alaibe: Only Call Context should go through the client. All the rest won't really change. I don't think 
`callFullTransactionByBlockNumberAndIndex` should use the circuit breaker, only: CallContext should

services/wallet/transfer/controller.go
```diff
@@ -178,35 +180,6 @@ func mapToList(m map[common.Address]struct{}) []common.Address {
 	return rst
 }
 
-func (c *Controller) LoadTransferByHash(ctx context.Context, rpcClient *rpc.Client, address common.Address, hash common.Hash) error {
```
alaibe, line 186: Are you sure this is not being used?

alaibe:
@dlipicar this PR is way tooooo big, this is barely possible to review correctly. Next time please split it

rpc/chain/rpc.go
dlipicar: Hmm I was simply trying to reproduce what's done for `TransactionInBlock`. Since `ethclient.Client` doesn't provide the function I need (get transaction by BlkNumber and Index, I had to write my "extension" here.

Not sure I fully understand your suggestion, do you want me to move most of this code to `FullTransactionByBlockNumberAndIndex` and only do the `CallContext` inside `makeCallSingleReturn`?

```
func (c *ClientWithFallback) TransactionInBlock(ctx context.Context, blockHash common.Hash, index uint) (*types.Transaction, error) {
	rpcstats.CountCall("eth_TransactionInBlock")

	transactions, err := c.makeCallSingleReturn(
		func() (any, error) { return c.main.TransactionInBlock(ctx, blockHash, index) },
		func() (any, error) { return c.fallback.TransactionInBlock(ctx, blockHash, index) },
		true,
	)

	if err != nil {
		return nil, err
	}

	return transactions.(*types.Transaction), nil
}

// TransactionInBlock returns a single transaction at index in the given block.
func (ec *Client) TransactionInBlock(ctx context.Context, blockHash common.Hash, index uint) (*types.Transaction, error) {
	var json *rpcTransaction
	err := ec.c.CallContext(ctx, &json, "eth_getTransactionByBlockHashAndIndex", blockHash, hexutil.Uint64(index))
	if err != nil {
		return nil, err
	}
	if json == nil {
		return nil, ethereum.NotFound
	} else if _, r, _ := json.tx.RawSignatureValues(); r == nil {
		return nil, fmt.Errorf("server returned transaction without signature")
	}
	if json.From != nil && json.BlockHash != nil {
		setSenderFromServer(json.tx, *json.From, *json.BlockHash)
	}
	return json.tx, err
}

```

rpc/chain/rpc.go
dlipicar: This is copied from `ethclient/signer.go`. Unfortunately that class is not exposed outside the module, but I want to keep the exact same behavior we have with the methods that return plain `Transaction` items.

services/wallet/transfer/controller.go
dlipicar: Not in desktop, but it is indeed used by mobile, thanks for the heads up.

dlipicar:
> @dlipicar this PR is way tooooo big, this is barely possible to review correctly. Next time please split it

Split into a couple different commits, hope it's more bearable now. Sorry about that!

rpc/chain/rpc.go
alaibe: So this is more an "extension" to the client? if that is the case then I guess it is fine to keep as is

services/wallet/transfer/commands.go
```diff
@@ -423,17 +426,87 @@ func (c *transfersCommand) Run(ctx context.Context) (err error) {
...
+			}
+			return id, nil
+		}
+	}
+
+	return NoMultiTransactionID, nil
+}
+
+func (c *transfersCommand) processMultiTransactions(ctx context.Context, allTransfers []Transfer) error {
+	// Organize subTxs by txHash
```
IvanBelyakoff: a function? :)

services/wallet/transfer/commands.go
```diff
@@ -524,11 +598,27 @@ func (c *findAndCheckBlockRangeCommand) Run(parent context.Context) (err error)
 			uniqHeaders = uniqueHeaders(allHeaders)
 		}
 
+		// Ensure only 1 erc20 transfer per transaction hash during block discovery.
```
IvanBelyakoff: make it a function?

IvanBelyakoff (approved):
LGTM. It is a huge work, both as for the project and in terms of code!

PR #3524, opened 2023-05-25
feat: a profile keypair name follows display name

saledjenic:
As part of this commit `UpdateKeypairName` endpoint added, will be used to rename all but the profile keypair.

### DIFF
--- a/multiaccounts/accounts/test_helper.go
+++ b/multiaccounts/accounts/test_helper.go
@@ -30,7 +30,7 @@ func GetWatchOnlyAccountsForTest() []*Account {
 	return []*Account{wo1, wo2, wo3}
 }
 
-func GetProfileKeypairForTest(onlyChatAndDefaultWalletAccount bool) *Keypair {
+func GetProfileKeypairForTest(includeChatAccount bool, includeDefaultWalletAccount bool, includeAdditionalAccounts bool) *Keypair {
 	kp := &Keypair{
 		KeyUID:      "0000000000000000000000000000000000000000000000000000000000000001",
 		Name:        "Profile Name",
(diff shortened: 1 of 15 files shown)


alaibe (approved):

multiaccounts/accounts/test_helper.go
```diff
@@ -30,47 +30,51 @@ func GetWatchOnlyAccountsForTest() []*Account {
...
-		PublicKey: types.Hex2Bytes("0x000000001"),
-		Name:      "Profile Name",
-		Operable:  AccountFullyOperable,
+	if includeChatAccount {
+		profileAccount := &Account{
+			Address:   types.Address{0x01},
+			KeyUID:    kp.KeyUID,
+			Wallet:    false,
+			Chat:      true,
+			Type:      AccountTypeGenerated,
```
qfrank, line 47: just noticed this change! it was empty for mobile chat account before. before this, type="" and type= AccountTypeGenerated have different meaning, now we just assume they are the same? cc @rasom @cammellos

multiaccounts/accounts/test_helper.go
```diff
@@ -30,47 +30,51 @@ func GetWatchOnlyAccountsForTest() []*Account {
...
-		Clock:     0,
-		Removed:   false,
-		Operable:  AccountFullyOperable,
+	if includeDefaultWalletAccount {
+		defaultWalletAccount := &Account{
+			Address:   types.Address{0x02},
+			KeyUID:    kp.KeyUID,
+			Wallet:    true,
+			Chat:      false,
+			Type:      AccountTypeGenerated,
```
qfrank, line 62: I have the same question as above

protocol/messenger_wallet.go
```diff
@@ -73,7 +77,30 @@ func (m *Messenger) watchWalletBalances() {
 	}()
 }
 
+func (m *Messenger) UpdateKeypairName(keyUID string, name string) error {
+	if keyUID == m.account.KeyUID && name != m.account.Name {
+		// profile keypair name must always follow profile display name
+		return ErrCannotChangeKeypairName
```
qfrank, line 83: ErrNameNotEqualAcountName sounds better? follow this way we don't need extra comment, ErrCannotChangeKeypairName is really not easy to understand for me

multiaccounts/accounts/test_helper.go
saledjenic: @qfrank that was already discussed in one of my previous PRs. 
Yes, they are the same and means that accounts with the same `AccountTypeGenerated` type are generated from a profile keypair.

protocol/messenger_wallet.go
saledjenic: Here is not a question whether is the name same or not, but more about requirement, so user cannot update profile keypair name directly, the only way to update it is by changing a display name. This error is descriptive enough and informs one who tries to use this `UpdateKeypairName` call to update a profile keypair that he "cannot change profile keypair name".

multiaccounts/accounts/test_helper.go
qfrank: so this is one thing mobile team need to align, could you summarise what extra need to be aligned for mobile team?

saledjenic: It's not only this and also I don't know what is used in which way in mobile app, but will create a note about what's changed that somebody from mobile team can go through it, analyze changes and align mobile app with them.

jakubgs:
WTF is this?
```
[nix-shell:~/workspace/status-go_prs_tests_PR-3524]$ go test -tags gowaku_skip_migrations -run ^TestTwitterLinkPreviewData$ github.com/status-im/status-go/protocol/urls/... -test.v
=== RUN   TestTwitterLinkPreviewData
-----------------------
GetLinkPreviewData: https://twitter.com/ethstatus/status/1419674733885407236
-----------------------
    urls_test.go:175: 
        	Error Trace:	/home/jenkins/workspace/status-go_prs_tests_PR-3524/protocol/urls/urls_test.go:175
        	            	invalid character '<' looking for beginning of value
        	Test:       	TestTwitterLinkPreviewData
--- FAIL: TestTwitterLinkPreviewData (0.32s)
FAIL
FAIL	github.com/status-im/status-go/protocol/urls	0.357s
FAIL
```
Why are we making requests to websites as part of unit tests? That's ridiculous.

protocol/messenger_wallet.go
Samyoul: May I express my thoughts here?

Generally the presence of any error communicates that an action failed. We make errors descriptive / have different types of error because we want to communicate why an action failed.

In this specific case, if we communicate why the action `UpdateKeypairName` failed the user of the function can make a correction to their use of the function. Otherwise we would just return an empty error, because we already know that the action failed due to the presence of an error.

The more details as to why something failed the more useful the error is at helping the user of the code to debug their issue.

saledjenic:
I created an issue for fixing a flaky test which I commented out in this PR.
Logged issue: https://github.com/status-im/status-go/issues/3529

multiaccounts/accounts/test_helper.go
saledjenic: > so this is one thing mobile team need to align, could you summarise what extra need to be aligned for mobile team?

@qfrank check this comment: https://github.com/status-im/status-go/pull/3475#issuecomment-1563932045

PR #3527, opened 2023-05-25
feat: add api to parse tx input

alaibe:
Issue [#10793](https://github.com/status-im/status-desktop/issues/10793)
Add an endpoint to parse data input and function name

### DIFF
--- a/services/wallet/thirdparty/fourbyte/client.go
+++ b/services/wallet/thirdparty/fourbyte/client.go
@@ -1,0 +1,111 @@
+package fourbyte
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"regexp"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/status-im/status-go/services/wallet/thirdparty"
+)
+
+type Signature struct {
+	ID   int    `json:"id"`
+	Text string `json:"text_signature"`
+}
+
+type ByID []Signature
+
+func (s ByID) Len() int           { return len(s) }
+func (s ByID) Less(i, j int) bool { return s[i].ID < s[j].ID }
+func (s ByID) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
+
+type SignatureList struct {
+	Count   int         `json:"count"`
+	Results []Signature `json:"results"`
+}
+
+type Client struct {
+	client *http.Client
+}
+
+func NewClient() *Client {
+	return &Client{client: &http.Client{Timeout: time.Minute}}
+}
+
+func (c *Client) DoQuery(url string) (*http.Response, error) {
+	resp, err := c.client.Get(url)
+
+	if err != nil {
+		return nil, err
+	}
+	return resp, nil
+}
+
+func (c *Client) Run(data string) (*thirdparty.DataParsed, error) {
+	if len(data) < 10 || !strings.HasPrefix(data, "0x") {
+		return nil, errors.New("input is badly formatted")
+	}
+	methodSigData := data[2:10]
+	url := fmt.Sprintf("https://www.4byte.directory/api/v1/signatures/?hex_signature=%s", methodSigData)
+	resp, err := c.DoQuery(url)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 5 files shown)


ghost:
### Pull Request Checklist

services/wallet/thirdparty/fourbyte/client.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if signatures.Count == 0 {
+		return nil, err
+	}
+
+	rgx := regexp.MustCompile(`\((.*?)\)`)
+	signature := signatures.Results[0].Text
```
saledjenic: @alaibe why do we handle only the first one, but not all signatures from the result?

services/wallet/thirdparty/fourbyte/client_test.go
```diff
@@ -0,0 +1,15 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestRun(t *testing.T) {
+	client := NewClient()
+	res, err := client.Run("0x40e8d703000000000000000000000000670dca62b3418bddd08cbc69cb4490a5a3382a9f0000000000000000000000000000000000000000000000000000000000000064")
+	require.Nil(t, err)
+	require.Equal(t, res.Signature, "processDepositQueue(address,uint256)")
+	require.Equal(t, res.Name, "processDepositQueue")
```
dlipicar, line 14: please add `Inputs` to test :D

services/wallet/thirdparty/fourbyte/client.go
dlipicar: (My understanding is) the API allows retrieving function signatures filtered by different criteria. For some of those, it makes sense to receive several results (filtering by text signature), but since we're using the hash, if there's an exact match it should be unique.

dlipicar (approved):
Awesome work! This looks highly cachable btw, not sure if we want to add it in this PR or in a separate one, but it should save us a lot of requests.

services/wallet/thirdparty/fourbyte/client.go
richard-ramos: @alaibe there are some edge cases that might happen with retrieving just the first result. for example:  the `balanceOf` function signature: https://www.4byte.directory/api/v1/signatures/?hex_signature=0x70a08231. 

Notice how using only the first result, we would display the function: `watch_tg_invmru_119a5a98(address,uint256,uint256)`, although most of the time an user would expect to see `balanceOf` instead. I think the heuristics for determining which function should be used must be changed, for example,  by using the result with the oldest id,  or attempting to decode the `data` with all the possible signatures until there's a match (however, i don't know if `method.Inputs.UnpackIntoMap` would fail if the data does not match the expected signature)

Interestingly enough, metamask only takes into account 4bytes directory when there's just 1 result: https://github.com/MetaMask/metamask-extension/blob/e3ea4f2cd044d48c61b6a35ef206fa942f3b43d1/ui/helpers/utils/transactions.util.js#L40

services/wallet/thirdparty/fourbyte/client.go
```diff
@@ -0,0 +1,91 @@
...
+	resp, err := c.client.Get(url)
+
+	if err != nil {
+		return nil, err
+	}
+	return resp, nil
+}
+
+func (c *Client) Run(data string) (*thirdparty.DataParsed, error) {
+	methodSigData := data[2:10]
```
richard-ramos, line 55: Is `data` always expected to contain a `0x` prefix, and at least 8 characters afterwards? or should a validation be added to this input

alaibe:
@dlipicar what are the request you are thinking of?

services/wallet/thirdparty/fourbyte/client.go
alaibe: you're right I should make a check to ensure it is correctly formatted

services/wallet/thirdparty/fourbyte/client.go
alaibe: Indeed i can try to check if the unpack into map fail or not first, if it does, I will then use it. If not, as u said I will use oldest ID
Good call!

alaibe:
@saledjenic @richard-ramos @dlipicar PR updated, pls look again

saledjenic (approved):

Issue #3529, opened 2023-05-25
Flaky `TestTwitterLinkPreviewData` test


saledjenic:
In this [commit](https://github.com/status-im/status-go/pull/3524/commits/e2887ab9d263f1d20f4ce0975f616192aab2ca1e) `TestTwitterLinkPreviewData` test is commented out, cause even after more than 10 tries (tests re-run) tests on CI didn't pass.

Weird thing is that test was passing in local:
```
[nix-shell:~/go/src/github.com/status-im/status-go]$ go test -tags gowaku_skip_migrations -run ^TestTwitterLinkPreviewData$ github.com/status-im/status-go/protocol/urls/... -v
=== RUN   TestTwitterLinkPreviewData
--- PASS: TestTwitterLinkPreviewData (0.81s)
PASS
ok      github.com/status-im/status-go/protocol/urls    0.813s
```

But on CI it was giving the following error:
```
jenkins@linux-04:~/workspace/status-go_prs_tests_PR-3524$ go test -tags gowaku_skip_migrations -run ^TestTwitterLinkPreviewData$ github.com/status-im/status-go/protocol/urls/... -v
=== RUN   TestTwitterLinkPreviewData
    urls_test.go:171: 
            Error Trace:    /home/jenkins/workspace/status-go_prs_tests_PR-3524/protocol/urls/urls_test.go:171
            Error:          Received unexpected error:
                            invalid character '<' looking for beginning of value
            Test:           TestTwitterLinkPreviewData
--- FAIL: TestTwitterLinkPreviewData (0.27s)
FAIL
FAIL    github.com/status-im/status-go/protocol/urls    0.275s
FAIL
```

All in all it should be fixed. 

Also one thing is vague here, what do we want to test with `TestTwitterLinkPreviewData` test?
Are we testing fetching data from remote? Or parsing the content of supported links?
Those 2 should be 2 independent tests and test for parsing the content should not depend on fetched data, but they should be mocked.

Samyoul:
@saledjenic if you want you can test in go using a mock server. I've implemented a number of tests that use go's inbuilt test server, see below for an example:

https://github.com/status-im/status-go/blob/2cac37c1fbc7eb15976dea04b7c599819df07cd9/images/decode_test.go#L89-L151

I think the biggest problem with splitting this testing is because the oembed structs hardcode their endpoints, which makes them hard to test.

[...]

PR #3530, opened 2023-05-25
Support unfurling more websites

ilmotta:
## Summary

- Related to issue https://github.com/status-im/status-mobile/issues/15918

This PR adds support for unfurling a wider range of websites. Most code changes are related to the implementation of a new `Unfurler`, an `OEmbedUnfurler`, which is necessary to get metadata for Reddit URLs using [oEmbed](oembed.com), since Reddit does not support OpenGraph meta tags. The new unfurler will also be useful for other websites, like Twitter. Also the *user agent* was changed, and now more websites consider status-go *reasonably* human.

Example hostnames that are now unfurleable: `reddit.com`, `open.spotify.com`, `music.youtube.com`

**No breaking changes in this PR** The PR in status-mobile will only bump the status-go version (not yet created).

## Other improvements

- [x] Better error handling, especially because I wasn't wrapping errors correctly. I also removed the unnecessary custom error `UnfurlErr`.
- [x] I made tests truly deterministic by parameterizing the `http.Client` instance and by customizing its `Transport` field (except for some failing conditions where it's even good to hit the real servers). I think the solution is pretty decent and highly flexible (although I hardcoded one or two things for the scope of this PR). Do take a look at the `StubTransport` type in the tests and its usages. The goal was to help our test suite not get worse than it already is in terms of reliability and speed. I know there are other places in status-go doing something similar, and I've been wondering if a more unified HTTP stubbing solution would benefit us all. Feedback is more than welcomed! :star2: 

## For the future

There is a lot more to be done, like better timeout handling, performance improvements when unfurling multiple URLs. There are even scenarios where websites such as Reddit return 429s if more than one request is made in a 2s window, which will happen if users try to unfurl more than one Reddit URL in the same message. Edge case or not, unfurling URLs is like a can of worms.

Sooner or later we'll need to add support for even more websites, like Twitter, maybe Amazon, etc. In this PR I found out that Amazon sucks (why am I surprised?). They don't support OpenGraph, nor oEmbed. They really just want product sellers to paste their own auto-generated embeddable links. So for the moment, I ignored support for Amazon. Twitter is different and requires more handling behind the scenes, even though it supports OpenGraph and oEmbed. Reddit I only added basic unfurling support, because they don't offer any thumbnail URL in their oEmbed response and they do not support OpenGraph.


protocol/linkpreview/linkpreview.go
```diff
@@ -20,72 +23,61 @@ import (
...
-	requestTimeout = 15000 * time.Millisecond
+	defaultRequestTimeout = 15000 * time.Millisecond
 
-	// Certain websites return an HTML error page if the user agent is unknown to
-	// them, e.g. IMDb.
-	defaultUserAgent = "Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/109.0"
+	// Without an user agent, many providers treat status-go as a gluttony bot,
+	// and either respond more frequently with a 429 (Too Many Requests), or
+	// simply refuse to return valid data.
+	defaultUserAgent = "status-go/v0.151.15"
```
cammellos: interesting, the two comments are a bit in contraposition :)

protocol/linkpreview/linkpreview.go
```diff
@@ -20,72 +23,61 @@ import (
...
 	defer func() {
-		if err = res.Body.Close(); err != nil {
-			logger.Error("Failed to close response body", zap.Error(err))
+		if res != nil {
+			if err = res.Body.Close(); err != nil {
+				logger.Error("failed to close response body", zap.Error(err))
+			}
 		}
 	}()
+	if err != nil {
```
cammellos, line 68: wondering if you swap this out, you still have to check `res` for nils, from the docs it seems like if `err` is `nil`, `res` is not,

so something like:

```
res, err := httpclient.Do(req)
if err != nil {
return nil, err
}

defer func() {
    if err = res.Body.Close(); err != nil {
        logger...
}
```

Also another thing, I would reinitialize the err in the defer function, golang behaves a bit inconsistently with regards of this. If we used a named parameter, the defer function would overwrite `err`, https://go.dev/play/p/0QB009PejeI . 
While if we didn't, as in this case, everything works as expected, https://go.dev/play/p/GCXcK_FTcKy 
it's a bit quirky I think, so just to be safe,
I would just do 
`if err := res.Body...`
 
Though your code is correct, it's just slightly safer in case we change to named parameters

protocol/linkpreview/linkpreview.go
```diff
@@ -136,79 +114,120 @@ func fetchThumbnail(logger *zap.Logger, url string) (common.LinkPreviewThumbnail
...
+func (u OEmbedUnfurler) unfurl() (common.LinkPreview, error) {
+	preview := newDefaultLinkPreview(u.url)
+
+	oembedURL, err := u.newOEmbedURL()
+	if err != nil {
+		return preview, err
+	}
+
+	headers := map[string]string{
+		"accept":          "application/json; charset=utf-8",
```
cammellos: maybe const?

protocol/linkpreview/linkpreview.go
```diff
@@ -136,79 +114,120 @@ func fetchThumbnail(logger *zap.Logger, url string) (common.LinkPreviewThumbnail
...
-		}
-	}()
-	if err != nil {
-		return preview, UnfurlError{
-			msg: "failed to get HTML page",
-			url: url.String(),
-			err: err,
-		}
+	headers := map[string]string{
+		"accept":          "text/html; charset=utf-8",
```
cammellos: const as well?

cammellos (approved):
Minor comments, looks great!

cammellos (approved):
Minor comments, looks great!

cammellos (approved):
Minor comments, looks great!

protocol/linkpreview/linkpreview.go
ilmotta: Oh, good catch. It should be "without a particular user agent". We really need one.

protocol/linkpreview/linkpreview.go
ilmotta: Thanks, very useful info @cammellos!

Samyoul (approved):
Lovely work @ilmotta.

Just a minor question, what impact if any does this functionality have on [protocol/urls/urls.go](https://github.com/status-im/status-go/blob/develop/protocol/urls/urls.go)?

protocol/linkpreview/linkpreview_test.go
```diff
@@ -88,106 +155,130 @@ func TestGetLinks(t *testing.T) {
...
 
-	var urls []string
-	for _, e := range examples {
-		urls = append(urls, e.url)
+	disabledStubs := false
+	if disabledStubs {
+		expected.Thumbnail.Width = 1280
+		expected.Thumbnail.Height = 720
+		expected.Thumbnail.DataURI = "data:image/jpeg;base64,/9j/4AAQSkZJRgABAQAAAQABAAD/2wCEAAUDBA8"
 	}
```
Samyoul: `disabledStubs` is always false, the if statement can never be true.

protocol/linkpreview/linkpreview_test.go
```diff
@@ -88,106 +155,130 @@ func TestGetLinks(t *testing.T) {
...
-			},
+func readAsset(t *testing.T, filename string) []byte {
+	b, err := ioutil.ReadFile("../../_assets/tests/" + filename)
+	require.NoError(t, err)
+	return b
+}
+
+func Test_UnfurlURLs_YouTube(t *testing.T) {
+	url := "https://www.youtube.com/watch?v=lE4UXdJSJM4"
+	thumbnailURL := "https://i.ytimg.com/vi/lE4UXdJSJM4/maxresdefault.jpg"
```
Samyoul, line 166: Do these tests need to consider segregating getting data from parsing data?

See https://github.com/status-im/status-go/issues/3529

protocol/linkpreview/linkpreview_test.go
ilmotta: Indeed @Samyoul, I used this flag during development, as a way of quickly switching on/off between stubbed and real HTTP calls. Since it's causing some confusion, I'll remove this.

Edit: Done in https://github.com/status-im/status-go/pull/3530/commits/20ca457dd48a245512889f079d9d629380f0abd3

protocol/linkpreview/linkpreview_test.go
ilmotta: Good question @Samyoul, but I'm inferring a little bit what you mean, please correct me if anything is off. So far, I've focused on testing only the public function `UnfurlURLs`, but because it does basically everything under the hood, its tests are more sociable than solitary.

I think I could do a better job of testing some of the private functions too, since they have a narrower responsibility, and it would be easier to test this segregation of data fetching vs parsing you mention.

ilmotta:
> Lovely work @ilmotta.
> 
> Just a minor question, what impact if any does this functionality have on [protocol/urls/urls.go](https://github.com/status-im/status-go/blob/develop/protocol/urls/urls.go)?

No impact in theory @Samyoul. In the first PR for the new unfurling implementation, I decided to not touch the `urls` package because there were too many differences and I didn't want to cause a regression on the Desktop app. Now that the Desktop team will be starting the work on using the new endpoints, eventually we'll be able to remove the `urls` implementation.

protocol/linkpreview/linkpreview_test.go
ilmotta: Slightly off topic, I saw many examples in the code about using a mock server, but in retrospect, I don't remember anymore why I didn't go that route, which in my experience (in other languages) works really well too. In the end, I opted for direct control over a RoundTripper to give me more control (?). Maybe overkill? I must say I've had good results with both approaches.

And your comments on the issue about the flaky tests are on point :)

protocol/linkpreview/linkpreview_test.go
```diff
@@ -1,14 +1,81 @@
...
+		}
+	}
+
+	if t.fallbackToDefaultTransport {
+		return http.DefaultTransport.RoundTrip(req)
+	}
+
+	return nil, fmt.Errorf("no HTTP matcher found")
+}
+
```
Samyoul, line 58: image

protocol/linkpreview/linkpreview_test.go
Samyoul: You know, I really have enjoyed reviewing this PR. Testing http requests are always tricky because the http client interface is generally embedded and segregating the getting concerns from the parsing concerns isn't a focus.

A mock client request verses a mock server response is basically removing the need for outbound calls. I've always focused on making the server mockable so that the client behaves as normally as possible during tests. But in your case I think that the server is not required because you pass in (inject) the client dependency and so you only really test the data parsing concern of your functionality.

Samyoul: Actually I may have just become a convert to client centric testing.

PR #3534, opened 2023-05-26
feat(categories): Add muted type to category muting shortcut

borismelnik:
Update and introduce `V2` method for muting category shortcut

need for [this issue in desktop](https://github.com/status-im/status-desktop/issues/9369)

### DIFF
--- a/services/ext/api.go
+++ b/services/ext/api.go
@@ -295,12 +295,12 @@ func (api *PublicAPI) DeleteChat(parent context.Context, chatID string) error {
 	return api.service.messenger.DeleteChat(chatID)
 }
 
-func (api *PublicAPI) MuteCommunityCategory(communityID string, categoryID string) error {
-	return api.service.messenger.SetMutePropertyOnChatsByCategory(communityID, categoryID, true)
+func (api *PublicAPI) MuteCommunityCategory(request *requests.MuteCategory) error {
+	return api.service.messenger.SetMutePropertyOnChatsByCategory(request, true)
 }
 
 func (api *PublicAPI) UnmuteCommunityCategory(communityID string, categoryID string) error {
-	return api.service.messenger.SetMutePropertyOnChatsByCategory(communityID, categoryID, false)
+	return api.service.messenger.SetMutePropertyOnChatsByCategory(&requests.MuteCategory{CommunityID: communityID, CategoryID: categoryID, MutedType: protocol.Unmuted}, false)
 }
 
 func (api *PublicAPI) MuteChatV2(parent context.Context, request *requests.MuteChat) (time.Time, error) {
--- a/protocol/chat.go
+++ b/protocol/chat.go
@@ -57,6 +57,7 @@ const (
 	MuteFor1Week
 	MuteTillUnmuted
 	MuteTill1Min
+	Unmuted
 )
 
 const pkStringLength = 68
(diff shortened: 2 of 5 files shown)


ghost:
### Pull Request Checklist

services/ext/api.go
```diff
@@ -294,13 +294,16 @@ func (api *PublicAPI) ActiveChats(parent context.Context) []*protocol.Chat {
 func (api *PublicAPI) DeleteChat(parent context.Context, chatID string) error {
 	return api.service.messenger.DeleteChat(chatID)
 }
+func (api *PublicAPI) MuteCommunityCategoryV2(request *requests.MuteCategory) error {
+	return api.service.messenger.SetMutePropertyOnChatsByCategory(request, true)
+}
 
 func (api *PublicAPI) MuteCommunityCategory(communityID string, categoryID string) error {
-	return api.service.messenger.SetMutePropertyOnChatsByCategory(communityID, categoryID, true)
+	return api.service.messenger.SetMutePropertyOnChatsByCategory(&requests.MuteCategory{CommunityID: communityID, CategoryID: categoryID, MutedType: 0}, true)
```
jrainville: I'd use a constant here instead

jrainville (commented):
LGTM. Just one small thing

services/ext/api.go
borismelnik: I did it in the same way as it was for `chats`, but now I realized why `0` when should be `5` (`MuteTillUnmuted`).

services/ext/api.go
```diff
@@ -294,13 +294,16 @@ func (api *PublicAPI) ActiveChats(parent context.Context) []*protocol.Chat {
 func (api *PublicAPI) DeleteChat(parent context.Context, chatID string) error {
 	return api.service.messenger.DeleteChat(chatID)
 }
+func (api *PublicAPI) MuteCommunityCategoryV2(request *requests.MuteCategory) error {
```
ilmotta: A bit of documentation on this endpoint or on the the other `MuteCommunityCategory` could be useful to explain why v2 exists, or if the original endpoint is deprecated, etc.

I'm assuming we're duplicating because we don't want to break the mobile client? Or will Desktop need v1 and v2 simultaneously?

ilmotta (approved):

services/ext/api.go
borismelnik: @ilmotta the mobile never used it? :) if no - I will remove `V2`

ilmotta: @borismelnik, let me check. I'll get back to this thread

ilmotta: @borismelnik, I can confirm `MuteCommunityCategory` is not used by the mobile client, so we can remove the duplication as you said.

borismelnik: awesome! thanks.

VERSION
```diff
@@ -1 +1 @@
-0.152.2
+0.152.3
```
jrainville: Since it's a feat, I think it's fine to increase the minor version

jrainville (approved):

VERSION
borismelnik: fixed :)

ilmotta (approved):

jrainville (approved):

jrainville (approved):

ilmotta (approved):

PR #3535, opened 2023-05-26
feat: add updated_at for syncing activity center notification

qfrank:
mainly changes:
- replace hard delete with soft delete for table `activity_center_notifications` 
- add column `updated_at` into `activity_center_notifications`

this PR is part of implementing sync activity center notification.

status: ready.


protocol/activity_center.go
```diff
@@ -65,6 +65,9 @@ type ActivityCenterNotification struct {
 	Deleted                   bool                           `json:"deleted"`
 	Accepted                  bool                           `json:"accepted"`
 	ContactVerificationStatus verification.RequestStatus     `json:"contactVerificationStatus"`
+	//Used for synchronization. Each update should increment the clock.
+	//The value should represent the time when the update occurred.
+	Clock uint64 `json:"clock"`
```
ilmotta: Suggestion: the name could be more specific. For instance `UpdatedAt`. `UpdatedAt` is a common convention to mean the column is a timestamp and the name more precisely conveys the purpose of the column. If possible, consider renaming the DB column as well.

Another point is that we already have a `timestamp` column in the table `activity_center_notifications`, so I'm thinking here, could we use it instead of adding another one? The name `timestamp` is too vague (timestamp of what?), so I'm not sure if what I'm suggesting even makes sense.

protocol/migrations/sqlite/1684979808_sync_activity_center_notifications.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE activity_center_states ADD COLUMN clock INT DEFAULT 0;
+ALTER TABLE activity_center_notifications ADD COLUMN clock INT DEFAULT 0;
```
ilmotta: Do you think we should add an index or two to cover the new column? I haven't checked the SQL queries to see if any could be impacted, so I'm not sure the lack of an index can actually be a problem.

protocol/activity_center.go
qfrank: I have also thought about the naming issue of clock before. Its meaning is not accurate enough, but almost all synchronization implementations use clock. Would you still use UpdatedAt after knowing this situation? I am open-minded about this issue :). Timestamp cannot be used as a clock because it only records the time when the record was first inserted. If it is also used to record the last updated time, it seems inappropriate.

protocol/migrations/sqlite/1684979808_sync_activity_center_notifications.up.sql
qfrank: I don't think we need add an index to cover the new column `clock` , but if you see there is a reason we need, pls tell me 🙂.

protocol/activity_center.go
ilmotta: @qfrank, I think nobody will know what just "clock" means without jumping to the struct field documentation. Even though there are other places in the code using the word clock, many are accompanied by something else to convey its purpose, like `LastUpdateClock`, `ReadMessageAtClockValue`, `NextClockAndTimestamp`, so I think we could do something similar.

> Timestamp cannot be used as a clock because it only records the time when the record was first inserted. If it is also used to record the last updated time, it seems inappropriate.

Makes complete sense.

protocol/migrations/sqlite/1684979808_sync_activity_center_notifications.up.sql
ilmotta: Thanks @qfrank, it was just a question. If you think we're good we're good :)

protocol/activity_center.go
qfrank: I'll rename it to `updated_at` 🫶

ilmotta (approved):

richard-ramos (commented):
If possible do not include changes on migrations that were not really affected by this PR to keep the history of those files cleaner, i.e: migrations/bindata.go, migrationsprevnodecfg.go, mailservers/migrations/bindata.go etc..

qfrank:
> If possible do not include changes on migrations that were not really affected by this PR to keep the history of those files cleaner, i.e: migrations/bindata.go, migrationsprevnodecfg.go, mailservers/migrations/bindata.go etc..

absolutely possible,  but `make generate` is so simple to use 😂, maybe create a local command in my IDE would be better, I'll do it @richard-ramos

MishkaRogachev (approved):
Looks good!

PR #3537, opened 2023-05-26
Add LoginAccount endpoint

cammellos:
This commit adds LoginAccount endpoint.
This makes it consistent with CreateAccount and RestoreAccount as they use similar config.

The notable difference with the previous endpoint is the API, which is the same as CreateAccount/RestoreAccount, and the fact that it will override your networks configuration, and use the credentials passed by the client.

Storing them in the config is now not needed anymore, as that's always driven from the backend, and we won't allow custom networks in the new wallet.

We can now probably start using the same 3 endpoints, CreateAccount/RestoreAccount/LoginAccount on both desktop and mobile.

### DIFF
--- a/rpc/verif_proxy.go
+++ b/rpc/verif_proxy.go
@@ -6,14 +6,15 @@ package rpc
 import (
 	"context"
 	"fmt"
-)
 
-import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/log"
+
 	gethrpc "github.com/ethereum/go-ethereum/rpc"
+
 	proxy "github.com/siphiuel/lc-proxy-wrapper"
+
 	"github.com/status-im/status-go/params"
 )
 
--- a/rpc/verif_proxy_test.go
+++ b/rpc/verif_proxy_test.go
@@ -14,16 +14,15 @@ import (
 
 	"net/http"
 	"net/http/httptest"
-)
 
-import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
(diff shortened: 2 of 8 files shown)


richard-ramos (approved):
:medal_military:

alaibe (approved):

rpc/verif_proxy_test.go
```diff
@@ -14,16 +14,15 @@ import (
...
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
-	gethrpc "github.com/ethereum/go-ethereum/rpc"
-	"github.com/status-im/status-go/params"
 
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
+
+	gethrpc "github.com/ethereum/go-ethereum/rpc"
+	"github.com/status-im/status-go/params"
```
Samyoul, line 25: Looks like your editor has jumbled up the imports for some reason.

rpc/verif_proxy.go
```diff
@@ -6,14 +6,15 @@ package rpc
...
-import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/log"
+
 	gethrpc "github.com/ethereum/go-ethereum/rpc"
+
 	proxy "github.com/siphiuel/lc-proxy-wrapper"
+
 	"github.com/status-im/status-go/params"
```
Samyoul, line 18: Same import jumbling

Samyoul (approved):
Lovely.

rpc/verif_proxy_test.go
cammellos: interesting, that's the output of `make lint-fix`

PR #3538, opened 2023-05-26
Fix panic on receiving nil message at handleSyncKeypair

rasom:
At the moment I was able to reproduce it only on mobile app. When an existing account is restored via mnemonic phrase the panic occurs.
As a malicious user can exploit this to crash the app it is a security issue as well (per @cammellos). In general panic should be avoided.

Though this likely doesn't solve the root cause of the issue :)

status: ready


ilmotta (approved):
LGTM.

The approach I usually prefer is to handle the problem closer to where it first happened. In this case, https://github.com/status-im/status-go/blob/963f9e3dbd8b93e8e16ed42b93b65acc1040e3d7/protocol/messenger.go#L4235 and https://github.com/status-im/status-go/blob/963f9e3dbd8b93e8e16ed42b93b65acc1040e3d7/protocol/messenger.go#L4220 can both get a null key pair, so I would add checks there instead and handle the situation right there. This way, the system doesn't continue in an incoherent state.

saledjenic (approved):

qfrank (approved):

PR #3539, opened 2023-05-26
Fix chats reseting when a Sync signal comes for communities

### LINKED ISSUE #10264: Primary nav badge sometimes missing
# Bug Report

- the `#desktop` channel has unread messages but the primary nav doesn't show any badge
- its parent `Desktop` category doesn't show in bold(er) either as it should

image_11-04-2023_19-35-01

### Additional Information

- Operating System: linux

jrainville:
Fixes https://github.com/status-im/status-desktop/issues/10264

The desktop issue is that the badges sometimes disappeared on both communities and chats for that community.

Turns out the issue is the SyncCommunity signal. When we receive this from the owner, we check if we are no longer in the pending state for joining, and if not, we join. However, it never checked if we are already joined, so it would re-save all the chats and that would reset all their properties to the default values.

It took me a long time to finally figure it out because it's not something I expected happened. I was more looking at the places where we actually handle the counters. Anyway, glad I could find the issue.

This also fixes the issue where `muted` would be reset.

---

FYI, the second commit is a "fix" to the CommunityImport function. At first, I thought that was the issue. That we were importing the community right before doing the store fecthes, and since we used to **not** saving the chats, we didn't register the new unread messages in the counter and then the store fetched message would just be ignored since we already imported them.

Anyway, that turned out to be a fluke, but the fix is still useful IMO and I already had created a test for that, so might as well commit it as well.

You can review per commit if you prefer.


ilmotta (approved):
LGTM, thanks for the PR context @jrainville

caybro:
LGTM, tested it and it looks fine so far

protocol/messenger_communities.go
```diff
@@ -2100,6 +2100,9 @@ func (m *Messenger) handleSyncCommunity(messageState *ReceivedMessageState, sync
 			mr, err = m.joinCommunity(context.Background(), syncCommunity.Id)
 			if err != nil {
 				logger.Debug("m.joinCommunity error", zap.Error(err))
+				if err == communities.ErrOrgAlreadyJoined {
+					return nil
+				}
 				return err
```
```suggestion
			if err != nil && err != communities.ErrOrgAlreadyJoined {
				logger.Debug("m.joinCommunity error", zap.Error(err))
				return err
```

osmaczko (approved):

protocol/communities/manager.go
```diff
@@ -1924,6 +1924,10 @@ func (m *Manager) JoinCommunity(id types.HexBytes) (*Community, error) {
 	if community == nil {
 		return nil, ErrOrgNotFound
 	}
+	if community.Joined() {
+		// Nothing to do, we are already joined
+		return community, ErrOrgAlreadyJoined
```
borismelnik, line 1929: Based on error processing in this method, we not expect that community will be returned with error?

borismelnik: Also, test not expect it, so maybe good to send `nil`?

borismelnik (approved):

PR #3540, opened 2023-05-28
feat(wallet) extract json blobs and add custom migration support

stefandunca:
### Update status-desktop [#10746](https://github.com/status-im/status-desktop/pull/new/migrate_transfers_blobs-10746)

Extended the migration process with a generic way of applying custom migration code on top of the SQL files. This is needed to extract the status from the JSON blob receipt and store it in table

Extract and migrate existing status from receipt to the activity Extract the rest of the useful receipt data

Notes:

- there is already an index on transfers table, SQLite creates one for each unique constraint, therefore, add only status to a new index
- the planned refactoring and improvements to the database have been postponed due to time constraints. Got the time to migrate the data though; extracting it can be done later for a more efficient implementation


services/wallet/activity/TODO.md
```diff
@@ -0,0 +1,157 @@
+# DB extension and refactoring for activity
```
stefandunca, line 1: This is part of the bigger work package. I will remove it after finishing the filtering work.

saledjenic (approved):

alaibe (approved):

services/wallet/activity/TODO.md
```diff
@@ -0,0 +1,157 @@
...
+Dropped benchmark performance and move on using theoretical knowledge by adding indexes for what we know only
+Will leave the performance concerns for the next milestone
+
+- [ ] Joining DBs
+- [ ] One activity DB for all require metadata
+  - Pros:
+    - Faster to query (don't know the numbers)
+    - Simpler query will decrease maintenance
+  - Cons:
+    - have to migrate all data, extract and fill the activity on every download of updates for all activities
```
IvanBelyakoff, line 150: which basically means maintenance complexity and possibly more fragmentation and inconsistency to sum it up. Not arguing which is better, just worried. I feel like we are trading one complexity for another one.

services/wallet/activity/activity.go
```diff
@@ -442,13 +441,13 @@ func GetActivityEntries(db *sql.DB, addresses []eth.Address, chainIDs []common.C
 			// TODO: extend DB with status in order to filter by status. The status has to be extracted from the receipt upon downloading
 			activityStatus := FinalizedAS
 			activityType, filteredAddress := getActivityType(dbTrType)
-			entry = NewActivityEntryWithTransaction(SimpleTransactionPT,
+			entry = NewActivityEntryWithTransaction(false,
```
IvanBelyakoff: Honestly I favor the former approach, because you always know what the parameter means when you see the function call. Using 'false' and 'true' at the moment of the call is not readable at all, unless the method is self-explanatory like `setActive(false)`. Otherwise you always guess and check the signature to know what it is.
WDYT?

IvanBelyakoff (approved):
Good job @stefandunca . I'd still appreciate a minor fix that I mentioned, though not insisting.

The bigger thing that bothers me when I see such PRs is that - do we really need to stick to SQLite that hard to make it so painful to play with JSON and do these complex migration changes?
It feels like that a more native and clear approach would be to store some data, that is not sensistive - like blockchain data - that already has JSON structure in some document DB. If what we need to do is mostly store some data and retrieve it - like receipts, logs and such things, we don't need to think how to map the data to SQL rules and use the data natively. That simplifies code, improves readability and might be even not sacrificing performance due to a key-value type of data in many cases. I'd go for it even now, considering that you have many things to do to proceed with this implementation.
Also @alaibe @saledjenic @dlipicar and whoever sees it, WDYT?

stefandunca:
> ... do we really need to stick to SQLite that hard to make it so painful to play with JSON and do these complex migration changes?

If we want to keep the compatibility and not force everyone to start from scratch then yes.

> It feels like that a more native and clear approach would be to store some data, that is not sensistive - like blockchain data - that already has JSON structure in some document DB.

AFAIK there is no nosql DB to tick all the requirements we need, searched a few times in the last years for other projects. Some of the requirements that the current approach has:
- Encryption for privacy with a good performance and without having to think at a micro-level about which property needs privacy and which doesn't disclose any
- Maturity (well known and well documented)
- Mobile fit for low power devices

There was a discussion in the team at the last offsite about upgrading to a more feature-rich DB but it was discarded due to none meeting the requirements 

> If what we need to do is mostly store some data and retrieve it - like receipts, logs and such things, we don't need to think how to map the data to SQL rules and use the data natively. 

Current implementation already provides fetching the json documents in an efficient way but it doesn't provide indexed access for filtering. This is what the current implementation improves upon and extracts the data for future use also.

>  ... data that is not sensitive ...

This is a decision hard to make and the benefits of micro-managing the data will quickly be overwhelmed by the relative simplicity of the current approach.

services/wallet/activity/activity.go
stefandunca: Fixed; please have another look.

stefandunca:
With the latest push, I updated the implementation to provide better error reporting and fixed an issue with SQLite throwing errors when storing `uint64`

appdatabase/database.go
```diff
@@ -198,11 +188,144 @@ func MigrateEnsUsernames(db *sql.DB) error {
...
+
+func migrateWalletJSONBlobs(sqlTx *sql.Tx) error {
+	var batchEntries [][]interface{}
+
+	// Extract useful information from the receipt blob and store it as sql interpretable
+	//
+	// Dropped storing postState because it was replaced by the status after EIP 658
+	// Dropped duplicating logs until we have a better way to store them in a
+	// more structured way. They can be found in transfers.receipt still
+	// Dropped tx_hash because it is redundant with transaction_details.hash
```
dlipicar: Note about this: The `hash` column in the `transfers` table is not (always) the transaction hash. Each entry in that table could either be:
- A native token (ETH) transfer. This has no associated entry in the Event Log.
- An ERC20/ERC721 token transfer. Each of these has an associated entry in the Event Log with the `Transfer` signature. 

A particular transaction (with its unique transaction hash) could either contain
- 1 ETH transfer
- 1/many ERC20/ERC721 transfers.

The `hash` column contains:
- For ETH transfers, the transaction hash. This is the one you can use on Etherscan.
- For ERC20/ERC721 transfers, a hash built by us, composed of the transaction hash and the index of the Event Log entry that originated the transfer.

The "real" transaction hash for each entry is the one contained in the receipt.

appdatabase/database.go
```diff
@@ -198,11 +188,144 @@ func MigrateEnsUsernames(db *sql.DB) error {
...
+	batchSize = 1000
+)
+
+func migrateWalletJSONBlobs(sqlTx *sql.Tx) error {
+	var batchEntries [][]interface{}
+
+	// Extract useful information from the receipt blob and store it as sql interpretable
+	//
+	// Dropped storing postState because it was replaced by the status after EIP 658
+	// Dropped duplicating logs until we have a better way to store them in a
```
dlipicar: The `log` entry points to the log that originated the ERC20/ERC721 transfer. Agree about not duplicating, but if we do not store a copy of the log, we should at least store its index (`view.go` uses it to get the actual information we need about the transfer. IMO this kind of metadata should be stored in a DB -multitransaction?- and not generated on the fly)

appdatabase/database.go
```diff
@@ -198,11 +188,144 @@ func MigrateEnsUsernames(db *sql.DB) error {
...
+	// more structured way. They can be found in transfers.receipt still
+	// Dropped tx_hash because it is redundant with transaction_details.hash
+	// Dropped the bloom filter because in SQLite is not possible to use it in an
+	// efficient manner
+	// Extract useful information from the tx blob
+	// Dropped access_list, need a separate table for it
+	// Already there chain_id, type, hash
+	// Dropped v, r, s because I see no way to be useful as BLOBs
```
dlipicar, line 230: Is the plan to eventually drop the tx and receipt columns? Or will they always be available in case we need access to the full data?

appdatabase/database.go
```diff
@@ -198,11 +188,144 @@ func MigrateEnsUsernames(db *sql.DB) error {
...
+
+	// Extract useful information from the receipt blob and store it as sql interpretable
+	//
+	// Dropped storing postState because it was replaced by the status after EIP 658
+	// Dropped duplicating logs until we have a better way to store them in a
+	// more structured way. They can be found in transfers.receipt still
+	// Dropped tx_hash because it is redundant with transaction_details.hash
+	// Dropped the bloom filter because in SQLite is not possible to use it in an
+	// efficient manner
+	// Extract useful information from the tx blob
```
dlipicar, line 224: Another fun fact: The `blk_hash` we use for `eth` transfers is wrong. It gets locally calculated using the transaction header, but it doesn't match the real one. That's the reason why the `blocks` table has a bunch of entries with the same `blk_number` but different `blk_hash`. The real block hash is the one that's in the receipt.
This is getting fixed for new entries in my Swap PR since it was causing issues, but I'm not fixing what's already on the DB.

appdatabase/database.go
```diff
@@ -198,11 +188,144 @@ func MigrateEnsUsernames(db *sql.DB) error {
...
+		rows.Close()
+		if err = rows.Err(); err != nil {
+			return err
+		}
+		mightHaveRows = (curProcessed == batchSize)
+
+		// insert extracted data into the new columns
+		if len(batchEntries) > 0 {
+			var stmt *sql.Stmt
+			stmt, err = sqlTx.Prepare(`UPDATE transfers SET status = ?, receipt_type = ?, cumulative_gas_used = ?, contract_address = ?, gas_used = ?, tx_index = ?,
```
dlipicar: not sure if by `receipt_type` you meant "type obtained from the receipt" or "type of receipt", I think the proper name for this is "Transaction Type" (https://docs.infura.io/infura/networks/ethereum/concepts/transaction-types)

appdatabase/database.go
```diff
@@ -198,11 +188,144 @@ func MigrateEnsUsernames(db *sql.DB) error {
...
+				}
+			}
+
+			if nullableTx.Valid {
+				gasPrice := sqlite.BigIntToClampedInt64(tx.GasPrice())
+				gasTipCap := sqlite.BigIntToClampedInt64(tx.GasTipCap())
+				gasFeeCap := sqlite.BigIntToClampedInt64(tx.GasFeeCap())
+				value := sqlite.BigIntToPadded128BitsStr(tx.Value())
+
+				currentRow = append(currentRow, tx.Protected(), tx.Gas(), gasPrice, gasTipCap, gasFeeCap, value, tx.Nonce(), int64(tx.Size()))
```
dlipicar: There's a couple other fields I think we won't be able to reconstruct without resorting to the JSON blobs
```
	if t.BaseGasFees != "" {
		baseFee := new(big.Int)
		baseFee.SetString(t.BaseGasFees[2:], 16)
		tip := t.Transaction.EffectiveGasTipValue(baseFee)

		view.EffectiveTip = (*hexutil.Big)(tip)
		price := new(big.Int).Add(baseFee, tip)
		view.EffectiveGasPrice = (*hexutil.Big)(price)
	}
```
```
	view.Input = hexutil.Bytes(t.Transaction.Data())
```
```
		if t.Transaction.To() != nil {
			view.To = *t.Transaction.To()
		}
```

sqlite/fields_test.go
```diff
@@ -0,0 +1,59 @@
...
+			input:    big.NewInt(0),
+			expected: strToPtr("00000000000000000000000000000000"),
+		},
+		{
+			name:     "case very large",
+			input:    new(big.Int).Exp(big.NewInt(10), big.NewInt(26), nil),
+			expected: strToPtr("000000000052b7d2dcc80cd2e4000000"),
+		},
+		{
+			name:     "case max",
```
dlipicar, line 36: please add a test with a negative number

appdatabase/database.go
stefandunca: Thanks for the heads up @dlipicar. So the hash we have is our own generated has (transfer entry ID). Will add tx_has moving forward.

appdatabase/database.go
stefandunca: > The log entry points to the log that originated the ERC20/ERC721 transfer. Agree about not duplicating, but if we do not store a copy of the log, we should at least store its index (view.go uses it to get the actual information we need about the transfer

Indeed log entry is duplicated; however, it makes sense to migrate the index now. WDYT?

> IMO this kind of metadata should be stored in a DB -multitransaction?- and not generated on the fly)

We discussed refactoring the entire data; however, given the time constraints, I decided that this would be too much work that won't benefit the short-term activity feature and leave it for some further reorganization based on the new requirements later on.

Regarding why no multi-transaction for the new fields is because not all transactions in transfers are recorded in multi-transaction and for the current work there is not much difference.

appdatabase/database.go
stefandunca: Not sure yet. Ideally, a proper reorganization would be required with the main transaction information we use in the `transactions` table (now transfers) and everything else linked to it, including the raw data if we require it for future extension. The rest of the aggregated data is required by the activity in the activity (now multi-transaction) table.

As discussed last time we met, the complete re-organization is more consistent and would be better done without migration and before the wide release.

appdatabase/database.go
stefandunca: I extracted the original block hash and added it as a new `tx_hash` column.

sqlite/fields_test.go
stefandunca: That is undefined behaviour and should not be attempted. Currently, we don't provide negative values but only convert unsigned data, will add a comment to clarify this.

Samyoul:
Before I begin my review of this code I wanted to add my support to your points here https://github.com/status-im/status-go/pull/3540#issuecomment-1568352817 .

And particularly to agree that If we could find a nosql db that does meet the requirements highlighted this is something that we'd happily consider.

appdatabase/migrations/sql/1685134628_transfers_receipt_json_blob_out.up.sql
```diff
@@ -0,0 +1,23 @@
...
+-- ALTER TABLE transfers ADD COLUMN gas_limit UNSIGNED INT;
+-- ALTER TABLE transfers ADD COLUMN gas_price_clamped64 INT;
+-- ALTER TABLE transfers ADD COLUMN gas_tip_cap_clamped64 INT;
+-- ALTER TABLE transfers ADD COLUMN gas_fee_cap_clamped64 INT;
+-- ALTER TABLE transfers ADD COLUMN amount_padded128hex CHAR(32);
+-- ALTER TABLE transfers ADD COLUMN account_nonce INT;
+-- ALTER TABLE transfers ADD COLUMN size INT;
+
+-- CREATE INDEX transfers_status ON transfers (status);
+
```
Samyoul: I'm curious why this SQL migration is done in go code? Normally schema changes are done via migrations. I'm not saying we shouldn't, I am just wondering.

services/wallet/activity/TODO.md
Samyoul: This is a nice document @stefandunca

appdatabase/migrations/sql/1685134628_transfers_receipt_json_blob_out.down.sql
```diff
@@ -0,0 +1 @@
+-- This migration is in GO code. If GO migration failure this serves to rollback to the previous one
```
Samyoul: Just a note about our migrations, we don't ever roll a migration back. This is because we can't control the clients and so have no control over what version the users have.

Samyoul (approved):
Looks really good. Thank you for this work @stefandunca

appdatabase/database.go
stefandunca: I mean, the type from the type filed from the receipt `Receipt.Type`

stefandunca: > ... I think the proper name for this is "Transaction Type" (https://docs.infura.io/infura/networks/ethereum/concepts/transaction-types)

My understanding is that the Transaction Type is retrieved by the `Transaction.Type()` call which extracts it from `TxData.txType` and which has a different meaning than the type found in the receipt.

appdatabase/database.go
IvanBelyakoff: I would split ETH and ERC20 transfers into different tables. They are different in nature, ERC20's hash is calculated differently for our need, it has logs and maybe something else that I forgot to mention. I don't think there is a strong need to keep them in the same table, while splitting them may simplify tailoring the tables for specifics of these transfer types and easier migration, extension, maintenance. The drawback would be JOINing tables on queries. And in code we mostly handle them differently as well AFAIK, so since we are refactoring, imo we should split them.

stefandunca: > I would split ETH and ERC20 transfers into different tables.

@IvanBelyakoff I agree that a more extensive reorganization would fit the latest requirements in a better way. However, this PR already does more than its intended scope, so let's pick up this idea after the activity work is finished, which can be served with the changes in this PR.

appdatabase/database.go
stefandunca: I had a quick look, and this seems to be a task for the extensive DB reorganization we talked about recently that won't require data migration.

appdatabase/migrations/sql/1685134628_transfers_receipt_json_blob_out.up.sql
stefandunca: I tried to explain it in the comments here: https://github.com/status-im/status-go/blob/d8b24ca160bd52df99d49dc09a903d3cd6914590/sqlite/migrate.go#L26

The migration API doesn't support rolling back to the previous migration to leave the DB in a consistent state (in case of custom golang migration failing). Therefore, I use migration down to roll back to the previous migration (before the custom one). The `down` migration for the current custom migration can be done in SQL. Still, it is more complicated than a simple `tx.Rollback` in the go code (have to remove rows, create a temporary DB to drop the new columns and recopy data). However, for another type of migration, providing an `up` SQL file and a `down` SQL file that discards changes made by the `up` one, might be a better idea and it will work the same.

Hope this clarifies your concerns.

appdatabase/migrations/sql/1685134628_transfers_receipt_json_blob_out.down.sql
stefandunca: As mentioned in the previous answer, this is required only in the case of custom migrations as an alternative to the API limitations of the migration system to unroll the changes made by the custom code in case of failures. Hence, we can know if a Golang failed, rolled-back, or wasn't run. Otherwise, the DB migration will mark the current custom version as run after running the SQL "marker" migration. Hence after fixing the custom migration, it won't be rerun, and to fix will have to create another migration. Long story short, it is a feature to make fixing Golang migrations easy and optional.

stefandunca: With this approach, we have a relatively simple and robust way of running Golang migrations along when there is no way around it.

appdatabase/database.go
stefandunca: With the latest changes I migrated also the `log_index` by extracting it from the log column.

appdatabase/migrations/sql/1685134628_transfers_receipt_json_blob_out.down.sql
Samyoul: Ah ok, @stefandunca I see. Rollback in the event the go code fails, that makes sense to me. Thank you for your patience in explaining this to me.

stefandunca:
@dlipicar I extracted `token_address`, `token_id` and updated `type` columns in the migration process. Please have another look.
@anastasiyaig could you please retest the latest changes once again?

stefandunca:
Rebasing on the latest master fails migration tests, looking into it right now. It worked on the morning master though.

PR #3542, opened 2023-05-29
ci: add timeouts to all Jenkinsfiles

jakubgs:
This is especially important for tests.


richard-ramos (approved):

jakubgs:
The tests log takes up nearly 50 MB and is nearly 260k lines long. That's ridiculous...

PR #3544, opened 2023-05-29
fix spiff worflow data dir

### LINKED ISSUE #3541: Spiff worlflow issue with nickname
We are using go-waku with SpiffWorkflow to send notifications from the workflow engine to Status Desktop. We set the nickname of Waku node however it seems to forget this on redeployment.

This is apparent when we have a new status desktop account getting a contact request. For those people who have already accepted the contact request, it will stay as-is. This is an issue for us since new people using Spiff will not be able to identify who is sending the contact request, as it is random. We normally call the nickname API and set the name to "SpiffWorkflow".

We do have persistent storage so not sure why this is happening. Can chat to @jakubgs and also @burnettk about this more.

cammellos:
allow data dir to be set from command line

Resolves: https://github.com/status-im/status-go/issues/3541

### DIFF
--- a/cmd/spiff-workflow/main.go
+++ b/cmd/spiff-workflow/main.go
@@ -324,9 +324,9 @@ func defaultNodeConfig(installationID string) (*params.NodeConfig, error) {
 }
 
 func ImportAccount(seedPhrase string, backend *api.GethStatusBackend) error {
-	backend.UpdateRootDataDir("./tmp")
+	backend.UpdateRootDataDir(*dataDir)
 	manager := backend.AccountManager()
-	if err := manager.InitKeystore("./tmp"); err != nil {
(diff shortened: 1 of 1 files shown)


jakubgs (approved):
Thanks man.

jakubgs:
I have deployed this on `app-dev` environment with this config fix:

- https://github.com/status-im/infra-spiff-workflow/commit/7c63c10a - spiff-workflow: re-add -dir flag for waku node

And it appears to work:
```
admin@node-01.sm-eu-zur1.spiff.app-dev:/docker/spiff-workflow % sudo find wakunode/data | grep accounts
wakunode/data/accounts.sql-shm
wakunode/data/accounts.sql
wakunode/data/accounts.sql-wal
```

PR #3545, opened 2023-05-29
fix(messenger_handler): fix group chat not being set as active

### LINKED ISSUE #10818: [messenger] group chat keeps disappearing
# Bug Report
After every fresh start, the group chat is not visible. It only reappears later through the runtime when there's some new messages. You can see below that I even get notifications on startup from the said "UI Team" group chat.

image

And later it's back (after a new message arrived there):
image

### Additional Information

- Operating System: linux

jrainville:
Fixes https://github.com/status-im/status-desktop/issues/10818

Fixes an issue where if a group chat was first received from a non-contact, and later received from a contact, it still wouldn't save it as active.

That's because we checked if we were **newly** added instead of just if we were added. That meant that in the case I described above, the chat would then never have the chance to be set active.


Parveshdhull (approved):

caybro (approved):
Tested to work on my PC

osmaczko (approved):
:+1:

PR #3548, opened 2023-05-30
fix(databaseLocks): Fixing database lock errors on transactions

alexjba:
### Fixing: https://github.com/status-im/status-desktop/issues/10838

Repro steps:
```
tx, err := db.Begin() // Create a new transaction
err = tx.QueryRow("SELECT COUNT(*) FROM test").Scan(&count) // Read the DB
_, err = tx.Exec("INSERT INTO test (data) values(?)", dataToWrite) // Writing the DB results in "database is locked" error
```

The default transaction lock is `deferred`. This means that the transaction will automatically become read or write transaction based on the first DB operation. In case the first operation is `SELECT` the transaction becomes read transaction, otherwise write transaction. When a read transaction tries to write the DB sqlite will promote the transaction to a write transaction if there is no other transaction that holds a lock. When the promotion fails `database is locked` error is returned. The error is returned immediately and does not use the busy handler. In our case almost all read transaction would fail with `database is locked` error.

This fix is changing the default transaction lock to `IMMEDIATE`. It translates to `BEGIN IMMEDIATE` instead of `BEGIN`. In this mode, the transaction will be created as a write transaction no matter what DB operation will run as part of the transaction. The write transaction will try to obtain the DB lock immediately when `BEGIN IMMEDIATE` is called and the busy handler is used when the DB is locked by other transaction.

### DIFF
--- a/sqlite/sqlite.go
+++ b/sqlite/sqlite.go
@@ -5,8 +5,10 @@ import (
 	"database/sql/driver"
 	"errors"
 	"fmt"
+	"net/url"
 	"os"
 	"runtime"
+	"strings"
 
 	sqlcipher "github.com/mutecomm/go-sqlcipher" // We require go sqlcipher that overrides default implementation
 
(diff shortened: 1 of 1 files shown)


osmaczko:
> This fix is changing the default transaction lock to IMMEDIATE. It translates to BEGIN IMMEDIATE instead of BEGIN. In this mode, the transaction will be created as a write transaction no matter what DB operation will run as part of the transaction. The write transaction will try to obtain the DB lock immediately when BEGIN IMMEDIATE is called and the busy handler is used when the DB is locked by other transaction.

It is also worth noting that `_txlock=immediate` will not block read operations:
```
./build/status-go-db-playground
2023/05/31 08:26:01 Setting up db connection
2023/05/31 08:26:01 A: db.Begin()
2023/05/31 08:26:01 A: db.Begin() return
2023/05/31 08:26:01 A: tx.QueryRow(SELECT..)
2023/05/31 08:26:01 A: tx.QueryRow(SELECT..) return
2023/05/31 08:26:01 B: db.QueryRow()
2023/05/31 08:26:01 Setting up db connection
2023/05/31 08:26:01 B: db.QueryRow() return
2023/05/31 08:26:01 A: tx.Exec(UPDATE..)
2023/05/31 08:26:01 A: tx.Exec(UPDATE..) return
2023/05/31 08:26:01 A: tx.Commit()
2023/05/31 08:26:01 A: tx.Commit() return
```

Although, it will block on each `db.Begin()` if any other transaction is ongoing, even if transaction doesn't modify the database.

osmaczko (approved):

saledjenic (approved):

PR #3549, opened 2023-05-30
feat(communities): Adds muting interval to community

borismelnik:
Add muting interval to communities

Important changes:

Closes #


ghost:
### Pull Request Checklist

protocol/messenger_communities.go
```diff
@@ -528,8 +549,32 @@ func (m *Messenger) SpectateCommunity(communityID types.HexBytes) (*MessengerRes
...
+		MuteTill = time.Now().Add(MuteFor1WeekDuration)
+	default:
+		MuteTill = time.Time{}
+	}
+	muteTillTimeRemoveMs, err := time.Parse("2006-01-02T15:04:05Z", MuteTill.Format("2006-01-02T15:04:05Z"))
+
+	if err != nil {
+		return err
+	}
+	return m.communitiesManager.SetMuted(request.CommunityID, muted, muteTillTimeRemoveMs)
```
jrainville: I think if we are unmuting (so `mute == false`), we could skip the whole time parsing, since there is no time on unmuting

jrainville (commented):
Looks good. One small comment

protocol/requests/mute_community_request.go
```diff
@@ -0,0 +1,26 @@
...
+	"errors"
+
+	"github.com/status-im/status-go/eth-node/types"
+)
+
+var ErrInvalidMuteCommunityParams = errors.New("mute-community: invalid params")
+
+type MuteCommunity struct {
+	CommunityID types.HexBytes `json:"id"`
+	MutedType   MutingVariation
```
ilmotta: Is the missing JSON tag on `MutedType` intentional? If not, it would be better if we added `json:"mutedType"`

protocol/messenger_communities.go
```diff
@@ -276,6 +276,27 @@ func (m *Messenger) updateCommunitiesActiveMembersPeriodically() {
 	}()
 }
 
+func (m *Messenger) CheckCommunitiesToUnmute(response *MessengerResponse) {
```
ilmotta: It would be good to get at least a basic test coverage for this public function.

protocol/messenger_communities.go
```diff
@@ -276,6 +276,27 @@ func (m *Messenger) updateCommunitiesActiveMembersPeriodically() {
...
+	m.logger.Debug("watching communities to unmute")
+	communities, err := m.communitiesManager.All()
+	if err != nil {
+		m.logger.Info("err", zap.Any("Couldn't get all communities", err))
+		return
```
ilmotta: Another approach I like is to return a wrapped error instead of void, and leave the logging call to upstream functions (in this case `protocol.watchChatsAndCommunitiesToUnmute`). The idea is that returning an error brings more flexibility for callers to decide what to do, whereas a void function is more _opaque_. It also goes a long way to help with testing.

Something along these lines:

```go
if err != nil {
	return fmt.Errorf("couldn't get all communities: %w", err)
}
```

ilmotta (approved):

protocol/messenger_communities.go
borismelnik: Got it :) thanks!

protocol/requests/mute_community_request.go
borismelnik: I agree :)

borismelnik:
@jrainville @ilmotta just FYI, I will be sure that chat muting functionality works well and then will update this PR. Got all your suggestions and will fix it soon.

jrainville (approved):

VERSION
```diff
@@ -1 +1 @@
-0.156.0
+0.157.0
```
ibrkhalil: Shouldn't it be `0.156.1`?

ibrkhalil (approved):
Thanks for the wonderful work!

VERSION
borismelnik: as I understand its a `feature` and should up minor version

ilmotta (approved):

VERSION
ibrkhalil: Didn't know, Thanks!

protocol/messenger_communities.go
```diff
@@ -574,8 +595,35 @@ func (m *Messenger) SpectateCommunity(communityID types.HexBytes) (*MessengerRes
...
-func (m *Messenger) SetMuted(communityID types.HexBytes, muted bool) error {
-	return m.communitiesManager.SetMuted(communityID, muted)
+func (m *Messenger) SetMuted(request *requests.MuteCommunity) error {
+	if err := request.Validate(); err != nil {
+		return err
+	}
+	if request.MutedType == Unmuted {
+		return m.communitiesManager.SetMuted(request.CommunityID, false, time.Time{})
+	}
+	var MuteTill time.Time
```
jo-mut, line 605: this code is share between mute community and mute chats. Your could have the code into a separate function and call the function to get the MuteTill value 
```golang
func (m *Messenger) MuteDuration(mutedType requests.MutingVariation) (time.Time, error) {
	var MuteTill time.Time

	switch mutedType {
	case MuteTill1Min:
		MuteTill = time.Now().Add(MuteFor1MinDuration)
	case MuteFor15Min:
		MuteTill = time.Now().Add(MuteFor15MinsDuration)
	case MuteFor1Hr:
		MuteTill = time.Now().Add(MuteFor1HrsDuration)
	case MuteFor8Hr:
		MuteTill = time.Now().Add(MuteFor8HrsDuration)
	case MuteFor1Week:
		MuteTill = time.Now().Add(MuteFor1WeekDuration)
	default:
		MuteTill = time.Time{}
	}

	muteTillTimeRemoveMs, err := time.Parse(time.RFC3339, MuteTill.Format(time.RFC3339))
	if err != nil {
		return time.Time{}, err
	}

	return muteTillTimeRemoveMs, nil
}
```

jo-mut:
@borismelnik pr looks good.

protocol/messenger_communities.go
jo-mut: you can also just merge the pr as this refactor is already done in my pr

jo-mut (approved):

PR #3550, opened 2023-05-30
waku2: static shards

richard-ramos:
Communities can be assigned to an specific shard cluster and index.
To do so, after the community is created, in the node management section the following json rpc method could be executed:
```
{"method": "wakuext_setCommunityShard", "params": ["0xCommunityID", 16, 2]}
```
- The community id can be obtained with `{"method": "wakuext_communities"}` 
- The second parameter is the shard cluster. The [RFC](https://rfc.vac.dev/spec/57/) suggest using 16 for status communities
- The third parameter is the shard index. Should be a number between 0 and 1023. The [RFC](https://rfc.vac.dev/spec/57/) has some suggestions in which number should be used.

Once this method is executed, you can expect to receive the following response:
```json
{
  "jsonrpc":"2.0",
  "id":0,
  "result":{
    "protectedTopic":[
      {"pubsubTopic":"/waku/2/rs/<cluster>/<index>","publicKey":"0x......"}
    ]
  }
}
```
This information can be used to setup the fleet nodes to indicate which pubsub topic should be protected and which pubkey to use.

Notice that the code for obtaining the communities description from the smart contract and from NFTs is temporarily commented as these do not contain information that would let us determine the cluster and index assigned to a community


richard-ramos:
The rebase for this PR is somewhat annoying. I'm doing it on a separate PR and once i do a quick test in status-desktop and see that everything's fine, i'll reset this branch with https://github.com/status-im/status-go/pull/3944

cc: @ilmotta

richard-ramos:
Replaced by https://github.com/status-im/status-go/pull/3944

PR #3551, opened 2023-05-31
feat(wallet): SequentialFetchStrategy improvements

IvanBelyakoff:
- reverted a change that stopped looking for ERC20 transfers if no nonce and balance change found within a block range for ETH
- implemented sending EventRecentHistoryReady event at a proper time
- moved EventFetchingRecentHistory event to Strategy type as it does not make sense to send this event in loop
- moved iterating through blocks logic to inside of `loadTransfers` command, which now accepts a block range. It allows to handle each accounts transfer commands completion separately and fire event on finished per account basis
- reused `uniqueHeaders` function in commands.go
- cleaned up

### DIFF
--- a/services/wallet/transfer/commands.go
+++ b/services/wallet/transfer/commands.go
@@ -47,7 +47,6 @@ var (
 )
 
 type ethHistoricalCommand struct {
-	eth          Downloader
 	address      common.Address
 	chainClient  *chain.ClientWithFallback
 	balanceCache *balanceCache
@@ -69,7 +68,8 @@ func (c *ethHistoricalCommand) Command() async.Command {
 }
 
 func (c *ethHistoricalCommand) Run(ctx context.Context) (err error) {
-	log.Info("eth historical downloader start", "address", c.address, "from", c.from.Number, "to", c.to, "noLimit", c.noLimit)
+	log.Info("eth historical downloader start", "chainID", c.chainClient.ChainID, "address", c.address,
+		"from", c.from.Number, "to", c.to, "noLimit", c.noLimit)
 
 	start := time.Now()
 	if c.from.Number != nil && c.from.Balance != nil {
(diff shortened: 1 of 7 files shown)


alaibe (approved):

services/wallet/transfer/commands.go
```diff
@@ -485,22 +519,9 @@ func (c *findAndCheckBlockRangeCommand) Run(parent context.Context) (err error)
...
-				uniqHeadersByHash[header.Hash] = uniqHeader
-			} else {
-				uniqHeadersByHash[header.Hash] = header
-			}
-		}
-
 		uniqHeaders := []*DBHeader{}
-		for _, header := range uniqHeadersByHash {
-			uniqHeaders = append(uniqHeaders, header)
+		if len(allHeaders) > 0 {
```
dlipicar, line 523: doesn't seem like this check is needed

dlipicar (commented):
Not sure if I'm too tired, but I cannot find this change
`reverted a change that stopped looking for ERC20 transfers if no nonce and balance change found within a block range for ETH`

----

OK, I'm too tired, disregard this. GitHub was hiding a file with large diff.

services/wallet/transfer/commands_sequential.go
```diff
@@ -109,29 +113,28 @@ func (c *findBlocksCommand) checkRange(parent context.Context, from *big.Int, to
...
+		return nil, nil
+	}
 
-		allHeaders := append(ethHeaders, erc20Headers...)
+	allHeaders := append(ethHeaders, erc20Headers...)
 
-		if len(allHeaders) > 0 {
-			foundHeaders = uniqueHeaders(allHeaders)
-		}
+	if len(allHeaders) > 0 {
```
dlipicar, line 136: ditto

services/wallet/transfer/commands_sequential.go
```diff
@@ -186,17 +202,11 @@ func (c *findBlocksCommand) fastIndex(ctx context.Context, bCache *balanceCache,
 		chainClient:  c.chainClient,
 		balanceCache: bCache,
 		address:      c.account,
-		eth: &ETHDownloader{
```
dlipicar, line 189: oh wow, this was unused?

dlipicar (approved):

services/wallet/transfer/commands.go
IvanBelyakoff: Well, it is not necessary of course, on the other hand, we save a call, an unnecessary creation of a map and a slice and a couple of more checks, so it is more on the good side than on the bad imo

services/wallet/transfer/commands_sequential.go
IvanBelyakoff: replied above, I'd prefer to keep it

PR #3552, opened 2023-05-31
fix: profile social links updated

saledjenic:
- in the old implementation some handlings were broken, in terms that context was mixed between backed up and social links messages
- new requirements allows adding multiple links of the same type
- each change over social links is broadcasted to paired devices


protocol/messenger_identity.go
```diff
@@ -163,9 +163,9 @@ func (m *Messenger) SetSocialLinks(socialLinks *identity.SocialLinks) error {
 	}
 
 	err = m.withChatClock(func(chatID string, clock uint64) error {
-		for _, link := range *socialLinks {
-			if link.Clock == 0 {
-				link.Clock = clock
```
qfrank, line 168: I can’t believe this is the code I wrote. I even fixed similar code written by someone else before. 😂 Thank you 🙂

qfrank (approved):

saledjenic:
@qfrank I re-requested your review on this, cause at the end it turned out that the requirements have changed and I had to redo it a bit.

protocol/messenger_identity.go
```diff
@@ -158,40 +184,85 @@ func (m *Messenger) SetSocialLinks(socialLinks *identity.SocialLinks) error {
...
+	err := m.validateAndResolveClockForSocialLinks(socialLinks)
+	if err != nil {
 		return err
 	}
 
-	if err = m.resetLastPublishedTimeForChatIdentity(); err != nil {
+	err = m.settings.UpdateSocialLinkAtPosition(position, newText, newURL, (*socialLinks)[0].Clock)
+	if err != nil {
+		return err
+	}
```
osmaczko: I believe the order should be opposite. 

#### Right now:
- validateAndResolveClockForSocialLinks -> publishContactCode
- settings.UpdateSocialLinkAtPosition

#### As it should be:
- settings.UpdateSocialLinkAtPosition
- publishContactCode

The issue with "Right now" approach is that it will propagate not updated social links.

protocol/protobuf/pairing.proto
```diff
@@ -322,7 +323,7 @@ message BackedUpProfile {
   string display_name = 2;
   uint64 display_name_clock = 3;
   repeated SyncProfilePicture pictures = 4;
-  repeated SyncSocialLinkSetting social_link_settings = 5;
+  repeated SocialLink social_links = 5;
   repeated SyncEnsUsernameDetail ens_username_details = 6;
```
Samyoul: This will cause a breaking change between versions. Older versions of `status-go` will expect position `5` to be of type `SyncSocialLinkSetting` and parse `SocialLink ` data as a `SyncSocialLinkSetting`.

Frustrating I know. The fix is to add a new entry on an unused position for type `repeated SocialLink`

saledjenic: @Samyoul thanks for the heads-up, I know for that, but opted in for this approach cause it is not not used anywhere. Mobile app doesn't use it yet and we're implementing this for the desktop app for the first time now. Want to say since syncing among devices was initially (which was recently) implemented on the status-go side was not in use.

saledjenic:
I set do not merge label as we need to sync all social links on every change. 
That's decided in `make sure everything that needs syncing is synced` group chat. :)

osmaczko (commented):
I'm not sure if this will be relevant to the new approach, but I'd like to point out a small issue with the current codebase.

protocol/protobuf/pairing.proto
Samyoul: Ah, in that case go ahead.

qfrank:
> @qfrank I re-requested your review on this, cause at the end it turned out that the requirements have changed and I had to redo it a bit.

is it ready to review now? cause i saw there was a dicussion on the solution of how to sync @saledjenic

saledjenic:
A changes we've discussed are applied. Now all social links are replaced in a single go if anything has changed to any of them, including their order. Sync clock is maintained in `social_links` column of `settings_sync_clock` table.

Samyoul (approved):
This is a nice implementation @saledjenic , you're a very thorough developer.

appdatabase/migrations/sql/1685880973_add_profile_links_settings_table.up.sql
```diff
@@ -0,0 +1,30 @@
...
+-- after discussion https://github.com/status-im/status-go/pull/3552#issuecomment-1573817044
+-- we decided to replace the entire db table content on every change (because of that position column is not needed)
+CREATE TABLE IF NOT EXISTS profile_social_links (
+  text VARCHAR NOT NULL CHECK (length(trim(text)) > 0),
+  url VARCHAR NOT NULL CHECK (length(trim(url)) > 0),
+  PRIMARY KEY (text, url)
+);
+
+-- Create new column to keep clock of the last change of profile social links (since content is replaced all row share the same clock value)
+ALTER TABLE settings_sync_clock ADD COLUMN social_links INTEGER NOT NULL DEFAULT 0;
```
qfrank, line 12: i feel not good while we adding a column `social_links` into `settings_sync_clock` but we don't use a corresponding field from table `settings` , it would be better to create a separate table like `activity_center_states` although it doesn't contain column `clock`

multiaccounts/settings_social_links/database.go
```diff
@@ -89,57 +89,53 @@ func (s *SocialLinksSettings) SetSocialLinks(links *identity.SocialLinks) (err e
...
 
-	stmt, err := tx.Prepare("UPDATE social_links_settings SET link_text = ?, link_url = ?, clock = ? WHERE link_text = ? AND clock < ?")
-	if err != nil {
-		return err
-	}
-	_, err = stmt.Exec(link.Text, link.URL, link.Clock, link.Text, link.Clock)
-	if err != nil {
-		return err
-	}
-	return stmt.Close()
```
qfrank, line 144: should we keep `stmt.Close()`?

multiaccounts/settings_social_links/database_test.go
```diff
@@ -27,91 +27,207 @@ func openTestDB(t *testing.T) (*SocialLinksSettings, func()) {
...
+	require.Len(t, links, 0)
+
+	clock := uint64(1)
+	// add profile social links
+	profileSocialLinks := profileSocialLinks()
+	err = socialLinkSettings.AddOrReplaceSocialLinksIfNewer(profileSocialLinks, clock)
+	require.NoError(t, err)
+
+	clock = 2
+	// test sociial link update
```
qfrank: social

protocol/messenger.go
```diff
@@ -6397,62 +6397,71 @@ func (m *Messenger) syncDeleteForMeMessage(ctx context.Context, rawMessageDispat
...
-			return err
-		}
+
+	dbClock, err := m.settings.GetSocialLinksClock()
+	if err != nil {
+		return err
 	}
-	return nil
+
+	_, chat := m.getLastClockWithRelatedChat()
```
qfrank, line 6415: if we only need `chat`, it would be better to use other method rather than `getLastClockWithRelatedChat`

appdatabase/migrations/sql/1685880973_add_profile_links_settings_table.up.sql
```diff
@@ -0,0 +1,30 @@
+-- Create new table where we will store profile social links,
+-- after discussion https://github.com/status-im/status-go/pull/3552#issuecomment-1573817044
+-- we decided to replace the entire db table content on every change (because of that position column is not needed)
+CREATE TABLE IF NOT EXISTS profile_social_links (
+  text VARCHAR NOT NULL CHECK (length(trim(text)) > 0),
+  url VARCHAR NOT NULL CHECK (length(trim(url)) > 0),
+  PRIMARY KEY (text, url)
+);
```
osmaczko: Just wondering, don't we need to maintain `position` anymore?

AFAIK the order of data returned by a SELECT statement is not guaranteed to be in the order of insertion or any other order.

osmaczko (approved):
Nice :+1:

multiaccounts/settings_social_links/database.go
saledjenic: it's already there line `119`

appdatabase/migrations/sql/1685880973_add_profile_links_settings_table.up.sql
saledjenic: Why I opted in for table, cause table was already there 🙂 check old `social_links_settings` they look more less the same, just that new one has more appropriate name doesn't have a `clock` column.

appdatabase/migrations/sql/1685880973_add_profile_links_settings_table.up.sql
saledjenic: Yes, was thinking about the same, had `position` in previous implementation, but when I was changing the logic completely to what we've discussed (sending everything on each change) and since it was not added in an implementation which was here even before I completely forgot about it. Very well spotted, thanks! :)

saledjenic: Will add it now.

PR #3553, opened 2023-05-31
Make profile photos visible by default

### LINKED ISSUE #16060: Make profile photos visible by default
There is a setting in the application to enable/disable showing profile picture to all other users by default
and another similar setting to enable/disable showing profile pictures from everyone else.

Currently both of these are disabled by default. There is a request to enable them by default

cc @churik

alwx:
Fixes https://github.com/status-im/status-mobile/issues/16060

### DIFF
--- a/api/defaults.go
+++ b/api/defaults.go
@@ -33,7 +33,8 @@ func defaultSettings(generatedAccountInfo generator.GeneratedAccountInfo, derive
 	s.BackupEnabled = true
 	logLevel := "INFO"
 	s.LogLevel = &logLevel
-	s.ProfilePicturesShowTo = settings.ProfilePicturesShowToContactsOnly
+	s.ProfilePicturesShowTo = settings.ProfilePicturesShowToEveryone
+	s.ProfilePicturesVisibility = settings.ProfilePicturesVisibilityEveryone
 	s.KeyUID = generatedAccountInfo.KeyUID
 	s.Address = types.HexToAddress(generatedAccountInfo.Address)
 	s.WalletRootAddress = types.HexToAddress(derivedAddresses[pathWalletRoot].Address)
@@ -62,7 +63,6 @@ func defaultSettings(generatedAccountInfo generator.GeneratedAccountInfo, derive
 
 	s.PreviewPrivacy = true
 	s.Currency = "usd"
-	s.ProfilePicturesVisibility = 1
 	s.LinkPreviewRequestEnabled = true
 
 	visibleTokens := make(map[string][]string)
--- a/cmd/populate-db/main.go
+++ b/cmd/populate-db/main.go
@@ -330,37 +330,37 @@ var paths = []string{pathWalletRoot, pathEIP1581, pathDefaultChat, pathDefaultWa
 func defaultSettings(generatedAccountInfo generator.GeneratedAccountInfo, derivedAddresses map[string]generator.AccountInfo, mnemonic *string) (*settings.Settings, error) {
 	chatKeyString := derivedAddresses[pathDefaultChat].PublicKey
 
-	settings := &settings.Settings{}
-	settings.KeyUID = generatedAccountInfo.KeyUID
-	settings.Address = types.HexToAddress(generatedAccountInfo.Address)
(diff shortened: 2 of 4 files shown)


cmd/populate-db/main.go
```diff
@@ -359,7 +359,7 @@ func defaultSettings(generatedAccountInfo generator.GeneratedAccountInfo, derive
 
 	settings.PreviewPrivacy = true
 	settings.Currency = "usd"
-	settings.ProfilePicturesVisibility = 1
+	settings.ProfilePicturesVisibility = 2
```
cammellos: it should be probably using some const that we defined somewhere (I know it's not your code, but mine :) ), would you mind pointing it at it?

thanks

alwx: No problem, I changed it to use the constant.
However, to avoid shadowing I had to change the name `settings` used for the object here (there is a package under that name that is already imported). `defaultSettings` seems to be an appropriate name because that's exactly what we're creating

rasom (approved):

api/defaults.go
```diff
@@ -33,7 +33,8 @@ func defaultSettings(generatedAccountInfo generator.GeneratedAccountInfo, derive
 	s.BackupEnabled = true
 	logLevel := "INFO"
 	s.LogLevel = &logLevel
-	s.ProfilePicturesShowTo = settings.ProfilePicturesShowToContactsOnly
+	s.ProfilePicturesShowTo = settings.ProfilePicturesShowToEveryone
```
Samyoul, line 36: Thank you for your work @alwx . I do have an objection with the nature of this PR, but not with the implementation and your work. If our messenger app is privacy first I'd argue that we should default to a sensible privacy focused setting, `settings.ProfilePicturesShowToContactsOnly` seems to me to strike that balance.

Otherwise we the concept of `ProfilePicturesShowTo` and `ProfilePicturesVisibility` becomes a bit of a gimmick.

The problem of conforming to other messenger conventions and managing the user's expectations should be an exercise in design and communication, not merely giving the user the loosest privacy levels and hoping they find the privacy toggle.

Samyoul (commented):
Should we be giving the user the lowest privacy level by default? This approach fixes an expectation without giving the user an opportunity learn about our approach to privacy.

@cammellos @John-44 @pedro-et what are you thoughts on this?

John-44:
> Should we be giving the user the lowest privacy level by default? This approach fixes an expectation without giving the user an opportunity learn about our approach to privacy.
> 
> @cammellos @John-44 @pedro-et what are you thoughts on this?

@alwx profile photos shouldn't be visible by default, this 'profile photo visibility' setting should be removed enterally (so that profile photos are always visible).  This setting has already been removed from the desktop UI about 6 months ago, and is not featured in the new Mobile designs.  So time to get rid of this enterally! 

@Samyoul there isn't a privacy concern here, because the reason this setting was originally added was nothing to do with privacy (I remember the conversations that led to this setting being added in the first place).  The reason this setting exists at all was due to a specific request from our legal team that was only relevant in the context of 'public group chats'.  Now that the 'public group chat' functionality has been removed from our product, this legal concern is no longer relevant, hence why this setting should be fully removed.

One other comment on the privacy prospective: we treat user avatar images in exactly the same way we treat usernames.  A user is free to choose whatever username they want, and they are free to choose whatever avatar image they want or not choose an avatar image at all.  If a user doesn't want other users to see their avatar image, they have the option of just not adding an avatar image, or if they really want the user also has an option to use a black or white circle or something like that ;-)

Samyoul:
Oh man @John-44 what a plonker I am 🤦, I even remember that concern now that you mention it.

I'm sorry @alwx (and @John-44 )I have made a mistake here.

Maybe we can approve this feature and then add removal of the option to another issue.

Samyoul:
Oh man @John-44 what a plonker I am 🤦, I even remember that concern now that you mention it.

I'm sorry @alwx (and @John-44 )I have made a mistake here.

Maybe we can approve this feature and then add removal of the option to another issue.

Samyoul (approved):

PR #3554, opened 2023-05-31
Added refactor example for urls

Samyoul:
Just some example code of how we could refactor oembed parsing and getting functionality to make it more modular and testable.

See https://github.com/status-im/status-go/issues/3529


igor-sirotin:
Closing as this code has changed a lot since then

PR #3557, opened 2023-05-31
fix(mute): Fix time calculation and skip empty response sending

borismelnik:
Fixes wrong time calculation and adds check for empty message response to avoid unnecessary event sending

Important changes:


jrainville (approved):

ilmotta (approved):
:rocket:

PR #3558, opened 2023-05-31
fix(contacts): add contact to ban and unban responses

jrainville:
Needed for https://github.com/status-im/status-desktop/issues/10501

We were not sending the contact when banning or unbanning, so the client needed to change the contact properties manually, possibly making mistakes and creating an inconsistency between the front and the back end.

I don't think this should be a breaking change, but I did make the unban return a response instead of just the error, so let me know if mobile needs to adapt to it cc @cammellos @ilmotta


ilmotta:
> Needed for [status-im/status-desktop#10501](https://github.com/status-im/status-desktop/issues/10501)
> 
> We were not sending the contact when banning or unbanning, so the client needed to change the contact properties manually, possibly making mistakes and creating an inconsistency between the front and the back end.
> 
> I don't think this should be a breaking change, but I did make the unban return a response instead of just the error, so let me know if mobile needs to adapt to it cc @cammellos @ilmotta

I'll take a look today @jrainville :)

ilmotta:
@jrainville, it's safe to merge this PR from the mobile perspective. The endpoint's response is currently ignored.

ilmotta (approved):

jrainville:
> @jrainville, it's safe to merge this PR from the mobile perspective. The endpoint's response is currently ignored.

Great, thank you for checking

PR #3559, opened 2023-05-31
chore(upgradeSQLCipher): Upgrading SQLCipher to version 5.4.5

alexjba:
Based on benchmarking data we can get the best performance using the latest `sqlcipher` with the improved crypto implementation and increase the DB page size. (see the chart below)

In this PR the `sqlcipher` version is bumped to latest release (v5.4.5) and the DB connection will be opened using v3 compatibility mode (1024 page size and keeping the improved SHA1 and AES)

### github.com/mutecomm/go-sqlcipher
1. The improved crypto argorighms from go-sqlcipher v3 are merged in v4 
Tags: 
v4.4.2-status.1 - merge `burn_stack` improvement 
v4.4.2-status.2 - merge `SHA1` improvement 
v4.4.2-status.4- merge 'AES' improvement

2. Fixed `go-sqlcipher` to support v3 database in compatibility mode (`sqlcipher` already supports this) (Tag: v4.4.2-status.3)
3. Upgrade `sqlcipher` to v5.4.5 (Tag: v4.5.4-status.1)

### github.com/status-im/migrate/v4
1. Upgrade `go-sqlcipher` version in `github.com/status-im/migrate/v4`

### status-go
1. Upgrade `go-sqlcipher` and `migrate` modules in status-go
2. Configure the DB connections to open the DB in v3 compatibility mode

Screenshot 2023-05-26 at 15 46 46


vendor/golang.org/x/net/context/context.go
```diff
@@ -1,56 +0,0 @@
-// Copyright 2014 The Go Authors. All rights reserved.
```
alexjba, line 1: @cammellos Not sure why `make vendor` removed `golang.org/x/net/context`. Should I leave it like this?

sqlite/sqlite.go
```diff
@@ -71,6 +71,20 @@ func EncryptDB(unencryptedPath string, encryptedPath string, key string, kdfIter
 		return err
 	}
 
+	if _, err := db.Exec("PRAGMA encrypted.cipher_page_size = 1024"); err != nil {
```
alexjba, line 74: @cammellos how can I test this flow? I see it's not used in StatusDesktop

jrainville (approved):

richard-ramos (approved):

osmaczko (approved):
Cool!

caybro (approved):

alexjba:
Updated the PR to fix failing tests.
The failure was caused by a go-sqlcipher bug where querying a null blob value in a `sql.NullString` would result in a NullString that is valid and not null.

This is the fix:
https://github.com/status-im/status-go/pull/3559/commits/f461a1b1a72ceca0655beafae5bf15c18711a566

alexjba:
> Updated the PR to fix failing tests. The failure was caused by a go-sqlcipher bug where querying a null blob value in a `sql.NullString` would result in a NullString that is valid and not null.
> 
> This is the fix: [f461a1b](https://github.com/status-im/status-go/commit/f461a1b1a72ceca0655beafae5bf15c18711a566)

Now there are other failures (sometimes different failures on different runs) that can't be reproduced locally:
FAIL: TestMessengerBackupSuite (15.04s)
FAIL    github.com/status-im/status-go/wakuv2    19.679s

I hope the reliability issue is not introduced by this PR

Issue #452, opened 2017-11-09
Fix datarace in geth/signal 


b00ris:
# Problem
Race detector have warnings like this:
  
````
WARNING: DATA RACE
Write at 0x000001ec94e0 by [32mINFO [0m[11-08|17:43:36] Node manager notifed app, that node has stopped [32mgeth[0m=StatusIM
goroutine 23:
  github.com/status-im/status-go/e2e/accounts.(*AccountsTestSuite).TearDownTest()
      <autogenerated>:170 +0x5e
  github.com/status-im/status-go/vendor/github.com/stretchr/testify/suite.Run.func2.1()
      /home/b00ris/go/src/github.com/status-im/status-go/vendor/github.com/stretchr/testify/suite/suite.go:97 +0xe0
  github.com/status-im/status-go/vendor/github.com/stretchr/testify/suite.Run.func2()
      /home/b00ris/go/src/github.com/status-im/status-go/vendor/github.com/stretchr/testify/suite/suite.go:102 +0x320
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:657 +0x107

Previous read at 0x000001ec94e0 by goroutine 91:
  github.com/status-im/status-go/geth/signal.NotifyNode()
      /home/b00ris/go/src/github.com/status-im/status-go/geth/signal/signals.go:74 +0x5c
````

ResetDefaultNodeNotificationHandler, SetDefaultNodeNotificationHandler are using only in tests

# Acceptance Criteria
 No data races warnings related to geth/signal.NotifyNode, ResetDefaultNodeNotificationHandler, SetDefaultNodeNotificationHandler

b00ris:
Fixed

PR #3560, opened 2023-06-01
Fix/mobile issue 16043 

qfrank:
update color hash/id of account on login , relate mobile [issue](https://github.com/status-im/status-mobile/issues/16043)


ghost:
### Pull Request Checklist

api/geth_backend.go
```diff
@@ -201,6 +201,24 @@ func (b *GethStatusBackend) GetAccounts() ([]multiaccounts.Account, error) {
 	return b.multiaccountsDB.GetAccounts()
 }
 
+func (b *GethStatusBackend) getAccountByKeyUID(keyUID string) (*multiaccounts.Account, error) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	if b.multiaccountsDB == nil {
+		return nil, errors.New("accounts db wasn't initialized")
```
vitvly, line 208: Sorry, probably dumb question - when might it happen that `multiaccountsDB == nil`?

rpc/verif_proxy.go
```diff
@@ -6,14 +6,15 @@ package rpc
 import (
 	"context"
 	"fmt"
-)
```
vitvly: Thanks for this. This might have been some linting issue right?

api/geth_backend.go
```diff
@@ -416,6 +444,30 @@ func (b *GethStatusBackend) OverwriteNodeConfigValues(conf *params.NodeConfig, n
 	return conf, nil
 }
 
+func (b *GethStatusBackend) updateAccountColorHashAndColorID(keyUID string, accountsDB *accounts.Database) (*multiaccounts.Account, error) {
```
vitvly, line 447: Might make sense to pass a multiAccount as a param instead of `keyUID`, as in 2 places this function is invoked, we have the multiAccount already, so this would save a DB call. Also, as the 2 invocations already include a conditional on `ColorHash` and `ColorID`, maybe we can also remove the superfluous conditional in the function body itself (line #452)

vitvly (approved):

api/geth_backend.go
qfrank: Not dumb at all 🙂 It depends on if `OpenAccounts` get called correctly, but whatever, keeping it is better than missing it, when invoking function on a nil object would cause strange error stacktrace

rpc/verif_proxy.go
qfrank: yes 👍

api/geth_backend.go
qfrank: Yeah, I thought about it like this before, but I don’t trust the multiAccount that already exists. I just want to make sure it’s from the database and not something passed from the frontend or constructed temporarily like [this](https://github.com/status-im/status-go/blob/fad98fe4a4a331d78c073acc39fac1145e21fb8e/api/geth_backend.go#L571) 🙂 @siphiuel

cammellos (approved):

cammellos (approved):

PR #3561, opened 2023-06-01
Stop loading transfers for deleted accounts

IvanBelyakoff:
Previously no event was sent using account feed, when an account was removed.
Introduced events for adding and removing accounts and handling them properly.

### DIFF
--- a/services/accounts/accounts.go
+++ b/services/accounts/accounts.go
@@ -15,6 +15,7 @@ import (
 	"github.com/status-im/status-go/multiaccounts/accounts"
 	"github.com/status-im/status-go/params"
 	"github.com/status-im/status-go/protocol"
+	"github.com/status-im/status-go/services/accounts/accountsevent"
 )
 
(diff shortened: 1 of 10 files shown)


dlipicar (approved):
LGTM!

I don't think we're currently doing a proper cleanup when an account gets removed (that is, different tables, like `blocks` and `transfers`, remain populated with data associated with an account that's not there anymore). There's a DB size and performance hit, so we should perhaps consider that improvement in the future.

IvanBelyakoff:
> LGTM!
> 
> I don't think we're currently doing a proper cleanup when an account gets removed (that is, different tables, like `blocks` and `transfers`, remain populated with data associated with an account that's not there anymore). There's a DB size and performance hit, so we should perhaps consider that improvement in the future.

you are right, initially I was going to fix removing data from these tables for removed accounts, but that should be done on per strategy basis as the sequential strategy, for example, uses one extra table that it needs to clear, so it might need some refactoring, so I left it out of scope of this PR

PR #3565, opened 2023-06-02
Feature/local pairing preflight

Samyoul:
## What's Changed?

Added a new package `server/pairing/preflight` and a new API endpoint `LocalPairingPreflightOutboundCheck`:

`LocalPairingPreflightOutboundCheck` creates a local TLS server accessible via an outbound network address. The function creates a client and makes an outbound network call to the local server. This function should be triggered to ensure that the device has permissions to access its LAN or to make outbound network calls.

The presumption is that by targeting the `http.Listener` to the device's network address and binding the TLS certificate to the same address that the `http.Client` will be forced to call the OS network interface and make an attempt to make a LAN call. Therefore if the OS requires permissions to make outbound LAN calls this preflight will trigger this before any critical path is interrupted.

## Why Change?

An attempt to address iOS network interface detecting outbound private network calls and triggering a permissions request. See https://stackoverflow.com/a/64242745 .


siddarthkay (approved):

qfrank (approved):
LGTM except the lint issue❤️

smohamedjavid (approved):

PR #3566, opened 2023-06-02
feat: expose account created at

alaibe:


multiaccounts/accounts/database.go
```diff
@@ -338,7 +341,9 @@ func (db *Database) processKeypairs(rows *sql.Rows) ([]*Keypair, error) {
 		if accClock.Valid {
 			acc.Clock = uint64(accClock.Int64)
 		}
-
+		if accCreatedAt.Valid {
+			acc.CreatedAt = accCreatedAt.Time.Unix()
```
saledjenic: @alaibe wdyt of using `UnixMilli()` instead for higher precision?

saledjenic (approved):

PR #3568, opened 2023-06-02
feat: add configurable throttling mechanism for importing msgs

osmaczko:
part of: status-im/status-desktop#10815

Introduce slow and fast mode of messages import.
- slow mode: up to 10 msgs/sec
- fast mode: up to 1000 msgs/sec

It is up to the client to switch between these two modes, the default is slow mode.

In fast mode, the import time of a 4.2GB archive on my machine significantly dropped from around 2 hours to approximately 20 minutes, maintaining a stable CPU usage at about 5-7% without any spikes.
```
$ du -sh Status/data/archivedata 
4.2G    Status/data/archivedata
```

### DIFF
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -12,6 +12,8 @@ import (
 	"sync"
 	"time"
 
+	"golang.org/x/time/rate"
+
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/ethclient"
 
@@ -46,6 +48,19 @@ var messageArchiveInterval = 7 * 24 * time.Hour
 var updateActiveMembersInterval = 24 * time.Hour
 
 const discordTimestampLayout = "2006-01-02T15:04:05+00:00"
+
+var importRateLimiter = rate.NewLimiter(rate.Every(importSlowRate), 1)
+
+const (
+	importSlowRate          = time.Second / 1
+	importFastRate          = time.Second / 100
+	importMessagesChunkSize = 10
+)
+
+const (
+	maxChunkSizeMessages = 1000
+	maxChunkSizeBytes    = 1500000
+)
 
 func (m *Messenger) publishOrg(org *communities.Community) error {
 	m.logger.Debug("publishing org", zap.String("org-id", org.IDString()), zap.Any("org", org))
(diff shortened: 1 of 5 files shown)


jrainville (approved):
nice

protocol/messenger_communities.go
```diff
@@ -47,6 +49,19 @@ var updateActiveMembersInterval = 24 * time.Hour
...
+
+const (
+	importSlowRate          = time.Second / 1
+	importFastRate          = time.Second / 100
+	importMessagesChunkSize = 10
+)
+
+const (
+	maxChunkSizeMessages = 1000
+	maxChunkSizeBytes    = 1500000
```
0x-r4bbit, line 62: This one doesn't seem to be used any longer..

0x-r4bbit: Actually, both are no longer used here

protocol/messenger_communities.go
```diff
@@ -3637,3 +3676,42 @@ func (m *Messenger) CheckPermissionsToJoinCommunity(request *requests.CheckPermi
...
+			end = len(slice)
+		}
+
+		chunks = append(chunks, slice[i:end])
+	}
+
+	return chunks
+}
+
+func chunkAttachmentsByByteSize(slice []*protobuf.DiscordMessageAttachment, maxFileSizeBytes uint64) [][]*protobuf.DiscordMessageAttachment {
```
0x-r4bbit, line 3697: This function doesn't seem to be used anymore.

Is this not an issue with discord file attachments that are big?
One of the reasons this chunking was introduced was because we would run into scenarios where we'd process many messages that have discord message attachments, which would take a long time to save.

0x-r4bbit (commented):
Changes LGTM just a question:

Why exactly did we move the chunking before the message handling, instead of doing it inside the message handling?

Was that a reason for slow down?

protocol/messenger_communities.go
osmaczko: It is still used when importing the community in `RequestImportDiscordCommunity` which is a bit different than importing messages from archives.

protocol/messenger_communities.go
osmaczko: It is still used when importing the community in `RequestImportDiscordCommunity` which is a bit different than importing messages from archives.

osmaczko:
> Was that a reason for slow down?

The throttling mechanism we had implemented previously came with several drawbacks. We were handling a large volume of messages at once, often ~1k, before resting for 2 seconds. Processing these messages all at once frequently resulted in the database locking up during processing. New approach uses a rate limiter, leading to a more linear processing timeline, as it also takes into account the time of messages handling itself. Moreover, chunks are much smaller now, allowing for breathing room for other database operations more frequently.

> Why exactly did we move the chunking before the message handling, instead of doing it inside the message handling?

I moved the logic to the earliest possible point where we could break down the archive messages into smaller chunks to avoid logic duplication in separate handlers.

protocol/messenger_communities.go
0x-r4bbit: Ah sorry, forgot

0x-r4bbit:
Thank you for clarifying @osmaczko !

0x-r4bbit (approved):

PR #3569, opened 2023-06-02
feat: community admin permission functionality

### LINKED ISSUE #10114: Enable Admin role (in addition to Owner role)
Now that we have Community level token permissions working, we need to add the Admin role so that a Community Owner can appoint multiple admins to help them manage the community.  

The differences between the Owner and Admin roles are as follows:

| Ability  | Owner | Owner, using install without community private key | Admin |
| :------------- | :------------- | :------------- | :------------- |
| Ability to access the Community admin area | Yes | Yes | Yes |
| Ability to access Community admin functions outside the Admin area | Yes | Yes | Yes |
| Edit Community (inc. edit tags, logo, banner, request to join required setting, pin setting, etc...) | Yes | Yes | Yes |
|  Create, edit and delete channels or categories | Yes | Yes |  Yes |
|  Reorder channels and categories | Yes | Yes | Yes |
|  Create, edit and delete 'become member' permissions | Yes | Yes |  Yes |
|  Create, edit and delete 'become admin' permissions | Yes | Yes | No |
| Create, edit and delete 'channel' permissions  | Yes | Yes | Yes |
| Receive 'request to join' notifications, and ability to Accept or Reject (see *Note 1* below)  | Yes | Yes | Yes |
| Kick and ban members (see *Note 1* below)  | Yes | Yes | Yes |
| Delete any message in the Community | Yes | Yes  | Yes |
| Pin messages, if 'Any member can pin a message' is switched off | Yes | Yes | Yes |
| View statistics on Overview screen  | Yes | Yes | Yes |
|  Mint Tokens | Yes | Yes | No |
|  Airdrop Tokens | Yes | Yes | No |
| Access Community private key  | Yes | No | No |

*Note 1:* When an Admin accepts a 'request to join', or kicks or bans a member, this command is sent to the Owner Node which then performs the required action. If the Owner Node is not online, the required action will be performed by the Owner Node the next time the Owner Node comes online.

*Note 2:* The "Owner, using install without community private key" abilities are for the scenario where (for example) the Owner has the community private key in their Status Desktop instance running on their Desktop computer, but is accessing their community from a Status Desktop instance running on their laptop that *does not* contain the community private key.

*Note 3:* In the community admin area, the "Mint tokens" and "Airdrop" sections in the secondary nav should be hidden from Admins (e.g. only Owners should see these two sections in the admin area secondary nav).

mprakhov:
- created community admin events
- implemented sending community admin events
- implemented handling community admin events
- added tests
- replaced `admin` bool on a member role

Important changes:
- [ ] Please, check the JSON data (replacing `admin` bool), which we are returning, does not impact the Mobile team

Closes: https://github.com/status-im/status-desktop/issues/10114
status-desktop PR: https://github.com/status-im/status-desktop/pull/10909


ghost:
### Pull Request Checklist

mprakhov:
@cammellos please assign someone from mobile team to take a look a PR

protocol/communities/community.go
```diff
@@ -191,7 +191,7 @@ func (o *Community) MarshalJSON() ([]byte, error) {
 	}
 	communityItem := struct {
 		ID                          types.HexBytes                                `json:"id"`
-		Admin                       bool                                          `json:"admin"`
+		MemberRole                  protobuf.CommunityMember_Roles                `json:"memberRole"`
```
0x-r4bbit, line 194: @cammellos this is a breaking change in the messenger response if I'm not mistaken, so mobile needs to know about this.

jrainville:
FYI, the builds are not working with error `multiaccounts/database.go:53:3: unknown field 'ColorID' in struct literal of type protobuf.MultiAccount`

protocol/communities/community.go
```diff
@@ -788,25 +774,29 @@ func (o *Community) removeMemberFromOrg(pk *ecdsa.PublicKey) {
 	for _, chat := range o.config.CommunityDescription.Chats {
 		delete(chat.Members, key)
 	}
-
-	o.increaseClock()
```
jrainville, line 792: I see that this line was moved to the functions calling `removeMemberFromOrg`. What is the goal? Not always we want to increase the clock?

protocol/communities/community.go
```diff
@@ -1849,6 +1897,44 @@ func (o *Community) AddMemberWallet(memberID string, addresses []string) (*Commu
 	return changes, nil
 }
 
+func (o *Community) createDeepCopy() Community {
```
jrainville, line 1900: I'm afraid that this might be forgotten at one point and we add a new property and forget to update this function, then we'd have a copy of the object that is missing the property.

Is there any way to alleviate that risk?

Does Go have a deep copy function for objects that does it automatically?

protocol/communities/community.go
```diff
@@ -1903,3 +1989,160 @@ func (d sortSlice) Swap(i, j int) {
...
+
+	prevPermissions := o.TokenPermissionsByType(protobuf.CommunityTokenPermission_BECOME_MEMBER)
+
+	if len(newPermissions) < len(prevPermissions) {
+		for _, permission := range newPermissions {
+			if permissionExists(permission.Id, prevPermissions) {
+				delete(o.config.CommunityDescription.TokenPermissions, permission.Id)
+			}
+		}
+	}
```
jrainville, line 2090: I feel like this function is missing the part where we set the permissions to the new one. We seem to only be deleting. Maybe I'm missing something

protocol/communities/community_admin_event.go
```diff
@@ -0,0 +1,347 @@
...
+
+	if adminEvent.Clock <= o.config.CommunityDescription.Clock {
+		return nil, errors.New("clock for admin event is outdated")
+	}
+
+	if adminEvent.Type == protobuf.CommunityAdminEvent_UNKNOWN {
+		return nil, errors.New("unknown admin event")
+	}
+
+	// create a deep copy of current community so we can patch CoommunityDescription
```
jrainville:
```suggestion
	// create a deep copy of current community so we can patch CommunityDescription
```

mprakhov:
> FYI, the builds are not working with error `multiaccounts/database.go:53:3: unknown field 'ColorID' in struct literal of type protobuf.MultiAccount`

I rebased on the development HEAD ~1, it should fix the build until status-go breaking commit will be fixed/reverted

protocol/communities/community_admin_event.go
```diff
@@ -0,0 +1,347 @@
...
+		if len(newPermissions) > len(prevPermissions) {
+			copy.addBecomeMemberTokenPermissions(newPermissions)
+			break
+		}
+
+		// update existing permissions
+		copy.updateTokenPermissions(newPermissions)
+		break
+	case protobuf.CommunityAdminEvent_COMMUNITY_MEMBER_TOKEN_PERMISSION_DELETE:
+		copy.replaceBecomeMemberTokenPermissions(adminEvent.TokenPermissions)
```
jrainville: Oh, is this function to remove a permission? If so, I'd rename it to `removeBecomeMemberTokenPermissions`. Replace means to remove something and put something else in its place.

protocol/communities/community_admin_event.go
```diff
@@ -0,0 +1,347 @@
...
+		// the default case, which will error out
+		break
+
+	case protobuf.CommunityAdminEvent_COMMUNITY_MEMBER_KICK:
+		pk, err := common.HexToPubkey(adminEvent.MemberToAction)
+		if err != nil {
+			return nil, err
+		}
+
+		if copy.IsMemberOwner(pk) {
```
jrainville, line 317: Do we also need to check other admins? Can an admin kick another admin?

protocol/communities/manager.go
```diff
@@ -620,62 +626,112 @@ func (m *Manager) EditCommunityTokenPermission(request *requests.EditCommunityTo
...
 					return err
 				}
+				isAdmin = true
+			} else if !permissionResponse.Satisfied && isAdmin {
+				removeAdminRole = true
+				isAdmin = false
+			}
+		}
+
+		// Remove admin role if user do not pass admin perrmissions or we do not have admin permissions but have an admin role
```
```suggestion
		// Remove admin role if user does not pass admin permissions or we do not have admin permissions but have an admin role
```

protocol/communities/manager.go
```diff
@@ -1346,14 +1565,31 @@ func (m *Manager) AcceptRequestToJoin(request *requests.AcceptRequestToJoinCommu
 			walletAddresses = append(walletAddresses, gethcommon.HexToAddress(walletAddress))
 		}
 
-		permissionResponse, err := m.checkPermissionToJoin(becomeMemberPermissions, walletAddresses, true)
-		if err != nil {
-			return nil, err
+		// admin token permissions requred to became an admin must not cancel request to join
```
```suggestion
		// admin token permissions required to become an admin must not cancel request to join
```

protocol/communities/manager.go
```diff
@@ -1550,19 +1819,34 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request
 			verifiedAddresses = append(verifiedAddresses, gethcommon.HexToAddress(walletAddress))
 		}
 
-		permissionResponse, err := m.checkPermissionToJoin(becomeMemberPermissions, verifiedAddresses, true)
-		if err != nil {
-			return nil, err
+		// admin token permissions requred to became an admin must not cancel request to join
```
```suggestion
		// admin token permissions required to become an admin must not cancel request to join
```

protocol/communities/manager.go
```diff
@@ -1550,19 +1819,34 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request
...
+		// if requirements were not met
+		if len(becomeAdminPermissions) > 0 {
```
jrainville: This code is there twice it seems. Any way we can extract it to another function?

jrainville (commented):
Wow that is massive.

Great work you two.

I didn't find anything big, but it's hard to find everything in such a big PR

protocol/communities/community.go
mprakhov: When the admin sends any event, he sends clock changes. When the client accepts admin events, he patched CommunityDescription by clock (applying admin event clock time) and by actions, but during this patching, I don't want to increase the clock as it is not necessary, I can apply admin clock. So we have a fn which has mutex lock and clock increase and implementation, which has the code of CommunityDescription update

protocol/communities/community.go
mprakhov: As I understood - nothing from the box.
ChatGPT tells me that alternatively, I can use external packages like github.com/mitchellh/copystructure or github.com/ulule/deepcopier that provide deep copying functionality. 
If will be ok, I can try to test it

protocol/communities/community.go
mprakhov: Bad naming, it is deleting, will rename

protocol/communities/community.go
jrainville: Yeah those sound very nice. Maintaining our own deep copy functions sounds like a future trap haha

protocol/communities/community_admin_event.go
0x-r4bbit: I believe we've decided that admins should not be allowed to kick other admins. So we should probably extend this check with `IsMemberAdmin()`

mprakhov: new ticket was created

protocol/communities/community.go
mprakhov: Tested module https://github.com/jinzhu/copier and it seems it does not do the full copy. The tests were not passing.
I found out that we can easily copy the protobuf data, so I copied via Clone() CommunityDescription, all other by fields.

protocol/communities/community_admin_event.go
```diff
@@ -0,0 +1,346 @@
...
+
+func (o *Community) PatchCommunityDescriptionByAdminEvent(adminEvent *protobuf.CommunityAdminEvent) (*protobuf.CommunityDescription, error) {
+	o.mutex.Lock()
+	defer o.mutex.Unlock()
+
+	if adminEvent.Clock <= o.config.CommunityDescription.Clock {
+		return nil, errors.New("clock for admin event is outdated")
+	}
+
+	if adminEvent.Type == protobuf.CommunityAdminEvent_UNKNOWN {
```
cammellos, line 203: This is tricky, it makes the protocol non-upgradable, something worth thinking about (have to better look at the code to judge), but say we add a new event, clients will refuse to process it and return an error (which might be fine, as long as that doesn't prevent us to process further events).

protocol/communities/community_admin_event.go
```diff
@@ -0,0 +1,346 @@
...
+		return nil, errors.New("unknown admin event")
+	}
+
+	// create a deep copy of current community so we can patch CommunityDescription
+	copy := o.createDeepCopy()
+
+	copy.config.CommunityDescription.Clock = adminEvent.Clock
```
cammellos, line 210: I think this maybe needs to be adjusted, if we go for events, you don't want the last one wins, you should be able to process them regardless of the order.

just to make sure, if I read the code correctly, let's take this example:

1) `admin A` changes say the name of a channel at `clock` 3
2)   `admin B` changes the name of a different channel at `clock` 4.
3) Client `C` retrieves and process 2) and 1) in that order

Here, seems like 1) would be discarded, since the clock would be `4`, but the last received event has clock `3`

Is that correct?

I think the desired behavior is that both are processed

cammellos (commented):
Sorry, going through it a bit slowly, left some questions that I think we want to clarify, thanks for the PR!

protocol/communities/community_admin_event.go
mprakhov: 1 admin event == 1 msg, from my understanding, if there will be an error, we will proceed next msg

cammellos: cool, that's good

protocol/communities/community_admin_event.go
mprakhov: Yes, but I was orienting on `func (o *Community) UpdateCommunityDescription`, where we do clock validation
Does it mean, that if instead of 2 admins we as an example take 2 owners (and you can transfer ownership) - if 2 owners send some update with the scenario above, we'll get the same result? So it is not an admin-only issue, but a global one for 2 owners

cammellos: 2 owners I think we are fine with `last one wins`, as we don't really expect to have multiple owners for community.

When it comes to admins, is a bit different, we do expect multiple admins to make changes to the community at different times

mprakhov: Any suggestions? :)
Do not track the clock? Just increase it to 1 from the current client CommunityDescription clock?

mprakhov: but again, this can bring another issue, when admin action will increase the clock and the owner msg can be rejected later if he'll not receive an admin event before sending his msg...

cammellos: > Any suggestions? :) Do not track the clock? Just increase it to 1 from the current client CommunityDescription clock?

I think we need to separate clocks for owner and admins, for private group chats, you "build" the group chat from a stream of events, but you don't keep track of the clock in a different piece of state ( the clock of the events is the latest clock in the event stream).
So I would not mix the two, basically a community is:

CommunityDescription -> only modifiable by a owner, has its own clock, when receiving an admin update, this clock should not be modified
CommunityAdminEvents -> each one has its own clock, they should not modify directly the community description

AmendedCommunityDescription = CommunityDescription + Process Any Event with clock > CommunityDescription.Clock

The issue is the owner republishing the community description / making changes without having received all the admin events, that might result in some events being ignored/discarded, I don't think there's a way around it, but admin can always republish events if they see that's the case
That would be the solution I'd be leaning towards
This is sort of similar, you apply events on top of the latest community description.

> when admin action will increase the clock and the owner msg can be rejected later if he'll not receive an admin event before sending his msg...

I believe this should not happen, if we don't mix clocks

protocol/communities/manager.go
```diff
@@ -627,59 +636,109 @@ func (m *Manager) EditCommunityTokenPermission(request *requests.EditCommunityTo
...
+
+		if memberKey == common.PubkeyToHex(&m.identity.PublicKey) || community.IsMemberOwner(memberPubKey) {
+			continue
+		}
+
+		isAdmin := community.IsMemberAdmin(memberPubKey)
+		memberHasWallet := len(member.RevealedAccounts) > 0
+
+		// Check if user was not treated as an admin without wallet in open community
+		// or user threated as a member without wallet in closed community
```
0x-r4bbit:
```suggestion
		// or user treated as a member without wallet in closed community
```

protocol/communities/manager.go
```diff
@@ -627,59 +636,109 @@ func (m *Manager) EditCommunityTokenPermission(request *requests.EditCommunityTo
...
-			hasPermission := permissionResponse.Satisfied
+		accountsAndChainIDs := revealedAccountsToAccountsAndChainIDsCombination(member.RevealedAccounts)
 
-			if !hasPermission {
-				pk, err := common.HexToPubkey(memberKey)
-				if err != nil {
-					return err
-				}
-				_, err = community.RemoveUserFromOrg(pk)
+		// Check if user is still an admin or can became an admin and do update of member role
```
0x-r4bbit:
```suggestion
		// Check if user is still an admin or can become an admin and do update of member role
```

protocol/communities/manager.go
```diff
@@ -815,7 +896,7 @@ func (m *Manager) ExportCommunity(id types.HexBytes) (*ecdsa.PrivateKey, error)
 		return nil, err
 	}
 
-	if !community.IsAdmin() {
+	if community.PrivateKey() == nil {
```
0x-r4bbit:
```suggestion
	if !community.IsOwner() {
```

0x-r4bbit: Might as well replace the custom error below with one of the community errors we have

protocol/communities/manager.go
```diff
@@ -1224,6 +1344,108 @@ func (m *Manager) HandleCommunityDescriptionMessage(signer *ecdsa.PublicKey, des
...
+		return nil, errors.New("signer can't be nil")
+	}
+
+	community, err := m.persistence.GetByID(&m.identity.PublicKey, adminEvent.CommunityId)
+	if err != nil {
+		return nil, err
+	}
+
+	if !community.IsMemberAdmin(signer) {
+		return nil, errors.New("user is not an admin")
```
0x-r4bbit, line 1360: Could we also use one of the community errors here that we already have?

protocol/communities/manager.go
```diff
@@ -1342,25 +1564,37 @@ func (m *Manager) AcceptRequestToJoin(request *requests.AcceptRequestToJoinCommu
...
 		}
 
 		accountsAndChainIDs := revealedAccountsToAccountsAndChainIDsCombination(revealedAccounts)
 
-		permissionResponse, err := m.checkPermissionToJoin(becomeMemberPermissions, accountsAndChainIDs, true)
-		if err != nil {
-			return nil, err
-		}
-		hasPermission := permissionResponse.Satisfied
+		// admin token permissions required to became an admin must not cancel request to join
```
0x-r4bbit, line 1586:
```suggestion
		// admin token permissions required to become an admin must not cancel request to join
```

protocol/communities/manager.go
```diff
@@ -1342,25 +1564,37 @@ func (m *Manager) AcceptRequestToJoin(request *requests.AcceptRequestToJoinCommu
...
+		// if requirements were not met
+		hasPermission := m.isWaleltsHasAdminPermission(becomeAdminPermissions, accountsAndChainIDs)
```
0x-r4bbit:
```suggestion
		hasPermission := m.walletHasAdminPermission(becomeAdminPermissions, accountsAndChainIDs)
```

protocol/communities/manager.go
```diff
@@ -1498,47 +1778,36 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request
...
+	if (len(becomeMemberPermissions) > 0 || len(becomeAdminPermissions) > 0) && len(request.RevealedAccounts) > 0 {
+		accountsAndChainIDs := revealedAccountsToAccountsAndChainIDsCombination(request.RevealedAccounts)
 
-		// verify if revealed addresses indeed belong to requester
-		for _, revealedAccount := range request.RevealedAccounts {
-			recoverParams := account.RecoverParams{
-				Message:   types.EncodeHex(crypto.Keccak256(crypto.CompressPubkey(signer), community.ID(), requestToJoin.ID)),
-				Signature: types.EncodeHex(revealedAccount.Signature),
-			}
+		// admin token permissions required to became an admin must not cancel request to join
```
0x-r4bbit:
```suggestion
		// admin token permissions required to become an admin must not cancel request to join
```

protocol/communities/manager.go
```diff
@@ -1498,47 +1778,36 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request
...
+		// if requirements were not met
+		hasPermission = m.isWaleltsHasAdminPermission(becomeAdminPermissions, accountsAndChainIDs)
```
0x-r4bbit:
```suggestion
		hasPermission = m.walletHasAdminPermission(becomeAdminPermissions, accountsAndChainIDs)
```

protocol/communities/manager.go
```diff
@@ -1609,16 +1869,12 @@ func (c *CheckPermissionToJoinResponse) calculateSatisfied() {
...
-				satisfied = false
-				break
+			if criteria {
+				c.Satisfied = true
+				return
 			}
 		}
-		if satisfied {
-			c.Satisfied = true
-		}
```
0x-r4bbit: These changes above do not look right to me.
In fact I've fixed them recently in https://github.com/status-im/status-go/pull/3494#discussion_r1220134354

**All** criteria must be satisfied for a permission to be satisfied.
However, only **one** permission must be satisfied to pass the check.

protocol/communities/manager.go
```diff
@@ -3623,3 +3900,13 @@ func revealedAccountsToAccountsAndChainIDsCombination(revealedAccounts []*protob
 	}
 	return accountsAndChainIDs
 }
+
+func (m *Manager) isWaleltsHasAdminPermission(becomeAdminPermissions []*protobuf.CommunityTokenPermission, accounts []*AccountChainIDsCombination) bool {
```
0x-r4bbit:
```suggestion
func (m *Manager) walletHasAdminPermission(becomeAdminPermissions []*protobuf.CommunityTokenPermission, accounts []*AccountChainIDsCombination) bool {
```

Also, as an FYI, I've introduced `checkPermissions()` which is essentially the same as `checkPermissionToJoin()` because functionality-wise, they do exactly the same thing, as you've probably figured out here as well..

https://github.com/status-im/status-go/pull/3605/files#diff-0c22647108517d38489ae5faa65a914ade3be02ea7a7e37355a82739da9aeb69R1640-R1641

services/chat/api.go
```diff
@@ -88,7 +86,7 @@ type ChannelGroup struct {
 	Chats                   map[string]*Chat                         `json:"chats"`
 	Categories              map[string]communities.CommunityCategory `json:"categories"`
 	EnsName                 string                                   `json:"ensName"`
-	Admin                   bool                                     `json:"admin"`
+	MemberRole              protobuf.CommunityMember_Roles           `json:"memberRole"`
```
0x-r4bbit, line 89: Need to make sure mobile team is aware of this change as this is breaking. /cc @cammellos

0x-r4bbit (changes requested):
I gave this a thorough review now and all in all this looks really good.
Just some mistakes in the comments and then there's one major logical bug that needs fixing, with regards to calculation of permission satisfaction.

services/chat/api.go
0x-r4bbit: Just realised I've already made such a comment here https://github.com/status-im/status-go/pull/3569#discussion_r1214544056

Sorry for the double ping

protocol/communities/community.go
mprakhov: it was communicated before

0x-r4bbit (approved):

jrainville (approved):

jrainville:
Merging now to avoid further conflicts. The issue/refactor proposed will be implemented in another PR

PR #3573, opened 2023-06-05
feat: clear balance cache after usage

alaibe:


ghost:
### Pull Request Checklist

services/wallet/transfer/commands.go
```diff
@@ -305,7 +305,7 @@ func (c *controlCommand) Run(parent context.Context) error {
 			BlockNumber: target,
 		})
 	}
-
+	bCache.Clear()
```
IvanBelyakoff: I think it is safe to clear even earlier - at line 272 before loading transfers

services/wallet/transfer/balance_cache.go
```diff
@@ -24,9 +30,19 @@ type balanceCache struct {
...
+func newBalanceCache() *balanceCache {
+	return &balanceCache{
+		balances:     make(map[common.Address]map[uint64]*big.Int),
+		nonces:       make(map[common.Address]map[uint64]*int64),
+		nonceRanges:  make(map[common.Address]map[int64]nonceRange),
+		sortedRanges: make(map[common.Address][]nonceRange),
+	}
+}
+func (b *balanceCache) Clear() {
+	b.balances = make(map[common.Address]map[uint64]*big.Int)
```
IvanBelyakoff: I think that we might want to keep first and last items for nonces and balances, as at least for OnDemand strategy, one side of the checked block range should hit the cache on next check.

IvanBelyakoff (approved):
LGTM

PR #3581, opened 2023-06-07
Feat/sync activity center notification

qfrank:
This PR is a further supplement and improvement to [#3535](https://github.com/status-im/status-go/pull/3535), to implement the synchronization function for the activity_center_notifications and activity_center_states tables. This synchronization only occurs after the devices are paired.

status: ready.


qfrank:
I have discussed the implementation issue with mishkarogachev before. The conclusion at that time was that since we have not implemented the function of synchronizing messages, we can only synchronize the following three types of notifications: ActivityCenterNotificationTypeCommunityKicked, ActivityCenterNotificationTypeCommunityMembershipRequest, and ActivityCenterNotificationTypeCommunityRequest. They are all unrelated to messages.

But I have a question. Our activity_center_notifications table only has two fields related to messages, message/reply_message, and they are both blob types. Is there any problem if we synchronize all types of notifications?

cc @cammellos @Samyoul

protocol/messenger_activity_center.go
```diff
@@ -318,11 +399,126 @@ func (m *Messenger) handleActivityCenterAccepted(state *ReceivedMessageState, me
...
+func (m *Messenger) handleSyncActivityCenterNotificationState(state *ReceivedMessageState, a *protobuf.SyncActivityCenterNotificationState) error {
+	s := &ActivityCenterState{
+		HasSeen:   a.HasSeen,
+		UpdatedAt: a.UpdatedAt,
+	}
+	_, err := m.persistence.UpdateActivityCenterNotificationState(s)
+	if err != nil {
+		return err
+	}
+	state.Response.SetActivityCenterState(s)
```
saledjenic: @qfrank we will get to this line even `s.UpdatedAt` is equal or less then the `updated_at` column at the db, which should not be the case, cause sending this signal will cause UI updated on the client side. It should be sent only when the db is really updated.

protocol/messenger_activity_center.go
```diff
@@ -318,11 +399,126 @@ func (m *Messenger) handleActivityCenterAccepted(state *ReceivedMessageState, me
...
 
 	return state.Response.Merge(resp)
 }
+
+func (m *Messenger) handleSyncActivityCenterNotificationState(state *ReceivedMessageState, a *protobuf.SyncActivityCenterNotificationState) error {
+	s := &ActivityCenterState{
+		HasSeen:   a.HasSeen,
+		UpdatedAt: a.UpdatedAt,
+	}
+	_, err := m.persistence.UpdateActivityCenterNotificationState(s)
```
saledjenic: @qfrank if this handling is `handleSyncActivityCenterNotificationState`, so it's about singular, not plural, how come that we don't set anywhere any notification id? I mean how do we know which notification should be updated?

protocol/messenger_activity_center.go
```diff
@@ -318,11 +399,126 @@ func (m *Messenger) handleActivityCenterAccepted(state *ReceivedMessageState, me
...
+		if err != nil {
+			return err
+		}
+		err = m.persistence.SaveActivityCenterNotification(notification, false)
+		if err != nil {
+			return err
+		}
+		notifications = append(notifications, notification)
+	}
+	response, err := m.processActivityCenterNotifications(notifications, true)
```
saledjenic, line 438: @qfrank the same issue is here, we're adding all notifications to the response, regardless they are updated in the db or skipped because of the older clock.

protocol/messenger_activity_center.go
```diff
@@ -60,16 +63,70 @@ func (m *Messenger) ActivityCenterNotificationsCount(request ActivityCenterCount
...
+	}
+	return m.sendToPairedDevices(context.TODO(), common.RawMessage{
+		Payload:             encodedMessage,
+		MessageType:         protobuf.ApplicationMetadataMessage_SYNC_ACTIVITY_CENTER_NOTIFICATION_STATE,
+		ResendAutomatically: true,
+	})
+}
+
+func (m *Messenger) syncActivityCenterNotifications(notifications []*ActivityCenterNotification) (err error) {
+	if notifications == nil {
```
saledjenic, line 94: @qfrank this check here is unnecessary, you can remove it.

saledjenic (changes requested):

protocol/messenger_activity_center.go
qfrank: there is an [API](https://github.com/status-im/status-go/blob/3bc07aac5a887595245393dd5626e9b726723c3c/services/ext/api.go#L1136) which mark all notifications as seen, corresponding table is `activity_center_states`, we have only 1 record in this table. there is another table `activity_center_notifications` which store different notifications

saledjenic: Ahh ok if it refers to all notifications, then ok (though the naming is a bit misleading).

protocol/messenger_activity_center.go
qfrank: I think it make sense in some cases, but i can't remember which case ATM .. check is always good than not 🙂

protocol/messenger_activity_center.go
```diff
@@ -60,16 +63,70 @@ func (m *Messenger) ActivityCenterNotificationsCount(request ActivityCenterCount
...
+		return nil
+	}
+	var s []*protobuf.SyncActivityCenterNotification
+	for _, n := range notifications {
+		if n == nil {
```
ilmotta: I checked all calls to `syncActivityCenterNotifications`, and in all cases the slice can't have nils inside. Maybe I missed something. Could you please share the offending call that can pass a slice of notifications with nil(s) inside?

Unexpected nils can be a more serious problem in the system, so if we can pinpoint the place that's building the `notifications` slice with nils, I think we should fix it, and not allow the slice to propagate further in the system.

qfrank:
hi @saledjenic , does PR look good to you now?

protocol/messenger_activity_center.go
```diff
@@ -318,11 +399,130 @@ func (m *Messenger) handleActivityCenterAccepted(state *ReceivedMessageState, me
...
+	}
+	response, err := m.processActivityCenterNotifications(notifications, true)
+	if err != nil {
+		return err
+	}
+	return state.Response.Merge(response)
+}
+
+func convertActivityCenterNotificationToProtobuf(n *ActivityCenterNotification) (p *protobuf.SyncActivityCenterNotification, err error) {
+	if n == nil {
```
ilmotta: This is a good example of a function I would prefer to receive a non-pointer value, because it means the compiler will guarantee `n` can't be nil, so we can remove this check. It leads to less defensive code.

Now callers of `convertActivityCenterNotificationToProtobuf` will (possibly) need to deal with the nil value, which is actually good because the *nil handling* will happen closer to where the problem is (similar to how error handling in Go is better handled closer to where it happened). This approach can also greatly help pinpoint issues in the system, because now we can log the unexpected nil issue right where it happened.

It's also interesting to accept a non-pointer value because the name of the function starts with `convert`, which strongly indicates `n` is not mutated.

This is a minor point @qfrank, and I'm aware not everyone agrees with it.

protocol/messenger_activity_center.go
```diff
@@ -318,11 +399,130 @@ func (m *Messenger) handleActivityCenterAccepted(state *ReceivedMessageState, me
...
+			return
+		}
+	}
+	if n.ReplyMessage != nil {
+		replyMessage, err = json.Marshal(n.ReplyMessage)
+		if err != nil {
+			return
+		}
+	}
+	p = &protobuf.SyncActivityCenterNotification{
```
ilmotta: Not a huge problem here, but I think it's a bit better to give more expressive names to named parameters. First of all because of the scope of the code is not short (as recommended [here](https://google.github.io/styleguide/go/decisions#variable-names)).

The other reason is that it's relatively easy to shadow single letter variables, and for a named return this can cause a silent bug in the system.

Again, not much of a problem here, just something I try to follow as a guide.

protocol/messenger.go
```diff
@@ -3833,6 +3833,40 @@ func (m *Messenger) handleRetrievedMessages(chatWithMessages map[transport.Filte
 							continue
 						}
 
+					case protobuf.SyncActivityCenterNotifications:
```
ilmotta, line 3836: [Light suggestion] This `handleRetrieveMessages` function has more than 1000 lines of code already :cry: :godmode: It would be ideal if we stop growing it like a god function IMO.

ilmotta (approved):
Looks good @qfrank! I made only non-blocking comments.

May I ask why we want to merge `feat/sync/activity_center_notification2` into `feat/sync_activity_center_notification` instead of the usual *merge into develop* process? Is this a strategy for merging code without causing regressions while the feature is under construction? If that's the case I much prefer a feature toggle because then we don't need to deal with merge hell later on.

saledjenic (approved):

protocol/messenger.go
Samyoul: A blasphemy has been spoken! The ways of Yog-Sothoth are unknowable, beyond even the merest glimpse.

Also, think of this function as a register, we can do some basic refactoring of this function, but fundamentally we will still need to pass a protobuf into something and then make a decision about what to do with that protobuf.

ilmotta: hauhauahau, I think I understand what you mean @Samyoul. But still, this function has too many responsibilities, and it has too many reasons to change. Each case could be handled by individual functions, easily testable in isolation if we so wanted. It would be a win to me if we could make `handleRetrievedMessages` as dumb as possible, so that it essentially becomes a long list of match by type + dispatcher.

A simple extract refactor for the 2 cases added in this PR would be something I would personally do, to start to break free from the holy pattern! But other ideas could follow in the future, so more blasphemy can be spoken.

Some could say, don't change a winning team, so perhaps `handleRetrievedMessages` is such a team ;)

Samyoul: Yes @ilmotta, I agree with you and I have floated the idea of refactoring this function in particular in the past.

- "make the `handleRetrievedMessages` as dumb as possible" - Winning ✅ 
- "Each case could be handled by individual functions" - Winning ✅ 
- "easily testable in isolation" - Winning ✅ 
- "more blasphemy" - [EPIC WINNING!! ✅ ](https://www.youtube.com/watch?v=9QS0q3mGPGg)

qfrank:
> May I ask why we want to merge feat/sync/activity_center_notification2 into feat/sync_activity_center_notification instead of the usual merge into develop process? Is this a strategy for merging code without causing regressions while the feature is under construction? If that's the case I much prefer a feature toggle because then we don't need to deal with merge hell later on.

no, just want PR to be easy to review 🙂 @ilmotta

protocol/messenger_activity_center.go
qfrank: my bad, i'll remove this as it confused you 😅

protocol/messenger_activity_center.go
qfrank: Agree, great point!

protocol/messenger_activity_center.go
ilmotta: It's not as if it's causing a big problem @qfrank, but more on the side of we usually don't need such checks when iterating because most slices are made of non-nil elements. Thanks for taking this in consideration.

protocol/messenger.go
qfrank: Hahaha, actually, i saw this issue when I trying to create my first PR relate to `handleRetrievedMessages` , it's too big! I was trying pick up reflection of tutorial to see what i can do 😅

protocol/messenger_activity_center.go
qfrank: Thank you for your valuable review as always👍

protocol/messenger.go
ilmotta: We are on the same page...

Samyoul: Although I would say that the difficulty(?) with breaking each case into a dedicated function is that each handler is basically the testable unique kernel. But the rest of the functionality is varying forms of pre and post handling, some cases use identical set ups, others use a combination of common elements and other have completely custom handling.

I'd be very interested in seeing the solution / proposals.

Samyoul (approved):
Thank you @qfrank for the work and thought you've put into this PR.

protocol/messenger.go
Samyoul: @qfrank @ilmotta I've added an issue to track this goal https://github.com/status-im/status-go/issues/3602

PR #3585, opened 2023-06-07
chore: `forceFetchingBackup` param added to `RequestAllHistoricMessages` and `RequestAllHistoricMessagesWithRetries` functions

saledjenic:
If `forceFetchingBackup` is set to `true` it will force `syncBackup` function call.

### DIFF
--- a/protocol/messenger_mailserver.go
+++ b/protocol/messenger_mailserver.go
@@ -294,12 +294,12 @@ func (m *Messenger) resetFiltersPriority(filters []*transport.Filter) {
 	}
 }
 
-func (m *Messenger) RequestAllHistoricMessagesWithRetries() (*MessengerResponse, error) {
-	return m.performMailserverRequest(m.RequestAllHistoricMessages)
+func (m *Messenger) RequestAllHistoricMessagesWithRetries(forceFetchingBackup bool) (*MessengerResponse, error) {
+	return m.performMailserverRequest(func() (*MessengerResponse, error) { return m.RequestAllHistoricMessages(forceFetchingBackup) })
 }
 
 // RequestAllHistoricMessages requests all the historic messages for any topic
-func (m *Messenger) RequestAllHistoricMessages() (*MessengerResponse, error) {
+func (m *Messenger) RequestAllHistoricMessages(forceFetchingBackup bool) (*MessengerResponse, error) {
 	shouldSync, err := m.shouldSync()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


ghost:
### Pull Request Checklist

saledjenic:
@cammellos @qfrank I guess that mobile app is using at least one of these 2 functions (`RequestAllHistoricMessagesWithRetries`, `RequestAllHistoricMessages`), so please update them with appropriate parameter in mobile app.

Cuteivist (approved):

alaibe (approved):

protocol/messenger_mailserver.go
```diff
@@ -294,12 +294,12 @@ func (m *Messenger) resetFiltersPriority(filters []*transport.Filter) {
...
 
-func (m *Messenger) RequestAllHistoricMessagesWithRetries() (*MessengerResponse, error) {
-	return m.performMailserverRequest(m.RequestAllHistoricMessages)
+func (m *Messenger) RequestAllHistoricMessagesWithRetries(forceFetchingBackup ...bool) (*MessengerResponse, error) {
+	return m.performMailserverRequest(func() (*MessengerResponse, error) { return m.RequestAllHistoricMessages(forceFetchingBackup...) })
 }
 
 // RequestAllHistoricMessages requests all the historic messages for any topic
-func (m *Messenger) RequestAllHistoricMessages() (*MessengerResponse, error) {
+func (m *Messenger) RequestAllHistoricMessages(forceFetchingBackup ...bool) (*MessengerResponse, error) {
```
qfrank: can we replace `forceFetchingBackup ...bool` with `forceFetchingBackup bool` ?

qfrank:
> @cammellos @qfrank I guess that mobile app is using at least one of these 2 functions (`RequestAllHistoricMessagesWithRetries`, `RequestAllHistoricMessages`), so please update them with appropriate parameter in mobile app.

mobile only used `RequestAllHistoricMessagesWithRetries`, I'll update relate calls on mobile

protocol/messenger_mailserver.go
saledjenic: Well, ofc we can do whatever we want, but this way I made it optional, means if it's provided then provided value will be used, otherwise it defaults to `false`.

qfrank: why we need optional

saledjenic: @qfrank well, firstly the default behavior of this function remains as it was and that is an intended behaviour, setting a `forceFetchingBackup` flag is very specific case, hence not sure why should we bother one who is going to use this function with setting a `forceFetchingBackup` flag, but those who really need it for their work they will know about it and will set it if they really want.   

Why do you think that it's so crucial to make it non optional flag?

qfrank: because optional is not very clear for me, confused me ... it also could be an array,  I think it would be better to pass false/true explictly

saledjenic: > because optional is not very clear for me, confused me

I hope it's clear for you now 😉  your wishes are granted.

qfrank (approved):
I'll align these changes next Monday morning

PR #3588, opened 2023-06-08
feat(BurnCollectibles): Add burn collectibles functionality

### LINKED ISSUE #10816: Burn collectibles logic
Update smart contract and add new functionality to the existing UI.

endulab:
Change smart contract with new API.
Update gas amount for deployment.
Add Burn() and EstimateBurn() functions.
Add RemainingSupply() functions.

**Do not review contracts/collectibles/collectibles.go. It is autogenerated.**

Issue #10816


jrainville (approved):

services/collectibles/api.go
```diff
@@ -235,6 +203,20 @@ func (api *API) RemoteBurn(ctx context.Context, chainID uint64, contractAddress
 	return tx.Hash().Hex(), nil
 }
 
+func (api *API) EstimateRemoteBurn(ctx context.Context, chainID uint64, contractAddress string, tokenIds []*bigint.BigInt) (uint64, error) {
```
dlipicar, line 206: Hmm we won't be able to get this listed in the transactions list on the sender side unless this generates a MultiTransaction (we need a new type like `MultiTransactionDestroy`), since that's not information we can easily fetch from the blockchain (as opposed to the receiver side, where we simply need to look for a transfer to address `0x0`).

endulab: Sorry @dlipicar, I do not understand. Do you mean EstimateRemoteBurn function? It works correctly. What is the problem?

dlipicar: Clarified in chat, it's all fine!

dlipicar (approved):

PR #3589, opened 2023-06-08
chore: resume archives import with delay

### LINKED ISSUE #10937: [Performance] start archives import with delay

osmaczko:
closes: status-im/status-desktop#10937

### DIFF
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -49,12 +49,11 @@ var updateActiveMembersInterval = 24 * time.Hour
 
 const discordTimestampLayout = "2006-01-02T15:04:05+00:00"
 
-var importRateLimiter = rate.NewLimiter(rate.Every(importSlowRate), 1)
-
 const (
 	importSlowRate          = time.Second / 1
 	importFastRate          = time.Second / 100
 	importMessagesChunkSize = 10
+	importInitialDelay      = time.Minute * 5
 )
 
 const (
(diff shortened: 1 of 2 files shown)


protocol/messenger_communities.go
```diff
@@ -2315,6 +2316,9 @@ func (m *Messenger) resumeHistoryArchivesImport(communityID types.HexBytes) erro
 
 	go func() {
 		defer task.Waiter.Done()
+		if delayed {
+			time.Sleep(importResumeDelay)
```
jrainville: Will this hold the thread for 5 minutes? I wonder if it can impact anything (one less thread available?). Though I guess that doing a timer thread would do the same, so I'm probably a bit ignorant in this field.

0x-r4bbit: I think we have two options generally:

1. Use `time.AfterFunc()` to delay the call to `resumeHistoryArchivesImport()`
2. Call `resumeHistoryArchivesImport()` immediately and use `time.Sleep()` as done here

I think 1) would be cleaner, but both are fine IMO.

Only thing to consider, within those 5 minutes, it's very possible that your node receives a magnetlink, and starts processing it, which could leading to import resuming earlier than the 5 minute threshold.

Of course, we can work around that too, but I'd say it's not worth the effort right now..

osmaczko: > Only thing to consider, within those 5 minutes, it's very possible that your node receives a magnetlink, and starts processing it, which could leading to import resuming earlier than the 5 minute threshold.

Good point! I solved it with new approach, please re-check.

> I think 1) would be cleaner, but both are fine IMO.

I tried it but Staticcheck didn't like it:

> S1037-Elaborate way of sleeping: 
> 
> Using a select statement with a single case receiving from the result of time.After is a very elaborate way of  sleeping that can much simpler be expressed with a simple call to time.Sleep.

0x-r4bbit: Ah great! Thanks for running after this!

0x-r4bbit (approved):
Sweet, I like this

protocol/messenger_communities.go
osmaczko: > Will this hold the thread for 5 minutes? I wonder if it can impact anything (one less thread available?). Though I guess that doing a timer thread would do the same, so I'm probably a bit ignorant in this field.

From [docs](https://pkg.go.dev/time#Sleep): "Sleep **pauses** the current goroutine for at least the duration d."

I assume the sleeping goroutine will not hold an OS thread or consume CPU cycles. In other words, the Go scheduler will not execute sleeping goroutines.

jrainville (approved):
Nice

PR #3591, opened 2023-06-08
perf(sqlCipher): Increase cipher page size to 8192

### LINKED ISSUE #10753: [Performance] Test different DB configurations (4096+ page size, new SQLCipher and SQLite versions)
### Description

See https://www.notion.so/Database-related-performance-1dbaa7ea894c461d885e55676683be07?pvs=4#65a0277e8fc243648355ab4478794cb6

Would the page size improve performance? We currently use 1024, but based on docs it was a good choice for 2003 systems. https://www.sqlite.org/pgszchng2016.html 

The page size configuration is one of the breaking changes in newer SQLCipher Versions. Increasing the page size to 4096 would allow us to use SQLCipher 4 with latest SQLite https://www.zetetic.net/blog/2018/11/30/sqlcipher-400-release/.

The goal of this task is to answer the following questions:

- [ ] What are the implications of migrating to latest SQLCipher 4? If migrating the DB to latest SQLCipher is not an option, how about migrating the lib to latest SQLCipher and open the DB in compatibility mode? See https://www.zetetic.net/sqlcipher/sqlcipher-api/#cipher_compatibility
- [ ] Quantify the performance improvement on different scenarios: increased page size, upgraded SQLCipher/SQLite versions, Migrate the DB to latest SQLCipher

alexjba:
Increasing the cipher page size to 8192 requires DB re-encryption. The process is as follows: 
```
//Login to v3 DB
PRAGMA key = 'key';
PRAGMA cipher_page_size = 1024"; // old Page size
PRAGMA cipher_hmac_algorithm = HMAC_SHA1";
PRAGMA cipher_kdf_algorithm = PBKDF2_HMAC_SHA1";
PRAGMA kdf_iter = kdfIterationsNumber";

//Create V4 DB with increased page size
ATTACH DATABASE 'newdb.db' AS newdb KEY 'key';
PRAGMA newdb.cipher_page_size = 8192; // new Page size PRAGMA newdb.cipher_hmac_algorithm = HMAC_SHA1"; // same as in v3 PRAGMA newdb.cipher_kdf_algorithm = PBKDF2_HMAC_SHA1"; // same as in v3 PRAGMA newdb.kdf_iter = kdfIterationsNumber"; // same as in v3 SELECT sqlcipher_export('newdb');
DETACH DATABASE newdb;

//Login to V4 DB
...
```

Worth noting:

- The DB migration will happen on the first successful login. 
- The new DB version will have a different name to be able to distinguish between different DB versions.Versions naming mirrors sqlcipher major version (naming conventions used by sqlcipher), meaning that we're migrating from V3 to V4 DB (even if we're not fully aligned with V4 standards). The DB is not migrated to the v4 standard `SHA512` due to performance reasons. Our custom `SHA1` implementation is fully optimised for performance.
- If the migration is successful a new DB file will be created and the old DB file will be deleted

Closes https://github.com/status-im/status-desktop/issues/10753

### DIFF
--- a/api/geth_backend.go
+++ b/api/geth_backend.go
@@ -248,6 +248,9 @@ func (b *GethStatusBackend) DeleteMultiaccount(keyUID string, keyStoreDir string
 		filepath.Join(b.rootDataDir, fmt.Sprintf("%s.db", keyUID)),
 		filepath.Join(b.rootDataDir, fmt.Sprintf("%s.db-shm", keyUID)),
 		filepath.Join(b.rootDataDir, fmt.Sprintf("%s.db-wal", keyUID)),
+		filepath.Join(b.rootDataDir, fmt.Sprintf("%s-v4.db", keyUID)),
+		filepath.Join(b.rootDataDir, fmt.Sprintf("%s-v4.db-shm", keyUID)),
+		filepath.Join(b.rootDataDir, fmt.Sprintf("%s-v4.db-wal", keyUID)),
 	}
 	for _, path := range dbFiles {
 		if _, err := os.Stat(path); err == nil {
@@ -713,23 +716,12 @@ func (b *GethStatusBackend) ExportUnencryptedDatabase(acc multiaccounts.Account,
 		return errors.New("root datadir wasn't provided")
 	}
 
-	// Migrate file path to fix issue https://github.com/status-im/status-go/issues/2027
-	oldPath := filepath.Join(b.rootDataDir, fmt.Sprintf("app-%x.sql", acc.KeyUID))
-	newPath := filepath.Join(b.rootDataDir, fmt.Sprintf("%s.db", acc.KeyUID))
-
-	_, err := os.Stat(oldPath)
-	if err == nil {
-		err := os.Rename(oldPath, newPath)
-		if err != nil {
-			return err
-		}
-
-		// rename journals as well, but ignore errors
-		_ = os.Rename(oldPath+"-shm", newPath+"-shm")
-		_ = os.Rename(oldPath+"-wal", newPath+"-wal")
+	dbPath, err := b.runDBFileMigrations(acc, password)
+	if err != nil {
+		return err
 	}
 
-	err = appdatabase.DecryptDatabase(newPath, directory, password, acc.KDFIterations)
+	err = appdatabase.DecryptDatabase(dbPath, directory, password, acc.KDFIterations)
 	if err != nil {
 		b.log.Error("failed to initialize db", "err", err)
 		return err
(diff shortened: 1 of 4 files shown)


jrainville (commented):
Looks good. Does this migration take a lot of time like the password migration? If so, we need to make sure we can show a loading screen for users.

alexjba:
> Looks good. Does this migration take a lot of time like the password migration? If so, we need to make sure we can show a loading screen for users.

It takes less than 1 min on a 6GB DB. Do you think we need a custom loading screen for this, or the regular will do?
I we need a custom loading screen I guess first we need to implement signals in status-go to notify when the encryption starts and when it ends.

alexjba:
Fixed failing tests: https://github.com/status-im/status-go/pull/3591/commits/9c335c2147f9e9ec58370ae11c89a602e6f44496

jrainville:
> It takes less than 1 min on a 6GB DB. Do you think we need a custom loading screen for this, or the regular will do?
I we need a custom loading screen I guess first we need to implement signals in status-go to notify when the encryption starts and when it ends.

If by normal screen, you mean the basic spinner with no text when we login, then 100% we need a new one (or the one we created for password re-encryption).  

The reason is that if users are used to login in 10 seconds, as soon as it takes 30 seconds, they'll think something went badly and they'll close, causing the app to break.

I know that, because that's exactly what happened last time when the loading screen didn't work

alexjba:
> > It takes less than 1 min on a 6GB DB. Do you think we need a custom loading screen for this, or the regular will do?
> > I we need a custom loading screen I guess first we need to implement signals in status-go to notify when the encryption starts and when it ends.
> 
> If by normal screen, you mean the basic spinner with no text when we login, then 100% we need a new one (or the one we created for password re-encryption).
> 
> The reason is that if users are used to login in 10 seconds, as soon as it takes 30 seconds, they'll think something went badly and they'll close, causing the app to break.
> 
> I know that, because that's exactly what happened last time when the loading screen didn't work

The app shouldn't break because we're creating a new DB file for the re-encryption and only delete the old one if it succeeds.
But it sounds like a good approach nonetheless. On it! 👍

alexjba:
> > It takes less than 1 min on a 6GB DB. Do you think we need a custom loading screen for this, or the regular will do?
> > I we need a custom loading screen I guess first we need to implement signals in status-go to notify when the encryption starts and when it ends.
> 
> If by normal screen, you mean the basic spinner with no text when we login, then 100% we need a new one (or the one we created for password re-encryption).
> 
> The reason is that if users are used to login in 10 seconds, as soon as it takes 30 seconds, they'll think something went badly and they'll close, causing the app to break.
> 
> I know that, because that's exactly what happened last time when the loading screen didn't work

We have now two events to signal the re-encryption start/end: https://github.com/status-im/status-go/pull/3591/commits/2e0d71d0f9233cb61678206bf179782622ffd5a5

Edit: I've sent the events on password change as well. It's also a re-encryption under the hood.

alexjba:
Hey @jrainville @cammellos @osmaczko @richard-ramos, please please review when you get the chance 🍻
Would be nice to land this in status-desktop 0.13

osmaczko (approved):
nice

jrainville (approved):

PR #3592, opened 2023-06-08
Make sure the image URL is returned when a replied message is deleted

cammellos:
When we reply to our own message with an image, we didn't set the URL of the image of the message, which resulted in the image not being displayed correctly.


qfrank (approved):
LGTM

protocol/messenger_delete_message_test.go
```diff
@@ -518,3 +478,58 @@ func (s *MessengerDeleteMessageSuite) TestDeleteMessageAndChatIsAlreadyRead() {
...
+	// create an http server
+	mediaServer, err := server.NewMediaServer(nil, nil, nil)
+	s.Require().NoError(err)
+	s.Require().NotNil(mediaServer)
+	s.Require().NoError(mediaServer.Start())
+
+	theirMessenger.httpServer = mediaServer
+
+	// We reply to our own message with an image
+	image, err := buildImageWithoutAlbumIDMessage(*theirChat)
```
qfrank: oh, can we rename `image` to `imageMessage`?

cammellos: good point

jrainville (approved):

protocol/messenger_messages.go
```diff
@@ -229,7 +229,9 @@ func (m *Messenger) DeleteMessageAndSend(ctx context.Context, messageID string)
 		if err != nil {
 			return nil, err
 		}
+
 		response.AddMessages(updatedMessages)
+		m.prepareMessages(response.messages)
```
Samyoul, line 234: That's all it took? 😆

Samyoul (approved):

PR #3593, opened 2023-06-09
feat(wallet): make filter API async

stefandunca:
### Closes status-desktop [#10994](https://github.com/status-im/status-desktop/issues/10994)

Refactor the filter interface into an async call which returns the result using a wallet event.
A call to the filter API will cancel the ongoing filter and receive an error result event.
The filter also returns if there might be more info based on the heuristic of the query returning fewer results than the requested batch
For every call to the filter API, an async response is guaranteed success or failure.
Provide error codes for error state interpretation

Closes status-desktop #10994

### DIFF
--- a/services/wallet/activity/service.go
+++ b/services/wallet/activity/service.go
@@ -1,0 +1,123 @@
+package activity
+
+import (
+	"context"
+	"database/sql"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/event"
+	"github.com/ethereum/go-ethereum/log"
+
+	w_common "github.com/status-im/status-go/services/wallet/common"
+	"github.com/status-im/status-go/services/wallet/walletevent"
+)
+
+const (
+	// FilterResponse json is sent as a message in the EventActivityFilteringDone event
+	EventActivityFilteringDone walletevent.EventType = "wallet-activity-filtering-done"
+)
+
+type Service struct {
+	db        *sql.DB
+	eventFeed *event.Feed
+
+	context  context.Context
+	cancelFn context.CancelFunc
+	wg       sync.WaitGroup
+	mu       sync.Mutex
+}
+
+func NewService(db *sql.DB, eventFeed *event.Feed) *Service {
+	return &Service{
+		db:        db,
+		eventFeed: eventFeed,
+	}
+}
+
+type ErrorCode = int
+
+const (
+	ErrorCodeSuccess ErrorCode = iota + 1
+	ErrorCodeFilterCanceled
+	ErrorCodeFilterFailed
+)
+
+type FilterResponse struct {
+	Activities       []Entry   `json:"activities"`
+	ThereMightBeMore bool      `json:"thereMightBeMore"`
+	ErrorCode        ErrorCode `json:"errorCode"`
+}
+
+// FilterActivityAsync allows only one filter task to run at a time
+// and it cancels the current one if a new one is started
+// All calls will trigger an EventActivityFilteringDone event with the result of the filtering
+func (s *Service) FilterActivityAsync(ctx context.Context, addresses []common.Address, chainIDs []w_common.ChainID, filter Filter, offset int, limit int) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	// If a previous task is running, cancel it and wait to finish
+	if s.cancelFn != nil {
+		s.cancelFn()
+		s.wg.Wait()
+	}
+
+	if ctx.Err() != nil {
+		return fmt.Errorf("context error: %w", ctx.Err())
+	}
+
+	s.context, s.cancelFn = context.WithCancel(context.Background())
+
+	s.wg.Add(1)
+	go func() {
+		defer s.wg.Done()
+		defer func() {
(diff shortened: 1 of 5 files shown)


services/wallet/activity/service.go
```diff
@@ -0,0 +1,123 @@
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	// If a previous task is running, cancel it and wait to finish
+	if s.cancelFn != nil {
+		s.cancelFn()
+		s.wg.Wait()
+	}
+
+	if ctx.Err() == nil {
```
Cuteivist: Minor thing: Arrow anti-pattern can be prevented by simply 
```suggestion
	if ctx.Err() != nil {
	    return nil
	}
```

Cuteivist (approved):
Awesome job, just one personal minor styling thing. I approve it anyway.

services/wallet/activity/service.go
```diff
@@ -0,0 +1,123 @@
...
+		s.context, s.cancelFn = context.WithCancel(context.Background())
```
IvanBelyakoff: I think we should have some timeout for it. If for some reason we can't filter within some reasonable time -15 secs for example, because DB is locked or whatever other reason, user should see at least a timeout error and then they can report it. If they don't see anything, app just freezes, that is a worse UX I believe.

services/wallet/activity/service.go
```diff
@@ -0,0 +1,123 @@
...
+		s.wg.Wait()
+		s.cancelFn = nil
+	}
+}
+
+func (s *Service) sendResponseEvent(response FilterResponse) {
+	payload, err := json.Marshal(response)
+	if err != nil {
+		log.Error("Error marshaling response: %v", err)
+		payload = []byte(fmt.Sprintf(`{"errorCode": %d}`, ErrorCodeServiceFailed))
```
IvanBelyakoff: I think that if FilterResponse is refactored, marshalling it to json will still be as expected if we use FilterResponse here. Having it as directly as bytearray is more error prone and we might forget to handle it.

IvanBelyakoff (approved):
Great job, Stefan. Some minor comments, feel free to address them if you agree

services/wallet/activity/service.go
stefandunca: It would be nice to address this corner case. I would like to wait and see if this is a real issue in practice. Moving forward, we will index all the filterable columns and extract all required metadata in the same async call. This should speed up the process and cover the most extreme corner case.

Regarding UX, the user will see loading animations and switching from the activity tab or on filter reset the current long call will be aborted.

I would suggest waiting for field testing and seeing if this is a genuine concern before addressing it.

services/wallet/activity/service.go
stefandunca: I agree this error is an extreme corner-case, and I will remove it and rely on the empty response to signal this error-state.

services/wallet/activity/service.go
IvanBelyakoff: I see. I was not talking about the performance of the request rather then about other issues that we might have with DB. basically that is not hard to handle somehow that error, then just hope it will never happen

stefandunca: I was including stability concerns in the performance case, also. However, if the DB contention is too high or we have unexpected behaviour, we must address it instead of reporting the error to the user.

> ... about other issues that we might have with DB.

Without a specific use case, it looks like premature optimization. What do you think?

IvanBelyakoff: Well, when we are in production, users will be the best testers and might catch the errors more often then QA does due to a variety of reasons, so if we don't have a way for them to report the exact problem - like they see an error in UI, then they just report - app freezes. Ok, it tells us nothing. So we must handle all the issues we can find ofc, but we can't find them all, then we need users to report the issues. When I was on Windows, I used to see a lot of "Error <Some number here>" popups and I could google them and address them. So I do feel like propagating them is like super important.

Regarding premature optimization. Here is the case - currently we have not even switched to a separate DB, so we  can be all stuck if messenger team or one of our requests suddenly takes the write lock for long. Again, if users catch it, and see the error they can report it.

Well, as I see it, we should not rely on being able to catch all issues but have them all logged and propagated, when it matters for users like  why they are waiting forever to filter the activities.

stefandunca: I agree with you that it will be ideal to handle as many corner cases as possible. The question where do you draw the line? In this case, waiting for a filter query for more than expected is a good enough indication that something is at fault. For further escalation, we will see in the API logs that a filter was called, which never received a response.

> Regarding premature optimization. Here is the case - currently we have not even switched to a separate DB, so we can be all stuck if messenger team or one of our requests suddenly takes the write lock for long. Again, if users catch it, and see the error they can report it.

This is a specific generic DB performance problem that is in work already. AFAIK we have tried to enable concurrent calls optimization that didn't work and next, there is a plan to address it as you suggested.

> Well, as I see it, we should not rely on being able to catch all issues but have them all logged and propagated, when it matters for users like why they are waiting forever to filter the activities.

What do you think will be a good timeout for this DB call? Which is the extreme expected data corner case and how long will the user expect to wait for a result?

stefandunca: Hey @IvanBelyakoff, I created this [issue](https://github.com/status-im/status-desktop/issues/11036) to track and address the concerns raised by this conversation. Let's continue our discussion there.

PR #3594, opened 2023-06-09
Local pairing keycard

saledjenic:
First commit works fine, but in order to complete the task we need second commit also which transfers content of `pairings.json` from sender to receiver.

There is a video in in [this comment](https://github.com/status-im/status-desktop/pull/11015#issue-1750035856), displaying how it works if user is pairing with the device where Status user is a keycard user.


alaibe (approved):

Samyoul (approved):
Really nice work @saledjenic , I really liked the video in the desktop PR 👨‍🍳 🤌 💋

PR #3605, opened 2023-06-12
feat: introduce `CheckChannelPermissions` API

0x-r4bbit:
Similar to `CheckPermissionToJoin()` we now get
a `CheckChannelPermissions()` API.

It will rely on the same `PermissionResponse` types, but gives information about both `ViewOnlyPermissions` and
`ViewAndPostPermissions`.

### DIFF
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -1497,6 +1497,25 @@ func (o *Community) TokenPermissionsByType(permissionType protobuf.CommunityToke
 	return permissions
 }
 
+func (o *Community) ChannelTokenPermissionsByType(channelID string, permissionType protobuf.CommunityTokenPermission_Type) []*protobuf.CommunityTokenPermission {
+	permissions := make([]*protobuf.CommunityTokenPermission, 0)
+	for _, tokenPermission := range o.TokenPermissions() {
+		if tokenPermission.Type == permissionType && includes(tokenPermission.ChatIds, channelID) {
+			permissions = append(permissions, tokenPermission)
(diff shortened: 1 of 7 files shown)


protocol/communities/community.go
```diff
@@ -1451,6 +1451,25 @@ func (o *Community) TokenPermissionsByType(permissionType protobuf.CommunityToke
...
+	permissions := make([]*protobuf.CommunityTokenPermission, 0)
+	for _, tokenPermission := range o.TokenPermissions() {
+		if tokenPermission.Type == permissionType && includes(tokenPermission.ChatIds, channelID) {
+			permissions = append(permissions, tokenPermission)
+		}
+	}
+	return permissions
+}
+
+func includes(channelIDs []string, channelID string) bool {
```
mprakhov, line 1510: I think it would be nice to add this fn as a lambda inside ChannelTokenPermissionsByType fn?

mprakhov (approved):

protocol/communities/community.go
0x-r4bbit: Hm... that'd be redefining the function for every call though.. 
I can do it if there's strong opinions

jrainville: No strong opinion whichever way

jrainville (approved):
LGTM. Good job

0x-r4bbit:
LoginWithKey test keeps failing on CI (unrelated to this PR).
Have restarted it 3 or 4 times now

PR #3607, opened 2023-06-12
feat: Make DB re-encryption fault proof

### LINKED ISSUE #10353: Make DB re-encryption fault proof with a back-up
### Description

Right, if the Db re-encryption is in progress, if the app is closed by any means, the DB then becomes kaput.

The solution is to first copy the DB, then re-encrypt the new copy. When it's all done, we can rename the DB to the original name, overwriting the first one.

That way, if the app is closed, the original DB is not touched.

**Alternative option**: if that doesn't work because we can only re-encrypt the active DB, we should still create a backup. If the app is closed and the DB is broken, we can then take the backup and use that one instead.
Make sure to delete the backup at the end of any of the successful endings, that way we do not use an old backup by accident.

### Acceptance Criteria

When doing a DB re-encryption by changing the password, it is possible to close the app meanwhile and then reopen. 
The state should then come back to before the re-encryption, with the old password

alexjba:
Improve error management to avoid DB corruption in case of failure/app exit.

Changes:
Use sqlcipher_export instead of rekey to change the DB password. The advantage is that sqlcipher_export will operate on a new DB file and we don't need to modify the current account until the export is successful. Keeping the rekey requires to create a DB copy before trying to re-encrypt the DB, but the DB copy is risky in case the DB file changes while the copy is in progress. It could also lead to DB corruption.

Closes https://github.com/status-im/status-desktop/issues/10353

### DIFF
--- a/api/geth_backend.go
+++ b/api/geth_backend.go
@@ -749,8 +749,7 @@ func (b *GethStatusBackend) ImportUnencryptedDatabase(acc multiaccounts.Account,
 	return nil
 }
 
-func (b *GethStatusBackend) ChangeDatabasePassword(keyUID string, password string, newPassword string) error {
-	dbPath := filepath.Join(b.rootDataDir, fmt.Sprintf("%s-v4.db", keyUID))
+func (b *GethStatusBackend) reEncryptKeyStoreDir(currentPassword string, newPassword string) error {
 	config := b.StatusNode().Config()
 	keyDir := ""
 	if config == nil {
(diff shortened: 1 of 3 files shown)


ghost:
### Pull Request Checklist

jrainville (approved):
LGTM. 

> Keeping the rekey requires to create a DB copy before trying to re-encrypt the DB, but the DB copy is risky in case the DB file changes while the copy is in progress. It could also lead to DB corruption.

To be sure I understand, what you're saying here is that if we kept the old way to re-encrypt and just used a copy of the DB, we would have the problems above? And your solution doesn't have those problems? 
If so, great, nice find!

alexjba:
> LGTM.
> 
> > Keeping the rekey requires to create a DB copy before trying to re-encrypt the DB, but the DB copy is risky in case the DB file changes while the copy is in progress. It could also lead to DB corruption.
> 
> To be sure I understand, what you're saying here is that if we kept the old way to re-encrypt and just used a copy of the DB, we would have the problems above? And your solution doesn't have those problems? If so, great, nice find!

Yes, based on sqlite docs copying the DB while there is an open connection does not guarantee the backup integrity. SQLite has developed specific APIs for this and so did sqlcipher on top of sqlite API (https://www.sqlite.org/howtocorrupt.html, https://www.sqlite.org/backup.html).
I guess it could have been done in a safe way, but I found it too complicated to mitigate the risk of corruption when the app is closed or the DB changes.

The disadvantage of the implemented approach is that we need to update the sql.DB pointer in all status-go services in order to keep the user logged in.

richard-ramos (approved):

anastasiyaig:
@alexjba can we add corresponding PR to status-desktop repo to bump the version? Also, any hint how this change could be tested?

api/geth_backend.go
```diff
@@ -760,25 +759,74 @@ func (b *GethStatusBackend) ChangeDatabasePassword(keyUID string, password strin
...
+
+	err = b.reEncryptKeyStoreDir(password, newPassword)
+	if err != nil {
+		return err
+	}
+
+	// Replacing the old database with the new one requires closing all connections to the database
+	// This is done by stopping the node and restarting it with the new DB
+	appDBPath, _ := appdatabase.GetDBFilename(b.appDB)
+	changeCurrentAccountPassword := appDBPath == dbPath
```
saledjenic, line 806: @alexjba sorry for the late check, but why do we need `changeCurrentAccountPassword` here? I mean, we're here only when we actually want to change the pass? 

Also if your intention was, if in case there are more users to let all but the logged in user to change the pass and in case of logged in user to log him out first and then change the pass, that's not needed, cause the only way yo change the pass (reencrypt the db) is to be logged in.

alexjba: > I mean, we're here only when we actually want to change the pass?

Yes, that's true.

I was thinking it might not be needed with the current flows we have. But since the API is not clear on this, I didn't want to make an assumption here.

My intention was to allow password change without requiring login. Looking at the API, it can receive a keyUID, old password, new password. This means to me the  password could be changed for any keyUID as long as the old password is correct. And I could also see some reasoning behind this. Eg. Changing password from logins screen or Deep links. 

Now since you've challenged me to think more about it, I see there's a bug here. Changing the password won't work if the user is not logged-in because the keystore dir re-encryption will probably fail as it's using `b.accountManager.Keydir`

PR #3580, opened 2023-06-07
feat: always request revealed addresses for joining communities

MishkaRogachev:
Waits https://github.com/status-im/status-go/pull/3569

For https://github.com/status-im/status-desktop/pull/10967
Issue: https://github.com/status-im/status-desktop/issues/10853

Important changes:
- [x] Always request revealed addresses for joining communities

PR #3609, opened 2023-06-13
fix: save revealed accounts regardless of permissions granted

MishkaRogachev:
For https://github.com/status-im/status-desktop/pull/10967

Important changes:
- [x] save revealed accounts regardless of permissions granted

This PR does not contain checks for provided addresses, this should be fixed in https://github.com/status-im/status-go/pull/3580 after release

### DIFF
--- a/account/accounts.go
+++ b/account/accounts.go
@@ -62,7 +62,14 @@ type RecoverParams struct {
 	Signature string `json:"signature"`
 }
 
-// Manager represents account manager interface.
+// Interface represents account manager interface
+type Interface interface {
+	GetVerifiedWalletAccount(db *accounts.Database, address, password string) (*SelectedExtKey, error)
+	Sign(rpcParams SignParams, verifiedAccount *SelectedExtKey) (result types.HexBytes, err error)
+	Recover(rpcParams RecoverParams) (addr types.Address, err error)
+}
+
+// Manager represents account manager implementation
 type Manager struct {
 	mu         sync.RWMutex
 	rpcClient  *rpc.Client
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -69,7 +69,7 @@ type Manager struct {
 	subscriptions                  []chan *Subscription
 	ensVerifier                    *ens.Verifier
 	identity                       *ecdsa.PrivateKey
-	accountsManager                *account.GethManager
+	accountsManager                account.Interface
 	tokenManager                   TokenManager
 	logger                         *zap.Logger
 	stdoutLogger                   *zap.Logger
@@ -122,7 +122,7 @@ func (t *HistoryArchiveDownloadTask) Cancel() {
 }
 
 type managerOptions struct {
-	accountsManager      *account.GethManager
+	accountsManager      account.Interface
 	tokenManager         TokenManager
 	walletConfig         *params.WalletConfig
 	openseaClientBuilder openseaClientBuilder
@@ -168,7 +168,7 @@ func (m *DefaultTokenManager) GetBalancesByChain(ctx context.Context, accounts, 
 
 type ManagerOption func(*managerOptions)
 
-func WithAccountManager(accountsManager *account.GethManager) ManagerOption {
+func WithAccountManager(accountsManager account.Interface) ManagerOption {
 	return func(opts *managerOptions) {
 		opts.accountsManager = accountsManager
 	}
@@ -1571,15 +1571,15 @@ func (m *Manager) AcceptRequestToJoin(request *requests.AcceptRequestToJoinCommu
 
 	becomeAdminPermissions := community.TokenPermissionsByType(protobuf.CommunityTokenPermission_BECOME_ADMIN)
 	becomeMemberPermissions := community.TokenPermissionsByType(protobuf.CommunityTokenPermission_BECOME_MEMBER)
-	revealedAccounts := make([]*protobuf.RevealedAccount, 0)
+
+	revealedAccounts, err := m.persistence.GetRequestToJoinRevealedAddresses(dbRequest.ID)
+	if err != nil {
+		return nil, err
+	}
 
 	memberRole := protobuf.CommunityMember_ROLE_NONE
 
 	if len(becomeMemberPermissions) > 0 || len(becomeAdminPermissions) > 0 {
-		revealedAccounts, err := m.persistence.GetRequestToJoinRevealedAddresses(dbRequest.ID)
-		if err != nil {
-			return nil, err
-		}
 
 		accountsAndChainIDs := revealedAccountsToAccountsAndChainIDsCombination(revealedAccounts)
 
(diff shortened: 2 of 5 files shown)


jrainville (approved):

protocol/communities_messenger_test.go
```diff
@@ -126,33 +175,52 @@ func (s *MessengerCommunitiesSuite) newMessengerWithOptions(shh types.Waku, priv
...
+	// add wallet account with keypair
+	kp := accounts.GetProfileKeypairForTest(false, true, false)
+	kp.Accounts[0].Address = types.HexToAddress(walletAddress)
+	err = m.settings.SaveOrUpdateKeypair(kp)
+	s.Require().NoError(err)
+
+	walletAccounts, err := m.settings.GetAccounts()
+	s.Require().NoError(err)
+	s.Require().Len(walletAccounts, 1)
+	//	s.Require().Equal(walletAccounts[0].Type, accounts.AccountTypeGenerated)
```
jrainville: Leftover?

jrainville (approved):

protocol/communities/manager.go
```diff
@@ -1783,6 +1783,38 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request
 		return nil, err
 	}
 
+	// Save revealed addresses + signatures so they can later be added
+	// to the community member list when the request is accepted
+	if len(request.RevealedAccounts) > 0 {
+		err = m.persistence.SaveRequestToJoinRevealedAddresses(requestToJoin)
```
mprakhov: I think this should be at the end, only when account was verified

protocol/communities/manager.go
```diff
@@ -1845,7 +1845,38 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request
...
-		return requestToJoin, nil
+	}
+
+	// Save revealed addresses + signatures so they can later be added
+	// to the community member list when the request is accepted
+	if len(request.RevealedAccounts) > 0 {
+		err = m.persistence.SaveRequestToJoinRevealedAddresses(requestToJoin)
+		if err != nil {
+			return nil, err
+		}
```
0x-r4bbit: This looks a big backwards to me. Shouldn't we only save the revealed addresses, **after** we've verified the signatures?

protocol/communities_messenger_test.go
```diff
@@ -3585,3 +3654,27 @@ func (s *MessengerCommunitiesSuite) TestHandleImport() {
...
+	s.advertiseCommunityTo(community, s.bob)
+
+	s.joinCommunity(community, s.alice, AlicePassword)
+	s.joinCommunity(community, s.bob, BobPassword)
+
+	joinedCommunities, err := s.admin.communitiesManager.Joined()
+	s.Require().NoError(err)
+	s.Require().Len(joinedCommunities, 1)
+	community = joinedCommunities[0]
+	s.Require().NotNil(community)
```
0x-r4bbit: Another way to do this would be to do `s.admin.GetCommunityByID(community.ID())`

protocol/communities_messenger_test.go
```diff
@@ -3585,3 +3654,27 @@ func (s *MessengerCommunitiesSuite) TestHandleImport() {
...
+
+	s.Require().Equal(3, community.MembersCount())
+
+	for pubKey, member := range community.Members() {
+		if pubKey != common.PubkeyToHex(&s.admin.identity.PublicKey) {
+			s.Require().Len(member.RevealedAccounts, 1)
+			s.Require().Equal(member.RevealedAccounts[0].Address, walletAddress)
+		}
```
0x-r4bbit, line 3675: Since we're iterating over members here, do we expect both, `bob` and `alice` to reveal the exact same `walletAddress`?

protocol/communities/manager.go

protocol/communities_messenger_test.go
MishkaRogachev: Yes, i just what to check here we are actually joined

0x-r4bbit: `joined()` will return the communities you're joined. But admin has created that community, so it will always be joined.

protocol/communities_messenger_test.go
MishkaRogachev: Yes, it is done consciously. The problem is in `communities/manager.go:1860`:
```
	recovered, err := m.accountsManager.Recover(recoverParams)
	if err != nil {
		return nil, err
	}
	if recovered.Hex() != revealedAccount.Address {
```
We can have different wallet addresses in the test, but that requires forging that address every time we ask to join with a valid wallet address, and that's a pretty painful change. Moreover, tests becoming *more* fragile, because we are trying to fake behaviour of an object, that is far away from issue's scope

MishkaRogachev: The right way i see is to change `AccountManager::Recover` api to embed verification inside, but i would wait for @saledjenic to check it with him

protocol/communities_messenger_test.go
MishkaRogachev: fixed

mprakhov (approved):

0x-r4bbit (approved):

account/accounts.go
```diff
@@ -62,7 +62,14 @@ type RecoverParams struct {
 	Signature string `json:"signature"`
 }
 
-// Manager represents account manager interface.
+// Interface represents account manager interface
+type Interface interface {
+	GetVerifiedWalletAccount(db *accounts.Database, address, password string) (*SelectedExtKey, error)
+	Sign(rpcParams SignParams, verifiedAccount *SelectedExtKey) (result types.HexBytes, err error)
+	Recover(rpcParams RecoverParams) (addr types.Address, err error)
+}
```
Samyoul, line 70: This interface name doesn't describe the kind of interface, it is already declared as an interface type so this information doesn't add anything to the code.

If we look at `stdlib` interfaces they have names like `JSONMarshaller`, `Reader`, `Writer` and `Stringer`. The convention is to name the interfaces after the action they perform.

In this case the interfaces `Get`s a `VerifiedWalletAccount`, `Sign`s and `Recover`s, the name would be something that reflects this functionality. Example `AccountManager`, `KeyManager` or `KeyHandler`

protocol/communities_messenger_test.go
```diff
@@ -63,9 +108,9 @@ func (s *MessengerCommunitiesSuite) SetupTest() {
 	s.shh = gethbridge.NewGethWakuWrapper(shh)
 	s.Require().NoError(shh.Start())
 
-	s.admin = s.newMessenger()
-	s.bob = s.newMessenger()
-	s.alice = s.newMessenger()
+	s.admin = s.newMessenger(AdminPassword)
+	s.bob = s.newMessenger(BobPassword)
+	s.alice = s.newMessenger(AlicePassword)
```
Samyoul, line 113: For the tests do we need to define the password at this level? Could passwords be an empty string or the same password used all messengers? 

Function signatures have been changed to accommodate the concept of password, but the password is not the testable element.

Samyoul (commented):
@MishkaRogachev Thank you for this PR, I appreciate it is already been merged, but perhaps you could give some thought to the points I've raised here. Thank you again 😄

account/accounts.go
MishkaRogachev: Sure, i was confused by name `Manager`, so i  thought `accounts.Interface` and `accounts.Manager` is a good pair.
And i created an issue for fixing stuff after this PR: https://github.com/status-im/status-go/issues/3625

protocol/communities_messenger_test.go
MishkaRogachev: The problem is that join to community requires password, and if its not match the behaviour should be different.  An i suppose we should test that. If i set same password for each community member, test may not cover some edge case.

PR #3611, opened 2023-06-13
Handle race condition when sending envelopes

cammellos:
It happens that an envelope is sent before it's tracked, resulting in long delays before the envelope is marked as sent. This commit changes the behavior of the code so that order is now irrelevant.


siddarthkay (approved):
Looks like a good improvement!

smohamedjavid (approved):
LGTM 🚀

ibrkhalil (approved):

protocol/messenger.go
```diff
@@ -2090,7 +2090,7 @@ func (m *Messenger) SendChatMessages(ctx context.Context, messages []*common.Mes
 	return &response, nil
 }
 
-// SendChatMessage takes a minimal message and sends it based on the corresponding chat
+// sendChatMessage takes a minimal message and sends it based on the corresponding chat
```
Samyoul, line 2094: ❤️

Samyoul (approved):

vkjr (approved):

PR #3612, opened 2023-06-13
feat(wallet) filter activity by erc20

stefandunca:
### Updates status-desktop [#11025](https://github.com/status-im/status-desktop/issues/11025)

required by status-desktop [#11062](https://github.com/status-im/status-desktop/pull/11062)

Main changes:
    
- Refactor activity API to propagate token identities.
- Extend service to convert token identities to symbols for filtering multi-transaction
- Filter transfers, pending_transactions and multi-transactions based on the provided token identities
- Return involved token identities in activity API
- Test token filtering

Also:

- Fixed calling cancel on a filer activity completed task to release resources

Notes:

- Found limitations with the token identity, which complicates things by not allowing filtering by token groups (like token-code does)


services/wallet/activity/activity.go
```diff
@@ -96,7 +96,8 @@ func newActivityEntryWithTransaction(pending bool, transaction *transfer.Transac
 		timestamp:      timestamp,
 		activityType:   activityType,
 		activityStatus: activityStatus,
-		tokenType:      AssetTT,
+		// TODO: update tokenType
+		tokenType: Native,
```
dlipicar: Maybe this should be two fields "inToken" and "outToken"? Since swap will involve two different ones. Simple transfers should just fill in the appropriate one (in for Receive, out for Send). WDYT?
Same as with the chainID, the bridge operation will involve two different ones ("inChainID" and "outChainID")

services/wallet/token/token.go
```diff
@@ -186,6 +191,21 @@ func (tm *Manager) FindToken(network *params.Network, tokenSymbol string) *Token
 	return nil
 }
 
+func (tm *Manager) LookupTokenIdentity(chainID uint64, address common.Address, native bool) *Token {
+	network := tm.networkManager.Find(chainID)
+	if native {
+		return tm.ToToken(network)
+	}
+
+	allTokens := tm.getFullTokenList(chainID)
```
dlipicar: if I'm not missing something, this second part is identical to `FindTokenByAddress`

services/wallet/activity/activity.go
dlipicar: Actually wondering if we need to have the Native-Erc20 distinction instead of calling them something like "Fungible" with the "Native" option simply being represented by symbol "ETH" or an empty contract address. Other than the actual transactions being very different, I think we treat both transfer types identically.


With this approach, we would have:
-Fungible (ETH+ERC20) -> Identified by contract address or symbol name. Transaction uses `Amount` field
-Non-Fungible (ERC721) -> Identified by contract address and tokenID. Doesn't use `Amount` field (only 1 token per tokenID in existence)
-Semi-Fungible (ERC1155) -> Identified by contract address and tokenID. Uses `Amount` field

With this new definition, the `tokenType` is unique for each activty entry (instead of the current approach, where you can swap `Native` and `erc20` tokens). 

Just throwing ideas around!

stefandunca: > Maybe this should be two fields "inToken" and "outToken"?

Sounds great. However, won't `sentToken` and `receivedToken` bring more clarity?

> Actually wondering if we need to have the Native-Erc20 ...

I think I prefer the current explicit approach; the others are simpler but hide complexity by requiring to have the understanding you mentioned. Not sure if there are other benefits to the fungible/non-semi-fungible approach

services/wallet/token/token.go
stefandunca: 😱 :) Thanks for the heads up

IvanBelyakoff (approved):
LGTM

PR #3614, opened 2023-06-14
feat(wallet): Improved performance of fetching transfers for Sequential fetch strategy

IvanBelyakoff:
Before:
 - block fetching commands for different accounts were in the same wait group, making them dependent on each iteration.
 - transfers loading command was checking database for new unloaded blocks on timeout and was in the same wait group with block fetching, so it was often blocked until all block fetching commands finish for iteration.
 
Now:
 - block fetching commands run independently for each account
 - transfers fetching command is run once on startup for unloaded blocks from DB
 - fetch history blocks commands are launched once on startup for accounts with no full history loaded
 - transfers are loaded on each iteration of block range check without waiting for all ranges to be checked

`loadAllTransfersCommand` removed as being redundant now - replaced with updated `loadTransfersCommand`
Logs improvement

### DIFF
--- a/services/wallet/transfer/commands_sequential.go
+++ b/services/wallet/transfer/commands_sequential.go
@@ -15,6 +15,53 @@ import (
 	"github.com/status-im/status-go/services/wallet/walletevent"
 )
 
+type findNewBlocksCommand struct {
+	*findBlocksCommand
+}
+
+func (c *findNewBlocksCommand) Command() async.Command {
+	return async.InfiniteCommand{
+		Interval: 13 * time.Second, // TODO - make it configurable based on chain block mining time
+		Runable:  c.Run,
+	}.Run
+}
+
+func (c *findNewBlocksCommand) Run(parent context.Context) (err error) {
+	log.Debug("start findNewBlocksCommand", "account", c.account, "chain", c.chainClient.ChainID, "noLimit", c.noLimit)
+
+	headNum, err := getHeadBlockNumber(parent, c.chainClient)
+	if err != nil {
+		// c.error = err
+		return err // Might need to retry a couple of times
+	}
+
+	blockRange, err := loadBlockRangeInfo(c.chainClient.ChainID, c.account, c.blockRangeDAO)
+	if err != nil {
+		log.Error("findBlocksCommand loadBlockRangeInfo", "error", err)
+		// c.error = err
+		return err // Will keep spinning forever nomatter what
+	}
+
+	if blockRange != nil {
+		c.fromBlockNumber = new(big.Int).Add(blockRange.LastKnown, big.NewInt(1))
+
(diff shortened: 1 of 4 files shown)


services/wallet/transfer/commands_sequential.go
```diff
@@ -15,6 +15,53 @@ import (
...
+		Runable:  c.Run,
+	}.Run
+}
+
+func (c *findNewBlocksCommand) Run(parent context.Context) (err error) {
+	log.Debug("start findNewBlocksCommand", "account", c.account, "chain", c.chainClient.ChainID, "noLimit", c.noLimit)
+
+	headNum, err := getHeadBlockNumber(parent, c.chainClient)
+	if err != nil {
+		// c.error = err
```
stefandunca, line 34: Any plans to remove this and the following commented assignments?

services/wallet/transfer/commands_sequential.go
```diff
@@ -15,6 +15,53 @@ import (
 	"github.com/status-im/status-go/services/wallet/walletevent"
 )
 
+type findNewBlocksCommand struct {
+	*findBlocksCommand
+}
+
+func (c *findNewBlocksCommand) Command() async.Command {
+	return async.InfiniteCommand{
+		Interval: 13 * time.Second, // TODO - make it configurable based on chain block mining time
```
stefandunca, line 24: I'm thinking that we should start linking these TODOs with tasks to be tracked, otherwise might pile up. Basically, mentioning it in a follow-up GH issue and the other way around. What do you think?

stefandunca (approved):
Didn't check the logic but the code looks good. Also, please see some questions inline.

services/wallet/transfer/commands_sequential.go
IvanBelyakoff: Good point @stefandunca. I certainly planned it. Created an [issue](https://github.com/status-im/status-desktop/issues/11074) to track it. Thanx!

services/wallet/transfer/commands_sequential.go
IvanBelyakoff: you are right. same [issue](https://github.com/status-im/status-desktop/issues/11074) as above to track progress

dlipicar (approved):
LGTM, great work!
Did you check resource usage? We can expect it to be a bit higher with the new approach, right?

IvanBelyakoff:
> LGTM, great work! Did you check resource usage? We can expect it to be a bit higher with the new approach, right?

Thanx @dlipicar. I have not checked resource usage. I believe it should be higher, but at this point lets see if we notice any performance degradation.

alaibe (approved):

alaibe:
I think as soon as u don't have accounts with 100k transactions that should still be valuable. It is unlikely someone as many account with 100k tx. only good for us to stress test

PR #3619, opened 2023-06-14
fix(chat/api): don't count muted chats' unviewed count unless mentioned

jrainville:
Small issue where we counted muted unviewed messages from muted channels, so the badge appeared if the only messages were in muted channels

### DIFF
--- a/services/chat/api.go
+++ b/services/chat/api.go
@@ -149,7 +149,9 @@ func (api *API) getChannelGroups(ctx context.Context, channelGroupID string) (ma
 			if !chat.IsActivePersonalChat() {
 				continue
 			}
-			totalUnviewedMessageCount += int(chat.UnviewedMessagesCount)
+			if !chat.Muted || chat.UnviewedMentionsCount > 0 {
+				totalUnviewedMessageCount += int(chat.UnviewedMessagesCount)
+			}
 			totalUnviewedMentionsCount += int(chat.UnviewedMentionsCount)
 
 			c, err := api.toAPIChat(chat, nil, pubKey, true)
@@ -197,8 +199,9 @@ func (api *API) getChannelGroups(ctx context.Context, channelGroupID string) (ma
 			if chat.CommunityID != community.IDString() || !chat.Active {
 				continue
 			}
-
-			totalUnviewedMessageCount += int(chat.UnviewedMessagesCount)
+			if !chat.Muted || chat.UnviewedMentionsCount > 0 {
+				totalUnviewedMessageCount += int(chat.UnviewedMessagesCount)
+			}
 			totalUnviewedMentionsCount += int(chat.UnviewedMentionsCount)
 		}
 


services/chat/api.go
```diff
@@ -149,7 +149,9 @@ func (api *API) getChannelGroups(ctx context.Context, channelGroupID string) (ma
 			if !chat.IsActivePersonalChat() {
 				continue
 			}
-			totalUnviewedMessageCount += int(chat.UnviewedMessagesCount)
+			if !chat.Muted || chat.UnviewedMentionsCount > 0 {
```
osmaczko, line 152: Append total unviewed messages when _chat is muted and has mentions_ seems a bit off to me. What is the use case? Why not simply `if !chat.Muted {`?

jrainville: We still show badges when someone is mentioned.

I'm pretty sure currently, it would still work if I only put `if !chat.Muted {`, but as I found out with the chat badges and unread state, it's safer to say we have unviewed messages when there is a mention (which is true, if we have a mention, we have unviewed messages)

osmaczko: Shouldn't the logic be as follows then?

```
    if !chat.Muted {
        totalUnviewedMessageCount += int(chat.UnviewedMessagesCount)
    } else if chat.UnviewedMentionsCount > 0 {
        totalUnviewedMessageCount += int(chat.UnviewedMentionsCount)
    }

    totalUnviewedMentionsCount += int(chat.UnviewedMentionsCount)
```

This makes it clear that if the chat is not muted, we increment the total unviewed messages count. And separately, if there are any unviewed mentions, regardless of the muted state, we increment the count by mentions count. 

In other words, I believe, we should not add muted unviewed messages which are not mentions to `totalUnviewedMessageCount`.

jrainville: Yes and no. It would do the same end result because we don't really use the count of unviewed message in the app, we just do `> 0`. So yeah, it would work.

I append the `UnviewedMessagesCount` because it's more exact in case we ever actually use that count.

The thing is, the whole logic for unread badges on the sections is a bit messed up/confusing right now.

Tthe UI shows the unread badge if `UnviewedMessagesCount > 0`, but in the case of muted channels, it's not perfect, because there might be unread messages, but if they are in muted channels, we don't want the badge.

IMO, we should refactor the section handling in our app and status-go so that status-go passes the "showBadge" value on the section, so that we can keep `UnviewedMessagesCount` as the actual value, whether or not it's in muted channels.

I rambled a bit, but hopefully you understand what I mean :sweat_smile:

osmaczko: > IMO, we should refactor the section handling in our app and status-go so that status-go passes the "showBadge" value on the section, so that we can keep UnviewedMessagesCount as the actual value, whether or not it's in muted channels.

Yes, I agree, `UnviewedMessagesCount` being the actual value is the most intuitive. Based solely on the API response, I would expect it to be exactly that.

> I append the UnviewedMessagesCount because it's more exact in case we ever actually use that count.

I understand, but for me at least, it would be extremely confusing to see a count of 0 for unviewed messages when the channel is muted, and then suddenly see the count jump to 100, for example, when just one mention comes in.

After all, I reckon it’s not that serious; I can let this surprise slide :smile:

osmaczko (approved):

PR #3623, opened 2023-06-15
feat(wallet) add filter api to retrieve recipients of a wallet

stefandunca:
### Updates status-desktop [#10025](https://github.com/status-im/status-desktop/issues/10025)

The new API returns all known wallet recipients by sourcing transfers, pending_transactions and multi_transactions tables. The API is synchronous. Future work will be to make it async. In some corner cases, when watching a famous wallet, there can be too many recipients to be returned in one go. Offset and limit can be used to paginate through the results.

### DIFF
--- a/services/wallet/activity/filter.go
+++ b/services/wallet/activity/filter.go
@@ -1,7 +1,10 @@
 package activity
 
 import (
-	eth_common "github.com/ethereum/go-ethereum/common"
+	"context"
+	"database/sql"
+
+	eth "github.com/ethereum/go-ethereum/common"
 	"github.com/status-im/status-go/services/wallet/common"
 )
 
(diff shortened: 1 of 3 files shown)


services/wallet/activity/filter.go
```diff
@@ -53,31 +56,81 @@ type TokenCode string
...
-	Tokens                Tokens               `json:"tokens"`
-	CounterpartyAddresses []eth_common.Address `json:"counterpartyAddresses"`
+	Period                Period        `json:"period"`
+	Types                 []Type        `json:"types"`
+	Statuses              []Status      `json:"statuses"`
+	Tokens                Tokens        `json:"tokens"`
+	CounterpartyAddresses []eth.Address `json:"counterpartyAddresses"`
+}
+
+// TODO: consider sorting by saved address and contacts to offload the client from doing it at runtime
```
stefandunca, line 88: Not sure about this if it makes sense yet, didn't check how we resolve names in the frontend but if rules are manageable in SQL this might be the efficient way.

Cuteivist: @stefandunca FYI in UI we check in that order:
* contact (yet to be implemented)
* wallet address
* saved address

Cuteivist (approved):
LGTM

dlipicar (approved):

PR #3626, opened 2023-06-16
fix(wallet): properly parse multitransaction amounts

dlipicar:
To properly parse the amounts from the multitransactions table, the `0x` prefix needs to be removed. This is added in this PR.

### DIFF
--- a/services/wallet/activity/activity.go
+++ b/services/wallet/activity/activity.go
@@ -476,14 +476,18 @@ func getTrInAndOutAmounts(activityType Type, trAmount sql.NullString) (inAmount 
 
 func getMtInAndOutAmounts(dbFromAmount sql.NullString, dbToAmount sql.NullString) (inAmount *hexutil.Big, outAmount *hexutil.Big) {
 	if dbFromAmount.Valid && dbToAmount.Valid {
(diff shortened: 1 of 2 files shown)


Cuteivist (approved):

services/wallet/activity/activity.go
```diff
@@ -613,6 +617,7 @@ func getActivityEntries(ctx context.Context, db *sql.DB, addresses []eth.Address
 		} else if multiTxID.Valid {
 			mtInAmount, mtOutAmount := getMtInAndOutAmounts(dbMtFromAmount, dbMtToAmount)
 			activityType := multiTransactionTypeToActivityType(transfer.MultiTransactionType(dbMtType.Byte))
+			fmt.Println("getMtInAndOutAmounts", multiTxID.Int64, activityType, mtInAmount, mtOutAmount)
```
stefandunca, line 620: I will remove these leftovers in my token PR and move the amounts helpers after the main `getActivityEntries `. The query should be close to the implementation for reference purposes. WDYT?

dlipicar: Oops! Sure, thanks!

Issue #3547, opened 2023-05-30
Sync chats list


igor-sirotin:
# Problem

Currently we don't sync chats list. Therefore, after local pairing, the chats list is empty.

Here's a screenshot of profile I have:

![image](https://github.com/status-im/status-go/assets/25482501/5bb124b3-cdf1-43b7-89b3-ce07786220e4)

This is how it appears on another device after local pairing:

![image](https://github.com/status-im/status-go/assets/25482501/7132b351-11e4-4f98-ab41-7e088133c494)

# Acceptance Criteria

Chats list is synced.
After local pairing the user can see all chats as on origin installation.

igor-sirotin:
There's also an error during initial local pairing:
```bash
ERROR[07-04|13:27:08.291] failed to HandleSyncChatMessagesRead when HandleSyncRawMessages error="can't find chat"
```

This doesn't sound like a problem, but should be considered when implementing this task. We should check the order of transferring sync data, so that `Chats` go before `ChatsMessagesRead.

PR #3627, opened 2023-06-16
fix: don't store ourselves as a contact

### LINKED ISSUE #3667: [Local pairing] Application crash when pairing 3rd device
# Problem

The app crashes here during local pairing, because there's no `contactRequest.From` in `m.allContacts`:
https://github.com/status-im/status-go/blob/9c596343bef3981e49f91aad5b0c76eb42abb6c7/protocol/messenger_contacts.go#L282

At that point `contactRequest.From` is our public key. 

# Steps to reproduce

1. `Alice` [device-1]: 
    * create account
    * send CR to `Bob`
3. `Bob`:
    * accept CR from `Alice`
5. `Alice` [device-2]:
    * local pairing from [device-1]
7. `Alice` [device-3]: 
    * local pairing from [device-2]
    * <-- crash here during pairing

igor-sirotin:
Fixes  #3667

1. Added a short-circuit for our public key in `contactsMap`
I've listed all places where it was happening in a separate issue: https://github.com/status-im/status-go/issues/3719
2. Added a test for pairing three devices of an account with mutual contact
3. Reused one `messageState.AllConacts.Load(senderID)` in `messenger.go`
4. Moved some common testing functions to `messenger_testing_utils.go`
5. Added a check if contact exists in `updateAcceptedContactRequest`

The bug mentioned is fixed by this:
https://github.com/status-im/status-go/blob/4618e1d3f160086cade6e55ae178bd02710b1e84/protocol/messenger_maps.go#L62-L72

Now the `Contact` object for ourselves is generated on the fly, so it's never "not found".

### DIFF
--- a/protocol/messenger_contact_requests_test.go
+++ b/protocol/messenger_contact_requests_test.go
@@ -62,12 +62,7 @@ func (s *MessengerContactRequestSuite) newMessenger(shh types.Waku) *Messenger {
 }
 
 func (s *MessengerContactRequestSuite) findFirstByContentType(messages []*common.Message, contentType protobuf.ChatMessage_ContentType) *common.Message {
-	for _, message := range messages {
-		if message.ContentType == contentType {
-			return message
-		}
-	}
-	return nil
+	return FindFirstByContentType(messages, contentType)
 }
 
 func (s *MessengerContactRequestSuite) sendContactRequest(request *requests.SendContactRequest, messenger *Messenger) {
--- a/protocol/messenger_maps.go
+++ b/protocol/messenger_maps.go
@@ -1,6 +1,7 @@
 package protocol
 
 import (
+	"go.uber.org/zap"
 	"sync"
 
 	"github.com/status-im/status-go/protocol/encryption/multidevice"
@@ -53,10 +54,16 @@ func (cm *chatMap) Delete(chatID string) {
 */
 
 type contactMap struct {
-	sm sync.Map
+	sm     sync.Map
+	me     *Contact
+	logger *zap.Logger
 }
 
 func (cm *contactMap) Load(contactID string) (*Contact, bool) {
+	if contactID == cm.me.ID {
+		cm.logger.Warn("contacts map: loading own identity", zap.String("contactID", contactID))
+		return cm.me, true
+	}
 	contact, ok := cm.sm.Load(contactID)
 	if contact == nil {
 		return nil, ok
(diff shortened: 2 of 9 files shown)


Samyoul:
Thank you @igor-sirotin , this is an interesting bug.

This PR certainly will stop of the crash, but I wonder if it may also prevents the expected behaviour. Does this change ensure that the receiving device has the mutual contact in its contact list after pairing?

igor-sirotin:
At that point I was trying to make some hotfix. 
But now I faced this problem once again. Taking now a closer look.

igor-sirotin:
@cammellos @MishkaRogachev 
1. I've updated the PR description and it's ready for review now 🙂 
2. The promised list of places, when we load/store ourselves as a contact is here: 
https://github.com/status-im/status-go/issues/3719
3. The mentioned `chat_messages_read` and order of syncing is mentioned here: 
https://github.com/status-im/status-go/issues/3547#issuecomment-1619989728
4. A note about removing contact requests here: 
https://github.com/status-im/status-go/issues/3720

protocol/messenger_contact_requests_test.go
```diff
@@ -62,12 +62,7 @@ func (s *MessengerContactRequestSuite) newMessenger(shh types.Waku) *Messenger {
 }
 
 func (s *MessengerContactRequestSuite) findFirstByContentType(messages []*common.Message, contentType protobuf.ChatMessage_ContentType) *common.Message {
```
jrainville, line 64: Is this function still needed? It seems like we can just call `FindFirstByContentType`

protocol/messenger_maps.go
```diff
@@ -53,10 +55,16 @@ func (cm *chatMap) Delete(chatID string) {
...
 type contactMap struct {
-	sm sync.Map
+	sm     sync.Map
+	me     *Contact
+	logger *zap.Logger
 }
 
 func (cm *contactMap) Load(contactID string) (*Contact, bool) {
+	if contactID == cm.me.ID {
+		cm.logger.Warn("contacts map: loading own identity", zap.String("contactID", contactID))
```
jrainville, line 65: I wonder if `Debug` would be better. Is getting our own identity usually wrong?

jrainville (approved):
Good job!

protocol/messenger_contact_requests_test.go
igor-sirotin: yes, I just didn't want to overload this PR
I can do that as a separate issue, if that makes sense

protocol/messenger_maps.go
igor-sirotin: hmmm. Good question actually. We can check it together with https://github.com/status-im/status-go/issues/3719.

I would say that we shouldn't ask `contactsMap` for own identity and have this condition outside `Load` function. At the same time it's such a simple solution to use a cached built `Contact` internally, that I think we could keep it forever. If we stick to this, we can replace it with `Debug` or even remove at all.

MishkaRogachev (approved):
Great!

cammellos (approved):

PR #3636, opened 2023-06-19
feat(AssetsDeployment): Assets deployment

### LINKED ISSUE #10987: Assets deployment
Implement assets deployment.

endulab:
- Add ERC20 contract
- Add decimals field to community_tokens db table
- Adjusting API to handle assets deployment
- Add decimals field to CommunityTokenMetadata

Issue #10987


jrainville (approved):
LGTM

contracts/assets/CommunityERC20.sol
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    // Events
+
+    // External functions
+
+    function setMaxSupply(uint256 newMaxSupply) external onlyOwner {
+        require(newMaxSupply >= totalSupply(), "MAX_SUPPLY_LOWER_THAN_TOTAL_SUPPLY");
+        maxSupply = newMaxSupply;
+    }
```
0x-r4bbit, line 33: Are we sure we want `maxSupply` to be changable after deployment?

contracts/assets/CommunityERC20.sol
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * @dev Mint tokens for each address in `addresses` each one with
+     * an amount specified in `amounts`.
+     *
+     */
+    function mintTo(address[] memory addresses, uint256[] memory amounts) external onlyOwner {
+        require(addresses.length == amounts.length, "WRONG_LENGTHS");
```
0x-r4bbit, line 41: This error message is not very useful.
I'd recommend using custom errors instead, also saves gas as `require` strings are rather expensive

protocol/communities/persistence_test.go
```diff
@@ -375,7 +375,7 @@ func (s *PersistenceSuite) TestGetCommunityTokens() {
 	s.Require().NoError(err)
 	s.Require().Len(tokens, 0)
 
-	token := CommunityToken{
+	token721 := CommunityToken{
```
0x-r4bbit: No biggie, but I think if we **do** make the effort to rename this variable here to make it semantically more meaningful, we should probably call it at least something like `tokenERC721` or so

protocol/migrations/sqlite/1687856939_add_community_tokens_decimals.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE community_tokens ADD COLUMN decimals INT DEFAULT 0;
```
0x-r4bbit: Should we default to 18 instead?

services/collectibles/api.go
```diff
@@ -109,12 +111,43 @@ func (api *API) Deploy(ctx context.Context, chainID uint64, deploymentParameters
...
+
 // Returns gas units + 10%
 func (api *API) DeployCollectiblesEstimate(ctx context.Context) (uint64, error) {
 	gasAmount := uint64(1960645)
 	return gasAmount + uint64(float32(gasAmount)*0.1), nil
 }
 
+// Returns gas units + 10%
+func (api *API) DeployAssetsEstimate(ctx context.Context) (uint64, error) {
+	gasAmount := uint64(957483)
```
0x-r4bbit, line 147: Where does this number come from?
Can we be sure it's constant across different chains?

contracts/assets/CommunityERC20.sol
endulab: Yes, This is our way of burning functionality. The same idea is implemented in ERC721.
We do not have any free tokens in a bank. We just decrease maxSupply, so less tokens can be distributed.
FYI, I did not implement the smart contract. Andrea did.

contracts/assets/CommunityERC20.sol
endulab: FYI, @gravityblast

contracts/assets/CommunityERC20.sol
0x-r4bbit: Ah right, the burn. Gotcha, thanks for reminding

services/collectibles/api.go
endulab: It is computed by hardhat deployment testing script.
Yes, it is the same on all chains.

protocol/communities/persistence_test.go
endulab: Done

protocol/migrations/sqlite/1687856939_add_community_tokens_decimals.up.sql
endulab: Done

services/collectibles/api.go
0x-r4bbit: Okay, I think there's a tiny percentage of a chance that L2 clients can implement different gas costs for opcodes at which point this function could estimate incorrectly.

But that's an optimization we can handle if we run into that issue.

0x-r4bbit (approved):

PR #3637, opened 2023-06-19
Chore: improve mocking for community tests

### LINKED ISSUE #3625: Improve mocking for community tests
# Problem

A technical debt has arisen in https://github.com/status-im/status-go/pull/3609 

# Implementation

- [x] Rename `Interface` in accounts.  Discussion: https://github.com/status-im/status-go/pull/3609#discussion_r1231942747
- [x] Community tests setup should used different wallet addresses for different users. Discussion:  https://github.com/status-im/status-go/pull/3609#discussion_r1230542955

MishkaRogachev:

Important changes:
- [x] Rename `Interface` to `Manager` and  `Manager` ot  `DefaultManger` in accounts.  
- [x] Different wallet addresses in community tests

Closes #3625

### DIFF
--- a/account/accounts.go
+++ b/account/accounts.go
@@ -62,15 +62,15 @@ type RecoverParams struct {
 	Signature string `json:"signature"`
 }
 
-// Interface represents account manager interface
-type Interface interface {
+// Manager represents account manager interface
+type Manager interface {
 	GetVerifiedWalletAccount(db *accounts.Database, address, password string) (*SelectedExtKey, error)
 	Sign(rpcParams SignParams, verifiedAccount *SelectedExtKey) (result types.HexBytes, err error)
-	CheckMatchingRecovered(rpcParams RecoverParams, revealedAddress types.Address) (bool, error)
+	CanRecover(rpcParams RecoverParams, revealedAddress types.Address) (bool, error)
 }
 
-// Manager represents account manager implementation
-type Manager struct {
+// DefaultManager represents default account manager implementation
+type DefaultManager struct {
 	mu         sync.RWMutex
 	rpcClient  *rpc.Client
 	rpcTimeout time.Duration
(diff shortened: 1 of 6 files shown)


protocol/communities_messenger_test.go
```diff
@@ -108,9 +110,9 @@ func (s *MessengerCommunitiesSuite) SetupTest() {
 	s.shh = gethbridge.NewGethWakuWrapper(shh)
 	s.Require().NoError(shh.Start())
 
-	s.admin = s.newMessenger(AdminPassword)
-	s.bob = s.newMessenger(BobPassword)
-	s.alice = s.newMessenger(AlicePassword)
+	s.admin = s.newMessenger(adminPassword, adminAddress)
+	s.bob = s.newMessenger(bobPassword, bobAddress)
+	s.alice = s.newMessenger(alicePassword, aliceAddress)
```
Samyoul: The password and address is used to create a `AccountManagerMock`, the testable element is the `account.Manager` not the password or the address. So the customisable element should be on the `account.Manager` level not the password and address.

This is important as the current approach limits the reusability of `s.newMessenger` as it requires a password and address when they are not always needed.

```go
func makeAccountMockManager(password, walletAddress string){
	accountsManager := &AccountManagerMock{}
	accountsManager.AccountsMap = make(map[string]string)
	accountsManager.AccountsMap[walletAddress] = types.EncodeHex(crypto.Keccak256([]byte(password)))

}
```

```go
	s.admin = s.newMessenger()
	s.admin.accountsManager = makeAccountMockManager(adminPassword, adminAddress)
	s.bob = s.newMessenger(bobPassword, bobAddress)
	s.bob.accountsManager = makeAccountMockManager(bobPassword, bobAddress)
	s.alice = s.newMessenger(alicePassword, aliceAddress)
	s.alice.accountsManager = makeAccountMockManager(alicePassword, aliceAddress)
```

Samyoul (approved):
Thank you for these changes. I really appreciate it. I've added a comment for one recommendation to consider, and great work generally.

protocol/communities_messenger_test.go
MishkaRogachev: Thanks, thats a good point! I've added additional `newMessengerWithWallet` function, because `accountsManager` needed by `communityManager`,  which is inside of a `Messenger` and it can not be settled outside

account/accounts.go
```diff
@@ -724,7 +724,15 @@ func (m *Manager) Recover(rpcParams RecoverParams) (addr types.Address, err erro
 	return
 }
 
-func (m *Manager) Sign(rpcParams SignParams, verifiedAccount *SelectedExtKey) (result types.HexBytes, err error) {
+func (m *DefaultManager) CheckMatchingRecovered(rpcParams RecoverParams, revealedAddress types.Address) (bool, error) {
```
0x-r4bbit: Maybe we can rename this to `CanRecover` ?

0x-r4bbit (approved):
This LGTM but I think we should come up with a different name for the recover method.
Is it really needed in the first place?

If we do need to have it, I think we should name it as what it's doing.
Like `CanRecover()` or `CheckAccountOwnership` or something along those lines

MishkaRogachev:
> This LGTM but I think we should come up with a different name for the recover method. Is it really needed in the first place?
> 
> If we do need to have it, I think we should name it as what it's doing. Like `CanRecover()` or `CheckAccountOwnership` or something along those lines

Yep, i changed it to `CanRecover`. The reason to have it is that we should check addresses matching in account manager, not in the community manager.

PR #3639, opened 2023-06-19
fix: don't check permissions if revealed accounts aren't legit

0x-r4bbit:
This seems to be a bug that was introduced when two features, admin permissions and "always reveal wallet accounts" where merged.

We need to make sure we **first** check the revealed accounts and only **then** do we perform permission checks on them. Otherwise we can run into scenarios where fake addresses are used and users will be accepted to the community.


osmaczko (approved):

MishkaRogachev (approved):
Good, seems it was me who placed this check in a wrong place. Btw, i'm going to change `Recover` call here: https://github.com/status-im/status-go/pull/3637/files

PR #3640, opened 2023-06-19
Fix/mutual state messages behaviour

MishkaRogachev:
Fix issues related with mutual state system messages
For https://github.com/status-im/status-desktop/pull/11178
https://github.com/status-im/status-mobile/issues/16254

Reference flow from @John-44 
```
--- What Alice (  who sends the contact request to Bob ) sees in her 1on1 chat history with Bob, at the point in time after Bob has accepted the contact request ---

1) System message:  "You sent a contact request to Bob"

2) User message from Alice: [whatever text was included in the contact request sent to Bob]

3) System message after Bob accepts Alice's contact request: "Bob accepted your contact request"

--- What Bob sees in his 1on1 chat history with Alice, after he as accepted Alice's contact request ---

1) System message: "Alice sent you a contact request"

2) User message from Alice: [whatever text was included in the contact request sent to Bob]

3) System message after Bob accepts Alice's contact request: "You accepted Alice's contact request"

```
 
Important changes:
- [x] Don't remove sent mutual state messages on accepting a CR
- [x] Don't activate chat on a `sendContactRequest` for a new contacts
- [x] Replace generated text with enum, to provide translations on a client side
- [x] Change added messages to accepted
- [x] Fix CR and mutual update messages order


protocol/messenger_contacts.go
```diff
@@ -30,7 +30,7 @@ func (m *Messenger) prepareMutualStateUpdateMessage(contactID string, updateType
 		case MutualStateUpdateTypeSent:
 			text = "You sent a contact request to @" + to
 		case MutualStateUpdateTypeAdded:
-			text = "You added  @" + to + " as a contact"
+			text = "You accepted @" + to + "'s contact request"
```
saledjenic: @MishkaRogachev since this text needs to be translated, isn't better to set it a bit differently? Unique constant? Also since it's always a case only for the 1:1 chat there is no need to add "to"/"from" anywhere in it (the same below).

MishkaRogachev: fixed, btw there was a discussion on this topic here: https://github.com/status-im/status-go/pull/3519#discussion_r1206799703

protocol/messenger_contacts.go
```diff
@@ -333,19 +306,14 @@ func (m *Messenger) updateAcceptedContactRequest(response *MessengerResponse, co
 	response.AddMessage(contactRequest)
 	response.AddContact(contact)
 
-	// Remove last sent system message
-	err = m.removeLastMutualStateUpdateMessage(response, contact.ID)
-	if err != nil {
-		return nil, err
+	// Add mutual state update message for incoming contact request
+	chat, ok := m.allChats.Load(contact.ID)
+	if !ok {
```
ilmotta: @MishkaRogachev I think it would be better to move this check right after the `contact` variable is defined in the beginning of the function. This way, we avoid unnecessary side-effects that can happen between the definition of the variable on line 255 and the chat existence check on line 311.

The idea being: load as much data as possible in the beginning of the function and validate that, and only then proceed with the side-effects. This is a practice I like for status-go in particular, because in this repository, when things fail in the middle of processes, we usually don't attempt to rollback anything, we just short-circuit and return an error, leaving the system in a potentially inconsistent state. Eventual consistency seems to be our preferred approach, but some inconsistencies could be avoided altogether.

For example, if we validate the chat exists right after loading the `contact` in memory and if the check fails, then `dispatchMessage` would never be called and no notification would be needlessly generated by `addActivityCenterNotification`.

The same idea could be used to move the line that does `GetActivityCenterNotificationByID` to the beginning of the function, then the check for the existence of `notification` would happen earlier and this would prevent some side-effects.

My comment here is leaning more on the side of a general good practice for status-go, not so much for your PR. You may disagree of course :) Feel free to ignore my suggestion.

cammellos (approved):

ilmotta (approved):

protocol/messenger_contacts.go
MishkaRogachev: @ilmotta thanks for the tip, already fixed it!

protocol/protobuf/chat_message.proto
```diff
@@ -180,6 +180,11 @@ message ChatMessage {
     IDENTITY_VERIFICATION = 13;
     // Only local
     SYSTEM_MESSAGE_PINNED_MESSAGE = 14;
-    SYSTEM_MESSAGE_MUTUAL_STATE_UPDATE = 15;
+    // Only local
+    SYSTEM_MESSAGE_MUTUAL_EVENT_SENT = 15;
```
igor-sirotin, line 184: Is it safe to replace `SYSTEM_MESSAGE_MUTUAL_STATE_UPDATE` with `SYSTEM_MESSAGE_MUTUAL_EVENT_SENT`, keeping same value `15`? There won't be conflicts with older versions?

I wonder if it should be a new value, marking the old one as deprecated.

protocol/messenger_contact_requests_test.go
```diff
@@ -1141,14 +1126,20 @@ func (s *MessengerContactRequestSuite) TestBobSendsContactRequestAfterDecliningO
 }
 
 func (s *MessengerContactRequestSuite) TestBuildContact() {
-	contactID := types.EncodeHex(crypto.FromECDSAPub(&s.m.identity.PublicKey))
-	contact, err := s.m.BuildContact(&requests.BuildContact{PublicKey: contactID})
+	bob := s.newMessenger(s.shh)
+	_, err := bob.Start()
```
igor-sirotin: Actually, maybe a full started messenger is not needed for this test?
Looks like any generated PublicKey would be enough.

protocol/messenger_contacts.go
```diff
@@ -18,21 +18,32 @@ import (
 	"github.com/status-im/status-go/protocol/transport"
 )
 
+const outgoingMutualStateEventSentDefaultText = "You sent a contact request to @%s"
```
igor-sirotin, line 21: Oh, I didn't know these text lines are in status-go.
Shouldn't they be in UI (QML side) to have ability for cool things like
- text localisation
- interactive profile name (e.g. clicking opens profile popup)

?

igor-sirotin: Ok, actually we're still able to detect the `contentType` and override the message text if we want to.

igor-sirotin (approved):
LGTM 👍 
I tried to get into details, but it's quite hard. So I can't say I checked it very careful. If needed, we can jump into a call for deeper dive 🙂

protocol/protobuf/chat_message.proto
MishkaRogachev: I suppose it's safe because it was used by local system messages only. But fresh development builds can see wrong messages in history

protocol/messenger_contacts.go
MishkaRogachev: This topic was discussed here: https://github.com/status-im/status-go/pull/3519#discussion_r1206799703. In general this text i's a fallback for clients that does not provide translations

protocol/messenger_contact_requests_test.go
MishkaRogachev: fixed

jrainville (approved):

jrainville:
Merging this. @cammellos @ilmotta @flexsurfer let us know if it creates a bad side-effect for you guys. Implementing this one should make it work https://github.com/status-im/status-mobile/issues/16254

PR #3642, opened 2023-06-20
fix: correct transfer type detection

dlipicar:
Part of https://github.com/status-im/status-desktop/issues/11080

Columns `address` and `sender` from the `transfers` table were wrongly interpreted. New columns `tx_from_address` and `tx_to_address` were added, which always give you the right addresses for all `eth`, `erc20` and `erc721` trasnfers. These are now used for filtering and detecting transaction type.

As a bonus, we use the same custom migration to change the type of column `token_address` from string to blob (like we use for all other addresses).

Moved around some code to make sure the tests use the same functions as the application to insert blocks and transfers into their tables.

This fixes a bug where `Send` activity entries were categorized as `Receive`.

### DIFF
--- a/appdatabase/database.go
+++ b/appdatabase/database.go
@@ -23,6 +23,7 @@ const nodeCfgMigrationDate = 1640111208
 var customSteps = []sqlite.PostStep{
 	{Version: 1674136690, CustomMigration: migrateEnsUsernames},
 	{Version: 1686048341, CustomMigration: migrateWalletJSONBlobs, RollBackVersion: 1686041510},
+	{Version: 1687193315, CustomMigration: migrateWalletTransferFromToAddresses, RollBackVersion: 1686825075},
 }
 
 // InitializeDB creates db file at a given path and applies migrations.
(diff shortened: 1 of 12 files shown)


appdatabase/database.go
```diff
@@ -377,3 +378,101 @@ func migrateWalletJSONBlobs(sqlTx *sql.Tx) error {
...
+			if nullableTx.Valid {
+				if nullableL.Valid {
+					_, tokenAddress, _, _, txFrom, txTo = w_common.ExtractTokenIdentity(w_common.Type(entryType), l, tx)
+				} else {
+					txFrom = &sender
+					txTo = tx.To()
+				}
+			}
+
+			currentRow = append(currentRow, tokenAddress, txFrom, txTo)
```
stefandunca, line 442: Thanks for the token fix.

appdatabase/database_test.go
```diff
@@ -238,7 +257,7 @@ func TestMigrateWalletJsonBlobs(t *testing.T) {
 			return rows.Err()
 		}
 		err := rows.Scan(&status, &receiptType, &txHash, &logIndex, &blockHash, &cumulativeGasUsed, &dbContractAddress, &gasUsed, &txIndex,
-			&txType, &protected, &gasLimit, &gasPriceClamped64, &gasTipCapClamped64, &gasFeeCapClamped64, &amount128Hex, &accountNonce, &size, &dbTokenAddress, (*bigint.SQLBigIntBytes)(tokenID), &entryType, &isTokenIDNull)
+			&txType, &protected, &gasLimit, &gasPriceClamped64, &gasTipCapClamped64, &gasFeeCapClamped64, &amount128Hex, &accountNonce, &size, &tokenAddress, (*bigint.SQLBigIntBytes)(tokenID), &entryType, &txFrom, &txTo, &isTokenIDNull)
```
stefandunca, line 254: `dbTokenAddress` can be removed now that the fix writes the binary address in place.

services/wallet/transfer/database.go
```diff
@@ -466,21 +478,117 @@ func updateOrInsertTransfers(chainID uint64, creator statementCreator, transfers
...
+	txValue            *big.Int
+	txNonce            *uint64
+	txSize             *uint64
+	tokenAddress       *common.Address
+	tokenID            *big.Int
+	txFrom             *common.Address
+	txTo               *common.Address
+}
+
+func updateOrInsertTransfersDBFields(creator statementCreator, transfers []transferDBFields) error {
```
stefandunca, line 570: Nice improvements.

stefandunca (approved):
See the notes inline.

PR #3644, opened 2023-06-20
feat: proposal for parsing shared urls

MishkaRogachev:
For https://github.com/status-im/status-desktop/issues/10851

Implement parsing shared urls in a new format

Important changes:
- [ ] Parse url with a community key or community data (/c/ or /c#)
- [ ] Parse url with a community channel (/cc/)
- [ ] Parse url with a contact

### DIFF
--- a/protocol/messenger.go
+++ b/protocol/messenger.go
@@ -88,6 +88,8 @@ var communityAdvertiseIntervalSecond int64 = 60 * 60
 // messageCacheIntervalMs is how long we should keep processed messages in the cache, in ms
 var messageCacheIntervalMs uint64 = 1000 * 60 * 60 * 48
 
+var baseStatusUrl string = "https://status.app/"
+
 // Messenger is a entity managing chats and messages.
 // It acts as a bridge between the application and encryption
 // layers.
@@ -6464,4 +6466,32 @@ func (m *Messenger) handleSyncSocialLinks(message *protobuf.SyncSocialLinks, cal
 
 func (m *Messenger) GetDeleteForMeMessages() ([]*protobuf.DeleteForMeMessage, error) {
 	return m.persistence.GetDeleteForMeMessages()
+}
+
+func (m *Messenger) ParseSharedURL(url string) (*MessengerResponse, error) {
+	if !strings.HasPrefix(url, baseStatusUrl) {
+		return nil, errors.New("Url should start with '" + baseStatusUrl + "'")
+	}
+
+	group, data, found := strings.Cut(strings.TrimPrefix(url, baseStatusUrl), "/")
+	if !found {
+		return nil, errors.New("Unexpected url format")
+	}
+
+	switch group {
+	case "c":
+		return m.ParseCommuntySharedURL(data, false)
+	case "c#":
+		return m.ParseCommuntySharedURL(data, true)
+	case "cc":
+		return m.ParseCommuntyChannelSharedURL(data)
+	case "u":
+		return m.ParseProfileSharedURL(data)
+	}
+
+	return nil, errors.New("Unhandled url group")
+}
+
+func (m *Messenger) BaseShareURL(group string) string {
+	return baseStatusUrl + group
 }
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -26,6 +26,7 @@ import (
 	"github.com/meirf/gopart"
 
 	"github.com/status-im/status-go/account"
+	"github.com/status-im/status-go/api/multiformat"
 	"github.com/status-im/status-go/eth-node/crypto"
 	"github.com/status-im/status-go/eth-node/types"
 	"github.com/status-im/status-go/images"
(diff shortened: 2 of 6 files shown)


protocol/messenger.go
```diff
@@ -6463,3 +6465,27 @@ func (m *Messenger) handleSyncSocialLinks(message *protobuf.SyncSocialLinks, cal
...
+
+	group, data, found := strings.Cut(strings.TrimPrefix(url, baseStatusUrl), "/")
+	if !found {
+		return nil, errors.New("Unexpected url format")
+	}
+
+	switch group {
+	case "c":
+		return m.ParseCommuntyShareURL(data, true)
+	case "c#":
```
igor-sirotin, line 6484: Probably you had some input for this, but I couldn't find what's `c#` in this context?

protocol/messenger.go
```diff
@@ -88,6 +88,8 @@ var communityAdvertiseIntervalSecond int64 = 60 * 60
 // messageCacheIntervalMs is how long we should keep processed messages in the cache, in ms
 var messageCacheIntervalMs uint64 = 1000 * 60 * 60 * 48
 
+var baseStatusUrl string = "https://status.app/"
```
igor-sirotin, line 91: Should we maybe also support:
- `http://`
- `status.app/...` directly without url scheme
- upper/mixed case `https://STATUS.App/...`

?

protocol/messenger.go
```diff
@@ -6463,3 +6465,27 @@ func (m *Messenger) handleSyncSocialLinks(message *protobuf.SyncSocialLinks, cal
 func (m *Messenger) GetDeleteForMeMessages() ([]*protobuf.DeleteForMeMessage, error) {
 	return m.persistence.GetDeleteForMeMessages()
 }
+
+func (m *Messenger) ParseSharedURL(url string) (*MessengerResponse, error) {
+	if !strings.HasPrefix(url, baseStatusUrl) {
```
igor-sirotin, line 6472: Maybe convert `url` to lower case here? 
Or do we consider only everything in lower case as valid?

protocol/messenger_communities.go
```diff
@@ -1704,6 +1704,34 @@ func (m *Messenger) ShareCommunity(request *requests.ShareCommunity) (*Messenger
 	return response, nil
 }
 
+func (m *Messenger) ParseCommuntyShareURL(url string, byID bool) (*MessengerResponse, error) {
+	var community *communities.Community
+	if byID {
+		// TODO: decompress community key
+		communityId := types.HexBytes(url)
+		community, err := m.GetCommunityByID(communityId)
```
igor-sirotin: hmmm... This will only work for "known" communities.
But I suppose in most cases user will user shared url for unknown community. Probably we should start fetching community details here, os smth like that.

Same for profile below. There it will use `BuildContact` and return smth for unknown contact, but probably still fetching name/image is a good idea.

protocol/messenger_contacts.go
```diff
@@ -1202,6 +1202,27 @@ func (m *Messenger) BuildContact(request *requests.BuildContact) (*Contact, erro
...
 }
 
+func (m *Messenger) ParseProfileShareURL(url string) (*MessengerResponse, error) {
+	// TODO: decompress public key
+	pubKey, err := common.HexToPubkey(url)
+	if err != nil {
+		return nil, err
+	}
+
+	contact, err := m.BuildContact(&requests.BuildContact{PublicKey: types.EncodeHex(crypto.FromECDSAPub(pubKey))})
```
igor-sirotin, line 1212: Should we support ENS names in shared urls, like `status.app/u/vitalik.eth`? 🤔

igor-sirotin:
Basically the solution is great, just a bunch of minor questions 🙂

protocol/messenger_communities.go
MishkaRogachev: There are 2 groups: 'c#' which only provides community ID and `c` with community data in protobuf. Here i'm waiting for conversion functions from Boris

protocol/messenger.go
MishkaRogachev: Answered below :)

protocol/messenger.go
MishkaRogachev: Good questions! There are discussion on urls here: https://github.com/status-im/status-web/issues/327, lets raise those questions there

igor-sirotin: I think this one is the most up-to-date doc?
https://github.com/vacp2p/rfc/blob/43c07f5290b644d668d3d4e6701a14075ef2d67b/content/docs/rfcs/60/README.md

(from https://github.com/vacp2p/rfc/pull/600)

MishkaRogachev:
@igor-sirotin i'm closing this PR in favour of https://github.com/status-im/status-go/pull/3665
For your questions there are discussion and specs here:
https://github.com/vacp2p/rfc/pull/600
 https://github.com/status-im/status-web/issues/327

PR #3645, opened 2023-06-20
feat: add account position

alaibe:


appdatabase/migrations/sql/1687249080_add_position_accounts.up..sql
```diff
@@ -0,0 +1,43 @@
+CREATE TABLE IF NOT EXISTS keypairs_accounts_new (
```
stefandunca: Isn't `ALTER TABLE keypairs_accounts ADD COLUMN position INT;` working?

multiaccounts/accounts/database.go
```diff
@@ -974,3 +998,29 @@ func (db *Database) UpdateAccountToFullyOperable(keyUID string, address types.Ad
 	_, err = tx.Exec(`UPDATE keypairs_accounts SET operable = ?	WHERE address = ?`, AccountFullyOperable, address)
 	return err
 }
+
```
stefandunca, line 1002: How about a comment on this function's guarantees so that the user won't assume that the position is the corresponding index?

multiaccounts/accounts/database.go
```diff
@@ -974,3 +998,29 @@ func (db *Database) UpdateAccountToFullyOperable(keyUID string, address types.Ad
...
+			return
+		}
+		_ = tx.Rollback()
+	}()
+
+	if err != nil {
+		return err
+	}
+
+	_, err = tx.Exec("UPDATE keypairs_accounts SET position = position + 1 WHERE position >= ?", newPosition)
```
stefandunca: If we have accounts [1, 2, 3, 4, 5], moving 2 to 4 will result in [1, 3, 4, 5, 6] not keeping the strict position index. Is this intended?

stefandunca (approved):

appdatabase/migrations/sql/1687249080_add_position_accounts.up..sql
alaibe: It will work but it won't update previous position record, see the 
`ROW_NUMBER () OVER (ORDER BY created_at) as position`

multiaccounts/accounts/database.go
alaibe: Yeah, I didn't think it was important to keep index as 1,2,3,4......
As we won't have billions of accounts anyway

appdatabase/migrations/sql/1687249080_add_position_accounts.up..sql
stefandunca: Not sure I follow. As I see it, the create table doesn't do anything related to the update. The following insert into does the trick. Using ALTER table can also be simplified to update the position only instead of all the data. Do I miss something?

multiaccounts/accounts/database.go
stefandunca: That is why I added the previous comment to create this awareness. Otherwise, the `positon` column name might lead to an assumption that this is a real position and not just an order hint and can be used in other operations.

appdatabase/migrations/sql/1687249080_add_position_accounts.up..sql
alaibe: Let me try to do the insert only

multiaccounts/accounts/database.go
alaibe: Yes fine adding this comment, also not that if u delete an account i do not update the position too, cause it doesn't have to be

appdatabase/migrations/sql/1687249080_add_position_accounts.up..sql
IvanBelyakoff: I don't think I understand why creating a new table. If we add a position column to the existing table, we could do something like the following to assign sequential indices to it based on `created_at`:
```
UPDATE keypairs_accounts AS t1
SET position = t2.rn
FROM (
SELECT created_at, ROW_NUMBER() OVER (ORDER BY created_at) AS rn
  FROM keypairs_accounts) AS t2
WHERE t2.created_at = t1.created_at
```

alaibe: Yeah that's what Stefan is saying and what I am doing

multiaccounts/accounts/database_test.go
```diff
@@ -42,6 +42,27 @@ func TestGetAddresses(t *testing.T) {
...
+	require.Equal(t, accounts[0].Position, int64(1))
+	require.Equal(t, accounts[1].Position, int64(2))
+
+	err = db.UpdateAccountPosition(accounts[1].Address, 1)
+	require.NoError(t, err)
+
+	accounts, err = db.GetAccounts()
+	require.Equal(t, accounts[0].Position, int64(0))
+	require.Equal(t, accounts[1].Position, int64(1))
+}
```
stefandunca: The following extended tests fails a normal presentation layer usage

```suggestion
func TestUpdateAccountPosition(t *testing.T) {
	db, stop := setupTestDB(t)
	defer stop()
	accounts := []*Account{
		{Address: types.Address{0x01}, Chat: true, Wallet: true},
		{Address: types.Address{0x02}},
		{Address: types.Address{0x03}},
		{Address: types.Address{0x04}},
		{Address: types.Address{0x05}},
	}
	require.NoError(t, db.SaveOrUpdateAccounts(accounts))
	res, err := db.GetAccounts()
	require.NoError(t, err)
	validateOrderForRes := func(accOrd []int) {
		require.Equal(t, len(accounts), len(accOrd))
		for i, ord := range accOrd {
			var resJson []byte
			if accounts[ord].Address != res[i].Address {
				resJson, _ = json.Marshal(res)
			}
			require.Equal(t, accounts[ord].Address, res[i].Address, "account order broken at element %d; result content", i, string(resJson))
		}
	}
	validateOrderForRes([]int{0, 1, 2, 3, 4})

	err = db.UpdateAccountPosition(accounts[0].Address, 4)
	require.NoError(t, err)

	res, err = db.GetAccounts()
	require.NoError(t, err)
	validateOrderForRes([]int{1, 2, 3, 0, 4})

	err = db.UpdateAccountPosition(accounts[2].Address, 4)
	require.NoError(t, err)

	res, err = db.GetAccounts()
	require.NoError(t, err)
	validateOrderForRes([]int{1, 3, 0, 2, 4})
}
```

stefandunca (changes requested):
I thought the tests were not covering enough test cases, so I extended them, and they seemed to fail on the second use case, where we moved the second element into the fourth position (one-based), and it landed in the results on the third position (one-based).

If the intention of the position argument in the API is 0-based, then the first fails.

stefandunca (approved):

multiaccounts/accounts/database_test.go
stefandunca: @alaibe, sorry for the misunderstanding; following the PM discussion, I now understand that position argument in the API is not index ordering but relative ordering. So QML will have to maintain the order. It will be nice to extend tests to show this.

stefandunca: I also fixed the proposed tests to follow the way API should be used and they pass all right

```go
func TestUpdateAccountPosition(t *testing.T) {
	db, stop := setupTestDB(t)
	defer stop()
	accounts := []*Account{
		{Address: types.Address{0x01}, Chat: true, Wallet: true},
		{Address: types.Address{0x02}},
		{Address: types.Address{0x03}},
		{Address: types.Address{0x04}},
		{Address: types.Address{0x05}},
	}
	require.NoError(t, db.SaveOrUpdateAccounts(accounts))
	res, err := db.GetAccounts()
	require.NoError(t, err)
	validateOrderForRes := func(accOrd []int) {
		require.Equal(t, len(accounts), len(accOrd))

		sort.Slice(res, func(l, r int) bool {
			return res[l].Position < res[r].Position
		})

		for i, ord := range accOrd {
			var resJson []byte
			if accounts[ord].Address != res[i].Address {
				resJson, _ = json.Marshal(res)
			}
			require.Equal(t, accounts[ord].Address, res[i].Address, "account order broken at element %d; result content", i, string(resJson))
		}
	}
	validateOrderForRes([]int{0, 1, 2, 3, 4})

	err = db.UpdateAccountPosition(accounts[0].Address, res[3].Position)
	require.NoError(t, err)

	res, err = db.GetAccounts()
	require.NoError(t, err)
	validateOrderForRes([]int{1, 2, 3, 0, 4})

	err = db.UpdateAccountPosition(accounts[2].Address, res[3].Position)
	require.NoError(t, err)

	res, err = db.GetAccounts()
	require.NoError(t, err)
	validateOrderForRes([]int{1, 3, 0, 2, 4})
}
```

PR #3648, opened 2023-06-20
Make sure that messages are confirmed when sent

cammellos:
In some instances, we receive a confirmation before the RawMessage and Message are saved in the database. This commit addresses that issue.

This commit has two changes:

1) Makes handling raw messages asynchronous, so that the if the confirmation arrives first, is preserved in the database and not overwritten when adding the raw message.
2) It adds a `BeforeDispatch` function on raw messages to save the message before it's actually dispatched, in order to avoid a race condition where you would get the confirmation before the message has been dispatched, and it would fail to update the message. This is not the cleanest solution, it would be better to have a separate table for message outgoing status and join at runtime, but that's a much larger change and it's an expensive migration for clients, so I would defer doing that unless necessary.

### DIFF
--- a/protocol/common/message_sender.go
+++ b/protocol/common/message_sender.go
@@ -261,6 +261,11 @@ func (s *MessageSender) sendCommunity(
 	rawMessage.ID = types.EncodeHex(messageID)
 	messageIDs := [][]byte{messageID}
 
+	if rawMessage.BeforeDispatch != nil {
+		if err := rawMessage.BeforeDispatch(rawMessage); err != nil {
+			return nil, err
+		}
+	}
 	// Notify before dispatching, otherwise the dispatch subscription might happen
 	// earlier than the scheduled
 	s.notifyOnScheduledMessage(nil, rawMessage)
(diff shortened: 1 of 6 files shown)


protocol/common/message_sender.go
```diff
@@ -351,6 +356,12 @@ func (s *MessageSender) sendPrivate(
 	messageID := v1protocol.MessageID(&rawMessage.Sender.PublicKey, wrappedMessage)
 	rawMessage.ID = types.EncodeHex(messageID)
 
+	if rawMessage.BeforeDispatch != nil {
```
MishkaRogachev, line 359: Just wondering if there is something like js optional changing in go?

MishkaRogachev (approved):
Looks great!

protocol/common/message_sender.go
cammellos: I don't think there's anything like that I am afraid

protocol/common/raw_message.go
```diff
@@ -34,4 +34,5 @@ type RawMessage struct {
 	CommunityID           []byte
 	CommunityKeyExMsgType CommKeyExMsgType
 	Ephemeral             bool
+	BeforeDispatch        func(*RawMessage) error
```
Samyoul, line 37: I like this.

Samyoul (approved):

saledjenic (commented):
@cammellos are changes done in the following files really needed?
- protocol/migrations/sqlite/000002_blah.sql
- protocol/migrations/sqlite/1685964184_test.sql
- test.txt

cammellos:
good point :) i was testing something, need to remove them, thanks

saledjenic (approved):

PR #3649, opened 2023-06-20
Set installation name on account creation and pairing

rasom:
status: ready

needed for https://github.com/status-im/status-mobile/issues/15944


protocol/requests/create_account.go
```diff
@@ -15,6 +15,7 @@ type CreateAccount struct {
 	// BackupDisabledDataDir is the directory where backup is disabled
 	BackupDisabledDataDir string `json:"backupDisabledDataDir"`
 
+	DeviceName         string `json"deviceName"`
```
ibrkhalil: Typo, Should be `json:"deviceName"`

ibrkhalil:
Should the title be `Set device name on account creation and pairing`?

protocol/requests/create_account.go
rasom: You are right, though what is weird is that it works anyway.

rasom: https://cs.opensource.google/go/go/+/refs/tags/go1.20.5:src/reflect/type.go;l=1221 ok, the colon is not really needed by design, no miracles, the value will be detected by opening double quote https://cs.opensource.google/go/go/+/refs/tags/go1.20.5:src/reflect/type.go;l=1221

rasom:
@ibrkhalil it is kind of the same thing here, and both names are already used in the code. It is a device name, but used fro installation as just a `name` property... Well, doesn't really change anything

protocol/requests/create_account.go
ibrkhalil: I think it's better to be consistent.

rasom: well yep it is already fixed some time ago, I just wanted to verify whether what I tested could actually work

borismelnik (approved):
LGTM :) Is it required some changes on desktop app?

ibrkhalil (approved):

rasom:
@borismelnik , no it shouldn't

alaibe (approved):

Issue #3622, opened 2023-06-15
Introduce generate share urls api methods


borismelnik:
# Problem

in fact of [desktop issue](https://github.com/status-im/status-desktop/issues/10851)

# Implementation

there will be a 3 new methods to generate correct urls:
`GenerateCommuntyShareURL` which return generated url for community or encrypted community as well (`/c/` or `/c#`)
`GenerateCommunityChannelShareURL` which return url for channel (`/cc/`)
`GenerateProfileShareURL` - which return url for open 1-to-1 chat

new one method to parse the url:
`ParseSharedURL` which return embedded data from url in JSON

also, 4 tests for every each method.

igor-sirotin:
Already implemented:
https://github.com/status-im/status-go/blob/c02ebf957cce04bd9489cd69e45ff7ab0b61a7ba/services/ext/api.go#L1665-L1695

PR #3650, opened 2023-06-20
feat(share-links): Add protobuf and encode/decode url data methods

borismelnik:
- Adds new module for `brotli` compression 
- Adds `encode/decode` methods for url data

Important changes:

part of  #3622


jrainville (approved):
Looks good. Maybe later when we have the full API we'd need a test that shows that the encode and decode work as expected

borismelnik:
> Looks good. Maybe later when we have the full API we'd need a test that shows that the encode and decode work as expected

it will be in the next PR :)

MishkaRogachev (approved):

protocol/urls/urls.go
```diff
@@ -293,3 +293,11 @@ func GetLinkPreviewData(link string) (previewData LinkPreviewData, err error) {
 		return previewData, fmt.Errorf("link %s isn't whitelisted. Hostname - %s", link, u.Hostname())
 	}
 }
+
+func EncodeDataURL(data []byte) (string, error) {
+	return encodeDataURL(data)
+}
+
+func DecodeDataURL(data string) ([]byte, error) {
```
ilmotta, line 301: Unit tests for the public functions `EncodeDataURL` and `DecodeDataURL` seem appropriate.

protocol/urls/urls.go
```diff
@@ -293,3 +293,11 @@ func GetLinkPreviewData(link string) (previewData LinkPreviewData, err error) {
 		return previewData, fmt.Errorf("link %s isn't whitelisted. Hostname - %s", link, u.Hostname())
 	}
 }
+
```
ilmotta, line 296: Worth mentioning that I created a different package `linkpreview` in `protocol/` for dealing specifically with link previews and  the idea was to delete the `urls` package entirely since it doesn't fulfill clients' needs anymore.

I think the PR can be merged using the `urls` package, but I think we should refactor soon(ish) to use the `Unfurler` interface defined in `protocol/linkpreview` so that we don't grow the package that was supposed to be deleted (when the new link preview implementation is more complete featurewise).

**OR forget what I just said above**, and we can use/think of a completely different package and interface to deal with "domain" data (community, channel, user) because these can be seen as significantly different from unfurleable data from websites. So maybe using the package `urls` is okay. Food for thought :)

protocol/protobuf/url_data.proto
```diff
@@ -0,0 +1,39 @@
...
+ // Community, Channel, or User
+ bytes content = 1;
+}
+
+// Field on CommunityDescription, CommunityChat and ContactCodeAdvertisement
+message URLParams {
+ string encoded_url_data = 1;
+ // Signature of encoded URL data
+ string encoded_signature = 2;
+}
```
ilmotta, line 39: It's generally accepted that we should end files with *proper* lines. If a reference helps, according to the POSIX standard, proper lines should have a terminating newline character. If you allow me, I recommend configuring your text editor of choice to do this for you. All of them have this feature.

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206

protocol/urls/urls.go
Samyoul: Not wanting to dogpile you @borismelnik, I noticed your comment https://github.com/status-im/status-go/pull/3650#issuecomment-1599276139 , but we should never introduce code into this repo without proving that it does what it is expected to do.

protocol/protobuf/url_data.proto
```diff
@@ -0,0 +1,39 @@
+syntax = "proto3";
```
Samyoul, line 1: I notice that these protobufs are introduced but are not used anywhere in this PR. Am I correct in thinking that they perhaps belong in another PR?

borismelnik: @Samyoul Yes, this PR is a general PR for next 2 PRs

protocol/urls/urls.go
borismelnik: okay! I will add the tests in this PR

Samyoul: Thank you very much, I really appreciate that 🙏

borismelnik:
in fact of https://github.com/status-im/status-go/pull/3665

PR #3652, opened 2023-06-21
fix: identify ETH asset properly in swap transactions

dlipicar:
Swap operations are done between ERC20 tokens. When the user wants to use ETH as From/To asset, WETH (ETH wrapped in an erc20 token) is used. In such cases, a `Deposit`/`Withdraw` event (from the WETH contract) will be emitted in the same transaction.

This PR adds detection of these two events, to properly identify asset ETH instead of WETH in swap transactions.

image

### DIFF
--- a/services/wallet/common/log_parser.go
+++ b/services/wallet/common/log_parser.go
@@ -28,18 +28,29 @@ const (
 	unknownTransaction Type = "unknown"
 
 	// Event types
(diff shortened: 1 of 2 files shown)


ghost:
### Pull Request Checklist

stefandunca (approved):

PR #3653, opened 2023-06-21
fix(migration): Fixes mute_till migration

borismelnik:
Fix the problem with `mute_till` migration file:

The PR with `muting community with intervals` was merged with new migration file with an old timestamp which introduced the bug, when user can't login because of `mute_till` column not exists. But no error occurs when creating a new user account.
To cover all cases, this PR introduce the migration file which should recreate `communities_communities` table and copy all data.

Important changes:

Closes #


jrainville (commented):
Would `ALTER TABLE communities ADD COLUMN muted_till TIMESTAMP;` be enough

Samyoul:
@borismelnik would you be able to detail a little more in the PR description what the problem is and how this PR fixes that problem?

borismelnik:
> Would `ALTER TABLE communities ADD COLUMN muted_till TIMESTAMP;` be enough

After discussion with @cammellos I realized its not enough :(

osmaczko (commented):
I believe it would be enough to just change the migration timestamp and run `make generate`.

```diff
diff --git a/protocol/migrations/sqlite/1685383829_add_communities_mute_till.up.sql b/protocol/migrations/sqlite/1685964184_add_communities_mute_till.up.sql
similarity index 100%
rename from protocol/migrations/sqlite/1685383829_add_communities_mute_till.up.sql
rename to protocol/migrations/sqlite/1685964184_add_communities_mute_till.up.sql
```

osmaczko:
> > Would `ALTER TABLE communities ADD COLUMN muted_till TIMESTAMP;` be enough
> 
> After discussion with @cammellos I realized its not enough :(

Can you please elaborate on that?

cammellos:
> > > Would `ALTER TABLE communities ADD COLUMN muted_till TIMESTAMP;` be enough
> > 
> > 
> > After discussion with @cammellos I realized its not enough :(
> 
> Can you please elaborate on that?

It works for those accounts who didn't run the migration, but if your account did run it (a new account), that would result in an error, as it would try to run the same migration twice.

cammellos:
also, if you were right on that migration (which is very unlikely, since it was out of order), you would get "file not found", as the migration system looks for the latest migration, though in this case I don't think it would happen.

The safest way is always to add a migration in general, renaming the timestamp is less bullet proof, though it might work in some instances (in this case I don't think it would, since we can't make `ALTER TABLE` not fail if the db had already been migrated, if it's was a create table, you can use IF EXISTS)

borismelnik:
@Samyoul @cammellos I can't merge it without approves :D

jrainville (approved):

osmaczko:
> I believe it would be enough to just change the migration timestamp and run `make generate`.
> 
> ```diff
> diff --git a/protocol/migrations/sqlite/1685383829_add_communities_mute_till.up.sql b/protocol/migrations/sqlite/1685964184_add_communities_mute_till.up.sql
> similarity index 100%
> rename from protocol/migrations/sqlite/1685383829_add_communities_mute_till.up.sql
> rename to protocol/migrations/sqlite/1685964184_add_communities_mute_till.up.sql
> ```

In case anyone still wondering why it wouldn't work well.

image

protocol/migrations/migrations.go
osmaczko, line 1: Only this file should be re-generated, could you please remove others from the commit?

borismelnik: You mean after rebase? but I got a conflicts there

borismelnik:
> > I believe it would be enough to just change the migration timestamp and run `make generate`.
> > ```diff
> > diff --git a/protocol/migrations/sqlite/1685383829_add_communities_mute_till.up.sql b/protocol/migrations/sqlite/1685964184_add_communities_mute_till.up.sql
> > similarity index 100%
> > rename from protocol/migrations/sqlite/1685383829_add_communities_mute_till.up.sql
> > rename to protocol/migrations/sqlite/1685964184_add_communities_mute_till.up.sql
> > ```
> 
> In case anyone still wondering why it wouldn't work well.
> 
> image

I actually agree, we still not released

protocol/migrations/migrations.go
osmaczko: I mean the other `bindata.go` and `migrations.go` files update are not needed.

cammellos (approved):

jrainville (approved):

PR #3654, opened 2023-06-21
Add migration check scripts

cammellos:
This commit adds a script that checks that no out of order migrations are introduced in the codebase when merging/rebasing.

This check is done on Jenkins, I have also added a way to install git hooks so that if someone merges locally it will trigger the same.

You can install the scripts locally by running
`make install-git-hooks`

### DIFF
--- a/.github/github-bot.yml
+++ b/.github/github-bot.yml
@@ -3,13 +3,6 @@ _extends: probot-settings
 github-team:
   slug: 'go'
 
-prchecklist:
-  title: Pull Request Checklist
-  checklist:
-    - 'Have you updated the documentation, if impacted (e.g. [docs.status.im](https://status.im/docs/))?'
-    - 'Have you tested changes with mobile?'
-    - 'Have you tested changes with desktop?'
-
 stale:
   daysUntilStale: 90
   daysUntilPullRequestStale: 14
--- a/_assets/scripts/migration_check.sh
+++ b/_assets/scripts/migration_check.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+set -e
+set -o pipefail
+
+check_migration_order() {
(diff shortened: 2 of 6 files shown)


_assets/scripts/migration_check.sh
```diff
@@ -0,0 +1,28 @@
...
+      echo "Error: Migration files are out of order. Please ensure migrations are added in chronological order."
+      exit 1
+    fi
+
+    prev_migration="$current_migration"
+  done
+}
+
+committed_files=$(git ls-tree -r --name-only HEAD protocol/migrations/sqlite/*.sql | sort)
+staged_files=$(git diff --name-only develop protocol/migrations/sqlite/*.sql | sort)
```
jakubgs: I see this is failing with:
```
fatal: ambiguous argument 'develop': unknown revision or path not in the working tree.
```
Maybe we could use `origin/develop` here:
```suggestion
staged_files=$(git diff --name-only origin/develop protocol/migrations/sqlite/*.sql | sort)
```

jakubgs (approved):
Nice work.

.github/github-bot.yml
```diff
@@ -3,13 +3,6 @@ _extends: probot-settings
 github-team:
   slug: 'go'
 
-prchecklist:
```
Samyoul, line 6: A friendly member from the Holy Orders of the Emperor's Inquisition will be in touch.

Samyoul (approved):
Apart from the potential heresy, this is a vision of beauty.

jrainville (approved):
Nice

borismelnik (approved):
👍

osmaczko (approved):
Cool!

caybro (approved):
Very nice :muscle:

PR #3656, opened 2023-06-21
feat(communities): enable selecting addresses to pass when joining

jrainville:
Needed for https://github.com/status-im/status-desktop/issues/11154

Improves `RequestToJoinCommunity`  to accept `Addresses` in the request. If `Addresses` is not empty, we then only pass to the owner the selected addresses. The others are ignored. Does not validate that the addresses in the slice are part of the user's wallet. Those not part of the wallet are just ignored.

### DIFF
--- a/protocol/communities_messenger_test.go
+++ b/protocol/communities_messenger_test.go
@@ -48,7 +48,8 @@ const alicePassword = "qwerty"
 const bobPassword = "bob123"
 
 const adminAddress = "0x0100000000000000000000000000000000000000"
-const aliceAddress = "0x0200000000000000000000000000000000000000"
+const aliceAddress1 = "0x0200000000000000000000000000000000000000"
+const aliceAddress2 = "0x0210000000000000000000000000000000000000"
 const bobAddress = "0x0300000000000000000000000000000000000000"
 
 type AccountManagerMock struct {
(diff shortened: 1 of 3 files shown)


protocol/communities_messenger_test.go
jrainville, line 1: I improved the utils functions in this test to accept passing multiple addresses. That enabled me to create the new test that verifies that when passing an address in the request, we indeed only receive that address in the owner's messenger.

protocol/requests/request_to_join_community.go
```diff
@@ -7,17 +7,22 @@ import (
 )
 
 var ErrRequestToJoinCommunityInvalidCommunityID = errors.New("request-to-join-community: invalid community id")
+var ErrRequestToJoinCommunityMissingPassword = errors.New("request-to-join-community: password is necessary when sending a list of addresses")
 
 type RequestToJoinCommunity struct {
 	CommunityID types.HexBytes `json:"communityId"`
 	ENSName     string         `json:"ensName"`
 	Password    string         `json:"password"`
+	Addresses   []string       `json:"addresses"`
```
0x-r4bbit: How do you think about calling this `AddressesToReveal` ?

protocol/messenger_communities.go
```diff
@@ -706,6 +706,21 @@ func (m *Messenger) RequestToJoinCommunity(request *requests.RequestToJoinCommun
...
+							found = true
+							break
+						}
+					}
+					if !found {
+						// The user didn't select this address
+						continue
+					}
+				}
+
```
0x-r4bbit, line 722: Just another way to do it, but would be cool if we could change the check to something like

```golang
if wantsToRevealWalletAddress := addressToRevealMap[walletAccount.Address.Hex()]; hasAddressesToReveal && wantsToRevealWalletAddress {
 ...
}
```

0x-r4bbit (approved):
This looks fine to me.
I've added some comments but they can be ignored

protocol/messenger_communities.go
mprakhov: or for more readiness:

```
if len(request.Addresses) > 0 && !containsAddress(request.Addresses, walletAccount.Address.Hex()) {
	continue
}

func containsAddress(addresses []string, targetAddress string) bool {
	for _, address := range addresses {
		if address == targetAddress {
			return true
		}
	}
	return false
}
```

mprakhov (approved):

PR #3657, opened 2023-06-22
feat: store check channel permission responses

0x-r4bbit:
This commit adds new tables to the database and APIs in `Messenger` and communities `Manager` to store `CheckChannelPermissionsResponse`s.

The responses are stored whenever channel permissions have been checked.

The reason we're doing this is so that clients can retrieve the last known channel permission state before waiting for onchain checks to finish.


jrainville (approved):
Looks good

osmaczko (approved):

jrainville (approved):

PR #3663, opened 2023-06-22
Run tests separately

cammellos:
This commit adds:

- run tests separately to avoid the 20m timeout
- Call shutdown after creating a messenger, with defer, not checking for errors. This is not ideal, but it's much easier to spot issues and messengers not closed, as there were many instances where that was the case.
- Fix torrent test, since it could not run concurrently

I tried to figure out the root cause of the slow tests. From what I could gather, it looks like a problem of resources.

The heaviest tests are the encryption tests, they take up to 6/7 minutes. On my machine though, it takes 20s to run a suite, while on CI it takes up to 4/5 minutes for the same suite. Though that's not consistent, some runs are very fast on ci (4s), while others are really slow. 
This seems to indicate that the system is under stress. One of the reasons was that messengers were not shutdown, so you could see a bunch of messengers running while the encryption tests were running (which don't use messenger at all). 
I have changed the code so that we call `defer m.Shutdown()` after creating a messenger, is less error prone, this skip error checking as it runs outside the test suite. Probably it would be best to have something a bit more elegant, but for now I think that's acceptable.
Even with this change, tests run slow, so there's something else going on. 
Each file now is run in a different `go test`, so the 20m timeout applies for test file, rather for the whole test suite.

### DIFF
--- a/protocol/messenger_delete_message_test.go
+++ b/protocol/messenger_delete_message_test.go
@@ -23,6 +23,7 @@ func (s *MessengerDeleteMessageSuite) TestDeleteMessage() {
 	theirMessenger := s.newMessenger()
 	_, err := theirMessenger.Start()
 	s.Require().NoError(err)
+	defer theirMessenger.Shutdown() // nolint: errcheck
 
 	theirChat := CreateOneToOneChat("Their 1TO1", &s.privateKey.PublicKey, s.m.transport)
 	err = theirMessenger.SaveChat(theirChat)
@@ -73,6 +74,7 @@ func (s *MessengerDeleteMessageSuite) TestDeleteMessagePreviousLastMessage() {
 	theirMessenger := s.newMessenger()
 	_, err := theirMessenger.Start()
 	s.Require().NoError(err)
+	defer theirMessenger.Shutdown() // nolint: errcheck
 
 	theirChat := CreateOneToOneChat("Their 1TO1", &s.privateKey.PublicKey, s.m.transport)
 	err = theirMessenger.SaveChat(theirChat)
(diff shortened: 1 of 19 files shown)


ghost:
### Pull Request Checklist

protocol/messenger_delete_message_test.go
```diff
@@ -23,6 +23,7 @@ func (s *MessengerDeleteMessageSuite) TestDeleteMessage() {
 	theirMessenger := s.newMessenger()
 	_, err := theirMessenger.Start()
 	s.Require().NoError(err)
+	defer theirMessenger.Shutdown() // nolint: errcheck
```
Samyoul, line 26: I wonder if it is possible to customise the linter to ignore errors from a deferred function

cammellos: I think that might not be what we want, in tests or like in this case, it makes sense, but people might do it in places where they shouldn't I guess

jakubgs (approved):

Samyoul (approved):

protocol/messenger_delete_message_test.go
Samyoul: Having a quick look through this commit https://github.com/amnonbc/errcheck/commit/3df4ffcb5f4958171ceb8eb833574522560ed9ae#diff-4adfb6e9793bcec46facbdb73ff436ac5f2b9a060577c73167801a8812a8fb69 what I'm suggesting appears to be possible but would require changes to the `errcheck` source code.

`errcheck` can be configured to ignore all test files entirely but doesn't have the logic to only conditionally ignore defer functions that are in test files. Seems possible by updating the `visitor` state and checking if the `visitor` is in a test file and is inspecting a defer statement. But this functionality is not configurable from the config settings alone.

PR #3666, opened 2023-06-23
fix: admins are not allowed to kick or ban other admins

0x-r4bbit:


mprakhov (approved):

MishkaRogachev (approved):

protocol/communities_messenger_admin_test.go
```diff
@@ -623,6 +623,17 @@ func (s *AdminMessengerCommunitiesSuite) TestAdminReorderChannelsAndCategories()
...
 
+func (s *AdminMessengerCommunitiesSuite) TestAdminKickAdmin() {
+	community := s.setUpCommunityAndRoles()
+
+	// admin tries to kick the owner
+	_, err := s.admin.RemoveUserFromCommunity(
+		community.ID(),
+		common.PubkeyToHex(&s.admin.identity.PublicKey),
+	)
+	s.Require().Error(err)
```
cammellos, line 636: You probably want to match the error, it's a bit fragile otherwise, same in the test below

PR #3669, opened 2023-06-23
tests: make test-unit target more dynamic

jakubgs:
We are generating a list of unit tests we can run with `go list`, and then a dynamic target that can take a `test-unit_${UNIT_PATH}` as target.

This way we can run all tests using `make test-unit`, or we can run individual tests using the prefixed form:
```
 > make test-unit_github.com/status-im/status-go/appdatabase
TESTING: test-unit_github.com/status-im/status-go/appdatabase
=== RUN   Test_GetDBFilename
--- PASS: Test_GetDBFilename (0.29s)
=== RUN   TestMigrateWalletJsonBlobs
--- PASS: TestMigrateWalletJsonBlobs (0.09s)
=== RUN   TestGetNodeConfig
--- PASS: TestGetNodeConfig (0.20s)
=== RUN   TestSaveNodeConfig
--- PASS: TestSaveNodeConfig (0.20s)
=== RUN   TestMigrateNodeConfig
--- PASS: TestMigrateNodeConfig (0.20s)
PASS
ok      github.com/status-im/status-go/appdatabase    (cached)

real    0m0.281s
user    0m0.872s
sys    0m0.438s
```
It's also worth noting that if `make test-unit -j1` is used then the tests will run serially instead of in parallel.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -306,15 +306,20 @@ docker-test: ##@tests Run tests in a docker container with golang.
 
 test: test-unit ##@tests Run basic, short tests during development
 
-test-unit: UNIT_TEST_PACKAGES = $(shell go list ./...  | \
-	grep -v /vendor | \
-	grep -v /t/e2e | \
-	grep -v /t/benchmarks | \
-	grep -v /transactions/fake )
+UNIT_TEST_PACKAGES = $(shell go list ./...  | grep -v -e /vendor -e /t/e2e -e /t/benchmarks -e /transactions/fake)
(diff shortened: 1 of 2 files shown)


igor-sirotin:
@jakubgs I'll close this as it doesn't seem relevant to us. Please let me know if you think this is needed.

jakubgs:
Wow, I forgot this was a thing. Sure, I think too many things have changed since then to be worth saving.

PR #3678, opened 2023-06-27
chore: community admin events

mprakhov:
Refactored existing code, where the CommunityDescription clock was modified by the admins.
CommunityDescription can be modified and saved to the DB by clients nodes only when we are receiving an event from the community owner.
If we receive admin events, we collect an array of those events and patch CommunityDescription with these events for the other nodes.
When the owner received admin events - he certifies these events and sends the changed CommunityDescription which will be applied by all other nodes, including admin nodes

Closes # https://github.com/status-im/status-desktop/issues/11084


protocol/communities/community.go
```diff
@@ -2100,59 +2058,315 @@ func (o *Community) deleteChat(chatID string) *CommunityChanges {
...
+func permissionExists(id string, prevPermissions []*protobuf.CommunityTokenPermission) bool {
+	for _, prevPermission := range prevPermissions {
+		if prevPermission.Id == id {
+			return true
+		}
+	}
+	return false
+}
+
+func (o *Community) collectCommuntyChanges(description *protobuf.CommunityDescription) (*CommunityChanges, error) {
```
mprakhov: moved duplicate code into a separate fn

mprakhov:
@cammellos please, take a look when you'll have time.
Refactored the code to not update CommunityDescription clock if it is an admin event until the owner certifies admin events changes

protocol/communities/community.go
```diff
@@ -39,6 +39,7 @@ type Config struct {
 	RequestsToJoin                []*RequestToJoin
 	MemberIdentity                *ecdsa.PublicKey
 	SyncedAt                      uint64
+	AdminsEvents                  []CommunityAdminEvent
```
jrainville: Small detail, but technically a typo
```suggestion
	AdminEvents                  []CommunityAdminEvent
```

protocol/communities/community.go
```diff
@@ -971,191 +1045,21 @@ func (o *Community) MemberIdentity() *ecdsa.PublicKey {
...
-	if err != nil {
-		return nil, err
-	}
-
 	response := o.emptyCommunityChanges()
 
-	if description.Clock <= o.config.CommunityDescription.Clock {
+	// allowEqualClock == true only if this is community event
+	if (allowEqualClock && description.Clock < o.config.CommunityDescription.Clock) ||
+		description.Clock < o.config.CommunityDescription.Clock {
```
jrainville: If I understand the logic, it should be `<=` here.
```suggestion
		description.Clock <= o.config.CommunityDescription.Clock {
```

protocol/communities/community.go
```diff
@@ -1924,8 +1847,42 @@ func (o *Community) AddMemberRevealedAccounts(memberID string, accounts []*proto
...
+	isOwner := o.IsOwner()
+	isAdmin := o.IsAdmin()
+
+	if !isOwner && !isAdmin {
+		return nil, ErrNotAdmin
+	}
+
+	changes := o.addMemberWithRevealedAccounts(dbRequest.PublicKey, roles, accounts)
+
+	if isAdmin {
```
jrainville, line 1857: Are owners still considered admins? If so, that `if` and most of the other ones are not necessary.

protocol/communities/community_event_message.go
```diff
@@ -0,0 +1,257 @@
...
+)
+
+type CommunityAdminEvent struct {
+	AdminEventClock        uint64                                      `json:"clock"`
+	Type                   protobuf.CommunityAdminEvent_EventType      `json:"type"`
+	CommunityConfig        *protobuf.CommunityConfig                   `json:"community_config,omitempty"`
+	TokenPermission        *protobuf.CommunityTokenPermission          `json:"token_permissions,omitempty"`
+	CategoryData           *protobuf.CategoryData                      `json:"category_data,omitempty"`
+	ChannelData            *protobuf.ChannelData                       `json:"channel_data,omitempty"`
+	MemberToAction         string                                      `json:"member_to_action,omitempty"`
```
jrainville: We use camelCase for json names usually.

jrainville: Also, what is `MemberToAction`? The member to whom the action applies? If so, that name is a bit confusing.

protocol/communities/manager.go
```diff
@@ -2422,11 +2314,12 @@ func (m *Manager) HandleCommunityRequestToJoinResponse(signer *ecdsa.PublicKey,
 		return nil, err
 	}
 
-	if !common.IsPubKeyEqual(community.PublicKey(), signer) && !community.IsMemberAdmin(signer) {
+	isAdminSigner := community.IsMemberAdmin(signer)
+	if !common.IsPubKeyEqual(community.PublicKey(), signer) && !isAdminSigner {
```
jrainville: Should this be an `||`?

protocol/communities/manager.go
```diff
@@ -4095,3 +3954,35 @@ func (m *Manager) accountsHasAdminPermission(becomeAdminPermissions []*protobuf.
...
+
+		m.publish(&Subscription{Community: community})
+		return nil
+	}
+
+	if community.IsAdmin() {
+		// Admins sends only all admins events. Based on on those events, clients construct CommunityDesctiption
+		// Constructed CommunityDescription won't be saved on the clients nodes, only when Owner node will
+		// apply these changes to CommunityDescription, new CommunityDescription with the increased
+		// clock will be send to all clients
```
```suggestion
		// clock will be sent to all clients
```

protocol/communities/manager.go
```diff
@@ -4095,3 +3954,35 @@ func (m *Manager) accountsHasAdminPermission(becomeAdminPermissions []*protobuf.
...
+		if err != nil {
+			return err
+		}
+
+		m.publish(&Subscription{Community: community})
+		return nil
+	}
+
+	if community.IsAdmin() {
+		// Admins sends only all admins events. Based on on those events, clients construct CommunityDesctiption
```
```suggestion
		// Admins send only all admins events. Based on on those events, clients construct CommunityDesctiption
```

jrainville (commented):
Huge work, good job.

protocol/communities/community_event_message.go
mprakhov: yes, for ban/unban, kick member

protocol/communities/community.go
mprakhov: we have a separate roles: Owner and Admin
Also, when this fn will be executed, owner will publish CommunityDescription, while Admin CommunityEventMessage with the community events

protocol/communities/community.go
mprakhov: good catch, thank you

protocol/communities/manager.go
mprakhov: no...
If the signer is not a community owner and not an admin, then return an error

protocol/communities/community.go
```diff
@@ -971,191 +1045,21 @@ func (o *Community) MemberIdentity() *ecdsa.PublicKey {
...
-
-	err := ValidateCommunityDescription(description)
-	if err != nil {
-		return nil, err
-	}
-
 	response := o.emptyCommunityChanges()
 
-	if description.Clock <= o.config.CommunityDescription.Clock {
+	// allowEqualClock == true only if this is community event
```
0x-r4bbit: Thank you for adding this comment. Can we make it a bit more specific though? "community event" sounds very general and can mean anything. Maybe add "community event emitted/caused by community admin" ?

protocol/communities/community.go
```diff
@@ -2100,59 +2058,306 @@ func (o *Community) deleteChat(chatID string) *CommunityChanges {
...
 
-func permissionExists(id string, prevPermissions []*protobuf.CommunityTokenPermission) bool {
-	for _, prevPermission := range prevPermissions {
-		if prevPermission.Id == id {
-			return true
+func (o *Community) DeclineRequestToJoin(dbRequest *RequestToJoin) error {
+	// typically, community's clock is increased implicitly when making changes
+	// to it, however in this scenario there are no changes in the community, yet
+	// we need to increase the clock to ensure the owner event is processed by other
+	// nodes.
```
0x-r4bbit: This comment doesn't seem to belong here and should probably be put somewhere around L:2358

protocol/communities/community_event.go
```diff
@@ -0,0 +1,430 @@
...
+			}
+		}
+		prevAdminClock = communityEvent.CommunityEventClock
+		communityEvent.Applied = true
+	}
+	return nil
+}
+
+func (o *Community) updateCommuntyDescriptionByAppliedEvents() error {
+	prevAdminClock := o.config.CommunityDescription.Clock + 1
```
0x-r4bbit: Is this being increased by 1 to ensure it's not equal to the previous one so it'll be processed properly?
If yes, or any other specific reason, I'd add a comment about that here so others will know when they run into this line.

protocol/communities/community_event.go
```diff
@@ -0,0 +1,430 @@
...
+		}
+		prevAdminClock = communityEvent.CommunityEventClock
+	}
+	return nil
+}
+
+func (o *Community) updateCommuntyDescriptionByCommunityEvent(communityEvent CommunityEvent, prevAdminClock uint64) error {
+	if communityEvent.CommunityEventClock <= o.config.CommunityDescription.Clock {
+		return errors.New("clock for admin event is outdated")
+	}
```
0x-r4bbit: I believe this check here confirms my comment above

protocol/communities/manager.go
```diff
@@ -574,92 +574,50 @@ func (m *Manager) CreateCommunity(request *requests.CreateCommunity, publish boo
 	return community, nil
 }
 
-func (m *Manager) CreateCommunityTokenPermission(request *requests.CreateCommunityTokenPermission) (*Community, *CommunityChanges, error) {
+func (m *Manager) CreateCommunityTokenPermission(request *requests.CreateCommunityTokenPermission) (*Community, *CommunityChanges, bool, error) {
```
0x-r4bbit: I think for now this change is fine, but maybe in the future we could consider adding this to the `CommunityChanges`

protocol/communities/manager.go
```diff
@@ -745,6 +703,13 @@ func (m *Manager) checkMemberPermissions(community *Community, removeAdmins bool
 	return nil
 }
 
+func (m *Manager) CheckIfStopCheckingPermissionsPeriodically(community *Community) {
+	if cancel, exists := m.periodicMemberPermissionsTasks.Load(community.IDString()); exists &&
+		len(community.TokenPermissions()) == 0 {
+		close(cancel.(chan struct{})) // Need to cast to the chan
```
0x-r4bbit, line 726: Isn't `close(cancel)` doing the trick?

protocol/messenger_communities.go
```diff
@@ -1455,21 +1455,37 @@ func (m *Messenger) CreateCommunityTokenPermission(request *requests.CreateCommu
...
+
+	err = m.communitiesManager.SaveAndPublish(community)
 	if err != nil {
 		return nil, err
 	}
 
-	if response == nil {
-		response = &MessengerResponse{}
-		response.AddCommunity(community)
+	if community.IsOwner() {
```
0x-r4bbit, line 1598: Shouldn't `admins` do the same thing?

protocol/messenger_communities.go
```diff
@@ -1505,22 +1536,45 @@ func (m *Messenger) DeleteCommunityTokenPermission(request *requests.DeleteCommu
...
 	if err != nil {
 		return nil, err
 	}
 
-	if response == nil {
-		response = &MessengerResponse{}
-		response.AddCommunity(community)
+	// check if members still fulfill the token criteria
+	// We do this in a separate routine to not block this function
+	if community.IsOwner() {
```
0x-r4bbit, line 1659: Same here, I believe the idea was that admins would help with this no?

0x-r4bbit (commented):
Massive PR and very clean work.
I've left some comments and questions.

One question I want to ask here right away:

Am I understanding this correctly that other member nodes will only apply admin changes once they were certified by the owner? Meaning, if and admin accepts someone's request to join, it'll not actually happen until the owner has certified this change?

If that is true, aren't we then back to square one with being ultimately dependent on the control node being online (I thought we wanted to offload that work with admin nodes)?

protocol/communities/manager.go
mprakhov: We have such an implementation in other place, so I just reuse it

protocol/communities/community_event.go
```diff
@@ -0,0 +1,430 @@
...
+		return errors.New("unknown admin community event")
+	}
+	return nil
+}
+
+func (o *Community) NewCommunityEventClock() uint64 {
+	if len(o.config.Events) == 0 {
+		return o.config.CommunityDescription.Clock + 1
+	}
+	return o.config.Events[len(o.config.Events)-1].CommunityEventClock + 1
```
cammellos: I think you want to use a `timestamp` as a seed, just like clock, rather than a straight lamport timestamp, similarly as we do with Clock values (it helps when you make changes and you might not have the latest state)

cammellos (commented):
Still reviewing, sorry a bit slow, but I will take time tomorrow, apologies

protocol/communities/community_event.go
mprakhov: I was following the pattern which we have for the CommunityDescription Clock now
During `increaseClock` clock we just add +1 to the current Clock... 
Also, it allows me to track that we got all admin updates and apply admins events to the CommunityDescription

protocol/migrations/sqlite/1687874359_create_communities_events_table.up.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE IF NOT EXISTS communities_events (
+  id BLOB NOT NULL PRIMARY KEY ON CONFLICT REPLACE,
+  rawEvents BLOB NOT NULL
```
cammellos: we tend to use snake case for tables

protocol/communities/community.go
```diff
@@ -971,191 +1045,21 @@ func (o *Community) MemberIdentity() *ecdsa.PublicKey {
...
-
-		for id, permission := range description.TokenPermissions {
-			if _, ok := o.config.CommunityDescription.TokenPermissions[id]; !ok {
-				if response.TokenPermissionsAdded == nil {
-					response.TokenPermissionsAdded = make(map[string]*protobuf.CommunityTokenPermission)
-				}
-				response.TokenPermissionsAdded[id] = permission
-			}
-		}
+	response, err := o.collectCommuntyChanges(description)
```
cammellos: typo

protocol/communities/community_event.go
cammellos: I see, it's not super important, we can keep a straight lamport timestamp, but we'll have to break ties.

Though not sure about the last sentence "Also, it allows me to track that we got all admin updates and apply admins events to the CommunityDescription"
Basically we can't ensure that we have all admin events, and we should not expect events to be strictly in sequence, in group chat we piggy back the history, so that's probably something we should be consider doing here as well.

For example: 
1) admin-1 sends event admin-1-event
2) admin-2 receives event admin-1-event
3) admin-2 sends event admin-2-event

In this case, admin-2 will also publish together admin-1-event & admin-2-event, therefore preserving the history.
Once the owner collates the events, that won't be necessary anymore.
That's an option worth considering.

protocol/communities/community_event_message.go
```diff
@@ -0,0 +1,257 @@
...
+		if len(communityEvent.MemberToAction) == 0 {
+			return errors.New("invalid community member ban event")
+		}
+
+	case protobuf.CommunityEvent_COMMUNITY_MEMBER_UNBAN:
+		if len(communityEvent.MemberToAction) == 0 {
+			return errors.New("invalid community member unban event")
+		}
+
+	default:
```
cammellos: a new event will result in an error from this function, which means no further processing will happen, that makes upgrading the protocol (i.e adding new events), tricky

protocol/communities/community_event.go
```diff
@@ -0,0 +1,430 @@
...
+	}
+	return nil
+}
+
+func (o *Community) updateCommuntyDescriptionByCommunityEvent(communityEvent CommunityEvent, prevAdminClock uint64) error {
+	if communityEvent.CommunityEventClock <= o.config.CommunityDescription.Clock {
+		return errors.New("clock for admin event is outdated")
+	}
+
+	if !(prevAdminClock == communityEvent.CommunityEventClock || prevAdminClock+1 == communityEvent.CommunityEventClock) {
```
cammellos: it seems like we expect events to be strictly in sequence, but unfortunately that's not possible, since there's no consensus mechanism,which might lead to inconsistencies.

For example:

1) admin-1 publishes event admin-1-event with clock 2
2) admin-2 at the same time publish admin-2-event with clock 2

Client-1 will receive 1 first, and process it, discarding 2.
Client-2 will receive 2 first, and process it, discarding 1.

Did I get that right that this is what the code would do?

In this case, the correct behavior I believe is for Client-1 to process both events (in a pre-established order, in this case you want to break ties deterministically, say by ordering them by clock-pk), as they are not causally related. We need to think of potential conflict, those can be sorted out by some arbitrary rules (i.e if admin-1 makes X an admin and admin-2 removes X from the community, all the clients will agree on what's the outcome based on some rules we decide).

cammellos (commented):
@mprakhov Ok did a first pass, apologies I am a bit slow, fighting a flu :) 

It seems like we are moving in the right direction, left some comments on processing and piggy backing events.

Also Community is often misspelt as Communty (neatpick)

Great work!

protocol/communities/community_event.go
mprakhov: We merge all admin events in one list, so if admin-1 and admin-2 will publish 2 different events with clock 2, clients will merge them and update ([logic inside](https://github.com/status-im/status-go/pull/3678/files#diff-837cc1410fc2cdd0d31b2a9c8b2e03f37ae09a824a55261fb70b2aaffeaf4513R100-R115))
In your case, Client-1 and Client-2 will megre both admin events and apply both

The idea of this check was not to allow apply events in order: 1-2-4, where 3 is unknown, but now I looked at this case from another corner and I think it is impossible as all new admin events will be +1 from previous or from CommunityDescription clock
P.S. This check can be removed

protocol/communities/community_event_message.go
mprakhov: Great point, thank you 👍

protocol/communities/community_event.go
mprakhov: Yes, the current logic is that admin-2  will send admin-1-event & send admin-2-event
But I just realized another problem:
1. admin-1 sends event admin-1-event
2. admin-2 receives event admin-1-event
3. admin-2 sends event admin-1-event & admin-2-event

Owner accepts admin-1 event from the admin-1. He certifies it and publishes updated CommunityDescription with updated Clock
Admin-2 event will be rejected as the owner increased the CommunityDescription clock.

It seems we should not bind admin events to the CommunityDescription clock at all

I need to think about the solution

cammellos: I think admin events should probably be tied to a specific version of a community description (which could be identiified by a clock value), but as you mentioned it should not be bound to it

mprakhov:
@jrainville , @cammellos 
Updated PR, ready for merge
What needs to be done (created a separate tickets, will start working after merging this one)
1. Validate the signature of CommunityDescription which we send during sending events
2. Validate received CommunityDescription from CommunityEventsMessage on the client's side
3. Owner must acknowledge the admin about the rejected event due to the CommunityDescription clock change
4. Owner must 5 min wait for admin events on a startup before updating CommunityDescription

cammellos (approved):
I did my best to look through it, apologies I didn't have so much time, and I will be on holiday for a couple of weeks, it looks ok, I think it's worth taking it for a spin and see how it goes, as long as the cases are all tested I think that will help us drive through the implementation. But best to get it in to avoid having to constantly rebase, it should have little impact on existing code I believe, great job, that's a really tough one!

protocol/communities/community_event.go
```diff
@@ -0,0 +1,461 @@
...
+		if err != nil {
+			return err
+		}
+
+		if metadata.Type != protobuf.ApplicationMetadataMessage_COMMUNITY_DESCRIPTION {
+			return ErrInvalidMessage
+		}
+
+		// TODO: check the signature
+		// Right now test are failing due to signer == nil, but this can be an issue in test
```
jrainville, line 401: Is this todo gonna be addressed in a later PR?

protocol/communities/manager.go
```diff
@@ -1636,12 +1554,14 @@ func (m *Manager) AcceptRequestToJoin(request *requests.AcceptRequestToJoinCommu
...
 	}
 
-	changes, err := community.AddMember(pk, role)
-	if err != nil {
-		return nil, err
-	}
+	// _, err = community.AddMember(pk, role)
+	// if err != nil {
+	// 	return nil, err
+	// }
```
jrainville, line 1560: leftover?

protocol/communities/manager.go
```diff
@@ -1636,12 +1554,14 @@ func (m *Manager) AcceptRequestToJoin(request *requests.AcceptRequestToJoinCommu
...
 
-	_, err = community.AddMemberRevealedAccounts(dbRequest.PublicKey, revealedAccounts, dbRequest.Clock)
+	// _, err = community.AddMemberRevealedAccounts(dbRequest.PublicKey, revealedAccounts, dbRequest.Clock)
+	// TODOM: resolve
```
jrainville, line 1563: leftover?

jrainville (approved):
Great work. 
Maybe just some leftovers.
Clean it up if needed, rebase (squash if you want) and merge that shiii.

protocol/communities/community_event.go
mprakhov: yes, yesterday created the tickets

protocol/communities/community_changes.go
```diff
@@ -73,176 +73,3 @@ func (c *CommunityChanges) HasMemberLeft(identity string) bool {
 	_, ok := c.MembersRemoved[identity]
 	return ok
 }
-
-func EvaluateCommunityChanges(origin, modified *protobuf.CommunityDescription) *CommunityChanges {
```
osmaczko, line 77: @mprakhov I am wondering why it was removed from here and moved back to community.go?

mprakhov: From my understanding it was the same fn as I had, but I wrapped a little bit more code inside

osmaczko: Restored already with: https://github.com/status-im/status-go/pull/3765

mprakhov: yes, I saw and approved that PR :)

PR #3681, opened 2023-06-27
feat(communities): enable RequestCommunityInfo to use a privateKey

jrainville:
Needed for  https://github.com/status-im/status-desktop/issues/11273

Enables `RequestCommunityInfoFromMailserver` to be called with a community private key as well as a public key like before.

### DIFF
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -1854,10 +1854,25 @@ func (m *Messenger) findCommunityInfoFromDB(communityID string) (*communities.Co
 	return community, nil
 }
 
+func (m *Messenger) GetCommunityIDFromKey(communityKey string) string {
+	// Check if the key is a private key
+	privateKey, err := crypto.HexToECDSA(communityKey[2:])
+	communityID := ""
+	if err != nil {
+		// Not a private key, use the public key
+		communityID = communityKey
+	} else {
+		// It is a privateKey
+		communityID = types.HexBytes(crypto.CompressPubkey(&privateKey.PublicKey)).String()
+	}
+	return communityID
+}
+
 // RequestCommunityInfoFromMailserver installs filter for community and requests its details
 // from mailserver. It waits until it has the community before returning it.
 // If useDatabase is true, it searches for community in database and does not request mailserver.
-func (m *Messenger) RequestCommunityInfoFromMailserver(communityID string, useDatabase bool) (*communities.Community, error) {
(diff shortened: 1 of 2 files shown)


protocol/messenger_communities.go
```diff
@@ -1854,10 +1854,28 @@ func (m *Messenger) findCommunityInfoFromDB(communityID string) (*communities.Co
 	return community, nil
 }
 
+func getCommunityIdFromKey(communityKey string) (string, error) {
```
osmaczko: No need to return error, the function always returns nil error.

```suggestion
func getCommunityIdFromKey(privateOrPublicKey string) (string, error) {
```

protocol/messenger_communities.go
```diff
@@ -1854,10 +1854,28 @@ func (m *Messenger) findCommunityInfoFromDB(communityID string) (*communities.Co
...
+	// Check if the key is a private key
+	privateKey, err := crypto.HexToECDSA(communityKey[2:])
+	communityID := ""
+	if err != nil {
+		// Not a private key, use the public key
+		communityID = communityKey
```
osmaczko, line 1863: Shouldn't we check if pubkey is valid here?

osmaczko (approved):

protocol/messenger_communities.go
0x-r4bbit: I also think this should be checked. Although, I do see we haven't made such a check before either. This would ultimately fail when it doesn't find a community given that (invalid) key.

0x-r4bbit (commented):
While I see that there's an issue in desktop to add this here, there doesn't seem to be any info on why this is needed.

If a node happens to own the private key of a community, it can' derive its public key so is there really a need to accept private keys here?

jrainville:
> While I see that there's an issue in desktop to add this here, there doesn't seem to be any info on why this is needed.
> 
> If a node happens to own the private key of a community, it can' derive its public key so is there really a need to accept private keys here?

Yeah sorry the issue not very clear. It's to support the new Import flow design: https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?type=design&node-id=16038-386120&mode=design&t=JvtAGWoVXzdSwx8Q-0

We want to show the community Info before importing.

0x-r4bbit (approved):
Thanks for clarifying

jrainville:
I'm gonna wait to merge this. Seems like the design might change and this might not be needed in the end.

jrainville:
Merging now since we won't be doing tokenized ownership this milestone

PR #3683, opened 2023-06-28
fix: default to attaching all support chainIDs in membership requests

### LINKED ISSUE #11255: Ensure not empty chain ID list is attached to revealed accounts
There are a few changes that happened to the request membership flow of communities:

1. Originally, when a community had permissions set up that users need to fulfill to become a member of the community, user nodes would reveal their wallet addresses to the owner node to proof ownership of the necessary requirements.
2. Then, we've improved this mechanism by letting user nodes figure out on which networks (chainIDs) they own the needed funds and attach that information to the request to join. This then allows the owner/admin node only check on those networks instead all by default.
3. Due to the complexity of possible scenarios when it comes to communities that change from non-permissioned to permissioned and member with revealed or no revealed addresses, we've decided that users now have to reveal their addresses regardless. In short, every request to join requires a user to attach wallet addresses (+ signatures to the request), even if the community doesn't have permissions set up (but might in the future)

@osmaczko has now pointed out the following scenario he's observed that is most likely a bug:

> 1) alice creates open community
> 2) bob joins open community and reveals his addresses
> 3) alice creates permissions to join which bob satisfies
> 4) bob is removed from community and must request to join again with revealed addresses
> At first glance, the issue is at point 2), in case there are no permissions for community, bob sends revealed addresses with empty array of chain IDs.
> Or am I wrong and 4) is expected?

The important part here is **bob sends revealed addresses with empty array of chain IDs**.

This state should only happen if a) the community has permissions set up that need to be satisfied to join the community at the time the request to join was made and b) the user sending the request **does not** have the funds.

The result is that the revealed accounts don't provide chainIDs on which the requirements are satisfied, which is that empty list.

**If however**, the community **didn't** have these permissions set up, meaning it's an open community as described in @osmaczko scenario above, then the default should always be the cartesian product of wallet addresse x all supported chain IDs.
We also have a utility function that does that: https://github.com/status-im/status-go/blob/develop/protocol/communities/manager.go#L4043-L4052

0x-r4bbit:
There's only one scenario in which a `RevealedAccount` will have an empty `ChainIDs` list attached to it:

When the community in question requires users to satisfy certain criteria to join, and the user's wallet does not own the necessary funds on any of the supported chains.

If there are **no** permissions to join on the community, then we want to reveal all (selected) accounts with all supported chainIDs.

This is necessary so that, once the community *does* become permissioned, it'll have address + chain information from all joined members.

Closes: https://github.com/status-im/status-desktop/issues/11255

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -1550,6 +1550,17 @@ func (m *Manager) CheckPermissionToJoin(id []byte, addresses []gethcommon.Addres
 	}
 
 	accountsAndChainIDs := combineAddressesAndChainIDs(addresses, allChainIDs)
+
+	if len(permissionsToJoin) == 0 {
+		// There are no permissions to join on this community at the moment,
+		// so we reveal all accounts + all chain IDs
+		response := &CheckPermissionsResponse{
+			Satisfied:         true,
+			Permissions:       make(map[string]*PermissionTokenCriteriaResult),
+			ValidCombinations: accountsAndChainIDs,
+		}
+		return response, nil
+	}
 	return m.checkPermissionToJoin(permissionsToJoin, accountsAndChainIDs, false)
 }
 


stefandunca (approved):
It looks like the code does what you intended and simplifies the user experience.

PR #3685, opened 2023-06-28
fix: properly initialize NFTMetadataProvider in wallet service

dlipicar:
Wallet Service initialization was moved above WakuV2 Service initialization in https://github.com/status-im/status-go/pull/3443/files#diff-fa95f2b61a88804406e520993c3df20fb3fd543dbff521b5c2b01d0873b2135dR108

Since the Wallet uses Waku to fetch Community NFT metadata, this caused a crash when an account in the wallet held such an NFT. 

This PR moves the Wallet Service initialization back to where it was, and a check was added to make sure the provider is available before attempting to use it.

With this, issue https://github.com/status-im/status-desktop/issues/11244 is fixed.
image

### DIFF
--- a/node/status_node_services.go
+++ b/node/status_node_services.go
@@ -105,11 +105,6 @@ func (b *StatusNode) initServices(config *params.NodeConfig, mediaServer *server
 		services = append(services, wakuext)
 	}
 
-	if config.WalletConfig.Enabled {
-		walletService := b.walletService(accDB, accountsFeed)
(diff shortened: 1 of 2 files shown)


IvanBelyakoff (approved):

stefandunca (approved):
Great catch!

endulab (approved):

cammellos (approved):

PR #3687, opened 2023-06-28
fix: ROW_NUMBER() not work on ios for some reason

qfrank:
because ROW_NUMBER() not work on ios for some reason, i create a temp table used `AUTOINCREMENT` as the replacement.

relate mobile [issue](https://github.com/status-im/status-mobile/issues/16421) 

relate geth.log:
```
t=2023-06-28T08:11:56+0000 lvl=eror msg="failed to create account" error="failed to migrate up: near \"AS\": syntax error in line 0: ALTER TABLE keypairs_accounts ADD COLUMN position INT NOT NULL DEFAULT 0;\n\nUPDATE keypairs_accounts AS ka SET position = ka2.rowNumber FROM (\n  SELECT created_at, ROW_NUMBER() OVER (ORDER BY created_at) AS rowNumber FROM keypairs_accounts) AS ka2\nWHERE ka.created_at = ka2.created_at;"
```


appdatabase/migrations/sql/1687249080_add_position_accounts.up..sql
```diff
@@ -1,5 +1,76 @@
-ALTER TABLE keypairs_accounts ADD COLUMN position INT NOT NULL DEFAULT 0;
```
cammellos, line 1: Filename looks wrong, `..sql`

cammellos: but we can't change it, otherwise it will break existing migrated clients, nevermind

cammellos (commented):
Looks good to me, it'd be good to have someone from the wallet team reviewing this

pavloburykh:
> Looks good to me, it'd be good to have someone from the wallet team reviewing this

Will post it also here: current PR results in a new error on login (now for both Android and IOS platforms) https://github.com/status-im/status-mobile/pull/16426#issuecomment-1611420837

photo_2023-06-28 16 31 48

rasom (approved):

PR #3689, opened 2023-06-28
Fixed concurrent write to blockchain status map

IvanBelyakoff:
Fix for:

```
fatal error: concurrent map writes
fatal error: concurrent map writes

goroutine 62964 [running]:
github.com/status-im/status-go/services/wallet.NewService.func1(0x66eed, {0x7f21469ad372, 0x2})
        /home/blade/work/projects/Status/status-desktop/vendor/status-go/services/wallet/service.go:76 +0x127
github.com/status-im/status-go/rpc/chain.(*ClientWithFallback).SetIsConnected(0x1c00185a900, 0x1)
        /home/blade/work/projects/Status/status-desktop/vendor/status-go/rpc/chain/client.go:147 +0xed
github.com/status-im/status-go/rpc/chain.(*ClientWithFallback).makeCallSingleReturn.func1()
        /home/blade/work/projects/Status/status-desktop/vendor/status-go/rpc/chain/client.go:210 +0x67
github.com/afex/hystrix-go/hystrix.Go.func1({0x1c001e35e90?, 0x7f21458ada86?})
        /home/blade/work/projects/Status/status-desktop/vendor/status-go/vendor/github.com/afex/hystrix-go/hystrix/hystrix.go:57 +0x1b
github.com/afex/hystrix-go/hystrix.GoC.func3()
        /home/blade/work/projects/Status/status-desktop/vendor/status-go/vendor/github.com/afex/hystrix-go/hystrix/hystrix.go:160 +0x1b5
created by github.com/afex/hystrix-go/hystrix.GoC
        /home/blade/work/projects/Status/status-desktop/vendor/status-go/vendor/github.com/afex/hystrix-go/hystrix/hystrix.go:116 +0x3ea

```

### DIFF
--- a/services/wallet/service.go
+++ b/services/wallet/service.go
@@ -3,6 +3,7 @@ package wallet
 import (
 	"database/sql"
 	"encoding/json"
+	"sync"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
@@ -59,7 +60,11 @@ func NewService(
 		Publisher: walletFeed,
 	}
 	blockchainStatus := make(map[uint64]string)
+	mutex := sync.Mutex{}
 	rpcClient.SetWalletNotifier(func(chainID uint64, message string) {
+		mutex.Lock()
+		defer mutex.Unlock()
+
 		if len(blockchainStatus) == 0 {
 			networks, err := rpcClient.NetworkManager.Get(false)
 			if err != nil {


services/wallet/service.go
```diff
@@ -59,7 +60,11 @@ func NewService(
 		Publisher: walletFeed,
 	}
 	blockchainStatus := make(map[uint64]string)
+	mutex := sync.Mutex{}
```
dlipicar, line 63: hmm having doubts about instancing the mutex as a local variable and using it inside the callback. Are we sure this works like we expect it to work?
Also seems like blockchainStatus works like some kind of implicit member variable, perhaps we should explicitly define this and the mutex as members for clarity

services/wallet/service.go
```diff
@@ -59,7 +60,11 @@ func NewService(
...
 	rpcClient.SetWalletNotifier(func(chainID uint64, message string) {
```
stefandunca, line 64: The best is to move this listener code outside the init function to have a cleaner implementation and make mutex a member of the service and call it `rpcClientMutex`

stefandunca (approved):

stefandunca:
I approve the quick fix; however, wondering why the service manages the RPC state notifications.
I also see that SetWalletNotifier triggers for every chaiID individually, and we transform this into all enabled networks event, can't this be improved?

services/wallet/service.go
IvanBelyakoff: yeah guys, just wanted a quick fix since I noticed it and it seems like it happens quite rare, so did not spend time on refactoring. Ok let me refactor it a bit later

services/wallet/service.go
IvanBelyakoff: No, Dario, I am not 100% sure, though I think it should work. Just was going to make a quick fix, will refactor when I've got more time

IvanBelyakoff:
> I approve the quick fix; however, wondering why the service manages the RPC state notifications. I also see that SetWalletNotifier triggers for every chaiID individually, and we transform this into all enabled networks event, can't this be improved?

I don't think we should handle RPC notifications in wallet service. maybe we have a proper service for that? Though not sure if we've got events there and if they are handled on NIM side.
The crash happened when 2 different chains got disconnected or connected simultaneously

PR #3693, opened 2023-06-29
make sure profile image is saved before node started

qfrank:
relate mobile [issue](https://github.com/status-im/status-mobile/issues/15794)


VERSION
```diff
@@ -1 +1 @@
-0.159.9
\ No newline at end of file
+0.159.11
```
saledjenic: @qfrank don't we need to set to `0.160.0` after `0.159.9`? That's how we used to do in the past.

saledjenic (approved):

VERSION
cammellos: that's not how we did it, maybe there's some confusion, it should be 0.159.10, 0.159.11...., 0.159.100 etc
we use a slightly modified version of semver (x.y.z), so you bump y to indicate a breaking change (which I don't believe is the case), while z for a non breaking change, we don't bump y once z reaches 9

cammellos (approved):

PR #3694, opened 2023-06-29
Chore/community permissions integration tests

osmaczko:
- cleaned up tests a bit
- added TestBecomeMemberPermissions integration test
    - if bob satisfy criteria, then bob can read channel messages
    - if bob doesn't satisfy criteria, then bob can't read channel messages
- fixed a bug, where ratchet key would be propagated to all members before doing permissions check


0x-r4bbit (approved):
Nice fixes and great test refactoring as well.
I couldn't find anything that raised a red flag.

PR #3698, opened 2023-06-29
chore: synchronization improvements applied to keypairs/accounts

saledjenic:
This is the first step of improvements over keypairs/keycards/accounts.
- `SyncKeypairFull` protobuf removed
- `SyncKeypair` protobuf is used for syncing all but the watch only accounts
- `SyncAccount` is used only for syncing watch only accounts
- related keycards are synced together with a keypair
- on any keypair change (either it's just a keypair name or any change made over an account which belongs to that keypair) entire keypair is synced including related keycards
- on any watch only account related change, that account is synced with all its details

The second part of improvements will be about keycards only, that will be a new PR, just for easier reviewing the code. In the next step we will sync entire keypair on each keycard related update, just like it's now for any keypair/wallet account related change.


protocol/protobuf/application_metadata_message.proto
```diff
@@ -78,9 +78,8 @@ message ApplicationMetadataMessage {
     SYNC_KEYCARD_ACTION = 63;
     SYNC_SOCIAL_LINKS = 64;
     SYNC_ENS_USERNAME_DETAIL = 65;
-    SYNC_FULL_KEYPAIR = 66;
```
saledjenic, line 81: I just simply made changes in these 2 .proto files without taking care of backwardcompatibility just because we have not launched the MVP yet, once that's done being backward compatible will be mandatory. Hope you're ok with that.

protocol/messenger_handler.go
```diff
@@ -2996,21 +2996,27 @@ func mapSyncAccountToAccount(message *protobuf.SyncAccount, accountOperability a
 	}
 }
 
-func (m *Messenger) resolveAccountOperability(keyUID string, defaultWalletAccount bool, accountReceivedFromRecovering bool) (accounts.AccountOperable, error) {
-	knownKeycardsForKeyUID, err := m.settings.GetKeycardByKeyUID(keyUID)
-	if err != nil {
-		if err == accounts.ErrDbKeypairNotFound {
+func (m *Messenger) resolveAccountOperability(keyUID string, syncKpMigratedToKeycard bool, chatAccount bool, defaultWalletAccount bool, accountReceivedFromLocalPairing bool) (accounts.AccountOperable, error) {
+	if !syncKpMigratedToKeycard {
+		// We're here when we receive a keypair form the paired device which is either:
```
rasom: probably "from" not "form"

protocol/messenger_wallet.go
```diff
@@ -273,32 +249,72 @@ func (m *Messenger) syncKeypair(keypair *accounts.Keypair, fullKeypairSync bool,
...
+		return nil
+	}
+
+	var (
+		clock    uint64
+		kpKeyUID string
+	)
+	_, chat := m.getLastClockWithRelatedChat()
+
+	switch obj := instance.(type) {
```
rasom: I'm wondering whether we could just add some interface with `KeyUID` and `Clock` fields and avoid having `switch` here. Not a big deal, but it might have looked a bit simpler then.

rasom (approved):

protocol/messenger_wallet.go
saledjenic: I agree that could be another option, but in that case we would need to update `syncWalletAccount` function as well. This way we use `obj` in the line `292`, cause in case of a watch only account we don't fetch it from the db in order to dispatch it, but use the `obj`.

alaibe (approved):

qfrank (approved):

saledjenic:
Guys, since your last review I had to update code a bit, but the general idea/concept is the same, what was changed since you reviewed it is:
- `deleteKeystoreFileForAddress` function added, which actually part of code which was previously in accounts api, now is moved to messenger and identified as a new function, cause it needs to be called for synced account to remove local keystore file if needed when accounts is synced
- `resolveAndSetAccountPropsMaintainedByBackend` function added, it takes care about account position and account operability (client doesn't need to set that explicitly)
- test from `TestMessengerSyncKeycardsStateSuite` are commented will be resolved in the second step keypair improvements

Basically that's it, if you want to review it once again feel free to do it, if not I will open an issue for mobile that you can align the mobile app with these changes and going to merge this and continue with second step of improvements which will be mostly keycard related (an easier one).
@rasom @alaibe @qfrank 

Mobile app issue: https://github.com/status-im/status-mobile/issues/16493

PR #3699, opened 2023-06-29
feat(wallet) make remaining filter APIs async

stefandunca:
### Updates status-desktop [#11170](https://github.com/status-im/status-desktop/issues/11170)

Implement `activity.Scheduler` to serialize and limit the number of calls on the activity service. This way, we protect from inefficient parallel queries and easily support async and rate limiting based on the API requirements.

Refactor the activity APIs async and use the Scheduler for managing the activity service calls configured with one of the two rules: cancel ignore.

Also, add new `OrderedMap` direct dependency [github.com/wk8/go-ordered-map/v2](https://github.com/wk8/go-ordered-map/blob/85ca4a2b29d3241fa4513f82be3d38fe2392a791/go.mod#L1)


alaibe (approved):

stefandunca:
The last push implements the TODO and makes the TaskType carry also the policy

services/wallet/activity/scheduler.go
dlipicar, line 1: awesome util! If I see correctly there's no dependency on activity stuff, perhaps it could be moved to some common directory.

stefandunca: It is used only for activity; it makes sense to stay close, right? How about moving it when we need it by some other package?

Issue #3684, opened 2023-06-28
Storing account data when creating new account


flexsurfer:
Currently, when we create a new account (profile) we store its data in two different tables, one is unencrypted which is available before login, and another is encrypted, and it starting to be confusing, we might store some in one table and others in another table, and in client we merge these two maps, it would be great if we could do it in status go, store each field only in one table and merge tables when they are sent to client

PR #3701, opened 2023-06-30
store customization color in settings

flexsurfer:
store customization color in settings

### DIFF
--- a/api/geth_backend.go
+++ b/api/geth_backend.go
@@ -1044,6 +1044,7 @@ func (b *GethStatusBackend) generateOrImportAccount(mnemonic string, request *re
 
 	settings.DeviceName = request.DeviceName
 	settings.DisplayName = request.DisplayName
+	settings.CustomizationColor = request.CustomizationColor
 	settings.PreviewPrivacy = request.PreviewPrivacy
 	settings.CurrentNetwork = request.CurrentNetwork
 
--- a/appdatabase/migrations/bindata.go
+++ b/appdatabase/migrations/bindata.go
@@ -76,6 +76,7 @@
 // 1687269871_add_device_name.up.sql (108B)
 // 1687506642_include_watch_only_account_setting.up.sql (81B)
 // 1688022264_add_include_watch_only_account_to_settings_sync_clock.up.sql (98B)
+// 1688113564_add_customization_color.up.sql (124B)
 // doc.go (74B)
 
 package migrations
@@ -86,6 +87,7 @@ import (
 	"crypto/sha256"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
@@ -95,7 +97,7 @@ import (
 func bindataRead(data []byte, name string) ([]byte, error) {
 	gz, err := gzip.NewReader(bytes.NewBuffer(data))
 	if err != nil {
-		return nil, fmt.Errorf("read %q: %w", name, err)
+		return nil, fmt.Errorf("read %q: %v", name, err)
 	}
 
 	var buf bytes.Buffer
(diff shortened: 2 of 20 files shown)


qfrank:
I thought we only store customization color into table `accounts.customizationColor` ? 🤔 cc @cammellos

cammellos:
> I thought we only store customization color into table `accounts.customizationColor` ? thinking cc @cammellos

we should probably store it in a single place, whether `accounts` or `settings` depends whether is used pre-login (`accounts` or post-login `settings`).

qfrank:
I prefer only store it in `accounts` , implementation of syncing customization color is based on such assuming 😅, 
[draft of sync customization color](https://github.com/status-im/status-go/pull/3702) , currently we need customization color before login, right ?  cc @cammellos

flexsurfer:
> we should probably store it in a single place, whether `accounts` or `settings` depends whether is used pre-login (`accounts` or post-login `settings`).

its needed before and after, same as name and image, but name and image stored in both tables and color not, so i just added color, i created an issue https://github.com/status-im/status-go/issues/3684 for changing this , but for now i just want to make it work similar to other fields

cammellos:
> 

Ok, so probably we want to stop storing name/image in both tables, and just save it in one only (`accounts` if pre-login & only in `settings` for post-login)

qfrank:
FYI, My draft PR added an API for updating customization color, i plan to finish it today or tomorrow @flexsurfer

flexsurfer:
so currently we have an issue, we don't have color in settings, what we gonna do? @qfrank will it be fixed by your PR ?

qfrank:
my two cents, if client want get and display customization color, endpoints `OpenAccounts` should already returned it, if we need update it, we can use [UpdateAccountCustomizationColor](https://github.com/status-im/status-go/blob/396cb285c054edffcd5e4c0eaa8c696cfa726236/services/accounts/multiaccounts.go#L33) , WDYT? cc @flexsurfer

flexsurfer:
> my two cents, if client want get and display customization color, endpoints `OpenAccounts` should already returned it, if we need update it, we can use [UpdateAccountCustomizationColor](https://github.com/status-im/status-go/blob/396cb285c054edffcd5e4c0eaa8c696cfa726236/services/accounts/multiaccounts.go#L33) , WDYT? cc @flexsurfer

OpenAccounts is used only once when we open the app, then we use settings when we login into account, so doesn't looks as a good API, status-go should care about this, client doesn't care where and how status-go stores props, accounts is just part of settings but unencrypted , and im not sure `CustomizationColorUpdatedAt` should be stored in it, its more like a settings, we don't need it before login, but anyway, my suggestion to join tables when we request settings , so we get full settings after login

flexsurfer:
for example when we create account, we login and load settings but there is no account's props in settings, so we have to load accounts get account by settings uid and merge some settings , this doesn't look great, we would expect to have all settings when we request them from status-go

qfrank:
Frontend proberbly don't need to know existence of `CustomizationColorUpdatedAt`, it's used for syncing. 
If we only store CustomizationColor in `accounts` , can you adjust your PR so that we don't have two places store it? And I'll remove the endpoint `UpdateAccountCustomizationColor` in my PR

> for example when we create account, we login and load settings but there is no account's props in settings, so we have to load accounts get account by settings uid and merge some settings , this doesn't look great, we would expect to have all settings when we request them from status-go

Thank you for clarification! 🙂 @flexsurfer

flexsurfer:
> can you adjust your PR so that we don't have two places store it?

i can't adjust it because its needed in both places, but i have an issues for that https://github.com/status-im/status-go/issues/3684 if you are interested you can take it please :)

flexsurfer:
because we need need similar functionality for all "duplicated" fields like name image identical etc , so it should be done all in one PR, and as you (@qfrank) are working on syncing color, name and images are also should be synced, so we need to implement them the same way

qfrank:
> if you are interested you can take it

okay, I'll take it when i got time 🙂

regarding adjusting, let's see what @cammellos would say ❤️

flexsurfer:
hey @qfrank thank you, we decided to pause this PR, and @cammellos will take a look at the issue

PR #3702, opened 2023-06-30
Feat/sync customization color

qfrank:
Major changes:
- Added ablity to sync account customization color, but it will only sync after user login. Sync it before login not implemented yet as there's something big(e.g. ability to fetch data using network before login) need to be addressed before it. I didn't add the endpoint to change customization color yet, we can add it later simply if we need it, because there is a [TBD](https://github.com/status-im/status-go/pull/3701#issuecomment-1614567955)

- made timesource singleton and not as attached service with node
- removed configuration item `EnableNTPSync` since we don't really give a user a choice

status: ready.


qfrank:
Hi @saledjenic , can you help check if desktop team used [UpdateAccount](https://github.com/status-im/status-go/blob/2c89047834f9cc244302d2b5bb7f5ba0f5258652/services/accounts/multiaccounts.go#L27) to update customization color? Because I found [TestAccountsUpdate](https://github.com/status-im/status-go/blob/2c89047834f9cc244302d2b5bb7f5ba0f5258652/multiaccounts/database_test.go#L39) failed. If we want update `customization color`, it would be better to use a separate endpoint IMO. If being used, could you tell me when it will be used? before or after login? thank you

saledjenic:
@qfrank no, the desktop team doesn't use [UpdateAccount](https://github.com/status-im/status-go/blob/2c89047834f9cc244302d2b5bb7f5ba0f5258652/services/accounts/multiaccounts.go#L27), at least not yet.

qfrank:
Hi @cammellos , I guess we have this [switch](https://github.com/status-im/status-go/blob/99674ba4e517a2730fe22a75e957a89c8baddaf2/params/config.go#L425) is because privacy, but we need timesource available to make sure time accurate, WDYT?  I prefer removing this switch and make timesource singleton and not as attached service to status node(currently not implemented) so that login/logout won't impact timesource. cc @Samyoul

mobile/status.go
```diff
@@ -280,6 +282,12 @@ func CreateAccountAndLogin(requestJSON string) string {
 
 	api.RunAsync(func() error {
 		log.Debug("starting a node and creating config")
+		var now uint64
+		if now, err = timesource.GetCurrentTimeInMillis(); err != nil {
+			log.Error("CreateAccountAndLogin. failed to get current time in millis", "error", err)
+			return err
+		}
+		request.CustomizationColorUpdatedAt = now
```
cammellos: I don't think this is the right place for updating the customization color, it should go in `CreateAccountAndLogin` and not sure it should be part of the request, since it's not user controlled?

multiaccounts/database.go
```diff
@@ -122,7 +125,7 @@ func (db *Database) GetAccountKDFIterationsNumber(keyUID string) (kdfIterationsN
 }
 
 func (db *Database) GetAccounts() (rst []Account, err error) {
-	rows, err := db.db.Query("SELECT  a.name, a.loginTimestamp, a.identicon, a.colorHash, a.colorId, a.customizationColor, a.keycardPairing, a.keyUid, a.kdfIterations, ii.name, ii.image_payload, ii.width, ii.height, ii.file_size, ii.resize_target, ii.clock FROM accounts AS a LEFT JOIN identity_images AS ii ON ii.key_uid = a.keyUid ORDER BY loginTimestamp DESC")
+	rows, err := db.db.Query("SELECT  a.name, a.loginTimestamp, a.identicon, a.colorHash, a.colorId, a.customizationColor, a.customizationColorUpdatedAt, a.keycardPairing, a.keyUid, a.kdfIterations, ii.name, ii.image_payload, ii.width, ii.height, ii.file_size, ii.resize_target, ii.clock FROM accounts AS a LEFT JOIN identity_images AS ii ON ii.key_uid = a.keyUid ORDER BY loginTimestamp DESC")
```
cammellos: we use snake_case

cammellos: or we don't :)

cammellos: I think though it should be called something that refers to the fact that is a `clock` value, since we would treat it as such in the future, once we allow modifying the color (if `customizationColorUpdatedAt` is > `time.now` if an edit occours locally, we would do `customizationColorUpdatedAt++`, so it would be a clock value)

cammellos (approved):

PR #3704, opened 2023-06-30
Feat/pending tx refactor

### LINKED ISSUE #10474: Wallet: make deleting pending transaction robust
Delete pending transactions in `status-go` when the complete or failed downloaded transaction is first written to the `transfer` DB.

The pending transactions are deleted on the Nim side when a new transaction state is received from the `status-go`, which can cause side effects in corner cases.

IvanBelyakoff:
- Moved pending transactions to `transactions` module
- Handle creation and deletion of pending transactions automatically on status-go side.

Fixed [#10474](https://github.com/status-im/status-desktop/issues/10474)

### DIFF
--- a/api/geth_backend.go
+++ b/api/geth_backend.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/imdario/mergo"
 
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	ethcrypto "github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/log"
@@ -30,18 +31,19 @@ import (
 	"github.com/status-im/status-go/logutils"
 	"github.com/status-im/status-go/multiaccounts"
 	"github.com/status-im/status-go/multiaccounts/accounts"
-	"github.com/status-im/status-go/multiaccounts/common"
+	multiacccommon "github.com/status-im/status-go/multiaccounts/common"
 	"github.com/status-im/status-go/multiaccounts/settings"
 	"github.com/status-im/status-go/node"
 	"github.com/status-im/status-go/nodecfg"
 	"github.com/status-im/status-go/params"
 	"github.com/status-im/status-go/protocol"
-	identityUtils "github.com/status-im/status-go/protocol/identity"
+	identityutils "github.com/status-im/status-go/protocol/identity"
 	"github.com/status-im/status-go/protocol/identity/colorhash"
 	"github.com/status-im/status-go/protocol/requests"
 	"github.com/status-im/status-go/rpc"
 	"github.com/status-im/status-go/services/ext"
 	"github.com/status-im/status-go/services/personal"
+	"github.com/status-im/status-go/services/rpcfilters"
 	"github.com/status-im/status-go/services/typeddata"
 	"github.com/status-im/status-go/signal"
 	"github.com/status-im/status-go/sqlite"
@@ -1062,7 +1064,7 @@ func (b *GethStatusBackend) generateOrImportAccount(mnemonic string, request *re
 	account := multiaccounts.Account{
 		KeyUID:             info.KeyUID,
 		Name:               request.DisplayName,
-		CustomizationColor: common.CustomizationColor(request.CustomizationColor),
+		CustomizationColor: multiacccommon.CustomizationColor(request.CustomizationColor),
 		KDFIterations:      sqlite.ReducedKDFIterationsNumber,
 	}
 	if request.ImagePath != "" {
@@ -1253,7 +1255,7 @@ func enrichMultiAccountBySubAccounts(account *multiaccounts.Account, subaccs []*
 			}
 			account.ColorHash = colorHash
 
-			colorID, err := identityUtils.ToColorID(pk)
+			colorID, err := identityutils.ToColorID(pk)
 			if err != nil {
 				return err
 			}
@@ -1274,7 +1276,7 @@ func enrichMultiAccountByPublicKey(account *multiaccounts.Account, publicKey typ
 	}
 	account.ColorHash = colorHash
 
-	colorID, err := identityUtils.ToColorID(pk)
+	colorID, err := identityutils.ToColorID(pk)
 	if err != nil {
 		return err
 	}
@@ -1584,7 +1586,12 @@ func (b *GethStatusBackend) SendTransaction(sendArgs transactions.SendTxArgs, pa
 		return
 	}
 
-	go b.statusNode.RPCFiltersService().TriggerTransactionSentToUpstreamEvent(hash)
+	go b.statusNode.RPCFiltersService().TriggerTransactionSentToUpstreamEvent(&rpcfilters.PendingTxInfo{
+		Hash:    common.Hash(hash),
+		Type:    string(transactions.WalletTransfer),
+		From:    common.Address(sendArgs.From),
+		ChainID: b.transactor.NetworkID(),
+	})
 
 	return
 }
@@ -1600,7 +1607,12 @@ func (b *GethStatusBackend) SendTransactionWithChainID(chainID uint64, sendArgs 
 		return
 	}
 
-	go b.statusNode.RPCFiltersService().TriggerTransactionSentToUpstreamEvent(hash)
+	go b.statusNode.RPCFiltersService().TriggerTransactionSentToUpstreamEvent(&rpcfilters.PendingTxInfo{
+		Hash:    common.Hash(hash),
+		Type:    string(transactions.WalletTransfer),
+		From:    common.Address(sendArgs.From),
+		ChainID: b.transactor.NetworkID(),
+	})
 
 	return
 }
@@ -1611,7 +1623,12 @@ func (b *GethStatusBackend) SendTransactionWithSignature(sendArgs transactions.S
 		return
 	}
 
-	go b.statusNode.RPCFiltersService().TriggerTransactionSentToUpstreamEvent(hash)
+	go b.statusNode.RPCFiltersService().TriggerTransactionSentToUpstreamEvent(&rpcfilters.PendingTxInfo{
+		Hash:    common.Hash(hash),
+		Type:    string(transactions.WalletTransfer),
+		From:    common.Address(sendArgs.From),
+		ChainID: b.transactor.NetworkID(),
+	})
 
 	return
 }
--- a/appdatabase/database.go
+++ b/appdatabase/database.go
@@ -26,6 +26,35 @@ var customSteps = []sqlite.PostStep{
 	{Version: 1687193315, CustomMigration: migrateWalletTransferFromToAddresses, RollBackVersion: 1686825075},
 }
 
+func doMigration(db *sql.DB) error {
+	lastMigration, migrationTableExists, err := sqlite.GetLastMigrationVersion(db)
+	if err != nil {
+		return err
+	}
+
+	if !migrationTableExists || (lastMigration > 0 && lastMigration < nodeCfgMigrationDate) {
+		// If it's the first time migration's being run, or latest migration happened before migrating the nodecfg table
+		err = migrationsprevnodecfg.Migrate(db)
+		if err != nil {
+			return err
+		}
+
+		// NodeConfig migration cannot be done with SQL
+		err = nodecfg.MigrateNodeConfig(db)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Run all the new migrations
+	err = migrations.Migrate(db, customSteps)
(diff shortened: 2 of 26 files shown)


transactions/pending.go
```diff
@@ -0,0 +1,306 @@
...
+	commandContext, cancel := context.WithTimeout(ctx, 10*time.Minute)
+	defer cancel()
+
+	err := watchTxCommand.Command()(commandContext)
+	if err != nil {
+		log.Error("watchTxCommand error", "error", err, "chainID", client.ChainID, "hash", transactionHash)
+		return err
+	}
+
+	return tm.DeletePending(client.ChainID, transactionHash)
```
IvanBelyakoff: Deleting pending tx automatically now

services/wallet/transfer/transaction.go
```diff
@@ -78,231 +87,110 @@ type MultiTransactionCommandResult struct {
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
-const multiTransactionColumns = "from_address, from_asset, from_amount, to_address, to_asset, to_amount, type, timestamp"
+func createPendingTransactions(hashes map[uint64][]types.Hash, data []*bridge.TransactionBridge,
+	multiTransactionID MultiTransactionIDType, multiTransaction *MultiTransaction) []*transactions.PendingTransaction {
```
dlipicar: no need to pass `multiTransactionID` separately, it's part of the `multiTransaction` object

transactions/pending.go
dlipicar: Will the entry be deleted if the application is restarted between creating it and fetching it from the blockchain?

services/wallet/transfer/transaction.go
IvanBelyakoff: > no need to pass `multiTransactionID` separately, it's part of the `multiTransaction` object

hm. I saw it not being set in code. And I was not sure if it should be and when. Ok, will check that once again

IvanBelyakoff: > > no need to pass `multiTransactionID` separately, it's part of the `multiTransaction` object
> 
> hm. I saw multiTransactionID not being set for multiTransaction in code. And I was not sure if it should be and when. Ok, will check that once again

dlipicar: It didn't before, but now it gets set when you insert the item into the DB. Anyway, there's no functional change, just a cleanup.

transactions/pending.go
IvanBelyakoff: Good question. Currently client watches for pending transactions, watch is never called by itself on status-go side. So based on this logic, client is responsible to check for pending transactions and call `watch` after restart. Which should cause proper deletion.
But that is a good catch, as I missed it on client side. Wdyt?

dlipicar: We can leave it for later if it's still being handled client-side! I just thought the client was not supposed to be doing that anymore after these changes.

appdatabase/database.go
```diff
@@ -26,34 +26,58 @@ var customSteps = []sqlite.PostStep{
 	{Version: 1687193315, CustomMigration: migrateWalletTransferFromToAddresses, RollBackVersion: 1686825075},
 }
 
-// InitializeDB creates db file at a given path and applies migrations.
-func InitializeDB(path, password string, kdfIterationsNumber int) (*sql.DB, error) {
-	db, err := sqlite.OpenDB(path, password, kdfIterationsNumber)
+// initializeUnencryptedDB creates unencrypted db file at a given path and applies migrations.
```
dlipicar: hmm where is this used? I thought we determined having any unencrypted data could lead to privacy issues.

dlipicar (approved):

transactions/pending.go
IvanBelyakoff: I think I did not make it clear. Client will not call `deletePendingTransaction` or `storePendingTransaction` any more like it did before.  But `watch` will still be done from client side for now as we discussed during some standup meeting.

appdatabase/database.go
IvanBelyakoff: It is not used in code. I used it for testing to check DB, because it is very annoying to set all the options and password to open an encrypted DB each time. I specifically made the function private to the module so that noone would use it by mistake from outside.
But a dev while working can make it public for his purposes and just use it.
I can add a comment to never use it in production or make public to make it more clear

dlipicar (approved):

api/geth_backend.go
```diff
@@ -30,7 +31,7 @@ import (
 	"github.com/status-im/status-go/logutils"
 	"github.com/status-im/status-go/multiaccounts"
 	"github.com/status-im/status-go/multiaccounts/accounts"
-	"github.com/status-im/status-go/multiaccounts/common"
+	multiAccCommon "github.com/status-im/status-go/multiaccounts/common"
```
stefandunca: 👮‍♂️`macommon` might work better. This looks like a variable name, see https://go.dev/blog/package-names

services/wallet/api.go
```diff
@@ -237,86 +238,68 @@ func (api *API) DeleteSavedAddress(ctx context.Context, address common.Address,
...
 	log.Debug("call to get pending outbound transactions by address")
-	rst, err := api.s.transactionManager.GetPendingByAddress([]uint64{api.s.rpcClient.UpstreamChainID}, address)
+	rst, err := api.s.pendingTxManager.GetPendingByAddress([]uint64{api.s.rpcClient.UpstreamChainID}, address)
 	log.Debug("result from database for pending transactions by address", "len", len(rst))
 	return rst, err
 }
 
-func (api *API) GetPendingOutboundTransactionsByAddressAndChainID(ctx context.Context, chainIDs []uint64, address common.Address) ([]*transfer.PendingTransaction, error) {
+func (api *API) GetPendingOutboundTransactionsByAddressAndChainID(ctx context.Context, chainIDs []uint64,
+	address common.Address) ([]*transactions.PendingTransaction, error) {
```
stefandunca, line 281: This doesn't look very clear, I had to look closer that it is not a code line. Maybe more indentation? Or leave the line wrap to do its thing?

services/wallet/transfer/transaction.go
```diff
@@ -453,7 +223,74 @@ func (tm *TransactionManager) UpdateMultiTransaction(multiTransaction *MultiTran
 	return updateMultiTransaction(tm.db, multiTransaction)
 }
 
-func (tm *TransactionManager) CreateMultiTransactionFromCommand(ctx context.Context, command *MultiTransactionCommand, data []*bridge.TransactionBridge, bridges map[string]bridge.Bridge, password string) (*MultiTransactionCommandResult, error) {
+func (tm *TransactionManager) CreateMultiTransactionFromCommand(ctx context.Context, command *MultiTransactionCommand,
+	data []*bridge.TransactionBridge, bridges map[string]bridge.Bridge, password string) (*MultiTransactionCommandResult, error) {
+
+	multiTransaction := multiTransactionFromCommand(command)
+
+	multiTransactionID, err := insertMultiTransaction(tm.db, multiTransaction)
```
stefandunca, line 231: How about adding a follow-up task to get all these operations as a SQL transaction?

transactions/pending.go
stefandunca: I agree with @dlipicar here; this deletion should be done within the same SQL transaction as the save/update of the original transaction (success or failure). I was expecting clients to have no more access to the raw delete pending transaction API.

services/rpcfilters/transaction_sent_to_upstream_event.go
```diff
@@ -4,23 +4,29 @@ import (
...
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
-
-	"github.com/status-im/status-go/eth-node/types"
 )
 
+type PendingTxInfo struct {
+	Hash    common.Hash
+	Type    string
+	From    common.Address
```
stefandunca, line 14: Not sure why the address matter for the transfer identity. Why not the usual ChainID + Hash?

stefandunca (approved):
Nice refactoring. There is an improvement in moving the mechanism of deleting transactions from the client side (Nim).

However, I'm not sure that doing it as a response to an event like any client is the best way. Going even further and making it an internal SQL transaction will make the operation robust.

Left some comments inline.

api/geth_backend.go
IvanBelyakoff: Thank you @stefandunca, nice catch.

transactions/pending.go
IvanBelyakoff: @stefandunca I overlooked that `DeletePending` is still public. I made it private now, so no more raw access to delete pending transaction. Though as we discussed it before, watching for pending transactions and downloading new ones are two different separate processes. One change we could make is downloading the confirmed pending tx from network to fill in all the details and make it a regular tx, bypassing the scheduler. But that complicates the logic, not sure if it is really worth it. So I don't see how to combine save/update as you say with deletion of pending tx without this hackish way

services/wallet/api.go
IvanBelyakoff: All these long lines without wrapping really make me crazy. You can't open several files on a single screen without having to scroll the window to see the function signature. I think wrapping long lines is somewhat industry standard. Do you have a different opinion?  I think we could make some project linting setting to have a limit of characters for line.

services/wallet/transfer/transaction.go
IvanBelyakoff: Well, I don't see why we should do that. There are 2 DB operations logically unrelated. And they are separated by `Send` function call. They must happen in order and not proceed if any error previously happened. Able to store multitransaction? Then do send. Able to send? then store it as pending to be able to watch for it.

services/rpcfilters/transaction_sent_to_upstream_event.go
IvanBelyakoff: This is because of `NOT NULL` restriction in `pending_transactions` table

transactions/pending.go
stefandunca: > ... Though as we discussed it before, watching for pending transactions and downloading new ones are two different separate processes. ...

This is where I see issues by having these dependent events as independent. Maybe I missed some context, so let me explain my understanding, and hope you can clarify where I got it wrong.

Transaction lifetime:
1. When a user posts a transaction, the blockchain acknowledges it, and the transaction enters the `pending` state (added to the mempool and not processed).
2. transaction is `confirmed` when it is selected from the mempool to include in the next block
3. transaction is `successful` or `failed` when the block in which it was confirmed is added to the blockchain.

My understanding of requirements:
a. at step `1.` we should have the transaction in pending table 
b. at step `2.` and `3.` it should be in the transfers table
c. at any point, it should not be in both tables at the same time (pending and transfers)

My understanding of the current implementation is that because of the independent pending and transaction download events triggering saving and updating of the transaction in `transfers` and deleting the transaction from the `pending_transactions` in an independent manner, the following breaking of the requirement `c.` can happen:
- For short periods, have the same transaction in `pending` and `transfers` tables
- If, for some reason, the process is interrupted in between the two independent events, we might have a dangling entry in the pending transaction (I can't see where we reconcile having the same transaction in pending and transfers at the same time at boot)

Hence from all these considerations, the requirements enforce that upon downloading the transaction for the first time (confirmed or failed) we should save the transaction in the transfer table and remove it from the pending table simultaneously. Hence clients should have no chance of interfering with this transformation event (by stopping the process) and register for these consecutive events if they are interested. This way, we can simplify the triggering of `pending update` on deletion because it coincides with the `new transaction` event.

services/wallet/api.go
stefandunca: Now I understand, you don't use line wrap. I haven't considered in a long time the line limitation. I thought we don't have a line limitation in status-go and nobody complained. Let's discuss this particular requirement in our next meeting.

> You can't open several files on a single screen without having to scroll the window to see the function signature.

I personally prefer the automation of line wrap from `vscode` and company (with provide good line hints) and get more vertical lines of code instead.

Still I find the following formatting clearer than current approach
image

transactions/pending.go
IvanBelyakoff: @stefandunca you are right about what you have written. I am not arguing that having separate handling of pending transactions by transactions manager and new transactions by reactor (scheduler) is troublesome, we indeed need to integrate it.
What I missed though is that this needs to be done in scope of [#11139](https://github.com/status-im/status-desktop/issues/11139), as this one and [#10474](https://github.com/status-im/status-desktop/issues/10474) were intersecting for scope of works, so I was not going to fix it here in some hackish way, but there as it needs some thinking since pending transactions handling was moved to a separate manager, since it handles non-transfer transactions as well. I updated the scopes of both issues to be more precise.
Thank you for heads up!

PR #3705, opened 2023-07-03
feat: add cropping support for community token assets

0x-r4bbit:
This commit extends the `AddCommunityToken` API to also expect an optional `CroppedImage`, which will be used instead of the `ImageBase64` path provided by `CommunityToken`, to calculate the actual base64 encoded image.

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -4003,7 +4003,7 @@ func (m *Manager) ImageToBase64(uri string) string {
 	return base64img
 }
 
-func (m *Manager) AddCommunityToken(token *CommunityToken) (*CommunityToken, error) {
+func (m *Manager) AddCommunityToken(token *CommunityToken, croppedImage *images.CroppedImage) (*CommunityToken, error) {
 
 	community, err := m.GetByIDString(token.CommunityID)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


protocol/communities/manager.go
```diff
@@ -4003,7 +4003,7 @@ func (m *Manager) ImageToBase64(uri string) string {
 	return base64img
 }
 
-func (m *Manager) AddCommunityToken(token *CommunityToken) (*CommunityToken, error) {
+func (m *Manager) AddCommunityToken(token *CommunityToken, croppedImage *images.CroppedImage) (*CommunityToken, error) {
```
0x-r4bbit, line 4006: I was considering putting the `CroppedImage` into the already provided `CommunityToken`.

However, it turns out that `CommunityToken` doesn't actually use that data. It's only interested in the base64 encoded image data, which atm is first provided via `Base64Image` as image path, which is then overridden with an actual base64 string.

So to not pollute `CommunityToken` with unnecessary data, I think it's better to keep `CroppedImage` separate.
There's an argument to be made that one could rather put these in a `request` object on the messenger level.

Lemme know what you think.

protocol/communities/manager.go
```diff
@@ -4013,7 +4013,26 @@ func (m *Manager) AddCommunityToken(token *CommunityToken) (*CommunityToken, err
...
+			croppedImage.X,
+			croppedImage.Y,
+			croppedImage.X+croppedImage.Width,
+			croppedImage.Y+croppedImage.Height,
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		base64img, err := images.GetPayloadDataURI(identityImages[0].Payload)
```
0x-r4bbit: Just going for the `SmallSize` identity image here, as I think it's enough. Can go large too, if preferred.

mprakhov (approved):

protocol/communities/manager.go
endulab: There is also OpenAndAdjustImage function which takes CroppedImage and do some adjustments.

protocol/communities/manager.go
endulab: Can you elaborate on putting in a request object? Where could it be?

0x-r4bbit: So the idea would be the `Messenger` would receive a `requests.AddCommunityToken` object that has a `CommunityToken` and ` CroppedImage`, similar to how we do `request` objects in all other messenger APIs.

Manager would still receive separate arguments I think.

protocol/communities/manager.go
0x-r4bbit: Thanks! Will make use of that one instead.

endulab (approved):

PR #3706, opened 2023-07-03
Feat/10791 detect ETH bridge transactions

dlipicar:
Part of https://github.com/status-im/status-desktop/issues/10791

This PR adds support for parsing Hop L1->L2, L2->L1 and L2->L2 ETH bridge operations. ERC20 Token bridging support is left for a separate PR.

Screenshot 2023-06-30 at 4 43 01 PM
image

The first commit is just some code reorg to make supporting the new transaction type easier.
The second commit contains the logic to parse L1->L2 bridges.
The third commit adds support for L2->L1 and L2->L2 bridges.

Changes to our go-ethereum fork were needed: https://github.com/status-im/go-ethereum/commit/20a73bc473f16f74b706448f9fed54059049d02d, since some of the data is present in `DepositTx` type transactions, which we couldn't parse before.

Bridges consist of two transactions in different chains, which are tied together with some common parameter (which I call `cross_tx_id`). Either the Origin tx or the Destination tx could be processed first, it makes no difference, but you need both of them to get the full bridge info. The needed data is fetched from each transaction, which results in a full Multitransaction entry.

For L1->L2 bridges, the blocks containing both the L1 and L2 transactions were already marked for processing, since both cause a change in the account's balance. However, actually processing the transactions was a bit trickier:
- For the L1 transaction, there's no `Transfer` event but the `From` address is our account's, so we just need to add subTx processing in case it was not done before.
- For the L2 transactions, there's no `Transfer` event and neither the `From` or the `To` address matches our account's. Moreover, the transaction is of type `DepositTx` (Opt/Arb) or `RetryTx` (Arb). Since getting the receipt is an expensive process, we only do it for transactions where the data contains the account's address.

For L2->L1 and L2->L2 transactions, we need to detect event `WithdrawalBonded` which comes in a plain transaction type, so we end up checking those as well for any data field containing the account's address.


dlipicar:
@alaibe with regards to the go-ethereum changes, basically
image

I took the code mostly from Arbitrum and Optimism forks of go-ethereum and stuff seems to work. If the changes in that branch look good to you, how would I go about doing a proper release?

alaibe:
@dlipicar you have to create a PR in https://github.com/status-im/go-ethereum/ then create a new tag and finally update status-go  mod file

stefandunca (approved):
👍🏼

dlipicar:
`go-ethereum` dependency now points to a proper release tag

PR #3716, opened 2023-07-04
fix(wallet) returning wrong address identity in filter

stefandunca:
### Stumble upon it while implementing status-desktop [#11233](https://github.com/status-im/status-desktop/issues/11233)

This fixes returning address identity. The issue was introduced after changing the real to/from fix. Previously the address was wrongly used as from.


Cuteivist (approved):
Tested on specific case I had 100% reproducibility. Work like a charm.
Great job!

Khushboo-dev-cpp (approved):

PR #3721, opened 2023-07-05
feat(wallet) deliver pending changed events to clients

stefandunca:
### Updates status-desktop [#11233](https://github.com/status-im/status-desktop/issues/11233)

Add wallet events feed to TransactionManager and send pending changed events on add and delete
    
Also

- Remove hardcoded values in the filter query
- Small improvement in query

### DIFF
--- a/services/wallet/transfer/transaction.go
+++ b/services/wallet/transfer/transaction.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/event"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/status-im/status-go/account"
 	"github.com/status-im/status-go/eth-node/types"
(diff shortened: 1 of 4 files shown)


services/wallet/transfer/transaction.go
```diff
@@ -252,11 +260,30 @@ func (tm *TransactionManager) AddPending(transaction PendingTransaction) error {
...
+			Accounts: addresses,
+			At:       int64(timestamp),
+		})
+	}
+}
+
 func (tm *TransactionManager) DeletePending(chainID uint64, hash common.Hash) error {
 	_, err := tm.db.Exec(`DELETE FROM pending_transactions WHERE network_id = ? AND hash = ?`, chainID, hash)
+	if err == nil {
+		tm.notifyPendingTransactionListeners(chainID, nil, 0)
```
dlipicar: shouldn't we get the from/to addresses from the deleted transaction to put them in the notification?

services/wallet/transfer/transaction.go
```diff
@@ -21,31 +22,38 @@ import (
 	"github.com/status-im/status-go/services/wallet/bigint"
 	"github.com/status-im/status-go/services/wallet/bridge"
 	wallet_common "github.com/status-im/status-go/services/wallet/common"
+	"github.com/status-im/status-go/services/wallet/walletevent"
 	"github.com/status-im/status-go/transactions"
 )
 
+const (
+	// EventNewTransfers emitted when new block was added to the same canonical chan.
```
dlipicar: typo `canonical chan`

dlipicar (approved):

services/wallet/transfer/transaction.go
stefandunca: I intentionally left out this optimization given that this should be temporary and improve with the implementation of dynamic updates in the future.

stefandunca: I followed your advice in the end; thanks for the heads up. But also found that the to_address in pending is wrong. I will follow up on that issue.

PR #3722, opened 2023-07-05
Send account/settings on login

cammellos:
This PR adds sending account/settings in the login signal.

I had to change the way we signal database encryption because of a circular dependency (sqlite depended on signals, signals depends on high level stuff which in turns depends on sqlite). 
The change is to pass a couple of function onStart onEnd, which makes sense as sqlite is low level while signal is higher abstraction.


alexjba (approved):
LGTM!

flexsurfer (approved):

PR #3723, opened 2023-07-05
feat(community): add api to edit shared address

jrainville:
Needed for https://github.com/status-im/status-desktop/issues/11153

Adds an API for users to edit the wallet addresses they shared with an Owner of a community.

You need to be part of the community for it to work


protocol/communities/manager.go
```diff
@@ -1900,6 +1900,52 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request
...
+	}
+
+	err = m.persistence.SaveCommunity(community)
+	if err != nil {
+		return err
+	}
+
+	if community.IsOwner() {
+		m.publish(&Subscription{Community: community})
+	}
```
0x-r4bbit, line 1944: Will admins not handle these kind of requests?

protocol/messenger_communities.go
```diff
@@ -661,14 +661,70 @@ func (m *Messenger) SetMutePropertyOnChatsByCategory(request *requests.MuteCateg
 	return nil
 }
 
+// getAccountsToShare is used to geth the wallet accounts to share either when requesting to join a community or when editing
```
0x-r4bbit:
```suggestion
// getAccountsToShare is used to get the wallet accounts to share either when requesting to join a community or when editing
```

protocol/messenger_communities.go
```diff
@@ -661,14 +661,70 @@ func (m *Messenger) SetMutePropertyOnChatsByCategory(request *requests.MuteCateg
...
+
 func (m *Messenger) RequestToJoinCommunity(request *requests.RequestToJoinCommunity) (*MessengerResponse, error) {
 	logger := m.logger.Named("RequestToJoinCommunity")
 	if err := request.Validate(); err != nil {
 		logger.Debug("request failed to validate", zap.Error(err), zap.Any("request", request))
 		return nil, err
 	}
 
-	// verify wallet password if there
+	// verify wallet password is there
```
0x-r4bbit: That comment might as well be removed as it adds zero value really

protocol/messenger_communities.go
```diff
@@ -841,6 +897,90 @@ func (m *Messenger) RequestToJoinCommunity(request *requests.RequestToJoinCommun
...
+
+	community, err := m.communitiesManager.GetByID(request.CommunityID)
+	if err != nil {
+		return nil, err
+	}
+
+	// verify wallet password is there
+	if request.Password == "" {
+		return nil, errors.New("password is necessary to use this function")
+	}
```
0x-r4bbit: Few thoughts here:

- First, I think it's unexpected that this API expects a `requests.RequestToJoinCommunity`, maybe we should consider introducing a `requests.RequestToChangeRevealedAddresses` or something along those lines
- Given that `Password` is mandatory, we should probably check that one first things first and return early. Not reason to do any other work if the password isn't supplied.
- For 2) we could do this in a `Validate()` function of `RequestToChangeRevealedAddresses`

protocol/messenger_communities.go
```diff
@@ -841,6 +897,90 @@ func (m *Messenger) RequestToJoinCommunity(request *requests.RequestToJoinCommun
...
+	// signatures to request
+	revealedAccounts, revealedAddresses, err := m.getAccountsToShare(request, []byte{})
+	if err != nil {
+		return nil, err
+	}
+
+	checkPermissionResponse, err := m.communitiesManager.CheckPermissionToJoin(community.ID(), revealedAddresses)
+	if err != nil {
+		return nil, err
+	}
```
0x-r4bbit, line 911: If I'm not mistaken, this API will only be call if one is already part of that community. 
Typically we call `CheckPermissionToJoin()` on the sender side to ensure the UI prevents users from entering a community if they don't satisfy the permissions.

Given that this is not the case here, I'm wondering if we need to do this check in the first place, or if we just want to send the addresses regardless.

0x-r4bbit: Ah, on a second thought... I think it's good we do it, also to offload some work.
Ignore the comment above

protocol/protobuf/communities.proto
```diff
@@ -136,6 +136,11 @@ message CommunityRequestToJoin {
   repeated RevealedAccount revealed_accounts = 6;
 }
 
+message CommunityEditRevealedAccounts {
+  bytes community_id = 4;
+  repeated RevealedAccount revealed_accounts = 6;
```
0x-r4bbit: Don't we want  clock here?

0x-r4bbit: Also, shouldn't the fields start with `1` (it's not enforced but probably more consistant)

protocol/messenger_communities.go
```diff
@@ -841,6 +897,90 @@ func (m *Messenger) RequestToJoinCommunity(request *requests.RequestToJoinCommun
...
+		}
+	}
+
+	requestToEditRevealedAccountsProto := &protobuf.CommunityEditRevealedAccounts{
+		CommunityId:      community.ID(),
+		RevealedAccounts: make([]*protobuf.RevealedAccount, 0),
+	}
+	// find wallet accounts and attach wallet addresses and
+	// signatures to request
+	revealedAccounts, revealedAddresses, err := m.getAccountsToShare(request, []byte{})
```
0x-r4bbit: Should we make use of that functoin in `RequestToJoinCommunity` as well?

protocol/communities/manager.go
jrainville: Maybe I misunderstood this code. I think only Owners can publish an updated community description? The admins will still process the change and save it locally, but only the owner can propagate the change. Make me wonder if maybe we actually need to publish it. Normal users don't actually need to know the shared addresses

protocol/messenger_communities.go
jrainville: > First, I think it's unexpected that this API expects a requests.RequestToJoinCommunity, maybe we should consider introducing a requests.RequestToChangeRevealedAddresses or something along those lines

Good point. I forgot to ask that question in the PR description.
I reused the same Object because it's basically the same props. It simplifies the code a lot and makes it so I can reuse `getAccountsToShare` for both function (which you correctly identified that I forgot to do :facepalm: )

Anyway, I was gonna propose that maybe I could rename `RequestToJoinCommunity` to something else so that it makes sense that it's used by multiple parts. Maybe `RequestToJoinCommunityOrEditSharedAddresses`. Doesn't exactly roll off the tongue though.

protocol/protobuf/communities.proto
jrainville: I didn't add a clock because I didn't think it might need it since you don't clash with anyone else when doing this call, but you could clash with yourself if you send multiple ones, so I'll add one.

As for the field numbers, that's the perils of copy paste lol. I'll fix.

protocol/communities/manager.go
0x-r4bbit: > Normal users don't actually need to know the shared addresses

Admins will receive those updates the same way, so they should probably be notified to help offloading the onchain checks

protocol/messenger_communities.go
0x-r4bbit: Another way would be to have `EditSharedAddresses` to inherit `RequestToJoinCommunity` (or vice versa is probably better).

protocol/communities/manager.go
mprakhov: via `m.publish(&Subscription{Community: community})` only owner can send a CommunityDescription as it requires a community private key for signature. If we want admins to do this job too - we need a separate admin event (it should be done in a separate ticket)

protocol/communities_messenger_token_permissions_test.go
```diff
@@ -445,6 +445,97 @@ func (s *MessengerCommunitiesTokenPermissionsSuite) TestJoinedCommunityMembersSe
...
+	community, err = s.owner.GetCommunityByID(community.ID())
+	s.Require().NoError(err)
+
+	for pubKey, member := range community.Members() {
+		if pubKey != common.PubkeyToHex(&s.owner.identity.PublicKey) {
+			s.Require().Len(member.RevealedAccounts, 1)
+
+			switch pubKey {
+			case common.PubkeyToHex(&s.alice.identity.PublicKey):
+				s.Require().Equal(member.RevealedAccounts[0].Address, aliceAddress1)
```
jrainville: I found that the test works 90% of the time, but sometimes, it fails here. It seems to fail because the Owner didn't retrieve the message to edit the community. How can I fix that? cc @0x-r4bbit @mprakhov

mprakhov: Hard to say without debugging :) 

I would suggest you check that you received the correct response first here https://github.com/status-im/status-go/pull/3723/files#diff-19758c8021697444e5216815e7a24d0537cd971a0018fc4b0116517b8930b961R479-R489 (do the check of your member in the response). If it is present in the response, but not in the db (in this for loop you validate community extracted from the DB), then it could be issue in the saving/restoring from the DB

Also, the problem can be that before you get your edit event - you are receiving some other event and handling this event instead of your edit event

mprakhov: Also, check `WaitOnMessengerResponse` fn, maybe it makes sense to reuse it as it uses merge

protocol/messenger_communities.go
jrainville: I tried the inheritance, but turns out Go doesn't do polymorphism (at least not easily), so I just copied the Class and pass the individual properties to the `getAccountsToShare` function instead of the object. It works fine.

protocol/communities_messenger_token_permissions_test.go
jrainville: Good tips, I'll try

jrainville: I managed to fix with your tip in the first comment. Thanks @mprakhov . 
I think you're right that another event came in first.

jrainville:
You can review. The PR is ready and fixed @0x-r4bbit @mprakhov

protocol/communities_messenger_token_permissions_test.go
```diff
@@ -445,6 +445,110 @@ func (s *MessengerCommunitiesTokenPermissionsSuite) TestJoinedCommunityMembersSe
...
+				case common.PubkeyToHex(&s.alice.identity.PublicKey):
+					if member.RevealedAccounts[0].Address != aliceAddress1 {
+						return errors.New("address not yet updated")
+					}
+				default:
+					s.Require().Fail("pubKey does not match expected keys")
+				}
+			}
+		}
+		return nil
```
mprakhov: minor: this code repeated 3 times, can be moved to lambda and reused where we need it

protocol/communities/community.go
```diff
@@ -685,6 +685,10 @@ func (o *Community) getMember(pk *ecdsa.PublicKey) *protobuf.CommunityMember {
 	return member
 }
 
+func (o *Community) GetMember(pk *ecdsa.PublicKey) *protobuf.CommunityMember {
+	return o.getMember(pk)
+}
```
mprakhov, line 690: not sure that it is needed, I guess you can call o.getMember(pk) where you need this fn

mprakhov (approved):
LGTM, few minor issues

protocol/communities/community.go
jrainville: I can't use `getMember` in messenger_communities because lower case methods are private.

PR #3726, opened 2023-07-06
chore: synchronization improvements applied to keycards

saledjenic:
This is the second step of improvements over keypairs/keycards/accounts (changes made in the first step are here https://github.com/status-im/status-go/pull/3698).

- `SyncKeycardAction` protobuf removed
- `SyncKeypair` protobuf is used for syncing keycards state as well as for all
keycards related changes
- `last_update_clock` column removed from `keypairs` table cause as well as
for accounts, any keycard related change is actually a change made on a related
keypair, thus a keypair's clock keeps the clock of the last change
- `position` column added to `keypairs` table, needed to display keycards in
the same order across devices

Changed endpoints:
- `AddKeycardOrAddAccountsIfKeycardIsAdded` changed to `SaveOrUpdateKeycard`
- `RemoveMigratedAccountsForKeycard` changed to `DeleteKeycardAccounts`
- `GetKeycardByKeyUID` changed to `GetKeycardsWithSameKeyUID`
- `GetAllKnownKeycardsGroupedByKeyUID` removed
- `GetKeycardByKeycardUID` added


alaibe (approved):

multiaccounts/accounts/database.go
```diff
@@ -902,6 +894,7 @@ func (db *Database) SaveOrUpdateAccounts(accounts []*Account, updateKeypairClock
 	return err
 }
 
+// This function doesn't save/update keycards.
```
rasom: Although it probably makes sense in terms of this PR to have a comment like this, I would prefer to state the obvious at first and write what it actually does (just to emphasise why mentioning of what it doesn't do is important) because this function doesn't do an infinite number of things :)

Definitely not a deal breaker

rasom (approved):

multiaccounts/accounts/database.go
saledjenic: Comment updated. ;)

PR #3728, opened 2023-07-07
Refactor/community naming cleanup

osmaczko:
- use `IsOwner` instead of `config.privateKey != nil`
- improve community functions naming

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -517,8 +517,8 @@ func (m *Manager) DeletedCommunities() ([]*Community, error) {
 	return m.persistence.DeletedCommunities(&m.identity.PublicKey)
 }
 
-func (m *Manager) Created() ([]*Community, error) {
-	return m.persistence.CreatedCommunities(&m.identity.PublicKey)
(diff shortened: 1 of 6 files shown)


0x-r4bbit (approved):
Nice cleanup, thank you

mprakhov (approved):
LGTM

protocol/communities/community.go
```diff
@@ -1409,24 +1409,24 @@ func (o *Community) MarshaledDescription() ([]byte, error) {
...
+		}
 
-	// This should not happen, as we can only serialize on our side if we
-	// created the community
-	if o.config.PrivateKey == nil && len(o.config.MarshaledCommunityDescription) == 0 {
-		return nil, ErrNotAdmin
-	}
-
-	//We are not owner or admin, use the received serialized version
-	if !o.IsOwnerOrAdmin() {
```
mprakhov: admins should also be available to marshal description

osmaczko: Admins don't have private key so how can they sign the description?

mprakhov: signing will be skipped, I think it is ok. At least if key is empty, we do not return an error

PR #3730, opened 2023-07-07
feat: rehydrate community_tokens table upon account recovery

### LINKED ISSUE #11188: Fill community_tokens table when restoring account
After restoring account the community owner should see his deployed contracts.
We need to check all community tokens info from CommunityDescription and fill the community_tokens table.

0x-r4bbit:
Closes https://github.com/status-im/status-desktop/issues/11188

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -25,6 +25,7 @@ import (
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
 
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/status-im/status-go/account"
@@ -33,11 +34,13 @@ import (
 	"github.com/status-im/status-go/images"
 	"github.com/status-im/status-go/params"
 	"github.com/status-im/status-go/protocol/common"
+	community_token "github.com/status-im/status-go/protocol/communities/token"
 	"github.com/status-im/status-go/protocol/encryption"
 	"github.com/status-im/status-go/protocol/ens"
 	"github.com/status-im/status-go/protocol/protobuf"
 	"github.com/status-im/status-go/protocol/requests"
 	"github.com/status-im/status-go/protocol/transport"
+	"github.com/status-im/status-go/services/collectibles"
 	"github.com/status-im/status-go/services/wallet/bigint"
 	walletcommon "github.com/status-im/status-go/services/wallet/common"
 	"github.com/status-im/status-go/services/wallet/thirdparty"
@@ -78,6 +81,7 @@ type Manager struct {
 	torrentConfig                    *params.TorrentConfig
 	torrentClient                    *torrent.Client
 	walletConfig                     *params.WalletConfig
+	collectiblesService              *collectibles.Service
 	historyArchiveTasksWaitGroup     sync.WaitGroup
 	historyArchiveTasks              sync.Map // stores `chan struct{}`
 	periodicMembersReevaluationTasks sync.Map // stores `chan struct{}`
@@ -112,6 +116,7 @@ type managerOptions struct {
 	tokenManager        TokenManager
 	collectiblesManager CollectiblesManager
 	walletConfig        *params.WalletConfig
+	collectiblesService *collectibles.Service
 }
 
 type TokenManager interface {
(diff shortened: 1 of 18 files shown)


0x-r4bbit:
Putting this up for early feedback @endulab 

Also, there's an issue with the linter that it's unhappy about certain imports, even though I didn't make any relevant changes there.
Anyone an idea?

jrainville:
> Putting this up for early feedback @endulab
> 
> Also, there's an issue with the linter that it's unhappy about certain imports, even though I didn't make any relevant changes there. Anyone an idea?

I think there might be a cyclic import in the `collectibles/database_test.go`. You import the collectibles service in the communities files, but that test imports `"github.com/status-im/status-go/protocol/communities"`

protocol/communities/manager.go
```diff
@@ -4058,3 +4073,62 @@ func (m *Manager) saveAndPublish(community *Community) error {
...
+
+func (m *Manager) HandleCommunityTokensMetadata(communityID string, communityTokens []*protobuf.CommunityTokenMetadata) error {
+	for _, tokenMetadata := range communityTokens {
+		for chainID, address := range tokenMetadata.ContractAddresses {
+			exists, err := m.persistence.HasCommunityToken(communityID, address, int(chainID))
+			if err != nil {
+				return err
+			}
+			if !exists {
+				contract, err := m.collectiblesService.API().GetContractInstance(chainID, address)
```
0x-r4bbit: Need to check for token type here to get correct contract instance

protocol/messenger_communities.go
```diff
@@ -2446,6 +2447,20 @@ func (m *Messenger) handleSyncCommunity(messageState *ReceivedMessageState, sync
...
 	}
 
+	id := crypto.CompressPubkey(orgPubKey)
+	savedCommunity, err := m.communitiesManager.GetByID(id)
+	if err != nil {
+		return err
+	}
+
+	if savedCommunity.HasPermissionToSendCommunityEvents() || savedCommunity.IsControlNode() {
+		err := m.handleCommunityTokensMetadata(savedCommunity.IDString(), savedCommunity.CommunityTokensMetadata())
```
jrainville, line 2457: This function might be slow because it calls the chain sometimes, but `handleSyncCommunity` is already handled in the background, so I guess it's not an issue?

I'm just asking to be sure all is good in terms of responsiveness and timeliness.

jrainville (approved):
Looks good.

I wonder if we could add a test for `HandleCommunityTokensMetadata`? Is mocking the contracts already there?

endulab:
LGTM.
Just FYI, In my current pr I added "deployer address" field to the database.
I am thinking about adding transaction hash to CommunityTokenMetadata, because w/o this
finding deployer address would be very expensive (run through all transactions).

endulab (approved):

0x-r4bbit:
> Just FYI, In my current pr I added "deployer address" field to the database.

@endulab if you do this, then is there still a need for adding the tx hash for the deployment to the database? Or would you need tx data for anything else after deployment was done.

Generally not opposing, probably good to have that data around for inspection later on.

protocol/messenger_communities.go
0x-r4bbit: > This function might be slow because it calls the chain sometimes, but handleSyncCommunity is already handled in the background, so I guess it's not an issue?

Not sure if there's so much we can do about that. We will have to reach out to chain regardless, but in reality I think it's not going to be that often anyways.

Also, this is part of recovering an account which already has a 5(?)min timer. So it will be part of that. Sync at runtime indeed happens in the background.

0x-r4bbit:
> I wonder if we could add a test for HandleCommunityTokensMetadata? Is mocking the contracts already there?

@endulab do you know by any chance?

endulab:
> > Just FYI, In my current pr I added "deployer address" field to the database.
> 
> @endulab if you do this, then is there still a need for adding the tx hash for the deployment to the database? Or would you need tx data for anything else after deployment was done.
> 
> Generally not opposing, probably good to have that data around for inspection later on.

No, I do not need tx hash in the database, only in CommunityTokenMetadata. However, we have a map <chain,address> to handle potential multichain contracts. Probably I need to have a map <chain, hash>. @0x-r4bbit?

endulab:
> > I wonder if we could add a test for HandleCommunityTokensMetadata? Is mocking the contracts already there?
> 
> @endulab do you know by any chance?

We do not mock/test contracts now in status-go.

PR #3732, opened 2023-07-09
[status-mobile-16467] Fix delete for me on receiver side using wrong chatID

ibrkhalil:
Just tiding up the code and adding test for an edge case (Found by invalid input)
Closes #16467 on status-mobile


cammellos (changes requested):
I think this will likely break things, since `albumMessages` is used in 3 places at least, did you verify it's not breaking anything? (unfortunately those places are not covered by test it looks like)
Also this should come with a test

cammellos (approved):

borismelnik (approved):
👍

ilmotta (approved):

PR #3738, opened 2023-07-11
fix: fix returning correctly when the password is incorrect

jrainville:
This PR (https://github.com/status-im/status-go/pull/3722) introduced a small bug where entering an incorrect password would make the login process just spin forever (not returning).

I don't know the logic super well @cammellos so please check it out (it's pretty small)


cammellos:
@jrainville my bad, I am a bit confused by the code, that's why I changed it, I should have been more careful, good catch

cammellos (approved):

PR #3739, opened 2023-07-11
fix(wallet) fix scan of default NULL multi_transactions.*_network_id

stefandunca:
The [migration setup](https://github.com/status-im/status-go/blob/3ce851b84b6a54fc2ea2098ad6fb8cedea45cb4a/appdatabase/migrations/sql/1688054680_add_columns_to_multitransaction.up.sql#L1-L2) left the default value to null, hence old entries have null that can't be scanned in `unit64` values.


dlipicar (approved):

dlipicar:
Oops... We could give it default value 0, I think that's not a valid chainID

IvanBelyakoff (approved):

alaibe (approved):

PR #3746, opened 2023-07-12
fix: ensure correct initialization order for wallet and mesenger services

dlipicar:
We've got a circular dependency between Messenger and Wallet.

```
	// CollectiblesManager needs the WakuExt service to get metadata for
	// Community collectibles.
	// Messenger needs the CollectiblesManager to get the list of collectibles owned
	// by a certain account and check community entry permissions.
```

There's probably a more proper way of handling this, but for now we simply inject the NFTMetadataProvider (WakuExt) instance to the WalletService after both objects have been instanced.


jrainville (approved):

dlipicar:
The ios build is for some reason failing at the upload stage (timeout). Since everything else passes, I'm merging this anyway.

PR #3747, opened 2023-07-12
Feat/multi chain collectible ownership provider

dlipicar:
The OpenSea V1 API only provides info for the Ethereum Main/Test chains (no Arbitrum and Optimism). This caused membership requests for Optimism/Arbitrum NFT-gated communities not to be sent, due to the checks not passing, even when the user owned the needed collectibles on those chains.

For now, we use Infura+Alchemy endpoints we already support to supplement OpenSea in the new method `FetchBalancesByOwnerAndContractAddress`. This is a bit inefficient, since those endpoints return the full list of owners for a given contract instead of just the wallet address we're interested in. Finding and using a more proper endpoint is left for a future PR.

We also now use the Wallet Service instead of a new instance of the OpenSea client in the Community manager.

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -27,7 +27,6 @@ import (
 
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
-	"github.com/ethereum/go-ethereum/event"
 	"github.com/status-im/status-go/account"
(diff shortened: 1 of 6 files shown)


mprakhov (approved):
LGTM

IvanBelyakoff (approved):
LGTM

0x-r4bbit (approved):
Good catch and very nice changes!

PR #3748, opened 2023-07-12
fix: only save/send request to join when the permissions are satisfied

jrainville:
There were two problems. 

1. We were saving the request to join to the user's DB before the permissions were checked, so if those errored or were not satisfied, we still saved it, creating a weird scenario where the user would think the request was sent, but it wasn't.
2. We still sent the request to join even if the user didn't satisfy the permission to join. In theory, the UI blocks that behaviour, but the backend should enforce it. 

I upgraded the "join with permission" test to validate those behaviours above.

### DIFF
--- a/protocol/communities_messenger_token_permissions_test.go
+++ b/protocol/communities_messenger_token_permissions_test.go
@@ -578,6 +578,15 @@ func (s *MessengerCommunitiesTokenPermissionsSuite) TestBecomeMemberPermissions(
 	err = <-waitOnCommunityEncryptionErrCh
 	s.Require().NoError(err)
 
+	_, err = WaitOnMessengerResponse(
+		s.bob,
+		func(r *MessengerResponse) bool {
+			return len(r.Communities()) > 0
+		},
+		"no community",
+	)
+	s.Require().NoError(err)
+
 	// bob should be kicked from the community,
 	// because he doesn't meet the criteria
 	community, err = s.owner.communitiesManager.GetByID(community.ID())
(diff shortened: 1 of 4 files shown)


borismelnik (approved):

protocol/communities_messenger_token_permissions_test.go
```diff
@@ -578,6 +578,11 @@ func (s *MessengerCommunitiesTokenPermissionsSuite) TestBecomeMemberPermissions(
 	err = <-waitOnCommunityEncryptionErrCh
 	s.Require().NoError(err)
 
+	// Make sure Bob gets the new community with the new permission
+	community, err = s.owner.GetCommunityByID(community.ID())
+	s.Require().NoError(err)
+	s.advertiseCommunityTo(community, s.bob)
```
osmaczko: No need to advertise community explicitly, `CreateCommunityTokenPermission` should do that already.
```suggestion
	_, err := WaitOnMessengerResponse(
		s.bob,
		func(r *MessengerResponse) bool {
			return len(r.Communities()) > 0
		},
		"no community",
	)
	s.Require().NoError(err)
```

osmaczko (approved):
Nice catch :+1:

protocol/communities_messenger_token_permissions_test.go
jrainville: I think I tried that, but it didn't work, but that was when I was writing the test, so I assume I had still issues in the code itself.

Now it works fine. Updated. Thanks.

Issue #3423, opened 2023-04-25
Add local pairing start/end events


igor-sirotin:
# Problem

In desktop I have a `LocalPairingState` enum. This makes it easier to implement the UI.

https://github.com/status-im/status-desktop/blob/9361584027030529e8e3b8284e1ec2a7b3f80308/src/app_service/service/devices/dto/local_pairing_status.nim#L6-L11

Because we don't really have such state in status-go (and I don't think we need to), I have to guess the state from incoming local-pairing-events. Here comes 2 problems:

1. It's not reliable.
The logic should come from status-go. It shouldn't be in the app. Even worse if it's duplicated on mobile too.

2. It's not quite obvious. 😐

https://github.com/status-im/status-desktop/blob/9361584027030529e8e3b8284e1ec2a7b3f80308/src/app_service/service/devices/dto/local_pairing_status.nim#L47-L79

# Implementation

I think it would be great to add `EventPairingStart` and `EventPairingEnd`.
It should be easy to implement with clients with `defer`. But I wasn't sure how to make it with server's handlers system. So I didn't do it myself 🙂 

@Samyoul wdyt of this?

Samyoul:
Hey @igor-sirotin yes, this seems a reasonable feature to have, particularly as the life cycle of almost all other phases of local pairing are communicated via signals.

The device that receives data will know when the process has been completed. The sender will require some additional information from the receiver, but this is not difficult to perform.

Issue #3306, opened 2023-03-16
Investigate unlinking `pairing` from `api.GethStatusBackend` and `Messenger`


Samyoul:
Investigate and implement separation between the dependency link between `api.GethStatusBackend` and `Messenger`.
If this is reasonable, we could use new pairing interfaces that Messenger implements to break the direct link between the `protocol`, `api` and `pairing` packages.

This will simplify the code and reduce future chances of import loops.

I'll add more details, for some background see https://github.com/status-im/status-go/pull/3248

PR #3750, opened 2023-07-12
Improve Local Pairing Idempotence

### LINKED ISSUE #16548: Error is not displayed as per design when scanning previously used Sync QR code
It might be nice to fix it together with https://github.com/status-im/status-mobile/issues/15945

### Steps to reproduce:
1. `User_A` Generate a valid sync QR code.
2. `User_B` Scan the sync QR code
3. `User_C` Scan the same sync QR code

### Actual result:
When scanning a previously used sync QR code, 'Oops something went wrong' page is shown

### Expected result:
When scanning a previously used sync QR code, toast with error message should be displayed
https://www.figma.com/file/o4qG1bnFyuyFOvHQVGgeFY/Onboarding-for-Mobile?type=design&node-id=4477-328513&mode=design&t=qEoB2LPz5HkB3QXX-0

### Logs:
[geth.log](https://github.com/status-im/status-mobile/files/12003329/geth.log)
[Status.log](https://github.com/status-im/status-mobile/files/12003330/Status.log)

### ENV:
[Nightly 10 Jul 2023]

Samyoul:
## What's Changed?

I've added a `pairing.ProcessStateManager` that is used in the backend to track the state of any ongoing. 2 main things are tracked.

- Is the application currently attempting to pair locally?
- Has the given QR code `ConnectionParams` string been successfully used to pair previously?
- If the answer to both of these questions is "No" then proceed with local pairing.

## Why Change?

Local pairing API functions are quite fragile when being fuzzed, this fix attempts to resolve a few issues:

- Prevent a client device from pairing with more than one device at a time
- Prevent a client device from attempting to re-pair / re-sync using a previously successful QR code `ConnectionParams` strings.
- Allows for the reuse of QR code `ConnectionParams` strings that are still valid and were not previously successful.

Possibly resolves https://github.com/status-im/status-mobile/issues/16548


ibrkhalil (approved):

saledjenic (approved):

server/pairing/statecontrol/state_management.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	psm.SetPairing(true)
+	return nil
+}
+
+// StopPairing takes a sessionName and an error, if the error is nil the sessionName will be recorded as successful
+// the pairing state of the ProcessStateManager is set to false.
+func (psm *ProcessStateManager) StopPairing(sessionName string, err error) {
+	if err == nil {
```
igor-sirotin, line 89: So when a session finishes with an error, we don't consider it as completed? 🤔 
Is it always safe to reuse the client/server that failed to pair with an error?

api/geth_backend.go
```diff
@@ -113,7 +113,8 @@ func (b *GethStatusBackend) initialize() {
 	b.personalAPI = personalAPI
 	b.statusNode.SetMultiaccountsDB(b.multiaccountsDB)
 	b.log = log.New("package", "status-go/api.GethStatusBackend")
-	b.localPairing = false
+	b.localPairingManager = new(statecontrol.ProcessStateManager)
```
igor-sirotin: Sounds like `localPairingStateManager` to me, wdyt?

Though having a `LocalPairingManager` I think is a good idea. I'will explain it below in comments.

mobile/status.go
```diff
@@ -1046,9 +1045,9 @@ func GetConnectionStringForBeingBootstrapped(configJSON string) string {
 		return makeJSONResponse(fmt.Errorf("no config given, PayloadSourceConfig is expected"))
 	}
 
-	statusBackend.SetLocalPairing(true)
+	statusBackend.LPM().SetPairing(true)
```
igor-sirotin: Now it looks like we save some state in `PairingStateManager` and have the actual state of the pairing Client/Server. And these states are synced only inside these top-level functions in `mobile/status.go`.

I think that now we could create a top-level class `ProcessStateManager` (I'd call it `LocalPairer`, but I don't think there's such a word 😄) and incapsulate this logic in it.

```go
type LocalPairingManager struct {
    backend          *api.GethStatusBackend
    stateManager *pairing.ProcessStateManager
}

func (m *LocalPairingManager) GetConnectionStringForBeingBootstrapped(configJSON) string {
	m.stateManager.SetPairing(true)
	defer func() {
		m.stateManager.SetPairing(false)
	}()
	cs, err := pairing.StartUpReceiverServer(m.backend, configJSON)
	if err != nil {
		return makeJSONResponse(err)
	}
	err = m.backend.Logout()
	if err != nil {
		return makeJSONResponse(err)
	}
	return cs
}
```

We could create `ProcessStateManager` in geth backend, and in `mobile/status.go` we would just call this function. Same for other 3 functions.

Having this done, we would be able to create reliable dependencies between `Server` and `StateManager`. For example, change state with this server timeout callback.
https://github.com/status-im/status-go/blob/ff0628c23b12166e21d95b58d9a7abf696fd902a/server/server.go#L79-L84

_____

This would allow us to:
- Show errors like `This sync code has expired`.
- https://github.com/status-im/status-go/issues/3423

mobile/status.go
```diff
@@ -1046,9 +1045,9 @@ func GetConnectionStringForBeingBootstrapped(configJSON string) string {
...
 	defer func() {
-		statusBackend.SetLocalPairing(false)
+		statusBackend.LPM().SetPairing(false)
```
igor-sirotin: Just thought about it: does this make sense here?
We don't actually start the pairing here, only connection string is created. And it's definitely not finished when we get out of this function.

igor-sirotin (approved):
Cool, thank you! 👍 
Left some minor questions and a proposal in comments

mobile/status.go
Samyoul: Hmmm, you are right. This whole API function is actually in a race condition that it loses almost every time.

Hmmm, I have an issue related to decoupling the backend dependency from the `pairing` package https://github.com/status-im/status-go/issues/3306, which should be kept in mind. However `Receiver`s should probably be given a callback function for when the completion state is met. I'll raise an issue for this.

Samyoul: https://github.com/status-im/status-go/issues/3752

server/pairing/statecontrol/state_management.go
Samyoul: Well we don't know what the error is, we can be very strict and prevent ANY reuse of a QR code if we want to. My approach with this functionality at least gives the user and the application the option to try again without having to regenerate a new QR code.

The failure could be anything from a flaky network connection to a fundamental bug in the code. This path just gives the user a bit more flexibility on what their next step should be after receiving the failure error. I'm open to suggestion on this, I made a best guess as to what the most helpful UX should be.

What do you think @igor-sirotin ?

api/geth_backend.go
Samyoul: Yes I agree I'll make this change

---

Done

mobile/status.go
saledjenic: > Show errors like `This sync code has expired`

We have countdown screen (displaying to a user when the connection is going to expire) on the client side, if we move that timer to status-go (which is good) then we need to emit signal from status-go updating client how many seconds remain till the expiration, cause otherwise is bad to have 2 counters for the same thing especially that their counters may differ.

Samyoul: Yes, I agree and `LocalPairer` sounds like a fine name to me. In fact I was mulling this very idea over as I was implementing this functionality. I can also make an issue for this, as state management of Local pairing and local pairing management should be coupled.

- [x] Made the issue

server/pairing/statecontrol/state_management.go
igor-sirotin: I agree, it should probably be fine to reuse the code. 
We can see later if this leads to some strange behaviour 🙂

mobile/status.go
igor-sirotin: > We have countdown screen on the client side

@saledjenic you're talking about sender-side. I was talking about the receiver, if someone scans an expired code, the server would return an "Sync code expired" error. 

Though now I think this might be impossible, because the server shuts down after timeout, so you won't be able to connect to a server with expired connection string.

Samyoul: @saledjenic sorry I missed your comment, for some reason it wasn't there.

So I agree with what you are saying, we need to keep the timeframe of the pairing session in sync between the sender device and the receiver device. I have an idea that I was thinking about yesterday to give the user some meaningful feedback in the case the QR code is too old, but it will help in this case also.

We can add to the `ConnectionParams` string a new parameter set by the device that creates the QR code, this will be `expiresAt` value that allows all other parts of the application on both devices to synchronise against. That means you can have an internal clock, a UI clock, set timeouts and certificate expiry all using a single common value. I'll raise an issue for this.

- [x] Made the issue

Samyoul: `LocalPairer` issue https://github.com/status-im/status-go/issues/3758

Samyoul: `expiresAt` `ConnectParams` string issue https://github.com/status-im/status-go/issues/3759

api/geth_backend.go
```diff
@@ -2106,6 +2107,6 @@ func (b *GethStatusBackend) SwitchFleet(fleet string, conf *params.NodeConfig) e
 	return nil
 }
 
-func (b *GethStatusBackend) SetLocalPairing(value bool) {
-	b.localPairing = value
+func (b *GethStatusBackend) LPSM() *statecontrol.ProcessStateManager {
```
qfrank: Do you think replace `LPSM` to `LocalPairingStateManager` or `GetLocalPairingStateManager` looks better? LPSM looks too short for me ... 😥 @Samyoul

qfrank (approved):
Nice new concept `ProcessStateManager` 👍

api/geth_backend.go
Samyoul: Ok, Just for you @qfrank I will replace this function with an exported field on the `GethStatusBackend` struct called `LocalPairingStateManager`

Samyoul: This is done @qfrank

PR #3751, opened 2023-07-13
Set timeout on shutdown

cammellos:
Logout takes a long time in some cases https://github.com/status-im/status-mobile/issues/16557 
that's due to filters request being synchronous. 
On logout we are not particularly interested in waiting for a response, so a short context is passed over.


igor-sirotin:
Closing as currently out of scope

PR #3754, opened 2023-07-13
Fix: Ignore contact requests from mutual and declined contacts

MishkaRogachev:
For https://github.com/status-im/status-desktop/issues/11483

Because of contact request message can be lost in network, we can not limit client code to send a contact request to a mutual contact. Instead, we are ignoring contact request message on the receiver side.

This behaviour is need by SpiffWorkflow bot [here](https://github.com/status-im/spiffworkflow-connector/blob/master/connectors/connector-waku/connector_waku/commands/sendMessage.py#L104).

### DIFF
--- a/protocol/messenger_contact_requests_test.go
+++ b/protocol/messenger_contact_requests_test.go
@@ -403,6 +403,62 @@ func (s *MessengerContactRequestSuite) TestReceiveAcceptAndRetractContactRequest
 	s.retractContactRequest(contactID, theirMessenger)
 }
 
+// The scenario tested is as follow:
+// 1) Alice sends a contact request to Bob
+// 2) Bob declines the contact request
+// 3) Alice fails to send a new contact request to Bob
+func (s *MessengerContactRequestSuite) TestAliceTriesToSpamBobWithContactRequests() {
+	messageTextAlice := "You wanna play with fire, Bobby?!"
+	alice := s.m
+
+	bob := s.newMessenger()
+	_, err := bob.Start()
+	s.Require().NoError(err)
+	defer bob.Shutdown() // nolint: errcheck
+
+	bobID := types.EncodeHex(crypto.FromECDSAPub(&bob.identity.PublicKey))
+
+	// Alice sends a contact request to Bob
+	request := &requests.SendContactRequest{
+		ID:      bobID,
+		Message: messageTextAlice,
+	}
+	s.sendContactRequest(request, alice)
+
+	contactRequest := s.receiveContactRequest(messageTextAlice, bob)
+	s.Require().NotNil(contactRequest)
+
+	// Bob declines the contact request
+	s.declineContactRequest(contactRequest, bob)
+
+	// Alice sends a new contact request
+	resp, err := alice.SendContactRequest(context.Background(), request)
+	s.Require().NoError(err)
+	s.Require().NotNil(resp)
+
+	// Check CR and mutual state update messages
+	s.Require().Len(resp.Messages(), 2)
+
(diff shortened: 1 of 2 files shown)


protocol/messenger_contact_requests_test.go
```diff
@@ -403,6 +403,61 @@ func (s *MessengerContactRequestSuite) TestReceiveAcceptAndRetractContactRequest
...
+	// Check CR and mutual state update messages
+	s.Require().Len(resp.Messages(), 2)
+
+	contactRequest = s.findFirstByContentType(resp.Messages(), protobuf.ChatMessage_CONTACT_REQUEST)
+	s.Require().NotNil(contactRequest)
+
+	s.Require().Equal(common.ContactRequestStatePending, contactRequest.ContactRequestState)
+	s.Require().Equal(request.Message, contactRequest.Text)
+
+	// We should not receive a CR from a mutual contact
```
osmaczko: Bob and Alice are not mutual contacts as Bob declined the contact request, no?

protocol/messenger_contact_requests_test.go
```diff
@@ -403,6 +403,61 @@ func (s *MessengerContactRequestSuite) TestReceiveAcceptAndRetractContactRequest
...
+	_, err = WaitOnMessengerResponse(
+		bob,
+		func(r *MessengerResponse) bool {
+			return len(r.Messages()) > 0
```
osmaczko: Would be better to check for the exact type of message otherwise this test may be flaky if different message comes in (e.g. contact code advertisement).

osmaczko:
> Because of contact request message can be lost in network, we can not limit client code to send a contact request to a mutual contact. Instead, we are ignoring contact request message on the receiver side.

Not sure I understand this case. If contact request message is lost in the network than client can't be a mutual contact. In other words, if they are mutual contacts, it means no contact request was lost in the network :thinking:

protocol/messenger_contact_requests_test.go
MishkaRogachev: Fixed

protocol/messenger_contact_requests_test.go
MishkaRogachev: Fixed

MishkaRogachev:
> Not sure I understand this case. If contact request message is lost in the network than client can't be a mutual contact. In other words, if they are mutual contacts, it means no contact request was lost in the network 🤔

If contact request message can be lost in the network, we should not limit client code to send an extra CR. That means we should handle it on the receiver's side. So we are ignoring extra CRs for mutual and declined contacts, but they can still be useful for pending state to update CR's message

osmaczko:
> If contact request message can be lost in the network, we should not limit client code to send an extra CR.

That's understood. However, this PR is NOT about enabling clients to send extra CRs, but rather about ignoring extra sent CRs if the sender is already a mutual or declined contact. The description is somewhat confusing  :thinking:

osmaczko (approved):

MishkaRogachev:
> > If contact request message can be lost in the network, we should not limit client code to send an extra CR.
> 
> That's understood. However, this PR is NOT about enabling clients to send extra CRs, but rather about ignoring extra sent CRs if the sender is already a mutual or declined contact. The description is somewhat confusing 🤔

Yep, i'll fix that. This was a leftover, as I originally wanted to throw an error when sending a contact request to a pending contact

MishkaRogachev:
Screenshot 2023-07-14 at 13 11 28
Tested with bot

cammellos (approved):

PR #3756, opened 2023-07-13
feat(communities): add airdropAddress param to SharedAccounts

jrainville:
Needed for [#11479](https://github.com/status-im/status-desktop/issues/11479)

Adds airdropAddress to the request to join params and a is_airdrop_address flag in the communities_requests_to_join_revealed_addresses table. 

This airdropAddress is used by the owner to know which address to use when airdropping

Based on top of https://github.com/status-im/status-go/pull/3748 because I didn't want to conflict with myself.


protocol/messenger_communities.go
```diff
@@ -709,10 +709,16 @@ func (m *Messenger) getAccountsToShare(addressesToReveal []string, communityID t
 
 		revealedAddress := gethcommon.HexToAddress(verifiedAccount.Address.Hex())
 		revealedAddresses = append(revealedAddresses, revealedAddress)
+		address := verifiedAccount.Address.Hex()
+		isAirdropAddress := address == airdropAddress
+		if airdropAddress == "" {
+			isAirdropAddress = i == 0
```
ilmotta, line 715: @jrainville, if the parameter `airdropAddress` was validated to never be the empty string in `EditSharedAddresses` and on `RequestToJoinCommunity` it was validated to not be empty if there are addresses to reveal, then why do we need to fallback to the first available address? Shouldn't we return an error or simply not test this condition if the validation exists?

Maybe I misinterpreted the request validations.

ilmotta (approved):

protocol/messenger_communities.go
jrainville: Good eye. The code is a bit confusing. It's because we accept two cases:
1. Manually selecting addresses: then we need to have a manually select airdrop address as well
2. No selected addresses: in that case, the community manager will parse the addresses and select the wallet accounts automatically for the user. Therefore, we need to assign one of those accounts as the airdrop address, so we select the first account (this is the line we are commenting on). 

When we edit the shared addresses, then we are in the first case, because that comes from the user wanting to manually select addresses. So in the case of edit, we always want the airdrop address to be given (that's why we validate differently there).

Hopefully I explained clearly. Do you want me to add comments to the lines that confused you?

ilmotta: Makes perfect sense @jrainville, thanks :)

At first I thought selecting the first available address could be a problem or be undesirable, but it's fine really.

> Do you want me to add comments to the lines that confused you?

Up to you, but it's clear to me now.

osmaczko (approved):
:+1:

PR #3757, opened 2023-07-14
Update go waku

cammellos:
Updates go waku to avoid a crash, it's using an off brach version as per latest version some tests in waku2 are failing, and we are close to release cutting and we want to avoid extra changes.

The changes include recovering from a panic (bad but the other solution proposed was locking in tests), and soft fail on stop when unsubscribing from peers, since it currently was trying to wait for the peers to acknowledge, but that's no necessary when logging out and increase logout time by up to 30s.


protocol/messenger.go
```diff
@@ -521,7 +521,17 @@ func NewMessenger(
 			pushNotificationClient.Stop,
 			communitiesManager.Stop,
 			encryptionProtocol.Stop,
-			transp.ResetFilters,
+			func() error {
+				ctx, cancel := context.WithTimeout(context.Background(), time.Second*1)
```
Samyoul: Nothing major, but you don't need to multiply by 1 🤷 

```suggestion
				ctx, cancel := context.WithTimeout(context.Background(), time.Second)
```

Samyoul (approved):
That was a lot of CPU and memory for basically 1 function worth of review 🤣

protocol/messenger.go
cammellos: ouch :)

qfrank:
any chance to write a test to reproduce the issue and prove it get solved? @cammellos

qfrank (approved):

cammellos:
@qfrank it was caught by our mobile integration tests, so it's partially cover there, I wasn't quite able to replicate it here, but I can try again, though not on this PR as this is a release blocker and would like to get it in as soon as possible

qfrank:
in my memory, it was the second time we got the like issue 🥹

cammellos:
> in my memory, it was the second time we got the like issue 🥹

different issue, albeit similar, both times was caught by our integration tests, the previous one was sending on a closed channel, this one was closing a closed channel.
The previous one was addressed by making sure the owner of a channel is also the reader (it should always be like that), this one is due to some concurrency where start/stop/start/stop are interleaved I believe, but could not replicate yet

qfrank:
It's highly possible that we have different memories 😉. I'm referring to this: https://github.com/waku-org/go-waku/pull/553. The similarity is that there's a delay when logging out.

cammellos:
oh @qfrank the dead loop is not in master, it's still a PR that we were trying to avoid the crash

PR #3765, opened 2023-07-17
feat: add CommunityEncryptionKeyActions and its evaluation logic

osmaczko:
Added utility that evaluates necessary encryption key actions based on community changes.

part of: status-im/status-desktop#10998

### DIFF
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -1018,6 +1018,14 @@ func (o *Community) UpdateCommunityDescription(description *protobuf.CommunityDe
 	o.mutex.Lock()
 	defer o.mutex.Unlock()
 
+	// This is done in case tags are updated and a client sends unknown tags
+	description.Tags = requests.RemoveUnknownAndDeduplicateTags(description.Tags)
+
+	err := ValidateCommunityDescription(description)
+	if err != nil {
+		return nil, err
+	}
+
 	response := o.emptyCommunityChanges()
 
 	// allowEqualClock == true only if this was a description from the handling request to join sent by an admin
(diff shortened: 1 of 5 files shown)


protocol/communities/community_encryption_key_action.go
```diff
@@ -0,0 +1,105 @@
...
+		return result
+	}
+
+	// open community/channel does not require any actions
+	if len(modifiedPermissions) == 0 {
+		return result
+	}
+
+	if len(membersRemoved) > 0 {
+		result.ActionType = EncryptionKeyRekey
```
jrainville, line 106: Is there a possibility that a permission would be modified at the same time as a member would be removed? If so, maybe we should move this condition at the top, since it seems to have more impact than `EncryptionKeyAdd` for example.

jrainville (commented):
Very cool. Nice tests too.
I just had one question

protocol/communities/community.go
```diff
@@ -1719,6 +1720,38 @@ func (o *Community) AddMember(publicKey *ecdsa.PublicKey, roles []protobuf.Commu
 	return changes, nil
 }
 
+func (o *Community) AddMemberToChat(chatID string, publicKey *ecdsa.PublicKey, roles []protobuf.CommunityMember_Roles) (*CommunityChanges, error) {
```
mprakhov, line 1825: It is not clear why we need it as it is used only in tests and even in tests we only check that this will not return an error, not check that` o *Community `was updated

osmaczko: This is needed since channels can have different member list than the whole community due to channel permissions.

It is indirectly tested here: https://github.com/status-im/status-go/blob/af9eecd7afdb07fd258921a0478e16166cb9f2df/protocol/communities/community_encryption_key_action_test.go#L774-L778

It is used in tests only for now. I am building the utility that I will later use in the production.

protocol/communities/community_encryption_key_action.go
osmaczko: This utility determines if:
a) permission was just added (there was no permission before) - we send new key to everyone
b) permission was just removed (there was permission before, there is none anymore) - we "remove" the key (what we really do we simply stop encryption)
c) members changes - either rekey when any member was removed or send existing key to new members

If permission is modified at the same time as a member is removed it doesn't matter. All what matters are members changes, i.e. members will be removed or added depending if they satisfied the modified permission.

jrainville (approved):
:tada:

mprakhov (approved):

PR #3769, opened 2023-07-18
fix: send HistoryArchiveSupportEnabled setting with CommunityEditAdminEvent

MishkaRogachev:
For https://github.com/status-im/status-desktop/pull/11560

Important changes:
- [x] Add `HistoryArchiveSupportEnabled` setting to CommunityEditAdminEvent.


protocol/communities/manager.go
```diff
@@ -1399,7 +1407,37 @@ func (m *Manager) HandleCommunityAdminEvent(signer *ecdsa.PublicKey, adminEvent
...
+		if adminEvent.Clock > settings.Clock {
+			settings.HistoryArchiveSupportEnabled = adminEvent.CommunityConfig.HistoryArchiveSupportEnabled
+			settings.Clock = adminEvent.Clock
+		}
+
+		err = m.persistence.SaveCommunitySettings(*settings)
+		if err != nil {
+			return nil, err
+		}
+		response.CommunitySettings = settings
```
MishkaRogachev, line 1438: @mprakhov @0x-r4bbit @jrainville 
Here I'm trying to sync `HistoryArchiveSupportEnabled` between community admins and it's coming out pretty messy. Is there a reason why we keep this setting separate in CommunitySettings but not in CommunityDescription?

On the other hand, can we use `SyncCommunitySettings` not only with other devices, but but with other admins?

MishkaRogachev: As i see [here](https://github.com/status-im/status-desktop/blob/fix/issue-11089-community-toggling-history-service-as-admin-is-not-reflected-to-the-owner-node/src/app_service/service/node_configuration/service.nim#L59) this setting make sense only for desktop owner node

0x-r4bbit: >  Is there a reason why we keep this setting separate in CommunitySettings but not in CommunityDescription?

Right. `community_settings` are settings that belong to an individual account/node. They are not settings that are meant to be shared.

Every account keeps track of community settings for themselves.

MishkaRogachev:
Not needed  for https://github.com/status-im/status-desktop/pull/11560

PR #3770, opened 2023-07-18
feat: Add contract deployment tx type

Cuteivist:
Task: #9047

Desktop changes: https://github.com/status-im/status-desktop/pull/11570

* Added new transaction type - contract deployment

### DIFF
--- a/services/wallet/activity/activity.go
+++ b/services/wallet/activity/activity.go
@@ -645,11 +645,12 @@ func getActivityEntries(ctx context.Context, deps FilterDependencies, addresses 
 		var dbMtType, dbTrType sql.NullByte
 		var toAddress, fromAddress eth.Address
 		var toAddressDB, ownerAddressDB, contractAddressDB sql.RawBytes
-		var tokenAddress *eth.Address
+		var tokenAddress, contractAddress *eth.Address
 		var aggregatedStatus int
 		var dbTrAmount sql.NullString
 		var dbMtFromAmount, dbMtToAmount, contractType sql.NullString
 		var tokenCode, fromTokenCode, toTokenCode sql.NullString
+		var transferType *TransferType
 		err := rows.Scan(&transferHash, &pendingHash, &chainID, &multiTxID, &timestamp, &dbMtType, &dbTrType, &fromAddress,
 			&toAddressDB, &ownerAddressDB, &dbTrAmount, &dbMtFromAmount, &dbMtToAmount, &aggregatedStatus, &aggregatedCount,
 			&tokenAddress, &tokenCode, &fromTokenCode, &toTokenCode, &outChainIDDB, &inChainIDDB, &contractType, &contractAddressDB)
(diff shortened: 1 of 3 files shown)


services/wallet/activity/activity.go
```diff
@@ -623,6 +624,10 @@ func getActivityEntries(ctx context.Context, deps FilterDependencies, addresses
 		getActivityType := func(trType sql.NullByte) (activityType Type, filteredAddress eth.Address) {
 			if trType.Valid {
 				if trType.Byte == fromTrType {
+					match, _ := regexp.MatchString("0x0+$", toAddress.String())
```
alaibe: Could you try doing `common.HexToAddress("0x0") == toAddress `

dlipicar: Hmm a contract deployment activity should be IMO:
- A `Send` activity
- That has `type` == `eth`
- And `contract_address` != null address

The toAddress == null address condition could be a token burn, for example

stefandunca: This will work; however, something like `if toAddress == ZeroAddress` might be more explicit.

stefandunca (approved):
Great job. Left a recommendation inline.

services/wallet/activity/activity.go
Cuteivist: Used `ZeroAddress` that @stefandunca  proposed. Much clearer check.

Cuteivist: Awesome hint! Added additional checks from @dlipicar

Cuteivist: Updated checks, thanks for pointing that out !

stefandunca (approved):

services/wallet/activity/activity.go
```diff
@@ -675,6 +678,10 @@ func getActivityEntries(ctx context.Context, deps FilterDependencies, addresses
 				tokenOut = deps.tokenFromSymbol(&cID, tokenCode.String)
 			}
 
+			if activityType == SendAT && ZeroAddress == toAddress && tokenOut.TokenType == Native {
```
dlipicar: Doesn't look that we're checking whether the ContractAddress field is non-zero here, this could be an ETH burn operation.

dlipicar: Example: https://etherscan.io/tx/0xc48210684a5705188e9456f3abfa4baa5bc9647973f59d64ccf8b221b310091a

Cuteivist: Updated. Detects contract deployment, while ignoring burn.

services/wallet/activity/activity.go
```diff
@@ -652,6 +655,9 @@ func getActivityEntries(ctx context.Context, deps FilterDependencies, addresses
 				involvedToken = &Token{TokenType: Native, ChainID: common.ChainID(chainID.Int64)}
 			}
 			if activityType == SendAT {
+				if toAddress == ZeroAddress && involvedToken.TokenType == Native && *contractAddress != ZeroAddress {
```
dlipicar: some DB entries might have a `NULL` `contractAddress` I believe, better to check that `contractAddress` is not `nil` as well. Also, we're doing the same check in two different places, probably a good idea to move it to a separate function

Cuteivist: @dlipicar Good point. I updated the checks

stefandunca:
@Cuteivist please rebase and use `contract_address` added here https://github.com/status-im/status-go/pull/3775

PR #3772, opened 2023-07-18
feat: add api to remove private key

MishkaRogachev:
For https://github.com/status-im/status-desktop/issues/11475

Important changes:
- [x] API for removing community private key
- [x] Separate `IsOwner` from `HasPrivateKey`
- [x] Remove Private key from syncing


jrainville (commented):
Looks good so far

protocol/communities/manager.go
```diff
@@ -607,13 +607,13 @@ func (m *Manager) CreateCommunityTokenPermission(request *requests.CreateCommuni
 		return nil, nil, err
 	}
 
-	if community.IsOwner() {
+	if community.IsControlNode() {
```
0x-r4bbit: For all of these where we publish a community description, I think we still want to just check for `isOwner()` no?

protocol/communities/community.go
```diff
@@ -233,6 +237,10 @@ func (o *Community) MarshalJSON() ([]byte, error) {
 	}{
 		ID:                          o.ID(),
 		MemberRole:                  o.MemberRole(o.MemberIdentity()),
+		IsAdmin:                     o.IsAdmin(),
+		IsOwner:                     o.IsOwner(),
+		IsControlNode:               o.IsControlNode(),
+		HasAdminPrivileges:          o.IsOwnerOrAdmin(),
```
osmaczko: The properties `IsAdmin` and `IsOwner` are mutually exclusive and are already better(imo) represented by the `MemberRole`.

osmaczko: I am wondering if we really need to expose that, can't clients simply check if member role is either owner or admin? I wouldn't consider it as crucial logic duplication among clients.

protocol/protobuf/pairing.proto
```diff
@@ -138,7 +138,8 @@ message SyncInstallationPublicChat {
 message SyncCommunity {
   uint64 clock = 1;
   bytes id = 2;
-  bytes private_key = 3;
+  // Don't sync private_key vecause we want to have only one control node
+  bytes private_key = 3 [deprecated = true];
```
osmaczko, line 142: :+1:

protocol/communities/community.go
mprakhov: +1, MemberRole will be enough
P.S. IsControlNode is ok

protocol/protobuf/pairing.proto
```diff
@@ -138,7 +138,8 @@ message SyncInstallationPublicChat {
 message SyncCommunity {
   uint64 clock = 1;
   bytes id = 2;
-  bytes private_key = 3;
+  // Don't sync private_key vecause we want to have only one control node
```
jrainville:
```suggestion
  // Don't sync private_key because we want to have only one control node
```

jrainville (approved):
LGTM

mprakhov (approved):
LGTM, all other stuff I'll fix in scope of my task where owner without privateKey must behave like an admin
For this purposes I'll need your MR :)

protocol/communities/community.go
MishkaRogachev: returned back!

osmaczko (approved):

PR #3773, opened 2023-07-18
fix(handleSyncKeypair): skip empty keypairs

igor-sirotin:
This is a hotfix for https://github.com/status-im/status-desktop/issues/11533.

As it turned out, after pairing mobile->desktop the latter would receive an and empty `SyncKeypair` message. 
This was leading to a status-desktop crash at some point.

I couldn't quickly find the source of such empty sync keypairs being sent. 
But I also couldn't reproduce the issue on latest status-desktop master. So I suppose it's ok to push such a hotfix with 0.13.1.


jrainville (approved):

jrainville:
@igor-sirotin can you add a PR on dekstop too to use that status-go commit. Thanks

PR #3774, opened 2023-07-19
feat: introduce CommunitiesKeyDistributor

osmaczko:
Introduced `CommunitiesKeyDistributor`

This component decouples key distribution from the Messenger, enhancing code maintainability, extensibility and testability. It also alleviates the need to impact all methods potentially affecting encryption keys.
Moreover, it allows key distribution inspection for integration tests.

part of: status-im/status-desktop#10998

### DIFF
--- a/protocol/common/raw_message.go
+++ b/protocol/common/raw_message.go
@@ -35,4 +35,5 @@ type RawMessage struct {
 	CommunityKeyExMsgType CommKeyExMsgType
 	Ephemeral             bool
 	BeforeDispatch        func(*RawMessage) error
+	HashRatchetGroupID    []byte
 }
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -2222,6 +2222,7 @@ func (o *Community) deleteTokenPermission(permissionID string) (*CommunityChange
 	delete(o.config.CommunityDescription.TokenPermissions, permissionID)
 
 	changes := o.emptyCommunityChanges()
+
 	changes.TokenPermissionsRemoved[permissionID] = permission
 	return changes, nil
 }
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -681,6 +681,20 @@ func (o *Community) GetMember(pk *ecdsa.PublicKey) *protobuf.CommunityMember {
 	return o.getMember(pk)
 }
 
+func (o *Community) getChatMember(pk *ecdsa.PublicKey, chatID string) *protobuf.CommunityMember {
+	if !o.hasMember(pk) {
+		return nil
+	}
+
+	chat, ok := o.config.CommunityDescription.Chats[chatID]
+	if !ok {
+		return nil
+	}
+
+	key := common.PubkeyToHex(pk)
+	return chat.Members[key]
+}
+
 func (o *Community) hasMember(pk *ecdsa.PublicKey) bool {
 
 	member := o.getMember(pk)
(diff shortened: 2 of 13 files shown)


protocol/messenger.go
```diff
@@ -460,19 +461,23 @@ func NewMessenger(
...
+		transport:              transp,
+		encryptor:              encryptionProtocol,
+		sender:                 sender,
+		anonMetricsClient:      anonMetricsClient,
+		anonMetricsServer:      anonMetricsServer,
+		telemetryClient:        telemetryClient,
+		pushNotificationClient: pushNotificationClient,
+		pushNotificationServer: pushNotificationServer,
+		communitiesManager:     communitiesManager,
+		communitiesKeyDistributor: &CommunitiesKeyDistributorImpl{
```
0x-r4bbit, line 488: Just cosmetics, but we could have a `NewKeyDistributor()` factory fn for this

protocol/messenger_communities.go
```diff
@@ -211,11 +211,45 @@ func (m *Messenger) handleCommunitiesHistoryArchivesSubscription(c chan *communi
...
+			m.logger.Warn("failed to retrieve orgs", zap.Error(err))
+			return result
+		}
+
+		for _, org := range ownedOrgs {
+			result[org.IDString()] = org
+		}
+
+		return result
+	}()
```
0x-r4bbit, line 231: Any reason this is wrapped in an IIFE? Nothing that needs change, just curious

0x-r4bbit (approved):

protocol/communities/community.go
```diff
@@ -1363,6 +1363,7 @@ func (o *Community) AddTokenPermission(permission *protobuf.CommunityTokenPermis
 	}
 
 	o.config.CommunityDescription.TokenPermissions[permission.Id] = permission
+	o.updateEncrypted()
```
mprakhov: If I understood correctly, if you'll do updateEcrypted, `UpdateCommunityEncryption` will always return nil due to
```
if community.Encrypted() == isEncrypted {
		return nil
	}
```

mprakhov: ok, I see, it was removed

mprakhov: What about admin events? What if the admin will add member permission? Add/Delete/Update token CommunityEvent must include encryption information in this case

protocol/messenger_communities.go
osmaczko: It represents the intention a bit better, you can read straight away "this piece of code is responsible for obtaining recentlyPublishedOrgs". Also, this way I don't pollute the scope with unneeded `ownedOrgs` variable.

protocol/messenger.go
osmaczko: Was thinking about factory as well, but since this struct doesn't need any special initialization, I left it as is.

jrainville (approved):

osmaczko:
Waits for QA: https://github.com/status-im/status-desktop/pull/11657

protocol/common/raw_message.go
```diff
@@ -35,4 +35,5 @@ type RawMessage struct {
 	CommunityKeyExMsgType CommKeyExMsgType
 	Ephemeral             bool
 	BeforeDispatch        func(*RawMessage) error
+	HashRatchetGroupID    []byte
```
Samyoul, line 38: This is a nice change.

protocol/communities_messenger_test.go
```diff
@@ -3447,47 +3447,49 @@ func (s *MessengerCommunitiesSuite) TestStartCommunityRekeyLoop() {
 	s.Require().False(c.Encrypted())
 	// TODO some check that there are no keys for the community. Alt for s.Require().Zero(c.RekeyedAt().Unix())
 
-	// Update the community to use encryption and check the values
```
Samyoul, line 3453: were your new changes causing this test to fail?

Samyoul (approved):
Thank you for your patience and your work here with this PR. 🙏

protocol/communities_messenger_test.go
osmaczko: It didn't compile because I removed `UpdateCommunityEncryption`. I will address that in follow-up.

PR #3779, opened 2023-07-19
feat(new-urls-format): Usage the chat key instead of signature

borismelnik:
On the last meeting about new urls formats we have agreed to use `chat key` instead of signature
This PR replace signatures to chat keys in url generators

Important changes:

### DIFF
--- a/protocol/messenger_share_urls.go
+++ b/protocol/messenger_share_urls.go
@@ -1,7 +1,6 @@
 package protocol
 
 import (
-	"crypto/ecdsa"
 	"fmt"
 	"regexp"
 	"strings"
@@ -24,6 +23,7 @@ type CommunityURLData struct {
 	MembersCount uint32   `json:"membersCount"`
 	Color        string   `json:"color"`
 	TagIndices   []uint32 `json:"tagIndices"`
+	CommunityID  string   `json:"communityId"`
 }
 
 type CommunityChannelURLData struct {
@@ -31,11 +31,13 @@ type CommunityChannelURLData struct {
 	DisplayName string `json:"displayName"`
 	Description string `json:"description"`
 	Color       string `json:"color"`
+	ChannelUUID string `json:"channelUuid"`
 }
 
 type ContactURLData struct {
 	DisplayName string `json:"displayName"`
 	Description string `json:"description"`
+	PublicKey   string `json:"publicKey"`
 }
 
 type URLDataResponse struct {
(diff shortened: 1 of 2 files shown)


protocol/messenger_share_urls.go
```diff
@@ -119,23 +123,13 @@ func (m *Messenger) parseCommunityURLWithChatKey(urlData string) (*URLDataRespon
 	}, nil
 }
 
-func (m *Messenger) prepareEncodedRawData(rawData []byte, privateKey *ecdsa.PrivateKey) (string, string, error) {
+func (m *Messenger) prepareEncodedRawData(rawData []byte, privateKey *ecdsa.PrivateKey) (string, error) {
```
MishkaRogachev: If there is no signing, there is no sense in `prepareEncodedRawData`, it can be just replaced with `urls.EncodeDataURL(rawData)`

MishkaRogachev (commented):
Look great! Just one question, at least we need to remove `privateKey` mentioning here

protocol/messenger_share_urls_test.go
```diff
@@ -369,20 +369,20 @@ func (s *MessengerShareUrlsSuite) TestShareUserURLWithData() {
 	url, err := s.m.ShareUserURLWithData(contact.ID)
 	s.Require().NoError(err)
 
-	userData, signature, err := s.m.prepareEncodedUserData(contact)
+	userData, chatKey, err := s.m.prepareEncodedUserData(contact)
 	s.Require().NoError(err)
 
-	expectedURL := fmt.Sprintf("%s/u/%s#%s", baseShareURL, userData, signature)
+	expectedURL := fmt.Sprintf("%s/u/%s#%s", baseShareURL, userData, chatKey)
 	s.Require().Equal(expectedURL, url)
```
jrainville, line 377: I think the asserts could improved or at leats maybe we could add a new assert.

The one you do right now basically just checks that `ShareUserURLWithData` returns the URL with the right order, because both use `prepareEncodedUserData`.

It would be good to test that the encoding methods do the right thing by having a "hardcoded" result.

eg:
```go
s.Require().Equal("https://status.app/cc/dfiohf32hf92hfhf30jksapofpok", url)
```

jrainville (commented):
Looks good. Just check Mishka's comment and also maybe we caa improve the tests.

protocol/messenger_share_urls_test.go
borismelnik: yeah, that was a first plan :) but when we create community or contact we should get different ids. At the end, `Shared***Data` methods trying to get community or contact from db. The one thing which I can improve - check that generated data is equal to hardcoded string (except channel data cos it embeds `uuid`)

jrainville (approved):

MishkaRogachev (approved):
LGTM!

PR #3781, opened 2023-07-20
fix: discard backed up messages if `ProcessBackedupMessages` is not set to `true`

saledjenic:
If user followed onboarding flow to recover his account using seed phrase or keycard, then `ProcessBackedupMessages` property of node config json object should be set to `true`, otherwise it should be set to `false` or be omitted.

This should be handled on mobile side as well, especially if mobile app is using `LoginWithKeycard` call.


alaibe (approved):

qfrank (approved):
LGTM, but this PR contains break change(LoginWithKeycard) to mobile , can you wait to merge it after mobile align this. I can do it next Monday @saledjenic

saledjenic:
> can you wait to merge it after mobile align this

@qfrank yes sure, thanks and please let me know when is the right time for the merge.

qfrank:
> when is the right time for the merge.

will notify you next Monday 🙂

PR #3782, opened 2023-07-20
feat: add preferred networks

alaibe:


saledjenic (approved):

multiaccounts/accounts/database.go
```diff
@@ -52,22 +52,24 @@ type Keypair struct {
...
+	Name                  string                    `json:"name"`
+	Emoji                 string                    `json:"emoji"`
+	ColorID               common.CustomizationColor `json:"colorId,omitempty"`
+	Hidden                bool                      `json:"hidden"`
+	Clock                 uint64                    `json:"clock,omitempty"`
+	Removed               bool                      `json:"removed,omitempty"`
+	Operable              AccountOperable           `json:"operable"` // describes an account's operability (read an explanation at the top of this file)
+	CreatedAt             int64                     `json:"createdAt"`
+	Position              int64                     `json:"position"`
+	ProdPreferredChainIDs string                    `json:"prodPreferredChainIds"`
```
dlipicar, line 75: I'd use a slice of uint64s everywhere other than the DB instead of moving around the encoded string

dlipicar (approved):

multiaccounts/accounts/database.go
```diff
@@ -804,6 +827,17 @@ func (db *Database) saveOrUpdateAccounts(tx *sql.Tx, accounts []*Account, update
 			}
 			keyUID = &acc.KeyUID
 		}
+		var exists bool
+		err = tx.QueryRow("SELECT EXISTS (SELECT 1 FROM keypairs_accounts WHERE address = ?)", acc.Address).Scan(&exists)
+		// Apply default values if account is new and not a watch only
+		if !exists && acc.Type != AccountTypeWatch {
+			if acc.ProdPreferredChainIDs == "" {
```
saledjenic, line 958: @alaibe cannot precisely recall, but I swear I saw somewhere that you moved this logic from here to messenger?

alaibe: @saledjenic After check with feature requirement, it shouldn't be possible to have them empty so changing here is working in all cases

PR #3785, opened 2023-07-20
fix(communities): fix revealed addresses string comparisons

jrainville:
Turns out the addresses are not always in the same format. I converted to lower case but maybe there is a better way to do this cc @alaibe

### DIFF
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -762,7 +762,7 @@ func (m *Messenger) getAccountsToShare(addressesToReveal []string, airdropAddres
 
 	containsAddress := func(addresses []string, targetAddress string) bool {
 		for _, address := range addresses {
-			if address == targetAddress {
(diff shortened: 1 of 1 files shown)


osmaczko (commented):
```
address1 := "0x71c7656ec7ab88b098defb751b7401b5f6d8976f"
ethAddress1 := common.HexToAddress(address1)

address2 := "0x71c7656ec7ab88b098defb751b7401b5f6d8976F"
ethAddress2 := common.HexToAddress(address2)

fmt.Println(ethAddress1 == ethAddress2)  // Will print: true
```

alaibe:
Exacly what @osmaczko said, use the address type!

jrainville:
Thanks. I updated

osmaczko (approved):

alaibe (approved):

PR #3786, opened 2023-07-20
fix: remove redundancy in permissions check when handling request to join

osmaczko:
- Fixed redundant permissions check. If community is set to auto-accept, then permissions would be checked twice, in `HandleCommunityRequestToJoin` and `AcceptRequestToJoinCommunity`. Mitigated it by returning from `HandleCommunityRequestToJoin` immediately in case of auto-accept.

- Extracted `accountsSatisfyPermissionsToJoin` to remove code duplication and simplify the logic.

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -1488,6 +1488,27 @@ func (m *Manager) CheckPermissionToJoin(id []byte, addresses []gethcommon.Addres
 	return m.checkPermissionToJoin(permissionsToJoin, accountsAndChainIDs, false)
 }
 
+func (m *Manager) accountsSatisfyPermissionsToJoin(community *Community, accounts []*protobuf.RevealedAccount) (satisfy bool, isAdmin bool, err error) {
+	accountsAndChainIDs := revealedAccountsToAccountsAndChainIDsCombination(accounts)
+	becomeAdminPermissions := community.TokenPermissionsByType(protobuf.CommunityTokenPermission_BECOME_ADMIN)
+	becomeMemberPermissions := community.TokenPermissionsByType(protobuf.CommunityTokenPermission_BECOME_MEMBER)
+
+	if len(becomeAdminPermissions) > 0 && m.accountsHasAdminPermission(becomeAdminPermissions, accountsAndChainIDs) {
+		return true, true, nil
+	}
+
+	if len(becomeMemberPermissions) > 0 {
+		permissionResponse, err := m.checkPermissionToJoin(becomeMemberPermissions, accountsAndChainIDs, true)
+		if err != nil {
+			return false, false, err
+		}
+
+		return permissionResponse.Satisfied, false, nil
+	}
+
+	return true, false, nil
+}
+
 func (m *Manager) AcceptRequestToJoin(request *requests.AcceptRequestToJoinCommunity) (*Community, error) {
 	dbRequest, err := m.persistence.GetRequestToJoin(request.ID)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


protocol/communities/manager.go
```diff
@@ -1707,10 +1701,6 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request
 			if !matching {
 				// if ownership of only one wallet address cannot be verified,
 				// we mark the request as cancelled and stop
-				err = m.markRequestToJoinAsCanceled(signer, community)
-				if err != nil {
-					return nil, err
-				}
 				requestToJoin.State = RequestToJoinStateDeclined
```
mprakhov, line 1704: Not sure that this is not correct. You are removing markRequestToJoinAsCanceled while
Declined != Cancelled

protocol/communities/manager.go
```diff
@@ -1724,68 +1714,26 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request
...
-			return nil, err
-		}
-		requestToJoin.State = RequestToJoinStateAccepted
+	permissionsSatisfied, _, err := m.accountsSatisfyPermissionsToJoin(community, request.RevealedAccounts)
+	if err != nil {
+		return nil, err
+	}
+	if !permissionsSatisfied {
+		requestToJoin.State = RequestToJoinStateDeclined
+		return requestToJoin, nil
```
0x-r4bbit: Doesn't look like that returning here is necessary, as it's the last operation before returning anyways.

protocol/communities/manager.go
```diff
@@ -1707,10 +1701,6 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request
 			if !matching {
 				// if ownership of only one wallet address cannot be verified,
 				// we mark the request as cancelled and stop
-				err = m.markRequestToJoinAsCanceled(signer, community)
-				if err != nil {
-					return nil, err
-				}
```
0x-r4bbit, line 1713: This looked suspicious first.. I wonder why we marked the request to join as `canceled` here.
Should've probably rather been marked as `declined` in the first place.

protocol/communities/manager.go
```diff
@@ -1724,68 +1714,26 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request
...
-		err = m.markRequestToJoinAsCanceled(signer, community)
+	if acceptAutomatically {
+		err = m.markRequestToJoin(signer, community)
 		if err != nil {
 			return nil, err
 		}
-		requestToJoin.State = RequestToJoinStateDeclined
+		requestToJoin.State = RequestToJoinStateAccepted
 		return requestToJoin, nil
 	}
```
mprakhov: In this case, if `acceptAutomatically && memberPermission` is set, I, as a control node, will accept the member without validating new member permission which is not correct.
For me, as a reader, it is not clear, that we will do this check further in another function, so please, add a comment

0x-r4bbit (approved):

protocol/communities/manager.go
0x-r4bbit: Yea I saw that one too. But even if it's marked as cancelled (which is already... unexpected), it'll be marked as declined later on.

So removing markedAsCancelled si fine here.

protocol/communities/manager.go
osmaczko: It was redundant, it will be indirectly invoked further in the processing pipeline, here:
https://github.com/status-im/status-go/blob/5543a29f6250d922eee70571e2b1590209b798d5/protocol/messenger_handler.go#L1425-L1434

protocol/communities/manager.go
osmaczko: Exactly as @0x-r4bbit said.

protocol/communities/manager.go
osmaczko: Sure, I will add the comment.

protocol/communities/manager.go
0x-r4bbit: That's correct, I still wonder why it was "mark as cancelled" in the first place

mprakhov (approved):

protocol/communities/manager.go
osmaczko: I presume that the redundancy was unintentionally introduced. The tests are passing, so I'll take that as an indication that there's no scenario where it's genuinely necessary.

PR #3787, opened 2023-07-21
chore: wallet accounts improvements - soft delete

saledjenic:


multiaccounts/accounts/database.go
```diff
@@ -602,57 +639,100 @@ func (db *Database) getAccountByAddress(tx *sql.Tx, address types.Address) (*Acc
...
 }
 
+// Returns keypair if it is not marked as removed and its accounts which are not marked as removed.
 func (db *Database) GetKeypairByKeyUID(keyUID string) (*Keypair, error) {
 	return db.getKeypairByKeyUID(nil, keyUID)
 }
 
-func (db *Database) GetAccounts() ([]*Account, error) {
-	return db.getAccounts(nil, types.Address{})
+func (db *Database) GetAccounts(includeRemoved bool) ([]*Account, error) {
```
dlipicar: Seems like there's only a few specific cases (only one?) where somebody would want to get removed accounts on top of current ones, 99% of times we just want to get the current list of accounts.
For this reason, I'd suggest keeping the old functions signatures working like they were before and add new functions for the new particular case. It can be done without duplicating code, and you'll be able to avoid most file changes that way.

saledjenic: I was thinking about that and couldn't decide which approach to follow and then opted in for that one (with 51:49 :) ), where developer decides whether to get removed accounts as well or not at the place of usage, instead having two functions like `getAccounts` and `getAllAccounts` or `getAccountsIncludingRemoved`. 

Anyhow I don't mind to change that, but let's hear from other reviewers what do they think about this?
(personally I would keep it as it's now, but let's have a majority)

dlipicar: Sure! This is just an opinion, I normally also like to have a single API with parameters and "force" the user to know what data he is (or isn't) getting back, but in this particular case I think including removed accounts would be unexpected and undesirable for almost all users.

dlipicar (approved):

api/geth_backend.go
```diff
@@ -694,7 +694,7 @@ func (b *GethStatusBackend) MigrateKeyStoreDir(acc multiaccounts.Account, passwo
 	if err != nil {
 		return err
 	}
-	accounts, err := accountDB.GetAccounts()
+	accounts, err := accountDB.GetAccounts(false)
```
IvanBelyakoff: @saledjenic I am fine with both approaches for `GetAccounts` - using params or a separate method. I'd go with params in C++ with a default param `includeRemoved=false` to cover 99% cases, but since Golang does not directly support it, I don't have strong opinion on that. Though I don't quite like function invokations, when I don't understand what function accepts like this:
`accountDB.GetAccounts(false)`
 or sometimes it is even worse when there are more params:
`accountDB.GetAccounts(false, true, 5, false)`
when you can't understand the signature without fancy tools like IDE.
Normally they say that a boolean parameter clearly shows that a function is low cohesive - does 2 different things based on the parameter, so it should be split into 2 different functions. So if it was not boolean, I could go with parameter. Since it is bool and the only parameter, i'd go with
a separate function like `GetAccountsWithRemoved` which is more readable and more cohesive to me.

IvanBelyakoff (approved):
Good job, Sale!
I expressed my opinion on what @dlipicar raised in an inline comment. Please have a look

multiaccounts/accounts/database.go
rasom: I'm for a separate function, whereas the old one can be renamed to `getActiveAccounts` or something like this, and a new one `getAllAccounts`. Another approach would be to have a separate call which would only fetch removed accounts and we deal with merging of both collections on the client side (if necessary). It might help with less confusing names though adds some work on the caller side.

saledjenic:
@all I updated functions' names, thanks.

protocol/messenger_sync_wallets_test.go
```diff
@@ -146,12 +146,12 @@ func (s *MessengerSyncWalletSuite) TestSyncWallets() {
 	woAccounts := accounts.GetWatchOnlyAccountsForTest()
 	err = s.m.settings.SaveOrUpdateAccounts(woAccounts, false)
 	s.Require().NoError(err)
-	dbWoAccounts1, err := s.m.settings.GetWatchOnlyAccounts()
+	dbWoAccounts1, err := s.m.settings.GetWatchOnlyAccounts(false)
```
dlipicar, line 149: Don't we want to follow the same approach with this one?

saledjenic: Not sure how I missed it, was in a rush, anyhow here is a PR handling this function as well, https://github.com/status-im/status-go/pull/3807. Thanks.

PR #3788, opened 2023-07-21
chore: make the owner without the community private key behave like a…

### LINKED ISSUE #11477: Make the owner without the community private key behave like an admin (ownership transfer)
### Description

First step of Owners without private key.

We want to support the flow where a user creates a community on a computer, then imports the same account on another computer with the same community (imports the community key), then forgets the key in the main computer.

That would make it so that the main computer remains an owner node, but **not** a control node.

So we need that main computer node to still have admin rights. In theory, the owner would still have the right to mint tokens and other token flows as well (as seen here: https://github.com/status-im/status-desktop/issues/10114), but as a first step, let's keep it to admin rights.

### Acceptance Criteria

- When an owner is no longer a control node, they should still have admin rights as per this list https://github.com/status-im/status-desktop/issues/10114

mprakhov:
Make the owner without the community private key behave like an admin (send CommunityEvents instead of directly updating CommunityDescription)

Closes [#11477](https://github.com/status-im/status-desktop/issues/11477)

### DIFF
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -507,14 +507,14 @@ func (o *Community) CreateChat(chatID string, chat *protobuf.CommunityChat) (*Co
 	o.mutex.Lock()
 	defer o.mutex.Unlock()
 
-	isOwner := o.IsOwner()
-	isAdmin := o.IsAdmin()
+	isControlNode := o.IsControlNode()
+	allowedToSendEvents := o.IsOwnerOrAdmin()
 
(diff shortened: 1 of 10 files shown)


jrainville (commented):
Looks good. Could you just add a test that validates an owner without private key ca indeed do admin stuff. Thanks

osmaczko (approved):

0x-r4bbit (approved):

PR #3793, opened 2023-07-24
Move wallet tables to a dedicated DB

### LINKED ISSUE #10964: Wallet dedicated db
Create a dedicated db for wallet information. Mostly data that comes from the blockchain
1 - On account creation -> create the new db
2 - Separate migration for wallet and init the db with current wallet structure
3 - Migrate data
4 - Migration to remove structure from main db
5 - Link wallet to this new db
6 - Handle clean/change password

Currently the choice of db is sqlite/sqlcipher

- [x] Check if statusd/main.go needs changing
- [x] Implement setting addresses for activity.go to avoid attaching main db to query for `keypairs_accounts`

IvanBelyakoff:
## What has been done:

- created a dedicated DB for wallet next to app db with scheme `<keyUID>-wallet.db` and the same password as app DB
- moved wallet-related tables to a dedicated db:
  - blocks
  - blocks_ranges
  - blocks_ranges_sequential
  - pending_transactions
  - saved_addresses
  - token_balances
  - tokens
  - visible_tokens
  - currency_format_cache
  - multi_transactions
  - balance_history
  - price_cache
  - collectibles_ownership_cache
  - transfers

- test fixes:
  - moved test db setup to a common place 't/helpers'
  - created interface for initializing db, which is implemented for appdatabase and
walletdatabase. TBD for multiaccounts DB.
  - unified DB initialisation for all tests using helpers and new interface.
  - reduced sqlcipher kdf iterations for all tests to `1` to improve performance

- sqlite migration improvements. While working on the feature, I tried to do in-code migration first, which led to some refactoring of migration functions. I dropped in-code migration in favor of sql files, because it did not allow standalone wallet db instantiation for tests, but I kept some related code for improvement purposes.
  - some functions split for more cohesion
  - custom PostSteps are stored as pointers to allow their retrieval and parameters passing
in runtime if needed (some extra work is dropped and TBD when needed)

### NOTE:
Data is not copied from app db to wallet db. This mostly is due to some bug (probably in sqlcipher/go-sqlcipher) that caused an error on detaching an attached database after copying data.

Please review single commit at a time for clarity.

Closes [#10964](https://github.com/status-im/status-desktop/issues/10964)


appdatabase/database.go
```diff
@@ -487,3 +507,211 @@ func migrateWalletTransferFromToAddresses(sqlTx *sql.Tx) error {
...
+// 3) detach database did not work after it - DB locked
+
+func copyDataToNewTables(sqlTx *sql.Tx, walletDbName string, tableNamesToMove []string) error {
+	fmt.Println("copyDataToNewTables start")
+
+	for _, tableName := range tableNamesToMove {
+		query := fmt.Sprintf("INSERT INTO %s.%s SELECT * FROM main.%s;", walletDbName, tableName, tableName)
+
+		fmt.Println("query start: ", query)
+		_, err := sqlTx.Exec(query)
```
IvanBelyakoff: This function fails. I thought my test DB (it is my real status app DB) might have too many records, but I tried to use `LIMIT 10` and `OFFSET`, but it did not help - same error `out of memory` during the first iteration.

```
createDatabaseSchema, query:
createDatabaseSchema end
AttachDatabaseWithDefaultSettings start
database is attached success
copyDataToNewTables start
query start:  INSERT INTO wallet.blocks SELECT * FROM main.blocks;
query failed:  INSERT INTO wallet.blocks SELECT * FROM main.blocks;  err: out of memory
    /home/blade/work/projects/Status/status-desktop/vendor/status-go/appdatabase/database_test.go:520:
        	Error Trace:	/home/blade/work/projects/Status/status-desktop/vendor/status-go/appdatabase/database_test.go:520
        	            	failed to rollback migration to version 1689248269: transaction start failed in line 0:  (details: out of memory)
        	Test:       	Test_migrateWalletTablesToDedicatedDB
--- FAIL: Test_migrateWalletTablesToDedicatedDB (1.71s)
FAIL
FAIL	github.com/status-im/status-go/appdatabase	1.732s
```

appdatabase/database.go
```diff
@@ -487,3 +507,213 @@ func migrateWalletTransferFromToAddresses(sqlTx *sql.Tx) error {
...
+		}
+
+		tableNamesWhereClause += "name='" + tableName + "'"
+	}
+
+	return tableNamesWhereClause
+}
+
+// func createAllTablesQueries(sqlTx *sql.Tx, whereClause string) ([]string, error) {
+func createAllTablesQueries(sqlTx *sql.Tx, tableNames []string) ([]string, error) {
```
stefandunca: @IvanBelyakoff I think this table specs copying automation is great. How will the automated copying of the primary key auto-index work? You can't create a "manual" index until you have the tables. However, when you create a table, SQLite automatically indexes the primary key (composed or not). What happens when we copy the previously automatically created indexes?

walletdatabase/migrations/sql_queries/0022_pending_transfers.up.sql
```diff
@@ -0,0 +1,22 @@
+ALTER TABLE pending_transactions RENAME TO pending_transactions_old;
```
stefandunca: How about having only the initial migration that creates the current wallet tables state and one custom migration to copy the data?

stefandunca (commented):
You did a great job of figuring out how to migrate data from one table to another. However, can't see a reason for the out-of-memory error. It might be down to how SQLite manages the memory when mapping two DBs. Or maybe a limitation of some sort with sqlchiper?

Following your work, I think it would be best to manually create the tables using the new wallet migrations scripts in one go. Running the old migrations on the new DB given that we don't have data might be pointless or confusing. This will be some extra work from your side, but as a result, we will have it documented properly in the first new wallet migration scripts.

The scope of custom migration is to do only what is not possible with the migration scripts: copy the data.

In short, I would expect that this custom migration only copies the data and creation of tables and indexes is visible in the first migration script.

IvanBelyakoff:
@stefandunca thank you for your feedback. It is great to receive it at this quite early point. 
I considered the default migration way using sql files, but decided to do this migration in code for several reasons:
- automatic extraction of table creation schemas: While working on this PR, I might miss some changes/migrations of tables, and will need to maintain and update them while working and rebase as soon as some new migration appears - that is troublesome - whether I will have to extract and combine all the migration sql files in a single one of copy all the related sql files to a separate `wallet` directory, which in my opinion only pollutes the project and it is still hard to track all the changes, while we need only the latest ones . With in-code approach I receive always the latest schema for all the tables and the only thing I need to keep track of is my own migration files timestamp.
- Instead of running all the migrations sql file to get to the final version, I run only single CREATE command for each table which is faster, more precise and cleaner approach.
- Splitting this operation into several steps - sql migration files and in-code requires more work, harder to understand and imho does not worth the efforts.

Though I have not quite understoodd the drawbacks that you mentioned of this way of doing it, sorry. If you could elaborate more on that, that would be awesome. We could have a call instead.

appdatabase/database.go
IvanBelyakoff: Stefan, my understanding is that "CREATE INDEX" command will be executed to create an index on table and once data will be copied to the wallet db, the database will index the data. Is it not correct in your opinion? I don't think I copy the indices. I copy the commands to create tables and indices. Then I use the same commands for the new DB.

walletdatabase/migrations/sql_queries/0022_pending_transfers.up.sql
IvanBelyakoff: omg, sorry for confusion, this is a drafty draft only intended to ask you teammates, if you spot any obvious issue why I get out of mem issue, so this was committed by mistake, it wont be there

stefandunca:
> * and will need to maintain and update them while working and rebase as soon *

I agree that the small benefits to have the current configuration snapshot might not be worth the effort though.

> * Instead of running all the migrations sql file to get to the final version, I run only single CREATE command for each table which is faster, more precise and cleaner approach.

I'm not very familiar with this approach so great to understand from you that it is consistent and complete

appdatabase/database.go
stefandunca: I just checked my DB `sqlite_master`, and I see that the `sql` misses the index entries listed by running `PRAGMA index_list('<table_name>');`. Maybe see if there is a good way to keep this using the introspection method.

IvanBelyakoff: @stefandunca Good point, indeed the indices were missing with the command, that I have in code. I actually noticed that, but did not get to it yet, while fixing other issues. The working command for me:
sqlite> SELECT sql FROM sqlite_master WHERE type='index' AND tbl_name="transfers";
```
CREATE INDEX idx_transfers_blk_loaded ON transfers(blk_number, loaded)
CREATE INDEX idx_transfers_filter ON transfers (status, token_address, token_id)
sqlite> 
sqlite> pragma index_list("transfers");
0|idx_transfers_filter|0|c|0
1|idx_transfers_blk_loaded|0|c|0
2|sqlite_autoindex_transfers_1|1|u|0
```
I believe `index_list` does not list auto indices, not created manually. But I will double check with tests that the indices on source and destination DBs match

walletdatabase/migrations/sql/doc.go
```diff
@@ -0,0 +1,3 @@
+package sql
+
+//go:generate go-bindata -pkg migrations -o ../bindata.go ./
```
alaibe, line 3: shall you modify the makefile too that it generate this?

IvanBelyakoff: done, thank you

appdatabase/migrations/sql/1690734500_move_wallet_tables_to_wallet_db.down.sql
```diff
@@ -0,0 +1,160 @@
...
+blk_hash BIGINT NOT NULL,
+loaded BOOL DEFAULT FALSE,
+CONSTRAINT unique_mapping_for_account_to_block_per_network UNIQUE (address,blk_hash,network_id)
+);
+
+CREATE TABLE blocks_ranges (
+network_id UNSIGNED BIGINT NOT NULL,
+address VARCHAR NOT NULL,
+blk_from BIGINT NOT NULL,
+blk_to BIGINT NOT NULL
```
dlipicar: weird formatting here

appdatabase/migrations/sql/1690734500_move_wallet_tables_to_wallet_db.down.sql
```diff
@@ -0,0 +1,160 @@
...
+  gas_price BLOB,
+  gas_limit BLOB,
+  value BLOB,
+  data TEXT,
+  type VARCHAR,
+  additional_data TEXT, multi_transaction_id INT,
+  PRIMARY KEY (network_id, hash)
+) WITHOUT ROWID;
+
+CREATE TABLE "saved_addresses" (
```
dlipicar: I'm guessing the quotes do nothing here, I think it's the first time I've seen them used

appdatabase/migrations/sql/1690734500_move_wallet_tables_to_wallet_db.down.sql
```diff
@@ -0,0 +1,160 @@
...
+
+CREATE TABLE multi_transactions (
+    from_address VARCHAR NOT NULL,
+    from_asset VARCHAR NOT NULL,
+    from_amount VARCHAR NOT NULL,
+    to_address VARCHAR NOT NULL,
+    to_asset VARCHAR NOT NULL,
+    type VARCHAR NOT NULL,
+    timestamp UNSIGNED BIGINT NOT NULL
+, to_amount VARCHAR, from_network_id UNSIGNED BIGINT, to_network_id UNSIGNED BIGINT, cross_tx_id VARCHAR DEFAULT "", from_tx_hash BLOB, to_tx_hash BLOB);
```
dlipicar: not sure if this is github misbehaving, seeing weird formatting here
image

appdatabase/migrations/sql/1690734500_move_wallet_tables_to_wallet_db.down.sql
```diff
@@ -0,0 +1,160 @@
...
+
+CREATE TABLE collectibles_ownership_cache (
+    chain_id UNSIGNED BIGINT NOT NULL,
+    contract_address VARCHAR NOT NULL,
+    token_id BLOB NOT NULL,
+    owner_address VARCHAR NOT NULL
+);
+
+CREATE TABLE transfers (
+network_id UNSIGNED BIGINT NOT NULL,
```
dlipicar: no indentation, as opposed to all other create statements

appdatabase/migrations/sql/1690734500_move_wallet_tables_to_wallet_db.up.sql
```diff
@@ -0,0 +1,18 @@
+DROP TABLE blocks;
+DROP TABLE blocks_ranges;
+DROP TABLE blocks_ranges_sequential;
+DROP TABLE pending_transactions;
+DROP TABLE saved_addresses;
+-- token_balances is the only table that was placed by mistake in nodeconfig
+-- migrations and in tests it is missing if nodeconfig migration is not used.
```
dlipicar: OOC, what's supposed to go in the nodeconfig migrations?

appdatabase/migrations/sql/1690734500_move_wallet_tables_to_wallet_db.down.sql
IvanBelyakoff: Yeah, I actually had a piece of code that just select all create table statements from app db and dumps them into a file. So the formatting comes from the sql files originally. I did not touch it to avoid reformatting on each rebase. But I can do that before I merge

appdatabase/migrations/sql/1690734500_move_wallet_tables_to_wallet_db.down.sql
IvanBelyakoff: That is right, it does nothing, just taken from sql migration file automatically

services/wallet/activity/activity.go
```diff
@@ -896,3 +907,28 @@ func contractTypeFromDBType(dbType string) (transferType *TransferType) {
...
+	}
+
+	addresses, err := accountsDb.GetWalletAddresses()
+	if err != nil {
+		log.Error("failed to get wallet addresses", "err", err)
+		return err
+	}
+
+	for _, address := range addresses {
+		_, err = db.Exec(fmt.Sprintf("INSERT OR IGNORE INTO %s (address) VALUES (?)", keypairAccountsTable), address)
```
dlipicar: possibly more efficient to do a Begin..Commit here

dlipicar (approved):
awesome work!

appdatabase/migrations/sql/1690734500_move_wallet_tables_to_wallet_db.up.sql
IvanBelyakoff: Oh, I should fix the comment. `token_balances` migration file was placed in protocol migrations, not in nodeconfig migrations. As I see prevnodeconfig migration files date earlier and are now superseeded by plain migrations folder. But not sure about `protocol` migrations - probably just related to some waku-related things judging by the code, but in the end, the files are merged and applied in correct order. But in tests if you want to achieve the same, you should do some work I believe

dlipicar (approved):

alaibe (approved):

Makefile
```diff
@@ -392,6 +392,10 @@ migration:
 migration-check:
 	bash _assets/scripts/migration_check.sh
 
+migration-wallet: DEFAULT_WALLET_MIGRATION_PATH := walletdatabase/migrations/sql
```
stefandunca, line 400: Wondering why two entries for `migration-wallet `

services/wallet/activity/activity.go
```diff
@@ -655,7 +659,14 @@ func getActivityEntries(ctx context.Context, deps FilterDependencies, addresses
 		return strconv.Itoa(int(t))
 	})
 
-	queryString := fmt.Sprintf(queryFormatString, involvedAddresses, toAddresses, assetsTokenCodes, assetsERC20, networks,
+	// Since the filter query needs addresses which are in a different database, we need to update the
+	// keypairs_accounts table in the current database with the latest addresses from the accounts database
+	err := updateKeypairsAccountsTable(deps.accountsDb, deps.db)
```
stefandunca, line 664: A good workaround for now, however, will be nice to have a task, for follow-up, linked to this change and remove this requirement entirely and rely on the filtered addresses as we discussed in private.

sqlite/migrate.go
```diff
@@ -36,7 +38,7 @@ var migrationTable = "status_go_" + sqlcipher.DefaultMigrationsTable
 //
 // untilVersion, for testing purposes optional parameter, can be used to limit the migration to a specific version.
 // Pass nil to migrate to the latest available version.
-func Migrate(db *sql.DB, resources *bindata.AssetSource, customSteps []PostStep, untilVersion *uint) error {
+func Migrate(db *sql.DB, resources *bindata.AssetSource, customSteps []*PostStep, untilVersion *uint) error {
```
stefandunca, line 41: Not sure what the intention was here, but if it was about optimizing, I doubt using reference in the slice will lead to any optimizations.

walletdatabase/migrations/sql/1691753758_initial.up.sql
```diff
@@ -0,0 +1,191 @@
+CREATE TABLE blocks (
```
stefandunca, line 1: Excellent reference and summarization of the wallet DB structure requirements.

stefandunca (approved):
Great job. I am looking forward to seeing the results in practice.

sqlite/migrate.go
IvanBelyakoff: Stefan, I had a function `getCustomStepByName(name)` and added parameters field for custom steps. So I passed the parameter to the step in runtime. For it to work I needed a pointer. I dropped some changes, because unused now as I switched to using sql migration files instead,  but left some changes like this one, which by itself does not look as imrovement without the context

Makefile
IvanBelyakoff: @stefandunca I dont understand this comment. Can you elaborate more?

services/wallet/activity/activity.go
IvanBelyakoff: https://github.com/status-im/status-go/issues/3881
Thanx for the reminder, Stefan!

Makefile
stefandunca: I meant `migration-wallet`, wrong copy-pasta

IvanBelyakoff: Stefan, I am not an expert on make, but I remember trying to combine setting a variable and the next `touch` command into a single line or under a single target, but it did not work for me. So probably that is how it should be done to make it work. But if you know how to change that, please share

stefandunca: I have no clue of make, just wondering :) Thanks for clarifying.

PR #3794, opened 2023-07-24
feat: introduce messenger API to remove community tokens

0x-r4bbit:
This is to allow users to remove a community token if the deployment transaction has been rejected.

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -3937,6 +3937,10 @@ func (m *Manager) UpdateCommunityTokenSupply(chainID int, contractAddress string
 	return m.persistence.UpdateCommunityTokenSupply(chainID, contractAddress, supply)
 }
 
+func (m *Manager) RemoveCommunityToken(chainID int, contractAddress string) error {
+	return m.persistence.RemoveCommunityToken(chainID, contractAddress)
+}
+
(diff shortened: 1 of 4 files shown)


mprakhov (approved):
LGTM
Have one question - who will have a possibility to remove CommunityToken?

0x-r4bbit:
This removes the community token from your own database. Everyone can do it. But you only want to do it when you know there's a deployment that went sideways.

Deploying a community token means:

1. Deploy contract
2. Create/add community token to database
3. Wait for transaction to go through
4. Update community token deploy status when tx status changes

If tx fails to deploy, then you end up with a community token entry with deploy status "failed".
In this case you can decide to either "remove" or "retry" the deployment.
This change is purely for removing.

0x-r4bbit:
@mprakhov actually, you might be onto something here.
I just realized that `AddCommunityToken()` not only adds community token metadata to the database but also already publishes it on the community description, even though the community token hasn't been deployed yet (tx might fail, or is still pending).

So... either we also remove the community token from the metadata here, or, if we expect we can only remove token metadata for tokens that haven't been successfully deploy, we shouldn't add them to the community description in the first place.

Will discuss this with @endulab

0x-r4bbit:
Created https://github.com/status-im/status-go/pull/3798 accordingly

0x-r4bbit:
Updated this PR to sit on top of #3798

PR #3795, opened 2023-07-24
feat(wallet) add request id to activity events

stefandunca:
Add and use the optional chainID in the wallet event structure.

Required by https://github.com/status-im/status-desktop/pull/11339

### DIFF
--- a/services/wallet/activity/service.go
+++ b/services/wallet/activity/service.go
@@ -23,20 +23,21 @@ const (
 	EventActivityGetOldestTimestampDone walletevent.EventType = "wallet-activity-get-oldest-timestamp-result"
 )
 
-var (
-	filterTask = async.TaskType{
-		ID:     1,
-		Policy: async.ReplacementPolicyCancelOld,
-	}
-	getRecipientsTask = async.TaskType{
-		ID:     2,
-		Policy: async.ReplacementPolicyIgnoreNew,
-	}
-	getOldestTimestampTask = async.TaskType{
-		ID:     3,
-		Policy: async.ReplacementPolicyCancelOld,
-	}
+const (
+	filterTaskID                 = int32(1)
+	filterTaskPolicy             = async.ReplacementPolicyCancelOld
+	getRecipientsTaskID          = int32(2)
+	getRecipientsTaskPolicy      = async.ReplacementPolicyIgnoreNew
+	getOldestTimestampTaskID     = int32(3)
+	getOldestTimestampTaskPolicy = async.ReplacementPolicyCancelOld
 )
+
+func makeTaskType(requestID int32, originalID int32, policy async.ReplacementPolicy) async.TaskType {
+	return async.TaskType{
+		ID:     int64(requestID)<<32 | int64(originalID),
+		Policy: policy,
+	}
+}
 
 type Service struct {
 	db           *sql.DB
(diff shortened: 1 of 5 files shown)


services/wallet/api.go
```diff
@@ -515,24 +515,24 @@ func (api *API) FetchAllCurrencyFormats() (currency.FormatPerSymbol, error) {
 	return api.s.currency.FetchAllCurrencyFormats()
 }
 
-func (api *API) FilterActivityAsync(ctx context.Context, addresses []common.Address, chainIDs []wcommon.ChainID, filter activity.Filter, offset int, limit int) error {
-	log.Debug("wallet.api.FilterActivityAsync", "addr.count", len(addresses), "chainIDs.count", len(chainIDs), "offset", offset, "limit", limit)
+func (api *API) FilterActivityAsync(requestID int, addresses []common.Address, chainIDs []wcommon.ChainID, filter activity.Filter, offset int, limit int) error {
```
dlipicar: huh I though the context was passed as first argument to every call from the client

stefandunca: It is optional. In this case, we don't do any cancelable processing, after the async conversion, so it makes no sense to pass it.

services/wallet/activity/service.go
```diff
@@ -74,7 +74,7 @@ type FilterResponse struct {
 // FilterActivityAsync allows only one filter task to run at a time
 // and it cancels the current one if a new one is started
 // All calls will trigger an EventActivityFilteringDone event with the result of the filtering
-func (s *Service) FilterActivityAsync(ctx context.Context, addresses []common.Address, chainIDs []w_common.ChainID, filter Filter, offset int, limit int) {
+func (s *Service) FilterActivityAsync(requestID int, addresses []common.Address, chainIDs []w_common.ChainID, filter Filter, offset int, limit int) {
 	s.scheduler.Enqueue(filterTask, func(ctx context.Context) (interface{}, error) {
```
dlipicar: IMO we want async calls from separate requestIDs not to affect each other. As it is, a filter request from the Send modal might cancel a previous request from the activity tab, right?

stefandunca: Right, I missed this; thanks for the heads up.

dlipicar (approved):
Nice work! I'll probably need to use a similar mechanism for the collectibles service. Will probably include the requestID mechanism in the scheduler if it doesn't look too problematic.

services/wallet/activity/service.go
```diff
@@ -22,21 +22,22 @@ const (
...
+	filterTaskPolicy             = ReplacementPolicyCancelOld
+	getRecipientsTaskID          = int32(2)
+	getRecipientsTaskPolicy      = ReplacementPolicyIgnoreNew
+	getOldestTimestampTaskID     = int32(3)
+	getOldestTimestampTaskPolicy = ReplacementPolicyCancelOld
 )
 
+func makeTaskType(requestID int32, originalID int32, policy ReplacementPolicy) TaskType {
+	return TaskType{
+		ID:     int64(requestID)<<32 | int64(originalID),
```
IvanBelyakoff, line 37: Is there any benefit on using single field here instead of using 2? My opinion is that it makes it less obvious for a reader that it has 2 meanings. And reverse operation is implicit right?

stefandunca: > Is there any benefit on using single field here instead of using 2?

Mainly the simplicity of implementation. There is a need for a unique identity for each task that has to be queued. Brings no benefit to having two fields for the scheduler. I see no benefit in passing the request-id responsibility to the generic queued scheduler.

> My opinion is that it makes it less obvious for a reader that it has 2 meanings.

Unicity is a combination of request ID and task ID which is defined by the service requirements, and it is the responsibility of the caller, `activity.Service` in this case.
Extending the scheduler with these two IDs won't bring any benefit outside the bigger space for each identifier that is unnecessary for this implementation. I mean both ID types have local meanings: request ID represents the unique clients (two for now) and task ID represents the unique tasks supported by the service (three for now). 

> And reverse operation is implicit right?

Because of the only unique identities requirements, there is no need to extract back the values, they just need to provide uniqueness; see `service_test.go/Test_makeTaskType`

Does it make sense? How do you think I can clarify this without complicating the implementation?

Issue #3719, opened 2023-07-05
Avoid getting/storing ourselves as contact


igor-sirotin:
# Problem

As found out in https://github.com/status-im/status-go/issues/3667, we were loading (and if not found, storing) ourselves to `ContactsMap`, which shouldn't be done.
I've added required conditions in https://github.com/status-im/status-go/pull/3627, but also tracked down all places we were doing this.

To gather this info I've added callstack logging (04675d8ab3f764d69a705912d2ea3e9b1b2de942) and a python script to analyse output of `make test`.
<details>
  <summary>Show me the python script</summary>
  
  ```python
import json
import re


class SetEncoder(json.JSONEncoder):
    def default(self, obj):
        if isinstance(obj, set):
            return list(obj)
        return json.JSONEncoder.default(self, obj)

file_name = 'vendor/status-go/tests_5.log'
file1 = open(file_name, 'r')
lines = file1.readlines()
file1.close()

full_dict = {}
loads_count = 0
stacks_dict = {}

for index, line in enumerate(lines):

    # FIXME: Fix logger format difference
    #        Example at lines: {47540, 47556, 47570, 47630, 47644, 47658, 47770, 47779, 47803, 47806, 47807}
    #        diff = set(loads.keys()) ^ set(loads2.keys())
    match_all = re.match(r'.*contacts map: loading own identity.*', line)
    if match_all is None:
        continue

    loads_count += 1
    m = re.match(r'.*contacts map: loading own identity.*"stack": (?P<stack>.*)}', line)
    if m is None:
        continue

    stack_json = m.group('stack')
    stack = json.loads(stack_json)
    full_dict[index] = stack

    stack_key = []
    for stack_item in stack[1:4]:
        stack_key.append(stack_item['file'])
    stack_key_string = json.dumps(stack_key)

    if stack_key_string not in stacks_dict:
        stacks_dict[stack_key_string] = {'count': 1, 'last': stack, 'message_types': set()}
    else:
        stacks_dict[stack_key_string]['count'] += 1
        stacks_dict[stack_key_string]['last'] = stack
        # Analyze previous line for "processing message"
        prev_line = lines[index - 1]
        m2 = re.match(r'.*processing message.*{"type": "(?P<message_type>.*)", "senderID".*', prev_line)
        if m2 is not None:
            message_type = m2.group('message_type')
            stacks_dict[stack_key_string]['message_types'].add(message_type)

print("Total lines found:   {}".format(len(lines)))
[...]

PR #3799, opened 2023-07-25
Fix syncing `blocked` contacts

### LINKED ISSUE #3804: When unblocked, we show 'Alice removed you as contact'
# Problem

After Bob's unblocked by Alice, he gets a `@Alice removed you as contact` system message in response, just as he was blocked.

Reproducible in tests: https://github.com/status-im/status-go/pull/3799

# Acceptance Criteria

Not sure if we should generate any system message here at all.
But there definitely shouldn't be such message.

igor-sirotin:
Fixes https://github.com/status-im/status-go/issues/3804
Required for https://github.com/status-im/status-desktop/issues/11523

1. Test for blocking/unblocking contact from paired devices
2. Fix syncing contact `blocked` state
3. Fix showing `Alice removed you as contact` when unblocking

Related bugs:
- https://github.com/status-im/status-go/issues/3800
- https://github.com/status-im/status-go/issues/3801
- https://github.com/status-im/status-go/issues/3803
- https://github.com/status-im/status-go/issues/3804

Though the test will currently pass, there's a bunch of `fixme`/`warning` to be updated when the bugs above fixed.


igor-sirotin:
- [x] clean up `sync_device_test.go`

protocol/messenger_contacts.go
```diff
@@ -810,71 +809,82 @@ func (m *Messenger) blockContact(response *MessengerResponse, contactID string,
...
-		return nil, nil, err
+		return err
 	}
 
-	return contact, chats, nil
+	return nil
 }
 
-func (m *Messenger) BlockContact(contactID string) (*MessengerResponse, error) {
+func (m *Messenger) BlockContact(contactID string, syncing bool) (*MessengerResponse, error) {
```
MishkaRogachev: that's really good idea!

protocol/messenger.go
```diff
@@ -3544,8 +3544,16 @@ func (m *Messenger) handleRetrievedMessages(chatWithMessages map[transport.Filte
 				}
 
 				senderID := contactIDFromPublicKey(publicKey)
+				ownID := contactIDFromPublicKey(m.IdentityPublicKey())
 				m.logger.Info("processing message", zap.Any("type", msg.Type), zap.String("senderID", senderID))
 
+				if senderID == ownID {
```
MishkaRogachev, line 3560: Can it be combined in one statement?

MishkaRogachev (approved):
Great!

protocol/messenger.go
igor-sirotin: Weell yes, but I left them separate intentionally, because there will probably be more message types to skip later.

Related to https://github.com/status-im/status-go/issues/3719.

protocol/messenger_contacts.go
Samyoul: Great job. Do you think that the purpose of this bool would be clearer if it was called something like `fromSyncingMessage` or `fromSync` or `shouldSync`? Just so other developers know the context a bit better.

Samyoul (approved):
This looks good. Thank you for doing this work 🙏

protocol/messenger_contacts.go
igor-sirotin: Yeah, that makes sense 👍 I just copied it from other places we have.

TODO:
- [x] rename `syncing` flag

PR #3802, opened 2023-07-25
feat: Collecting community messages count metrics

MishkaRogachev:
For https://github.com/status-im/status-desktop/issues/11152

Implement both strategies for fetching community messaging metrics:

- [x] `CommunityMetricsRequestMessagesTimestamps` to collect messages timestamps by intervals for all community chats
- [x] `CommunityMetricsRequestMessagesCount` to collect messages count by intervals for all community chats

### DIFF
--- a/protocol/messenger_community_metrics.go
+++ b/protocol/messenger_community_metrics.go
@@ -2,6 +2,7 @@ package protocol
 
 import (
 	"errors"
+	"fmt"
 	"sort"
 
 	"github.com/status-im/status-go/eth-node/types"
@@ -39,8 +40,8 @@ func (m *Messenger) collectCommunityMessagesTimestamps(request *requests.Communi
 		return nil, err
 	}
 
-	intervals := []MetricsIntervalResponse{}
-	for _, sourceInterval := range request.Intervals {
+	intervals := make([]MetricsIntervalResponse, len(request.Intervals))
+	for i, sourceInterval := range request.Intervals {
 		// TODO: messages count should be stored in special table, not calculated here
 		timestamps, err := m.persistence.SelectMessagesTimestampsForChatsByPeriod(chatIDs, sourceInterval.StartTimestamp, sourceInterval.EndTimestamp)
 		if err != nil {
@@ -50,11 +51,11 @@ func (m *Messenger) collectCommunityMessagesTimestamps(request *requests.Communi
 		// there is no built-in sort for uint64
 		sort.Slice(timestamps, func(i, j int) bool { return timestamps[i] < timestamps[j] })
 
-		intervals = append(intervals, MetricsIntervalResponse{
+		intervals[i] = MetricsIntervalResponse{
 			StartTimestamp: sourceInterval.StartTimestamp,
 			EndTimestamp:   sourceInterval.EndTimestamp,
 			Timestamps:     timestamps,
-		})
+		}
 	}
 
 	response := &CommunityMetricsResponse{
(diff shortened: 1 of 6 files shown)


protocol/persistence_metrics.go
```diff
@@ -0,0 +1,76 @@
...
+		if err == nil {
+			err = tx.Commit()
+			return
+		}
+		// don't shadow original error
+		_ = tx.Rollback()
+	}()
+
+	var timestamps []uint64
+	for _, chatID := range chatIDs {
```
MishkaRogachev: Here I get all the chat id's for the community, but I can probably rely on the fact that the channel id is communityId+uuid, and so I can just select messages with the chat id starting with the community id

alexjba (approved):
LGTM! 👍

jrainville (approved):
Nice job. This is very clean

protocol/messenger_community_metrics.go
```diff
@@ -0,0 +1,111 @@
...
+	return community.ChatIDs(), nil
+}
+
+func (m *Messenger) collectCommunityMessagesTimestamps(request *requests.CommunityMetricsRequest) (*CommunityMetricsResponse, error) {
+	chatIDs, err := m.getChatIdsForCommunity(request.CommunityID)
+	if err != nil {
+		return nil, err
+	}
+
+	intervals := []MetricsIntervalResponse{}
```
ilmotta: Whenever possible, it's a good idea to initialize slices with a max size (2nd arg to `make`). In this case the length of `request.Intervals`

protocol/persistence_metrics.go
```diff
@@ -0,0 +1,58 @@
...
+	query := fmt.Sprintf(selectTimestampsQuery, querySeveralChats(chatIDs))
+
+	rows, err := db.db.Query(query, startTimestamp, endTimestamp)
+	if err != nil {
+		return []uint64{}, err
+	}
+	defer rows.Close()
+
+	var timestamps []uint64
+	for rows.Next() {
```
ilmotta, line 34: The section `Fetching Data from the Database` is quite useful, as is the whole page actually http://go-database-sql.org/retrieving.html. It says we should ideally always check for errors after looping with `rows.Err()`. It's not enough to just `defer rows.Close()`.

protocol/requests/community_metrics_request.go
```diff
@@ -0,0 +1,44 @@
...
+	EndTimestamp   uint64 `json:"endTimestamp"`
+}
+
+type CommunityMetricsRequest struct {
+	CommunityID types.HexBytes              `json:"communityId"`
+	Type        CommunityMetricsRequestType `json:"type"`
+	Intervals   []MetricsIntervalRequest    `json:"intervals"`
+}
+
+func (r *CommunityMetricsRequest) Validate() error {
```
ilmotta, line 32: We could have some tests for the validations, but it's fine without since it's fairly simple.

ilmotta (approved):
Very well done @MishkaRogachev! 🚀

protocol/messenger_community_metrics.go
MishkaRogachev: fixed

protocol/persistence_metrics.go
MishkaRogachev: fixed

protocol/requests/community_metrics_request.go
MishkaRogachev: added small test case checking invalid request intervals

PR #3806, opened 2023-07-26
feat(mentions): add callID for `OnChangeText`

qfrank:
relate mobile [PR1](https://github.com/status-im/status-mobile/pull/16780) [PR2](https://github.com/status-im/status-mobile/pull/21171)

### DIFF
--- a/protocol/messenger_mention.go
+++ b/protocol/messenger_mention.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+	"sync"
 	"unicode"
 	"unicode/utf8"
 
@@ -167,6 +168,10 @@ type ChatMentionContext struct {
 	MentionState       *MentionState
 	PreviousText       string // user input text before the last change
 	NewText            string
+
+	CallID    uint64
+	mu        *sync.Mutex
+	MaxCallID uint64
 }
 
 func NewChatMentionContext(chatID string) *ChatMentionContext {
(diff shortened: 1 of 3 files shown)


vkjr (approved):

ilmotta (approved):
Interestingly @vkjr @qfrank, I solved a similar problem at the re-frame event layer for generating link previews. Basically, by the time any re-frame event is dispatched, it is processed asynchronously in an event queue, and by the time status-go response arrives to the client, new events may have been dispatched already.

So in effect, the pattern I used is a known one, which is to use a unique request ID to allow event de-duplication. The client only changes the application state (aka app-db) if the response ID is the same as the originating request ID. One difference is that I used a UUID to generate unique request IDs.

If you are curious, here's how I implemented this (very roughly I must say) in status-mobile https://github.com/status-im/status-mobile/blob/238e35a281a108ec2cac2091ca1f1ef84e106e80/src/status_im2/contexts/chat/composer/link_preview/events.cljs#L43

There are interesting ways to create a general solution in the client side too, because high-frequency client events that happen faster than the backend can reply *can usually* suffer from the problem this PR solved and the problem I solved on the client.

Well, thanks for reading this much :)

igor-sirotin:
@qfrank is this PR still needed?

qfrank:
> @qfrank is this PR still needed?

I'd say yes and now one of our focus is stability, this will improve the stability of the application. Time to merge after manual QA @igor-sirotin

igor-sirotin:
> I'd say yes

@qfrank can you please rebase it then and request manual QA?

qfrank:
> > I'd say yes
> 
> @qfrank can you please rebase it then and request manual QA?

ofc, I planned today, but I got something more important to do today, I'll do it tomorrow

PR #3808, opened 2023-07-26
chore: deprecate InviteUsersToCommunity

osmaczko:
Add deprecation info to InviteUsersToCommunity

### DIFF
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -1870,6 +1870,8 @@ func (m *Messenger) ImportCommunity(ctx context.Context, key *ecdsa.PrivateKey) 
 	return response, nil
 }
 
+// Deprecated: Community invites are no longer sent to users.
+// Instead, the community is just shared and access requests is required from users.
 func (m *Messenger) InviteUsersToCommunity(request *requests.InviteUsersToCommunity) (*MessengerResponse, error) {
 	if err := request.Validate(); err != nil {
 		return nil, err
--- a/services/ext/api.go
+++ b/services/ext/api.go
@@ -481,6 +481,9 @@ func (api *PublicAPI) DeleteCommunityChat(communityID types.HexBytes, chatID str
 }
 
 // InviteUsersToCommunity invites the users with pks to the community with ID
+//
+// Deprecated: Community invites are no longer sent to users.
+// Instead, the community is just shared and access requests is required from users.
 func (api *PublicAPI) InviteUsersToCommunity(request *requests.InviteUsersToCommunity) (*protocol.MessengerResponse, error) {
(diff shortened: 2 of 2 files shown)


osmaczko:
@Samyoul do you use `InviteUsersToCommunity` on mobile any longer? Maybe we could remove it altogether.

0x-r4bbit (approved):
Also +1 for removing API if mobile doesn't use it (anymore)

Samyoul:
@osmaczko Thank you, looking at the `status-mobile` code we only seem to have an adaptor function for this API method:

```clojure
(rf/defn invite-users
  {:events [::invite-people-confirmation-pressed]}
  [cofx user-pk contacts]
  (let [community-id (fetch-community-id-input cofx)
        pks          (if (seq user-pk)
                       (conj contacts user-pk)
                       contacts)]
    (when (seq pks)
      {:json-rpc/call [{:method      "wakuext_inviteUsersToCommunity"
                        :params      [{:communityId community-id
                                       :users       pks}]
                        :js-response true
                        :on-success  #(re-frame/dispatch [::people-invited %])
                        :on-error    #(do
                                        (log/error "failed to invite-user community" %)
                                        (re-frame/dispatch [::failed-to-invite-people %]))}]})))
```

https://github.com/status-im/status-mobile/blob/109061a112795e32d59052e138e4b4aced5e74d2/src/status_im/communities/core.cljs#L329-L345

The event `invite-people-confirmation-pressed` is only triggered once, but only in the old UI https://github.com/status-im/status-mobile/blob/develop/src/status_im/ui/screens/communities/invite.cljs#L87

However for dealing with deprecated functions we sure mark them `Deprecated` and ultimately removed them once they are no longer being used. We have a kind of tiered deprecation in `status-go`:

- `A`functionality we no longer use and will never use again, have no internal or external dependency.
  - ✅ Safe to remove
- `B` functionality that has become redundant due to other functionality
  - ✅ Safe to remove
- `C` functionality that is partially used and could / should be replace with a better replacement
- `D` functionality we've previously implemented but want to develop further or keep around
- `E` functionality we've introduced but been asked to remove, but subsequently asked to reimplement, then asked to remove again 🙂 
  - Or functionality that is fully formed and we suspect may be asked to be reintroduced 
  - See @igor-sirotin's PR here https://github.com/status-im/status-go/pull/3809 for an example of this case.

I think in this particular case with this PR's deprecation this is some mix of `A` and `B`, so we can open a PR on `status-mobile` and remove it safely.

osmaczko:
@Samyoul thank you very much for detailed description. I will create a separate PR with the removal of `InviteUsersToCommunity` which has been deprecated for less than a week :sweat_smile:

PR #3809, opened 2023-07-26
fx: deprecate profile and timeline chats

### LINKED ISSUE #3800: Deprecate profile/timeline chats
# Problem

Although `ChatTypeProfile` and `ChatTypeTimeline` are deprecated, they still appear in UI somehow (https://github.com/status-im/status-desktop/issues/11523). Instead of fixing that bug directly, would be better to fairly deprecate these chat types and stop creating them 🙂 

# Implementation

As discussed with @Samyoul, we should deprecate them in the following way:
1. Mark as deprecated in the standard Go way
2. `return nil` in the beginning of functions to be deprecated.
This would protect us from using deprecated code.
3. The code is not removed in case we want to reincarnate such functionality.

# Acceptance Criteria

Types `ChatTypeProfile`/`ChatTypeTimeline` and all related code is marked as deprecated.
No profile/timeline chats is created.

igor-sirotin:
Closes https://github.com/status-im/status-go/issues/3800

Profile and timeline chats have been not supported for some time already, but were still created.
- Marked all related enums/variables/functions as deprected
- Introduce new `deprecation` package with deprecation flags
- Added early-return `nil` in deprecated functions
- Hide calls to deprecated functions with corresponding deprecation flag
- Update tests checking chats


Samyoul:
I can see that you've deprecation of all the functions related to profile and timeline chat, I think that is appropriate.
We also have the API function that exposes `CreateProfileChat`

```go
func (api *PublicAPI) CreateProfileChat(parent context.Context, request *requests.CreateProfileChat) (*protocol.MessengerResponse, error) {
    return api.service.messenger.CreateProfileChat(request)
}
```

Which I think is the main function we should deprecate

Samyoul:
The early returns as you've said are causing a problem with the linter, and we can resolve this and the function not used lint error with a deprecation flag.
Example:

```go
const (
  ChatProfileDeprecated = true
  ChatTimelineDeprecated = true
)
```

```go
// Timeline and profile chats are deprecated.
// This code can be removed after some reasonable time.

// upsert timeline chat
if !ChatTimelineDeprecated {
  err = m.ensureTimelineChat()
    if err != nil {
    return err
  }
}

// upsert profile chat
if !ChatProfileDeprecated {
  err = m.ensureMyOwnProfileChat()
  if err != nil {
    return err
  }
}
```

We can apply the same logic anywhere you've commented out code. This way we can avoid the function not used lint issue
Early returns can be updated like below

```go
// Deprecated: profile chats are no more supported
func (m *Messenger) ensureMyOwnProfileChat() error {
    // Return error to prevent usage of deprecated function
    if ChatProfileDeprecated {
        return errors.New("profile chats are deprecated")
    }

    chatID := common.PubkeyToHex(&m.identity.PublicKey)
    _, ok := m.allChats.Load(chatID)
    if ok {
        return nil
    }
    chat := m.buildProfileChat(chatID)
    chat.Active = true
    // Save topics
    _, err := m.Join(chat)
    if err != nil {
        return err
    }
    return m.saveChat(chat)
}
```

Samyoul:
Ah one other thing that I'm trying ensure we introduce everywhere is correct Go documentation comments. See [Go Doc Comments - The Go Programming Language](https://tip.golang.org/doc/comment)

Basically for the sake of this PR would you mind updating the deprecation notices to follow the pattern:

```go
// Deprecated: <func / var name> <message>
```

Example:
```go
// Deprecated: ensureMyOwnProfileChat profile chats are no more supported
```

No need to add function descriptions beyond what you've already written, that is definitely not your responsibility, but adding the func / var name allows Go tooling to identify that the comment is a documentation comment about a specific func / var.

Samyoul (approved):
This is nice work, thank you.

PR #3812, opened 2023-07-26
feat(communities): add supply to CommunityTokenMetadata

jrainville:
Needed for https://github.com/status-im/status-desktop/issues/11437


protocol/communities/manager.go
```diff
@@ -3920,6 +3920,8 @@ func (m *Manager) AddCommunityToken(communityID string, chainID int, address str
 		TokenType:         token.TokenType,
 		Name:              token.Name,
 		Decimals:          uint32(token.Decimals),
+		Supply:            token.Supply.String(),
+		InfiniteSupply:    token.InfiniteSupply,
```
0x-r4bbit, line 3924: I believe, originally, we decided not to put all of that information into the metadata as we can recover it from the contract on-chain.
Just out of curiosity, what's the reason we're putting these back in?
Probably bandwidth/performance?

If we land this, this could simplify the work done in https://github.com/status-im/status-go/pull/3730 as we then don't need to reach out to chain nodes anymore (provided we'd put `remoteDestruct` and `transferable` in there as well)

jrainville: To be honest, I'm not sure. I saw that we used the CommunityTokenMetadata to generate the token list on the front-end, but later I found that we also have `getCommunityTokens` that gets the tokens from another table in the DB. Maybe I can reuse that instead? I'll investigate and report back when I know

jrainville: Ok, I can use `getCommunityTokens` instead. Using the `CommunityTokenMetadata` to generate the token list wasn't a good source of truth, so I'm using `getCommunityTokens` instead and it works great. I'll close this PR

PR #3813, opened 2023-07-27
Fix/no such column FALSE

qfrank:
when creating account on ios simulator, we got following error:
```
lvl=eror msg="failed to create account" error="no such column: FALSE"
```
this PR fixed this error.


saledjenic (approved):

saledjenic:
@qfrank again iOS issue? Thanks for the fix.

PR #3814, opened 2023-07-27
chore: include token masters in recipients community requests

0x-r4bbit:
Token masters will contribute in accepting/rejecting member requests the same way admins do, so users will have to send requests to them as well.

### DIFF
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -1217,6 +1217,8 @@ func tokenMasterRolePermissions() map[protobuf.CommunityMember_Roles]bool {
 func (o *Community) MemberRole(pubKey *ecdsa.PublicKey) protobuf.CommunityMember_Roles {
 	if o.IsMemberOwner(pubKey) {
 		return protobuf.CommunityMember_ROLE_OWNER
+	} else if o.IsMemberTokenMaster(pubKey) {
+		return protobuf.CommunityMember_ROLE_TOKEN_MASTER
 	} else if o.IsMemberAdmin(pubKey) {
 		return protobuf.CommunityMember_ROLE_ADMIN
 	} else if o.CanManageUsers(pubKey) {


protocol/communities/community.go
```diff
@@ -1189,15 +1201,29 @@ func canDeleteMessageForEveryonePermissions() map[protobuf.CommunityMember_Roles
 	return roles
 }
 
+func (o *Community) GetAdminsAndTokenMasters() []*ecdsa.PublicKey {
+	return append(o.GetMemberTokenMasters(), o.GetMemberAdmins()...)
+}
+
 func (o *Community) GetMemberAdmins() []*ecdsa.PublicKey {
```
mprakhov: I think it is better to name it
GetAdminMembers
The same for GetTokenMasterMembers

protocol/messenger_communities.go
```diff
@@ -907,9 +907,8 @@ func (m *Messenger) RequestToJoinCommunity(request *requests.RequestToJoinCommun
 		return nil, err
 	}
 
-	// send request to join also to community admins
-	communityAdmins := community.GetMemberAdmins()
-	for _, communityAdmin := range communityAdmins {
+	adminsAndTokenMasters := community.GetAdminsAndTokenMasters()
```
mprakhov: what about owners without commuunity private key?

mprakhov (approved):
LGTM, few comments

protocol/messenger_communities.go
0x-r4bbit: Good point, will add!
Will they be able to handle requests to join atm?

mprakhov: Not fully understood the question, but the owner without a community private key will behave the same as the admin

protocol/messenger_communities.go
```diff
@@ -924,9 +924,8 @@ func (m *Messenger) RequestToJoinCommunity(request *requests.RequestToJoinCommun
 		return nil, err
 	}
 
-	// send request to join also to community admins
-	communityAdmins := community.GetMemberAdmins()
-	for _, communityAdmin := range communityAdmins {
+	adminsAndTokenMasters := community.GetPrivilegedMembers()
```
0x-r4bbit: @mprakhov this includes owners as well now

protocol/communities/community.go
```diff
@@ -1209,15 +1221,33 @@ func canDeleteMessageForEveryonePermissions() map[protobuf.CommunityMember_Roles
...
+func (o *Community) GetAdminMembers() []*ecdsa.PublicKey {
+	return o.getMembersByPredicate(o.IsMemberAdmin)
+}
+
+func (o *Community) GetTokenMasterMembers() []*ecdsa.PublicKey {
+	return o.getMembersByPredicate(o.IsMemberTokenMaster)
+}
+
+func (o *Community) GetOwnerMembers() []*ecdsa.PublicKey {
+	return o.getMembersByPredicate(o.IsMemberOwner)
```
mprakhov, line 1237: do we need to filter out the owner, who is the control node in order not to send 2 requests to join?

0x-r4bbit:
There's a test that has been failing a bunch of times on CI but I couldn't reproduce locally yet. Rerunning CI tests again

0x-r4bbit:
This is now failing with another test.
So seems like this is flaky.

Need to investigate this more.
Only idea I have is that this could cause a problem because we're now sending the request to join to 
-> control node
-> then, not only admins, but also owners (which we don't know if they are control node as well)

protocol/communities/community.go
```diff
@@ -1234,6 +1236,31 @@ func canDeleteMessageForEveryonePermissions() map[protobuf.CommunityMember_Roles
...
+func (o *Community) getMembersByPredicate(filterFn predicate) []*ecdsa.PublicKey {
+	result := make([]*ecdsa.PublicKey, 0)
+	members := o.GetMemberPubkeys()
+	for _, member := range members {
+		if filterFn(member) {
+			result = append(result, member)
+		}
+	}
+	return result
+}
```
mprakhov: this is not needed anymore
I added `GetPrivilegedMembers` which will do the same

0x-r4bbit: Removed!

PR #3815, opened 2023-07-27
chore: don't send membership requests to admins of auto-accept communities

0x-r4bbit:
No need to send requests to admins if owner will auto accept anyways.

### DIFF
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -1217,6 +1217,8 @@ func tokenMasterRolePermissions() map[protobuf.CommunityMember_Roles]bool {
 func (o *Community) MemberRole(pubKey *ecdsa.PublicKey) protobuf.CommunityMember_Roles {
 	if o.IsMemberOwner(pubKey) {
 		return protobuf.CommunityMember_ROLE_OWNER
+	} else if o.IsMemberTokenMaster(pubKey) {
+		return protobuf.CommunityMember_ROLE_TOKEN_MASTER
 	} else if o.IsMemberAdmin(pubKey) {
 		return protobuf.CommunityMember_ROLE_ADMIN
 	} else if o.CanManageUsers(pubKey) {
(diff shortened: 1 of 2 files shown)


protocol/messenger_communities.go
```diff
@@ -924,11 +924,13 @@ func (m *Messenger) RequestToJoinCommunity(request *requests.RequestToJoinCommun
...
-		_, err := m.sender.SendPrivate(context.Background(), communityAdmin, &rawMessage)
-		if err != nil {
-			return nil, err
+	if !community.AcceptRequestToJoinAutomatically() {
+		adminsAndTokenMasters := community.GetPrivilegedMembers()
+		for _, communityAdmin := range adminsAndTokenMasters {
+			_, err := m.sender.SendPrivate(context.Background(), communityAdmin, &rawMessage)
+			if err != nil {
+				return nil, err
+			}
```
mprakhov, line 933: Here you do not send requests to join as a sender, I think we need also to filter out messages on the HandleRequestToJoin side

protocol/messenger_communities.go
```diff
@@ -924,11 +924,13 @@ func (m *Messenger) RequestToJoinCommunity(request *requests.RequestToJoinCommun
 		return nil, err
 	}
 
-	adminsAndTokenMasters := community.GetPrivilegedMembers()
-	for _, communityAdmin := range adminsAndTokenMasters {
-		_, err := m.sender.SendPrivate(context.Background(), communityAdmin, &rawMessage)
-		if err != nil {
-			return nil, err
```
mprakhov: What if someone will modify his client and send you such a message?
I think we need additionally to filter out messages on client side (HandleRequestToJoin)

0x-r4bbit: Yes, need to drop the message as discussed

mprakhov (approved):

PR #3816, opened 2023-07-27
fix(messenger): fix handling reply with only one image

jrainville:
Needed for https://github.com/status-im/status-desktop/issues/11695 and https://github.com/status-im/status-mobile/issues/16800

If a message is sent with only 1 image, the album is not generated (no albumID), so then, in the notification handling code, it didn't use the right ID, because it thought it had to use the AlbumID for the message ID


borismelnik (approved):
👍

ilmotta (approved):
Fix makes sense, thanks for the explanation and for improving test coverage.

PR #3819, opened 2023-07-28
fix: owner without community private key and token master was not able to send some admin events

mprakhov:
Unit tests for the owner without a community private key and for TokenMaster were configured incorrectly, so the tests were false positives
Fixed tests and failed logic

### DIFF
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -809,7 +809,7 @@ func (o *Community) RemoveUserFromOrg(pk *ecdsa.PublicKey) (*protobuf.CommunityD
 		return nil, ErrNotAdmin
 	}
 
-	if allowedToSendEvents && o.IsMemberOwnerOrAdmin(pk) {
+	if !isControlNode && o.IsPrivilegedMember(pk) {
 		return nil, ErrCannotRemoveOwnerOrAdmin
 	}
 
@@ -879,7 +879,7 @@ func (o *Community) BanUserFromCommunity(pk *ecdsa.PublicKey) (*protobuf.Communi
 		return nil, ErrNotAdmin
 	}
 
-	if allowedToSendEvents && o.IsMemberOwnerOrAdmin(pk) {
+	if !isControlNode && o.IsPrivilegedMember(pk) {
 		return nil, ErrCannotBanOwnerOrAdmin
 	}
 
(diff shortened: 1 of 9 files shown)


protocol/communities/community.go
```diff
@@ -809,7 +809,7 @@ func (o *Community) RemoveUserFromOrg(pk *ecdsa.PublicKey) (*protobuf.CommunityD
 		return nil, ErrNotAdmin
 	}
 
-	if allowedToSendEvents && o.IsMemberOwnerOrAdmin(pk) {
+	if !isControlNode && o.IsPrivilegedMember(pk) {
```
jrainville, line 812: The second part of the condition is not needed. As soon as we are not a control node, we cannot do this I think.

protocol/communities/community.go
```diff
@@ -879,7 +879,7 @@ func (o *Community) BanUserFromCommunity(pk *ecdsa.PublicKey) (*protobuf.Communi
...
```
jrainville, line 882: Same here

protocol/communities/community_event.go
```diff
@@ -341,8 +341,8 @@ func (o *Community) updateCommunityDescriptionByCommunityEvent(communityEvent Co
 			return err
 		}
 
-		if o.IsMemberOwnerOrAdmin(pk) {
-			return errors.New("attempt to kick an owner or admin of the community from the admin side")
+		if !o.IsControlNode() && o.IsPrivilegedMember(pk) {
```
jrainville, line 344: same

protocol/communities/community_event.go
```diff
@@ -353,8 +353,8 @@ func (o *Community) updateCommunityDescriptionByCommunityEvent(communityEvent Co
 			return err
 		}
 
-		if o.IsMemberOwnerOrAdmin(pk) {
-			return errors.New("attempt to ban an owner or admin of the community from the admin side")
+		if !o.IsControlNode() && o.IsPrivilegedMember(pk) {
```
jrainville, line 356: same

protocol/communities/manager.go
```diff
@@ -1327,8 +1327,8 @@ func (m *Manager) HandleCommunityEventsMessage(signer *ecdsa.PublicKey, message
 		return nil, ErrOrgNotFound
 	}
 
-	if !community.IsMemberAdmin(signer) {
-		return nil, errors.New("user is not an admin")
+	if !community.IsPrivilegedMember(signer) {
+		return nil, errors.New("user has not permissions to send events")
```
jrainville, line 1331:
```suggestion
		return nil, errors.New("user does not have permissions to send events")
```

protocol/communities/community.go
mprakhov: this is the fix, this is not changing the behavior when only the control node can manage users. For this, we have a separate ticket and it requires some additional implementation

jrainville: Let's say a normal member tries to do this action, wouldn't this condition let them do it, since they are not priviledged?

mprakhov: no, because a few lines upper we have a check if it is a control node or a privileged user 
https://github.com/status-im/status-go/pull/3819/files#diff-6d2a8416a631183a7cbd5f7919c6afa15e8576af6be1af1e3f9e5f6f1466516dR805-R810

jrainville: aaah ok, got it. In a way, we could still get rid of the second condition because it's a bit redundant, but whatever.

jrainville (approved):

PR #3821, opened 2023-07-29
[status-mobile-16457] Fix messages containing Giphy URLs not being sent

### LINKED ISSUE #16457: Messages containing Giphy/Tenor links are not delivered to receiver
Originally reported here https://github.com/status-im/status-mobile/pull/15891#issuecomment-1552899794

Example of giphy link: https://giphy.com/gifs/this-is-fine-QMHoU66sBXqqLqYvGO
Example of tenor link: https://tenor.com/view/wolf-of-wall-street-media-leonardo-di-caprio-tapping-table-happy-gif-17819377

**Steps:** 

1. User A sends given link in chat
2. See if User B received this link

**Actual result:** link is not delivered to User B. Any message containing such link will not be delivered to receiver. 

https://github.com/status-im/status-mobile/assets/97245802/f23bd88d-7cfb-43b4-9628-f261b52972c4

### Additional Information

- Status version: nightly 

ibrkhalil:
When trying to calculate an image for a preview using incorrect field names, It resulted in the message not to be sent

Closes [#16457](https://github.com/status-im/status-mobile/issues/16457) on status-mobile


protocol/linkpreview/linkpreview.go
```diff
@@ -130,7 +130,7 @@ type OEmbedUnfurler struct {
 
 type OEmbedResponse struct {
 	Title        string `json:"title"`
-	ThumbnailURL string `json:"thumbnail_url"`
```
ilmotta, line 133: The rename seems unnecessary, in fact the struct tags here are unnecessary because `OEmbedResponse` is not encoded as JSON anywhere.

protocol/linkpreview/linkpreview.go
```diff
@@ -256,6 +257,21 @@ func newUnfurler(logger *zap.Logger, httpClient http.Client, url *neturl.URL) Un
...
 		}
+	case "giphy.com", "media.giphy.com":
+		return OEmbedUnfurler{
+			oembedEndpoint: "https://giphy.com/services/oembed",
+			url:            url,
+			logger:         logger,
+			httpClient:     httpClient,
+		}
+	case "tenor.com", "media.tenor.com":
+		// TODO: Investigate a way to unfurl previews that contain iframes (Webviews?)
```
ilmotta: I would say this TODO should be removed. This question can be answered & implemented by a proper issue. The answer is yes @ibrkhalil. The old implementation used to do this ugly hack for Twitter URLs, i.e. extract metadata from the `html` field returned in the oembed response, and there's no need for iframes or webviews.

It's a sad story that certain providers like Twitter, Reddit and others don't easily expose the metadata in the oEmbed response. They really want people to inject their HTML snippets into websites so they can load their `script` tags...

protocol/linkpreview/linkpreview_test.go
```diff
@@ -207,6 +207,54 @@ func Test_UnfurlURLs_YouTube(t *testing.T) {
...
+			"width": 400,
+			"author_name": "Happy Place",
+			"author_url": "https://giphy.com/happyplaceshow",
+			"provider_name": "GIPHY",
+			"provider_url": "https://giphy.com/",
+			"type": "photo"
+		}
+		`),
+	)
+	transport.fallbackToDefaultTransport = true
```
ilmotta: The idea is to not commit `fallbackToDefaultTransport = true` in the test. Otherwise what can happen is that if the developer set up an incorrect matcher in the transport (which is quite easy to happen), the test will make a real network call. This is useful during development, but we don't want that behavior in the CI because we want the tests to be fast and deterministic as much as possible.

protocol/linkpreview/linkpreview.go
ibrkhalil: I think some endpoints return JSON, Like giphy.
That's why the rename fixed the giphy issue.

ibrkhalil: I did it in a different way, Please take a look when you have the time.

protocol/linkpreview/linkpreview.go
```diff
@@ -168,19 +172,38 @@ func (u OEmbedUnfurler) unfurl() (common.LinkPreview, error) {
...
 	}
-	err = json.Unmarshal(oembedBytes, &oembedResponse)
+
+	type responseOembed struct {
+		title string
+		url   string
+	}
+
+	var response responseOembed
+	if strings.Contains(u.url.String(), "giphy") {
```
ilmotta: @ibrkhalil, this kind of code defeats the purpose of the generic oembed interface. We already detected in a reliable way if it's a giphy URL in `newUnfurler` via a normalized hostname. Besides that, this check here is not reliable. After an unfurler is created by `newUnfurler` the code should be agnostic of any given host.

If the oembed response can vary between providers (which is definitely possible) then I think we should simply evolve the abstractions & interfaces to support this need.

ibrkhalil: But if every gif provider has his own API and is not unifiying their response structure, Should we try to expect what they'll return in their JSON?

ilmotta: They follow the oEmbed protocol at the very minimum, so I think we should go back to it and look at the specs https://oembed.com/.

By no means the current implementation is perfectly implementing oEmbed. I didn't care about a lot of stuff there in the past. So now in your issue you are simply seeing the limitations of the original implementation.

I checked Giphy responses and the oEmbed response is of type `photo` and the spec says the required fields for type `photo` are `url`, `width`, and `height`. This explains why Giphy wasn't working and why you had to rename the struct tag to `url`.

But just renaming to `url` is incorrect according to the specs. `thumbnail_url` is also correct. The fix IMO should be to evolve the code to more strictly handle oEmbed responses considering the `type` returned in the oEmbed JSON response. So the code in the implementation of `unfurl` for `OEmbedUnfurler` should be kept generic and respect the oEmbed specs (i.e. no concrete statements about the hostname).

ibrkhalil: Please take another look Icaro when you have the time

protocol/linkpreview/linkpreview.go
```diff
@@ -162,25 +204,35 @@ func (u OEmbedUnfurler) unfurl() (common.LinkPreview, error) {
...
+	switch oembedResponse.Type {
+	case "photo":
+		handlePhotoOembedType(&preview, oembedResponse)
+
+	case "video":
+		handleVideoOembedType(&preview, oembedResponse)
+
+	case "rich":
+		handleRichOembedType(&preview, oembedResponse)
+	default:
```
ilmotta, line 253: To be consistent with formatting we should have a blank line before `default`.

protocol/linkpreview/linkpreview.go
```diff
@@ -149,9 +162,38 @@ func (u OEmbedUnfurler) newOEmbedURL() (*neturl.URL, error) {
...
+	if response.Width != 0 {
+		preview.Thumbnail.Width = response.ThumbnailWidth
+	}
+	if response.Height != 0 {
+		preview.Thumbnail.Height = response.ThumbnailHeight
+	}
+}
+
+func handleRichOembedType(preview *common.LinkPreview, response OEmbedBaseResponse) {
+	if response.ThumbnailURL != "" {
```
ilmotta: It seems all these checks could be removed, also in the other `handle*` functions. I changed the code locally and the tests pass (except for the failure which you're aware of in the CI as well).

The zero-values of the response object are basically empty strings or zero for numbers so we can just always set struct field anyway. 

To give you an example for this `response.ThumbnailURL != ""` check (but the reasoning should hold for the other checks):

`preview.Thumbnail.URL` is initialized as the zero-value of the string, thus empty. Checking if `response.ThumbnailURL != ""` is unnecessary because if we set `preview.Thumbnail.URL = response.ThumbnailURL` it will be the empty string, just as the zero value of the struct field. This same logic could be used for the other checks.

Maybe I'm missing something.

ibrkhalil: Removed, So I get it nil is the ones that sometimes is problematic, That's why we tend to check for it.

protocol/linkpreview/linkpreview.go
ibrkhalil: Done

protocol/linkpreview/linkpreview.go
ilmotta: I thought a `string` struct field cannot be nil, so there's never a case where `OEmbedBaseResponse` has nil values.

ilmotta (approved):

ilmotta:
Hey @ibrkhalil, is there anything holding this PR from advancing? I'm already working on changes that will touch the same files, so we'll face conflicts for sure.

ibrkhalil:
> Hey @ibrkhalil, is there anything holding this PR from advancing? I'm already working on changes that will touch the same files, so we'll face conflicts for sure.

Hey Icaro, I am awaiting a reply from design team on [this](https://github.com/status-im/status-mobile/pull/16814#issuecomment-1670085820) comment

ibrkhalil:
> > Hey @ibrkhalil, is there anything holding this PR from advancing? I'm already working on changes that will touch the same files, so we'll face conflicts for sure.
> 
> Hey Icaro, I am awaiting a reply from design team on [this](https://github.com/status-im/status-mobile/pull/16814#issuecomment-1670085820) comment

I posted it on their Discord server

ilmotta:
@ibrkhalil, interesting proposal https://github.com/status-im/status-mobile/pull/16814#issuecomment-1670085820 for processing gifs.

The feature I'm working on is to add support for static image URLs, which, if we go the route you and @cammellos suggested of taking a snapshot of a single gif frame, means image URLs and gif URLs will share a lot in common code-wise and behavior-wise.

The agreed flow with image URLs is that the image will be unfurled once on the sender side in a lower resolution and compressed image, and then on the client side, the user will need to explicitly press on the preview thumbnail to see the full image, a la Discord.

Just one question @cammellos @ibrkhalil, why can't we unfurl and persist the full gif (heavily preprocessed of course) and then serve it to clients? Are there technical limitations? Or is it just for simplicity at this stage of the product?

The ideal behavior in my head would be:

1. Preprocess gifs: lower their frame rate & resolution, and compress them.
2. Persist on waku the preprocessed gifs and the URL to the original resource (in case the user wants to see the gif in its full glory).
3. On clients, by default, show the preprocessed gifs (not just a static frame), but allow users to see the original resource if they so want. This step is essentially the same we'll go for static images.

The benefit, I believe, is that the majority of users will be happy to see only the preprocessed (also interactive) gifs, and they will never pay the privacy cost because they won't bother seeing the full gif in the original website. If we decide to show only a static frame of the gif on clients, it kind of diminishes the value of the gif feature, because I believe a lot of users will still press on the image to the see the interactive gif, but this will cost them on privacy. You see, the static gif snapshot sort of invites users to interact with centralized servers because humans are curious and they want to see more frames 🤷🏼

cammellos:
> 

@ilmotta If we can keep gifs size lower than 300/400KB I think sending through waku is fine and preferable, @Samyoul has worked on the compression, not sure it supports gifs. 
If the size is higher though, we might want to do as discussed (click to view gif).
A play button though is also useful for performance, as a side note, so that they don't consume cpu cycles by sending multiple gifs that are all animating, but that's a different issue.

ilmotta:
> > 
> 
> @ilmotta If we can keep gifs size lower than 300/400KB I think sending through waku is fine and preferable, @Samyoul has worked on the compression, not sure it supports gifs. If the size is higher though, we might want to do as discussed (click to view gif).

I sincerely have no idea if average smartphones can quickly process gifs to make them fit this sizing constraint. Some gifs are almost like full short videos, so preprocessing them could take a considerable amount of time on mobile devices. That's why I asked if there's any technical limitation, but I guess we would need to quickly try on status-go and see? Unless status-go isn't equipped now to do this sort of processing on gifs, then it could be too much effort for the MVP (?).

> A play button though is also useful for performance, as a side note, so that they don't consume cpu cycles by sending multiple gifs that are all animating, but that's a different issue.

100%. Addtionally, so many users hate autoplay gifs.

Samyoul:
Hey @ilmotta / @cammellos sorry for the late reply.

### Gifs in `status-go`.

Gifs manipulation is not supported currently in `status-go`, but it can be. Gifs manipulation in Go is slightly different to jpg or other raster image format types. A gif (as you may expect) is an array of frames, and each frame needs to be processed individually and then the gif wrapper data needs to be constituted (which can be fiddly, or at least isn't straight forward).

Back in the day I did do some basic work on gif support in `status-go`, but the biggest issue is that the native mobile image picker would "typecast" the image as a JPEG and send only the first the frame of the gif, with all the gif metadata stripped out. I didn't want to spend time working more on this so I parked the feature.

If we are pulling the gif from a URL the image being mangled by the image selector isn't a problem. But it would be really nice to also fix that issue as many phones have gif selection options built into the keyboard and inject the image data as if it was selected from the phone's storage.

### Issues about computational power, that is a tricky one.

For processing the image data, in general most image manipulation is handled easily by even low powered phones, we really don't do anything complex. Cropping is very fast (basically trimming arrays), resize is also quick (2D array nearest neighbour(s) value averaging). These are the 2 main processes we use. In the jpg specific context we also run many multiple quality level renders to discover the optimal jpg quality level to ideal file size, this is also fast even for low spec devices.

For running the gifs that is a different scenario, it really depends on the gif and the device. So maybe auto-stop playing once a gif is out of frame, and default no-auto play (Although I like seeing all the gifs play like a nausea inducing circus of chaos.).

ilmotta:
Thank you for the detailed explanation @Samyoul.
 
> Back in the day I did do some basic work on gif support in `status-go`, but the biggest issue is that the native mobile image picker would "typecast" the image as a JPEG and send only the first the frame of the gif, with all the gif metadata stripped out. I didn't want to spend time working more on this so I parked the feature.
> 
> If we are pulling the gif from a URL the image being mangled by the image selector isn't a problem. But it would be really nice to also fix that issue as many phones have gif selection options built into the keyboard and inject the image data as if it was selected from the phone's storage.

Agreed 👍🏼 

From what I understood, it seems we have some unknowns that may turn out to be blockers. In this case, the static image approach to gifs is starting to grow on me :)

> Gifs manipulation is not supported currently in `status-go`, but it can be. Gifs manipulation in Go is slightly different to jpg or other raster image format types. A gif (as you may expect) is an array of frames, and each frame needs to be processed individually and then the gif wrapper data needs to be constituted (which can be fiddly, or at least isn't straight forward).

Since animated wepb can deliver similar quality for lower sizes compared to gif, I wonder if mobile supports animated webp, and if that could be of any help. This link is quite useful https://developers.google.com/speed/webp/faq#why_should_i_use_animated_webp This could be interesting anyway for the desktop app.

> For processing the image data, in general most image manipulation is handled easily by even low powered phones

I'm impressed, cool!

-----

@cammellos, that was all! I guess we can evolve gif previews in later iterations.

igor-sirotin:
@ilmotta do you know if this PR is still needed?
https://github.com/status-im/status-mobile/issues/16457 seems to be resolved, but there's some long discussion here 🤔

ilmotta:
> @ilmotta do you know if this PR is still needed? [status-im/status-mobile#16457](https://github.com/status-im/status-mobile/issues/16457) seems to be resolved, but there's some long discussion here 🤔

Good finding @igor-sirotin. We can close the issue because there are too many open points and, at least on Mobile, it's not in our priorities and probably won't be for a long time.

PR #3827, opened 2023-07-31
Regularly update the list of curated communities

vkjr:
Related to [status-mobile bug](https://github.com/status-im/status-mobile/issues/16346)

Problem: when featured communities are requested by a client, some of them may be not known yet and their details should be requested from mailserver. But the client won't know this because the `CuratedCommunities()` result already returned without unknown communities.

Solution: now `status-go` every 2 minutes (random timeout, can be changed) polls the featured communities and signals the client about them. If there are unknown communities in the latest poll, the next one will happen in 15 seconds (hoping that communities info already will be fetched from mailserver).

### DIFF
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -403,6 +403,46 @@ func (m *Messenger) SpectatedCommunities() ([]*communities.Community, error) {
 	return m.communitiesManager.Spectated()
 }
 
+// Regularly gets list of curated communities and signals them to client
+func (m *Messenger) startCuratedCommunitiesUpdateLoop() {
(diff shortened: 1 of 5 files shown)


jo-mut (approved):
looks good

protocol/messenger_communities.go
```diff
@@ -455,6 +495,11 @@ func (m *Messenger) CuratedCommunities() (*communities.KnownCommunitiesResponse,
 			backend = b
 		}
 	}
+
+	if backend == nil {
+		return nil, errors.New("failed to initialize backend before requesting curated communities")
```
qfrank, line 470: change `backend` to `ethclient` looks better for me

qfrank (approved):

PR #3837, opened 2023-08-01
chore: reduce batch size for goerli arbitrum/optimism

alaibe:
Reduce batch size for arbitrum/optimism goerli to 10K max rather than 100k for arbitrum and 500k for optimism. As requested per infura team.

This is causing availability issue to infura team


rasom (approved):

jakubgs (approved):
Thank you.

PR #3839, opened 2023-08-01
fix(wallet) propagate multi-transactions IDs to transfers

stefandunca:
### Updates status-desktop [#11754](https://github.com/status-im/status-desktop/issues/11754), [#11404](https://github.com/status-im/status-desktop/issues/11404)

Mainly refactor API to have control on pending_transactions operations.
Use the new API to migrate the multi-transaction ID from to transfers
in one SQL transaction.
The refactoring was done to better mirror the purpose of pending_transactions

Pending API tracker usage (`PendingTxTracker`):
- Track a pending transaction `TrackPendingTransaction`
- Watch `EventPendingTransactionStatusChanged` for changes
- Optionally, auto delete is off, confirm the transaction change was acknowledged `DeleteBySQLTx` or `Delete`

Also:
- Externalize TransactionManager from WalletService to be used by other services
- Extract walletEvent as a dependency for all services that need to propagate events
- Batch chain requests
- Remove unused APIs
- Add auto delete option for clients that fire and forget transactions

Commit order

- fix(wallet) propagate multi-transactions IDs to transfers
- feat(wallet) add scope and auto_delete to pending_transactions
- chore(wallet) add stretchr/testify/mock to dependencies
- fix(wallet) fix test on checking order of async task

[Initial plan](https://github.com/status-im/status-go/blob/aad0af2bc1a9c997c296922254fc4d92562c97f3/transactions/TODO-pending_transition.md)


rpc/chain/client.go
```diff
@@ -25,6 +25,10 @@ type FeeHistory struct {
 	BaseFeePerGas []string `json:"baseFeePerGas"`
 }
 
+type EthClient interface {
```
IvanBelyakoff: BatchCaller ? seems like a better name and following interface naming conventions

appdatabase/migrations/sql/1691258252_add_scope_to_pending.up.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE pending_tx_status (
```
IvanBelyakoff: Stefan, why do we need this static table?

services/wallet/transfer/transaction.go
```diff
@@ -329,10 +329,13 @@ func createPendingTransactions(hashes map[uint64][]types.Hash, data []*bridge.Tr
...
 				Data:               tx.Data().String(),
 				Type:               transactions.WalletTransfer,
-				ChainID:            tx.ChainID,
+				ChainID:            wallet_common.ChainID(tx.ChainID),
 				MultiTransactionID: int64(multiTransaction.ID),
 				Symbol:             multiTransaction.FromAsset,
+				AutoDelete:         new(bool),
 			}
+			// Transaction downloader will delete pending transaction as soon as it is confirmed
+			*pendingTransaction.AutoDelete = false
```
IvanBelyakoff, line 338: Why not set it on instantiation?

transactions/conditionalrepeater.go
IvanBelyakoff, line 1: Good job. Just curious, if you considered async package from wallet and why it did not fit your needs?

IvanBelyakoff (approved):
Great job! Just some minor comments/questions

appdatabase/migrations/sql/1691258252_add_scope_to_pending.up.sql
dlipicar: Same question... Doesn't sound like one would normally use a SQL table for this

appdatabase/migrations/sql/1691258252_add_scope_to_pending.up.sql
```diff
@@ -0,0 +1,20 @@
...
+INSERT INTO pending_tx_status (id, name) VALUES
+(0, 'Pending'),
+(1, 'Confirmed'),
+(2, 'Failed');
+
+ALTER TABLE pending_transactions ADD status INTEGER NOT NULL DEFAULT 0;
+
+-- The watcher will auto delete the pending txs that are confirmed or failed
+-- Else the producer module will have to delete them manually on processing
+ALTER TABLE pending_transactions ADD COLUMN auto_delete BOOLEAN NOT NULL DEFAULT 1;
```
dlipicar: hmm how would we handle pending txs with `auto_delete` set to 0 if the app is closed and re-opened? Would the producer be in charge of loading any pending transactions of a particular type and listening for state changes? Do we really need this or can we leave processing of these entries fully in charge of the new module?

rpc/chain/client.go
stefandunca: This is the Client abstract interface that `ClientWithFallback` implements. It is to be extended with mocked values used in tests. Not sure if the fact that we now have only one function abstraction makes sense to name it as such. Still, its scope is to abstract the `ethclient.Client`. Maybe you will like `ClientInteface` better?

appdatabase/migrations/sql/1691258252_add_scope_to_pending.up.sql
stefandunca: I was searching for a way to document the enum values without storing the name of the enum values. It looked to me like this way might bring some advantages:

- keep only enum values (index) instead of duplicate string names
- you can still query and visualize the content in a human-readable way
```sql
SELECT SUBSTR(HEX(from_address), 0, 5), pending_tx_status.name, auto_delete
FROM pending_transactions
JOIN pending_tx_status ON pending_transactions.status = pending_tx_status.id;
```
- because this column is expected to keep an enum that is found in status-go, we can have a test to validate that the DB is consistent with the code expectations

Maybe you find this approach unnatural. Let me know what you think. For sure, this experiment can be dropped and keep the strings. The space save for `pending_transactions` is negligible, and the changes of changing enum values are low and not very error-prone.

services/wallet/transfer/transaction.go
stefandunca: Do you mean `new(delete)`? I find it less readable, but for sure will work.

transactions/conditionalrepeater.go
stefandunca: I like to have the implementation close to the context where it is used. It makes sense to move it to `async` as soon as it is externalised (as we did with `async.Scheduler`). For now, this is the only place we need it for.

appdatabase/migrations/sql/1691258252_add_scope_to_pending.up.sql
stefandunca: > hmm how would we handle pending txs with auto_delete set to 0 if the app is closed and re-opened?

By default, all are auto-deleted (set to `true`) as it is implied now just that on a restart, the service will clean them up if they are not pending anymore. Note that now it is done by `WatchTransaction` blocking call; therefore, in case of an interruption, like a long pending state, we leave dangling pending transactions without a chance to clean up.

> Would the producer be in charge of loading any pending transactions of a particular type and listening for state changes?

That is right. For now only `WalletTransfer` have auto-delete to false and the transfer downloader has the responsibility to delete ("confirm") that the pending state was acknowledged.

> Do we really need this or can we leave processing of these entries fully in charge of the new module?

I see no other simple way to ensure consistency by propagating information like multi-transaction ID. Also, other clients keep their own table of pending transactions. They can switch to this unified way in the future.

I hope this clarifies the intention a bit.

dlipicar (approved):

rpc/chain/client.go
IvanBelyakoff: No, If you want to keep your initial meaning for it, then no sense to change the interface name, up to you. What I mean is that I'd use interface for clear purpose that it declares. So this one is indeed BatchCaller. If you will add some related function to it, it is still a BatchCaller, if you want to add an unrelated function to it, then you make another interface for that purpose and let your EthClient implement it. That is how I see what interfaces are for. Not mixing unrelated functionality in an interface.
That is just a suggestion, up to you

services/wallet/transfer/transaction.go
IvanBelyakoff: I think I missed last time that you have pointer to bool, not bool. Did not expect it to be a pointer to bool actually, so was curious, why you initialized `AutoDelete = false` not inline with the rest fields. But bool is still false by default. Do you just make it explicit for devs that for this case we have it set to false or I miss some golang specs?

appdatabase/migrations/sql/1691258252_add_scope_to_pending.up.sql
IvanBelyakoff: I see your point. Still it feels to me like an overkill. You keep the declaration of the constants in migration files, which I believe even IDE can't reference to and not an obvious place for devs to find them. And I believe it is confusing for many devs raising more questions than answers. Moreover the need to upgrade the database to change status enums does not look like a good idea as this involves DB upgrade for all users that we will have and DB upgrade is a lot more risky and troublesome thing that depends on environment than code upgrade (speaking about constants, this does not depend on any environment and a way more safe thing that does not happen in runtime, but at build time).
So thinking it over once again, I am quite strongly against this peace of code

services/wallet/transfer/transaction.go
stefandunca: I use reference to extend the generic struct with an optional value specifically for tracking pending transactions and not users in regular transactions; There are other ways to be more explicit but this was simpler.

rpc/chain/client.go
stefandunca: Because the `Client` object is one of our main wallet dependencies and the most important one I think it should have been named `ClientInterface` or `ClientIf` from the start and used everywhere instead of `go-ethereum.Client` so it can be easily mocked to test the client. However, I think it is confusing to have intermediate names just because we do it incrementally, now for one call and extend later with other calls as needed.

appdatabase/migrations/sql/1691258252_add_scope_to_pending.up.sql
stefandunca: Thanks for the great feedback @IvanBelyakoff, I'm glad to see your perspective on this experiment and will fall back to using enum names.

> ... Still it feels to me like an overkill. You keep the declaration of the constants in migration files, which I believe even IDE can't reference to and not an obvious place for devs to find them. And I believe it is confusing for many devs raising more questions than answers.

I saw it as if it is in the migration files then it is documented. Because migrations are run the link between enum values and entries is in the DB data also. Because we have a test to link the enum consistency in the code with the DB then it is also checked and in sync with the code enumerations.  The only missing link is that I didn't add a foreign key reference.

> Moreover the need to upgrade the database to change status enums does not look like a good idea as this involves DB upgrade for all users that we will have and DB upgrade is a lot more risky and troublesome thing that depends on environment than code upgrade (speaking about constants, this does not depend on any environment and a way more safe thing that does not happen in runtime, but at build time).

I see the free-roaming strings even worse in case of needing to extend the enum. If we need to change the enum (add/remove elements) we still have to migrate the data to keep the DB normalize, just that with strings there is nothing to rise a flag that the required migration was not done and DB is inconsistent.

> So thinking it over once again, I am quite strongly against this peace of code

You were heard, I will remove the reference from the DB

PR #3841, opened 2023-08-02
fix: converting profile to/from keycard/regular keypair handling

saledjenic:


api/geth_backend.go
```diff
@@ -1127,10 +1137,13 @@ func (b *GethStatusBackend) ConvertToKeycardAccount(account multiaccounts.Accoun
...
 			kc.AccountsAddresses = append(kc.AccountsAddresses, acc.Address)
 		}
-		err = accountDB.SaveOrUpdateKeycard(kc, uint64(time.Now().Unix()), true)
+		err = messenger.SaveOrUpdateKeycard(context.Background(), &kc)
 		if err != nil {
 			return err
 		}
+		// After `ConvertToKeycardAccount` function call the next step is logout, we need to give some time to the app (cause the
+		// password will be changed below) to dispatch the sync keypair message to paired devices, sent by `SaveOrUpdateKeycard` call.
+		time.Sleep(1 * time.Second)
```
IvanBelyakoff: I don't like sleeps. Do you need it? From what I see, all the internal calls to send a message to paired devices look synchronous, so after this function call, the message should be sent.

api/geth_backend.go
```diff
@@ -1384,10 +1402,13 @@ func (b *GethStatusBackend) ConvertToRegularAccount(mnemonic string, currPasswor
...
 	}
 
-	err = db.DeleteAllKeycardsWithKeyUID(accountInfo.KeyUID, uint64(time.Now().Unix()))
+	err = messenger.DeleteAllKeycardsWithKeyUID(context.Background(), accountInfo.KeyUID)
 	if err != nil {
 		return err
 	}
+	// After `ConvertToRegularAccount` function call the next step is logout, we need to give some time to the app (cause the password
+	// will be changed below) to dispatch the sync keypair message to paired devices, sent by `DeleteAllKeycardsWithKeyUID` call.
+	time.Sleep(1 * time.Second)
```
IvanBelyakoff: I believe same here. If you believe the inner calls are async and the message is not yet sent after this function returns, is there any chance to make a blocking call instead?

IvanBelyakoff (commented):
Sale, could you please address the inline comments?

api/geth_backend.go
saledjenic: @IvanBelyakoff I don't like that either, but the reason is stated in the comments in those 2 places.

If you have a better idea of how to do that let me know?

saledjenic: `ChangeDatabasePassword` does the logout, so no other way.

IvanBelyakoff: Sale, you have not explained why you think the message will not be sent yet on return from this line:
`err = messenger.SaveOrUpdateKeycard(context.Background(), &kc)`

saledjenic: Ahh I've thought you got it.

saledjenic: @IvanBelyakoff so... when we're converting a profile keypair to:
- Start using a Keycard or
- Stop using a Keycard

sync keypair message is emitted, it should be received by all paired devices. That way paired devices know that they need to align with the change made on another device.

Now since synced messages are dispatched every 300ms without that "sleep" `ChangeDatabasePassword` does the `Logout` and synced messages for notifying paired devices about profile keypair migration to/from keycard are never emitted.

That's the point of that sleep, if you have a better idea I am eager to hear.

IvanBelyakoff: Sale, the only thing that I did not get is 300ms timeout. I was asking you why you think the message is not sent yet, because I've looked into the code before asking and it looked to me, that the messages are sent immediately. I have not seen any delay. I grepped 300 for status-go and I can't find that delay either. I will ask you in DM about this

saledjenic: You can check this function and see, https://github.com/status-im/status-go/blob/c3d2ff072319d12e7739e94854813ac7d75e4921/protocol/common/message_sender.go#L353

saledjenic: @rasom @qfrank do you know if there is a better way to achieve the same (to be sure that the sync message is sent before `Logout` call) without using `sleep`?

saledjenic: @IvanBelyakoff messages are added to the store of vacp2p here:
https://github.com/status-im/status-go/blob/c3d2ff072319d12e7739e94854813ac7d75e4921/protocol/common/message_sender.go#L810

and sent async from here:
https://github.com/vacp2p/mvds/blob/d6b27d3843ffae66806cc969596af342c6e7a839/node/node.go#L324

alaibe: Could we leverage go coroutine and a wait here?

alaibe (approved):

api/geth_backend.go
saledjenic: Not sure what could be an event to trigger the end of go routine.

This function from vacp2p is called when we add a message to the store to be dispatched:
https://github.com/vacp2p/mvds/blob/d6b27d3843ffae66806cc969596af342c6e7a839/node/node.go#L271

This function from vacp2p is called when a message is dispatched for real:
https://github.com/vacp2p/mvds/blob/d6b27d3843ffae66806cc969596af342c6e7a839/node/node.go#L324

So all we can have is the message ID, but that's in the moment when we add it to the store, we don't know the moment when it's dispatched for real. The trick is that we do `Logout` from `ChangeDatabasePassword` call before the sync message gets dispatched and the paired device never receives it.

protocol/messenger_handler.go
```diff
@@ -3236,6 +3236,42 @@ func (m *Messenger) handleSyncAccountsPositions(message *protobuf.SyncAccountsPo
...
+	}
+
+	migrationNeeded := dbKeypair.MigratedToKeycard() && len(message.Keycards) == 0 || // `true` if profile keypair was migrated to the app on one of paired devices
+		!dbKeypair.MigratedToKeycard() && len(message.Keycards) > 0 // `true` if profile keypair was migrated to a Keycard on one of paired devices
+	err = m.settings.SaveSettingField(settings.ProfileMigrationNeeded, migrationNeeded)
+	if err != nil {
+		return false, err
+	}
+	state.Response.AddSetting(&settings.SyncSettingField{SettingField: settings.ProfileMigrationNeeded, Value: migrationNeeded})
+	if migrationNeeded {
```
qfrank: seems can be simplified with `return migrationNeeded, nil`

qfrank (commented):
Most LGTM, but i won't accept using sleep ... @cammellos should have some suggestions

qfrank:
maybe we can check `rawMessage.Sent`? not sure. https://github.com/status-im/status-go/blob/721386334a235c42d3ea63f7ae4e4ca76fcae5f5/protocol/messenger.go#L604  cc @cammellos

saledjenic:
As said above, I do not like either using "sleep", but I don't see an easy way to overcome that. Actually, I have an idea but that's a job by itself, which implies that we need to find a way to stop any user's input/activity and dispatch all pending messages before logout/closing the app.

Also in this case it's easily noticeable, but any other sync message might not be sent if the user immediately after an action closes the app fast enough. 

Dummy example... if a user sends/accepts a friend request and in a moment closes the app, then the sync message won't be dispatched.

Maybe we can log a new issue for that?

api/geth_backend.go
cammellos: waiting is not an option, as the same condition would happen if the user is offline for example, but messages should be retried even after login (at least we do for 1-to-1, and we should be doing the same pairing). This is assuming the message is scheduled at the datasync layer.

@saledjenic mentioned that even after login the message is not resent, so maybe there's some bug or timing issue, that's where we should have a look, but I think we can avoid the time.sleep

saledjenic:
@cammellos I removed "sleep" and did another round of tests for all above described scenarios devices A and B are paired:

- **Scenario 1**:
  1. the app on device **A is running**, the app on device **B is running** and the **internet is available** to both devices
  2. profile migrated on device A
  3. the app is closed on device A after the migration
  4. no sync messages have been received on device B at this moment
  5. profile logged in on device A, while the app on device B is still running
  6. sync message is received on device B and the migration process is run

- **Scenario 2**:
  1. the app on device **A is running**, the app on device **B is closed** and the **internet is available** to both devices
  2. profile migrated on device A
  3. app closed on device A after the migration
  4. profile logged in on device B, no sync messages have been received at this moment
  5. the app closed on device B
  6. profile logged in on device A, while the app on device B is still closed
  7. the app closed on device A
  8. profile logged in on device B, sync message is received on device B and the migration process is run
 
- **Scenario 3**:
  1. the app on device **A is running**, the app on device **B is running** and **no internet** on both devices
  2. profile migrated on device A
  3. app closed on device A after the migration
  4. no sync messages have been received on device B at this moment
  5. internet is available to both devices now
  6. no sync messages have been received on device B at this moment
  7. profile logged in on device A, while the app on device B is still running
  8. sync message is received on device B and the migration process is run

- **Scenario 4**:
  1. the app on device **A is running**, the app on device **B is running** and **no internet** on both devices
  2. profile migrated on device A
  3. app closed on device A after the migration
  4. no sync messages have been received on device B at this moment
  5. internet is still not available to both devices now
  6. no sync messages have been received on device B at this moment
  7. profile logged in on device A, while the app on device B is still running
  8. no sync messages have been received on device B at this moment
  9. internet is available to both devices now
  10. sync message is received on device B and the migration process is run

Now based on all that, I see that after successful migration on the device, a sync message gets dispatched from the device on the next profile logging in.

 I didn't know about that and that's why I used "sleep" to dispatch a sync message before app restart on the device where the migration was performed.

@cammellos if above described scenarios work for us, to have a sync message dispatched on the next logging-in and to not have it dispatched if the user never logs in after the migration then I am ready to remove "sleep" from this PR?

cammellos:
@saledjenic great work testing this!
Looks like sleep can be safely removed then.

saledjenic:
@cammellos "sleep" is removed now, thanks for the heads-up.

@benjthayer since this is something I didn't know before, maybe we could add a message like `Make sure you login again on this device to sync the migration with the other devices`, if you think that's needed, to these two screens:
- https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?node-id=7258%3A182469&mode=dev
- https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?node-id=18332%3A433949&mode=dev

qfrank (approved):
LGTM 👍

PR #3842, opened 2023-08-02
feat: change network selection strategy

alaibe:
1) Only fetch history for networks that match test mode enabled
Trade Off: it will be only refetch in 12 hours so changing test mode won't trigger a refetch until app is restarted or 12 hours. I think it is ok as change test mode is not a common use case

 2) Do not consider networks that are enabled or not as this can be change more often than every 12 hours


stefandunca (approved):
Great catch! This will improve the UX and save some requests.

rasom (approved):

PR #3843, opened 2023-08-02
refactor: EventSenders forward RequestToJoin decision to control node

0x-r4bbit:
This is a bigger change in how community membership requests are handled among admins, token masters, owners, and control nodes.

Prior to this commit, all privileged users, also known as `EventSenders`, were able to accept and reject community membership requests and those changes would be applied by all users.

This commit changes this behaviour such that:

1. EventSenders can make a decision (accept, reject), but merely forward their decision to the control node, which ultimately has to confirm it
2. EventSenders are no longer removing or adding members to and from communities
3. When an eventsender signaled a decision, the membership request will enter a pending state (acceptedPending or rejectedPending)
4. Once a decision was made by one eventsender, no other eventsender can override that decision

This implementation is covered with a bunch of tests:

- Ensure that decision made by event sender is shared with other event senders
  - `testAcceptMemberRequestToJoinResponseSharedWithOtherEventSenders()`
  - `testRejectMemberRequestToJoinResponseSharedWithOtherEventSenders()`
- Ensure memebrship request stays pending, until control node has confirmed decision by event senders
  - `testAcceptMemberRequestToJoinNotConfirmedByControlNode()`
  - `testRejectMemberRequestToJoinNotConfirmedByControlNode()`
- Ensure that decision made by event sender cannot be overriden by other event senders
  - `testEventSenderCannotOverrideRequestToJoinState()`

These test cases live in three test suites for different event sender types respectively

- `OwnerWithoutCommunityKeyCommunityEventsSuite`
- `TokenMasterCommunityEventsSuite`
- `AdminCommunityEventsSuite`

In addition to the changes mentioned above, there's also a smaller changes that ensures membership requests to *not* attached revealed wallet addresses when the requests are sent to event senders (in addition to control nodes).

Requests send to a control node will still include revealed addresses as the control node needs them to verify token permissions.

This commit does not yet handle the case of event senders attempting to kick and ban members.

Similar to accepting and rejecting membership requests, kicking and banning need a new pending state. However, we don't track such state in local databases yet so those two cases will be handled in future commit to not have this commit grow larger.

### DIFF
--- a/protocol/activity_center.go
+++ b/protocol/activity_center.go
@@ -37,6 +37,8 @@ const (
 	ActivityCenterMembershipStatusPending
 	ActivityCenterMembershipStatusAccepted
 	ActivityCenterMembershipStatusDeclined
+	ActivityCenterMembershipStatusAcceptedPending
+	ActivityCenterMembershipStatusDeclinedPending
 )
 
 type ActivityCenterQueryParamsRead uint
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -753,7 +753,6 @@ func (o *Community) RemoveUserFromOrg(pk *ecdsa.PublicKey) (*protobuf.CommunityD
 	}
 
 	o.removeMemberFromOrg(pk)
-
 	if isControlNode {
 		o.increaseClock()
 	}
@@ -1940,31 +1939,14 @@ func (o *Community) AddMemberWithRevealedAccounts(dbRequest *RequestToJoin, role
 	defer o.mutex.Unlock()
 
 	isControlNode := o.IsControlNode()
-	allowedToSendEvents := o.HasPermissionToSendCommunityEvents()
 
-	if !isControlNode && !allowedToSendEvents {
+	if !isControlNode {
 		return nil, ErrNotAdmin
 	}
 
 	changes := o.addMemberWithRevealedAccounts(dbRequest.PublicKey, roles, accounts, dbRequest.Clock)
 
-	if allowedToSendEvents {
-		acceptedRequestsToJoin := make(map[string]*protobuf.CommunityRequestToJoin)
-		acceptedRequestsToJoin[dbRequest.PublicKey] = dbRequest.ToCommunityRequestToJoinProtobuf()
-
-		adminChanges := &CommunityEventChanges{
-			CommunityChanges:       changes,
-			AcceptedRequestsToJoin: acceptedRequestsToJoin,
-		}
-		err := o.addNewCommunityEvent(o.ToCommunityRequestToJoinAcceptCommunityEvent(adminChanges))
-		if err != nil {
-			return nil, err
-		}
-	}
-
-	if isControlNode {
-		o.increaseClock()
-	}
+	o.increaseClock()
 
 	return changes, nil
 }
@@ -2128,16 +2110,6 @@ func (o *Community) deleteChat(chatID string) *CommunityChanges {
 
 	delete(o.config.CommunityDescription.Chats, chatID)
 	return changes
-}
-
-func (o *Community) addCommunityMember(pk *ecdsa.PublicKey, member *protobuf.CommunityMember) {
-
-	if o.config.CommunityDescription.Members == nil {
-		o.config.CommunityDescription.Members = make(map[string]*protobuf.CommunityMember)
-	}
-
-	memberKey := common.PubkeyToHex(pk)
-	o.config.CommunityDescription.Members[memberKey] = member
 }
 
 func (o *Community) addTokenPermission(permission *protobuf.CommunityTokenPermission) (*CommunityChanges, error) {
--- a/protocol/communities/community_event.go
+++ b/protocol/communities/community_event.go
@@ -160,7 +160,6 @@ func (o *Community) ToCommunityRequestToJoinAcceptCommunityEvent(changes *Commun
 	return &CommunityEvent{
 		CommunityEventClock:    o.NewCommunityEventClock(),
 		Type:                   protobuf.CommunityEvent_COMMUNITY_REQUEST_TO_JOIN_ACCEPT,
-		MembersAdded:           changes.MembersAdded,
 		AcceptedRequestsToJoin: changes.AcceptedRequestsToJoin,
 	}
 }
@@ -321,26 +320,6 @@ func (o *Community) updateCommunityDescriptionByCommunityEvent(communityEvent Co
 		_, err := o.reorderCategories(communityEvent.CategoryData.CategoryId, int(communityEvent.CategoryData.Position))
 		if err != nil {
 			return err
-		}
-
-	case protobuf.CommunityEvent_COMMUNITY_REQUEST_TO_JOIN_ACCEPT:
-		for pkString, addedMember := range communityEvent.MembersAdded {
-			pk, err := common.HexToPubkey(pkString)
-			if err != nil {
-				return err
-			}
-			if !o.HasMember(pk) {
-				o.addCommunityMember(pk, addedMember)
-			}
-		}
-
-	case protobuf.CommunityEvent_COMMUNITY_REQUEST_TO_JOIN_REJECT:
-		for pkString := range communityEvent.RejectedRequestsToJoin {
-			pk, err := common.HexToPubkey(pkString)
-			if err != nil {
-				return err
-			}
-			o.removeMemberFromOrg(pk)
 		}
 
 	case protobuf.CommunityEvent_COMMUNITY_MEMBER_KICK:
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -287,6 +287,8 @@ type Subscription struct {
 	DownloadingHistoryArchivesFinishedSignal *signal.DownloadingHistoryArchivesFinishedSignal
 	ImportingHistoryArchiveMessagesSignal    *signal.ImportingHistoryArchiveMessagesSignal
 	CommunityEventsMessage                   *CommunityEventsMessage
+	AcceptedRequestsToJoin                   []types.HexBytes
+	RejectedRequestsToJoin                   []types.HexBytes
 }
 
 type CommunityResponse struct {
@@ -1353,42 +1355,8 @@ func (m *Manager) HandleCommunityEventsMessage(signer *ecdsa.PublicKey, message 
 
 func (m *Manager) handleAdditionalAdminChanges(community *Community) error {
 
-	saveOrUpdateRequestToJoin := func(signer string, request *protobuf.CommunityRequestToJoin, state RequestToJoinState) error {
-		requestToJoin := &RequestToJoin{
-			PublicKey:        signer,
-			Clock:            request.Clock,
-			ENSName:          request.EnsName,
-			CommunityID:      request.CommunityId,
-			State:            state,
-			RevealedAccounts: request.RevealedAccounts,
-		}
-
-		requestToJoin.CalculateID()
-
-		existingRequestToJoin, err := m.persistence.GetRequestToJoin(requestToJoin.ID)
-		if err != nil && err != sql.ErrNoRows {
-			return err
-		}
-
-		if existingRequestToJoin != nil {
-			// node already knows about this request to join, so let's compare clocks
-			// and update it if necessary
-			if existingRequestToJoin.Clock <= requestToJoin.Clock {
-				pk, err := common.HexToPubkey(existingRequestToJoin.PublicKey)
-				if err != nil {
-					return err
-				}
-				err = m.persistence.SetRequestToJoinState(common.PubkeyToHex(pk), community.ID(), state)
-				if err != nil {
-					return err
-				}
-			}
-		} else {
-			err := m.persistence.SaveRequestToJoin(requestToJoin)
-			if err != nil {
-				return err
-			}
-		}
+	if !(community.IsControlNode() || community.HasPermissionToSendCommunityEvents()) {
+		// we're a normal user/member node, so there's nothing for us to do here
 		return nil
 	}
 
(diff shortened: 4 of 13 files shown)


protocol/activity_center.go
```diff
@@ -37,6 +37,8 @@ const (
 	ActivityCenterMembershipStatusPending
 	ActivityCenterMembershipStatusAccepted
 	ActivityCenterMembershipStatusDeclined
+	ActivityCenterMembershipStatusAcceptedPending
+	ActivityCenterMembershipStatusDeclinedPending
```
0x-r4bbit, line 41: @status-im/desktop-devs these are new states that desktop UI should be aware of

protocol/communities/community.go
```diff
@@ -824,9 +824,8 @@ func (o *Community) RemoveUserFromOrg(pk *ecdsa.PublicKey) (*protobuf.CommunityD
 		}
 	}
 
-	o.removeMemberFromOrg(pk)
-
 	if isControlNode {
+		o.removeMemberFromOrg(pk)
```
0x-r4bbit: This is moved into the condition because admins should not be able to remove users like that, We'll need a pending state for this.

For now, we at least prevent the member to be removed by the admin

protocol/communities/community.go
```diff
@@ -1995,31 +1994,14 @@ func (o *Community) AddMemberWithRevealedAccounts(dbRequest *RequestToJoin, role
 	defer o.mutex.Unlock()
 
 	isControlNode := o.IsControlNode()
-	allowedToSendEvents := o.HasPermissionToSendCommunityEvents()
 
-	if !isControlNode && !allowedToSendEvents {
+	if !isControlNode {
```
0x-r4bbit, line 1943: Admins should not make changes to members directly, so only control node is allowed to do stuff here.
Also, we want to move away from attaching revealed addresses to members in the community description (as this is public metadata).

Next step is to remove usage of this function entirely.

protocol/communities/community_event.go
```diff
@@ -160,7 +160,6 @@ func (o *Community) ToCommunityRequestToJoinAcceptCommunityEvent(changes *Commun
 	return &CommunityEvent{
 		CommunityEventClock:    o.NewCommunityEventClock(),
 		Type:                   protobuf.CommunityEvent_COMMUNITY_REQUEST_TO_JOIN_ACCEPT,
-		MembersAdded:           changes.MembersAdded,
```
0x-r4bbit, line 163: No need attach `MembersAdded` if admins can't actually do that anymore. 
Now we're just signalling which request to join is marked as accepted

protocol/communities/community_event.go
```diff
@@ -315,26 +314,6 @@ func (o *Community) updateCommunityDescriptionByCommunityEvent(communityEvent Co
...
-		}
-
-	case protobuf.CommunityEvent_COMMUNITY_REQUEST_TO_JOIN_REJECT:
-		for pkString := range communityEvent.RejectedRequestsToJoin {
-			pk, err := common.HexToPubkey(pkString)
-			if err != nil {
-				return err
-			}
-			o.removeMemberFromOrg(pk)
-		}
```
0x-r4bbit, line 344: Removal/addition of members via admin events is now done later in `handleAdditionalAdminChanges()` where the control node makes the ultimate decision

protocol/communities/manager.go
```diff
@@ -1370,65 +1371,144 @@ func (m *Manager) HandleCommunityEventsMessage(signer *ecdsa.PublicKey, message
...
+			// We only want to do this when `requestUpdated == true`, if the control node
+			// didn't know this request before, it means it just learned about it via this
+			// community event, which in turn means, we don't have any revealed addresses
+			// for this request yet, so we don't actually want to perform an
+			// accept action by the control node
+			acceptedRequestsToJoin = append(acceptedRequestsToJoin, requestToJoin.ID)
+		}
+	}
+	if community.IsControlNode() {
+		m.publish(&Subscription{AcceptedRequestsToJoin: acceptedRequestsToJoin})
```
0x-r4bbit, line 1459: This event causes `Messenger` to `AcceptCommunityRequestToJoin()` for all accepted requests

protocol/communities/manager.go
```diff
@@ -1601,54 +1685,85 @@ func (m *Manager) AcceptRequestToJoin(request *requests.AcceptRequestToJoinCommu
...
+			return community, ErrNoPermissionToJoin
+		}
+
+		memberRoles := []protobuf.CommunityMember_Roles{}
+		if isAdmin {
+			memberRoles = []protobuf.CommunityMember_Roles{protobuf.CommunityMember_ROLE_ADMIN}
+		}
+
+		// TODO: Don't add member to community, instead store revealed accounts in database
+		_, err = community.AddMemberWithRevealedAccounts(dbRequest, memberRoles, revealedAccounts)
```
0x-r4bbit, line 1730: This needs to be removed as part of https://github.com/status-im/status-desktop/issues/11573

protocol/communities/request_to_join.go
```diff
@@ -16,6 +16,8 @@ const (
 	RequestToJoinStateDeclined
 	RequestToJoinStateAccepted
 	RequestToJoinStateCanceled
+	RequestToJoinStateAcceptedPending
+	RequestToJoinStateDeclinedPending
```
0x-r4bbit, line 20: @status-im/desktop-devs also new states we probably want to use in the UI

protocol/messenger_communities.go
```diff
@@ -284,6 +284,18 @@ func (m *Messenger) handleCommunitiesSubscription(c chan *communities.Subscripti
...
+					for _, requestID := range sub.AcceptedRequestsToJoin {
+						accept := &requests.AcceptRequestToJoinCommunity{
+							ID: requestID,
+						}
+						_, err := m.AcceptRequestToJoinCommunity(accept)
+						if err != nil {
+							m.logger.Warn("failed to accept request to join ", zap.Error(err))
+						}
+					}
+				}
```
0x-r4bbit, line 282: This will be triggered as a side effect in `HandleCommunityEventsMessage` when there were accepted requests to join

protocol/activity_center.go
osmaczko: I am wondering why we need this notification. Isn't pending state shared only among admins and control node? If that's the case, I believe we don't need a notification for that :thinking:

protocol/communities/manager.go
```diff
@@ -1370,65 +1371,144 @@ func (m *Manager) HandleCommunityEventsMessage(signer *ecdsa.PublicKey, message
...
+			acceptedRequestsToJoin = append(acceptedRequestsToJoin, requestToJoin.ID)
+		}
+	}
+	if community.IsControlNode() {
+		m.publish(&Subscription{AcceptedRequestsToJoin: acceptedRequestsToJoin})
+	}
+	return nil
+}
+
+func (m *Manager) handleCommunityEventRequestRejected(community *Community, communityEvent *CommunityEvent) error {
```
mprakhov, line 1464: Does the control node need to notify privileged users about the declined request, as you do for an accepted one?

protocol/activity_center.go
MishkaRogachev: How we should show those states in the UI?

0x-r4bbit: Everything related to UI is discussed here: https://github.com/status-im/status-desktop/issues/11649

protocol/communities_events_owner_without_community_key_test.go
```diff
@@ -40,6 +41,10 @@ func (s *OwnerWithoutCommunityKeyCommunityEventsSuite) GetEventSender() *Messeng
 	return s.ownerWithoutCommunityKey
 }
 
+func (s *OwnerWithoutCommunityKeyCommunityEventsSuite) GetEventSender2() *Messenger {
+	return s.ownerWithoutCommunityKey2
+}
+
```
mprakhov: I wonder if it is possible to create this user in the common test scenario. The same as we did for the `User` in requests to join tests

mprakhov: Like here
https://github.com/status-im/status-go/pull/3843/files#diff-2990ef21a6e7432ed6cdc95cc05b9f47439c44742fb1da050adb30d976ea1aa7R840-R845

mprakhov (commented):
Great work.
Left few comments

protocol/communities/manager.go
jrainville: They should notify the privileged users yes, so that they can delete the request and stop being in the pending state.
We still shouldn't send it to the user requesting however.

protocol/communities/manager.go
```diff
@@ -1370,65 +1371,144 @@ func (m *Manager) HandleCommunityEventsMessage(signer *ecdsa.PublicKey, message
...
+		}
+
+		if requestUpdated {
+			// We only want to do this when `requestUpdated == true`, if the control node
+			// didn't know this request before, it means it just learned about it via this
+			// community event, which in turn means, we don't have any revealed addresses
+			// for this request yet, so we don't actually want to perform an
+			// accept action by the control node
+			acceptedRequestsToJoin = append(acceptedRequestsToJoin, requestToJoin.ID)
+		}
```
osmaczko: What if we receive many (same) requests to join through community events?

protocol/communities/manager.go
```diff
@@ -1370,65 +1371,144 @@ func (m *Manager) HandleCommunityEventsMessage(signer *ecdsa.PublicKey, message
...
+
+	existingRequestToJoin, err := m.persistence.GetRequestToJoin(requestToJoin.ID)
+	if err != nil && err != sql.ErrNoRows {
+		return updated, err
+	}
+
+	if existingRequestToJoin != nil {
+		// node already knows about this request to join, so let's compare clocks
+		// and update it if necessary
+		if existingRequestToJoin.Clock <= requestToJoin.Clock {
```
osmaczko, line 1395:
```suggestion
		if existingRequestToJoin.Clock < requestToJoin.Clock {
```

I know it was like that before, but I believe same clock should be ignored.

protocol/communities/manager.go
```diff
@@ -1601,54 +1685,85 @@ func (m *Manager) AcceptRequestToJoin(request *requests.AcceptRequestToJoinCommu
...
+		if !permissionsSatisfied {
+			return community, ErrNoPermissionToJoin
+		}
+
+		memberRoles := []protobuf.CommunityMember_Roles{}
+		if isAdmin {
+			memberRoles = []protobuf.CommunityMember_Roles{protobuf.CommunityMember_ROLE_ADMIN}
+		}
+
+		// TODO: Don't add member to community, instead store revealed accounts in database
```
osmaczko: Not sure I got it right. When/where do we add members to community then?

protocol/communities/manager.go
```diff
@@ -1822,22 +1959,36 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request
...
+			requestToJoin.State = RequestToJoinStateAccepted
+		} else {
+			err = m.markRequestToJoinAsAcceptedPending(signer, community)
+			if err != nil {
+				return nil, err
+			}
+			requestToJoin.State = RequestToJoinStateAcceptedPending
 		}
-		// Don't check permissions here,
-		// it will be done further in the processing pipeline.
```
osmaczko, line 1804: This comment could still be useful.

protocol/communities/manager.go
0x-r4bbit: That's a good point. I also just realized we're not updating other admins to update their pending state on accepted either.

protocol/communities_events_owner_without_community_key_test.go
0x-r4bbit: I will take a look at this!

osmaczko (commented):
Nice work, although I have few questions/uncertainties.

protocol/communities/manager.go
0x-r4bbit: The first one wins. But this can indeed cause a state where 2 admins look at different pending states until the ultimate state was confirmed by the control node.

To avoid that, we'd have to also store which account made which decision

protocol/communities/manager.go
0x-r4bbit: They need to be store in a separated database table instead of the community descriptoin so they aren't shared with public metadata.

Then, revealed addresses only locally known need to be backed up on waku such that control nodes can recover them.

protocol/communities/manager.go
0x-r4bbit: Okay will undo this

protocol/communities/manager.go
osmaczko: > They need to be store in a separated database table instead of the community descriptoin so they aren't shared with public metadata.

How will clients then construct the member list?

0x-r4bbit: Ah sorry, the comment has a mistake. This is not about the members but about the revealed addresses

protocol/communities/manager.go
osmaczko: I was wondering about the scenario where two or more admin events come to control node before request to join:

1) acceptedRequestToJoin by adminA at timestamp 1
2) acceptedRequestToJoin by adminB at timestamp 2

at 2) `acceptedRequestsToJoin` will be populated with request that don't have revealed addresses. As a result, control node will reject the request before receiving addresses.

0x-r4bbit: So what I'd expect to happen in this case (and I'll add a test for this to verify this):

1. adminA decision reaches control node. Control node saves this request to join as it has seen it for the first time (at this point the request is in state acceptedPending
2. adminB's decision reaches control node. Control node sees that it already have a request with the same ID **and** it's in some pending state (acceptedPending or rejectedPending), so it'll ignore it altogether
3. Ultimately, control node will receive requestToJoin from user and perform the necessary checks. If there is already a request to join from an admin (meaning, the request exists in the database and is in any pending state coming from an admin), then it knows it can automatically accept (or reject) the request based on that

The important part here is that in 3) the request comes in via normal `HandleRequestToJoin()` while 1) and 2) are handled via `HandleCommunityEventsMessage`.

osmaczko: > 2. adminB's decision reaches control node. Control node sees that it already have a request with the same ID and it's in some pending state (acceptedPending or rejectedPending), so it'll ignore it altogether

I can't find this part in the code, could you please point me where this check happens?

0x-r4bbit: Ah, I'm relying on this https://github.com/status-im/status-go/pull/3843/files#diff-0c22647108517d38489ae5faa65a914ade3be02ea7a7e37355a82739da9aeb69R1411 which doesn't check for pending state. It'll basically ignore any additional incoming decision from an admin, regardless of the state of existing request. Should probably add a check for pending states there

osmaczko: > It'll basically ignore any additional incoming decision from an admin, regardless of the state of existing request

On the contrary, it will update the existing request if the new one (with higher clock value) comes :thinking: 

https://github.com/status-im/status-go/blob/bf60cd6a265495c677de9f0fce06e238670672d7/protocol/communities/manager.go#L1411-L1421

0x-r4bbit: > On the contrary, it will update the existing request if the new one (with higher clock value) comes 🤔

That's correct, but we expect it to be the same requestToJoin coming from different admins, so the clock will be the same no?

If you cancel your reques to join and send a new one, then it's in fact a new one which is fine (unless I'm missing something

osmaczko: > so the clock will be the same no?

Oh, I missed that. I assumed the clock is altered once admin accepts/rejects the request. All fine then, thanks for patience.

> The first one wins. But this can indeed cause a state where 2 admins look at different pending states until the ultimate state was confirmed by the control node.

That's acceptable to me.

osmaczko (approved):

protocol/communities/manager.go
0x-r4bbit: Thank you for exploring this in detail! I'm going to add a test case for this regardless :)

protocol/communities/manager.go
0x-r4bbit: As discussed in https://github.com/status-im/status-go/pull/3843/files#r1282298019 we expect it to be the same clock, so we don't want to ignore this

protocol/communities/community.go
0x-r4bbit: Undoing this as I'm not handling kicking/banning members in this PR.

protocol/communities/manager.go
0x-r4bbit: After a few discussions with @osmaczko , this signal should be part of https://github.com/status-im/status-desktop/issues/11715, so that has to land first.

protocol/activity_center.go
igor-sirotin: Should it be `...AcceptPending` and `..DeclinePending` (without `ed`)?
Or am I misunderstanding smth?

0x-r4bbit: Well the state is in "accepted pending" meaning, an admin has accepted the request but the decision is still pending until a control node confirms. Not sure if `-ed` is an issue here

protocol/communities_events_owner_without_community_key_test.go
0x-r4bbit: I've updated the PR to make `additionalEventSenders` parameterizable so there's no need for `GetEventSender2()`, however, `setUpOnRequestCommunity()` now needs a `[]*Messenger{}` list of additional event senders where needed.

protocol/communities_events_utils_test.go
```diff
@@ -363,9 +364,20 @@ func setUpOnRequestCommunityAndRoles(base CommunityEventsTestsInterface, role pr
...
+	for _, eventSender := range additionalEventSenders {
+		advertiseCommunityTo(s, community, base.GetControlNode(), eventSender)
+		joinOnRequestCommunity(s, community, base.GetControlNode(), eventSender)
+
+		grantPermission(s, community, base.GetControlNode(), eventSender, role)
+		checkPermissionGranted = func(response *MessengerResponse) error {
+			return checkRolePermissionInResponse(response, eventSender.IdentityPublicKey(), role)
+		}
+		waitOnMessengerResponse(s, WaitCommunityCondition, checkPermissionGranted, base.GetMember())
+		time.Sleep(500 * time.Millisecond)
```
mprakhov: this should work without sleep

0x-r4bbit: Ah, this has been flaky so I've added the sleep here, but didn't mean to leave it in.
I'm still making changes to this PR.

Any other ideas what can be done to make this less flaky?

mprakhov: 0. Need to detect where it is failing exactly
1. WaitCommunityCondition change on condition, where you will wait for a community + specific permission for a specific user
2. Call waitOnMessengerResponse for all other members in order for they have the latest CommunityDescription

P.S. If there will be any problems, ping me, I'll take a look at logs

protocol/communities/manager.go
```diff
@@ -1812,22 +1958,42 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request
...
 		return requestToJoin, nil
 	}
 
-	permissionsSatisfied, _, err := m.accountsSatisfyPermissionsToJoin(community, request.RevealedAccounts)
-	if err != nil {
-		return nil, err
-	}
-	if !permissionsSatisfied {
-		requestToJoin.State = RequestToJoinStateDeclined
+	if community.IsControlNode() && len(request.RevealedAccounts) > 0 {
```
jrainville, line 1963: if `len(request.RevealedAccounts) == 0` and the community has permissions to join, we should auto set it to `RequestToJoinStateDeclined`. I'm not sure if the code does this.

protocol/messenger_communities.go
```diff
@@ -284,6 +284,33 @@ func (m *Messenger) handleCommunitiesSubscription(c chan *communities.Subscripti
...
+				if sub.AcceptedRequestsToJoin != nil {
+					for _, requestID := range sub.AcceptedRequestsToJoin {
+						accept := &requests.AcceptRequestToJoinCommunity{
+							ID: requestID,
+						}
+						_, err := m.AcceptRequestToJoinCommunity(accept)
+						if err != nil {
+							m.logger.Warn("failed to accept request to join ", zap.Error(err))
+						}
+						// TODO INFORM ADMINS
```
jrainville, line 266: Are those TODOs done or to be done in another PR?

jrainville (commented):
Very good work. I had 2 questions.

protocol/messenger_communities.go
0x-r4bbit: This needs to be done as part of https://github.com/status-im/status-desktop/issues/11715

protocol/communities/manager.go
0x-r4bbit: Happening in https://github.com/status-im/status-go/pull/3853

protocol/communities/manager.go
0x-r4bbit: Ah actually, the check on `len(request.RevealedAccounts)` is a left over but it probably wouldn't hurt. It ends up being marked as `Declined` a few line below latest after the satisfaction check was done

MishkaRogachev (approved):
Great work! And no questions from my side :)

mprakhov (approved):

alexjba:
There is one UI requirement that might need backed support and I can't spot it in this PR:

[Only the admin that made the initial decision can change the decision. (Other admins will not see the alternate button on hover).](https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba⎜Desktop?type=design&node-id=3113-356327&mode=design&t=Jij9C3tclFyDpj6t-0). Since only the initiator can see the alternate button and change the membership request state after accepting/rejecting it, we need to identify the source of membership state change. 

Was it considered for this PR?

0x-r4bbit:
@alexjba this is indeed not covered in this PR. The PR assumes that, once a state has been introduced by an admin, it can no longer be changed. 

@jrainville this would need change in the database and a migration script. Should this happen in this PR as well or should we create a separate issue for it so this can be tackled in follow-up work?

jrainville:
> @jrainville this would need change in the database and a migration script. Should this happen in this PR as well or should we create a separate issue for it so this can be tackled in follow-up work?

It can be done in another ticket

0x-r4bbit:
> It can be done in another ticket

@jrainville added an issue describing what needs to be done here: https://github.com/status-im/status-desktop/issues/11842

Might be a good task for Teodor to get his feet wet with status-go

PR #3845, opened 2023-08-02
Optimized counter increment logic for 1-1 chat messages

### LINKED ISSUE #16696: Optimized counter increment logic for 1-1 chat messages
See the original feature description provided by @John-44 in this conversation (see point B) https://discord.com/channels/624634427930312714/928646594314784859/1129407293256122429

Currently, in a 1-1 chat, each message sent increments the counter, regardless of the proximity of the messages in time. To enhance the user experience we should implement an optimized logic for the counter incrementation.

The suggested behavior is as follows: When a 1-1 chat message is sent, the counter will be incremented by 1. However, if another message is sent within 2 minutes of the previous message, the counter will not be incremented, and the 2-minute window will reset. Only when no further messages are received within the subsequent 2 minutes after the most recent message, the counter will increment again upon sending a subsequent message.

By implementing this logic, the counter will accurately reflect the number of distinct messaging events rather than incrementing for each individual message sent within a short time frame. This optimization will provide a more meaningful representation of chat activity while avoiding unnecessary counter increments and reducing visual clutter.

alwx:
Fixes https://github.com/status-im/status-mobile/issues/16696

### DIFF
--- a/protocol/messenger_handler.go
+++ b/protocol/messenger_handler.go
@@ -39,6 +39,13 @@ const (
 	requestAddressForTransactionDeclinedMessage = "Request address for transaction declined"
 )
 
+const (
+	// IncreaseUnviewedMessagesCountTimeout
+	// this timeout indicates how long the time between received messages should be
+	// for a new message to increase the unviewed messages counter
+	IncreaseUnviewedMessagesCountTimeout = 1000 * 60 * 2
+)
+
 var (
 	ErrMessageNotAllowed                     = errors.New("message from a non-contact")
 	ErrMessageForWrongChatType               = errors.New("message for the wrong chat type")
@@ -2993,7 +3000,7 @@ func (m *Messenger) isMessageAllowedFrom(publicKey string, chat *Chat) (bool, er
 }
 
 func (m *Messenger) updateUnviewedCounts(chat *Chat, message *common.Message) {
-	if message.Timestamp > chat.LastMessage.Timestamp+1000*60*2 {
+	if message.Timestamp > chat.LastMessage.Timestamp+IncreaseUnviewedMessagesCountTimeout {
 		chat.UnviewedMessagesCount += 1
 	}
 	if message.Mentioned || message.Replied {
--- a/protocol/messenger_handler.go
+++ b/protocol/messenger_handler.go
@@ -396,7 +396,7 @@ func (m *Messenger) handleCommandMessage(state *ReceivedMessageState, message *c
 
 	// Increase unviewed count
 	if !common.IsPubKeyEqual(message.SigPubKey, &m.identity.PublicKey) {
-		m.updateUnviewedCounts(chat, message.Mentioned || message.Replied)
+		m.updateUnviewedCounts(chat, message)
 		message.OutgoingStatus = ""
 	} else {
 		// Our own message, mark as sent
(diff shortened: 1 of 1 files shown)


protocol/messenger_handler.go
```diff
@@ -3015,9 +3015,11 @@ func (m *Messenger) isMessageAllowedFrom(publicKey string, chat *Chat) (bool, er
 	return contact.added(), nil
 }
 
-func (m *Messenger) updateUnviewedCounts(chat *Chat, mentionedOrReplied bool) {
-	chat.UnviewedMessagesCount++
-	if mentionedOrReplied {
+func (m *Messenger) updateUnviewedCounts(chat *Chat, message *common.Message) {
+	if message.Timestamp > chat.LastMessage.Timestamp+1000*60*2 {
```
jrainville: Maybe we should put that in a constant? It's possible we might reuse it and it would be clearer what it achieves.

jrainville (commented):
Small comments.

Regarding the issue itself, I'm not sure if it's really something we want. As a user, if I see a marker that says "2 new messages" and then open the chat and see more than 2 new messages, I would think there was some kind of bug.

I couldn't open the discord link in the issue, so I don't have full context there.

John is on vacation, so if we want, we can merge this and discuss it with him later.

ilmotta (approved):
I agree with @jrainville, this requirement to increment the counter in a specified window of time can be confusing to users. A part of me would prefer to only solve this (potential) problem after gathering feedback from lots of users.

The 2min window is also questionable. Why not 30s, 15s, 1min if the goal is to reduce noise? If you extract to a constant as @jrainville suggested it'll be easier for us to change in the future.

But anyway, the requirements are clear on Discord and the code implements it correctly. Here's the excerpt from Discord for those without access:

> In the Messaging section of the app, the unread messages *counter* should
> always be incremented when:
>
> A) every time the user is @ mentioned in an ad-hoc group chat
>
> B) when the user receives a new message in a one on one chat.  There is also some logic so that multiple messages sent within a rolling time window in a 1on1 chat only increment the counter a single time, instead of incrementing the counter for each message.  I can't remember the exact values offhand, but it's something "when a 1on1 message is sent, it increments the counter by 1.  If another message is sent within 2min of the first message being sent, the counter is not incremented and the 2min window is reset.  If no further messages are received in the next 2min after the most recent message, when a subsequent message is eventually sent the counter is incremented again.
>
> EDIT: spotted that I'd forgotten to mention replies to the user's messages.
> All replies to the user's messages in an ad-hoc group chat should be treated
> in exactly the same way as an @ mention
>
> Re. the grey dot 'unread messages' indicator, this should only be shown for
> ad-hoc group chats when:
>
> A) there is a message in the ad-hoc group chat that doesn't @ mention the user
>
> AND
> 
> B) the unread message indicator for that ad-hoc group chat is "0" and is
> therefor not being displayed.
>
> For 1on1 chats, the grey dot unread message indicator should *never* be
> displayed, because there is never any need to display it as all 1on1 messages
> are treated as something that is counted in the counter, similar to @
> mentions.

jrainville (approved):

pavloburykh:
Hey @alwx ! Thanx for the PR. Just wondering why it has been merged into develop before it has been tested on client side and got QA approval?

cammellos:
@alwx reverted in the meantime

alwx:
oops that's totally my fault — merged the wrong PR :( thanks for reverting @cammellos

PR #3847, opened 2023-08-03
feat: fetch collection metadata when missing

### LINKED ISSUE #11782: Missing collection metadata for Arbitrum

dlipicar:
Fixes https://github.com/status-im/status-desktop/issues/11782

Collection data is now fetched separately when missing from the FullCollectibleData struct.

This is now mainly used for Infura (which, in turn, is used to fetch collectibles from Arbitrum Mainnet), which doesn't provide Collection metadata when fetching owned NFTs.

image

Collection names being shown means we now have collection metadata.

### DIFF
--- a/services/wallet/thirdparty/opensea/client.go
+++ b/services/wallet/thirdparty/opensea/client.go
@@ -119,6 +119,40 @@ func (o *Client) FetchAllCollectionsByOwner(chainID walletCommon.ChainID, owner 
 	return collections, nil
 }
 
+func (o *Client) FetchCollectionsDataByContractID(ids []thirdparty.ContractID) ([]thirdparty.CollectionData, error) {
+	ret := make([]thirdparty.CollectionData, 0, len(ids))
+
+	for _, id := range ids {
+		path := fmt.Sprintf("asset_contract/%s", id.Address.String())
+		url, err := o.urlGetter(id.ChainID, path)
+		if err != nil {
+			return nil, err
+		}
+
+		body, err := o.client.doGetRequest(url, o.apiKey)
+		if err != nil {
+			o.connectionStatus.SetIsConnected(false)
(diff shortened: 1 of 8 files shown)


services/wallet/thirdparty/opensea/client.go
```diff
@@ -119,6 +119,40 @@ func (o *Client) FetchAllCollectionsByOwner(chainID walletCommon.ChainID, owner
...
+	for _, id := range ids {
+		path := fmt.Sprintf("asset_contract/%s", id.Address.String())
+		url, err := o.urlGetter(id.ChainID, path)
+		if err != nil {
+			return nil, err
+		}
+
+		body, err := o.client.doGetRequest(url, o.apiKey)
+		if err != nil {
+			o.connectionStatus.SetIsConnected(false)
```
IvanBelyakoff, line 134: @dlipicar What bothers me is that we don't have any retry with backoff implementation in case of occasional network issues AFAIK for collectibles. So if server responds an error - we do retry, but for other reasons - no, we just set status to disconnected. And I personally see a lot of "no connection" banners in the app. Do we have any issue to implement that for collectibles?

IvanBelyakoff (approved):
LGTM.

services/wallet/thirdparty/opensea/client.go
dlipicar: Yes! Two tasks for that:
Error state + Manual retry: https://github.com/status-im/status-desktop/issues/11636
Implement main/fallback providers: https://github.com/status-im/status-desktop/issues/11800

dlipicar: Until now, we only used OpenSea. Now we have more than 1 provider for some chains, so it's possible to implement that.

PR #3848, opened 2023-08-03
feat: Allow TokenMaster and Owner without private key to add Tokens to CommunityDescription

mprakhov:
The Token Master role, part 2

- Allow TokenMaster and Owner without private key to add Tokens to CommunityDescription
- Added CommunityTokenPermission_BECOME_TOKEN_MASTER
- Check TokenMaster role permission during the reevaluating member permissions
- Unit test

Ticket: https://github.com/status-im/status-desktop/issues/11275


protocol/communities/manager.go
```diff
@@ -4129,3 +4117,51 @@ func (m *Manager) fixupChannelMembers() error {
...
+
+func (m *Manager) ReevaluatePrivelegedMember(community *Community, tokenPermissions []*protobuf.CommunityTokenPermission,
+	accountsAndChainIDs []*AccountChainIDsCombination, memberPubKey *ecdsa.PublicKey,
+	privilegedRole protobuf.CommunityMember_Roles, hasPrivilegedRole bool) (bool, error) {
+
+	hasPrivilegedRolePermissions := len(tokenPermissions) > 0
+	removeCurrentRole := false
+
+	if hasPrivilegedRolePermissions {
+		permissionResponse, err := m.checkPermissionToJoin(tokenPermissions, accountsAndChainIDs, true)
```
osmaczko: That's misleading, AFAICT we are not checking permissions to join here.
```suggestion
		permissionResponse, err := m.checkPermissions(tokenPermissions, accountsAndChainIDs, true)
```

protocol/communities/manager.go
```diff
@@ -4129,3 +4117,51 @@ func (m *Manager) fixupChannelMembers() error {
 
 	return nil
 }
+
+func (m *Manager) ReevaluatePrivelegedMember(community *Community, tokenPermissions []*protobuf.CommunityTokenPermission,
+	accountsAndChainIDs []*AccountChainIDsCombination, memberPubKey *ecdsa.PublicKey,
+	privilegedRole protobuf.CommunityMember_Roles, hasPrivilegedRole bool) (bool, error) {
```
osmaczko: These two arguments, `privilegedRole` and `hasPrivilegedRole`, next to each other are a bit confusing.
```suggestion
	privilegedRole protobuf.CommunityMember_Roles, alreadyHasPrivilegedRole bool) (bool, error) {
```

protocol/protobuf/community_update.proto
```diff
@@ -36,6 +37,7 @@ message CommunityEvent {
     COMMUNITY_MEMBER_KICK = 14;
     COMMUNITY_MEMBER_BAN = 15;
     COMMUNITY_MEMBER_UNBAN = 16;
+    COMMUNITY_TOKEN_ADD = 17;
```
osmaczko, line 40: Can TokenMasters and Owners only add tokens? Can't they remove the token?

protocol/communities_messenger_token_permissions_test.go
```diff
@@ -936,3 +936,261 @@ func (s *MessengerCommunitiesTokenPermissionsSuite) TestViewChannelPermissions()
...
+	community, err = s.owner.communitiesManager.GetByID(community.ID())
+	s.Require().NoError(err)
+	s.Require().True(community.HasTokenPermissions())
+
+	s.advertiseCommunityTo(community, s.alice)
+
+	var tokenPermission *protobuf.CommunityTokenPermission
+	for _, tokenPermission = range community.TokenPermissions() {
+		break
+	}
```
osmaczko, line 980: that's magic, lol :exploding_head: 
```suggestion
	s.Require().Len(community.TokenPermissions(), 1)
	tokenPermission := community.TokenPermissions()[0]
```

protocol/communities_messenger_token_permissions_test.go
```diff
@@ -936,3 +936,261 @@ func (s *MessengerCommunitiesTokenPermissionsSuite) TestViewChannelPermissions()
...
+	case protobuf.CommunityTokenPermission_BECOME_ADMIN:
+		s.Require().True(community.IsMemberAdmin(&s.alice.identity.PublicKey))
+	case protobuf.CommunityTokenPermission_BECOME_TOKEN_MASTER:
+		s.Require().True(community.IsMemberTokenMaster(&s.alice.identity.PublicKey))
+	default:
+		s.Require().True(false, "Unknown permission, please, update the test")
+	}
+
+	err = s.owner.communitiesManager.ReevaluateMembers(community)
+	s.Require().NoError(err)
```
osmaczko, line 993: Why this is done here?

protocol/communities_messenger_token_permissions_test.go
osmaczko, line 1: Props for these tests, really nice you added them!! :rocket:

osmaczko (approved):
Good job :muscle:

protocol/communities/community.go
```diff
@@ -836,11 +836,25 @@ func (o *Community) RemoveUserFromOrg(pk *ecdsa.PublicKey) (*protobuf.CommunityD
 func (o *Community) AddCommunityTokensMetadata(token *protobuf.CommunityTokenMetadata) (*protobuf.CommunityDescription, error) {
 	o.mutex.Lock()
 	defer o.mutex.Unlock()
-	if !o.IsControlNode() {
-		return nil, ErrNotControlNode
+
+	isControlNode := o.IsControlNode()
+	canManageTokens := o.IsOwnerWithoutCommunityKey() || o.IsTokenMaster()
```
jrainville: This name might be confusing, because in theory, the controlNode can also manage tokens. I'm not sure what else we can name it though.

One option is to change this to
```go
canManageTokens := o.IsControlNode() || o.IsOwnerWithoutCommunityKey() || o.IsTokenMaster()
``` 

Then on line 846, you can change that condition to `!isControlNode` instead.

The condition on line 842 can change to only `!canManageTokens` as well. 

What do you think?

protocol/protobuf/community_update.proto
jrainville: Not in the current design. Also, the token cannot be deleted from the blockchain, so it would only unlink from the community.

jrainville (approved):
Very good work! I only have a naming comment

protocol/communities/community.go
mprakhov: I would like to have `isControlNode` in a separate variable, as it was done for all other community "actions". Mixing will complicate the readiness of the code
I can rename it to `allowedToSendEvents` as it is on other community actions, or some other name :)

protocol/communities_messenger_token_permissions_test.go
mprakhov: one problem - this is a `map[string]*protobuf.CommunityTokenPermission` where key is unknown... you solution work with array only 😞

protocol/communities_messenger_token_permissions_test.go
mprakhov: In order to validate the scenario, when the user joined the community and after some time control node will call `ReevaluateMembers`. Here we are checking, that `ReevaluateMembers` fn works correctly when the user fits the role and his role is not changed.

protocol/communities/community.go
jrainville: I think `allowedToSendEvents` is slightly better, but in theory, the control ndoe can also send events if it wanted to.

Anyway, as long as everyone understands the code. What is you guys' preference? cc @osmaczko @0x-r4bbit

mprakhov: To be honest, I don't think that the control node should have the possibility to send community events, he must edit CommunityDescription directly

jrainville: Yeah, ok that makes sense. We can call it `allowedToSendEvents` then

protocol/communities_messenger_token_permissions_test.go
osmaczko: Can't we use a key then?

protocol/communities_messenger_token_permissions_test.go
osmaczko: > fn works correctly when the user fits the role and **his role is not changed**.

My point was that only `err` is checked here. Shouldn't we also check that roles have not changed?

protocol/communities/community.go
osmaczko: I would opt to be as explicit as possible -> `canSendTokenEvents`/`allowedToSendTokenEvents`. `allowedToSendEvents` is too generic, as an example: admins can send events but can't send token events.

protocol/communities_messenger_token_permissions_test.go
mprakhov: The key is unknown. This way I just extract the first element from the map with 1 element

osmaczko: Right, now the magic makes sense :magic_wand:

PR #3851, opened 2023-08-04
fix: Query token id for transactions

Cuteivist:
issue https://github.com/status-im/status-desktop/issues/11791

status-desktop PR: https://github.com/status-im/status-desktop/pull/11792

Passes the token ID to transaction result

### DIFF
--- a/services/wallet/activity/activity.go
+++ b/services/wallet/activity/activity.go
@@ -392,6 +392,7 @@ const (
 		1 AS agg_count,
 
 		transfers.token_address AS token_address,
+		transfers.token_id AS token_id,
 		NULL AS token_code,
 		NULL AS from_token_code,
 		NULL AS to_token_code,
@@ -468,6 +469,7 @@ const (
 		1 AS agg_count,
 
 		NULL AS token_address,
+		NULL AS token_id,
 		pending_transactions.symbol AS token_code,
 		NULL AS from_token_code,
 		NULL AS to_token_code,
@@ -522,6 +524,7 @@ const (
 		COALESCE(tr_status.count, 0) + COALESCE(pending_status.count, 0) AS agg_count,
 
 		NULL AS token_address,
+		NULL AS token_id,
 		NULL AS token_code,
 		multi_transactions.from_asset AS from_token_code,
 		multi_transactions.to_asset AS to_token_code,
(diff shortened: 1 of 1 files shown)


services/wallet/activity/activity.go
```diff
@@ -392,6 +393,11 @@ const (
 		1 AS agg_count,
 
 		transfers.token_address AS token_address,
+		transfers.token_id AS token_id,
+		CASE
+			WHEN transfers.token_id IS NULL THEN 1
+			ELSE 0
+		END as token_id_status,
```
stefandunca: It looks to me that it would be simpler and cleaner to read only token_id and handle `NULL` in `Scan` instead. Either extend `SQLBigIntBytes` to include a `Valid` member or scan `sql.RawBytes` and check fro its `len` (see nullifiable types like `sql.NullString`)

stefandunca (approved):
Great finding. Please see my notes inline.

services/wallet/activity/activity.go
Cuteivist: @stefandunca Good point. I updated it to much simpler way.

services/wallet/activity/activity.go
```diff
@@ -734,9 +742,9 @@ func getActivityEntries(ctx context.Context, deps FilterDependencies, addresses
 			// Extract tokens and chains
 			var involvedToken *Token
 			if tokenAddress != nil && *tokenAddress != ZeroAddress {
-				involvedToken = &Token{TokenType: Erc20, ChainID: common.ChainID(chainID.Int64), Address: *tokenAddress}
+				involvedToken = &Token{TokenType: Erc20, ChainID: common.ChainID(chainID.Int64), TokenID: tokenID, Address: *tokenAddress}
```
dlipicar, line 745: tokenID is exclusive to ERC721 (and ERC1155), shouldn't we be passing a nil here?

Cuteivist: Well if tokenID will be null in DB then it will be nil here too.

dlipicar: Sure, just saying since we know we're dealing with either an ERC20 or an ETH transfer, the concept of `tokenID` doesn't make sense, we shouldn't be using whatever we got from the db. 
Yeah it should be nil if whoever wrote to the DB did everything right, and it shouldn't be used by the client for these types of transfers anyway. What I mean is, somebody reading this code might thing there's some useful data in that field since we're making the effort of filling it in, which isn't true.

Cuteivist: @dlipicar  I agree. I will prepare small fix today.

PR #3853, opened 2023-08-04
chore: don't expose `RevealedAccount`s on community description

### LINKED ISSUE #11573: Members should NOT have access to other members' revealed accounts
# Description

Right now, the members' shared addresses are inside the community description. They should not, as that means that all members see them.

Only the control node should have access to those addresses (and later Token Masters and owners without the community private key)

# Acceptance criteria

Here is the list of things to change/do (kindof ordered):

1. Remove revealed accounts from the community description (member chat key is still needed)
2. Add a new SQL table for the control node to store the revealed accounts
3. Back up that table's content to waku, using the community key (can only be decrypted by control nodes)  in case the control node is lost
4. Change the way Admins receive and process requests to join
    - Members no longer send the full request to join to admins, they only send the abridged version that contains the chat key and other necessary params (without any address)
    - When an admin makes a decision about that request to join, it does **not** get propagated to users yet, it goes to the control node and other admins
    - The request is shown as "pending" for the admin afterwards (eg "Rejection pending...") @benjthayer will have designs later, but use a placeholder for now
    - Other admins receive that signal and display the same pending message as above. They cannot change the decision
    - The control node receives the decision and does normal control node stuff; it checks the permissions if needed, updates the description and sends it
    - The human owner doesn't have do do anything, this is automatic by the control node
    - When admins receive the new description, they can get rid of the pending request
5. All other admin actions that touch permissions or members use that same flow
    - They go in pending state until the control node executes the action

0x-r4bbit:
Prior to this commit a control node would add the revealed addresses to the member struct on the community description, which exposes all those addresses to the public.

We don't want that. Revealed addresses are exclusively shared with control nodes and should stay there (although, they might be privately shared among token masters, see
https://github.com/status-im/status-desktop/issues/11610).

In this commit, we no longer add the revealed addresses to the community description. The addresses are already stored in the requestToJoin database table so we can take them from there if we need them.

Closes: https://github.com/status-im/status-desktop/issues/11573

### DIFF
--- a/protocol/protobuf/communities.proto
+++ b/protocol/protobuf/communities.proto
@@ -23,6 +23,7 @@ message CommunityMember {
     ROLE_TOKEN_MASTER = 5;
   }
   repeated Roles roles = 1;
+  // deprecated, do not put accounts here
   repeated RevealedAccount revealed_accounts = 2;
   uint64 last_update_clock = 3;
 }
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -628,7 +628,7 @@ func (m *Manager) ReevaluateMembers(community *Community) error {
 
 	hasMemberPermissions := len(becomeMemberPermissions) > 0
 
-	for memberKey, member := range community.Members() {
+	for memberKey := range community.Members() {
 		memberPubKey, err := common.HexToPubkey(memberKey)
 		if err != nil {
 			return err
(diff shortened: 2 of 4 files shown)


protocol/protobuf/communities.proto
```diff
@@ -23,6 +23,7 @@ message CommunityMember {
     ROLE_TOKEN_MASTER = 5;
   }
   repeated Roles roles = 1;
+  // deprecated, do not put accounts here
   repeated RevealedAccount revealed_accounts = 2;
```
osmaczko, line 27:
```suggestion
  repeated RevealedAccount revealed_accounts = 2 [deprecated=true];
```
Then regenerate protobuf, it will add deprecated comment to golang generated code.

osmaczko (approved):

jrainville (approved):
Looks good. I only reviewed the second commit, since I think this need a rebase.

So in the front-end, I used check the community description to show members and their airdrop address. I guess I'll need to ask for them in the requests to join now?

0x-r4bbit:
> So in the front-end, I used check the community description to show members and their airdrop address. I guess I'll need to ask for them in the requests to join now?

I guess it'd make sense to introduce a new messenger API for `GetRevealedAddresses(memberPk, communityID)`.
I've already added that to `communitiesManager` so one can take it from there.

The underlying data lives in the `communities_requests_to_join_revealed_addresses` table, so I'm not sure if you have that data already available in desktop via requests-to-join

0x-r4bbit:
> . I guess I'll need to ask for them in the requests to join now?

@jrainville this is the API that can be used to retrieve revealed addresses given a member + community ID
https://github.com/status-im/status-go/pull/3853/files#diff-0c22647108517d38489ae5faa65a914ade3be02ea7a7e37355a82739da9aeb69R4310-R4313

PR #3857, opened 2023-08-07
fix: Extract `PairDevices` func

igor-sirotin:
Just a minor refactoring to use a common `PairDevices` function everywhere, instead of copying it all around.


qfrank (approved):
LGTM

Samyoul (commented):
This is nice. Good catch. 🙈 There are probably a lot of instances of code duplication in the tests.

PR #3860, opened 2023-08-07
Chore/11592 move collectibles data cache to db

### LINKED ISSUE #11592: Move collectibles data cache to DB

dlipicar:
Closes https://github.com/status-im/status-desktop/issues/11592

Collectibles data is now cached in the DB instead of memory.
A couple bugs were discovered and fixed in the thirdparty provider clients.

image

image


appdatabase/migrations/sql/1691173699_add_collectibles_and_collections_data_cache.up.sql
```diff
@@ -0,0 +1,29 @@
+CREATE TABLE IF NOT EXISTS collectible_data_cache (
```
IvanBelyakoff, line 1: @dlipicar It is advised to not use `IF NOT EXISTS` in migration files because if something wrong happens and the migration will run again, you ll see the error, while with `IF NOT EXISTS` you wont.
[Mentioned here](https://github.com/golang-migrate/migrate/blob/master/GETTING_STARTED.md)

appdatabase/migrations/sql/1691173699_add_collectibles_and_collections_data_cache.up.sql
```diff
@@ -0,0 +1,29 @@
...
+    permalink VARCHAR NOT NULL,
+    image_url VARCHAR NOT NULL,
+    animation_url VARCHAR NOT NULL,
+    animation_media_type VARCHAR NOT NULL,
+    traits BLOB NOT NULL,
+    background_color VARCHAR NOT NULL,
+    token_uri VARCHAR NOT NULL
+);
+
+CREATE UNIQUE INDEX IF NOT EXISTS collectible_data_identify_entry ON collectible_data_cache (chain_id, contract_address, token_id);
```
IvanBelyakoff, line 16: The same for index I believe

appdatabase/migrations/sql/1691173699_add_collectibles_and_collections_data_cache.up.sql
```diff
@@ -0,0 +1,29 @@
...
+
+CREATE TABLE IF NOT EXISTS collection_data_cache (
```
IvanBelyakoff, line 31: same

appdatabase/migrations/sql/1691173699_add_collectibles_and_collections_data_cache.up.sql
```diff
@@ -0,0 +1,29 @@
...
+    chain_id UNSIGNED BIGINT NOT NULL,
+    contract_address VARCHAR NOT NULL,
+    provider VARCHAR NOT NULL,
+    name VARCHAR NOT NULL,
+    slug VARCHAR NOT NULL,
+    image_url VARCHAR NOT NULL,
+    traits BLOB NOT NULL
+);
+
+CREATE UNIQUE INDEX IF NOT EXISTS collection_data_identify_entry ON collection_data_cache (chain_id, contract_address);
```
IvanBelyakoff, line 40: same

services/wallet/collectibles/collectible_data_db.go
```diff
@@ -0,0 +1,164 @@
...
+	tx, err = o.db.Begin()
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err == nil {
+			err = tx.Commit()
+			return
+		}
+		_ = tx.Rollback()
```
IvanBelyakoff, line 152: Do we need to rollback here? I mean if we've got data for multiple collectibles and managed to update some successfully, do we need to revert the successful changes?
I might be wrong about the type of data though

appdatabase/migrations/sql/1691173699_add_collectibles_and_collections_data_cache.up.sql
stefandunca: Thanks @IvanBelyakoff, for the docs hint. However, it doesn't look like clear recommendations in there:
- "Consider [making your migrations idempotent](https://github.com/golang-migrate/migrate/blob/856ea12df9d230b0145e23d951b7dbd6b86621cb/GETTING_STARTED.md?plain=1#L17C1-L17C44) - we can **run the same sql code twice in a row** with the same result. This makes our migrations more robust."
- "On the other hand, it causes slightly less control over database schema - e.g. let's say you forgot to drop the table in **down migration**."

Given that we don't do down migrations, I think the first recommendation might fit better our way of migrating. What do you think?

services/wallet/collectibles/collectible_data_db.go
```diff
@@ -0,0 +1,164 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (o *CollectibleDataDB) SetData(collectibles []thirdparty.CollectibleData) (err error) {
+	var (
```
stefandunca: Just a personal note that i find this not very readable with only one var. Why not the simple and concise `tx, err := o.db.Begin()`. It will still be captured given that it is the first definition.

services/wallet/collectibles/collection_data_db.go
```diff
@@ -0,0 +1,145 @@
...
+		Traits: make(map[string]thirdparty.CollectionTrait),
+	}
+	err := row.Scan(
+		&c.ID.ChainID,
+		&c.ID.Address,
+		&c.Provider,
+		&c.Name,
+		&c.Slug,
+		&c.ImageURL,
+		&JSONBlob{&c.Traits},
```
stefandunca: No way to split that into separate columns or linked tables? Or is there no need for future inquiry into data?

services/wallet/collectibles/collectible_data_db.go
```diff
@@ -0,0 +1,164 @@
...
+
+func NewCollectibleDataDB(sqlDb *sql.DB) *CollectibleDataDB {
+	return &CollectibleDataDB{
+		db: sqlDb,
+	}
+}
+
+const collectibleDataColumns = "chain_id, contract_address, token_id, provider, name, description, permalink, image_url, animation_url, animation_media_type, traits, background_color, token_uri"
+
+func insertOrReplaceCollectiblesData(creator statementCreator, collectibles []thirdparty.CollectibleData) error {
```
alaibe: nit: upsertCollecitblesData

services/wallet/collectibles/db_types.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/status-im/status-go/services/wallet/thirdparty"
+)
+
+// JSONBlob type for marshaling/unmarshaling inner type to json.
+type JSONBlob struct {
+	data interface{}
+}
+
+// Scan implements interface.
+func (blob *JSONBlob) Scan(value interface{}) error {
```
stefandunca: We already have two similar implementations. Would you be able to use one of them?

services/wallet/collectibles/collectible_data_db.go
```diff
@@ -0,0 +1,164 @@
...
+	insertCollectible, err := creator.Prepare(fmt.Sprintf(`INSERT OR REPLACE INTO collectible_data_cache (%s) 
+																				VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`, collectibleDataColumns))
```
alaibe: is it me or indent is odd?

stefandunca (approved):
Great job. I left some questions inline.

services/wallet/collectibles/collectibles.go
```diff
@@ -2,12 +2,12 @@ package collectibles
```
alaibe, line 2: nit file name: manager?

alaibe (approved):

appdatabase/migrations/sql/1691173699_add_collectibles_and_collections_data_cache.up.sql
IvanBelyakoff: I see your point @stefandunca. I will address the second "we don't do down migration" statement first. I actually provided a down migration file for my wallet DB PR, but after you mentioned it now, I realised there is no code for doing down migration. Well, I am not very well aware of why we don't do down migrations for some cases, when it makes sense and prefer having broken databases for users instead, but I believe we might want to think it over.
Which leads to point 1 - I don't think we should prefer possible errors in migration files over the `idempotentness` they mention. I believe migration should give the same results, unless there were errors before.
But ok, since you found it arguable, I agree that it is up to @dlipicar if he wants to keep it.

dlipicar: Thanks for these comments! Interesting points on both sides. Not at all an expert on db-related topics, so my opinion should be taken with a huge grain of salt.
I believe the DB structure should be fully predictable at any point in time and any sort of "unexpected state" during migration should be inadmissible. This brings its own problems, basically if any problem were to occur during migration you'll need to roll back to a previous version of the app for it to work, but well, it's a tradeoff. If we all shared this opinion, then the "IF NOT EXISTS" would keep us from detecting unexpected states of the DB, which is undesirable.

However, I don't think that's how we're handling our DB changes in general (you can see a LOT of these `IF NOT EXISTS` in the migration files.

So, to summarize:
1) Now that we'll have our own DB for the wallet, it would be good to have a team-wide discussion (brainstorm?) about how we want to manage migration errors. My opinion leans towards Ivan's in this regard.
2) For now, I think it's better to be fault-tolerant since that's how we seem to have been (intentionally or by default) handling this everywhere else in the app.

services/wallet/collectibles/db_types.go
dlipicar: Wlil move this to a more common spot

services/wallet/collectibles/collectible_data_db.go
dlipicar: hmm the format is auto-set every time I save, there's a tab in there

services/wallet/collectibles/collectible_data_db.go
dlipicar: Hmm interesting point. Is the table guaranteed to be in a correct state if we don't rollback? That is, will it have all the latest updated entries until the one that made it fail (and no residue of the failing entry)?

services/wallet/collectibles/collection_data_db.go
dlipicar: Separate columns won't do, since each collectible/collection could own any number of traits. I could move this data to a separate table though. Don't think it's really needed unless in the future we want to filter collectibles by some trait (unlikely), though it's probably better for table data inspection to have it that way.

dlipicar: Done!

services/wallet/collectibles/collectible_data_db.go
IvanBelyakoff: Good question. I was trying to dig into the code and find some info on that, but could not do it with a reasonable amount of time. But my point is the following:
1) I expect db.Exec() to begin a transaction and commit or rollback at the end of the statement execution. If we want to make several queries or statements that we want all or nothing, then we have to explicitly  do sql.Tx.Begin() and commit or rollback later.
2) In most cases we just log an error and/or display a popup in case of db.Exec() error. We don't call `panic()` or some `fatal` and just let the app proceed, which leads me to thinking that db is most likely in a consistent state after such errors, though no proof for now, sorry.
Feel free to correct me if I am wrong.

IvanBelyakoff (approved):
Good changes, @dlipicar

appdatabase/migrations/sql/1691173699_add_collectibles_and_collections_data_cache.up.sql
```diff
@@ -0,0 +1,51 @@
...
+    trait_type VARCHAR NOT NULL,
+    trait_value VARCHAR NOT NULL,
+    display_type VARCHAR NOT NULL,
+    max_value VARCHAR NOT NULL,
+    FOREIGN KEY(chain_id, contract_address, token_id) REFERENCES collectible_data_cache(chain_id, contract_address, token_id) 
+      ON UPDATE CASCADE 
+      ON DELETE CASCADE
+);
+
+CREATE TABLE IF NOT EXISTS collection_data_cache (
```
stefandunca, line 31: How about using a ROWID as a collection ID and add an optional `collectionID` in `collectible_data_cache` for a more explicit collection association?

stefandunca (approved):
Great job getting rid of JSON blob.

appdatabase/migrations/sql/1691173699_add_collectibles_and_collections_data_cache.up.sql
dlipicar: hmm I left them separate because some providers have separate API endpoints for collectible and collection data, so I might be able to have one without having the other.

dlipicar: oh that's why you said `optional`. Hmm is there a way I can link the "real" IDs somehow instead (that is, to say columns `chain_id` and `contract_address` from both collectibles and collections tables represent the same contract ID)? If not I'll have to manually manage the new ID column every time I add a collectible/collection.

services/wallet/collectibles/collectible_data_db.go
dlipicar: As a general rule, I also use Begin->Commit/Rollback for performance reasons whenever I'm updating the DB inside of a loop, to do a single batch-update of the DB instead of committing on each run of the loop.

appdatabase/migrations/sql/1691173699_add_collectibles_and_collections_data_cache.up.sql
dlipicar: Since collectible/collection metadata updates are independent, I'm gonna leave it as-is for now and merge this. I'll consider establishing a more elaborate relationship between the two later, if needed.

PR #3867, opened 2023-08-09
feat: sign community events

### LINKED ISSUE #11834: CommunityEventsMessage should aggregate signed events
### Description

Currently, `CommunityEventsMessage` is validated only based on the signature of the sender. The aggregated events within the message aren't signed, making them unverifiable by the control node. This vulnerability allows roles such as the owner, token master to be potentially compromised. For example, an admin could fabricate an event (like creating a token) that they don't have the rights for, and the propagated` CommunityEventsMessage` would still be accepted by the control node.

#### Acceptance criteria:
- [x] Sign each event individually.
- [x] ~~Control~~ Every node should verify both the signature and roles for each event.

osmaczko:
closes: status-im/status-desktop#11834


protocol/protobuf/community_update.proto
```diff
@@ -64,14 +64,25 @@ message ChannelData {
...
-  
+
   // Events base CommunityDescription with owner signature on top of which events were generated
   bytes events_base_community_description = 2;
-  
+
   // A list of admins events for the channel in bytes
-  repeated bytes events = 3;
+  // Deprecated: use signed_events instead.
+  repeated bytes events = 3 [deprecated=true];
```
mprakhov, line 84: why can't we just remove it?

protocol/communities/manager.go
```diff
@@ -1299,17 +1299,66 @@ func (m *Manager) handleCommunityDescriptionMessageCommon(community *Community,
 	}, nil
 }
 
+func (m *Manager) signEvents(community *Community) error {
```
mprakhov, line 1302: I think we should sign during the event adding (`addNewCommunityEvent`) and not iterate over all events, check its signature and sign at the end, before sending

protocol/communities/roles_authorization.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func RolesAuthorizedToPerformEvent(roles []protobuf.CommunityMember_Roles, event *CommunityEvent) bool {
+	if !canRolesPerformEvent(roles, event.Type) {
+		return false
+	}
+
+	if event.Type == protobuf.CommunityEvent_COMMUNITY_MEMBER_TOKEN_PERMISSION_CHANGE ||
+		event.Type == protobuf.CommunityEvent_COMMUNITY_MEMBER_TOKEN_PERMISSION_DELETE {
+		return canRolesModifyPermission(roles, event.TokenPermission.Type)
```
mprakhov, line 93: Here we need to check additionally, that Admin did not try to create CommunityTokenPermission_BECOME_TOKEN_MASTER or CommunityTokenPermission_BECOME_TOKEN_ADMIN

Is it TODO?

protocol/protobuf/community_update.pb.go
```diff
@@ -417,24 +417,76 @@ func (m *ChannelData) GetChannel() *CommunityChat {
...
 type CommunityEventsMessage struct {
 	CommunityId []byte `protobuf:"bytes,1,opt,name=community_id,json=communityId,proto3" json:"community_id,omitempty"`
 	// Events base CommunityDescription with owner signature on top of which events were generated
 	EventsBaseCommunityDescription []byte `protobuf:"bytes,2,opt,name=events_base_community_description,json=eventsBaseCommunityDescription,proto3" json:"events_base_community_description,omitempty"`
 	// A list of admins events for the channel in bytes
-	Events               [][]byte `protobuf:"bytes,3,rep,name=events,proto3" json:"events,omitempty"`
-	XXX_NoUnkeyedLiteral struct{} `json:"-"`
-	XXX_unrecognized     []byte   `json:"-"`
-	XXX_sizecache        int32    `json:"-"`
+	// Deprecated: use signed_events instead.
```
mprakhov, line 476: I'm not sure that it is correct to edit pb.go files

protocol/protobuf/community_update.proto
osmaczko: > If you [update](https://protobuf.dev/programming-guides/proto3/#updating) a message type by entirely deleting a field, or commenting it out, future developers can reuse the field number when making their own updates to the type. This can cause severe issues, as described in [Consequences of Reusing Field Numbers](https://protobuf.dev/programming-guides/proto3/#consequences).

Long story short, removing fields from protobufs is not safe. Anyway, I think I will go with `reserved` approach rather than with `[deprecated=true]`.

protocol/communities/manager.go
osmaczko: I tried that but `addNewCommunityEvent` does not have access to member's private key. I preferred to delegate signing to manager instead of propagating each `Community` instance with private key.

protocol/communities/roles_authorization.go
osmaczko: It is not TODO, this method covers exactly this.

https://github.com/status-im/status-go/blob/2a8438e64c0e2bcafb64ee4b49a24e9f89252d18/protocol/communities/roles_authorization.go#L39-L43

protocol/protobuf/community_update.pb.go
osmaczko: This is auto-generated, comments are copied from *.proto files.

protocol/protobuf/community_update.proto
osmaczko: > Non-required fields can be removed, as long as the field number is not used again in your updated message type. You may want to rename the field instead, perhaps adding the prefix “OBSOLETE_”, or make the field number [reserved](https://protobuf.dev/programming-guides/proto2/#fieldreserved), so that future users of your .proto can’t accidentally reuse the number.

More formal way (and better IMO) of adding "OBSOLETE_" is `[deprecated=true]`.

mprakhov: hm... I was changing it before...

mprakhov (approved):
LGTM 👍

protocol/communities/manager.go
mprakhov: the only cons I see right now is that if somehow you'll get an unsigned event - you will sign it from your name
But ideally this should not happen :)

osmaczko: Right, good concern. In current codebase the events are validated when they are received and the ones without signature or with invalid signature are filtered out, so it should not happen.

protocol/protobuf/community_update.proto
osmaczko: Changing the name is ok as long as type and semantics stays the same. Removing a field and reusing its number for something different will most likely lead to issues. Maybe you were lucky because you changed message that nobody uses.

Once you realize that older serialized messages may still contain the deleted field and that fields are identified by numbers, then you will understand why deleting fields in protobufs is not safe.

PR #3868, opened 2023-08-09
wallet acc syncing updates

saledjenic:
- accounts keep the origin operability when local pairing
- synced from field updated if recovered keypairs made fully operable


rasom (approved):

protocol/protobuf/pairing.proto
```diff
@@ -301,6 +301,7 @@ message SyncAccount {
   int64 position = 13;
   string prodPreferredChainIDs = 14;
   string testPreferredChainIDs = 15;
+  string operable = 16;
```
alaibe, line 304: why do u need to sync this? this should be local to each device no?

saledjenic: An explanation is placed in the comment above `MarkKeypairFullyOperable` function.

The reason why I added it is because of local pairing of recovered keypairs, cause in that case we need to transfer the state exactly as it's on another device, including account's operability.

alaibe (approved):

PR #3870, opened 2023-08-09
fix(communities_manager): fix joining community with admin permission

### LINKED ISSUE #11853: Unable to send requests to join when there is an admin permission
# Bug Report

## Description

If a community has only a permission to become admin, or a join permission and  a permission to become admin, even if the user meets the permission to join, the request will not be sent.

## Steps to reproduce

1. Create a community with a permission to be an admin (mine was with an NFT, but I'm pretty sure any token works)
2. Invite a user to it
3. The user clicks the button and gets the error that permissions are not met

#### Expected behavior

The request is sent correctly

#### Actual behavior

The request is not sent. There is an error about permissions not being met.

### Additional Information

- Operating System: Ubuntu 20

jrainville:
Fixes https://github.com/status-im/status-desktop/issues/11853

Fixes an issue where if a community had an admin permission, it would be impossible to join as a normal member because the admin permission wasn't satisfied.

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -1671,7 +1671,7 @@ func (m *Manager) CheckPermissionToJoin(id []byte, addresses []gethcommon.Addres
 
 	accountsAndChainIDs := combineAddressesAndChainIDs(addresses, allChainIDs)
 
-	if len(permissionsToJoin) == 0 {
+	if len(becomeMemberPermissions) == 0 || len(permissionsToJoin) == 0 {
 		// There are no permissions to join on this community at the moment,
 		// so we reveal all accounts + all chain IDs
 		response := &CheckPermissionsResponse{
@@ -1682,6 +1682,7 @@ func (m *Manager) CheckPermissionToJoin(id []byte, addresses []gethcommon.Addres
 		return response, nil
 	}
 	return m.checkPermissions(permissionsToJoin, accountsAndChainIDs, false)
+
 }
 
 func (m *Manager) accountsSatisfyPermissionsToJoin(community *Community, accounts []*protobuf.RevealedAccount) (bool, protobuf.CommunityMember_Roles, error) {
--- a/protocol/communities_messenger_token_permissions_test.go
+++ b/protocol/communities_messenger_token_permissions_test.go
@@ -698,6 +698,150 @@ func (s *MessengerCommunitiesTokenPermissionsSuite) TestBecomeMemberPermissions(
 	s.Require().Equal(msg.Text, response.Messages()[0].Text)
 }
 
+func (s *MessengerCommunitiesTokenPermissionsSuite) TestJoinCommunityWithAdminPermission() {
+	community, _ := s.createCommunity()
+
+	// setup become admin permission
+	permissionRequest := requests.CreateCommunityTokenPermission{
+		CommunityID: community.ID(),
+		Type:        protobuf.CommunityTokenPermission_BECOME_ADMIN,
+		TokenCriteria: []*protobuf.TokenCriteria{
+			&protobuf.TokenCriteria{
+				Type:              protobuf.CommunityTokenType_ERC20,
+				ContractAddresses: map[uint64]string{testChainID1: "0x123"},
+				Symbol:            "TEST",
+				Amount:            "100",
+				Decimals:          uint64(18),
+			},
+		},
+	}
+
+	response, err := s.owner.CreateCommunityTokenPermission(&permissionRequest)
+	s.Require().NoError(err)
+	s.Require().Len(response.Communities(), 1)
+
+	s.advertiseCommunityTo(community, s.bob)
+
+	// Bob should still be able to join even if there is a permission to be an admin
+	s.joinCommunity(community, s.bob, bobPassword, []string{})
+
+	// Verify that we have Bob's revealed account
+	revealedAccounts, err := s.owner.GetRevealedAccounts(community.ID(), common.PubkeyToHex(&s.bob.identity.PublicKey))
+	s.Require().NoError(err)
+	s.Require().Len(revealedAccounts, 1)
+	s.Require().Equal(bobAddress, revealedAccounts[0].Address)
+}
+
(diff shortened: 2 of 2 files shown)


jrainville:
Please let me know if there might be side effects that I might have missed. I didn't know at which level to fix the issue. 

`checkPermissions` seems more generic, so putting logic about admin vs member permissions didn't seem right, but maybe it would be better because it's the one that's used everywhere?

`CheckPermissionToJoin` kinda makes sense because it's really what is used when trying to join the community specifically.

I added two tests and the first test failed before my fix and works now.

protocol/communities/manager.go
```diff
@@ -1681,7 +1681,33 @@ func (m *Manager) CheckPermissionToJoin(id []byte, addresses []gethcommon.Addres
 		}
 		return response, nil
 	}
-	return m.checkPermissions(permissionsToJoin, accountsAndChainIDs, false)
+	checkPermissionsResponse, err := m.checkPermissions(permissionsToJoin, accountsAndChainIDs, false)
```
mprakhov: I think it can be done simpler


```
if len(becomeMemberPermissions) == 0 {
		// There is no permission to join, we can just say it's satisfied
		checkPermissionsResponse.Satisfied = true
		return checkPermissionsResponse, nil
}

return m.checkPermissions(permissionsToJoin, accountsAndChainIDs, true) // change false to true
```

protocol/communities/manager.go
```diff
@@ -1681,7 +1681,33 @@ func (m *Manager) CheckPermissionToJoin(id []byte, addresses []gethcommon.Addres
...
+				for _, criteria := range checkedPermission.Criteria {
+					if criteria == true {
+						checkPermissionsResponse.Satisfied = true
+						return checkPermissionsResponse, nil
+					}
+				}
+			}
+		}
+	}
+	return checkPermissionsResponse, nil
```
osmaczko: > **If a community has only a permission to become admin**, or a join permission and a permission to become admin, even if the user meets the permission to join, the request will not be sent.

The bold part could be simply fixed by the following:

```diff
diff --git a/protocol/communities/manager.go b/protocol/communities/manager.go
index ecd9f8120..cce86bc56 100644
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -1726,7 +1726,7 @@ func (m *Manager) CheckPermissionToJoin(id []byte, addresses []gethcommon.Addres
 	}
 	accountsAndChainIDs := combineAddressesAndChainIDs(addresses, allChainIDs)
 
-	if len(permissionsToJoin) == 0 {
+	if len(becomeMemberPermissions) == 0 || len(permissionsToJoin) == 0 {
 		// There are no permissions to join on this community at the moment,
 		// so we reveal all accounts + all chain IDs
 		response := &CheckPermissionsResponse{
```

--- 

> If a community has only a permission to become admin, **or a join permission and a permission to become admin**, even if the user meets the permission to join, the request will not be sent.

This indicates an issue inside the `checkPermissions`.  `checkPermissions` is permission type agnostic so it shouldn't matter what type of permissions are passed there. It should return `Satisfied == true` if at least one of the permissions is satisfied.

osmaczko (changes requested):

protocol/communities/manager.go
osmaczko: Changing false to true is not the solution, it will only affect shortcircuit parameter `if shortcircuit is true, it will stop as soon as we know the answer`. If user is expecting full response, to check what type of permissions it satisfies exactly, then it will break its expectations.

protocol/communities/manager.go
jrainville: > > If a community has only a permission to become admin, or a join permission and a permission to become admin, even if the user meets the permission to join, the request will not be sent.

> This indicates an issue inside the checkPermissions. checkPermissions is permission type agnostic so it shouldn't matter what type of permissions are passed there. It should return Satisfied == true if at least one of the permissions is satisfied.

I tried this scenario again (commented most of my new code) and ran my new tests and they still pass. I just kept the small change you suggested. Maybe there was something else happening when i manually tested or maybe my tests are not good?

jrainville:
You guys can re-review. I simplified a lot like @osmaczko suggested. I also check the revealed accounts just in case in the tests.

The tests still pass, so it's possible that I had a false negative with regards of admin + join permission failing :shrug:

jrainville:
I found a small issue in my test. I was using the same address for both tokens. I fixed that. The tests still pass.

I'll try to do the manual join test now

protocol/communities_messenger_token_permissions_test.go
```diff
@@ -698,6 +698,150 @@ func (s *MessengerCommunitiesTokenPermissionsSuite) TestBecomeMemberPermissions(
...
+	s.Require().Len(response.Communities(), 1)
+
+	// make bob satisfy the member criteria
+	s.makeAddressSatisfyTheCriteria(testChainID1, bobAddress, permissionRequestMember.TokenCriteria[0])
+
+	s.advertiseCommunityTo(community, s.bob)
+
+	// Bob should still be able to join even though he doesn't satisfy the admin requirement
+	// because he satisfies the member one
+	s.joinCommunity(community, s.bob, bobPassword, []string{})
```
mprakhov, line 781: should you pass the address here?
As I understand, everyone should provide address

jrainville:
I tested manually in Desktop and it works fine. Both with only admin and admin + join permissions. I guess I had a false negative before :shrug:

protocol/communities/manager.go
mprakhov: as an option, try to comment on your changes and run tests. Accordingly to the MR title, `TestJoinCommunityWithAdminPermission` must fail

protocol/communities_messenger_token_permissions_test.go
jrainville: Not necessarily. The code is made so that if no address is passed, we will just pass all addresses. In the UI, that's the right button called just "Join" or something (without manually selecting addresses)

mprakhov (approved):

osmaczko (approved):

PR #3872, opened 2023-08-10
Add preview and counters for CR system messages #16898

flexsurfer:
1. Add counters for "you were removed from contact" and "now is contact" system messages
Steps:

2. Add preview for "you were removed from contact"/"you removed from contact" system messages

status-react :https://github.com/status-im/status-mobile/issues/16898


protocol/messenger_handler.go
```diff
@@ -938,6 +938,11 @@ func (m *Messenger) handleAcceptContactRequestMessage(state *ReceivedMessageStat
 		}
 		state.Response.AddMessage(updateMessage)
 
+		chat.UnviewedMessagesCount++
```
MishkaRogachev: Isn't  better to increase counter after `UpdateFromMessage` because it can exit with error here?

MishkaRogachev (approved):
Great work! Here is one question you can ignore

PR #3873, opened 2023-08-10
fix: ensure CommunityDescription msg kept in db is signed

### LINKED ISSUE #11859: Don't keep unsigned CommunityDescription message in the database
> * do not keep unsigned CommunityDescription (this will require on each community request from db to restore the current state and track that we never save or backup/sync restored community)

I am opting for that option. As I've been saying from the start, we shouldn't store unsigned `CommunityDescription` message in the database. This not only causes syncing issues but I also believe it can lead to problems in various scenarios. I haven't tested this, but I am pretty sure that sharing community won't work if we have unsigned `CommunityDescription` in the database. We use `ShareCommunity` when inviting users to the community. AFAIK admins can invite users to the community. Consider the following stack trace:
```
community.ToBytes
sendChatMessage
SendChatMessage
SendChatMessages
ShareCommunity
```

If an admin sends an unsigned `CommunityDescription`, it will fail on the receiver's end, thus preventing users from viewing the invitation.

_Originally posted by @osmaczko in https://github.com/status-im/status-desktop/issues/11713#issuecomment-1662067681_

osmaczko:
closes: status-im/status-desktop#11859

### DIFF
--- a/protocol/communities/community_event.go
+++ b/protocol/communities/community_event.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/status-im/status-go/protocol/common"
 	"github.com/status-im/status-go/protocol/protobuf"
-	"github.com/status-im/status-go/protocol/v1"
 )
 
 func (o *Community) ToCreateChannelCommunityEvent(channelID string, channel *protobuf.CommunityChat) *CommunityEvent {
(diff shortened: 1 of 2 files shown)


jrainville (approved):
LGTM but I have no idea of the implications

mprakhov (approved):

PR #3874, opened 2023-08-10
chore: Common `createOtherDevice` for community_messenger_test

igor-sirotin:
Just a minor test refactoring:
- created a common `createOtherDevice` function not to copy the code
- fixed a type `send an community message`


osmaczko (approved):
Thanks

PR #3876, opened 2023-08-10
feat: auto re-apply rejected community events

### LINKED ISSUE #11552: Owner must acknowledge the admin about the rejected event due to the CommunityDescription clock change
### Description

We can't guarantee, that 2 admins won't send 2 events at the same time. If it will happen, we can face with next situation:
1. A1 sent E1
2. A2 sent E2
3. Owner receives A1 E1 and updates CommunityDescription clock and publishes CommunityDescription msg
4. Owner receives A2 E2 and rejects it as the CommunityDescription clock is outdated

If the owner will reject the A2 E2 event, A2 must acknowledge, that his CommunityEventMessage was rejected and try to resend it with an updated CommunityDescription clock

osmaczko:
closes: status-im/status-desktop#11552


protocol/v1/status_message.go
```diff
@@ -328,7 +328,9 @@ func (m *StatusMessage) HandleApplication() error {
 		return m.unmarshalProtobufData(new(protobuf.SyncActivityCenterNotifications))
 	case protobuf.ApplicationMetadataMessage_SYNC_ACTIVITY_CENTER_NOTIFICATION_STATE:
 		return m.unmarshalProtobufData(new(protobuf.SyncActivityCenterNotificationState))
-	case protobuf.ApplicationMetadataMessage_COMMUNITY_ADMIN_MESSAGE:
+	case protobuf.ApplicationMetadataMessage_COMMUNITY_EVENTS_MESSAGE:
+		return m.unmarshalProtobufData(new(protobuf.CommunityEventsMessage))
+	case protobuf.ApplicationMetadataMessage_COMMUNITY_EVENTS_MESSAGE_REJECTED:
```
cammellos, line 333: Wondering whether we should have a different protobuf for this (or add some fields to the existing one)? 
I would add a `Reason` for the rejection for example (most of the cases would be that is out of date, but it could also be lack of permissions for example).
Either just a small wrapper, that way we don't have to have a nested decision in `handle...`, where we check the `type` of the metadata message in a nested decision,
or just adding a field to the existing protobuf.

What do you think?

protocol/protobuf/application_metadata_message.proto
```diff
@@ -79,9 +79,10 @@ message ApplicationMetadataMessage {
     SYNC_ENS_USERNAME_DETAIL = 64;
     SYNC_ACTIVITY_CENTER_NOTIFICATION = 65;
     SYNC_ACTIVITY_CENTER_NOTIFICATION_STATE = 66;
-    COMMUNITY_ADMIN_MESSAGE = 67;
+    COMMUNITY_EVENTS_MESSAGE = 67;
```
mprakhov, line 82: I think now you can safely remove
`repeated bytes events = 3 [deprecated=true];` from protobuf

osmaczko: Not sure I am following :thinking: Why?

protocol/v1/status_message.go
osmaczko: Right, makes sense, having a custom protobuf will make it more flexible in the future. Will do :+1:

protocol/protobuf/application_metadata_message.proto
mprakhov: The previous protobuf can be recognized by the clients only if MessageType is COMMUNITY_ADMIN_MESSAGE
Taking into account, that you changed MessageType, your clients will not handle COMMUNITY_ADMIN_MESSAGE, they'll handle COMMUNITY_EVENTS_MESSAGE, and this MessageType is new in the network and the network doesn't know anything about the previous `protobuf.CommunityEventsMessage` format

cammellos: This change  is only local, it's just renaming a variable, but has not impact on what's passed over the wire, @osmaczko to confirm just to make sure, so protocol-wise everything is the same

mprakhov: ah... sorry, we will send an integer...

osmaczko: ~~Fields in a protobuf message are identified by field numbers, not by field names. Changing the name of the message does not have any impact, it is still the same message.~~

It is just an enum, the rename won't affect anything that goes over the wire.

protocol/v1/status_message.go
osmaczko: Updated, please take a second look.

protocol/messenger_communities.go
```diff
@@ -84,28 +84,51 @@ func (m *Messenger) publishOrg(org *communities.Community) error {
...
-	payload, err := adminMessage.Marshal()
+	rawMessage := common.RawMessage{
+		Payload: payload,
+		Sender:  m.identity,
+		// we don't want to wrap in an encryption layer message
+		SkipProtocolLayer: true,
+		MessageType:       protobuf.ApplicationMetadataMessage_COMMUNITY_EVENTS_MESSAGE,
+	}
+
+	_, err = m.sender.SendPublic(context.Background(), types.EncodeHex(msg.CommunityID), rawMessage)
```
cammellos, line 104: I am not sure these are retried, we should check probably, just so we know and we can address in the future (say the user goes offline just after retrieving the event)

cammellos (approved):

protocol/messenger_communities.go
osmaczko: Added TODOs

PR #3877, opened 2023-08-11
added nil checking for communitiesManager.GetByID

qfrank:
this should fix relate mobile [issue](https://github.com/status-im/status-mobile/issues/16983)


protocol/communities_messenger_test.go
```diff
@@ -2803,6 +2805,12 @@ func (s *MessengerCommunitiesSuite) TestCheckCommunitiesToUnmute() {
 
 }
 
+func (s *MessengerCommunitiesSuite) TestCommunityNotInDB() {
+	community, err := s.alice.communitiesManager.GetByID([]byte("0x123"))
+	require.Nil(s.T(), err)
```
cammellos: `s.Require().Nil(err)`
and
`s.Require().Nil(community)` 
so you don't have to import `testify/require`

cammellos (approved):

vkjr (approved):

PR #3878, opened 2023-08-11
Handle remove event out of order in private group chats

cammellos:
When we received a remove event from a private group chat out of order, the chat would not be created.
This was causing some issues if later on we received the previous event. This commit changes the behavior so that a chat is created.

### DIFF
--- a/protocol/messenger_group_chat_test.go
+++ b/protocol/messenger_group_chat_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/golang/protobuf/proto"
 	"github.com/stretchr/testify/suite"
 
 	userimage "github.com/status-im/status-go/images"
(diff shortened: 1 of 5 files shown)


osmaczko (approved):

smohamedjavid (approved):

protocol/messenger_group_chat_test.go
```diff
@@ -409,3 +410,47 @@ func (s *MessengerGroupChatSuite) TestGroupChatHandleDeleteMemberMessage() {
 	defer s.NoError(admin.Shutdown())
 	defer s.NoError(member.Shutdown())
 }
+
+func (s *MessengerGroupChatSuite) TestGroupChatMembersRemoval2() {
```
ibrkhalil: Should we rename this test to better describe the issue that was happening?

cammellos: oh yes, sorry, forgot to do that

ibrkhalil (approved):

PR #3879, opened 2023-08-11
feat: community privileged user sync message

mprakhov:
Introduce the `CommunityPrivilegedUserSyncMessage` message type that allows for sharing information between privileged members in a secure manner

Part 1: admin reject and
- Control node notify about the accepted request to join all privileged members
- Control node notify about rejected requests to join all privileged members
- Fixed tests

Related to # https://github.com/status-im/status-desktop/issues/11715

Part 2 with sharing CommunityToken will be in a separate MR


protocol/messenger_communities.go
```diff
@@ -315,6 +316,30 @@ func (m *Messenger) handleCommunitiesSubscription(c chan *communities.Subscripti
...
+					for _, requestID := range sub.RejectedRequestsToJoin {
+						reject := &requests.DeclineRequestToJoinCommunity{
+							ID: requestID,
+						}
+						_, err := m.DeclineRequestToJoinCommunity(reject)
+						if err != nil {
+							m.logger.Warn("failed to decline request to join ", zap.Error(err))
+						}
+					}
+				}
```
mprakhov: hm... I wonder why it is displayed as a new code... I only removed 2 lines

protocol/communities/manager.go
```diff
@@ -1478,6 +1486,64 @@ func (m *Manager) HandleCommunityEventsMessageRejected(signer *ecdsa.PublicKey,
...
+
+	switch message.Type {
+	case protobuf.CommunityPrivilegedUserSyncMessage_CONTROL_NODE_ACCEPT_REQUEST_TO_JOIN:
+		fallthrough
+	case protobuf.CommunityPrivilegedUserSyncMessage_CONTROL_NODE_REJECT_REQUEST_TO_JOIN:
+		var state RequestToJoinState
+		if message.Type == protobuf.CommunityPrivilegedUserSyncMessage_CONTROL_NODE_ACCEPT_REQUEST_TO_JOIN {
+			state = RequestToJoinStateAccepted
+		} else {
+			state = RequestToJoinStateDeclined
```
jrainville, line 1528: I think if the message is from someone other than the control node, the state should be `PendingDeclined` as per Pascal's work on having all member messages need to hit the control node before being official

protocol/communities_events_owner_without_community_key_test.go
```diff
@@ -116,13 +116,6 @@ func (s *OwnerWithoutCommunityKeyCommunityEventsSuite) TestOwnerAcceptMemberRequ
 	testAcceptMemberRequestToJoinResponseSharedWithOtherEventSenders(s, community, user, additionalOwner)
 }
 
-func (s *OwnerWithoutCommunityKeyCommunityEventsSuite) TestOwnerAcceptMemberRequestToJoinNotConfirmedByControlNode() {
-	community := setUpOnRequestCommunityAndRoles(s, protobuf.CommunityMember_ROLE_OWNER, []*Messenger{})
-	// set up additional user that will send request to join
-	user := s.newMessenger("", []string{})
-	testAcceptMemberRequestToJoinNotConfirmedByControlNode(s, community, user)
-}
```
jrainville, line 124: I think we still want those tests

protocol/communities_events_utils_test.go
```diff
@@ -875,26 +875,27 @@ func testAcceptMemberRequestToJoin(base CommunityEventsTestsInterface, community
...
 
-	// user receives community admin event without being a member yet
-	response, err = WaitOnMessengerResponse(
+	// at this point, the request to join is marked as accepted by GetEventSender node
+	acceptedRequestsPending, err := base.GetEventSender().AcceptedPendingRequestsToJoinForCommunity(community.ID())
+	s.Require().NoError(err)
+	s.Require().Len(acceptedRequestsPending, 1)
+	s.Require().Equal(acceptedRequestsPending[0].PublicKey, common.PubkeyToHex(&user.identity.PublicKey))
+
+	// user should not receives community admin event without being a member yet
```
jrainville:
```suggestion
	// user should not receive community admin event without being a member yet
```

protocol/communities_events_utils_test.go
```diff
@@ -912,68 +913,39 @@ func testAcceptMemberRequestToJoin(base CommunityEventsTestsInterface, community
...
-	response, err := user.RequestToJoinCommunity(requestToJoin)
-	s.Require().NoError(err)
-	s.Require().NotNil(response)
-	s.Require().Len(response.RequestsToJoinCommunity, 1)
-
-	sentRequest := response.RequestsToJoinCommunity[0]
 
-	// event sender receives request to join
-	response, err = WaitOnMessengerResponse(
+	// event sender receives update community
```
jrainville:
```suggestion
	// event sender receives updated community
```

protocol/messenger_communities.go
jrainville: The code was duplicated. See above, it's there twice.

protocol/messenger_communities.go
```diff
@@ -1432,11 +1494,53 @@ func (m *Messenger) DeclineRequestToJoinCommunity(request *requests.DeclineReque
...
+	if err != nil {
+		return nil, err
+	}
+
+	community, err := m.communitiesManager.DeclineRequestToJoin(dbRequest)
 	if err != nil {
 		return nil, err
 	}
 
+	if community.IsControlNode() {
```
jrainville, line 1481: If it's an admin, it should send the message to other privileged users too, so that they can either put it in pending declined or decline it if it's the control node.

jrainville (commented):
Nice work.

I think I might have misunderstood the changes, but it looks like the the Pending state when non-control node privileged users accept/reject a membership request.

The Privileged message was in part so that admins could send messages to each other about who they approved/rejected without having the user receiving them.

protocol/communities/manager.go
```diff
@@ -1324,7 +1324,8 @@ func (m *Manager) signEvents(community *Community) error {
 	for i := range community.config.EventsData.Events {
 		communityEvent := &community.config.EventsData.Events[i]
 		if communityEvent.Signature == nil || len(communityEvent.Signature) == 0 {
-			err := communityEvent.Sign(m.identity)
+			var err error
+			communityEvent.Signature, err = crypto.Sign(crypto.Keccak256(communityEvent.Payload), m.identity)
```
osmaczko: Conflicts resolution went wrong?

protocol/communities/manager.go
```diff
@@ -1337,9 +1338,16 @@ func (m *Manager) validateAndFilterEvents(community *Community, events []Communi
...
+			return errors.New("missing signature")
+		}
+
+		signer, err := crypto.SigToPub(
+			crypto.Keccak256(event.Payload),
+			event.Signature,
+		)
 		if err != nil {
-			return err
+			return errors.New("failed to recover signer")
```
osmaczko: Same here?

protocol/communities/manager.go
mprakhov: this message type will be received only from the control node by all other privileged users.
If the admin accepts the request to join, admins will modify the state in their databases during handling community events

protocol/communities_events_utils_test.go
```diff
@@ -912,68 +913,39 @@ func testAcceptMemberRequestToJoin(base CommunityEventsTestsInterface, community
...
 
-	acceptRequestToJoin := &requests.AcceptRequestToJoinCommunity{ID: sentRequest.ID}
-	response, err = base.GetEventSender().AcceptRequestToJoinCommunity(acceptRequestToJoin)
-	s.Require().NoError(err)
-	s.Require().NotNil(response)
-	s.Require().Len(response.Communities(), 1)
-	// we don't expect `user` to be a member already, because `eventSender` merely
-	// forwards its accept decision to the control node
-	s.Require().False(response.Communities()[0].HasMember(&user.identity.PublicKey))
+	for i := 0; i <= 5; i++ {
```
osmaczko: `WaitOnMessengerResponse` already uses a retry mechanism (`RetryWithBackOff`) internally. Is this for loop really necessary?

protocol/communities_events_owner_without_community_key_test.go
mprakhov: The basic scenario with accepting/rejecting the request to join covers these steps, this is just a code duplication

protocol/messenger_communities.go
mprakhov: they will change the state during handling community events. It is already implemented and we have tests for this scenario

protocol/communities/manager.go
jrainville: Ok, so if I understand correctly, the new message type for privileged users is **only** sent by the control node?

I don't think that's exactly what the issue was asking. I think it was a type of message that privileged users can send each other, like a small group chat that normal members are not part of. 

Maybe @0x-r4bbit can confirm?

mprakhov: no, it can be sent by any privileged users, the sender decides who will receive this message :) 
In this case, it makes sense to verify that this 2 types of msg (accepted/rejected) comes from the control node

protocol/communities/manager.go
mprakhov: ah...s sorry for that, didn't notice

jrainville (approved):

protocol/communities/manager.go
osmaczko: Somehow this was not resolved properly :disappointed:

osmaczko (commented):
Late into the party, just wanted to say it looks ok :+1:

PR #3887, opened 2023-08-14
feat: transferring keystore files for selected keypair via local network

saledjenic:
There is a desktop app feature where we need to transfer keystore files for selected keypair/s only (of course, which are not migrated to a keycard, otherwise we wouldn't need to do that) via local network using QR code.

### DIFF
--- a/mobile/status.go
+++ b/mobile/status.go
@@ -1140,6 +1140,36 @@ func InputConnectionStringForBootstrappingAnotherDevice(cs, configJSON string) s
 	return makeJSONResponse(err)
 }
 
+// GetConnectionStringForExportingKeypairsKeystores starts a pairing.SenderServer
+// then generates a pairing.ConnectionParams. Used when the device is Logged in and therefore has Account keys
+// and the device might not have a camera, to transfer kestore files of provided key uids.
+func GetConnectionStringForExportingKeypairsKeystores(configJSON string) string {
+	if configJSON == "" {
+		return makeJSONResponse(fmt.Errorf("no config given, SendingServerConfig is expected"))
+	}
+
+	cs, err := pairing.StartUpSenderServer(statusBackend, configJSON)
+	if err != nil {
+		return makeJSONResponse(err)
+	}
+	return cs
+}
+
+// InputConnectionStringForImportingKeypairsKeystores starts a pairing.ReceiverClient
+// The given server.ConnectionParams string will determine the server.Mode
+// Used when the device is Logged in and has Account keys and has a camera to read a QR code
+//
+// Example: A mobile device (device with a camera) receiving account data from
+// a device with a screen (mobile or desktop devices)
+func InputConnectionStringForImportingKeypairsKeystores(cs, configJSON string) string {
+	if configJSON == "" {
+		return makeJSONResponse(fmt.Errorf("no config given, ReceiverClientConfig is expected"))
+	}
+
+	err := pairing.StartUpReceivingClient(statusBackend, cs, configJSON)
+	return makeJSONResponse(err)
+}
+
 func ValidateConnectionString(cs string) string {
 	err := pairing.ValidateConnectionString(cs)
 	if err == nil {
--- a/server/pairing/client.go
+++ b/server/pairing/client.go
@@ -429,31 +429,33 @@ func (c *ReceiverClient) sendInstallationData() error {
 }
 
 // setupReceivingClient creates a new ReceiverClient after parsing string inputs
-func setupReceivingClient(backend *api.GethStatusBackend, cs, configJSON string) (*ReceiverClient, error) {
+func setupReceivingClient(backend *api.GethStatusBackend, cs, configJSON string) (*ReceiverClient, *ReceiverClientConfig, error) {
 	ccp := new(ConnectionParams)
 	err := ccp.FromString(cs)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	conf := NewReceiverClientConfig()
 	err = json.Unmarshal([]byte(configJSON), conf)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
-	err = validateAndVerifyNodeConfig(conf, conf.ReceiverConfig)
+
+	err = validateReceiverClientConfig(backend, conf)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	conf.ReceiverConfig.DB = backend.GetMultiaccountDB()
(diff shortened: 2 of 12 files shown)


rasom (approved):

api/geth_backend.go
```diff
@@ -2120,3 +2120,7 @@ func (b *GethStatusBackend) SwitchFleet(fleet string, conf *params.NodeConfig) e
 func (b *GethStatusBackend) SetLocalPairing(value bool) {
 	b.localPairing = value
 }
+
+func (b *GethStatusBackend) GetLocalPairing() bool {
```
qfrank: no usage anywhere?

mobile/status.go
```diff
@@ -1140,6 +1140,50 @@ func InputConnectionStringForBootstrappingAnotherDevice(cs, configJSON string) s
 	return makeJSONResponse(err)
 }
 
+// GetConnectionStringForExportingKeypairsKeystores starts a pairing.SenderServer
+// then generates a pairing.ConnectionParams. Used when the device is Logged in and therefore has Account keys
+// and the device might not have a camera, to transfer kestore files of provided key uids.
+func GetConnectionStringForExportingKeypairsKeystores(configJSON string) string {
```
qfrank, line 1146: what's the difference with `GetConnectionStringForBootstrappingAnotherDevice` except the validation? can we reuse `GetConnectionStringForBootstrappingAnotherDevice`? Can we move the validations into file `server/pairing/common.go` ? 🙂

mobile/status.go
```diff
@@ -1140,6 +1140,50 @@ func InputConnectionStringForBootstrappingAnotherDevice(cs, configJSON string) s
...
+	return cs
+}
+
+// InputConnectionStringForImportingKeypairsKeystores starts a pairing.ReceiverClient
+// The given server.ConnectionParams string will determine the server.Mode
+// Used when the device is Logged in and has Account keys and has a camera to read a QR code
+//
+// Example: A mobile device (device with a camera) receiving account data from
+// a device with a screen (mobile or desktop devices)
+func InputConnectionStringForImportingKeypairsKeystores(cs, configJSON string) string {
```
qfrank, line 1164: same as above

server/pairing/client.go
```diff
@@ -429,31 +429,40 @@ func (c *ReceiverClient) sendInstallationData() error {
...
-		return nil, err
-	}
-	err = validateAndVerifyNodeConfig(conf, conf.ReceiverConfig)
-	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
-	conf.ReceiverConfig.DB = backend.GetMultiaccountDB()
+	if conf.ReceiverConfig.TransferringKeystoreFiles {
```
qfrank: I believe this and following validation can be moved into `server/pairing/common.go` 🤔

server/pairing/payload_receiver.go
```diff
@@ -185,11 +191,14 @@ func (aps *AccountPayloadStorer) storeKeys(keyStorePath string) error {
...
 		} else {
-			return ErrKeyFileAlreadyExists
+			if !aps.transferringKeystoreFiles {
+				return ErrKeyFileAlreadyExists
+			}
 		}
 	}
 
 	for name, data := range aps.keys {
+		fmt.Println("----STORE...", keyStorePath, "  name: ", name)
```
qfrank: maybe use log instead?

qfrank (approved):
Most of it looks good to me, I've left minor comments. ❤️

mobile/status.go
saledjenic: We can, but let's hear from @Samyoul cause he is working on some improvements, maybe he already moved it somewhere.

saledjenic:
@qfrank all your comments are addressed, validations moved to `common.go`, thanks.

qfrank:
> @qfrank all your comments are addressed, validations moved to `common.go`, thanks.

thank you! but i still see changes in `status.go` ?

saledjenic:
> thank you! but i still see changes in status.go ?

@qfrank yes, cause I moved some checks from there common.go.

Samyoul:
Thank you for your work on this @saledjenic , one thing I notice with this PR is that we now have conditional logic in the pairing process to accommodate the sending of key files only. But also the configs are conditionally set, which suggests that there should really be a dedicated process flow for only sending / receiving keys.

It makes more sense to have a dedicated "selected key transfer" flow, and a dedicated Sender type. Local pairing consists of multiple senders / receiver types, but you are not handling pairing here, so invoking the full pairing structure is not required. You'll notice that the structures in pairing are build compositionally, so you could create your own KeyPayload handlers, Sender / Receivers.

I notice that you've created new API endpoints, but use identical functionality to the existing API methods, presumably the only difference is the config values. I think that you are on to the right idea here, but instead of having new API methods use a config input to change behaviour, be explicit in the behaviour with a separate process flow for selected key only transfers.

I'm more than happy to discuss this further. What are your thoughts?

saledjenic:
@Samyoul firstly was thinking about creating a new type just for sending keystore files only, but then adding a property to an existing config was a faster solution and then went for it. But I agree that would be better.

I can change the approach, not a problem at all, if you have any other suggestions please share them, thanks.

saledjenic:
@Samyoul I've created a new PR could you check it https://github.com/status-im/status-go/pull/3911?

qfrank:
> Thank you for your work on this @saledjenic , one thing I notice with this PR is that we now have conditional logic in the pairing process to accommodate the sending of key files only. But also the configs are conditionally set, which suggests that there should really be a dedicated process flow for only sending / receiving keys.
> 
> It makes more sense to have a dedicated "selected key transfer" flow, and a dedicated Sender type. Local pairing consists of multiple senders / receiver types, but you are not handling pairing here, so invoking the full pairing structure is not required. You'll notice that the structures in pairing are build compositionally, so you could create your own KeyPayload handlers, Sender / Receivers.
> 
> I notice that you've created new API endpoints, but use identical functionality to the existing API methods, presumably the only difference is the config values. I think that you are on to the right idea here, but instead of having new API methods use a config input to change behaviour, be explicit in the behaviour with a separate process flow for selected key only transfers.
> 
> I'm more than happy to discuss this further. What are your thoughts?

You made an extrodinary comment 👍, i should learn the way of communication you've done , i didn't make a clarification for this [comment](https://github.com/status-im/status-go/pull/3887#discussion_r1295866376) @Samyoul

saledjenic:
Closing this in favor of this one https://github.com/status-im/status-go/pull/3911

PR #3889, opened 2023-08-14
Optimized counter increment logic for 1-1 chat messages

### LINKED ISSUE #16696: Optimized counter increment logic for 1-1 chat messages

alwx:
Fixes https://github.com/status-im/status-mobile/issues/16696

An updated version of https://github.com/status-im/status-go/pull/3845 (which was reverted)

### DIFF
--- a/protocol/common/message.go
+++ b/protocol/common/message.go
@@ -627,6 +627,18 @@ func (m *Message) PrepareContent(identity string) error {
 	return m.parseAudio()
 }
 
+func (m *Message) IsSystemMessage() bool {
+	return m.ContentType == protobuf.ChatMessage_UNKNOWN_CONTENT_TYPE ||
+		m.ContentType == protobuf.ChatMessage_SYSTEM_MESSAGE_CONTENT_PRIVATE_GROUP ||
+		m.ContentType == protobuf.ChatMessage_SYSTEM_MESSAGE_GAP ||
+		m.ContentType == protobuf.ChatMessage_CONTACT_REQUEST ||
+		m.ContentType == protobuf.ChatMessage_IDENTITY_VERIFICATION ||
+		m.ContentType == protobuf.ChatMessage_SYSTEM_MESSAGE_PINNED_MESSAGE ||
+		m.ContentType == protobuf.ChatMessage_SYSTEM_MESSAGE_MUTUAL_EVENT_SENT ||
+		m.ContentType == protobuf.ChatMessage_SYSTEM_MESSAGE_MUTUAL_EVENT_ACCEPTED ||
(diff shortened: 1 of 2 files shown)


protocol/messenger_handler.go
```diff
@@ -3008,9 +3015,14 @@ func (m *Messenger) isMessageAllowedFrom(publicKey string, chat *Chat) (bool, er
...
 }
 
-func (m *Messenger) updateUnviewedCounts(chat *Chat, mentionedOrReplied bool) {
-	chat.UnviewedMessagesCount++
-	if mentionedOrReplied {
+func (m *Messenger) updateUnviewedCounts(chat *Chat, message *common.Message) {
+	if chat == nil {
+		return
+	}
+	if chat.LastMessage == nil || chat.LastMessage.IsSystemMessage() || message.Timestamp > chat.LastMessage.Timestamp+IncreaseUnviewedMessagesCountTimeout {
```
alwx: An important update here compared to the previous version:  all the (potentially important) system messages always increase counter. That's done mainly because of the bug we had in the previous PR (https://github.com/status-im/status-go/pull/3845): all the messages sent in two minutes after the contact request was accepted were not increasing the counter, now it's fixed.

protocol/common/message.go
```diff
@@ -627,6 +627,18 @@ func (m *Message) PrepareContent(identity string) error {
 	return m.parseAudio()
 }
 
+func (m *Message) IsSystemMessage() bool {
+	return m.ContentType == protobuf.ChatMessage_UNKNOWN_CONTENT_TYPE ||
```
jrainville: Should Unknown really be considered a system message?

jrainville (approved):
LGTM.

ilmotta (approved):

protocol/common/message.go
alwx: Good point. Removed it from the list.

Issue #3675, opened 2023-06-26
⚠️ Group Encryption Key IDs Loop Every 50 Day ⚠️


Samyoul:
There is a tricky little type conversion bug in group encryption key ID generation. See the source code below:

https://github.com/status-im/status-go/blob/57b2432290347f15e0e452e6a38d76acef601bc5/protocol/encryption/encryptor.go#L612-L624

### TLDR:

Key IDs loop every 4294967296 ms or roughly every 50 days because of a value overflow on int64 to uint32 type conversion.

### Proof:

The key IDs looked wrong and so I looked into them:

```go
func Test(t *testing.T) {
	umill := time.Now().UnixMilli()
	umillU32 := (uint32)(umill)

	spew.Dump(umill, umillU32)

	found := []int64{}

	for {
		umill++
		if uint32(1) > (uint32)(umill) {
			found = append(found, umill)
			spew.Dump("wrap", umill, (uint32)(umill), "--------")

			if len(found) == 20 {
				break
			}
		}
	}

	spew.Dump(found)

	for i := range found {
		if i == 0 {
			continue
		}

		spew.Dump(found[i] - found[i-1])
	}

}
```

```text
(int64) 1687793068391
(uint32) 4165888359
(string) (len=4) "wrap"
(int64) 1687922147328
(uint32) 0
(string) (len=8) "--------"
(string) (len=4) "wrap"
(int64) 1692217114624
(uint32) 0
(string) (len=8) "--------"
(string) (len=4) "wrap"
(int64) 1696512081920
(uint32) 0
(string) (len=8) "--------"
(string) (len=4) "wrap"
(int64) 1700807049216
(uint32) 0
(string) (len=8) "--------"
(string) (len=4) "wrap"
(int64) 1705102016512
(uint32) 0
(string) (len=8) "--------"
(string) (len=4) "wrap"
(int64) 1709396983808
(uint32) 0
(string) (len=8) "--------"
(string) (len=4) "wrap"
(int64) 1713691951104
(uint32) 0
(string) (len=8) "--------"
(string) (len=4) "wrap"
(int64) 1717986918400
(uint32) 0
(string) (len=8) "--------"
(string) (len=4) "wrap"
(int64) 1722281885696
(uint32) 0
(string) (len=8) "--------"
(string) (len=4) "wrap"
(int64) 1726576852992
(uint32) 0
(string) (len=8) "--------"
(string) (len=4) "wrap"
(int64) 1730871820288
(uint32) 0
[...]

PR #3897, opened 2023-08-16
Fix Group Encryption Key ID Timestamp Overflow

Samyoul:
Description to follow:

Addresses #3675


igor-sirotin:
Closing as the root issue https://github.com/status-im/status-go/issues/3675 is closed

Issue #3884, opened 2023-08-12
Local Pairing - Network Address / Peer Discovery


Samyoul:
## Network Address / Peer Discovery

Revamp the preferred network discovery process

### Self IP address acquisition

1. Ask network interface for its preferred outbound IP address
    - The process currently relied on
    - ~~Add a required a self "confirmation" ping, to ensure the given address is network addressable.~~
      - ~~This code exists [see makeOutboundCheck()](#makeoutboundcheck)~~
      - This doesn't prove anything. Don't do this to "confirm" the IP address is network accessible.
    - On fail, fall through to stage 2
2. ~~device gets a list of its network addresses.~~
   - ~~Use `netIfaces, err := net.Interfaces()` [See snippet](#get-all-network-addresses)~~
   - ~~run filter for only local private net addrs~~
   - ~~open listener on 0.0.0.0~~
   - ~~make an outbound network call to all ips~~
      - ~~call should include the addressing ip so device knows which address is network addressable~~
   - ~~record all ip addresses that are able to ping the listener.~~
      - ~~hopefully the list will only ever contain 1 IP address~~
   - ~~On fail, fall through to stage 3~~
   - ^^^ This is a **NOT** a good idea! This will not prove anything, because the device IS a part of ALL these networks and so technically could self address via any of the networks.^^^
3. Attempt discovery via UDP multicast
    - The basic code for this functionality exists [see StartSearchForLocalPairingPeers()](#startsearchforlocalpairingpeers)
    - **Question :** How do **both** devices know when to start UPD discovery?
      - Perhaps the QR code device can add a UDP discovery search flag to the `ConnectionParams`
    - no guarantee that both devices get an IP address
      - [See UDP discovery failure mitigation](#udp-discovery-failure-mitigation)

###### makeOutboundCheck()

https://github.com/status-im/status-go/blob/e620633eb745cffd54e16717ad5ac8c23e0550de/server/pairing/preflight/preflight.go#L86-L119

###### StartSearchForLocalPairingPeers()

https://github.com/status-im/status-go/blob/6cf0162877d3be58254efd1188c57b42a10e1403/mobile/status.go#L1029-L1042

###### UDP discovery failure mitigation

 - Both devices listen on a tcp port chosen by the network interface
 - Attempt UDP discovery
 - In the UDP packet include:
   - the device's listener port number
   - ensure the notifier's id is stored for later connections via tcp
[...]

PR #3898, opened 2023-08-16
feat: network functions for local pairing

igor-sirotin:
Required for https://github.com/status-im/status-go/issues/3884

Implemented:
- `GetLocalAddressesForPairingServer`: provides a list of IP addresses to be used for local pairing server
- `FindReachableAddressesForPairingClient`: takes a list of server's IP addresses and returns a list of reachable IP address to be used for client connection


server/ips.go
```diff
@@ -20,3 +20,143 @@ func GetOutboundIP() (net.IP, error) {
...
+	nis, err := net.Interfaces()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, ni := range nis {
+		addrs, err := ni.Addrs()
+		if err != nil {
+			return nil, err
+		}
```
Samyoul: I wonder if this error should be logged and then `continue` the loop. You don't want an error with your first addr to terminate the whole process.

server/pairing/connection_test.go
```diff
@@ -1,9 +1,12 @@
 package pairing
 
 import (
+	"fmt"
+	"net"
 	"testing"
 
 	"github.com/stretchr/testify/suite"
+	"go.uber.org/zap"
```
Samyoul: I think that this will give you compile errors, for unused packages.

server/ips_test.go
```diff
@@ -0,0 +1,266 @@
+package server
+
+import (
+	"fmt"
+	"github.com/status-im/status-go/server/servertest"
+	"net"
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+	"go.uber.org/zap"
```
Samyoul:
```suggestion
	"fmt"
	"net"
	"testing"

	"github.com/stretchr/testify/suite"
	"go.uber.org/zap"
	
	"github.com/status-im/status-go/server/servertest"
```

This order should satisfy the linter

server/ips.go
```diff
@@ -20,3 +20,143 @@ func GetOutboundIP() (net.IP, error) {
...
+
+	var ips [][]net.IP
+
+	for _, ni := range nis {
+		var niIps []net.IP
+
+		addrs, err := ni.Addrs()
+		if err != nil {
+			return nil, err
+		}
```
Samyoul: Same as https://github.com/status-im/status-go/pull/3898/files#r1296946449

server/ips.go
```diff
@@ -20,3 +20,143 @@ func GetOutboundIP() (net.IP, error) {
...
+
+// FindReachableAddresses returns a filtered remoteIps array,
+// in which each IP matches one or more of given localNets.
+func FindReachableAddresses(remoteIps []net.IP, localNets []net.IPNet) []net.IP {
+	var result []net.IP
+	for _, localNet := range localNets {
+		for _, remoteIP := range remoteIps {
+			if localNet.Contains(remoteIP) {
+				result = append(result, remoteIP)
+			}
```
Samyoul, line 130: Is it possible that both devices may each have different networks but that both these networks contain a the same ip?

- Device a 
  - network a (with ip 1.2.3.4)
  - network b (with ip 5.6.7.8)
  - network c (with ip 9.0.1.2)
- Device b
  - network d (with ip 3.4.5.6)
  - network e (with ip 7.8.9.0)
  
Now imagine that network d also has an ip address `1.2.3.4` . This function would in theory give us a false positive as the devices don't share a network.

I think we need a 2nd step of attempting to call each `result` ip address, to make sure that we can communicate with the other device.

Samyoul (approved):
Good work, I like the overall approach, just a few points to consider.

server/ips.go
igor-sirotin: 🤔

igor-sirotin: First of all, afaik, it's an invalid network configuration when 2 devices has same IP address.
Probably we could (should?) still handle such cases and not consider same IPs as reachable. Do you think this would be good?

> I think we need a 2nd step of attempting to call each result ip address, to make sure that we can communicate with the other device.

This is why I eventually decided to return an array of IPs.
We can add another procedure to attempt to connect to each of them, but I thought maybe LP Client would do that 🤔

Samyoul: > Now imagine that network d also has an ip address `1.2.3.4` . This function would in theory give us a false positive as the devices don't share a network.

Ah, I wasn't clear. "*Now imagine that network d also has an ip address `1.2.3.4`*", what I should have said is "*Now imagine that network d also has an ip address `1.2.3.4`* **belonging to a device c**"

> We can add another procedure to attempt to connect to each of them, but I thought maybe LP Client would do that 🤔

Yes doing that in another stage is fine, I remember that part of the conversation;

igor-sirotin: Ok, I get you now 
Yes, that is possible and this algorithm will definitely take this as a reachable address. I thinkg this is a good case when UDP discovery should come into the house 🙂

PR #3899, opened 2023-08-16
refactor: simplify events creation logic

osmaczko:
- use `roles_authorization.go` APIs as the ultimate source of truth
- unify&simplify community modifications in terms of events creation

### DIFF
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -483,31 +483,26 @@ func (o *Community) CreateChat(chatID string, chat *protobuf.CommunityChat) (*Co
 	o.mutex.Lock()
 	defer o.mutex.Unlock()
 
-	isControlNode := o.IsControlNode()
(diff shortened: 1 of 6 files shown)


jrainville (approved):
LGTM

mprakhov (commented):
I don't have any objections, but the previous variant was more readable for me (IMHO).
You start investigate fn and see that you do not fit into requirements and have early return.

Here my brain need to parse if/else if/else in order to understand if I can execute this logic

osmaczko:
> I don't have any objections, but the previous variant was more readable for me (IMHO). You start investigate fn and see that you do not fit into requirements and have early return.
> 
> Here my brain need to parse if/else if/else in order to understand if I can execute this logic

Indeed, I came to the same conclusion and corrected it just before you posted the comment.

mprakhov (approved):

PR #3901, opened 2023-08-16
feat: Unfurl image URLs

### LINKED ISSUE #3761: Unfurl image URLs
# Problem

New URL unfurling scheme was implemented in https://github.com/status-im/status-go/pull/3471.
We should also unfurl image URLs on sender side and include a compressed preview in the message.

Link example: https://placehold.co/600x400@2x.png

# Acceptance Criteria

`UnfurlURLs` should unfurl image URLs. 
By "Image URLs" I suppose we mean that response `content-type` is `image/*`.

# Future Steps

Support the solution in desktop/mobile.

ilmotta:
Closes #3761 

### Summary

This PR adds support for unfurling static image URLs (not GIFs, not animated WebPs), such as https://placehold.co/600x400@2x.png. It also compresses images before returning them as data URIs to clients.

I'll run `generate` after a round of review to avoid polluting the PR.

### Demo

Screenshot showing the mobile app unfurling both WebP and JPEG image URLs (please, ignore UI inconsistencies).

### Important

- This PR is a potential breaking change because it adds a field `Type` to `common.message.LinkPreview` and a validation to check it exists. There's no fallback type. In the mobile client, I had to make a tiny change, which will become a PR (_to be added here_).
- I'm _assuming_ there's no data to migrate pertaining unfurled links (column `user_messages.unfurled_links`). This is a known pain point ever since https://github.com/status-im/status-go/pull/3471 was opened. Given that "unfurling" is still evolving, I'd say it's wise to finally move away from the serialized column approach to a separate table. I'd be interested to work on this problem soon (and maybe finish it before the next desktop release lands).

#### About compression

The compression strategy implemented in this PR leverages the existing function `images.CompressToFileLimits`. I didn't implement a more comprehensive logic to consider the possibility of multiple image URLs being unfurled simultaneously, and the requirement that all of them fit within the target message size. For now, I believe it's good enough.

#### About animated WebP

Quick note about animated WebP: if you try to copy **raw** Giphy URLs, very often you'll get animated WebP resources, and not true GIF ones, for example https://i.giphy.com/media/IaWMz9Ln8OWvf66z6k/giphy.webp. As far as I checked, we don't support animated WebP in status-go.

The requirements for GIFs are still under discussion (see https://github.com/status-im/status-go/pull/3821). Once all requirements are clear, we can come back to animated WebP URLs.


protocol/linkpreview/linkpreview.go
```diff
@@ -241,13 +245,105 @@ func (u OpenGraphUnfurler) unfurl() (common.LinkPreview, error) {
...
+	return preview, nil
+}
+
 func normalizeHostname(hostname string) string {
 	hostname = strings.ToLower(hostname)
 	re := regexp.MustCompile(`^www\.(.*)$`)
 	return re.ReplaceAllString(hostname, "$1")
 }
 
+// isSupportedImageContentType detects whether a URL ends with one of the
```
cammellos: Wrong function name I believe, `isSupportImageURL`

protocol/linkpreview/linkpreview.go
```diff
@@ -241,13 +245,105 @@ func (u OpenGraphUnfurler) unfurl() (common.LinkPreview, error) {
...
+// supported image extensions. It provides a quick way to identify whether URLs
+// should be unfurled as images without needing to retrieve the full response
+// body first.
+func isSupportedImageURL(url *neturl.URL) bool {
+	re := regexp.MustCompile(`(?i)^.+(png|jpg|jpeg|webp)$`)
```
cammellos: Worth considering not using a regex here and compare char by char maybe? I run a few benchmarks and it would be around 10/14 times faster

```
func isSupportedImageURL(url *url.URL) bool {
	path := url.Path
	l := len(path)
	
	return (l > 5 && lower(path[l-1]) == 'g' && lower(path[l-2]) == 'n' && lower(path[l-3]) == 'p' && lower(path[l-4]) == '.') ||
		(l > 4 && lower(path[l-1]) == 'g' && lower(path[l-2]) == 'e' && lower(path[l-3]) == 'p' && lower(path[l-4]) == 'j' && lower(path[l-5]) == '.') ||
		(l > 5 && lower(path[l-1]) == 'g' && lower(path[l-2]) == 'e' && lower(path[l-3]) == 'p' && lower(path[l-4]) == 'j' && lower(path[l-5]) == '.') ||
		(l > 5 && lower(path[l-1]) == 'p' && lower(path[l-2]) == 'b' && lower(path[l-3]) == 'e' && lower(path[l-4]) == 'w' && lower(path[l-5]) == '.') ||
		(l > 5 && lower(path[l-1]) == 'p' && lower(path[l-2]) == 'b' && lower(path[l-3]) == 'e' && lower(path[l-4]) == 'w' && lower(path[l-5]) == '.') ||
		(l > 5 && lower(path[l-1]) == 'o' && lower(path[l-2]) == 'b' && lower(path[l-3]) == 'e' && lower(path[l-4]) == 'w' && lower(path[l-5]) == '.') ||
		(l > 5 && lower(path[l-1]) == 'a' && lower(path[l-2]) == 'b' && lower(path[l-3]) == 'e' && lower(path[l-4]) == 'w' && lower(path[l-5]) == '.') ||
		(l > 5 && lower(path[l-1]) == 'f' && lower(path[l-2]) == 'b' && lower(path[l-3]) == 'e' && lower(path[l-4]) == 'w' && lower(path[l-5]) == '.')
}

func lower(c byte) byte {
	if c >= 'A' && c <= 'Z' {
		return c + ('a' - 'A')
	}
	return c
}
``` 

(chatgpt generated)
a bit ugly, but could be worth it if used many times and if it's a vector of attack

protocol/common/message.go
```diff
@@ -758,10 +759,20 @@ func (m *Message) LoadImage() error {
 	return nil
 }
 
+func isValidLinkPreviewThumbnail(thumbnail LinkPreviewThumbnail) bool {
+	return (thumbnail.DataURI == "" && thumbnail.Width == 0 && thumbnail.Height == 0) ||
+		(thumbnail.DataURI != "" && thumbnail.Width > 0 && thumbnail.Height > 0)
+}
+
 func isValidLinkPreviewForProto(preview LinkPreview) bool {
-	return preview.Title != "" && preview.URL != "" &&
```
cammellos, line 762: This would be the breaking change mentioned in the description, I think it's worth handling backward compatibility in this case, otherwise clients won't be able to see unfurled links across versions (not sure if you meant to say that you will address backward compatibility in a separate PR).
But it seems we just want to have a case where `Type` is not set, and use the old condition for those, "fingerprinting" the old version. 
We can drop compatibility when enough people are on the compatible build, what do you think?

protocol/linkpreview/linkpreview.go
ilmotta: Good catch. I don't know why I assumed the language server would rename the comment too. I thought that was a feature.

Edit: the language server actually does update the comment if I rename a function. So it seems to not be as reliable as I thought 🤷🏼

protocol/linkpreview/linkpreview.go
ilmotta: @cammellos, I think this level of performance is fine, because unfurling needs to happen every few hundred milliseconds at most (debounced), since the slowest path is IO bound, so the savings from this check should be negligible. Now that you mention about performance, I think I can optimize it enough by avoiding the recompilation of the regex all the time, so I can extract the compiled regex to a var.

About the security aspect, I understand historically regexps have been the source of some issues (not sure about Go's history), but if I can't use a regexp here, are you saying that in Go we shouldn't use regular expressions if the input string is out of our control?

protocol/common/message.go
ilmotta: Exactly, I opted for not having a fallback to the `link` type. Now that you mention, I think it's better to go that route. Will do it.

The backward compatibility issue I mentioned in the PR is more related to migrations. Currently it's not possible to easily migrate existing data with just SQL because the column is JSON serialized.

protocol/linkpreview/linkpreview.go
cammellos: > About the security aspect, I understand historically regexps have been the source of some issues (not sure about Go's history), but if I can't use a regexp here, are you saying that in Go we shouldn't use regular expressions if the input string is out of our control?

In general, we try to avoid it, as performance can vary and it's possible at times to find a weak spot in a regex to dos the app. (also input in our case is when it comes from the network, not from the user using the app)
This is probably not the case, I think you'll be the best judge at that since you'll know the usage better, and seems like the regex is probably ok

ilmotta: Thanks for the feedback, made me think for sure. For this particular case, the input string are URLs that already went through the URL parser, and, for the majority of time, have a reasonably small length. So with that in mind I'll keep the simpler version @cammellos, but I'll extract the regexp compilation to help a little bit.

protocol/linkpreview/linkpreview.go
ilmotta: Done

protocol/linkpreview/linkpreview.go
ilmotta: Done

protocol/linkpreview/linkpreview_test.go
```diff
@@ -276,3 +295,63 @@ func Test_UnfurlURLs_CommonFailures(t *testing.T) {
...
+		{url: "https://placehold.co/600x400@2x.JPG", expected: true},
+		{url: "https://placehold.co/600x400@2x.jpeg", expected: true},
+		{url: "https://placehold.co/600x400@2x.Jpeg", expected: true},
+		{url: "https://placehold.co/600x400@2x.webp", expected: true},
+		{url: "https://placehold.co/600x400@2x.WebP", expected: true},
+		{url: "https://placehold.co/600x400@2x.PnGs", expected: false},
+		{url: "https://placehold.co/600x400@2x.tiff", expected: false},
+	}
+
+	for _, e := range examples {
```
igor-sirotin, line 316: Minor thing, but it's a great case to use "go test cases" feature.
`MailserverSuite TestMailServer`as an example.

https://github.com/status-im/status-go/blob/6c92025099f410223c9548374af175e9c57d2898/transactions/transactor_test.go#L143-L212

ilmotta:
For reviewers: I couldn't find a way to satisfy the linter when using the repo's nix shell, and because that's all I can use in my system, the PR is currently blocked because I can't run a *good* `make generate`.

I tried to go back to all other revisions that changed the nixpkgs revision in `shell.nix`, but all of them fail with the same errors. These are the steps to reproduce:

```bash
git checkout <any rev in status-go>
nix-shell --pure
make generate
make lint
# => lots of errors about copyLocks
```

I talked with @cammellos about this problem, and hopefully he will be able to solve the copyLocks errors 🙏🏼 Thanks a lot for that!

~It seems that if the CI was integrating code into `develop` using nix, this sort of problem would never happen.~ This is wrong, @jakubgs corrected me, and the CI is actually using Nix.

cc'ing @jakubgs as an FYI

protocol/common/message.go
ilmotta: Done

cammellos (approved):

protocol/linkpreview/linkpreview.go
```diff
@@ -241,13 +247,104 @@ func (u OpenGraphUnfurler) unfurl() (common.LinkPreview, error) {
...
+// isSupportedImage returns true when payload is one of the supported image
+// types. In the future, we should differentiate between animated and
+// non-animated WebP because, currently, only static WebP can be processed by
+// functions in the status-go/images package.
+func isSupportedImage(payload []byte) bool {
+	return images.IsJpeg(payload) || images.IsPng(payload) || images.IsWebp(payload)
+}
+
 func newUnfurler(logger *zap.Logger, httpClient http.Client, url *neturl.URL) Unfurler {
+	if isSupportedImageURL(url) {
```
igor-sirotin, line 340: Hmmm. This way we won't be able to unfurl image links that don't end with required extension, e.g. https://placekitten.com/100/200. And Discord supports such links.

I think it would be great to decide if it's an image from HTTP header `Content-Type: image/jpeg`. Any chance it's possible to check the header without fetching the whole image?

igor-sirotin (commented):
Great job, thank you! I will take this into desktop tomorrow 👌 
Just one concert about no-extension image links.

protocol/linkpreview/linkpreview.go
ilmotta: I started coding with this idea in mind @igor-sirotin, but in the end I decided against it in this PR for simplicity sake. To cleanly implement this logic we would need to change the logic that decides which unfurler to create, because the HTTP request would always need to be performed first. And because the request would have been made already, the unfurling logic of all unfurlers would change as well because the response body would already be available.

This should all work fine, unless I face issues related to the user agent. For example, requests to unfurl OpenGraph URLs need to use specific headers, otherwise some websites stubbornly fail. Hopefully the same headers can be used for image URLs.

Since you commented about this limitation, I now have another reason to implement this. I can do it on Monday (I'll be off this Friday).

ilmotta: > Any chance it's possible to check the header without fetching the whole image?

If we knew all servers understood HEAD requests, we'd be able to get responses without a body. So I think the best we can do is always request upfront and use that whole information to decide which unfurler to instantiate and pass the response to other functions.

It should work fine because we always fallback to OpenGraph to unfurl, so always requesting upfront won't incur in additional requests, since we would request something anyway.

protocol/linkpreview/linkpreview_test.go
ilmotta: It is indeed! The inconvenience with this approach, at least in my tooling, is that I can't run the tests individually, only the whole suite.

For example, I can't run `TestGasValues` by itself. This becomes a huge pain for me in files where the suite is bigger. I'm pretty sure it's a limitation of my tool, and I should probably take a look at improving this.

Thanks for the feedback, I'll see if I can make my Emacs run the test under the cursor when the test is part of a suite.

ilmotta:
> I will take this into desktop tomorrow

@igor-sirotin, I can't push a clean PR due to code generation issues. If I push the results of a `make generate`, the PR becomes a monstrosity of 15k lines and the linter fails. See https://github.com/status-im/status-go/pull/3901#issuecomment-1682974663

So I think this affects the desktop integration because you won't be able to build the library pointing to this branch. This is currently preventing me from opening a good PR in status-mobile too. Let's see if tomorrow the problem can be fixed, otherwise I'll need to ask somebody else to run `make generate` in their machine and push to this branch 😢

protocol/linkpreview/linkpreview_test.go
igor-sirotin: Oh I see. I wouldn't say it's very important to fix 🙂 

I can only confirm that VSCode supports executing a single test (like `TestGasValues`).
And JetBrains GoLand even allows to run a single test case 👍

igor-sirotin: hmmm, I just noted that you're referring to "can't run a single test of suite". But it's still possible to implement test cases even without a suite. `TestTopicHistorySameRange` is an example. 

https://github.com/status-im/status-go/blob/6c92025099f410223c9548374af175e9c57d2898/db/history_test.go#L29-L71

It's still not very important to fix, just sharing my knowledge 🙂

igor-sirotin:
> I can't push a clean PR due to code generation issues
> ... otherwise I'll need to ask somebody else to run make generate in their machine and push to this branch 😢

@ilmotta, `make generate` works fine for me, so I will push the updates to this branch, not to block this PR. We kinda have a milestone today 😄

protocol/linkpreview/linkpreview.go
igor-sirotin: No prob! 
I think it makes sense to skip it for now then and add this later. It will be a status-go update with no breaking changes for desktop/mobile.

igor-sirotin (approved):

igor-sirotin:
Tested in desktop, works like a charm 👍 Thank you, @ilmotta  
https://github.com/status-im/status-desktop/pull/11940

protocol/linkpreview/linkpreview.go
ilmotta: Perfect!

ilmotta:
> Tested in desktop, works like a charm 👍 Thank you, @ilmotta [status-im/status-desktop#11940](https://github.com/status-im/status-desktop/pull/11940)

Sweet! 💯

protocol/linkpreview/linkpreview_test.go
ilmotta: Yes @igor-sirotin, using a manual approach like the one you described is quite useful! Such a technique is used in some of the tests related to link previews too.

But I still think I can make Emacs understand which test of a suite it should run. Thanks again for the feedback!

igor-sirotin:
Took an opportunity to rebase on latest develop to fix conflicts.

PR #3902, opened 2023-08-16
Fix pending join requests + API to get them

jrainville:
Needed for https://github.com/status-im/status-desktop/issues/11851

Fixes a couple of issues with pending community join requests.
I'll annotate inline to explain.

Also adds a API to get the pending approved and dismissed requests.

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -2110,7 +2110,7 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, request 
 		return nil, err
 	}
 
-	if existingRequestToJoin != nil {
+	if existingRequestToJoin != nil && existingRequestToJoin.State != RequestToJoinStateCanceled {
 		// request to join was already processed by an admin and waits to get
 		// confirmation for its decision
 		//
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -1553,14 +1553,13 @@ func (m *Manager) HandleAddCommunityTokenPrivilegedUserSyncMessage(message *prot
 }
 
 func (m *Manager) handleAdditionalAdminChanges(community *Community) (*CommunityResponse, error) {
+	communityResponse := CommunityResponse{
+		RequestsToJoin: make([]*RequestToJoin, 0),
+	}
 
 	if !(community.IsControlNode() || community.HasPermissionToSendCommunityEvents()) {
 		// we're a normal user/member node, so there's nothing for us to do here
-		return nil, nil
-	}
-
-	communityResponse := CommunityResponse{
-		RequestsToJoin: make([]*RequestToJoin, 0),
+		return &communityResponse, nil
 	}
 
 	for i := range community.config.EventsData.Events {
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -308,8 +308,9 @@ type Subscription struct {
 }
 
 type CommunityResponse struct {
-	Community *Community        `json:"community"`
-	Changes   *CommunityChanges `json:"changes"`
+	Community      *Community        `json:"community"`
+	Changes        *CommunityChanges `json:"changes"`
+	RequestsToJoin []*RequestToJoin  `json:"requestsToJoin"`
 }
 
 type CommunityEventsMessageInvalidClockSignal struct {
@@ -1422,7 +1423,7 @@ func (m *Manager) HandleCommunityEventsMessage(signer *ecdsa.PublicKey, message 
 		return nil, err
 	}
 
-	err = m.handleAdditionalAdminChanges(changes.Community)
+	additionalCommunityResponse, err := m.handleAdditionalAdminChanges(changes.Community)
 	if err != nil {
 		return nil, err
 	}
@@ -1449,8 +1450,9 @@ func (m *Manager) HandleCommunityEventsMessage(signer *ecdsa.PublicKey, message 
 	}
 
 	return &CommunityResponse{
-		Community: changes.Community,
-		Changes:   changes,
+		Community:      changes.Community,
+		Changes:        changes,
+		RequestsToJoin: additionalCommunityResponse.RequestsToJoin,
 	}, nil
 }
 
@@ -1550,32 +1552,42 @@ func (m *Manager) HandleAddCommunityTokenPrivilegedUserSyncMessage(message *prot
 	return nil
 }
 
-func (m *Manager) handleAdditionalAdminChanges(community *Community) error {
+func (m *Manager) handleAdditionalAdminChanges(community *Community) (*CommunityResponse, error) {
 
 	if !(community.IsControlNode() || community.HasPermissionToSendCommunityEvents()) {
 		// we're a normal user/member node, so there's nothing for us to do here
-		return nil
+		return nil, nil
+	}
+
+	communityResponse := CommunityResponse{
+		RequestsToJoin: make([]*RequestToJoin, 0),
 	}
 
 	for i := range community.config.EventsData.Events {
 		communityEvent := &community.config.EventsData.Events[i]
 		switch communityEvent.Type {
 		case protobuf.CommunityEvent_COMMUNITY_REQUEST_TO_JOIN_ACCEPT:
-			err := m.handleCommunityEventRequestAccepted(community, communityEvent)
+			requestsToJoin, err := m.handleCommunityEventRequestAccepted(community, communityEvent)
 			if err != nil {
-				return err
+				return nil, err
+			}
+			if requestsToJoin != nil {
+				communityResponse.RequestsToJoin = append(communityResponse.RequestsToJoin, requestsToJoin...)
 			}
 
 		case protobuf.CommunityEvent_COMMUNITY_REQUEST_TO_JOIN_REJECT:
-			err := m.handleCommunityEventRequestRejected(community, communityEvent)
+			requestsToJoin, err := m.handleCommunityEventRequestRejected(community, communityEvent)
 			if err != nil {
-				return err
+				return nil, err
+			}
+			if requestsToJoin != nil {
+				communityResponse.RequestsToJoin = append(communityResponse.RequestsToJoin, requestsToJoin...)
 			}
 
 		default:
 		}
 	}
-	return nil
+	return &communityResponse, nil
 }
 
 func (m *Manager) saveOrUpdateRequestToJoin(signer string, communityID types.HexBytes, requestToJoin *RequestToJoin) (bool, error) {
@@ -1609,8 +1621,10 @@ func (m *Manager) saveOrUpdateRequestToJoin(signer string, communityID types.Hex
 	return updated, nil
 }
 
-func (m *Manager) handleCommunityEventRequestAccepted(community *Community, communityEvent *CommunityEvent) error {
+func (m *Manager) handleCommunityEventRequestAccepted(community *Community, communityEvent *CommunityEvent) ([]*RequestToJoin, error) {
 	acceptedRequestsToJoin := make([]types.HexBytes, 0)
+
+	requestsToJoin := make([]*RequestToJoin, 0)
 
 	for signer, request := range communityEvent.AcceptedRequestsToJoin {
 		requestToJoin := &RequestToJoin{
@@ -1624,7 +1638,7 @@ func (m *Manager) handleCommunityEventRequestAccepted(community *Community, comm
 
 		existingRequestToJoin, err := m.persistence.GetRequestToJoin(requestToJoin.ID)
 		if err != nil && err != sql.ErrNoRows {
-			return err
+			return nil, err
 		}
 
 		if community.IsControlNode() {
@@ -1643,7 +1657,7 @@ func (m *Manager) handleCommunityEventRequestAccepted(community *Community, comm
 
 		requestUpdated, err := m.saveOrUpdateRequestToJoin(signer, community.ID(), requestToJoin)
 		if err != nil {
-			return err
+			return nil, err
 		}
 
 		if community.IsControlNode() && requestUpdated {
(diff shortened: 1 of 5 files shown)


protocol/communities/manager.go
```diff
@@ -1478,107 +1480,90 @@ func (m *Manager) HandleCommunityEventsMessageRejected(signer *ecdsa.PublicKey,
 	return reapplyEventsMessage, nil
 }
 
-func (m *Manager) HandleCommunityPrivilegedUserSyncMessage(signer *ecdsa.PublicKey, message *protobuf.CommunityPrivilegedUserSyncMessage) error {
```
jrainville, line 1505: I split this function into two functions, one for requests to join and one for the tokens. The reason is that when we receive those messages from other privileges users, we didn't update the client at all. Now, `HandleRequestToJoinPrivilegedUserSyncMessage` returns the requestsToJoin so that we can add them to the Response.

protocol/communities/manager.go
```diff
@@ -1478,107 +1480,90 @@ func (m *Manager) HandleCommunityEventsMessageRejected(signer *ecdsa.PublicKey,
...
+	communityResponse := CommunityResponse{
+		RequestsToJoin: make([]*RequestToJoin, 0),
 	}
 
 	for i := range community.config.EventsData.Events {
 		communityEvent := &community.config.EventsData.Events[i]
 		switch communityEvent.Type {
 		case protobuf.CommunityEvent_COMMUNITY_REQUEST_TO_JOIN_ACCEPT:
-			err := m.handleCommunityEventRequestAccepted(community, communityEvent)
+			requestsToJoin, err := m.handleCommunityEventRequestAccepted(community, communityEvent)
```
jrainville, line 1569: Similarly here, when handling additional admin changes, I return the requests to join so that I can update the activity center in the messenger.

protocol/communities/persistence.go
```diff
@@ -374,7 +374,7 @@ func (p *Persistence) SaveRequestToJoin(request *RequestToJoin) (err error) {
 		return ErrOldRequestToJoin
 	}
 
-	_, err = tx.Exec(`INSERT INTO communities_requests_to_join(id,public_key,clock,ens_name,chat_id,community_id,state) VALUES (?, ?, ?, ?, ?, ?, ?)`, request.ID, request.PublicKey, request.Clock, request.ENSName, request.ChatID, request.CommunityID, request.State)
+	_, err = tx.Exec(`INSERT OR REPLACE INTO communities_requests_to_join(id,public_key,clock,ens_name,chat_id,community_id,state) VALUES (?, ?, ?, ?, ?, ?, ?)`, request.ID, request.PublicKey, request.Clock, request.ENSName, request.ChatID, request.CommunityID, request.State)
```
jrainville, line 377: This is needed so that we don't have multiple times the same request.

protocol/messenger_communities.go
```diff
@@ -1308,6 +1308,19 @@ func (m *Messenger) CancelRequestToJoinCommunity(request *requests.CancelRequest
 		return nil, err
 	}
 
+	if !community.AcceptRequestToJoinAutomatically() {
+		// send cancelation to community admins also
```
jrainville, line 1312: We were not sending the cancellation to the admins, so they kept the request to join.

protocol/messenger_communities.go
```diff
@@ -1354,17 +1367,18 @@ func (m *Messenger) AcceptRequestToJoinCommunity(request *requests.AcceptRequest
 		return nil, err
 	}
 
-	pk, err := common.HexToPubkey(requestToJoin.PublicKey)
-	if err != nil {
-		return nil, err
-	}
+	if community.IsControlNode() {
```
jrainville, line 1370: This is needed, because otherwise, the user received the acceptance/declining from admins.

protocol/messenger_communities.go
```diff
@@ -2143,6 +2157,26 @@ func (m *Messenger) PendingRequestsToJoinForCommunity(id types.HexBytes) ([]*com
 	return m.communitiesManager.PendingRequestsToJoinForCommunity(id)
 }
 
+func (m *Messenger) AllPendingRequestsToJoinForCommunity(id types.HexBytes) ([]*communities.RequestToJoin, error) {
```
jrainville, line 2178: New function for the API

protocol/messenger_communities.go
```diff
@@ -2592,6 +2626,41 @@ func (m *Messenger) handleCommunityResponse(state *ReceivedMessageState, communi
 		return err
 	}
 
+	for _, requestToJoin := range communityResponse.RequestsToJoin {
+		// Activity Center notification
```
jrainville, line 2649: The AC handling for request to join updates

protocol/messenger_communities.go
```diff
@@ -2623,7 +2692,55 @@ func (m *Messenger) handleCommunityEventsMessageRejected(state *ReceivedMessageS
 }
 
 func (m *Messenger) handleCommunityPrivilegedUserSyncMessage(state *ReceivedMessageState, signer *ecdsa.PublicKey, message protobuf.CommunityPrivilegedUserSyncMessage) error {
-	return m.communitiesManager.HandleCommunityPrivilegedUserSyncMessage(signer, &message)
```
jrainville, line 2642: Moving a lot of the logic from the manager to the messenger so that I can use the requests to the Response

protocol/messenger_handler.go
```diff
@@ -1376,6 +1376,15 @@ func (m *Messenger) HandleCommunityCancelRequestToJoin(state *ReceivedMessageSta
 
 // HandleCommunityRequestToJoin handles an community request to join
 func (m *Messenger) HandleCommunityRequestToJoin(state *ReceivedMessageState, signer *ecdsa.PublicKey, requestToJoinProto protobuf.CommunityRequestToJoin) error {
+	if state.Response.RequestsToJoinCommunity != nil {
+		for _, request := range state.Response.RequestsToJoinCommunity {
+			if request.PublicKey == common.PubkeyToHex(signer) {
+				// We have already handled this request to join
```
jrainville: Users send the request to all admins and owners + the control node, so the owner might receive the same request twice

protocol/messenger_communities.go
```diff
@@ -2623,7 +2692,55 @@ func (m *Messenger) handleCommunityEventsMessageRejected(state *ReceivedMessageS
...
+	err = m.communitiesManager.ValidateCommunityPrivilegedUserSyncMessage(&message)
+	if err != nil {
+		return err
+	}
+
+	switch message.Type {
+	case protobuf.CommunityPrivilegedUserSyncMessage_CONTROL_NODE_ACCEPT_REQUEST_TO_JOIN:
+		fallthrough
+	case protobuf.CommunityPrivilegedUserSyncMessage_CONTROL_NODE_REJECT_REQUEST_TO_JOIN:
+		if !common.IsPubKeyEqual(community.PublicKey(), signer) {
```
mprakhov: !isControlNodeMsg

protocol/communities/manager.go
```diff
@@ -308,8 +308,9 @@ type Subscription struct {
 }
 
 type CommunityResponse struct {
-	Community *Community        `json:"community"`
-	Changes   *CommunityChanges `json:"changes"`
+	Community      *Community        `json:"community"`
+	Changes        *CommunityChanges `json:"changes"`
+	RequestsToJoin []*RequestToJoin  `json:"requestsToJoin"`
```
mprakhov, line 313: sorry, but why do we need it? Can't we use members added/removed from the changes?

jrainville: I guess I didn't think about that, but it doesn't feel right since the member isn't added yet, since it's an admin event.

Also, the ActivityCenter function needs a request ID, so it would be a bit of a mess to reconvert to a request in the messenger IMO

protocol/messenger_communities.go
```diff
@@ -1308,6 +1308,19 @@ func (m *Messenger) CancelRequestToJoinCommunity(request *requests.CancelRequest
 		return nil, err
 	}
 
+	if !community.AcceptRequestToJoinAutomatically() {
+		// send cancelation to community admins also
+		rawMessage.Payload = payload
+
+		privilegedMembers := community.GetPrivilegedMembers()
+		for _, privilegedMember := range privilegedMembers {
+			_, err := m.sender.SendPrivate(context.Background(), privilegedMember, &rawMessage)
```
mprakhov, line 1317: check, that you are not sending to yourself

```
if privilegedMember.Equal(&m.identity.PublicKey) {
				continue
}
```

protocol/messenger_handler.go
mprakhov: if it will be received twice - it will be just replaced, correct? (after fix in DB querry)
Also maybe we can filter it out by the receiver. As I understand, in the case of the control node it will be a community

protocol/messenger_communities.go
```diff
@@ -2623,7 +2692,55 @@ func (m *Messenger) handleCommunityEventsMessageRejected(state *ReceivedMessageS
...
+		fallthrough
+	case protobuf.CommunityPrivilegedUserSyncMessage_CONTROL_NODE_REJECT_REQUEST_TO_JOIN:
+		if !common.IsPubKeyEqual(community.PublicKey(), signer) {
+			return errors.New("accepted/requested to join sync messages can be send only by the control node")
+		}
+		requestsToJoin, err := m.communitiesManager.HandleRequestToJoinPrivilegedUserSyncMessage(&message, community.ID())
+		if err != nil {
+			return nil
+		}
+		state.Response.AddRequestsToJoinCommunity(requestsToJoin)
```
mprakhov, line 2752: What will this do? Will it send a response to the sender?

protocol/communities/manager.go
```diff
@@ -1425,8 +1426,9 @@ func (m *Manager) HandleCommunityEventsMessage(signer *ecdsa.PublicKey, message
 	}
 
 	return &CommunityResponse{
-		Community: changes.Community,
-		Changes:   changes,
+		Community:      changes.Community,
+		Changes:        changes,
+		RequestsToJoin: additionalCommunityResponse.RequestsToJoin,
```
osmaczko, line 1455: That looks strange. One would wonder why not to merge two community responses together than? It would be cleaner to make `m.handleAdditionalAdminChanges` return `[]*RequestToJoin` instead of `*CommunityResponse`.

protocol/communities/persistence.go
osmaczko: Isn't it already solved by: `PRIMARY KEY (id) ON CONFLICT REPLACE`?

protocol/messenger_communities.go
```diff
@@ -1308,6 +1308,19 @@ func (m *Messenger) CancelRequestToJoinCommunity(request *requests.CancelRequest
 		return nil, err
 	}
 
+	if !community.AcceptRequestToJoinAutomatically() {
```
osmaczko, line 1311: Why do we send cancellation to admins only under this condition?

protocol/communities/manager.go
jrainville: Yeah I thought of that too.

I didn't find a function to merge the community responses, but yeah it would be better.

I could also just return `[]*RequestToJoin` instead of `*CommunityResponse` like you said, but I didn't know if it might change in the future. `handleAdditionalAdminChanges` right now only handles request to join, but maybe later it would handle some more? I guess in that case, we can just add a new return type.

protocol/communities/persistence.go
jrainville: Yeah I guess so. I found that we had the requests duplicated, but maybe that was just a UI thing. I can try reverting and see if I can reproduce.

protocol/messenger_communities.go
jrainville: Because if we `AcceptRequestToJoinAutomatically`, that means that admins won't ever see that request to join. It will be handled automatically by the Owner. I think we literally don't send the requests to the Admins in that case as well.
It saves some network bandwidth

protocol/messenger_communities.go
jrainville: No, it sends a signal to the client. Then the client can update its requests to join to the right state.

protocol/messenger_handler.go
jrainville: Yeah, the DB only saved once. I did this so we don't add the request to join twice in the Response (signal to the client), because the client (desktop) is a bit dumb right now and just adds the request blindly. 

I guess you're right, we should just add some logic in the client to not have duplicate requests.

protocol/messenger_communities.go
osmaczko: Makes sense. Thanks

protocol/communities/manager.go
osmaczko: Right, I believe I'll need to propagate also handled permissions and `CommunityResponse` would be a good fit to store that info. Having a Merge method for `CommunityResponse`  would simplify. Here is an example of Merge for `MessengerResponse`: https://github.com/status-im/status-go/blob/3a41a2550aa28fce8c53746f9bf9168edf13beba/protocol/messenger_response.go#L294

PS. I'll implement it in the scope of my PR.

protocol/messenger_communities.go
jrainville: I can add it, but I don't think there is a scenario where you can send it to yourself, since if you cancel a request to join, that means you're not part of the community, hence not a privileged user.

protocol/communities/persistence.go
osmaczko:
```
CREATE TABLE communities_requests_to_join  (
  id BLOB NOT NULL,
  public_key VARCHAR NOT NULL,
  clock INT NOT NULL,
  ens_name VARCHAR NOT NULL DEFAULT "",
  chat_id VARCHAR NOT NULL DEFAULT "",
  community_id BLOB NOT NULL,
  state INT NOT NULL DEFAULT 0,
  PRIMARY KEY (id) ON CONFLICT REPLACE <-----
);
```

Yes, I don't think we need that.

osmaczko (approved):

protocol/communities/persistence.go
jrainville: Yeah. 
I think I still prefer the query with the `INSERT OR REPLACE` though, because it makes it explicit that we will replace the old one if it exists, without having to check the `CREATE TABLE ` statement.

PR #3903, opened 2023-08-17
feat: enable passing ring width from mobile

yqrashawn:
-------

mobile issue https://github.com/status-im/status-mobile/issues/15037
mobile pr https://github.com/status-im/status-mobile/pull/17034

ident ring width can be 2 or 4 in design system
this PR enables reading ring width from mobile via image server handler

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -72,6 +72,7 @@ type ImageParams struct {
 	UppercaseRatio  float64
 	Theme           ring.Theme
 	Ring            bool
+	RingWidth       float64
 	IndicatorSize   float64
 	IndicatorBorder float64
 	IndicatorColor  color.Color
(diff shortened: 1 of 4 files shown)


server/handlers.go
```diff
@@ -278,6 +289,12 @@ func handleAccountImagesImpl(multiaccountsDB *multiaccounts.Database, logger *za
 		logger.Error("handleAccountImagesImpl: failed to load image.", zap.String("keyUid", parsed.KeyUID), zap.String("imageName", parsed.ImageName), zap.Error(err))
 		return
 	}
+
+	if parsed.Ring && parsed.RingWidth == 0 {
+		logger.Error("handleAccountImagesImpl: no ringWidth.")
+		return
```
qfrank, line 293: this return makes me feel following fallback useless. if `RingWidth` is required, perhaps we can validate it within function `DrawRing` and return the error.
```
if param.RingWidth == 0 {
    ringPxSize = math.Max(2.0, float64(param.Width/16.0))
}
```

qfrank (approved):
LGTM, left minor comment ❤️

server/handlers.go
yqrashawn: this is for other usage of `DrawRing`, like this one https://github.com/status-im/status-go/blob/f7530f3cd0224ae9b13728741938980446505337/server/handlers.go#L724
although I don't know if this handler is being used anywhere

qfrank: proberbly we don't use it anymore , we can remove relate code to `status-go` safely, we can check the git history for file `src/utils/image_server.cljs` , relate commit id: `ce322c9537dff1905d60ce70adc9fd378cf885eb` , relate [PR](https://github.com/status-im/status-mobile/pull/15815)

yqrashawn: you are right, i checked mobile and desktop code. `handleIdenticon` is not used. I added a commit to remove it.

alwx (approved):

PR #3905, opened 2023-08-17
Set compressed key as alias on contact

cammellos:
3 random words names are not to be used anymore, the fallback will be the compressed pubkey, this commit adds the compressed pubkey with ellipsis.

cc @jonathanbarker


Samyoul (approved):
Thank you

jrainville (approved):
LGTM

PR #3907, opened 2023-08-17
feat: Switch to ConnectionString V2

igor-sirotin:
Required for https://github.com/status-im/status-go/issues/3884

This PR switches local pairing to ConnectionString version 2 🥳 
For information I already tested with:
- mac-desktop -> mac-desktop (on a single MacOS device)
- mac-desktop -> win-desktop
- win-desktop -> mac-desktop

Didn't test with mobile yet, not sure how to build status-mobile with a not-yet-merged status-go branch.

We can do more tests when merging `feat/issue-3884-lp-cs-v2` to `develop`. Code review is mostly required now.

### DIFF
--- a/server/pairing/client.go
+++ b/server/pairing/client.go
@@ -6,7 +6,6 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"encoding/pem"
-	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -50,7 +49,7 @@ func NewBaseClient(c *ConnectionParams) (*BaseClient, error) {
 
 		serverCert, err = getServerCert(u)
 		if err != nil {
-			certErrs = errors.Join(certErrs, err)
+			certErrs = fmt.Errorf("%wconnecting to '%s' failed: %w; ", certErrs, u, err)
 			continue
 		}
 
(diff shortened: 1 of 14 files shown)


server/pairing/client.go
```diff
@@ -36,17 +36,37 @@ type BaseClient struct {
...
+
+	for i := range c.netIPs {
+		u, err := c.URL(i)
+		if err != nil {
+			return nil, err
+		}
+
+		serverCert, err = getServerCert(u)
+		if err != nil {
+			certErrs = fmt.Errorf("%wconnecting to '%s' failed: %w; ", certErrs, u, err)
```
Samyoul, line 52: Nice use of wrapped errors.

server/pairing/client.go
```diff
@@ -36,17 +36,37 @@ type BaseClient struct {
 
 // NewBaseClient returns a fully qualified BaseClient from the given ConnectionParams
 func NewBaseClient(c *ConnectionParams) (*BaseClient, error) {
-	u, err := c.URL()
-	if err != nil {
-		return nil, err
+
+	var url *url.URL
```
Samyoul: This variable shadows the package name `net/url`, meaning calls to the `net/url` package will be ambiguous after this line. That is why the old variable was named `u`, as this name wouldn't shadow / collide with the the `net/url`

The variable name doesn't need to be `u`, something else is totally fine.

server/pairing/client.go
igor-sirotin: At first I used `errors.Join`, but Go version is older than `1.21` in CI 🙁

server/pairing/client.go
igor-sirotin: indeed. Nice catch, will fix 👌

server/pairing/client.go
Samyoul: Yeah, I much prefer `errors.Join` but it is too new

Samyoul (approved):
Lovely. Thank you @igor-sirotin , really nice

PR #3904, opened 2023-08-17
feat: ConnectionParams v2 serialize/parse functions

igor-sirotin:
NOTE: This PR has `feat/issue-3884-lp-cs-v2` branch destination. 
I will gather all changes there and then make a single PR from branch to `develop`, with testing.

Part of https://github.com/status-im/status-go/issues/3884
__________

- Added new connection string version
- Added serialize/parse functions

This PR doesn't yet make the use of new connection parameters.

PR #3909, opened 2023-08-17
Local Pairing: connection string version 2 with sharing multiple server addresses

### LINKED ISSUE #3825: Local Pairing Doesn't Work With Windows Desktop
- [Summary](#summary)
- [OS and app version](#os-and-app-version)
- [Screenshots](#screenshots)
- [Device network data](#device-network-data)

## Summary

A Windows Desktop app will not pair with another device. Windows device targets the default gateway IP of `192.168.56.1` as its preferred outbound IP, other OSes return the device's network addressable IP.

## OS and app version

```
Desktop version 0.13.1 / OS windows 10
Mobile version 1.24.0 / Pixel 7a 
```

## Screenshots

<table>
<tr>
<td></td>
<td></td>
</tr>
</table>

## Device network data

```
The computer IPv4 address is: 192.168.245.121

Where my mobile IP address is: 192.168.245.101
```

<table>
<tr>
<th>Desktop</th>
<th>Mobile</th>
</tr>
<tr>
<td><pre>
Link speed (Receive/Transmit):    1000/1000 (Mbps)
Link-local IPv6 address:    fe80::e3d5:cfcd:46a4:3ee8%3
IPv4 address:    192.168.245.121
IPv4 DNS servers:    192.168.245.1
Primary DNS suffix:    lan
DNS suffix search list:    lan
Manufacturer:    Intel
Description:    Intel(R) Ethernet Connection (2) I219-V
Driver version:    12.13.17.7
Physical address (MAC):    30-5A-3A-5A-22-A8
</pre>
<br/>
<pre>C:\Users\********>ipconfig

Windows IP Configuration

Ethernet adapter Ethernet:

   Connection-specific DNS Suffix  . : lan
   Link-local IPv6 Address . . . . . : fe80::e3d5:cfcd:46a4:3ee8%3
   IPv4 Address. . . . . . . . . . . : 192.168.245.121
   Subnet Mask . . . . . . . . . . . : 255.255.255.0
   Default Gateway . . . . . . . . . : 192.168.245.1

Ethernet adapter Ethernet 2:

   Connection-specific DNS Suffix  . :
   Link-local IPv6 Address . . . . . : fe80::4016:62be:48c7:fc5c%18
   IPv4 Address. . . . . . . . . . . : 192.168.56.1
   Subnet Mask . . . . . . . . . . . : 255.255.255.0
   Default Gateway . . . . . . . . . : 
</pre>
</td>
<td></td>
</tr>
</table>

### LINKED ISSUE #3884: Local Pairing - Network Address / Peer Discovery

igor-sirotin:
Closes https://github.com/status-im/status-go/issues/3825
Closes https://github.com/status-im/status-go/issues/3884

Code has already partially reviewed in previous PRs for this branch:
- https://github.com/status-im/status-go/pull/3898
- https://github.com/status-im/status-go/pull/3904
- https://github.com/status-im/status-go/pull/3907

_____

This PR updates Local Pairing process
- server uses all available private IP addresses instead of a single IP
- to make this work, we introduce ConnectionString v2. It's backwards compatible with v1


server/ips.go
```diff
@@ -20,3 +24,147 @@ func GetOutboundIP() (net.IP, error) {
...
+			ips = append(ips, niIps)
+		}
+	}
+
+	return ips, nil
+}
+
+// GetLocalAddressesForPairingServer is a high-level func
+// that returns a list of addresses to be used by local pairing server.
+func GetLocalAddressesForPairingServer() ([]net.IP, error) {
```
qfrank, line 114: It seems we can rename some functions for clarity:

GetLocalAddresses -> getLocalAddresses
FilterAddressesForPairingServer -> filterAddressesForPairingServer
Using lowercase function names indicates which functions are intended for external usage, while uppercase names imply internal-only functions. This naming convention would help differentiate between public and private methods at a glance.

qfrank (approved):
Nice work!

igor-sirotin:
Unfortunately, older versions won't see a `Not supported sync code version`. Instead, there's a general `This doesn't look like a sync code` message.

To fix this (and fix it right) we would need to implement error codes. Would be good to do that, but doesn't seem very important to implement here and now.

image

server/ips.go
igor-sirotin: Thanks for this! I was actually planning to do this, but left this for the end of implementation. And then forgot to update the names 😞

igor-sirotin:
I was able to conduct such a test:

| Device X | Device Y |
|-|-|
| Network A 🔵  `192.168.1.33` | Network A 🔵  `192.168.1.36` |
| Network B 🟠   `192.168.2.42` |  Network C 🔴 `192.168.2.53` | 
 
Networks B 🟠  and C 🔴  are local virtual networks, though for deices X and Y they would look the same.
And there was no problem for the devices to pair 👌

Samyoul (approved):
Really nice! This PR covers a number of failure cases.

Samyoul:
Just a follow up on things I've tested this PR with.

- ✅ Windows device with multiple private networks:
- ✅ Windows device with multiple valid networks
  - eg a device with multiple networks shared with both devices
- ✅ Windows device with multiple false positive networks
  - eg valid private network address that exists on multiple networks, but neither addresses are on a share network of both devices.
- ✅ Windows device with both multiple valid networks and false positive networks
- ✅ Windows device with only wifi network
- ✅ Windows device with only ethernet connection LAN
- ✅ Windows device with both wifi and ethernet connection LAN

Sample outputs:

```go
 netIPs: ([]net.IP) (len=7 cap=8) {
  (net.IP) (len=16 cap=16) 172.26.240.1,
  (net.IP) (len=16 cap=16) 172.28.192.1,
  (net.IP) (len=16 cap=16) 192.168.56.1,
  (net.IP) (len=16 cap=16) 192.168.33.1,
  (net.IP) (len=16 cap=16) 192.168.65.1,
  (net.IP) (len=16 cap=16) 192.168.159.1,
  (net.IP) (len=16 cap=16) 192.168.0.19
 },
```

```
cs3:2YAYR1henvaYDaYKVtXiknsZsb3wFg51MFeR7z2ZM:Jg9:z6VQ8GezNA1afgmR14po8AeFoHsYdtjMqANPKS14BdEJ:5gvvWVPu2MAVMTKjDeLdDAMmE5RgJUQjiRDU77ai1s5y
```

Samyoul:
> Unfortunately, older versions won't see a `Not supported sync code version`. Instead, there's a general `This doesn't look like a sync code` message.
> 
> To fix this (and fix it right) we would need to implement error codes. Would be good to do that, but doesn't seem very important to implement here and now.
> 
> image

@igor-sirotin Hmm, those errors don't seem to come from `status-go`. Is the desktop UI returning custom errors when an error is returned from `status-go`?

But I see the value in error codes, I'll raise an issue for this.

igor-sirotin:
> @igor-sirotin Hmm, those errors don't seem to come from status-go. Is the desktop UI returning custom errors when an error is returned from status-go?

This one is hard-coded in UI. I show it in any case of `validateConnectionString` fails.
It's not really good to show the status-go error message directly in UI (for i18n/l10n reasons). An returned error code would make it possible to show different error messages in UI.

Samyoul:
> This one is hard-coded in UI. I show it in any case of `validateConnectionString` fails. It's not really good to show the status-go error message directly in UI (for i18n/l10n reasons). An returned error code would make it possible to show different error messages in UI.

Yes I understand, I've created the issue https://github.com/status-im/status-go/issues/3933 . So we can do localised messages etc.

PR #3911, opened 2023-08-18
feat: transferring keystore files for selected keypair via local network

saledjenic:
This PR is a replacement for the work done here https://github.com/status-im/status-go/pull/3887, in an improved way.

There is a desktop app feature where we need to transfer keystore files for selected keypair/s only (of course, which are not migrated to a keycard, otherwise we wouldn't need to do that) via a local network using a QR code.

### DIFF
--- a/mobile/status.go
+++ b/mobile/status.go
@@ -1140,6 +1140,36 @@ func InputConnectionStringForBootstrappingAnotherDevice(cs, configJSON string) s
 	return makeJSONResponse(err)
 }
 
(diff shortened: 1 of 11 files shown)


IvanBelyakoff (approved):
LGTM

qfrank (approved):
LGTM! I don't have that [feeling](https://github.com/status-im/status-go/pull/3887#discussion_r1295866376) now 👍

Samyoul (approved):
Thank you very much for implementing this PR. I really like your first instincts on this.

PR #3912, opened 2023-08-18
feat: automatically add owner and token master permissions after minting OwnerToken and TokenMasterToken

mprakhov:
After minting OwnerToken and TokenMasterToken automatically add Owner and TokenMaster community permissions

Added a new API for reevaluating community members' permissions. 

Closes # https://github.com/status-im/status-desktop/issues/11874

status-desktop PR: https://github.com/status-im/status-desktop/pull/11942


protocol/messenger_communities.go
```diff
@@ -4244,6 +4260,38 @@ func (m *Messenger) AddCommunityToken(communityID string, chainID int, address s
...
+
+		contractAddresses := make(map[uint64]string)
+		contractAddresses[uint64(communityToken.ChainID)] = communityToken.Address
+
+		tokenCriteria := &protobuf.TokenCriteria{
+			ContractAddresses: contractAddresses,
+			Type:              protobuf.CommunityTokenType_ERC721,
+			Symbol:            communityToken.Symbol,
+			Name:              communityToken.Name,
+			Amount:            "1.0000000000000000",
```
endulab: These tokens are Erc721 so "1" should be sufficient.

endulab (approved):

osmaczko (commented):
I am wondering why do we need this endpoint at all? Can't we invoke `ReevaluateMembers(community)` after minting is done in status-go?

mprakhov:
> I am wondering why do we need this endpoint at all? Can't we invoke `ReevaluateMembers(community)` after minting is done in status-go?

In order to instantly update user permission during airdropping or destructing airdrop If the airdropped token is used in community permissions

Example - you airdrop TokenMasterToken... 
The status will be received in status-desktop, we will check if this token is used in community permissions and if yes - a check will be requested

osmaczko:
> > I am wondering why do we need this endpoint at all? Can't we invoke `ReevaluateMembers(community)` after minting is done in status-go?
> 
> In order to instantly update user permission during airdropping or destructing airdrop If the airdropped token is used in community permissions
> 
> Example - you airdrop TokenMasterToken... The status will be received in status-desktop, we will check if this token is used in community permissions and if yes - a check will be requested

My question was why do we need to do that on status-desktop side? Isn't airdrop process done in status-go?

Example - you airdrop TokenMasterToken... ~~The status will be received in status-desktop~~ Then as a post-action in status go, we will check if this token is used in community permissions and if yes - a check will be ~~requested~~ done

endulab:
> > > I am wondering why do we need this endpoint at all? Can't we invoke `ReevaluateMembers(community)` after minting is done in status-go?
> > 
> > 
> > In order to instantly update user permission during airdropping or destructing airdrop If the airdropped token is used in community permissions
> > Example - you airdrop TokenMasterToken... The status will be received in status-desktop, we will check if this token is used in community permissions and if yes - a check will be requested
> 
> My question was why do we need to do that on status-desktop side? Isn't airdrop process done in status-go?
> 
> Example - you airdrop TokenMasterToken... ~The status will be received in status-desktop~ Then as a post-action in status go, we will check if this token is used in community permissions and if yes - a check will be ~requested~ done

Airdrop is done in status-go, however we need to wait for airdrop transaction to be minted. The transaction system which we use to listen to transactions has some limitations and we can get the signal in status-desktop. Then depending on the status of transaction we can proceed with other operations.

mprakhov:
@endulab right... I can do it, but only if I will patch the wallet API and do community permissions verifications there after the token will be airdropped/destroyed.
Here's monitoring api
```
func (api *API) WatchTransaction(ctx context.Context, transactionHash common.Hash) error
func (api *API) WatchTransactionByChainID(ctx context.Context, chainID uint64, transactionHash common.Hash) error
```
But I don't think that this is a good idea
So better to do it via status-desktop and a separate API call

osmaczko:
> The transaction system which we use to listen to transactions has some limitations and we can get the signal in status-desktop.

Ok :+1:  this justifies things a bit.

protocol/communities/manager.go
```diff
@@ -821,6 +821,35 @@ func (m *Manager) DeleteCommunityTokenPermission(request *requests.DeleteCommuni
...
+		return nil, ErrOrgNotFound
+	}
+
+	// TODO: Control node needs to be notified to do a permission check if TokenMasters did airdrop
+	// of the token which is using in a community permissions
+	if !community.IsControlNode() {
+		return nil, ErrNotEnoughPermissions
+	}
+
+	err = m.ReevaluateMembers(community)
```
osmaczko, line 840: I believe we could do it async, it would save us from parsing the separate response on clients, because when members are reevaluated changes are signaled to the clients anyway.
```
go func() {
    err = m.ReevaluateMembers(community)
    if err != nil {
        m.logger.Warn("failed to evaluate members")
    }
}()

return nil
```

More that that it would simplify client's code because we will be able to call this API synchronously.

protocol/protobuf/communities.proto
```diff
@@ -71,6 +71,7 @@ message CommunityTokenPermission {
     CAN_VIEW_CHANNEL = 3;
     CAN_VIEW_AND_POST_CHANNEL = 4;
     BECOME_TOKEN_MASTER = 5;
+    BECOME_TOKEN_OWNER = 6;
```
jrainville, line 74: This role has a confusing name. What does it do? As it is worded, it sounds like having this role means you can own any token

mprakhov: This is token owner community permission (CommunityTokenPermission_BECOME_TOKEN_OWNER). We need it for displaying owner token community permission

protocol/communities/manager.go
mprakhov: I was following the pattern we used before for API and I do not remember that we do it async on a backend.
Status-desktop already has it async, also async on status-desktop allows us to show the errors in the console, while doing it here - will require you to do certain manipulations in order to display status logs

mprakhov: that's how I catched API calls limitations to the request

osmaczko: You can easily read logs in `geth.log`. That was just a suggestion which I believe would simplify the workflow a bit. I am okay with current impl.

PR #3871, opened 2023-08-10
ci: generate junit reports for unit tests

yakimant:
We want to make failed tests more visible without reading the logs and also use Jenkins features, like for tracking test history.
We need to have an integration between `go test` and Jenkins through junit xml files. This is an industry standard for tests-ci integration.

This PR adds too instalation and junit reports generation.

PR #3913, opened 2023-08-18
fix: exitcode for unit tests

yakimant:
#3871 broke exitcode of failing `go test`. `pipefail` will to the trick.


jakubgs (approved):
Thanks.

cammellos (approved):

PR #3914, opened 2023-08-18
Always use protobufs by reference

cammellos:
This commit does a few things, maybe too many, but they are somewhat related:

1) Update all the protobufs to be used by reference instead of by value. Protobuf 3.20 introduces a sync.Mutex inside the generated files. This means that it cannot be passed by value anymore otherwise a copy of the lock is made and the linter complains. This commit updates all the structs to pass a reference.

2) v1/statusmessage and messenger.go handlers both used reflection to check the protobuf value. As far as I can see there was no way to keep the same method without converting the protobuf to a value, as `ParsedMessage` would have to be converted to a value in the process. So the code was removed and replaced with a generator of handlers.

The generator of handles is in `generate_handlers.go`, and generates a `dispatchToHandler` method that calls a method based on the protobuf enum in application metadata. For example for:
`ApplicationMetadataMessage_CHAT_MESSAGE` it will generate and call a method:
`m.HandleChatMessageProtobuf(state, message, statusMessage)` which will unmarshal the correct protobuf and expects a method `m.HandleChatMessage(state,message,statusMessage)` to be defined.

This force handling of each protobuf that we create and forces consistency in naming methods/signature. Performance wise it should be faster since it doesn't use any reflection, and the cost of reflection is at generation time. 
It also moves the check from runtime, to compile time, which should be safer.
Also, forces the same name between the `applicationmetadata_enum` and the protobuf name (there were a few inconsistencies).

`make generate` will re-generate the handlers file, so when changing/adding a protobuf, you will have to add the handler otherwise the code won't compile, so it's more explicit than before where you had to modify two places. 

I have kept the code as much as it was before to minimize changes, but further refactoring could be done. 

Things still left to do:
1) plug `make generate` so that when run it re-generates handlers
2) various polish
3) Linting

Some questionable choices:
1) If the method starts with `SYNC_` it will be only allowed if the pk of the sender is coming from us. We could have a registry instead but then it's two places to edit and it still needs to be remember, not sure what's best.


qfrank (commented):
> If the method starts with SYNC_ it will be only allowed if the pk of the sender is coming from us. We could have a registry instead but then it's two places to edit and it still needs to be remember, not sure what's best.

i prefer no registry! and using generator is a great idea 👍

cammellos:
I believe this is ready to be reviewed, most likely just fixing some tests from now on cc @Samyoul @osmaczko @qfrank

osmaczko (approved):
Massive one, nice!

> If the method starts with SYNC_ it will be only allowed if the pk of the sender is coming from us. We could have a registry instead but then it's two places to edit and it still needs to be remember, not sure what's best.

We could also handle it manually inside the not-generated hadlers. I'll opt for SYNC_ prefix as it enforces the conistency in the naming.

protocol/messenger_handlers.go
```diff
@@ -0,0 +1,1607 @@
...
+		return m.handleSyncAccountsPositionsProtobuf(messageState, protoBytes, msg, filter)
+        
+           case protobuf.ApplicationMetadataMessage_COMMUNITY_EVENTS_MESSAGE_REJECTED:
+		return m.handleCommunityEventsMessageRejectedProtobuf(messageState, protoBytes, msg, filter)
+        
+           case protobuf.ApplicationMetadataMessage_COMMUNITY_PRIVILEGED_USER_SYNC_MESSAGE:
+		return m.handleCommunityPrivilegedUserSyncMessageProtobuf(messageState, protoBytes, msg, filter)
+        
+	default:
+		m.logger.Info("protobuf type not found", zap.String("type", string(msg.Type)))
```
ilmotta, line 230: @cammellos, could you expand a bit on why you think it's better to log + return nil instead of generating an error when no match is found?

protocol/messenger_handlers.go
```diff
@@ -0,0 +1,1607 @@
...
+
+	return m.HandleChatMessage(messageState, p, msg)
+	
+}
+
+
+func (m *Messenger) handleContactUpdateProtobuf(messageState *ReceivedMessageState, protoBytes []byte, msg *v1protocol.StatusMessage, filter transport.Filter) error {
+	m.logger.Info("handling ContactUpdate")
+	
+
```
ilmotta, line 258: There are a bunch of extra lines here and there in this file :)

ilmotta (approved):

protocol/messenger_handlers.go
cammellos: oh good point, that's the old code (it was run in a loop instead of a separate function), I think it's safe to return an error now

protocol/messenger_handlers.go
cammellos: This file won't be linted, since it's generated by a template, and it's hard to get both in a readable state (if I remove the extra lines the template becomes a bit unreadable), since I have to inline the for loops, hope that's ok

qfrank (approved):

protocol/messenger_handlers.go
ilmotta: I see now, thanks for the clarification.

_assets/ci/Jenkinsfile
```diff
@@ -18,7 +18,7 @@ pipeline {
 
   options {
     /* Prevent Jenkins jobs from running forever */
-    timeout(time: 30, unit: 'MINUTES')
+    timeout(time: 40, unit: 'MINUTES')
```
Samyoul, line 21: No specific objection to this, but I feel that we should probably think about optimising our tests so that we don't keep increasing the timeout ad infinitum. I think it used to be 5 minutes 😆

Samyoul (commented):
An enjoyable PR. Cool work. I love the use of more `make generate` functionality.

PR #3916, opened 2023-08-18
ci: drop wrongly calling make install-lint

jakubgs:
Linting tools should already be installed by Nix, and we should use them.

Also switching most shells to pure mode to see what happens.
And set Go cache to workspace tmp folder to avoid cache poisoning.


ilmotta (approved):
Nice!

cammellos (approved):

PR #3917, opened 2023-08-18
chore: Moved link preview unfurlers to a separate package

igor-sirotin:
Requires https://github.com/status-im/status-go/pull/3901
Note that it uses `feat/link-previews-for-image-urls` as target branch. I will change the base when https://github.com/status-im/status-go/pull/3901 is merged.

_____

1. Moved `linkpreviews` to `Messenger`
- This is required for unfurling Status SharedURL. That functionality is implemented in `Messenger`
- Because of this, I also reworked tests to a TestSuite, because we need a `Messenger` there

2. Clean up
- Added a separate `unfurlers` "sub"-package
- Moved each unfurler to a separate file

### DIFF
--- a/protocol/linkpreview/linkpreview.go
+++ b/protocol/linkpreview/linkpreview.go
@@ -3,13 +3,14 @@ package linkpreview
 import (
 	"errors"
 	"fmt"
-	"github.com/status-im/markdown"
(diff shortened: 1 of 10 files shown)


ilmotta (approved):
LGTM @igor-sirotin!

PR #3922, opened 2023-08-21
fix workdir path in case it has code

yakimant:
`make test` was failing (see https://github.com/status-im/status-go/pull/3914) in case of code in `github.com/status-im/status-go` (from `go list ./...` output`)

~~This PR adds `./` in front of subfolder.~~
Found a better solution to find a dir of package: `go list -f {{.Dir}} ${package}`
Also fixed variable names.


cammellos (approved):

jakubgs (approved):
Looks better. Especially not using `path` as a variable name =].

Even lowercase it looks scary.

PR #3925, opened 2023-08-22
Feat/pending state token permissions handling

osmaczko:
add pending state for token permissions

iterates: https://github.com/status-im/status-desktop/issues/11852

### DIFF
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -104,26 +104,26 @@ func (o *Community) MarshalPublicAPIJSON() ([]byte, error) {
 		return nil, errors.New("member identity not set")
 	}
 	communityItem := struct {
-		ID                      types.HexBytes                                `json:"id"`
-		Verified                bool                                          `json:"verified"`
-		Chats                   map[string]CommunityChat                      `json:"chats"`
-		Categories              map[string]CommunityCategory                  `json:"categories"`
-		Name                    string                                        `json:"name"`
-		Description             string                                        `json:"description"`
-		IntroMessage            string                                        `json:"introMessage"`
-		OutroMessage            string                                        `json:"outroMessage"`
-		Tags                    []CommunityTag                                `json:"tags"`
-		Images                  map[string]images.IdentityImage               `json:"images"`
-		Color                   string                                        `json:"color"`
-		MembersCount            int                                           `json:"membersCount"`
-		EnsName                 string                                        `json:"ensName"`
-		Link                    string                                        `json:"link"`
-		CommunityAdminSettings  CommunityAdminSettings                        `json:"adminSettings"`
-		Encrypted               bool                                          `json:"encrypted"`
-		BanList                 []string                                      `json:"banList"`
-		TokenPermissions        map[string]*protobuf.CommunityTokenPermission `json:"tokenPermissions"`
-		CommunityTokensMetadata []*protobuf.CommunityTokenMetadata            `json:"communityTokensMetadata"`
-		ActiveMembersCount      uint64                                        `json:"activeMembersCount"`
+		ID                      types.HexBytes                       `json:"id"`
+		Verified                bool                                 `json:"verified"`
+		Chats                   map[string]CommunityChat             `json:"chats"`
+		Categories              map[string]CommunityCategory         `json:"categories"`
+		Name                    string                               `json:"name"`
+		Description             string                               `json:"description"`
+		IntroMessage            string                               `json:"introMessage"`
+		OutroMessage            string                               `json:"outroMessage"`
+		Tags                    []CommunityTag                       `json:"tags"`
+		Images                  map[string]images.IdentityImage      `json:"images"`
+		Color                   string                               `json:"color"`
+		MembersCount            int                                  `json:"membersCount"`
+		EnsName                 string                               `json:"ensName"`
+		Link                    string                               `json:"link"`
+		CommunityAdminSettings  CommunityAdminSettings               `json:"adminSettings"`
+		Encrypted               bool                                 `json:"encrypted"`
+		BanList                 []string                             `json:"banList"`
+		TokenPermissions        map[string]*CommunityTokenPermission `json:"tokenPermissions"`
+		CommunityTokensMetadata []*protobuf.CommunityTokenMetadata   `json:"communityTokensMetadata"`
+		ActiveMembersCount      uint64                               `json:"activeMembersCount"`
 	}{
 		ID:         o.ID(),
 		Verified:   o.config.Verified,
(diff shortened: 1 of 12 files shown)


protocol/communities/community_event.go
```diff
@@ -246,23 +237,19 @@ func (o *Community) updateCommunityDescriptionByCommunityEvent(communityEvent Co
...
 	case protobuf.CommunityEvent_COMMUNITY_MEMBER_TOKEN_PERMISSION_CHANGE:
-		_, exists := o.config.CommunityDescription.TokenPermissions[communityEvent.TokenPermission.Id]
-		if exists {
-			_, err := o.updateTokenPermission(communityEvent.TokenPermission)
-			if err != nil {
-				return err
-			}
-		} else {
-			_, err := o.addTokenPermission(communityEvent.TokenPermission)
+		if o.IsControlNode() {
```
osmaczko, line 240:
```suggestion
		if o.IsPrivilegedMember(o.MemberIdentity() {
```
Other admins should see pending changes.

protocol/communities/community.go
```diff
@@ -1320,20 +1321,64 @@ func (o *Community) Categories() map[string]*protobuf.CommunityCategory {
...
-func (o *Community) TokenPermissions() map[string]*protobuf.CommunityTokenPermission {
-	return o.config.CommunityDescription.TokenPermissions
+func (o *Community) tokenPermissions() map[string]*CommunityTokenPermission {
+	result := make(map[string]*CommunityTokenPermission, len(o.config.CommunityDescription.TokenPermissions))
+	for _, tokenPermission := range o.config.CommunityDescription.TokenPermissions {
+		result[tokenPermission.Id] = NewCommunityTokenPermission(tokenPermission)
+	}
+
+	// Non-privileged members should not see pending permissions
+	if o.config.EventsData == nil || !o.IsPrivilegedMember(o.MemberIdentity()) {
```
osmaczko, line 1326: `!o.IsPrivilegedMember(o.MemberIdentity())` this check shouldn't be needed

TODO: reject token permission events for ordinary members

protocol/communities/manager.go
```diff
@@ -1393,35 +1396,35 @@ func (m *Manager) HandleCommunityEventsMessage(signer *ecdsa.PublicKey, message
...
 	if err != nil {
 		return nil, err
 	}
 
-	// Control node cerifies community events and publish changes
-	// all other nodes only apply changes to the community
-	if changes.Community.IsControlNode() {
-		changes.Community.increaseClock()
-		err = m.persistence.SaveCommunity(changes.Community)
+	// Control node applies events and publish updated CommunityDescription
```
jrainville, line 1462:
```suggestion
	// Control node applies events and publishes updated CommunityDescription
```

protocol/communities/community_token_permission.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"reflect"
+
+	"github.com/status-im/status-go/protocol/protobuf"
+)
+
+type TokenPermissionState uint8
+
+const (
+	TokenPermissionApproved TokenPermissionState = iota
```
jrainville, line 12: Do we need a `TokenPermissionRejected` too? Maybe if an admin tries to create a TokenMaster permission or something? (they'd need to hack the client to do that, but still, we need to reject those)

jrainville (approved):
Nice job

protocol/communities/community_token_permission.go
osmaczko: I don't think so.

> Maybe if an admin tries to create a TokenMaster permission

Then it will not go through the validation process and be silently ignored by everyone (control node, other admins and members).

One scenario where a token permission event can be explicitly rejected is when control node has newer `CommunityDescription` than the one received in events message. We have a generic solution for that already: https://github.com/status-im/status-go/pull/3876

protocol/communities/community_changes.go
```diff
@@ -79,7 +79,35 @@ func (c *CommunityChanges) HasMemberLeft(identity string) bool {
...
+			changes.TokenPermissionsRemoved[id] = originPermission
+		}
+	}
+
+	// Check for added token permissions
+	for id, permission := range modifiedTokenPermissions {
+		if _, ok := originTokenPermissions[id]; !ok {
+			changes.TokenPermissionsAdded[id] = permission
+		}
+	}
```
mprakhov, line 104: why did you remove it from the evaluateCommunityChangesByDescription if this data from CommunityDescription?

protocol/communities/community.go
```diff
@@ -1315,20 +1316,64 @@ func (o *Community) Categories() map[string]*protobuf.CommunityCategory {
...
 
-func (o *Community) TokenPermissions() map[string]*protobuf.CommunityTokenPermission {
-	return o.config.CommunityDescription.TokenPermissions
+func (o *Community) tokenPermissions() map[string]*CommunityTokenPermission {
+	result := make(map[string]*CommunityTokenPermission, len(o.config.CommunityDescription.TokenPermissions))
+	for _, tokenPermission := range o.config.CommunityDescription.TokenPermissions {
+		result[tokenPermission.Id] = NewCommunityTokenPermission(tokenPermission)
+	}
+
+	// Non-privileged members should not see pending permissions
```
mprakhov, line 1325: I think this is wrong and tokenPermissions must return all data it has related to the tokenPermissions and not modify the content to return

If permission in a pending state, it must be ignored by privileged members during CommunityEvent handling, wait until Control node handle this event and publish CommunityDescription with the update

mprakhov (commented):
Ok, it seems I was late :)

protocol/communities/community.go
osmaczko: https://github.com/status-im/status-go/pull/3925#discussion_r1302011133

protocol/communities/community_changes.go
osmaczko: Because permissions are now obtained not only from the community description but also from events. See: https://github.com/status-im/status-go/blob/c3cac93229f25446c6203d4dea40b5d2eae960e9/protocol/communities/community.go#L1319

protocol/communities/community.go
mprakhov: ok, but you already skip it for non-control nodes
https://github.com/status-im/status-go/pull/3925/files#diff-ad79863baafd661e9f1626e2e7ebc0e5db3489ce93c5d5859237446a801b2af2R239-R245

osmaczko: Permissions are now obtained also from events, this part is skipping only `CommunityDescription` update (`updateCommunityDescriptionByCommunityEvent`).

osmaczko: > ok, but you already skip it for non-control nodes https://github.com/status-im/status-go/pull/3925/files#diff-ad79863baafd661e9f1626e2e7ebc0e5db3489ce93c5d5859237446a801b2af2R239-R245

Permissions are now obtained also from events, this part is skipping only `CommunityDescription` update (`updateCommunityDescriptionByCommunityEvent`).

mprakhov: Permissions were obtained from events before, I wrote unit tests for this scenario. All members, who received these permissions, were adding them to their community description. 
From my understanding, your changes should be to ignore from member-side permissions, which are not confirmed by a Control Node.

When you request Community from the database, it should apply all community events on top of the community description. Previously this fn was the same as it was for patching CommunityDescription during handling community events

So you, as a simple member, will not receive patched CommunityDescription with non-approved CommunityPermission. All other admins/token masters must receive it and see it in a pending state

PR #3928, opened 2023-08-22
chore: bumped go-waku with improved peer mgmt

chaitanyaprem:
Bumped go-waku to recent commit which includes enhanced peer-management per https://github.com/waku-org/go-waku/issues/594

It include restricting number of in peer relay connections as well as out peer connections based on max-connections configuration that is set. It also ensures connections to sufficient relay peers are maintained.

We want to validate the new peer management by dogfooding with status-desktop and this PR is a step towards the same.

Important changes:
- [x] Bumped up go-waku to recent commit on master which includes updated peer management
- [x] Updated API changes as per new API's

- Tested locally using status-desktop by connecting to another desktop client. 
- Tested 1-1 chat and creating a community and messaging in the community. 
- Enabled Node management and validated peer count
- Enable light mode and performed 1-1 and community chat tests

@richard-ramos , few points i wanted your input on. 
- I had noticed some other dependencies that have got updated which i am not sure of. Maybe due to some other changes made in go-waku from last update. Please take a look and confirm if these are ok.
- I had also noticed that previously 40 was passed as minPeers count to discovery(which means we could potentially maintain connections with 40 peers), which i have reused as max-connections count(which means each status-desktop will try to have 10 connections for outRelayPeers and 22 for inRelay peers). Have any tests been done before with status-desktop as to how many peer connections are required without hitting any performance/latency issues in message delivery? Because i still think 40 is too many peer connections for a desktop client and max-connections can be set to 20 .


richard-ramos (approved):

richard-ramos:
I dont think we've done such tests before!
I do agree that 40 is excessive! I wonder if some value got overriden in a commit? because I do remember setting up a discovery value to limit discovery so if you had 20something peers (i dont remember the number), then it would not actively try to discover more peers.  The number of peers could keep growing because of no code to limit incoming connections. 

 In wakuV1, i do remember that the connections seemed to stabilize at 23-25.

chaitanyaprem:
> I dont think we've done such tests before! I do agree that 40 is excessive! I wonder if some value got overriden in a commit? because I do remember setting up a discovery value to limit discovery so if you had 20something peers (i dont remember the number), then it would not actively try to discover more peers. The number of peers could keep growing because of no code to limit incoming connections.
> 
> In wakuV1, i do remember that the connections seemed to stabilize at 23-25.

So, I have modified the config to 20 maxConnections which means a max outConnections of 10 and maxIn connections of 6 would be maintained. Which means a total of 16 relay peers can be connected which is well within upper range of gossipSub D limit.

richard-ramos (approved):

vitvly (approved):

chaitanyaprem:
Closing this PR as underlying changes are already brought into status-go.

PR #3931, opened 2023-08-22
feat: share CommunityTokens between privileged users

mprakhov:
Add new CommunityToken to the database of privileged members during:
- request to join the response
- reevaluate community member's roles
- update CommunityDescription with new CommunityToken
- handling community events

Refactored collectibles service in order to make it possible to use it in tests

P.S. PR split on 2 commits, all `make generate` in a separate commit

Closes #[11943](https://github.com/status-im/status-desktop/issues/11943)


jrainville (approved):
LGTM

protocol/communities/manager.go
```diff
@@ -4605,43 +4605,26 @@ func (m *Manager) HandleCommunityTokensMetadata(communityID string, communityTok
...
 						return err
 					}
-					communityToken.Supply = &bigint.BigInt{Int: totalSupply}
-					communityToken.InfiniteSupply = collectibles.GetInfiniteSupply().Cmp(totalSupply) == 0
+
+					communityToken.Supply = contractData.TotalSupply
+					communityToken.InfiniteSupply = contractData.InfiniteSupply
 				}
 
 				err = m.persistence.AddCommunityToken(communityToken)
```
endulab, line 4665: We miss 2 fields here: PrivilegeLevel and Deployer. It will be hard to get deployer since we need to parse blockchain transactions. PrivilegeLevel can be taken from permissions.

services/collectibles/manager.go
```diff
@@ -0,0 +1,99 @@
+package collectibles
```
endulab, line 1: Why do you need Manager layer ? What is the idea ?

mprakhov: In order not to increase the Service interface.
Also, the Manager can be passed to the service directly (if it will be necessary) and it can be mocked

PR #3938, opened 2023-08-23
Fix contract call for communities

cammellos:


protocol/messenger_communities.go
```diff
@@ -507,30 +506,8 @@ func (m *Messenger) CuratedCommunities() (*communities.KnownCommunitiesResponse,
...
-			}
-			ethClient = e
-		}
-	}
-
-	if ethClient == nil {
-		return nil, errors.New("failed to initialize backend before requesting curated communities")
-	}
 
-	directory, err := m.contractMaker.NewDirectoryWithBackend(chainID, ethClient)
```
alaibe, line 535: Can we get rid of NewDirectoryWithBackend? I think it is unused now?

rasom (approved):

protocol/messenger_communities.go
```diff
@@ -507,30 +506,8 @@ func (m *Messenger) CuratedCommunities() (*communities.KnownCommunitiesResponse,
 	// Revert code to https://github.com/status-im/status-go/blob/e6a3f63ec7f2fa691878ed35f921413dc8acfc66/protocol/messenger_communities.go#L211-L226 once the curated communities contract is deployed to mainnet
 
 	chainID := uint64(420) // Optimism Goerli
-	sDB, err := accounts.NewDB(m.database)
-	if err != nil {
-		return nil, err
-	}
-	nodeConfig, err := sDB.GetNodeConfig()
```
alaibe, line 516: Shall we add a todo in this getNodeConfig function to update and use network manager or this unexpected empty rpc url will be come back?

alaibe (approved):

protocol/messenger_communities.go
cammellos: added comment & removed method

PR #3939, opened 2023-08-23
feat(wallet/activity): Added API for tx and multiTx details

Cuteivist:
task #11897
API for tx and multi tx details.

Note that pending TX was not handled, it is because pending tx doesn't contain details in DB, so empty details object is returned on client side.

Details:
    * protocol type (BridgeAT and SwapAT)
    * nonce
    * block number
    * tx hash
    * input data
    * contract address
    * max total fees

### API video

UI code is still WIP, but results are already visible

https://github.com/status-im/status-go/assets/11396062/d51015ff-1ae9-4b35-9212-d6643eb33703

### DIFF
--- a/services/wallet/activity/activity.go
+++ b/services/wallet/activity/activity.go
@@ -48,61 +48,58 @@ const (
 )
 
 type Entry struct {
-	payloadType     PayloadType
-	transaction     *transfer.TransactionIdentity
-	id              transfer.MultiTransactionIDType
-	timestamp       int64
-	activityType    Type
-	activityStatus  Status
-	amountOut       *hexutil.Big // Used for activityType SendAT, SwapAT, BridgeAT
-	amountIn        *hexutil.Big // Used for activityType ReceiveAT, BuyAT, SwapAT, BridgeAT
-	tokenOut        *Token       // Used for activityType SendAT, SwapAT, BridgeAT
-	tokenIn         *Token       // Used for activityType ReceiveAT, BuyAT, SwapAT, BridgeAT
-	sender          *eth.Address
-	recipient       *eth.Address
-	chainIDOut      *common.ChainID
-	chainIDIn       *common.ChainID
-	transferType    *TransferType
-	contractAddress *eth.Address
+	payloadType    PayloadType
+	transaction    *transfer.TransactionIdentity
+	id             transfer.MultiTransactionIDType
+	timestamp      int64
+	activityType   Type
+	activityStatus Status
+	amountOut      *hexutil.Big // Used for activityType SendAT, SwapAT, BridgeAT
+	amountIn       *hexutil.Big // Used for activityType ReceiveAT, BuyAT, SwapAT, BridgeAT
+	tokenOut       *Token       // Used for activityType SendAT, SwapAT, BridgeAT
+	tokenIn        *Token       // Used for activityType ReceiveAT, BuyAT, SwapAT, BridgeAT
+	sender         *eth.Address
+	recipient      *eth.Address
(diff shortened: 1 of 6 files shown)


services/wallet/activity/activity.go
```diff
@@ -960,3 +976,164 @@ func updateKeypairsAccountsTable(accountsDb *accounts.Database, db *sql.DB) erro
...
+		fallthrough
+	case common.HopBridgeTo:
+		*protocolType = ProtocolHop
+	default:
+		return nil
+	}
+	return protocolType
+}
+
+func getMultiTxDetails(ctx context.Context, db *sql.DB, id string) (*EntryDetails, error) {
```
stefandunca: Does it make sense to have details entry in a separate file? `details.go` maybe?

services/wallet/activity/activity.go
```diff
@@ -960,3 +976,164 @@ func updateKeypairsAccountsTable(accountsDb *accounts.Database, db *sql.DB) erro
...
+		var transferHashDB, contractAddressDB sql.RawBytes
+		var blockNumberDB int64
+		var nonceDB uint64
+		tx := &types.Transaction{}
+		nullableTx := sqlite.JSONBlob{Data: tx}
+		err := rows.Scan(&transferHashDB, &blockNumberDB, &contractTypeDB, &nonceDB, &nullableTx, &contractAddressDB)
+		if err != nil {
+			return nil, err
+		}
+		if len(transferHashDB) > 0 {
```
stefandunca: What happens if it is a swap or a bridge with multiple entries?

services/wallet/activity/service.go
```diff
@@ -101,6 +101,14 @@ func (s *Service) FilterActivityAsync(requestID int32, addresses []common.Addres
 	})
 }
 
+func (s *Service) GetMultiTxDetails(ctx context.Context, id string) (*EntryDetails, error) {
```
stefandunca: We should use the original type as much as possible to documenting purpose and type checking. Please change the ID here and in `GetTxDetails`

stefandunca (approved):
Great job. Please see my notes and questions inline.

services/wallet/activity/activity.go
Cuteivist: At this moment I tested a lot of Bridges and Swaps and everyone of them had multiple entries, and each entry had same transfer hash.
From UI perspective there should be second hash, from other network, e.g. Bridge L1 -> L2, then there should be L1 hash and L2 hash. At this moment we only have L1 hash.
This task is to get existing data from DB to UI, in future I'm more than happy to figure out how to get second hash.

stefandunca: How about processing only the first entries then?

Cuteivist: First that wouldn't work for Protocol, where type will be eth for most of transfer and only 1 or 2 will have Uniswap or Hop, so that require iterating.
Using sql won't great too, because I'm not sure that all data we require will be available in single transfer, so ultimately I decided to query all (I think at most I found 6 transfers for one multi transfer), then get first value.

My plan is to further adjust this implementation when adding additional data or if there's some new data we can query.

services/wallet/activity/activity.go
Cuteivist: Moved it to separate file. Makes more sense that way.

services/wallet/activity/service.go
Cuteivist: Great catch! I also added second parameter, so now returned details object can determine whether it's multiTx or Tx.

PR #3942, opened 2023-08-23
Fix/permissions pending state

osmaczko:
- fix: remove fixupChannelMembers
This manual migration causes problems because it updates `CommunityDescription` without publishing it. All new communities will still function properly. We definitely need a better mechanism for `CommunityDescription` migrations.

- fix: expose token permission from events
Ensure token permission from events is returned by `tokenPermissions`. Otherwise, edited permissions waiting for approval will show the original value.


osmaczko:
@jrainville, once @iurimatias migrates to 0.14 RC1 the [fix: remove fixupChannelMembers](https://github.com/status-im/status-go/pull/3942/commits/fa53622a076bea62010859e9a386b47879b9b245) should be applied to the RC branch.

jrainville (approved):

PR #3855, opened 2023-08-04
fix encoding of status.app links

felicio:
> note: this is just a draft for others who worked on go implemtation to take over, feel free to push commits to it if the settings allow otherwise please let me know

- [ ] fix go implementation
- [ ] use same test data as js implementation
- [ ] have tests displaying/asserting full urls (snapshots) too

---

fixes https://github.com/status-im/status-go/pull/3665
relates https://github.com/status-im/status-web/issues/459
relates https://github.com/vacp2p/rfc/pull/602
relates https://github.com/vacp2p/rfc/pull/600
closes https://github.com/status-im/status-go/issues/3713

PR #3944, opened 2023-08-23
waku2: static shards

richard-ramos:
# Description

Communities can be assigned to an specific shard cluster and index.
To do so, after the community is created, in the node management section the following json rpc method could be executed:
```
{"method": "wakuext_setCommunityShard", "params": [{"communityId": "0xCommunityID", "shard": { "cluster": 16, "index": 128}, "privateKey": "0xsomePrivateKey or null"}]}
```
- The community id can be obtained with {"method": "wakuext_communities"}
- The second parameter is the shard cluster. The [RFC](https://rfc.vac.dev/spec/57/) suggest using 16 for status communities
- The third parameter is the shard index. Should be a number between 0 and 1023. The [RFC](https://rfc.vac.dev/spec/57/) has some suggestions in which number should be used.
- The fourth parameter is optional, and if not specified, it will auto-generate a private key/public key

Once this method is executed, you can expect to receive the following response:
```
{
  "jsonrpc":"2.0",
  "id":0,
  "result":{
    "protectedTopic":[
      {"pubsubTopic":"/waku/2/rs/<cluster>/<index>","publicKey":"0x......"}
    ]
  }
}
```

This information can be used to setup the fleet nodes to indicate which pubsub topic should be protected and which pubkey to use.

Notice that the code for obtaining the communities description from the smart contract and from NFTs is temporarily commented as these do not contain information that would let us determine the cluster and index assigned to a community

**There's no discovery functionality yet**. For dogfooding, peers supporting a shard should be added manually. For this we can use this RPC method in node management section:
```
{"method": "wakuext_dialPeer", "params": ["peer_multiaddress_goes_here"]}
```

In https://github.com/status-im/infra-shards/issues/2, the following shards will be configured on a new fleet:
- `/waku/2/rs/16/128`
- `/waku/2/rs/16/256`

Once we have it configured, besides adding manually the nodes with `wakuext_dialPeer`, we should use `wakuext_setCommunityShard` and pass the parameters `[{"communityId": "0xCommunityID", "shard": { "cluster": 16, "index": 128 or 256}, "privateKey": "0xsomePrivateKey"}]`. 
(The private keys are already generated. Let me know if you need them)

# Changes
- use protected topics for communities
- associate chats to pubsub topics and populate these depending if the chat belongs to a community or not
- mailserver functions should be aware of pubsub topics
- generate private key for pubsub topic protection when creating a community
- add shard cluster and index to communities
- setup shards for existing communities
- distribute pubsubtopic password


services/ext/service.go
```diff
@@ -567,7 +568,13 @@ func (s *Service) FetchCollectibleMetadata(id thirdparty.CollectibleUniqueID, to
 	}
 
 	// Try to fetch metadata from Messenger communities
-	community, err := s.messenger.RequestCommunityInfoFromMailserver(communityID, true)
+
+	// TODO: we need the shard information in the collectible to be able to retrieve info for
+	// communities that have specific shards
+
+	var shardCluster *uint // TODO: get this value from token
+	var shardIndex *uint   // TODO: get this value from token
```
richard-ramos: This requires smart contract work to obtain the shard cluster and index from the collectibles

protocol/messenger_communities.go
```diff
@@ -558,8 +558,15 @@ func (m *Messenger) CuratedCommunities() (*communities.KnownCommunitiesResponse,
...
+
+	// TODO: The curated communities smart contract should also store the community shard cluster and index
+
 	for _, c := range featuredCommunities {
-		response.ContractFeaturedCommunities = append(response.ContractFeaturedCommunities, types.HexBytes(c).String())
+		response.ContractFeaturedCommunities = append(response.ContractFeaturedCommunities, communities.CommunityShard{
+			CommunityID: types.HexBytes(c).String(),
+			// Cluster:     c.Cluster, // TODO: obtain this value
+			// Index:       c.Index, // TODO: obtain this value
+		})
```
richard-ramos: The curated communities smart contract must be modified to also store the shard cluster and index for a community

richard-ramos:
In line 88 of `messenger_share_url.go` there's this function:
```
func (m *Messenger) ShareCommunityURLWithChatKey(communityID types.HexBytes) (string, error) {
	shortKey, err := m.SerializePublicKey(communityID)
	if err != nil {
		return "", err
	}
	return fmt.Sprintf("%s/c#%s", baseShareURL, shortKey), nil
}
```
Which I see it's used in status-desktop. What should we do for the cases when the community has a specific shard assigned? Should I  do something like this? `fmt.Sprintf("%s/c#%s/%d/%d", shortKey, theCluster, theIndex)`

cc: @jrainville @felicio

felicio:
> In line 88 of `messenger_share_url.go` there's this function:
> 
> ```
> func (m *Messenger) ShareCommunityURLWithChatKey(communityID types.HexBytes) (string, error) {
> 	shortKey, err := m.SerializePublicKey(communityID)
> 	if err != nil {
> 		return "", err
> 	}
> 	return fmt.Sprintf("%s/c#%s", baseShareURL, shortKey), nil
> }
> ```
> 
> Which I see it's used in status-desktop. What should we do for the cases when the community has a specific shard assigned? Should I do something like this? `fmt.Sprintf("%s/c#%s/%d/%d", shortKey, theCluster, theIndex)`
> 
> cc: @jrainville @felicio

@richard-ramos were you able to form an opinion on https://github.com/vacp2p/rfc/pull/600#discussion_r1305672545 yet?

Based on that, if you say it's okay if those fields are exposed to servers, I'd deprecate this types of methods (i.e. without encoded data).

felicio:
And changes to the other methods (i.e. with encoded data) should be done similarly to https://github.com/status-im/status-go/pull/3855.

eth-node/bridge/geth/waku.go
```diff
@@ -69,8 +69,14 @@ func (w *gethWakuWrapper) SubscribeToPubsubTopic(topic string, optPublicKey *ecd
 	return errors.New("not available in WakuV1")
 }
 
+func (w *gethWakuWrapper) RetrievePubsubTopicKey(topic string) (*ecdsa.PrivateKey, error) {
+	// not available in WakuV1
+	return nil, nil
```
ilmotta: Other implementations for v1 in this file return an error like `errors.New("not available in WakuV1")`, so it seems we should follow the pattern to stay consistent.

protocol/common/message_sender.go
```diff
@@ -187,6 +189,40 @@ func (s *MessageSender) SendCommunityMessage(
 	return s.sendCommunity(ctx, &rawMessage)
 }
 
+// SendCommunityMessage takes encoded data, encrypts it and sends through the wire
```
ilmotta: The function name in the comment seems to be outdated.

protocol/transport/filters_manager.go
```diff
@@ -151,7 +153,11 @@ func (f *FiltersManager) InitCommunityFilters(communityFiltersToInitialize []Com
 	defer f.mutex.Unlock()
 
 	for _, cf := range communityFiltersToInitialize {
-		pubsubTopic := GetPubsubTopic(cf.CommunityID)
+		if cf.PrivKey == nil {
+			continue
+		}
+
+		pubsubTopic := GetPubsubTopic(cf.ShardCluster, cf.ShardIndex)
```
ilmotta: The shard's cluster and index seem to be highly correlated, so much so that both numbers appear almost always together. It would be nice to capture this correlation as a single type.

I can tell you would need to change a bunch of function signatures in this PR if you wanted to apply this idea consistently 🤷🏼

protocol/protobuf/community_shard_key.proto
```diff
@@ -0,0 +1,14 @@
...
+option go_package = "./;protobuf";
+
+package protobuf;
+
+message CommunityShardKey {
+  bytes communityID = 1;
+  bytes privateKey = 2;
+  string topic = 3;
+  uint64 clock = 4;
+  int32 shard_cluster = 5;
```
ilmotta: In status-go I see a lot of protos with a mix of camel case and snake case for field names. Not that one style is better than the other, but for consistency sake, @richard-ramos @cammellos do you know if we follow Google's style guide https://protobuf.dev/programming-guides/style/#message-field-names? If we do, then that means all field names should use the snake case.

protocol/protobuf/community_shard_key.proto
```diff
@@ -0,0 +1,14 @@
...
+  int32 shard_index = 6;
+}
```
ilmotta: Missing trailing newline character.

protocol/communities/manager.go
```diff
@@ -853,6 +871,36 @@ func (m *Manager) DeleteCommunity(id types.HexBytes) error {
 	return m.persistence.DeleteCommunitySettings(id)
 }
 
+func (m *Manager) UpdateShard(community *Community, shardCluster, shardIndex *uint) error {
+	community.config.ShardCluster = shardCluster
+	community.config.ShardIndex = shardIndex
+	return m.persistence.SaveCommunity(community)
+}
+
+// EditCommunity takes a description, updates the community with the description,
```
ilmotta: The function name in the comment seems to be outdated.

protocol/communities/persistence.go
```diff
@@ -32,7 +32,7 @@ var ErrOldRequestToLeave = errors.New("old request to leave")
 
 const OR = " OR "
 const communitiesBaseQuery = `
-	SELECT c.id, c.private_key, c.description, c.joined, c.spectated, c.verified, c.muted, c.muted_till, r.clock, ae.raw_events, ae.raw_description
+	SELECT c.id, c.private_key, c.description, c.joined, c.spectated, c.verified, c.muted, c.muted_till, r.clock, ae.raw_events, ae.raw_description, c.shard_cluster, c.shard_index
```
ilmotta, line 36: Worth checking if we should also create indices for both columns in the migration file.

protocol/messenger_communities.go
```diff
@@ -1947,6 +1973,90 @@ func (m *Messenger) CreateCommunity(request *requests.CreateCommunity, createDef
 	return response, nil
 }
 
+func (m *Messenger) SetCommunityShard(communityID types.HexBytes, clusterShard *uint, clusterIndex *uint, privateKey *string) (*MessengerResponse, error) {
```
ilmotta: I like how some endpoints follow the convention of accepting a `request` struct that implements the `Validate` function and then the first thing the implementation does is essentially validate the request sent from clients.

Here we are proceeding, assuming the values are valid based solely on their type, but we know, for instance, that the index is valid only from 0-1023. Sure things will fail later on if let's say the index is 2000, but I think it's more robust if we fail as early as we can without incurring on possible side-effects.

Example endpoint implementing this pattern:

- Endpoint: https://github.com/status-im/status-go/blob/09a988607deaa0282d4330f71bbc93053977d621/services/ext/api.go#L269
- Early validation: https://github.com/status-im/status-go/blob/09a988607deaa0282d4330f71bbc93053977d621/protocol/messenger_chats.go#L300

protocol/messenger_communities.go
```diff
@@ -1947,6 +1973,90 @@ func (m *Messenger) CreateCommunity(request *requests.CreateCommunity, createDef
...
+	community, err := m.communitiesManager.SetShard(communityID, clusterShard, clusterIndex)
```
ilmotta: What are your thoughts about capturing some of the shard allocation recommendations found in https://rfc.vac.dev/spec/57/#shard-allocation as code in status-go? For instance, 0-15 is reserved in the Status cluster.

My perspective is that clients can be complex, buggy beasts, and they easily can, and often do send bad data to status-go.

protocol/messenger_communities.go
```diff
@@ -1947,6 +1973,90 @@ func (m *Messenger) CreateCommunity(request *requests.CreateCommunity, createDef
...
+
+	return response, nil
+}
+
+func (m *Messenger) UpdateCommunityFilters(community *communities.Community, privKey *ecdsa.PrivateKey) error {
+	if err := m.transport.StorePubsubTopicKey(community.PubsubTopic(), privKey); err != nil {
+		return err
+	}
+
+	var publicFiltersToInit []transport.FiltersToInitialize
```
ilmotta: Here we can use `make` with the capacity `len(community.DefaultFilters()) + len(community.Chats())` to declare `publicFiltersToInit`.

protocol/messenger_communities.go
```diff
@@ -2269,6 +2379,56 @@ func (m *Messenger) RemoveUserFromCommunity(id types.HexBytes, pkString string)
...
+		Clock:       m.getTimesource().GetCurrentTime(),
+		CommunityID: community.ID(),
+		PrivateKey:  crypto.FromECDSA(key),
+		Topic:       community.PubsubTopic(),
+	}
+
+	if community.ShardIndex() != nil {
+		communityShardKey.ShardIndex = int32(*community.ShardIndex())
+	} else {
+		communityShardKey.ShardIndex = -1
```
ilmotta: Wdyt we capture `-1` as a constant instead, like `InvalidShardIndex` and update all references to it?

protocol/messenger_response.go
```diff
@@ -211,6 +222,17 @@ func (r *MessengerResponse) CommunitiesSettings() []*communities.CommunitySettin
 	return settings
 }
 
+func (r *MessengerResponse) ProtectedTopics() []ProtectedTopic {
+	var protectedTopics []ProtectedTopic
```
ilmotta: It's better to avoid the overhead of dynamically growing the slice and define it with `make` and the capacity of `len(r.protectedTopics)`.

protocol/messenger_response.go
```diff
@@ -103,6 +112,7 @@ func (r *MessengerResponse) MarshalJSON() ([]byte, error) {
 		Notifications                 []*localnotifications.Notification   `json:"notifications"`
 		Communities                   []*communities.Community             `json:"communities,omitempty"`
 		CommunitiesSettings           []*communities.CommunitySettings     `json:"communitiesSettings,omitempty"`
+		ProtectedTopics               []ProtectedTopic                     `json:"protectedTopic,omitempty"`
```
ilmotta: Shouldn't the JSON key be in the plural (`protectedTopics`)?

protocol/messenger_handlers.go
```diff
@@ -1608,3 +1611,19 @@ func (m *Messenger) handleCommunityPrivilegedUserSyncMessageProtobuf(messageStat
...
+	p := &protobuf.CommunityShardKey{}
+	err := proto.Unmarshal(protoBytes, p)
+	if err != nil {
+		return err
+	}
+
+	m.outputToCSV(msg.TransportMessage.Timestamp, msg.ID, messageState.CurrentMessageState.Contact.ID, filter.ContentTopic, filter.ChatID, msg.Type, p)
+
+	return m.HandleCommunityShardKey(messageState, p, msg)
+
```
ilmotta, line 1628: Extra line.

protocol/communities/persistence.go
richard-ramos: Not sure if necessary as we usually retrieve the full community information and wont do any filtering based on shard fields

protocol/protobuf/community_shard_key.proto
richard-ramos: Having a quick look, it seems we're not being completely consistent with the usage of snake_case across .proto files. Since it seems that convention is used for the majority of protobuffers of status-go, i'll change this file to also use snake_case

protocol/common/shard.go
```diff
@@ -0,0 +1,22 @@
...
+
+import "github.com/status-im/status-go/protocol/transport"
+
+type Shard struct {
+	Cluster uint16 `json:"cluster"`
+	Index   uint16 `json:"index"`
+}
+
+func (s *Shard) TransportShard() *transport.Shard {
+	if s == nil {
```
cammellos, line 25: This looks suspicious, the caller should make sure that the callee is not `nil` in general

protocol/communities/community.go
```diff
@@ -347,6 +363,14 @@ func (o *Community) DescriptionText() string {
 	return ""
 }
 
+func (o *Community) Shard() *common.Shard {
+	if o != nil && o.config != nil {
```
cammellos, line 378: Same here, though now I am starting to think that we do that somewhere else and that's why you used the same pattern :)

protocol/messenger_communities.go
```diff
@@ -1407,6 +1424,11 @@ func (m *Messenger) AcceptRequestToJoinCommunity(request *requests.AcceptRequest
 			return nil, err
 		}
 
+		err = m.SendCommunityShardKey(community, []*ecdsa.PublicKey{pk})
```
cammellos: I would not send a separate message, I think it would be best if this message was sent over with the community accepted message

protocol/messenger_communities.go
```diff
@@ -1947,6 +1974,95 @@ func (m *Messenger) CreateCommunity(request *requests.CreateCommunity, createDef
...
+	for chatID := range community.Chats() {
+		communityChatID := community.IDString() + chatID
+		_, err := m.transport.RemoveFilterByChatID(communityChatID)
+		if err != nil {
+			return err
+		}
+		publicFiltersToInit = append(publicFiltersToInit, transport.FiltersToInitialize{ChatID: communityChatID, PubsubTopic: community.PubsubTopic()})
+	}
+
+	_, err := m.transport.InitPublicFilters(publicFiltersToInit)
```
cammellos, line 2079: maybe we could refactor this function? it seems the same logic could be used somewhere else?

protocol/messenger_communities.go
```diff
@@ -2761,7 +2918,59 @@ func (m *Messenger) HandleCommunityEventsMessageRejected(state *ReceivedMessageS
...
+	err = m.publishCommunityEvents(community, reapplyEventsMessage)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// HandleCommunityShardKey handles the private keys for the community shards
+func (m *Messenger) HandleCommunityShardKey(state *ReceivedMessageState, message *protobuf.CommunityShardKey, statusMessage *v1protocol.StatusMessage) error {
```
cammellos, line 2950: This is risky, it does not seem to support out of order messages (say that the community changes shards twice, last one wins, but we don't check clock etc), I think we need to think about this a bit more

protocol/messenger_communities.go
```diff
@@ -2761,7 +2918,59 @@ func (m *Messenger) HandleCommunityEventsMessageRejected(state *ReceivedMessageS
...
+	community, err := m.communitiesManager.GetByID(message.CommunityId)
```
cammellos, line 2956: getbyid returns nil if the community is not in the db, so we need to handle it, how, we need to think about :)

protocol/messenger_handlers.go
```diff
@@ -1608,3 +1611,18 @@ func (m *Messenger) handleCommunityPrivilegedUserSyncMessageProtobuf(messageStat
 }
 
 
+func (m *Messenger) handleCommunityShardKeyProtobuf(messageState *ReceivedMessageState, protoBytes []byte, msg *v1protocol.StatusMessage, filter transport.Filter) error {
```
cammellos, line 1614: just to make sure, this was generated by the `make generate` command right?

protocol/messenger_response.go
```diff
@@ -34,6 +37,11 @@ type ClearedHistory struct {
 	ClearedAt uint64 `json:"clearedAt"`
 }
 
+type ProtectedTopic struct {
```
cammellos, line 40: does the client need this info?

protocol/migrations/sqlite/1689931301_community_shard.up.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE communities_communities ADD COLUMN shard_cluster INT DEFAULT NULL;
+ALTER TABLE communities_communities ADD COLUMN shard_index INT DEFAULT NULL;
+
+UPDATE communities_communities SET shard_cluster = NULL; -- TODO confirm this
```
cammellos: is it not null by default? same above

cammellos (commented):
@richard-ramos left some comments,
main point of concern is the handling/sending of the pubsubptopic key, I think that needs to be a bit more robust. Happy to merge if it's not a breaking change, though you mentioned that it would be, can we make it a compatible change?

protocol/common/shard.go
richard-ramos: I agree that it is somewhat suspicious but it was a deliverate choice because using this approach does have the advantage of returning `nil` immediately instead of having the callee check if the shard is nil. (Since `.TransportShard()` is called in a lot of places, it saves us having to  repeat this in a lot of places, or have to create an function just to do:
```
var pubsubTopic := relay.DefaultPubsubTopic
if community.Shard() != nil {
   pubsubTopic = transport.GetPubsbuTopic(community.Shard().TransportShard())
}
```
While with this approach I can do this directly:
```
pubsubTopic = transport.GetPubsubTopic(community.Shard().TransportShard())
```
and `pubsubTopic` will have either the default pubsub topic or the shard topic

protocol/communities/community.go
richard-ramos: Ah :-) indeed, this is the result of copy-pasting another accesor for communities

protocol/messenger_handlers.go
richard-ramos: TIL!

protocol/messenger_response.go
richard-ramos: Yes, this information will need to be sent to infra to setup the shard and protected topic in the fleet

protocol/migrations/sqlite/1689931301_community_shard.up.sql
richard-ramos: yeah! just confirmed it. I'll fix it

richard-ramos:
@cammellos i think it should be good to merge now. I ended up adding these functions to avoid introducing breaking changes

- `RequestCommunityInfoFromMailserverWithShard`
- `RequestCommunityInfoFromMailserverAsyncWithShard`

pavloburykh:
Hi @richard-ramos! Could you please rebase and resolve conflicts in current status-go branch so we can perform testing on mobile side https://github.com/status-im/status-mobile/pull/17305 ? Thank you!

richard-ramos:
> Hi @richard-ramos! Could you please rebase and resolve conflicts in current status-go branch so we can perform testing on mobile side [status-im/status-mobile#17305](https://github.com/status-im/status-mobile/pull/17305) ? Thank you!

Took a long time, but it finally has green lights. Will update status-mobile PR now

pavloburykh:
> Took a long time, but it finally has green lights. Will update status-mobile PR now

@richard-ramos thank you! 

Please take a look at the issue. 

### ISSUE 1 Mobile app crashes on joining community

**Steps:**

1. User A (mobile) requests joining community
2. See if User A can join community 

**Actual result:** app crashes on joining community.

[Status-debug-logs (59).zip](https://github.com/status-im/status-mobile/files/12657006/Status-debug-logs.59.zip)

https://github.com/status-im/status-mobile/assets/97245802/17f59bd1-eb20-4dbd-8a69-9f4f15956217

cammellos (approved):

PR #3947, opened 2023-08-24
feat: added an activity center notification that a new keypair was added on a paired device

saledjenic:


MishkaRogachev (approved):
Notification looks ok so far

alaibe (approved):

PR #3948, opened 2023-08-24
feat: Introduce KickedPending and BannedPending states

MishkaRogachev:
For https://github.com/status-im/status-desktop/pull/12068

Important changes:
- [x] Apply kick and ban actions only if control node 
- [x] Add pending states into response

Closes # https://github.com/status-im/status-desktop/issues/11795

### DIFF
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -106,6 +106,15 @@ type CommunityTag struct {
 	Emoji string `json:"emoji"`
 }
 
+type CommunityMemberState uint8
+
+const (
+	CommunityMemberBanned CommunityMemberState = iota
+	CommunityMemberBanPending
+	CommunityMemberUnbanPending
+	CommunityMemberKickPending
+)
+
 func (o *Community) MarshalPublicAPIJSON() ([]byte, error) {
 	if o.config.MemberIdentity == nil {
 		return nil, errors.New("member identity not set")
@@ -127,7 +136,7 @@ func (o *Community) MarshalPublicAPIJSON() ([]byte, error) {
 		Link                    string                               `json:"link"`
 		CommunityAdminSettings  CommunityAdminSettings               `json:"adminSettings"`
 		Encrypted               bool                                 `json:"encrypted"`
-		BanList                 []string                             `json:"banList"`
+		PendingAndBannedMembers map[string]CommunityMemberState      `json:"pendingAndBannedMembers"`
 		TokenPermissions        map[string]*CommunityTokenPermission `json:"tokenPermissions"`
 		CommunityTokensMetadata []*protobuf.CommunityTokenMetadata   `json:"communityTokensMetadata"`
 		ActiveMembersCount      uint64                               `json:"activeMembersCount"`
@@ -169,11 +178,11 @@ func (o *Community) MarshalPublicAPIJSON() ([]byte, error) {
 		}
 
 		communityItem.TokenPermissions = o.tokenPermissions()
+		communityItem.PendingAndBannedMembers = o.PendingAndBannedMembers()
 		communityItem.MembersCount = len(o.config.CommunityDescription.Members)
 		communityItem.Link = fmt.Sprintf("https://join.status.im/c/0x%x", o.ID())
 		communityItem.IntroMessage = o.config.CommunityDescription.IntroMessage
 		communityItem.OutroMessage = o.config.CommunityDescription.OutroMessage
-		communityItem.BanList = o.config.CommunityDescription.BanList
 		communityItem.CommunityTokensMetadata = o.config.CommunityDescription.CommunityTokensMetadata
 		communityItem.ActiveMembersCount = o.config.CommunityDescription.ActiveMembersCount
 
@@ -234,7 +243,7 @@ func (o *Community) MarshalJSON() ([]byte, error) {
 		MuteTill                    time.Time                            `json:"muteTill,omitempty"`
 		CommunityAdminSettings      CommunityAdminSettings               `json:"adminSettings"`
 		Encrypted                   bool                                 `json:"encrypted"`
-		BanList                     []string                             `json:"banList"`
+		PendingAndBannedMembers     map[string]CommunityMemberState      `json:"pendingAndBannedMembers"`
 		TokenPermissions            map[string]*CommunityTokenPermission `json:"tokenPermissions"`
 		CommunityTokensMetadata     []*protobuf.CommunityTokenMetadata   `json:"communityTokensMetadata"`
 		ActiveMembersCount          uint64                               `json:"activeMembersCount"`
@@ -288,11 +297,11 @@ func (o *Community) MarshalJSON() ([]byte, error) {
 			communityItem.Chats[id] = chat
 		}
 		communityItem.TokenPermissions = o.tokenPermissions()
+		communityItem.PendingAndBannedMembers = o.PendingAndBannedMembers()
 		communityItem.Members = o.config.CommunityDescription.Members
 		communityItem.Permissions = o.config.CommunityDescription.Permissions
 		communityItem.IntroMessage = o.config.CommunityDescription.IntroMessage
 		communityItem.OutroMessage = o.config.CommunityDescription.OutroMessage
-		communityItem.BanList = o.config.CommunityDescription.BanList
 		communityItem.CommunityTokensMetadata = o.config.CommunityDescription.CommunityTokensMetadata
 		communityItem.ActiveMembersCount = o.config.CommunityDescription.ActiveMembersCount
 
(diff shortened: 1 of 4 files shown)


protocol/communities/community.go
```diff
@@ -162,11 +171,11 @@ func (o *Community) MarshalPublicAPIJSON() ([]byte, error) {
 		}
 
 		communityItem.TokenPermissions = o.tokenPermissions()
+		communityItem.PendingAndBannedMembers = o.PendingAndBannedMembers()
```
0x-r4bbit: I don't think we want to store this here. We're polluting the `CommunityDescription` with information that is only relevant for admins and the control node.

I think what should rather happen is:

1. Create an new table for community members that tracks their status
2. Update a member's entry state when admin marks as kicked/ban pending
3. Send that info to control node and other admins
4. Other admins will update their own table accordingly (also want to store **who** made that decision, so someone who's done a decision can undo it again)
5. Control node will process the pending state message and sends decision to other admins

^ Notice that the pending state here never needed to be processed or seen by non-admin member nodes.

jrainville: We do not want to make the banned members public (so not part of the description)

0x-r4bbit: To be fair, we already have currently banned members in the `BanList` which lives in the community description.
We should likely remove it from there and use the same mechanism as described above for sharing this information with other admins.

MishkaRogachev: 1. All right, I'll get it out of here. I put it here because the bannedList was here originally, but if we don't show banned users to regular community members, I'll remove it.
2.  I relied on https://github.com/status-im/status-go/pull/3925 to implement this PR. Pending states completely rely on the state obtained from community events and I think a separate table is not needed here. If the ability to cancel requests is needed, it seems reasonable to me to add a new community event that will remove another specified community event from the queue. I would prefer to do this in a different PR.

protocol/communities_events_utils_test.go
```diff
@@ -616,12 +620,62 @@ func kickMember(base CommunityEventsTestsInterface, communityID types.HexBytes,
...
+		}
+
+		return nil
+	}, base.GetMember())
+
+	// 4. control node should handle event and actually kick member
+	waitOnMessengerResponse(s, WaitCommunityCondition, checkKicked, base.GetControlNode())
+
+	// 5. event sender get removed member
+	waitOnMessengerResponse(s, WaitCommunityCondition, checkKicked, base.GetEventSender())
```
jrainville: Does this also make sure that there is no longer the kickedPending state?

jrainville (commented):
LGTM. 

I just want to make sure that once the control processes the kick or ban, the admins no longer have the pending state.

For the clients, you might be missing the Responses containing the new state, but maybe that's already handled by the event handler.

protocol/communities/community_event.go
```diff
@@ -298,35 +298,29 @@ func (o *Community) updateCommunityDescriptionByCommunityEvent(communityEvent Co
 		}
 
 	case protobuf.CommunityEvent_COMMUNITY_MEMBER_KICK:
-		pk, err := common.HexToPubkey(communityEvent.MemberToAction)
-		if err != nil {
-			return err
-		}
-
-		if !o.IsControlNode() && o.IsPrivilegedMember(pk) {
-			return errors.New("attempt to kick an control node or privileged user from non-control node side")
```
mprakhov, line 307: here or in validate we should check that someone didn't create a vulnerable admin event when we're kicking the owner

protocol/communities_events_utils_test.go
MishkaRogachev: Yes,. checkKicked will fail if we still in pending state

protocol/communities/community_event.go
MishkaRogachev: I was about to add those checks in `validateAndGetEventsMessageCommunityDescription`. But next discovered there is already `validateCommunityEvent`. @mprakhov @osmaczko should not we somehow generalise it?

protocol/communities/community.go
```diff
@@ -1371,6 +1374,33 @@ func (o *Community) tokenPermissions() map[string]*CommunityTokenPermission {
...
+	// Non-privileged members should not see pending and banned members
+	if o.config.EventsData == nil || !o.IsPrivilegedMember(o.MemberIdentity()) {
+		return result
+	}
+
+	for _, bannedMemberID := range o.config.CommunityDescription.BanList {
+		result[bannedMemberID] = CommunityMemberBanned
+	}
+
+	for _, event := range o.config.EventsData.Events {
```
osmaczko, line 1459: To be consistent with how events are processed when obtaining the state of pending permissions, the first event should be counted rather than the last one. See: https://github.com/status-im/status-go/blob/6bcf5f1289f9160168574290cbd6f90dede3f8f6/protocol/communities/community.go#L1350

protocol/communities/community_event.go
osmaczko: It could be done as a part of [validateAndFilterEvents](https://github.com/status-im/status-go/blob/6bcf5f1289f9160168574290cbd6f90dede3f8f6/protocol/communities/manager.go#L1371), more precisely there: https://github.com/status-im/status-go/blob/6bcf5f1289f9160168574290cbd6f90dede3f8f6/protocol/communities/community.go#L2148

osmaczko (approved):

protocol/communities/community_event.go
MishkaRogachev: done

protocol/communities/community.go
MishkaRogachev: fixed

protocol/communities/community.go
```diff
@@ -1371,6 +1374,40 @@ func (o *Community) tokenPermissions() map[string]*CommunityTokenPermission {
...
+		return result
+	}
+
+	for _, bannedMemberID := range o.config.CommunityDescription.BanList {
+		result[bannedMemberID] = CommunityMemberBanned
+	}
+
+	processedEvents := make(map[protobuf.CommunityEvent_EventType]bool)
+	for _, event := range o.config.EventsData.Events {
+		if processedEvents[event.Type] {
```
osmaczko: I think that's wrong, there may be multiple same type events for different users. The keys of the map should be the `memberID`.

osmaczko (changes requested):

protocol/communities/community.go
```diff
@@ -2154,12 +2191,20 @@ func (o *Community) ValidateEvent(event *CommunityEvent, signer *ecdsa.PublicKey
...
-	member := o.getMember(signer)
-	if member == nil {
+	eventSender := o.getMember(signer)
+
+	pk, err := common.HexToPubkey(event.MemberToAction)
+	if err != nil {
+		return err
+	}
+
+	member := o.getMember(pk)
```
osmaczko:
```suggestion
	eventTarget := o.getMember(pk)
```
Could increase readability a bit IMO.

protocol/communities/roles_authorization.go
```diff
@@ -56,36 +60,51 @@ var rolesToAuthorizedPermissionTypes = map[protobuf.CommunityMember_Roles][]prot
...
 
 	if event.Type == protobuf.CommunityEvent_COMMUNITY_MEMBER_TOKEN_PERMISSION_CHANGE ||
 		event.Type == protobuf.CommunityEvent_COMMUNITY_MEMBER_TOKEN_PERMISSION_DELETE {
-		return canRolesModifyPermission(roles, event.TokenPermission.Type)
+		return canRolesModifyPermission(senderRoles, event.TokenPermission.Type)
+	}
+
+	if event.Type == protobuf.CommunityEvent_COMMUNITY_MEMBER_BAN ||
+		event.Type == protobuf.CommunityEvent_COMMUNITY_MEMBER_KICK {
+		return canRolesKickOrBanMember(senderRoles, memberRoles)
```
osmaczko, line 116: nice :+1:

protocol/communities/community.go

protocol/communities/community.go

osmaczko (approved):

protocol/communities/roles_authorization.go
```diff
@@ -56,36 +60,51 @@ var rolesToAuthorizedPermissionTypes = map[protobuf.CommunityMember_Roles][]prot
...
+		return true
+	}
+
+	if (slices.Contains(senderRoles, protobuf.CommunityMember_ROLE_ADMIN) ||
+		slices.Contains(senderRoles, protobuf.CommunityMember_ROLE_TOKEN_MASTER)) &&
+		!(slices.Contains(memberRoles, protobuf.CommunityMember_ROLE_ADMIN) ||
+			slices.Contains(memberRoles, protobuf.CommunityMember_ROLE_TOKEN_MASTER) ||
+			slices.Contains(memberRoles, protobuf.CommunityMember_ROLE_OWNER)) {
+		return true
+	}
```
jrainville, line 98: This is a bit more complex than this.
Admin can only kick normal members.
TokenMaster can also kick Admins
Owner can kick everyone.

MishkaRogachev: fixed

mprakhov (approved):
Good job 👍

jrainville (approved):
Looks good :+1:

PR #3952, opened 2023-08-25
continue process after `SaveSyncDisplayName` get `ErrNewClockOlderTha…

qfrank:
when recovering account use seed phrase on status mobile, app will ask user to fill display name, and then will process backedup data, but `SaveSyncDisplayName` will get `ErrNewClockOlderThanCurrent` and causing ens usernames not restored, i believe this is not what we want, this PR fixed this issue.


saledjenic:
@qfrank changes look good, but don't we follow the same recovering flow in both mobile/desktop apps? We have design inconsistency among desktop and mobile apps in that case. I mean why can the user set a display name if he follows the recover an account flow?

qfrank:
> @qfrank changes look good, but don't we follow the same recovering flow in both mobile/desktop apps? We have design inconsistency among desktop and mobile apps in that case. I mean why can the user set a display name if he follows the recover an account flow?

cc @cammellos

saledjenic:
Will try to explain now how the desktop app works in case of Status account recovery. The profile data are mandatory, which means the desktop app considers that the recovery failed only if the profile data are not fetched, otherwise recovery succeeded regardless if all backed-up data are fetched or not.

Now there are 3 scenarios:
- profile data fetched AND all other data are fetched:
  - the only available option is "Continue", which leads the user to the app where everything will be set as it was at the moment when the most recent backed-up data were made,
- profile data fetched AND all other data are partially fetched or not fetched at all: 
  - the only available option is "Continue", which leads the user to the app where all received data are properly set, but in this case, any backed-up data that may come later will be applied later regardless they didn't come during the timeframe we were expecting them to be received,
- profile data are not fetched AND all other data are partially fetched or not fetched at all: 
  - there are 2 available options in this case "Try to fetch the profile again" or "Create new profile with the same chatkey"
    - option 1 simply re-requests data from waku and waits 2 more minutes with the hope that at least profile data will be fetched
    - option 2 points the user to the screen where he is able to set a new display name and photo and after that continues to the app, in this case it looks like a brand new created Status account, but for the known key (seed phrase) and in this case we discard all backed-up data which may come later.


That's how the Status account recovery flow is implemented on the desktop.
Let's see how it should be (what's the real use case for this) and align the desktop and mobile apps.
@John-44 @benjthayer

cammellos:
@saledjenic it's a different user flow, we currently don't support the flow you described. 

The one @qfrank mentioned is when you click on "I am new to status" but you enter the seed phrase. In this case you are setting your profile, but you might still be retrieving from backup.
 
This one https://www.figma.com/file/o4qG1bnFyuyFOvHQVGgeFY/Onboarding-for-Mobile?node-id=2539%3A281163&mode=dev 

The one you mentioned is not implemented in mobile as discussed with @John-44 before

saledjenic:
If that's the issue faced in `"I am new to status"` flow, then I guess `ProcessBackedupMessages` property of the `NodeConfig` should be set to `false` which is actually the default value for it. It's set to `true` on the status-go side just in a few recovery-related cases. So if that's set to `false` all backed-up messages will be discarded, that's what we're using on the desktop app side.
@cammellos @qfrank

qfrank:
emm.. I set `ProcessBackedupMessages` to true in this case before, because user is trying to recover an account. @saledjenic @cammellos

saledjenic:
@qfrank well, then it's not `"I am new to status"` flow, but recovering flow.

Then it's the case described in this [comment](https://github.com/status-im/status-go/pull/3952#issuecomment-1693346180).

In that comment `ProcessBackedupMessages` should be always set to `true` except in described scenario 3, option 2.

qfrank:
> @qfrank well, then it's not `"I am new to status"` flow, but recovering flow.
> 
> Then it's the case described in this [comment](https://github.com/status-im/status-go/pull/3952#issuecomment-1693346180).
> 
> In that comment `ProcessBackedupMessages` should be always set to `true` except in described scenario 3, option 2.

I'm sure it's a recovering flow(use recovery phrase), let me show you the screenshot @saledjenic 
image

saledjenic:
Yes, you can recover either using a seed phrase or a keycard, so in that case we should start fetching backed-up data without asking the user for a display name, if fetching profile data fails then offer two options which is scenario 3 from [here](https://github.com/status-im/status-go/pull/3952#issuecomment-1693346180). 

That way user will be able to enter the display name only if he proceeds with option 2 and only in that case `ProcessBackedupMessages` should be `false` in all other scenarios from that comment including scenario 3, option 1 it should be set to `true`.

cammellos:
@qfrank @saledjenic is correct, is a recovery flow, but you have chosen `I am new to Status`, therefore we could discard backups in that case. 

Given that we don't have the other flow, not sure whether it's desirable at the moment, since it's the only way for a user to recover backup if they are to re-install. In my opinion is fine to keep it as it is, until we implement the other flow, at which point we can set that to false, but happy to change if anyone has a strong opinion.

qfrank:
> @qfrank @saledjenic is correct, is a recovery flow, but you have chosen `I am new to Status`, therefore we could discard backups in that case.
> 
> Given that we don't have the other flow, not sure whether it's desirable at the moment, since it's the only way for a user to recover backup if they are to re-install. In my opinion is fine to keep it as it is, until we implement the other flow, at which point we can set that to false, but happy to change if anyone has a strong opinion.

Yeah, we don't have another flow implemented to support recovery with backedup data ): , i want merge this PR, and i can revert `ProcessBackedupMessages` back when another flow is ready @saledjenic @cammellos 

Btw, we have the option `use recovery phrase` in the flow of `i am new to Status` which confused me. Not sure how you feel ...

saledjenic (approved):
@qfrank ok, np, these changes do not affect the desktop app, I just wanted to discuss the reason, cause at the end both apps should be aligned.

PR #3953, opened 2023-08-25
fix(wallet): balance history used a wrong db for accessing accounts.

IvanBelyakoff:
Balance history was not checked for all chains if no history on some chain.
Removed `SetInitialRange` from wallet API as internal implementation. This method was called on adding a brand new Status account to initialize `blocks_range` table for `OnDemand` transfer fetching strategy.

And [issue](https://github.com/status-im/status-go/issues/3971) created as a follow up to handle a case for a brand new accounts and initial history fetching for ERC20 tokens.

### DIFF
--- a/services/wallet/transfer/reactor.go
+++ b/services/wallet/transfer/reactor.go
@@ -57,6 +57,30 @@ type HistoryFetcher interface {
 		limit int64, fetchMore bool) ([]Transfer, error)
 }
 
+func NewOnDemandFetchStrategy(
+	db *Database,
+	blockDAO *BlockDAO,
+	feed *event.Feed,
+	transactionManager *TransactionManager,
+	pendingTxManager *transactions.PendingTxTracker,
+	tokenManager *token.Manager,
+	chainClients map[uint64]*chain.ClientWithFallback,
+	accounts []common.Address,
+) *OnDemandFetchStrategy {
+	strategy := &OnDemandFetchStrategy{
+		db:                 db,
+		blockDAO:           blockDAO,
+		feed:               feed,
+		transactionManager: transactionManager,
+		pendingTxManager:   pendingTxManager,
+		tokenManager:       tokenManager,
+		chainClients:       chainClients,
+		accounts:           accounts,
+	}
+
+	return strategy
+}
+
 type OnDemandFetchStrategy struct {
 	db                 *Database
 	blockDAO           *BlockDAO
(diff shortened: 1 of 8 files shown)


IvanBelyakoff:
@J-Son89 @qfrank I removed `SetInitialBlockRange` method from wallet API, as it required me to do some refactoring and it feels like it should be internal method that does some `block_range` table initialization. Please confirm that it is OK with you. Feel free to add a person to reviewers who is more aware about it

dlipicar (approved):
LGTM

services/wallet/transfer/reactor.go
```diff
@@ -294,16 +349,8 @@ func (r *Reactor) createFetchStrategy(chainClients map[uint64]*chain.ClientWithF
...
-		db:                 r.db,
-		feed:               r.feed,
-		blockDAO:           r.blockDAO,
-		transactionManager: r.transactionManager,
-		pendingTxManager:   r.pendingTxManager,
-		tokenManager:       r.tokenManager,
-		chainClients:       chainClients,
-		accounts:           accounts,
-	}
+	return NewOnDemandFetchStrategy(r.db, r.feed, r.blockDAO, r.transactionManager, r.pendingTxManager,
```
qfrank: Not enough arguments in call to 'NewOnDemandFetchStrategy'

qfrank (commented):
The mobile app used to call this API (setInitialBlocksRange), but it can handle/tolerate failure when the API call fails. I saw that `CheckRecentHistory` will call `setInitialBlocksRange` internally in your PR (frontend knows less API now, good move) and mobile will call `CheckRecentHistory`, i don't see there is a problem except the compile issue. Thank you for inviting!  cc @rasom

J-Son89:
> @J-Son89 @qfrank I removed `SetInitialBlockRange` method from wallet API, as it required me to do some refactoring and it feels like it should be internal method that does some `block_range` table initialization. Please confirm that it is OK with you. Feel free to add a person to reviewers who is more aware about it

Unfortunately I know almost no knowledge of the inner workings of status go. Perhaps @cammellos or @Samyoul can provide some insight if your concerns have not been addressed yet

qfrank (approved):

IvanBelyakoff:
> > @J-Son89 @qfrank I removed `SetInitialBlockRange` method from wallet API, as it required me to do some refactoring and it feels like it should be internal method that does some `block_range` table initialization. Please confirm that it is OK with you. Feel free to add a person to reviewers who is more aware about it
> 
> Unfortunately I know almost no knowledge of the inner workings of status go. Perhaps @cammellos or @Samyoul can provide some insight if your concerns have not been addressed yet

Thank you, @rasom reached out to me and confirmed generally it is fine to remove the method from API, but we still should call it internally fro accounts that have been just created and have no history, so that we avoid fetching transfers history for these accounts

dlipicar:
> Thank you, @rasom reached out to me and confirmed generally it is fine to remove the method from API, but we still should call it internally fro accounts that have been just created and have no history, so that we avoid fetching transfers history for these accounts

you might find `services/wallet/accounts/watcher.go` useful

stefandunca (approved):
👍🏼

PR #3963, opened 2023-08-28
update generate_handlers_template.txt

qfrank:
Because [Filter.Topic](https://github.com/status-im/status-go/blob/e732000d3168888a3fb3b0c1aa1e56bf75c09d27/protocol/transport/filter.go#L19) removed, we need update `generate_handlers_template.txt` as well to make `make generate` output the right content.


cammellos (approved):

richard-ramos (approved):
TIL about these handlers generators

PR #3964, opened 2023-08-28
Removed settings.Usernames and reactivate PreferredName

qfrank:
this PR removed `settings.Usernames` and reactivate sync `settings.PreferredName` , mobile will use the same API relate to ens as desktop now, e.g. [`Register`](https://github.com/status-im/status-go/blob/ca959467be8932d4dabc1b33b5cdd93fbe0af0e9/services/ens/api.go#L410) 

relate mobile [issue](https://github.com/status-im/status-mobile/issues/13204)
relate mobile [PR](https://github.com/status-im/status-mobile/pull/17127)


saledjenic (approved):
I logged the issue to remove it from the desktop app implementation, here https://github.com/status-im/status-desktop/issues/12017.

protocol/protobuf/sync_settings.proto
```diff
@@ -31,8 +31,7 @@ message SyncSetting {
     DISPLAY_NAME = 13;
     BIO = 14;
     MNEMONIC_REMOVED = 15;
-    ENS_USERNAMES = 16;
-    INCLUDE_WATCHONLY_ACCOUNT = 17;
+    INCLUDE_WATCHONLY_ACCOUNT = 16;
```
igor-sirotin: hmm, shouldn't we keep it `17`?

igor-sirotin (commented):
Thanks for this, @qfrank!
Just one question in comments.

protocol/protobuf/sync_settings.proto
saledjenic: There is no need to be backward compatible till the MVP, otherwise, we will unnecessarily bloat the code.

cammellos: that's not quite it, otherwise it gets crazy, we test things against desktop on mobile to make sure releases are compatible, especially syncing, we need to keep some level of compatibility, otherwise we spend to much time investigating bugs that might be bugs or are just breaking changes. 
Avoid breaking changes when necessary, even if we are pre-mvp (if the change wasn't released etc, there's no issue, but if it was released, and it's a minor change like this, seems very unnecessary to break compatibility).

igor-sirotin: And in this case it's actually a very straight, simple and right way of making things compatible.
This explicit protobuf enum nubmering was created for exactly such case 🙂

qfrank: thank you guys! I heard more voices support my old view now, i should insist it 😅 fixed @igor-sirotin

igor-sirotin:
Thanks for the update 👍

I also just digged a bit more into this topic and found a good approach for marking deleted fields: 

Feel free to add it or not, I'm just sharing what I found 🙂

igor-sirotin (approved):
Thanks for the update 👍

I also just digged a bit more into this topic and found a good approach for marking deleted fields: 

Feel free to add it or not, I'm just sharing what I found 🙂

protocol/protobuf/sync_settings.proto
```diff
@@ -31,6 +31,7 @@ message SyncSetting {
     DISPLAY_NAME = 13;
     BIO = 14;
     MNEMONIC_REMOVED = 15;
+    reserved 16; //do not use this value anymore as it was removed
```
igor-sirotin, line 34: ❤️

protocol/messenger_backup_handler.go
```diff
@@ -205,12 +205,22 @@ func (m *Messenger) handleBackedUpSettings(message *protobuf.SyncSetting) error
 		return nil
 	}
 
-	if settingField != nil && m.config.messengerSignalsHandler != nil {
-		response := wakusync.WakuBackedUpDataResponse{
-			Setting: settingField,
+	if settingField != nil {
+		if message.GetType() == protobuf.SyncSetting_PREFERRED_NAME && message.GetValueString() != "" {
```
saledjenic, line 209: @qfrank the changes look good, I have only 2 general questions:
- Do we need to display the ens name at all on the login screen or we display the display name and use the preferred ens name after login? (This is more about whether is that a request or not?)
- What in case the user changes the setting or removes the ens name but the backup still refers to the preferred ens name which is not in use anymore?

qfrank: > Do we need to display the ens name at all on the login screen or we display the display name and use the preferred ens name after login? (This is more about whether is that a request or not?)

I had the same feeling, and @cammellos made a clarification [here](https://github.com/status-im/status-mobile/pull/17127#issuecomment-1717879690)

> What in case the user changes the setting or removes the ens name but the backup still refers to the preferred ens name which is not in use anymore?

No process in this case , mobile doesn't support remove ENS ATM @saledjenic

PR #3969, opened 2023-08-29
Chore: Calculate finalize status for transactions

Cuteivist:
* Added calculation for finalize status for transactions and multi transactions
* Added ability to filter by finalized status

Video:

https://github.com/status-im/status-go/assets/11396062/925a2d4a-a4ac-493f-bc11-a55c97175829

### DIFF
--- a/services/wallet/activity/activity.go
+++ b/services/wallet/activity/activity.go
@@ -324,6 +324,8 @@ type FilterDependencies struct {
 	tokenSymbol func(token Token) string
 	// use the chainID and symbol to look up token.TokenType and token.Address. Return nil if not found
 	tokenFromSymbol func(chainID *common.ChainID, symbol string) *Token
+	// use to get current timestamp
+	currentTimestamp func() int64
 }
 
 // getActivityEntries queries the transfers, pending_transactions, and multi_transactions tables based on filter parameters and arguments
@@ -379,6 +381,11 @@ func getActivityEntries(ctx context.Context, deps FilterDependencies, addresses 
 		networks = joinItems(chainIDs, nil)
 	}
 
+	layer2Chains := []uint64{common.OptimismMainnet, common.OptimismGoerli, common.ArbitrumMainnet, common.ArbitrumGoerli}
+	layer2Networks := joinItems(layer2Chains, func(chainID uint64) string {
+		return fmt.Sprintf("%d", chainID)
+	})
+
 	startFilterDisabled := !(filter.Period.StartTimestamp > 0)
 	endFilterDisabled := !(filter.Period.EndTimestamp > 0)
 	filterActivityTypeAll := len(filter.Types) == 0
@@ -408,7 +415,7 @@ func getActivityEntries(ctx context.Context, deps FilterDependencies, addresses 
 	})
 
 	queryString := fmt.Sprintf(queryFormatString, involvedAddresses, toAddresses, assetsTokenCodes, assetsERC20, networks,
-		joinedMTTypes)
+		layer2Networks, joinedMTTypes)
 
 	// The duplicated temporary table UNION with CTE acts as an optimization
 	// As soon as we use filter_addresses CTE or filter_addresses_table temp table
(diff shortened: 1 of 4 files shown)


services/wallet/activity/activity.go
```diff
@@ -674,10 +708,13 @@ func getActivityEntries(ctx context.Context, deps FilterDependencies, addresses
 		fromTrType, toTrType,
 		filterAllAddresses, filterAllToAddresses,
 		includeAllStatuses, filterStatusCompleted, filterStatusFailed, filterStatusFinalized, filterStatusPending,
-		FailedAS, CompleteAS, PendingAS,
+		FailedAS, CompleteAS, FinalizedAS, PendingAS,
 		includeAllTokenTypeAssets,
 		includeAllNetworks,
 		transactions.Pending,
+		time.Now().Unix(),
+		604800, // 7 days in seconds for layer 2 finalization
```
IvanBelyakoff: @Cuteivist I don't think it is that simple. Do we handle chains reorgs when transaction actually becomes failed? In this case time enough time has passed, but from what I see, it will be considered as finalized.
And also, maybe it would be a better idea to use block numbers for comparison instead of time to be precise? Because 12s is somewhat approximate time, from what I know.

Cuteivist: @IvanBelyakoff 
1. That's topic I wasn't aware and assumed that we will handle it by setting status to failed, what will result in FailedAS.
2. I am moving block approximation to status-go, so anyway we would use approximated block number, refreshed every 10 minutes. I drafted this idea as it should suffice and it's much faster than to create calculations on SQL side for each transaction.
@stefandunca  Do you think we should use latest block number anyway or this approximation is sufficient?

IvanBelyakoff: @Cuteivist honestly I am not so proficient in finalization process, whether 64 blocks is a sliding window or we should split the number of blocks to 64 to get epochs. And I guess for different chains (L1 vs L2 and even between L2 optimistic vs zk-rolllups) finalization is different, so not sure if the suggested solution covers it.
@alaibe do you have a better understanding of it?

IvanBelyakoff: @Cuteivist Do you see any difficulties in using block number instead of time? In terms of complexity of implementation? Because if all we need is to pass current head block number instead of timestamp and compare transfers block number with it, and if this will improve accuracy, I think it is worth it

Cuteivist: @IvanBelyakoff  I based this on calculations we used in UI.
So for L1 it's 64 block
For L2 we only use hardcoded timestamp like here. If there's better way to calculate it using block number I didn't found it anywhere.

stefandunca: >    @stefandunca  Do you think we should use latest block number anyway or this approximation is sufficient?

AFAIK the real finalization for different chains varies and is impractical to predict in SQL for filtering. However, here we only want to provide an approximation that is fast and good enough. I think this implementation for the sake of filtering is sufficient and should be further discussed with @John-44 if we should indeed give more importance than this.

> Do we handle chains reorgs when transaction actually becomes failed?

@IvanBelyakoff, I'm not sure what you mean by this.

stefandunca (approved):
This looks sufficient for the dynamic filter. Maybe we can exaggerate it or add an "unsure" state so the user can get a more elaborate result while going into details. Let's talk about this more in a follow up discussion.

services/wallet/activity/activity.go
IvanBelyakoff: > @IvanBelyakoff, I'm not sure what you mean by this.

I mean that in case a transaction is included in an orphaned block, it might be reverted. So if we don't handle this case, you will see it happily as finalized. And that scares me, because it is absolutely not correct though quite rare. It might be worth a separate PR though.

Regarding approximation, well, would love to hear from @John-44 how sure we should be about tx finalization. Imo, when I see a bank statement that says "transaction is done", I expect it to be not less then "done", this is not a matter of probability.

Cuteivist:
@stefandunca  @IvanBelyakoff I updated the code according to @John-44 mention about avoiding false positives.
L1 layer requires `80` blocks instead of exactly `64` blocks to be marked as finalized
L2 layer require `7.5` day instead of exactly `7 days` to be marked as finalised

alaibe (approved):

IvanBelyakoff (approved):

PR #3970, opened 2023-08-29
feat(CommunityTokens): CommunityTokens API for handling signer pubkey

### LINKED ISSUE #11954: [Community Ownership] deploy Owner Token through Deployer contract
The Owner Token must be deployed by the community address. This is to prevent malicious owners from deploying multiple Owner Tokens with potentially harmful intentions in the future.

This will effectively bind the contract to the community. Clients will verify the owner token contract address by reviewing the transaction history of the community address; only the first deployment is valid.

Edit: Use Deployer contract method described in comments below

---

Ignore this if the way above works

To facilitate the deployment by the community address, it needs sufficient funds to cover the gas fees. Therefore, the procedure should be:
1. The owner transfers ETH to the community address to cover gas fees.
2. The community address deploys the Owner Token contract.
3. The community address refunds any remaining ETH to the owner.
4. The owner [sets signerPublicKey](https://github.com/status-im/communities-contracts/blob/8d3f185bf4d382d522815cff7cb2b70902cfffa8/contracts/mvp/OwnerToken.sol#L42) with its chat PublicKey.

---

### Acceptance criteria:
- [ ] Owner Token contract is deployed by community address
- [ ] Owner Token's [signerPublicKey](https://github.com/status-im/communities-contracts/blob/8d3f185bf4d382d522815cff7cb2b70902cfffa8/contracts/mvp/OwnerToken.sol#L10) property points to Owner's chat PublicKey

#### Implementation:
Ideally, these four steps should be executed in a single atomic transaction. This ensures that either all operations succeed, or none are executed, preventing any intermediate and undesirable state. If executing all in one transaction is not feasible, a robust recovery/retry mechanism should be implemented in the event that any of the transactions fail.

#### Considerations:
Extend Owner Token's constructor with `signerPublicKey` parameter to avoid step 4.

endulab:
New contracts and contract go functions.
Adjust owner&master tokens deployment flow.
Create deployment signature (needed for deployment)
CommunityTokens API for handling signer pubkey.

_Please do not review contracts and autogenerated go files._


services/communitytokens/api.go
```diff
@@ -244,20 +328,105 @@ func (api *API) DeployAssets(ctx context.Context, chainID uint64, deploymentPara
 
 // Returns gas units + 10%
 func (api *API) DeployCollectiblesEstimate(ctx context.Context) (uint64, error) {
-	gasAmount := uint64(2091605)
+	// TODO investigate why the code below does not return correct values
```
osmaczko, line 331: Is there a task for it?

services/communitytokens/api.go
```diff
@@ -679,3 +868,91 @@ func (api *API) estimateMethod(ctx context.Context, chainID uint64, contractAddr
...
+	}
+	signerPubKey, err := contractInst.SignerPublicKey(callOpts)
+	if err != nil {
+		return "", err
+	}
+	return common.Bytes2Hex(signerPubKey), nil
+}
+
+// Gets signer public key directly from deployer contract
+func (api *API) SafeGetSignerPubKey(ctx context.Context, chainID uint64, communityID string) (string, error) {
```
osmaczko, line 887: Why is it "safe"?

osmaczko (approved):
Nice work :+1:

jrainville (approved):
Good job

services/communitytokens/api.go
endulab: There are 2 safe functions: SafeGetSignerPubKey and SafeGetOwnerTokenAddress. I called them "safe" because I wanted to indicate that that these functions use directly deployer contract so we can trust this information.
If it is not clear enough I can change it of course. 
Another way is to get owner token address from, eg. CommunityDescription and use GetSignerPubKey(chainID, contractAddress) if we trust that the contract address in CommunityDescription was not malformed.

services/communitytokens/api.go
endulab: I've just created: https://github.com/status-im/status-go/issues/4084

services/communitytokens/api.go
osmaczko: Is there a scenario where clients of the public API would be interested in the "non-safe" version? If not, what do you think about renaming the methods as follows:
```
func (api *API) getSignerPubKey(ctx context.Context, chainID uint64, contractAddress string) (string, error) {...}
func (api *API) GetSignerPubKey(ctx context.Context, chainID uint64, communityID string) (string, error) {...}
```

endulab: Honestly, I am not sure. I created the API before the logic.
However, I am open to any changes. At this moment I am not sure which function will be used.

PR #3973, opened 2023-08-30
feat: Estimate block number

Cuteivist:
task https://github.com/status-im/status-desktop/issues/11564

Change based on https://github.com/status-im/status-desktop/pull/11522 , it was removed on desktop by mistake as part of refactoring
* add api to estimate block number
* refresh block estimation for each chain in the background every 10 minutes

### DIFF
--- a/services/wallet/blockchainstate.go
+++ b/services/wallet/blockchainstate.go
@@ -1,0 +1,149 @@
+package wallet
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/ethereum/go-ethereum/log"
+
+	"github.com/status-im/status-go/multiaccounts/accounts"
+	"github.com/status-im/status-go/rpc"
+	"github.com/status-im/status-go/services/wallet/async"
+	"github.com/status-im/status-go/services/wallet/common"
+)
+
+const (
+	fetchLatestBlockNumbersInterval = 10 * time.Minute
+)
+
+type fetchLatestBlockNumberCommand struct {
+	state      *BlockChainState
+	rpcClient  *rpc.Client
+	accountsDB *accounts.Database
+}
+
+func (c *fetchLatestBlockNumberCommand) Command() async.Command {
+	return async.InfiniteCommand{
+		Interval: fetchLatestBlockNumbersInterval,
+		Runable:  c.Run,
+	}.Run
+}
+
+func (c *fetchLatestBlockNumberCommand) Run(parent context.Context) (err error) {
+	log.Debug("start fetchLatestBlockNumberCommand")
+
+	networks, err := c.rpcClient.NetworkManager.Get(false)
+	if err != nil {
+		return nil
+	}
+	areTestNetworksEnabled, err := c.accountsDB.GetTestNetworksEnabled()
+	if err != nil {
+		return
+	}
+	ctx := context.Background()
+	for _, network := range networks {
+		if network.IsTest != areTestNetworksEnabled {
+			continue
+		}
+		_, _ = c.state.fetchLatestBlockNumber(ctx, network.ChainID)
+	}
+	return nil
+}
+
+type LatestBlockData struct {
+	blockNumber   uint64
+	timestamp     time.Time
+	blockDuration time.Duration
+}
+
+type BlockChainState struct {
+	rpcClient          *rpc.Client
+	accountsDB         *accounts.Database
+	blkMu              sync.RWMutex
+	latestBlockNumbers map[uint64]LatestBlockData
+	group              *async.Group
+	cancelFn           context.CancelFunc
+	sinceFn            func(time.Time) time.Duration
+}
+
+func NewBlockChainState(rpcClient *rpc.Client, accountsDb *accounts.Database) *BlockChainState {
+	return &BlockChainState{
+		rpcClient:          rpcClient,
+		accountsDB:         accountsDb,
+		blkMu:              sync.RWMutex{},
+		latestBlockNumbers: make(map[uint64]LatestBlockData),
+		sinceFn:            time.Since,
+	}
+}
+
+func (s *BlockChainState) Start() {
+	ctx, cancel := context.WithCancel(context.Background())
+	s.cancelFn = cancel
+	s.group = async.NewGroup(ctx)
+
+	command := &fetchLatestBlockNumberCommand{
+		state:      s,
+		accountsDB: s.accountsDB,
+		rpcClient:  s.rpcClient,
+	}
+	s.group.Add(command.Command())
+}
+
+func (s *BlockChainState) Stop() {
+	if s.cancelFn != nil {
+		s.cancelFn()
+		s.cancelFn = nil
+	}
+	if s.group != nil {
+		s.group.Stop()
+		s.group.Wait()
+		s.group = nil
+	}
+}
+
+func (s *BlockChainState) GetEstimatedLatestBlockNumber(ctx context.Context, chainID uint64) (uint64, error) {
+	blockNumber, ok := s.estimateLatestBlockNumber(chainID)
+	if ok {
+		return blockNumber, nil
+	}
+	return s.fetchLatestBlockNumber(ctx, chainID)
+}
+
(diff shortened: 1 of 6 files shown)


services/wallet/activity/service.go
```diff
@@ -39,23 +42,46 @@ var (
...
+		db:                 db,
+		accountsDB:         accountsDb,
+		tokenManager:       tokenManager,
+		eventFeed:          eventFeed,
+		rpcClient:          rpcClient,
+		scheduler:          async.NewMultiClientScheduler(),
+		blkMu:              sync.Mutex{},
+		latestBlockNumbers: make(map[uint64]LatestBlockData),
 	}
+	go func() {
```
dlipicar: You might want to start and stop this along with the service (`func (s *Service) Stop()`, `Start` needs to be added).
You could use `InfiniteCommand` + `Group` from the `async` package. Handles that, graceful thread stop on app exit, etc.

services/wallet/activity/service.go
```diff
@@ -39,23 +42,46 @@ var (
...
+		allChains := []uint64{w_common.EthereumMainnet, w_common.EthereumGoerli, w_common.OptimismMainnet, w_common.OptimismGoerli, w_common.ArbitrumMainnet, w_common.ArbitrumGoerli}
```
dlipicar: do we want to have this running for testnets even if we're in mainnet mode? (and viceversa)

anyway, instead of hardcoding the list, better to use

```
	networks, err := c.networkManager.Get(false)
	if err != nil {
		return err
	}

	areTestNetworksEnabled, err := c.accountsDB.GetTestNetworksEnabled()
	if err != nil {
		return err
	}

	for _, network := range networks {
		if network.IsTest != areTestNetworksEnabled {
			continue
		}
                 ............
	}
```

services/wallet/activity/service.go
```diff
@@ -109,6 +135,40 @@ func (s *Service) GetTxDetails(ctx context.Context, id string) (*EntryDetails, e
...
+	blockDuration, found := w_common.AverageBlockDurationForChain[w_common.ChainID(chainID)]
+	if !found {
+		blockDuration = w_common.AverageBlockDurationForChain[w_common.ChainID(w_common.UnknownChainID)]
+	}
+	timeDiff := time.Since(latestBlockData.timestamp)
+	blockNumber := latestBlockData.blockNumber + uint64(timeDiff/blockDuration)
+	return blockNumber, nil
+}
+
+func (s *Service) fetchLatetBlockNumber(ctx context.Context, chainID uint64) (uint64, error) {
```
dlipicar: typo Latet -> Latest

services/wallet/activity/service.go
```diff
@@ -109,6 +135,40 @@ func (s *Service) GetTxDetails(ctx context.Context, id string) (*EntryDetails, e
 	return getTxDetails(ctx, s.db, id)
 }
 
+func (s *Service) GetLatestBlockNumber(ctx context.Context, chainID uint64) (uint64, error) {
+	s.blkMu.Lock()
```
dlipicar: personal suggestion, I usually like to use only `[R]Lock()` and `defer [R]Unlock()`, so I wrap access to mutex-protected resources in their own function.
In this case, getting or setting latestBlockNumbers should be their own function so we don't have to mess with the mutex in too many places.

dlipicar:
Would be nice to have some UTs to make sure estimation + correction works as expected

stefandunca (commented):
I agree with @dlipicar improvement recommendations. Also, how about keeping it in its own file and independent of activity (e.g. `blockchainstate.go`)?

services/wallet/activity/service.go
```diff
@@ -109,6 +135,40 @@ func (s *Service) GetTxDetails(ctx context.Context, id string) (*EntryDetails, e
 	return getTxDetails(ctx, s.db, id)
 }
 
+func (s *Service) GetLatestBlockNumber(ctx context.Context, chainID uint64) (uint64, error) {
```
stefandunca: How about a name closer to its purpose e.g.`EstimatedCurrentBlockNumber`?

services/wallet/activity/service.go
```diff
@@ -109,6 +135,40 @@ func (s *Service) GetTxDetails(ctx context.Context, id string) (*EntryDetails, e
...
+	s.blkMu.Lock()
+	latestBlockData, ok := s.latestBlockNumbers[chainID]
+	s.blkMu.Unlock()
+	if !ok {
+		return s.fetchLatetBlockNumber(ctx, chainID)
+	}
+	blockDuration, found := w_common.AverageBlockDurationForChain[w_common.ChainID(chainID)]
```
stefandunca: If we already fetch it regularly, why not keep and use the history instead of the hardcoded? We can drop the hard coding and use the fetched altogether.

Cuteivist: @stefandunca 
So do you mean creating our own estimation instead of using hardcoded estimations?

That could be done, but I think it is out of the scope of this task, but I'm more than happy to create low priority one for the future.

services/wallet/activity/service.go
Cuteivist: Thank you for pointing that out @dlipicar , neat API to use.

Updated code

services/wallet/activity/service.go
Cuteivist: You are right, updated the code.

services/wallet/activity/service.go
Cuteivist: Updated

services/wallet/activity/service.go
Cuteivist: Good point, updated

Cuteivist:
> I agree with @dlipicar improvement recommendations. Also, how about keeping it in its own file and independent of activity (e.g. `blockchainstate.go`)?

@stefandunca  Moved to separate file

> Would be nice to have some UTs to make sure estimation + correction works as expected

@dlipicar  Added UTs, although I couldn't figure out good way to handle more because of dependencies.

services/wallet/blockchainstate.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func (s *BlockChainState) setLatestBlockDataForChain(chainID uint64, latestBlockData LatestBlockData) {
+	s.blkMu.Lock()
+	defer s.blkMu.Unlock()
+	s.latestBlockNumbers[chainID] = latestBlockData
+}
+
+func (s *BlockChainState) estimateLatestBlockNumber(chainID uint64) (uint64, bool) {
+	s.blkMu.Lock()
```
dlipicar: `RLock`, `RUnlock`

services/wallet/blockchainstate_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+	state.setLatestBlockDataForChain(1, LatestBlockData{
+		blockNumber:   uint64(100),
+		timestamp:     time.Now().Add(-31 * time.Second),
+		blockDuration: 10 * time.Second,
+	})
+
+	state.setLatestBlockDataForChain(2, LatestBlockData{
+		blockNumber:   uint64(200),
+		timestamp:     time.Now().Add(-5 * time.Second),
```
dlipicar: not super happy about the UTs depending on real time measurements... If there was a big slowdown for whatever reason they could begin to randomly fail.
Not sure if we've got a nice mock for the `time` module somewhere, but if not you could add a member `TimeGetFn` to `BlockChainState`, which normally gets populated with `time.Now()` and will let you override it (ie. set any arbitrary current time value) for testing purposes.

services/wallet/blockchainstate.go
Cuteivist: Thanks, that's new feature for me to know :) Fixed

services/wallet/blockchainstate_test.go
Cuteivist: Good point. I added new method to calculate since mocked date.

dlipicar (approved):
Awesome work!

stefandunca (approved):

PR #3935, opened 2023-08-22
Link previews not working on message edit

ibrkhalil:
Allows the message preview data to be sent to status-go on message edit.

Closes #[16728 on status-mobile](https://github.com/status-im/status-mobile/issues/16728)

PR #3974, opened 2023-08-30
Link previews not working on message edit

ibrkhalil:
Allows the message preview data to be sent to status-go on message edit.

Closes #https://github.com/status-im/status-mobile/issues/16728


ibrkhalil:
Accidentally closed https://github.com/status-im/status-go/pull/3935, So I opened this new one.

protocol/messenger_edit_message_test.go
```diff
@@ -445,3 +445,76 @@ func (s *MessengerEditMessageSuite) TestEditMessageWithMention() {
...
+	}
+
+	sendResponse, err = theirMessenger.EditMessage(context.Background(), editedMessage)
+
+	s.Require().NoError(err)
+	s.Require().Len(sendResponse.Messages(), 1)
+	s.Require().NotEmpty(sendResponse.Messages()[0].LinkPreviews)
+	response, err = WaitOnMessengerResponse(
+		s.m,
+		func(r *MessengerResponse) bool { return len(r.messages) > 0 },
```
cammellos: If you are expecting exactly one message, you want to check for exactly one message here

protocol/messenger_messages.go
```diff
@@ -50,7 +50,11 @@ func (m *Messenger) EditMessage(ctx context.Context, request *requests.EditMessa
 
 	response := &MessengerResponse{}
 
-	for _, message := range messages {
+	for index, message := range messages {
+		//Add LinkPreviews only to first message
+		if len(request.LinkPreviews) > 0 && index == 0 {
```
cammellos: mmh, does not look great, you should rather match by id rather than relaying on the order of messages, since you have the message id I believe.

Also, we should update link previews on all messages in a image album? or not? have you tested it?

Could you also please rename `getConnectedMessages` to something like `getOtherMessagesInAlbum` (or something similar, since that's what it does? (I know you didn't write it, but it took me ages to figure out what it was doing, as connected messages as a different meaning in the rest of the codebase)

cammellos (commented):
Looks good for the most part, just a couple of comments

protocol/messenger_edit_message_test.go
ibrkhalil: Done

protocol/messenger_messages.go
ibrkhalil: Renamed getConnectedMessages, And removed thh index check
Thanks Andrea, Thinking more about it, I think it's better to apply it to all images since if one images was removed we don't lose the preview data

cammellos: do test it though in the frontend, just to make sure

cammellos (approved):

protocol/messenger_messages.go
ibrkhalil: Turns out we don't support editing image messages on status-mobile yet.

PR #3976, opened 2023-08-30
fix(wallet) fix pending transactions notification

stefandunca:
Also, add regression test.


endulab (approved):
I am not familiar with the code but I tested transaction and I confirm that the notification is received now.

IvanBelyakoff (approved):
Good job!

PR #3978, opened 2023-08-31
ci: hide `go test` output to the file on CI

### LINKED ISSUE #3543: Huge volume of logs and number of stack traces
### Problem

The volume of logs from the `tests` CI job is ridiculous:
```
 > wc -l status_go_tests.log
258778 status_go_tests.log

 > du -hs status_go_tests.log
49M	status_go_tests.log
```
It's almost not readable in the browser, and `vim` requires a `-u NONE` or `'+no syntax'` allow the file to load.

### Future Steps

One thing that would be great to do would be to split the logs from tests into separate files depending on test.
In a perfect world we could also prefix those log files with `failure_` or `success_` to easily find the issue.
This might be very difficult or impossible to do with Go test logging, so we should try to reduce the volume first.

yakimant:
Let's hide stdout of go tests to the files.
In case of failed test - it should be included in junit report anyway and shown in Jenkins.

Closes #3543

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,6 @@ GIT_COMMIT = $(shell git rev-parse --short HEAD)
 AUTHOR ?= $(shell git config user.email || echo $$USER)
 
 ENABLE_METRICS ?= true
-BUILD_TAGS ?= 
 BUILD_FLAGS ?= $(shell echo "-ldflags='\
 	-X github.com/status-im/status-go/params.Version=$(RELEASE_TAG:v%=%) \
 	-X github.com/status-im/status-go/params.GitCommit=$(GIT_COMMIT) \
@@ -48,7 +47,6 @@ BUILD_FLAGS_MOBILE ?= $(shell echo "-ldflags='\
 	-X github.com/status-im/status-go/params.IpfsGatewayURL=$(IPFS_GATEWAY_URL)'")
 
 networkid ?= StatusChain
-gotest_extraflags =
 
 DOCKER_IMAGE_NAME ?= statusteam/status-go
 BOOTNODE_IMAGE_NAME ?= statusteam/bootnode
@@ -311,28 +309,23 @@ docker-test: ##@tests Run tests in a docker container with golang.
 
 test: test-unit ##@tests Run basic, short tests during development
 
-test-unit: SHELL := /bin/bash -o pipefail
-test-unit: UNIT_TEST_PACKAGES = $(shell go list ./...  | \
+test-unit: export BUILD_TAGS ?=
+test-unit: export UNIT_TEST_PACKAGES ?= $(shell go list ./... | \
 	grep -v /vendor | \
 	grep -v /t/e2e | \
 	grep -v /t/benchmarks | \
 	grep -v /transactions/fake )
 test-unit: ##@tests Run unit and integration tests
-	for package in $(UNIT_TEST_PACKAGES); do \
-		set -e; \
-		package_dir=$$(go list -f {{.Dir}} $${package}); \
-		go test -tags '$(BUILD_TAGS)' -timeout 30m -v -failfast $${package} $(gotest_extraflags) | \
-		  go-junit-report -iocopy -out $${package_dir}/report.xml; \
-	done
+	./_assets/scripts/run_unit_tests.sh
 
-test-unit-race: gotest_extraflags=-race
+test-unit-race: export GOTEST_EXTRAFLAGS=-race
 test-unit-race: test-unit ##@tests Run unit and integration tests with -race flag
 
 test-e2e: ##@tests Run e2e tests
 	# order: reliability then alphabetical
 	# TODO(tiabc): make a single command out of them adding `-p 1` flag.
 
-test-e2e-race: gotest_extraflags=-race
+test-e2e-race: export GOTEST_EXTRAFLAGS=-race
 test-e2e-race: test-e2e ##@tests Run e2e tests with -race flag
 
 canary-test: node-canary
--- a/_assets/scripts/run_unit_tests.sh
+++ b/_assets/scripts/run_unit_tests.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+set -o pipefail
+
+GIT_ROOT=$(cd "${BASH_SOURCE%/*}" && git rev-parse --show-toplevel)
+
+source "${GIT_ROOT}/_assets/scripts/colors.sh"
+
+for package in ${UNIT_TEST_PACKAGES}; do
+  echo -e "${GRN}Testing:${RST} ${package}"
+  package_dir=$(go list -f "{{.Dir}}" "${package}")
+  output_file=${package_dir}/test.log
+
+  go test -tags "${BUILD_TAGS}" -timeout 30m -v -failfast "${package}" ${GOTEST_EXTRAFLAGS} | \
+     if [ "${CI}" = "true" ]; then cat > "${output_file}"; else tee "${output_file}"; fi
+  go_test_exit=$?
+
+  if [ "${CI}" = "true" ]; then
+    go-junit-report -in "${output_file}" -out "${package_dir}"/report.xml
+  fi
+
+  if [ ${go_test_exit} -ne 0 ]; then
+    echo -e "${YLW}Failed, see the log:${RST} ${BLD}${output_file}${RST}"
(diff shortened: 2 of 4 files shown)


yakimant:
Failure looks like that:
```
[2023-08-31T11:49:39.455Z] Testing github.com/status-im/status-go/api
[2023-08-31T11:50:03.590Z] make: *** [Makefile:319: test-unit] Error 1
```
https://ci.status.im/blue/organizations/jenkins/status-go%2Fprs%2Ftests/detail/PR-3978/1/pipeline

With stacktrace and debug saved in
https://ci.status.im/blue/organizations/jenkins/status-go%2Fprs%2Ftests/detail/PR-3978/1/tests/

With second commit the output is better, log file destination is printed with failure message.

_assets/scripts/run_unit_tests.sh
```diff
@@ -0,0 +1,28 @@
...
+set -o pipefail
+
+UNIT_TEST_PACKAGES=$(go list ./...  | \
+  grep -v /vendor | \
+  grep -v /t/e2e | \
+  grep -v /t/benchmarks | \
+  grep -v /transactions/fake )
+
+for package in ${UNIT_TEST_PACKAGES}; do
+  echo "Testing ${package}"
```
jakubgs: Maybe some more color here too?
```suggestion
  echo -e "${GRN}Testing:${RST} ${package}"
```

_assets/scripts/run_unit_tests.sh
```diff
@@ -0,0 +1,28 @@
...
+  go test -tags "${BUILD_TAGS}" -timeout 30m -v -failfast "${package}" ${GOTEST_EXTRAFLAGS} | \
+     if [ "${CI}" = "true" ]; then cat > "${output_file}"; else tee "${output_file}"; fi
+  go_test_exit=$?
+
+  if [ "${CI}" = "true" ]; then
+    go-junit-report -in "${output_file}" -out "${package_dir}"/report.xml
+  fi
+
+  if [ ${go_test_exit} -ne 0 ]; then
+    echo "  Failed, see the log: ${output_file}"
```
jakubgs: This is good, but it would be nice to make it a bit more visible for people:
```suggestion
    echo -e "${YLW}Failed, see the log:${RST} ${BLD}${output_file}${RST}"
```
Of course we'd need the colors:
```sh
export YLW='\033[1;33m'
export RED='\033[0;31m'
export GRN='\033[0;32m'
export BLU='\033[0;34m'
export BLD='\033[1m'
export RST='\033[0m'
```

_assets/scripts/run_unit_tests.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
+set -o pipefail
+
+UNIT_TEST_PACKAGES=$(go list ./...  | \
```
jakubgs: I wonder if maybe we should keep this in the `Makefile` and use `?=` so it's overridable. What do you think?

jakubgs (approved):
Nice. This will make logs much easier to follow.

_assets/scripts/run_unit_tests.sh
jakubgs: The colors could be in a separate file, as we do it in `status-mobile`:
https://github.com/status-im/status-mobile/blob/develop/scripts/colors.sh

yakimant: Good idea, let me try, how it will work on CI and locally.

_assets/scripts/run_unit_tests.sh
yakimant: Yeah, I think it's a good idea to leave developers an option to override it and run a limited scope of tests.

Makefile
```diff
@@ -311,28 +311,24 @@ docker-test: ##@tests Run tests in a docker container with golang.
 
 test: test-unit ##@tests Run basic, short tests during development
 
-test-unit: SHELL := /bin/bash -o pipefail
-test-unit: UNIT_TEST_PACKAGES = $(shell go list ./...  | \
+test-unit: export BUILD_TAGS ?= BUILD_TAGS
+test-unit: export GOTEST_EXTRAFLAGS ?= GOTEST_EXTRAFLAGS
```
jakubgs: This seems pretty redundant.

Makefile
```diff
@@ -48,7 +48,7 @@ BUILD_FLAGS_MOBILE ?= $(shell echo "-ldflags='\
 	-X github.com/status-im/status-go/params.IpfsGatewayURL=$(IPFS_GATEWAY_URL)'")
 
 networkid ?= StatusChain
-gotest_extraflags =
+GOTEST_EXTRAFLAGS ?=
```
jakubgs: Whta is this for even?

Makefile
yakimant: I can remove `BUILD_TAGS`, because it is not set in `Makefile` currently.
As for `GOTEST_EXTRAFLAGS` - it is needed to passthrough the env variable, when set in `Makefile`, for example here:
https://github.com/status-im/status-go/blob/81b94b7a4e7ed81f6d0ae78cf3dcbda950e5d278/Makefile#L328-L329
Or do you want to pass it as an argument to the script?

jakubgs: I think the line 51 is pointless. And that variable should be set per-target using `?=`.

Makefile
```diff
@@ -311,28 +309,24 @@ docker-test: ##@tests Run tests in a docker container with golang.
...
-		set -e; \
-		package_dir=$$(go list -f {{.Dir}} $${package}); \
-		go test -tags '$(BUILD_TAGS)' -timeout 30m -v -failfast $${package} $(gotest_extraflags) | \
-		  go-junit-report -iocopy -out $${package_dir}/report.xml; \
-	done
-
-test-unit-race: gotest_extraflags=-race
+	export BUILD_TAGS=$(BUILD_TAGS) ; \
+	export GOTEST_EXTRAFLAGS=$(GOTEST_EXTRAFLAGS) ; \
+	./_assets/scripts/run_unit_tests.sh
```
jakubgs: This is silly. Just use `export` keyword where the variable is defined.
```suggestion
	./_assets/scripts/run_unit_tests.sh
```

Makefile
```diff
@@ -311,28 +309,24 @@ docker-test: ##@tests Run tests in a docker container with golang.
...
+
+test-unit-race: GOTEST_EXTRAFLAGS=-race
```
jakubgs: Just use `export` here, there's no need to add `export` lines within the target.
```suggestion
test-unit-race: export GOTEST_EXTRAFLAGS = -race
```
And use spaces, they get stripped.

Makefile
```diff
@@ -311,28 +309,24 @@ docker-test: ##@tests Run tests in a docker container with golang.
...
 test-unit-race: test-unit ##@tests Run unit and integration tests with -race flag
 
 test-e2e: ##@tests Run e2e tests
 	# order: reliability then alphabetical
 	# TODO(tiabc): make a single command out of them adding `-p 1` flag.
 
-test-e2e-race: gotest_extraflags=-race
+test-e2e-race: GOTEST_EXTRAFLAGS=-race
```
jakubgs: Same here, just use `export` here:
```suggestion
test-e2e-race: export GOTEST_EXTRAFLAGS = -race
```

Makefile
yakimant: Agree, removed the extra exports

Makefile
```diff
@@ -35,7 +35,6 @@ GIT_COMMIT = $(shell git rev-parse --short HEAD)
 AUTHOR ?= $(shell git config user.email || echo $$USER)
 
 ENABLE_METRICS ?= true
-BUILD_TAGS ?=
```
jakubgs, line 38: If you're going to remove `BUILD_TAGS` from here then you need to add them to every target that uses it.

Otherwise just leave it here, but add `export`.

yakimant: As I understand, `BUILD_TAGS` is only set outside of `Makefile` and `make` sets `$(BUILD_TAGS)` according to it.

If `BUILD_TAGS` is not set as env var, `$(BUILD_TAGS)` is empty.

jakubgs: Yeah. You're right. I think it's just useful to have an explicit indication that `BUILD_TAGS` is being used by a target, so lets add an `export BUILD_TAGS ?=` to the test target to indicate that the script uses it.

Issue #3801, opened 2023-07-25
Improper AC notifications when accepting contact request after unblocking.


igor-sirotin:
# Problem

Checkout `MessengerContactRequestSuite/TestBlockedContactSyncing` in https://github.com/status-im/status-go/pull/3799:
1. setup Bob profile
2. setup Alice on 2 paired devices
3. make Alice and Bob mutual contacts
4. Alice: block Bob
5. Alice: unblock Bob
6. Repeat 3-5.

(probably same can be reproduced with simply removing the contact, not block/unblock).

During this, `acceptContactRequest` is called twice.

First time, Alice-2 response will contain:
1. Messages:
    - `@<user> accepted your contact request`, ContentType: `SYSTEM_MESSAGE_MUTUAL_EVENT_ACCEPTED`
    - `hello-1`, ContentType: `CONTACT_REQUEST`
2. ActivityCenterNotifications:
    - Message: `<hello-1>` , Type: CONTACT_REQUEST.

Second time, Alice-2 response will contain:
1. Messages:
    - `@<user> accepted your contact request`. ContentType: `SYSTEM_MESSAGE_MUTUAL_EVENT_ACCEPTED`
    - `hello-2`. ContentType: `CONTACT_REQUEST`
2. ActivityCenterNotifications:
    - Message: `<gen-pubkey-0x20>`.	Type: 5 (CR).
    - Message: `<hello-1>`.	 		Type: 5 (CR).
    - Message: `<hello-2>`.	 		Type: 5 (CR).

Note 3 AC notifications. This is wrong.

# Acceptance Criteria

After any number of cr-block-unblock iterations, the result should be the same.

igor-sirotin:
Probably happens because of AC syncing.
For now will fix the test with `acceptContactRequest` flag if AC notifications should be checked.

PR #3979, opened 2023-08-31
only sync decisions/state of AC notifications

### LINKED ISSUE #3949: Don't sync AC notifications, only decisions/state
# Problem

It seems that currently we fully sync AC notifications.
We should only sync decisions/state but not the actual notification.

https://github.com/status-im/status-go/blob/dd3e408a4e5ca412ecd6ed69a14c1fc8e456c038/protocol/messenger_handler.go#L2281

# Acceptance Criteria

- AC notifications are always generated locally and never synced.
- AC notifications state is synced
- AC notifications decision is synced

Related bug: https://github.com/status-im/status-go/issues/3801

qfrank:
Closes #3949


protocol/messenger.go
```diff
@@ -4102,16 +4102,11 @@ func (m *Messenger) markAllRead(chatID string, clock uint64, shouldBeSynced bool
...
 }
 
-func (m *Messenger) MarkAllRead(chatID string) error {
-	notifications, err := m.persistence.DismissAllActivityCenterNotificationsFromChatID(chatID, m.getCurrentTimeInMillis())
+func (m *Messenger) MarkAllRead(ctx context.Context, chatID string) error {
+	_, err := m.DismissAllActivityCenterNotificationsFromChatID(ctx, chatID, m.GetCurrentTimeInMillis())
 	if err != nil {
 		return err
 	}
-	err = m.syncActivityCenterNotifications(notifications)
```
saledjenic, line 4123: @qfrank in case of an activity like this one, we need to sync it, so marking all or particular notifications as read/unread manually is an action that should be synced.

protocol/messenger.go
```diff
@@ -4156,12 +4151,6 @@ func (m *Messenger) MarkAllReadInCommunity(communityID string) ([]string, error)
 	if err != nil {
 		return chatIDs, err
 	}
-
-	err = m.syncActivityCenterNotifications(notifications)
```
saledjenic, line 4173: @qfrank the same here.

protocol/messenger_communities.go
```diff
@@ -1724,16 +1719,17 @@ func (m *Messenger) CheckAndDeletePendingRequestToJoinCommunity(sendResponse boo
...
+					updatedAt := m.GetCurrentTimeInMillis()
+					notification.UpdatedAt = updatedAt
 					err = m.persistence.DeleteActivityCenterNotificationByID(types.FromHex(requestToJoin.ID.String()), notification.UpdatedAt)
 					if err != nil {
 						m.logger.Error("failed to delete notification from activity center", zap.Error(err))
 						return nil, err
 					}
 					notification.Deleted = true
-					err = m.syncActivityCenterNotifications([]*ActivityCenterNotification{notification})
+					err = m.syncActivityCenterDeletedByIDs(ctx, []types.HexBytes{notification.ID}, updatedAt)
```
saledjenic: @qfrank I see this is not your code, it was already here, but I think this is wrong if I understand the context correctly. If a pending request to join a community gets deleted, we need to send that info and once it's received on paired devices will remove the pending request and delete the related notification from the activity center.

But, in my opinion, we don't need to delete a related activity center notification at all. Not sure why we're doing that. So we're adding a notification once a request is made, if it gets deleted in the meantime a click on that notification should open a popup saying "The request doesn't exist anymore" or something like that.

protocol/messenger_contacts.go
```diff
@@ -847,14 +847,14 @@ func (m *Messenger) blockContact(response *MessengerResponse, contactID string,
 		}
 
 		// We remove anything that's related to this contact request
-		notifications, err := m.persistence.DeleteChatContactRequestActivityCenterNotifications(contact.ID, m.getCurrentTimeInMillis())
+		updatedAt := m.GetCurrentTimeInMillis()
+		notifications, err := m.persistence.DeleteChatContactRequestActivityCenterNotifications(contact.ID, updatedAt)
```
saledjenic, line 852: @qfrank similar to above, we don't need to delete a previous notification, but simply open a popup of that contact displaying it in the blocked state.

saledjenic (commented):
@qfrank in general code looks good, I don't have anything to say about changes, my concerns are more use-case oriented in terms of notifications added to the activity center.

We discussed and agreed that we need to sync an event and that event, once received on paired devices, will add a notification to paired devices. That's ok.

But there are some events we need to sync for activity center notifications like marking notifications read/unread, that's something that should be synced and the same across all paired devices.

Also updating/deleting notifications that are already added looks odd to me. For example (T1 < T2 < T3 < .... < Tn) you added me as a friend, I see a notification at the T1 `Frank added me as a friend` then I got some new notifications after that at time T2, T3, and then at the moment T4 when you accept my request I should get a new request at the time T4 saying `Frank accepted my request`, but not updating the request received at time T1. 

Similarly, if a request got deleted, I should still be able to see added notification to the activity center, received notifications should not be deleted.

protocol/messenger.go
qfrank: emm.. did you check `DismissAllActivityCenterNotificationsFromChatID`? i moved the sync action into `DismissAllActivityCenterNotificationsFromChatID`, and we don't sync the whole notification right now, just the decision `SyncActivityCenterDismissed` @saledjenic

protocol/messenger_communities.go
qfrank: not sure about this, this PR just want to implement what we've agreed on "we should be syncing decisions/states and not actual notifications" , what's your opinion? @ilmotta cc @cammellos

protocol/messenger_contacts.go
qfrank: I found this comment `// We remove anything that's related to this contact request` comes from @MishkaRogachev , WDYT about sale's comment? @MishkaRogachev

qfrank:
> But there are some events we need to sync for activity center notifications like marking notifications read/unread, that's something that should be synced and the same across all paired devices.

I think i implemented these in: https://github.com/status-im/status-go/blob/4f3f7e63c9e63a550f6cfc6cc9ac709d8f3d71c7/protocol/messenger_activity_center.go#L196 
and https://github.com/status-im/status-go/blob/4f3f7e63c9e63a550f6cfc6cc9ac709d8f3d71c7/protocol/messenger_activity_center.go#L150

> Similarly, if a request got deleted, I should still be able to see added notification to the activity center, received notifications should not be deleted.

Maybe @ilmotta has something to say about this? because i saw it seems it was you introduced `DeleteActivityCenterNotifications` 🙂

> Also updating/deleting notifications that are already added looks odd to me. For example (T1 < T2 < T3 < .... < Tn) you added me as a friend, I see a notification at the T1 `Frank added me as a friend` then I got some new notifications after that at time T2, T3, and then at the moment T4 when you accept my request I should get a new request at the time T4 saying `Frank accepted my request`, but not updating the request received at time T1.

Could you explain this more , not sure what you said, thank you @saledjenic

saledjenic:
@qfrank yes, ofc, will try to explain better now in short, at least 2 things are something I am not sure about that we're handing in the right way.

1. When any notification is added to the activity center we should not update it after that, but add a new notification describing what's happened. For example if you added me as a friend and that's added to the AC as a new notification `Frank added me as a friend` then you decide to cancel that friend request before I confirm it, I should still be able to see that notification, we don't need to delete it on my side. But if I click on it, then I should not be able to confirm the request (the confirmation button should be disabled). 

That's the pattern we should use for all notifications, adding new notifications for each event, not updating existing ones.

2. There is a feature in AC to mark all notifications as read/unread and to mark a particular notification as read/unread. That action should be synced across all paired devices and in case of marking all read/unread, that's an easy thing to do. But in the case of marking a particular notification that's not as easy as sending a single sync message saying a notification with id `X` is marked as read/unread, cause now notifications are generated locally and the same notification do not have the same id on all paired devices.

protocol/messenger.go
saledjenic: In this particular case all is good cause this is called when the user is marking all notifications coming from the chat with chatID as read and that's ok, but just wanted to say about the general approach, but it might be that I was not clear enough, please read this, more detailed [comment](https://github.com/status-im/status-go/pull/3979#issuecomment-1702403086).

protocol/activity_center_persistence.go
```diff
@@ -1027,9 +1024,29 @@ func (db sqlitePersistence) MarkActivityCenterNotificationsUnread(ids []types.He
 		return nil, err
 	}
 
+	if len(notifications) == 0 {
+		return notifications, nil
+	}
+
+	tx, err2 := db.db.BeginTx(context.Background(), &sql.TxOptions{})
```
ilmotta: I think ideally, the whole `MarkActivityCenterNotificationsUnread` should run inside the same transaction, so that the query which finds notifications is consistent with the sql updates.

protocol/protobuf/application_metadata_message.proto
```diff
@@ -77,13 +77,15 @@ message ApplicationMetadataMessage {
     SYNC_KEYPAIR = 62;
     SYNC_SOCIAL_LINKS = 63;
     SYNC_ENS_USERNAME_DETAIL = 64;
-    SYNC_ACTIVITY_CENTER_NOTIFICATIONS = 65;
+    reserved 65; // don't use this value since we want keep backwards compatibility
```
ilmotta: I see the comment as unnecessary, `reserved` is self explanatory in protobufs.

ilmotta: Besides the number, it's also a good practice to reserve the name.

ilmotta (commented):
@qfrank, PR seems reasonable and has simplified some things. I prefer not to approve it because it's not clear to me how the system should behave in some cases.

I agree with the general idea shared by @saledjenic that we shouldn't be deleting notifications automatically.

In my **very personal** view of how the system should behave, I think notifications should be accrued, in an almost immutable fashion (with the exception of some state changes, like read/unread). Users should never see notifications disappearing.

A real example where I think the in-place edits we do to notifications are bad for users:

- Alice and Bob are friends.
- Alice owns a closed community named Acme.
- Bob sends a request to join Acme.
- Bob sees in his AC the notification "Request pending. You requested to join
  Acme."
- Alice accepts the join request.
- Bob sees his previous notification be mutated to be "Request accepted. Now you are a member of Acme." Even the timestamp changes, so for Bob, the original notification disappeared and a new one was generated. If behind the scenes we mutated the notification it doesn't matter to him, because in practice, he lost his previous notification.

And things get worse. If now Bob leaves the community and requests to join again, he will see his previous notification "Request accepted. Now you are a member of Acme" be replaced by "Request pending. You requested to join Acme."

So this flow is bad for users, they can't track their actions in the Activity Center because they lose information over time.

One thing I've noticed about our notification system is that we keep getting back to fix things because the business rules are too complicated.

I think @MishkaRogachev has coded more stuff in the AC than myself. I added him as a reviewer in case it helps clarify some points.

cammellos:
I think it would be useful to distinguish functional requirements from technical implementation, from the conversation above is not clear whether we are talking about one or the other, and looks like the two are intersecting.

### Functional requirements

https://www.figma.com/file/eDfxTa9IoaCMUy5cLTp0ys/Shell-for-Mobile?node-id=4537%3A515620&mode=dev

Here we can see captures some rules and we need to clarify what the user sees. We can see that notifications are updated (visually, not in terms of implementation) when their state change (I send a request to A, A accepts, visually the notification changes on both devices), no new notification is shown. Note that this can be implemented either ways.

The case that @saledjenic is highlighting:

1) A sends a notification to B
2) A retracts the notification before B can answer it

Is not captured I believe in the designs. So that's the first thing we want to clarify. 
@MishkaRogachev might have done already part of this work and might know the answer, probably because of the complexity, it would be good to capture this somewhere.

### Technical implementation

We can only answer so much questions without fully understanding the functional requirements, so I would first clarify those so we can make a better informed decision on how we want it to be implemented.


> There is a feature in AC to mark all notifications as read/unread and to mark a particular notification as read/unread. That action should be synced across all paired devices and in case of marking all read/unread, that's an easy thing to do. But in the case of marking a particular notification that's not as easy as sending a single sync message saying a notification with id X is marked as read/unread, cause now notifications are generated locally and the same notification do not have the same id on all paired devices.

@saledjenic this should be an exception, the same notification should have the same id on multiple devices, if generated from the same event, I can't see a case where that should not hold, but maybe I am missing something, so syncing the id of the notification should be fine in theory.

protocol/messenger_contacts.go
MishkaRogachev: I checked with @John-44 , in case of blocking contact, we have to delete all notifications associated with this contact.

saledjenic:
> @saledjenic this should be an exception, the same notification should have the same id on multiple devices, if generated from the same event, I can't see a case where that should not hold, but maybe I am missing something, so syncing the id of the notification should be fine in theory.

I am not involved in activity center notifications at all, but I've raised that as a potential concern cause I saw in multiple places setting id like this `ID:        types.FromHex(uuid.New().String()),` for `ActivityCenterNotification` type. That definitely means the notification id is local to the device and there is no way to sync read/unread state for such notification.

Also I see id set like this `ID:                        types.FromHex(vr.ID),` for `ActivityCenterNotification` type, where `vr` refers to `verification.Request` type. In this case, if we decide to not update the contact request notification but generate a new notification whenever the state between contacts changes then we will need to have additional argument not only notification ID (which is verification request id) when we want to sync the read/unread state of such notification.

MishkaRogachev:
@qfrank, Thanks for this PR, the code looks good!

I apologize for the long reply. I share the opinion that the state (Deleted, Seen and UpdatedAt) should be synchronized between devices, but I see one problem: id for some notifications will be different between devices and at least these are contact requests (other types need to be checked, for most everything should be fine already). As far as I remember there are difficulties with contact synchronization, where we generate AC notification and set a new id [here](https://github.com/status-im/status-go/blob/develop/protocol/messenger_handler.go#L467-L468). @cammellos wdyt to finally get rid of default contact request id here and make AC notifications for contact requests having same on all synced devices?

MishkaRogachev:
> > @saledjenic this should be an exception, the same notification should have the same id on multiple devices, if generated from the same event, I can't see a case where that should not hold, but maybe I am missing something, so syncing the id of the notification should be fine in theory.

Yes,  that problem affects not only contact requests, as i mentioned before. We need to provide how to generate AC notification IDs so that we can replicate this on the synchronized device

cammellos:
> I apologize for the long reply. I share the opinion that the state (Deleted, Seen and UpdatedAt) should be synchronized between devices, but I see one problem: id for some notifications will be different between devices and at least these are contact requests (other types need to be checked, for most everything should be fine already). As far as I remember there are difficulties with contact synchronization, where we generate AC notification and set a new id [here](https://github.com/status-im/status-go/blob/develop/protocol/messenger_handler.go#L467-L468). @cammellos wdyt to finally get rid of default contact request id here and make AC notifications for contact requests having same on all synced devices?

@MishkaRogachev 

We can't completely get rid of the default contact request. 

The case we need to handle is (one of the many):

1) Alice & Bob are contacts
2) Alice restores her device, backup for whatever reason is not retrieved, all the state is lost
3) Bob sends a message to Alice

The message in this case would be ignored effectively, since Alice is not in contact with Bob. Bob is not able to send a contact request to Alice since they believe they are mutual contacts, therefore no communication can take place until Alice explicitly tries to contact Bob.
Default contact request are used in this case so that Alice infers that Bob was in her contacts. It could be something else of course, or the shape we want, but a graceful recovery mechanism in my opinion is required, otherwise we will get into a place where communication can't happen (this is why we never wanted to implement mutual contacts in the first place and throughout the history of the product, it's just a bad idea given our technology).

There are many of these examples, especially when syncing is taken into consideration, as there's lot of timing issues when multiple devices are involved and no order of messages is guaranteed.

We can always change the format of the default contact request id etc, of course, though not sure that's the issue, since it's predictable and the same on all devices:

```
func defaultContactRequestID(contactID string) string {
	return "0x" + types.Bytes2Hex(append(types.Hex2Bytes(contactID), 0x20))
}
```

Are you sure you are referring to that one when you say?

" and make AC notifications for contact requests having same on all synced devices"

MishkaRogachev:
> We can't completely get rid of the default contact request.
> 

I agree, I meant that if we rely on id's to sync notifications, we need to ensure that the id's for contact requests are the same. If we didn't have a requirement to show the history of requests in notifications and for each contact we would always have one notification showing the current state of the last contact request, we could always use defaultContactID and a lot of problems would go away. On the other hand, do we really need to sync the state of such notifications?

cammellos:
> I agree, I meant that if we rely on id's to sync notifications, we need to ensure that the id's for contact requests are the same. If we didn't have a requirement to show the history of requests in notifications and for each contact we would always have one notification showing the current state of the last contact request, we could always use defaultContactID and a lot of problems would go away. On the other hand, do we really need to sync the state of such notifications?

Either we sync the state of the notification, or the state of the contact (or both), but in that case, default contact request should be shown as accepted if the contacts are mutual, so either way it should work.

qfrank:
Thanks to the awesome comments everyone did, i got an agreement that "do exactly the same as Discord e.g. the user doesn't see the retracted notification at all as if it didn't happen." from PM  and design team. @saledjenic @ilmotta 

Currently, for the 3 of `Acceptance Criteria` #3949 mentioned, this PR almost fit the requirement, but i found 
`AC notifications state` is not processed well i think, e.g. `MarkActivityCenterNotificationsUnread` won't update `state` , i will try to fix it in this PR today. 

cc @cammellos @MishkaRogachev @igor-sirotin

protocol/activity_center_persistence.go
```diff
@@ -980,32 +1085,89 @@ func (db sqlitePersistence) AcceptActivityCenterNotificationsForInvitesFromUser(
...
-	return notifications, nil
+	return notifications, updateActivityCenterState(tx, updatedAt)
 }
 
 func (db sqlitePersistence) MarkAllActivityCenterNotificationsRead(updatedAt uint64) error {
-	_, err := db.db.Exec(`UPDATE activity_center_notifications SET read = 1, updated_at = ? WHERE NOT read AND NOT deleted`, updatedAt)
+	var tx *sql.Tx
+	var err error
+
+	tx, err = db.db.BeginTx(context.Background(), &sql.TxOptions{})
```
igor-sirotin: Can't we use `:=` here?

```suggestion
	tx, err := db.db.BeginTx(context.Background(), &sql.TxOptions{})
```

protocol/messenger.go
igor-sirotin: @qfrank It seems to me too that this is wrong here now. 
`Mark all as read` changes the state of the notification and this fact should be synced. And I don't see any syncing in `DismissAllActivityCenterNotificationsFromChatID`.

Or is it some new UI approach you was talking about?

protocol/messenger.go
igor-sirotin: same here. unresolving just to point out.

protocol/messenger_contacts.go
```diff
@@ -887,14 +887,9 @@ func (m *Messenger) BlockContact(contactID string, fromSyncing bool) (*Messenger
 	//		 This should be considered separately, I'm not sure if that's safe.
 	//		 https://github.com/status-im/status-go/issues/3720
 	if !fromSyncing {
-		notifications, err := m.persistence.DismissAllActivityCenterNotificationsFromUser(contactID, m.getCurrentTimeInMillis())
-		if err != nil {
-			return nil, err
-		}
-
-		err = m.syncActivityCenterNotifications(notifications)
```
igor-sirotin, line 896: I don't remember this very good, but I think `fromSyncing` flag is not needed anymore if we remove `syncActivityCenterNotifications` from here.  

Ping me to discuss this if needed.

igor-sirotin: Also, shouldn't `dismiss` action be synced from here?

protocol/messenger_contacts.go
```diff
@@ -904,30 +899,23 @@ func (m *Messenger) BlockContact(contactID string, fromSyncing bool) (*Messenger
...
 	}
 
-	response, err = m.DeclineAllPendingGroupInvitesFromUser(response, contactID)
+	response, err = m.DeclineAllPendingGroupInvitesFromUser(ctx, response, contactID)
 	if err != nil {
 		return nil, err
 	}
 
-	notifications, err := m.persistence.DismissAllActivityCenterNotificationsFromUser(contactID, m.getCurrentTimeInMillis())
+	_, err = m.DismissAllActivityCenterNotificationsFromUser(ctx, contactID, m.GetCurrentTimeInMillis())
```
igor-sirotin: same here, should maybe sync `dismiss`?

igor-sirotin: I see there're more places with `Dismiss...` below. This applies to all cases.

protocol/messenger_group_chat.go
```diff
@@ -806,7 +799,7 @@ func (m *Messenger) DeclineAllPendingGroupInvitesFromUser(response *MessengerRes
 	}
 
 	// Decline group invites from activity center notifications
-	notifications, err := m.persistence.AcceptActivityCenterNotificationsForInvitesFromUser(userPublicKey, m.getCurrentTimeInMillis())
+	notifications, err := m.AcceptActivityCenterNotificationsForInvitesFromUser(ctx, userPublicKey, m.GetCurrentTimeInMillis())
```
igor-sirotin, line 802: Same here, shouldn't we sync accept/decline result?

protocol/protobuf/application_metadata_message.proto
```diff
@@ -77,13 +77,16 @@ message ApplicationMetadataMessage {
     SYNC_KEYPAIR = 62;
     SYNC_SOCIAL_LINKS = 63;
     SYNC_ENS_USERNAME_DETAIL = 64;
-    SYNC_ACTIVITY_CENTER_NOTIFICATIONS = 65;
+    reserved 65;
```
igor-sirotin, line 81: ❤️

protocol/protobuf/pairing.proto
```diff
@@ -177,59 +177,19 @@ message SyncActivityCenterDismissed {
   repeated bytes ids = 2;
 }
 
-message SyncActivityCenterNotificationState {
-  uint64 updated_at = 1;
-  bool has_seen = 2;
+message SyncActivityCenterDeleted {
+  uint64 clock = 1;
+  repeated bytes ids = 3;
```
igor-sirotin: `= 2`?

protocol/messenger_sync_activity_center_test.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func TestMessengerSyncActivityCenter(t *testing.T) {
+	suite.Run(t, new(MessengerSyncActivityCenterSuite))
+}
+
+func (s *MessengerSyncActivityCenterSuite) TestSyncUnRead() {
+	theirMessenger, err := newMessengerWithKey(s.shh, s.privateKey, s.logger, nil)
+	s.NoError(err)
+	PairDevices(&s.Suite, theirMessenger, s.m)
```
igor-sirotin, line 24: I know it's no obvious but it should be manually done in both directions.
It won't probably affect the test, but better to fix to catch any loop syncing.
```suggestion
	PairDevices(&s.Suite, theirMessenger, s.m)
	PairDevices(&s.Suite, s.m, theirMessenger)
```

protocol/messenger_sync_activity_center_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+	_, err = WaitOnMessengerResponse(theirMessenger, func(r *MessengerResponse) bool {
+		return r.ActivityCenterState() != nil
+	}, "activity center notification state not received")
+	s.NoError(err)
+	notificationByID, err = theirMessenger.persistence.GetActivityCenterNotificationByID(id)
+	s.NoError(err)
+	s.False(notificationByID.Read)
+
+	s.NoError(theirMessenger.Shutdown())
```
igor-sirotin: would be better to have it `defer` in the beginning in case the test failed earlier?

protocol/messenger_sync_activity_center_test.go
```diff
@@ -0,0 +1,121 @@
...
+	_, err = WaitOnMessengerResponse(theirMessenger, func(r *MessengerResponse) bool {
+		return r.ActivityCenterState() != nil
+	}, "activity center notification state not received")
+	s.NoError(err)
+	notificationByID, err = theirMessenger.persistence.GetActivityCenterNotificationByID(id)
+	s.NoError(err)
+	s.True(notificationByID.Deleted)
+
+	s.NoError(theirMessenger.Shutdown())
+}
```
igor-sirotin, line 92: I think there're more actions on notifications available?
Would be cool to have tests for them as well. Probably our comments above about missing syncing will be reproduced.

igor-sirotin (commented):
Thank you, @qfrank, great job!
I left some doubts in comments, not sure that everything that's needed is synced.

_____

Also, I think this PR should fix this bug: https://github.com/status-im/status-go/issues/3801
There're 2 warnings in the code pointing to the issue. Maybe worth checking this right here?
https://github.com/status-im/status-go/blob/287ae899408bc2fe13ae5158355c631b442ba7d6/protocol/messenger_contact_requests_test.go#L255
https://github.com/status-im/status-go/blob/287ae899408bc2fe13ae5158355c631b442ba7d6/protocol/messenger_contact_requests_test.go#L291

protocol/activity_center_persistence.go
```diff
@@ -594,6 +594,9 @@ func (db sqlitePersistence) HasPendingNotificationsForChat(chatID string) (bool,
 }
 
 func (db sqlitePersistence) GetActivityCenterNotificationsByID(ids []types.HexBytes) ([]*ActivityCenterNotification, error) {
+	if len(ids) == 0 {
+		return nil, nil
```
MishkaRogachev: Suggestion: return an empty list instead of `nil` to avoid extra checks

protocol/messenger.go
```diff
@@ -4117,16 +4117,11 @@ func (m *Messenger) markAllRead(chatID string, clock uint64, shouldBeSynced bool
 	return m.persistence.SaveChats([]*Chat{chat})
 }
 
-func (m *Messenger) MarkAllRead(chatID string) error {
-	notifications, err := m.persistence.DismissAllActivityCenterNotificationsFromChatID(chatID, m.getCurrentTimeInMillis())
+func (m *Messenger) MarkAllRead(ctx context.Context, chatID string) error {
+	_, err := m.DismissAllActivityCenterNotificationsFromChatID(ctx, chatID, m.GetCurrentTimeInMillis())
```
MishkaRogachev, line 4119: I case we change the state of some notifications, i think we should add those notifications to the response for the client

protocol/messenger_contacts.go
MishkaRogachev: It's better to add dismissed notifications to the response here, too.

MishkaRogachev (commented):
Great work, thanks @qfrank!
There are a few minor question, it in general it's a great PR! I will give an approval after Igor's questions would be resolved :)

protocol/messenger.go
qfrank: It's here: 
https://github.com/status-im/status-go/blob/95048404de53733dd1772b64d628c71a65434532/protocol/messenger_activity_center.go#L452

@igor-sirotin

protocol/messenger.go
qfrank: it's here:
https://github.com/status-im/status-go/blob/95048404de53733dd1772b64d628c71a65434532/protocol/messenger_activity_center.go#L436

@igor-sirotin

protocol/messenger_group_chat.go
igor-sirotin: Ok, @qfrank pointed me to that it's synced in `AcceptActivityCenterNotificationsForInvitesFromUser`. No idea how I missed that 🤦

protocol/messenger_contacts.go
qfrank: > It's better to add dismissed notifications to the response here, too.

maybe separate PR would be better, this PR mainly focused on #3949 🙂

protocol/messenger.go
qfrank: Hi @MishkaRogachev ,  I can update it in a separate PR since it may need break changes and this PR already took much time, WDYT?

igor-sirotin (approved):
Thank you very much, nice job 👍

MishkaRogachev (approved):
Great work!

igor-sirotin:
🎉

Issue #3835, opened 2023-08-01
Wallet Merge balance history and transfer process


alaibe:
# Problem

Check if possible to merge balance history and transfer processes

@IvanBelyakoff @stefandunca tagging you in order to help

IvanBelyakoff:
AFAIK we store balance history for balance chart and for that purpose we care about transfers that change ETH or token balances, store the timestamp, block and balance. @stefandunca implemented this and used `balance_history` table for that. For transfers we currently use two strategies - `OnDemanFetchStrategy` (mobile v1, that is obsolete now) - the one that was there since the beginning, and a new one - `SequentialFetchStrategy` (used on Desktop) - where we load all transfers sequentially from latest block till history start. `OnDemandFetchStrategy` stores pretty much the same info as balance history fetcher - block ranges and balance in `blocks_ranges` table (`block_dao.go: ProcessBlock`). So we might want to use a single table for that. `SequentialFetchStrategy` does not store block ranges in a similar way, so nothing to unify here in terms of db columns imo, but while looking for ETH transfers, a downloader checks ranges and finds blocks with ETH transfers (`transfers/concurrent.go:findBlocksWithEthTransfers`) and balances and those are the data, that is the same imo that we store in balance_history, so I believe that would be nice to reuse this data  for balance history.
As for tokens, downloader also uses `getLogs` which I believe balance history fetcher uses as well to get the same data, so we should reuse the requests as well.

stefandunca:
> AFAIK we store balance history for balance chart and for that purpose we care about transfers that change ETH or token balances, store the timestamp, block and balance.

You summarized well. In addition, it tries to estimate and fetch regular intervals to have a uniform sampling point. Also samples are snapped between intervals to reuse requests

> OnDemandFetchStrategy stores pretty much the same info as balance history fetcher - block ranges and balance in blocks_ranges table

If we do cover the needed granularity by the balance history, for the supported intervals, it sounds like this strategy could be used for balance history also.

> As for tokens, downloader also uses `getLogs` which I believe balance history fetcher uses as well to get the same data, so we should reuse the requests as well.

Balance history only uses `HeaderByNumber` to find the graph entry block timestamp and `BalanceAt` for token values. So no logs.

rasom:
@stefandunca @IvanBelyakoff @alaibe @John-44 
To summarise this note once again https://notes.status.im/TSPntaw0SCeV44Z4ENM1sQ?view#Merging-of-balancetransfers-history-algos 

[...]

PR #3983, opened 2023-09-04
feat(wallet): move balance cache to a common place

IvanBelyakoff:
And make it a parameter to share between transfers and balance history in an upcoming commit.
Had to refactor its interface for that reason.

- `BalanceCache` interface is renamed to `BalanceCacher` which requires `BalanceReader` to fetch balance data from chain. Added an additional `Cache()` method to retrieve `CacheIface` interface. It is used whenever we should get data  whether it is already cached or needs fetching from chain.
- `CacheIface` interface is used to get and add balance and nonce to/from cache (forced by linter to remove `Balance` prefix for the interface name, since the type is in `balance` package) . This is for operations on cache itself without fetching anything from chain.
- `Cache` type implements both above interfaces (forced by linter to remove `Balance` prefix for the type name)

Updates [#3835](https://github.com/status-im/status-go/issues/3835)

### DIFF
--- a/services/wallet/balance/balance_cache.go
+++ b/services/wallet/balance/balance_cache.go
@@ -1,4 +1,4 @@
-package transfer
+package balance
 
 import (
 	"context"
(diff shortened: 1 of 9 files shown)


stefandunca (approved):
👍🏼

PR #3984, opened 2023-09-04
fix(wallet-tests): possible fix for TestPendingTxTracker_MultipleClients

IvanBelyakoff:
This test occasionally fails on CI (saw it multiple times):

```
[2023-09-04T08:13:19.059Z] === RUN   TestPendingTxTracker_MultipleClients
[2023-09-04T08:13:19.888Z]     pendingtxtracker_test.go:293: timeout waiting for event
[2023-09-04T08:13:19.888Z] --- FAIL: TestPendingTxTracker_MultipleClients (1.00s)
[2023-09-04T08:13:19.888Z] FAIL
```
I can't reproduce it locally, but tracking it down, I think I found a possible root cause - see comments.

### DIFF
--- a/transactions/pendingtxtracker_test.go
+++ b/transactions/pendingtxtracker_test.go
@@ -265,17 +265,29 @@ func TestPendingTxTracker_MultipleClients(t *testing.T) {
 		(*res)["blockNumber"] = transactionSuccessStatus
 	})
 
-	for i := range txs {
-		err := m.TrackPendingTransaction(txs[i].ChainID, txs[i].Hash, txs[i].From, txs[i].Type, true)
-		require.NoError(t, err)
-	}
+	// If we call TrackPendingTransaction immediately, there is a chance that some events
+	// will be emitted before we reach select, so occasionally select fails on timeout.
+	go func() {
+		for i := range txs {
+			err := m.TrackPendingTransaction(txs[i].ChainID, txs[i].Hash, txs[i].From, txs[i].Type, true)
+			require.NoError(t, err)
+		}
+	}()
 
 	eventChan := make(chan walletevent.Event)
 	sub := eventFeed.Subscribe(eventChan)
 
-	err := m.Start()
-	require.NoError(t, err)
+	// events caused by addPending
+	for i := 0; i < 2; i++ {
+		select {
+		case we := <-eventChan:
+			require.Equal(t, EventPendingTransactionUpdate, we.Type)
+		case <-time.After(1 * time.Second):
+			t.Fatal("timeout waiting for event")
+		}
+	}
 
+	// events caused by tx deletion on fetch
 	for i := 0; i < 2; i++ {
 		for j := 0; j < 2; j++ {
 			select {
@@ -295,7 +307,7 @@ func TestPendingTxTracker_MultipleClients(t *testing.T) {
 		}
 	}
 
-	err = m.Stop()
+	err := m.Stop()
 	require.NoError(t, err)
 
 	res, err := m.GetAllPending()


transactions/pendingtxtracker_test.go
```diff
@@ -265,17 +265,29 @@ func TestPendingTxTracker_MultipleClients(t *testing.T) {
 		(*res)["blockNumber"] = transactionSuccessStatus
 	})
 
-	for i := range txs {
-		err := m.TrackPendingTransaction(txs[i].ChainID, txs[i].Hash, txs[i].From, txs[i].Type, true)
-		require.NoError(t, err)
-	}
+	// If we call TrackPendingTransaction immediately, there is a chance that some events
+	// will be emitted before we reach select, so occasionally select fails on timeout.
```
stefandunca, line 269: The subscribe is done after `TrackingPendingTransaction` to avoid processing the `addPending` events. The fact that we have a timeout means there are fewer events processed than expected.

transactions/pendingtxtracker_test.go
```diff
@@ -265,17 +265,29 @@ func TestPendingTxTracker_MultipleClients(t *testing.T) {
...
+	go func() {
```
stefandunca, line 270: Starting to do it async is not a good idea for a test. We miss the control flow and increase the chance of mixing the update with `EventPendingTransactionUpdate` with `EventPendingTransactionStatusChanged`.

transactions/pendingtxtracker_test.go
```diff
@@ -265,17 +265,29 @@ func TestPendingTxTracker_MultipleClients(t *testing.T) {
...
+		}
+	}()
 
 	eventChan := make(chan walletevent.Event)
 	sub := eventFeed.Subscribe(eventChan)
 
-	err := m.Start()
-	require.NoError(t, err)
+	// events caused by addPending
+	for i := 0; i < 2; i++ {
```
stefandunca, line 281: It makes sense to have this only if we were to have emitted more events than we processed. As stated in the first comment we are on the other side where we emit fewer events than expected.

transactions/pendingtxtracker_test.go
IvanBelyakoff: Yes, right. But from my understanding it is not guaranteed that `fetchTransactions` is not executed before `select` and that will cause not enough events

transactions/pendingtxtracker_test.go
IvanBelyakoff: I don't agree with the conclusion. We receive less events than expected, not emit less than expected. So this way I ensure I catch all the events that are expected. And I believe all should be received now.
I mean that some emitted events were not received for some reason, and the reason I think was that they were emitted before they could be received

transactions/pendingtxtracker_test.go
stefandunca: Sorry, I just now realized the base for this PR is not the latest master. I extended the test to validate the implementation expectations in the latest master. Let's monitor the last change and see if that was enough.

> ... is not guaranteed that fetchTransactions is not executed before select and that will cause not enough events

You were right, we might have missed events because the subscription could have happened before `fetchTransactions` started. I fixed this in the latest master by subscribing before `TrackPendingTransaction`.

transactions/pendingtxtracker_test.go
stefandunca: Now I understand better your changes. However, the last improvements I added to the master branch to subscribe before calling the APIs and increase the channel buffer accordingly is doing the same as you proposed in a different way.

IvanBelyakoff:
Abandoning it as stale

PR #3986, opened 2023-09-04
feat: add sharding information when importing/exporting a community

richard-ramos:
While dogfooding the sharding PR, i noticed that importing/exporting a community was missing the shard cluster/index and protected topic. Notice this is a breaking change due to the import/export community API modifications


services/ext/api.go
```diff
@@ -442,22 +442,13 @@ func (api *PublicAPI) SetCommunityShard(request *requests.SetCommunityShard) (*p
...
 // ImportCommunity imports a community with the given private key in hex
-func (api *PublicAPI) ImportCommunity(ctx context.Context, hexPrivateKey string) (*protocol.MessengerResponse, error) {
-	// Strip the 0x from the beginning
-	privateKey, err := crypto.HexToECDSA(hexPrivateKey[2:])
-	if err != nil {
-		return nil, err
-	}
-	return api.service.messenger.ImportCommunity(ctx, privateKey)
+func (api *PublicAPI) ImportCommunity(ctx context.Context, importData *communities.CommunityImportInfo) (*protocol.MessengerResponse, error) {
+	return api.service.messenger.ImportCommunity(ctx, importData)
```
ilmotta, line 477: @richard-ramos, I'm getting a type signature failure because `api.service.messenger.ImportCommunity` still expects a `ecdsa.PrivateKey`.

richard-ramos: Forgot to add a change to the commit. Sorry!

protocol/communities/manager.go
```diff
@@ -980,7 +980,13 @@ func (m *Manager) RemovePrivateKey(id types.HexBytes) (*Community, error) {
 	return community, nil
 }
 
-func (m *Manager) ExportCommunity(id types.HexBytes) (*ecdsa.PrivateKey, error) {
+type CommunityImportInfo struct {
+	Shard        *common.Shard   `json:"shard"`
+	TopicKey     *types.HexBytes `json:"protectedTopicPrivateKey,omitempty"`
```
ilmotta: `TopicKey` could be `json:"topicKey"` and `CommunityKey` could be `json:"communityKey"`? Or is there a reason to not apply this convention?

protocol/communities/manager.go
```diff
@@ -1007,19 +1041,26 @@ func (m *Manager) ImportCommunity(key *ecdsa.PrivateKey) (*Community, error) {
...
 		if err != nil {
 			return nil, err
 		}
 	} else {
-		community.config.PrivateKey = key
+		community.config.PrivateKey = communityKey
+		community.config.Shard = importData.Shard
+	}
+
+	err = m.transport.StorePubsubTopicKey(community.PubsubTopic(), protectedTopicKey)
```
ilmotta: Would it make sense here to check if `protectedTopicKey` is not nil first, before calling `StorePubsubTopicKey`? The reason is that `protectedTopicKey` is only set if `importData.TopicKey` is present, hence if `TopicKey` is nil we will call `StorePubsubTopicKey` with a nil private key.

PR #3990, opened 2023-09-05
chore: Use proper Gwei decimals

Cuteivist:
Use proper decimal points for Gwei, so that Gwei and Gwei converted to ETH is displayed correctly.

image

https://goerli.etherscan.io/tx/0xb3b9675a8ff138feb0654dcb3fcc52daac8e94731381244e1540038b9d904f0c


alaibe (approved):

stefandunca (approved):
Good workaround. Not sure if mixing token symbols with denominations is a good idea long term though. Maybe we should look into moving this to the presentation layer.

Issue #3991, opened 2023-09-05
Discord Import Tool: add `PinMessage.ThirdPartyMessage_ID` property


igor-sirotin:
The issue was originally faced here: https://github.com/status-im/status-desktop/issues/11854

# Problem

Currently when importing messages from Discord, we save `ThirdPartyID` property here:
https://github.com/status-im/status-go/blob/7531dbeaef7d067db832a1a175af500428e40c98/protocol/messenger_communities.go#L4353

This gives us ability to distinguish _imported_ messages when processing archives:
https://github.com/status-im/status-go/blob/7531dbeaef7d067db832a1a175af500428e40c98/protocol/messenger_handler.go#L1339-L1343

Unfortunately, we don't save this information for `PinMessage`s. Neither we save a `discord_message_id` to `pin_messages` table.
It's not a problem until we generate a `SYSTEM_MESSAGE_PINNED_MESSAGE`.
➡️  We don't mark this message as `seen`, although it's imported and should be `seen`:
https://github.com/status-im/status-go/blob/7531dbeaef7d067db832a1a175af500428e40c98/protocol/messenger_handler.go#L837-L850

# Implementation

Save `ThirdPartyID` in `pinMessagesToWakuMessages` function:
https://github.com/status-im/status-go/blob/7531dbeaef7d067db832a1a175af500428e40c98/protocol/messenger_communities.go#L4292

PR #3992, opened 2023-09-05
Hotfix: mark `Chat` and `Pin` archive messages as seen

igor-sirotin:
Related: https://github.com/status-im/status-desktop/issues/11854.

1. Add option to force `seen` property for generated `SYTEM_MESSAGE_PIN_MESSAGE`
https://github.com/status-im/status-go/blob/7531dbeaef7d067db832a1a175af500428e40c98/protocol/messenger_handler.go#L852-L854

2. Apply workaround: Mark all `ChatMessage` and `PinMessage` from archives as `seen`.
Currently we only mark "imported" messages as seen. But this doesn't cover imported `PinMessage`s and therefore wasn't possible to mark generated `SYTEM_MESSAGE_PIN_MESSAGE` as `Seen`. 
    - We can remove this workaround when https://github.com/status-im/status-desktop/issues/10911 is implemented.
    - More details here: https://github.com/status-im/status-go/issues/3991

3. Add processing of `PinMessage` when handling imported messages.
I could introduce it together with [#3991](https://github.com/status-im/status-go/issues/3991), but it totally makes sense here anyway. And adds backward compatibility for the future.
https://github.com/status-im/status-go/blob/7531dbeaef7d067db832a1a175af500428e40c98/protocol/messenger.go#L3422-L3423

4. Subscribe to spectated communities.
Without it's easy to get into "infinite chat loading animation" in desktop app. 
https://github.com/status-im/status-go/blob/7531dbeaef7d067db832a1a175af500428e40c98/protocol/messenger_communities.go#L2584

### DIFF
--- a/_assets/generate_handlers/generate_handlers.go
+++ b/_assets/generate_handlers/generate_handlers.go
@@ -19,11 +19,12 @@ type EnumType struct {
 
 // MethodInfo holds information about a method
 type MethodInfo struct {
-	ProtobufName string
-	MethodName   string
-	EnumValue    string
-	ProcessRaw   bool
-	SyncMessage  bool
+	ProtobufName   string
+	MethodName     string
+	EnumValue      string
+	ProcessRaw     bool
+	SyncMessage    bool
+	FromArchiveArg bool
 }
 
 func main() {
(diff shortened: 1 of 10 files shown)


cammellos (approved):

MishkaRogachev (approved):
LGTM!

PR #3995, opened 2023-09-06
fix(wallet): added ChainID to balance cache, as now it is shared between

IvanBelyakoff:
services and contains balances for all addresses and chains.
Made rpc chain client return ChainID property on `NetworkID()` method

Updates #3835

### DIFF
--- a/services/wallet/balance/balance_cache.go
+++ b/services/wallet/balance/balance_cache.go
@@ -2,6 +2,7 @@ package balance
 
 import (
 	"context"
+	"math"
 	"math/big"
 	"sort"
 	"sync"
@@ -23,6 +24,7 @@ type Reader interface {
 	NonceAt(ctx context.Context, account common.Address, blockNumber *big.Int) (uint64, error)
 	HeaderByNumber(ctx context.Context, number *big.Int) (*types.Header, error)
 	FullTransactionByBlockNumberAndIndex(ctx context.Context, blockNumber *big.Int, index uint) (*chain.FullTransaction, error)
+	NetworkID() uint64
 }
 
 // Cacher interface for caching balance to BalanceCache. Requires BalanceReader to fetch balance.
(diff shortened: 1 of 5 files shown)


services/wallet/balance/balance_cache.go
```diff
@@ -35,145 +36,176 @@ type Cacher interface {
...
-				maxBlock = key
+		for chainID, cache := range chainCache {
+			if len(cache) == 0 {
+				continue
 			}
-			if key < minBlock {
-				minBlock = key
+
+			var maxBlock uint64 = 0
+			var minBlock uint64 = 18446744073709551615
```
stefandunca: I wonder where is this constant coming from. Can it be sourced for example, from MaxUint64 from a reliable source?

services/wallet/balance/balance_cache.go
```diff
@@ -35,145 +36,176 @@ type Cacher interface {
...
-		var minBlock uint64 = 18446744073709551615
-		for key := range cache {
-			if key > maxBlock {
-				maxBlock = key
-			}
-			if key < minBlock {
-				minBlock = key
+		for chainID, cache := range chainCache {
+			var maxBlock uint64 = 0
+			var minBlock uint64 = 18446744073709551615
```
stefandunca: Share this magic number, maybe?

stefandunca (approved):
I left some questions inline in case you can clarify them.

saledjenic (approved):

services/wallet/balance/balance_cache.go
IvanBelyakoff: @stefandunca Good point! Fixed.

services/wallet/balance/balance_cache.go
IvanBelyakoff: Fixed.

PR #3998, opened 2023-09-07
chore: mark api deprecated

alaibe:


IvanBelyakoff (approved):
LGTM. But I have not checked every method

saledjenic (approved):

services/wallet/api.go
```diff
@@ -248,6 +260,7 @@ func (api *API) GetPendingTransactionsForIdentities(ctx context.Context, identit
 	return
 }
 
+// @deprecated
 // TODO - #11861: Remove this and replace with EventPendingTransactionStatusChanged event and Delete to confirm the transaction where it is needed
 func (api *API) WatchTransactionByChainID(ctx context.Context, chainID uint64, transactionHash common.Hash) (err error) {
```
stefandunca, line 265: This is still used by Nim `transaction.Service` for watching a transaction. The [#11861](https://github.com/status-im/status-desktop/issues/11861) is supposed to address this legacy API and replace it with the new event-based.

services/wallet/api.go
```diff
@@ -225,6 +236,7 @@ func (api *API) DeleteSavedAddress(ctx context.Context, address common.Address,
 	return err
 }
 
+// @deprecated
 func (api *API) GetPendingTransactions(ctx context.Context) ([]*transactions.PendingTransaction, error) {
```
stefandunca, line 240: @Cuteivist isn't this used to fetch details for pending transactions?

services/wallet/api.go
```diff
@@ -332,26 +348,31 @@ func (api *API) GetOpenseaAssetsByOwnerAndCollection(ctx context.Context, chainI
...
 }
 
+// @deprecated
 func (api *API) GetCollectiblesByOwnerAndCollectionWithCursor(ctx context.Context, chainID wcommon.ChainID, owner common.Address, collectionSlug string, cursor string, limit int) (*thirdparty.FullCollectibleDataContainer, error) {
 	log.Debug("call to GetCollectiblesByOwnerAndCollectionWithCursor")
 	return api.s.collectiblesManager.FetchAllAssetsByOwnerAndCollection(chainID, owner, collectionSlug, cursor, limit)
 }
 
+// @deprecated
 func (api *API) GetCollectiblesByOwnerWithCursor(ctx context.Context, chainID wcommon.ChainID, owner common.Address, cursor string, limit int) (*thirdparty.FullCollectibleDataContainer, error) {
```
stefandunca, line 358: @dlipicar I see this and the following ar still referenced on the Nim side; not used, though. Is there any plan to still use them in the future?
- GetCollectiblesByOwnerAndContractAddressWithCursor
- GetCollectiblesByUniqueID

services/wallet/api.go
```diff
@@ -332,26 +348,31 @@ func (api *API) GetOpenseaAssetsByOwnerAndCollection(ctx context.Context, chainI
...
 }
 
+// @deprecated
 func (api *API) GetCollectiblesByUniqueID(ctx context.Context, uniqueIDs []thirdparty.CollectibleUniqueID) ([]thirdparty.FullCollectibleData, error) {
 	log.Debug("call to GetCollectiblesByUniqueID")
 	return api.s.collectiblesManager.FetchAssetsByCollectibleUniqueID(uniqueIDs)
 }
 
+// @deprecated
 func (api *API) GetCollectibleOwnersByContractAddress(chainID wcommon.ChainID, contractAddress common.Address) (*thirdparty.CollectibleContractOwnership, error) {
```
stefandunca: This is used by `community_tokens.Service.fetchCommunityOwners`

services/wallet/api.go
```diff
@@ -376,26 +397,31 @@ func (api *API) GetEthereumChains(ctx context.Context) ([]*network.CombinedNetwo
 	return api.s.rpcClient.NetworkManager.GetCombinedNetworks()
 }
 
+// @deprecated
 func (api *API) FetchPrices(ctx context.Context, symbols []string, currencies []string) (map[string]map[string]float64, error) {
```
stefandunca, line 400: I see this is used by Nim's `token.Service.getTokenPrice`

services/wallet/api.go
```diff
@@ -376,26 +397,31 @@ func (api *API) GetEthereumChains(ctx context.Context) ([]*network.CombinedNetwo
...
 }
 
+// @deprecated
 func (api *API) FetchMarketValues(ctx context.Context, symbols []string, currency string) (map[string]thirdparty.TokenMarketValues, error) {
 	log.Debug("call to FetchMarketValues")
 	return api.s.marketManager.FetchTokenMarketValues(symbols, currency)
 }
 
+// @deprecated
 func (api *API) GetHourlyMarketValues(ctx context.Context, symbol string, currency string, limit int, aggregate int) ([]thirdparty.HistoricalPrice, error) {
```
stefandunca: Used in building the asset historical value graph.

services/wallet/api.go
```diff
@@ -376,26 +397,31 @@ func (api *API) GetEthereumChains(ctx context.Context) ([]*network.CombinedNetwo
...
 	log.Debug("call to GetHourlyMarketValues")
 	return api.s.marketManager.FetchHistoricalHourlyPrices(symbol, currency, limit, aggregate)
 }
 
+// @deprecated
 func (api *API) GetDailyMarketValues(ctx context.Context, symbol string, currency string, limit int, allData bool, aggregate int) ([]thirdparty.HistoricalPrice, error) {
```
stefandunca: Used for the asset graph

services/wallet/api.go
```diff
@@ -411,6 +437,7 @@ func (api *API) GetEstimatedLatestBlockNumber(ctx context.Context, chainID uint6
 	return api.s.blockChainState.GetEstimatedLatestBlockNumber(ctx, chainID)
 }
 
+// @deprecated
 func (api *API) GetTransactionEstimatedTime(ctx context.Context, chainID uint64, maxFeePerGas *big.Float) (TransactionEstimation, error) {
```
stefandunca, line 438: Seems not to be use, still need a follow up PR to remove its reference and others from the Nim side

stefandunca (approved):
Please see the comments inline

alaibe:
@stefandunca I agree some are still in use in desktop but that doesn't make them "deprecated".

FetchPrices for example should not be used on its own

services/wallet/api.go
alaibe: So that means it is deprecated? still in use but nevertheless shouldn't be right?

services/wallet/api.go
alaibe: I thought, we unified API for that?

services/wallet/api.go
alaibe: Yes referenced but are still deprecated i think

services/wallet/api.go
alaibe: Indeed

services/wallet/api.go
alaibe: indeed

services/wallet/api.go
alaibe: We should have everything we need inside the wallet token?

services/wallet/api.go
stefandunca: Yes, this should go away.

services/wallet/api.go
dlipicar: We shouldn't use this, since we're now using the new Async API. It's just a matter of removing the definitions from the `backend.nim` file

services/wallet/api.go
dlipicar: Yeap, this one shouldn't be deprecated

services/wallet/api.go
dlipicar: I believe this was previously used to fetch the current price in the send modal if the cached price is older than X. Not sure if that's still the case though

PR #4006, opened 2023-09-07
fix(archives): Skip importing the archives for community when user not a member

borismelnik:
This PR should skip side effect when the app firing log messages that the archives can't be decrypted.
In short, if user not a member of community he doesn't have decryption key.
[desktop issue](https://github.com/status-im/status-desktop/issues/12005) 

Important changes:

Closes #

### DIFF
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -3090,6 +3090,21 @@ func (m *Messenger) enableHistoryArchivesImportAfterDelay() {
 	}()
 }
 
+func (m *Messenger) checkIfIMemberOfCommunity(communityID types.HexBytes) error {
+	community, err := m.communitiesManager.GetByID(communityID)
+	if err != nil {
+		m.communitiesManager.LogStdout("couldn't get community to import archives", zap.Error(err))
+		return err
+	}
+
+	if !community.HasMember(&m.identity.PublicKey) {
+		m.communitiesManager.LogStdout("can't import archives when user not a member of community")
+		return ErrUserNotMember
+	}
+
+	return nil
+}
+
 func (m *Messenger) resumeHistoryArchivesImport(communityID types.HexBytes) error {
 	archiveIDsToImport, err := m.communitiesManager.GetMessageArchiveIDsToImport(communityID)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


protocol/messenger_communities.go
```diff
@@ -3168,6 +3168,17 @@ importMessageArchivesLoop:
...
+			community, err := m.communitiesManager.GetByID(communityID)
+			if err != nil {
+				m.communitiesManager.LogStdout("couldn't get community to import archives", zap.Error(err))
+				return err
+			}
+
+			if !community.HasMember(&m.identity.PublicKey) {
+				m.communitiesManager.LogStdout("can't import archives when user not a member of community")
+				break importMessageArchivesLoop
+			}
```
0x-r4bbit: I think generally the check is fine, but I think it should not happen inside of this loop. We know the communityID before we run into this part here, so I think we shouldn't even call `importHistoryArchives()` if we know we aren't a member of the community.

If, however, we are a member, the loop is running and our memership state changes to not being a member, then we need to ensure that all tasks related to this community are stopped.

borismelnik: Good point! thanks!

protocol/messenger_communities.go
```diff
@@ -3138,6 +3138,17 @@ func (m *Messenger) SlowdownArchivesImport() {
...
+	community, err := m.communitiesManager.GetByID(communityID)
+	if err != nil {
+		m.communitiesManager.LogStdout("couldn't get community to import archives", zap.Error(err))
+		return err
+	}
+
+	if !community.HasMember(&m.identity.PublicKey) {
+		m.communitiesManager.LogStdout("can't import archives when user not a member of community")
+		return ErrUserNotMember
+	}
```
0x-r4bbit: Again, I think if we know that we're not a member we shouldn't be caling `importHistoryArchives` in the first place.
Returning here early is fine, but we should also check the places when membership status changes.

If the client receives a signal that we're no longer a member then we shouldn't be calling `importHistoryArchives` in the first place. At the very least we should check if any import tasks are running and cancel/stop them in case we're no longer a member.

borismelnik: Oh sorry, I didn't get it correctly (thought I should do it before starting the loop. Okay I will check

0x-r4bbit: Returning early here is perfectly fine I think.
Then we don't have the condition everywhere where we call this function (there might only be one place, I don't actually know from the top of my head).

But we should for sure double check if we're correctly stopping this task in case a membership state changes.

0x-r4bbit (approved):
I'm going to approve this now as I don't want to block the changes made here.

However, as mentioned in https://github.com/status-im/status-go/pull/4006#discussion_r1321849368, we should ensure that, when we're being kicked from a community (that is, when we receive a community description in which we're no longer a member), we should make sure we're stopping any archive tasks.

jrainville (approved):
Like Pascal said, it would be good to cancel the task if we leave the community or are kicked.

jrainville (approved):

PR #4009, opened 2023-09-08
feat: Add activity filtering by contract deploy and minting

Cuteivist:
task: https://github.com/status-im/status-desktop/issues/12093
Corresponding status-desktop change with video:  https://github.com/status-im/status-desktop/pull/12109

* Added filtering by `Contract Deployment` and `Mint` type
ContractDeployment is subtype of Send and Mint is subtype of Receive, that required hiding those specific transfers when filtering by Send or Receive.
* Fixed missing `chainID` for `Contract Deployment`
* Fixed issue with transfer unit tests utils using wrong value for block table
* Fixed crash in rare cases when `TokenID` could be `nil`


stefandunca (approved):
Great implementation

PR #4010, opened 2023-09-08
fix mobile issue #17223 

qfrank:
Major changes:
- add 1 minute to the before time, because we need the cert to be valid for the duration of the timeout period.
- retry connections when failed to get server cert

relate [mobile issue](https://github.com/status-im/status-mobile/issues/17223)


siddarthkay (approved):

server/pairing/client.go
```diff
@@ -69,6 +76,12 @@ func NewBaseClient(c *ConnectionParams) (*BaseClient, error) {
 
 	if serverCert == nil {
 		certErrs = fmt.Errorf("failed to connect to any of given addresses. %w", certErrs)
+		if retryCount < maxRetries {
```
igor-sirotin: This makes it total of 4 attempts, is this right?

server/pairing/client.go
```diff
@@ -69,6 +76,12 @@ func NewBaseClient(c *ConnectionParams) (*BaseClient, error) {
...
+			time.Sleep(2 * time.Second)
+			logger.Info("failed to connect to any of given addresses. Retrying...", zap.Error(certErrs))
+			retryCount++
+			goto loop
+		}
 		signal.SendLocalPairingEvent(Event{Type: EventConnectionError, Error: certErrs.Error(), Action: ActionConnect})
 		return nil, certErrs
 	}
```
Samyoul: I wonder if we need to break some of this code into a new function:

```go
func findServerCert(c *ConnectionParams) (*x509.Certificate, error) {
	netIps, err := server.FindReachableAddressesForPairingClient(c.netIPs)
	if err != nil {
		return nil, err
	}

	for i := range netIps {
		u, err := c.URL(i)
		if err != nil {
			return nil, err
		}
		serverCert, err = getServerCert(u)
		if err != nil {
			var certErr string
			if certErrs != nil {
				certErr = certErrs.Error()
			}
			certErrs = fmt.Errorf("%sconnecting to '%s' failed: %s; ", certErr, u, err.Error())
			continue
		}
		baseAddress = u
		break
	}

	if serverCert == nil {
		certErrs = fmt.Errorf("failed to connect to any of given addresses. %w", certErrs)
		return nil, certErrs
	}
	
	return serverCert, nil
}
```

Usage, would be something like this.
```go
maxRetries := 3

for i := 0; i< maxRetries; i++ {
	serverCert, err = findServerCert(c)
	if err == nil {
		break
	}
}

if err != nil {
	signal.SendLocalPairingEvent(Event{Type: EventConnectionError, Error: certErrs.Error(), Action: ActionConnect})
	return nil, certErrs
}
```

What do you think @qfrank ?

Edited because my suggestion looked wonky.

server/pairing/client.go
```diff
@@ -69,6 +76,12 @@ func NewBaseClient(c *ConnectionParams) (*BaseClient, error) {
 
 	if serverCert == nil {
 		certErrs = fmt.Errorf("failed to connect to any of given addresses. %w", certErrs)
+		if retryCount < maxRetries {
+			time.Sleep(2 * time.Second)
+			logger.Info("failed to connect to any of given addresses. Retrying...", zap.Error(certErrs))
```
igor-sirotin: Maybe logger.Warn?

Samyoul (approved):
This is look great thank you.

igor-sirotin (approved):
Look good, thank you!
We want to take this into the desktop release, right?

server/pairing/client.go
qfrank: I may accept both since the loop body is short enough, but i can change it for you 🙂 @Samyoul

server/pairing/client.go
qfrank: yes if all failed, do you think it's too many? 🤔 @igor-sirotin

igor-sirotin: It's fine by me. I just got it wrong at first from "retries := 3" 🙂

server/pairing/preflight/preflight.go
```diff
@@ -34,8 +34,9 @@ func preflightHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func makeCert(address net.IP) (*tls.Certificate, []byte, error) {
-	notBefore := time.Now()
-	notAfter := notBefore.Add(time.Minute)
+	now := time.Now()
+	notBefore := now.Add(-time.Minute)
```
cammellos: can we add `time.Minute` into a variable? and we can call it something like:
`certificateMaxClockDrift` or something similar?

cammellos (approved):

qfrank:
> We want to take this into the desktop release, right?

yes , pls ❤️ @igor-sirotin

PR #2745, opened 2022-07-08
feat: share image/username on separate topic on a community

osmaczko:
- `publishContactCode` now also publishes on joined communities separate topic
- `publishIdentityImage` is now triggered each time user joins a new community, this is to ensure other community members got updated with user details immediately

closes: #2707

PR #4011, opened 2023-09-08
Publish image and contact code when joining community

cammellos:
We publish a contact code as soon as we join a community, so that other users in the community can retrieve it


protocol/messenger_communities.go
```diff
@@ -791,6 +791,11 @@ func (m *Messenger) joinCommunity(ctx context.Context, communityID types.HexByte
 		if err = m.subscribeToCommunityShard(community.ID(), community.Shard()); err != nil {
 			return nil, err
 		}
+
+		if err = m.publishContactCodeInCommunity(community); err != nil {
```
osmaczko, line 795: Contact code is already triggered few lines below: 
https://github.com/status-im/status-go/blob/cddfe163cf0bea9ae81b747fa8f35153164c54a3/protocol/messenger_communities.go#L818

:thinking:

osmaczko: It was already addressed in this PR: https://github.com/status-im/status-go/pull/2745
> `publishIdentityImage` is now triggered each time user joins a new community, this is to ensure other community members got updated with user details immediately

igor-sirotin:
Closing as resolved

PR #4013, opened 2023-09-11
fix: upsert community customer token

alaibe:


micieslak (approved):
Tested, works as expected!

PR #4014, opened 2023-09-12
feat: register and maintain keycard local pairing file by `status-go`

### LINKED ISSUE #4003: Register and maintain keycard local pairing file by `status-go`
Similar to `initKeystore` at the app start or re-init it on the Status account change we need to init keycard local pairing file to `status-go`. Actually, we need to let `status-go` know in which location the keycard pairing file resides. For the difference from keystore files, it's maintained per installation, not per Status account.

Once it's registered on the `status-go` side, we need to:
1. minimize the likelihood of running into "max pairing slots reached" error by sending the `pairings.json` file content along with the `SyncKeypair` message and that way ensure that all paired devices have the same `pairings.json` file

2. remove `keycardPairings` property from the local pairing mechanism cause a client doesn't need to send it explicitly, since that will be handled on the `status-go` side

saledjenic:
A few things to note here:

1. `KeycardPairingDataFile` property needs to be set on the client side (represents the full path to the `pairings.json` file) 
 and passed to status-go via `NodeConfig`, it's a required property now
2. Content of the pairings.json file (used to keep the keycard pairing details) is now sent together with the `SyncKeypair` sync message to paired devices, that way we're minimizing the possibility of running into "no free pairing slots" error
3. Since the status-go part is aware of the `pairings.json` location, we don't need to explicitly set `keycardPairings` when local pairing, that is completely handled by the status-go (second commit)

Closes: #4003


services/wallet/service.go
```diff
@@ -257,3 +261,29 @@ func (s *Service) Protocols() []p2p.Protocol {
...
+	if os.IsNotExist(err) {
+		dir, _ := filepath.Split(s.keycardPairingsFile)
+		err = os.MkdirAll(dir, 0700)
+		if err != nil {
+			return err
+		}
+	}
+
+	return ioutil.WriteFile(s.keycardPairingsFile, content, 0600)
+}
```
alaibe, line 265: Service is entry point only to real logic, you can see how everything is being dispatch from the service. Could you move this into a keycard_paring_manager as we do elsewhere?

saledjenic: @alaibe updated.

alaibe (approved):

saledjenic:
@rasom would be good if somebody could add `KeycardPairingDataFile` param to `NodeConfig` for the mobile app?

rasom:
@saledjenic 
as it is set in `defaultNodeConfig` we should be covered on mobile as well

rasom (approved):

PR #4015, opened 2023-09-12
perf: avoid call to market manager

alaibe:


dlipicar (approved):
So, if I understand correctly we're now only fetching prices for whatever tokens will be displayed on the wallet (hardcoded list or balance >0). Sounds like a reasonable change! 
Nice job

alaibe:
@dlipicar exactly!

PR #4016, opened 2023-09-12
Fix/mobile issue 17255 (unable to reset password on ios)

qfrank:
On iOS, the rootDataDir value does not contain a trailing slash. This is causing an incorrectly formatted temporary file path to be generated, leading to an "operation not permitted" error.

relate mobile [issue](https://github.com/status-im/status-mobile/issues/17255)


api/geth_backend.go
```diff
@@ -1030,7 +1030,14 @@ func (b *GethStatusBackend) createTempDBFile(pattern string) (tmpDbPath string,
 		err = errors.New("root datadir wasn't provided")
 		return
 	}
-	file, err := os.CreateTemp(filepath.Dir(b.rootDataDir), "*-"+pattern)
+	rootDataDir := b.rootDataDir
+	if !strings.HasSuffix(rootDataDir, "/") {
```
cammellos, line 1039: Could you please leave a comment explaining the issue?

cammellos (approved):

api/geth_backend.go
qfrank: sure

siddarthkay (approved):

PR #4021, opened 2023-09-13
chore(wallet) remove activity filter dependency on accounts

### LINKED ISSUE #3881: Use filtered addresses that are passed from UI to for filter quiery in activity, instead of syncing `keypairs_accounts` table
# Problem

Currently for query in activity to work I created a temporary table `keypairs_accounts` in wallet db and sync addresses from main db with this table.

https://github.com/status-im/status-go/blob/5fab5947327c1d278a2dc6b7997e812d7a5c02ae/services/wallet/activity/activity.go#L664

Instead we should use the addresses passed from UI for that and avoid databases sync.

stefandunca:
### Updates status-desktop [#11980](https://github.com/status-im/status-desktop/issues/11980)

The activity filter must pass at least one account address to extract the type sent/received.

Had to keep the duplicated addresses with a temporary table for UNION with CTE acting as an optimization. As soon as we use `filter_addresses` CTE or `filter_addresses_table` temp table or switch them alternatively for `JOIN` or `IN` clauses the performance drops significantly


services/wallet/activity/activity.go
```diff
@@ -407,23 +407,24 @@ func getActivityEntries(ctx context.Context, deps FilterDependencies, addresses
...
-	// Since the filter query needs addresses which are in a different database, we need to update the
-	// keypairs_accounts table in the current database with the latest addresses from the accounts database
-	err := updateKeypairsAccountsTable(deps.accountsDb, deps.db)
+	queryString := fmt.Sprintf(queryFormatString, involvedAddresses, toAddresses, assetsTokenCodes, assetsERC20, networks,
+		joinedMTTypes)
+
+	// The duplicated temporary table UNION with CTE acts as an optimization
+	// As soon as we use filter_addresses CTE or filter_addresses_table temp table
+	// or switch them alternatively for JOIN or IN clauses the performance drops significantly
+	_, err := deps.db.Exec(fmt.Sprintf("DROP TABLE IF EXISTS filter_addresses_table; CREATE TEMP TABLE filter_addresses_table (address VARCHAR PRIMARY KEY); INSERT OR IGNORE INTO filter_addresses_table (address) VALUES %s;\n", involvedAddresses))
```
IvanBelyakoff, line 416: Stefan, I dont not quite understand the issue with performance drops, sorry. But I wonder how fast is table dropping and creation on every DB query for activities. Can we have a temp table which we could update on each request?

IvanBelyakoff: Another question, I heard during the meeting that you were going to delete `allAddresses` parameter from API. Am I missing something, because I don't know how to drop it if you dont have access to the main db. And you are not dropping it as I see. Just to clarify

stefandunca: > Stefan, I dont not quite understand the issue with performance drops, sorry.

In the past, we had the accounts table `UNION ALL` with the filtered addresses [here in `filter.sql`](https://github.com/status-im/status-go/blob/5e8300d6a12fe22a97c2219dac17888499003b8f/services/wallet/activity/filter.sql#L48) as a CTE (Common Table Expression) like `WITH filter_addresses AS (...)`.
This was for the optimization not to pass all addresses in the case of the "all addresses selected" filter call and fetch them from the naturally available `addresses` table.

Because we don't have the `addresses` table in the same DB the optimization doesn't help anymore and we had to copy cross-DBs the addresses, which we did even in the case of on address already passed anyway.

Naturally, I wanted to drop this data duplication and require only to pass the addresses list using formatted string for the query `%s`. However, I noticed a drastic performance drop.
After a painful investigation, I discovered that having only one source of addresses slows down the query tenfold. I tried a lot of options (using only the temp DB, or selectively the CTE for e.g. `tx_to_address IN addresses` and table for join; also the other way around. The query is still slow.

Therefore, I gave up and left the same construction with a CTE that sources alternatively the same data from the table DB and plain inserted addresses duplication. Not pretty, but I already spent too much time figuring this out.

> But I wonder how fast the table dropping and creation on every DB query for activities. Can we have a temp table which we could update on each request?

Given how often and how much it slows don to do the query, it is negligible. However, I also thought of a similar optimization to cache every static input as a session-based, but I still see this as a minor (under 1%) gain to do it as early.

> Another question, I heard during the meeting that you were going to delete allAddresses parameter from API.

There must be a misunderstanding. I dropped the old way of passing `all addresses selected` as empty addresses, sourced them from the status DB, and added this `allAddresses` boolean that you explicitly set along the passed addresses, which we already had to do in case only one address was selected.

I hope this clarifies your questions.

IvanBelyakoff (approved):
Great job!

dlipicar (approved):

PR #4022, opened 2023-09-13
Implemented balance history based on transfers

### LINKED ISSUE #3835: Wallet Merge balance history and transfer process

IvanBelyakoff:
### Description
Before: Balance history was fetched independently from fetching transfers.
Now: 
  - Balance history utilizes balances that are cached during fetching transfers. Those balances that are missing (for ERC20 transfers) are fetched for the blocks where the transfers were made.
  - Additional points added at the edges of the requested interval for UI to draw at least a line.
  - Additional padding points are added to the result value for the chart to be more smooth (if there were no balance changes during the requested interval, and UI has only two points - at the beginning and the end of the interval, it draws a straight line, despite that market price for the token might have changed which should be reflected on the chart. A straight line by itself looks confusing for users, imo)
  - Balance cache uses TTL to clear the items that are not accessed during an interval (5 minutes for now)
  - Balance cache is now shared between history and transfers strategy

### Notes
There is an issue with loading balances and transfers on adding a new account caused by exceeding providers RPC requests per second limits. Without some throttling it is hard to see the balance on adding account. I filed an [issue](https://github.com/status-im/status-go/issues/4096) and **made a `sleep` for now.**

Closes #3835


services/wallet/balance/balance_cache.go
```diff
@@ -41,253 +35,187 @@ type CacheIface interface {
...
+	_, exists := b.nonces[account]
+	if !exists {
+		b.nonces[account] = make(map[uint64]N)
 	}
 
-	_, exists := b.sortedRanges[account]
+	_, exists = b.nonces[account][chainID]
 	if !exists {
-		b.sortedRanges[account] = make(map[uint64][]nonceRange)
+		b.nonces[account][chainID] = reflect.New(reflect.TypeOf(b.nonces[account][chainID]).Elem()).Interface().(N)
```
rasom, line 132: I have no idea how common it is to use reflection in such cases in Go. Still I have some discomfort while seeing it and would prefer having simple builder functions provided by `genericCache` implementation instead. 

It (reflection) might be an idiomatic way to do things here (again idk what are performance implications), so if it is fine just ignore the comment.

rasom: provided we have simple builders for all three cache instances we could probably avoid parametrisation of `genericCache`. That also might make it a bit more straightforward not only in terms of getting rid of some reflection code but just general readability

IvanBelyakoff: Hey @rasom , thanks for your feedback. It might be that I mad it too complex while working on the task and there might be another way of doing it. I can see now, that there are less methods that require parameterization that I initially thought and actually did. I would like to clarify if you mean to implement duplicate `AddBalance` and `AddNonce` methods for specific cache types?

rasom: we have 
```go
type genericCache[B cacheIface[uint64, *big.Int], N cacheIface[uint64, *int64], NR cacheIface[int64, nonceRange]] struct {
	nonceRangeCache[NR]

	// balances maps an address and chain to a cache of a block number and the balance of this particular address on the chain
	balances addressChainMap[B]
	nonces   addressChainMap[N]
	rw       sync.RWMutex
}
```

which could be something like this for removing parameters from `genericCache` and getting rid of reflection. It is not going to be less verbose, just arguably a bit more straightforward. 

```go
type cacheBuilderInterface {
    newBalance() cacheIface[uint64, *big.Int]
    newNonce() cacheIface[uint64, *int64]
    newNonceRange() cacheIface[int64, nonceRange]
}

type genericCache struct {
	nonceRangeCache[cacheIface[int64, nonceRange]]

	// balances maps an address and chain to a cache of a block number and the balance of this particular address on the chain
	balances addressChainMap[cacheIface[uint64, *big.Int]]
	nonces   addressChainMap[cacheIface[uint64, *int64]]
        builder cacheBulderInterface   
	rw       sync.RWMutex
}
```

rasom:
```go
//then
b.nonces[account][chainID] = reflect.New(reflect.TypeOf(b.nonces[account][chainID]).Elem()).Interface().(N)
//becomes
b.nonces[account][chainID] = b.builder.newNonce()
```

rasom (approved):
I have no strong objections against reflection (as I don't know how common/idiomatic it is), otherwise no comments from my side. Good luck with conflicts :D

services/wallet/history/service.go
```diff
@@ -306,242 +329,207 @@ func tokenToValue(tokenCount *big.Int, mainDenominationValue float32, weisInOneM
...
-func (i timeIdentity) dataPoint(data map[chainIdentity][]*DataPoint) *DataPoint {
-	return data[i.chain][i.index]
-}
+			if balance == nil {
+				balance, err = client.BalanceAt(ctx, common.Address(address), entry.block)
+				if balance == nil {
+					log.Error("Error getting balance", "chainID", network.ChainID, "address", address.String(), "err", err, "unwrapped", errors.Unwrap(err))
+					return err
+				}
+				time.Sleep(50 * time.Millisecond) // TODO Remove this sleep after fixing exceeding rate limit
```
IvanBelyakoff, line 404: Please pay attention to this sleep. I created an [issue](https://github.com/status-im/status-go/issues/4096) for that. Without it, it is hard to see balance history at all for new accounts.
Please shout if you think this is still inappropriate

services/wallet/balance/balance_cache.go
IvanBelyakoff: Thank you for your suggestion. I will consider it in a follow up [PR](https://github.com/status-im/status-go/issues/4097)

services/wallet/transfer/block_test.go
```diff
@@ -182,3 +182,34 @@ func TestGetNewRanges(t *testing.T) {
 	require.Equal(t, int64(50), newRange.to.Int64())
 	require.Equal(t, 4, len(d))
 }
+
+func TestInsertZeroBalance(t *testing.T) {
+	// b, stop := setupTestTransferDB(t)
```
stefandunca: There are a few commented lines of code, leftovers?

services/wallet/history/service.go
stefandunca: It works for now, and it would be great if we have a task to track this workaround.
Indeed, at first glance, requests/second quota should not be part of the responsibilities of RPC clients and should be offloaded to a throttler who knows the API requirements for each service and handles this transparently.

stefandunca (approved):
Great job. I am looking forward to seeing it in action.

services/wallet/history/service.go
IvanBelyakoff: Thank you, Stefan. Are you saying about some other task rather than the one I mentioned in my comment?

stefandunca: The one you mentioned in your comment.

PR #4027, opened 2023-09-14
Feat/collectibles ownership update improvements

dlipicar:
Part of https://github.com/status-im/status-desktop/issues/11379

Several collectibles improvements, split into commits:

1. Implemented collectibles ownership update timestamp (to know if/when the collectibles list was last updated for a certain address+chainID)

2a. Individual accounts+chain periodic fetch. Allows individually starting/restarting the command. Now, when a new account is added only that account gets fetched.

2b. If an account+chainID is fetched for the first time, the DB is updated as each "chunk" is downloaded. Moreover, an  event is emitted indicating partial update progress. This helps the UI start showing collectibles faster when an account holds a lot of collectibles. Subsequent fetches after the first one will only update the DB + emit the event when the whole list of owned collectibles is fetched. 

3. Implement collectible ownership status. Allows the UI to know if ownership for a given address+chainID is updating, and when (if) the last update occurred.

### DIFF
--- a/services/wallet/collectibles/commands.go
+++ b/services/wallet/collectibles/commands.go
@@ -137,6 +137,7 @@ func (c *loadOwnedCollectiblesCommand) Run(parent context.Context) (err error) {
 
 	pageNr := 0
 	cursor := thirdparty.FetchFromStartCursor
+	start := time.Now()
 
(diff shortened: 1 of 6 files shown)


stefandunca (approved):
Great improvements and implementation.

PR #4028, opened 2023-09-15
feat: Filter by collectibles

Cuteivist:
Added filtering by collectibles.

It only handles transfers, because right now multi tx  and pending tx 

Video:

https://github.com/status-im/status-go/assets/11396062/1062a102-7545-4146-b350-96e8f43ede3c

### DIFF
--- a/services/wallet/activity/activity.go
+++ b/services/wallet/activity/activity.go
@@ -311,8 +311,9 @@ const (
 	fromTrType = byte(1)
 	toTrType   = byte(2)
 
-	noEntriesInTmpTableSQLValues           = "(NULL)"
-	noEntriesInTwoColumnsTmpTableSQLValues = "(NULL, NULL)"
+	noEntriesInTmpTableSQLValues             = "(NULL)"
+	noEntriesInTwoColumnsTmpTableSQLValues   = "(NULL, NULL)"
+	noEntriesInThreeColumnsTmpTableSQLValues = "(NULL, NULL, NULL)"
 )
 
 //go:embed filter.sql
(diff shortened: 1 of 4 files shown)


services/wallet/activity/activity.go
```diff
@@ -368,6 +369,21 @@ func getActivityEntries(ctx context.Context, deps FilterDependencies, addresses
 		}
 	}
 
+	includeAllCollectibles := len(filter.Collectibles) == 0 && !filter.FilterOutCollectibles
+	assetsERC721 := noEntriesInThreeColumnsTmpTableSQLValues
+	if !includeAllCollectibles && !filter.FilterOutCollectibles {
+		assetsERC721 = joinItems(filter.Collectibles, func(item Token) string {
+			// SQL HEX() (Blob->Hex) conversion returns uppercase digits with no 0x prefix
+			tokenID := strings.ToUpper(item.TokenID.String()[2:])
+			address := strings.ToUpper(item.Address.Hex()[2:])
```
IvanBelyakoff, line 384: address as blob is well unmarshalled into `common.Address` if the value is a valid `common.Address` (not NULL, and proper length), so please use common.Address directly wherever possible as it allows you to avoid handling conversion edge cases.
As for tokenID, I think it is well converted to big.Int, which you can convert to hexutil.Big.

services/wallet/activity/filter.sql
```diff
@@ -278,6 +283,19 @@ WHERE
...
 	)
+	AND (
+		includeAllCollectibles
+		OR (
+			transfers.type = "erc721"
+			AND (
+				(
+					transfers.network_id,
+					HEX(transfers.token_id),
+					HEX(transfers.token_address)
```
IvanBelyakoff, line 307: Just use the address as is without HEX(). Please check if the same is valid for token_id.

IvanBelyakoff (commented):
Good job, @Cuteivist. Please have a look at inline comments

services/wallet/activity/activity.go
stefandunca: @IvanBelyakoff, here we prepare the string query and inline the filter items as a CTE (common table expression) in SQL for quick lookup using `IN`. Hence, the format has to be the same as the one taken from the DB (for blobs, we usually use `HEX`, which returns the hexadecimal string representation).

I'm looking now, with [this task](https://github.com/status-im/status-desktop/issues/11036), if I can do the reverse to insert binary and not use HEX in the query as an optimization. Still don't have an answer, though. If this is not possible, I plan to measure if there are speed improvements for injecting all blob filter values using temporary in-memory tables, as in my previous PR. However, the first experiments I did with addresses didn't result any performance gain but a drastic degradation.

stefandunca (approved):
Great job.

alaibe (approved):

PR #4030, opened 2023-09-15
feat: Expose raw token's balance via `GetWalletToken`

micieslak:
Exposes token balance in base units via `GetWalletToken`. The amount is served as a string representing integer decimal. It allows UI to use that value instead of the float value which can be distorted and therefore not ideal for e.g. comparing amounts.

Important changes:
- [x] `RawBalance` field added to `ChainBalance`
- [x] migration created adding `row_balance` column to  `token_balances` table
- [x] `persistance_test` adjusted and fixed to use properly initialized wallet database (with migrations)

### DIFF
--- a/services/wallet/reader.go
+++ b/services/wallet/reader.go
@@ -72,10 +72,11 @@ type TokenMarketValues struct {
 }
 
 type ChainBalance struct {
-	Balance  *big.Float     `json:"balance"`
-	Address  common.Address `json:"address"`
-	ChainID  uint64         `json:"chainId"`
-	HasError bool           `json:"hasError"`
+	RawBalance string         `json:"rawBalance"`
+	Balance    *big.Float     `json:"balance"`
+	Address    common.Address `json:"address"`
+	ChainID    uint64         `json:"chainId"`
+	HasError   bool           `json:"hasError"`
 }
 
 type Token struct {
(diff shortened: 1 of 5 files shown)


alaibe (approved):

services/wallet/reader.go
```diff
@@ -72,10 +72,11 @@ type TokenMarketValues struct {
 }
 
 type ChainBalance struct {
-	Balance  *big.Float     `json:"balance"`
-	Address  common.Address `json:"address"`
-	ChainID  uint64         `json:"chainId"`
-	HasError bool           `json:"hasError"`
+	RawBalance string         `json:"rawBalance"`
```
MishkaRogachev, line 75: Just curious why raw balance is string?

MishkaRogachev (approved):
Great work!

services/wallet/reader.go
micieslak: initially I wanted to use `*big.Int` but it turns out that then serialization doesn't work properly - looks like nim tries to deserialize to a plain number which can be insufficient to handle amount.

stefandunca: > initially I wanted to use `*big.Int` but it turns out that then serialization doesn't work properly - looks like nim tries to deserialize to a plain number which can be insufficient to handle amount.

I think this is the main reason why we usually expose `big.Int` as [*hexutil.Big](https://github.com/status-im/status-go/blob/2f539d3bd23694f1d0db7fdbf1ffde81f42e8d7d/vendor/github.com/ethereum/go-ethereum/common/hexutil/json.go) in the status-go API.

PR #4033, opened 2023-09-15
feature: Unfurl status links

### LINKED ISSUE #3762: Unfurl Status deep links
# Problem

New URL unfurling scheme was implemented in https://github.com/status-im/status-go/pull/3471.
We should also unfurl status deep links on sender side and include a preview in the message.

`UnfurlURLs` should follow this flow for Status deep links:
1. If the URL contains data, decode it first
I'm not sure if this step makes sense, because we plan to unfurl it in step 2.
At the same time, the receiver will update the preview on reception anyway. so maybe this step is important and step 2 is redundant.
3. Request the profile/community info directly from waku
    - `RequestContactInfoFromMailserver` for profile
    - `RequestCommunityInfoFromMailserver` for community
    - Take a think about channels, as it might contain sensitive data that the sender has access to and the receiver doesn't.
    There should be no more info than a Status deep link with data would contain.

# Acceptance Criteria

`UnfurlURLs` should unfurl Status deep links for:
- Profile
- Communiy
- Community channel

# Future Steps

Support the solution in desktop/mobile.

igor-sirotin:
closes #3762
desktop PR: https://github.com/status-im/status-desktop/pull/12303

This PR contains https://github.com/status-im/status-go/pull/4110. Will remove before merging.

# Description

This PR introduces support for unfurling Status contact/community/channel links.

## Storing the data

As we discussed, status links are unfurled in a separate database column. 

So `UnfurlURLs` will now return a `UnfurlURLsResponse` with both link types:
https://github.com/status-im/status-go/blob/70ed74b1a196c7dd8cb804749e069e9b8afa7cd2/protocol/messenger_linkpreview.go#L19-L22

And a separate field is added to `Message`:
https://github.com/status-im/status-go/blob/70ed74b1a196c7dd8cb804749e069e9b8afa7cd2/protocol/protobuf/chat_message.proto#L219

## Contained data

The preview itself contains more data than the original encoded URL.
We add more data (e.g. icons). if it's available locally in the database, this is required for the expected designs.

We could also fetch data from Waku, if it's not available locally, but that might require more time and eventually might fail Which would be a strange behaviour, having that there's some data simply encoded in the URL. We would need some special mechanism of "supplementing the preview":
- in the first stage, decode the url and return it to the user immediately
- at the same time start fetching from waku and if more info is fetched before the message is sent, add it to the preview
- maybe add some special loading UI for this case?

This is much more to do, but it's a not a bad idea I think.

## Minor updates to shared urls

@MishkaRogachev adding you as reviewer for this part

- use pointers
https://github.com/status-im/status-go/blob/70ed74b1a196c7dd8cb804749e069e9b8afa7cd2/protocol/messenger_share_urls.go#L43-L47
- expose `IsStatusSharedUrl` method
https://github.com/status-im/status-go/blob/70ed74b1a196c7dd8cb804749e069e9b8afa7cd2/protocol/messenger_share_urls.go#L533-L535


igor-sirotin:
It's ready to review, I will fix the tests meanwhile. Probably it's linter issues only.

protocol/common/message_test.go
```diff
@@ -250,6 +250,339 @@ func TestConvertFromProtoToLinkPreviews(t *testing.T) {
...
+	invalidPreview.Contact = invalidContactPreview
+	message.StatusLinkPreviews = []StatusLinkPreview{invalidPreview}
+	_, err = message.ConvertStatusLinkPreviewsToProto()
+	require.ErrorContains(t, err, "invalid status link preview, url='https://status.app/u/'")
+}
+
+func TestConvertFromProtoToStatusLinkPreviews(t *testing.T) {
+
+	contact := &protobuf.UnfurledStatusContactLink{
+		PublicKey:   "1",
```
MishkaRogachev: I'd replace the numbers with readable names so that if the test fails it's clear what it's on.

protocol/linkpreview_unfurler_status.go
```diff
@@ -0,0 +1,194 @@
...
+	thumbnail.DataURI, err = image.GetDataURI()
+	if err != nil {
+		return fmt.Errorf("failed to get data uri: %w", err)
+	}
+
+	return nil
+}
+
+func (u *StatusUnfurler) buildContactData(contactData *ContactURLData) (*common.StatusContactLinkPreview, error) {
+	c := new(common.StatusContactLinkPreview)
```
MishkaRogachev: it's better to move the creation to the end, after all the checks are passed

protocol/linkpreview_unfurler_status.go
```diff
@@ -0,0 +1,194 @@
...
+
+	// Now check if there's newer information in the database
+	communityID, err := types.DecodeHex(data.CommunityID)
+	if err != nil {
+		return c, fmt.Errorf("failed to decode community id: %w", err)
+	}
+
+	community, err := u.m.GetCommunityByID(communityID)
+	if err != nil {
+		return c, nil
```
MishkaRogachev: shouldn't we return err?

protocol/linkpreview_unfurler_status.go
```diff
@@ -0,0 +1,194 @@
...
+	contactID, err := multiformat.DeserializeCompressedKey(contactData.PublicKey)
+	if err != nil {
+		return nil, err
+	}
+
+	contact := u.m.GetContactByID(contactID)
+	if contact == nil {
+		return c, nil
+	}
+
```
MishkaRogachev, line 62: if we got a contact object, we can update DisplayName and Description

protocol/linkpreview_unfurler_status.go
```diff
@@ -0,0 +1,194 @@
...
+	// First, fill the output with the data from URL
+	c.CommunityID = data.CommunityID
+	c.DisplayName = data.DisplayName
+	c.Description = data.Description
+	c.MembersCount = data.MembersCount
+	c.Color = data.Color
+	c.TagIndices = data.TagIndices
+
+	// Now check if there's newer information in the database
```
MishkaRogachev: if we get an url with community data it really can be outdated. WDYT about updating all the fields from object?

MishkaRogachev (commented):
Thanks, Igor! Great work!

igor-sirotin:
I missed to implement the message editing, which was kindly caught by `TestEditMessageWithLinkPreviews`.
I'm fixing that now, but that won't change the idea of PR, so I think it makes sense to start reviewing anyway.

UPD: fixed.

protocol/common/message_test.go
igor-sirotin: I was a bit lazy to create all the values and wanted to keep the diversity at the same time 😄 
I'll switch it to smth like `PublicKey_1`, sound good?

protocol/linkpreview_unfurler_status.go
igor-sirotin: This is a good point!

@John-44, it's probably a question for you.
1. When unfurling a Status URL, we first have the data (`PublicKey`, `DisplayName`, `Description`) from the the URL itself.
2. Then we lookup the database for the such contact. If found, we also add the contact icon to the link preview.

But as @MishkaRogachev pointed out, we are able to update the `DisplayName` and `Description` with database values.

Same question about community and channel links.

pros: 
-  most _probably_ database values are more up-to-date 
cons:
- the URL in the message text will still contain the old data, so there will be an inconsistency
- (I think) we don't have a way to distinguish which values are newer



WDYT?

protocol/linkpreview_unfurler_status.go
igor-sirotin: Good question, will discuss all together here: 
https://github.com/status-im/status-go/pull/4033/files/69545bc6e44e27c644c15ed0ef17752d61d77230#r1342465007

protocol/linkpreview_unfurler_status.go
igor-sirotin: Here I think we shouldn't, but we probably should above, if `types.DecodeHex` failed 🙂

protocol/common/message.go
```diff
@@ -772,20 +809,106 @@ func (m *Message) LoadImage() error {
 	return nil
 }
 
-func isValidLinkPreviewThumbnail(thumbnail LinkPreviewThumbnail) bool {
-	return (thumbnail.DataURI == "" && thumbnail.Width == 0 && thumbnail.Height == 0) ||
-		(thumbnail.DataURI != "" && thumbnail.Width > 0 && thumbnail.Height > 0)
+func isValidLinkPreviewThumbnail(thumbnail LinkPreviewThumbnail) error {
```
ilmotta: Since you are changing the function to return an error instead of a boolean, it's better if the function is renamed to `validateLinkPreviewThumbnail`.

Same improvement can be applied to `isValidLinkPreviewForProto` and `isValidForProto`, they would become `validateLinkPreviewsForProto` and `validateForProto`, respectively.

protocol/common/message.go
```diff
@@ -853,11 +976,219 @@ func (m *Message) ConvertFromProtoToLinkPreviews(makeMediaServerURL func(msgID s
...
+			mediaURL = makeMediaServerURL(m.ID, URL, imageId)
+		}
+		return LinkPreviewThumbnail{
+			Width:  int(thumbnail.Width),
+			Height: int(thumbnail.Height),
+			URL:    mediaURL,
+		}
+	}
+
+	convertCommunityFromProto := func(c *protobuf.UnfurledStatusCommunityLink, URL string, thumbnailPrefix string) *StatusCommunityLinkPreview {
```
ilmotta: I think function `ConvertFromProtoToStatusLinkPreviews` can be cleaned up quite a bit if we move the inlined functions `createThumbnail` and `convertCommunityFromProto` outside. 

- `convertCommunityFromProto` doesn't need to close over any value.
- `createThumbnail` closes over `m.ID` and `makeMediaServerURL`, but both can be easily passed as parameters once we extract `createThumbnail` to its own function outside `ConvertFromProtoToStatusLinkPreviews`. To avoid having to repeat the function signature of `makeMediaServerURL` yet again, you could give a name to this function type.

protocol/common/message.go
```diff
@@ -853,11 +976,219 @@ func (m *Message) ConvertFromProtoToLinkPreviews(makeMediaServerURL func(msgID s
...
+				DisplayName: c.DisplayName,
+				Description: c.Description,
+			}
+			if icon := c.GetIcon(); icon != nil {
+				lp.Contact.Icon = createThumbnail(icon, link.Url, "contact-icon")
+			}
+		}
+
+		if c := link.GetCommunity(); c != nil {
+			lp.Community = convertCommunityFromProto(c, link.Url, "community-")
```
ilmotta: The prefixes like `community-`, etc could be extracted to constants to reduce the chance that somebody will copy & paste these values around, for some reason.

protocol/linkpreview_unfurler_status.go
```diff
@@ -0,0 +1,195 @@
...
+	thumbnail.DataURI, err = image.GetDataURI()
+	if err != nil {
+		return fmt.Errorf("failed to get data uri: %w", err)
+	}
+
+	return nil
+}
+
+func (u *StatusUnfurler) buildContactData(contactData *ContactURLData) (*common.StatusContactLinkPreview, error) {
+
```
ilmotta: Sometimes functions in this PR don't start with a blank line, sometimes they do. In terms of consistency, I think it's more common in status-go to not start or end functions with blank lines, so it would be nice if you could make the style more consistent.

protocol/linkpreview_unfurler_status.go
```diff
@@ -0,0 +1,195 @@
...
+	c.DisplayName = contactData.DisplayName
+	c.Description = contactData.Description
+
+	contact := u.m.GetContactByID(contactID)
+	if contact == nil {
+		return c, nil
+	}
+
+	// TODO: Should we try to fetch from waku?
+	//if contact == nil {
```
ilmotta: I'm on the camp that prefers when commented code is not merged, except very rare cases, but I don't know how strong of a convention this in status-go, so maybe it's fine. Usually the code lurks around after the merge and nobody updates them later.

protocol/linkpreview_unfurler_status.go
```diff
@@ -0,0 +1,195 @@
...
+		return preview, err
+	}
+
+	// If a URL has been successfully parsed,
+	// any further errors should not be returned, only logged.
+
+	if resp.Contact != nil {
+		preview.Contact, err = u.buildContactData(resp.Contact)
+		u.logger.Warn("error when building contact data: ", zap.Error(err))
+		u.logger.Info("<<< StatusUnfurler::Unfurl", zap.Any("contact", preview.Contact))
```
ilmotta: I see some C++ tentacles in the log with the double colon syntax ;) Just kidding

protocol/messenger_linkpreview_test.go
```diff
@@ -378,3 +372,191 @@ func (s *MessengerLinkPreviewsTestSuite) Test_UnfurlURLs_Image() {
...
+	s.Require().NoError(err)
+	s.Require().NotNil(c)
+
+	pubkey, err := c.PublicKey()
+	s.Require().NoError(err)
+
+	shortKey, err := s.m.SerializePublicKey(crypto.CompressPubkey(pubkey))
+	s.Require().NoError(err)
+
+	payload, err := images.GetPayloadFromURI("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAADIAAAAyCAYAAAAeP4ixAAAAiklEQVR4nOzWwQmFQAwG4ffEXmzLIizDImzLarQBhSwSGH7mO+9hh0DI9AthCI0hNIbQGEJjCI0hNIbQxITM1YfHfl69X3m2bsu/8i5mIobQGEJjCI0hNIbQlG+tUW83UtfNFjMRQ2gMofm8tUa3U9c2i5mIITSGqEnMRAyhMYTGEBpDaO4AAAD//5POEGncqtj1AAAAAElFTkSuQmCC")
```
ilmotta: You could make this slightly more readable without requiring horizontal scrolling by breaking the string:

```go
payload, err := images.GetPayloadFromURI("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAADIAAAAyCAYAAAAeP4ixAAAAiklE" +
	"QVR4nOzWwQmFQAwG4ffEXmzLIizDImzLarQBhSwSGH7mO+9hh0DI9AthCI0hNIbQGEJjCI0hNIb" +
	"QxITM1YfHfl69X3m2bsu/8i5mIobQGEJjCI0hNIbQlG+tUW83UtfNFjMRQ2gMofm8tUa3U9c2i5" +
	"mIITSGqEnMRAyhMYTGEBpDaO4AAAD//5POEGncqtj1AAAAAElFTkSuQmCC")
```

server/handlers.go
```diff
@@ -980,3 +995,127 @@ func handleLinkPreviewThumbnail(db *sql.DB, logger *zap.Logger) http.HandlerFunc
...
+				return nil, fmt.Errorf("channel community banner is empty")
+			}
+			return channel.Community.Banner.Payload, nil
+		}
+	}
+
+	return nil, fmt.Errorf("invalid query parameter 'image-id' value")
+}
+
+func handleStatusLinkPreviewThumbnail(db *sql.DB, logger *zap.Logger) http.HandlerFunc {
```
ilmotta: This function (and indirectly `getStatusLinkThumbnailPayload`) should be covered with unit tests because it's our best chance to make sure the endpoint works well, and the handler does quite a lot with all the conditional checks. You can base the work off of the tests I wrote for `handleLinkPreviewThumbnail` in `TestHandleLinkPreviewThumbnail`.

services/ext/api.go
```diff
@@ -1157,7 +1157,7 @@ func (api *PublicAPI) GetTextURLs(text string) []string {
 // be removed from the response.
 //
 // This endpoint expects the client to send URLs normalized by GetTextURLs.
-func (api *PublicAPI) UnfurlURLs(urls []string) ([]common.LinkPreview, error) {
+func (api *PublicAPI) UnfurlURLs(urls []string) (protocol.UnfurlURLsResponse, error) {
```
ilmotta, line 1169: Just a reminder that this is a breaking change for the mobile client. Before merging this PR it's important to create a mobile PR that is reviewed, tested, and ready to be merged before you merge your PR.

I can implement the changes in `status-mobile` for us since it's trivial, but it will need to go over QA, and that can take some time at the moment.

server/handlers.go
```diff
@@ -980,3 +995,127 @@ func handleLinkPreviewThumbnail(db *sql.DB, logger *zap.Logger) http.HandlerFunc
...
+				return nil, fmt.Errorf("channel community is empty")
+			}
+			if channel.Community.Banner == nil {
+				return nil, fmt.Errorf("channel community banner is empty")
+			}
+			return channel.Community.Banner.Payload, nil
+		}
+	}
+
+	return nil, fmt.Errorf("invalid query parameter 'image-id' value")
```
ilmotta: This error will be returned if `links.UnfurledStatusLinks` is empty, but that's not quite right. Perhaps better is to change the code to 1. loop and find the `UnfurledStatusLink` that matches `URL` 2. Check if it was found and return a different error if it wasn't and 3. proceed as usual.

```go
var links protobuf.UnfurledStatusLinks
err = proto.Unmarshal(result, &links)
if err != nil {
	return nil, fmt.Errorf("failed to unmarshal protobuf.UrlPreview: %w", err)
}

var foundLink *protobuf.UnfurledStatusLink
for _, p := range links.UnfurledStatusLinks {
	if p.Url == URL {
		foundLink = p
	}
}
if foundLink == nil {
	// return error not found
}

switch imageID {
	...
}
```

protocol/migrations/sqlite/1695331492_add_status_link_previews.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE user_messages ADD COLUMN unfurled_status_links BLOB;
```
ilmotta, line 1: Just sharing, nothing to do now, but I still fear status links have far more reasons to change than the relatively stable API of normal unfurled links, which could mean painful migrations. But since the `user_messages` table should never have its data migrated (according to our past discussions) then things can get tricky in the future.

A part of me still thinks this more "fragile" blob should be captured in a separate table, with explicit columns, although the join could affect the performance (or maybe not).

Don't worry, I'm just rambling here.

protocol/messenger_linkpreview.go
```diff
@@ -127,27 +128,44 @@ func GetURLs(text string) []string {
...
-			m.logger.Info("failed to unfurl", zap.String("url", url), zap.Error(err))
+			m.logger.Warn("failed to unfurl", zap.String("url", url), zap.Error(err))
 			continue
 		}
-		previews = append(previews, p)
+		r.LinkPreviews = append(r.LinkPreviews, p)
 	}
 
-	return previews, nil
+	m.logger.Info("<<< UnfurlURLs", zap.Any("response", r))
```
ilmotta: The response can be extremely large for logging since there can be a lot of image data encoded. I might be wrong, but I think we should be careful, maybe remove this or log a subset of the response without the encoded data URIs. In my past experiments I only temporarily logged responses due to their size.

protocol/linkpreview_unfurler_status.go
igor-sirotin: Sure, I'll drop it.
I've added a question about this to the PR anyway. And you remember I also discussed this with you, it will probably require more app-backend signals communication, so we're not going to do this now I think.

protocol/linkpreview_unfurler_status.go
igor-sirotin: Actually everything with `<<<` is just a debugging-purposes output, I will remove this 🙂

protocol/messenger_linkpreview.go
igor-sirotin: forgot to remove 👌

protocol/linkpreview_unfurler_status.go
ilmotta: Thanks @igor-sirotin. Yes, I remember our conversation :) I just commented to remind you to consider removing the commented code. In `status-mobile` we tend to avoid TODO comments and prefer to rely on GH issues, but of course this is not necessarily a practice we follow in status-go.

protocol/common/message.go
igor-sirotin: Agree 👍 
There's one problem though. It's not possible to use `server` package here, it creates a cycle dependency 🤦 

```go
package github.com/status-im/status-go/server
	imports github.com/status-im/status-go/ipfs
	imports github.com/status-im/status-go/params
	imports github.com/status-im/status-go/protocol/pushnotificationserver
	imports github.com/status-im/status-go/protocol/common
	imports github.com/status-im/status-go/server: import cycle not allowed
```

And I don't think it's a good idea to store those const in 2 places.
I'll take a look if it's possible to break this import chain somewhere.

ilmotta: Ouch 🤦

igor-sirotin: Fixed one cycle, got another 😄 
```go
package github.com/status-im/status-go/server
	imports github.com/status-im/status-go/signal
	imports github.com/status-im/status-go/multiaccounts/settings
	imports github.com/status-im/status-go/protocol/common
	imports github.com/status-im/status-go/server: import cycle not allowed
```

igor-sirotin: oh there's actually a direct dependency 🤦 
```go
# github.com/status-im/status-go/server
package github.com/status-im/status-go/server
	imports github.com/status-im/status-go/protocol/common: import cycle not allowed in test
```

igor-sirotin: Well... I don't think there's a fast way of removing this dependency now.
I will probably just move the imageId constants to `common` package and leave a comment there.
@ilmotta do you see any better way?

or maybe there's a suitable package for such constants 🤔

ilmotta: @igor-sirotin, I can't think of any other way you haven't already tried.

services/ext/api.go
ilmotta: FYI: @igor-sirotin, I have a mobile branch with the fix for the new response `UnfurlURLsResponse`. I'll open the mobile PR as soon as this status-go PR is finished.

protocol/migrations/sqlite/1695331492_add_status_link_previews.up.sql
igor-sirotin: I agree.

Actually there's a good news that we're going to update it on receiver side anyway. Only IDs are required to fetch the data. So we could migrate it later to some simple table, where we have IDs as separate columns, but all other data as blob 🤔

igor-sirotin:
@ilmotta, thank you very much for your review and detailed comments, I appreciate that ❤️ 
l I've addressed all of you comments, please, check it out.

I decided to refactor things on the way, therefore you'll find some new separate files.
I also took an opportunity to refactor `handlers_test` to be a Suite with test cases.

Finnaly, as discussed, I will update the unfurling logic in this PR.

protocol/linkpreview_unfurler_status.go
igor-sirotin: I've discussed this with Space Assasians crew. Here's the decision for unfurling status links:
- ignore the URL-encoded data (except IDs)
- take data from database (95% cases it will be there, because link is generated locally)
- if no data yet, fetch it from waku

I will correspondingly update this PR.

Thanks for bringing this up, @MishkaRogachev!

protocol/linkpreview_unfurler_status.go
igor-sirotin: Aaaand now this code will be uncommented 😄

ilmotta:
> @ilmotta, thank you very much for your review and detailed comments, I appreciate that ❤️ l I've addressed all of you comments, please, check it out.
> 
> I decided to refactor things on the way, therefore you'll find some new separate files. I also took an opportunity to refactor `handlers_test` to be a Suite with test cases.
> 
> Finnaly, as discussed, I will update the unfurling logic in this PR.

Thank you @igor-sirotin. I'm going to review today again.

igor-sirotin:
Pushed the change:
- ignore URL data
- fetch from waku if nothing's available locally

protocol/common/message.go
```diff
@@ -772,98 +754,6 @@ func (m *Message) LoadImage() error {
 	return nil
 }
 
-func isValidLinkPreviewThumbnail(thumbnail LinkPreviewThumbnail) bool {
```
MishkaRogachev, line 766: I appreciate these changes

protocol/linkpreview_unfurler_status.go
```diff
@@ -0,0 +1,187 @@
...
+
+func NewStatusUnfurler(URL string, messenger *Messenger, logger *zap.Logger) *StatusUnfurler {
+	return &StatusUnfurler{
+		m:      messenger,
+		logger: logger,
+		url:    URL,
+	}
+}
+
+func buildThumbnail(image *images.IdentityImage, thumbnail *common.LinkPreviewThumbnail) error {
```
MishkaRogachev: mb `updateThumbnail` if it's not in the arguments

MishkaRogachev (approved):
Great work!

protocol/linkpreview_unfurler_status.go
ilmotta: I second `updateThumbnail`, because `build` implies the function is usually immutable.

ilmotta (approved):
Great work in this PR @igor-sirotin. And thank you for organizing meetings to openly discuss the features.

Since the code in this PR is relatively stable now, I'll move forward with the mobile PR today (Friday) and get it checked by QAs. The mobile QA team is quite busy at the moment. I'll keep you posted when it's approved so you can merge this one.

pavloburykh:
@igor-sirotin hello! Could you please rebase current branch? We need it for proper testing of correspondent mobile PR https://github.com/status-im/status-mobile/pull/17573 Thank you!

igor-sirotin:
@pavloburykh sure, done 👌

pavloburykh:
> @pavloburykh sure, done 👌

thank you @igor-sirotin!

protocol/protobuf/chat_message.proto
```diff
@@ -126,6 +133,46 @@ message UnfurledLink {
   }
 }
 
+message UnfurledStatusContactLink {
+  string public_key = 1;
```
cammellos: This could be the compressed pub key to save on space (bytes)

protocol/protobuf/chat_message.proto
```diff
@@ -126,6 +133,46 @@ message UnfurledLink {
...
+  string display_name = 2;
+  string description = 3;
+  UnfurledLinkThumbnail icon = 4;
+}
+
+message UnfurledStatusCommunityLink {
+  string community_id = 1;
```
cammellos: could also be bytes, to save space

cammellos (approved):

igor-sirotin:
🎉

Issue #3971, opened 2023-08-29
Implement ERC20 transfers initial history check based on supported tokens balances and nonce


IvanBelyakoff:
# Problem

Old implementation used `setInitialRange` method to set `block_ranges` table's `from=0`, `to=$head` values for OnDemandStrategy for transfers, which allowed to avoid history check for brand new Status accounts to save some RPC calls to providers. The method was removed from API as it bypassed strategy and revealed internal implementation to the client.
Currently it appears that there is no way to know on status-go side if account is a brand new Status acc. Consider either passing "brandNewAcc" option in node config when creating a new account or drop this case in favor of general implementation of initial looking for transfers:

# Implementation
1) Currently for OnDemandStrategy we search for a range of blocks that contains 20 < n < 25 number  (nonce delta)  of  outgoing ERC20 transfers using binary search.
2) Currently for SequentialFetchStrategy we search for blocks with transfers using `getLogs` and default block chunk range == 10K blocks until the `start` block for ETH transfers is found (first ever transfer of ETH).

We need to implement checking all supported ERC20 tokens balances and ETH nonce prior to calling `getLogs` to know if there is any ERC20  transfers history for the account at all:
1) If nonce == 0, but some balance is not 0, check:
  - if it is possible to use binary search by balances of tokens to search for ERC20 transfers?
  - if it is cheaper then using `getLogs` ?
  - search for the transfers using one of the above methods
2) if nonce == 0 and all balances == 0 then set blocks ranges accordingly as fully checked to avoid further search.
3) If nonce != 0, nothing to be done for this task, as we will search for ETH transfers anyways and will be looking for ERC20 transfers as well.

rasom:
this one is done

PR #4034, opened 2023-09-15
Basic test for findBlocksCommand

rasom:
needed for https://github.com/status-im/status-go/issues/3971, otherwise a bit difficult to iterate with new changes there 

status: ready

### DIFF
--- a/rpc/chain/client.go
+++ b/rpc/chain/client.go
@@ -25,8 +25,23 @@ type FeeHistory struct {
 	BaseFeePerGas []string `json:"baseFeePerGas"`
 }
 
+type BatchCallClient interface {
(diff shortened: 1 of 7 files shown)


IvanBelyakoff (approved):
LGTM.
Just some minor comments - you replaced some Debug logs with Info. I know that `go test` does not display Debug logs by default (not sure if there is any option to override this), but it displays Info. But generally I think those ones should be only visible when user toggles "debug" toggle in the app. I used a lot of Info when developing some of that code because I was not sure which log level to select, but now I have plans to replace Info with Debug for many cases.

rasom:
@IvanBelyakoff no worries about that, I planed to change them back to `Debug` before merging, just leftovers

PR #4035, opened 2023-09-15
Make 1-1 messages appear as mentions each 2 minutes

jrainville:
Iterates over https://github.com/status-im/status-go/pull/3889

Needed for https://github.com/status-im/status-desktop/issues/10195

Fixes and improves the work done by @alwx 

Makes it so that 1-1 messages count as a mention every 2 minutes.

I didn't add the activity center portion of the issue, because the chat is saved before the AC notification can be created, so we lose the last message timestamp.
It can be done later.

### DIFF
--- a/protocol/chat.go
+++ b/protocol/chat.go
@@ -27,6 +27,13 @@ var chatColors = []string{
 	"#d37ef4", // purple
 }
 
+const (
+	// IncreaseUnviewedMessagesCountTimeoutMs
+	// this timeout indicates how long the time between received messages should be
+	// for a new message to increase the unviewed messages counter
+	IncreaseUnviewedMessagesCountTimeoutMs = 1000 * 60 * 2
+)
+
 type ChatType int
 
 const (
@@ -389,6 +396,14 @@ func (c *Chat) UpdateFromMessage(message *common.Message, timesource common.Time
 		c.LastClockValue = message.Clock
 	}
 	return nil
+}
+
+func (c *Chat) ShouldIncreaseMessageCount(message *common.Message) bool {
+	return c.LastMessage == nil ||
+		c.LastMessage.IsSystemMessage() ||
+		!c.LastMessage.New ||
+		c.LastMessage.Seen ||
+		message.Timestamp > c.LastMessage.Timestamp+IncreaseUnviewedMessagesCountTimeoutMs
 }
 
 func (c *Chat) UpdateFirstMessageTimestamp(timestamp uint32) bool {
--- a/protocol/common/message.go
+++ b/protocol/common/message.go
@@ -635,8 +635,6 @@ func (m *Message) PrepareContent(identity string) error {
 func (m *Message) IsSystemMessage() bool {
 	return m.ContentType == protobuf.ChatMessage_SYSTEM_MESSAGE_CONTENT_PRIVATE_GROUP ||
 		m.ContentType == protobuf.ChatMessage_SYSTEM_MESSAGE_GAP ||
-		m.ContentType == protobuf.ChatMessage_CONTACT_REQUEST ||
-		m.ContentType == protobuf.ChatMessage_IDENTITY_VERIFICATION ||
 		m.ContentType == protobuf.ChatMessage_SYSTEM_MESSAGE_PINNED_MESSAGE ||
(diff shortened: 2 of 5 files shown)


protocol/chat.go
```diff
@@ -391,6 +398,14 @@ func (c *Chat) UpdateFromMessage(message *common.Message, timesource common.Time
 	return nil
 }
 
+func (c *Chat) ShouldIncreaseMessageCount(message *common.Message) bool {
```
jrainville, line 401: I moved the checks in this function so that it can be reused in the AC service later if needed

protocol/messenger.go
```diff
@@ -4092,6 +4092,8 @@ func (m *Messenger) markAllRead(chatID string, clock uint64, shouldBeSynced bool
 	chat.UnviewedMessagesCount = 0
 	chat.UnviewedMentionsCount = 0
 
+	chat.LastMessage.Seen = true
```
jrainville: This fixes the MarkAllRead behaviour to make sure the LastMessage is marked as Seen too

protocol/chat.go
```diff
@@ -391,6 +398,14 @@ func (c *Chat) UpdateFromMessage(message *common.Message, timesource common.Time
...
+	return c.LastMessage == nil ||
+		c.LastMessage.IsSystemMessage() ||
+		!c.LastMessage.New ||
+		c.LastMessage.Seen ||
```
jrainville, line 405: This line was missing in the original PR. Without it, if you saw the last message and went to another chat, the marker wouldn't increase.

protocol/chat.go
```diff
@@ -27,6 +27,13 @@ var chatColors = []string{
 	"#d37ef4", // purple
 }
 
+const (
+	// IncreaseUnviewedMessagesCountTimeout
+	// this timeout indicates how long the time between received messages should be
+	// for a new message to increase the unviewed messages counter
+	IncreaseUnviewedMessagesCountTimeout = 1000 * 60 * 2
```
ilmotta: Maybe better to add the unit to the name, like `...CountTimeoutMs` to avoid any confusion when doing math.

protocol/messenger_handler.go
```diff
@@ -3075,10 +3075,16 @@ func (m *Messenger) isMessageAllowedFrom(publicKey string, chat *Chat) (bool, er
 	return contact.added(), nil
 }
 
-func (m *Messenger) updateUnviewedCounts(chat *Chat, mentionedOrReplied bool) {
-	chat.UnviewedMessagesCount++
-	if mentionedOrReplied {
-		chat.UnviewedMentionsCount++
+func (m *Messenger) updateUnviewedCounts(chat *Chat, message *common.Message) {
+	if chat == nil {
```
ilmotta: If `chat` could be nil, the calls to `updateUnviewedCounts` would fail before reaching this line because they call `chat.ID` a few lines before. So perhaps we can remove this defensive mechanism?

ilmotta (approved):

protocol/common/message.go
```diff
@@ -632,6 +632,17 @@ func (m *Message) PrepareContent(identity string) error {
 	return m.parseAudio()
 }
 
+func (m *Message) IsSystemMessage() bool {
+	return m.ContentType == protobuf.ChatMessage_SYSTEM_MESSAGE_CONTENT_PRIVATE_GROUP ||
+		m.ContentType == protobuf.ChatMessage_SYSTEM_MESSAGE_GAP ||
+		m.ContentType == protobuf.ChatMessage_CONTACT_REQUEST ||
```
MishkaRogachev: Not sure Contact request is needed here, it's more like a plain text message

MishkaRogachev (approved):

protocol/common/message.go
jrainville: Is `ChatMessage_CONTACT_REQUEST` the message that the user wrote when sending the request or another message generated by the system?

MishkaRogachev: It's the message that the user wrote when sending the request, but it can be generated with default text "Please, add me to your contacts"

jrainville: Right. I don't think that qualifies as system message. I'll remove it.

PR #4039, opened 2023-09-18
Hotfix: Don't send members for unpermitted channels

MishkaRogachev:
Hot fix for [this](https://github.com/status-im/status-desktop/issues/12114#issuecomment-1721426909) problem.

Important changes:
- [x] Empty community channels member list, if there is no permissions settled up for that channels when sending a community description message
- [x] Restore community channels member list from community member list if there is no permissions settled up

Actual fix should be implemented here: https://github.com/status-im/status-desktop/issues/12188

### DIFF
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -1231,7 +1231,7 @@ func (o *Community) Description() *protobuf.CommunityDescription {
 }
 
 func (o *Community) marshaledDescription() ([]byte, error) {
-	// Clear members list for each channel what don't have permissions
+	// Clear members list for channels that don't have permissions
 	// TMP: should be fixed in https://github.com/status-im/status-desktop/issues/12188
 	clonedDescritpion := proto.Clone(o.config.CommunityDescription).(*protobuf.CommunityDescription)
 	for chatID, chat := range clonedDescritpion.Chats {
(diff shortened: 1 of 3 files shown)


protocol/communities/community.go
```diff
@@ -1231,7 +1231,16 @@ func (o *Community) Description() *protobuf.CommunityDescription {
 }
 
 func (o *Community) marshaledDescription() ([]byte, error) {
-	return proto.Marshal(o.config.CommunityDescription)
+	// Clear members list for each channel what don't have permissions
```
jrainville:
```suggestion
	// Clear members list for channels that don't have permissions
```

protocol/communities/manager.go
```diff
@@ -1272,6 +1272,14 @@ func (m *Manager) HandleCommunityDescriptionMessage(signer *ecdsa.PublicKey, des
 }
 
 func (m *Manager) handleCommunityDescriptionMessageCommon(community *Community, description *protobuf.CommunityDescription, payload []byte) (*CommunityResponse, error) {
+	// For each channel what don't have permissions, we use community members list instead
```
```suggestion
	// For each channel that doesn't have permissions, we use community members list instead
```

jrainville (approved):

osmaczko (approved):

protocol/communities/community.go

protocol/communities/manager.go

PR #4041, opened 2023-09-18
ci: parametrize DB port for unit tests

### LINKED ISSUE #4040: Remove hard-coded postgres name/port
# Problem

Tests seems to be failing to run in parallel in CI:
```
docker: Error response from daemon: Conflict. The container name "/status-go-test-db" is already in use by container "2723e6e32c65e7a5130fe5ad94d63f3697608879a860dff2f714b71294fcc6dd". You have to remove (or rename) that container to be able to reuse that name.
```

# Implementation

Parametrize:
1. docker container name
https://github.com/status-im/status-go/blob/ea6107b0ba0a508617e38b2a98a715e200354a9c/_assets/ci/Jenkinsfile.tests#L28
2. postgres port
https://github.com/status-im/status-go/blob/ea6107b0ba0a508617e38b2a98a715e200354a9c/postgres/helpers.go#L10-L11

jakubgs:
Otherwise we can't run tests in parallel on the same host. Also the container name has to be different depending on executor.

Resolves errors like:
```
docker: Error response from daemon: Conflict.
    The container name "/status-go-test-db" is already in use by container "123...".
    You have to remove (or rename) that container to be able to reuse that name.
```

Resolves: https://github.com/status-im/status-go/issues/4040

### DIFF
--- a/_assets/ci/Jenkinsfile.tests
+++ b/_assets/ci/Jenkinsfile.tests
@@ -25,7 +25,8 @@ pipeline {
 
   environment {
     TARGET   = 'tests'
-    DB_CONT  = 'status-go-test-db'
+    DB_CONT  = "status-go-test-db-${env.EXECUTOR_NUMBER.toInteger() + 1}"
+    DB_PORT  = "${5432 + env.EXECUTOR_NUMBER.toInteger()}"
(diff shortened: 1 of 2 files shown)


cammellos (approved):

igor-sirotin (approved):
Thank you!

PR #4044, opened 2023-09-19
Prevent Logged In Account Local Pairing / Syncing With Another Account

### LINKED ISSUE #4024: Prevent Logged In Account Local Pairing / Syncing With Another Account.
Currently it is possible to start local pairing / syncing between 2 devices but one device is logged into a different account. See the below process flow:

1. Device 1 logs into Account 1
1. Device 2 logs into Account 2
1. Device 2 Generates sync QR of Account 2
1. Device 1 scans generated Account 2 sync code while logged into Account 1

This process causes the following bug / horrible db coruption https://github.com/status-im/status-mobile/issues/17270

We have a check for this very scenario but for some reason it doesn't seem to prevent this issue.

image

qfrank:
RT.

Closes #4024

relate mobile [issue](https://github.com/status-im/status-mobile/issues/17270)


server/pairing/server.go
```diff
@@ -245,6 +245,7 @@ func MakeFullReceiverServer(backend *api.GethStatusBackend, config *ReceiverServ
 		return nil, err
 	}
 
+	// ignore err because we don't care if there is no active account
```
siddarthkay: hi @qfrank : did you add this comment by mistake?

qfrank: Hi @siddarthkay , i'd say no, it should allow no active account here, sorry my expression confused you, i'll update it to `ignore err because we allow no active account here` ? @siddarthkay WDYT?

server/pairing/sync_device_test.go
```diff
@@ -1167,3 +1167,99 @@ func (s *SyncDeviceSuite) TestTransferringKeystoreFilesAfterStopUisngKeycard() {
...
+			Password:     s.password,
+		},
+		ServerConfig: new(ServerConfig),
+	}
+	configBytes, err := json.Marshal(config)
+	s.NoError(err)
+	cs, err := StartUpSenderServer(serverBackend, string(configBytes))
+	s.NoError(err)
+
+	expectedKDFIterations := 2048
```
siddarthkay: nit: would be good to have expected KDF values for mobile and desktop in some constant :)

siddarthkay (approved):

server/pairing/server.go
siddarthkay: yes all good, I just saw a comment with no code change and assumed you either forgot to change code in this file or added the comment by mistake :D 
Thanks for explaining!

server/pairing/client.go
```diff
@@ -496,6 +496,12 @@ func setupReceivingClient(backend *api.GethStatusBackend, cs, configJSON string)
 		return nil, err
 	}
 
+	// ignore err because we don't care if there is no active account
+	activeAccount, _ := backend.GetActiveAccount()
+	if activeAccount != nil {
+		conf.ReceiverConfig.LoggedInKeyUID = activeAccount.KeyUID
```
saledjenic, line 499: @qfrank as far as I know the only way to create an account on another device, by syncing from the device where that account is already created, is possible only if there is no info about that account on another device at that moment. That means `backend.GetActiveAccount()` will always be `nil`. `KeyUID` on the receiver side will be set during the pairing when the user enters the connection string, that is made on the sender side, on the receiver side.

Not sure why we need to know key uid on the receiver side and set `conf.ReceiverConfig.LoggedInKeyUID` when the user won't be log in ever before the pairing process is done. There is no feature to run pairing/syncing explicitly when the same user is logged in on both devices.

qfrank: @saledjenic it's also possible if there is no info about that account on another device.
you don't have to set `conf.ReceiverConfig.LoggedInKeyUID` in this case in the receiver side, it's optional, it's only needed to be set when user logged in so that we can check if the account we received is different, we don't allow this [case](https://github.com/status-im/status-go/issues/4024) happen ATM.

Samyoul (approved):
Wonderful thank you very much @qfrank

igor-sirotin (approved):
LGTM 👍 
Just waiting for @saledjenic to reply on our question about `prepareBackendWithoutAccount`.

saledjenic:
@igor-sirotin I don't see any question about `prepareBackendWithoutAccount`?

igor-sirotin:
@saledjenic I DMed you

Issue #4029, opened 2023-09-15
Do not send "curated.communities.update" signal if there is no changes


flexsurfer:
"curated.communities.update" signal should be sent only if there are changes

flexsurfer:
cc @vkjr

PR #4046, opened 2023-09-19
chore(messenger_communities): up unknown curated communities timeout

jrainville:
Needed for https://github.com/status-im/status-desktop/issues/12207

The issue is that the curation Dapp is not up yet, so we hard coded the contract with communities.
However, one of the curated communities is very old and not used anymore, so it is no longer broadcasted on Waku.

That means that we enter the "unknown community timeout" loop of 3 seconds. It ends up spamming a little bit.

I upped the timer to 1 minute for now. We can lower it back down when we deploy the real Dapp and contract.

### DIFF
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -499,7 +499,9 @@ func (m *Messenger) startCuratedCommunitiesUpdateLoop() {
 
 	const errorTimeout = 10 * time.Second
 	const successTimeout = 120 * time.Second
-	const unknownCommunitiesFoundTimeout = 3 * time.Second
+	// TODO lower this back again once the real curated community contract is up
+	// The current contract contains communities that are no longer accessible on waku
+	const unknownCommunitiesFoundTimeout = 60 * time.Second
 
 	go func() {
 		for {


protocol/messenger_communities.go
```diff
@@ -470,7 +470,9 @@ func (m *Messenger) startCuratedCommunitiesUpdateLoop() {
 
 	const errorTimeout = 10 * time.Second
 	const successTimeout = 120 * time.Second
-	const unknownCommunitiesFoundTimeout = 3 * time.Second
+	// TODO lower this back again once the real curated community contract is up
+	// The current contract contains communities that are no longer accessible on waku
+	const unknownCommunitiesFoundTimeout = 60 * time.Second
 
 	go func() {
 		for {
```
jrainville, line 507: I can't comment below enough, but maybe we could make the `SendCuratedCommunitiesUpdate` function smarter too. Right now, it sends all the communities every time. Maybe it should only be called when there is an actual update?

osmaczko (approved):
Looks ok to me but please wait for @vkjr approval, maybe the lower timeout was introduced for a reason.

protocol/messenger_communities.go
vkjr: @jrainville, thanks for the changes! Yes, you are absolutely right! There is already an issue to not send unnecessary updates #4029

vkjr (approved):

vkjr:
> Looks ok to me but please wait for @vkjr approval, maybe the lower timeout was introduced for a reason.

Sorry for the late reply. 
No, the low timeout was introduced only because I believed that the community will be fetched anyway, so the quicker we check it again, the quicker mobile app will receive the information. But this is not always true as I know after this PR)

PR #4048, opened 2023-09-19
feat(wallet): add community info to collectibles

dlipicar:
Part of https://github.com/status-im/status-desktop/issues/12155

This adds community information needed to properly follow the designs for how Community collectibles should be displayed on the wallet.

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -4588,53 +4588,62 @@ func (m *Manager) HandleCommunityTokensMetadata(community *Community) error {
 				return err
 			}
 			if !exists {
-				communityToken := &community_token.CommunityToken{
-					CommunityID:        communityID,
-					Address:            address,
-					TokenType:          tokenMetadata.TokenType,
-					Name:               tokenMetadata.Name,
-					Symbol:             tokenMetadata.Symbol,
-					Description:        tokenMetadata.Description,
-					Transferable:       true,
-					RemoteSelfDestruct: false,
-					ChainID:            int(chainID),
-					DeployState:        community_token.Deployed,
-					Base64Image:        tokenMetadata.Image,
-					Decimals:           int(tokenMetadata.Decimals),
-				}
-
-				switch tokenMetadata.TokenType {
-				case protobuf.CommunityTokenType_ERC721:
-					contractData, err := m.communityTokensService.GetCollectibleContractData(chainID, address)
-					if err != nil {
-						return err
-					}
-
-					communityToken.Supply = contractData.TotalSupply
-					communityToken.Transferable = contractData.Transferable
-					communityToken.RemoteSelfDestruct = contractData.RemoteBurnable
-					communityToken.InfiniteSupply = contractData.InfiniteSupply
-
-				case protobuf.CommunityTokenType_ERC20:
-					contractData, err := m.communityTokensService.GetAssetContractData(chainID, address)
-					if err != nil {
-						return err
-					}
-
-					communityToken.Supply = contractData.TotalSupply
-					communityToken.InfiniteSupply = contractData.InfiniteSupply
-				}
-
-				communityToken.PrivilegesLevel = getPrivilegesLevel(chainID, address, community.TokenPermissions())
-
-				err = m.persistence.AddCommunityToken(communityToken)
+				// Fetch community token to make sure it's stored in the DB, discard result
+				communityToken, err := m.FetchCommunityToken(community, tokenMetadata, chainID, address)
 				if err != nil {
 					return err
 				}
+
+				return m.persistence.AddCommunityToken(communityToken)
 			}
 		}
 	}
 	return nil
+}
+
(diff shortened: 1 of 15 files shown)


protocol/communities/manager.go
```diff
@@ -4588,53 +4588,62 @@ func (m *Manager) HandleCommunityTokensMetadata(community *Community) error {
...
+	communityToken := &community_token.CommunityToken{
+		CommunityID:        communityID,
+		Address:            contractAddress,
+		TokenType:          tokenMetadata.TokenType,
+		Name:               tokenMetadata.Name,
+		Symbol:             tokenMetadata.Symbol,
+		Description:        tokenMetadata.Description,
+		Transferable:       true,
+		RemoteSelfDestruct: false,
+		ChainID:            int(chainID),
```
Cuteivist, line 4616: In wallet we use `type ChainID uint64`. Maybe `uint64` should be used here too?

services/ext/service.go
```diff
@@ -556,46 +560,223 @@ func (s *Service) CanProvideCollectibleMetadata(id thirdparty.CollectibleUniqueI
...
+}
+
+func (s *Service) fetchCommunityCollectiblePermission(community *communities.Community, id thirdparty.CollectibleUniqueID) (*communities.CommunityTokenPermission, error) {
+	permissionTypes := []protobuf.CommunityTokenPermission_Type{
+		protobuf.CommunityTokenPermission_BECOME_TOKEN_OWNER,
+		protobuf.CommunityTokenPermission_BECOME_TOKEN_MASTER,
+	}
+
+	// Check permission types of interest
+	for _, permissionType := range permissionTypes {
```
Cuteivist, line 751: Generally I don't mind it, but here is really extreme example of arrow anti-pattern. Do you think it could be reimplemented in more readable form? Wdyt @dlipicar  ?

services/ext/service.go
```diff
@@ -556,46 +560,223 @@ func (s *Service) CanProvideCollectibleMetadata(id thirdparty.CollectibleUniqueI
...
+					}
+				}
 			}
 		}
 	}
 
 	return nil, nil
 }
+
+func boolToString(value bool) string {
```
Cuteivist, line 784: Wouldn't it be easier to parse `0` and `1` instead of `Yes` and `No` ?

Cuteivist (approved):
Look good. Awesome job.

Leaving some minor points to think about before merging,.

endulab (approved):

services/ext/service.go
dlipicar: Traits are generic string pairs (Title + Value) that do not follow any kind of standard. We don't do any parsing on the client side, we just display them as they are. If we want this to look like in Figma, we need to set it in its text form.

Cuteivist: I don't know the design you are referring, but do you mean that this `Yes` and `No` is directly displayed in the UI ?
If that's the case then any translations won't work on those strings.

services/ext/service.go
dlipicar: Done

services/ext/service.go
dlipicar: Hmm we don't translate traits from other collectibles, not sure if we want to handle these in a special way. I'll ask Ben!

services/ext/service.go
Cuteivist: Great job!

protocol/communities/manager.go
dlipicar: messenger folks have their own data types... Trying to use ours wherever reasonable, but I'd rather not change theirs

services/ext/service.go
IvanBelyakoff: Yeah, agree with @Cuteivist that it'd make sense to pass a boolean to UI and handle the interpretation of the value on client side. That would also simplify the logic on client side if we need to use those flags in code  - checking for ` === "Yes"` and be sure that it is not "yes", "YES", or we need to always call `toLowercase()` instead of just `isTransferable` enforces client to know more about the passed data format and changing those strings will require changing both sides.
Unless there are caveats that I am not aware of

IvanBelyakoff (approved):
Great job.
If you don't have any good reason not to pass flags as boolean rather than strings in collectible traits, fixing this would be nice.

services/ext/service.go
dlipicar: So... My logic here was to try to make this work as much as possible as a standard collectible and not require a bunch of special logic on the client side. 

These traits are just cosmetic elements, we're not supposed to use them for any business logic. If we needed to use the data for some functionality, I'd transfer it in some different struct.
Moreover, we're not translating these strings just like we don't translate trait types/values for other collectibles. I discussed this with Ben and he said:
> It feels practical at least for MVP that we opt to keep it in English just as Opensea does, even if the user changes their language setting. I think the fact NFT communities often have their own quirky terminology and language for traits makes it more ok for us to avoid devling into translation - examples attached - eg skelly cig, puffer up

I'm gonna keep it as-is for now since there's some other work that depends on this, we can make further adjustments down the road.

PR #4050, opened 2023-09-20
Fetch erc20 history tail

### LINKED ISSUE #3930: Almost proper scan of transfers history tail
# Problem

The current implementation of `SequentialFetchStartegy` stops making `eth_getLogs` requests as soon as the oldest account's ETH transfer is found. The reason is that the history's tail might be millions of blocks and `etc_getLogs` range is limited (by API providers like POKT/Infura) to 10k blocks sometimes, thus we will likely make a lot of heavy RPC calls for nothing. There is an alternative way to scan that tail for ERC20 transfers.

# Implementation

Instead of scanning tail with `eth_getLogs` requests we only need to check whether balances for a set of ERC20 tokens at the block which precedes the one with the oldest ETH transfer. If those balances are zero no other actions are required. If some of them are non-zero we need to make a binary search of blocks with balance change for those specific tokens. 

The mentioned set of ERC20 contains all tokens with non-zero balance (in the latest block) per [`GetTokensBalancesForChainIDs`](https://github.com/status-im/status-go/blob/develop/services/wallet/token/tokenstore.go#L55) ([for all tokens we have in the list, this is already checked anyway](https://github.com/status-im/status-go/blob/develop/services/wallet/token/tokenstore.go#L55)) along with any other ERC20 which we find to be used by account after the oldest block with ETH transfer. If there are no such tokens we can skip checking the tail for transfers at all.

# Notes

- it would be nice to figure out whether we can use `[BalanceScanner`](https://github.com/status-im/status-go/blob/develop/contracts/ethscan/ethscan.go#L285) for historical data in order to minimise number of requests.
- for binary search there is no need to start with zero block, the minimum one to check is a block of contract creation

rasom:
fix (initially) #3930 

Problem: we do not check incoming ERC20 transfers in case If we already reached the first ETH transfer while checking history.

The result is that some account will be missing ERC20 transfers in history (for instance I found one by randomly checking ERC20 transfers in 2 mins. Either I was lucky or that's not such a rare case. https://etherscan.io/address/0x1f01d365c17112f9574683e4318d2625231a9a13#tokentxns) 

The reason for not checking ERC20 tail via `eth_getLogs` right away is that those requests are slow, heavy for RPC API providers (as a result for some networks we are limited to a range like 10K blocks per request) and also we have to make quite a lot of requests to check empty or almost empty account (with only relatively recent transaction). With 18M blocks and 100K blocks per range (that's how it works in older version of the algorithm) it would cost 18M/10K*2=360 requests.
**In most cases though that ERC20 tail is going to be absent and we will likely confirm this with via single RPC request.** But when the tail exist we have to do some extra work.

Algorithm:
1. When the first ETH transfer is found we check ERC20 balances for known tokens (we have a list, at the moment it is limited to that list) before this block.
2. if there are non zero balances we do binary search of the range which is less or equal than the range limit for `eth_getLogs` call (currently 500K on [mainnet](https://github.com/status-im/status-go/blob/feature%2F%233930-erc20-history-tail/services/wallet/transfer/commands.go#L48) ).
3. Then we make two `etc_getLogs` in order to find transfers

18M blocks to check and a single ERC20 transfers somewhere in history:
1. old implementation would make **18M/100K*2=360  `etc_getLogs` requests** and find that transfer
2. new implementation would make **2 `etc_getLogs`** request and find that transfer only if it happened in that latest 100K range
3. this PRs implementation would make **2+1+(1+log2(18M/500K))+2=~12 request (for of them `eth_getLogs`), rest contract calls** (2 `eth_getLogs` on initial scan, single contact call to get balance for all known tokens, 1+log2(18M/500K)=~6 contract calls to narrow the range, 2 `eth_getLogs` calls to find blocks)

TODO in separate PRs:
1. Make sure we merge ranges for different tokens when those intersect, to avoid fetching transactions twice
4. Figure out whether we still need to check outgoing ERC20 in those ranges
5. In case if there are many tokens in the tail it might be cheaper to execute eth_getLogs, depends on network (range limits) and number of tokens. 

status: ready

### DIFF
--- a/rpc/client.go
+++ b/rpc/client.go
@@ -48,9 +48,9 @@ type Client struct {
 	UpstreamChainID uint64
 
 	local           *gethrpc.Client
-	upstream        *chain.ClientWithFallback
+	upstream        chain.ClientInterface
 	rpcClientsMutex sync.RWMutex
-	rpcClients      map[uint64]*chain.ClientWithFallback
+	rpcClients      map[uint64]chain.ClientInterface
 
 	router         *router
 	NetworkManager *network.Manager
@@ -84,7 +84,7 @@ func NewClient(client *gethrpc.Client, upstreamChainID uint64, upstream params.U
 		local:          client,
 		NetworkManager: networkManager,
 		handlers:       make(map[string]Handler),
-		rpcClients:     make(map[uint64]*chain.ClientWithFallback),
+		rpcClients:     make(map[uint64]chain.ClientInterface),
 		log:            log,
 	}
 
@@ -112,12 +112,12 @@ func (c *Client) SetWalletNotifier(notifier func(chainID uint64, message string)
 	c.walletNotifier = notifier
 }
 
-func (c *Client) getClientUsingCache(chainID uint64) (*chain.ClientWithFallback, error) {
+func (c *Client) getClientUsingCache(chainID uint64) (chain.ClientInterface, error) {
 	c.rpcClientsMutex.Lock()
 	defer c.rpcClientsMutex.Unlock()
 	if rpcClient, ok := c.rpcClients[chainID]; ok {
-		if rpcClient.WalletNotifier == nil {
-			rpcClient.WalletNotifier = c.walletNotifier
+		if rpcClient.GetWalletNotifier() == nil {
+			rpcClient.SetWalletNotifier(c.walletNotifier)
 		}
 		return rpcClient, nil
 	}
@@ -150,7 +150,7 @@ func (c *Client) getClientUsingCache(chainID uint64) (*chain.ClientWithFallback,
 }
 
 // Ethclient returns ethclient.Client per chain
-func (c *Client) EthClient(chainID uint64) (*chain.ClientWithFallback, error) {
+func (c *Client) EthClient(chainID uint64) (chain.ClientInterface, error) {
 	client, err := c.getClientUsingCache(chainID)
 	if err != nil {
 		return nil, err
@@ -169,8 +169,8 @@ func (c *Client) AbstractEthClient(chainID common.ChainID) (chain.BatchCallClien
 	return client, nil
 }
 
-func (c *Client) EthClients(chainIDs []uint64) (map[uint64]*chain.ClientWithFallback, error) {
-	clients := make(map[uint64]*chain.ClientWithFallback, 0)
+func (c *Client) EthClients(chainIDs []uint64) (map[uint64]chain.ClientInterface, error) {
+	clients := make(map[uint64]chain.ClientInterface, 0)
 	for _, chainID := range chainIDs {
 		client, err := c.getClientUsingCache(chainID)
 		if err != nil {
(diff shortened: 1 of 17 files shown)


rpc/client.go
```diff
@@ -182,6 +182,13 @@ func (c *Client) EthClients(chainIDs []uint64) (map[uint64]*chain.ClientWithFall
 	return clients, nil
 }
 
+// SetClient strictly for testing purposes
```
IvanBelyakoff, line 185: Do you think we could move this method to client_test.go instead? Then noone would be tempted to use it in production.

services/wallet/token/token.go
```diff
@@ -330,7 +331,7 @@ func (tm *Manager) GetAllTokens() ([]*Token, error) {
 
 	tokens, err := tm.GetCustoms()
 	if err != nil {
-		log.Error("can't fetch custom tokens", "error", err)
+		log.Error("can't fetch custom tokens 1", "error", err)
```
IvanBelyakoff: leftover numbers here and later?

services/wallet/transfer/reactor.go
```diff
@@ -291,17 +291,17 @@ func (r *Reactor) stop() {
...
 	r.stop()
 	return r.start(chainClients, accounts, loadAllTransfers)
 }
 
-func (r *Reactor) createFetchStrategy(chainClients map[uint64]*chain.ClientWithFallback,
+func (r *Reactor) createFetchStrategy(chainClients map[uint64]chain.ClientInterface,
 	accounts []common.Address, loadAllTransfers bool) HistoryFetcher {
 
-	if loadAllTransfers {
+	if true /*loadAllTransfers*/ {
```
IvanBelyakoff: revert this?

IvanBelyakoff (approved):
Awesome work, Roman! Appreciate you wrote the tests for sequential strategy and fixing it.
I have only minor comments, which you probably was going to fix before converting the draft PR to a normal one.

Regarding point 2) Outgoing ERC20 transfers should change the nonce. So I believe once we reached the first ETH transfer, there wont be any more outgoing ERC20 transfers farther back.

rpc/client.go
rasom: but I it will be available outside test even in this case, no?

services/wallet/transfer/reactor.go
rasom: yep, that was for testing on mobile, thanks

services/wallet/token/token.go
rasom: yeah kind of but also it would be nice to have different message as sometimes it makes it difficult to debug from logs. I will remove those though.

rpc/client.go
IvanBelyakoff: Yes, it will, just noone would directly see it in client.go. And those who find it in client_test.go would know it is meant for testing. Up to you though, just thought it'd be good to separate this method even more

services/wallet/token/token.go
IvanBelyakoff: I did not notice the same error messages. Fine with me.

rasom: no that's a different message actually, so I'm mixing things. Should be fixed

PR #4054, opened 2023-09-21
fix: Don't store chat members in the community description

MishkaRogachev:
This PR should fix community channels members list:
https://github.com/status-im/status-desktop/issues/12114#issuecomment-1726335732

### DIFF
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -155,7 +155,7 @@ func (o *Community) MarshalPublicAPIJSON() ([]byte, error) {
 				Emoji:       c.Identity.Emoji,
 				Description: c.Identity.Description,
 				Permissions: c.Permissions,
-				Members:     c.Members,
+				Members:     o.getChatMembers(id),
 				CanPost:     canPost,
 				CategoryID:  c.CategoryId,
 				Position:    int(c.Position),
@@ -275,7 +275,7 @@ func (o *Community) MarshalJSON() ([]byte, error) {
 				Color:       c.Identity.Color,
 				Description: c.Identity.Description,
 				Permissions: c.Permissions,
-				Members:     c.Members,
+				Members:     o.getChatMembers(id),
 				CanPost:     canPost,
 				CategoryID:  c.CategoryId,
 				Position:    int(c.Position),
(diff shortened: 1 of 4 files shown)


protocol/communities/community.go
```diff
@@ -1256,7 +1266,16 @@ func (o *Community) Description() *protobuf.CommunityDescription {
...
+	// Clear members list for channels that don't have permissions
+	// TMP: should be fixed in https://github.com/status-im/status-desktop/issues/12188
+	clonedDescritpion := proto.Clone(o.config.CommunityDescription).(*protobuf.CommunityDescription)
+	for chatID, chat := range clonedDescritpion.Chats {
+		if !CheckIfChannelHasAnyPermissions(chatID, clonedDescritpion) {
+			chat.Members = map[string]*protobuf.CommunityMember{}
+		}
+	}
+
+	return proto.Marshal(clonedDescritpion)
```
osmaczko: This shouldn't be needed anymore.

protocol/communities/community.go
```diff
@@ -658,7 +668,7 @@ func (o *Community) RemoveUserFromChat(pk *ecdsa.PublicKey, chatID string) (*pro
 	if !o.IsControlNode() {
 		return nil, ErrNotControlNode
 	}
-	if !o.hasMember(pk) {
+	if !o.hasMember(pk) || !CheckIfChannelHasAnyPermissions(chatID, o.config.CommunityDescription) {
```
osmaczko: This seems to be a wrong place for this condition. We should simply check if `chat.Members` contains the pk.

protocol/communities/utils.go
```diff
@@ -57,3 +59,12 @@ func ExtractTokenCriteria(permissions []*CommunityTokenPermission) (erc20TokenCr
 	}
 	return
 }
+
+func CheckIfChannelHasAnyPermissions(chatID string, description *protobuf.CommunityDescription) bool {
```
osmaczko: There is `ChannelHasTokenPermissions` already defined.

protocol/communities/community.go
```diff
@@ -1874,7 +1897,10 @@ func (o *Community) PopulateChatWithAllMembers(chatID string) (*CommunityChanges
 		return o.emptyCommunityChanges(), ErrNotControlNode
 	}
 
-	return o.populateChatWithAllMembers(chatID)
+	if CheckIfChannelHasAnyPermissions(chatID, o.config.CommunityDescription) {
```
osmaczko: `PopulateChatWithAllMembers` is not needed anymore, instead we should remove all members of a channel when permission is removed from it (see the usage of `PopulateChatWithAllMembers`).

osmaczko (changes requested):

PR #4055, opened 2023-09-22
fix: Prevent (reject event <-> resend event) loop

mprakhov:
Prevent (reject event <-> resend event) loop

Send in COMMUNITY_EVENTS_MESSAGE_REJECTE the last actual CommunityDescription and prevent sending by privileged users new event if CommunityDescription differs on the control node client and on the privileged member client

Closes #https://github.com/status-im/status-desktop/issues/11997

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -1425,6 +1425,8 @@ func (m *Manager) HandleCommunityEventsMessage(signer *ecdsa.PublicKey, message 
 	err = community.UpdateCommunityByEvents(eventsMessage)
 	if err != nil {
 		if err == ErrInvalidCommunityEventClock && community.IsControlNode() {
+			// send updated CommunityDescription to the event sender on top of which he must apply his changes
+			eventsMessage.EventsBaseCommunityDescription = community.config.CommunityDescriptionProtocolMessage
 			m.publish(&Subscription{
 				CommunityEventsMessageInvalidClock: &CommunityEventsMessageInvalidClockSignal{
 					Community:              community,
@@ -1491,6 +1493,17 @@ func (m *Manager) HandleCommunityEventsMessageRejected(signer *ecdsa.PublicKey, 
 	if err != nil {
 		return nil, err
 	}
+
+	communityDescription, err := validateAndGetEventsMessageCommunityDescription(eventsMessage.EventsBaseCommunityDescription, signer)
+	if err != nil {
+		return nil, err
+	}
+	// the privileged member did not receive updated CommunityDescription so his events
+	// will be send on top of outdated CommunityDescription
+	if communityDescription.Clock != community.Clock() {
+		return nil, errors.New("resend rejected community events aborted, client node has outdated community description")
+	}
+
 	eventsMessage.Events = m.validateAndFilterEvents(community, eventsMessage.Events)
 
 	myRejectedEvents := make([]CommunityEvent, 0)
@@ -4812,4 +4825,8 @@ func (m *Manager) shareAcceptedRequestToJoinWithPrivilegedMembers(community *Com
 
 func (m *Manager) GetCommunityRequestsToJoinWithRevealedAddresses(communityID types.HexBytes) ([]*RequestToJoin, error) {
 	return m.persistence.GetCommunityRequestsToJoinWithRevealedAddresses(communityID)
+}
+
+func (m *Manager) SaveCommunity(community *Community) error {
+	return m.persistence.SaveCommunity(community)
 }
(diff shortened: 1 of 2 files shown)


protocol/communities/manager.go
```diff
@@ -1425,6 +1425,8 @@ func (m *Manager) HandleCommunityEventsMessage(signer *ecdsa.PublicKey, message
 	err = community.UpdateCommunityByEvents(eventsMessage)
 	if err != nil {
 		if err == ErrInvalidCommunityEventClock && community.IsControlNode() {
```
MishkaRogachev, line 1427: To clarify: we return `nil, err` even for `ErrInvalidCommunityEventClock`?

protocol/communities/manager.go
```diff
@@ -4813,3 +4826,7 @@ func (m *Manager) shareAcceptedRequestToJoinWithPrivilegedMembers(community *Com
 func (m *Manager) GetCommunityRequestsToJoinWithRevealedAddresses(communityID types.HexBytes) ([]*RequestToJoin, error) {
 	return m.persistence.GetCommunityRequestsToJoinWithRevealedAddresses(communityID)
 }
+
+func (m *Manager) SaveCommunity(community *Community) error {
```
MishkaRogachev, line 4830: Minor thing, but i would be better to move it closer to other persistence wrappers: `SetPrivateKey`, `SaveCommunitySettings` and etc

MishkaRogachev (approved):

protocol/communities/manager.go
mprakhov: yes, we should not allow to process further as CommunityDescription is outdated

protocol/communities/manager.go
mprakhov: I think we just need to split manager.go :)

protocol/communities/manager.go
```diff
@@ -1491,6 +1493,17 @@ func (m *Manager) HandleCommunityEventsMessageRejected(signer *ecdsa.PublicKey,
...
 	}
+
+	communityDescription, err := validateAndGetEventsMessageCommunityDescription(eventsMessage.EventsBaseCommunityDescription, signer)
+	if err != nil {
+		return nil, err
+	}
+	// the privileged member did not receive updated CommunityDescription so his events
+	// will be send on top of outdated CommunityDescription
+	if communityDescription.Clock != community.Clock() {
+		return nil, err
```
jrainville: `err` is `nil`. Is that what you want?

jrainville (commented):
Looks good, I just have one question in line

protocol/communities/manager.go
mprakhov: Good catch, added error

jrainville (approved):

osmaczko (commented):
Late to the party, but I just wanted to say that this is a very nice PR.

I have one concern: if admins don't re-apply rejected events due to an outdated community description, how or when are they re-applied? Is it acceptable if they are never re-applied?

mprakhov:
From my old discussion with @cammellos I remember that it should be okay if the event is rejected due to outdated community description. Control node and admins should have the same CommunityDescripton (the same for clients)

osmaczko:
> From my old discussion with @cammellos I remember that it should be okay if the event is rejected due to outdated community description. Control node and admins should have the same CommunityDescripton (the same for clients)

In the case admin has outdated CommunityDescripton, then events once rejected will be lost forever. Consider following scenario:
- control node is offline
- adminA does changes A,B,C
- adminB does changes D,E,F
- adminC does changed G,H,J
- control node comes online and accepts changes A,B,C, then propagates updated CommunityDescription
- control node rejects  D,E,F changes
- control node rejects G,H,J changes
- adminB receives the events rejected message and does nothing, changes D,E,F are lost forever
- adminC receives the events rejected message and does nothing, changes G,H,J are lost forever

In other words, the rejection mechanism does not make sense anymore because admins are not re-sending their events. Am I missing something?

mprakhov:
I just follow the rule, that community events can't be applied to outdated CommunityDescription, also they will be rejected by clients, who have new CommunityDescription (the same by the Control node), so in your case - they will be lost if admin AdminB and AdminC will not have updated CommunityDescription

If they have (a control node publishing an updated community description and it will do it after accepting AdminA events) - they will do updates on top of the CommunityDescripton with AdminA events, confirmed by the control node

Also, your case means, that AdminA did not receive events from AdminB and AdminC, and the same for other admins (they did not receive events from other admins)

I'm open to a discussion about how you see it should be, we touch on this question each time we are trying to figure out how it must work and still, we do not have any rules/patterns, so as I told before, I follow the rules: community events can't be applied to outdated CommunityDescription

osmaczko:
> I just follow the rule, that community events can't be applied to outdated CommunityDescription, also they will be rejected by clients, who have new CommunityDescription (the same by the Control node), 

That's ok. The whole idea was to re-apply the events on updated CommunityDescription. The rule you mentioned wouldn't be broken.

> Also, your case means, that AdminA did not receive events from AdminB and AdminC, and the same for other admins (they did not receive events from other admins)

You are right, it was rather an edge case. How many events will be lost depends really on what control node processes first.

Let's observe how the app behaves with current solution and adapt if needed in the future.

PR #4056, opened 2023-09-22
feat(messenger_communities): only send updated curated communities

### LINKED ISSUE #4029: Do not send "curated.communities.update" signal if there is no changes

jrainville:
Fixes #4029

### DIFF
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"reflect"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


vkjr (approved):

protocol/messenger_communities.go
```diff
@@ -503,16 +504,38 @@ func (m *Messenger) startCuratedCommunitiesUpdateLoop() {
...
+					!reflect.DeepEqual(c.UnknownCommunities, response.UnknownCommunities) {
+					// One of the communities is different, send the updated response
+					m.config.messengerSignalsHandler.SendCuratedCommunitiesUpdate(response)
+
+					// Update the values
+					mu.Lock()
+					c.ContractCommunities = response.ContractCommunities
+					c.ContractFeaturedCommunities = response.ContractFeaturedCommunities
+					c.UnknownCommunities = response.UnknownCommunities
+					mu.Unlock()
```
osmaczko: The check-and-update mechanism for the `c` variable is not atomic, which might lead to race conditions if there are other parts of the code concurrently reading/writing to `c`. You're using sync.RWMutex, but you are only locking for the update part.

jrainville: I only put it around the update part since `RWMutex` enables multiple readers, but just to not take chances, I put the lock around the read as well now.

osmaczko (approved):

PR #4059, opened 2023-09-25
Revert "Optimized counter increment logic for 1-1 chat messages"

alwx:
Reverts status-im/status-go#3889

### DIFF
--- a/protocol/common/message.go
+++ b/protocol/common/message.go
@@ -632,17 +632,6 @@ func (m *Message) PrepareContent(identity string) error {
 	return m.parseAudio()
 }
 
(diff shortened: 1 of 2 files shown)


vitvly:
I was able to add contacts and send messages between them using status-mobile pointing to this PR.

vitvly (approved):

vitvly:
However, this doesn't seem to be the root cause of issues here https://github.com/status-im/status-mobile/pull/17344#issuecomment-1733537072, so most likely this particular change (counter increment logic) is not the culprit.

PR #4064, opened 2023-09-26
fix(messenger_communities): block messages and reactions to token gated or spectated communities

### LINKED ISSUE #11915: Able to send reaction in token gated communities
Members not yet approved in the communities are able to send reaction to message

Expected: they should not be able to react

shinnok:
Right now, if the QML UI control checks are disabled, the user can send messages to spectated, non-joined and, possibly, token-gated communities with _invalid tokens_?

Right now the PR is in draft stage with possible code culprits.

---
Closes https://github.com/status-im/status-desktop/issues/11915

### DIFF
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -1722,72 +1722,54 @@ func (o *Community) VerifyGrantSignature(data []byte) (*protobuf.Grant, error) {
 
 func (o *Community) CanPost(pk *ecdsa.PublicKey, chatID string, grantBytes []byte) (bool, error) {
 	if o.config.CommunityDescription.Chats == nil {
-		o.config.Logger.Debug("canPost, no-chats")
+		o.config.Logger.Debug("Community.CanPost: no-chats")
 		return false, nil
 	}
 
 	chat, ok := o.config.CommunityDescription.Chats[chatID]
 	if !ok {
-		o.config.Logger.Debug("canPost, no chat with id", zap.String("chat-id", chatID))
+		o.config.Logger.Debug("Community.CanPost: no chat with id", zap.String("chat-id", chatID))
 		return false, nil
 	}
 
-	// creator can always post
+	// community creator can always post, return immediately
 	if common.IsPubKeyEqual(pk, o.config.ID) {
 		return true, nil
 	}
 
-	// if banned cannot post
 	if o.isBanned(pk) {
+		o.config.Logger.Debug("Community.CanPost: user is banned", zap.String("chat-id", chatID))
 		return false, nil
 	}
 
-	// If both the chat & the org have no permissions, the user is allowed to post
-	if o.config.CommunityDescription.Permissions.Access == protobuf.CommunityPermissions_NO_MEMBERSHIP && chat.Permissions.Access == protobuf.CommunityPermissions_NO_MEMBERSHIP {
-		return true, nil
+	if o.config.CommunityDescription.Members == nil {
+		o.config.Logger.Debug("Community.CanPost: no members in org", zap.String("chat-id", chatID))
+		return false, nil
 	}
 
-	if chat.Permissions.Access != protobuf.CommunityPermissions_NO_MEMBERSHIP {
-		if chat.Members == nil {
-			o.config.Logger.Debug("canPost, no members in chat", zap.String("chat-id", chatID))
+	// If community member, also check chat membership next
+	_, ok = o.config.CommunityDescription.Members[common.PubkeyToHex(pk)]
+	if !ok {
+		o.config.Logger.Debug("Community.CanPost: not a community member", zap.String("chat-id", chatID))
+		return false, nil
+	}
+
+	if chat.Members == nil {
+		o.config.Logger.Debug("Community.CanPost: no members in chat", zap.String("chat-id", chatID))
+		return false, nil
+	}
+
+	// Need to also be a chat member to post
+	if !o.IsMemberInChat(pk, chatID) {
+		if grantBytes == nil {
+			o.config.Logger.Debug("Community.CanPost: not a chat member:", zap.String("chat-id", chatID))
 			return false, nil
 		}
-
-		_, ok := chat.Members[common.PubkeyToHex(pk)]
-		// If member, we stop here
-		if ok {
-			return true, nil
-		}
-
-		// If not a member, and not grant, we return
-		if !ok && grantBytes == nil {
-			o.config.Logger.Debug("canPost, not a member in chat", zap.String("chat-id", chatID))
-			return false, nil
-		}
-
-		// Otherwise we verify the grant
 		return o.canPostWithGrant(pk, chatID, grantBytes)
 	}
 
-	// Chat has no membership, check org permissions
-	if o.config.CommunityDescription.Members == nil {
-		o.config.Logger.Debug("canPost, no members in org", zap.String("chat-id", chatID))
-		return false, nil
-	}
-
-	// If member, they can post
-	_, ok = o.config.CommunityDescription.Members[common.PubkeyToHex(pk)]
-	if ok {
-		return true, nil
-	}
-
-	// Not a member and no grant, can't post
-	if !ok && grantBytes == nil {
-		o.config.Logger.Debug("canPost, not a member in org", zap.String("chat-id", chatID), zap.String("pubkey", common.PubkeyToHex(pk)))
-		return false, nil
-	}
-
-	return o.canPostWithGrant(pk, chatID, grantBytes)
+	// all conditions satisfied, user can post after all
+	return true, nil
 }
 
 func (o *Community) canPostWithGrant(pk *ecdsa.PublicKey, chatID string, grantBytes []byte) (bool, error) {
--- a/protocol/communities/community_test.go
+++ b/protocol/communities/community_test.go
@@ -511,7 +511,7 @@ func (s *CommunitySuite) TestCanPost() {
 			name:    "no-membership org with no-membeship chat",
 			config:  s.configNoMembershipOrgNoMembershipChat(),
 			member:  notMember,
-			canPost: true,
+			canPost: false,
 		},
 		{
 			name:    "no-membership org with invitation only chat-not-a-member",
@@ -529,7 +529,7 @@ func (s *CommunitySuite) TestCanPost() {
 			name:    "no-membership org with invitation only chat-not-a-member valid grant",
 			config:  s.configNoMembershipOrgInvitationOnlyChat(),
 			member:  notMember,
-			canPost: true,
+			canPost: false,
 			grant:   validGrant,
 		},
 		{
@@ -555,7 +555,7 @@ func (s *CommunitySuite) TestCanPost() {
 			name:    "membership org with no-membership chat  not-a-member valid grant",
 			config:  s.configOnRequestOrgNoMembershipChat(),
 			member:  notMember,
-			canPost: true,
+			canPost: false,
 			grant:   validGrant,
 		},
 		{
--- a/protocol/communities_messenger_test.go
+++ b/protocol/communities_messenger_test.go
@@ -442,18 +442,29 @@ func (s *MessengerCommunitiesSuite) TestCommunityContactCodeAdvertisement() {
 func (s *MessengerCommunitiesSuite) TestPostToCommunityChat() {
 	community, chat := s.createCommunity()
 
-	s.advertiseCommunityTo(community, s.owner, s.alice)
-	s.joinCommunity(community, s.owner, s.alice)
-
-	ctx := context.Background()
-
 	chatID := chat.ID
 	inputMessage := common.NewMessage()
 	inputMessage.ChatId = chatID
 	inputMessage.ContentType = protobuf.ChatMessage_TEXT_PLAIN
 	inputMessage.Text = "some text"
 
+	ctx := context.Background()
+
+	s.advertiseCommunityTo(community, s.owner, s.alice)
+
+	// Send message without even spectating fails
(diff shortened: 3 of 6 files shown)


protocol/communities/community.go
```diff
@@ -902,6 +902,7 @@ func (o *Community) Leave() {
 
 func (o *Community) Spectate() {
 	o.config.Spectated = true
+	o.config.Joined = false // this doesn't seem to fix anything, but looks proper to do
```
jrainville: This is probably not needed. It defaults to false. We can keep it or not.

protocol/communities/community.go
```diff
@@ -1605,6 +1606,11 @@ func (o *Community) CanPost(pk *ecdsa.PublicKey, chatID string, grantBytes []byt
 		return false, nil
 	}
 
+	// this resolves the issue, however, I'm not sure how it should play with grants and tokens below
+	if !o.isMember() {
```
jrainville: That looks right. We don't need to check the tokens or whatever if this condition doesn't pass. If someone is not a member, they cannot post.

protocol/messenger.go
```diff
@@ -2218,13 +2218,18 @@ func (m *Messenger) sendChatMessage(ctx context.Context, message *common.Message
...
 		}
+	} else if len(message.AudioPath) != 0 {
+		err := message.LoadAudio()
+		if err != nil {
+			return nil, err
+		}
+	}
 
-	} else if len(message.CommunityID) != 0 {
+	if len(message.CommunityID) != 0 {
```
jrainville: I think this code path is only to send an invitation to the community. So it's probably not the culprit

jrainville (commented):
The line in `CanPost` seems like the right culprit for the Sender side.

We should also do a similar check on the receiver side.
The entry point for that is in `messenger_handlers.go` > `dispatchToHandler`. 
That's where we dispatch the received message to the right handler. 
In this case, it would be `handleChatMessageProtobuf`.

Pro tip: using `fmt.Println("...")`  is a good way to add logs to status-go that will show up in the console when running Status Desktop.

But even better is using the Go tests. You'll find a lot of useful tests in `community_test.go`. For example `TestCanPost`. `TestValidateRequestToJoin` might be a good one to validate that messages only work for those joined, but maybe it's better to create a new test case.

protocol/communities/community.go
shinnok: Reverted.

protocol/messenger.go
shinnok: I tried to split the logic here a bit, so it's more obvious what we're checking for. Will keep this bit.

shinnok:
I can't get the `SendPinMessage()` call to go through even though I have `PinMessageAllMembersEnabled` set to **true**.

[`39789d9` (#4064)](https://github.com/status-im/status-go/pull/4064/commits/39789d9a80d212ccd611835a20ece38e61da1704#diff-d4f4b1d3dcfd1f9dda7f6d7a397650c30b499042a7770a1df9a6fe80d61f04feR441)

https://github.com/status-im/status-go/blob/fix/11915/send-message-to-gated-community/protocol/messenger_pin_messages.go#L42

Not sure how critical it is to get the `TestPinMessageInCommunityChat` working for this particular issue, thus I'm enabling this PR for review and also welcome to help on investigating why isn't the community message pin working in the test suite, when it is fine otherwise.

protocol/communities_messenger_test.go
```diff
@@ -436,12 +436,37 @@ func (s *MessengerCommunitiesSuite) TestCommunityContactCodeAdvertisement() {
 	s.Require().NoError(err)
 }
 
-func (s *MessengerCommunitiesSuite) TestPostToCommunityChat() {
+func (s *MessengerCommunitiesSuite) TestPinMessageInCommunityChat() {
 	community, chat := s.createCommunity()
+	community.AdminSettings().PinMessageAllMembersEnabled = true // FIXME: this doesn't work
 
 	s.advertiseCommunityTo(community, s.alice)
+	s.alice.SpectateCommunity(community.ID()) // TODO: also test in stages like TestPostToCommunityChat
```
osmaczko: This line seems unnecessary, there is no strict requirement to spectate before joining.

protocol/communities_messenger_test.go
```diff
@@ -436,12 +436,37 @@ func (s *MessengerCommunitiesSuite) TestCommunityContactCodeAdvertisement() {
 	s.Require().NoError(err)
 }
 
-func (s *MessengerCommunitiesSuite) TestPostToCommunityChat() {
+func (s *MessengerCommunitiesSuite) TestPinMessageInCommunityChat() {
```
osmaczko: No one can prevent a peer from sending a pinned message in the community channel (someone might modify the status-go source code for their client). We should also check on the receiver's side to ensure that pinned messages from non-approved members are rejected.

protocol/communities_messenger_test.go
```diff
@@ -436,12 +436,37 @@ func (s *MessengerCommunitiesSuite) TestCommunityContactCodeAdvertisement() {
...
 	community, chat := s.createCommunity()
+	community.AdminSettings().PinMessageAllMembersEnabled = true // FIXME: this doesn't work
```
osmaczko: This doesn't work because the changes aren't saved. When a community is advertised to the user, only the community ID is propagated. Then, when constructing the community message, it is fetched from the database.

osmaczko (commented):
Looks good in general :+1: some minor things to address.

protocol/messenger.go
```diff
@@ -2241,11 +2244,6 @@ func (m *Messenger) sendChatMessage(ctx context.Context, message *common.Message
 		message.Payload = &protobuf.ChatMessage_Community{Community: wrappedCommunity}
 
 		message.ContentType = protobuf.ChatMessage_COMMUNITY
-	} else if len(message.AudioPath) != 0 {
-		err := message.LoadAudio()
-		if err != nil {
-			return nil, err
-		}
```
jrainville: The changes above can be reverted

jrainville (commented):
Code looks good, though there is a small cleanup needed.

As we discussed in our 1-1, the Pin message test can be added in another PR, because it seems like the Setting to let users pin in a community is not being saved correctly on the receiver side, so it's worth another issue.

osmaczko:
> it seems like the Setting to let users pin in a community is not being saved correctly on the receiver side

I believe the root cause is different: https://github.com/status-im/status-go/pull/4064#discussion_r1350782701

jrainville:
> > it seems like the Setting to let users pin in a community is not being saved correctly on the receiver side
> 
> I believe the root cause is different: [#4064 (comment)](https://github.com/status-im/status-go/pull/4064#discussion_r1350782701)

Yeah I know. When I was in the 1-1 with him, we did add the `saveCommunity` call. Even though I checked that the owner's DB did have the right value afterwards, Alice's DB didn't have it. So, it seems like maybe members don,t save settings correctly?

Or maybe we do not publish community settings with `advertiseCommunity`?

osmaczko:
> > > it seems like the Setting to let users pin in a community is not being saved correctly on the receiver side
> > 
> > 
> > I believe the root cause is different: [#4064 (comment)](https://github.com/status-im/status-go/pull/4064#discussion_r1350782701)
> 
> Yeah I know. When I was in the 1-1 with him, we did add the `saveCommunity` call. Even though I checked that the owner's DB did have the right value afterwards, Alice's DB didn't have it. So, it seems like maybe members don,t save settings correctly?
> 
> Or maybe we do not publish community settings with `advertiseCommunity`?

The `saveCommunity` should be executed on the control node side before it is advertised to the users. The admin settings are a part of the `CommunityDescription`. The description sent by the control node should be identical on the receiver side because the proto lib handles unmarshaling. I would try to investigate what admin settings (`CommunityDescription`) is distributed exactly.

protocol/communities_messenger_test.go
shinnok: Moved the `TestPinMessageInCommunityChat` and afferent investigation to https://github.com/status-im/status-go/pull/4139.

protocol/communities_messenger_test.go
shinnok: This is fixed for both sender and receiver in my testing. Check out the consolidated fixes in final commit dace1b28d8510bc9514ff4661928d2fb5c0302e6.

protocol/messenger.go
shinnok: Done. Check consolidated changes in dace1b28d8510bc9514ff4661928d2fb5c0302e6.

shinnok:
I've created issue https://github.com/status-im/status-go/issues/4138 to further investigate the channel message pin issue and forked the new test method `TestPinMessageInCommunityChat` to PR https://github.com/status-im/status-go/pull/4139.

---

The current squashed commit contains the sufficient changes to block messages, reactions and pins to channels for which the user is not a member, both for him and other members. Please review again.

osmaczko (approved):

jrainville (approved):

protocol/communities/community_test.go
```diff
@@ -483,18 +483,30 @@ func (s *CommunitySuite) TestValidateRequestToJoin() {
...
+
+//FIXME: add complete permissions matrix
+// | Community Perm | Chat Perm | Chat Member | Grant   | Can Post |
+// |----------------|-----------|-------------|---------|----------|
+// | MEMBERSHIP     | x         | Yes         | -       | Yes      |
+// | MEMBERSHIP     | x         | No          | -       | No       |
+// | NO-MEMBERSHIP  | x         | Yes         | -       | Yes      |
+// | NO-MEMBERSHIP  | x         | No          | -       | No       |
+// | NO-MEMBERSHIP  | x         | No          | Valid   | Yes?     |
+// | NO-MEMBERSHIP  | x         | No          | Invalid | No       |
```
jrainville: @cammellos can you shortly explain what a grant is? I feel like it's probably something from an old requirement, but I don't remember seeing it in any of the new requirements or code.

I'm of the opinion that if someone is not part of the member list, they shouldn't be able to post ever. However, this test below seems to imply that with a grant, it should be possible. Is that still something we want? If not, might as well remove all the grant code I guess?

cc @John-44 to clarify requirements. Basically, am I correct that you 100% need to be a member of a community/chat to be able to post to it?

protocol/communities_messenger_test.go
shinnok: To expand on the matter a bit, both a community and individual chats of access permissions of these type:
https://github.com/status-im/status-go/blob/fix/11915/send-message-to-gated-community/protocol/protobuf/communities.pb.go

On top of that, there is the distinction that a user is a member, or not, of a community and then an individual chat. A user becomes a member of a Community when it is first advertised by another user B and A requests to join. Then what happens to A's request to join depends on _Community.Access.Permission_. After joining, when A attempt a post, the protocol code decides whether to allow a post, a reaction or a pin, based on _Community.Chat.Access.Permission._ Please correct if any of these is incorrect.

Now, where do grants fit in? As I could not find either documentation nor help pages describing what it is, nor the code hints seems to be implemented fully:

[`156bf39` (#4064)](https://github.com/status-im/status-go/pull/4064/commits/156bf3942a7d2ecf855628fa6aad82c03072f50d)

protocol/communities/community_test.go
jrainville: I just discussed this with @cammellos and he said:
> so, the idea is that when you join a community, you get a "grant"
and that's a proof you have been added
so when you send a message, you send also the grant
so other clients can optimistically process the message, if the community description hasn't reached them yet

But it turns out we never fully supported it anyway:
> we never really supported it

He then told me it's fine to remove it.

I'll open an issue in a few.

@shinnok that means you can just set the grant tests to `canPost: false`, so that the test will pass. Then someone will have the chore of removing the grant code that's no longer useful later.

jrainville: I added the issue to remove the grant code here: https://github.com/status-im/status-go/issues/4180

shinnok: I adjusted the expected result for the sub-tests that are supposed to fail in TestCanPost().

protocol/communities/community.go
```diff
@@ -1742,6 +1742,12 @@ func (o *Community) CanPost(pk *ecdsa.PublicKey, chatID string, grantBytes []byt
 		return false, nil
 	}
 
+	// if not a member can't post or react on message
+	if !o.IsMemberInChat(pk, chatID) {
```
jrainville: The condition `if o.config.CommunityDescription.Members == nil {` at line 1780 (the comment on it seems not ok) should be moved above this one or merged with this one.

jrainville (commented):
Looks good, the order of conditions in `canPost` can be improved a little

protocol/messenger_activity_center_test.go
osmaczko, line 1: Thanks for cleanup.

osmaczko (approved):
:+1:

shinnok:
Attaching a video showing the changes in effect, the top left Status instance has adjustments in this PR in place, the bottom right doesn't, but instead has the QML UI checks disabled for sending, reacting and pinning. This effectively verifies that the receiving side effectively blocks non-member actions, however, once applied, this patch also denies non-member actions on the sender's side for efficiency. Testing this exact same workflow without this patch on either side, but with the [UI checks disabled](https://github.com//status-im/status-desktop/tree/fix/11915/send-message-to-gated-community), will allow a non-member to send, react and pin before being approved in a on-request community or before spectating for no-request community.

https://github.com/status-im/status-go/assets/544446/870f70ec-0a10-4ba0-97a5-327b3f95243d

* video is encoded to H.265/HEVC due to size/length, might help to download and play with VLC/Mplayer instead of the browser *

PR #4068, opened 2023-09-27
Feat/11962/unique control node across devices

osmaczko:
Ensure there is only one control node across devices by syncing the installation_id of control node.
For now control node can be only changed by import.

Based on #4083 , otherwise tests will fail due to wrongly setup protocol database.


protocol/communities/adaptors.go
```diff
@@ -29,11 +36,13 @@ func (o *Community) ToSyncInstallationCommunityProtobuf(clock uint64, communityS
 	return &protobuf.SyncInstallationCommunity{
 		Clock:          clock,
 		Id:             o.ID(),
+		PrivateKey:     pkb,
```
jrainville: What's the goal of adding the PrivateKey?

protocol/communities/manager.go
```diff
@@ -1032,12 +1045,21 @@ func (m *Manager) ImportCommunity(key *ecdsa.PrivateKey) (*Community, error) {
 
 	if community == nil {
 		description := &protobuf.CommunityDescription{
-			Permissions: &protobuf.CommunityPermissions{},
+			Permissions: &protobuf.CommunityPermissions{
+				Access: protobuf.CommunityPermissions_ON_REQUEST,
```
jrainville: Is this a leftover of testing?

protocol/communities/persistence.go
```diff
@@ -1401,3 +1413,36 @@ func toRevealedAccount(rawAddress sql.NullString, rawChainIDsStr sql.NullString,
...
+		}
+		return nil, err
+	}
+
+	return result, nil
+}
+
+func (p *Persistence) SaveSyncControlNode(communityID types.HexBytes, clock uint64, installationID string) error {
+	_, err := p.db.Exec(
+		`INSERT INTO communities_control_node (
```
jrainville, line 1552: Is this new table necessary? Could we just have added the installation_id in the communities table (with an update clock if needed)

jrainville (commented):
Nice job. I have a few questions

protocol/communities/adaptors.go
osmaczko: For open communities, once you lose access to the control node device, the community could be lost forever. We eliminated private key syncing primarily because we aimed to ensure that there is only one control node when a community is imported onto another device. Now, only one device (identified by its installation_id) can act as the control node at any given time, and this is guaranteed by the protocol. When a community is imported onto a different device, that device notifies the previous control node that it has now taken over control. This has the advantage of ensuring that the community isn't lost if the device is misplaced.

protocol/communities/manager.go
osmaczko: No, it's intentional. Without it, [the validation](https://github.com/status-im/status-go/blob/17cef5f211b6bd599592cbc27bcf40907688d238/protocol/communities/community.go#L943) for 'CommunityDescription' will fail, leading to all community updates being rejected by clients. I discovered this when my newly-written test failed due to an invalid 'CommunityDescription'

P.S. I bet imported communities never truly functioned 🤷‍♂️

protocol/communities/persistence.go
osmaczko: Yes. The clock for community and for sync messages is different (and should be different).

protocol/communities_messenger_test.go
```diff
@@ -2333,9 +2329,12 @@ func (s *MessengerCommunitiesSuite) TestSyncCommunity() {
...
 
 	s.True(newCommunity.IsControlNode())
-	s.False(newCommunity.IsOwnerWithoutCommunityKey())
-	s.True(tnc.IsOwnerWithoutCommunityKey())
+	s.True(newCommunity.IsOwner())
+
+	// Even though synced device have the private key, it is not the control node
+	// There can be only one control node
 	s.False(tnc.IsControlNode())
+	s.True(tnc.IsOwner())
```
osmaczko, line 2447: @jrainville here: private key doesn't determine the control node anymore

protocol/communities/adaptors.go
jrainville: Ok, I get it. I'm not a fan of putting private keys on waku, but it's encrypted and also once the OwnerToken is minted, it's then no longer used.

protocol/communities/manager.go
jrainville: That's weird. We really should test imported communities. This looks like it makes it so that all imported communities become on request. It's not the biggest deal, but still.

protocol/communities/persistence.go
jrainville: Yeah I get it. We can add another field called `installation_id_updated_at` though if we want.

protocol/communities/manager.go
osmaczko: Among all options, `CommunityPermissions_ON_REQUEST` seems as a good default when importing without previous community information. I am open to change it to something else if that makes more sense.

protocol/communities/persistence.go
osmaczko: Right, it would work as well. I didn't want to pollute communities table with something that is relevant only to control nodes.

jrainville: I don't have a strong opinion. I think adding a new table can get a little more confusing, but it's also a bit cleaner like you said since it's only for owners. I'll let the other reviewer decide :stuck_out_tongue:

osmaczko:
@jrainville @cammellos I am re-requesting review, some things have changed after rebase.

@jrainville I have removed the commit that recovered the private key syncing as it would sync the private key of the owner if the community was token-gated.

jrainville (approved):

PR #4072, opened 2023-09-28
chore: use lamport timestamp for communities

### LINKED ISSUE #11961: [Community Ownership] use timestamped hinted lamport clock for communities
- The CommunityDescription's Clock might conflict with other control node clocks. Consider the following scenario:
UserA is the control node.
UserA broadcasts communityDescrX (clock==0).
UserA broadcasts communityDescrY (clock==1).
The control node changes in the contract, and now the control node is UserB.
Now, when UserB comes online (assuming UserB did not receive communityDescrY), UserB will broadcast communityDescrZ (clock==1), which will be ignored by other members who received communityDescrY earlier.
    - Solution: Utilize timestamped hinted lamport clocks (as suggested by @cammellos). This would require a substantial refactor because we need to propagate the timesource to `Communty.nextClock()` through the communities manager.

_Originally posted by @osmaczko in https://github.com/status-im/status-desktop/issues/11371#issuecomment-1624206189_

osmaczko:
closes: status-im/status-desktop#11961

### DIFF
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -51,13 +51,18 @@ type EventsData struct {
 }
 
 type Community struct {
-	config *Config
-	mutex  sync.Mutex
+	config     *Config
+	mutex      sync.Mutex
+	timesource common.TimeSource
 }
 
-func New(config Config) (*Community, error) {
+func New(config Config, timesource common.TimeSource) (*Community, error) {
 	if config.MemberIdentity == nil {
 		return nil, errors.New("no member identity")
+	}
+
+	if timesource == nil {
+		return nil, errors.New("no timesource")
 	}
 
 	if config.Logger == nil {
@@ -68,7 +73,7 @@ func New(config Config) (*Community, error) {
 		config.Logger = logger
 	}
 
-	community := &Community{config: &config}
+	community := &Community{config: &config, timesource: timesource}
 	community.initialize()
 	return community, nil
 }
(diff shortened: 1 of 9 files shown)


jrainville (approved):
LGTM

protocol/communities/community.go
```diff
@@ -1820,7 +1821,20 @@ func (o *Community) RequestedToJoinAt() uint64 {
 }
 
 func (o *Community) nextClock() uint64 {
-	return o.config.CommunityDescription.Clock + 1
+	if o.timesource == nil {
+		return o.config.CommunityDescription.Clock + 1
```
cammellos: when would this be the case, only testing? maybe otherwise it would have been best to throw an error, as it might mean something is misconfigured (though is likely fine either ways)

protocol/communities_messenger_test.go
```diff
@@ -2635,7 +2635,6 @@ func (s *MessengerCommunitiesSuite) TestSyncCommunity_Leave() {
 
 	// Check that the joined community has the correct values
 	s.Equal(community.ID(), aCom.ID())
-	s.Equal(uint64(0x2), aCom.Clock())
```
cammellos, line 2638: maybe we can compare to non 0, or greater than?

cammellos (approved):
looks good, only minor comments

protocol/communities/community.go
osmaczko: good point :+1:  addressed, please re-check

protocol/communities_messenger_test.go
osmaczko: addressed

cammellos (approved):

PR #4073, opened 2023-09-28
feat(backup): add 1-1 chats and group chats to backup

### LINKED ISSUE #10324: When a user performs a fresh install and restores via seed phrase, 1on1 and ad-hoc group chat conversations that happened in the last 30 days should be restored
Version 0.11, Window 10

To reproduce:

1. Perform a fresh install of Status Desktop, and restore an account that's been active in the last 30 days via Seed Phrase

2. When the user lands inside Status after restoring, all of the user's contacts have been successfully restored from the backup to waku (Yay! :-) ).  This can be checked by navigating to Settings/Messaging/Contacts and viewing the restored contacts

3. Navigate to the Messages section of the app.

--- What currently incorrectly happens ---

4. The Messages secondary nav will be empty, with no 1on1 and ad-hoc group chats listed

--- Desired functionality ---

4. All 1on1 and ad-hoc group chats that have had activity in them in the last 30 days should be restored.  By 'restored', I mean that they should be listed in the Messages section secondary nav, in recency order (e.g. the chat with the most recent message at the top, the chat with the second most recent message 2nd from the top, etc..).  I think that due to 'perfect forward secrecy' the messages themselves in the 1on1 and ad-hoc group chats cannot be restored via fetching the user's last 30 days of messages from a waku mailserver (but I would love to be proven wrong about this!), but at the minimum we should restore the list of 1on1 and ad-hoc group chats into the Messages section secondary nav.

One way to implement this would be to simply add a list of all 1on1 and ad-hoc group chats in the Messages section in recency to the list of user data that's: 1) backed up to Waku every 8 hours and 2) restored to a new device during initial pairing in the device syncing flow.

--------------------

This issue is a more detailed write up of **Issue 4.** in the following github issue https://github.com/status-im/status-desktop/issues/10276

Assigning to Milestone 0.13, but feel free to pull into Milestone 0.12 if there is time.

jrainville:
Fixes https://github.com/status-im/status-desktop/issues/10324

Backs up 1-1 chats and groups chats so they can be synced back when importing with seed phrase.

Only syncs the chat metadata, not the messages. That way, the chat list is populated when you rei-mport.

[import-chats.webm](https://github.com/status-im/status-go/assets/11926403/0c241a66-6ee6-4f1a-8f9d-b699e20ba5d3)


protocol/messenger_backup.go
```diff
@@ -308,6 +327,44 @@ func (m *Messenger) backupCommunities(ctx context.Context, clock uint64) ([]*pro
 	return backupMessages, nil
 }
 
+func (m *Messenger) backupChats(ctx context.Context, clock uint64) []*protobuf.Backup {
+	var oneToOneAndGroupChats []*protobuf.SyncChat
+	m.allChats.Range(func(chatID string, chat *Chat) (shouldContinue bool) {
+		if chat.Active && (chat.OneToOne() || chat.PrivateGroupChat()) {
```
igor-sirotin: Can we please use a early `return`here?
```go
if !chat.Active || !(chat.OneToOne() || chat.PrivateGroupChat()) 
    return true
// do stuff here with less spaghetti
```

protocol/messenger_handler.go
```diff
@@ -493,6 +493,53 @@ func (m *Messenger) HandleSyncInstallationAccount(state *ReceivedMessageState, m
...
+			Name:                     syncChat.Name,
+			Timestamp:                time.Now().Unix(),
+			ReadMessagesAtClockValue: 0,
+			Active:                   true,
+			Joined:                   time.Now().Unix(),
+			ChatType:                 ChatType(syncChat.ChatType),
+			Highlight:                false,
+		}
+		if chat.OneToOne() {
+			chat.Name = "" // Emptying since it contains non useful data
```
igor-sirotin: Maybe we shouldn't even send the chat name in this case?
Not big deal, but less data in the network 😄

protocol/messenger_backup_test.go
```diff
@@ -857,3 +857,68 @@ func (s *MessengerBackupSuite) TestBackupWatchOnlyAccounts() {
...
+	// Backup
+	_, err = bob1.BackupData(context.Background())
+	s.Require().NoError(err)
+
+	// Wait for the message to reach its destination
+	response, err = WaitOnMessengerResponse(
+		bob2,
+		func(r *MessengerResponse) bool {
+			return r.BackupHandled
+
```
igor-sirotin: 😠

igor-sirotin (approved):
Cool, thank you! 👍 

Do you think this could be fixed together: https://github.com/status-im/status-go/issues/3547?

jrainville:
> Do you think this could be fixed together: https://github.com/status-im/status-go/issues/3547?

Are you talking about the comment or the issue itself? Because I think this PR fixes the issue, but I haven't seen that log about the `HandleSyncChatMessagesRead`, though I guess it's in geth.log and I didn't check that.

jrainville:
Ah nvm, I just saw that your issue is about local pairing. I just tested local pairing and it didn't work. I guess it uses a different protobuf or something?

I'll try to do it in another PR

igor-sirotin:
> Ah nvm, I just saw that your issue is about local pairing. I just tested local pairing and it didn't work. I guess it uses a different protobuf or something?

yeah it's a different thing, but definitely related and shouldn't be forgotten 🙂 

> I'll try to do it in another PR

Cool, thanks!

jrainville:
@saledjenic @Samyoul friendly reminder to review. Thanks

jrainville:
@saledjenic can you review this one too, thanks

protocol/messenger_backup.go
```diff
@@ -308,6 +327,46 @@ func (m *Messenger) backupCommunities(ctx context.Context, clock uint64) ([]*pro
 	return backupMessages, nil
 }
 
+func (m *Messenger) backupChats(ctx context.Context, clock uint64) []*protobuf.Backup {
+	var oneToOneAndGroupChats []*protobuf.SyncChat
+	m.allChats.Range(func(chatID string, chat *Chat) (shouldContinue bool) {
```
saledjenic:
```suggestion
	m.allChats.Range(func(chatID string, chat *Chat) bool {
```

protocol/messenger_backup.go
```diff
@@ -308,6 +327,46 @@ func (m *Messenger) backupCommunities(ctx context.Context, clock uint64) ([]*pro
...
+		if !chat.Active || (!chat.OneToOne() && !chat.PrivateGroupChat()) {
```
saledjenic: @jrainville don't we need to backup inactive chats as well? I guess that `Active` flag is `false` if the chat is removed (soft deleted), if so we should backup such chats as well otherwise those removed chats may appear again after recovering.

protocol/messenger_handler.go
saledjenic: If the `Name` is not marked as omitempty when marshalling, will be sent anyhow. 😉

protocol/messenger_backup.go
```diff
@@ -308,6 +327,46 @@ func (m *Messenger) backupCommunities(ctx context.Context, clock uint64) ([]*pro
...
+			return true
+		}
+		syncChat := protobuf.SyncChat{
+			Id:       chatID,
+			ChatType: uint32(chat.ChatType),
```
saledjenic, line 338: I guess we should backup props like color, emoji, active, muted, and maybe other relevant props as well?

saledjenic (commented):
@jrainville sorry for the late reply, the changes look good, I left just a few questions that should be considered before merging this.

protocol/messenger_handler.go
saledjenic: If the `Name` is not marked as omitempty when marshalling, will be sent anyhow. 😉

protocol/messenger_backup.go
jrainville: Updated. I added the Active property. I also added a clock, so that we only sync the latest

protocol/messenger_backup.go
jrainville: I tried to keep it as light weight as possible.
The color and image props for group chats come from the `MembershipUpdateEvent` (the name is confusing)
I didn't think of syncing the muted property. I guess it makes sense, I'll add it

saledjenic (approved):

PR #4078, opened 2023-09-29
feat: added functionality to ask a client to sign a transaction and an endpoint to continue sending using provided signature

saledjenic:
This functionality is needed in case the user wants to send a transaction and signs it using the signature provided by the keycard (or any other compatible way).

### DIFF
--- a/services/wallet/bridge/cbridge.go
+++ b/services/wallet/bridge/cbridge.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	ethTypes "github.com/ethereum/go-ethereum/core/types"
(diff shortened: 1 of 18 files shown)


stefandunca (approved):
🤩

services/wallet/bridge/cbridge.go
```diff
@@ -289,13 +289,25 @@ func (s *CBridge) Send(sendArgs *TransactionBridge, verifiedAccount *account.Sel
...
+	tx, err := s.send(sendArgs, getSigner(sendArgs.ChainID, sendArgs.CbridgeTx.From, verifiedAccount))
 	if err != nil {
 		return types.HexToHash(""), err
 	}
 
 	return types.Hash(tx.Hash()), nil
 }
 
+func (s *CBridge) SendWithSignature(chainID uint64, tx *ethTypes.Transaction, signature []byte) (types.Hash, error) {
+	return s.transactor.SendBuiltTransactionWithSignature(chainID, tx, signature)
```
alaibe: Do we need a SendWithSignature being copied over all bridges and doing the same, this could simply be trigger from transfer/transaction

As we don't have custom logic per bridge?

(Comment here is about the interface so valid for all bridges)

services/wallet/bridge/cbridge.go
```diff
@@ -244,33 +245,32 @@ func (s *CBridge) GetContractAddress(network *params.Network, token *token.Token
 	return nil
 }
 
-func (s *CBridge) Send(sendArgs *TransactionBridge, verifiedAccount *account.SelectedExtKey) (types.Hash, error) {
+func (s *CBridge) send(sendArgs *TransactionBridge, signerFn bind.SignerFn) (*ethTypes.Transaction, error) {
```
alaibe: nit: `sendOrBuild`? (valid comment for all bridges)

services/wallet/transfer/transaction.go
```diff
@@ -302,6 +352,60 @@ func (tm *TransactionManager) CreateMultiTransactionFromCommand(ctx context.Cont
 	}, nil
 }
 
+func (tm *TransactionManager) ProceedWithTransactionsSignatures(ctx context.Context, signatures map[string]SignatureDetails) (*MultiTransactionCommandResult, error) {
+	if tm.multiTransactionForKeycardSigning == nil {
+		return nil, errors.New("no multi transaction to proceed with")
+	}
+	if len(tm.transactionsBridgeData) == 0 {
+		return nil, errors.New("no transactions brifge data to proceed with")
```
alaibe: nit: brifge => bridge

alaibe (approved):

services/wallet/bridge/cbridge.go
saledjenic: Yes, we can, I didn't know that all of them would do the same thing at the beginning and added that to the interface, now when I see they do exactly the same thing, it can be moved.

services/wallet/bridge/cbridge.go
saledjenic: Yes, it can be renamed.

saledjenic:
@alaibe suggestions applied.

PR #4079, opened 2023-09-29
[#17288] fix: Mention is displayed as public key in PNs

mohsen-ghafouri:
Mention is displayed as public key in PNs message

related to status-mobile issue https://github.com/status-im/status-mobile/issues/17396


protocol/messenger.go
```diff
@@ -204,6 +204,30 @@ type EnvelopeEventsInterceptor struct {
...
+		return ensName, nil
+	}
+	return m.settings.DisplayName()
+}
+
+func (m *Messenger) GetCanonicalName(mentionID string) (string, error) {
+	if mentionID == m.myHexIdentity() {
+		return m.CanonicalName()
+	}
+	fmt.Println("m.contacts", m.allContacts)
```
ilmotta: Debugging left over maybe?

protocol/messenger_test.go
```diff
@@ -2447,3 +2449,38 @@ type testTimeSource struct{}
...
+	_, err := bob.Start()
+	s.Require().NoError(err)
+	defer bob.Shutdown() // nolint: errcheck
+
+	// Set Bob's profile keypair And set display names for Bob and Alice
+	bobProfileKp := accounts.GetProfileKeypairForTest(true, false, false)
+	bobProfileKp.KeyUID, bobProfileKp.Accounts[0].KeyUID = bob.account.KeyUID, bob.account.KeyUID
+	s.Require().NoError(bob.settings.SaveOrUpdateKeypair(bobProfileKp))
+	s.Require().NoError(bob.SetDisplayName("bobby"))
+	s.Require().NoError(alice.settings.SetDisplayName("Alice"))
```
ilmotta: We could remove `SetDisplayName` and `SetNotificationEnabled` functions because they're being used just for testing.

```diff
modified   protocol/messenger_test.go
@@ -24,6 +24,7 @@ import (
 	enstypes "github.com/status-im/status-go/eth-node/types/ens"
 	"github.com/status-im/status-go/images"
 	"github.com/status-im/status-go/multiaccounts/accounts"
+	"github.com/status-im/status-go/multiaccounts/settings"
 	"github.com/status-im/status-go/protocol/common"
 	"github.com/status-im/status-go/protocol/protobuf"
 	"github.com/status-im/status-go/protocol/requests"
@@ -2462,8 +2463,8 @@ func (s *MessengerSuite) TestSendMessageMention() {
 	bobProfileKp.KeyUID, bobProfileKp.Accounts[0].KeyUID = bob.account.KeyUID, bob.account.KeyUID
 	s.Require().NoError(bob.settings.SaveOrUpdateKeypair(bobProfileKp))
 	s.Require().NoError(bob.SetDisplayName("bobby"))
-	s.Require().NoError(alice.settings.SetDisplayName("Alice"))
-	s.Require().NoError(alice.settings.SetNotificationEnabled(true))
+	s.Require().NoError(alice.settings.SaveSettingField(settings.DisplayName, "Alice"))
+	s.Require().NoError(alice.settings.SaveSettingField(settings.NotificationsEnabled, true))
```

protocol/messenger.go
```diff
@@ -204,6 +204,30 @@ type EnvelopeEventsInterceptor struct {
 	Messenger             *Messenger
 }
 
+func (m *Messenger) CanonicalName() (string, error) {
```
ilmotta: @mohsen-ghafouri, I'm confused by the names of the functions. One is `CanonicalName` and the other is `GetCanonicalName`, both get the same "thing" based on their name, but one returns either the ENS name or display name and the other the primary name.

So in the end, what is a canonical name? Should the `m.settings.GetPreferredUsername` be part of the logic to return a canonical name?

mohsen-ghafouri: actually `CanonicalName` is for current user and should return user name/ens/nickname, and i gave priority to ens if it has any. `GetCanonicalName` func but return contact primaryName by mentionId. i can change `CanonicalName` to for example `GetUserPrefferredName` maybe? @ilmotta

mohsen-ghafouri: I pushed some changes please check again. thanks

ilmotta: Thanks @mohsen-ghafouri, but I still don't understand from the code alone why the canonical name of the current account means something different than the canonical name of some other contact.

The function `GetCanonicalName` could be read as `GetCanonicalNameIfCurrentAccountOtherwisePrimaryName`. Of course this is an absurd function name, but I'm just trying to illustrate why it's so confusing to me. The point is that `GetCanonicalName` doesn't return a canonical name if it's not the current user, it returns the PrimaryName, two very different things, so why name both as if they are the same thing?

I hope my confusion makes some sense :)

mohsen-ghafouri: Yes, it makes sense. So, basically, if the mention ID is for the current user, it should look at the settings. When it's a contact, it should return the contact's primary name. I'm new to this codebase, so maybe I made a mistake because I thought that ENS name / display name is equal to the contact's primary name. please let me know your suggestion to remove this confusion @ilmotta

mohsen-ghafouri: if it's all about func name, what do you think about `GetOwnCanonicalName` and `ResolveCanonicalName` for the second func ?

ilmotta: @mohsen-ghafouri, even though I'm not new to the code, the naming and what they mean is still confusing to me, like primary name, secondary name, display name, preferred name, nickname, it's a lot of *names, and now we'll have "canonical name" :) ENS name is clear of course. Maybe others in the team are also confused just as me...

Maybe @cammellos can help here?

cammellos: We can rename to `PrimaryName` -> intead of `Canonical`?
The only difference between the user logged in and any contacts, is that in contacts you might have a `LocalNickname`, which takes precendence over ENS/Display name, while for that's not available for yourself (i.e you can nickname yourself), and therefore it's ENS/Display name

Would that clear the confusion?

mohsen-ghafouri: thanks, so i can rename funcs to `GetOwnPrimaryName` and `ResolvePrimaryName` , i hope that it can help

ilmotta: It cleared up a little bit, thanks. I found this unit test which made me better understand your suggestion @cammellos https://github.com/status-im/status-go/blob/7d03ae8272f856cb4a7ac6539a3d295b10b3b269/protocol/contact_test.go#L727

mohsen-ghafouri: thanks for your suggestion, resolved

ilmotta (approved):

PR #4080, opened 2023-09-29
fix(test): fix flaky permission test

### LINKED ISSUE #4069: Sometimes `TestJoinCommunityAsAdminWithMemberAndAdminPermission` failed
The `TestJoinCommunityAsAdminWithMemberAndAdminPermission` test case appears to exhibit inconsistent results, with intermittent failures.

To reproduce the issue, you may need to rerun the test multiple times locally until a failure is observed.

Relate error log:
```
communities_messenger_helpers_test.go:324: 
          Error Trace:  /Users/frank/development/go/src/github.com/status-im/status-go/protocol/communities_messenger_helpers_test.go:324
                              /Users/frank/development/go/src/github.com/status-im/status-go/protocol/communities_messenger_token_permissions_test.go:176
                              /Users/frank/development/go/src/github.com/status-im/status-go/protocol/communities_messenger_token_permissions_test.go:166
                              /Users/frank/development/go/src/github.com/status-im/status-go/protocol/communities_messenger_token_permissions_test.go:831
          Error:        Received unexpected error:
                        permission to join not satisfied
          Test:         TestMessengerCommunitiesTokenPermissionsSuite/TestJoinCommunityAsAdminWithMemberAndAdminPermission
```

branch: latest `develop` (`17cef5f211b6bd599592cbc27bcf40907688d238`)

jrainville:
Fixes #4069

The problem was that there was a race condition where `ReevaluateCommunityMembersPermissions`  from the first permission creation would finish after the second permission was created, thus overwriting the second permission.

Because of that, the test would fail because the Admin permission would no longer exist and Bob would not have access to the community.

### DIFF
--- a/protocol/communities_messenger_token_permissions_test.go
+++ b/protocol/communities_messenger_token_permissions_test.go
@@ -748,6 +748,13 @@ func (s *MessengerCommunitiesTokenPermissionsSuite) TestJoinCommunityAsMemberWit
 	s.Require().NoError(err)
 	s.Require().Len(response.Communities(), 1)
 
+	waitOnCommunityPermissionCreated := waitOnCommunitiesEvent(s.owner, func(sub *communities.Subscription) bool {
+		return sub.Community.HasTokenPermissions()
+	})
+
+	err = <-waitOnCommunityPermissionCreated
+	s.Require().NoError(err)
+
 	// setup become admin permission
 	permissionRequestAdmin := requests.CreateCommunityTokenPermission{
 		CommunityID: community.ID(),
@@ -766,10 +773,17 @@ func (s *MessengerCommunitiesTokenPermissionsSuite) TestJoinCommunityAsMemberWit
 	s.Require().NoError(err)
 	s.Require().Len(response.Communities(), 1)
 
+	waitOnCommunityPermissionCreated = waitOnCommunitiesEvent(s.owner, func(sub *communities.Subscription) bool {
+		return len(sub.Community.TokenPermissions()) == 2
+	})
+
+	err = <-waitOnCommunityPermissionCreated
+	s.Require().NoError(err)
+
 	// make bob satisfy the member criteria
 	s.makeAddressSatisfyTheCriteria(testChainID1, bobAddress, permissionRequestMember.TokenCriteria[0])
 
-	s.advertiseCommunityTo(community, s.bob)
+	s.advertiseCommunityTo(response.Communities()[0], s.bob)
 
 	// Bob should still be able to join even though he doesn't satisfy the admin requirement
 	// because he satisfies the member one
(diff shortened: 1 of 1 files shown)


protocol/messenger_communities.go
```diff
@@ -1966,7 +1966,7 @@ func (m *Messenger) CreateCommunity(request *requests.CreateCommunity, createDef
 	return response, nil
 }
 
-func (m *Messenger) CreateCommunityTokenPermission(request *requests.CreateCommunityTokenPermission) (*MessengerResponse, error) {
+func (m *Messenger) CreateCommunityTokenPermissionWithSkipParam(request *requests.CreateCommunityTokenPermission, skipReevaluation bool) (*MessengerResponse, error) {
```
mprakhov: I'm not sure that bool flags look right here. In my understanding, we are adding CommunityDescription with the permissions and saving it to the DB. After it is saved, we need to do a permission verification based on the permissions in the CommunityDescription

What could be wrong:
1.  CommunityDescription was not saved
2. We pass the old community description, which does not contain new Permissions

Anyway, reevaluating members' permission must be done on top of the CommunityDescription with permissions

jrainville: I only use that param in tests, though I get that it's not that clear and it's not clear that it shouldn't be used normally.

I can try to change the tests to wait for the reevaluation.

mprakhov: Okay, checked the test and I think I found the issue
When you are calling `s.advertiseCommunityTo(community, s.bob)` - you are sending an outdated community. You must request community from the DB or extract it from the last response

P.S. But maybe that's not the reason... at least in my test I faced with such a mistake before :)

mprakhov: nah... check where it is used and that's not the case :(

jrainville: Alright, I used `waitOnCommunityPermissionCreated` after the permission is created to wait for the re-evaluation.

It seems to work well and is way simpler.

You can review again

protocol/communities_messenger_token_permissions_test.go
```diff
@@ -820,12 +841,25 @@ func (s *MessengerCommunitiesTokenPermissionsSuite) TestJoinCommunityAsAdminWith
 	response, err = s.owner.CreateCommunityTokenPermission(&permissionRequestAdmin)
 	s.Require().NoError(err)
 	s.Require().Len(response.Communities(), 1)
+	s.Require().Len(response.Communities()[0].TokenPermissionsByType(protobuf.CommunityTokenPermission_BECOME_ADMIN), 1)
+	s.Require().Len(response.Communities()[0].TokenPermissions(), 2)
 
-	// make bob satisfy the admin criteria
-	s.makeAddressSatisfyTheCriteria(testChainID1, bobAddress, permissionRequestAdmin.TokenCriteria[0])
+	waitOnCommunityPermissionCreated = waitOnCommunitiesEvent(s.owner, func(sub *communities.Subscription) bool {
+		return sub.Community.HasTokenPermissions()
```
mprakhov: Can we do it more strictly and check that we have 2 permissions?

mprakhov (approved):

qfrank (approved):
thank you for your PR, the test works well on my local machine now.  @jrainville

PR #4020, opened 2023-09-13
Add filter debugging logs and map locking

vitvly:
This PR adds more extensive logging for wakuv2 filter and refactors the code that checks peer connectivity health.

Relies on changes from https://github.com/waku-org/go-waku/pull/750 and https://github.com/waku-org/go-waku/pull/762.

Items included:
- extend filter logging
- log PeerStats in `onConnStatus` listener
- add locking for filter operations, relying on mutex in `Filters`
- remove OnNewEnvelopes `envelopeErrors` return value, as it's empty for now
- refactor subscriptions, add checkFilterHealth() fn
- use unique sets of TopicType for Filter.ContentTopics
- modify single-filter unsubscribe (https://github.com/waku-org/go-waku/pull/762)
- update go-waku to latest version

PR #4082, opened 2023-10-02
Filter debugging 2

vitvly:
A version of https://github.com/status-im/status-go/pull/4020 with most recent status-go (382fcde74e74e914eed63804fde00d906c6f0b62) and go-waku (https://github.com/waku-org/go-waku/commit/dcc828749f67e2414125c2b94d4873181a48d08d)

It uses a different implementation of peer health checker for filter, moved into a separate file (filter_manager.go).

Quoting FilterManager's comment:
```
Methods on FilterManager maintain filter peer health
runFilterLoop is the main event loop
Filter Install/Uninstall events are pushed onto eventChan
Subscribe, UnsubscribeWithSubscription, IsSubscriptionAlive calls
are invoked from goroutines and request results pushed onto eventChan
```

Also, better log output for all points in filter lifecycle.


wakuv2/common/filter.go
```diff
@@ -89,28 +98,34 @@ func (fs *Filters) Install(watcher *Filter) (string, error) {
...
-	if fs.watchers[id] != nil {
-		fs.removeFromTopicMatchers(fs.watchers[id])
+	fs.Lock()
+	defer fs.Unlock()
+	watcher := fs.watchers[id]
+	if watcher != nil {
+		fs.removeFromTopicMatchers(watcher)
 		delete(fs.watchers, id)
+
+		fs.logger.Info("Filters.Uninstall()", zap.String("id", id))
```
cammellos: Comments generally start from lower case, maybe Info is too high? debug?

wakuv2/filter_manager.go
```diff
@@ -0,0 +1,360 @@
...
+	"github.com/waku-org/go-waku/waku/v2/protocol"
+	"github.com/waku-org/go-waku/waku/v2/protocol/filter"
+	"github.com/waku-org/go-waku/waku/v2/protocol/relay"
+	"github.com/waku-org/go-waku/waku/v2/protocol/subscription"
+	"go.uber.org/zap"
+	"golang.org/x/exp/maps"
+)
+
+const (
+	FilterEventAdded             int = 0
```
cammellos: You can use `iota` for this, also, generally you don't want to start from `0`, since in go is hard to tell whether something is not initialized or actually set to 0, if it's a non nullable value like in this case.

wakuv2/filter_manager.go
```diff
@@ -0,0 +1,360 @@
...
+	tempID    string
+	sub       *subscription.SubscriptionDetails
+	ch        chan FilterSubs
+}
+
+// Methods on FilterManager maintain filter peer health
+//
+// runFilterLoop is the main event loop
+//
+// Filter Install/Uninstall events are pusehd onto eventChan
```
cammellos: pushed typo

wakuv2/filter_manager.go
```diff
@@ -0,0 +1,360 @@
...
+				}
+				// Delete subs for removed peer
+				for filterID, subs := range mgr.filterSubs {
+					for _, sub := range subs {
+						if sub == nil {
+							// Skip temp subs
+							continue
+						}
+						if sub.PeerID == ev.peerID {
+							mgr.logger.Debug("FILTER sub is inactive", zap.String("filterId", filterID), zap.String("subID", sub.ID))
```
cammellos: lowercase

wakuv2/filter_manager.go
```diff
@@ -0,0 +1,360 @@
...
+			}
+		}
+	}
+}
+
+func (mgr *FilterManager) subscribeToFilter(filterID string, peer peer.ID, tempID string) {
+
+	f := mgr.getFilter(filterID)
+	if f == nil {
+		mgr.logger.Error("FILTER subscribeToFilter: No filter found", zap.String("id", filterID))
```
cammellos: lowercase

wakuv2/filter_manager.go
```diff
@@ -0,0 +1,360 @@
...
+
+	subDetails, err := mgr.node.FilterLightnode().Subscribe(ctx, contentFilter, filter.WithPeer(peer))
+	success := err == nil
+	var sub *subscription.SubscriptionDetails
+	if err != nil {
+		mgr.logger.Warn("FILTER could not add wakuv2 filter for peer", zap.String("filterId", filterID), zap.Stringer("peer", peer), zap.Error(err))
+	} else {
+		mgr.logger.Debug("FILTER subscription success", zap.String("filterId", filterID), zap.Stringer("peer", peer), zap.String("pubsubTopic", contentFilter.PubsubTopic), zap.Strings("contentTopics", contentFilter.ContentTopicsList()))
+		sub = subDetails[0]
+
```
cammellos: extra line

wakuv2/api_test.go
```diff
@@ -43,12 +43,12 @@ func TestMultipleTopicCopyInNewMessageFilter(t *testing.T) {
...
 
-	t1 := [4]byte{0xde, 0xea, 0xbe, 0xef}
-	t2 := [4]byte{0xca, 0xfe, 0xde, 0xca}
+	t1 := common.TopicType([4]byte{0xde, 0xea, 0xbe, 0xef})
+	t2 := common.TopicType([4]byte{0xca, 0xfe, 0xde, 0xca})
 
 	crit := Criteria{
 		SymKeyID:      keyID,
-		ContentTopics: []common.TopicType{common.TopicType(t1), common.TopicType(t2)},
+		ContentTopics: []common.TopicType{t1, t2},
```
siddarthkay, line 51: nice

wakuv2/filter_manager.go
```diff
@@ -0,0 +1,360 @@
...
+	mgr.eventChan <- FilterEvent{eventType: FilterEventSubscribeResult, filterID: filterID, tempID: tempID, sub: sub, result: success}
+}
+
+func (mgr *FilterManager) unsubscribeFromFilter(filterID string, sub *subscription.SubscriptionDetails) {
+	mgr.logger.Debug("FILTER unsubscribe from filter node", zap.String("filterId", filterID), zap.String("subId", sub.ID), zap.Stringer("peer", sub.PeerID))
+	// Unsubscribe on light node
+	ctx, cancel := context.WithTimeout(mgr.ctx, requestTimeout)
+	defer cancel()
+	_, err := mgr.node.FilterLightnode().UnsubscribeWithSubscription(ctx, sub)
+	success := err == nil
```
cammellos: I would move this line below the error checking, it always looks a bit odd using a value before error checking, though the code is perfectly fine

wakuv2/common/filter.go
```diff
@@ -89,28 +98,34 @@ func (fs *Filters) Install(watcher *Filter) (string, error) {
 	}
 
 	watcher.id = id
+
 	fs.watchers[id] = watcher
 	fs.addTopicMatcher(watcher)
+
+	fs.logger.Info("Filters.Install()", zap.String("id", id))
```
siddarthkay: I've seen a bunch of Info level logs, do you still need them?

wakuv2/filter_manager.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+// Check whether each of the installed filters
+// has enough alive subscriptions to peers
+func (mgr *FilterManager) pingPeers() {
+	mgr.logger.Debug("FILTER pingPeers")
+
+	distinctPeers := make(map[peer.ID]struct{})
+	for filterID, subs := range mgr.filterSubs {
+		if len(subs) == 0 { // TODO !!!!!!!
```
cammellos: is this still a TODO?

wakuv2/filter_manager.go
```diff
@@ -0,0 +1,360 @@
...
+	return protocol.ContentFilter{
+		PubsubTopic:   pubsubTopic,
+		ContentTopics: protocol.NewContentTopicSet(contentTopics...),
+	}
+}
+
+// Find suitable peer(s)
+func (mgr *FilterManager) findFilterPeers() []peer.ID {
+	allPeers := mgr.node.Host().Peerstore().Peers()
+	//mgr.logger.Debug("Peerstore peers", zap.Stringers("peers", allPeers))
```
siddarthkay: maybe we can delete this

wakuv2/filter_manager.go
```diff
@@ -0,0 +1,360 @@
...
+				continue
+			}
+			_, found := distinctPeers[sub.PeerID]
+			if found {
+				continue
+			}
+			distinctPeers[sub.PeerID] = struct{}{}
+			mgr.logger.Debug("FILTER ping peer", zap.Stringer("peerId", sub.PeerID))
+			go func(sub *subscription.SubscriptionDetails) {
+				err := mgr.isFilterSubAlive(sub)
```
cammellos, line 270: matching the exact error would feel safer here :) but maybe there's many and they all indicate a dead connection?

siddarthkay (approved):
Apart from a bunch of tiny nit picks this PR looks fabulous!
Thanks for all your hard work

wakuv2/filter_manager.go
cammellos: leftover

wakuv2/waku.go
```diff
@@ -1038,43 +933,41 @@ func (w *Waku) Subscribe(f *common.Filter) (string, error) {
...
+		w.filterManager.eventChan <- FilterEvent{eventType: FilterEventRemoved, filterID: id}
+	}
+
 	return nil
 }
 
+// Used for testing
+func (w *Waku) getFilterStats() FilterSubs {
+	ch := make(chan FilterSubs)
+	w.filterManager.eventChan <- FilterEvent{eventType: FilterEventGetStats, ch: ch}
```
cammellos, line 965: This looks suspicious, as it might end up in a panic if the channel is closed, I'll look at the code to make sure, just a note for now

wakuv2/waku.go
```diff
@@ -1038,43 +933,41 @@ func (w *Waku) Subscribe(f *common.Filter) (string, error) {
...
-		}
-	}
-
 	ok := w.filters.Uninstall(id)
 	if !ok {
 		return fmt.Errorf("failed to unsubscribe: invalid ID '%s'", id)
 	}
+
+	if w.settings.LightClient {
+		w.filterManager.eventChan <- FilterEvent{eventType: FilterEventRemoved, filterID: id}
```
cammellos, line 956: This looks suspicious, as it might end up in a panic if the channel is closed, I'll look at the code to make sure, just a note for now

wakuv2/filter_manager.go
```diff
@@ -0,0 +1,360 @@
...
+// Filter Install/Uninstall events are pusehd onto eventChan
+// Subscribe, UnsubscribeWithSubscription, IsSubscriptionAlive calls
+// are invoked from goroutines and request results pushed onto eventChan
+//
+// filterSubs is the map of filter IDs to subscriptions
+
+type FilterManager struct {
+	ctx              context.Context
+	filterSubs       FilterSubs
+	eventChan        chan (FilterEvent)
```
cammellos, line 61: when are we closing this? might not be need to be closed, but worth checking

cammellos: it's also unbuffered, so if an event is pushed after the loop is done (for whatever reason), it will lock the publisher, is that something you thought about?

wakuv2/filter_manager.go
```diff
@@ -0,0 +1,360 @@
...
+					}
+					go mgr.unsubscribeFromFilter(ev.filterID, sub)
+				}
+				delete(mgr.filterSubs, ev.filterID)
+
+			case FilterEventPingResult:
+				if ev.result {
+					break
+				}
+				if ev.filterID != "" {
```
cammellos: it's not immediately clear why a `filterID` being set indicates that a filter has not peers, maybe we could elaborate with a comment?, or rather the opposite, why `filterID` would be empty?

wakuv2/filter_manager.go
```diff
@@ -0,0 +1,360 @@
...
+					if sub == nil {
+						// Skip temp subs
+						continue
+					}
+					go mgr.unsubscribeFromFilter(ev.filterID, sub)
+				}
+				delete(mgr.filterSubs, ev.filterID)
+
+			case FilterEventPingResult:
+				if ev.result {
```
cammellos: maybe `result` -> `success`?

wakuv2/filter_manager.go
```diff
@@ -0,0 +1,360 @@
...
+							delete(subs, sub.ID)
+							go mgr.unsubscribeFromFilter(filterID, sub)
+						}
+					}
+					mgr.resubscribe(filterID)
+				}
+
+			case FilterEventSubscribeResult:
+				subs, found := mgr.filterSubs[ev.filterID]
+				if ev.result {
```
cammellos: `success` sounds better than result to me, a bit clearer

cammellos (commented):
Left some comments, looks good, much clearer, but I'll have to do another pass I think to dig into the logic

wakuv2/common/filter.go
vitvly: Nope, thanks for spotting this, fixed.

wakuv2/filter_manager.go
vitvly: Deleted.

wakuv2/common/filter.go
vitvly: Fixed.

wakuv2/filter_manager.go
vitvly: Ah, yes, thanks.

wakuv2/filter_manager.go
vitvly: Fixed.

wakuv2/filter_manager.go
vitvly: Fixed.

wakuv2/filter_manager.go
vitvly: Fixed.

wakuv2/filter_manager.go
vitvly: Removed.

wakuv2/filter_manager.go
vitvly: Moved.

wakuv2/filter_manager.go
vitvly: Nope, leftover.

wakuv2/filter_manager.go
vitvly: Yes, we don't really care about the error type, just if it's there or not. Maybe @richard-ramos can weight in and correct my understanding, if necessary.

wakuv2/filter_manager.go
vitvly: Fixed.

wakuv2/waku.go
vitvly: Channel is not closed anywhere. Not sure if this is right or not. Noticed that in `wakuv2/waku.go` we don't close many channels too.

wakuv2/waku.go
vitvly: Channel is not closed anywhere.

wakuv2/filter_manager.go
vitvly: Loop is done only when context is closed. So if push to the channel is performed afterwards, the goroutine will hang?

wakuv2/filter_manager.go
vitvly: Added an extra comment.

wakuv2/filter_manager.go
vitvly: Yes, ok.

wakuv2/filter_manager.go
vitvly: Ok

vitvly:
@siddarthkay @cammellos addressed review comments. All fixed, except for the ones related to `eventChan` - not sure if this should be closed or not.

wakuv2/waku.go
cammellos: it's fine not to close a channel if you think it will be recycled by the garbage collector, we probably want to figure out the lifecycle, just to make sure, otherwise we'd have a memory leak

wakuv2/filter_manager.go
cammellos: if that' the case, we might leak memory, since we'll have a goroutine (or multiple) hanging

vitvly: Added `eventChan` buffer.

wakuv2/waku.go
vitvly: Added a buffer to `eventChan` to address an issue of goroutines hanging around.

vitvly:
@cammellos added a buffer to `eventChan`. Regarding the lifecycle - yes, it's entirely possible for some `Subscribe`/`Unsubscribe` call inside a goroutine to last for 10+ secs for instance (seen this in testing). So by then `runFilterLoop` might already exit, and therefore no one will read from `eventChan`. Hopefully buffering addresses this issue (to an extent no one will attempt to push onto a full buffer which is unlikely).

cammellos (approved):
It'd be good to figure out if we should close the channel, but otherwise looks good! Non blocking comment

wakuv2/common/filter.go
```diff
@@ -124,16 +139,15 @@ func (fs *Filters) AllTopics() []TopicType {
...
 		fs.allTopicsMatcher[watcher] = struct{}{}
 	} else {
 		filtersPerContentTopic, ok := fs.topicMatcher[watcher.PubsubTopic]
 		if !ok {
 			filtersPerContentTopic = make(ContentTopicToFilter)
 		}
 
-		for _, t := range watcher.Topics {
-			topic := BytesToTopic(t)
+		for topic := range watcher.ContentTopics {
```
richard-ramos, line 150: I'm curious about this change. Wouldnt `topic` be an integer here?

wakuv2/waku.go
```diff
@@ -1267,6 +1160,7 @@ func (w *Waku) Start() error {
 			case c := <-w.connStatusChan:
 				w.connStatusMu.Lock()
 				latestConnStatus := formatConnStatus(w.node, c)
+				w.logger.Info("PeerStats", zap.Any("stats", latestConnStatus))
```
richard-ramos: This might add too much noise to the logs, consider using debug instead

wakuv2/waku.go
```diff
@@ -1373,21 +1279,19 @@ func (w *Waku) Stop() error {
...
-		return nil, nil
+		return nil
 	}
 
-	envelopeErrors := make([]common.EnvelopeError, 0)
-
 	logger := w.logger.With(zap.String("hash", recvMessage.Hash().Hex()))
 
-	logger.Debug("received new envelope")
+	logger.Info("received new envelope")
```
richard-ramos: Consider setting this as Debug, as it will add a log for each message received.

wakuv2/waku.go
```diff
@@ -1427,10 +1331,10 @@ func (w *Waku) add(recvMessage *common.ReceivedMessage) (bool, error) {
 	}
 
 	if alreadyCached {
-		w.logger.Debug("w envelope already cached", zap.String("envelopeHash", recvMessage.Hash().Hex()))
+		w.logger.Info("w envelope already cached", zap.String("envelopeHash", recvMessage.Hash().Hex()))
 		common.EnvelopesCachedCounter.WithLabelValues("hit").Inc()
 	} else {
-		w.logger.Debug("cached w envelope", zap.String("envelopeHash", recvMessage.Hash().Hex()))
+		w.logger.Info("cached w envelope", zap.String("envelopeHash", recvMessage.Hash().Hex()))
```
richard-ramos: Should these be Debug too?

wakuv2/waku.go
```diff
@@ -1462,7 +1366,7 @@ func (w *Waku) processQueue() {
 
 			// If not matched we remove it
 			if !matched {
-				w.logger.Debug("filters did not match", zap.String("hash", e.Hash().String()), zap.String("contentTopic", e.ContentTopic.ContentTopic()))
+				w.logger.Info("filters did not match", zap.String("hash", e.Hash().String()), zap.String("contentTopic", e.ContentTopic.ContentTopic()))
```
richard-ramos: Debug?

richard-ramos (approved):
Nice! :rocket:

wakuv2/common/filter.go
vitvly: `ContentTopics` is now a map, so `topic` iterates through the map's keys.

wakuv2/waku.go
vitvly: Fixed.

wakuv2/waku.go
vitvly: Yes, fixed.

wakuv2/waku.go
vitvly: Agree, yes.

wakuv2/waku.go
vitvly: Fixed.

igor-sirotin:
In case someone is facing errors with IDEs after this, add `gowaku_no_rln` to custom build tags.
And in case of GoLand, also go to `Edit configurations template` and check `Use all custom build tags` .

Ping me for more instructions if needed.

protocol/communities_messenger_token_permissions_test.go
```diff
@@ -108,7 +108,7 @@ func (tckd *TestCommunitiesKeyDistributor) waitOnKeyDistribution(condition func(
 }
 
 func TestMessengerCommunitiesTokenPermissionsSuite(t *testing.T) {
-	suite.Run(t, new(MessengerCommunitiesTokenPermissionsSuite))
+	//suite.Run(t, new(MessengerCommunitiesTokenPermissionsSuite))
```
osmaczko, line 111: Was it merged on purpose :thinking: ?

vitvly: Not quite. The test is quite unreliable, had to comment it away in order to merge. Wanted to look into the test and restore it in the separate PR, but got delayed a bit. Sorry for not communicating.

Added a draft PR https://github.com/status-im/status-go/pull/4123, trying out some things. I don't yet understand why community tests use wakuv1 though.

vitvly: Ah, i see, probably just because it'll use same instance for different messengers.

osmaczko: > Not quite. The test is quite unreliable, had to comment it away in order to merge. Wanted to look into the test and restore it in the separate PR, but got delayed a bit. Sorry for not communicating.

That's what I thought, thanks for update :pray:

PR #4083, opened 2023-10-02
fix: use proper migrations for protocol's test database

osmaczko:
- use `appdatabse.DbInitializer{}` in tests to ensure consistent migrations

- remove protocol's open database functions due to improper initialization caused by missing node config migration

- introduce `PushNotificationServerConfig` to resolve cyclic dependency issues


jrainville (approved):
LGTM though I'd wait for someone else to review since I'm not an expert at all in any of what you changed :sweat_smile:

IvanBelyakoff (commented):
Great work, Patrik! I have one question though. Before protocol tests only ran protocol migrations without appDB migrations. Obviously they are independent. I believe that was faster and more clean way of testing protocol migrations only. Now appDB migrations will run first and then protocol migrations. Though this mimics real app behavior, it slows down testing and make it dependent on app db migrations. I doubt myself which approach I like more, both have its benefits. But I'd go with the former one, just would implelement `DbInitializer`  interface for protocol which would call Open (and migrate internally) like it was before. It would unify the old approach with the new one, while keeping protocll migrations separated.
Wdyt?

osmaczko:
> Great work, Patrik! I have one question though. Before protocol tests only ran protocol migrations without appDB migrations. Obviously they are independent.

These changes were motivated by an issue I encountered while working on another PR. The problem is that protocol migrations are no longer independent of appDB migrations. In other words, to function properly, protocol migrations require appDB migrations. Here's the protocol migration I introduced: [migration](https://github.com/status-im/status-go/blob/57ffb03df432d90dc3bbd024cae886a090e7e3dd/protocol/migrations/sqlite/1695318401_community_control_node_sync.up.sql) and this is the error I encountered:
```
database_test.go:89: 
        	Error Trace:	/home/jenkins/workspace/status-go_prs_tests_PR-4068/services/communitytokens/database_test.go:89
        	            				/home/jenkins/workspace/status-go_prs_tests_PR-4068/vendor/github.com/stretchr/testify/suite/suite.go:187
        	            	no such table: shhext_config in line 0: CREATE TABLE communities_control_node (
        	            	    community_id BLOB NOT NULL PRIMARY KEY ON CONFLICT REPLACE,
        	            	    clock INT NOT NULL,
        	            	    installation_id VARCHAR NOT NULL
        	            	);
        	            	
        	            	INSERT INTO
        	            	    communities_control_node (community_id, clock, installation_id)
        	            	SELECT
        	            	    c.id,
        	            	    1 AS clock,
        	            	    s.installation_id
        	            	FROM
        	            	    communities_communities AS c
        	            	    JOIN shhext_config AS s
        	            	WHERE
        	            	    c.private_key IS NOT NULL
        	            	    AND c.private_key != '';
        	            	
        	            	failed to migrate
```

IvanBelyakoff (approved):

IvanBelyakoff:
It would be nice to implement `DbInitializer` interface for protocol which calls `appdatabase.DbInitializer{}` and `sqlite.Migrate()` to hide the logic and avoid code duplication in tests

PR #4089, opened 2023-10-03
feat: use Alchemy API key for Ethereum main/goerli

dlipicar:
Part of https://github.com/status-im/status-desktop/issues/12312

Use the new Alchemy API keys for Ethereum Mainnet and Goerli

Also implemented some small fixes:
- Failure to fetch community metadata is considered a "soft failure" (owner node could be down, shouldn't prevent user from loading remaining collecitbles)
- Properly parse Alchemy struct field with non-strict type


jakubgs (approved):
Always good to have more logging for issues like that.

PR #4092, opened 2023-10-03
feat: add sepolia toggle

alaibe:


stefandunca (approved):
👌🏼

rpc/network/network.go
```diff
@@ -5,9 +5,14 @@ import (
 	"database/sql"
 	"fmt"
 
+	"github.com/status-im/status-go/multiaccounts/accounts"
 	"github.com/status-im/status-go/params"
 )
 
+const SepoliaChainID = 11155111
+
+var GoerliChainIDs = []uint64{5} //, 420, 421613}
```
Cuteivist, line 15: Looks like a bug

alaibe: It is not, goal is to extend this

Cuteivist: Ok, make sense. Maybe you could add some detailed TODO comment here about that?

Cuteivist (approved):

services/wallet/api.go
```diff
@@ -428,7 +428,11 @@ func (api *API) DeleteEthereumChain(ctx context.Context, chainID uint64) error {
 
 func (api *API) GetEthereumChains(ctx context.Context) ([]*network.CombinedNetwork, error) {
 	log.Debug("call to GetEthereumChains")
-	return api.s.rpcClient.NetworkManager.GetCombinedNetworks()
+	isSepoliaEnabled, err := api.s.accountsDB.GetIsSepoliaEnabled()
```
saledjenic: @alaibe this is not needed here, you have access to `nm.accountsDB` from the Manager and that can be resolved there.

saledjenic: Actually, that's what you're doing, seems you just forgot to remove `isSepoliaEnabled` param from the `GetCombinedNetworks` function declaration, since it's not used there.

multiaccounts/settings/columns.go
```diff
@@ -392,6 +392,11 @@ var (
 		dBColumnName:   "test_networks_enabled",
 		valueHandler:   BoolHandler,
 	}
+	IsSepoliaEnabled = SettingField{
```
saledjenic, line 395: @alaibe do we want/plan to sync this across devices or that's just a local setting?

alaibe: Just local setting!

services/wallet/api.go
alaibe: indeed 🤦

saledjenic (approved):

PR #4093, opened 2023-10-03
feat: send community requests in non protected pubsub topic

richard-ramos:
Instead of using the default pubsub topic, with this PR, community requests for communities with shards will arrive in an specific pubsub topic

### DIFF
--- a/protocol/common/shard.go
+++ b/protocol/common/shard.go
@@ -3,7 +3,12 @@ package common
 import (
 	"github.com/status-im/status-go/protocol/protobuf"
 	"github.com/status-im/status-go/protocol/transport"
+	"github.com/waku-org/go-waku/waku/v2/protocol/relay"
 )
+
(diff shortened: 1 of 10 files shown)


protocol/common/shard.go
```diff
@@ -43,5 +48,15 @@ func (s *Shard) Protobuffer() *protobuf.Shard {
 	}
 }
 
-const MainStatusShard = 16
-const UndefinedShardValue = 0
+func DefaultNonProtectedPubsubTopic(shard *Shard) string {
```
vitvly, line 51: Not really important - but do we plan to extract some data from `shard` variable in the future? Otherwise we only care whether its `nil` or not.

vitvly (approved):

protocol/common/shard.go
richard-ramos: for now we're only interested in it being nil or not!

PR #4095, opened 2023-10-04
[#3930] Make sure eth_getLogs for outgoing transfers are not executed…

rasom:
needed for #3930

Tiny optimisation to avoid separate `eth_getLogs` request for outgoing transfers during scanning of ERC20 tail. 
Also PR adds an option to check the number of RPC requests in tests.

status: ready

### DIFF
--- a/services/wallet/transfer/commands_sequential.go
+++ b/services/wallet/transfer/commands_sequential.go
@@ -132,7 +132,7 @@ func (c *findBlocksCommand) ERC20ScanByBalance(parent context.Context, fromBlock
 			if fromBalance.Cmp(toBalance) != 0 {
 				diff := new(big.Int).Sub(to, from)
 				if diff.Cmp(batchSize) <= 0 {
-					headers, err := c.fastIndexErc20(parent, from, to)
+					headers, err := c.fastIndexErc20(parent, from, to, true)
 					if err != nil {
 						return nil, err
 					}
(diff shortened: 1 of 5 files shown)


services/wallet/transfer/commands_sequential.go
```diff
@@ -132,7 +132,7 @@ func (c *findBlocksCommand) ERC20ScanByBalance(parent context.Context, fromBlock
 			if fromBalance.Cmp(toBalance) != 0 {
 				diff := new(big.Int).Sub(to, from)
 				if diff.Cmp(batchSize) <= 0 {
-					headers, err := c.fastIndexErc20(parent, from, to)
+					headers, err := c.fastIndexErc20(parent, from, to, true)
```
IvanBelyakoff, line 135: @rasom could you clarify please. We take a batch that is less than batchSize, which means this is the earliest (tail) batch we are checking, but if I get it right, it still can have outgoing transfers too. But you are checking for incoming only. Am I missing something?

rasom: @IvanBelyakoff Those erc20 transfers happen before the first outgoing transfer (including erc20 transfers), which means that only incoming erc20 can occur here. I'm not sure how `batchSize` is relevant to this, that's just the maximum range we allow for `eth_getLogs` request.

IvanBelyakoff (approved):

PR #4102, opened 2023-10-04
feat(pairing): sync 1-1 chats and groups on pairing devices

### LINKED ISSUE #3547: Sync chats list

jrainville:
Fixes #3547

Based on top of https://github.com/status-im/status-go/pull/4073

[pairing-chats.webm](https://github.com/status-im/status-go/assets/11926403/a263bf2e-b7cc-4b17-9043-c51a565b4d9e)


protocol/messenger_chats.go
```diff
@@ -487,10 +477,10 @@ func (m *Messenger) saveChat(chat *Chat) error {
 		chat.Identicon = identicon
 	}
 
-	// Sync chat if it's a new active public chat, but not a timeline chat
-	if !ok && chat.Active && chat.Public() && !chat.ProfileUpdates() && !chat.Timeline() {
-
-		if err := m.syncPublicChat(context.Background(), chat, m.dispatchMessage); err != nil {
+	// Sync chat if it's a new active public, 1-1 or group chat, but not a timeline chat
+	isPublicChat := !chat.Timeline() && !chat.ProfileUpdates() && chat.Public()
+	if !ok && chat.Active && (isPublicChat || chat.OneToOne() || chat.PrivateGroupChat()) {
```
igor-sirotin: Looks like this condition is duplicated here and in `SyncDevices`.
Maybe create a `Chat` method?
```go
func (c *Chat) shouldBeSynced() bool  {
    isPublicChat := !chat.Timeline() && !chat.ProfileUpdates() && chat.Public()
    return chat.Active && (isPublicChat || chat.OneToOne() || chat.PrivateGroupChat())
}
```

protocol/messenger_handler.go
```diff
@@ -766,18 +764,19 @@ func (m *Messenger) HandleSyncInstallationPublicChat(state *ReceivedMessageState
...
+	existingChat.Joined = int64(message.Clock)
 	state.AllChats.Store(chat.ID, chat)
 
 	state.Response.AddChat(chat)
+
+	// TODO do we need to re-register notif or something?
 	// We join and re-register as we want to receive mentions from the newly joined public chat
-	_, err := m.createPublicChat(chat.ID, state.Response)
-	return err
+	// _, err := m.createPublicChat(chat.ID, state.Response)
```
igor-sirotin: hmmm, I have no idea about this.

protocol/protobuf/application_metadata_message.proto
```diff
@@ -27,7 +27,7 @@ message ApplicationMetadataMessage {
     DECLINE_REQUEST_TRANSACTION = 11;
     SYNC_INSTALLATION_CONTACT_V2 = 12;
     SYNC_INSTALLATION_ACCOUNT = 13;
-    SYNC_INSTALLATION_PUBLIC_CHAT = 14;
+    SYNC_INSTALLATION_CHAT = 14;
```
igor-sirotin: I think this is breaking backwards compatibility.

With this change new instances still might receive `SYNC_INSTALLATION_PUBLIC_CHAT` from an older instance and will try to parse it as `SYNC_INSTALLATION_CHAT`, because it's also `14`.

- `SYNC_INSTALLATION_CHAT` should be defined with a new sequential number.
- `14` can be marked as `reserved`, if we're deprecate it, or leave as is if we want some backwards compatibility. But then we should also keep `message SyncInstallationPublicChat` and the it's handler.

igor-sirotin (changes requested):
Thank you, @jrainville!
I'm basically happy with the implementation, but worried about backwards compatibility. Please, check out inline comments.

protocol/protobuf/application_metadata_message.proto
jrainville: Good point. I fixed it by setting 14 as reserved. I don't think it makes sense to still support it, as it's for public chats and we probably don't support public chats anyway.

protocol/chat.go
```diff
@@ -282,6 +282,10 @@ func (c *Chat) IsActivePersonalChat() bool {
 	return c.Active && (c.OneToOne() || c.PrivateGroupChat() || c.Public()) && c.CommunityID == ""
 }
 
+func (c *Chat) IsPublicChat() bool {
```
igor-sirotin: I'm a bit confused with `IsPublicChat()` and `Public()` 😵‍💫
What I meant is to have `ShouldBeSynced` method. But I guess you have some objections 🙂

igor-sirotin (approved):

protocol/chat.go
jrainville: > What I meant is to have ShouldBeSynced method. But I guess you have some objections 🙂

My bad, I didn't understand you the first time. I didn't check the code snippet lol

jrainville:
@saledjenic you can review this one too please

protocol/chat.go
```diff
@@ -275,6 +275,11 @@ func (c *Chat) IsActivePersonalChat() bool {
 	return c.Active && (c.OneToOne() || c.PrivateGroupChat() || c.Public()) && c.CommunityID == ""
 }
 
+func (c *Chat) shouldBeSynced() bool {
+	isPublicChat := !c.Timeline() && !c.ProfileUpdates() && c.Public()
+	return c.Active && (isPublicChat || c.OneToOne() || c.PrivateGroupChat())
```
saledjenic: @jrainville the same as in the previous review, we should sync deleted chats as well.

protocol/messenger.go
```diff
@@ -2527,15 +2527,14 @@ func (m *Messenger) SyncDevices(ctx context.Context, ensName, photoPath string,
...
-			err = m.syncPublicChat(ctx, chat, rawMessageHandler)
+		if chat.shouldBeSynced() {
+			err = m.syncChat(ctx, chat, rawMessageHandler)
 			if err != nil {
 				return false
 			}
 		}
 
-		if (isPublicChat || chat.OneToOne() || chat.PrivateGroupChat()) && !chat.Active && chat.DeletedAtClockValue > 0 {
+		if (chat.Public() || chat.OneToOne() || chat.PrivateGroupChat()) && !chat.Active && chat.DeletedAtClockValue > 0 {
```
saledjenic: I am pretty sure that we can remove entire this condition (lines 2537 - 2547), cause all that will be properly synced via `err = m.syncChat(ctx, chat, rawMessageHandler)` in the line `2531`.

The same states for the 2 blocks below, that can be removed as well:
- lines 2551 - 2556
- lines 2558 - 2563

Also in terms of local pairing these messages should arrive to the destination in the order they are sent, but this `SyncDevice` function may be used to broadcast sync data over the wire in that case no guarantee of the message order and in that case handling "messages read" or "clear history" related sync messages doesn't have point if we're not sure that the chat those sync messages are referring to is not received.

saledjenic (commented):
@jrainville in overall code looks good, the comments I added are more for a short discussion, just to confirm if the points I raised make sense or not.

protocol/messenger.go
jrainville: I removed the block at 2537. I agree that it's useless now that we add `Active` in the Chat itself.

I get your point about the order of message for "message read" and "clear history", but in case it does arrive in order, isn't it better to have it than not?

saledjenic: > in case it does arrive in order, isn't it better to have it than not?

I would say that we don't need them at all, but also keeping them won't ruin anything. The reason why I think so is cause nothing has changed in the meantime. 

If we're setting up device B (that has an empty state at this moment) from device A (that has some data, like chats, contacts, accounts, whatever...) when local pairing (syncing) starts there is no way anything to be changed by user's activity that can change the context of the data being synced. Anything changed after the syncing process will be synced, cause at that point in time devices A and B are paired. That means we only need to sync the state, not the events. Sync messages sent via `syncChatMessagesRead` and `syncClearHistory` are initiated by the user's activity.

That was my point, why keep the code, that won't have any effect on the other side, just send a few messages more.

jrainville: Ok yeah I get your point. It makes sense.

jrainville:
@saledjenic I updated as you suggested. You can review again. Thanks

saledjenic (approved):
Thanks 👍

PR #4104, opened 2023-10-05
fix: token balance

alaibe:


stefandunca (approved):
Great finding!

IvanBelyakoff (approved):

dlipicar (approved):

rasom (approved):
My fault

services/wallet/token/token.go
```diff
@@ -783,7 +783,10 @@ func (tm *Manager) GetBalancesAtByChain(parent context.Context, clients map[uint
 		mu.Unlock()
 	}
 
-	for _, client := range clients {
+	for clientIdx := range clients {
+		// Keep the reference to the client. DO NOT USE A LOOP, the client will be overrident in the coroutine
```
rasom: overriden

rasom: same for other comments

rasom:
Just wondering, is there a safer way to achieve same results (not like I fully understand the reasoning here), or that's an idiomatic way to handle this referencing issue?

alaibe:
@rasom from what I know there is no other way. When there is a coroutine, you need assign variable otherwise they will be rewrite.
That's a bit naughty but according to golang this is an optimization

@stefandunca shared this with me: https://golang50shad.es/index.html#range_val_update which may explain the problem better than i would ever do

PR #4108, opened 2023-10-05
Validate community description async for token owner communities

cammellos:
This commit changes the behavior of the system so that if a community description with a token owner is detected, it will queue it up and validates it in the background. 
It will check ownership against the contract and if the ownership validates, it will process the description and signal the user.

I am missing one test at the messenger level to validate the behavior, I will be adding that, but in the meantime it would be good to get some eyes on it.
There's also an issue with out of order messages (that's pre-existing), when a community description will arrive after some change of membership occurred. We talked about re-fetching from mailserver, but I would address this separately as it's connected as well on how communities are encrypted, I will pick it up next. 

This will have to be throughlty tested and validated, I might need some help setting things up.

Built on Patryk's code as you can see from the first commit.


api/defaults.go
```diff
@@ -26,6 +26,20 @@ const defaultKeycardPairingDataFile = "/ethereum/mainnet_rpc/keycard/pairings.js
...
 var paths = []string{pathWalletRoot, pathEIP1581, pathDefaultChat, pathDefaultWallet}
 
+const (
+	statusProdFleet = "status.prod"
+	statusTestFleet = "status.test"
+	wakuv2ProdFleet = "wakuv2.prod"
+	wakuv2TestFleet = "wakuv2.test"
+)
+
+var defaultWakuNodes = map[string][]string{
```
osmaczko: Rebase gone wrong? :thinking: Seems unrelated to the changes

cammellos: Sorry, yes, I have added some changes I had locally, will remove

protocol/communities/manager.go
```diff
@@ -348,19 +360,93 @@ func (m *Manager) Start() error {
...
 
 func (m *Manager) runENSVerificationLoop() {
+	ticker := time.NewTicker(500 * time.Millisecond)
+	defer ticker.Stop()
 	go func() {
 		for {
 			select {
 			case <-m.quit:
-				m.logger.Debug("quitting ens verification loop")
+				m.logger.Debug("quitting owner verification loop")
```
```suggestion
				m.logger.Debug("quitting ens verification loop")
```

protocol/communities/manager.go
```diff
@@ -348,19 +360,93 @@ func (m *Manager) Start() error {
...
+	}()
+}
+
+func (m *Manager) runOwnerVerificationLoop() {
+	m.logger.Info("starting owner verification loop")
+	go func() {
+		for {
+			select {
+			case <-m.quit:
+				m.logger.Debug("quitting ens verification loop")
```
```suggestion
				m.logger.Debug("quitting owner verification loop")
```

protocol/communities/manager.go
jrainville: I think this log is still ens

protocol/communities/manager.go
jrainville: this one should say owner

protocol/communities/manager.go
```diff
@@ -348,19 +360,93 @@ func (m *Manager) Start() error {
...
+						signer, description, err := m.unwrapCommunityDescriptionMessage(communityToValidate.payload)
+						if err != nil {
+							m.logger.Error("failed to unwrap community", zap.Error(err))
+							continue
+						}
+
+						chainID := CommunityDescriptionTokenOwnerChainID(description)
+						if chainID == 0 {
+							// This should not happen
+							m.logger.Error("chain id is 0, ignoring")
```
jrainville, line 425: Should you `continue` here?

protocol/communities/manager.go
```diff
@@ -348,19 +360,93 @@ func (m *Manager) Start() error {
...
+							continue
+						}
+
+						response, err := m.HandleCommunityDescriptionMessage(signer, description, communityToValidate.payload, ownerPK)
+						if err != nil {
+							m.logger.Error("failed to handle community", zap.Error(err))
+							continue
+						}
+						if response != nil {
+							m.publish(&Subscription{Community: response.Community, TokenCommunityValidated: response})
```
jrainville: Just to be sure, this `publish` is local only right? It doesn't publish to peers?

protocol/communities/manager.go
```diff
@@ -623,6 +709,23 @@ func (m *Manager) ControlledCommunities() ([]*Community, error) {
 	return communities, nil
 }
 
+func (m *Manager) Owned() ([]*Community, error) {
+	communities, err := m.persistence.CommunitiesWithPrivateKey(&m.identity.PublicKey)
```
jrainville, line 741: This function name makes it sound like we it only returns communities for which we have the community's private key. Is that the case?
If so, maybe there is a problem, because you can be the owner of a community because of the owner token while not having its private key.

Maybe this is all just a little naming problem

protocol/communities/manager.go
```diff
@@ -4771,6 +4943,91 @@ func (m *Manager) createCommunityTokenPermission(request *requests.CreateCommuni
...
+	community.setControlNode(pubKey)
+	err = m.persistence.SaveCommunity(community)
+	if err != nil {
+		return nil, err
+	}
+
+	return community, nil
+}
+
+// FIXME: temporary for POC's tests, remove me once communities use timestamp hinted lamport clocks
```
jrainville: I think @osmaczko 's PR was merged, so maybe this can be removed?

protocol/communities/manager.go
```diff
@@ -4771,6 +4943,91 @@ func (m *Manager) createCommunityTokenPermission(request *requests.CreateCommuni
...
+
+	err = m.persistence.SaveCommunity(community)
+	if err != nil {
+		return nil, err
+	}
+
+	return community, nil
+}
+
+// TODO: implement me
```
jrainville: Do you plan to implement this in this PR or should we open a new issue and do it after?

jrainville (commented):
Partial review because i need to take a break. 

This is massive work. Good job.

I'll finish reviewing later.

protocol/communities/manager.go
cammellos: Correct, it publishes only to `messenger` that can handle the message then and signal to the client

protocol/communities/manager.go
cammellos: Yes, you would, but you would have the private key of the community, or rather, the contract will be set to your (or any public key you have access), and then you can `SetPrivateKey` of the community. 
So the function is checking that the community has private key set, and that you are the owner of that community (i.e the controlNode is the public key corresponding to the private key you have set on the community)
hope that it's clear, not sure if I explained it very clearly :)

protocol/communities/manager.go
cammellos: oh, it's implemented in a different function, can be removed, good point

protocol/communities/manager.go
cammellos: Yes :)

protocol/communities/manager.go
jrainville: Ok I get it. The name is  a bit confusing when you know the old way of doing things, but it's ok

protocol/communities/manager.go
```diff
@@ -1325,21 +1462,48 @@ func (m *Manager) HandleCommunityDescriptionMessage(signer *ecdsa.PublicKey, des
...
+	} else {
+		// we only queue if the clock is greater
+		if shouldQueue && community.config.CommunityDescription.Clock < description.Clock {
+			return nil, m.Queue(signer, community, description, payload)
+		}
+	}
+
+	// Override verified owner
+	if verifiedOwner != nil {
+		community.config.ControlNode = verifiedOwner
```
```suggestion
		community.setControlNode(verifiedOwner)
```

protocol/communities/manager.go
osmaczko: Yes, this should be removed.

protocol/communities/manager_test.go
```diff
@@ -1538,3 +1547,350 @@ func (s *ManagerSuite) buildCommunityWithChat() (*Community, string, error) {
...
+		case event := <-subscription:
+			if event.Community == nil || event.TokenCommunityValidated == nil {
+				continue
+			}
+			published = true
+		case <-time.After(2 * time.Second):
+			s.FailNow("no subscription")
+		}
+	}
+
```
osmaczko, line 1633: IMO `community.ControlNode() == owner`` would be worth checking here as well.

protocol/communities/manager.go
```diff
@@ -360,7 +374,80 @@ func (m *Manager) runENSVerificationLoop() {
...
+						chainID := CommunityDescriptionTokenOwnerChainID(description)
+						if chainID == 0 {
+							// This should not happen
+							m.logger.Error("chain id is 0, ignoring")
+							continue
+						}
+
+						m.logger.Info("validating community", zap.String("id", types.EncodeHex(communityToValidate.id)), zap.String("signer", common.PubkeyToHex(signer)))
 
+						ctx, cancel := context.WithTimeout(context.Background(), time.Second*3)
```
jrainville, line 431: Is this a 3 seconds "sleep"? What is it for?

jrainville (commented):
Second round of reviews done. Turns out most of the remaining files were protobufs, so it was pretty easy to review :smile: 

Anyway, it all looks very good. Nice work

osmaczko (commented):
Very nice! Thanks for providing solid tests :ok_hand:

protocol/communities/manager.go
osmaczko: It is a timeout for the contract call.

protocol/communities/community.go
```diff
@@ -1255,6 +1256,22 @@ func (o *Community) PrivateKey() *ecdsa.PrivateKey {
...
+
+func (o *Community) ControlNode() *ecdsa.PublicKey {
+	return o.config.ControlNode
+}
+
+func (o *Community) setControlNode(pubKey *ecdsa.PublicKey) {
+	if pubKey != nil {
+		o.config.ControlNode = pubKey
+	}
+}
```
mprakhov, line 1298: will this be changed? In my understanding, control node can be only one device, but in this case all devices will be treated as a control node because they all will have the same pubkey

mprakhov (commented):
Great work!

protocol/communities/manager.go
```diff
@@ -1310,12 +1417,42 @@ func (m *Manager) DeleteCategory(request *requests.DeleteCommunityCategory) (*Co
 	return changes.Community, changes, nil
 }
 
-func (m *Manager) HandleCommunityDescriptionMessage(signer *ecdsa.PublicKey, description *protobuf.CommunityDescription, payload []byte) (*CommunityResponse, error) {
+func (m *Manager) Queue(signer *ecdsa.PublicKey, community *Community, description *protobuf.CommunityDescription, payload []byte) error {
```
mprakhov: description is not used here

protocol/communities/manager.go
```diff
@@ -1310,12 +1417,42 @@ func (m *Manager) DeleteCategory(request *requests.DeleteCommunityCategory) (*Co
...
+
+	m.logger.Info("queuing community", zap.String("id", community.IDString()), zap.String("signer", common.PubkeyToHex(signer)))
+
+	communityToValidate := communityToValidate{
+		id:         community.ID(),
+		clock:      community.config.CommunityDescription.Clock,
```
mprakhov: Not clear which role the clock is playing here. I do not see it being used in the run loop, also this is a previous community clock and not a new one

protocol/communities/community.go
cammellos: I believe that's fine @osmaczko is working on some changes that would allow this. The key is set to the chat key of the user, so it's already shared across devices.

protocol/communities/manager.go
cammellos: We pull them from the database in reverse clock order, and as soon as one validates, we stop validating any other with the same id, good point about the clock, I'll fix it :)

cammellos:
@jrainville @osmaczko @endulab @mprakhov Updated and addressed feedback.

I have added a test that should cover the whole flow https://github.com/status-im/status-go/blob/9a5e86049e1d12dc61d9422d8d35ad6c36a71a34/protocol/communities_messenger_signers_test.go#L110 

Let me know if I should cover more cases etc

protocol/communities/manager.go
```diff
@@ -4771,6 +4956,70 @@ func (m *Manager) createCommunityTokenPermission(request *requests.CreateCommuni
...
+
+	return community, nil
+}
+
+func (m *Manager) UpdatePrivateKeyAndControlNode(communityID types.HexBytes, pk *ecdsa.PrivateKey) (*Community, error) {
+	community, err := m.UpdatePrivateKey(communityID, pk)
+	if err != nil {
+		return nil, err
+	}
+	keyBytes := common.PubkeyToHexBytes(&pk.PublicKey)
```
endulab: Maybe I do not see something but does it make sense to marshal pubkey here and then unmarshal in UpdateControlNode?

protocol/communities/community.go
```diff
@@ -2312,6 +2312,31 @@ func (o *Community) ValidateEvent(event *CommunityEvent, signer *ecdsa.PublicKey
 }
 
 func (o *Community) MemberCanManageToken(member *ecdsa.PublicKey, token *community_token.CommunityToken) bool {
-	return o.IsMemberOwner(member) || (o.IsMemberTokenMaster(member) &&
+	return o.IsMemberOwner(member) || o.IsControlNode() || (o.IsMemberTokenMaster(member) &&
```
mprakhov, line 2329: I don't think that we need to do a check of `o.IsControlNode()` as controlNode must be by default and Owner. It if is not right - then there is a bug somewhere

protocol/communities_messenger_helpers_test.go
```diff
@@ -103,11 +118,21 @@ func (c *CollectiblesServiceMock) GetAssetContractData(chainID uint64, contractA
 }
 
 func (c *CollectiblesServiceMock) SetMockCollectibleContractData(chainID uint64, contractAddress string, collectible *communitytokens.CollectibleContractData) {
+	fmt.Println("SETTING", collectible)
```
mprakhov: do we need this fmt.Println? (here and below)

protocol/communities_messenger_signers_test.go
```diff
@@ -128,102 +154,93 @@ func (s *MessengerCommunitiesSignersSuite) TestControlNodeUpdate() {
...
 		func(r *MessengerResponse) bool {
-			return len(r.Communities()) > 0 && r.Communities()[0].Description().Identity.Description == bobDescr
+			return published
 		},
 		"no communities",
 	)
 	s.Require().NoError(err)
 
-	// alice accepts community update
+	// We check the control node is correctly set on john
```
mprakhov: to alice

protocol/communities/manager.go
```diff
@@ -1325,21 +1467,55 @@ func (m *Manager) HandleCommunityDescriptionMessage(signer *ecdsa.PublicKey, des
 	// Workaround for https://github.com/status-im/status-desktop/issues/12188
 	HydrateChannelsMembers(types.EncodeHex(id), description)
 
+	// We should queue only if the community has a token owner, and the owner has been verified
```
osmaczko, line 1523:
```suggestion
	// We should queue only if the community has a token owner, and the owner has not been verified
```

protocol/communities_messenger_signers_test.go
```diff
@@ -128,102 +154,93 @@ func (s *MessengerCommunitiesSignersSuite) TestControlNodeUpdate() {
...
+	johnCommunity, err := s.john.communitiesManager.GetByIDString(community.IDString())
+	s.Require().NoError(err)
+	s.Require().NotNil(johnCommunity)
+	s.Require().True(common.IsPubKeyEqual(johnCommunity.ControlNode(), &s.alice.identity.PublicKey))
+	s.Require().False(johnCommunity.IsControlNode())
```
mprakhov, line 225: need to check, that Alice became an Owner, as the control node always has a role Owner
and john should loose his role and get it based on the community permission

protocol/communities_messenger_helpers_test.go
```diff
@@ -95,6 +109,7 @@ func (c *CollectiblesServiceMock) GetCollectibleContractData(chainID uint64, con
 }
 
 func (c *CollectiblesServiceMock) GetAssetContractData(chainID uint64, contractAddress string) (*communitytokens.AssetContractData, error) {
+	fmt.Println("CHAINID", chainID, "contract", contractAddress, c.Assets)
```
osmaczko: :broom:

protocol/communities_messenger_helpers_test.go
osmaczko: :broom:

protocol/communities_messenger_helpers_test.go
```diff
@@ -103,11 +118,21 @@ func (c *CollectiblesServiceMock) GetAssetContractData(chainID uint64, contractA
...
 	if c.Collectibles == nil {
 		c.Collectibles = make(map[uint64]map[string]*communitytokens.CollectibleContractData)
 	}
 	c.Collectibles[chainID] = make(map[string]*communitytokens.CollectibleContractData)
 	c.Collectibles[chainID][contractAddress] = collectible
+	fmt.Println(c.Collectibles)
+}
+
+func (c *CollectiblesServiceMock) SafeGetSignerPubKey(ctx context.Context, chainID uint64, communityID string) (string, error) {
+	fmt.Println("GETTING SIGNER PUBKEY", c.Signers)
```
osmaczko: :broom:

protocol/communities_messenger_signers_test.go
```diff
@@ -0,0 +1,246 @@
...
+	s.Require().NoError(err)
+
+	// make alice the control node
+	transaction := transactions.SendTxArgs{}
+	_, err = s.alice.SetCommunitySignerPubKey(context.Background(), community.ID(), chainID, communityAddress, transaction, "password", common.PubkeyToHex(&s.alice.identity.PublicKey))
+	s.Require().NoError(err)
+
+	var published bool
+
+	// We check the updated description is published as a signal
```
osmaczko, line 198: Why do we check it as signal? Shouldn't John and Bob populate `MessengerResponse` with updated community in `RetrieveAll`?

protocol/messenger_communities.go
```diff
@@ -231,6 +233,25 @@ func (m *Messenger) handleCommunitiesHistoryArchivesSubscription(c chan *communi
 				if sub.ImportingHistoryArchiveMessagesSignal != nil {
 					m.config.messengerSignalsHandler.ImportingHistoryArchiveMessages(sub.ImportingHistoryArchiveMessagesSignal.CommunityID)
 				}
+				if sub.TokenCommunityValidated != nil {
```
osmaczko: Not sure if it belongs to `handleCommunitiesHistoryArchivesSubscription`

protocol/messenger_communities.go
```diff
@@ -297,6 +318,7 @@ func (m *Messenger) handleCommunitiesSubscription(c chan *communities.Subscripti
 					return
 				}
 				if sub.Community != nil {
+					fmt.Println("PUBLISHING ORG", sub.Community.Description().TokenPermissions)
```
osmaczko: :broom:

protocol/communities/manager.go
```diff
@@ -360,7 +377,82 @@ func (m *Manager) runENSVerificationLoop() {
 					return
 				}
 				m.logger.Info("received records", zap.Any("records", records))
+			}
+		}
+	}()
+}
+
+func (m *Manager) runOwnerVerificationLoop() {
```
mprakhov, line 392: could you please check `TestTokenMasterReceiveOwnerTokenFromControlNode` test?
I was unable to make it work after the control node deployed ownerToken and owner permission
It seems this loop is not starting
P.S. I did additional set up for test case:
```
s.collectiblesServiceMock.SetSignerPubkeyForCommunity(community.ID(), common.PubkeyToHex(&s.controlNode.identity.PublicKey))
communities.SetValidateInterval(50 * time.Millisecond)
```

cammellos: I need to fix the test, yes, it's because you don't get it through retrieveall, but it will be published eventually as a signal, so I think I need to adjust the test, I will look into it, well spotted, thank you

protocol/communities_messenger_signers_test.go
cammellos: that is a bit less clear, we discussed with @osmaczko , but it's alice responsibility to then update the description and remove John from the member list as having token ownership (as I understand, there's some cases where we might want to have a separate owner from control node), this PR is not currently addressing that though (maybe in a separate?)

protocol/communities_messenger_signers_test.go
cammellos: no, maybe I misunderstand the question, but in this case, the check is asynchronous (it's queued), and so it won't be published through RetrieveAll, but as a signal.
Did I understand the question correctly?

osmaczko: Ah, you are right :+1:

FYI: I have created a helper for that case: https://github.com/status-im/status-go/blob/19771c422da460e870aa53701589be80cde1ff52/protocol/messenger_testing_utils.go#L53 but the PR with that is in deep draft. I'll use it here once I merge it.

cammellos:
@endulab @osmaczko @mprakhov @jrainville I have addressed (I hope I haven't missed any) the feedback, it should be ready to be reviewed again, thank you!

protocol/communities_messenger_signers_test.go
cammellos: neat, we need to test more signals!

protocol/communities_messenger_signers_test.go
mprakhov: Agree, not in this. I have PR with kicking all members

osmaczko (approved):

osmaczko:
Massive work, thanks @cammellos :rocket:

mprakhov:
Could you please help a little: how can I listen for a signal in tests (do we have a working example)? Maybe I will have a possibility to help you with fixing the tests

cammellos:
> Could you please help a little: how can I listen for a signal in tests (do we have a working example)? Maybe I will have a possibility to help you with fixing the tests

I think the test that was broken now is testing a bit less (we only validate that the token is received, but we don't check the exact type), but the test I have added covers that flow completely (a token owner is sent, and we make sure it's received, and processed correctly). Or there's still some concern on your part?

mprakhov:
> > Could you please help a little: how can I listen for a signal in tests (do we have a working example)? Maybe I will have a possibility to help you with fixing the tests
> 
> I think the test that was broken now is testing a bit less (we only validate that the token is received, but we don't check the exact type), but the test I have added covers that flow completely (a token owner is sent, and we make sure it's received, and processed correctly). Or there's still some concern on your part?

No concerns, if you think it is covered already, it can be outdated in this case and we can remove it
I just wonder if `TestTokenMasterReceiveOwnerTokenFromControlNode` pass?

osmaczko:
> Could you please help a little: how can I listen for a signal in tests (do we have a working example)? Maybe I will have a possibility to help you with fixing the tests

https://github.com/status-im/status-go/pull/4108#discussion_r1358317429

cammellos:
> No concerns, if you think it is covered already, it can be outdated in this case and we can remove it
I just wonder if TestTokenMasterReceiveOwnerTokenFromControlNode pass?

Oh I missed that one, I updated it in the same way, so we check that the community description as in queued, similarly to the other test that was failing, the full flow it should be covered in the new test (otherwise we would have to stub/mock all the interactions with blockchain)

mprakhov (approved):
Looks great 👍

jrainville (approved):
Awesome work! Let's ship it! :rocket:

endulab (approved):

protocol/communities/manager.go
```diff
@@ -1355,12 +1475,42 @@ func (m *Manager) DeleteCategory(request *requests.DeleteCommunityCategory) (*Co
...
+	err := m.persistence.SaveCommunityToValidate(communityToValidate)
+	if err != nil {
+		m.logger.Error("failed to save community", zap.Error(err))
+		return err
+	}
+
+	return nil
+}
+
+func (m *Manager) HandleCommunityDescriptionMessage(signer *ecdsa.PublicKey, description *protobuf.CommunityDescription, payload []byte, shard *common.Shard, verifiedOwner *ecdsa.PublicKey) (*CommunityResponse, error) {
```
mprakhov, line 1498: please, double-check who is calling this fn...
I have a crash in one of my tests in `handleChatMessage`, `m.handleWrappedCommunityDescriptionMessage` returns nil due to a queue

PR #4109, opened 2023-10-05
fix(wallet): new blocks were not found

IvanBelyakoff:
New block were not found caused by range value not set


dlipicar (approved):

rasom (approved):

rasom:
Although obviously I forgot to add it there, still would be caught if we'd have any test which would cover that piece of code.

PR #4110, opened 2023-10-05
chore: improve requesting contact/community info from mailserver

igor-sirotin:
1. Use `defer` to forget requests
2. Return error in case of timeout
3. Minor refactoring


igor-sirotin:
finally test passed 😅

jrainville (approved):

PR #4118, opened 2023-10-06
fix(communities): make sure to join the community if we are a member

### LINKED ISSUE #12368: Client doesn't recognize that user is still member of a community after re-import from seed
# Bug Report

## Description

Experienced by Jakub.

He re-imported his account from seed phrase on 0.14.0 and even though he still appears in the member list for the Status CC community, he cannot post and has to request access again.

## Steps to reproduce

0. Be part of a community
1. Import that account in another instance using seed phrase
2. Import the community or have it shared with you
3. See that you are part of the member's list but still can't post

#### Expected behavior

You get access to the community without having to request access again

#### Actual behavior

You need to request to join again

### Additional Information

- Status desktop version: 0.14.0
- Operating System: NixOS

jrainville:
Fixes https://github.com/status-im/status-desktop/issues/12368

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -1417,6 +1417,11 @@ func (m *Manager) handleCommunityDescriptionMessageCommon(community *Community, 
 	}
 	community.config.EventsData = nil
 
+	// Set Joined if we are part of the member list
+	if !community.Joined() && community.hasMember(&m.identity.PublicKey) {
+		changes.ShouldMemberJoin = true
+	}
+
 	err = m.persistence.SaveCommunity(community)
 	if err != nil {
 		return nil, err
--- a/protocol/communities_messenger_test.go
+++ b/protocol/communities_messenger_test.go
@@ -2018,7 +2018,7 @@ func (s *MessengerCommunitiesSuite) TestShareCommunity() {
 
 	inviteMessage := "invite to community testing message"
 
-	// Create an community chat
+	// Create an community
 	response, err := s.bob.CreateCommunity(description, true)
 	s.Require().NoError(err)
 	s.Require().NotNil(response)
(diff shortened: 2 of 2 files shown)


protocol/communities/manager.go
```diff
@@ -1417,6 +1417,11 @@ func (m *Manager) handleCommunityDescriptionMessageCommon(community *Community,
 	}
 	community.config.EventsData = nil
 
+	// Set Joined if we are part of the member list
+	if !community.Joined() && community.hasMember(&m.identity.PublicKey) {
```
MishkaRogachev, line 1421: If one of the two conditions doesn't work here, I think the problem may be elsewhere

protocol/communities_messenger_test.go
```diff
@@ -2018,6 +2018,61 @@ func (s *MessengerCommunitiesSuite) TestShareCommunity() {
...
+	s.Require().Len(response.Messages(), 1)
+
+	// Add bob to contacts so it does not go on activity center
+	bobPk := common.PubkeyToHex(&s.bob.identity.PublicKey)
+	request := &requests.AddContact{ID: bobPk}
+	_, err = s.alice.AddContact(context.Background(), request)
+	s.Require().NoError(err)
+
+	// Pull message and make sure org is received
+	err = tt.RetryWithBackOff(func() error {
```
osmaczko, line 2047: `WaitOnMessengerResponse` should be used instead, it is superior to `RetryWithBackOff`.

osmaczko (approved):

ilmotta (approved):

MishkaRogachev (approved):

protocol/communities/manager.go
jrainville: Are you saying I should do this elsewhere? I don't fully get your point

PR #4123, opened 2023-10-08
Tests: CommunitiesTokenPermissionsSuite - use wakuv2 transport

vitvly:
Use waku v2 for transport in one set of community tests.

The Suite struct contains a `useLocalWaku` flag (by default set to true). If on, only local libp2p peers will be used. At least in local testing, suite test runtime decreases from ~110 seconds to ~57 seconds.

### DIFF
--- a/protocol/communities_messenger_helpers_test.go
+++ b/protocol/communities_messenger_helpers_test.go
@@ -5,6 +5,7 @@ import (
 	"crypto/ecdsa"
 	"encoding/json"
 	"errors"
+	"os"
 
 	"io/ioutil"
 	"sync"
@@ -20,6 +21,7 @@ import (
 	"github.com/status-im/status-go/account"
 	"github.com/status-im/status-go/account/generator"
 	"github.com/status-im/status-go/appdatabase"
+	gethbridge "github.com/status-im/status-go/eth-node/bridge/geth"
 	"github.com/status-im/status-go/eth-node/crypto"
 	"github.com/status-im/status-go/eth-node/types"
 	"github.com/status-im/status-go/multiaccounts"
@@ -34,6 +36,7 @@ import (
 	"github.com/status-im/status-go/services/communitytokens"
 	walletToken "github.com/status-im/status-go/services/wallet/token"
 	"github.com/status-im/status-go/t/helpers"
+	waku "github.com/status-im/status-go/wakuv2"
 	"github.com/status-im/status-go/walletdatabase"
 )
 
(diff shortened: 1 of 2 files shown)


protocol/communities_messenger_helpers_test.go
```diff
@@ -122,6 +126,78 @@ func (c *CollectiblesServiceMock) DeploymentSignatureDigest(chainID uint64, addr
 	return gethcommon.Hex2Bytes("ccbb375343347491706cf4b43796f7b96ccc89c9e191a8b78679daeba1684ec7"), nil
 }
 
+func findFreeUDPPort() (int, error) {
```
igor-sirotin: Hmmm, I see some problems with this.

1. Range of TCP and UDP ports are different. A free TCP port can be busy for UDP.
2. I don't think it makes sense to have an "attempts loop". There're thousands of ports available locally, it's a very rare case if we don't have a free one.

### There's another function available

I found an function here which seems to be a better one:
https://github.com/status-im/status-go/blob/212e74fc1fe7aea31f736f3fbfe32686f73291df/wakuv2/waku.go#L160-L170

Though I think it's better to replace `net.IPv4zero` with `net.IP{127,0,0,1}`. Even though it's quite a fast operation, there's no need to make it publicly available, localhost is safer an more then enough.

### The approach 

Currently we go with this approach:
1. open a socket with port `0`
2. os gives us a free port
3. we remember it
4. we close the socket
5. we pass the remembered port to waku config
6. waku opens a udp socket on given port

I guess this will work fine in 99,999% of cases, but there's still a possibility that between clauses 4 and 6 the OS will give this port to some other socket. Then Waku will fail binding to this port.

I'm not sure what's the purpose of this port (I guess it's a discovery port?), probably it makes sense to have it fixed from the top level. But maybe for tests sake we could update waku library so that it assumes `port 0` as to select on automatically? And then if we need the port on top level, we just take it out with a method.

This would fix that 0.001% cases of problems. But what's more important, it will make the code clearer 🙂

igor-sirotin: Wow, it's actually already done and even used by your code!. `getUsableUDPPort` is automatically called inside `waku.New(...)`:
https://github.com/status-im/status-go/blob/212e74fc1fe7aea31f736f3fbfe32686f73291df/wakuv2/waku.go#L188-L193

vitvly: Thank you, wow, i missed that func! So i guess it's fine to just rely on `getUsableUDPPort` inside `Waku.New`, correct? And leave `Config.UDPPort` at 0.

protocol/communities_messenger_helpers_test.go
```diff
@@ -122,6 +126,78 @@ func (c *CollectiblesServiceMock) DeploymentSignatureDigest(chainID uint64, addr
...
+		case <-time.After(30 * time.Second):
+			s.Require().Fail("timeout elapsed")
+		case <-connStatusChan:
+			// proceed, peers found
+			close(connStatusChan)
+		}
+	}
+
+	wakuWrapper := gethbridge.NewGethWakuV2Wrapper(shh)
+	return wakuWrapper
```
igor-sirotin:
```suggestion
	return gethbridge.NewGethWakuV2Wrapper(shh)
```

protocol/communities_messenger_helpers_test.go
igor-sirotin: yep, looks like that!
Though this should be done for the first instance only. For 2 other instances we should set the port as it's set in the first one.

vitvly: Sorry, don't understand - why should we set identical udp ports for different Waku instances?

protocol/communities_messenger_helpers_test.go
vitvly: Fixed.

protocol/communities_messenger_helpers_test.go
vitvly: Removed the `findFreeUDPPort` func, relying on Waku's builtin func.

igor-sirotin: > Sorry, don't understand - why should we set identical udp ports for different Waku instances?

Oh, never mind, I think I had a moment of misunderstanding 😄

igor-sirotin (approved):
Cool, thank you! 👍 

I think we should make the process of creating local waku nodes simpler in the future.
E.g. maybe create a function that returns a given number of connected nodes, this would simplify things a lot.

```go
func connectWakuNodes([]*waku.Waku) {
  // move the connecting loop you wrote here
}

func createWakuNetwork(count int,) []*waku.Waku {
   // Create `count` new nodes
   connectWakuNodes(nodes)
}
```

vitvly:
> Cool, thank you! 👍
> 
> I think we should make the process of creating local waku nodes simpler in the future. E.g. maybe create a function that returns a given number of connected nodes, this would simplify things a lot.
> 
> ```go
> func connectWakuNodes([]*waku.Waku) {
>   // move the connecting loop you wrote here
> }
> 
> func createWakuNetwork(count int,) []*waku.Waku {
>    // Create `count` new nodes
>    connectWakuNodes(nodes)
> }
> ```

Yes, good idea, added `createWakuNetwork` func.

PR #4128, opened 2023-10-09
feat: url unfurling settings

### LINKED ISSUE #4130: Add URL unfurling settings
# Problem

Settings logic and requirements explained here:
https://www.figma.com/file/Mr3rqxxgKJ2zMQ06UAKiWL/%F0%9F%92%AC-Chat%E2%8E%9CDesktop?type=design&node-id=23648-94820&mode=design&t=HhJjXK8OTQ9HhPuV-4

igor-sirotin:
Required for https://github.com/status-im/status-desktop/issues/10866
Closes https://github.com/status-im/status-go/issues/4130
_____

1. Add new `url_unfurling_mode` setting
2. Ensure the new property is synced and backed up
3. `UnfurlURLs` automatically handles unfurling settings


multiaccounts/settings/enums.go
```diff
@@ -22,3 +22,12 @@ const (
...
 	ProfilePicturesShowToNone
 )
+
+type URLUnfurlingModeType int
+
+const (
+	URLUnfurlingAlwaysAsk URLUnfurlingModeType = iota + 1
+	URLUnfurlingEnableAll
+	URLUnfurlingDisableAll
+	URLUnfurlingSpecifyForSite // TODO: post-mvp
```
ilmotta: I think we can go by without the dummy constant. We don't know which constants we will need.

I usually call this kind of code zombie code 🧟, because it adds complexity but does nothing. Just adding a nice comment or better yet, creating a GitHub issue is clearer in my experience to document whatever needs to be done in the future.

multiaccounts/settings/sync_protobuf_factories.go
```diff
@@ -539,3 +541,28 @@ func includeWatchOnlyAccountProtobufFactory(value interface{}, clock uint64, cha
...
+	pb := &protobuf.SyncSetting{
+		Type:  protobuf.SyncSetting_URL_UNFURLING_MODE,
+		Value: &protobuf.SyncSetting_ValueInt64{ValueInt64: v},
+		Clock: clock,
+	}
+	rm, err := buildRawSyncSettingMessage(pb, chatID)
+	return rm, pb, err
+}
+
+func urlUnfurlingModeProtobufFactory(value interface{}, clock uint64, chatID string) (*common.RawMessage, *protobuf.SyncSetting, error) {
```
ilmotta: Since Go 1.18 `any` is a type alias to `interface{}`.

protocol/messenger_linkpreview.go
```diff
@@ -137,6 +137,29 @@ func (m *Messenger) UnfurlURLs(httpClient *http.Client, urls []string) ([]common
...
+	switch s.URLUnfurlingMode {
+	case settings.URLUnfurlingDisableAll:
+		return nil, fmt.Errorf("url unfurling is disabled")
+	case settings.URLUnfurlingEnableAll:
+		break
+	case settings.URLUnfurlingAlwaysAsk:
+		// This mode should be handled on the app side
+		// and is considered as equal to URLUnfurlingEnableAll in status-go.
+		break
+	case settings.URLUnfurlingSpecifyForSite:
```
ilmotta: I understand the motivation @igor-sirotin, but well, I think we should only merge code that has some tangible value in production. This dummy code 🧟‍♂️ should never be reached because clients won't use the mode `URLUnfurlingSpecifyForSite`.

protocol/messenger_linkpreview_test.go
```diff
@@ -378,3 +370,53 @@ func (s *MessengerLinkPreviewsTestSuite) Test_UnfurlURLs_Image() {
...
+	requestsCount = 0
+	err = s.m.settings.SaveSettingField(settings.URLUnfurlingMode, settings.URLUnfurlingEnableAll)
+	s.Require().NoError(err)
+
+	linkPreviews, err = s.m.UnfurlURLs(&stubbedClient, []string{u})
+	s.Require().NoError(err)
+	s.Require().Len(linkPreviews, 1)
+
+	// Test `DisableAll`
+	requestsCount = 0
```
ilmotta, line 559: The `requestsCount` is being reset, but we're not asserting anything about it. I guess we want to do the same as `s.Require().Equal(requestsCount, 1)`?

ilmotta (approved):
🚀

protocol/messenger_linkpreview_test.go
igor-sirotin: Hmmm, I'm pretty sure I wrote that, good catch! Now it's back there 👌

protocol/messenger_linkpreview.go
igor-sirotin: 👍 I agree, removed!

multiaccounts/settings/sync_protobuf_factories.go
igor-sirotin: nice, didn't know!

multiaccounts/settings/enums.go
igor-sirotin: I agree, I actually was thinking of removing it, but forgot in the end 🙂 
Dropped now 👍

PR #4134, opened 2023-10-10
Fix checking of ERC20 tail on new blocks scanning

rasom:
This case was not covered by tests. It likely works even now but makes unnecessary requests.

status: ready


stefandunca (approved):
👍🏼

PR #4139, opened 2023-10-11
Add test method TestPinMessageInCommunityChat

### LINKED ISSUE #4138: Tests - setting community.PinMessageAllMembersEnabled is not propagated
# Problem

Trying to set `community.AdminSettings().PinMessageAllMembersEnabled` to *true* in a test method, is not propagated properly to members, resulting in failed `s.user.SendPinMessage()` calls.

shinnok:
Add TestPinMessageInCommunityChat to test that a pin message type can be sent correctly for community\chat members or denied for non-members.

~~Need to fix `community.AdminSettings().PinMessageAllMembersEnabled` setting first, which seems to be ignored for non-owner members.~~

Closes #4138

### DIFF
--- a/protocol/communities_messenger_test.go
+++ b/protocol/communities_messenger_test.go
@@ -498,6 +498,82 @@ func (s *MessengerCommunitiesSuite) TestPostToCommunityChat() {
 	s.Require().True(found)
 }
 
+func (s *MessengerCommunitiesSuite) TestPinMessageInCommunityChat() {
+	ctx := context.Background()
+
+	// Create a community
+	description := &requests.CreateCommunity{
+		Membership:                  protobuf.CommunityPermissions_AUTO_ACCEPT,
+		Name:                        "status",
+		Color:                       "#ffffff",
+		Description:                 "status community description",
+		PinMessageAllMembersEnabled: true,
+	}
+
+	response, err := s.owner.CreateCommunity(description, true)
+	s.Require().NoError(err)
+	s.Require().NotNil(response)
+	s.Require().Len(response.Communities(), 1)
+
+	community := response.Communities()[0]
+	s.Require().NotNil(community)
+	s.Require().Equal(community.AllowsAllMembersToPinMessage(), true)
+
+	// Create a community chat
+	orgChat := &protobuf.CommunityChat{
+		Permissions: &protobuf.CommunityPermissions{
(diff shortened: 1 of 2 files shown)


protocol/communities_messenger_test.go
```diff
@@ -495,6 +495,42 @@ func (s *MessengerCommunitiesSuite) TestPostToCommunityChat() {
...
+	inputMessage.ChatId = chat.ID
+	inputMessage.ContentType = protobuf.ChatMessage_TEXT_PLAIN
+	inputMessage.Text = "pinned message"
+
+	ctx := context.Background()
+	sendResponse, err := s.alice.SendChatMessage(ctx, inputMessage)
+	s.Require().NoError(err)
+	s.Require().Len(sendResponse.Messages(), 1)
+
+	community, err = s.admin.communitiesManager.GetByID(community.ID())
```
shinnok: Grabbing the community again, either via s.admin or via s.[alice|bob] doesn't not yield the CommunityAdminSettings member which was set initially at community creation.

shinnok:
To be merged after https://github.com/status-im/status-go/pull/4064 .

protocol/communities_messenger_test.go
```diff
@@ -484,6 +484,74 @@ func (s *MessengerCommunitiesSuite) TestPostToCommunityChat() {
...
+	pinMessage.ChatId = chat.ID
+	pinMessage.MessageId = inputMessage.ID
+	pinMessage.Pinned = true
+	sendResponse, err = s.bob.SendPinMessage(ctx, pinMessage)
+	s.Require().NoError(err)
+	s.Require().Len(sendResponse.PinMessages(), 1)
+
+	// alice did not join the community, so she should not be able to send the pin message
+	s.advertiseCommunityTo(community, s.alice)
+	sendResponse, err = s.alice.SendPinMessage(ctx, pinMessage)
```
jrainville: This needs Alice to Spectate the community too, so she has access to the Chat.

Also, I recommend creating a new `pinMessage`, as sending the same message with the same ID could cause a clash.

Finally, you can validate that the error message is the one you expect at the end, something like "user can't post".

shinnok: Done. With the latest change, this should fail in the CI until #4064 counterpart is merged to develop.

osmaczko (approved):

PR #4140, opened 2023-10-11
feat: arbitrum sepolia

alaibe:


rpc/network/network.go
```diff
@@ -226,8 +226,17 @@ func (nm *Manager) Get(onlyEnabled bool) ([]*params.Network, error) {
 
 	var results []*params.Network
 	for _, network := range networks {
-		if !isSepoliaEnabled && network.ChainID == SepoliaChainID {
-			continue
+		if !isSepoliaEnabled {
```
dlipicar, line 229: could be a separate function

services/wallet/token/tokenstore.go
```diff
@@ -1679,6 +1679,14 @@ func newDefaultStore() *DefaultStore {
 				Decimals: 18,
 				ChainID:  11155111,
 			},
+			&Token{
```
dlipicar, line 1682: hmm is this the only "important" token available in that chain? not even DAI?

dlipicar (approved):

services/wallet/token/tokenstore.go
alaibe: I haven't found DAI!

saledjenic (approved):

PR #4147, opened 2023-10-11
feat: emit event on received community collectibles

dlipicar:
Whenever a new community collectible is detected on an account, a wallet event will be emitted containing some collectible+community information

### DIFF
--- a/services/wallet/collectibles/commands.go
+++ b/services/wallet/collectibles/commands.go
@@ -23,6 +23,14 @@ const (
 
 type OwnershipState = int
 
+type OwnedCollectibles struct {
+	chainID walletCommon.ChainID
+	account common.Address
+	ids     []thirdparty.CollectibleUniqueID
+}
+
+type OwnedCollectiblesCb func(OwnedCollectibles)
+
 const (
 	OwnershipStateIdle OwnershipState = iota + 1
 	OwnershipStateDelayed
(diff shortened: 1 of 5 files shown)


services/wallet/collectibles/commands.go
```diff
@@ -103,9 +121,14 @@ func (c *periodicRefreshOwnedCollectiblesCommand) loadOwnedCollectibles(ctx cont
 	c.group.Add(command.Command())
 
 	select {
+	case ownedCollectibles := <-receivedCollectiblesCh:
+		if (c.receivedCollectiblesCb) != nil {
```
rasom: why are parentheses needed at `(c.receivedCollectiblesCb)`?

services/wallet/collectibles/commands.go
```diff
@@ -114,24 +137,32 @@ func (c *periodicRefreshOwnedCollectiblesCommand) loadOwnedCollectibles(ctx cont
...
+	walletFeed             *event.Feed
+	receivedCollectiblesCh chan<- OwnedCollectibles
 
 	// Not to be set by the caller
 	partialOwnership []thirdparty.CollectibleUniqueID
 	err              error
 }
 
-func newLoadOwnedCollectiblesCommand(manager *Manager, ownershipDB *OwnershipDB, walletFeed *event.Feed, chainID walletCommon.ChainID, account common.Address) *loadOwnedCollectiblesCommand {
+func newLoadOwnedCollectiblesCommand(
```
rasom, line 152: I'm a bit puzzled how
```
command := newLoadOwnedCollectiblesCommand(c.manager, c.ownershipDB, c.walletFeed, c.chainID, c.account, receivedCollectiblesCh)
```
is better than
```
command := &loadOwnedCollectiblesCommand{c.manager, c.ownershipDB, c.walletFeed, c.chainID, c.account, receivedCollectiblesCh}
```

and why we actually need an extra `newSomething` method in case if there is nothing except passing fields to the structure. But well if that's an idiomatic way to do things then that's just my problem. I also know it was not introduced in this specific PR, just that those lines have to be changed because of this pattern.

services/wallet/collectibles/commands.go
dlipicar: they aren't, removed

services/wallet/collectibles/commands.go
rasom: ok I see there are fields which
```
// Not to be set by the caller
```
still named fields could be used. Anyways though, that's likely just my problem :D

rasom (approved):

services/wallet/collectibles/commands.go
dlipicar: I don't really know what's universally considered good golang... That being said, I kind of like encouraging use of the constructor methods so changes to the struct and its initialization are limited to one spot. Moreover, if any new required field for initialization appears you can know if it's missing in any call to `newX` at compile time, while you might miss it if you instantiate the struct directly somewhere in the code.

services/wallet/collectibles/controller.go
```diff
@@ -215,12 +216,14 @@ func (c *Controller) startPeriodicalOwnershipFetchForAccountAndChainID(address c
 		c.commands[address] = make(commandPerChainID)
 	}
 
+	receivedCollectiblesCb := c.notifyCommunityCollectiblesReceived
```
rasom: considering `receivedCollectiblesCb` is used only once we could probably avoid having it

services/wallet/collectibles/commands.go
rasom: > to the struct and its initialization are limited to one spot

yep I get it and agree, just seeing lots of functions with 3+ arguments make me uncomfortable, but again my problem

stefandunca (approved):
✨

PR #4148, opened 2023-10-12
optimised finding server cert

qfrank:
mainly changes:
- set timeout when requesting server cert 
- trying get server cert with different ip concurrently

this PR should help resolving [mobile issue](https://github.com/status-im/status-mobile/issues/17458)

### DIFF
--- a/server/ips.go
+++ b/server/ips.go
@@ -2,8 +2,6 @@ package server
 
 import (
 	"net"
-	"runtime"
-	"strings"
 
 	"go.uber.org/zap"
 
@@ -99,7 +97,7 @@ func getAndroidLocalIP() ([][]net.IP, error) {
 func getLocalAddresses() ([][]net.IP, error) {
 	// TODO until we can resolve Android errors when calling net.Interfaces() just return the outbound local address.
 	//  Sorry Android
-	if strings.HasPrefix(runtime.GOOS, common.AndroidPlatform) {
+	if common.OperatingSystemIs(common.AndroidPlatform) {
 		return getAndroidLocalIP()
 	}
 
@@ -193,7 +191,7 @@ func getAllAvailableNetworks() ([]net.IPNet, error) {
 // that returns a reachable server's address to be used by local pairing client.
 func FindReachableAddressesForPairingClient(serverIps []net.IP) ([]net.IP, error) {
 	// TODO until we can resolve Android errors when calling net.Interfaces() just noop. Sorry Android
-	if strings.HasPrefix(runtime.GOOS, common.AndroidPlatform) {
+	if common.OperatingSystemIs(common.AndroidPlatform) {
 		return serverIps, nil
 	}
 
--- a/server/ips.go
+++ b/server/ips.go
@@ -3,6 +3,7 @@ package server
 import (
 	"net"
 	"runtime"
+	"strings"
 
 	"go.uber.org/zap"
 
@@ -98,7 +99,7 @@ func getAndroidLocalIP() ([][]net.IP, error) {
 func getLocalAddresses() ([][]net.IP, error) {
 	// TODO until we can resolve Android errors when calling net.Interfaces() just return the outbound local address.
 	//  Sorry Android
-	if runtime.GOOS == common.AndroidPlatform {
+	if strings.HasPrefix(runtime.GOOS, common.AndroidPlatform) {
 		return getAndroidLocalIP()
 	}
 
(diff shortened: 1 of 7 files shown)


server/ips.go
```diff
@@ -192,7 +193,7 @@ func getAllAvailableNetworks() ([]net.IPNet, error) {
 // that returns a reachable server's address to be used by local pairing client.
 func FindReachableAddressesForPairingClient(serverIps []net.IP) ([]net.IP, error) {
 	// TODO until we can resolve Android errors when calling net.Interfaces() just noop. Sorry Android
-	if runtime.GOOS == common.AndroidPlatform {
+	if strings.HasPrefix(runtime.GOOS, common.AndroidPlatform) {
```
igor-sirotin: As discussed in dm, I found a way to suppress a single GoLand warning. 
Though it doesn't look pretty, I think with such solution the intentions are more clear. And we don't add any runtime operations.

```suggestion
	if //goland:noinspection GoBoolEpressions
	runtime.GOOS == common.AndroidPlatform {
```

Also, there one more place with such warning, could you please update it as well?

https://github.com/status-im/status-go/blob/6894295ac3b21a2739e68c7a8bce54f70a52854a/t/utils/utils.go#L225

server/pairing/client.go
```diff
@@ -38,31 +39,53 @@ type BaseClient struct {
...
-			var certErr string
-			if certErrs != nil {
-				certErr = certErrs.Error()
+	var once sync.Once
+	errCh := make(chan struct {
+		ip  net.IP
+		err error
+	}, len(reachableIPs))
+	certCh := make(chan *x509.Certificate, len(reachableIPs))
+	urlCh := make(chan *url.URL, len(reachableIPs))
```
igor-sirotin: Maybe use a `struct`  for success channel too?
```suggestion
	certCh := make(chan struct {
	    cert *x509.Certificate,
	    url *url.URL
	}, len(reachableIPs))
```

And actually it seems that it can also be merged with `done` channel?

server/pairing/client.go
```diff
@@ -38,31 +39,53 @@ type BaseClient struct {
...
+	var combinedErrors string
+	for {
+		select {
+		case <-done:
+			baseAddress = <-urlCh
+			serverCert = <-certCh
+			return baseAddress, serverCert, nil
+		case ipErr := <-errCh:
+			errorCount++
+			combinedErrors += fmt.Sprintf("IP %s: %s; ", ipErr.ip, ipErr.err)
```
igor-sirotin, line 81: `%w` can be used for error directly

igor-sirotin (approved):
Thanks for the improvements, looks great! 🚀

server/pairing/client.go
qfrank: I don't think we should use `%w` here since they're not `wrap` relation, WDYT? 🙂

siddarthkay (approved):

Samyoul (approved):
Thank you very much @qfrank, I like the use of `OperatingSystemIs()` 😉

server/pairing/client.go
Samyoul: Unfortunately wrapping only applies to errors on different call depths. In `go 1.20` you can use `errors.Join` https://pkg.go.dev/errors@master#Join for errors generated on the same call depth.

At the moment `status-go` is using `go 1.19`, so we don't have access to `errors.Join` which would be a nice solution.

saledjenic: The desktop app is using `go1.20.4`.

server/pairing/client.go
```diff
@@ -71,13 +92,19 @@ func NewBaseClient(c *ConnectionParams, logger *zap.Logger) (*BaseClient, error)
...
+	if err != nil {
+		logger.Error("[local pair client] failed to find reachable addresses", zap.Error(err), zap.Any("netIPs", netIPs))
+		signal.SendLocalPairingEvent(Event{Type: EventConnectionError, Error: err.Error(), Action: ActionConnect})
+		return nil, err
+	}
+
 	maxRetries := 3
 	for i := 0; i < maxRetries; i++ {
-		baseAddress, serverCert, certErrs = findServerCert(c)
+		baseAddress, serverCert, certErrs = findServerCert(c, netIPs)
```
saledjenic, line 104: Is it possible for the search for a server certificate to last longer?
Like, if we want to check 5 addresses, 4 of them return err and the fifth needs time. Is that scenario possible at all?

qfrank: each address may last up to 1 second

PR #4153, opened 2023-10-12
chore: Remove include watch only accounts settings as it is out dated and not needed anymore

Khushboo-dev-cpp:

Important changes:

Closes #


saledjenic (approved):

churik:
FYI: introduced critical issue on mobile side https://github.com/status-im/status-mobile/issues/17646

appdatabase/migrations/sql/1697123140_drop_include_watch_only_accounts.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE settings DROP COLUMN include_watch_only_account;
```
saledjenic, line 1: @qfrank could you check if this causes issues on the iOS side?

qfrank: ios support `DROP COLUMN` now, so no more changes needed for this PR

PR #4157, opened 2023-10-15
Fix encryption id

cammellos:
This commit changes the format of the encryption id to be based off 3 things:

1) The group id
2) The timestamp
3) The actual key

Previously this was solely based on the timestamp and the group id, but this might lead to conflicts. Moreover the format of the key was an uint32 and so it would wrap periodically.

The migration is a bit tricky, so first we cleared the cache of keys, that's easier than migrating, and second we set the new field hash_id to the concatenation of group_id / key_id.
This might lead on some duplication in case keys are re-received, but it should not have an impact on the correctness of the code.

I have added 2 tests covering compatibility between old/new clients, as this should not be a breaking change.


protocol/common/message_sender.go
```diff
@@ -1078,11 +1078,11 @@ func (s *MessageSender) StartDatasync() {
 }
 
 // GetCurrentKeyForGroup returns the latest key timestampID belonging to a key group
```
vitvly, line 1080: Probably func comment should be updated.

protocol/common/message_sender.go
```diff
@@ -1078,11 +1078,11 @@ func (s *MessageSender) StartDatasync() {
...
-func (s *MessageSender) GetCurrentKeyForGroup(groupID []byte) (uint32, error) {
+func (s *MessageSender) GetCurrentKeyForGroup(groupID []byte) (*encryption.HashRatchetKeyCompatibility, error) {
 	return s.protocol.GetCurrentKeyForGroup(groupID)
 }
 
 // GetKeyIDsForGroup returns a slice of key IDs belonging to a given group ID
```
vitvly, line 1085: Also here.

protocol/encryption/encryptor.go
```diff
@@ -337,7 +337,18 @@ func (s *encryptor) DecryptPayload(myIdentityKey *ecdsa.PrivateKey, theirIdentit
...
 
-		decryptedPayload, err := s.decryptWithHR(header.GroupId, header.KeyId, header.SeqNo, payload)
+		ratchet := &HashRatchetKeyCompatibility{
+			GroupID: header.GroupId,
+			// NOTE: this would be nil in the old format
+			keyID: header.KeyId,
+		}
+
+		// Old key format
+		if header.DeprecatedKeyId != 0 {
```
vitvly, line 347: `HashRatchetKeyCompatibility` has a `IsOldFormat` method, would it make sense to have a similar method on `HRHeader`?

protocol/encryption/encryptor.go
```diff
@@ -609,51 +620,37 @@ func (s *encryptor) EncryptPayload(theirIdentityKey *ecdsa.PublicKey, myIdentity
...
+		return nil, err
 	}
 
-	err = s.persistence.SaveHashRatchetKey(groupID, keyID, hrKeyBytes)
-
-	return keyID, err
+	return key, s.persistence.SaveHashRatchetKey(key)
 }
 
 // EncryptHashRatchetPayload returns a new EncryptedMessageProtocol with a given payload encrypted, given a group's key
```
vitvly, line 642: Comment update needed.

protocol/encryption/encryptor.go
```diff
@@ -609,51 +620,37 @@ func (s *encryptor) EncryptPayload(theirIdentityKey *ecdsa.PublicKey, myIdentity
...
-func (s *encryptor) EncryptHashRatchetPayload(groupID []byte, keyID uint32, payload []byte) (map[string]*EncryptedMessageProtocol, error) {
+func (s *encryptor) EncryptHashRatchetPayload(ratchet *HashRatchetKeyCompatibility, payload []byte) (map[string]*EncryptedMessageProtocol, error) {
 	logger := s.logger.With(
 		zap.String("site", "EncryptHashRatchetPayload"),
-		zap.Any("group-id", groupID),
-		zap.Any("key-id", keyID))
+		zap.Any("group-id", ratchet.GroupID),
+		zap.Any("key-id", ratchet.keyID))
 
 	s.mutex.Lock()
```
vitvly, line 649: Didn't quite understand what the `mutex` is for. Access to what resources does it guard? Can the lock be narrower? Maybe a comment for mutex declaration would be nice.

protocol/encryption/persistence.go
```diff
@@ -766,63 +767,157 @@ func (s *sqlitePersistence) GetHashRatchetKeyByID(groupID []byte, keyID uint32,
...
+	currentTime := GetCurrentTime()
+	if h.Timestamp < currentTime {
+		ratchet.Timestamp = bumpKeyID(currentTime)
+	} else {
+		ratchet.Timestamp = h.Timestamp + 1
+	}
+
+	ratchet.Key = hrKeyBytes
+
+	_, err = ratchet.GetKeyID()
```
vitvly, line 870: Here we are implicitly checking for `len(h.GroupID) == 0 || h.Timestamp == 0 || len(h.Key) == 0` i think. Probably would be better to move this check into a separate func.

protocol/encryption/persistence.go
```diff
@@ -766,63 +767,157 @@ func (s *sqlitePersistence) GetHashRatchetKeyByID(groupID []byte, keyID uint32,
...
 
 	switch err {
 	case sql.ErrNoRows:
-		return 0, nil
+		return ratchet, nil
 	case nil:
-		return keyID, nil
+		ratchet.Key = key
+		ratchet.Timestamp = timestamp
+		_, err = ratchet.GetKeyID()
```
vitvly, line 904: Also here, an implicit check for `len(h.GroupID) == 0 || h.Timestamp == 0 || len(h.Key) == 0`.

protocol/encryption/protocol.go
```diff
@@ -517,12 +514,27 @@ func (p *Protocol) HandleHashRatchetKeys(groupID, encodedKeys []byte) ([]*HashRa
...
 	}
 	for _, key := range keys.Keys {
+		ratchet := &HashRatchetKeyCompatibility{
+			GroupID:   groupID,
+			Timestamp: key.Timestamp,
+			Key:       key.Key,
+		}
+
+		// If there's no timestamp, is coming from an older client
+		if key.Timestamp == 0 {
```
vitvly, line 524: `IsOldFormat` method here? But then it'd be only a single-usage func, maybe not good.

vitvly (commented):
Minor comments on `comments` :), tiny refactor suggestions, and a question about `encryptor.mutex`.

protocol/encryption/persistence.go
```diff
@@ -766,63 +767,157 @@ func (s *sqlitePersistence) GetHashRatchetKeyByID(groupID []byte, keyID uint32,
...
+	}
+	hrKeyBytes := crypto.FromECDSA(hrKey)
+
+	if err != nil {
+		return nil, err
+	}
+
+	currentTime := GetCurrentTime()
+	if h.Timestamp < currentTime {
+		ratchet.Timestamp = bumpKeyID(currentTime)
```
osmaczko, line 863: I am wondering why do we bump here?

protocol/migrations/sqlite/1697038980_add_timestamp_hash_ratchet.up.sql
```diff
@@ -0,0 +1,28 @@
+
+CREATE TABLE hash_ratchet_encryption_v2 (
+  group_id BLOB NOT NULL,
+  key_id INT NOT NULL,
```
osmaczko, line 4:
```suggestion
  deprecated_key_id INT NOT NULL,
```
To be consistent with code naming.

protocol/migrations/sqlite/1697038980_add_timestamp_hash_ratchet.up.sql
```diff
@@ -0,0 +1,28 @@
...
+  key BLOB NOT NULL,
+  key_timestamp BLOB,
+  hash_id BLOB NOT NULL,
```
osmaczko, line 7:
```suggestion
  key_id BLOB NOT NULL,
```
To be consistent with code naming.

protocol/encryption/persistence.go
```diff
@@ -729,34 +730,34 @@ func (s *sqliteSessionStorage) Load(id []byte) (*dr.State, error) {
...
-       select ? AS group_id, ? AS key_id, ? as seq_no
+func (s *sqlitePersistence) GetHashRatchetKeyByID(ratchet *HashRatchetKeyCompatibility, seqNo uint32) (*HRCache, error) {
+	stmt, err := s.DB.Prepare(`WITH input AS (
+       select ? AS group_id, ? AS key_id, ? as seq_no, ? AS old_key_id
      ),
      cec AS (
        SELECT e.key, c.seq_no, c.hash FROM hash_ratchet_encryption e, input i
-			 LEFT JOIN hash_ratchet_encryption_cache c ON e.group_id=c.group_id AND e.key_id=c.key_id
-       WHERE e.key_id=i.key_id AND e.group_id=i.group_id),
+			 LEFT JOIN hash_ratchet_encryption_cache c ON e.group_id=c.group_id AND (e.hash_id=c.key_id OR e.key_id=c.key_id)
```
osmaczko, line 750:
```suggestion
			 LEFT JOIN hash_ratchet_encryption_cache c ON e.group_id=c.group_id AND (e.hash_id=c.hash OR e.key_id=c.key_id)
```
:thinking:

protocol/encryption/protocol_message.proto
```diff
@@ -56,27 +56,36 @@ message X3DHHeader {
...
   bytes group_id = 3;
+  // group key ID
+  bytes key_id = 4;
 }
 
+// Rekey the whole community
+// hash of pk, and bytes of concatenated strings
+//message RekeyAll {
+//  map<string, bytes>
+//}
```
osmaczko, line 73: Leftover?

protocol/encryption/persistence.go
```diff
@@ -766,63 +767,157 @@ func (s *sqlitePersistence) GetHashRatchetKeyByID(groupID []byte, keyID uint32,
 	var key, hash []byte
 	var seqNoPtr *uint32
 
-	err = stmt.QueryRow(groupID, keyID, seqNo).Scan(&key, &seqNoPtr, &hash)
+	oldFormat := ratchet.IsOldFormat()
+	if oldFormat {
```
osmaczko, line 771: In the flow: `DecryptPayload` -> `decryptWithHR` -> `GetHashRatchetKeyByID` it will be always true, even if the header indicates new format. That's because `ratchet.Key` is nil. Is that expected?

osmaczko (commented):
Nice :+1: have some minor questions

cammellos:
@osmaczko sorry, this PR I will close in favor of https://github.com/status-im/status-go/pull/4182 which includes it

PR #4158, opened 2023-10-16
refactor: simplify community requests logic

osmaczko:
With the recent introduction of pending states, the community requests logic became more complex. This commit simplifies the flow and appropriately delegates logic to its corresponding abstraction levels: messenger, manager and community. Additionally, it eliminates redundancies in notifications and request-saving mechanism.

### DIFF
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -1020,12 +1020,19 @@ func (o *Community) ValidateRequestToJoin(signer *ecdsa.PublicKey, request *prot
 	o.mutex.Lock()
 	defer o.mutex.Unlock()
 
-	// If we are not admin, fuggetaboutit
-	if !o.IsControlNode() && !o.HasPermissionToSendCommunityEvents() {
+	if o.IsControlNode() {
+		// TODO: Enable this once mobile supports revealed addresses.
+		// if len(request.RevealedAccounts) == 0 {
+		// 	return errors.New("no addresses revealed")
+		// }
+	} else if o.HasPermissionToSendCommunityEvents() {
+		if o.AcceptRequestToJoinAutomatically() {
+			return errors.New("auto-accept community requests can only be processed by the control node")
+		}
+	} else {
 		return ErrNotAdmin
 	}
 
-	// If the org is ens name only, then reject if not present
 	if o.config.CommunityDescription.Permissions.EnsOnly && len(request.EnsName) == 0 {
 		return ErrCantRequestAccess
 	}
@@ -1037,6 +1044,19 @@ func (o *Community) ValidateRequestToJoin(signer *ecdsa.PublicKey, request *prot
 	err := o.validateRequestToJoinWithoutChatID(request)
 	if err != nil {
 		return err
+	}
+
+	if o.isBanned(signer) {
+		return ErrCantRequestAccess
+	}
+
+	timeNow := uint64(time.Now().Unix())
+	requestTimeOutClock, err := AddTimeoutToRequestToJoinClock(request.Clock)
+	if err != nil {
+		return err
+	}
+	if timeNow >= requestTimeOutClock {
+		return errors.New("request is expired")
 	}
 
 	return nil
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -2119,25 +2119,17 @@ func (m *Manager) DeclineRequestToJoin(dbRequest *RequestToJoin) (*Community, er
 	return community, nil
 }
 
-func (m *Manager) isUserRejectedFromCommunity(signer *ecdsa.PublicKey, community *Community, requestClock uint64) (bool, error) {
-	declinedRequestsToJoin, err := m.persistence.DeclinedRequestsToJoinForCommunity(community.ID())
+func (m *Manager) shouldUserRetainDeclined(signer *ecdsa.PublicKey, community *Community, requestClock uint64) (bool, error) {
+	requestID := CalculateRequestID(common.PubkeyToHex(signer), types.HexBytes(community.IDString()))
+	request, err := m.persistence.GetRequestToJoin(requestID)
 	if err != nil {
+		if err == sql.ErrNoRows {
+			return false, nil
+		}
 		return false, err
 	}
 
-	for _, req := range declinedRequestsToJoin {
-		if req.PublicKey == common.PubkeyToHex(signer) {
-			dbRequestTimeOutClock, err := AddTimeoutToRequestToJoinClock(req.Clock)
-			if err != nil {
-				return false, err
-			}
-
-			if requestClock < dbRequestTimeOutClock {
-				return true, nil
-			}
-		}
-	}
-	return false, nil
+	return request.ShouldRetainDeclined(requestClock)
 }
 
 func (m *Manager) HandleCommunityCancelRequestToJoin(signer *ecdsa.PublicKey, request *protobuf.CommunityCancelRequestToJoin) (*RequestToJoin, error) {
@@ -2149,11 +2141,11 @@ func (m *Manager) HandleCommunityCancelRequestToJoin(signer *ecdsa.PublicKey, re
 		return nil, ErrOrgNotFound
 	}
 
-	isUserRejected, err := m.isUserRejectedFromCommunity(signer, community, request.Clock)
+	retainDeclined, err := m.shouldUserRetainDeclined(signer, community, request.Clock)
 	if err != nil {
 		return nil, err
 	}
-	if isUserRejected {
+	if retainDeclined {
 		return nil, ErrCommunityRequestAlreadyRejected
 	}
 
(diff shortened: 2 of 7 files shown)


protocol/communities/community.go
```diff
@@ -1020,12 +1020,18 @@ func (o *Community) ValidateRequestToJoin(signer *ecdsa.PublicKey, request *prot
 	o.mutex.Lock()
 	defer o.mutex.Unlock()
 
-	// If we are not admin, fuggetaboutit
-	if !o.IsControlNode() && !o.HasPermissionToSendCommunityEvents() {
+	if o.IsControlNode() {
+		if len(request.RevealedAccounts) == 0 {
+			return errors.New("no addresses revealed")
```
jrainville: I know we were waiting for mobile to also support sharing addresses before setting this as a requirement. I'm not sure if they did yet cc @cammellos

protocol/communities/request_to_join.go
```diff
@@ -74,6 +74,19 @@ func (r *RequestToJoin) Empty() bool {
 	return len(r.ID)+len(r.PublicKey)+int(r.Clock)+len(r.ENSName)+len(r.ChatID)+len(r.CommunityID)+int(r.State) == 0
 }
 
+func (r *RequestToJoin) ShouldRetainDeclined(clock uint64) (bool, error) {
+	if r.State != RequestToJoinStateDeclined {
+		return false, nil
+	}
+
+	declineExpiryClock, err := AddTimeoutToRequestToJoinClock(r.Clock)
```
jrainville, line 82: This timeout is for admins I'm guessing? For how long they should retain a request to join before the control node validates it?
If so, why not for approvals too?

jrainville (commented):
Nice work. It does look cleaner

protocol/communities/manager.go
```diff
@@ -2119,25 +2119,17 @@ func (m *Manager) DeclineRequestToJoin(dbRequest *RequestToJoin) (*Community, er
 	return community, nil
 }
 
-func (m *Manager) isUserRejectedFromCommunity(signer *ecdsa.PublicKey, community *Community, requestClock uint64) (bool, error) {
-	declinedRequestsToJoin, err := m.persistence.DeclinedRequestsToJoinForCommunity(community.ID())
+func (m *Manager) shouldUserRetainDeclined(signer *ecdsa.PublicKey, community *Community, requestClock uint64) (bool, error) {
```
mprakhov, line 2122: IMHO - not the best naming, will appreciate if it can be renamed

protocol/communities/manager.go
```diff
@@ -2149,7 +2141,7 @@ func (m *Manager) HandleCommunityCancelRequestToJoin(signer *ecdsa.PublicKey, re
 		return nil, ErrOrgNotFound
 	}
 
-	isUserRejected, err := m.isUserRejectedFromCommunity(signer, community, request.Clock)
+	isUserRejected, err := m.shouldUserRetainDeclined(signer, community, request.Clock)
```
mprakhov: should we rename the variable as well?

protocol/communities/community.go
osmaczko: I have commented it out for now.

protocol/communities/request_to_join.go
osmaczko: No, this is a cool-down period for declined requests to prevent spam. Once a user is rejected from the community, all their subsequent requests should be automatically declined for a week.

The logic was already there; I just extracted it and, in my opinion, gave it a more precise name: `shouldUserRetainDeclined` over `isUserRejectedFromCommunity`.

protocol/communities/manager.go
osmaczko: I believe the name is more precise than before. However, I'm open to discussion. Could you please suggest any alternatives?

mprakhov: I saw your comments to Jo, now it has become clear what it should do. Maybe a small comment, as it confuses not only me

protocol/communities/manager.go
```diff
@@ -2119,25 +2119,17 @@ func (m *Manager) DeclineRequestToJoin(dbRequest *RequestToJoin) (*Community, er
 	return community, nil
 }
 
-func (m *Manager) isUserRejectedFromCommunity(signer *ecdsa.PublicKey, community *Community, requestClock uint64) (bool, error) {
-	declinedRequestsToJoin, err := m.persistence.DeclinedRequestsToJoinForCommunity(community.ID())
+func (m *Manager) shouldUserRetainDeclined(signer *ecdsa.PublicKey, community *Community, requestClock uint64) (bool, error) {
+	requestID := CalculateRequestID(common.PubkeyToHex(signer), types.HexBytes(community.IDString()))
+	request, err := m.persistence.GetRequestToJoin(requestID)
```
mprakhov, line 2124: What if we request a declined request directly? As I understand, if it is not declined - it will return `false, nil` anyway (the same as empty)

protocol/communities/manager.go
mprakhov: How about:
- IsApprovalDelayedAfterDecline
- IsNotYetApprovableAfterDecline
- IsApprovalBlockedAfterDecline
- IsApprovalNotAllowedAfterDecline

If change returns a result from false to true it can be renamed to
- IsEligibleForApprovalAfterDecline
- CanApproveAfterDecline
- IsApprovalAllowedAfterDeclineTimeout
- HasDeclineCooldownElapsed
- IsApprovalPermittedAfterDecline

protocol/communities/request_to_join.go
jrainville: > Once a user is rejected from the community, all their subsequent requests should be automatically declined for a week.

Do you remember the reason why we only decline for the next 7 days? I thought we always rejected requests after that. 
I'm guessing it's mitigate lost messages or something?

osmaczko: > Do you remember the reason why we only decline for the next 7 days? I thought we always rejected requests after that.

I have no idea, the logic to reject new requests if previous one was declined less than a week ago was introduced with this commit: 0eff61c57ae8fa1e80a3032e1ba8ada583598747

@smohamedjavid could you please shed more light on it?

jrainville: I dug through the PR and issues and found this:
> If the request to join a community is ignored or unattended by the Admin/Community Manager for more than seven (7) days, the request should be marked as expired, and the community "Join request" notification from Admin/Community Manager's Activity Center should be deleted.

This makes sense to me. If a request was not replied to in 7 days, the user can request again.

However, the code as it is right now doesn't reflect that. The code seems about declined requests, but maybe that's just the name? It's worth double checking @osmaczko

osmaczko: I refactored the code by preserving the logic. I believe fixing the requirements is out of scope for this PR.

Take a look at the code before the refactoring:
https://github.com/status-im/status-go/blob/0eff61c57ae8fa1e80a3032e1ba8ada583598747/protocol/communities/manager.go#L1383-L1389

and after the refactoring:
https://github.com/status-im/status-go/blob/4d0d508f206a97e4f61f7ede8a41e1f48532f352/protocol/communities/manager.go#L2200-L2206

It does exactly the same but with different naming.

osmaczko: Take a closer look at the [changes](https://github.com/status-im/status-go/commit/0eff61c57ae8fa1e80a3032e1ba8ada583598747#diff-0c22647108517d38489ae5faa65a914ade3be02ea7a7e37355a82739da9aeb69) to `isUserRejectedFromCommunity` I mentioned - it can be translated to: "should retain declined"

smohamedjavid: Hey @osmaczko 

> @smohamedjavid could you please shed more light on it?

This is a bit complex as the expiry (7 days) helps in a couple of scenarios:

### Scenario 1:

When a user’s request to join was accidentally/intentionally declined by the admin for some reason. The declined response is not sent to the requester (as we don’t want to abruptly say you have been declined for UX reasons). For the requester, it would appear that the request is awaiting a response from the admin. 

Later some time, If the admin finds the requester trustworthy, the admin can’t update the existing request. The user can’t make another request as the request is still pending on his side. Thus, expiry helps the user to request again.

### Scenario 2:
For any user who has requested to join a community and the request is left unattended for 7 days by the admin (as he/she is away/busy). We should automatically clear the request on both the requester’s and admin’s side. So that, the user can request again if they want.

A few more scenarios are described in mobile PR: https://github.com/status-im/status-mobile/pull/15627

<hr />

The cooldown/expiry period is introduced to reduce spamming the admin and eventually, give a good user experience for the requester.

smohamedjavid: > Take a closer look at the [changes](https://github.com/status-im/status-go/commit/0eff61c57ae8fa1e80a3032e1ba8ada583598747#diff-0c22647108517d38489ae5faa65a914ade3be02ea7a7e37355a82739da9aeb69) to isUserRejectedFromCommunity I mentioned - it can be translated to: "should retain declined"

You are right. The PR looks good to me in terms of the request to join handling.

protocol/communities/manager.go
osmaczko: Thank you for the suggestions. I would prefer to stick with `shouldUserRetainDeclined` if you don't mind, as it conveys the same meaning as the proposed names.

protocol/communities/manager.go
osmaczko: Not sure I understand, could you please elaborate?

mprakhov: re-reviewed, it is ok
P.S. The idea was to request declined directly from database

mprakhov (approved):

jrainville (approved):

PR #4160, opened 2023-10-16
feat(discord): Import single channel from discord

borismelnik:
Import single channel from discord functionality.

Important changes:

Closes #

### DIFF
--- a/protocol/messenger.go
+++ b/protocol/messenger.go
@@ -147,6 +147,7 @@ type Messenger struct {
 	cancel context.CancelFunc
 
 	importingCommunities map[string]bool
+	importingChannels    map[string]bool
 	importRateLimiter    *rate.Limiter
 	importDelayer        struct {
 		wait chan struct{}
@@ -540,6 +541,7 @@ func NewMessenger(
 		requestedContactsLock:    sync.RWMutex{},
 		requestedContacts:        make(map[string]*transport.Filter),
 		importingCommunities:     make(map[string]bool),
+		importingChannels:        make(map[string]bool),
 		importRateLimiter:        rate.NewLimiter(rate.Every(importSlowRate), 1),
 		importDelayer: struct {
 			wait chan struct{}
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -3773,6 +3773,740 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
 	}()
 }
 
+func (m *Messenger) saveDiscordAuthorIfNotExists(discordAuthor *protobuf.DiscordMessageAuthor) *discord.ImportError {
+	exists, err := m.persistence.HasDiscordMessageAuthor(discordAuthor.GetId())
+	if err != nil {
+		m.logger.Error("failed to check if message author exists in database", zap.Error(err))
+		return discord.Error(err.Error())
+	}
+
+	if !exists {
+		err := m.persistence.SaveDiscordMessageAuthor(discordAuthor)
+		if err != nil {
+			return discord.Error(err.Error())
+		}
+	}
+
+	return nil
+}
+
+func (m *Messenger) convertDiscordMessageTimeStamp(discordMessage *protobuf.DiscordMessage, timestamp time.Time) *discord.ImportError {
+	discordMessage.Timestamp = fmt.Sprintf("%d", timestamp.Unix())
+
+	if discordMessage.TimestampEdited != "" {
+		timestampEdited, err := time.Parse(discordTimestampLayout, discordMessage.TimestampEdited)
+		if err != nil {
+			m.logger.Error("failed to parse discord message timestamp", zap.Error(err))
+			return discord.Warning(err.Error())
+		}
+		// Convert timestamp to unix timestamp
+		discordMessage.TimestampEdited = fmt.Sprintf("%d", timestampEdited.Unix())
+	}
+
+	return nil
+}
+
+func (m *Messenger) createPinMessageFromDiscordMessage(message *common.Message, pinnedMessageID string, channelID string, community *communities.Community) (*common.PinMessage, *discord.ImportError) {
+	pinMessage := protobuf.PinMessage{
+		Clock:       message.WhisperTimestamp,
+		MessageId:   pinnedMessageID,
+		ChatId:      message.LocalChatID,
+		MessageType: protobuf.MessageType_COMMUNITY_CHAT,
+		Pinned:      true,
+	}
+
+	encodedPayload, err := proto.Marshal(&pinMessage)
+	if err != nil {
+		m.logger.Error("failed to parse marshal pin message", zap.Error(err))
+		return nil, discord.Warning(err.Error())
+	}
+
+	wrappedPayload, err := v1protocol.WrapMessageV1(encodedPayload, protobuf.ApplicationMetadataMessage_PIN_MESSAGE, community.PrivateKey())
+	if err != nil {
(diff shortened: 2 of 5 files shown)


protocol/messenger_communities.go
```diff
@@ -3730,6 +3730,741 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+			cancel <- request.DiscordChannelID
+			return
+		}
+
+		m.config.messengerSignalsHandler.DiscordCommunityImportFinished(request.DiscordChannelID)
+		close(done)
+	}()
+}
+
 func (m *Messenger) RequestImportDiscordCommunity(request *requests.ImportDiscordCommunity) {
```
jrainville, line 4510: I guess most of the code above is the split functions like you talked about. So we can delete the original function, which is `RequestImportDiscordCommunity` I think, when it's ready?

protocol/messenger_communities.go
```diff
@@ -3730,6 +3730,741 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+					importProgress.AddTaskError(discord.ChannelsCreationTask, discord.Error(errmsg))
+					importProgress.StopTask(discord.ChannelsCreationTask)
+					progressUpdates <- importProgress
+					cancel <- request.DiscordChannelID
+					return
+				}
+
+				community = changes.Community
+				for chatID, chat := range changes.ChatsAdded {
+					newChat = CreateCommunityChat(request.CommunityID.String(), chatID, chat, m.getTimesource())
```
jrainville, line 4150: What does this function do? We already do `m.communitiesManager.CreateChat` above.

protocol/messenger_communities.go
```diff
@@ -3730,6 +3730,741 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+				// We're multiplying `chunksCount` by `0.25` so we leave 25% for additional save operations
+				// 0.5 are the previous 50% of progress
+				currentCount := ii + 1
+				progressValue := calculateProgress(i+1, totalImportChunkCount, 0.5+(float32(currentCount)/float32(chunksCount))*0.25)
+				importProgress.UpdateTaskProgress(discord.ImportMessagesTask, progressValue)
+				progressUpdates <- importProgress
+
+				// We slow down the saving of message chunks to keep the database responsive
+				if currentCount < chunksCount {
+					time.Sleep(2 * time.Second)
```
jrainville, line 4210: I think this is no longer needed since the changes to the DB + the manual slow down we already do when the app is in focus cc @osmaczko

protocol/messenger_communities.go
```diff
@@ -3730,6 +3730,741 @@ func (m *Messenger) RequestExtractDiscordChannelsAndCategories(filesToImport []s
...
+				}
+			}
+
+			// Get slice of all values in `messagesToSave` map
+
```
```suggestion
			// Get slice of all values in `messagesToSave` map
```

protocol/messenger_communities.go
```diff
@@ -4581,11 +5316,20 @@ func (m *Messenger) MarkDiscordCommunityImportAsCancelled(communityID string) {
 	m.importingCommunities[communityID] = true
 }
 
+func (m *Messenger) MarkDiscordChannelImportAsCancelled(channelID string) {
+	m.importingChannels[channelID] = true
```
jrainville, line 5362: We should probably rename this and `importingCommunities` to something like `canceledImportChannels`, because the name is confusing.

jrainville (commented):
Looks good. Does it need more work, since it's marked as draft?

protocol/messenger_communities.go
osmaczko: This is different. The speedup/slowdown was implemented for importing archives, not for importing Discord communities/channels.

The same approach could be applied here as well, but it was never a priority because importing a community is a one-time action. I would stick with the naive `time.Sleep` for now, unless we decide to undertake a significant refactor of the complex `RequestImportDiscordCommunity`.

protocol/messenger_communities.go
jrainville: Opened this issue to do it later https://github.com/status-im/status-desktop/issues/12503

protocol/messenger_communities.go
jrainville: Ok, I checked and it's more a factory function to get the right type of object. It doesn't call the DB or anything

protocol/messenger_communities.go
jrainville: I'm guessing this will be fixed here: https://github.com/status-im/status-desktop/issues/12504

protocol/messenger.go
```diff
@@ -146,6 +146,7 @@ type Messenger struct {
 	cancel context.CancelFunc
 
 	importingCommunities map[string]bool
+	importingChannels    map[string]bool
```
0x-r4bbit, line 150: I'm wondering whether this map is necessary...
The equivalent `importingCommunities` was originally introduce to keep track of whether an import of a community was marked as cancelled. And because there could be theoretically multiple imports happen at the same time, we needed a map for that.

The UI however, restricts this to only having one community import at a time (which actually means there's no need for a map in the first place, but it was already there and I think there were discussions around this actually being enabled for multiple imports running in parallel).

The `importingChannels` map hence implies that there can be multiple channels being imported into various communities at the same time.

Do we allow for that in the UI?

Also, I see we use `DiscordImportChannelMarkedAsCancelled` in various places where we used to us `DiscordImportMarkedAsCancelled`.

Notice that we used to initialize a map entry in `importingCommunities` at the beginning of each import.
This was to ensure so we can mark it as cancelled and cleanup work afterwards.

Cleanup particularly happens in https://github.com/status-im/status-go/pull/4160/files#diff-e8ae84ba629f58c08abd2b2b3864896c92dd9b3898cf7b450d378f76004bf312R5371-R5391

I'm seeing most of the `markDiscordCommunityImportAsCancelled()` calls have been removed (and replaced with the channel import equivalent). This is a problem because this potentially prevents the communityID entry in the `importingCommunity` map, which needs to exist to do the `cleanup()`.

Otherwise you keep creating zombie communities.

So I think we need to figure out whether the import has originally started as "community import" and if yes, then we need to mark it as such.

jrainville: > The importingChannels map hence implies that there can be multiple channels being imported into various communities at the same time.
> Do we allow for that in the UI?

No we don't. At one point we will like to support that, but we block it to keep it simple for now.

> Also, I see we use DiscordImportChannelMarkedAsCancelled in various places where we used to us DiscordImportMarkedAsCancelled.

I think `DiscordImportMarkedAsCancelled` is still there. Boris just duplicated a lot of the code to make importing a channel work. I don't see code deletion, so the old code for the community import should still be all there.

I added an issue to clean up this code because of that https://github.com/status-im/status-desktop/issues/12503

I could be wrong though. Boris is off until Thursday, that's why I'm answering, since I want to move this along and merged soon.

0x-r4bbit: > I think DiscordImportMarkedAsCancelled is still there. Boris just duplicated a lot of the code to make importing a channel work. I don't see code deletion, so the old code for the community import should still be all there.

That is a very good point. I was searching through the changes and could only see one usage of `MarkDiscordImportCommunityAsCancelled` (I think in handling authors data code, that's why I thought it was removed everywhere else.

But you're absolutely right, this is only adding stuff.

Ignore my comment then

jrainville:
@osmaczko @MishkaRogachev can you review this please

0x-r4bbit (approved):

jrainville (approved):

PR #4161, opened 2023-10-16
feat: replace DefaultPubsubTopic by Shard 32

richard-ramos:
This PR removes the usage of default pubsub topic. It is not possible to mix named and static sharding, and the default pubsub topic is a named shard. With this PR, I'm using shard rs/16/32 as the default shard for all non-community messages.

This is somewhat problematic because it's a breaking change and compatibility between diff versions would be lost? not sure how to solve this. Perhaps the status.prod fleet should also listen to shard rs/16/32 before this PR is merged?


richard-ramos:
For some reason `make vendor` included A LOT of files

vitvly (approved):
Went through, all good to me!

vitvly:
Question: in this PR i've noticed lines like `community.Shard().TransportShard()`, that do not check for `nil` return value of `Shard()`. Which contrasts with some other places in code, like `SaveCommunity()` or `MarshalPublicAPIJSON()` in `protocol/communities` package, that do this check.

This func will only return `nil` if `Community.config` is `nil` - but when can this happen?

richard-ramos:
> This func will only return nil if Community.config is nil - but when can this happen?

I'm not sure when this specific condition will this happen, tbh, but something that can indeed happen is that `o.config.Shard` is `nil` which happens when no shard has been assigned to a community. Since this is such a common scenario, in `TransportShard` i do check if the shard is `nil`. That way I don't need to do a comparison for nil whenever `.Shard()` is used, since `TransportShard` will automatically know what to do

richard-ramos:
@cammellos, @vitvly 
I did try to get around the backward compatibility issue with having some hardcoded nodes and it is problematic because even if we add some hardcoded peers, the fact that the messages get published on separate pubsub topics means that every message would have to get published on both the default pubsub topic, and in a static shard, meaning that there wouldn't be any bandwidth improvements. If anything, the bandwidth usage would increase even more due to duplicated messages being posted on separate pubsub topics.

Also the bootnodes currently cant subscribe to both named and static shards at same time, meaning that if we want to have harcoded nodes just to support backward compatibility, we can't reuse those from shards.test fleet and would require a separate fleet of nodes with a subscription to the default pubsub topic

Issue #4087, opened 2023-10-03
Optimize new account fetching balance


alaibe:
When it is a new account never used
0 balance
nonce is 0

It means the account doesn't have any history to fetch

PR #4163, opened 2023-10-17
[#4087] Omit history scanning on generated multiacc creation

rasom:
fix (partially) #4087 

When we generate a new account we skip checking history, it saves 5 requests per network at the moment (we have only one request left).
~~Changes on client side are required to make it work, account which are new should be marked as new. Otherwise it gets a bit complicated with all that keycard and sync stuff.~~ Not really, only desktop requites some changes on client side which will be later removed after switching to `CreateAcoountAndLogin`/`RestoreAccountAndLogin`.


IvanBelyakoff (approved):
Great job!

PR #4164, opened 2023-10-17
[WIP] lets not track changes in vendor at all

siddarthkay:
*WIP do not review*

This PR is an attempt to not track vendor folder in git.
The vendor folder will still exist locally and can be generated locally via `make vendor`

status: WIP


jakubgs (commented):
Keep mind that this change will most probably also require changes to how `status-go` is built by Nix in `status-mobile`.

jakubgs:
As far as I can tell the main problem with this is that we'd have to switch the [`status-mobile` derivation for `status-go`](https://github.com/status-im/status-mobile/blob/develop/nix/status-go/mobile/build.nix) from using `buildGoPackage` to use `buildGoModule`, which in turn requires the `vendorSha256` parameter. This is problematic because it would have to be updated every time `status-go` version is updated, as it is in `go-waku` for example:
```nix
          # FIXME: This needs to be manually changed when updating modules.
          vendorSha256 = "sha256-4xChSKAkwwrFp5/ZMnhtvsR4drVfw1cLE3YXwVHeW0A=";
```
https://github.com/status-im/go-waku/blob/bfee9964f65d6169165ac7d6335541c42f134f6b/flake.nix#L31-L32

I even made a thread on Nix Discourse about this:

* https://discourse.nixos.org/t/how-should-i-build-a-go-package-from-local-source/19490

siddarthkay:
latest is https://discourse.nixos.org/t/how-should-i-build-a-go-package-from-local-source/19490/12
converting this PR to draft till we figure out a proper way forward.

PR #4165, opened 2023-10-18
feat: add token list

alaibe:
Add the token list

+ update the uniswap token list with latest


services/wallet/api.go
```diff
@@ -179,6 +179,13 @@ func (api *API) GetBalanceHistoryRange(ctx context.Context, chainIDs []uint64, a
 	return api.s.history.GetBalanceHistory(ctx, chainIDs, address, tokenSymbol, currencySymbol, fromTimestamp)
 }
 
+func (api *API) GetTokenList(ctx context.Context) ([]*token.TokenList, error) {
+	log.Debug("call to get token list")
+	rst := api.s.tokenManager.GetTokenList()
+	log.Debug("result from token list", "len", len(rst))
```
micieslak, line 168: Are those debug logs here intentionally?

alaibe: Yes we wrap all function around with debug log

services/wallet/token/tokenstore.go
```diff
@@ -32,1665 +33,1869 @@ func GetTokenPegSymbol(symbol string) string {
...
-				Symbol:   "USDC",
-				Color:    "#f8f8f8",
-				Decimals: 6,
-				ChainID:  421614,
+		TokenListID: []*Token{
+			&Token{
+				Address:     common.HexToAddress("0x6b175474e89094c44da98b954eedeac495271d0f"),
+				Name:        "Dai Stablecoin",
+				Symbol:      "DAI",
+				Color:       "#f8f8f8",
```
Khushboo-dev-cpp: @alaibe can we remove this color property? it has no significance for tokens right

services/wallet/token/tokenstore.go
```diff
@@ -32,1665 +33,1869 @@ func GetTokenPegSymbol(symbol string) string {
...
+				Decimals:    18,
+				ChainID:     1,
+				TokenListID: "status",
```
Khushboo-dev-cpp, line 65: can we remove this as we are getting rid of the other two apis and only this single one exists and that makes this redundant?

alaibe: Feel free to ignore but that can be useful no if we want to display which token belong to which token list?

services/wallet/token/tokenstore.go
alaibe: yes we can

services/wallet/token/token.go
```diff
@@ -121,60 +93,23 @@ func mergeTokenLists(sliceLists [][]*Token) []*Token {
...
-func (tm *Manager) fetchTokens() {
-	tokenList := make([]*Token, 0)
+func NewTokenManager(
+	db *sql.DB,
+	RPCClient *rpc.Client,
+	networkManager *network.Manager,
+) *Manager {
+	maker, _ := contracts.NewContractMaker(RPCClient)
+	stores := []store{newUniswapStore(), newDefaultStore()}
+	tokens := make([]*Token, 0)
```
IvanBelyakoff, line 107: What confuses me a bit is moving the code from a separate function that fills in a token list from stores to a token manager constructor fucntion, which make it long and prevents writing a test for filling-in the tokens functionality.

services/wallet/token/uniswap_tokenstore.go
```diff
@@ -8,5 +8,25 @@ func newUniswapStore() *uniswapStore {
 }
 
 func (ts *uniswapStore) GetTokens() []*Token {
+	for _, token := range uniswapTokens {
```
IvanBelyakoff, line 11: I think this method can be called many times during the app lifetime. But we need to set it only once, so we could optimize it for reading and set TokenListID only once in `newUniswapStore()`

alaibe: As this is "internal API" i'd rather optimize at a higher level. We don't need to cache those list at all level, just where they are used.
WDYT?

services/wallet/token/token.go
alaibe: Before we had some logic to lazy load token list which leads to multiple call like this:
```
if !notTokenLoaded {
  loadToken()
}
```

I don't think it was ideal especially if only purpose was for testing.
It is better I feel to create a function which return the token but not bind to the manager in that case?

services/wallet/token/uniswap_tokenstore.go
IvanBelyakoff: Frankly, this method is external and can be called from other packages the way they need it, so most likely we will end up calling it many times, because the caller does not assume on the implementation and it will be just slower in the end. While I don't see the downsides of making it in the constructor and we are guaranteed to return the list immediately.
I think getter should be just getter for better cohesion as well, while construction method will do construction

alaibe: The language doesn't allow to make it private due to usage of API but this is a private one by design, the only usage should be the token manager. It is also that way in our wallet codebase.

The type for example is private: uniswapStore

Khushboo-dev-cpp (approved):

PR #4168, opened 2023-10-18
fix: setup owner as CommunityDescription signer after minting community owner token

mprakhov:
When we mint the owner token, clients will start to verify the signature of the CommunityDescription from the contract owner, so CommunityDescription must be signed by the owner privateKey and not by generated community private key

Closes #https://github.com/status-im/status-desktop/issues/12473

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -4701,6 +4701,10 @@ func (m *Manager) AddCommunityToken(token *community_token.CommunityToken) (*Com
 		if err != nil {
 			return nil, err
 		}
+
+		if token.PrivilegesLevel == community_token.OwnerLevel {
+			m.promoteSelfToControlNode(community)
+		}
 	}
 
 	return community, m.saveAndPublish(community)
@@ -5027,21 +5031,21 @@ func (m *Manager) PromoteSelfToControlNode(communityID types.HexBytes) (*Communi
 	if err != nil {
 		return nil, err
 	}
+
 	if community == nil {
 		return nil, ErrOrgNotFound
 	}
 
+	m.promoteSelfToControlNode(community)
+
+	return community, m.saveAndPublish(community)
+}
+
+func (m *Manager) promoteSelfToControlNode(community *Community) {
 	community.setPrivateKey(m.identity)
 	if !community.ControlNode().Equal(&m.identity.PublicKey) {
 		community.setControlNode(&m.identity.PublicKey)
 	}
-
-	err = m.saveAndPublish(community)
-	if err != nil {
-		return nil, err
-	}
-
-	return community, nil
 }
 
 func (m *Manager) shareRequestsToJoinWithNewPrivilegedMembers(community *Community, newPrivilegedMembers map[protobuf.CommunityMember_Roles][]*ecdsa.PublicKey) error {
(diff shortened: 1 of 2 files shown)


protocol/messenger_handler.go
```diff
@@ -2296,6 +2296,11 @@ func (m *Messenger) handleChatMessage(state *ReceivedMessageState, forceSeen boo
 		if err != nil {
 			return err
 		}
+
+		if communityResponse == nil {
+			return nil
+		}
```
mprakhov, line 2302: fixed the crash, `handleWrappedCommunityDescriptionMessage` can return nil, nil

protocol/communities/manager.go
```diff
@@ -4701,6 +4701,13 @@ func (m *Manager) AddCommunityToken(token *community_token.CommunityToken) (*Com
 		if err != nil {
 			return nil, err
 		}
+
+		if token.PrivilegesLevel == community_token.OwnerLevel {
+			community, err = m.updatePrivateKeyAndControlNode(community, m.identity)
```
jrainville: FYI, @osmaczko updated this method, so you'll have a conflict (whoever merges first)

jrainville (approved):
Looks good. Just be careful with the possible conflict with Patryk's PR

endulab (approved):

protocol/communities/manager.go
osmaczko: The method is named `PromoteSelfToControlNode` now, there is no need to pass `m.identity`.

osmaczko (approved):

mprakhov:
https://github.com/status-im/status-go/pull/4168#discussion_r1364517504
Is this done in some other PR? I do not observe such a fn name

P.S. Found, rebased, adapted to my needs

protocol/communities/manager.go
```diff
@@ -4701,6 +4701,10 @@ func (m *Manager) AddCommunityToken(token *community_token.CommunityToken) (*Com
 		if err != nil {
 			return nil, err
 		}
+
+		if token.PrivilegesLevel == community_token.OwnerLevel {
+			m.promoteSelfToControlNode(community)
+		}
```
osmaczko, line 4707: @mprakhov I am analyzing the code in that area now, and after deeper thought, I'm struggling to understand why this change was necessary. Shouldn't only the control node be able to mint the owner token? This means the node has to already be a control node at that stage and `m.promoteSelfToControlNode(community)` is redundant :thinking:

mprakhov: Due to the signature of the CommunityDescription. When we create the community, the community private key is different from the control node.
So when we deploy the contract owner (receiving owner token) - we change the community private key to be a private key of the contract owner. Otherwise, when you extract the signer through `unwrapCommunityDescriptionMessage` it will be signed by the community private key and not the owner private key and you'll not pass `HandleCommunityDescriptionMessage` due to `ErrNotAuthorized` error

mprakhov: So yes, only the control node will mint the owner token, but as soon as we mint the owner token, all CommunityDescriptions verifications on the client side will be done through the loop, and loop check chat contact owner and signer of the CommunityDescription is always the same

osmaczko: Right, you're correct. Sorry, I had a brain fart :sweat_smile:

PR #4169, opened 2023-10-18
fix: shared links and link previews contain full self information

### LINKED ISSUE #4154: Link previews and shared URLs doesn't contain data for self contact link
# Problem

When calling a `ShareUserURLWithData` or `StatusUnfurler.buildContactData` with self public key, the result won't contain any data like `DisplayName`, `Description`,  `thumbnail`. 

# Implementation

In both cases:
- https://github.com/status-im/status-go/blob/157e2de3ab8813765399ca6413fd911ac0857649/protocol/messenger_share_urls.go#L500
- https://github.com/status-im/status-go/blob/157e2de3ab8813765399ca6413fd911ac0857649/protocol/linkpreview_unfurler_status.go#L54

... the contact is requested with `messenger.allContacts.Load`, which has custom handling for self-requests:
https://github.com/status-im/status-go/blob/244b4273de49a28a9bc54da51c849d13fe452486/protocol/messenger_maps.go#L63-L73

The problem is that `contactMap.me` is created with no properties and is never updated.

I think there might be more affected places.

# Notes

The thing is that _yourself is not a contact_. The original change (https://github.com/status-im/status-go/pull/3627) for `ContactsMap` still makes sense. We should avoid storing ourselves as a contact. 

There're 2 ways out of this:
1. Keep `ContactsMap` as is and update `messenger.GetContactByID` so that it checks if self-request if made, build and return a valid contact from there.
2. Keep `ContactsMap` up-to-date. This will require updating it on each change like `SetDisplayName`, `SetBio`, etc.

I tend to go with option `1`.

igor-sirotin:
Fixes #4154 

# Description

1. Added a new `messenger.selfContact` property.
2. `selfContact` is automatically updated with any of these fields change (these are the only ones mutable ones):
    - `displayName`
    - `bio`
    - `ensName` (preferredName)
    - `socialLinks`
    - `identityImages`
3. `GetContactByID` now automatically returns `selfContact` when needed.
This way I kept `ContactsMap` to be contacts-only and not handling self-contact. 
Later we can:
    - track all places using `allContacts.Load`
    - replace the ones that need to handle self-contact with `GetContactByID`
    - remove the temporary check for self-contact in `ContactMap` and return `nil` for such cases.
4. Replaced `allContacts.Load` with `GetContactByID` in generating Shared URLs

# Accompanying changes

1. `multiaccounts.Database.identityImageSubscriptions` now contains a `publishExpected` flag
Previously `publish` property was handled inside, but I needed to notify Messenger of image changes. And wanted to do it without another channel.
3. Added `multiaccounts.settings.Database.changesSubscriptions`
This is needed to subscribe to settings changes and update `messenger.selfContact`. Though it was possible to handle some properties in `messenger.SetDisplayName` and `messenger.SetBio`, it wasn't possible to track `PreferredName`.
4. Fixed a bug that a contact without an icon (or an invalid one) would result in an empty `LinkPreview`
5. Added `identityImages` argument to `syncProfilePictures`.
Just not to request the database twice, the data is already available in most cases.

# Tests

1. Added `Test_SelfContact` for the introduced functionality
2. Added `Test_UnfurlURLs_SelfLink` for unfurling "self-link"

Also, I tested desktop app:
(I can see there's a bug that chat input link preview card doesn't show contact icon, but that's a different issue)

https://github.com/status-im/status-go/assets/25482501/532c29ba-5f55-43ff-a0a2-9d44f3d71cd1


protocol/messenger_sync_settings.go
```diff
@@ -159,3 +159,24 @@ func (m *Messenger) startSyncSettingsLoop() {
...
+		for {
+			select {
+			case s := <-channel:
+				switch s.GetReactName() {
+				case settings.DisplayName.GetReactName():
+					m.selfContact.DisplayName = s.Value.(string)
+				case settings.PreferredName.GetReactName():
+					m.selfContact.EnsName = s.Value.(string)
+				case settings.Bio.GetReactName():
+					m.selfContact.Bio = s.Value.(string)
```
jrainville, line 175: Is there a chance that somewhere else we might modify those values? I'm asking in case we need a lock

jrainville (commented):
Looks good, just one question

protocol/messenger_sync_settings.go
igor-sirotin: You mean `m.selfCotnact.EnsName = ...`? Then no, we only set those values here.
We also set it during initialisation, but that should be fine.

jrainville: Yeah that's what I meant. Sounds good then.

jrainville (approved):

jrainville:
It would be good to have someone more knowledgeable of contacts to review as well, as I'm not that versed in this side

igor-sirotin:
@cammellos @Samyoul would be nice to have your review here.
It's kind of changing the logic here towards "not storing ourselves in `allContacts`", I want to make sure this is the right thing to do.

protocol/contact.go
```diff
@@ -386,6 +388,41 @@ func buildContact(publicKeyString string, publicKey *ecdsa.PublicKey) (*Contact,
...
+			}
+		}
+		if socialLinks, err := settings.GetSocialLinks(); err != nil {
+			c.SocialLinks = socialLinks
+		}
+	}
+
+	if multiAccounts != nil && account != nil {
+		if identityImages, err := multiAccounts.GetIdentityImages(account.KeyUID); err != nil {
+			imagesMap := make(map[string]images.IdentityImage)
```
MishkaRogachev, line 414: Suggestion: `make(map[string]images.IdentityImage, len(identityImages))`

MishkaRogachev (approved):
Great work! Unfortunately i don't have much awareness here for review either :)

cammellos (approved):
Looks ok, myself I would invert the dependency between settings / messenger, so we don't have to publish / etc, but that's a larger change.
One thing I didn't see, I assume we receive the data from sync as well, do we update it in that case?

igor-sirotin:
> Looks ok, myself I would invert the dependency between settings / messenger, so we don't have to publish / etc, but that's a larger change.

yeah, I thought the same. Currently both desktop and mobile apps directly call `SaveSetting` method, so settings are the source of truth here for now.

> One thing I didn't see, I assume we receive the data from sync as well, do we update it in that case?

Yeah, that should work fine 👍  But I'll double check

igor-sirotin:
Tested with syncing, works fine 👌

PR #4173, opened 2023-10-19
Chore/use cached control node

osmaczko:
- chore: omit `CommunityDescription` queuing if owner is already verified
- fix: prevent nil dereference in `IsControlNode` for existing communities

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -1545,36 +1545,36 @@ func (m *Manager) HandleCommunityDescriptionMessage(signer *ecdsa.PublicKey, des
 		}
 
 		// A new community, we need to check if we need to validate async.
-		// That would be the case if it has a contract. We queue everything and process separately
-
+		// That would be the case if it has a contract. We queue everything and process separately.
 		if shouldQueue {
 			return nil, m.Queue(signer, community, description.Clock, payload)
 		}
-
 	} else {
-		// we only queue if the clock is greater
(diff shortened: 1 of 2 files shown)


protocol/communities/manager.go
```diff
@@ -1545,36 +1545,35 @@ func (m *Manager) HandleCommunityDescriptionMessage(signer *ecdsa.PublicKey, des
...
 		if shouldQueue {
 			return nil, m.Queue(signer, community, description.Clock, payload)
 		}
-
 	} else {
-		// we only queue if the clock is greater
-		if shouldQueue && community.config.CommunityDescription.Clock < description.Clock {
+		// only queue if already known control node is different than the signer
+		// and if the clock is greater
+		shouldQueue = shouldQueue && !common.IsPubKeyEqual(community.ControlNode(), signer) && community.config.CommunityDescription.Clock < description.Clock
```
jrainville: Shouldn't this condition be part of `shouldQueue`'s original assignment?

jrainville (approved):
Looks good, just a small possible improvement?

cammellos (approved):

protocol/communities/manager.go
osmaczko: Not really, the `community` may be nil there and I din't want to repeat `community != nil` check there.

mprakhov (approved):

PR #4174, opened 2023-10-19
fix(chat/api): make community chat creation more forgiving and accurate

jrainville:
Fixes https://github.com/status-im/status-desktop/issues/12418

That code had two issues, that are symptoms of the same root issue.

The root issue is unknown to me. Basically, the community had a channel deleted, so the description was updated. However, Stefan's chat's DB wasn't updated. The community chat called `Test` was still there and set as active.

So the two fixes I applied to make the app still work are:
- Don't return an error when the chat is not found in the community when populating the members
  - This would leave the member list empty, but it's better than having no messenger sections
- Loop the community chats directly instead of looping all channels and find the ones that have the community ID
  - This one makes sure that we only show the channels that are still in the community. So we shouldn't have a ghost channel with no members.

Obviously, fixing the root cause would be even better (no discrepancy between chat and community tables), however, I have no idea how it happened. Stefan had a lot of issues with that community because of the description size issues + the missing encryption keys, so it might be a previous bug that messed up the DB a bit.
Also, I don't think anyone else had this issue, so I'm hoping it was just a one off issue.

The changes in the PR at least make sure that if it happens again, we will be resilient against it.

### DIFF
--- a/services/chat/api.go
+++ b/services/chat/api.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"strings"
 
+	"github.com/ethereum/go-ethereum/log"
+
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/status-im/status-go/eth-node/crypto"
 	"github.com/status-im/status-go/eth-node/types"
@@ -106,12 +108,14 @@ type ChannelGroup struct {
 
 func NewAPI(service *Service) *API {
 	return &API{
-		s: service,
+		s:   service,
+		log: log.New("package", "status-go/services/chat.API"),
 	}
 }
 
 type API struct {
-	s *Service
+	s   *Service
+	log log.Logger
 }
 
 func unique(communities []*communities.Community) (result []*communities.Community) {
@@ -246,6 +250,12 @@ func (api *API) getChannelGroups(ctx context.Context, channelGroupID string) (ma
 
 		for _, chat := range channels {
 			if chat.CommunityID == community.IDString() && chat.Active {
+				_, exists := community.Chats()[chat.CommunityChatID()]
+				if !exists {
+					api.log.Warn("Chat not found in the community", "chat.ID", chat.ID)
+					continue
+				}
+
 				c, err := api.toAPIChat(chat, community, pubKey, true)
 				if err != nil {
 					return nil, err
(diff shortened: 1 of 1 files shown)


stefandunca (approved):
Thanks, @jrainville, for the great fix. I tested it, and it worked wonders on my dogfooding user folder that I backed up before starting from scratch.

services/chat/api.go
```diff
@@ -244,14 +244,19 @@ func (api *API) getChannelGroups(ctx context.Context, channelGroupID string) (ma
 			}
 		}
 
-		for _, chat := range channels {
-			if chat.CommunityID == community.IDString() && chat.Active {
+		// Looping The community's chats first makes sure we only get the chats that are reallay part of the community
+		for communityChatID := range community.Chats() {
+			for _, chat := range channels {
+				if communityChatID != chat.CommunityChatID() || !chat.Active {
+					continue
```
osmaczko, line 256: Could you please add a warning here to determine if that was indeed the issue? This would also let us understand better what happened. Once we have the chatID that is the culprit we could investigate the database.

osmaczko (approved):

services/chat/api.go
jrainville: Added the log. Please check the code again since I switched back to the old loop

services/chat/api.go
```diff
@@ -421,7 +431,8 @@ func getChatMembers(sourceChat *protocol.Chat, community *communities.Community,
 	if community != nil {
 		channel, exists := community.Chats()[sourceChat.CommunityChatID()]
 		if !exists {
-			return result, communities.ErrChatNotFound
+			// Skip unknown community chats. They might be channels that were deleted. We shouldn't get here
+			return result, nil
```
osmaczko, line 435: Does it still make sense to keep this change if we already have `!exists` check in `getChannelGroups`?

services/chat/api.go
osmaczko: Looks good :+1:

osmaczko (approved):

MishkaRogachev (approved):

services/chat/api.go
jrainville: It's a double security. In case someone calls `getChatMembers` from somewhere else. But yeah, you're right that in the current context, this should never be called, hence why I added `We shouldn't get here` in the comment.

PR #4176, opened 2023-10-20
feat: add check that member with the role Owner is a contract in the community with minted owner token

mprakhov:
During ReevaluateMember for the community with minted owner token check that we do not have members with incorrect Owner role and current contract owner has the correct Owner role

Closes #https://github.com/status-im/status-desktop/issues/11945

### DIFF
--- a/protocol/communities/errors.go
+++ b/protocol/communities/errors.go
@@ -41,3 +41,4 @@ var ErrNotEnoughPermissions = errors.New("not enough permissions for this commun
 var ErrCannotRemoveOwnerOrAdmin = errors.New("not allowed to remove admin or owner")
 var ErrCannotBanOwnerOrAdmin = errors.New("not allowed to ban admin or owner")
 var ErrInvalidManageTokensPermission = errors.New("no privileges to manage tokens")
+var ErrNoPermissionsForReevaluate = errors.New("no permissions to reevaluate members")
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -449,13 +449,14 @@ func (m *Manager) runOwnerVerificationLoop() {
 							m.logger.Error("failed to handle community", zap.Error(err))
 							continue
 						}
+
 						if response != nil {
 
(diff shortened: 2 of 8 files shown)


osmaczko (changes requested):
I believe there's a more optimal approach to this problem.

Currently, we have a loop that checks for verified signers. When the control node changes, we receive a `CommunityDescription` from the new control node. As a consequence, we unmark ourselves as the control node. See: https://github.com/status-im/status-go/blob/fc8ce915afc8518584e87d0501c49e5008fb2a18/protocol/communities/manager.go#L1567-L1570

When a new control node takes ownership, it's crucial that the owner role is updated appropriately. I suggest embedding this logic here: https://github.com/status-im/status-go/blob/fc8ce915afc8518584e87d0501c49e5008fb2a18/protocol/communities/manager.go#L5044

mprakhov:
> I believe there's a more optimal approach to this problem.
> 
> Currently, we have a loop that checks for verified signers. When the control node changes, we receive a `CommunityDescription` from the new control node. As a consequence, we unmark ourselves as the control node. See:
> 
> https://github.com/status-im/status-go/blob/fc8ce915afc8518584e87d0501c49e5008fb2a18/protocol/communities/manager.go#L1567-L1570
> 
> When a new control node takes ownership, it's crucial that the owner role is updated appropriately. I suggest embedding this logic here:
> 
> https://github.com/status-im/status-go/blob/fc8ce915afc8518584e87d0501c49e5008fb2a18/protocol/communities/manager.go#L5044

1. If you are not an owner - that means that you are a simple member. So you must not change CommunityDescription
2. Putting to `promoteSelfToControlNode` - you can't mint an owner token if you are not a control node, and the role Owner is assigned to the control node on the community creation. So during the call promoteSelfToControlNode, you'll already be a control node. But, I have this fix in my ticket, where I drop all members and update the Community :) 

I added the code to reevaluate in order we do not miss configuring us as an owner (everything can happen) + in order if someone somehow cheats (for example through the events) and sets up himself as an Owner

mprakhov:
Discussed, closing as not needed and outdated

PR #4179, opened 2023-10-20
Feat/wait on signaled messenger response

osmaczko:
- chore: add `WaitOnSignaledMessengerResponse` test utility
- fix: use `messengerSignalsHandler` instead of global object

### DIFF
--- a/protocol/communities_messenger_signers_test.go
+++ b/protocol/communities_messenger_signers_test.go
@@ -3,8 +3,6 @@ package protocol
 import (
 	"context"
 	"crypto/ecdsa"
-	"encoding/json"
-	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


mprakhov (approved):
👍

protocol/communities_messenger_signers_test.go
```diff
@@ -193,25 +190,11 @@ func (s *MessengerCommunitiesSignersSuite) TestControlNodeUpdateSigner() {
...
-		published = strings.Contains(string(data), community.IDString())
-	})
-
 	// john accepts community update
-	_, err = WaitOnMessengerResponse(
+	_, err = WaitOnSignaledMessengerResponse(
 		s.john,
 		func(r *MessengerResponse) bool {
-			return published
+			return len(r.Communities()) > 0 && r.Communities()[0].IDString() == community.IDString()
```
mprakhov17, line 197: do we need check on `r != nil` ?

osmaczko: We could certainly do it. However, theoretically, we should never signal a 'nil' response. I think it's safe to proceed without it. If a nil response is ever emitted, then the test will crash, which is also acceptable.

PR #4182, opened 2023-10-23
Rekey with a single message

### LINKED ISSUE #11211: Messages are not received in the encrypted community
# Bug Report

edit: it turns out the issue is about not receiving messages. See comments below

Messages sent by me are not getting delivered in status CC community channels. Direct messages sent to contacts seem to be working fine.

## Description
I have recently installed Status Desktop on my macbook and joined the Status CC Community as i am a new joinee. 
I am seeing messages sent by others in the same community channel. 

Attaching screenshots of a channel seen by me and my colleague.
Screenshot as seen in my app: Screenshot 2023-06-21 at 7 04 40 PM
Secreenshot as seen in my colleague's app: image

As you can see, my colleague doesn't see the message sent by me, whereas it is shown in my view. Similarly he is not able to see the message sent by another person SidActual whereas i am able to see it. 

Similarly the intro message i have posted in intros channel also is not delivered, as my colleague is not able to see it.

Looks like something has gone wrong wrt communities.

I have even restarted the status app, but still noticing the same behaviour.

## Steps to reproduce

Nothing specific, already highlighted in description.

#### Expected behavior
My messages should be sent to community channels once my request to join has been accepted.

#### Actual behavior
None of my messages are being sent out/delivered.

### Additional Information

- Status desktop version: 0.12.0
- Operating System: MacOS 13.4 (22F66) on M2 MBP

cammellos:
This commit builds on top of https://github.com/status-im/status-go/pull/4157

it adds a protocol message that rekeys multiple  users with a single message. it maintains for now compatibility.
it also enables the rekey loop, and tests encrypted communities, both with the old format and the new rekey format.


protocol/migrations/sqlite/1697738980_add_timestamp_hash_ratchet.up.sql
```diff
@@ -0,0 +1,28 @@
+
+CREATE TABLE hash_ratchet_encryption_v2 (
+  group_id BLOB NOT NULL,
+  key_id INT NOT NULL,
```
osmaczko:
```suggestion
  deprecated_key_id INT NOT NULL,
```

To be consistent with code naming.

protocol/migrations/sqlite/1697738980_add_timestamp_hash_ratchet.up.sql
```diff
@@ -0,0 +1,28 @@
...
+  key BLOB NOT NULL,
+  key_timestamp BLOB,
+  hash_id BLOB NOT NULL,
```
osmaczko:
```suggestion
  key_id BLOB NOT NULL,
```
To be consistent with code naming.

protocol/encryption/persistence.go
```diff
@@ -729,34 +730,34 @@ func (s *sqliteSessionStorage) Load(id []byte) (*dr.State, error) {
...
-       select ? AS group_id, ? AS key_id, ? as seq_no
+func (s *sqlitePersistence) GetHashRatchetKeyByID(ratchet *HashRatchetKeyCompatibility, seqNo uint32) (*HRCache, error) {
+	stmt, err := s.DB.Prepare(`WITH input AS (
+       select ? AS group_id, ? AS key_id, ? as seq_no, ? AS old_key_id
      ),
      cec AS (
        SELECT e.key, c.seq_no, c.hash FROM hash_ratchet_encryption e, input i
-			 LEFT JOIN hash_ratchet_encryption_cache c ON e.group_id=c.group_id AND e.key_id=c.key_id
-       WHERE e.key_id=i.key_id AND e.group_id=i.group_id),
+			 LEFT JOIN hash_ratchet_encryption_cache c ON e.group_id=c.group_id AND (e.hash_id=c.key_id OR e.key_id=c.key_id)
```
osmaczko:
```suggestion
			 LEFT JOIN hash_ratchet_encryption_cache c ON e.group_id=c.group_id AND (e.hash_id=c.hash OR e.key_id=c.key_id)
```
:thinking:

protocol/communities_key_distributor.go
```diff
@@ -104,6 +104,7 @@ func (ckd *CommunitiesKeyDistributorImpl) distributeKey(community *communities.C
 }
 
 func (ckd *CommunitiesKeyDistributorImpl) sendKeyExchangeMessage(community *communities.Community, hashRatchetGroupID []byte, pubkeys []*ecdsa.PublicKey, msgType common.CommKeyExMsgType) error {
+	// NOTE: should the sender by set to the community control node?
```
jrainville: What do you mean here? In theory, only the Owner should be able to send encryption keys. As for Owner vs  control node, unless we send the installation ID, it's both the same pubkey.

protocol/encryption/persistence.go
```diff
@@ -766,63 +767,157 @@ func (s *sqlitePersistence) GetHashRatchetKeyByID(groupID []byte, keyID uint32,
...
+	}
+	hrKeyBytes := crypto.FromECDSA(hrKey)
+
+	if err != nil {
+		return nil, err
+	}
+
+	currentTime := GetCurrentTime()
+	if h.Timestamp < currentTime {
+		ratchet.Timestamp = bumpKeyID(currentTime)
```
osmaczko, line 863: I am wondering why do we bump here?

protocol/communities_messenger_test.go
```diff
@@ -384,13 +386,13 @@ func (s *MessengerCommunitiesSuite) createCommunity() (*communities.Community, *
 	return createCommunity(&s.Suite, s.owner)
 }
 
-func (s *MessengerCommunitiesSuite) advertiseCommunityTo(community *communities.Community, user *Messenger) {
-	advertiseCommunityTo(&s.Suite, community, s.owner, user)
+func (s *MessengerCommunitiesSuite) advertiseCommunityTo(community *communities.Community, user *Messenger, admin *Messenger) {
```
jrainville: Can we keep using the word `owner` here? Since `admin` can mean the permission, which is under owner level.

protocol/encryption/persistence.go
```diff
@@ -766,63 +767,157 @@ func (s *sqlitePersistence) GetHashRatchetKeyByID(groupID []byte, keyID uint32,
 	var key, hash []byte
 	var seqNoPtr *uint32
 
-	err = stmt.QueryRow(groupID, keyID, seqNo).Scan(&key, &seqNoPtr, &hash)
+	oldFormat := ratchet.IsOldFormat()
+	if oldFormat {
```
osmaczko, line 771: In the flow: `DecryptPayload` -> `decryptWithHR` -> `GetHashRatchetKeyByID` it will be always true, even if the header indicates new format. That's because `ratchet.Key` is nil. Is that expected?

protocol/communities_messenger_test.go
```diff
@@ -3272,49 +3285,269 @@ func (s *MessengerCommunitiesSuite) TestStartCommunityRekeyLoop() {
...
-		for i := 0; i < 5; i++ {
-			time.Sleep(s.admin.communitiesManager.RekeyInterval * 2)
-			keys, err = s.admin.sender.GetKeyIDsForGroup(c.ID())
-			s.Require().NoError(err)
-			s.Require().Greater(len(keys), keyCount)
-			keyCount = len(keys)
-		}
-	*/
+func (s *MessengerCommunitiesSuite) TestCommunityRekeyAfterBan() {
+	admin := s.newMessenger()
```
jrainville: I'm curious, why can't you use the basic `s.owner`?

protocol/common/message_sender.go
```diff
@@ -320,17 +324,71 @@ func (s *MessageSender) sendCommunity(
...
 		}
 
-		for i, spec := range keyExMessageSpecs {
-			recipient := rawMessage.Recipients[i]
-			_, _, err = s.sendMessageSpec(ctx, recipient, spec, messageIDs)
+		if forceRekey {
+
+			var ratchet *encryption.HashRatchetKeyCompatibility
+			// We have just rekeyed, pull the latest
+			if forceRekey && RekeyCompatibility {
```
osmaczko:
```suggestion
			if RekeyCompatibility {
```
`forceRekey` is redundant

protocol/messenger_communities.go
```diff
@@ -4997,12 +4990,13 @@ func (m *Messenger) rekeyAllCommunities(logger *zap.Logger) {
...
-	} else {
-		rki = m.communitiesManager.RekeyInterval
-	}
+	/*
+		var rki time.Duration
+		if m.communitiesManager.RekeyInterval == 0 {
+			rki = time.Hour
+		} else {
+			rki = m.communitiesManager.RekeyInterval
+		}*/
```
jrainville, line 5007: Still needed?

protocol/messenger_communities.go
```diff
@@ -5011,7 +5005,7 @@ func (m *Messenger) rekeyAllCommunities(logger *zap.Logger) {
 		return
 	}
 	for _, c := range cs {
-		keyTimestampID, err := m.GetCurrentKeyForGroup(c.ID())
+		//keyTimestampID, err := m.GetCurrentKeyForGroup(c.ID())
```
jrainville: leftover

protocol/communities/manager.go
```diff
@@ -99,6 +96,7 @@ type Manager struct {
 	historyArchiveDownloadTasks      map[string]*HistoryArchiveDownloadTask
 	stopped                          bool
 	RekeyInterval                    time.Duration
+	PermissionChecker                PermissionChecker
```
osmaczko, line 99:
```suggestion
	permissionChecker                PermissionChecker
```

jrainville (commented):
Nice job! I'm far from an expert on the encryption level, so I just asked a few questions.

protocol/communities/permission_checker.go
osmaczko, line 1: Nice extraction :+1:

protocol/communities_messenger_test.go
cammellos: some tests wants an aggressive rekey interval in the loop, while this one wants it disabled, and it's done on creation (before start), so I though the best would be to create a new one

protocol/messenger_communities.go
cammellos: not sure :) it wasn't used at all, so I left it as it was (the only usage was a printf), I would leave it for now just because I didn't quite look at it

protocol/communities_key_distributor.go
cammellos: it looks like we are always using m.identity to send the encryption keys, so not the pk of the community, in case is not tokenized (if it' stokenized, we use the right key as the owner is set to the chat key)

jrainville: Aaah, I get it. I guess it would be more rigorous to set it to the community key while the Owner token is not minted, but maybe you can keep the code simpler by using the Owner's pubkey at all times?

osmaczko: Received encryption keys aren't validated, so anyone could send them. Once validation is implemented, the sender must indeed be the community itself for non-tokenized communities.

protocol/communities_messenger_test.go
osmaczko: Also could you please invert the order? It is easier to reason about if `owner` is first (owner advertises community to user).

protocol/messenger_handler.go
```diff
@@ -3683,3 +3682,7 @@ func (m *Messenger) addNewKeypairAddedOnPairedDeviceACNotification(keyUID string
 	}
 	return nil
 }
+
+func (m *Messenger) HandleRekeyGroup(state *ReceivedMessageState, message *protobuf.CommunityMessageArchiveMagnetlink, statusMessage *v1protocol.StatusMessage) error {
+	return nil
```
osmaczko: Missing implementation?

protocol/messenger_handlers.go
osmaczko, line 1: Generated file unnecessarily formatted.

protocol/encryption/persistence.go
cammellos: hash is the current hash of original key + seqNo (i.e the original key hashed `sedNo` key), here we want to match with the old key_id format (deprecated_key_id as per your suggestion) and with the new format (`hash_id`), as we want to get the original key I believe

protocol/encryption/persistence.go
cammellos: I can't remember,  I guess we want it to bump of at least one, but not sure, it would make sense to bump of a random interval to avoid conflicts, but after this change conflicts are not possible/not damaging

protocol/communities_key_distributor.go
cammellos: probably I should make that change now (not the validation), but at least send from control node, otherwise once we add it, it will not validate anymore

protocol/communities/manager.go
cammellos: I think I am using it outside the namespace, but I will check

protocol/messenger_handler.go
cammellos: it's a left over, I didn't create a new application method after all, but did at some point, I will remove it

protocol/communities/manager.go
cammellos: Yes, I use it in tests to replace it at runtime for mocking

protocol/encryption/persistence.go
cammellos: mmh, not good :), it works fine since we maintain compatibility, but we should prefer the new key format, i'll fix it

cammellos: wait, is it, isOldFormat 
```
func (h *HashRatchetKeyCompatibility) IsOldFormat() bool {
	return len(h.keyID) == 0 && len(h.Key) == 0
}
```
so keyid would have to be empty for it to be in the old format, but we do set it

```

		ratchet := &HashRatchetKeyCompatibility{
			GroupID: header.GroupId,
			// NOTE: this would be nil in the old format
			keyID: header.KeyId,
		}

```
so should work, or am I missing something?

cammellos:
@jrainville @osmaczko I have updated the pr and addressed the feedback, it should be ready to be reviewed again, thank you!

protocol/communities_key_distributor.go
osmaczko: +1

protocol/encryption/persistence.go
osmaczko: I got confused with hash_id

protocol/messenger_communities.go
```diff
@@ -5021,7 +5014,6 @@ func (m *Messenger) rekeyAllCommunities(logger *zap.Logger) {
 		//  c.RekeyedAt().Add(rki).Before(time.Now())
 		//  keyTimestampID + rki < time.Now()
 		//  Just using the vars that will be used later
```
osmaczko, line 5024: With this PR the rekey loop has been enabled with a default interval of 5 minutes. Without this segment of code, the community would rekey every 5 minutes, which I believe is excessively frequent. The intention was to have a rekeying process every hour (refer to line 4996), but to initiate a check on what to rekey every 5 minutes (see line 4966).

osmaczko: We could keep the loop disabled and address it as followup to not block this PR anymore.

osmaczko (approved):
:+1: 

One important concern about rekey loop.

protocol/messenger_communities.go
cammellos: oh I see, I will add a flag to disable and have it disabled for now, yes, good catch

jrainville (approved):
Great job

PR #4184, opened 2023-10-23
fix: publish group member info

mohsen-ghafouri:
Publish group member info just after group creation

related to status-mobile issue https://github.com/status-im/status-mobile/issues/17230


protocol/messenger_group_chat_test.go
```diff
@@ -454,3 +455,48 @@ func (s *MessengerGroupChatSuite) TestGroupChatMembersRemovalOutOfOrder() {
...
+				return false
+			}
+			contact, ok := memberA.allContacts.Load(common.PubkeyToHex(&memberB.identity.PublicKey))
+			return ok && contact.DisplayName == "memberB"
+		},
+		"DisplayName is not the same",
+	)
+	s.Require().NoError(err)
+
+	defer s.NoError(admin.Shutdown())
```
cammellos: no need to defer if you run it at the end of the file

cammellos (approved):

protocol/messenger_group_chat_test.go
mohsen-ghafouri: done

protocol/v1/membership_update_message.go
```diff
@@ -448,6 +448,26 @@ func (g Group) Members() []string {
 	return g.members.List()
 }
 
+func (g Group) MemberPublicKeys() ([]*ecdsa.PublicKey, error) {
+	var publicKeys []*ecdsa.PublicKey
+	for _, memberPublicKey := range g.Members() {
+		publickKey, err := hexToPubkey(memberPublicKey)
```
ilmotta: Typo in `publickKey`.

protocol/v1/membership_update_message.go
```diff
@@ -448,6 +448,26 @@ func (g Group) Members() []string {
 	return g.members.List()
 }
 
+func (g Group) MemberPublicKeys() ([]*ecdsa.PublicKey, error) {
+	var publicKeys []*ecdsa.PublicKey
```
ilmotta: It's a good idea to initialize with the capacity of `len(g.Members())`.

ilmotta (approved):

protocol/v1/membership_update_message.go
mohsen-ghafouri: Sorry i didn't get that, can you give me an example?

mohsen-ghafouri: like this? @ilmotta 
```suggestion
	var publicKeys := make([]*ecdsa.PublicKey, 0, len(g.Members()))
```

ilmotta: Exactly @mohsen-ghafouri, if the max capacity of the slice is known, it's better to initialize it with that value.

mohsen-ghafouri: Thanks, done

PR #4187, opened 2023-10-24
feat: kick all members after ownership change and auto-accept after sharing the address

### LINKED ISSUE #11969: [Community Ownership] new Owner should ask members for addresses
Dependent on https://github.com/status-im/status-desktop/issues/11972

At present, only the control node stores the revealed addresses of members. When ownership is transferred to a new owner, this new owner doesn't have access to these revealed addresses. To perform permissions checks, the new owner must request members to reveal their addresses again. ~~Maybe we could introduce a cool-down period, potentially 24 hours, during which members are not kicked for not having revealed their addresses immediately after an ownership transfer.~~ All members should be kicked and asked to rejoin with revealed addresses. This should happen automatically without members interaction. The community will rebuild itself silently when members become online. If control node is offline members should be informed with something like: "The community owner has changed, you will be added back to the community when the control node goes back online".

NOTE: _Although the new owner can approach the previous owner to obtain the revealed addresses, this method poses a security risk. The former owner could send malicious addresses or may no longer be trusted._

### Acceptance criteria:
- [ ] new control node should kick all members and ask them to rejoin with revealed addresses
    - [ ] clients should send request to join automatically without user interaction
- [ ] users should be notified if they have been removed due to owner change and are waiting for control node to become online to accept their request to join
~~- [ ] permissions check should be done after cool-down period, so that members have time to send addresses~~

mprakhov:
- new community message to share an address with the new community owner on the ownership change
- new request to join state RequestToJoinStateOwnershipChanged. On the client side - indicate, that the user was kicked by the owner and required to share the revealed address for the client. On the owner's side, this state will be used to auto-accept requests to join from the client
- API promotes self as a control node
- API to get awaiting address requests to join
- remove CommunityDescription to validate from the validation loop if HandleCommunityDescription returns an error

status-desktop PR: https://github.com/status-im/status-desktop/pull/12560

Closes: https://github.com/status-im/status-desktop/issues/11969


mprakhov:
P.S. Not sure that some fn naming is good, so I'm open for a renamings

protocol/communities/request_to_join.go
```diff
@@ -18,6 +18,7 @@ const (
 	RequestToJoinStateCanceled
 	RequestToJoinStateAcceptedPending
 	RequestToJoinStateDeclinedPending
+	RequestToJoinStateOwnershipChanged
```
osmaczko:
```suggestion
	RequestToJoinStateAwaitingAddresses
```

protocol/communities/persistence.go
```diff
@@ -828,6 +828,10 @@ func (p *Persistence) DeclinedPendingRequestsToJoinForCommunity(id []byte) ([]*R
 	return p.RequestsToJoinForCommunityWithState(id, RequestToJoinStateDeclinedPending)
 }
 
+func (p *Persistence) OwnershipChangedRequestsToJoinForCommunity(id []byte) ([]*RequestToJoin, error) {
```
osmaczko:
```suggestion
func (p *Persistence) AwaitingAddressesRequestsToJoinForCommunity(id []byte) ([]*RequestToJoin, error) {
```

protocol/protobuf/anon_metrics.pb.go
```diff
@@ -1,28 +1,32 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
+// versions:
+// 	protoc-gen-go v1.31.0
+// 	protoc        v4.24.4
```
osmaczko: The `protoc-gen-go` we should use is 1.3.4, see:
```
install-protobuf: ##@install Install Protobuf Generation Tools
	GO111MODULE=on go install github.com/kevinburke/go-bindata/go-bindata@v3.13.0
	GO111MODULE=on go install github.com/golang/protobuf/protoc-gen-go@v1.3.4
```

Otherwise everyone will produce massive changes to all *.pb.go files, making it hard to reveiw.

protocol/protobuf/application_metadata_message.proto
```diff
@@ -91,5 +91,6 @@ message ApplicationMetadataMessage {
     SYNC_CHAT = 74;
     SYNC_ACTIVITY_CENTER_DELETED = 75;
     SYNC_ACTIVITY_CENTER_UNREAD = 76;
+    COMMUNITY_NEW_OWNERSHIP_REQUEST_SHARED_ADDRESS = 77;
```
osmaczko:
```suggestion
    COMMUNITY_REQUEST_SHARED_ADDRESS = 77;
```

Can't we generalize it? It happens that we request shared addresses on ownership change but in the future we may have different triggers.

protocol/protobuf/communities.proto
```diff
@@ -205,3 +205,7 @@ message WakuMessageArchiveIndexMetadata {
 message WakuMessageArchiveIndex {
   map<string, WakuMessageArchiveIndexMetadata> archives = 1;
 }
+
+message CommunityNewOwnershipRequestSharedAddress {
```
osmaczko:
```suggestion
message CommunityRequestSharedAddress {
```

services/ext/api.go
```diff
@@ -646,6 +646,10 @@ func (api *PublicAPI) DeleteCommunityCategory(request *requests.DeleteCommunityC
 	return api.service.messenger.DeleteCommunityCategory(request)
 }
 
+func (api *PublicAPI) CommunityNewOwnershipKickAllAndRequestRejoin(communityID types.HexBytes) (*protocol.MessengerResponse, error) {
```
osmaczko: I believe we don't want explicit endpoint for that, this should happen automatically when we set ourselves as control node.

protocol/communities/manager.go
```diff
@@ -2473,7 +2519,10 @@ func (m *Manager) HandleCommunityRequestToJoin(signer *ecdsa.PublicKey, receiver
 		// If user is already a member, then accept request automatically
 		// It may happen when member removes itself from community and then tries to rejoin
 		// More specifically, CommunityRequestToLeave may be delivered later than CommunityRequestToJoin, or not delivered at all
-		acceptAutomatically := community.AcceptRequestToJoinAutomatically() || community.HasMember(signer)
+		// If community ownership changed, then accept request automatically
+		acceptAutomatically := community.AcceptRequestToJoinAutomatically() || community.HasMember(signer) ||
+			(existingRequestToJoin != nil && existingRequestToJoin.State == RequestToJoinStateOwnershipChanged)
```
osmaczko: Could you please move it few lines up into the `if existingRequestToJoin != nil {` scope?

osmaczko (commented):
Nice :+1: I have few naming improvement proposals and concerns

protocol/protobuf/anon_metrics.pb.go
mprakhov: Huge thanks, it seems I got some auto-updates, will fix that

services/ext/api.go
mprakhov: I thought that we were planning to call it directly from the status-desktop side.
So far it is not clear to me how we will set up as an owner. Will we have a separate API for this? As I understand - we will listen for a community token on the status-desktop side

Do you mean it must be put in `PromoteSelfToControlNode` ?
I'm asking as I have status-desktop changes where I call this API async and if we do not need it - then I need to remove it

P.S. Anyway, it was used by me for testing

osmaczko: The action will be triggered manually by the status-desktop client. The objective is to have a single API: `PromoteSelfToControlNode`. The specifics of what this API does should be determined internally in status-go logic. If the control node changes to a new user, the API will make the current user the new control node, update the contract with itself as a signer, remove all others, and then request addresses. If the control node changes among devices, the API will make the current installation the control node.

mprakhov: Got it, thank you. Should I implement this API or it is already in progress in some other task?

osmaczko: No problem. To the best of my knowledge, it hasn't been implemented yet. It would be great if you could address it within the scope of this PR 🙏

jrainville: @endulab has a PR with the call to setPubkey (to make yourself owner), you can use the same API or create a new one that's called `PromoteSelfToControlNode`

osmaczko:
A small reminder: please try to keep commits self-contained. If someone checks out 4a3cc376a0196b3c17ff7092aa14527409af7bdd, it will not compile because it requires the subsequent commit that runs make generate. Every commit should be healthy so that it compiles and tests pass.

Otherwise, things like `git bisect`  will become almost impossible.

protocol/communities/persistence.go
jrainville: Small variation on Patryk's idea to make it sound better
```suggestion
func (p *Persistence) RequestsToJoinForCommunityAwaitingAddresses(id []byte) ([]*RequestToJoin, error) {
```

protocol/communities_messenger_signers_test.go
```diff
@@ -204,3 +207,221 @@ func (s *MessengerCommunitiesSignersSuite) TestControlNodeUpdateSigner() {
...
+	s.Require().NoError(err)
+
+	_, err = WaitOnMessengerResponse(
+		s.bob,
+		func(r *MessengerResponse) bool {
+			return len(r.Communities()) > 0 && r.Communities()[0].HasMember(&s.bob.identity.PublicKey)
+		},
+		"Bob was auto-accepted",
+	)
+	s.Require().NoError(err)
```
jrainville, line 239: We could add some more checks here to make sure everything is right, like getting the community from Alice's DB and confirming that everyone is a member and that we have Bob's address (John's address will be missing sadly, we need to fix it in another PR)

protocol/messenger_communities.go
```diff
@@ -1417,6 +1417,54 @@ func (m *Messenger) DeleteCommunityCategory(request *requests.DeleteCommunityCat
...
 }
 
+func (m *Messenger) CommunityNewOwnershipKickAllAndRequestRejoin(communityID types.HexBytes) (*MessengerResponse, error) {
+	clock, _ := m.getLastClockWithRelatedChat()
+	changes, err := m.communitiesManager.PromoteSelfToControlNodeAndKickAllMembers(communityID, clock)
+	if err != nil {
+		return nil, err
+	}
+
+	err = m.communitiesManager.CommunityNewOwnershipGenerateRequestsToJoinForAutoApproval(changes.Community.ID(), changes.MembersRemoved)
```
jrainville: Possible new name
Basically, keeping the functions as verbs. Also, I don't think it's necessary to add `Community` in the name, since we're already in the `messenger_communities` file.
```suggestion
	err = m.communitiesManager.GenerateRequestsToJoinForAutoApprovalOnNewOwnership(changes.Community.ID(), changes.MembersRemoved)
```

protocol/messenger_communities.go
```diff
@@ -1417,6 +1417,54 @@ func (m *Messenger) DeleteCommunityCategory(request *requests.DeleteCommunityCat
 	return &response, nil
 }
 
+func (m *Messenger) CommunityNewOwnershipKickAllAndRequestRejoin(communityID types.HexBytes) (*MessengerResponse, error) {
```
jrainville: Possible new name
```suggestion
func (m *Messenger) KickAllAndRequestRejoinOnNewOwnership(communityID types.HexBytes) (*MessengerResponse, error) {
```

mprakhov:
> A small reminder: please try to keep commits self-contained. If someone checks out [4a3cc37](https://github.com/status-im/status-go/commit/4a3cc376a0196b3c17ff7092aa14527409af7bdd), it will not compile because it requires the subsequent commit that runs make generate. Every commit should be healthy so that it compiles and tests pass.
> 
> Otherwise, things like `git bisect` will become almost impossible.

I always do `squash and merge` during the merge, so this problem will be not actual in develop branch. Here I'm trying to split mage generate "spam" and implementation. Also if some changes are done, from time to time I try to put them into a separate commit in order to see the changes

protocol/messenger_communities.go
```diff
@@ -1417,6 +1417,54 @@ func (m *Messenger) DeleteCommunityCategory(request *requests.DeleteCommunityCat
...
+	if err != nil {
+		return nil, err
+	}
+
+	err = m.communitiesManager.CommunityNewOwnershipGenerateRequestsToJoinForAutoApproval(changes.Community.ID(), changes.MembersRemoved)
+	if err != nil {
+		return nil, err
+	}
+
+	newOwnershipRejoinRequest := &protobuf.CommunityNewOwnershipRequestSharedAddress{
```
```suggestion
	newOwnershipRejoinRequest := &protobuf.RequestSharedAddressOnNewOwnership{
```

protocol/messenger_handler.go
```diff
@@ -1542,6 +1542,10 @@ func (m *Messenger) HandleCommunityRequestToJoin(state *ReceivedMessageState, re
 	case communities.RequestToJoinStateCanceled:
 		// cancellation is handled by separate message
 		fallthrough
+	case communities.RequestToJoinStateOwnershipChanged:
+		// ownership changed request is handled only if onwer kick members and saved
```
jrainville:
```suggestion
		// ownership changed request is handled only if owner kicked members and saved
```

protocol/protobuf/application_metadata_message.proto
jrainville: Good point. The members should accept all requests for shared addresses, as long as they come from the rightful owner.

jrainville (commented):
Nice job!

protocol/protobuf/application_metadata_message.proto
mprakhov: removed as I found another solution during dev testing (solution with the signal has a drawback)

services/ext/api.go
mprakhov: Added API PromoteSelfToControlNode

protocol/communities_messenger_signers_test.go
mprakhov: done

mprakhov:
Found a bug with resending requests to join during the testing. When we have a new owner, CommunityDescription can be verified with a delay (due to loop interval) and it can cause the situation when the client received CommunityDescription but did not verify it, and then it receives a resend request to join message, but he is still not kicked and current contract owner != controlNode() pubkey from his old CommunityDescription

Taking into account, that the community description will be added to the verification loop **only when there will be a new control node**, the clients during publishing `TokenCommunityValidated` can detect themself if community has a new control node (owner) and automatically send revealed addresses to him. In this case, we do not need additional signal from control node to all members

protocol/communities/manager.go
```diff
@@ -3240,7 +3271,11 @@ func (m *Manager) PendingRequestsToJoin() ([]*RequestToJoin, error) {
 }
 
 func (m *Manager) PendingRequestsToJoinForUser(pk *ecdsa.PublicKey) ([]*RequestToJoin, error) {
-	return m.persistence.PendingRequestsToJoinForUser(common.PubkeyToHex(pk))
+	return m.persistence.RequestsToJoinForUserByState(common.PubkeyToHex(pk), RequestToJoinStatePending)
+}
+
+func (m *Manager) AwaitingRequestsToJoinForUser(pk *ecdsa.PublicKey) ([]*RequestToJoin, error) {
```
```suggestion
func (m *Manager) AwaitingAddressesRequestsToJoinForUser(pk *ecdsa.PublicKey) ([]*RequestToJoin, error) {
```

Otherwise `AwaitingRequestsToJoinForUser` vs `PendingRequestsToJoinForUser` is very confusing.

protocol/migrations/sqlite/1698329296_add_signature_to_revealed_addresses.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE communities_requests_to_join_revealed_addresses ADD COLUMN signature BLOB;
```
osmaczko: Just a note: `signature` will be empty for existing requests to join. That's ok I suppose? Does Status CCs Community control node have revelead addresses yet?

services/ext/api.go
```diff
@@ -556,6 +556,11 @@ func (api *PublicAPI) MyPendingRequestsToJoin() ([]*communities.RequestToJoin, e
 	return api.service.messenger.MyPendingRequestsToJoin()
 }
 
+// MyAwaitingRequestsToJoin returns requests to join, which must be auto-accepted when control node will be online
+func (api *PublicAPI) MyAwaitingRequestsToJoin() ([]*communities.RequestToJoin, error) {
```
osmaczko:
```suggestion
func (api *PublicAPI) MyAwaitingAddressesRequestsToJoin() ([]*communities.RequestToJoin, error) {
```

protocol/messenger_communities.go
```diff
@@ -5928,18 +5994,28 @@ func (m *Messenger) SetCommunitySignerPubKey(ctx context.Context, communityID []
...
 	clock, _ := m.getLastClockWithRelatedChat()
-
-	community, err := m.communitiesManager.PromoteSelfToControlNode(communityID, clock)
+	changes, err := m.communitiesManager.PromoteSelfToControlNode(communityID, clock)
 	if err != nil {
 		return nil, err
 	}
 
-	err = m.syncCommunity(context.Background(), community, m.dispatchMessage)
+	err = m.syncCommunity(context.Background(), changes.Community, m.dispatchMessage)
```
osmaczko: `syncCommunity` should be invoked last otherwise sync message will lack the generated requests to join.

protocol/communities_messenger_signers_test.go
```diff
@@ -144,7 +150,7 @@ func (s *MessengerCommunitiesSignersSuite) TestControlNodeUpdateSigner() {
 	_, err = WaitOnSignaledMessengerResponse(
 		s.bob,
 		func(r *MessengerResponse) bool {
-			return len(r.Communities()) > 0 && len(r.Communities()[0].CommunityTokensMetadata()) == 1
+			return len(r.Communities()) > 0 && len(r.Communities()[0].TokenPermissions()) == 1
```
osmaczko, line 153: Why this change?

protocol/communities_messenger_signers_test.go
```diff
@@ -154,53 +160,294 @@ func (s *MessengerCommunitiesSignersSuite) TestControlNodeUpdateSigner() {
...
+		},
+		"Bob was auto-accepted",
+	)
+	s.Require().NoError(err)
+
+	community = validateResults(s.bob)
+	s.Require().False(community.IsControlNode())
+	s.Require().False(community.IsOwner())
+
+	// TODO: uncomment when ex-owner will start sharing request to join with revealed address
```
osmaczko, line 245: Shouldn't it happen already with improved implementation of listening to control node change?

protocol/messenger_communities.go
```diff
@@ -365,6 +366,17 @@ func (m *Messenger) handleCommunitiesSubscription(c chan *communities.Subscripti
 						m.logger.Error("failed to save data and prepare response")
 					}
 
+					// if control node changed, the client will be kicked from the community
+					// if owner token will be mint, the client will stay the member of the community
+					if communityResponse.Changes.ShouldMemberLeave {
```
osmaczko: This will be a bit more explicit.

```suggestion
					if communityResponse.Changes.ControlNodeChanged != nil && communityResponse.Changes.ShouldMemberLeave {
```

protocol/messenger_communities.go
```diff
@@ -365,6 +366,17 @@ func (m *Messenger) handleCommunitiesSubscription(c chan *communities.Subscripti
 						m.logger.Error("failed to save data and prepare response")
 					}
 
+					// if control node changed, the client will be kicked from the community
+					// if owner token will be mint, the client will stay the member of the community
```
```suggestion
					// Control node changed and we were kicked out. It now awaits our addresses.
```

protocol/communities_messenger_signers_test.go
mprakhov: Having community token metadata is not a strict requirement for the test. You can have a token metadata, but if you don't have Owner permissions set - then you test will fail

protocol/communities_messenger_signers_test.go
mprakhov: No, the community owner does not keep in the database his own request to join with revealed accounts. We discussed it on Friday meeting, I created a ticket to solve it, it on me

protocol/migrations/sqlite/1698329296_add_signature_to_revealed_addresses.up.sql
mprakhov: some of the users - don't. We have a ticket https://github.com/status-im/status-desktop/issues/11974 so if the signature is absent - we can show the kicked members this msg

jrainville (approved):

osmaczko (approved):

protocol/communities/community.go
```diff
@@ -990,6 +1021,10 @@ func (o *Community) UpdateCommunityDescription(description *protobuf.CommunityDe
 	o.config.CommunityDescription = description
 	o.config.CommunityDescriptionProtocolMessage = rawMessage
 
+	if newControlNode != nil {
+		o.setControlNode(newControlNode)
+	}
+
```
mprakhov, line 1025: @osmaczko please, take a look in this [commit](https://github.com/status-im/status-go/pull/4187/commits/59dd24f19abc0aa99131deaa8b11a2e95cdc99e5) changes

Previously, `ControlNodeChanged` was not set correctly from the loop as we updated the original `Community` ,while we need to update the modified `Community`

PR #4188, opened 2023-10-24
feat: add AC messages for setting signer operations

### LINKED ISSUE #11964: [Community Ownership] enable ownership change
Once the Owner Token is transferred, the new owner needs to set the signer (control node).

### Acceptance criteria:
- [ ] Add `setupControlNode(communityID)` endpoint in messenger that will:
    - update community's Owner Token contract with new signer -> [setSignerPublicKey](https://github.com/status-im/communities-contracts/blob/8d3f185bf4d382d522815cff7cb2b70902cfffa8/contracts/mvp/OwnerToken.sol#L42)(identity.Public())
    - distribute CommunityDescription signed by new control node 
    - notify other devices about this device being the control node (this should be done in scope of https://github.com/status-im/status-desktop/issues/11962)
    - create and distribute encryption keys (this should be done in scope of https://github.com/status-im/status-desktop/issues/11963)

endulab:
Issue [#11964](https://github.com/status-im/status-desktop/issues/11964)

Adding registering of AC messages for different phases of ownership transfer.
Exposing PromoteSelfToControlMode.


jrainville (approved):

protocol/messenger_communities.go
```diff
@@ -5145,3 +5145,46 @@ func (m *Messenger) PromoteSelfToControlNode(communityID types.HexBytes) (*commu
...
+	err := m.addActivityCenterNotification(response, notification, nil)
+	if err != nil {
+		m.logger.Error("failed to save notification", zap.Error(err))
+		return response, err
+	}
+
+	return response, nil
+}
+
+func (m *Messenger) RegisterOwnerTokenReceivedNotification(communityID string) (*MessengerResponse, error) {
```
MishkaRogachev: Just suggestion: you can reduce this PR to one function, if you pass `acType ActivityCenterType` in rpc arguments from nim

MishkaRogachev (approved):

protocol/activity_center.go
```diff
@@ -29,6 +29,11 @@ const (
 	ActivityCenterNotificationTypeContactVerification
 	ActivityCenterNotificationTypeContactRemoved
 	ActivityCenterNotificationTypeNewKeypairAddedToPairedDevice
+	ActivityCenterNotificationTypeOwnerTokenReceived
+	ActivityCenterNotificationTypeSetSignerSuccess
+	ActivityCenterNotificationTypeSetSignerFailed
+	ActivityCenterNotificationTypeSetSignerDeclined
+	ActivityCenterNotificationTypeLostOwnership
```
osmaczko: Don't we need yet another notification type -> `ActivityCenterNotificationTypeBecameControlNode`? (ActivityCenterNotificationTypeSetSignerSuccess != ActivityCenterNotificationTypeBecameControlNode)

https://www.figma.com/file/qHfFm7C9LwtXpfdbxssCK3/Kuba%E2%8E%9CDesktop---Communities?type=design&node-id=37206-87325&mode=design&t=Klfu2omdUgTGSnWN-0

services/ext/api.go
```diff
@@ -1546,6 +1546,35 @@ func (api *PublicAPI) CreateTokenGatedCommunity() (*protocol.MessengerResponse,
...
+
+// Returns response with AC notification when setting signer is failed
+func (api *PublicAPI) RegisterSetSignerFailedNotification(communityID string) (*protocol.MessengerResponse, error) {
+	return api.service.messenger.RegisterSetSignerFailedNotification(communityID)
+}
+
+// Returns response with AC notification when setting signer is declined
+func (api *PublicAPI) RegisterSetSignerDeclinedNotification(communityID string) (*protocol.MessengerResponse, error) {
+	return api.service.messenger.RegisterSetSignerDeclinedNotification(communityID)
+}
```
osmaczko: These shouldn't be needed as API, it should be determined behind the scenes. See: https://github.com/status-im/status-go/pull/4187#discussion_r1370598171

services/ext/api.go
```diff
@@ -1546,6 +1546,35 @@ func (api *PublicAPI) CreateTokenGatedCommunity() (*protocol.MessengerResponse,
 	return api.service.messenger.CreateTokenGatedCommunity()
 }
 
+// Returns response with AC notification when owner token is received
+func (api *PublicAPI) RegisterOwnerTokenReceivedNotification(communityID string) (*protocol.MessengerResponse, error) {
```
osmaczko, line 1635: I am wondering why do we need it as an API? I guess it is because we don't have defined way to talk from wallet to messenger, right?

services/ext/api.go
```diff
@@ -1546,6 +1546,35 @@ func (api *PublicAPI) CreateTokenGatedCommunity() (*protocol.MessengerResponse,
...
+func (api *PublicAPI) RegisterSetSignerFailedNotification(communityID string) (*protocol.MessengerResponse, error) {
+	return api.service.messenger.RegisterSetSignerFailedNotification(communityID)
+}
+
+// Returns response with AC notification when setting signer is declined
+func (api *PublicAPI) RegisterSetSignerDeclinedNotification(communityID string) (*protocol.MessengerResponse, error) {
+	return api.service.messenger.RegisterSetSignerDeclinedNotification(communityID)
+}
+
+func (api *PublicAPI) RegisterLostOwnershipNotification(communityID string) (*protocol.MessengerResponse, error) {
```
osmaczko, line 1655: I am wondering why do we need it as an API? I guess it is because we don't have defined way to talk from wallet to messenger, right?

protocol/activity_center.go
endulab: Actually, I can rename ActivityCenterNotificationTypeSetSignerSuccess to ActivityCenterNotificationTypeBecameControlNode or ActivityCenterNotificationTypeReceivedOwnership.
Because we do not need these two in AC, but only one in the end of the whole process (promoting to control node happens just after setting signer).

services/ext/api.go
endulab: Yes, however I already talked to Stefan and I will create a task for this. We need some kind of listener on wallet feed in the messenger. It should add AC messages on some concrete wallet actions. Also it should promoteToControlNode when signer is set (by observing setting signer transaction). It should happen w/o the client interaction. But this is for next milestone.

PR #4189, opened 2023-10-24
feat: Added ERC20 community id

### LINKED ISSUE #12447: Add community path to erc20
* [x] update the contract with a new attribute
* [x] populate the new attribute when deploying the contract
* [x] add to the custom token table with the new community id attributes
* [x] when receiving contract from community populate this attribute (the custom token is already created)
* [x] when receiving a transaction check if this attribute exist and if it does populate the custom attribute
* [x] Only return custom token that have a community id
* [x] Add the community id attribute to the token being exposed

Cuteivist:
closes https://github.com/status-im/status-desktop/issues/12447

* Use community ERC20 contract to detect community ID when receiving community token mint tx
* Added column to `tokens` table. It is NULL by default to populate it with empty string for non-community tokens to prevent further checks.


jrainville (approved):

services/wallet/reader.go
```diff
@@ -200,6 +212,8 @@ func (r *Reader) GetWalletToken(ctx context.Context, addresses []common.Address)
 		allTokens = append(allTokens, r.tokenManager.ToToken(network))
 	}
 
+	allTokens = removeCommunityTokens(allTokens)
```
alaibe: why remove?

services/wallet/transfer/commands.go
```diff
@@ -582,7 +582,10 @@ func (c *transfersCommand) processUnknownErc20CommunityTransactions(ctx context.
 	for _, tx := range allTransfers {
 		if tx.Type == w_common.Erc20Transfer {
 			// Find token in db or if this is a community token, find its metadata
-			_ = c.tokenManager.FindOrCreateTokenByAddress(ctx, tx.NetworkID, *tx.Transaction.To())
+			token := c.tokenManager.FindOrCreateTokenByAddress(ctx, tx.NetworkID, *tx.Transaction.To())
```
alaibe: This should be also run when seeing community on the network, see https://github.com/status-im/status-go/pull/4013/files

services/wallet/transfer/commands.go
```diff
@@ -582,7 +582,10 @@ func (c *transfersCommand) processUnknownErc20CommunityTransactions(ctx context.
...
+			if token != nil && token.CommunityID == nil {
+				c.tokenManager.DiscoverTokenCommunityID(ctx, token, *tx.Transaction.To())
+			}
```
alaibe: Why not doing it in one go?

services/wallet/reader.go
Cuteivist: AFAIK community tokens aren't showed as normal tokens.

There is separate API where we get all the tokens (default and customs in which community can be found) - getTokens, called for each chainID.

Here tokens are collected and their history is calculated. Community tokens don't have price, so they can be skipped from that list.

services/wallet/transfer/commands.go
Cuteivist: A little optimisation. 

So this will be called for any ERC20 transaction that is discovered. At that point we might already marked token as community or non community token (that's why community_id in DB is NULL, so we can easily deduct that token with empty community_id is not community token, but was checked)

So this if statement prevents duplicated check every time there is new tx with token that has community link checked.

Cuteivist: On second thought I moved that if statement inside the `DiscoverTokenCommunityID`.

services/wallet/transfer/commands.go
Cuteivist: @alaibe Thanks for pointing that out. To prevent that issue in the future I moved Discover method inside the `FindOrCreateToken` function.

services/wallet/token/token.go
```diff
@@ -288,9 +294,58 @@ func (tm *Manager) FindOrCreateTokenByAddress(ctx context.Context, chainID uint6
 		return nil
 	}
 
+	tm.DiscoverTokenCommunityID(context.Background(), token, address)
 	return token
 }
 
+func (tm *Manager) DiscoverTokenCommunityID(ctx context.Context, token *Token, address common.Address) {
```
alaibe: nit: can be private with a lowercase `d`?

alaibe (approved):

services/wallet/reader.go
Cuteivist: Removed after discussion.

services/wallet/token/token.go
Cuteivist: Good point!

PR #4193, opened 2023-10-24
fix: SetSignerPublicKey return address with 0x prefix

mprakhov:
SetSignerPublicKey returned PubKey as a string without `0x` prefix before
This lead to converting string -> PubKey error

### DIFF
--- a/eth-node/types/bytes.go
+++ b/eth-node/types/bytes.go
@@ -39,3 +39,8 @@ func Hex2Bytes(str string) []byte {
 	h, _ := hex.DecodeString(str)
 	return h
 }
+
+// ToHex returns the hex string representation of bytes with 0x prefix.
+func ToHex(bytes []byte) string {
+	return "0x" + Bytes2Hex(bytes)
+}
--- a/services/communitytokens/service.go
+++ b/services/communitytokens/service.go
@@ -13,6 +13,7 @@ import (
 	"github.com/status-im/status-go/account"
 	"github.com/status-im/status-go/contracts/community-tokens/ownertoken"
 	communityownertokenregistry "github.com/status-im/status-go/contracts/community-tokens/registry"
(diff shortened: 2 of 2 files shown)


jrainville (approved):

vendor/github.com/ethereum/go-ethereum/common/bytes.go
```diff
@@ -36,6 +36,15 @@ func FromHex(s string) []byte {
 	return Hex2Bytes(s)
 }
 
+// ToHex returns the hex string representation of bytes with 0x prefix.
+func ToHex(bytes []byte) string {
```
osmaczko: Changing anything in `vendor/*` is a bad idea, it will be vanished with `make vendor`

osmaczko (changes requested):
Good catch but needs different approach.

vendor/github.com/ethereum/go-ethereum/common/bytes.go
mprakhov: yes, pipeline already pointed me on my mistake :) 
Fixed

jrainville (approved):

eth-node/types/bytes.go
```diff
@@ -39,3 +39,12 @@ func Hex2Bytes(str string) []byte {
...
 	return h
 }
+
+// ToHex returns the hex string representation of bytes with 0x prefix.
+func ToHex(bytes []byte) string {
+	str := Bytes2Hex(bytes)
+	if !has0xPrefix(str) {
+		str += "0x" + str
+	}
+	return str
```
osmaczko:
```suggestion
	return "0x" + Bytes2Hex(bytes)
```

`Bytes2Hex` never returns with prefix, does it?

osmaczko (approved):

Issue #3833, opened 2023-08-01
Usage of transfer history fetch


alaibe:
# Problem

Calculate the usage of full transfer history fetch

If acceptable for mobile, the on demand strategy can removed

rasom:
**fetching tx history for one of my accounts** (which has 550+ tx per etherscan (that excludes incoming ERC20 transfers in most of cases) and something around 50 incoming ERC20) **takes 65-85MB of data and more than 7k requests**

what is a bit suspicious (although I don't know in details how that works) is that each time I run it i get a different number of `eth_TransactionReceipt` and `eth_FullTransactionByBlockAndIndex` (maybe @IvanBelyakoff knows whether that's expected)

IvanBelyakoff:
@rasom How different are they? Are you using `rpcstats.CountCall` for that? Did you remove the data from DB each time? Because we don't detele the data if account is removed.

The only thing that comes to my mind irrespective of how this is tested and measured is that every time it is measured new blocks have been generated and history checked intervals are different which may result in slight variations of calls to find the blocks that have ETH transfers. For ERC20 I think that the difference can be even smaller if there is.

rasom:
@IvanBelyakoff 
> How different are they?

so I logged three runs so far, I have 367/221, 422/282, 306/247 for `eth_TransactionReceipt`/`eth_FullTransactionByBlockAndIndex`

> Are you using rpcstats.CountCall for that?

yes

>  Did you remove the data from DB each time? 

I do remove entire multiacc (status account, not wallet account) and from data usage point of view it looks like the database is also removed (or not used after account recovery via phrase), otherwise it would be hard to explain 65-85MB of traffic.

> history checked intervals are different which may result in slight variations of calls to find the blocks that have ETH transfers

Yes I'm aware of it, that's why I wrote specifically about calls which counts shouldn't vary per my understanding.

rasom:
I also just noticed that for that run which had `306/247` there were only 264 `eth_BlockByNumber` which is likely not enough to scan entire history of that wallet account (500+ transfer and most of them in different blocks), so it looks like that scanning stalls on mobile at some point. I have to figure out what causes this and then rerun tests.

rasom:
sooo, it stops loading historical data after the first "check recent blocks" request like
[...]

PR #4197, opened 2023-10-25
[#3833] Increase intreval of new blocks detection command

rasom:
related to #3833

currently an idle RPC load on a newly created multipack with a single wallet acc and single network looks like this
```Clojure
;; => {:total 1883,
;;     :methods
;;     ([:eth_FilterLogs 556]
;;      [:eth_CallContract 453]
;;      [:eth_HeaderByNumber 282]
;;      [:eth_BalanceAt 278]
;;      [:eth_NonceAt 278]
;;      [:eth_CodeAt 18]
;;      [:eth_BlockNumber 18])}
```
of which all calls except `eth_CallContract` are idle load of the routine which looks for new blocks with transfers, 1430 requests per hour, ~34k requests per day. By increasing routine's interval to 2m from 13s we reduce the number of requests around 8-10 times, depends on network

```Clojure
;; => {:total 635,
;;     :methods
;;     ([:eth_CallContract 442]
;;      [:eth_FilterLogs 60]
;;      [:eth_HeaderByNumber 36]
;;      [:eth_BalanceAt 31]
;;      [:eth_NonceAt 30]
;;      [:eth_CodeAt 18]
;;      [:eth_BlockNumber 18])}
```
of which 193 calls are related to the routine, projects to ~4.6k requests per day.

That is still a lot, but there are other ways to improve this number without changing the interval.

The cons is that in the worst case a new transfer will be detected with 2 minutes delay (comparing to 13s at the moment), but in most cases it is acceptable and if necessary can be adjusted in some specific situations (like sending of tx initiated by the user or additional scan initiated by opening wallet tab, etc).

status: ready

### DIFF
--- a/services/wallet/transfer/commands_sequential.go
+++ b/services/wallet/transfer/commands_sequential.go
@@ -23,7 +23,10 @@ type findNewBlocksCommand struct {
 
 func (c *findNewBlocksCommand) Command() async.Command {
 	return async.InfiniteCommand{
-		Interval: 13 * time.Second, // TODO - make it configurable based on chain block mining time
+		// TODO - make it configurable based on chain block mining time
+		// NOTE(rasom): ^ it is unclear why each block has to be checked,
+		// that is rather undesirable, as it causes a lot of RPC requests
+		Interval: 2 * time.Minute,
 		Runable:  c.Run,
 	}.Run
 }


stefandunca (approved):
It sounds like a good idea to have longer polling times but refresh on request (if the activity view is visible, on user operations ...).

PR #4208, opened 2023-10-25
fix: ensure receivedMessage.CommunityID in chat message handler is set

osmaczko:
This enables clients to obtain invitation's community ID even when description processing is queued.

part of: status-im/status-desktop#12481

### DIFF
--- a/protocol/messenger_handler.go
+++ b/protocol/messenger_handler.go
@@ -1689,11 +1689,6 @@ func (m *Messenger) HandleCommunityRequestToLeave(state *ReceivedMessageState, r
 	return nil
 }
 
-// handleWrappedCommunityDescriptionMessage handles a wrapped community description
-func (m *Messenger) handleWrappedCommunityDescriptionMessage(payload []byte, shard *common.Shard) (*communities.CommunityResponse, error) {
-	return m.communitiesManager.HandleWrappedCommunityDescriptionMessage(payload, shard)
-}
-
 func (m *Messenger) handleEditMessage(state *ReceivedMessageState, editMessage EditMessage) error {
 	if err := ValidateEditMessage(editMessage.EditMessage); err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


protocol/messenger_handler.go
```diff
@@ -2282,20 +2277,17 @@ func (m *Messenger) handleChatMessage(state *ReceivedMessageState, forceSeen boo
 	if receivedMessage.ContentType == protobuf.ChatMessage_COMMUNITY {
 		m.logger.Debug("Handling community content type")
 
-		communityResponse, err := m.handleWrappedCommunityDescriptionMessage(receivedMessage.GetCommunity(), common.ShardFromProtobuff(receivedMessage.Shard))
+		signer, description, err := communities.UnwrapCommunityDescriptionMessage(receivedMessage.GetCommunity())
```
jrainville, line 2280: Do we still handle the shard somewhere?

osmaczko: Good point :thinking:

osmaczko: https://github.com/status-im/status-go/blob/3805662a1834dd65c2b317156b2ce4e1ee2a3950/protocol/messenger_communities.go#L2859

It seems we don't use shard when we handle community description :man_shrugging: - nil is hardcoded

protocol/messenger_handler.go
```diff
@@ -2282,20 +2277,22 @@ func (m *Messenger) handleChatMessage(state *ReceivedMessageState, forceSeen boo
...
+			return err
 		}
 
-		community := communityResponse.Community
-		receivedMessage.CommunityID = community.IDString()
-
-		state.Response.AddCommunity(community)
-		state.Response.CommunityChanges = append(state.Response.CommunityChanges, communityResponse.Changes)
+		if len(description.ID) != 0 {
+			receivedMessage.CommunityID = description.ID
```
mprakhov, line 2291: I do not understand how it can fix the issue... will all necessary data be fetched from the mailserver by `receivedMessage.CommunityID` in this case? 
Here it is a new community with a minted owner token, so this community description will be added to the queue, while the user will receive a message without a description and members count

protocol/messenger_handler.go
jrainville: Interesting. @richard-ramos do we need the shard?

richard-ramos: I have a TODO for that: https://github.com/status-im/status-go/blob/develop/protocol/communities/manager.go#L464  but i have done some dogfooding past week and I think the shard ain't necessary in here. I can send a PR to remove it completely from handleWrappedCommunityDescriptionMessage

osmaczko: Thanks @richard-ramos , I removed `handleWrappedCommunityDescriptionMessage` completely with this PR, there is no need to address that.

protocol/messenger_handler.go
osmaczko: From the description: "This enables clients to obtain invitation's community ID even when description processing is queued."

When we receive this message, we display the card in a pending state and listen for async responses. Since we now possess the community ID, we know which community to monitor in the response, ensuring that the card is eventually displayed in its fully rendered state.

mprakhov: Thank you, the proposed solution sounded more scared than the implementation 😄

mprakhov (approved):

Issue #4191, opened 2023-10-24
Remove generated files from source control


osmaczko:
# Problem

Remove the output of  `make generate` from source control. Nowadays, almost every PR produces a ridiculous number of changes because contributors use different versions of the tools responsible for generation. This makes PRs difficult to review since most of the changes are unrelated to the actual issue.

# Acceptance Criteria

- [x] remove all generated files from source control
- [ ] enforce the use of the Nix shell in the Makefile to ensure uniform tool usage

osmaczko:
@jakubgs I would love to hear your opinion. Is there anything else I should think about?

jakubgs:
I'm all for not storing generated files in the repo, that is silly. We can always do that at build time.

But the fundamental issue is that we're using varied version of tooling, and this could be easily remedied by using an already defined Nix shell on dev environments:
https://github.com/status-im/status-go/blob/develop/shell.nix

We could achieve this the same way we do it in `status-mobile`, by enforcing use of the Nix shell in the `Makefile`, like so:
```Makefile
# WARNING: This has to be located right before all the targets.
SHELL := ./nix/scripts/shell.sh

shell: export TARGET ?= default
shell: ##@prepare Enter into a pre-configured shell
ifndef IN_NIX_SHELL
	@ENTER_NIX_SHELL
else
	@echo "${YELLOW}Nix shell is already active$(RESET)"
endif
```
https://github.com/status-im/status-mobile/blob/ca2b3abb1c3091712fc2d5e4300185394a6fb77f/Makefile#L67-L76

Of course some shells would not use Nix shell, but that can be always overriden using the `SHELL` variable, for example:
```Makefile
_fix-node-perms: SHELL := /bin/sh
_fix-node-perms: ##@prepare Fix permissions so that directory can be cleaned
	$(shell test -d node_modules && chmod -R 744 node_modules)
	$(shell test -d node_modules.tmp && chmod -R 744 node_modules.tmp)
```
https://github.com/status-im/status-mobile/blob/ca2b3abb1c3091712fc2d5e4300185394a6fb77f/Makefile#L123-L126

It would be fairly simple to adopt Nix setup from `status-mobile` to `status-go`, me and @yakimant can help.

osmaczko:
[...]

PR #4209, opened 2023-10-25
feat: Profile showcase data sharing

### LINKED ISSUE #4143: Profile showcase data sharing
# Problem

Screenshot 2023-10-10 at 19 32 37

A user can present communities, wallet accounts, assets and collectibles in [profile showcase](https://github.com/status-im/status-desktop/issues/11081). Those preferences can have four levels of visibility:

-  NoOne
-  IDVerifiedContacts
-  Contacts 
-  Everyone

Profile preferences should be available to other users based on their level of visibility. For example, mutual contacts can see the `Contacts` and `Everyone` groups, but not `IDVerifiedContacts`.

# Implementation

I propose to discuss the implementation of access levels in the commentary. We have previously discussed two approaches:
 - Send broadcast messages encrypting `Contacts` and `IDVerifiedContacts` for mutual contacts and for id verified contacts. (Will probably need to add two key pairs here).
 - Send `Contacts` and `IDVerifiedContacts` levels in 1-to-1 mode, running through the contacts list.

# Acceptance Criteria

1. Add broadcasting message for profile showcase preferences 
2. Send messages after showcase preferences were changed, contact added or id verified
3. Add table with latests obtained showcase preferences for each known user
4. Add endpoint for getting showcase preferences by contactID

MishkaRogachev:
Provide broadcasting of profile showcase changes for 3 categories: everyone, mutual contacts, and ID verified contacts.

Important changes:
- [x] Add .proto messages for profile showcase updates
- [x] Dispatch profile showcase updates on profile preference changes
- [x] Send unencrypted profile showcase updates for everyone 
- [x] Send encrypted profile showcase updates for mutual contacts and ID verified contacts
- [x] Receive and handle profile showcase updates
- [x] Decrypt showcase updates for mutual contacts and ID verified contacts
- [x] Store profile showcase updates for contacts in the database
- [x] Promote changes to UI

Closes #4143


MishkaRogachev:
FYI:  this PR contains temporary commit with protobuf binaries, see https://github.com/status-im/status-go/issues/4191

MishkaRogachev:
@saledjenic is it better to add synchronisation and backup for showcase in another PR?

protocol/messenger_profile_showcase.go
```diff
@@ -86,3 +124,193 @@ func (m *Messenger) GetProfileShowcasePreferences() (*ProfileShowcasePreferences
...
+		Assets:       toProfileShowcaseUpdateEntries(preferences.Assets, ProfileShowcaseVisibilityIDVerifiedContacts),
+	}
+
+	mutualContacts := []*Contact{}
+	iDVerifiedContacts := []*Contact{}
+
+	m.allContacts.Range(func(_ string, contact *Contact) (shouldContinue bool) {
+		if contact.mutual() {
+			mutualContacts = append(mutualContacts, contact)
+			if contact.IsVerified() {
```
saledjenic, line 265: @MishkaRogachev isn't a verified contact at the same time the mutual contact as well? I guess we should be mutually exclusive here.
```
			if contact.IsVerified() {
							iDVerifiedContacts = append(iDVerifiedContacts, contact)
			} else if contact.mutual() {
							mutualContacts = append(mutualContacts, contact)
			}
```

protocol/messenger_profile_showcase.go
```diff
@@ -86,3 +124,193 @@ func (m *Messenger) GetProfileShowcasePreferences() (*ProfileShowcasePreferences
...
+	}
+
+	return &protobuf.ProfileShowcase{
+		ForEveryone:           forEveryone,
+		ForContacts:           forContactsEncrypted,
+		ForIdVerifiedContacts: forIDVerifiedContactsEncrypted,
+	}, nil
+}
+
+func (m *Messenger) BuildProfileShowcaseFromIdentity(senderPubKey *ecdsa.PublicKey, message *protobuf.ProfileShowcase) (*identity.ProfileShowcase, error) {
```
saledjenic: @MishkaRogachev do we protect somehow that messages meant to be delivered to verified contacts can not be revealed by mutual contacts?

protocol/protobuf/profile_showcase.proto
```diff
@@ -0,0 +1,28 @@
+syntax = "proto3";
+
+option go_package = "./;protobuf";
+package protobuf;
+
+message ProfileShowcaseEntry {
+  string entry_id = 3;
```
saledjenic: @MishkaRogachev why don't we start from `1` here?

saledjenic:
> @saledjenic is it better to add synchronisation and backup for showcase in another PR?

Whichever you prefer more.

protocol/protobuf/profile_showcase.proto
MishkaRogachev: it's typo, fixed :)

protocol/messenger_profile_showcase.go
MishkaRogachev: If I understand the identity verification correctly, verifification avaliable only for mutual contacts, so here i explicitly ask for being mutual contacts for identity verified contacts:

| Visibility | For Everyone | For Contacts | For ID Verified Contacts 
| ------ | ------------- | ------------- | ------------- |
Everyone | true | false | false
Mutual contacts | true | true | false
Id verified mutual contacts | true | true | true

protocol/messenger_profile_showcase.go
MishkaRogachev: Yes, each group encrypted with with explicitly specified contact's keys, same as  in identity image flow

protocol/messenger_profile_showcase.go
saledjenic: I agree, but I see that as verified contacts are a subset of mutual contacts, right?

But what I'm pointing you here is that in the current code, the same verified contact, let's say X, will be added to both arrays, `mutualContacts` and `iDVerifiedContacts`. It means that contact X will receive a single message meant to be used for mutual and another message meant to be used for verified contacts, that further means contact X will have a profile showcase for a user that is carried by the last received message, regardless is it mutual to verified contact.

Is that correct?

protocol/messenger_profile_showcase.go
```diff
@@ -9,6 +20,32 @@ type ProfileShowcasePreferences struct {
 	Assets       []*ProfileShowcaseEntry `json:"assets"`
 }
 
+func toProfileShowcaseUpdateEntries(entries []*ProfileShowcaseEntry, visibility ProfileShowcaseVisibility) []*protobuf.ProfileShowcaseEntry {
+	result := []*protobuf.ProfileShowcaseEntry{}
+
+	for _, entry := range entries {
+		if entry.ShowcaseVisibility == visibility {
```
igor-sirotin: Maybe fast return here?
```go
if entry.ShowcaseVisibility != visibility {
    continue
}
// ...
```

protocol/messenger_profile_showcase.go
MishkaRogachev:
```
		if contact.IsVerified() {
			iDVerifiedContacts = append(iDVerifiedContacts, contact)
		} else if contact.mutual() {
			mutualContacts = append(mutualContacts, contact)
		}
```
With this implementation, a ID verified contact will be able to decrypt `forIDVerifiedContacts` part but will not be able to decrypt `forContacts` part because it won't contain his key. The test will fail [here](https://github.com/status-im/status-go/blob/f1dc7be2aa298323d2b9e29732b2e72666969016/protocol/messenger_profile_showcase_test.go#L359-L360). So i will need to encrypt `forContacts` message both with `iDVerifiedContacts` and `mutualContacts` keys. 

Each 'protobuf.ProfileShowcaseEntries' message contains only those items that are visible to the corresponding category. They are dispatched together in one `protobuf.ProfileShowcase` message, so end user will always use latest  message, regardless is it mutual, verified or neither of them.

saledjenic (approved):

protocol/messenger_profile_showcase.go
igor-sirotin: @saledjenic I guess it's also about the contained data.
A verified contact should process all of the 3 categories to get full showcase. `forIDVerifiedContacts` only contains part of the showcase that's available to verified users.

protocol/messenger_profile_showcase_test.go
```diff
@@ -1,34 +1,177 @@
...
+
+	// Send contact request
+	_, err := s.m.SendContactRequest(context.Background(), request)
+	s.Require().NoError(err)
+
+	// Wait for the message to reach its destination
+	_, err = WaitOnMessengerResponse(
+		theirMessenger,
+		func(r *MessengerResponse) bool {
+			return len(r.Contacts) > 0 && len(r.Messages()) > 0 && len(r.ActivityCenterNotifications()) > 0
```
igor-sirotin: I think `ActivityCenterNotifications` shouldn't be part of wait condition, because it's created on the fly. We should check it afterwards, yes, but no during wait.

protocol/messenger_profile_showcase_test.go
```diff
@@ -1,34 +1,177 @@
...
+
+	resp, err = s.m.VerifiedTrusted(context.Background(), &requests.VerifiedTrusted{ID: types.FromHex(verificationRequestID)})
+	s.Require().NoError(err)
+
+	s.Require().Len(resp.Messages(), 1)
+	s.Require().Equal(common.ContactVerificationStateTrusted, resp.Messages()[0].ContactVerificationState)
 }
 
-func (s *MessengerProfileShowcaseSuite) TestSetAndGetProfileShowcasePreferences() {
+func (s *TestMessengerProfileShowcase) prepareShocasePreferencs() ProfileShowcasePreferences {
```
igor-sirotin: 2 typos combo! 😄 
```suggestion
func (s *TestMessengerProfileShowcase) prepareShowcasePreferences() ProfileShowcasePreferences {
```

protocol/migrations/sqlite/1698396318_profile_showcase_contacts.up.sql
```diff
@@ -0,0 +1,9 @@
+DROP TABLE profile_showcase_contacts;
+
+CREATE TABLE IF NOT EXISTS profile_showcase_contacts (
```
igor-sirotin: I guess `IF NOT EXISTS` is not needed if we drop the table just one line earlier 😄

igor-sirotin (approved):
LGTM, nice job! 💪

igor-sirotin:
@MishkaRogachev this is not related to storing the profile of ourselves, right?
If it is, then, please, make sure that `messenger.selfContact` is properly updated 🙏

MishkaRogachev:
@saledjenic @igor-sirotin, thanks fort review! 
> If it is, then, please, make sure that messenger.selfContact is properly updated 🙏

No, as far as we don't use showcase data for user urls :)

protocol/identity_images.go
```diff
@@ -77,7 +77,7 @@ image:
 			// Decrypt the main encryption AES key with AES encryption using the DH key
 			dAESKey, err := common.Decrypt(empk, sharedKey)
 			if err != nil {
-				if err.Error() == "cipher: message authentication failed" {
+				if err.Error() == "cipher: message authentication failed" { // nolint: goconst
```
cammellos: we should probably const this instead of disabling lint :)

protocol/persistence.go
```diff
@@ -727,6 +727,12 @@ func (db sqlitePersistence) Contacts() ([]*Contact, error) {
 			}
 			contact.SocialLinks = append(contact.SocialLinks, link)
 		}
+
+		profileShowcase, err := db.GetProfileShowcaseForContact(contact.ID)
```
cammellos: this probably should be a join for performance reasons? (same with social links) at least we could do 
(1 for contacts, one for social links, one for profile showcase), instead of now that we do`2*nContacts` + 1 query, I think that's going to be slow maybe?

protocol/identity_images.go
MishkaRogachev: fixed :)

protocol/persistence.go
MishkaRogachev: The thing is, both for social links and profile entries, there are several rows for one contact id, so `LEFT JOIN` will trigger a whole query to have several rows. But challenge to optimise this func is interesting. Can I try to do it in separated PR ?

cammellos: sure, that's no problem

MishkaRogachev: https://github.com/status-im/status-go/issues/4239

MishkaRogachev:
@saledjenic @igor-sirotin @cammellos 

~~i've re-requested review because of add ing the meta field. I'm going to use it to share some extra metadata for ui proposal~~

I've decided to separate profile entries by category to be able provide custom fields

protocol/protobuf/profile_showcase.proto
```diff
@@ -0,0 +1,45 @@
...
+
+message ProfileShowcaseAccount {
+  string address = 1;
+  string name = 2;
+  string color_id = 3;
+  string emoji = 4;
+  uint32 order = 5;
+}
+
+message ProfileShowcaseCollectible {
```
MishkaRogachev, line 19: @dlipicar can you please check if there is a need to add any additional fields to assets and collectibles?

protocol/identity_images.go
```diff
@@ -9,6 +9,8 @@ import (
 	"github.com/status-im/status-go/protocol/protobuf"
 )
 
+const CipherMessageAutentificationFailed = "cipher: message authentication failed"
```
cammellos: `ErrCipher...`

cammellos (commented):
looks ok, main concern is performance of having to query multiple tables in separate queries, did you check though queries are covered by indexes? you are querying by contact-id most of the time, but primary key is composite, so depending on how the index is created (community first, contact last I guess), it could be slow if the community is popular, or viceversa if the user has many communities for example.

Also worth considering having a single table for your own? though no strong opinion on my side

protocol/identity_images.go
MishkaRogachev: fixed

protocol/protobuf/profile_showcase.proto
dlipicar: not sure what the "uid" is. To identify any collectible you need three separate components:
- chainID
- contract address
- tokenID

protocol/messenger_profile_showcase.go
```diff
@@ -1,88 +1,342 @@
...
 
-	allPreferences := []*ProfileShowcaseEntry{}
+	return m.publishContactCode()
+}
+
+func (m *Messenger) GetProfileShowcasePreferences() (*ProfileShowcasePreferences, error) {
+	return m.persistence.GetProfileShowcasePreferences()
+}
+
+func (m *Messenger) EncryptProfileShowcaseEntriesWithContactPubKeys(entries *protobuf.ProfileShowcaseEntries, contacts []*Contact) (*protobuf.ProfileShowcaseEntriesEncrypted, error) {
```
igor-sirotin, line 142: The list of contacts might be quite big actually, maybe pass it by pointer?

protocol/persistence_profile_showcase.go
```diff
@@ -49,7 +93,340 @@ func (db sqlitePersistence) InsertOrUpdateProfileShowcasePreference(entry *Profi
...
+		contactID,
+		community.UID,
+		community.Order,
+	)
+
+	if err != nil {
+		return err
+	}
+
+	return nil
```
igor-sirotin: Looks like it can be simplified 😄 
(there're a few such places in this file)
```suggestion
	return err
```

igor-sirotin (approved):
Nice job! LGTM 👍

protocol/messenger_profile_showcase.go
MishkaRogachev: it's a slice, so it's already a pointer

cammellos (approved):

PR #4212, opened 2023-10-26
Feat/wallet cache of community metadata

### LINKED ISSUE #12521: Keep local cache of collectibles community metadata
Client should only fetch information from the DB and get an "immediate" response. Any access to Waku/providers should be done in an async manner if desired.

Keep a local cache of the collectibles community metadata in the wallet DB, which is refreshed together with the ownership list.

dlipicar:
Fixes https://github.com/status-im/status-desktop/issues/12521

We now store a local cache of community metadata in the wallet.
Community metadata cache is refreshed every time collectible ownership is refreshed. As an additional improvement, if fetching for a given community fails, we skip fetching for other collectibles from the same community.
Cached data is only refreshed if all needed data for a given collectible can be properly fetched.
Only cached data is used when requesting collectibles from the client.

This could introduce a slight inconsistency when some community properties change (for example color) until the update is triggered. An improvement to make this less noticeable is planned for the future.

### DIFF
--- a/node/status_node_services.go
+++ b/node/status_node_services.go
@@ -106,7 +106,7 @@ func (b *StatusNode) initServices(config *params.NodeConfig, mediaServer *server
 	// Community collectibles.
 	// Messenger needs the CollectiblesManager to get the list of collectibles owned
 	// by a certain account and check community entry permissions.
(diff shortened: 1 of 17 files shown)


stefandunca (approved):
It looks great!

alaibe (approved):

PR #4216, opened 2023-10-26
chore: enable community rekey loop

### LINKED ISSUE #11211: Messages are not received in the encrypted community

osmaczko:
rekey will happen every 2 days

### DIFF
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -37,7 +37,6 @@ import (
 	"github.com/status-im/status-go/protocol/communities"
 	"github.com/status-im/status-go/protocol/communities/token"
 	"github.com/status-im/status-go/protocol/discord"
-	"github.com/status-im/status-go/protocol/encryption"
 	"github.com/status-im/status-go/protocol/protobuf"
 	"github.com/status-im/status-go/protocol/requests"
 	"github.com/status-im/status-go/protocol/transport"
@@ -509,6 +508,10 @@ func (m *Messenger) updateCommunityActiveMembers(communityID string) error {
 
 func (m *Messenger) Communities() ([]*communities.Community, error) {
 	return m.communitiesManager.All()
+}
+
+func (m *Messenger) ControlledCommunities() ([]*communities.Community, error) {
+	return m.communitiesManager.Controlled()
 }
 
 func (m *Messenger) JoinedCommunities() ([]*communities.Community, error) {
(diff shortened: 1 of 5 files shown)


jrainville (approved):

protocol/messenger_communities.go
```diff
@@ -5787,44 +5765,38 @@ func (m *Messenger) startCommunityRekeyLoop() {
...
-	// Get and loop over all communities in persistence
-	cs, err := m.Communities()
+	controlledCommunities, err := m.ControlledCommunities()
 	if err != nil {
 		logger.Error("error getting communities", zap.Error(err))
 		return
 	}
-	for _, c := range cs {
+	for _, c := range controlledCommunities {
+		key, err := m.sender.GetCurrentKeyForGroup(c.ID())
```
cammellos: not sure if that's the best way, maybe is safer to just check if the community is Encrypted?
It should not probably happen, but say that for whatever reason a key is not generated, if we check `Encrypted` it would self heal, while if we check whether there's already a key present, this will not trigger.
But I saw that you removed the Encrypted check, maybe there's a reason why we would want to use getCurrent... (would it not though rekey also non encrypted communities, in case encryption is toggled off?)

osmaczko: That's a good point. Updated, please recheck, also added handling for channels re-keying.

cammellos (approved):

PR #4217, opened 2023-10-26
fix(community): getting kicked out of a community should still spectate

### LINKED ISSUE #12558: Chats are not loading if user was kicked and restarted the app
# Bug Report

## Description

Chats are not loading if the user was kicked and restarted the app

## Steps to reproduce

https://github.com/status-im/status-desktop/assets/117639195/4358472b-ee16-4326-8238-4d16dd355c43

Can be healed by reopening community invite

- Status desktop version: latest master
- Operating System: macOS

jrainville:
Fixes https://github.com/status-im/status-desktop/issues/12558

When getting kicked out of  a community, before we used to leave the community completely, but just keep the filters on. That created a problem when reopening the app, because the community disappeared and could even create a problem in desktop where it tried to open the last opened community but it's no longer there.

The fix now is that when getting kicked out, we instead just remove ourselves from the community and set Joined to false, but we keep the community spectated.

### DIFF
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -1699,7 +1699,7 @@ func (m *Messenger) LeaveCommunity(communityID types.HexBytes) (*MessengerRespon
 		return nil, err
 	}
 
-	mr, err := m.leaveCommunity(communityID, true)
+	mr, err := m.leaveCommunity(communityID)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


mprakhov (approved):

protocol/messenger_communities.go
```diff
@@ -5878,11 +5888,7 @@ func (m *Messenger) processCommunityChanges(messageState *ReceivedMessageState)
 			}
 
 		} else if changes.ShouldMemberLeave {
```
osmaczko: Would be great to rename it as well to make the intention a bit clearer.
```suggestion
		} else if changes.Kicked {
```

osmaczko (approved):

PR #4219, opened 2023-10-27
refactor: base community key actions evaluation on Encrypted() property

osmaczko:


protocol/communities/community.go
```diff
@@ -1547,6 +1547,10 @@ func (o *Community) HasTokenPermissions() bool {
 	return len(o.tokenPermissions()) > 0
 }
 
+func (o *Community) ChannelEncrypted(channelID string) bool {
+	return o.ChannelHasTokenPermissions(o.IDString() + channelID)
```
jrainville: Maybe we should create a function to get the channel ID. I think we have this concatenation in multiple places

jrainville (approved):
LGTM

protocol/communities/community.go
osmaczko: :+1: extracted

jrainville (approved):

Issue #4221, opened 2023-10-27
Detect ERC1155 transfers


IvanBelyakoff:
# Problem

Those are not detected. They are missing in activity list and balance history chart is not correct.

- [x] Detect TransferSingle events
- [x] Detect TransferBatch events

# Future Steps

Provide these transfers to client to display on Activity tab

PR #4222, opened 2023-10-27
feat(wallet): Added detection of ERC1155 SingleTransfer events

IvanBelyakoff:
Those events are stored in wallet DB as `erc1155` type.

Updates #4221

### DIFF
--- a/services/wallet/common/log_parser.go
+++ b/services/wallet/common/log_parser.go
@@ -20,20 +20,24 @@ type EventType string
 
 const (
 	// Transaction types
-	EthTransfer        Type = "eth"
-	Erc20Transfer      Type = "erc20"
-	Erc721Transfer     Type = "erc721"
-	UniswapV2Swap      Type = "uniswapV2Swap"
-	UniswapV3Swap      Type = "uniswapV3Swap"
-	HopBridgeFrom      Type = "HopBridgeFrom"
-	HopBridgeTo        Type = "HopBridgeTo"
-	unknownTransaction Type = "unknown"
+	EthTransfer           Type = "eth"
+	Erc20Transfer         Type = "erc20"
+	Erc721Transfer        Type = "erc721"
+	Erc1155SingleTransfer Type = "erc1155"
+	Erc1155BatchTransfer  Type = "erc1155"
+	UniswapV2Swap         Type = "uniswapV2Swap"
+	UniswapV3Swap         Type = "uniswapV3Swap"
+	HopBridgeFrom         Type = "HopBridgeFrom"
+	HopBridgeTo           Type = "HopBridgeTo"
+	unknownTransaction    Type = "unknown"
 
 	// Event types
 	WETHDepositEventType                      EventType = "wethDepositEvent"
 	WETHWithdrawalEventType                   EventType = "wethWithdrawalEvent"
 	Erc20TransferEventType                    EventType = "erc20Event"
 	Erc721TransferEventType                   EventType = "erc721Event"
+	Erc1155TransferSingleEventType            EventType = "erc1155SingleEvent"
+	Erc1155TransferBatchEventType             EventType = "erc1155BatchEvent"
 	UniswapV2SwapEventType                    EventType = "uniswapV2SwapEvent"
 	UniswapV3SwapEventType                    EventType = "uniswapV3SwapEvent"
 	HopBridgeTransferSentToL2EventType        EventType = "hopBridgeTransferSentToL2Event"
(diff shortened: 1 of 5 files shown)


services/wallet/common/log_parser.go
```diff
@@ -49,10 +53,13 @@ const (
 
 	// Transfer (index_topic_1 address from, index_topic_2 address to, uint256 value)
 	// Transfer (index_topic_1 address from, index_topic_2 address to, index_topic_3 uint256 tokenId)
-	Erc20_721TransferEventSignature = "Transfer(address,address,uint256)"
+	Erc20_721TransferEventSignature     = "Transfer(address,address,uint256)"
+	Erc1155TransferSingleEventSignature = "TransferSingle(address,address,address,uint256,uint256)"
```
dlipicar: Please add a comment explaining the format of these events, like the signature types have

services/wallet/common/log_parser.go
```diff
@@ -149,7 +166,7 @@ func GetFirstEvent(logs []*types.Log) (EventType, *types.Log) {
 
 func IsTokenTransfer(logs []*types.Log) bool {
 	eventType, _ := GetFirstEvent(logs)
-	return eventType == Erc20TransferEventType
+	return eventType == Erc20TransferEventType || eventType == Erc721TransferEventType || eventType == Erc1155TransferSingleEventType || eventType == Erc1155TransferBatchEventType
```
dlipicar: better use a switch statement

IvanBelyakoff: Thanx Dario. I thought I will have to write many `fallthrough` statements after each case, so I did not see how it is better, but thanx to copilot, it is easy now to teach an old dog some new tricks.

services/wallet/common/log_parser.go
```diff
@@ -20,20 +20,24 @@ type EventType string
...
-	UniswapV2Swap      Type = "uniswapV2Swap"
-	UniswapV3Swap      Type = "uniswapV3Swap"
-	HopBridgeFrom      Type = "HopBridgeFrom"
-	HopBridgeTo        Type = "HopBridgeTo"
-	unknownTransaction Type = "unknown"
+	EthTransfer           Type = "eth"
+	Erc20Transfer         Type = "erc20"
+	Erc721Transfer        Type = "erc721"
+	Erc1155SingleTransfer Type = "erc1155"
+	Erc1155BatchTransfer  Type = "erc1155"
```
stefandunca, line 27: I understand these identify transfer types in the `transactions` table. Using the exact string for single and batched will lose this information that we have at runtime. Is this intended? Won't it be helpful to propagate the runtime info to DB by using different identifiers (e.g. `erc1155single` and `erc1155batch`)?

services/wallet/transfer/commands.go
```diff
@@ -166,7 +167,7 @@ func (c *erc20HistoricalCommand) Run(ctx context.Context) (err error) {
 	for !c.iterator.Finished() {
 		headers, _, _, err := c.iterator.Next(ctx)
 		if err != nil {
-			log.Error("failed to get next batch", "error", err)
+			log.Error("failed to get next batch", "error", err, "chainID", c.chainClient.NetworkID()) // TODO: stop inifinite command in case of an error that we can't fix like missing trie node
```
stefandunca, line 170: How about connecting this TODO with a GitHub issue to increase awareness? Similar to [this](https://github.com/status-im/status-desktop/blob/ccd8c5b65f973718aa935d37736edc90b763957f/ui/app/AppLayouts/Wallet/views/walletconnect/WalletConnectSDK.qml#L3). Otherwise all these todos might be fade out in the big codebase.

stefandunca (approved):
Nice detector improvements. Please see my inline questions.

services/wallet/common/log_parser.go
IvanBelyakoff: Good question, @stefandunca. I was thinking about the same, but was not sure if we will need it to distinguish between those at this level. It might depend on how we want to implement those and design UI as well. Basically from what I see, Batch transfers will just have several ids and corresponding values. I am not sure if we need to keep those as a single transfer in DB or as multiple transfers with the same multitransaction id, since we already have it. So I left it for now for the next PR for Batch transfers that I am working on. But we might want to think about it now to avoid fixing already fetched types in DB. Do you have any ideas?

services/wallet/transfer/commands.go
IvanBelyakoff: Stefan I already have an [issue](https://github.com/status-im/status-desktop/issues/11074) for that. Just have not started it yet.

rasom (approved):

PR #4223, opened 2023-10-27
fix order of drop statement

### LINKED ISSUE #12566: Failures during login
# Bug Report

## Description

Currently getting a variety of errors when attempting to log in with an existing account. Using a fresh Status db does work. Troubleshooting done so far suggests it may be related to this change: https://github.com/status-im/status-go/pull/4182/files#diff-47ca94cb01ad5f1c83b007b21d3ade014ad6cf2d2ace4c26800e2015f94c60e4

image
image
image
image
(password incorrect showing even when password is typed correctly)

### Additional Information

- Status desktop version: master (found in https://github.com/status-im/status-desktop/pull/12458 after rebase)
- Operating System: Mac

cc: @borismelnik

cammellos:
a foreign key constraint was impacted when dropping the table, apologies,, I think i tested by removing the data first.

This should be a safe change, users that already migrated, will not run it, users who got the error, will run it.


osmaczko (approved):

jrainville (approved):
I tested and it works

PR #4224, opened 2023-10-27
fix:  Update timestamp when upserting multi tx

### LINKED ISSUE #12500: Dates of transactions are incorrect
# Bug Report

Opened activity tab for account 0x333c5dba8179822056f2289bdede1b53a863f577 (NOT a test network)
While transactions are being loaded, "New transactions" button appears and on clicking it, the transactions appear, but their dates are wrong - seem to be the dates they were discovered by Status app.
Have a look at Optimism scanner for this address: https://optimistic.etherscan.io/address/0x333c5dba8179822056f2289bdede1b53a863f577

There are no transactions dated 20/10/2023
image

- Status desktop version: https://ci.infra.status.im/job/status-desktop/job/master/3543/
- Operating System: MacOS 13.5

Cuteivist:
Closes https://github.com/status-im/status-desktop/issues/12500

### DIFF
--- a/services/wallet/transfer/transaction.go
+++ b/services/wallet/transfer/transaction.go
@@ -188,6 +188,13 @@ func rowsToMultiTransactions(rows *sql.Rows) ([]*MultiTransaction, error) {
 	return multiTransactions, nil
 }
 
+func getMultiTransactionTimestamp(multiTransaction *MultiTransaction) uint64 {
+	if multiTransaction.Timestamp != 0 {
+		return multiTransaction.Timestamp
+	}
+	return uint64(time.Now().Unix())
+}
+
 // insertMultiTransaction inserts a multi transaction into the database and updates multi-transaction ID and timestamp
 func insertMultiTransaction(db *sql.DB, multiTransaction *MultiTransaction) (MultiTransactionIDType, error) {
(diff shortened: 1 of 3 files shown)


services/wallet/transfer/transaction.go
```diff
@@ -188,15 +188,21 @@ func rowsToMultiTransactions(rows *sql.Rows) ([]*MultiTransaction, error) {
 	return multiTransactions, nil
 }
 
+func getMultiTransactionTimestamp(multiTransaction *MultiTransaction) uint64 {
+	if multiTransaction.Timestamp != 0 {
+		return multiTransaction.Timestamp
+	}
+	return uint64(time.Now().Unix())
```
dlipicar, line 195: is there an expected, valid situation where the Tx shouldn't have a timestamp and we should return Now() instead? If not, perhaps add a log here.

dlipicar (approved):

services/wallet/transfer/transaction.go
Cuteivist: Idea here is to check if Timestamp was set. In case it was not, then return current time, like before. 

We don't want the case where the timestamp is 0, because it will point to year 1970. Blockchain didn't exist yet :D

PR #4226, opened 2023-10-30
refactor: remove always nil `shard` parameter in `HandleCommunityDesc…

osmaczko:
…riptionMessage`


cammellos (commented):
looks ok to me, not sure if at some point we will need the parameter @richard-ramos might want to have a look

richard-ramos:
I think it's okay to delete it for now!

richard-ramos (approved):

jrainville (approved):

PR #4229, opened 2023-10-30
feat: Send envelopes to telemetry service

### LINKED ISSUE #12430: [Telemetry] Improve Telemetry server code to accept envelopes
### Description

Part of https://github.com/status-im/status-desktop/issues/12429

Improve the telemetry server code [here](https://github.com/status-im/telemetry) so that it can accept receiving envelopes.

Right now, it is made to receive only messages that have been "fully received", meaning that they fit the topics the client had and they had the key to decrypt the message. That means that we could have received the message from Waku, but it wouldn't make it to the Telemetry server because of one of the above steps failed.

With this improvement, we'd have a better idea of where the problem in the chain of communication the unreliability is, if there is one.

### Acceptance Criteria

- The Telemetry service accepts receiving envelopes. Needed properties are
  - Envelope ID
  - topic 
  - if possible the state of the message (processed, ignored) and even better of we can know why it was ignored
  - the timestamp of when it was received (this will be used to know if it took a long time to receive the message)
  - (cc @richard-ramos )
- The Message objects that were previously already accepted now also contain the envelopeID so that we can match the envelopes to the messages later

### LINKED ISSUE #12431: [Telemetry] Send envelopes received by status-go to the Telemetry server
### Description

Part of https://github.com/status-im/status-desktop/issues/12429

Follow-up to https://github.com/status-im/status-desktop/issues/12430

In the status-go code [here](https://github.com/status-im/status-go/blob/54928bd7f503b7b3c18eaaeb2bd6758f0617bafa/protocol/messenger.go#L3585-L3587), we send the messages to the telemetry server when they are retrieved.

We need to do the same for envelopes. This will help track message reliability and see if we receives envelopes but just don't process them.

### Acceptance Criteria

- The status-go code sends the info about the envelopes received to the telemetry server.
  - Bonus point if we also send if we processed the envelope
  - 2x bonus points if we add a reason (can be an enum) like "no subscribed topic" or "no encryption key"

endulab:
1. Push envelopes to telemetry service
2. Update envelopes which were filtered but some error happened.

Related telemetry repo: https://github.com/status-im/telemetry/pull/12/files

Issue [#12430](https://github.com/status-im/status-desktop/issues/12430)

### DIFF
--- a/api/geth_backend.go
+++ b/api/geth_backend.go
@@ -2299,7 +2299,7 @@ func (b *GethStatusBackend) injectAccountsIntoWakuService(w types.WakuKeyManager
 	}
 
 	if st != nil {
-		if err := st.InitProtocol(b.statusNode.GethNode().Config().Name, identity, b.appDB, b.walletDB, b.statusNode.HTTPServer(), b.multiaccountsDB, acc, b.accountManager, b.statusNode.RPCClient(), b.statusNode.WalletService(), b.statusNode.CommunityTokensService(), logutils.ZapLogger()); err != nil {
+		if err := st.InitProtocol(b.statusNode.GethNode().Config().Name, identity, b.appDB, b.walletDB, b.statusNode.HTTPServer(), b.multiaccountsDB, acc, b.accountManager, b.statusNode.RPCClient(), b.statusNode.WalletService(), b.statusNode.CommunityTokensService(), b.statusNode.WakuV2Service(), logutils.ZapLogger()); err != nil {
 			return err
 		}
 		// Set initial connection state
--- a/protocol/messenger.go
+++ b/protocol/messenger.go
@@ -412,6 +412,9 @@ func NewMessenger(
 	var telemetryClient *telemetry.Client
 	if c.telemetryServerURL != "" {
 		telemetryClient = telemetry.NewClient(logger, c.telemetryServerURL, c.account.KeyUID, nodeName)
+		if c.wakuService != nil {
+			c.wakuService.SetStatusTelemetryClient(telemetryClient)
+		}
 	}
 
 	// Initialize push notification server
@@ -3625,6 +3628,9 @@ func (m *Messenger) handleRetrievedMessages(chatWithMessages map[transport.Filte
 
 			statusMessages, acks, err := m.sender.HandleMessages(shhMessage)
 			if err != nil {
+				if m.telemetryClient != nil {
+					go m.telemetryClient.UpdateEnvelopeProcessingError(shhMessage, err)
+				}
 				logger.Info("failed to decode messages", zap.Error(err))
 				continue
 			}
--- a/services/ext/service.go
+++ b/services/ext/service.go
@@ -55,6 +55,7 @@ import (
 	"github.com/status-im/status-go/services/wallet"
 	w_common "github.com/status-im/status-go/services/wallet/common"
 	"github.com/status-im/status-go/services/wallet/thirdparty"
+	"github.com/status-im/status-go/wakuv2"
 )
 
 const infinityString = "∞"
@@ -121,7 +122,7 @@ func (s *Service) GetPeer(rawURL string) (*enode.Node, error) {
 	return enode.ParseV4(rawURL)
 }
 
-func (s *Service) InitProtocol(nodeName string, identity *ecdsa.PrivateKey, appDb, walletDb *sql.DB, httpServer *server.MediaServer, multiAccountDb *multiaccounts.Database, acc *multiaccounts.Account, accountManager *account.GethManager, rpcClient *rpc.Client, walletService *wallet.Service, communityTokensService *communitytokens.Service, logger *zap.Logger) error {
+func (s *Service) InitProtocol(nodeName string, identity *ecdsa.PrivateKey, appDb, walletDb *sql.DB, httpServer *server.MediaServer, multiAccountDb *multiaccounts.Database, acc *multiaccounts.Account, accountManager *account.GethManager, rpcClient *rpc.Client, walletService *wallet.Service, communityTokensService *communitytokens.Service, wakuService *wakuv2.Waku, logger *zap.Logger) error {
 	var err error
 	if !s.config.ShhextConfig.PFSEnabled {
 		return nil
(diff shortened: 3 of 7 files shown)


api/geth_backend.go
```diff
@@ -2295,7 +2295,7 @@ func (b *GethStatusBackend) injectAccountsIntoWakuService(w types.WakuKeyManager
 	}
 
 	if st != nil {
-		if err := st.InitProtocol(b.statusNode.GethNode().Config().Name, identity, b.appDB, b.walletDB, b.statusNode.HTTPServer(), b.multiaccountsDB, acc, b.accountManager, b.statusNode.RPCClient(), b.statusNode.WalletService(), b.statusNode.CommunityTokensService(), logutils.ZapLogger()); err != nil {
+		if err := st.InitProtocol(b.statusNode.GethNode().Config().Name, identity, b.appDB, b.walletDB, b.statusNode.HTTPServer(), b.multiaccountsDB, acc, b.accountManager, b.statusNode.RPCClient(), b.statusNode.WalletService(), b.statusNode.CommunityTokensService(), b.statusNode.WakuV2Service(), logutils.ZapLogger()); err != nil {
```
endulab, line 2302: I am not sure if I reinvented the wheel here. I had to set in waku service the same telemetry client which was created in messenger, because I want to send data with the same receiverKeyUID and nodeName. Maybe there is a simpler way to access waku service from messenger.

protocol/messenger.go
```diff
@@ -3604,6 +3607,9 @@ func (m *Messenger) handleRetrievedMessages(chatWithMessages map[transport.Filte
 			}
 
 			statusMessages, acks, err := m.sender.HandleMessages(shhMessage)
+			if m.telemetryClient != nil && err != nil {
+				go m.telemetryClient.UpdateEnvelopeProcessingError(shhMessage, err)
```
endulab: Here, in case of error, I send envelope data and processing error to the service.

protocol/messenger.go
```diff
@@ -410,6 +410,9 @@ func NewMessenger(
 	var telemetryClient *telemetry.Client
 	if c.telemetryServerURL != "" {
 		telemetryClient = telemetry.NewClient(logger, c.telemetryServerURL, c.account.KeyUID, nodeName)
+		if c.wakuService != nil {
+			c.wakuService.SetStatusTelemetryClient(telemetryClient)
```
endulab, line 416: Here I set telemetryClient is waku service.

protocol/messenger.go
```diff
@@ -3604,6 +3607,9 @@ func (m *Messenger) handleRetrievedMessages(chatWithMessages map[transport.Filte
 			}
 
 			statusMessages, acks, err := m.sender.HandleMessages(shhMessage)
+			if m.telemetryClient != nil && err != nil {
```
jrainville: This condition can be put in side the `if err != nil {` below

telemetry/client.go
```diff
@@ -56,3 +57,44 @@ func (c *Client) PushReceivedMessages(filter transport.Filter, sshMessage *types
 		c.logger.Error("Error sending message to telemetry server", zap.Error(err))
 	}
 }
+
+func (c *Client) PushReceivedEnvelope(envelope *protocol.Envelope) {
+	c.logger.Debug("Pushing received envelope to telemetry server")
+	url := fmt.Sprintf("%s/received-envelope", c.serverURL)
+	postBody := map[string]interface{}{
+		"messageHash":    types.EncodeHex(envelope.Hash()),
```
jrainville, line 65: should this be `envelopeHash` or you're reusing a similar API to the message one?

telemetry/client.go
```diff
@@ -56,3 +57,44 @@ func (c *Client) PushReceivedMessages(filter transport.Filter, sshMessage *types
...
+		"sentAt":         uint32(envelope.Message().Timestamp / int64(time.Second)),
```
jrainville, line 66: I think it would be good to also have `receivedAt` to compare how long it took for it to be received

jrainville (commented):
Looks good. I don't know about your comment about passing the telemetry client, @richard-ramos probably knows better.

Otherwise, just comments about stuff that can be added.

telemetry/client.go
```diff
@@ -56,3 +57,44 @@ func (c *Client) PushReceivedMessages(filter transport.Filter, sshMessage *types
...
+	}
+	body, _ := json.Marshal(postBody)
+	_, err := c.httpClient.Post(url, "application/json", bytes.NewBuffer(body))
+	if err != nil {
+		c.logger.Error("Error sending envelope to telemetry server", zap.Error(err))
+	}
+}
+
+func (c *Client) UpdateEnvelopeProcessingError(shhMessage *types.Message, processingError error) {
+	c.logger.Debug("Pushing envelope update to telemetry server")
```
richard-ramos: I'd add the envelope hash to the log

services/ext/service.go
```diff
@@ -442,6 +444,7 @@ func buildMessengerOptions(
 		protocol.WithWalletConfig(&config.WalletConfig),
 		protocol.WithWalletService(walletService),
 		protocol.WithCommunityTokensService(communityTokensService),
+		protocol.WithWakuService(wakuService),
```
richard-ramos, line 447: Probably non an issue since we only use WakuV2, but i think this can be nil if using WakuV1

telemetry/client.go
```diff
@@ -56,3 +57,44 @@ func (c *Client) PushReceivedMessages(filter transport.Filter, sshMessage *types
 		c.logger.Error("Error sending message to telemetry server", zap.Error(err))
 	}
 }
+
+func (c *Client) PushReceivedEnvelope(envelope *protocol.Envelope) {
+	c.logger.Debug("Pushing received envelope to telemetry server")
```
richard-ramos: Might be a good idea to log the hash here

richard-ramos (approved):

endulab:
@jrainville I added related pr in telemetry repo: https://github.com/status-im/telemetry/pull/12/files

telemetry/client.go
endulab: Hmm, I called it messageHash also in database. Technically envelope's hash IS message's hash.
So I am not sure, if you think that envelopeHash would be better, then I can change it.

telemetry/client.go
endulab: In case of messages, there is a createdAt field which is set by the service when the message is received. I did the same for envelopes:  https://github.com/status-im/telemetry/pull/12/files#diff-18714ff6aeb7898a2395af4afe27885d30c8c23e7855823d2cff7d9bb8b5d039R21

The timestamp is created when the service receives information about a new envelope.

jrainville (approved):
Nice job

PR #4233, opened 2023-10-31
fix crash on x86_64 android emulators

siddarthkay:
Due to presence of syscall to `epoll_wait` android emulators after Android 10 would crash.
This PR has added amd64 fallback implementations in 2 spots

- tcp-shaker ref : https://github.com/status-im/tcp-shaker/commit/571aa46f35a5bb72b8bc8306fad2f87bb75def12
- notify ref : https://github.com/status-im/notify/commit/0eb36e6c1a02a5d8dd3c663a04dcb46ddfb49593

The build flags are updated for fallback implementation to only be used in case of x86_64 simulators.

status-mobile PR : https://github.com/status-im/status-mobile/pull/17773


vitvly (approved):
Probably should be also tested in a related status-mobile PR.

Samyoul (approved):
🧙 🪄 👻

PR #4234, opened 2023-10-31
[#4087] Omit history scanning on generated multiacc creation on deskt…

rasom:
Proposed by @saledjenic instead of making on client side.

status: ready

### DIFF
--- a/mobile/status.go
+++ b/mobile/status.go
@@ -352,7 +352,7 @@ func SaveAccountAndLogin(accountData, password, settingsJSON, configJSON, subacc
 		return makeJSONResponse(err)
 	}
 
-	if settings.Mnemonic != "" {
+	if *settings.Mnemonic != "" {
 		settings.OmitTransfersHistoryScan = true
 	}
 
--- a/mobile/status.go
+++ b/mobile/status.go
@@ -351,6 +351,11 @@ func SaveAccountAndLogin(accountData, password, settingsJSON, configJSON, subacc
 	if err != nil {
 		return makeJSONResponse(err)
 	}
+
+	if settings.Mnemonic != "" {
+		settings.OmitTransfersHistoryScan = true
+	}
+
 	var conf params.NodeConfig
 	err = json.Unmarshal([]byte(configJSON), &conf)
 	if err != nil {


mobile/status.go
```diff
@@ -351,6 +351,11 @@ func SaveAccountAndLogin(accountData, password, settingsJSON, configJSON, subacc
 	if err != nil {
 		return makeJSONResponse(err)
 	}
+
+	if *settings.Mnemonic != "" {
+		settings.OmitTransfersHistoryScan = true
```
saledjenic, line 356: @rasom could you point me to the piece of code where we're doing the history scan if new accounts are added on a later date, after the user backed up his mnemonic?

saledjenic: This change is ok, and that's the initial state when a user is creating a regular (non-keycard) Status account.
But then the user can back up the mnemonic, use the same seed phrase in other tools, then get back to Status and add the same account he used elsewhere. Do we handle that case?

rasom: We don't want to scan history right after account creation. It doesn't matter what happens with mnemonic afterwards in terms of this change.

saledjenic: Aha... so it's only for the default wallet account?

saledjenic: But anyway, I don't see code that avoids history scans based on `OmitTransfersHistoryScan` (not in this PR, but in `status-go` repo), do we have it or not yet?

saledjenic (approved):

mobile/status.go
saledjenic: Ok, I see this https://github.com/status-im/status-go/pull/4163/files#diff-856ac71f99a073eb293c1213dc5ad2e9e620b0174fdc369cd39951d50a5117b4R106-R112 
thanks for sharing that

qfrank:
Looks like mobile won't call `SaveAccountAndLogin` except `create-keycard-multiaccount` since mobile use new API `CreateAccountAndLogin` now, if you want to make this PR effect with mobile as well , proberbly additional changes are needed 🙂 @rasom

rasom:
@qfrank it is covered for mobile in previous PR

qfrank (approved):
Oh, i see, there you go!

PR #4238, opened 2023-11-01
feat: new configurable endpoint RequestCommunityInfoFromMailserverV2

igor-sirotin:
1. Add new API endpoint `RequestCommunityInfoFromMailserverV2`
It does nothing new, but has a single general `CommunityInfoRequest` argument, which:
  - adds ability to request community info with `useDatabase = false`
  - makes it simpler to extend request with more parameters if needed

2. Deprecate old `RequestCommunityInfo...` endpoints.
Currently they wrap the new `RequestCommunityInfoFromMailserverV2`.

3. Fixed `StatusUnfurler` to use `shard` property from URL data

4. Some logs added

### DIFF
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -68,10 +68,33 @@ const (
 )
 
 type FetchCommunityRequest struct {
-	CommunityID     string        `json:"communityId"`
+	// CommunityKey should be either a public or a private community key
+	CommunityKey    string        `json:"communityKey"`
 	Shard           *common.Shard `json:"shard"`
-	TryDatabase     bool          `json:"useDatabase"`
+	TryDatabase     bool          `json:"tryDatabase"`
 	WaitForResponse bool          `json:"waitForResponse"`
+}
+
+func (r *FetchCommunityRequest) Validate() error {
+	if len(r.CommunityKey) <= 2 {
+		return fmt.Errorf("community key is too short")
+	}
+	return nil
+}
+
+func (r *FetchCommunityRequest) getCommunityID() string {
+	return GetCommunityIDFromKey(r.CommunityKey)
+}
+
+func GetCommunityIDFromKey(communityKey string) string {
+	// Check if the key is a private key. strip the 0x at the start
+	if privateKey, err := crypto.HexToECDSA(communityKey[2:]); err == nil {
+		// It is a privateKey
+		return types.HexBytes(crypto.CompressPubkey(&privateKey.PublicKey)).String()
+	}
+
+	// Not a private key, use the public key
+	return communityKey
 }
 
 func (m *Messenger) publishOrg(org *communities.Community) error {
@@ -2336,7 +2359,7 @@ func (m *Messenger) ImportCommunity(ctx context.Context, key *ecdsa.PrivateKey) 
 	}
 
 	_, err = m.FetchCommunity(&FetchCommunityRequest{
-		CommunityID:     community.IDString(),
+		CommunityKey:    community.IDString(),
 		Shard:           community.Shard(),
 		TryDatabase:     false,
 		WaitForResponse: true,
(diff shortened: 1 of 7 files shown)


services/ext/api.go
```diff
@@ -1183,22 +1183,58 @@ func (api *PublicAPI) EnsVerified(pk, ensName string) error {
...
+		CommunityID:     communityID,
+		Shard:           shard,
+		UseDatabase:     true,
+		WaitForResponse: false,
+	}
+	_, err := api.RequestCommunityInfoFromMailserverV2(request)
+	return err
+}
+
+func (api *PublicAPI) RequestCommunityInfoFromMailserverV2(request *protocol.CommunityInfoRequest) (*communities.Community, error) {
```
osmaczko: Given that the request may not reach the mailserver at all, would it make sense to rename it as follows?
```suggestion
func (api *PublicAPI) FetchCommunity(request *protocol.FetchCommunityOpts) (*communities.Community, error) {
```

protocol/messenger_communities.go
```diff
@@ -2613,38 +2625,29 @@ func (m *Messenger) GetCommunityIDFromKey(communityKey string) string {
...
 
 // RequestCommunityInfoFromMailserver installs filter for community and requests its details
-// from mailserver. It waits until it has the community before returning it.
-// If useDatabase is true, it searches for community in database and does not request mailserver.
-func (m *Messenger) RequestCommunityInfoFromMailserver(privateOrPublicKey string, shard *common.Shard, useDatabase bool) (*communities.Community, error) {
-	communityID := m.GetCommunityIDFromKey(privateOrPublicKey)
-	if useDatabase {
+// from mailserver.
+//
+// If `request.UseDatabase` is true, it searches for community in database and does not request mailserver.
```
osmaczko: Wouldn't it request mailserver anyway if community is not in db?

protocol/messenger_communities.go
```diff
@@ -67,6 +67,13 @@ const (
 	maxChunkSizeBytes    = 1500000
 )
 
+type CommunityInfoRequest struct {
+	CommunityID     string        `json:"communityId"`
+	Shard           *common.Shard `json:"shard"`
+	UseDatabase     bool          `json:"useDatabase"`
```
```suggestion
	TryDatabase     bool          `json:"tryDatabase"`
```

It will make the intention clearer imo. It conveys the behavior of first consulting the local database before potentially reaching out to the mailserver.

osmaczko (commented):
Very cool improvement! Please consider my suggestions.

services/ext/api.go
igor-sirotin: Good idea! 
I'll keep stick to `FetchCommunityRequest` instead of `...Opts`, we use Request/Response for such thing in other places.

protocol/messenger_communities.go
igor-sirotin: Indeed. Seems that the comment was previously wrong all the time 😄

osmaczko (approved):

jrainville (approved):
Nice, much cleaner

protocol/messenger_communities.go
```diff
@@ -2612,42 +2624,34 @@ func (m *Messenger) GetCommunityIDFromKey(communityKey string) string {
...
-	if useDatabase {
+// FetchCommunity installs filter for community and requests its details
+// from mailserver.
+//
+// If `request.TryDatabase` is true, it first looks for community in database,
+// and requests from mailserver only if it wasn't found locally.
+// If `request.WaitForResponse` is true, it waits until it has the community before returning it.
+// If `request.WaitForResponse` is false, it installs filter for community and requests its details
+// from mailserver. When response received it will be passed through signals handler.
+func (m *Messenger) FetchCommunity(request *FetchCommunityRequest) (*communities.Community, error) {
```
jrainville, line 2644: I think we could validate the request here before doing operations on it, since it comes from the user.
Make sure the communityID is not nul, if there is a chard, that the values are ok, stuff like that.

richard-ramos:
It would be a good idea to open an issue in status-mobile in case they use `RequestCommunityInfoFromMailserver`

richard-ramos (approved):

PR #4240, opened 2023-11-01
feat(wallet) implement WalletConnect pair API

stefandunca:
### Updates status-desktop [#12551](https://github.com/status-im/status-desktop/issues/12551)

status-desktop PR [#12604](https://github.com/status-im/status-desktop/pull/12604)

### DIFF
--- a/services/wallet/api.go
+++ b/services/wallet/api.go
@@ -26,6 +26,7 @@ import (
 	"github.com/status-im/status-go/services/wallet/thirdparty"
 	"github.com/status-im/status-go/services/wallet/token"
 	"github.com/status-im/status-go/services/wallet/transfer"
+	wc "github.com/status-im/status-go/services/wallet/walletconnect"
 	"github.com/status-im/status-go/services/wallet/walletevent"
 	"github.com/status-im/status-go/transactions"
 )
@@ -626,7 +627,7 @@ func (api *API) GetActivityCollectiblesAsync(requestID int32, chainIDs []wcommon
 }
 
 func (api *API) FetchChainIDForURL(ctx context.Context, rpcURL string) (*big.Int, error) {
-	log.Debug("wallet.api.VerifyURL", rpcURL)
+	log.Debug("wallet.api.VerifyURL", "rpcURL", rpcURL)
 
 	rpcClient, err := gethrpc.Dial(rpcURL)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


services/wallet/api.go
```diff
@@ -635,3 +636,15 @@ func (api *API) FetchChainIDForURL(ctx context.Context, rpcURL string) (*big.Int
 	client := ethclient.NewClient(rpcClient)
 	return client.ChainID(ctx)
 }
+
+func (api *API) WCPairSessionProposal(ctx context.Context, sessionProposalJSON string) (*wc.PairSessionResponse, error) {
```
dlipicar, line 640: hmm i've never seen us pass arguments from the client explicitly as a JSON string, any reason why we do the unmarshaling manually here?

services/wallet/walletconnect/service.go
```diff
@@ -0,0 +1,53 @@
...
+		return s.networkManager.Find(chainID) != nil
+	})
+	if len(chains) != len(proposedChains) {
+		return nil, ErrorChainsNotSupported
+	}
+	namespace.Chains = eipChains
+
+	activeAccounts, err := s.accountsDB.GetActiveAccounts()
+	if err != nil {
+		return nil, ErrorChainsNotSupported
```
dlipicar, line 40: I don't imagine we'll get this error often, but we could generate a new one instead of `ErrorChainsNotSupported` since it's not really related

services/wallet/api.go
stefandunca: They come as `JSON` from the `WebView` through the `WalletConnect SDK`. It seemed pointless to parse it in Nim to forward them to status-go. Is there any value in having the parsing and logic in Nim?

services/wallet/walletconnect/service.go
stefandunca: Good point. I already fixed it in the following PR: https://github.com/status-im/status-go/blob/5ab9e8e4e7daa5f45a9f5db57a3c9e42889af04b/services/wallet/walletconnect/service.go#L52

services/wallet/api.go
dlipicar: Ohh alright, have yet to check the status-desktop one. I guess it makes sense if it's data we're passing around and not building ourselves

dlipicar (approved):

PR #4241, opened 2023-11-01
remove sync AC state

qfrank:
No need to sync AC state now since it doesn't make sense and we allow exist different number of ACs between paired devices.
Also we will update AC state for each AC notification update(including sync AC decision) now.


MishkaRogachev (approved):
That's a great PR! I love it when things get simpler

PR #4242, opened 2023-11-01
chore: upgrade to go 1.20

richard-ramos:
go-libp2p's latest version requires go 1.20 as minimum. Since I'm updating go-waku's dependencies, status-go will require also to run with 1.20


richard-ramos:
Looks like go 1.20 and gomobile don't play well, but we can't upgrade to go1.21 yet because of go-libp2p: 
```
# github.com/quic-go/quic-go/internal/qtls
../../go/pkg/mod/github.com/quic-go/quic-go@v0.36.4/internal/qtls/go121.go:5:13: cannot use "The version of quic-go you're using can't be built on Go 1.21 yet. For more details, please see https://github.com/quic-go/quic-go/wiki/quic-go-and-Go-versions." (untyped string constant "The version of quic-go you're using can't be built on Go 1.21 yet. F...) as int value in variable declaration
```
I'll create a branch in go-waku that uses go1.19 in the meantime and use that in status-go

Closing this PR

PR #4244, opened 2023-11-01
Increased timeouts when fetching curated communities

### LINKED ISSUE #4201: Possible `eth_CallContract` leakage
according to https://github.com/status-im/status-go/pull/4197 currently we have ~450 `eth_CallContract` calls per hour when idle, that alone costs over 10k requests per day.

Should be identified at first.

vkjr:
Fixed timeouts for fetching curated communities:
All timeouts increased and we don't fetch communities often in case of error or unknown communities.

New timeouts:
- if fetching succeeded: 300 seconds
- if fetching results contain unresolved communities first time: 60 seconds
- if fetching results contain unresolved communities two times in a row: 300 seconds
- if fetching failed: 30 seconds
- if fetching failed two times in a row: 300 seconds

Closes #4201

### DIFF
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -565,8 +565,8 @@ func (m *Messenger) SpectatedCommunities() ([]*communities.Community, error) {
 
 const (
 	fetchError       int = 0
-	fetchSuccess         = 1
-	fetchHasUnknowns     = 2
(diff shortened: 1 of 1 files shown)


rasom (approved):

endulab (approved):
LGTM, though I did not test.

PR #4246, opened 2023-11-01
fix(wakuv2): post envelopes previously cached but not processed

igor-sirotin:
In relay mode we cache envelopes, even if there're no filters found for them.
This affects later mailserver requests, because the envelopes that are found in cache were considered as "processed", even when there were not.

Now we actually save the `Processed` state of the envelopes in the cache.

### DIFF
--- a/wakuv2/common/message.go
+++ b/wakuv2/common/message.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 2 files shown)


vitvly (approved):
Missed the `logger.With` part, therefore removed earlier comments. Looks good!

richard-ramos (approved):
LGTM, 
Interesting usage of atomic booleans btw

jrainville (approved):

PR #4248, opened 2023-11-02
fix: Correct skip check statement

Cuteivist:
* `getWalletToken` API call didn't have `communityID` field populated
* discovering community ID didn't work because of wrong initial check statement


stefandunca (approved):
👍🏼

PR #4249, opened 2023-11-02
feat: Return only community tokens

Cuteivist:
Only return community tokens from custom tokens

image


alaibe (approved):
To confirm, this is not used to check if the token exists?

stefandunca (approved):
🍯

Issue #4088, opened 2023-10-03
Flaky test - TestMessengerCommunitiesSuite/TestDeletePendingRequestAccess


cammellos:
```
 communities_messenger_test.go:985: 
        	Error Trace:	/home/jenkins/workspace/status-go_prs_tests_PR-4079/protocol/communities_messenger_test.go:985
        	Error:      	"[%!s(*communities.RequestToJoin=&{[210 161 130 16 185 44 12 55 4 157 146 15 81 196 186 58 25 201 199 62 71 125 224 155 218 211 146 208 131 163 34 60] 0x04c45e6b110fff7fc0708f86ddf8d4806c5ac014cd24e917342c7152a4921777ffdfb9d81adc849545ea7efd52d8dd312f64be87bc44eeddb9a6ed86d25281b7dc 1696275279   [2 237 26 193 85 172 108 13 21 27 61 86 169 17 200 232 75 30 166 9 79 69 31 34 212 213 148 191 233 175 136 179 139] 1 false false []}) %!s(*communities.RequestToJoin=&{[210 161 130 16 185 44 12 55 4 157 146 15 81 196 186 58 25 201 199 62 71 125 224 155 218 211 146 208 131 163 34 60] 0x04c45e6b110fff7fc0708f86ddf8d4806c5ac014cd24e917342c7152a4921777ffdfb9d81adc849545ea7efd52d8dd312f64be87bc44eeddb9a6ed86d25281b7dc 1696275280   [2 237 26 193 85 172 108 13 21 27 61 86 169 17 200 232 75 30 166 9 79 69 31 34 212 213 148 191 233 175 136 179 139] 1 false false []})]" should have 1 item(s), but has 2
```

PR #4250, opened 2023-11-02
[Fixes: #4088] Fix pending notification test

cammellos:
There were 2 issues:

1) We hard delete requests, that means that on retransmission they will be recreated, the test has been changed to accommodate this behavior 

2) We always used time.now when updating timestamp in notification,
   sometimes time is the same so the notification is not updated, we
   changed to use what essentially is a clock value

### DIFF
--- a/protocol/activity_center.go
+++ b/protocol/activity_center.go
@@ -76,6 +76,16 @@ type ActivityCenterNotification struct {
 	AlbumMessages []*common.Message `json:"albumMessages"`
 }
 
+func (n *ActivityCenterNotification) IncrementUpdatedAt(timesource common.TimeSource) {
+	tNow := timesource.GetCurrentTime()
+	// If updatead at is greater or equal than time now, we bump it
+	if n.UpdatedAt >= tNow {
+		n.UpdatedAt++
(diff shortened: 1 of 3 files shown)


rasom (approved):

osmaczko (approved):

qfrank (commented):
Hi @cammellos , thank you for your fix! Apologize for brought the issue relate to `updated_at` , i was wondering if we still need this field as update notification condition, because we have changed mechanism for sync AC notification(only decision) rather than sync whole db record. There are 5 places (e.g. [1](https://github.com/status-im/status-go/blob/d3558d8e09e16199e5dc7758819298da36b7009e/protocol/activity_center_persistence.go#L1030) [2](https://github.com/status-im/status-go/blob/d3558d8e09e16199e5dc7758819298da36b7009e/protocol/activity_center_persistence.go#L147)) use it as update condition in `activity_center_persistence.go` , we can search with keywords `updated_at >` and `updated_at <`, after removing such update condition, we don't need `IncrementUpdatedAt` anymore, we can just leave it as normal recording field. WDYT?

PR #4253, opened 2023-11-02
fix: estimate gas function resoved that it uses real network estimation instead of hardcoded value

saledjenic:


stefandunca (approved):
Excellent job; I'm looking forward to seeing it in action.

saledjenic:
@all could you take another look into this (second commit) it fixes additional issues I found while I was working on this.

alaibe (approved):

PR #4255, opened 2023-11-02
fix(build): avoid removing untracked git changes by `make clean`

IvanBelyakoff:
Use `make deep-clean` for deep cleaning

Me and some other devs occasionally fall in trouble by using `make clean` command which imo unexpectedly removes untracked files, including not yet commited fresh sources. Better use `deep-clean` command for that.


jakubgs (approved):
Good idea.

dlipicar (approved):

PR #4260, opened 2023-11-03
fix: decompress any public keys in link previews

igor-sirotin:
Required for https://github.com/status-im/status-desktop/pull/12613.

it's good to transfer the keys compressed, but the app mainly uses the keys uncompressed.
At first I decompressed them on the client side, but that required a status-go call per each decompression, which is bad.

### DIFF
--- a/protocol/common/message_linkpreview.go
+++ b/protocol/common/message_linkpreview.go
@@ -3,7 +3,9 @@ package common
 import (
 	"fmt"
 	"net/url"
+	"strings"
 
+	"github.com/status-im/status-go/api/multiformat"
 	"github.com/status-im/status-go/images"
 	"github.com/status-im/status-go/protocol/protobuf"
 )
@@ -467,4 +469,27 @@ func (m *Message) ConvertFromProtoToStatusLinkPreviews(makeMediaServerURL func(m
 	}
 
 	return previews
+}
+
+func decompressPublicKey(input string) string {
+	decompressed, err := multiformat.DeserializePublicKey(input, "f")
+	if err != nil {
+		return input
+	}
+	const secp256k1Code = "fe701"
+	return "0x" + strings.TrimPrefix(decompressed, secp256k1Code)
+}
+
+func (m *Message) DecompressPublicKeys() {
+	for _, linkPreview := range m.StatusLinkPreviews {
+		if linkPreview.Contact != nil {
+			linkPreview.Contact.PublicKey = decompressPublicKey(linkPreview.Contact.PublicKey)
+		}
+		if linkPreview.Community != nil {
+			linkPreview.Community.CommunityID = decompressPublicKey(linkPreview.Community.CommunityID)
+		}
+		if linkPreview.Channel != nil && linkPreview.Channel.Community != nil {
+			linkPreview.Channel.Community.CommunityID = decompressPublicKey(linkPreview.Channel.Community.CommunityID)
+		}
+	}
 }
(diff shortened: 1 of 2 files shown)


protocol/common/message_linkpreview.go
```diff
@@ -468,3 +470,26 @@ func (m *Message) ConvertFromProtoToStatusLinkPreviews(makeMediaServerURL func(m
 
 	return previews
 }
+
+func decompressPublicKey(input string) string {
```
ilmotta, line 474: @igor-sirotin, should we use something like this instead? Do we need the new custom implementation?

https://github.com/status-im/status-go/blob/85f8c92cde612820f9ee960b246ab3de285bc644/protocol/common/message_sender.go#L428

protocol/common/message_linkpreview.go
```diff
@@ -468,3 +470,26 @@ func (m *Message) ConvertFromProtoToStatusLinkPreviews(makeMediaServerURL func(m
...
+		return input
+	}
+	const secp256k1Code = "fe701"
+	return "0x" + strings.TrimPrefix(decompressed, secp256k1Code)
+}
+
+func (m *Message) DecompressPublicKeys() {
+	for _, linkPreview := range m.StatusLinkPreviews {
+		if linkPreview.Contact != nil {
+			linkPreview.Contact.PublicKey = decompressPublicKey(linkPreview.Contact.PublicKey)
```
ilmotta, line 486: In a way, this feels like a workaround no? The public key is already compressed because that's what we want clients to use. Maybe it's totally fine, I'd just confirm with @cammellos

cammellos: Ideally we move away from decompressed keys and we use compressed keys everywhere, we can make status-go endpoints agnostic, though it's still a lot of work to get rid of them completely, so maybe do it gradually is best, i guess it depends on why we need it, and we can make a decision?

igor-sirotin: In status-desktop we have mostly work with decompressed keys, e.g. there're many maps like `map[decompressedKey]Community`. 

Also, we store decompressed keys in the database. So I thought that it's the main format we use.

@cammellos @ilmotta 
I'm fine with changing it everywhere, but that sounds like a different issue to do. 
Do you think this this PR makes sense for now?

> though it's still a lot of work to get rid of them completely

And this very true 😄

protocol/common/message_linkpreview.go
igor-sirotin: We have so many implementations of these, `DecompressPubkey` is not the only one 🙁 
I just went with the simplest solution for now, because it needs clean up everywhere anyway.

Also, `DecompressPubkey` returns `ecdsa.PublicKey`, which I will have to convert back to String after operations. I don't know if that's a difficult operation, but this code looked simpler to me.

protocol/common/message_linkpreview.go
cammellos: hard to say, I don't want you to commit to a bunch of extra work that is not strictly required, and similarly on mobile we do exactly the same, we index by uncompressed key. I am not sure it would be a priority to make the swap, and I can imagine being quite painful, so I would leave that decision up to you, you'd know best the complexity of the change against the priorities.

igor-sirotin: I'm very sure making this change is much simpler for now 😄

cammellos (commented):
looks good, maybe a test? but otherwise is all ok

protocol/common/message_linkpreview.go
ilmotta: Yeah, up to you really @igor-sirotin. It might be worth trying, because calling a simple function that's already implemented and converting to a string might be easier than having a custom implementation and adding the unit tests.

ilmotta (approved):

igor-sirotin:
Closing this PR as the keys should be decoded and sent as `bytes` on the sender side: https://github.com/status-im/status-go/issues/4266
Will open a corresponding PR to fix that.

PR #4261, opened 2023-11-03
fix: Show balance for all accounts

Cuteivist:
Status-desktop: https://github.com/status-im/status-desktop/pull/12632
Task: https://github.com/status-im/status-desktop/issues/12001

Added parsing balance for multiple addresses at once.
Map is used to store current balance for each account at specific timestamp.
Moving to next timestamp updates the balances and from that cumulative balance is calculated for all accounts for specific timestamp.

### DIFF
--- a/services/wallet/history/balance_db.go
+++ b/services/wallet/history/balance_db.go
@@ -2,6 +2,7 @@ package history
 
 import (
 	"database/sql"
+	"encoding/hex"
 	"fmt"
 	"math/big"
 
@@ -33,8 +34,21 @@ type entry struct {
 
 type assetIdentity struct {
 	ChainID     uint64
-	Address     common.Address
+	Addresses   []common.Address
 	TokenSymbol string
+}
+
+func (a *assetIdentity) addressesToString() string {
+	var addressesStr string
+	for i, address := range a.Addresses {
+		addressStr := hex.EncodeToString(address[:])
+		if i == 0 {
+			addressesStr = "X'" + addressStr + "'"
+		} else {
+			addressesStr += ", X'" + addressStr + "'"
+		}
+	}
+	return addressesStr
 }
 
 func (e *entry) String() string {
@@ -87,8 +101,9 @@ func (b *BalanceDB) getEntriesWithoutBalances(chainID uint64, address common.Add
 
 func (b *BalanceDB) getNewerThan(identity *assetIdentity, timestamp uint64) (entries []*entry, err error) {
 	// DISTINCT removes duplicates that can happen when a block has multiple transfers of same token
-	rawQueryStr := "SELECT DISTINCT block, timestamp, balance FROM balance_history WHERE chain_id = ? AND address = ? AND currency = ? AND timestamp > ? ORDER BY timestamp"
-	rows, err := b.db.Query(rawQueryStr, identity.ChainID, identity.Address, identity.TokenSymbol, timestamp)
+	rawQueryStr := "SELECT DISTINCT block, timestamp, balance, address FROM balance_history WHERE chain_id = ? AND address IN (%s) AND currency = ? AND timestamp > ? ORDER BY timestamp"
+	queryString := fmt.Sprintf(rawQueryStr, identity.addressesToString())
+	rows, err := b.db.Query(queryString, identity.ChainID, identity.TokenSymbol, timestamp)
 	if err == sql.ErrNoRows {
 		return nil, nil
 	} else if err != nil {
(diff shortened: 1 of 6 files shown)


services/wallet/history/service_test.go
```diff
@@ -271,11 +272,68 @@ func Test_entriesToDataPoints(t *testing.T) {
...
+					},
+				},
+			},
+			want: []*DataPoint{
+				{
+					Balance:   (*hexutil.Big)(big.NewInt(11)),
+					Timestamp: 1,
+				},
+				{
+					Balance:   (*hexutil.Big)(big.NewInt(7)),
```
stefandunca: Why isn't `6` expected instead of the current `7`? As I see it, chain `2` doesn't change balance, so there is still balance `5` in there, and with the new value of `1`, for chain `1` I expect to have value `6` in here.

Cuteivist: @stefandunca Thank you for pointing out I missed handling the different chains!

In that example actually instead 7 it should be 12, because
* Address 1 has 5 on chain 2
* Address 2 has 6 on chain 1
* Address 1 has 1 on chain 1

So at timestamp 2 it should sum all three, giving 12.
Next values also always add that balance of 5 on chain 2.

Pushed the fix.

stefandunca: You're right; the balance change is per chain and address identity.

services/wallet/history/balance_db.go
```diff
@@ -86,9 +87,20 @@ func (b *BalanceDB) getEntriesWithoutBalances(chainID uint64, address common.Add
 }
 
 func (b *BalanceDB) getNewerThan(identity *assetIdentity, timestamp uint64) (entries []*entry, err error) {
+	var addressesStr string
```
IvanBelyakoff: Would be good to make it a separate helper func `addressesToString`

stefandunca (approved):
👌🏼

services/wallet/history/service.go
```diff
@@ -137,34 +137,67 @@ func (s *Service) mergeChainsBalances(chainIDs []uint64, address common.Address,
...
+				ChainID: entry.chainID,
+			}
+			if _, exists := tmpBalanceMap[key]; !exists {
+				tmpBalanceMap[key] = entry.balance
+			} else {
+				tmpBalanceMap[key].Add(tmpBalanceMap[key], entry.balance)
+			}
+		}
+
+		for key, balance := range tmpBalanceMap {
```
IvanBelyakoff: It fees like you don't want to update the initial map here, but just return a tmp one, since you assign it to the initial map line 198

services/wallet/history/service.go
```diff
@@ -137,34 +137,67 @@ func (s *Service) mergeChainsBalances(chainIDs []uint64, address common.Address,
...
+	type AddressKey struct {
+		Address common.Address
+		ChainID uint64
+	}
+
+	sumBalances := func(balanceMap map[AddressKey]*big.Int) *big.Int {
 		sum := big.NewInt(0)
-		for _, entry := range entries {
-			sum.Add(sum, entry.balance)
+		for _, balance := range balanceMap {
```
IvanBelyakoff, line 159: since you dont care about the keys of the map (address and chain - and I believe that is correct), and use only values why creating a map if you dont need it? From what I see the whole point of creating a map was to calculate balances separately based on chain and address, but we don't do that

IvanBelyakoff (commented):
Great job, @Cuteivist. I would appreciate your response on some comments

services/wallet/history/balance_db.go
Cuteivist: Good point. I added the helper function instead

services/wallet/history/service.go
Cuteivist: I found out that I can simplify this function.
Previous version had only address as a key. I added the chains, but I missed that I can refactor this function.

Now this function updates the balances from grouped entries for specific timestamp.

services/wallet/history/service.go
Cuteivist: In this function I don't care for keys. It is just to sum all saved balances at this point of time.

Map itself is required because in grouped entries by timestamp there might not be update for each account + chain pair each time.
That means in any speciifc timestamp there can be from 1 to all balances for all account + chain pairs.

To get sum of all balances at any timestamp we need to preserve previous balances for each pair.

IvanBelyakoff (approved):
Great job!

PR #4267, opened 2023-11-06
fix: 'Join Community' notification is not getting dismissed

mohsen-ghafouri:
After joining a community, the 'Join Community' notification remains persistent in the notification center, even after accessing the community

related to status-mobile issue https://github.com/status-im/status-mobile/issues/17572


protocol/messenger.go
```diff
@@ -4154,6 +4154,11 @@ func (m *Messenger) MarkAllRead(ctx context.Context, chatID string) error {
 	return m.markAllRead(chatID, clock, true)
 }
 
+func (m *Messenger) MarkReadCommunityOverview(ctx context.Context, communityID string) error {
```
qfrank: we can just remove this function and call `DismissAllActivityCenterNotificationsFromCommunity` directly?
and `messenger.go` is too big, better to move logics relate to AC into `protocol/messenger_activity_center.go` if need

protocol/messenger_activity_center.go
```diff
@@ -447,6 +447,14 @@ func (m *Messenger) DismissAllActivityCenterNotificationsFromUser(ctx context.Co
 	return notifications, m.syncActivityCenterDismissed(ctx, notifications, updatedAt)
 }
 
+func (m *Messenger) DismissActivityCenterNotificationsCommunity(ctx context.Context, communityID string, updatedAt uint64) ([]*ActivityCenterNotification, error) {
```
qfrank: rename `DismissActivityCenterNotificationsCommunity` to `DismissActivityCenterNotificationsByCommunity` sounds better?

protocol/activity_center_persistence.go
```diff
@@ -845,6 +845,37 @@ func (db sqlitePersistence) DismissActivityCenterNotifications(ids []types.HexBy
 	return updateActivityCenterState(tx, updatedAt)
 }
 
+func (db sqlitePersistence) DismissAllActivityCenterNotificationsCommunity(communityID string, updatedAt uint64) ([]*ActivityCenterNotification, error) {
```
qfrank: rename `DismissAllActivityCenterNotificationsCommunity` to `DismissAllActivityCenterNotificationsByCommunity` sounds better?

services/ext/api.go
```diff
@@ -814,6 +814,10 @@ func (api *PublicAPI) MarkAllRead(ctx context.Context, chatID string) error {
 	return api.service.messenger.MarkAllRead(ctx, chatID)
 }
 
+func (api *PublicAPI) MarkCommunityNotificationRead(ctx context.Context, communityID string) error {
+	return api.service.messenger.MarkReadCommunityOverview(ctx, communityID)
+}
```
qfrank, line 825: why we have different names: `MarkCommunityNotificationRead` / `MarkReadCommunityOverview` / `DismissActivityCenterNotificationsCommunity` , i'd prefer `DismissActivityCenterNotificationsByCommunity` , pretty straight for me :)

protocol/messenger.go
mohsen-ghafouri: @qfrank that function will dismiss notification related to community channel's but we need another function for dismiss notification related to just community it self and not it's channels.

services/ext/api.go
mohsen-ghafouri: sorry @qfrank my bad. will fix names

protocol/messenger.go
qfrank: oh, sorry, i mean `m.DismissActivityCenterNotificationsCommunity` , not `DismissAllActivityCenterNotificationsFromCommunity` , my bad for the wrong paste @mohsen-ghafouri

mohsen-ghafouri: we don't have `DismissActivityCenterNotificationsCommunity` func @qfrank

services/ext/api.go
mohsen-ghafouri: @qfrank i applied changes, could you please recheck . thanks

protocol/messenger.go
qfrank: emm.. isn't this one? https://github.com/status-im/status-go/blob/2410c01ab72122ecd20a838a14cf63e3e78ed141/protocol/messenger_activity_center.go#L450

mohsen-ghafouri: in `develop` branch we don't have it

qfrank: Oh, i saw you already removed the changes to `protocol/messenger.go`, that's what i mean, thank you!

qfrank (approved):
LGTM

services/ext/api.go
```diff
@@ -814,6 +814,10 @@ func (api *PublicAPI) MarkAllRead(ctx context.Context, chatID string) error {
 	return api.service.messenger.MarkAllRead(ctx, chatID)
 }
 
+func (api *PublicAPI) MarkCommunityNotificationRead(ctx context.Context, communityID string) error {
```
ilmotta: The name of this endpoint was confusing to me because we already have an endpoint `MarkActivityCenterNotificationsRead` which does essentially one thing, set `read = 1` in the notification. But `MarkCommunityNotificationRead` is doing more than just marking as read.

I suggest we use another name, for instance `DismissActivityCenterNotificationsByCommunity`, exactly the same name as in the `messenger`.

ilmotta: Even if we only need one argument `communityID` today, it is preferable to accept a Request struct because then new keys can be added in a backwards compatible way. This has been a problem a few times in the past that required coordination between releases in desktop/mobile. Example https://github.com/status-im/status-go/blob/c27384680a119eeeefc887e2aa0591e56387f121/protocol/requests/send_contact_request.go#L14

Then you would also be able to follow the convention to define a `Validate()` function and use it to check if the `communityID` argument is valid, something currently missing in the PR.

ilmotta (approved):
LGTM too 🚀

PR #4270, opened 2023-11-07
fix: promote self to community control node event if we are not a community member

mprakhov:
Promote self-to-community control node event if we are not a community member during the ownership transfer
This fix includes:
1. Searching in the mailserver for a community in case it is absent in the database
2. Adding as a member and setting owner role

Closes # https://github.com/status-im/status-desktop/issues/12642


osmaczko (approved):
lgtm

PR #4271, opened 2023-11-07
networks configuration 

saledjenic:
The first commit removes unneeded code and fixes the migration issue we had.

The second commit sorts out networks.

### DIFF
--- a/rpc/network/network.go
+++ b/rpc/network/network.go
@@ -130,6 +130,14 @@ func (nm *Manager) Init(networks []params.Network) error {
 	// Add new networks and update related chain id for the old ones
 	for i := range networks {
 		found := false
+		// Update original urls if they are empty
+		if networks[i].OriginalRPCURL == "" {
+			networks[i].OriginalRPCURL = networks[i].RPCURL
+		}
+		if networks[i].OriginalFallbackURL == "" {
+			networks[i].OriginalFallbackURL = networks[i].FallbackURL
+		}
+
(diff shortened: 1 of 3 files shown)


rpc/network/network.go
```diff
@@ -140,7 +148,7 @@ func (nm *Manager) Init(networks []params.Network) error {
 						errors += fmt.Sprintf("error updating network fallback_url for ChainID: %d, %s", currentNetworks[j].ChainID, err.Error())
 					}
 				}
-				err := nm.UpdateOriginalURL(currentNetworks[j].ChainID, networks[i].RPCURL, networks[i].FallbackURL)
+				err := nm.UpdateRpcURLs(currentNetworks[j].ChainID, networks[i].RPCURL, networks[i].OriginalRPCURL, networks[i].FallbackURL, networks[i].OriginalFallbackURL)
```
alaibe: why updating all urls here? we only need to update the original no?

alaibe (approved):

rpc/network/network.go
saledjenic: We're using those that are not "originals" and in case the user set env network-related vars, previously they were just updating "originals" but the RPC requests were made using "not originals", so that didn't have an effect. 
Also, this `Init` function is called only once on the app start and it's important to replace both, otherwise we're revealing our default app infura/pokt tokens.

stefandunca (approved):
👍🏼

PR #4276, opened 2023-11-07
Fix: Prevent Unintended Clearing of Previous Community Notifications

ajayesivan:
fix for https://github.com/status-im/status-mobile/issues/17052

### Summary
This pull request resolves the issue where deleting the last mention or reply for everyone in a community chat inadvertently cleared previous notifications.

### DIFF
--- a/protocol/activity_center_persistence.go
+++ b/protocol/activity_center_persistence.go
@@ -60,10 +60,6 @@ func (db sqlitePersistence) DeleteActivityCenterNotificationForMessage(chatID st
 	}
 
 	for _, notification := range notifications {
-		if notification.LastMessage != nil && notification.LastMessage.ID == messageID {
-			withNotification(notification)
-		}
-
 		if notification.Message != nil && notification.Message.ID == messageID {
 			withNotification(notification)
 		}
--- a/protocol/activity_center_persistence_test.go
+++ b/protocol/activity_center_persistence_test.go
@@ -209,7 +209,7 @@ func (s *ActivityCenterPersistenceTestSuite) Test_DeleteActivityCenterNotificati
 	_, err = p.DeleteActivityCenterNotificationForMessage(chat.ID, messages[1].ID, currentMilliseconds())
 	s.Require().NoError(err)
 
-	for _, id := range []types.HexBytes{nID2, nID3} {
+	for _, id := range []types.HexBytes{nID1, nID2} {
 		notif, err = p.GetActivityCenterNotificationByID(id)
 		s.Require().NoError(err, notif.ID)
 		s.Require().True(notif.Deleted, notif.ID)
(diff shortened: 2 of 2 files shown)


protocol/activity_center_persistence.go
```diff
@@ -60,10 +60,6 @@ func (db sqlitePersistence) DeleteActivityCenterNotificationForMessage(chatID st
 	}
 
 	for _, notification := range notifications {
-		if notification.LastMessage != nil && notification.LastMessage.ID == messageID {
-			withNotification(notification)
-		}
-
```
ajayesivan, line 66: This logic was not included in the [initial implementation](https://github.com/status-im/status-go/pull/2319) done by @briansztamfater  but was later introduced in [this PR](https://github.com/status-im/status-go/pull/2854) by @rasom. I'm uncertain whether removing it will cause any issues that I'm not aware of. To the best of my knowledge, everything seems to be working correctly based on my testing. Please don't hesitate to inform me if I've overlooked any details.

ajayesivan: This check compares the `LastMessage.ID` of the notification with the ID of the deleted message. When deleting the last sent message, the ID of the deleted message matches the `LastMessage.ID` of all other messages in the activity center, causing the removal of all notifications. This is the reason I removed this part, as it resolves the issue.

ilmotta: I don't fully understand the implications, but doesn't deleting this code means we'll have a notification pointing to a orphaned message that has been marked to be deleted? See https://github.com/status-im/status-go/blob/eb437e9d8dd5bd1ca3b18fa05c4853d6ae8f4956/protocol/messenger_handler.go#L1870

VolodLytvynenko:
Hey, this PR has been in the review column for quite some time. Is it still under review, or might it have been overlooked?

ajayesivan:
@VolodLytvynenko Moved it to the contributor column. I will make necessary changes and update you once it's ready for testing.

VolodLytvynenko:
> @VolodLytvynenko Moved it to the contributor column. I will make necessary changes and update you once it's ready for testing.

thank you for update!

protocol/activity_center_persistence.go
ilmotta: @ajayesivan, GitHub sent me a notification to review this PR again. I can't be sure if the fix is good or not from the code alone, and since you tested, seems alright.

ilmotta (approved):

ajayesivan:
@ilmotta Thanks for the review! As far as I tested the fix works, but I'm not 100% sure if it will introduce other issues. I will pair with Andrea and verify this.

igor-sirotin:
Issue resolved in https://github.com/status-im/status-go/pull/5789

PR #4277, opened 2023-11-08
Feat/print origins for log

qfrank:
This pull request primarily introduces additional `file location` to each log message, enabling more precise filtering of logs.
[relate description](https://github.com/status-im/go-ethereum/pull/103#issue-1981026107)

**NOTE**: The new log format omits the 'year' component of the timestamp. If there is a strong need to retain this information, please share your feedback. Thank you.

Example output for geth.log with this PR:
geth.log example


siddarthkay (approved):
Can't think of cases where having year as part of timestamp will help.
Nice one @qfrank

saledjenic (approved):

PR #4280, opened 2023-11-08
fix: better `IsStatusSharedURL`

### LINKED ISSUE #4275: StatusUnfurler treats https://status.app/ as Status link
# Problem

This happens because of this function implementation:
https://github.com/status-im/status-go/blob/b65eda3f42c86428566e32e8536fb59842378497/protocol/messenger_share_urls.go#L514-L516

# Implementation

`IsStatusSharedURL` should be smarted and at least check for `/u`, `/c` or `/cc` path.

igor-sirotin:
Fixes https://github.com/status-im/status-go/issues/4275

Now `IsStatusSharedURL` method check for URL to begin with one of the options:
- `https://status.app/u#`
- `https://status.app/u/`
- `https://status.app/c#`
- `https://status.app/c/`
- `https://status.app/cc/`

So link preveiw is generated properly for https://status.app:
image

### DIFF
--- a/protocol/messenger_share_urls.go
+++ b/protocol/messenger_share_urls.go
@@ -48,6 +48,18 @@ type URLDataResponse struct {
 }
 
 const baseShareURL = "https://status.app"
+const userPath = "u#"
+const userWithDataPath = "u/"
+const communityPath = "c#"
+const communityWithDataPath = "c/"
+const channelPath = "cc/"
+
+const sharedUrlUserPrefix = baseShareURL + "/" + userPath
+const sharedUrlUserPrefixWithData = baseShareURL + "/" + userWithDataPath
+const sharedUrlCommunityPrefix = baseShareURL + "/" + communityPath
+const sharedUrlCommunityPrefixWithData = baseShareURL + "/" + communityWithDataPath
+const sharedUrlChannelPrefixWithData = baseShareURL + "/" + channelPath
+
 const channelUUIDRegExp = "^[0-9a-f]{8}-[0-9a-f]{4}-[0-5][0-9a-f]{3}-[089ab][0-9a-f]{3}-[0-9a-f]{12}$"
 
 func (m *Messenger) SerializePublicKey(compressedKey types.HexBytes) (string, error) {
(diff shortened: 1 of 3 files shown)


protocol/messenger_share_urls.go
```diff
@@ -511,13 +523,17 @@ func (m *Messenger) parseUserURLWithData(data string, chatKey string) (*URLDataR
...
+		strings.HasPrefix(url, sharedURLUserPrefixWithData) ||
+		strings.HasPrefix(url, sharedURLCommunityPrefix) ||
+		strings.HasPrefix(url, sharedURLCommunityPrefixWithData) ||
+		strings.HasPrefix(url, sharedURLChannelPrefixWithData)
 }
 
 func (m *Messenger) ParseSharedURL(url string) (*URLDataResponse, error) {
-	if !m.IsStatusSharedURL(url) {
-		return nil, fmt.Errorf("url should start with '%s'", baseShareURL)
+	if !IsStatusSharedURL(url) {
```
MishkaRogachev: it would be a little more obvious to my taste if you combine here checking and parsing urls, something like this:
```
	if chatKey := chatKeyFromUrl(url); isStatusCommunityUrl(url) {
		return m.parseCommunityURLWithChatKey(chatKey)
	}
    ...
```

igor-sirotin: nice idea!
refactored the function, looks better to me now 👍

protocol/messenger_share_urls.go
```diff
@@ -511,51 +525,68 @@ func (m *Messenger) parseUserURLWithData(data string, chatKey string) (*URLDataR
...
+		chatKey := strings.TrimPrefix(url, sharedURLUserPrefix)
+		if strings.HasPrefix(chatKey, "zQ3sh") {
+			return m.parseUserURLWithChatKey(chatKey)
+		}
+		return m.parseUserURLWithENS(chatKey)
 	}
 
-	if strings.HasPrefix(urlContents[0], "c/") {
-		return m.parseCommunityURLWithData(strings.TrimPrefix(urlContents[0], "c/"), urlContents[1])
+	if strings.HasPrefix(url, sharedURLUserPrefixWithData) {
```
MishkaRogachev, line 555: yeah, that's much better!

MishkaRogachev (approved):
Great!

PR #4281, opened 2023-11-08
fix: edit message even if converting link previews to proto

igor-sirotin:
We should keep editing the message, even if link previews conversion to proto failed.

We do the same when sending the message:

https://github.com/status-im/status-go/blob/7009a7d18a748097f3554a88dc8257f6087ae55a/protocol/messenger.go#L2312-L2327


ilmotta (approved):
Nice catch!

PR #4285, opened 2023-11-09
feat: enable messages segmentation

### LINKED ISSUE #12188: Split CommunityDescription message (and all messages too big)
## Description

At the moment, the [CommunityDescription](https://github.com/status-im/status-go/blob/develop/protocol/protobuf/communities.proto#L84-L101) message is too large, both in terms of meaningfulness and size in bytes. In [this](https://github.com/status-im/status-desktop/issues/12114#issuecomment-1721426909) experiment, we found that the size of this message for Status CCs reaches 2 mb. 

This [PR](https://github.com/status-im/status-go/pull/4039) is a temporary measure until this feature is implemented. When switching to split messages I think we should remove this optimisation, but in the list of channel members we should not send the members themselves, but only their IDs.

osmaczko:
closes: https://github.com/status-im/status-desktop/issues/12188

This PR introduces a segmentation layer positioned directly above the transport layer. It facilitates the chunking of messages into smaller, manageable segments before they are transmitted via Waku. The segmentation mechanism is designed to be payload-agnostic, ensuring compatibility with any message type without necessitating protocol modifications.

Important: Legacy clients lacking support for this feature will be unable to process segmented messages.

TODO:
- [x] use `segmentMessage` function within `MessageSender`
- [x] add tests to verify the successful exchange of segmented messages between two messengers

Good starting point for review would be: `TestHandleSegmentMessages` and `TestRetrieveBigCommunity`


osmaczko:
@cammellos @Samyoul @jrainville this is still in draft, but I would appreciate a review at this stage regardless, as the foundation is functional.

protocol/migrations/sqlite/1699554099_message_segments.up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS message_segments (
```
osmaczko, line 1: Also store the `SigPubKey` to prevent malicious peers from sending segments with invalid payloads, which would effectively invalidate the message reassembly process.

osmaczko:
@cammellos @Samyoul @jrainville the PR is ready for review :muscle:

protocol/common/message_sender.go
```diff
@@ -34,6 +37,7 @@ const (
 	// Roughly this is 50KB
 	largeSizeInBytes = 50000
 	whisperPoWTime   = 5
+	// whisperMaxPayloadSize = 150 * 1024 // 150 kilobytes
```
osmaczko: leftover

protocol/requests/create_community_request.go
```diff
@@ -96,7 +96,6 @@ func (c *CreateCommunity) ToCommunityDescription() (*protobuf.CommunityDescripti
 	if c.Image != "" || c.Banner.ImagePath != "" {
 		ciis := make(map[string]*protobuf.IdentityImage)
 		if c.Image != "" {
-			log.Info("has-image", "image", c.Image)
```
osmaczko: bring back

protocol/requests/create_community_request.go
```diff
@@ -114,7 +113,6 @@ func (c *CreateCommunity) ToCommunityDescription() (*protobuf.CommunityDescripti
 			ciis[img.Name] = adaptIdentityImageToProtobuf(*img)
 		}
 		ci.Images = ciis
-		log.Info("set images", "images", ci)
```
osmaczko: bring back

protocol/common/raw_messages_persistence.go
```diff
@@ -336,3 +337,78 @@ func (db RawMessagesPersistence) DeleteHashRatchetMessages(ids [][]byte) error {
...
+	}()
+
+	sigPubKeyBlob := crypto.CompressPubkey(sigPubKey)
+
+	_, err = tx.Exec("DELETE FROM message_segments WHERE hash = ? AND sig_pub_key = ?", hash, sigPubKeyBlob)
+	if err != nil {
+		return err
+	}
+
+	_, err = tx.Exec("INSERT INTO message_segments_completed (hash, sig_pub_key) VALUES (?,?)", hash, sigPubKeyBlob)
```
jrainville, line 408: Do you think we'd need to clean up that table at some point? After 30 days, the entries here are pretty useless.
They should be small, so it's not a big issue, but it does take some space and the bigger the table the longer queries will take.
It can be done in another PR too

jrainville (approved):
LGTM! Nice job!

protocol/common/raw_messages_persistence.go
osmaczko: Yes, that would make sense :+1:  I planned to create a similar follow-up task to clean up the `message_segments` table in case segments aren't completed in X days/hours, because it would occupy even more memory, since the message segments themselves are much heavier than a tuple of two hashes. I'll make sure that task includes your proposal as well.

osmaczko: @jrainville FYI: https://github.com/status-im/status-go/issues/4297

protocol/common/message_sender.go
```diff
@@ -427,16 +434,18 @@ func (s *MessageSender) sendCommunity(
...
 		if err != nil {
 			s.logger.Error("failed to send a community message", zap.Error(err))
 			return nil, errors.Wrap(err, "failed to send a message spec")
 		}
+	}
 
-		s.logger.Debug("sent community message ", zap.String("messageID", messageID.String()), zap.String("hash", types.EncodeHex(hash)))
+	for i, newMessage := range newMessages {
+		s.logger.Debug("sent community message ", zap.String("messageID", messageID.String()), zap.String("hash", types.EncodeHex(hashes[i])))
+		s.transport.Track(messageIDs, hashes[i], newMessage)
```
cammellos, line 446: Interesting, we probably need to update at some point what it means for a message to be dispatched received?

cammellos (approved):

protocol/common/message_sender.go
osmaczko: Thanks, valid point. I will investigate it as followup: https://github.com/status-im/status-go/issues/4310

PR #4286, opened 2023-11-09
Integrating keycard signing in case of buying stickers/ens username

saledjenic:
We should let the client decide where to sign a transaction and once the tx is signed let him submit the tx providing the signature. This PR allows exactly that.


saledjenic:
@ajayesivan @ilmotta could you please check if the mobile app is using any of the following calls and if yes, could you or somebody else align with the following changes done in this pr:

**Ens Service**

1. Instead of the old `ens_release` call we should do the following calls:
- `ens_prepareTxForReleasingRegisteredEnsUsername`
- decide where to sign the tx from the previous call on a keycard or by calling `ens_signPreparedTx` for regular Status user
- call `ens_sendPreparedTxWithSignature`

2. Instead of the old `ens_register` call we should do the following calls:
- `ens_prepareTxForRegisteringEnsUsername`
- decide where to sign the tx from the previous call on a keycard or by calling `ens_signPreparedTx` for regular Status user
- call `ens_sendPreparedTxWithSignature`

3. Instead of the old `ens_setPubKey` call we should do the following calls:
- `ens_prepareTxForSettingPublicKey`
- decide where to sign the tx from the previous call on a keycard or by calling `ens_signPreparedTx` for regular Status user
- call `ens_sendPreparedTxWithSignature`

**Stickers Service**

1. Instead of the old `stickers_buy` call we should do the following calls:
- `stickers_prepareTxForBuyingStickers`
- decide where to sign the tx from the previous call on a keycard or by calling `stickers_signPreparedTx` for regular Status user
- call `stickers_sendPreparedTxWithSignature`

alaibe (approved):

ilmotta:
Thank you for the detailed changelog @saledjenic. The mobile repo is definitely affected by the endpoint changes.

I'm going to create an issue in `status-mobile` and see if we have capacity to make the adaptations this week. I'm going to report back here with any findings.

services/ens/api.go
```diff
@@ -336,42 +348,134 @@ func (api *API) Price(ctx context.Context, chainID uint64) (string, error) {
 	return fmt.Sprintf("%x", price), nil
 }
 
-func (api *API) Release(ctx context.Context, chainID uint64, txArgs transactions.SendTxArgs, password string, username string) (string, error) {
-	registryAddr, err := api.usernameRegistrarAddr(ctx, chainID)
+// Signs a transaction using appropriate local keystore file that is decrypted with provided password and returns the signature in hex format.
+// The tx that will be signed is the one that was prepared by the last call to
+// PrepareTxForRegisteringEnsUsername or PrepareTxForSettingPublicKey or PrepareTxForReleasingRegisteredEnsUsername.
+func (api *API) SingPreparedTx(password string) (interface{}, error) {
```
endulab: Sing -> Sign

endulab: Why do you return interface{} and not string?

services/stickers/transactions.go
```diff
@@ -65,23 +71,99 @@ func (api *API) Buy(ctx context.Context, chainID uint64, txArgs transactions.Sen
...
+		packID:        packID,
+	}
+
+	signer := ethTypes.NewLondonSigner(new(big.Int).SetUint64(api.txSignDetails.chainID))
+	return signer.Hash(api.txSignDetails.txBeingSigned), nil
+}
+
+// Signs a transaction using appropriate local keystore file that is decrypted with provided password and returns the signature in hex format.
+// The tx that will be signed is the one that was prepared by the last call to PrepareTxForBuyingStickers method.
+func (api *API) SingPreparedTx(password string) (interface{}, error) {
```
endulab: The code is the same as in ens/api.go. Can you move it to eg. services/utils/utils.go? There is a GetSigner function which you can use also.

services/stickers/transactions.go
```diff
@@ -65,23 +71,99 @@ func (api *API) Buy(ctx context.Context, chainID uint64, txArgs transactions.Sen
...
+	hash := signer.Hash(api.txSignDetails.txBeingSigned)
+
+	signature, err := crypto.Sign(hash[:], selectedAccount.PrivateKey)
+
+	return types.EncodeHex(signature), err
+}
+
+// Sends a transaction using provided signature and returns the transaction hash.
+// Provided signature must be in hex format and must be the result of signing the tx that was prepared by the last call to PrepareTxForBuyingStickers method.
+func (api *API) SendPreparedTxWithSignature(ctx context.Context, signature string) (string, error) {
```
endulab, line 103: Isn't it the same (mostly) as in ens/api.go?

services/stickers/transactions.go
cammellos: Sign I take?

services/ens/api.go
saledjenic: While I was developing I used `[]byte`, then `string` and then it remained like that. Np, will change that.

services/stickers/transactions.go
saledjenic: > There is a GetSigner function which you can use also.

Well, it looks similar, but it doesn't do exactly the same thing as `GetSigner` does. 
- `GetSigner`:
  - makes a tx hash for signing
  - signs tx hash building a signature
  - sends a tx using prepared signature and returns tx of sent tx
- `SingPreparedTx`:
  - signs prepared tx and returns signature
  
> Can you move it to eg. services/utils/utils.go?

Yes, those 2 are very similar, basically they do the same thing, but over different tx. One signs a tx prepared in the context of ens context, other signs a tx prepared in the stickers context. But yes, I can move those few lines, the common part, to the utils.go file, I agree.

saledjenic:
@cammellos @endulab comments addressed, thanks. :)

services/stickers/transactions.go
saledjenic: It is not, different tx, different contract, different context.

stefandunca (approved):
👍🏼

services/stickers/transactions.go
```diff
@@ -65,23 +71,89 @@ func (api *API) Buy(ctx context.Context, chainID uint64, txArgs transactions.Sen
...
+
+	return utils.SignTx(api.txSignDetails.chainID, api.accountsManager, api.keyStoreDir,
+		types.Address(api.txSignDetails.from), password, api.txSignDetails.txBeingSigned)
+}
+
+// Sends a transaction using provided signature and returns the transaction hash.
+// Provided signature must be in hex format and must be the result of signing the tx that was prepared by the last call to PrepareTxForBuyingStickers method.
+func (api *API) SendPreparedTxWithSignature(ctx context.Context, signature string) (string, error) {
+	if api.txSignDetails == nil {
+		return "", errors.New("no known tx to send to register ens username")
```
endulab, line 105: it is stickers, not ens

services/stickers/transactions.go
endulab: If you look at SendPreparedTxWithSignature in ens service, most of the code is identical, only the end differs.

endulab: Aaa, sorry, different api.

endulab (approved):

flexsurfer:
hey @saledjenic on mobile we use `ens_registerPrepareTx` and `ens_setPubKeyPrepareTx` just wondering if we need to change this

qfrank:
Hi @saledjenic , can't we just keep the original implementation for endpoints: `ens_register` and `ens_setPubKey` ? cc @cammellos @flexsurfer

flexsurfer:
@qfrank  ens_register and ens_setPubKey are not used on mobile, why do you need them?

qfrank:
> @qfrank ens_register and ens_setPubKey are not used on mobile, why do you need them?

I thought you see the usages on `ens_register` and `ens_setPubKey` since i saw you removed them in your PR. But they shouldn't be removed otherwise ens registration flow won't function proberbly @flexsurfer

flexsurfer:
> I thought you see the usages on `ens_register` and `ens_setPubKey` since i saw you removed them in your PR

i removed them just because they are not used, we use ens_registerPrepareTx and ens_setPubKeyPrepareTx

qfrank:
> > I thought you see the usages on `ens_register` and `ens_setPubKey` since i saw you removed them in your PR
> 
> i removed them just because they are not used, we use ens_registerPrepareTx and ens_setPubKeyPrepareTx

actually they're used, we should keep them except you don't want mobile ens works as before

flexsurfer:
> actually they're used, we should keep them except you don't want mobile ens works as before

could you please give an example of where they are used, in my mobile PR, because the point of this PR is to replace original methods, they should be used on both desktop and mobile

qfrank:
> > actually they're used, we should keep them except you don't want mobile ens works as before
> 
> could you please give an example of where they are used, in my mobile PR, because the point of this PR is to replace original methods, they should be used on both desktop and mobile

added more review comments in your PR

qfrank:
> hey @saledjenic on mobile we use `ens_registerPrepareTx` and `ens_setPubKeyPrepareTx` just wondering if we need to change this

I think we can remove `ens_registerPrepareTx` and `ens_setPubKeyPrepareTx` @saledjenic ? because we have new implementation `ens_prepareTxForRegisteringEnsUsername` and `ens_prepareTxForSettingPublicKey` now cc @flexsurfer

saledjenic:
@qfrank @flexsurfer sorry for the late reply.

> hey @saledjenic on mobile we use ens_registerPrepareTx and ens_setPubKeyPrepareTx just wondering if we need to change this

@flexsurfer not sure how the mobile app uses them, but since `ens_register` and `ens_setPubKey` do not exist anymore, I am not sure if we need them at all anymore. Maybe we should remove those to as well?

> Hi @saledjenic , can't we just keep the original implementation for endpoints: ens_register and ens_setPubKey ? cc @cammellos @flexsurfer

@qfrank if we need them, we can, but why do we need to have many API endpoints for the same feature? I am more for aligning the desktop and mobile app on this.

> I think we can remove ens_registerPrepareTx and ens_setPubKeyPrepareTx @saledjenic ? because we have new implementation ens_prepareTxForRegisteringEnsUsername and ens_prepareTxForSettingPublicKey now cc @flexsurfer

Exactly, if there are no other places in the mobile app that are using those endpoints I would remove them? 
Could you confirm and let me know if is that ok by your side? @flexsurfer

qfrank:
> if we need them, we can, but why do we need to have many API endpoints for the same feature? I am more for aligning the desktop and mobile app on this.

It depends on how much time we can wait for aligning, the quickest way to keep mobile ens feature work as before is keep the original implementation so that we can align it later after you merge this PR. If time is not what we need concidering, then let's just aligning on this. @saledjenic cc @cammellos @flexsurfer

saledjenic:
Ok, I can revert those endpoints and just add those that are needed for the desktop app and you can align with them later.
Does that work for you better? @flexsurfer

flexsurfer:
hey @saledjenic i think it was already implemented and we have `ens_registerPrepareTx` and `ens_setPubKeyPrepareTx` and i don't understand the point of having `ens_signPreparedTx` and `ens_sendPreparedTxWithSignature`, and same for stickers, why don't we have just general `wallet_signPreparedTx` and `wallet_sendPreparedTxWithSignature`, so on mobile currently it works next way, we prepare transaction and then sign and send it with `sendTransaction` why do we want to change this?

so we have a general sign transaction dialog, which signs and sends any transaction with user password, and adding conditions to it, or having some special dialog for stickers and ens doesn't make sense

flexsurfer:
so on mobile for ens it works next way
1. `ens_registerPrepareTx` we get transaction object
2. sign and send it as any other transaction with a general wallet sign dialog
3. on success `ens_register` for adding a record in the app db and then
4. watch transaction progress

and it works fine, why would we want to change this? it also works great with the keycard, probably desktop team could just use this instead of creating new API?

saledjenic:
> and it works fine, why would we want to change this? it also works great with the keycard, probably desktop team could just use this instead of creating new API?

@flexsurfer if you want I can update the code, leaving functions as they were and just adding those we need for the desktop app. Is that ok?

We need the tx hash that needs to be signed, and I don't see any that returns the tx hash, among the methods you mentioned.

flexsurfer:
as we discussed in 1-1, we use hashTransaction for getting the hash for keycard sign

saledjenic:
> as we discussed in 1-1, we use hashTransaction for getting the hash for keycard sign

Yes, that was the point, wondering why the desktop app has never been used since it gives a lot of flexibility. Will align with the desktop app now to what/how the mobile app does and most likely we won't need this PR anymore.

saledjenic:
@all closing this one cause it's not supported the way the mobile app does.
I've created new general endpoints in the new wallet API and once the mobile app switches to it will align the way the desktop app is dealing with these calls (basically the approach is the same just that wallet API endpoints include chain id).

For the reference, these are related PRs:
- https://github.com/status-im/status-go/pull/4393
- https://github.com/status-im/status-go/pull/4397

PR #4287, opened 2023-11-09
fix: add missing error check in currency service

dlipicar:
Added missing error check in the currency service, that caused a crash on startup when generating the list of currency formats.


alaibe (approved):

stefandunca (approved):
✨

PR #4290, opened 2023-11-10
Feat/collectibles api generalization

dlipicar:
Part of https://github.com/status-im/status-desktop/issues/12578

Change collectible APIs so that the desired data type and the "fetch criteria" can be specified by the caller.

Fetch criteria allows refreshing the list of owned collectibles when certain conditions are met. This is useful when checking collectibles for accounts not added to the wallet.

We now use a single Collectibles data type with optional fields to prevent having to add case-specific structs with different fields.

### DIFF
--- a/services/wallet/collectibles/service.go
+++ b/services/wallet/collectibles/service.go
@@ -93,6 +93,7 @@ func NewService(
 		ownershipDB: NewOwnershipDB(db),
 		communityDB: community.NewDataDB(db),
 		walletFeed:  walletFeed,
+		scheduler:   async.NewMultiClientScheduler(),
 	}
 	s.controller.SetReceivedCollectiblesCb(s.notifyCommunityCollectiblesReceived)
 	return s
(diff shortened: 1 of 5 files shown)


stefandunca (approved):
Great job 🤩

services/wallet/collectibles/service.go
```diff
@@ -89,88 +115,184 @@ type OwnershipStatus struct {
...
+}
+
+func (s *Service) needsToFetch(chainID walletCommon.ChainID, address common.Address, fetchCriteria FetchCriteria) (bool, error) {
+	mustFetch := false
+	switch fetchCriteria.FetchType {
+	case FetchTypeAlwaysFetch:
+		mustFetch = true
+	case FetchTypeNeverFetch:
+		mustFetch = false
+	case FetchTypeFetchIfNotCached, FetchTypeFetchIfCacheOld:
```
IvanBelyakoff, line 186: Curious if `mustFetch` is set to true for FetchTypeFetchIfNotCached

IvanBelyakoff (approved):
LGTM. Feel free to respond or fix the comment if you find it valid

services/wallet/collectibles/service.go
dlipicar: It should. Not cached means `timestamp == InvalidTimestamp`, which is the first condition we check.

PR #4294, opened 2023-11-10
feat: new `GetTextURLsToUnfurl` endpoint which extends GetURLs

### LINKED ISSUE #4235: Extend `GetTextURLs` endpoint to add unfurling plan for each URL
> Keep the [idea of async API](https://github.com/status-im/status-go/issues/4192) for later. 
Extend `GetTextURLs` (or create a new separate endpoint). For each URL found add some metadata that can be user to make further  requests to `UnfurlURLs`  smarter. The endpoint should process the settings and add instructions for the app if this link should be unfurled. 

Any ideas are welcome!

igor-sirotin:
The new `GetURLsToUnfurl` endpoint checks user settings and correspondingly set `Permit` field, which should be used by the client app to decide whether to unfurl the URL or not.

Status shared URLs are always marked as unfurled.
Because I already check if it's a Status shared URL here, I decided to add `IsStatusSharedURL` to the response as well. It can be used by the app to unfurl those links separately e.g.

I will keep this a draft until I test it with desktop app.

Closes https://github.com/status-im/status-go/issues/4235


igor-sirotin:
There's currently failing a flaky test, I fixed it here: https://github.com/status-im/status-go/pull/4312

igor-sirotin:
Tested with desktop here: https://github.com/status-im/status-desktop/pull/12751

protocol/messenger_linkpreview.go
```diff
@@ -116,23 +145,57 @@ func GetURLs(text string) []string {
...
+		}
+
+		metadata := URLUnfurlingMetadata{
+			URL:               url,
+			IsStatusSharedURL: IsStatusSharedURL(url),
+		}
+
+		//if len(result.URLs) == UnfurledLinksPerMessageLimit {
+		//	metadata.Permission = URLUnfurlingForbiddenByLimit
+		//}
```
jrainville: leftover

jrainville (approved):
LGTM

protocol/messenger_linkpreview.go
```diff
@@ -116,23 +145,54 @@ func GetURLs(text string) []string {
...
-		// This is a temporary limitation solution,
-		// should be changed with https://github.com/status-im/status-go/issues/4235
-		if len(urls) == UnfurledLinksPerMessageLimit {
+		result.URLs = append(result.URLs, metadata)
+	}
+
+	return result
+}
+
+// Deprecated: GetURLs is deprecated in favor of more generic GetTextURLsToUnfurl.
```
ilmotta, line 182: The deprecation warning should ideally be in the public API too, or just there.

Speaking about this @igor-sirotin, if the new endpoint doesn't require too many changes on mobile, I can tell we will soon be able to remove the deprecated endpoint. Thanks for evolving status-go in a backwards compatible way!

VERSION
```diff
@@ -1 +1 @@
-0.171.12
+0.172.0
```
ilmotta: I don't quite understand our versioning scheme. @igor-sirotin, are you bumping to `0.172.0` because there's a breaking change? I'm asking this because I thought when there's no breaking change we just bump the patch version.

Anyway, how we define and more importantly use versioning in status-go is hazy, but it doesn't hurt to ask :)

ilmotta (approved):
🚀

VERSION
igor-sirotin: oh my bad, I'll change it to `0.171.13` 👌

protocol/messenger_linkpreview.go
igor-sirotin: Will add it there too 👍

igor-sirotin: @ilmotta, the new endpoint doesn't require much to do, but there's something to code.
dm me when you start working on this, I'll let you know of desktop solution with this endpoint 🙂

PR #4295, opened 2023-11-10
feat: detect flaky tests

### LINKED ISSUE #4264: Add nightly build for detection of flaky tests
# Problem

Lately, we have been grappling with a significant number of unstable tests. Currently, pinpointing these erratic tests requires manual effort, which involves observing which tests fail upon re-execution. This approach is far from ideal.

Implementing a nightly build process that executes each test multiple times and generates an accessible report would greatly assist in detecting unstable tests. Additionally, this could help in tracing back to commits that potentially increased the instability of existing tests.

yakimant:
Need to create a Jenkins job to
- run tests multiple times nightly
- don't fail immediatly on test failure
- generate junit report with multiple runs
- make sure it is correctly shown in Jenkins (not done in this PR)
- the best would be to send an email report (leave for later)

Closes #4264

### DIFF
--- a/_assets/scripts/run_unit_tests.sh
+++ b/_assets/scripts/run_unit_tests.sh
@@ -5,21 +5,53 @@ GIT_ROOT=$(cd "${BASH_SOURCE%/*}" && git rev-parse --show-toplevel)
 
 source "${GIT_ROOT}/_assets/scripts/colors.sh"
 
+if [[ $UNIT_TEST_FAILFAST == 'true' ]]; then
+  GOTEST_EXTRAFLAGS="${GOTEST_EXTRAFLAGS} --failfast"
+fi
+
+if [[ -z "${UNIT_TEST_COUNT}" ]]; then
+  UNIT_TEST_COUNT=1
+fi
+
+redirect_stdout() {
+  output_file=$1
+
+  if [[ "${CI}" == 'true' ]];
+  then
+    cat > "${output_file}";
+  else
+    tee "${output_file}";
+  fi
+}
+
+last_failing_exit_code=0
+
 for package in ${UNIT_TEST_PACKAGES}; do
   echo -e "${GRN}Testing:${RST} ${package}"
   package_dir=$(go list -f "{{.Dir}}" "${package}")
   output_file=${package_dir}/test.log
 
-  go test -tags "${BUILD_TAGS}" -timeout 30m -v -failfast "${package}" ${GOTEST_EXTRAFLAGS} | \
-     if [ "${CI}" = "true" ]; then cat > "${output_file}"; else tee "${output_file}"; fi
+  go test -count="${UNIT_TEST_COUNT}" -tags "${BUILD_TAGS}" -v "${package}" ${GOTEST_EXTRAFLAGS} | \
+    redirect_stdout "${output_file}"
   go_test_exit=$?
 
-  if [ "${CI}" = "true" ]; then
+  if [[ "${CI}" == 'true' ]]; then
     go-junit-report -in "${output_file}" -out "${package_dir}"/report.xml
   fi
 
-  if [ ${go_test_exit} -ne 0 ]; then
+  if [[ "${go_test_exit}" -ne 0 ]]; then
     echo -e "${YLW}Failed, see the log:${RST} ${BLD}${output_file}${RST}"
-    exit "${go_test_exit}"
+    if [[ "$UNIT_TEST_FAILFAST" == 'true' ]]; then
+      exit "${go_test_exit}"
+    fi
+    last_failing_exit_code="${go_test_exit}"
   fi
 done
+
+if [[ "${last_failing_exit_code}" -ne 0 ]]; then
+  if [[ "${UNIT_TEST_COUNT}" -gt 1 ]]; then
+    "${GIT_ROOT}/_assets/scripts/test_stats.py"
+  fi
+
+  exit "${last_failing_exit_code}"
+fi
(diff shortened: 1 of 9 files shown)


yakimant:
Looks like `go-junit-report` is not combining the results of multiple runs.
I also could not find it in specs:
- https://github.com/testmoapp/junitxml
- https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd

As I remember I was using some software able co combine multiple runs, maybe that one:
https://github.com/lyndsey-ferguson/fastlane-plugin-test_center/blob/master/docs/feature_details/collate_junit_reports.md
And then loading the resulting xml to Teamcity succesfully.

yakimant:
https://github.com/bhovhannes/junit-report-merger doesn't solve the issue with multirun junit report.

yakimant:
https://github.com/lyndsey-ferguson/fastlane-plugin-test_center/blob/master/docs/feature_details/collate_junit_reports.md
provides smth better, trying to merge test results with `retries` attribute.

yakimant:
Another option is to have our own html generation, see:
- https://github.com/vakenbolt/go-test-report (will not show failed test if it recovered)
- https://github.com/Thatooine/go-test-html-report (simmilar to jenkins handling, but worse UI)
- https://github.com/medyagh/gopogh/

And then integrate to Jenkins:
https://github.com/jenkinsci/htmlpublisher-plugin

yakimant:
gopogh is Okish, although looks ugly:

Screenshot 2023-11-16 at 15 27 50

---

Screenshot 2023-11-16 at 15 28 12

---

Screenshot 2023-11-16 at 15 28 28

yakimant:
For Jenkins email integration:
1. https://www.jenkins.io/doc/pipeline/steps/email-ext/
2. https://plugins.jenkins.io/mailer/

_assets/scripts/run_unit_tests.sh
```diff
@@ -1,25 +1,45 @@
 #!/usr/bin/env bash
 set -o pipefail
+source "${GIT_ROOT}/_assets/scripts/colors.sh"
```
jakubgs, line 6: Definition `GIT_ROOT` is below this.

_assets/scripts/run_unit_tests.sh
```diff
@@ -1,25 +1,45 @@
...
 
 for package in ${UNIT_TEST_PACKAGES}; do
   echo -e "${GRN}Testing:${RST} ${package}"
   package_dir=$(go list -f "{{.Dir}}" "${package}")
   output_file=${package_dir}/test.log
 
-  go test -tags "${BUILD_TAGS}" -timeout 30m -v -failfast "${package}" ${GOTEST_EXTRAFLAGS} | \
-     if [ "${CI}" = "true" ]; then cat > "${output_file}"; else tee "${output_file}"; fi
+  go test -count="${UNIT_TEST_COUNT}" -tags "${BUILD_TAGS}" -v "${package}" ${GOTEST_EXTRAFLAGS} | \
+     if [[ "${CI}" == 'true' ]]; then cat > "${output_file}"; else tee "${output_file}"; fi
```
jakubgs: This line looks like it could be a function.

_assets/scripts/test_stats.py
```diff
@@ -0,0 +1,34 @@
...
+for file in glob.glob("**/report.xml"):
+    tree = ET.parse(file)
+    root = tree.getroot()
+    for testcase in root.iter("testcase"):
+        test_name = testcase.attrib["name"]
+
+        if test_name not in test_stats:
+            test_stats[test_name] = {"failed": 0, "total": 0}
+
+        test_stats[test_name]["total"] += 1
```
jakubgs: You could use [setdefault](https://docs.python.org/2/library/stdtypes.html#dict.setdefault) for this.
```suggestion
        test_stats.setdefault(test_name, {"failed": 0, "total": 0})
        test_stats[test_name]["total"] += 1
```
But there is an even more Pythonic way using `defaultdict`:
```python
In [1]: from collections import defaultdict

In [2]: d = defaultdict(lambda: defaultdict(int))

In [3]: d['x']['z'] += 1

In [4]: d['x']['z'] += 1

In [5]: d['x']['y'] += 1

In [6]: d
Out[6]: 
defaultdict(<function __main__.<lambda>()>,
            {'x': defaultdict(int, {'z': 2, 'y': 1})})
```

_assets/scripts/test_stats.py
```diff
@@ -0,0 +1,34 @@
...
+
+        if testcase.find("failure") is not None:
+            test_stats[test_name]["failed"] += 1
+
+failing_test_stats = [
+    {"name": name, "failure_rate": stats["failed"] / stats["total"]}
+    for name, stats in test_stats.items() if stats["failed"] != 0
+]
```
jakubgs, line 25: Cute.

jakubgs (approved):
Looks simple enough.

_assets/scripts/test_stats.py
yakimant: Beautiful, thanks!

_assets/scripts/run_unit_tests.sh
yakimant: Done!

yakimant:
10 runs took 1h, but this is not stable.
Each run cat take up to 30-40 minutes, resulting in 6h in worst case.

jakubgs (approved):
I'm wondering id maybe contents of the `for` loop in the `run_unit_tests.sh` could be extracted into `run_single_unit_tests.sh` or at least a function.

Not mandatory tho.

PR #4302, opened 2023-11-13
local pair: check netIPs if empty before find server cert

qfrank:
this PR should fix relate [issue](https://github.com/status-im/status-mobile/issues/17634#issuecomment-1806768628)

empty netIPs will cause `findServerCert` hanging


siddarthkay (approved):

igor-sirotin (approved):
LGTM 👍

PR #4304, opened 2023-11-13
Fix/timesource offline

osmaczko:
closes: https://github.com/status-im/status-desktop/issues/12691

- fixes "panic: sync: negative WaitGroup counter"

- ensures `timesource.GetCurrentTime` is non-blocking and doesn't return error

Otherwise, users will encounter the following problems when offline:

- The application may freeze for approximately two minutes before the login screen appears, as it waits for a retry.
- Subsequent retry attempts may still result in an error, leading to the `openAccounts` error. This prevents users from accessing their accounts while offline.

### DIFF
--- a/timesource/timesource.go
+++ b/timesource/timesource.go
@@ -201,6 +201,7 @@ func (s *NTPTimeSource) updateOffset() error {
 			c.callback(now())
 			c.wg.Done()
 		}
+		s.callbacks = nil
 		return errUpdateOffset
 	}
 	log.Info("Difference with ntp servers", "offset", offset)
--- a/timesource/timesource.go
+++ b/timesource/timesource.go
@@ -7,8 +7,6 @@ import (
 	"sync"
 	"time"
 
-	"go.uber.org/zap"
-
 	"github.com/beevik/ntp"
 
 	"github.com/ethereum/go-ethereum/log"
@@ -145,11 +143,6 @@ func Default() *NTPTimeSource {
 	return ntpTimeSourceCreator()
 }
 
-type timeCallback struct {
-	wg       *sync.WaitGroup
-	callback func(time.Time)
-}
-
 // NTPTimeSource provides source of time that tries to be resistant to time skews.
 // It does so by periodically querying time offset from ntp servers.
 type NTPTimeSource struct {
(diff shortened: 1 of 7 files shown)


timesource/timesource.go
```diff
@@ -239,19 +203,29 @@ func (s *NTPTimeSource) runPeriodically(fn func() error) error {
 				}
```
qfrank, line 203: Thank you for your PR @osmaczko , I think the best way to solve the relate issue would be adjust logic `else if` here (e.g. add indicator that already tried first time but got error, so next invoking to `GetCurrentTime` should not wait and return time using latest offset), we should try our best to wait(up to 2 seconds as we have `DefaultRPCTimeout`) timesource to get the offset for the first time at least in some cases.

osmaczko: > we should try our best to wait(up to 2 seconds as we have DefaultRPCTimeout) timesource to get the offset for the first time at least in some cases.

I agree, this is exactly what I do here: https://github.com/status-im/status-go/blob/52cd6bbb2c871d48f788c54600cfe2dfc7748fe4/timesource/timesource.go#L218-L226

PS. I believe with that approach we don't need additional indicator and we can keep `runPeriodically` simple, wdyt?

qfrank: awesome, i didn't notice this change before my previous comment, LGTM! 👍

timesource/timesource_test.go
```diff
@@ -271,20 +272,39 @@ func TestGetCurrentTimeInMillis(t *testing.T) {
...
+			fastNTPSyncPeriod: 1 * time.Millisecond,
+			slowNTPSyncPeriod: 1 * time.Second,
+			timeQuery: func(string, ntp.QueryOptions) (*ntp.Response, error) {
+				return nil, errors.New("offline")
+			},
+		}
+		return ntpTimeSource
+	}
+	_ = GetCurrentTime()
+	_ = GetCurrentTime()
```
qfrank, line 311: It's not that appearently we are expecting this invoking should return quickly, maybe add a timeout for the second invoking ?

qfrank (approved):

jrainville (approved):

timesource/timesource_test.go
osmaczko: Not sure I understand the point, could you please elaborate?

PS. This was to test two consecutive calls to `GetCurrentTime` won't result in "panic: sync: negative WaitGroup counter" if offline.

qfrank: Ah, I see now, thank you for your explanation. It would be helpful if you could include a comment alongside the second invocation to clarify what is expected. Since we're unable to assert in this context, adding a descriptive comment would make the intent more transparent

PR #4306, opened 2023-11-13
Feat/erc1155 batch transfer detection

### LINKED ISSUE #4221: Detect ERC1155 transfers

IvanBelyakoff:
- Implemented detection of ERC1155 batch transfers (`TransferBatch` event):
- Refactored transfers loading to reduce blockchain RPC requests (removed unnecessary getBaseFee, getTransaction,  getTransactionReceipt calls) by reusing preloaded transactions, tx and receipt caches and block's fee
- Split extraction of subtransactions from logs and from ETH transfer into different methods.
- Refactored log_parser to extract sender and receiver addresses uniformly for different transfer types.
- Replaced info logs with debug where needed.

Closes #4221


services/wallet/transfer/downloader.go
```diff
@@ -188,20 +250,15 @@ func (d *ETHDownloader) getTransfersInBlock(ctx context.Context, blk *types.Bloc
...
+				receipt, err := getReceipt(address, tx.Hash())
 				if err != nil {
 					return nil, err
 				}
 
 				// Since we've already got the receipt, check for subTxs of
 				// interest in case we haven't already.
 				if !areSubTxsCheckedForTxHash[tx.Hash()] {
-					subtransactions, err := d.subTransactionsFromTransactionData(tx, receipt, blk, baseGasFee, address)
+					subtransactions, err := d.subTransactionsFromTransactionData(address, from, tx, receipt, blk)
```
IvanBelyakoff, line 261: @dlipicar Could you have a look please, how did we (did we?) filter these subtransactions (DynamicFee, Deposit, Retry) by address to make sure we don't add the ones unrelated to the account?

services/wallet/transfer/downloader.go
```diff
@@ -330,47 +383,83 @@ func (d *ETHDownloader) subTransactionsFromTransactionData(tx *types.Transaction
...
-			trFrom, trTo, _ := w_common.ParseErc721TransferLog(log)
-			if trFrom == address || trTo == address {
-				mustAppend = true
-			}
-		case w_common.Erc1155TransferSingleEventType:
-			_, trFrom, trTo, _, _ := w_common.ParseErc1155TransferSingleLog(log)
-			if trFrom == address || trTo == address {
-				mustAppend = true
-			}
-		case w_common.UniswapV2SwapEventType, w_common.UniswapV3SwapEventType:
```
IvanBelyakoff, line 355: @dlipicar we never seem to compare `address` to anything for this types of transactions. What am I missing?

services/wallet/transfer/downloader.go
dlipicar: I'll need to revisit this a later to be fully sure, but what I remember at this point is:
- A swap operation always involves an ETH/ERC20 transfer from/to the account address in the same transaction
- `subTransactionsFromTransactionHash` (for ERC20 transfers) or `subTransactionsFromTransactionData` (for ETH transfers) are called only after filtering transactions irrelevant to the wallet address at a previous stage.

Some sub transactions will not involve the wallet address, that's normally the case for ERC20 -> ERC20 swaps, where at some point some intermediate transfer will occur between swap pools. However, we still want to store this in the DB to have more information about the Swap.


-0-

For those other weird Tx types, which are needed to detect Bridge transactions, we look fro the wallet address in the Transaction data

```
				switch tx.Type() {
				case types.DynamicFeeTxType, types.OptimismDepositTxType, types.ArbitrumDepositTxType, types.ArbitrumRetryTxType:
					mustCheckSubTxs = !areSubTxsCheckedForTxHash[tx.Hash()] && w_common.TxDataContainsAddress(tx.Type(), tx.Data(), address)
				}
			}
```

services/wallet/transfer/downloader.go
dlipicar: See previous comment. I might be wrong, but I think that was the original intention.

rasom (approved):

services/wallet/common/log_parser.go
```diff
@@ -251,52 +247,153 @@ func ParseErc721TransferLog(ethlog *types.Log) (from, to common.Address, tokenID
...
+}
+
+func GetLogSubTxID(log types.Log) common.Hash {
+	// Get unique ID by using TxHash and log index
+	index := [4]byte{}
+	binary.BigEndian.PutUint32(index[:], uint32(log.Index))
+	return crypto.Keccak256Hash(log.TxHash.Bytes(), index[:])
+}
+
+func getLogSubTxIDWithTokenID(log types.Log, tokenID uint16) common.Hash {
```
dlipicar: hmm can't the same tokenID appear twice in the list of ids for a `TransferBatch`? That would result in the same subTxID for multiple transfer entries. I'd use the corresponding index of the `ids` array instead to make sure the hash is unique.

services/wallet/common/log_parser.go
```diff
@@ -537,32 +616,27 @@ func ExtractTokenIdentity(dbEntryType Type, log *types.Log, tx *types.Transactio
...
+		from, to, _ := ParseErc721TransferLog(log)
 		txFrom = &from
 		txTo = &to
-	case Erc1155SingleTransfer:
+	case Erc1155Transfer:
 		tokenAddress = new(common.Address)
 		*tokenAddress = log.Address
-		_, from, to, tokenID, value := ParseErc1155TransferSingleLog(log)
-		txTokenID = tokenID
+		_, from, to, _, _, err := ParseErc1155TransferLog(log, Erc1155TransferSingleEventType)
```
dlipicar: Hmm I understand it doesn't matter since `from` and `to` extraction is the same for the batch and single transfers, but hardcoding type `Erc1155TransferSingleEventType` here is a bit confusing. You could add a comment about that

services/wallet/common/log_parser.go
IvanBelyakoff: I think you are correct that it is possible to use any token ID as many times as one wants, thought practically that would be a very silly implementation of a contract or client to send the same token in the same transfer multiple times instead of using proper value for it. So if anyone does it we will have a single transfer entry in the activities list, but I think when we request balance of that token Id for the account of interest, we should get the correct value.

IvanBelyakoff: Still fixed it, since it was easy and I don't see any drawbacks. Thanx

PR #4312, opened 2023-11-13
fix: self-contact related flaky tests

igor-sirotin:
`Test_UnfurlURLs_SelfLink` and `Test_SelfContact` are now based on direct self-contact change events.

Previously they were tracking the `selfContact` change by origin events like `ultiAccounts.SubscribeToIdentityImageChanges` and `messenger.settings.SubscribeToChanges`. But because those events are processed by `messenger` in concurrently, it was leading to flaky tests.

### DIFF
--- a/protocol/messenger_contacts.go
+++ b/protocol/messenger_contacts.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/ethereum/go-ethereum/log"
 	"github.com/golang/protobuf/proto"
 	"go.uber.org/zap"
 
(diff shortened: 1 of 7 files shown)


protocol/messenger_contacts.go
```diff
@@ -1291,3 +1304,23 @@ func (m *Messenger) forgetContactInfoRequest(publicKey string) {
 
 	delete(m.requestedContacts, publicKey)
 }
+
+func (m *Messenger) GetDeleteForMeMessages() ([]*protobuf.SyncDeleteForMeMessage, error) {
```
jrainville: This isn't really related to contacts

protocol/messenger_contacts.go
```diff
@@ -1291,3 +1304,23 @@ func (m *Messenger) forgetContactInfoRequest(publicKey string) {
...
+	return m.persistence.GetDeleteForMeMessages()
+}
+
+func (m *Messenger) SubscribeToSelfContactChanges() chan *SelfContactChangeEvent {
+	s := make(chan *SelfContactChangeEvent, 10)
```
jrainville, line 1310: Why 10?

jrainville (commented):
Looks good, just two comments

protocol/messenger_contacts.go
igor-sirotin: lol indeed, I don't know how it got here 😄

protocol/messenger_contacts.go
igor-sirotin: Well I saw we mostly set it to `100`. But I thought that this one has relatively low usage and there's no need for such big buffer. Do you think it's better to keep it larger?

jrainville: No, I just forgot how `chan` works and I found it weird to see a random magic number. I don't know what's best here

igor-sirotin: 10 should be enough imo. And we actually only use this subscription in tests anyway.

jrainville (approved):

PR #4313, opened 2023-11-14
fix(discord): Handle multiple channels with the same name

borismelnik:
This PR fix handling multiple channels with the same name by adding `_number` to the end of the name.

Important changes:

Closes #

### DIFF
--- a/protocol/messenger_communities_import_discord.go
+++ b/protocol/messenger_communities_import_discord.go
@@ -1087,6 +1087,9 @@ func (m *Messenger) RequestImportDiscordCommunity(request *requests.ImportDiscor
 		processedChannelIds := make(map[string]string, 0)
 		processedCategoriesIds := make(map[string]string, 0)
 
+		// The map with counts of duplicated channel names
+		uniqueChatNames := make(map[string]int, 0)
+
 		for i, importFile := range request.FilesToImport {
 
 			exportData, errs := m.ExtractDiscordDataFromImportFiles([]string{importFile})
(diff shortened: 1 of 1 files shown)


protocol/messenger_communities_import_discord.go
```diff
@@ -1087,6 +1087,9 @@ func (m *Messenger) RequestImportDiscordCommunity(request *requests.ImportDiscor
 		processedChannelIds := make(map[string]string, 0)
 		processedCategoriesIds := make(map[string]string, 0)
 
+		// The map with counts of duplicated channel names
+		ununiqueChatNames := make(map[string]int, 0)
```
caybro:
```suggestion
		uniqueChatNames := make(map[string]int, 0)
```

borismelnik: probable you right :)

protocol/messenger_communities_import_discord.go
```diff
@@ -1184,12 +1187,20 @@ func (m *Messenger) RequestImportDiscordCommunity(request *requests.ImportDiscor
 			}
 
 			if !exists {
+				channelUniqueName := channel.Channel.Name
+				if count, ok := uniqueChatNames[channelUniqueName]; ok {
+					uniqueChatNames[channelUniqueName] = count + 1
+					channelUniqueName = fmt.Sprintf("%s_%d", channelUniqueName, uniqueChatNames[channelUniqueName])
```
MishkaRogachev, line 1193: I would put count+1 in a separate variable so that it doesn't have to be queried from the map again, but that's more a matter of taste.

MishkaRogachev (approved):

jrainville (approved):

PR #4316, opened 2023-11-14
Collectibles backend improvements

dlipicar:
Two improvements done to collectibles fetching:

1) Pass context to collectibles requests. This helps avoid blocking when some scheduled task involving collectibles gets cancelled.
2) When fetching some community info fails, wait a certain period of time before trying to fetch it again. This helps when the user owns collectibles minted by "dead" (owner node down) communities, the first community info fetch will timeout and for a while no further attempts will be made.


stefandunca (approved):
😍

PR #4320, opened 2023-11-14
Feat/cache curated communities

### LINKED ISSUE #12277: Cache Community portal results in a table so that we don't have to wait for the contract call
### Description

Split up version of https://github.com/status-im/status-desktop/issues/7418

Start async process to fetch curated communities as soon as the user logs in. Then once this data is fetched for the first time it should be cached, probably in a table, so that the next time the user logs in, if they navigate to the community directory, and the latest data has not been fetched, then the cached data should be displayed instead.

One way to do it is to maybe have a field in the communities table about whether the community is curated and/or featured.
Then just have a query that retrieves only those.

### Acceptance Criteria

- Curated communities are cached some way or  another
- Opening the app and going to the curated communities portal shows the previous results even before the contract answers

osmaczko:
- fetch contract every two minutes and save results in database
- always return database entries to clients

closes: https://github.com/status-im/status-desktop/issues/12277

### DIFF
--- a/protocol/communities/persistence.go
+++ b/protocol/communities/persistence.go
@@ -1400,7 +1400,7 @@ func decodeEventsData(eventsBytes []byte, eventsDescriptionBytes []byte) (*Event
 func (p *Persistence) GetCommunityRequestsToJoinWithRevealedAddresses(communityID []byte) ([]*RequestToJoin, error) {
 	requests := []*RequestToJoin{}
 	rows, err := p.db.Query(`
-	SELECT r.id, r.public_key, r.clock, r.ens_name, r.chat_id, r.state, r.community_id, 
+	SELECT r.id, r.public_key, r.clock, r.ens_name, r.chat_id, r.state, r.community_id,
 		a.address, a.chain_ids, a.is_airdrop_address, a.signature
 	FROM communities_requests_to_join r
 	LEFT JOIN communities_requests_to_join_revealed_addresses a ON r.id = a.request_id
@@ -1604,7 +1604,7 @@ func (p *Persistence) SaveRequestsToJoin(requests []*RequestToJoin) (err error) 
 		}
 	}()
 
-	stmt, err := tx.Prepare(`INSERT OR REPLACE INTO communities_requests_to_join(id,public_key,clock,ens_name,chat_id,community_id,state) 
+	stmt, err := tx.Prepare(`INSERT OR REPLACE INTO communities_requests_to_join(id,public_key,clock,ens_name,chat_id,community_id,state)
 		VALUES (?, ?, ?, ?, ?, ?, ?)`)
 	if err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


protocol/communities/persistence.go
```diff
@@ -1632,3 +1632,70 @@ func (p *Persistence) SaveRequestsToJoin(requests []*RequestToJoin) (err error)
...
+		if featured {
+			result.ContractFeaturedCommunities = append(result.ContractFeaturedCommunities, communityID)
+		}
+	}
+
+	return result, nil
+}
+
+func (p *Persistence) SetCuratedCommunities(communities *CuratedCommunities) error {
+	tx, err := p.db.BeginTx(context.Background(), &sql.TxOptions{})
```
MishkaRogachev, line 1663: i wish we can boilerplate transaction someday (not a comment, just sharing)

protocol/communities/persistence.go
```diff
@@ -1632,3 +1632,70 @@ func (p *Persistence) SaveRequestsToJoin(requests []*RequestToJoin) (err error)
...
+	}
+	defer stmt.Close()
+
+	featuredMap := make(map[string]bool)
+	for _, community := range communities.ContractFeaturedCommunities {
+		featuredMap[community] = true
+	}
+
+	for _, community := range communities.ContractCommunities {
+		_, err := stmt.Exec(community, featuredMap[community])
```
MishkaRogachev, line 1694: if we save all curated communities as featured, can it be just `_, err := stmt.Exec(community, true)` ?

MishkaRogachev (commented):
Great work!  Just one question about saving curated communities..

protocol/messenger_curated_communities.go
```diff
@@ -0,0 +1,147 @@
...
+				}
+
+				err = m.communitiesManager.SetCuratedCommunities(curatedCommunities)
+				if err == nil {
+					cache = curatedCommunities
+				} else {
+					logger.Error("failed to save curated communities", zap.Error(err))
+				}
+
+				response, err := m.fetchCuratedCommunities(curatedCommunities)
```
jrainville, line 59: Can we just pass `curatedCommunities` to `SendCuratedCommunitiesUpdate`. Maybe we can just wrap it in the Response?

jrainville (commented):
Nice job. Just one small thing

protocol/communities/persistence.go
osmaczko: We don't save all curated communities as featured, see how `featuredMap` is propagated, am I missing something :thinking: ?

protocol/communities/persistence.go
osmaczko: Wdym exactly?

protocol/messenger_curated_communities.go
osmaczko: Not sure I am following :thinking:  The response is a type crafted specifically for clients, including fetched community data, not only the IDs of curated and featured communities.

protocol/communities/persistence.go
MishkaRogachev: we have this code repeated dozens of times. Just curious, if it's possible to boilerplate it 
```
tx, err := p.db.BeginTx(context.Background(), &sql.TxOptions{})

if err != nil {
		return err
	}

defer func() {
		if err == nil {
			err = tx.Commit()
			return
		}
		// don't shadow original error
		_ = tx.Rollback()
	}()
```

protocol/communities/persistence.go
MishkaRogachev: Ah, i see, it's `communities.ContractFeaturedCommunities`

MishkaRogachev (approved):

protocol/communities/persistence.go
osmaczko: Ah, would be nice indeed. The most obvious would be to use helper function:

```
defer func() {
    FinalizeTransaction(tx, &err)
}()
```

PR #4321, opened 2023-11-15
Community member joins the community if he sends a request to join an…

ibrkhalil:
…d cancels it and admin is offline during both

An issue that happened when user requested to join a community and cancels his request while admin is offline, Admin received first message and allowed the user in, But the user had already cancelled his request, So we remove him from the community.

[Closes #17576 on status-mobile](https://github.com/status-im/status-mobile/issues/17576)


protocol/communities_messenger_test.go
```diff
@@ -3684,3 +3684,170 @@ func (s *MessengerCommunitiesSuite) TestRetrieveBigCommunity() {
...
+		if err != nil {
+			return err
+		}
+		if len(response.Communities()) == 0 {
+			return errors.New("message not received")
+		}
+		return nil
+	})
+
+	s.Require().NoError(err)
```
osmaczko:
```suggestion
	community, _ := s.createCommunity()
	s.advertiseCommunityTo(community, s.owner, s.alice)
```

protocol/communities_messenger_test.go
```diff
@@ -3684,3 +3684,170 @@ func (s *MessengerCommunitiesSuite) TestRetrieveBigCommunity() {
...
+
+	// Retrieve cancel request to join to owner
+	err = tt.RetryWithBackOff(func() error {
+		response, err = s.owner.RetrieveAll()
+		if err != nil {
+			return err
+		}
+		if len(response.RequestsToJoinCommunity) == 0 {
+			return errors.New("request to join community not received")
+		}
```
osmaczko: That may cause unexpected results, with this condition, there is no guarantee both requests will be received.

protocol/communities_messenger_test.go
```diff
@@ -3684,3 +3684,170 @@ func (s *MessengerCommunitiesSuite) TestRetrieveBigCommunity() {
...
+	s.Require().Equal(communities.RequestToJoinStateCanceled, response.RequestsToJoinCommunity[0].State)
+
+	// pull to make sure it has been saved
+	cancelRequestsToJoin, err := s.alice.MyCanceledRequestsToJoin()
+	s.Require().NoError(err)
+	s.Require().Len(cancelRequestsToJoin, 1)
+	s.Require().Equal(cancelRequestsToJoin[0].State, communities.RequestToJoinStateCanceled)
+
+	// Retrieve cancel request to join to owner
+	err = tt.RetryWithBackOff(func() error {
```
osmaczko: Please use `WaitOnMessengerResponse` instead, it is superior to `RetryWithBackOff` because it merges the responses.

protocol/communities_messenger_test.go
```diff
@@ -3684,3 +3684,170 @@ func (s *MessengerCommunitiesSuite) TestRetrieveBigCommunity() {
...
+	var response2 *MessengerResponse
+	// Fetch new messages for Alice
+	err = tt.RetryWithBackOff(func() error {
+		response2, err = s.alice.RetrieveAll()
+		if err != nil {
+			return err
+		}
+		if len(response2.Communities()) == 0 {
+			return errors.New("communities not received")
+		}
```
osmaczko: There are no guarantees in which order community updates will be retrieved by alice. This may be the first update where alice is still a member of a community.

protocol/messenger_handler.go
```diff
@@ -1443,10 +1443,38 @@ func (m *Messenger) HandleCommunityCancelRequestToJoin(state *ReceivedMessageSta
...
+			senderID,
+		)
+		if err != nil {
+			return err
+		}
+
+		if len(response.Communities()) > 0 {
+			state.Response.AddCommunity(response.Communities()[0])
+		}
+	}
```
osmaczko: I am wondering why not to put this logic directly into `communitiesManager.HandleCommunityCancelRequestToJoin`?

We would avoid double community reading from db -> `community, err := m.communitiesManager.GetByID(cancelRequestToJoinProto.CommunityId)`

osmaczko: I believe we should compare the clocks of community requests before removing a member from the community. Let's consider the following:

User actions:
RequestToJoin, clock: 1
CancelRequestToJoin, clock: 2
RequestToJoin, clock: 3

The control node comes back online and receives the requests in the following order:
RequestToJoin, clock: 1
RequestToJoin, clock: 3
CancelRequestToJoin, clock: 2

The user would be removed from the community even when they shouldn't be. To mitigate this, we should handle cancel requests only if they are newer than the latest request to join.

protocol/communities_messenger_test.go
```diff
@@ -3684,3 +3684,170 @@ func (s *MessengerCommunitiesSuite) TestRetrieveBigCommunity() {
 	}, "updated description not received")
 	s.Require().NoError(err)
 }
+
+func (s *MessengerCommunitiesSuite) TestRequestAndCancelCommunityAdminOffline() {
```
osmaczko, line 3778: I checked myself and the test seems to be flaky (possible culprits in other comments). It is crucial to have in mind that _messages in status-go tests may come out of order_.

In order to test if test is flaky you could use `-count=20`:
`go test -timeout 60s -tags gowaku_skip_migrations -count=20 -run ^TestMessengerCommunitiesSuite$ -testify.m ^\(TestRequestAndCancelCommunityAdminOffline\)$ github.com/status-im/status-go/protocol`

osmaczko (changes requested):
Thanks for the PR!

There are few issues I identified.

ibrkhalil:
> Thanks for the PR!
> 
> There are few issues I identified.

I'll convert it to a draft and address the comments, Thanks!

protocol/messenger_handler.go
ibrkhalil: I refactored it to only fetch the community data once.

ibrkhalil:
@osmaczko Can you please take another look, Thanks!

protocol/communities/persistence.go
```diff
@@ -741,6 +741,24 @@ func (p *Persistence) CanceledRequestsToJoinForUser(pk string) ([]*RequestToJoin
...
+func (p *Persistence) CanceledRequestToJoinForUserForCommunityID(pk string, communityID []byte) (*RequestToJoin, error) {
+
+	rows, err := p.db.Query(`SELECT id,public_key,clock,ens_name,chat_id,community_id,state FROM communities_requests_to_join WHERE state = ? AND public_key = ? AND community_id = ?`, RequestToJoinStateCanceled, pk, communityID)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	request := &RequestToJoin{}
+	for rows.Next() {
```
ilmotta: > You should always check for an error at the end of the for rows.Next() loop. If there’s an error during the loop, you need to know about it. Don’t just assume that the loop iterates until you’ve processed all the rows.
>
> -- http://go-database-sql.org/retrieving.html

protocol/communities/persistence.go
```diff
@@ -741,6 +741,24 @@ func (p *Persistence) CanceledRequestsToJoinForUser(pk string) ([]*RequestToJoin
 	return requests, nil
 }
 
+func (p *Persistence) CanceledRequestToJoinForUserForCommunityID(pk string, communityID []byte) (*RequestToJoin, error) {
+
+	rows, err := p.db.Query(`SELECT id,public_key,clock,ens_name,chat_id,community_id,state FROM communities_requests_to_join WHERE state = ? AND public_key = ? AND community_id = ?`, RequestToJoinStateCanceled, pk, communityID)
```
ilmotta: Raw literal strings can be multiline. Such long lines are unnecessary.

```go
rows, err := p.db.Query(`
  SELECT id, public_key, clock, ens_name, chat_id, community_id, state
  FROM communities_requests_to_join
  WHERE state = ? AND public_key = ? AND community_id = ?`,
  RequestToJoinStateCanceled, pk, communityID)
```

protocol/communities/persistence.go
```diff
@@ -741,6 +741,24 @@ func (p *Persistence) CanceledRequestsToJoinForUser(pk string) ([]*RequestToJoin
 	return requests, nil
 }
 
+func (p *Persistence) CanceledRequestToJoinForUserForCommunityID(pk string, communityID []byte) (*RequestToJoin, error) {
+
```
ilmotta, line 698: Extra line

protocol/messenger_handler.go
```diff
@@ -1442,7 +1442,41 @@ func (m *Messenger) HandleCommunityCancelRequestToJoin(state *ReceivedMessageSta
...
+		response, err := m.RemoveUserFromCommunity(
+			community.ID(),
+			senderID,
+		)
+		if err != nil {
+			return err
+		}
+
+		if len(response.Communities()) > 0 {
+			state.Response.AddCommunity(response.Communities()[0])
```
ilmotta: `RemoveUserFromCommunity` returns a community using `GetByID`, but the `community` instance returned by `m.communitiesManager.GetByID` a few lines above is already fetched from the DB, so I don't understand why we need to check if `response` contains a community, it's redundant.

Difficult to explain, here's the diff:

```diff
modified   protocol/messenger_handler.go
@@ -1443,10 +1443,10 @@ func (m *Messenger) HandleCommunityCancelRequestToJoin(state *ReceivedMessageSta
 	}
 
 	community, err := m.communitiesManager.GetByID(cancelRequestToJoinProto.CommunityId)
-
 	if err != nil {
 		return err
 	}
+	state.Response.AddCommunity(community)
 
 	senderID := state.CurrentMessageState.Contact.ID
 	senderPubKey, err := common.HexToPubkey(senderID)
@@ -1462,17 +1462,10 @@ func (m *Messenger) HandleCommunityCancelRequestToJoin(state *ReceivedMessageSta
 	}
 
 	if community.HasMember(senderPubKey) && cancelRequestToJoinProto.Clock > previousRequestToJoin.Clock {
-		response, err := m.RemoveUserFromCommunity(
-			community.ID(),
-			senderID,
-		)
+		_, err := m.RemoveUserFromCommunity(community.ID(), senderID)
 		if err != nil {
 			return err
 		}
-
-		if len(response.Communities()) > 0 {
-			state.Response.AddCommunity(response.Communities()[0])
-		}
 	}
 
 	requestToJoin, err := m.communitiesManager.HandleCommunityCancelRequestToJoin(signer, cancelRequestToJoinProto, community)
```

protocol/messenger_handler.go
```diff
@@ -1659,26 +1699,28 @@ func (m *Messenger) HandleCommunityRequestToJoinResponse(state *ReceivedMessageS
...
+				notification.MembershipStatus = ActivityCenterMembershipStatusAccepted
+				notification.Read = false
+				notification.Deleted = false
+			} else {
+				notification.MembershipStatus = ActivityCenterMembershipStatusDeclined
+			}
+			notification.UpdatedAt = m.GetCurrentTimeInMillis()
+			err = m.addActivityCenterNotification(state.Response, notification, nil)
+			if err != nil {
+				m.logger.Warn("failed to update notification", zap.Error(err))
```
ilmotta: The errors that can happen inside `addActivityCenterNotification` are logged at the error level, and should rarely happen, so it's reasonable this here should also be logged as an error.

protocol/communities/persistence.go
ibrkhalil: Isn't that what line 757-759 doing already?

protocol/messenger_handler.go
ibrkhalil: Wouldn't we want an indicator that a member has been removed in the response?

protocol/communities/persistence.go
ilmotta: The quote is about checking for errors at the end of the loop. The check you mention is done on a row by row basis, it's different.

See the code in section `Fetching Data from the Database`.

ibrkhalil: Thanks Icaro, Done

ilmotta: Thanks!

protocol/messenger_handler.go
ilmotta: Oh I see. So you are adding the community to mean "member removed"? If that's the case then I think it's really hard to grasp this because it's not obvious why having a community in the response is related.

But do we really need an indicator that the member has been removed? Do you mean the clients will use this information? Because if clients are not using this, than no code is better than any code we would come up with.

ibrkhalil: I think this response returns to the admin/privliged user of the community, And I think it might be necessary to let the admin client know someone is no longer a member, But it's just an assumption.
WDYT Icaro?

protocol/communities/manager.go
```diff
@@ -2381,11 +2381,7 @@ func (m *Manager) shouldUserRetainDeclined(signer *ecdsa.PublicKey, community *C
...
 
-func (m *Manager) HandleCommunityCancelRequestToJoin(signer *ecdsa.PublicKey, request *protobuf.CommunityCancelRequestToJoin) (*RequestToJoin, error) {
-	community, err := m.GetByID(request.CommunityId)
-	if err != nil {
-		return nil, err
-	}
+func (m *Manager) HandleCommunityCancelRequestToJoin(signer *ecdsa.PublicKey, request *protobuf.CommunityCancelRequestToJoin, community *Community) (*RequestToJoin, error) {
 	if community == nil {
 		return nil, ErrOrgNotFound
 	}
```
osmaczko: This doesn't make sense anymore, withouth the context of `GetByID`. Should be checked on the caller side of this func.

protocol/communities/manager.go
```diff
@@ -2381,11 +2381,7 @@ func (m *Manager) shouldUserRetainDeclined(signer *ecdsa.PublicKey, community *C
 	return request.ShouldRetainDeclined(requestClock)
 }
 
-func (m *Manager) HandleCommunityCancelRequestToJoin(signer *ecdsa.PublicKey, request *protobuf.CommunityCancelRequestToJoin) (*RequestToJoin, error) {
-	community, err := m.GetByID(request.CommunityId)
-	if err != nil {
-		return nil, err
-	}
+func (m *Manager) HandleCommunityCancelRequestToJoin(signer *ecdsa.PublicKey, request *protobuf.CommunityCancelRequestToJoin, community *Community) (*RequestToJoin, error) {
```
```suggestion
func (m *Manager) HandleCommunityCancelRequestToJoin(community *Community, signer *ecdsa.PublicKey, request *protobuf.CommunityCancelRequestToJoin) (*RequestToJoin, error) {
```

protocol/messenger_handler.go
```diff
@@ -1442,7 +1442,41 @@ func (m *Messenger) HandleCommunityCancelRequestToJoin(state *ReceivedMessageSta
...
+	}
+
+	previousRequestToJoin, err := m.communitiesManager.GetRequestToJoinByPkAndCommunityID(senderPubKey, community.ID())
+
+	if err != nil {
+		return err
+	}
+
+	if community.HasMember(senderPubKey) && cancelRequestToJoinProto.Clock > previousRequestToJoin.Clock {
+		response, err := m.RemoveUserFromCommunity(
```
osmaczko: There is still an issue with fetching community multiple times (RemoveUserFromCommunity fetches community by ID). Isn't it possible to move all this logic to `communitiesManager.HandleCommunityCancelRequestToJoin` as suggested previously?

protocol/messenger_handler.go
```diff
@@ -1603,7 +1643,7 @@ func (m *Messenger) HandleCommunityRequestToJoinResponse(state *ReceivedMessageS
 		state.Response.RequestsToJoinCommunity = append(state.Response.RequestsToJoinCommunity, updatedRequest)
 	}
 
-	if requestToJoinResponseProto.Accepted {
+	if requestToJoinResponseProto.Accepted && myCancelledRequestToJoin.State != communities.RequestToJoinStateCanceled {
```
osmaczko: Not sure I understand why we need `myCancelledRequestToJoin.State != communities.RequestToJoinStateCanceled`. Isn't `MyCanceledRequestToJoinForCommunityID` returning only requests in cancelled state?

protocol/communities_messenger_test.go
```diff
@@ -3684,3 +3684,115 @@ func (s *MessengerCommunitiesSuite) TestRetrieveBigCommunity() {
...
+	cancelRequestToJoin2 := response.RequestsToJoinCommunity[0]
+	s.Require().NotNil(cancelRequestToJoin2)
+	s.Require().Equal(community.ID(), cancelRequestToJoin2.CommunityID)
+	s.Require().False(cancelRequestToJoin2.Our)
+	s.Require().NotEmpty(cancelRequestToJoin2.ID)
+	s.Require().NotEmpty(cancelRequestToJoin2.Clock)
+	s.Require().Equal(cancelRequestToJoin2.PublicKey, common.PubkeyToHex(&s.alice.identity.PublicKey))
+	if cancelRequestsToJoin.Clock > requestToJoin1.Clock {
+		s.Require().False(response.Communities()[0].HasMember(&s.alice.identity.PublicKey))
+	}
```
osmaczko: It seems this condition makes the test not deterministic? Sometimes it checks what we are interested in (alice not being a member of community), the other time it is not.

protocol/communities_messenger_test.go
```diff
@@ -3684,3 +3684,115 @@ func (s *MessengerCommunitiesSuite) TestRetrieveBigCommunity() {
...
+
+	// pull to make sure it has been saved
+	cancelRequestsToJoin, err := s.alice.MyCanceledRequestToJoinForCommunityID(community.ID())
+	s.Require().NoError(err)
+	s.Require().NotNil(cancelRequestsToJoin)
+	s.Require().Equal(cancelRequestsToJoin.State, communities.RequestToJoinStateCanceled)
+
+	response, err = WaitOnMessengerResponse(
+		s.owner,
+		func(r *MessengerResponse) bool { return len(r.RequestsToJoinCommunity) >= 1 },
```
osmaczko: This seems to be wrong. Shouldn't owner receive 2 requests to join?

protocol/communities_messenger_test.go
```diff
@@ -3684,3 +3684,115 @@ func (s *MessengerCommunitiesSuite) TestRetrieveBigCommunity() {
...
+		"no requests to join community",
+	)
+
+	s.Require().NoError(err)
+
+	_, err = WaitOnMessengerResponse(
+		s.alice,
+		func(r *MessengerResponse) bool { return r != nil },
```
osmaczko: It does not check anything, is that intended?

osmaczko (changes requested):
Thanks for addressing previously found issues.

protocol/messenger_handler.go
osmaczko: Sorry, I don't see how it was resolved?

protocol/messenger_handler.go
ilmotta: I'm on the fence because I prefer to only add code in status-go that at least one client will use. If we're not sure this is going to be used by clients, then I would remove it.

protocol/communities_messenger_test.go
ibrkhalil: I thought it ensured correct message order.
Thinking of a workaround.

protocol/communities_messenger_test.go
ibrkhalil: Yes, But due to the issues we currently have with message reliability I didn't want this test to fail because of it.
In some cases (And it happens often, Two requests to join are not sent to Admin)

protocol/messenger_handler.go
ibrkhalil: If the user has cancelled his request and we have that value locally, Even if admin auto-approved him we shouldn't make him a member.

protocol/messenger_handler.go
ibrkhalil: Got it, Will do.
Thanks Icaro!

protocol/communities_messenger_test.go
ibrkhalil: Done.

protocol/messenger_handler.go
ibrkhalil: Done

protocol/communities/manager.go
ibrkhalil: Done

protocol/communities/manager.go
ibrkhalil: Done

protocol/communities/manager.go
```diff
@@ -2386,9 +2386,6 @@ func (m *Manager) HandleCommunityCancelRequestToJoin(signer *ecdsa.PublicKey, re
 	if err != nil {
 		return nil, err
 	}
-	if community == nil {
-		return nil, ErrOrgNotFound
-	}
```
osmaczko: Since `GetByID` has been reintroduced to this function, this check should also be included.

protocol/communities/manager.go
```diff
@@ -2408,6 +2410,18 @@ func (m *Manager) HandleCommunityCancelRequestToJoin(signer *ecdsa.PublicKey, re
 		return nil, err
 	}
 
+	if community.HasMember(signer) && request.Clock > previousRequestToJoin.Clock {
```
osmaczko: I think it makes sense to check the clock a bit earlier, before we mark request to join as canceled.
```
if request.Clock <= previousRequestToJoin.Clock {
    return nil, ErrInvalidClock
}
```

protocol/messenger_handler.go
osmaczko: Can `myCancelledRequestToJoin` be in different state than `RequestToJoinStateCanceled` ?

ibrkhalil:
https://github.com/status-im/status-go/pull/4321#discussion_r1405848075

If user didn't cancel his request, It'd be in another state than RequestToJoinStateCanceled

ibrkhalil:
Kindly take another look @osmaczko

protocol/communities/manager.go
```diff
@@ -2398,6 +2406,10 @@ func (m *Manager) HandleCommunityCancelRequestToJoin(signer *ecdsa.PublicKey, re
 		return nil, ErrCommunityRequestAlreadyRejected
 	}
 
+	if err != nil {
+		return nil, err
+	}
```
osmaczko: leftover?

protocol/messenger_handler.go
```diff
@@ -1594,21 +1595,29 @@ func (m *Messenger) HandleCommunityRequestToJoinResponse(state *ReceivedMessageS
...
+	myCancelledRequestToJoin, err := m.MyCanceledRequestToJoinForCommunityID(requestToJoinResponseProto.CommunityId)
+
 	if err != nil {
 		return err
 	}
 
-	if updatedRequest != nil {
-		state.Response.RequestsToJoinCommunity = append(state.Response.RequestsToJoinCommunity, updatedRequest)
+	// Updating the request if request it was cancelled breaks the test
+	if myCancelledRequestToJoin == nil {
```
osmaczko: Since `m.MyCanceledRequestToJoinForCommunityID` returns an error if there are no rows, the `if myCancelledRequestToJoin == nil {` will never be reached in that case, is that right?

protocol/communities/persistence.go
```diff
@@ -741,6 +741,33 @@ func (p *Persistence) CanceledRequestsToJoinForUser(pk string) ([]*RequestToJoin
...
+	}
+	defer rows.Close()
+
+	request := &RequestToJoin{}
+	for rows.Next() {
+		err = rows.Scan(&request.ID, &request.PublicKey, &request.Clock, &request.ENSName, &request.ChatID, &request.CommunityID, &request.State)
+		if err != nil {
+			return nil, err
+		}
+	}
```
osmaczko: Why do you iterate over the rows if the return value is not a collection of RequestToJoins? QueryRow should be used instead. Please pay attention to: `sql.ErrNoRows` error type, in this case we shouldn't return an error to the client.

osmaczko (commented):
Nice test rework :muscle:

protocol/messenger_handler.go
ibrkhalil: I just tested it, It returns `nil`

protocol/communities/persistence.go
ibrkhalil: Fixed

protocol/communities_messenger_test.go
ibrkhalil: Apologies, You were right.
Fixed :)

protocol/messenger_handler.go
ibrkhalil: You were right, Fixed.

protocol/communities/persistence.go
```diff
@@ -741,6 +741,24 @@ func (p *Persistence) CanceledRequestsToJoinForUser(pk string) ([]*RequestToJoin
...
+	WHERE
+	state = ? AND public_key = ? AND community_id = ?`,
+		RequestToJoinStateCanceled, pk, communityID)
+
+	request := &RequestToJoin{}
+
+	err := row.Scan(&request.ID, &request.PublicKey, &request.Clock, &request.ENSName, &request.ChatID, &request.CommunityID, &request.State)
+	if err != nil && err != sql.ErrNoRows {
+		return nil, err
+	}
```
osmaczko: I believe in case of `sql.ErrNoRows` we should return nil.

```suggestion
	if err == sql.ErrNoRows {
		return nil, nil
	} else if err != nil {
		return nil, err
	}
```

protocol/messenger_handler.go
```diff
@@ -1594,21 +1595,29 @@ func (m *Messenger) HandleCommunityRequestToJoinResponse(state *ReceivedMessageS
...
+	myCancelledRequestToJoin, err := m.MyCanceledRequestToJoinForCommunityID(requestToJoinResponseProto.CommunityId)
+
 	if err != nil {
 		return err
 	}
 
-	if updatedRequest != nil {
-		state.Response.RequestsToJoinCommunity = append(state.Response.RequestsToJoinCommunity, updatedRequest)
+	// Updating the request if request it was cancelled breaks the test
+	if myCancelledRequestToJoin.State == 0 {
```
osmaczko: 1) Please use `nil` assertion as suggested before. Comparing State to 0 is not readable at all.
2) Please use early return instead of populating each block with this condition.

```suggestion
	// Don't process response if we cancelled request to join.
	if myCancelledRequestToJoin != nil {
            return nil
    }
```

osmaczko (approved):

protocol/communities_messenger_test.go
```diff
@@ -3774,3 +3774,167 @@ func (s *MessengerCommunitiesSuite) TestRetrieveBigCommunity() {
...
+	s.Require().NotNil(requestToJoin1)
+	s.Require().Equal(community.ID(), requestToJoin1.CommunityID)
+	s.Require().True(requestToJoin1.Our)
+	s.Require().NotEmpty(requestToJoin1.ID)
+	s.Require().NotEmpty(requestToJoin1.Clock)
+	s.Require().Equal(requestToJoin1.PublicKey, common.PubkeyToHex(&s.alice.identity.PublicKey))
+	s.Require().Equal(communities.RequestToJoinStatePending, requestToJoin1.State)
+
+	// Make sure clock is not empty
+	s.Require().NotEmpty(requestToJoin1.Clock)
```
ilmotta: This check is already done a few lines before.

protocol/communities_messenger_test.go
```diff
@@ -3774,3 +3774,167 @@ func (s *MessengerCommunitiesSuite) TestRetrieveBigCommunity() {
...
+		Clock:       cancelRequestsToJoin.Clock,
+		CommunityId: community.ID(),
+		Accepted:    true,
+	}
+
+	err = s.alice.HandleCommunityRequestToJoinResponse(messageState, requestToJoinResponseProto, &statusMessage)
+	s.Require().NoError(err)
+	// Make sure alice is NOT a member of the community that she cancelled her request to join to
+	s.Require().False(community.HasMember(s.alice.IdentityPublicKey()))
+	// Make sure there's no PNs for Alice
```
ilmotta: The comment should probably be `Make sure there are no AC notifications for Alice`, not `PNs`, which means something else.

ilmotta (approved):

PR #4322, opened 2023-11-15
feat: wallet - emoji and color

OmarBasem:
This PR sets customization color and emoji for the wallet account when creating a Status account

status-mobile PR: https://github.com/status-im/status-mobile/pull/17906


qfrank (approved):
LGTM

cammellos:
tests please :)

cammellos:
@OmarBasem one thing, is the user passing  the emoji from the client or you are just choosing a random one in the client?
If you are just choosing a random one, maybe it makes more sense to generate on status-go, the least choices the frontend makes the better, but up to you

OmarBasem:
> @OmarBasem one thing, is the user passing the emoji from the client or you are just choosing a random one in the client? If you are just choosing a random one, maybe it makes more sense to generate on status-go, the least choices the frontend makes the better, but up to you

@cammellos The emoji is passed from the client. To generate it on status-go we would need to have all of the emojis data on status-go

cammellos:
@OmarBasem oh I see, and I take that's something we don't have already, that's cool then, thank you

vkjr (approved):

PR #4323, opened 2023-11-15
chore: cleanup message segments

### LINKED ISSUE #4297: Clean up incomplete message segments
- [x] clean up the `message_segments` table in case segments aren't completed in X days/hours, because it would occupy  a lot of memory
- [x] cleanup `message_segments_completed` table periodically, because retaining completed segment hashes that are older than the lifespan of store node messages is pointless

https://github.com/status-im/status-go/pull/4285#discussion_r1389925143

osmaczko:
closes: #4297

### DIFF
--- a/protocol/common/raw_messages_persistence.go
+++ b/protocol/common/raw_messages_persistence.go
@@ -347,11 +347,11 @@ func (db *RawMessagesPersistence) IsMessageAlreadyCompleted(hash []byte) (bool, 
 	return alreadyCompleted > 0, nil
 }
 
-func (db *RawMessagesPersistence) SaveMessageSegment(segment *protobuf.SegmentMessage, sigPubKey *ecdsa.PublicKey) error {
+func (db *RawMessagesPersistence) SaveMessageSegment(segment *protobuf.SegmentMessage, sigPubKey *ecdsa.PublicKey, timestamp int64) error {
 	sigPubKeyBlob := crypto.CompressPubkey(sigPubKey)
 
-	_, err := db.db.Exec("INSERT INTO message_segments (hash, segment_index, segments_count, sig_pub_key, payload) VALUES (?, ?, ?, ?, ?)",
-		segment.EntireMessageHash, segment.Index, segment.SegmentsCount, sigPubKeyBlob, segment.Payload)
+	_, err := db.db.Exec("INSERT INTO message_segments (hash, segment_index, segments_count, sig_pub_key, payload, timestamp) VALUES (?, ?, ?, ?, ?, ?)",
+		segment.EntireMessageHash, segment.Index, segment.SegmentsCount, sigPubKeyBlob, segment.Payload, timestamp)
 
 	return err
 }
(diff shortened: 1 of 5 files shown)


cammellos:
@osmaczko what happens if messages are coming from a store node? they should have an old timestamp, and they might get cleaned up too quickly, unless I am mistaking. Maybe we should use just time of insertion?

osmaczko:
> @osmaczko what happens if messages are coming from a store node? they should have an old timestamp, and they might get cleaned up too quickly, unless I am mistaking. Maybe we should use just time of insertion?

EDITED:
Well spotted, it makes more sense, will update, thanks :+1:

osmaczko:
> > @osmaczko what happens if messages are coming from a store node? they should have an old timestamp, and they might get cleaned up too quickly, unless I am mistaking. Maybe we should use just time of insertion?
> 
> EDITED: Well spotted, it makes more sense, will update, thanks 👍

@cammellos done

protocol/common/raw_messages_persistence.go
```diff
@@ -405,10 +410,15 @@ func (db *RawMessagesPersistence) CompleteMessageSegments(hash []byte, sigPubKey
...
+	_, err = tx.Exec("INSERT INTO message_segments_completed (hash, sig_pub_key, timestamp) VALUES (?,?,?)", hash, sigPubKeyBlob, timestamp)
 	if err != nil {
 		return err
 	}
 
 	return err
 }
+
+func (db *RawMessagesPersistence) RemoveMessageSegmentsCompletedOlderThan(timestamp int64) error {
+	_, err := db.db.Exec("DELETE FROM message_segments_completed WHERE timestamp < ?", timestamp)
```
cammellos, line 422: is there an index on timestamp?

osmaczko: it is now -> https://github.com/status-im/status-go/compare/4a75ec896b03a9c6dee7c1ae997f857ba66dd05f..146c4dbb478a4ff2fb7da34433ccc3af4e2ae99e

protocol/messenger.go
```diff
@@ -5724,3 +5725,26 @@ func (m *Messenger) handleSyncSocialLinks(message *protobuf.SyncSocialLinks, cal
...
+func (m *Messenger) startMessageSegmentsCleanupLoop() {
+	logger := m.logger.Named("messageSegmentsCleanupLoop")
+
+	go func() {
+		var interval time.Duration = 0
+		for {
+			select {
+			case <-time.After(interval):
+				// Immediate execution on first run, then set to regular interval
+				interval = 1 * time.Hour
```
jrainville, line 5738: Do we really need a loop? Just doing it once per launch should be enough? I guess in case someone never closes the app it's good :laughing:

jrainville (approved):
LGTM

protocol/messenger.go
cammellos: I think that's both ok, the one thing we might not want to do, is to schedule it right at launch time, I would give it say a minute or two, since at start up there's a lot happening, the drawback is that if someone just opens it for 59 seconds it might not run, but I think that's acceptable.

osmaczko: Valid point. Addressed: https://github.com/status-im/status-go/pull/4328

PR #4324, opened 2023-11-15
fix: add option to preserve history on chat delete

BalogunofAfrica:
The `Delete Chat` action was updated to `Close Chat` which is meant to preserve history. This PR adds the option to preserve the history when calling `DeactivateChat`.

Related mobile issue: https://github.com/status-im/status-mobile/issues/17912

### DIFF
--- a/protocol/messenger_chats.go
+++ b/protocol/messenger_chats.go
@@ -383,7 +383,9 @@ func (m *Messenger) DeactivateChat(request *requests.DeactivateChat) (*Messenger
 		return nil, err
 	}
 
-	return m.deactivateChat(request.ID, 0, true, true)
+	doClearHistory := !request.PreserveHistory
+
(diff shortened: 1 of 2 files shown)


ghost:
Hey @BalogunofAfrica, and thank you so much for making your first pull request in status-go! :heart: Please help us make your experience better by filling out this brief questionnaire https://goo.gl/forms/uWqNcVpVz7OIopXg2

jrainville (approved):
LGTM.
I assume `PreserveHistory` defaults at false, so if we want the same behaviour on desktop, we need to explicitly pass it at `true`?

BalogunofAfrica:
> LGTM. I assume `PreserveHistory` defaults at false, so if we want the same behaviour on desktop, we need to explicitly pass it at `true`?

Yes

qfrank (approved):

Issue #4200, opened 2023-10-25
Avoid transfers scanning for wallet accounts which have never been used


rasom:
related to https://github.com/status-im/status-go/issues/3833 

We don't need to check history per account if
1) the multiacc was generated and 12 words mnemonic has never been shown
2) the address of the account wasn't shown/copied/shared

tbd how syncing affects it

rasom:
places where address is shown/shared so far

Communities

Sharing addresses when joining a community:
https://www.figma.com/file/qHfFm7C9LwtXpfdbxssCK3/Kuba⎜Desktop---Communities?type=design&node-id=31461-563897&mode=design&t=i4DlMVX8BtMt3xe9-4

Wallet Settings

User has full visibility over their OWN full account strings anywhere in Wallet settings
https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/⚙%EF%B8%8F-Settings-%7C-Desktop?type=design&node-id=18442-536894&mode=design&t=bHA4aqootsgXZ99T-4

context menu allows copying address and showing address QR:
https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/⚙%EF%B8%8F-Settings-%7C-Desktop?type=design&node-id=18442-529258&mode=design&t=bHA4aqootsgXZ99T-4

Wallet
Address QR dialog
https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=13049-456457&mode=design&t=RaV4PRYApXbS7Lfg-4

Copying address from right-click of account list https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?type=design&node-id=13909-410577&mode=design&t=RaV4PRYApXbS7Lfg-4

PR #4325, opened 2023-11-15
[#4200] Avoid tx watching for addresses which were not exposed

rasom:
needed for #4200

Client side changes will be added separately, meanwhile is it hidden behind the `mnemonicCheckEnabled` flag, so the wallet works as usually.

- columns/props are named `MnemonicWasNotShown`/`AddressWasNotShown`, negation is used, to make sure that when it default to `false` everything still works as expected
- when local pairing happens `MnemonicWasNotShown` is set to `true`, tracking it after pairing might be too complicated
- `MnemonicWasNotShown` is false for keycard accounts, also for accounts which were restored via mnemonic


saledjenic:
> when local pairing happens MnemonicWasNotShown is set to true, tracking it after pairing might be too complicated

@rasom this depends if the mnemonic was revealed on the device the user is syncing from or not. So not always true, not always false.

api/geth_backend.go
```diff
@@ -1352,6 +1352,7 @@ func (b *GethStatusBackend) generateOrImportAccount(mnemonic string, customizati
 	if mnemonic == "" {
 		settings.Mnemonic = &info.Mnemonic
 		settings.OmitTransfersHistoryScan = true
+		settings.MnemonicWasNotShown = true
```
saledjenic: If the mnemonic is empty, means we're restoring an account using the seed phrase, which implies that the seed phrase is known, doesn't that need to set `settings.MnemonicWasNotShown` to `false` instead?

api/geth_backend.go
```diff
@@ -1377,6 +1378,10 @@ func (b *GethStatusBackend) generateOrImportAccount(mnemonic string, customizati
 		Name:      walletAccountDefaultName,
 	}
 
+	if mnemonic == "" {
+		walletAccount.AddressWasNotShown = true
```
saledjenic, line 1382: The same thing here.

multiaccounts/settings/columns.go
```diff
@@ -452,7 +452,11 @@ var (
 		dBColumnName:   "omit_transfers_history_scan",
 		valueHandler:   BoolHandler,
 	}
-
+	MnemonicWasNotShown = SettingField{
```
saledjenic, line 455: Do we need this field at all, cause non-empty `mnemonic` means the same, that mnemonic was not shown?

multiaccounts/settings/database.go
```diff
@@ -613,6 +616,14 @@ func (db *Database) MnemonicRemoved() (result bool, err error) {
 	return result, err
 }
 
+func (db *Database) GetMnemonicWasNotShown() (result bool, err error) {
+	err = db.makeSelectRow(MnemonicWasNotShown).Scan(&result)
```
saledjenic, line 620: As mentioned above, instead of adding new field we can just have this:
```
func (db *Database) GetMnemonicWasNotShown() (result bool, err error) {
	mnemonic, err := db.Mnemonic()
	if err != nil {
		return result, err 
	}
	result = len(mnemonic) > 0
	return
}
```

api/geth_backend.go
```diff
@@ -2294,6 +2299,19 @@ func (b *GethStatusBackend) GetActiveAccount() (*multiaccounts.Account, error) {
...
+	if b.account == nil {
+		return errors.New("master key account is nil in the GethStatusBackend")
+	}
+
+	accountDB, err := accounts.NewDB(b.appDB)
+	if err != nil {
+		return err
+	}
+
+	return accountDB.MnemonicWasShown()
```
saledjenic, line 2312: Covered in [this](https://github.com/status-im/status-go/pull/4325#issuecomment-1831926033).

api/geth_backend.go
rasom: on contrary, if this `mnemonic` is not empty then we are restoring an account, otherwise we will have it in `info` or something `¯\_(ツ)_/¯`

api/geth_backend.go
rasom: and same answer

multiaccounts/settings/columns.go
rasom: sure we do, because `MnemonicWasNotShown` should be set to false even if user has not finished backing up flow. So that if it had appeared on the screen it is wasted

multiaccounts/settings/database.go
rasom: answered there

rasom:
@saledjenic 
> @rasom this depends if the mnemonic was revealed on the device the user is syncing from or not. So not always true, not always false.

I realize that, but also tracking it after pairing seems a bit more complicated than I would want it to be, so if it ever will be needed we can work on it separately later. It is just simpler to mark it as seen when pairing happens.

api/geth_backend.go
saledjenic: I overlooked the function name, that's not the function the desktop app is using at all.

If it's as you said _"if this mnemonic is not empty then we are restoring an account"_ then all good.

multiaccounts/settings/columns.go
saledjenic: Yes, makes sense.

saledjenic:
> @saledjenic
> 
> > @rasom this depends if the mnemonic was revealed on the device the user is syncing from or not. So not always true, not always false.
> 
> I realize that, but also tracking it after pairing seems a bit more complicated than I would want it to be, so if it ever will be needed we can work on it separately later. It is just simpler to mark it as seen when pairing happens.

Sounds ok to me. It simply means that fetching history improvements will be applied only to the "main device", the one user used to create a Status account, but not for other devices synced from it.

saledjenic (approved):
@rasom don't forget to update the version file once you're about to merge it.

PR #4327, opened 2023-11-15
Explicitly set PubsubTopics when sending/subscribing to messages

vitvly:
During investigation of https://github.com/status-im/status-mobile/issues/17898 it turned out that:
  - messages are sent with pubsubTopic=`/waku/2/default-waku/proto`, which is set here: https://github.com/status-im/status-go/blob/a71c27b7f80c9879dda2169f5bed24572b800411/wakuv2/waku.go#L1054
  and possibly overridden in https://github.com/status-im/status-go/blob/a71c27b7f80c9879dda2169f5bed24572b800411/wakuv2/waku.go#L1016
  - messages are received via filter subscriptions, but clients subscribe to filters using sharded pubsubTopics like `/waku/2/rs/16/32`, set for instance here: https://github.com/status-im/status-go/blob/a71c27b7f80c9879dda2169f5bed24572b800411/protocol/transport/filters_manager.go#L388
  - overrides for pubsubTopics contained in received envelopes only work for relay, not light clients https://github.com/status-im/status-go/blob/a71c27b7f80c9879dda2169f5bed24572b800411/wakuv2/waku.go#L1319 as light clients are not going to receive those messages via subscriptions, them being removed at the filter server level

Most changed files are due to the removal of extra `transport.Shard` struct. Also `InitCommunityFilters()`/`DefaultFilters()` now consider `useShards` var so that not to create an extra filter for default non protected shards if shards are not used anyway.


richard-ramos (approved):

vitvly:
There are failures in `TestBecomeMemberPermissions` test, investigating. **EDIT**: found the issue, fixing.

richard-ramos (approved):
:rocket: :rocket: :rocket: :rocket: :rocket:

PR #4329, opened 2023-11-16
fix: send erc20

alaibe:


services/wallet/bridge/transfer.go
```diff
@@ -34,9 +41,45 @@ func (s *TransferBridge) CalculateFees(from, to *params.Network, token *token.To
 
 func (s *TransferBridge) EstimateGas(fromNetwork *params.Network, toNetwork *params.Network, from common.Address, to common.Address, token *token.Token, amountIn *big.Int) (uint64, error) {
 	// TODO fix for ERC20
```
J-Son89: is the TODO needed still?

alaibe: nope, fixed

dlipicar (approved):
Nice! I assume the estimation will be different for ERC721, we could add that one in the future

alaibe:
@dlipicar we already have it for erc721!

PR #4332, opened 2023-11-16
Feat/alchemy sepolia support

dlipicar:
Part of https://github.com/status-im/status-desktop/issues/12771

Enable support for Sepolia in the Alchemy client

Picking up keys for all chains, but only Eth and Arb are supported by Alchemy at the moment. Opt coming soon.

### DIFF
--- a/api/default_networks.go
+++ b/api/default_networks.go
@@ -9,14 +9,17 @@ import (
 )
 
 const (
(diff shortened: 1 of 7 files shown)


stefandunca (approved):
👍🏼

saledjenic (approved):

PR #4336, opened 2023-11-17
fix(wakuv2): logout / login delay

richard-ramos:
Fixes the issue described in https://github.com/status-im/status-mobile/issues/17899#issuecomment-1812584250 as well as reusing the IDService from go-waku instead of having status-go create its own, and also an error with a reference to a loop variable

### DIFF
--- a/protocol/transport/filters_manager.go
+++ b/protocol/transport/filters_manager.go
@@ -168,7 +168,8 @@ func (f *FiltersManager) InitCommunityFilters(communityFiltersToInitialize []Com
 		// TODO: requests to join / cancels are currently being sent into the default waku topic.
 		// They must be sent into an specific non protected shard
 		for _, pubsubTopic := range topics {
-			identityStr := PublicKeyToStr(&cf.PrivKey.PublicKey)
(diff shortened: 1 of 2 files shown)


protocol/transport/filters_manager.go
```diff
@@ -168,7 +168,8 @@ func (f *FiltersManager) InitCommunityFilters(communityFiltersToInitialize []Com
 		// TODO: requests to join / cancels are currently being sent into the default waku topic.
 		// They must be sent into an specific non protected shard
 		for _, pubsubTopic := range topics {
-			identityStr := PublicKeyToStr(&cf.PrivKey.PublicKey)
+			pk := &cf.PrivKey.PublicKey
```
vitvly, line 171: So here, given that we're taking an address of loop variable `cf`, this `pk` param would always point to the same var, correct?

vitvly (approved):

protocol/transport/filters_manager.go
richard-ramos: yeah!
Earlier I updated golangci-lint on my machine and when I executed `make lint` saw a warning!

PR #4340, opened 2023-11-18
Adding store wakuv2 test

### LINKED ISSUE #4317: Add test for WakuV2 Store
There is a `wakuv2.Config.EnableStore` flag that can be used to enable Store protocol on local Status node:
https://github.com/status-im/status-go/blob/010afb4b39e0d4aa2315cc438c0c5759a5742f6d/wakuv2/waku.go#L306

However, there are no associated tests. It would be good to have one in a manner similar to `TestBasicWakuV2` here: https://github.com/status-im/status-go/blob/010afb4b39e0d4aa2315cc438c0c5759a5742f6d/wakuv2/waku_test.go#L176

but using a local store node.

plopezlpz:
Adding a store wakuv2 test
Closes #4317

### DIFF
--- a/wakuv2/waku_test.go
+++ b/wakuv2/waku_test.go
@@ -358,7 +358,6 @@ func TestWakuV2Store(t *testing.T) {
 	// Send a message from the first node
 	msgTimestamp := w1.CurrentTime().UnixNano()
 	contentTopic := maps.Keys(filter.ContentTopics)[0]
-
 	_, err = w1.Send(relay.DefaultWakuTopic, &pb.WakuMessage{
 		Payload:      []byte{1, 2, 3, 4, 5},
 		ContentTopic: contentTopic.ContentTopic(),
@@ -380,8 +379,6 @@ func TestWakuV2Store(t *testing.T) {
 	storeResult, err := w1.query(context.Background(), w2.node.Host().ID(), relay.DefaultWakuTopic, []common.TopicType{contentTopic}, uint64(timestampInSeconds-int64(marginInSeconds)), uint64(timestampInSeconds+int64(marginInSeconds)), []store.HistoryRequestOption{
 		store.WithLocalQuery(),
 	})
-	if err != nil || len(storeResult.Messages) == 0 {
-		t.Fail()
-	}
-
+	require.NoError(t, err)
+	require.True(t, len(storeResult.Messages) > 0, "no messages received from store node")
 }
--- a/wakuv2/waku_test.go
+++ b/wakuv2/waku_test.go
@@ -387,6 +387,7 @@ func waitForPeerConnection(t *testing.T, peerID string, peerCh chan []string) {
 			}
 		case <-ctx.Done():
 			require.Fail(t, "timed out waiting for peer "+peerID)
+			return
 		}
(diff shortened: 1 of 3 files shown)


wakuv2/waku_test.go
```diff
@@ -280,3 +282,103 @@ func TestWakuV2Filter(t *testing.T) {
...
+
+func TestWakuV2Store(t *testing.T) {
+	// Configuration for the first Waku node
+	sql1, err := helpers.SetupTestMemorySQLDB(appdatabase.DbInitializer{})
+	require.NoError(t, err)
+	config1 := &Config{
+		Port:              0,
+		EnableDiscV5:      false,
+		DiscoveryLimit:    20,
+		EnableStore:       true,
```
vitvly: Do we need to enable store on the first Waku node?

wakuv2/waku_test.go
```diff
@@ -280,3 +282,103 @@ func TestWakuV2Filter(t *testing.T) {
...
+	err = tt.RetryWithBackOff(func() error {
+		if len(w1.Peers()) == 0 {
+			return errors.New("no peers discovered")
+		}
+		return nil
+	}, options)
+	require.NoError(t, err)
+
+	// Wait for the nodes to discover each other
+	time.Sleep(1 * time.Second)
```
vitvly: We use `Sleep` and `RetryWithBackOff` to wait for peers for discovered. I think this is often used in other tests too, however, just realised that we have this: https://github.com/status-im/status-go/blob/8e7d0a523e803f5f3b62f9b04deb3dd67449da16/wakuv2/waku.go#L1192
One can add a connection status subscription via `Waku.SubscribeToConnStatusChanges` and wait for `ConnStatus` structs to arrive.

wakuv2/waku_test.go
```diff
@@ -280,3 +282,103 @@ func TestWakuV2Filter(t *testing.T) {
...
+	_, err = w1.Send(relay.DefaultWakuTopic, &pb.WakuMessage{
+		Payload:      []byte{1, 2, 3, 4, 5},
+		ContentTopic: contentTopic.ContentTopic(),
+		Version:      0,
+		Timestamp:    msgTimestamp,
+	})
+	require.NoError(t, err)
+
+	// Wait for the message to be transferred
+	time.Sleep(1 * time.Second)
```
vitvly: In a similar fashion to connection status updates, we have this for envelope events: https://github.com/status-im/status-go/blob/8e7d0a523e803f5f3b62f9b04deb3dd67449da16/wakuv2/waku.go#L1419
One can subscribe to these envelope events via `Waku.SubscribeEnvelopeEvents`

vitvly: So maybe this way we can obviate the need for `Sleep` statements.

vitvly (changes requested):
Looks great, just some suggestions on possible ways to remove sleeps/retries in test code.

wakuv2/waku_test.go
plopezlpz: awesome, I'll try that

vitvly (approved):

PR #4341, opened 2023-11-19
New way of sharing communities

ibrkhalil:

Important changes:

Closes #status-mobile-17116


protocol/messenger_communities.go
```diff
@@ -2316,11 +2316,15 @@ func (m *Messenger) ShareCommunity(request *requests.ShareCommunity) (*Messenger
...
-		message.Text = fmt.Sprintf("Community %s has been shared with you", community.Name())
+		communityURL, err := m.ShareCommunityURLWithChatKey(community.ID())
+		if err != nil {
+			return nil, err
+		}
+		message.Text = communityURL
+		message.ContentType = protobuf.ChatMessage_COMMUNITY_INVITATION
 		if request.InviteMessage != "" {
-			message.Text = request.InviteMessage
+			message.Text = request.InviteMessage + "\n" + communityURL
```
MishkaRogachev: Suggestion: `fmt.Sprintf( %s\n%s", request.InviteMessage, communityURL)`

MishkaRogachev (commented):
There are staged extra changes, including migrations binaries. You can stage only your `.proto` and affected generated files (`chat_message.pg.go` i guess)

igor-sirotin:
Are we sure we need this fix? There was a decision to drop support for old "community invitation" concept and replace it with a simple community shared URL + new url unfurling mechanism.

We have a desktop issue for this here, I will start working on that soon: https://github.com/status-im/status-desktop/issues/12570
I already mentioned this here in comments, but it seems that it went unnoticed: https://github.com/status-im/status-mobile/issues/17777#issuecomment-1787336704

cc @ibrkhalil @cammellos @John-44

ibrkhalil:
> Are we sure we need this fix? There was a decision to drop support for old "community invitation" concept and replace it with a simple community shared URL + new url unfurling mechanism.
> 
> 
> 
> We have a desktop issue for this here, I will start working on that soon: https://github.com/status-im/status-desktop/issues/12570
> 
> I already mentioned this here in comments, but it seems that it went unnoticed: https://github.com/status-im/status-mobile/issues/17777#issuecomment-1787336704
> 
> 
> 
> cc @ibrkhalil @cammellos @John-44 

This PR adds a url in the message, But I figured I'd add a new protobuf type to differentiate it from normal text messages

igor-sirotin:
> This PR adds a url in the message, But I figured I'd add a new protobuf type to differentiate it from normal text messages

@ibrkhalil, the idea is not to use the `ChatMessage_COMMUNITY_INVITATION` at all.
To share a community the client app should send a usual text message with a URL and a link preview.

ibrkhalil:
> > This PR adds a url in the message, But I figured I'd add a new protobuf type to differentiate it from normal text messages
> 
> 
> 
> @ibrkhalil, the idea is not to use the `ChatMessage_COMMUNITY_INVITATION` at all.
> 
> To share a community the client app should send a usual text message with a URL and a link preview.

Understood, will fix.

cammellos:
@ibrkhalil thanks, @igor-sirotin has the whole context about this, so he will be able to help you on the solution we agreed on.

ibrkhalil:
@igor-sirotin Can you please take a look?

protocol/messenger_communities.go
```diff
@@ -2305,21 +2305,44 @@ func (m *Messenger) ShareCommunity(request *requests.ShareCommunity) (*Messenger
...
+
+	var messageStatusLinkPreview common.StatusLinkPreview
 	response := &MessengerResponse{}
 
 	community, err := m.communitiesManager.GetByID(request.CommunityID)
 	if err != nil {
 		return nil, err
 	}
 
+	communityURL, err := m.ShareCommunityURLWithChatKey(community.ID())
```
igor-sirotin: This should be `ShareCommunityURLWithData`.
Even though we don't use the shared URL data in desktop/mobile, we should still use those in case they will be opened in status-web someday.

protocol/messenger_communities.go
```diff
@@ -2305,21 +2305,44 @@ func (m *Messenger) ShareCommunity(request *requests.ShareCommunity) (*Messenger
...
+	if err != nil {
+		return nil, err
+	}
+
+	communityLinkPreview := &common.StatusCommunityLinkPreview{
```
igor-sirotin: This is interesting.

I like the simplicity of this solution. We simply fill `StatusCommunityLinkPreview` inplace.
But this duplicates the code from `StatusUnfurler`:
https://github.com/status-im/status-go/blob/1794b93c16b1f68cdb7125b4b66cc8926da9eed2/protocol/linkpreview_unfurler_status.go#L115-L121

I thought that we will call `UnfurlURLs` here. But that would be definitely an overkill here.

Maybe we could extract a smth like `func BuildStatusCommunityLinkPreview(community *communities.Community)`? This will make the solution both simple and non-duplicated.

protocol/messenger_communities.go
```diff
@@ -2305,21 +2305,44 @@ func (m *Messenger) ShareCommunity(request *requests.ShareCommunity) (*Messenger
...
 		message := common.NewMessage()
+		message.StatusLinkPreviews = []common.StatusLinkPreview{messageStatusLinkPreview}
 		message.ChatId = pk.String()
-		message.CommunityID = request.CommunityID.String()
 		message.Shard = community.Shard().Protobuffer()
-		message.Text = fmt.Sprintf("Community %s has been shared with you", community.Name())
+		message.ContentType = protobuf.ChatMessage_TEXT_PLAIN
+		if err != nil {
+			return nil, err
+		}
```
igor-sirotin: What `err` are we checking here? 🤔

protocol/messenger_communities.go
```diff
@@ -2305,21 +2305,44 @@ func (m *Messenger) ShareCommunity(request *requests.ShareCommunity) (*Messenger
...
+		message.Text = communityURL
 		if request.InviteMessage != "" {
+			fmt.Sprintf("%s\n%s", request.InviteMessage, communityURL)
 			message.Text = request.InviteMessage
```
igor-sirotin: Shouldn't it be a direct assignment?
```suggestion
			message.Text = fmt.Sprintf("%s\n%s", request.InviteMessage, communityURL)
```

protocol/communities_messenger_test.go
```diff
@@ -3684,3 +3684,239 @@ func (s *MessengerCommunitiesSuite) TestRetrieveBigCommunity() {
...
+	s.Require().True(aliceCommunity.Joined())
+
+	// Make sure the requests are not pending on either sides
+	requestsToJoin, err = s.bob.PendingRequestsToJoinForCommunity(community.ID())
+	s.Require().NoError(err)
+	s.Require().Len(requestsToJoin, 0)
+
+	requestsToJoin, err = s.alice.MyPendingRequestsToJoin()
+	s.Require().NoError(err)
+	s.Require().Len(requestsToJoin, 0)
```
igor-sirotin: I think this long part about joining the community doesn't make sense in this test.
IMO the test should be simple and only test the sharing functionality:
1. owner: create community, no matter of Membershiprules)
2. owner: share community to Alice
3. alice: check the received message

It doesn't even matter if Alice joined the community. If the URL was received, then we're good.

wdyt?

protocol/communities_messenger_test.go
```diff
@@ -3684,3 +3684,239 @@ func (s *MessengerCommunitiesSuite) TestRetrieveBigCommunity() {
...
+		}
+
+		if len(response.Messages()) == 0 {
+			return errors.New("Messages not received")
+		}
+
+		return nil
+	})
+
+	s.Require().Len(response.Messages(), 1)
```
igor-sirotin: Would be nice to check the message text here as well. I mean the `<invite-message>%s<shared-url>`.

As I wrote in comments, I think you have a bug here that `Come to status Bob!` is actually not prepended 🙂

igor-sirotin (changes requested):
Nice job, thank you! Left some important notes in comments.
WIll test with desktop tomorrow 👌

protocol/communities_messenger_test.go
ibrkhalil: The long test was to test if a non admin shared community.
Because of the admin was the only member who could share a community.
But I guess since this is a completely new implementation I can do your suggestion.

igor-sirotin:
@ibrkhalil just noticed that we have lots of tests that use `advertiseCommunityTo` functionality that depend on sharing community functionality with `ChatMessage_COMMUNITY`.

ibrkhalil:
> @ibrkhalil just noticed that we have lots of tests that use `advertiseCommunityTo` functionality that depend on sharing community functionality with `ChatMessage_COMMUNITY`.

ouch, I think versioning would be better here.

protocol/protobuf/chat_message.proto
```diff
@@ -149,6 +149,7 @@ message UnfurledStatusCommunityLink {
   string color = 5;
   UnfurledLinkThumbnail icon = 7;
   UnfurledLinkThumbnail banner = 8;
+  uint32 active_members_count = 9;
```
igor-sirotin: We shouldn't really have this here.
This number is "live" and there's no sense to rely on the information in the preview that was sent some time ago.

Client app should extend\* this link preview with the real live number of active users.

extend\*: Actually, it should replace all the received information with trusted data from waku, but this is a different question. Ping me if needed, I'll give you details on how we do this in desktop.

protocol/communities/manager.go
```diff
@@ -5134,3 +5134,54 @@ func (m *Manager) GetCuratedCommunities() (*CuratedCommunities, error) {
...
+		return nil, fmt.Errorf("failed to get data uri: %w", err)
+	}
+
+	thumbnail.Width = width
+	thumbnail.Height = height
+	thumbnail.DataURI = dataURI
+	return thumbnail, nil
+}
+
+func (m *Manager) BuildStatusCommunityLinkPreview(community *Community) (*common.StatusCommunityLinkPreview, error) {
```
igor-sirotin: Thanks for this!
We should also reuse this in `linkpreview_unfurler_status.go`, otherwise it's still a duplication.

_____

Also, do you think we could make it a method of `Community` type?
It would make much more sense than as a method of Manager I think.
```go
func (community *Community) toStatusLinkPreview() (*common.StatusCommunityLinkPreview, error)
```

... and same for images:
```go
func (image *images.IdentityImage) toLinkPreveiwThumbnail() (*common.LinkPreviewThumbnail, error)
```

Or make it the other way around a method `StatusCommunityLinkPreview` and `LinkPreviewThumbnail` correspondingly.

_______

If you want, you can leave this as is and I will refactor it after this PR.

protocol/communities/manager.go
```diff
@@ -5134,3 +5134,54 @@ func (m *Manager) GetCuratedCommunities() (*CuratedCommunities, error) {
...
+	communityLinkPreview := &common.StatusCommunityLinkPreview{}
+	if image, ok := community.Images()[images.SmallDimName]; ok {
+		if thumbnail, err := getImage(&images.IdentityImage{Payload: image.Payload}); err != nil {
+			m.logger.Warn("unfurling status link: failed to set community thumbnail", zap.Error(err))
```
igor-sirotin: We access `m.logger` here and probably this keeps us from making this a `Community` method. I think it's fine to drop this logging. It's not much needed anyway.

protocol/communities_messenger_test.go
igor-sirotin: So I still think that it makes no sense to test all of this here, it's overkill.

> But I guess since this is a completely new implementation

Yep, sounds very reasonable.

protocol/communities_messenger_test.go
```diff
@@ -3684,3 +3685,254 @@ func (s *MessengerCommunitiesSuite) TestRetrieveBigCommunity() {
...
+	s.Require().Contains(response.Messages()[0].Text, "Come to status Bob!\nhttps://status.app/c/GzIAAGS9bQ9sj5IgfISkFKWb9HDM_6A1mUUggqTTeYsdg9o011QUrSPim-1Teh7f25WhxjYA")
```
igor-sirotin: Why is the link hardcoded here? 🤔 
I think we should manually call `ShareCommunityURLWithData` here and check that the URL is equal.

Also,``https://status.app/c/GzIAAGS9bQ9sj5IgfISkFKWb9HDM_6A1mUUggqTTeYsdg9o011QUrSPim-1Teh7f25WhxjYA` doesn't look like a correct community shared URL.

protocol/communities_messenger_test.go
```diff
@@ -3684,3 +3685,254 @@ func (s *MessengerCommunitiesSuite) TestRetrieveBigCommunity() {
...
+	s.Require().NotNil(response.Messages()[0].UnfurledStatusLinks)
+	s.Require().Equal(response.Messages()[0].UnfurledStatusLinks.GetUnfurledStatusLinks()[0].GetCommunity().MembersCount, uint32(2))
+	s.Require().Equal(response.Messages()[0].UnfurledStatusLinks.GetUnfurledStatusLinks()[0].GetCommunity().DisplayName, "status")
+	s.Require().NotEmpty(response.Messages()[0].UnfurledStatusLinks.GetUnfurledStatusLinks()[0].GetCommunity().Banner)
+	s.Require().Len(response.Messages()[0].UnfurledStatusLinks.UnfurledStatusLinks, 1)
```
igor-sirotin: I believe that his check should go first

protocol/communities_messenger_test.go
```diff
@@ -3684,3 +3685,254 @@ func (s *MessengerCommunitiesSuite) TestRetrieveBigCommunity() {
...
+
+		return nil
+	})
+
+	s.Require().Len(response.Messages(), 1)
+	s.Require().Contains(response.Messages()[0].Text, "Come to status Bob!\nhttps://status.app/c/GzIAAGS9bQ9sj5IgfISkFKWb9HDM_6A1mUUggqTTeYsdg9o011QUrSPim-1Teh7f25WhxjYA")
+	s.Require().NotNil(response.Messages()[0].UnfurledStatusLinks)
+	s.Require().Equal(response.Messages()[0].UnfurledStatusLinks.GetUnfurledStatusLinks()[0].GetCommunity().MembersCount, uint32(2))
+	s.Require().Equal(response.Messages()[0].UnfurledStatusLinks.GetUnfurledStatusLinks()[0].GetCommunity().DisplayName, "status")
+	s.Require().NotEmpty(response.Messages()[0].UnfurledStatusLinks.GetUnfurledStatusLinks()[0].GetCommunity().Banner)
```
igor-sirotin: Actually this is tested as part of `MessengerLinkPreviewsTestSuite`, so IMO there's no sense to test it here too. It only adds cognitive load. 
The only important thing to check I think is the `...GetCommunity().CommunityID`.

protocol/messenger_communities.go
```diff
@@ -2351,10 +2351,71 @@ func (m *Messenger) ShareCommunity(request *requests.ShareCommunity) (*Messenger
 	return response, nil
 }
 
+func (m *Messenger) ShareCommunityV2(request *requests.ShareCommunity) (*MessengerResponse, error) {
```
igor-sirotin: Though versioning is a good idea, I think we're able to avoid this.
We introduce `ShareCommunityV2` because I warned you about `advertiseCommunityTo` used in tests. But I believe it should be possible to update it internally and not break anything.

This is the function I'm talking about:

https://github.com/status-im/status-go/blob/384543d3a6329fe88848f14daad1e560d5c89af8/protocol/communities_messenger_helpers_test.go#L322-L347

The function has access to both sender's and receiver's messengers. So we could just take required parts from the old functionality and force it's usage here. 

1. Take the community description from sender's messenger:
https://github.com/status-im/status-go/blob/384543d3a6329fe88848f14daad1e560d5c89af8/protocol/messenger.go#L2306-L2318

2. Push the community description to receiver's messenger:
https://github.com/status-im/status-go/blob/e32c5546e1e5d13f597d2c4d56c92cf831d35ea8/protocol/messenger_handler.go#L2298-L2306

wdyt?

igor-sirotin: Here comes a more general question: _should we deprecate this message type and usage of `Messenger.CommunityID` in this PR?_

We could either do it here and not introduce a new `ShareCommunityV2`, or I could pick it up after this is merged with new endpoint and updated all other status-go places to deprecate the old functionality.

I'm ok with both ways, but please open an tracking issue if we decide to go with second option 🙂

igor-sirotin (changes requested):
The main idea is good, but I still have concerns, left comments inline and in DM.

protocol/communities_messenger_test.go
ibrkhalil: Done

protocol/communities/manager.go
ibrkhalil: Did it for the community one, But fell for a circular dependency when doing the image one.

protocol/messenger_communities.go
ibrkhalil: Kindly take a look at the new changes, If you think they are okay then I'll remove the `protobuf.ChatMessage_COMMUNITY` type.

protocol/communities_messenger_test.go
ibrkhalil: I think on the tests I cannot exactly predict the community hashed data.
So previously I was checking a substring, But next day it failed 😆.

protocol/messenger_communities.go
ibrkhalil: And I still need to make a change to the logic that adds the community data so that it can work with unsigned user (non admin)
For example: This doesn't work (Non admin member does invitre)
```go
func (s *MessengerCommunitiesSuite) TestShareCommunity() {
	ctx := context.Background()
	description := &requests.CreateCommunity{
		Membership:  protobuf.CommunityPermissions_MANUAL_ACCEPT,
		Name:        "status",
		Description: "status community description",
		Color:       "#FFFFFF",
		Image:       "../_assets/tests/status.png",
		ImageAx:     0,
		ImageAy:     0,
		ImageBx:     256,
		ImageBy:     256,
		Banner: images.CroppedImage{
			ImagePath: "../_assets/tests/IMG_1205.HEIC.jpg",
			X:         0,
			Y:         0,
			Width:     160,
			Height:    90,
		},
	}

	response, err := s.owner.CreateCommunity(description, true)

	s.Require().NoError(err)
	s.Require().NotNil(response)
	s.Require().Len(response.Chats(), 1)
	community := response.Communities()[0]

	// Alice shares community with Bob
	response, err = s.owner.ShareCommunity(&requests.ShareCommunity{
		CommunityID:   community.ID(),
		Users:         []types.HexBytes{common.PubkeyToHexBytes(&s.alice.identity.PublicKey)},
		InviteMessage: "Come on alice, You'll like it here!",
	})

	s.Require().NoError(err)
	s.Require().NotNil(response)
	s.Require().Len(response.Messages(), 1)

	_, err = WaitOnMessengerResponse(s.alice, func(r *MessengerResponse) bool {
		return len(r.Communities()) > 0
	}, "Messages not received")
	s.Require().NoError(err)
	s.Require().Len(response.Messages(), 1)
	s.Require().Contains(response.Messages()[0].Text, "Come on alice, You'll like it here!\nhttps://status.app/c/") // The last bit of the link is auto generated to we cannot assert the whole link.
	s.Require().NotNil(response.Messages()[0].UnfurledStatusLinks)

	response, err = s.alice.JoinCommunity(ctx, community.ID(), false)
	s.Require().NoError(err)
	s.Require().NotNil(response)
	s.Require().True(response.Communities()[0].Joined())
	s.Require().Len(response.Chats(), 1)

	// Alice shares community with Bob
	response, err = s.alice.ShareCommunity(&requests.ShareCommunity{
		CommunityID:   community.ID(),
		Users:         []types.HexBytes{common.PubkeyToHexBytes(&s.bob.identity.PublicKey)},
		InviteMessage: "Come on alice, You'll like it here!",
	})

	_, err = WaitOnMessengerResponse(s.alice, func(r *MessengerResponse) bool {
		return len(r.Communities()) > 0
	}, "Communities not received")
}
```

ibrkhalil: Will continue investigating how to do it with any permissions tomorrow.

igor-sirotin: > For example: This doesn't work (Non admin member does invitre)

What exactly doesn't work in this test?

Just a note, this could be related: https://github.com/status-im/status-go/issues/4413

ibrkhalil: Cannot wrap community because sender is not admin.

protocol/communities_messenger_test.go
igor-sirotin:
```go
s.Require().Equal(response.Messages()[0].Text, fmt.Printf("Come to status Bob!\n%s", communityLink))
```

protocol/communities_messenger_test.go
```diff
@@ -2102,57 +2103,51 @@ func (s *MessengerCommunitiesSuite) TestLeaveAndRejoinCommunity() {
...
+	_, err = WaitOnMessengerResponse(s.alice, func(r *MessengerResponse) bool {
+		return len(r.Messages()) > 0
+	}, "Messages not received")
 	s.Require().NoError(err)
 	s.Require().Len(response.Messages(), 1)
-
-	message := response.Messages()[0]
-	s.Require().Equal(community.IDString(), message.CommunityID)
-	s.Require().Equal(inviteMessage, message.Text)
+	s.Require().Contains(response.Messages()[0].Text, sentMessageText)
```
igor-sirotin: Did you decide not to check `<message-text>\n<url>` structure?

igor-sirotin (approved):
Cool, thank you!
Will you please open an issue about deprecating community invitations? And assign it to me 🙂

protocol/communities_messenger_test.go
ibrkhalil: Sorry I forgot to add it into the test, Added.

ibrkhalil:
> Cool, thank you! Will you please open an issue about deprecating community invitations? And assign it to me 🙂

Sure

MishkaRogachev (approved):
I'm ok with this PR after all Igor's questions are closed

igor-sirotin:
@ibrkhalil maybe rename this PR? 🙂 
It's actually about changing the way we share a community now.

PR #4342, opened 2023-11-19
feat(wallet) WalletConnect: track pairings in DB

stefandunca:
### Closes status-desktop [#12794](https://github.com/status-im/status-desktop/issues/12794)

Add new APIs to track if valid pairings are available to be used by the application:

- to be used by activity filtering to show `dapp` interaction
- not to run WalletConnect SDK if not needed


services/wallet/walletconnect/database.go
```diff
@@ -0,0 +1,76 @@
...
+	err := row.Scan(&pairing.Topic, &pairing.Expiry, &pairing.Disconnected, &pairing.AppName, &pairing.URL, &pairing.Description, &pairing.Icon, &pairing.Verified.IsScam, &pairing.Verified.Origin, &pairing.Verified.VerifyURL, &pairing.Verified.Validation)
+	if err != nil {
+		return nil, err
+	}
+
+	return &pairing, nil
+}
+
+// GetActivePairings returns all active pairings (not disconnected and not expired) that have an expiry timestamp newer or equal to the given timestamp.
+func GetActivePairings(db *sql.DB, expiryNotOlderThanTimestamp int64) ([]Pairing, error) {
```
saledjenic, line 40:
```suggestion
func GetActivePairings(db *sql.DB, expiryNotOlderThanTimestamp uint64) ([]Pairing, error) {
```

services/wallet/walletconnect/database.go
```diff
@@ -0,0 +1,76 @@
...
+		err := rows.Scan(&pairing.Topic, &pairing.Expiry, &pairing.Disconnected, &pairing.AppName, &pairing.URL, &pairing.Description, &pairing.Icon, &pairing.Verified.IsScam, &pairing.Verified.Origin, &pairing.Verified.VerifyURL, &pairing.Verified.Validation)
+		if err != nil {
+			return nil, err
+		}
+		if err := rows.Err(); err != nil {
+			return nil, err
+		}
+		pairings = append(pairings, pairing)
+	}
+
```
saledjenic, line 64: @stefandunca we should check here for potential errors from calling `rows.Next()`, like this:
```
    return nil, err
}
```

services/wallet/api.go
saledjenic, line 1: @stefandunca should we update the version file?

saledjenic (approved):

walletdatabase/migrations/sql/1700414564_add_wallet_connect_pairings_table.up.sql
saledjenic, line 1: @stefandunca should we update the version file?

services/wallet/walletconnect/database.go
stefandunca: It looks to me that `in64` is a better type for epoch timestamp data due to various reasons (standardization, working as negative dates, compatibility, overflow/underflow in arithmetic), and I can't see many advantages to uint64 (allowing un-probable date representation)

PR #4344, opened 2023-11-20
makefile: fix linking Git hooks on MacOS

jakubgs:
Otherwise it results in:
```
ln: illegal option -- r
```
Because MacOS tools are such garbage they lack basic features. And get absolute path of the `Makefile` location to work regardless ofr from where the `Makefile` is called.

Also renamed `AUTHOR` to `GIT_AUTHOR` to reflect source.


igor-sirotin (approved):
Thanks ❤️

apentori (approved):

yakimant (approved):

PR #4345, opened 2023-11-20
feat(walletconnect)_: support the tx and personal signing from within the app or keycard

saledjenic:


stefandunca (approved):
It looks great

PR #4353, opened 2023-11-21
chore: fix sending request to join msg's to the new owner

mprakhov:
There are 2 problems:
1. When the ownership changed - if we continue to send the msg by `SendCommunityMessage` - wakuv2 api `Post` will set asymmetric key using `community.ID`.
We can handle it by using for asymetric key not `community.ID`, but `community.ControlNode()` if we minted the token, but here we receive the second problem

2. Filters - owner will not receive send request to join msg. This will require updating filters on both sides - on the owner and on the sender (client) side

Taking into account, that requests to join, leave, cancel request to join, and shared revealed accounts must be a private msg to the control node - I modified the existing code to `sendPrivate` msg in case we have a community with a minted owner token

Closes # https://github.com/status-im/status-desktop/issues/12795

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -2838,7 +2838,7 @@ func (m *Manager) HandleCommunityRequestToJoinResponse(signer *ecdsa.PublicKey, 
 		return nil, err
 	}
 
-	isControlNodeSigner := common.IsPubKeyEqual(community.PublicKey(), signer)
+	isControlNodeSigner := common.IsPubKeyEqual(community.ControlNode(), signer)
 	if !isControlNodeSigner {
 		return nil, ErrNotAuthorized
 	}
--- a/protocol/communities_messenger_signers_test.go
+++ b/protocol/communities_messenger_signers_test.go
@@ -1,6 +1,7 @@
 package protocol
 
 import (
+	"context"
 	"crypto/ecdsa"
 	"testing"
 	"time"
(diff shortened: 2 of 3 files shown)


mprakhov:
@cammellos this is the issue I wrote you about. 
Please take a look or assign someone to review the changes from the mobile team and verify possible drawbacks

protocol/communities/manager.go
```diff
@@ -2838,7 +2838,7 @@ func (m *Manager) HandleCommunityRequestToJoinResponse(signer *ecdsa.PublicKey,
 		return nil, err
 	}
 
-	isControlNodeSigner := common.IsPubKeyEqual(community.PublicKey(), signer)
+	isControlNodeSigner := common.IsPubKeyEqual(community.ControlNode(), signer)
```
jrainville, line 2841: What does `community.ControlNode()` return when the Owner token is not minted yet?

protocol/communities_messenger_signers_test.go
```diff
@@ -451,3 +452,102 @@ func (s *MessengerCommunitiesSignersSuite) TestAutoAcceptOnOwnershipChangeReques
...
+
+	inputMessage := common.NewMessage()
+	inputMessage.ChatId = chat.ID
+	inputMessage.Text = "some text"
+	inputMessage.CommunityID = community.IDString()
+
+	err = s.alice.SaveChat(chat)
+	s.Require().NoError(err)
+	_, err = s.alice.SendChatMessage(context.Background(), inputMessage)
+	s.Require().NoError(err)
```
jrainville, line 539: Can't you use `advertiseCommunityTo` instead of all this?

jrainville (commented):
Looks good. Just 2 questions

protocol/communities/manager.go
mprakhov: config.ID which is PublicKey of the community

protocol/communities_messenger_signers_test.go
mprakhov: in this case no, here community for the new member will be added to a verification loop on the client side, so I need to listen for a Signal. That's why I'm waiting for a `WaitOnSignaledMessengerResponse` instead of `WaitOnMessengerResponse`

mprakhov: as an option, the code can be modified to pass a "listener" to `advertiseCommunityTo` :)

jrainville (approved):
All good then!

protocol/messenger_communities.go
```diff
@@ -6148,3 +6148,11 @@ func (m *Messenger) CreateResponseWithACNotification(communityID string, acType
 
 	return response, nil
 }
+
+func (m *Messenger) SendCommunityMessage(community *communities.Community, rawMessage common.RawMessage) ([]byte, error) {
+	if community.ControlNode() != nil && !community.PublicKey().Equal(community.ControlNode()) {
```
```suggestion
	if !community.PublicKey().Equal(community.ControlNode()) {
```

ControlNode() can't be nil, can it? :thinking:

protocol/communities_key_distributor.go
```diff
@@ -79,6 +79,7 @@ func (ckd *CommunitiesKeyDistributorImpl) sendKeyExchangeMessage(community *comm
 		HashRatchetGroupID:    hashRatchetGroupID,
 		PubsubTopic:           community.PubsubTopic(), // TODO: confirm if it should be sent in community pubsub topic
 	}
+	// TODO: should we also send it private? Do I need additional logic in sendPrivate?
```
osmaczko: It is send as private under the hood, no need for adaptation here.

protocol/messenger_communities.go
```diff
@@ -6148,3 +6148,11 @@ func (m *Messenger) CreateResponseWithACNotification(communityID string, acType
 
 	return response, nil
 }
+
+func (m *Messenger) SendCommunityMessage(community *communities.Community, rawMessage common.RawMessage) ([]byte, error) {
```
```suggestion
func (m *Messenger) SendMessageToControlNode(community *communities.Community, rawMessage common.RawMessage) ([]byte, error) {
```

osmaczko (approved):
Looks good :+1:

protocol/messenger_communities.go
mprakhov: just added it for backward compatibility, I'm not sure that on the previous version of status it can't be nil
Removing

osmaczko: I don't understand the concern :thinking: you are not pushing this commit in the past, are you?

https://github.com/status-im/status-go/blob/c62957f621ff8169f5f4f0650ac4ccf71a1cafa9/protocol/communities/community.go#L1352-L1357

PR #4360, opened 2023-11-22
fix(wakuv2): store query pubsubTopic

richard-ramos:
Pubsub topic was arriving empty to the `query` function. This PR fixes that by using the default pubsub topic / shard if it is empty


vitvly (approved):

igor-sirotin (approved):

jrainville:
@richard-ramos do you know if this fix is needed for the release branch? ie is the bug in the release branch too?

If so, would you mind cherry-picking this fix to the release branch after merging. Thanks!

Issue #4192, opened 2023-10-24
refactor and fix `RequestCommunityInfoFromMailserver`


igor-sirotin:
# Problem

While investigating [problem](https://github.com/status-im/status-desktop/issues/12465) with `RequestCommunityInfoFromMailserver`, I found many other issues on the way:

1. Only a single community info request is possible at a time.     
This includes the regular batch requests with `requestCommunitiesInfoFromMailserver`.

2. When the given community is already requested, you won't get the result. 
This includes the regular batch requests with `requestCommunitiesInfoFromMailserver`. Easy to get into a conflict.
We should be able to simply wait for that already made request to finish.

4. After processing the mailserver batch, we run a loop where once in 200ms we check if the community appeared in the `CommunityManager`.
This is not quite beautiful and reliable, we could Instead create some kind of "community fetched event" and subscribe to it.

5. The code is duplicated in `requestCommunityInfoFromMailserver` and `requestCommunitiesInfoFromMailserver`.

6. We request a full MONTH of history.
This might be a lot of data, while we only need the latest communtiy description.
It's not that simple to fix it, but we could at least implement a stupid algorithm here. Like fetch 1 day, if no description found, fetch 2 days, then 4 days and so on.

7. Because the temporary created filter is added to `allFilters`, at some point it's picked up by regular history requests.
We should avoid this, temporal filters should only be processed with the requests from the function itself.
This makes a redundant duplicate mailserver request.

# Implementation

I'm thinking toward a separate `CommunityRequestsManager`.
I have even started a draft implementation.

# Acceptance Criteria

Would be _perfect_, if we can write tests for this function.
This requires to setup a local network of 3 nodes:
- a store node
- community owner node
- community "requestor"

igor-sirotin:
As @caybro mentioned about this issue: https://github.com/status-im/status-desktop/issues/12542, in desktop we face wrong behaviour of the `RequestCommunityInfoFromMailserver` endpoint. Though we keep list of requested communities on the application side, and filter out the ones that are being requested right now, we still might receive the "community already requested" response from status-go. This is probably happening because status-go is regularly self-requesting communities with `requestCommunitiesInfoFromMailserver`.

Which is exactly what I described in clauses 1 and 2.

igor-sirotin:
This also applies to `RequestContactInfoFromMailserver`

igor-sirotin:
Also, would be nice if this function could automatically wait for mailserver to become available before making the request.
[...]

PR #4364, opened 2023-11-22
test: request community from storenode

igor-sirotin:
Required for https://github.com/status-im/status-go/issues/4192

## Implemented `TestRequestCommunityInfo`

Here's the logic of the test:
- Create a waku store node (
- Create a non-store waku node and a messenger using it ("owner")
- Create a non-store waku node and a messenger using it ("bob")
- Connect both waku nodes to the created store node
- owner: create a community
- bob: fetch the community by id
- 🎉 

Tested that this works without internet connection.

## Other bugfixes and improvements

This wouldn't be possible without this:

1. Move `StartRetrieveMessagesLoop` and `PublishMessengerResponse` to Messenger.
This was needed to be able to start the retrieve loop in tests, where we don't have ext service.
Also there was a duplication of `PublishMessengerResponse` in `protocol/messenger_sync_raw_messages.go`.
Also with this I removed `onContactENSVerified`, which improved readability.

2. Added `mailservers` argument to `StartMailserverCycle`.
Previously we were gathering it on every mailserver cycle iteration

3. Added filter by waku version in `Messenger.allMailservers()`
I guess this makes sense no to connect to nodes of different version? cc @richard-ramos 

4. Move `newWakuV2` and `createWakuNetwork` functions to `messenger_testing_utils.go`

5. Remove `Messenger.mailservers`, as it was a duplication of `mailserversDatabase`

6. Refactored `StartMailserverCycle` to check waku version.

7. Removed code duplication in `findNewMailserver`

8. Reformatted and improved some logging.

### DIFF
--- a/protocol/messenger.go
+++ b/protocol/messenger.go
@@ -142,9 +142,10 @@ type Messenger struct {
 	browserDatabase            *browsers.Database
 	httpServer                 *server.MediaServer
 
-	quit   chan struct{}
-	ctx    context.Context
-	cancel context.CancelFunc
+	quit              chan struct{}
+	ctx               context.Context
+	cancel            context.CancelFunc
+	shutdownWaitGroup sync.WaitGroup
 
 	importingCommunities map[string]bool
 	importingChannels    map[string]bool
@@ -160,13 +161,12 @@ type Messenger struct {
 	requestedContactsLock sync.RWMutex
 	requestedContacts     map[string]*transport.Filter
 
-	connectionState                      connection.State
-	telemetryClient                      *telemetry.Client
-	contractMaker                        *contracts.ContractMaker
-	downloadHistoryArchiveTasksWaitGroup sync.WaitGroup
-	verificationDatabase                 *verification.Persistence
-	savedAddressesManager                *wallet.SavedAddressesManager
-	walletAPI                            *wallet.API
+	connectionState       connection.State
+	telemetryClient       *telemetry.Client
+	contractMaker         *contracts.ContractMaker
+	verificationDatabase  *verification.Persistence
+	savedAddressesManager *wallet.SavedAddressesManager
+	walletAPI             *wallet.API
 
 	// TODO(samyoul) Determine if/how the remaining usage of this mutex can be removed
 	mutex                     sync.Mutex
@@ -1884,7 +1884,7 @@ func (m *Messenger) Init() error {
 func (m *Messenger) Shutdown() (err error) {
 	close(m.quit)
 	m.cancel()
-	m.downloadHistoryArchiveTasksWaitGroup.Wait()
+	m.shutdownWaitGroup.Wait()
 	for i, task := range m.shutdownTasks {
 		m.logger.Debug("running shutdown task", zap.Int("n", i))
 		if tErr := task(); tErr != nil {
@@ -3272,16 +3272,14 @@ func (m *Messenger) RetrieveAll() (*MessengerResponse, error) {
 }
 
 func (m *Messenger) StartRetrieveMessagesLoop(tick time.Duration, cancel <-chan struct{}) {
+	m.shutdownWaitGroup.Add(1)
 	go func() {
+		defer m.shutdownWaitGroup.Done()
 		ticker := time.NewTicker(tick)
 		defer ticker.Stop()
 		for {
 			select {
 			case <-ticker.C:
-				// We might be shutting down here
-				if m == nil {
-					return
-				}
 				m.ProcessAllMessages()
 			case <-cancel:
 				return
--- a/protocol/messenger.go
+++ b/protocol/messenger.go
@@ -62,6 +62,7 @@ import (
 	"github.com/status-im/status-go/services/communitytokens"
 	ensservice "github.com/status-im/status-go/services/ens"
 	"github.com/status-im/status-go/services/ext/mailservers"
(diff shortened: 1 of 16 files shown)


protocol/messenger.go
```diff
@@ -1444,13 +1442,7 @@ func (m *Messenger) handleENSVerified(records []*ens.VerificationRecord) {
 		}
 	}
 
-	m.logger.Info("calling on contacts")
-	if m.config.onContactENSVerified != nil {
-		m.logger.Info("called on contacts")
-		response := &MessengerResponse{Contacts: contacts}
-		m.config.onContactENSVerified(response)
-	}
-
```
richard-ramos, line 1453: I see in the PR description that this was was removed to increase readability. But is the call to `m.config.oNContactENSVerified()` not needed anymore?

protocol/messenger.go
```diff
@@ -3279,6 +3271,46 @@ func (m *Messenger) RetrieveAll() (*MessengerResponse, error) {
...
+		ticker := time.NewTicker(tick)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ticker.C:
+				// We might be shutting down here
+				if m == nil {
+					return
+				}
+				m.ProcessAllMessages()
```
richard-ramos: This could still fail inside `m.ProcessAllMessages()` if `m` becomes `nil` while `ProcessAllMessages` is being executed. Perhaps it's a good idea to add a waitgroup property to `Messenger`, and replace line 3275 by:
```
m.waitGroup.Add(1)
go func() {
    defer m.waitGroup.Done()
```

and later in Shutdown:

```
// Shutdown takes care of ensuring a clean shutdown of Messenger
func (m *Messenger) Shutdown() (err error) {
	close(m.quit)
	m.cancel()
        m.waitgroup.Wait()   // <---------------------------------------------
```

And that way, this goroutine would be safer to execute, IMO

richard-ramos (approved):
Very cool!

richard-ramos:
> I guess this makes sense no to connect to nodes of different version? cc @richard-ramos

Yeah, it is a nice addition!

protocol/messenger.go
igor-sirotin: `onContactENSVerified ` was previously set to `publishMessengerResponse` in `WithENSVerificationConfig`.
Now when `PublishMessengerResponse` is accessibly from here, we just call it directly.

protocol/messenger.go
igor-sirotin: This is an interesting finding!
But I think this is not the only place where we start a goroutine that uses messenger. I can fix this one, but who will take care of the others? 😄

igor-sirotin: @richard-ramos renamed `downloadHistoryArchiveTasksWaitGroup` to `shutdownWaitGroup` and resued it here. It was doing just the thing you proposed 🙂

protocol/messenger_communities.go
```diff
@@ -2624,6 +2629,9 @@ func (m *Messenger) requestCommunityInfoFromMailserver(communityID string, shard
 		return nil, nil
 	}
 
+	// TODO: We can force to process all messages then we don't have to wait?
+	//m.ProcessAllMessages()
```
jrainville, line 2634: Is this still needed?

protocol/messenger_storenode_request_test.go
```diff
@@ -0,0 +1,143 @@
...
+func (s *MessengerStoreNodeRequestSuite) SetupTest() {
+	s.logger = tt.MustCreateTestLogger()
+
+	// Create store node
+
+	storeNodeLogger := s.logger.With(zap.String("name", "store-node-waku"))
+	s.wakuStoreNode = NewWakuV2(&s.Suite, storeNodeLogger, true, true)
+
+	//protocols := s.wakuStoreNode.Protocols()
+	//s.Require().Contains(protocols, store.StoreID_v20beta4)
```
jrainville: leftover

jrainville (approved):
Very good job!

protocol/messenger_communities.go
igor-sirotin: yep, I left it intentionally for myself not to forget when doing https://github.com/status-im/status-go/issues/4192

protocol/messenger_storenode_request_test.go
igor-sirotin: oh, this is actually a question that I forgot to ask @richard-ramos!
Is there a way to check that the waku node successfully supports store protocol?
I tried to use `.Protocols()`, but it's always empty 🤔

igor-sirotin: ok, Richard proposed some options, but this seems to be redundant to do.
I'll remove these lines in upcoming PRs, not to rerun tests here.

igor-sirotin:
While working further on more tests, figured out that this one is flaky.
Working on this now.

igor-sirotin:
The flaky test is now fixed, this was the issue:
https://github.com/status-im/status-go/blob/2d251e9a08eea85dd788b1126d3aee54562a080b/protocol/messenger_communities.go#L2599

The fix is to `ceil` the timestamp. Store node database uses nano-precision timestamps and there for wasn't returning anything if the community was added "in the same second".
https://github.com/status-im/status-go/blob/d78c8c6189eff06b3274ed32f5b1d66b5e782f8a/protocol/messenger_communities.go#L2605

PR #4365, opened 2023-11-22
Resend community offline updates

vitvly:
Builds on top of https://github.com/status-im/status-go/commit/f1d1726a4e12159013d72008c949fee77e48627c

Added a separate testing suite to deal with sending messages in a community under flaky network conditions.

In order to facilitate testing, an extra method `SetSendFailures` is added to `Waku` struct, and message resend params are configurable via Messenger Options.

Looking into failing tests.

Related to https://github.com/status-im/status-mobile/issues/17950

### DIFF
--- a/protocol/persistence.go
+++ b/protocol/persistence.go
@@ -856,7 +856,7 @@ func (db sqlitePersistence) ExpiredMessagesIDs(maxSendCount int) ([]string, erro
 			FROM
 				raw_messages
 			WHERE
-			message_type IN (?, ?) AND sent = ? AND send_count <= ?`,
+			(message_type IN (?, ?) OR resend_automatically) AND sent = ? AND send_count <= ?`,
 		protobuf.ApplicationMetadataMessage_CHAT_MESSAGE,
 		protobuf.ApplicationMetadataMessage_EMOJI_REACTION,
 		false,
--- a/wakuv2/waku.go
+++ b/wakuv2/waku.go
@@ -89,6 +89,7 @@ type settings struct {
 	EnableDiscV5        bool   // Indicates whether discv5 is enabled or not
 	DefaultPubsubTopic  string // Pubsub topic to be used by default for messages that do not have a topic assigned (depending whether sharding is used or not)
 	Options             []node.WakuNodeOption
+	SkipPublishToTopic  bool // used in testing
 }
 
 type ITelemetryClient interface {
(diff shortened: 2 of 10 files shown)


protocol/persistence.go
```diff
@@ -856,7 +856,7 @@ func (db sqlitePersistence) ExpiredMessagesIDs(maxSendCount int) ([]string, erro
 			FROM
 				raw_messages
 			WHERE
-			message_type IN (?, ?) AND sent = ? AND send_count <= ?`,
+			(message_type IN (?, ?) OR resend_automatically) AND sent = ? AND send_count <= ?`,
```
cammellos, line 859: I think we want an index on `resend_automatically` if not there, also I think it's safer for release to do the fix as you did, but for develop we can stop checking the message type, and only trust `resend_automatically`, what do you think?

vitvly: Ok, i will add the index for release! As for dropping message_type checks in develop, would it work if i verify them and fix tests (if any broken with this PR) separately?

churik:
@vitvly I tested fix in https://github.com/status-im/status-mobile/pull/17979, and it is working 🎉 
As I see some review comments, can you ping me when PR is approved for a quick re-check?

Thank you!

wakuv2/waku.go
```diff
@@ -995,13 +996,19 @@ func (w *Waku) UnsubscribeMany(ids []string) error {
...
+
 func (w *Waku) broadcast() {
 	for {
 		select {
 		case envelope := <-w.sendQueue:
 			var err error
 			logger := w.logger.With(zap.String("envelopeHash", hexutil.Encode(envelope.Hash())), zap.String("pubsubTopic", envelope.PubsubTopic()), zap.String("contentTopic", envelope.Message().ContentTopic), zap.Int64("timestamp", envelope.Message().Timestamp))
-			if w.settings.LightClient {
+			if w.settings.SendFailures {
+				err = errors.New("Test send failure")
```
kaichaosun, line 1030: How about change `SendFailures` to `SkipPublishToTopic` which may feel less test specific, and add a short comment explain it mostly used in test scenarios?
Wondering if `err` is needed here, I haven't look into how `EventEnvelopeExpired` is used.

vitvly: Good idea, thanks, renamed. The error value at least is useful in tests, it's visible in the output.

protocol/persistence.go
vitvly: Added index for `resend_automatically`.

kaichaosun (approved):
LGTM

PR #4367, opened 2023-11-23
Fix/duplicate accept cr messages

### LINKED ISSUE #17968: Duplicate 'Contact Added' Notifications When Pair-Device User Accepts Request
# Bug Report

## Problem 
First device is userA, second and third are userB and paired, when userA send CR to userB and userB accepted, userA will receive 2 duplicate messages: "xxx is now a contact"

#### Expected behavior
userA should receive message: "Bbbbb is now a contact" only once.

#### Actual behavior
userA received message: "Bbbbb is now a contact" twice.

### Reproduction

1. prepare 3 devices, 1 device for userA, the other 2 devices(paired) for userB
2. userA send contact request to userB
3. userB accept contact request with either device
4. observe messages received for userA

### Additional Information

- Status version: nightly 

qfrank:
this PR fix [mobile issue](https://github.com/status-im/status-mobile/issues/17968)


MishkaRogachev (approved):
LGTM!

protocol/messenger_sync_contact_request_decision_test.go
```diff
@@ -0,0 +1,115 @@
...
+	_, err = WaitOnMessengerResponse(s.m, func(r *MessengerResponse) bool {
+		if len(r.ActivityCenterNotifications()) > 0 {
+			for _, n := range r.ActivityCenterNotifications() {
+				if n.Type == ActivityCenterNotificationTypeContactRequest {
+					contactRequestMessageID = n.ID
+					return true
+				}
+			}
+		}
+		return false
```
igor-sirotin: I'd appreciate inverting the condition for better readability 
```suggestion
		if len(r.ActivityCenterNotifications()) == 0 {
		    return false
		}
	    for _, n := range r.ActivityCenterNotifications() {
			if n.Type == ActivityCenterNotificationTypeContactRequest {
			    contactRequestMessageID = n.ID
				return true
			}
		}
		return false
```

protocol/messenger_sync_contact_request_decision_test.go
```diff
@@ -0,0 +1,115 @@
...
+	}, "contact request not received on device 1")
+	s.Require().NoError(err)
+	_, err = WaitOnMessengerResponse(s.m2, func(r *MessengerResponse) bool {
+		if len(r.ActivityCenterNotifications()) > 0 {
```
igor-sirotin: same here

protocol/messenger_sync_contact_request_decision_test.go
igor-sirotin: Oh, it's even better actually.
The condition is not needed at all, the loop will simply make 0 iterations.

protocol/messenger_sync_contact_request_decision_test.go
```diff
@@ -0,0 +1,115 @@
...
+		r, err := userB.RetrieveAll()
+		if err != nil {
+			return err
+		}
+		for _, m := range r.Messages() {
+			if m.GetContentType() == protobuf.ChatMessage_SYSTEM_MESSAGE_MUTUAL_EVENT_ACCEPTED {
+				numAcceptMessageReceived++
+			}
+		}
+		// always return error so that we retry until timeout
```
igor-sirotin: I guess it's ok for now, but it's a bad practice.
Tests shouldn't rely on "being timed out". This  means that this test will always run for 10 seconds. And if someone increases the timeout, it will not change the test logic, but increase the run time.

We should probably test that the cr acceptance is not sent rather than that it's not received.

igor-sirotin: If we leave it like this for now, maybe consider it with a more obvious code?
```go
time.Sleep(10 * time.Second)
r, err := userB.RetrieveAll()
// check that r.Messages contain only 1 CR acceptance
```

igor-sirotin (approved):
Good finding, thank you!

protocol/messenger_sync_contact_request_decision_test.go
qfrank: emm.. I don't like Sleep, i can add a BackOffOption to restrict MaxElapsedTime, maybe 5 seconds? I doubt somebody would increase the timeout. I don't find a better way how to check cr acceptance is not sent, if you found one, do tell me! @igor-sirotin

igor-sirotin: I don't like `Sleep` either, but it's I think in fact  the same as current solution, but more straightforward.

As for checking the sent messages, we could modify `dispatchMessage` to use some debug callback that would only be set in tests. This callback could be set within tests to count the sent messages. This should be in sync with the call to `AcceptContactRequest`, so we won't need to wait after the call.

igor-sirotin: I've made a POC in this branch based on your PR: `fix/duplicate_accept_CR_messages-no-wait-poc`.

https://github.com/status-im/status-go/blob/cc4377f2145edc8083119b0a19b0aae665417a72/protocol/messenger_sync_contact_request_decision_test.go#L86-L106

The test takes 2 seconds in both cases of fail and success.

@qfrank  wdyt?

qfrank: interesting hack, i'll make an update

qfrank: done @igor-sirotin

igor-sirotin (approved):
Nice, thanks for the changes!
We saved ~8 seconds of tests runtime 💪

PR #4369, opened 2023-11-23
Aligning Namespaces with WalletConnect v2.0 protocol

saledjenic:


stefandunca (approved):
Great to see all methods supported 🤩

PR #4370, opened 2023-11-23
chore: add cluster_id to DB, move clusterID to config file and load clusterID when switching fleets

richard-ramos:


chaitanyaprem (approved):
LGTM

Although I am not familiar with the DB flow.

PR #4373, opened 2023-11-23
feat: show activity center notification if user must reveal addressed to join/rejoin the community

### LINKED ISSUE #11974: If a user is asked by the owner for their shared addresses and they have none set up, we should prompt the user
### Description

Dependent on https://github.com/status-im/status-desktop/issues/11969

In the above issue, the owner node will ask the users' nodes what their shared addresses are (all in the background). It,s possible however that the user never shared them (very old account), or more probable, they lost their account and re-imported.

If that's the case, when the user's node gets the signal to get the shared addresses, the app should add a notification of some sort asking to select the shared addresses.

I propose a notification in the activity center, but I'll leave it up to designers.

Once that prompt/notification is clicked by the user, it should make them go to the "edit shared addresses" modal, which will send the shared addresses correctly to the owner

### Acceptance Criteria

- There is a prompt/notification that is shown to users that have no shared addresses in their DB for a specific community when asked by the owner
- When clicked, that prompt goes to the edit shared addresses modal

mprakhov:
When the community ownership changed, all members are kicked. If this member does not have request to join with revealed addresses or revealed addresses is not signed - show the a notification in activity center to share their revealed accounts

Closes: https://github.com/status-im/status-desktop/issues/11974

status-desktop PR: https://github.com/status-im/status-desktop/pull/12856

### DIFF
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -3,6 +3,7 @@ package protocol
 import (
 	"context"
 	"crypto/ecdsa"
+	"database/sql"
 	"errors"
 	"fmt"
 	"sync"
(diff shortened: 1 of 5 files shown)


protocol/messenger_communities.go
```diff
@@ -3096,9 +3102,30 @@ func (m *Messenger) sendSharedAddressToControlNode(receiver *ecdsa.PublicKey, co
...
+	for _, account := range requestToJoin.RevealedAccounts {
+		revealedAccountSigned = len(account.Signature) > 0
+		if revealedAccountSigned {
+			break
+		}
+	}
+
+	if !revealedAccountSigned {
+		return nil, communities.ErrNoRevealedAccountsSignature
+	}
```
mprakhov, line 3127: it would be nice to include this ticket to RC 0.15.x or we need to move this selected code to RC 0.15.x

this is a bug when a user without a revealed address or without signature in the revealed address (old DB accounts) can send an incorrect request to join, which will be declined by an owner

jrainville: The rest of the code doesn't seem to introduce regressions, so it should be ok to put in the release.

jrainville (approved):
LGTM

PR #4377, opened 2023-11-24
feat: 200 years long media tls cert

yqrashawn:
-------
mobile issue https://github.com/status-im/status-mobile/issues/17789

media server TSL cert validity from 
now -> 1 year later
to
100 years before -> 100 years later

no idea if there's any security concern about this

### DIFF
--- a/server/certs.go
+++ b/server/certs.go
@@ -13,8 +13,6 @@ import (
 	"time"
 
 	"github.com/ethereum/go-ethereum/log"
-
-	"github.com/status-im/status-go/timesource"
 )
 
 var globalMediaCertificate *tls.Certificate = nil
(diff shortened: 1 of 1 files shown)


server/certs.go
```diff
@@ -82,8 +82,9 @@ func generateMediaTLSCert() error {
 		return nil
 	}
 
-	notBefore := timesource.GetCurrentTime()
-	notAfter := notBefore.Add(365 * 24 * time.Hour)
+	now := timesource.GetCurrentTime()
```
qfrank: If the time range has expanded significantly, we may no longer need an accurate time. In this case, using the built-in time.Now() function might be preferable because the timesource may require network access.

qfrank (approved):

server/certs.go
yqrashawn: fixed, thanks

PR #4381, opened 2023-11-27
chore: fix peer exchange and unit testing

### LINKED ISSUE #4343: Add test for Wakuv2: runPeerExchangeLoop
Add a test for `wakuv2/waku.go` to test the functionality of the waku's node peer exchange.
Posibly in here:
[err := w.node.PeerExchange().Request(w.ctx, peersToDiscover, peer_exchange.WithAutomaticPeerSelection(withThesePeers...))](https://github.com/status-im/status-go/blob/3140a0bc5380a4cb6266e3922969095204fb89dc/wakuv2/waku.go#L611)

kaichaosun:
Add unit test for peer exchange, and fix it.

Important changes:
- [x] add `ethdisc.Resolver` config for local DNS resolver
- [x] start PeerExchange when enabled in server side
- [x] unit test for PeerExchange scenario
- [ ]  ~not close peer if light client, **Note: not sure if this is the expected behaviour?**~

Closes https://github.com/status-im/status-go/issues/4343


vendor/github.com/waku-org/go-waku/waku/v2/dnsdisc/enr.go
```diff
@@ -44,16 +44,19 @@ func init() {
 }
 
 // RetrieveNodes returns a list of multiaddress given a url to a DNS discoverable ENR tree
-func RetrieveNodes(ctx context.Context, url string, opts ...DNSDiscoveryOption) ([]DiscoveredNode, error) {
+func RetrieveNodes(ctx context.Context, url string, resolver dnsdisc.Resolver, opts ...DNSDiscoveryOption) ([]DiscoveredNode, error) {
```
richard-ramos: Instead of adding a new mandatory `resolver` parameter, this can be implemented by adding a new `func WithResolver(resolver *net.Resolver) DNSDiscoveryOption`  in this file. That way if someone does not need to use an specific resolver, they don't have to even worry about figuring out what value to pass in `resolver`, and you would not need to have a `dnsResolver` attribute in the `Waku` type

wakuv2/waku.go
```diff
@@ -478,13 +480,13 @@ func (w *Waku) identifyAndConnect(ctx context.Context, isLightClient bool, peerI
...
-			}
-			return
-		}
+		// if isLightClient {
+		// 	err = w.node.Host().Network().ClosePeer(peerInfo.ID)
+		// 	if err != nil {
+		// 		w.logger.Error("could not close connections to peer", zap.Stringer("peer", peerInfo.ID), zap.Error(err))
+		// 	}
+		// 	return
+		// }
```
richard-ramos: Why is this commented?

richard-ramos (changes requested):
The idea looks good. These changes must be done in go-waku directly and not in the vendor/ folder. Also left few comments on how to implement these changes in go-waku.

wakuv2/waku.go
kaichaosun: When debugging the tests previously, this logic makes the connected peer always 0.
But now it actually not the case. I will bring it back.
I'm wondering, even the peer is close here, a light client can still connect the closed peers later which is stored in peer store manager, right?

richard-ramos: Yes

kaichaosun:
waiting for: https://github.com/status-im/status-go/pull/4407

richard-ramos (approved):

Issue #2203, opened 2021-04-15
Storing references into sync.Maps after loading and editing that reference


Samyoul:
For a little context see the discussion here https://github.com/status-im/status-go/pull/2199#discussion_r613871484

I've introduced a lot of comments into the code base that read `// TODO(samyoul) remove storing of an updated reference pointer?`.

We've recently changed `Messenger`'s use of `map`s to use custom `sync.Map`s and removed most usages of `Messenger.mutex`. While doing this I noticed that in a lot of cases we load a reference from a map and then perform some update on it and then save the reference back into the map.

In principle we are getting a reference from a map and updating the reference, then we give the reference back to the map, meaning we aren't updating the map's value and therefore don't need to store the reference back into the map.

We need to answer the question of what do we want to do?

- Do we want to store something after updating a reference pointer?
  - If so why?
- Do we want to remove all storing of updated references?
  - Are we happy without an explicit "update"?

igor-sirotin:
Closing as stale

PR #4382, opened 2023-11-27
feature(@desktop/chat) Enhance message context menu with mark as unread

kounkou:
### Description

This is the Golang repo PR matching PR : 12879

Feature to enhance message context menu with mark as unread.

[Figma design](https://www.figma.com/file/Mr3rqxxgKJ2zMQ06UAKiWL/%F0%9F%92%AC-Chat%E2%8E%9CDesktop?type=design&node-id=14632-460085&mode=design&t=9iVc38RXY5QVm5uz-0)

fixes #10329

### What does the PR do

The PR does the following : 

- Adds capacity to mark a message as unread
- Adds capacity to mark a message with mention as unread
- Adds persistence to the marking of the message (change can be seen at after reboot)
- Adds marking in right click contextual menu

### Affected areas

- Chat screen
- Contextual menu (Right click 

### Screenshot of functionality (including design for comparison)

output

### DIFF
--- a/protocol/message_persistence.go
+++ b/protocol/message_persistence.go
@@ -1891,6 +1891,99 @@ func (db sqlitePersistence) MarkMessagesSeen(chatID string, ids []string) (uint6
 	return countWithMentions + countNoMentions, countWithMentions, err
 }
 
+func (db sqlitePersistence) GetMessageIdsWithGreaterTimestamp(chatID string, messageID string) ([]string, error) {
+	var err error
+	var rows *sql.Rows
+	query := "SELECT id FROM user_messages WHERE local_chat_id = ? AND timestamp >= (SELECT timestamp FROM user_messages WHERE id = ?)"
+
+	rows, err = db.db.Query(query, chatID, messageID)
+
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	var ids []string
+
+	for rows.Next() {
+		var messageID string
+		err = rows.Scan(&messageID)
+		if err != nil {
+			return nil, err
+		}
+		ids = append(ids, messageID)
+	}
+
+	return ids, nil
+}
+
+func (db sqlitePersistence) MarkMessageAsUnread(chatID string, messageID string) (uint64, uint64, error) {
+	tx, err := db.db.BeginTx(context.Background(), &sql.TxOptions{})
+	if err != nil {
+		return 0, 0, err
+	}
+	defer func() {
+		if err == nil {
+			err = tx.Commit()
+			return
+		}
+		// don't shadow original error
+		_ = tx.Rollback()
+	}()
+
+	// TODO : Reduce number of queries for getting (total unread messages, total messages with mention)
+	// The function expected result is a pair (total unread messages, total messages with mention)
+	// Currently a 2 step operation is needed to obtain this pair
+	_, err = tx.Exec(`UPDATE user_messages SET seen = 1 WHERE local_chat_id = ? AND NOT(seen)`, chatID)
+	if err != nil {
+		return 0, 0, err
+	}
+
+	_, err = tx.Exec(
+		`UPDATE user_messages
+			SET seen = 0
+			WHERE local_chat_id = ?
+			AND seen = 1
+			AND (mentioned OR replied)
+			AND timestamp >= (SELECT timestamp FROM user_messages WHERE id = ?)`, chatID, messageID)
+	if err != nil {
+		return 0, 0, err
+	}
+
+	var countWithMentions uint64
+	row := tx.QueryRow("SELECT changes();")
+	if err := row.Scan(&countWithMentions); err != nil {
+		return 0, 0, err
+	}
+
+	_, err = tx.Exec(
+		`UPDATE user_messages 
+			SET seen = 0 
+			WHERE local_chat_id = ? 
+			AND seen = 1 
+			AND NOT(mentioned OR replied) 
+			AND timestamp >= (SELECT timestamp FROM user_messages WHERE id = ?)`, chatID, messageID)
+	if err != nil {
+		return 0, 0, err
+	}
+
+	var countNoMentions uint64
+	row = tx.QueryRow("SELECT changes();")
+	if err := row.Scan(&countNoMentions); err != nil {
+		return 0, 0, err
+	}
+
+	count := countWithMentions + countNoMentions
+
+	_, err = tx.Exec(
+		`UPDATE chats
+            SET unviewed_message_count = ?, unviewed_mentions_count = ?,
+			highlight = 0
+			WHERE id = ?`, count, countWithMentions, chatID)
+
+	return count, countWithMentions, err
+}
+
 func (db sqlitePersistence) UpdateMessageOutgoingStatus(id string, newOutgoingStatus string) error {
 	_, err := db.db.Exec(`
 		UPDATE user_messages
--- a/protocol/messenger.go
+++ b/protocol/messenger.go
@@ -4138,6 +4138,49 @@ func (m *Messenger) DeleteMessagesByChatID(id string) error {
 	return m.persistence.DeleteMessagesByChatID(id)
 }
 
+func (m *Messenger) MarkMessageAsUnreadImpl(chatID string, messageID string) (uint64, uint64, error) {
+	count, countWithMentions, err := m.persistence.MarkMessageAsUnread(chatID, messageID)
+
+	if err != nil {
+		return 0, 0, err
+	}
+
+	chat, err := m.persistence.Chat(chatID)
+	if err != nil {
+		return 0, 0, err
+	}
+	m.allChats.Store(chatID, chat)
+	return count, countWithMentions, nil
+}
+
+func (m *Messenger) MarkMessageAsUnread(chatID string, messageID string) (uint64, uint64, []*ActivityCenterNotification, error) {
+	count, countWithMentions, err := m.MarkMessageAsUnreadImpl(chatID, messageID)
+
+	if err != nil {
+		return 0, 0, nil, err
+	}
+
+	ids, err := m.persistence.GetMessageIdsWithGreaterTimestamp(chatID, messageID)
+
+	if err != nil {
+		return 0, 0, nil, err
+	}
+
+	hexBytesIds := []types.HexBytes{}
+	for _, id := range ids {
+		hexBytesIds = append(hexBytesIds, types.FromHex(id))
+	}
+
+	updatedAt := m.GetCurrentTimeInMillis()
+	notifications, err := m.persistence.MarkActivityCenterNotificationsUnread(hexBytesIds, updatedAt)
+
+	if err != nil {
+		return 0, 0, nil, err
+	}
+
+	return count, countWithMentions, notifications, nil
+}
+
 // MarkMessagesSeen marks messages with `ids` as seen in the chat `chatID`.
 // It returns the number of affected messages or error. If there is an error,
 // the number of affected messages is always zero.
--- a/protocol/messenger_test.go
+++ b/protocol/messenger_test.go
@@ -6,7 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"math/big"
 	"os"
 	"strings"
@@ -2472,7 +2472,7 @@ func buildImageWithAlbumIDMessage(chat Chat, albumID string) (*common.Message, e
 	}
 	defer file.Close()
 
-	payload, err := ioutil.ReadAll(file)
+	payload, err := io.ReadAll(file)
 	if err != err {
 		return nil, err
 	}
--- a/protocol/messenger_unread_test.go
+++ b/protocol/messenger_unread_test.go
@@ -1,0 +1,360 @@
+package protocol
+
+import (
+	"context"
+	"errors"
+
+	_ "github.com/mutecomm/go-sqlcipher/v4" // require go-sqlcipher that overrides default implementation
+
+	"github.com/status-im/status-go/eth-node/types"
+	"github.com/status-im/status-go/protocol/common"
+	"github.com/status-im/status-go/protocol/protobuf"
+	"github.com/status-im/status-go/protocol/requests"
+	"github.com/status-im/status-go/protocol/tt"
+)
+
+func (s *MessengerSuite) checkMessageSeen(messageID string, expectedSeen bool) {
+	message, err := s.m.MessageByID(messageID)
+
+	s.Require().NoError(err)
+	s.Require().Equal(expectedSeen, message.Seen)
+}
+
+func (s *MessengerSuite) retrieveAllWithRetry(errorMessage string) (*MessengerResponse, error) {
+	var response *MessengerResponse
+	var err error
+
+	retryFunc := func() error {
+		response, err = s.m.RetrieveAll()
+		if err != nil {
+			return err
+		}
+		if len(response.messages) == 0 {
+			return errors.New(errorMessage)
+		}
+		return nil
+	}
+
+	err = tt.RetryWithBackOff(retryFunc)
+	return response, err
+}
+
+func (s *MessengerSuite) TestMarkMessageAsUnreadWhenMessageListContainsSingleMessage() {
+	chat := CreatePublicChat("test-chat-1", s.m.transport)
+
+	chat.UnviewedMessagesCount = 2
+	chat.UnviewedMentionsCount = 2
+	chat.Highlight = true
+
+	err := s.m.SaveChat(chat)
+	s.Require().NoError(err)
+
+	inputMessage1 := buildTestMessage(*chat)
+	inputMessage1.ID = "1"
+	inputMessage1.Seen = true
+	inputMessage1.Mentioned = true
+
+	err = s.m.SaveMessages([]*common.Message{inputMessage1})
+	s.Require().NoError(err)
+
+	_, err = s.m.MarkAllRead(context.Background(), chat.ID)
+	s.Require().NoError(err)
+
+	chats := s.m.allChats
+
+	actualChat, ok := chats.Load(chat.ID)
+
+	s.Require().True(ok)
+	s.Require().Equal(uint(0), actualChat.UnviewedMessagesCount)
+	s.Require().Equal(uint(0), actualChat.UnviewedMentionsCount)
+	s.Require().False(actualChat.Highlight)
+	s.checkMessageSeen(inputMessage1.ID, true)
+
+	count, countWithMentions, notifications, err := s.m.MarkMessageAsUnread(chat.ID, inputMessage1.ID)
+	s.Require().NoError(err)
+	s.Require().Equal(uint64(1), count)
+	s.Require().Equal(uint64(1), countWithMentions)
+	s.Require().Len(notifications, 0)
+
+	chats = s.m.allChats
+
+	actualChat, ok = chats.Load(chat.ID)
+	s.Require().True(ok)
+	s.Require().Equal(uint(1), actualChat.UnviewedMessagesCount)
+	s.Require().Equal(uint(1), actualChat.UnviewedMentionsCount)
+	s.Require().False(actualChat.Highlight)
+	s.checkMessageSeen(inputMessage1.ID, false)
+}
+
+func (s *MessengerSuite) TestMarkMessageAsUnreadWhenMessageListContainsSeveralMessages() {
+	chat := CreatePublicChat("test-chat-2", s.m.transport)
+
+	chat.UnviewedMessagesCount = 0
+	chat.UnviewedMentionsCount = 0
+	chat.Highlight = true
+
+	err := s.m.SaveChat(chat)
+	s.Require().NoError(err)
+
+	inputMessage1 := buildTestMessage(*chat)
+	inputMessage1.ID = "1"
+	inputMessage1.Seen = true
+	inputMessage1.Mentioned = true
+	inputMessage1.Timestamp = 1
+
+	inputMessage2 := buildTestMessage(*chat)
+	inputMessage2.ID = "2"
+	inputMessage2.Seen = true
+	inputMessage2.Mentioned = false
+	inputMessage2.Timestamp = 2
+
+	inputMessage3 := buildTestMessage(*chat)
+	inputMessage3.ID = "3"
+	inputMessage3.Seen = true
+	inputMessage3.Mentioned = false
+	inputMessage3.Timestamp = 3
+
+	err = s.m.SaveMessages([]*common.Message{
+		inputMessage1,
+		inputMessage2,
+		inputMessage3,
+	})
+	s.Require().NoError(err)
+
+	_, err = s.m.MarkAllRead(context.Background(), chat.ID)
+	s.Require().NoError(err)
+
+	chats := s.m.allChats
+	actualChat, ok := chats.Load(chat.ID)
+
+	s.Require().True(ok)
+	s.Require().Equal(uint(0), actualChat.UnviewedMessagesCount)
+	s.Require().Equal(uint(0), actualChat.UnviewedMentionsCount)
+	s.Require().False(actualChat.Highlight)
+	s.checkMessageSeen(inputMessage1.ID, true)
+	s.checkMessageSeen(inputMessage2.ID, true)
+	s.checkMessageSeen(inputMessage3.ID, true)
+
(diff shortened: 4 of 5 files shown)


osmaczko (approved):
Nice, congrats for first status-go contribution :+1:

protocol/message_persistence.go
```diff
@@ -1891,6 +1891,58 @@ func (db sqlitePersistence) MarkMessagesSeen(chatID string, ids []string) (uint6
...
+		return 0, 0, err
+	}
+
+	var countWithMentions uint64
+	row := tx.QueryRow("SELECT changes();")
+	if err := row.Scan(&countWithMentions); err != nil {
+		return 0, 0, err
+	}
+
+	_, err = tx.Exec(`UPDATE user_messages SET seen = 0 WHERE seen = 1 AND NOT(mentioned) AND NOT(replied) AND id = ?`, messageId)
```
jrainville: I think the reason why the count of unread mentions doesn't update correctly is because you are only marking as unread the one message you selected.

I think you're only missing a SELET to get the timestamp of the message that was clicked, then instead of updating using the `id`, use a `> timestamp` to make sure all newer messages as **also** marked as unread!

jrainville (commented):
Great work! I added a comment inline that explains what might work to make the mentions count work correctly.

protocol/message_persistence.go
kounkou: > I think the reason why the count of unread mentions doesn't update correctly is because you are only marking as unread the one message you selected.
> 
> I think you're only missing a SELET to get the timestamp of the message that was clicked, then instead of updating using the `id`, use a `> timestamp` to make sure all newer messages as **also** marked as unread!

oh... that's interesting perspective. Thanks, let me have a look

kounkou:
Working on fixing the unittests. 

Edit : 
Seems unrelated failures

```
TestMessengerCommunitiesTokenPermissionsSuite – github.com/status-im/status-go/protocol
TestMessengerCommunitiesTokenPermissionsSuite/TestBecomeMemberPermissions – github.com/status-im/status-go/protocol
TestMessengerCommunitiesTokenPermissionsSuite/TestCommunityTokensMetadata – github.com/status-im/status-go/protocol
TestMessengerCommunitiesTokenPermissionsSuite/TestCreateTokenPermission – github.com/status-im/status-go/protocol
TestMessengerCommunitiesTokenPermissionsSuite/TestEditSharedAddresses – github.com/status-im/status-go/protocol
TestMessengerCommunitiesTokenPermissionsSuite/TestEditTokenPermission – github.com/status-im/status-go/protocol
```

Those tests are flaky. After multiple retries, the tests are Pass

protocol/messenger.go
```diff
@@ -4138,6 +4138,10 @@ func (m *Messenger) DeleteMessagesByChatID(id string) error {
 	return m.persistence.DeleteMessagesByChatID(id)
 }
 
+func (m *Messenger) MarkMessageAsUnread(chatID string, messageID string) (uint64, uint64, error) {
+	return m.persistence.MarkMessageAsUnread(chatID, messageID)
```
igor-sirotin: Should we update `m.allChats` here as done in `markMessagesSeenImpl`?
https://github.com/status-im/status-go/blob/6197ed1bdb4b14ec7454e34086bafb2cf0f4952c/protocol/messenger.go#L4139-L4143

protocol/message_persistence.go
```diff
@@ -1891,6 +1891,74 @@ func (db sqlitePersistence) MarkMessagesSeen(chatID string, ids []string) (uint6
...
+	defer func() {
+		if err == nil {
+			err = tx.Commit()
+			return
+		}
+		// don't shadow original error
+		_ = tx.Rollback()
+	}()
+
+	_, err = tx.Exec(`UPDATE user_messages SET seen = 1 WHERE NOT(seen)`)
```
igor-sirotin: Looks like this line will mark ALL of the messages as `seen`.
Even the ones from different chats.

protocol/message_persistence.go
```diff
@@ -1891,6 +1891,74 @@ func (db sqlitePersistence) MarkMessagesSeen(chatID string, ids []string) (uint6
...
+	if err != nil {
+		return 0, 0, err
+	}
+
+	_, err = tx.Exec(
+		`UPDATE user_messages 
+			SET seen = 0 
+			WHERE seen = 1 
+			AND (mentioned OR replied)
```
igor-sirotin: It took me awhile to get why we do 2 separate requests for `mentioned or replied` messages.
Maybe instead we could do it in a more straightforward way?
1. `UPDATE ... SET seen = 1 WHERE timestamp > ...`
2. `SELECT COUNT(1) FROM user_messages WHERE seen = 0`
3. `SELECT COUNT(1) FROM user_messages WHERE seen = 1`

IMO the intentions would be more obvious like thie.
Also, if I get it right, we will cut the number of db requests from 4 to 3 (if `SELECT changes()` are counted).

igor-sirotin (changes requested):
Good job, congrats on first status-go PR! 🎉 
Left some comments, some of them look very important.

Also, do you think we could implement a test for this feature?
Looks like not too complicated one, should be a good option for a first status-go test as well 🙂

protocol/message_persistence.go
kounkou: Thanks for the comment, I got some doubts, and as discussed offline, I went into testing this. As a prerequesite, I followed the notion document to be able to decrypt messages in database.
Here are the steps I went through for the testing:


I made a test account test-jacques have 2 contacts (Jacques and Tester1) to chat with.
I created 2 chats, to get messages from those 2 accounts
Then I sent some messages,
Here I ran a query after decrypting messages in database :
        
```bash
sqlite> select text, seen from user_messages;
You sent a contact request to @0x04eeb4914b7eb91000de56f00fcb78845d461d0cacd84ba6cb2ba1c149665326fafcfb4c00cb7fa88c3c95b245eaaf8ae1a98bb89c67d98ef46a1c70c7bb4c131e|1
Hi !|1
Tester1 : in here|1
Tester1 : ahah|1
test-jacques : being the boss|0
test-jacques : I am the one|0
Tester1 : hello|1
Tester1 : I|1
test-jacques : hello|0
test-jacques : yo!|0
test-jacques : ahah|0
Tester1 : am|1
Tester1 : the boss|1

```

then from test-jacques, I went into one of the chats with Test1 and moved the marker around,
the expectation is that all the messages will be marked as seen

```bash
sqlite> select text, seen from user_messages;
You sent a contact request to @0x04eeb4914b7eb91000de56f00fcb78845d461d0cacd84ba6cb2ba1c149665326fafcfb4c00cb7fa88c3c95b245eaaf8ae1a98bb89c67d98ef46a1c70c7bb4c131e|1
Hi !|1
Tester1 : in here|1
Tester1 : ahah|0
test-jacques : being the boss|0
test-jacques : I am the one|0
Tester1 : hello|1
Tester1 : I|1
test-jacques : hello|0
test-jacques : yo!|0
test-jacques : ahah|0
Tester1 : am|1
Tester1 : the boss|1
```

Conclusion : 

We can see we do not impact the other chat here for which the messages start with `test-jacques`
So why is it the case ? 
The reason is that at the end of the my function I do the following which restrict my query to the provided chatId.
I hope this answers your concern

```sql
	// Update denormalized count
	_, err = tx.Exec(
		`UPDATE chats
              	SET unviewed_message_count =
		   (SELECT COUNT(1)
		   FROM user_messages
		   WHERE local_chat_id = ? AND seen = 0),
		   unviewed_mentions_count =
		   (SELECT COUNT(1)
		   FROM user_messages
		   WHERE local_chat_id = ? AND seen = 0 AND (mentioned OR replied)),
                   highlight = 0
		WHERE id = ?`, chatID, chatID, chatID)
	return countWithMentions + countNoMentions, countWithMentions, err
```

protocol/message_persistence.go
kounkou: Ah, I was following the same pattern as other code I learnt from, but let me test having a single query here 👍🏾

protocol/messenger.go
kounkou: I am not sure why we need this? Could you explain why this is needed please?

protocol/message_persistence.go
osmaczko: > The reason is that at the end of the my function I do the following which restrict my query to the provided chatId.

Isn't it the restriction for different query? :thinking:

protocol/messenger.go
kounkou: After second thought I understand why we need to store the the chat in the map. Everything is explained here : https://github.com/status-im/status-go/issues/2203
This is legacy code that we might want to work on to as it takes mental effort to understand why we should put this line in the first place.

protocol/message_persistence.go
kounkou: I will add another layer of protection by using the chatId.
But let me know if you see any flaws in my testing please
I will allocate time for myself to understand why the experiment doesn't show any impact at all

protocol/message_persistence.go
kounkou: I tried as my best on this one, but I couldn't make the provided idea work. Let's discuss offline if needed, for now I have added a TODO note for this item, and submitted changes for the other comments to make progress.

kounkou:
> Good job, congrats on first status-go PR! 🎉 Left some comments, some of them look very important.
> 
> Also, do you think we could implement a test for this feature? Looks like not too complicated one, should be a good option for a first status-go test as well 🙂

Implemented unit-tests in the latest revision, and added some comments to get everybody up to speed for understanding what's happening. cheers!

protocol/message_persistence.go
kounkou: I have created a test :
see : 
`s.T().Run("Test marking a message as unread in a chat does not impact other chat", func(t *testing.T) {` 

Which helps me show that if we do not make the queries specific to a chatID, we can impact other chats as the messages inside the user_messages table contains messages from different chats.
I have also launched faulty commands 1 by 1, and could see how not providing the chatID impacts other chats.
So I believe this topic can be closed.

protocol/messenger_test.go
```diff
@@ -2540,3 +2540,237 @@ func (s *MessengerSuite) TestSendMessageMention() {
...
+
+		err = s.m.SaveMessages([]*common.Message{inputMessage1})
+		s.Require().NoError(err)
+
+		_, err = s.m.MarkAllRead(context.Background(), chat.ID)
+		s.Require().NoError(err)
+
+		chats := s.m.Chats()
+
+		for idx := range chats {
```
igor-sirotin: This function can be used to avoid using a loop:
https://github.com/status-im/status-go/blob/3ab1afaae8c01c3228a8349fb606b767e7d697dc/protocol/messenger_chats.go#L135-L139

protocol/messenger_test.go
```diff
@@ -2540,3 +2540,237 @@ func (s *MessengerSuite) TestSendMessageMention() {
...
+
+		chats = s.m.Chats()
+
+		for idx := range chats {
+			if chats[idx].ID == chat.ID {
+				s.Require().Equal(uint(1), chats[idx].UnviewedMessagesCount)
+				s.Require().Equal(uint(1), chats[idx].UnviewedMentionsCount)
+				s.Require().Equal(false, chats[idx].Highlight)
+			}
+		}
```
igor-sirotin: Here you check 2 things:
- return value of `MarkMessageAsUnread`
- chat's `UnviewedMessagesCount` and `UnviewedMentionsCount`

It's good, but I think the most important part is to check that the messages `Seen` value was indeed updated. And that it's correct for each message. You can use `MessageByID` function for this. I think you have IDs of all saved messages.

protocol/messenger_test.go
```diff
@@ -2540,3 +2540,237 @@ func (s *MessengerSuite) TestSendMessageMention() {
...
+		chat1.UnviewedMessagesCount = 2
+		chat1.UnviewedMentionsCount = 0
+		chat1.Highlight = true
+
+		inputMessage1 := buildTestMessage(*chat1)
+		inputMessage1.ID = "1"
+		inputMessage1.Seen = false
+		inputMessage1.Mentioned = false
+
+		time.Sleep(1000 * time.Millisecond)
```
igor-sirotin: a leftover or a hack?

protocol/messenger_test.go
```diff
@@ -2540,3 +2540,237 @@ func (s *MessengerSuite) TestSendMessageMention() {
...
+		inputMessage3.Seen = false
+		inputMessage3.Mentioned = false
+
+		err = s.m.SaveMessages([]*common.Message{inputMessage3})
+		s.Require().NoError(err)
+
+		count, countWithMentions, err := s.m.MarkMessageAsUnread(chat1.ID, inputMessage1.ID)
+
+		s.Require().NoError(err)
+		s.Require().Equal(uint(1), chat2.UnviewedMessagesCount)
```
igor-sirotin: hmm, I don't think it makes sense to check this, would this value update automatically?
Even though `chat2` is a pointer, we replace it in `allChats` with a new Chat later.

protocol/messenger_test.go
```diff
@@ -2540,3 +2540,237 @@ func (s *MessengerSuite) TestSendMessageMention() {
...
+		s.Require().Equal(uint64(0), countWithMentions)
+
+		chats := s.m.Chats()
+
+		for idx := range chats {
+			if chats[idx].ID == chat1.ID {
+				s.Require().Equal(uint(2), chats[idx].UnviewedMessagesCount)
+				s.Require().Equal(uint(0), chats[idx].UnviewedMentionsCount)
+				s.Require().Equal(false, chats[idx].Highlight)
+			} else if chats[idx].ID == chat2.ID {
```
igor-sirotin: This is not enough here.
Since you don't update chat2 in `messenger.allChats`, the value will surely remain the same.
We should check that the `Seen` property for the message in the database in correct.

protocol/messenger_test.go
```diff
@@ -2540,3 +2540,237 @@ func (s *MessengerSuite) TestSendMessageMention() {
 	s.Require().NoError(err)
 	s.Require().Equal("Alice talk to bobby", response.Notifications()[0].Message)
 }
+
+func (s *MessengerSuite) TestMarkMessageAsUnread() {
```
igor-sirotin: Overall the test look very good, thank you! I hope you enjoyed this 🙂 

I would recommend moving this to a separate `messenger_unread_test.go`, 2500 lines of code is more than enough here already 😄  And if you create a separate `MessengerUnreadSuite`, you can implement each test in a separate function and you won't have to call `s.T().Run(...)`.

protocol/message_persistence.go
```diff
@@ -1891,6 +1891,78 @@ func (db sqlitePersistence) MarkMessagesSeen(chatID string, ids []string) (uint6
...
+		return 0, 0, err
+	}
+
+	// Update denormalized count
+	_, err = tx.Exec(
+		`UPDATE chats
+              	SET unviewed_message_count =
+		   (SELECT COUNT(1)
+		   FROM user_messages
+		   WHERE local_chat_id = ? AND seen = 0),
```
igor-sirotin: hmm, but we have the `countNoMentions` and `countWithMentions` already, maybe just push it here as a parameter? Should be better on performance I guess?

protocol/message_persistence.go
igor-sirotin: I wonder why this didn't work. Let's discuss in DM.

igor-sirotin (commented):
I dont' have any concerns on the implementation now, but have some important comments on the test. Not a blocker, but please check out 🙂 

Thanks for writing a test! ❤️

protocol/messenger_test.go
kounkou: sure, I can update this. Thanks

protocol/messenger_test.go
kounkou: Fair enough, updated the test. Thanks

protocol/messenger_test.go
kounkou: was a leftover, removed. Thanks

protocol/messenger_test.go
kounkou: This test is useful for detecting regression. From my tests, if I remove the `WHERE local_chat_id = ?` from the query, I could see impact on the second chat.
I have improved the test also based on previous comments to test the .Seen property

protocol/messenger_test.go
kounkou: that's right, updated all tests to account for the same, meaning testing the .Seen property.

protocol/message_persistence.go
kounkou: Fair enough, updated the code

protocol/messenger_unread_test.go
```diff
@@ -0,0 +1,277 @@
...
+	s.Require().NoError(err)
+
+	_, err = s.m.MarkAllRead(context.Background(), chat.ID)
+	s.Require().NoError(err)
+
+	chats := s.m.allChats
+
+	actualChat, ok := chats.Load(chat.ID)
+
+	s.Require().Equal(true, ok)
```
igor-sirotin:
```suggestion
	s.Require().True(ok)
```

protocol/messenger_unread_test.go
```diff
@@ -0,0 +1,277 @@
...
+	s.Require().Equal(uint(0), actualChat.UnviewedMessagesCount)
+	s.Require().Equal(uint(0), actualChat.UnviewedMentionsCount)
+	s.Require().Equal(false, actualChat.Highlight)
```
igor-sirotin:
```suggestion
	s.Require().False(actualChat.Highlight)
```

igor-sirotin: this and all similar places below

protocol/messenger_unread_test.go
```diff
@@ -0,0 +1,277 @@
...
+
+	err := s.m.SaveChat(chat)
+	s.Require().NoError(err)
+
+	inputMessage1 := buildTestMessage(*chat)
+	inputMessage1.ID = "1"
+	inputMessage1.Seen = true
+	inputMessage1.Mentioned = true
+
+	time.Sleep(1000 * time.Millisecond)
```
igor-sirotin: still leftover? 🙂

protocol/messenger_unread_test.go
```diff
@@ -0,0 +1,277 @@
...
+	actualChat, ok := chats.Load(chat.ID)
+
+	s.Require().Equal(true, ok)
+	s.Require().Equal(uint(0), actualChat.UnviewedMessagesCount)
+	s.Require().Equal(uint(0), actualChat.UnviewedMentionsCount)
+	s.Require().Equal(false, actualChat.Highlight)
+
+	actualMessage1, err := s.m.MessageByID(inputMessage1.ID)
+	s.Require().NoError(err)
+	s.Require().Equal(true, actualMessage1.Seen)
```
igor-sirotin: So you could make a separate function and simplify the code a lot.

```go
func (s *MessengerSuite) checkMessageSeen(messageID string, expectedSeen) {
    message, err := s.m.MessageByID(messageID)
	s.Require().NoError(err)
	s.Require().Equal(expectedSeen, message.Seen)
}

s.checkMessageSeen(inputMessage1.ID, true)
s.checkMessageSeen(inputMessage2.ID, true)
s.checkMessageSeen(inputMessage3.ID, true)
```

protocol/messenger_unread_test.go
```diff
@@ -0,0 +1,277 @@
...
+	err = s.m.SaveMessages([]*common.Message{inputMessage1, inputMessage2})
+	s.Require().NoError(err)
+
+	chat2.UnviewedMessagesCount = 1
+	chat2.UnviewedMentionsCount = 0
+	chat2.Highlight = true
+
+	inputMessage3 := buildTestMessage(*chat2)
+	inputMessage3.ID = "3"
+	inputMessage3.Seen = false
```
igor-sirotin, line 242: hmmm, but it's `false` from the start of the test 🤔  Shouldn't it be `true` here?
And same for `inputMessage1` and `inputMessage2`.

protocol/messenger_unread_test.go
kounkou: I used it to simulate the timestamp as the timestamp is created automatically when creating the message and to account for the speed of unittest, but I understand I can directly overid the Timestamp to make the test more controllable here. I'll update this. thanks

protocol/messenger_unread_test.go
kounkou: This is chat2, so it's never touched 😄 so, it should be false, not true. Nonetheless I have made it obvious now that for chat1 the seen states are changing as improvement.

protocol/messenger_unread_test.go
kounkou: Sure, updated. thanks

protocol/message_persistence.go
```diff
@@ -1891,6 +1891,73 @@ func (db sqlitePersistence) MarkMessagesSeen(chatID string, ids []string) (uint6
 	return countWithMentions + countNoMentions, countWithMentions, err
 }
 
+func (db sqlitePersistence) MarkMessageAsUnread(chatID string, messageID string) (uint64, uint64, error) {
```
MishkaRogachev, line 1920: just suggestion: you can parametrise `MarkMessagesSeen` with boolean seen/unseen flag instead of this function

protocol/messenger.go
```diff
@@ -4138,6 +4138,31 @@ func (m *Messenger) DeleteMessagesByChatID(id string) error {
...
+	return count, countWithMentions, nil
+}
+
+func (m *Messenger) MarkMessageAsUnread(chatID string, messageID string) (uint64, uint64, error) {
+	count, countWithMentions, err := m.MarkMessageAsUnreadImpl(chatID, messageID)
+
+	if err != nil {
+		return 0, 0, err
+	}
+
```
MishkaRogachev, line 4162: i think it's right to also check activity center here. If persistence has AC notification with `id == messageID` it also should be marked as unseen. Basically same as here: https://github.com/status-im/status-go/blob/develop/protocol/messenger.go#L4173

igor-sirotin (approved):
🎉

protocol/message_persistence.go
kounkou: MarkAsUnread is more than just making a bunch of items as unread. The difference is that with MarkAsUnread, we also need to mark every items with smaller timestamp than the current element as read. The name mark as read is still valid though. So that reason I think a separate function makes more sense.

protocol/messenger.go
kounkou: > i think it's right to also check activity center here. If persistence has AC notification with `id == messageID` it also should be marked as unseen. Basically same as here: https://github.com/status-im/status-go/blob/develop/protocol/messenger.go#L4173

Let me have a look.

MishkaRogachev: Agh i probably got the idea, `persistence.MarkMessageAsUnread` function marks as unread not the message with provided id, but all seen in the chat starting with a timestamp of the message with provided id. Name is quite confusing..

Wouldn't it be easier with this approach:
1) Rename the function to `MarkMessagesAsUnreadSinceTimestamp`, and pass the desired timestamp to the status-go instead of the message id
2) Call `persistence.MarkMessagesAsUnreadSinceTimestamp`, which will update the `user_messages` and `chats` tables and return either a slice of messages or a slice of message id, offected by updates.
3) Call `persistence.MarkActivityCenterNotificationsUnread`, with the message ids we got in the previous call

kounkou: > Agh i probably got the idea, `persistence.MarkMessageAsUnread` function marks as unread not the message with provided id, but all seen in the chat starting with a timestamp of the message with provided id. Name is quite confusing..
> 
> Wouldn't it be easier with this approach:
> 
> 1. Rename the function to `MarkMessagesAsUnreadSinceTimestamp`, and pass the desired timestamp to the status-go instead of the message id
> 2. Call `persistence.MarkMessagesAsUnreadSinceTimestamp`, which will update the `user_messages` and `chats` tables and return either a slice of messages or a slice of message id, offected by updates.
> 3. Call `persistence.MarkActivityCenterNotificationsUnread`, with the message ids we got in the previous call

1.
It's better to pass the messageId because passing the timestamp can be flawed. The source of truth for the timestamps is in the table. That's why I pass the messageId

Thanks for the suggestions

2.
For performance, it's better to keep those 2 separate because getting the messageIds or the Ids it requires to go over the results. To keep the `MarkMessageAsUnread` simple with regards to performance, it's better to not include the getTimestamps.

3. This is done already

MishkaRogachev (approved):
LGTM!

PR #4384, opened 2023-11-27
ci: publish tests-nightly log/xml for debug

yakimant:


_assets/ci/Jenkinsfile.tests
```diff
@@ -115,6 +115,11 @@ pipeline {
         reportName:             'Reports',
         reportTitles:           'Test Stats'
       ]
+      script {
+        if (isTestNightlyJob()) {
+          archiveArtifacts artifacts: '**/report.xml, **/test.log'
```
jakubgs, line 120: You could just do:
```suggestion
          archiveArtifacts(artifacts: '**/report.xml, **/test.log', allowEmptyArchive: true)
```
And avoid the conditional entirely.
Also, the whole `always` block might as well start with a `script` block.

_assets/ci/Jenkinsfile.tests
```diff
@@ -125,22 +130,30 @@ pipeline {
...
 } // pipeline
 
-def getDefaultUnitTestCount() {
+def isTestNightlyJob() {
   if (env.JOB_BASE_NAME == 'tests-nightly') {
+    return true
+  } else {
+    return false
+  }
+}
```
jakubgs: What if we make those lambdas for brevity?
```suggestion
def isTestNightlyJob = { env.JOB_BASE_NAME == 'tests-nightly' ? true : false }
```
Same for other functions.

jakubgs (commented):
Cool.

_assets/ci/Jenkinsfile.tests
yakimant: I don't think we need to publish xml and log files for PR runs.
They use space, that's why we decided to hide them from STDOUT.

_assets/ci/Jenkinsfile.tests
yakimant: Maybe this would work even: `def isTestNightlyJob = { env.JOB_BASE_NAME == 'tests-nightly' }`

yakimant: Hm closure didn't work for some reason:
```
java.lang.NoSuchMethodError: No such DSL method 'getDefaultTimeout' found among steps
```

```groovy
timeout(time: getDefaultTimeout(), unit: 'MINUTES')
...
def getDefaultTimeout = { isTestNightlyJob() ? 5*60 : 40 }
```

_assets/ci/Jenkinsfile.tests
jakubgs: I see, fair enough. How big are they tho?

_assets/ci/Jenkinsfile.tests
jakubgs: Yes, it would, i was more showing it as an example for the non-boolean functions that return numbers.

And why it fails... weird. I can verify it works with just groovy:
```sh
 > cat lambda.groovy 
#!/usr/bin/env nix-shell
#!nix-shell -i groovy -p groovy

def testFunc = { "x" == "y" }
println(testFunc())

def testArgFunc = { x -> "string: ${x}" }
println(testArgFunc("wat"))

 > ./lambda.groovy 
false
string: wat
```

_assets/ci/Jenkinsfile.tests
```diff
@@ -125,23 +130,10 @@ pipeline {
...
-}
+def getDefaultUnitTestCount() { isTestNightlyJob() ? '20' : '1' }
 
-def getDefaultTimeout() {
-  if (env.JOB_BASE_NAME == 'tests-nightly') {
-    return 5*60
-  }
-  return 40
-}
+def getDefaultUnitTestFailfast() { isTestNightlyJob() ? false : true }
```
jakubgs, line 137: I wonder if you could just use `!isTestNightlyJob()` directly in `defaultValue`. If not then just here.

_assets/ci/Jenkinsfile.tests
yakimant: I think they will be twice big as they were in stdout before:
https://github.com/status-im/status-go/issues/3543#issuecomment-1717264766
So I estimate 100M per PR build.

For nightly it will be 100M * 20 = 2G

_assets/ci/Jenkinsfile.tests
yakimant: Yeah, I tried in `groovysh` too, but got an error in Jenkins.

I dindn't investigate and research why, just changed to functions and it worked again:

from
```
def getDefaultTimeout = { isTestNightlyJob() ? 5*60 : 40 }
```
to
```
def getDefaultTimeout() { isTestNightlyJob() ? 5*60 : 40 }
```

Takes the same space and same readability.
Do you think it's worth to research on closures failure?

_assets/ci/Jenkinsfile.tests
yakimant: Let me try.

_assets/ci/Jenkinsfile.tests
jakubgs: Then the conditional makes sense. Thanks.

_assets/ci/Jenkinsfile.tests
jakubgs: Probably not. It might be something stupid tho, like order of operations.

I kinda like the functions being defined at the end tho.

jakubgs (approved):
I'm kinda surprised this format for functions works without a `return`.

PR #4387, opened 2023-11-27
Allow to sync chat for 1 month & add convenience endpoints for settings

cammellos:
This pr adds a `SyncChat` method that sync the chat for 30 days.
It also adds some custom endpoints to set some of the settings fields, instead of having to set the whole nodeconfig.
This endpoint hits messenger first, instead of the various services, to be consistent with the strategy of having messenger above any other component in the dependency chain.


protocol/messenger_mailserver.go
```diff
@@ -945,3 +945,53 @@ func (m *Messenger) ConnectionChanged(state connection.State) {
 
 	m.connectionState = state
 }
+
+func (m *Messenger) SyncChatOneMonth(chatID string) (uint32, error) {
```
jrainville: Maybe it would be better to call this just `SyncChat` and add a param for the time to sync? I just feel like if at one point our store nodes handle 2 months of data, then having to do a new function called `SyncChatTwoMonths` is a bit silly  haha.

jrainville (approved):

protocol/messenger_mailserver.go
cammellos: that's true :)

qfrank (approved):

osmaczko (approved):

PR #4388, opened 2023-11-27
Use Nix shell in the Makefile for builds

apentori:
# Description

Use nix  in Makefile and remove generated files from source control

Linked to :
* https://github.com/status-im/status-go/issues/4191

# Steps 

1. [o] Add nix scripts & config
	1. [x] Scripts 
	2. [x] Config file
	3. [x] Test build with new configuration
2. Update Makefile
	1. [x] Add nix shell definition as default shell
	2. [x] Set simple shell for tasks which doesn't need Nix Shell
	4. [x] Remove tasks to install dependency provided by the nix shell
3. [ ] Update gitignore file to remove the generated stuff


ghost:
Hey @apentori, and thank you so much for making your first pull request in status-go! :heart: Please help us make your experience better by filling out this brief questionnaire https://goo.gl/forms/uWqNcVpVz7OIopXg2

jakubgs:
I would't attempt to remove generated files in this PR. First I would try to get this to work in CI as-is.

nix/overlay.nix
```diff
@@ -0,0 +1,28 @@
...
+#
+# For more details see:
+# - https://nixos.wiki/wiki/Overlays
+# - https://nixos.org/nixos/nix-pills/callpackage-design-pattern.html
+
+self: super:
+
+let
+  /* FIXME not sure if the stdenv and other are usefull*/
+  inherit (super) stdenv stdenvNoCC callPackage;
```
jakubgs: Since none of those are used in this file then inheriting them from `super` is pointless.

Also, we might want to change the arguments for the overlay from `self` and `super` to ones that actually make some sense, for example `final` and `prev`, like the [wiki article describes](https://nixos.wiki/wiki/Overlays). That way it's more clear that:

* The `final` argument represents the state of nixpkgs after overlay is applied.
* The `prev` argument represents the state of nixpkgs before overlay is applied.

nix/pkgs.nix
```diff
@@ -0,0 +1,29 @@
...
+
+  # Status specific configuration defaults
+  default_config = {
+      allowUnfree = true;
+      android_sdk.accept_license = true;
+    }
+  # Override some packages and utilities
+  pkgsOverlay = import ./overlay.nix;
+
+
```
jakubgs: Unnecessary newlines.

nix/README.md
```diff
@@ -0,0 +1,63 @@
...
+Here is the list of variables you can use to override defaults:
+```shell
+# Build type (influences which .env file gets used for feature flags)
+export BUILD_TYPE="pr"
+export STATUS_GO_IPFS_GATEWAY_URL="https://ipfs.status.im/"
+```
+You can see the defaults in code:
+- [`nix/status-go/default.nix`](./status-go/default.nix)
+
+<!-- TODO confirm / update-->
```
jakubgs: Yes, this file will need updating.

nix/shell.nix
```diff
@@ -1,37 +1,11 @@
...
-}:
+{ config ? {}
+, pkgs ? import ./pkgs.nix { inherit config; } }:
 
 let
   inherit (pkgs) lib stdenv;
 
+
+
   /* No Android SDK for Darwin aarch64. */
```
jakubgs, line 6: Unnecessary newlines.

nix/shell.nix
```diff
@@ -1,37 +1,11 @@
...
-            "extras;android;m2repository"
-            "extras;google;m2repository"
-          ];
-        };
-      })
-    ];
-  }
-}:
+{ config ? {}
+, pkgs ? import ./pkgs.nix { inherit config; } }:
```
jakubgs, line 2: Since you are moving this file here you will need to import it at the root level of the repo using `shell.nix`.

Also, the move of `nix.conf` will also require changes in the `Makefile` to make work. And it need to be updated too.

nix/shell.nix
```diff
@@ -0,0 +1,38 @@
...
+  gomobileMod = pkgs.gomobile.override {
+    inherit xcodeWrapper;
+    withAndroidPkgs = !isMacM1;
+  };
+in pkgs.mkShell {
+  name = "status-go-shell";
+
+  buildInputs = with pkgs; [
+    git jq which
+    go_1_19 golangci-lint go-junit-report gopls go-bindata # gomobileMod
```
apentori: `gomobileMod` generate an error:

```
while evaluating attribute 'buildInputs' of derivation 'status-go-shell'
         at /nix/store/vcj90ksw5vk6ia960b73vpncxa9pw826-source/pkgs/stdenv/generic/make-derivation.nix:350:7:
          349|       depsHostHost                = lib.elemAt (lib.elemAt dependencies 1) 0;
          350|       buildInputs                 = lib.elemAt (lib.elemAt dependencies 1) 1;
             |       ^
          351|       depsTargetTarget            = lib.elemAt (lib.elemAt dependencies 2) 0;
       (stack trace truncated; use '--show-trace' to show the full trace)
       error: value is a string with context while a set was expected
```

nix/pkgs.nix
```diff
@@ -0,0 +1,47 @@
...
+      allowUnfree = true;
+      android_sdk.accept_license = true;
+  };
+  # Override some packages and utilities
+  # FIXME: doesn't work due to the use of nixpkgsSrc in androidenv. 
+  # See use it in `nix/overlay.nix`
+  #  pkgsOverlay = import ./overlay.nix { inherit nixpkgsSrc; };
+  pkgsOverlay = [
+    (self: super: {
+      androidPkgs = nixpkgsSrc.androidenv.composeAndroidPackages {
```
jakubgs:
```suggestion
      androidPkgs = super.androidenv.composeAndroidPackages {
```
This is causing this error:
```
       error: value is a string with context while a set was expected
```

nix/default.nix
```diff
@@ -0,0 +1,10 @@
+{
+    config ? {},
+    pkgs ? import ./pkgs.nix { inherit config; }
+}:
+ let
```
jakubgs, line 5:
```suggestion
let
```

nix/default.nix
```diff
@@ -0,0 +1,10 @@
...
+  # put all main targets and shells together for easy import
+  shells = pkgs.callPackage ./shell.nix { };
+in {
+  inherit pkgs shells;
```
jakubgs, line 9: Actually you could avoid the `let`/`in` block entirely by doing this:
```suggestion
  inherit pkgs;
  shell = pkgs.callPackage ./shell.nix { };
```
Also, we have only one shell, not multiple.

nix/overlay.nix
```diff
@@ -0,0 +1,25 @@
+# Override some packages and utilities in 'pkgs'
+# and make them available globally via callPackage.
+#
+# For more details see:
+# - https://nixos.wiki/wiki/Overlays
+# - https://nixos.org/nixos/nix-pills/callpackage-design-pattern.html
+self: super:
+let
+  inherit (super);
```
jakubgs: Completely unnecessary. And please use `final` and `prev`. With `final`/`prev` it's kinda obvious that `prev` is the version of `nixpkgs` before the overlay was applied and `final` is the version after the overlay was applied.

nix/pkgs.nix
```diff
@@ -0,0 +1,47 @@
...
+
+  # Status specific configuration defaults
+  defaultConfig = {
+      allowUnfree = true;
+      android_sdk.accept_license = true;
+  };
+  # Override some packages and utilities
+  # FIXME: doesn't work due to the use of nixpkgsSrc in androidenv. 
+  # See use it in `nix/overlay.nix`
+  #  pkgsOverlay = import ./overlay.nix { inherit nixpkgsSrc; };
```
jakubgs: Please remember to use this import.

nix/pkgs.nix
```diff
@@ -0,0 +1,47 @@
...
+        ];
+      };
+    })
+  ];
+
+in
+  # import nixpkgs with a config override
+  (import nixpkgsSrc) {
+    config = defaultConfig // config;
+    overlays =  pkgsOverlay;
```
jakubgs: If you're going to use the overlay from the file, you need to treat it as it is, a single overlay, and not a list.
```suggestion
    overlays = [ pkgsOverlay ];
```

shell.nix
```diff
@@ -1,66 +1,9 @@
...
-  '';
-
-  # Sandbox causes Xcode issues on MacOS. Requires sandbox=relaxed.
-  # https://github.com/status-im/status-mobile/pull/13912
-  __noChroot = stdenv.isDarwin;
-}
+  main = import ./nix { inherit config; };
+in
+  # use the default shell when calling nix-shell without arguments
+  main.shells
```
jakubgs: Without `config` - which we don't use anyway - and simplifying further we can reduce `shell.nix` just one line:
```
~/work/status-go using-nix-shell-in-makefile*
 > cat shell.nix 
import ./nix { }

~/work/status-go using-nix-shell-in-makefile*
 > nix-shell    

[nix-shell:~/work/status-go]$
```

apentori:
Building with make seems to work
```bash
 make statusgo
Missing TARGET, assuming default target. See nix/README.md for more details.
Configuring Nix shell for target 'default'...
Missing TARGET, assuming default target. See nix/README.md for more details.
Configuring Nix shell for target 'default'...
Missing TARGET, assuming default target. See nix/README.md for more details.
Configuring Nix shell for target 'default'...
go build -mod=vendor -v \
	-tags 'gowaku_no_rln' -ldflags=' -X github.com/status-im/status-go/params.Version=0.175.1 -X github.com/status-im/status-go/params.GitCommit=cda4aca61 -X github.com/status-im/status-go/params.IpfsGatewayURL=https://ipfs.status.im/ -X github.com/status-im/status-go/vendor/github.com/ethereum/go-ethereum/metrics.EnabledStr=true' \
	-o ./build/bin/statusd ./cmd/statusd
Missing TARGET, assuming default target. See nix/README.md for more details.
Configuring Nix shell for target 'default'...
google.golang.org/protobuf/internal/flags
google.golang.org/protobuf/internal/set
[...]
github.com/status-im/status-go/cmd/statusd
Missing TARGET, assuming default target. See nix/README.md for more details.
Configuring Nix shell for target 'default'...
Compilation done.
Missing TARGET, assuming default target. See nix/README.md for more details.
Configuring Nix shell for target 'default'...
Run "build/bin/statusd -h" to view available commands.
```

* Need to understand why the `Missing TAGET` message appear 3 times.

jakubgs:
@apentori because in a `Makefile` every line in a given target starts a new instance of a shell:
https://github.com/status-im/status-go/blob/cda4aca611ac2c42c64f03fd1c11d04493828bd0/Makefile#L132-L137
In this case there's 3 lines, `go build` and two `echo` lines. Since they are not joined together with a `&& \` or `; \` they are separate commands that spawn separate shells.

This is one of the main reasons why people use the `&& \` hack. For example if they set env variables within a `Makefile` target and want to use them in the next command within the same target.

nix/shells.nix
```diff
@@ -0,0 +1,20 @@
...
+  inherit (pkgs) lib callPackage;
+
+  default = callPackage ./shell.nix { };
+
+  shells = {
+    inherit default;
+
+  # Other shell for makefile or CI
+  };
+  mergeDefaultShell = (_: val: lib.mergeSh default [ val ]);
```
jakubgs: Do we even needs this `mergeSh`? Or the `shells.nix` file at all? We don't have multiple shells, and we probably won't need to merge multiple shells either.

apentori: I was creating a new shell for the test to include in a `shellHook` all the variables defined in the Makefile under `test-unit`.

jakubgs: I don't think we need to do that. Lets keep it simple. One shell for now.

apentori:
- `make statusgo` : OK
- `make statusgo-android` : KO - need to investigate on why
- `make statusgo-library` : OK
- `make statusgo-shared-library`: OK
- `make statusgo-ios` : KO - can we build outside of macos ?
- `make test`: KO - need to confirm if its normal or not

Makefile
```diff
@@ -85,6 +85,47 @@ GO_CMD_PATHS := $(filter-out library, $(wildcard cmd/*))
...
+export NIX_USER_CONF_FILES = $(PWD)/nix/nix.conf
+# Location of symlinks to derivations that should not be garbage collected
+export _NIX_GCROOTS = /nix/var/nix/gcroots/per-user/$(USER)/status-go
+
+
+# MacOS root is read-only, read nix/README.md for details
+UNAME_S := $(shell uname -s)
+ifeq ($(UNAME_S),Darwin)
+export NIX_IGNORE_SYMLINK_STORE=1
+endif
```
jakubgs: I'm not sure this is still necessary. This was a fix added in 2019, probably no longer needed.

* https://github.com/status-im/status-mobile/pull/9218

jakubgs: * https://github.com/status-im/status-mobile/pull/18952

default.nix
```diff
@@ -0,0 +1,10 @@
+# for passing build options, see nix/README.md
+{ config ? { } }:
```
jakubgs: I don't think we need this file at all right now, since we're not building any targets.

Makefile
```diff
@@ -85,6 +85,47 @@ GO_CMD_PATHS := $(filter-out library, $(wildcard cmd/*))
...
+
+nix-repl: SHELL := /bin/sh
+nix-repl: ##@nix Start an interactive Nix REPL
+	nix repl default.nix
+
+nix-gc-protected: SHELL := /bin/sh
+nix-gc-protected:
+	@echo -e "$(YELLOW)The following paths are protected:$(RESET)" && \
+	ls -1 $(_NIX_GCROOTS) | sed 's/^/ - /'
```
jakubgs, line 122: I don't see a `nix-gc`, `nix-purge`, or `nix-upgrade` targets.

nix/overlay.nix
```diff
@@ -0,0 +1,36 @@
+# Override some packages and utilities in 'pkgs'
+# and make them available globally via callPackage.
+#
+# For more details see:
+# - https://nixos.wiki/wiki/Overlays
+# - https://nixos.org/nixos/nix-pills/callpackage-design-pattern.html
+prev: final:
+let 
+  inherit (prev);
+in
```
jakubgs: What is this for?

jakubgs (changes requested):

Makefile
```diff
@@ -85,6 +85,47 @@ GO_CMD_PATHS := $(filter-out library, $(wildcard cmd/*))
...
+endif
+#----------------
+# Nix targets
+#----------------
+
+
+SHELL := ./nix/scripts/shell.sh
+
+shell: export TARGET ?= default
+shell: ##@prepare Enter into a pre-configured shell
```
jakubgs, line 108: This file doesn't have the script that provides help message from `make` so those are just comments.

nix/overlay.nix
```diff
@@ -0,0 +1,36 @@
+# Override some packages and utilities in 'pkgs'
+# and make them available globally via callPackage.
+#
+# For more details see:
+# - https://nixos.wiki/wiki/Overlays
+# - https://nixos.org/nixos/nix-pills/callpackage-design-pattern.html
+prev: final:
```
jakubgs: This is wrong. See doc at https://nixos.wiki/wiki/Overlays :
```suggestion
final: prev:
```
That's the source of your infinite recursion error.

nix/overlay.nix
```diff
@@ -0,0 +1,35 @@
...
+    platformVersions = [ "31" ];
+    cmakeVersions = [ "3.18.1" ];
+    ndkVersion = "22.1.7171670";
+    includeNDK = true;
+    includeExtras = [
+      "extras;android;m2repository"
+      "extras;google;m2repository"
+    ];
+  };
+  # FIXME infinite recursion encountered
```
jakubgs: This comment should explain why this override is necessary.

jakubgs:
Also, with this PR working we can remove all of those:
https://github.com/status-im/status-go/blob/bcb1fa1063009b194b7667be58a21973eb59f5f7/Makefile#L273-L316
Except for one that's missing from `nixpkgs`, which is `modvendor` package.
But it should be fairly easy to add it to the `overlay.nix`.

jakubgs:
Also, your `nix/scripts/shell.nix` appears to be missing parts that prevent nesting of shells.

apentori:
> But it should be fairly easy to add it to the `overlay.nix`.

I don't see how to do it :/

Shoud it be something like this ?
```
goModVendor = buildGoModule rec {
  pname = "modvendor";
  version = "0.5.0";

  src = fetchFromGitHub {
    owner = "goware";
    repo = "nodvendor";
    rev = "v${version}";
    hash = "sha256-somesha";
  };
  }
```

jakubgs:
>Shoud it be something like this ?

Yes, essentially yes. You can see examples of this in `status-mobile`:
```nix
  # Custom packages
  aapt2 = callPackage ./pkgs/aapt2 { };
  patchMavenSources = callPackage ./pkgs/patch-maven-srcs { };
  goMavenResolver = callPackage ./pkgs/go-maven-resolver { };
  xcbeautify = callPackage ./pkgs/xcbeautify { };
  idb-companion = callPackage ./pkgs/idb-companion { };
```
https://github.com/status-im/status-mobile/blob/05a810b051605bd15cc92c502e955a896f98d26f/nix/overlay.nix#L83-L88

nix/shell.nix
```diff
@@ -0,0 +1,38 @@
...
+in pkgs.mkShell {
+  name = "status-go-shell";
+
+  buildInputs = with pkgs; [
+    git jq which
+    go_1_19 golangci-lint go-junit-report gopls go-bindata gomobileMod
+    mockgen protobuf3_20 protoc-gen-go gotestsum
+   ] ++ lib.optionals (stdenv.isDarwin) [ xcodeWrapper ];
+
+>>>>>>> 3d3e0f1f6 (nix: fix androidPackages issue)
```
jakubgs: Merge artifact.

apentori: Fixed

nix/overlay.nix
```diff
@@ -0,0 +1,35 @@
+# Override some packages and utilities in 'pkgs'
+# and make them available globally via callPackage.
+#
+# For more details see:
+# - https://nixos.wiki/wiki/Overlays
+# - https://nixos.org/nixos/nix-pills/callpackage-design-pattern.html
+final: prev:
+let
+  inherit (prev) callPackage;
+in {
```
jakubgs: Not sure if this inherit makes sense considering we use it just once.

nix/scripts/shell.sh
```diff
@@ -0,0 +1,62 @@
...
+    nixArgs+=("--keep ${_NIX_KEEP//,/ --keep }")
+fi
+
+# Not all builds are ready to be run in a pure environment
+if [[ -n "${_NIX_PURE}" ]]; then
+    nixArgs+=("--pure")
+    pureDesc='pure '
+fi
+
+echo -e "${GRN}Configuring ${pureDesc}Nix shell for target ...${RST}" 1>&2
```
jakubgs: Where's no reference to `${TARGET}` here, why?

Makefile
```diff
@@ -85,6 +85,55 @@ GO_CMD_PATHS := $(filter-out library, $(wildcard cmd/*))
...
+
+
+#----------------
+# Nix targets
+#----------------
+
+
+SHELL := ./nix/scripts/shell.sh
+
+shell: ##@prepare Enter into a pre-configured shell
```
jakubgs, line 108: We're missing a default target. And you've made changes to `nix/scripts/shell.sh` that I don't understand.
```suggestion
shell: export TARGET ?= default
shell: ##@prepare Enter into a pre-configured shell
```

shell.nix
```diff
@@ -1,77 +1,9 @@
...
-            sha256 = "sha256-s4XVjACmpd10C5k+P3vtcS/aWxI6UkSUPyxzLhD2vRI=";
-          };
-        });
-      })
-    ];
-  }
-}:
+# for passing build optionsm see nix/README
+# TODO complet nix/README
+{ config ? { } }:
```
jakubgs, line 3: Do we even need `config` for anything? Maybe we can get rid of it entirely.

This whole file could just be:
```nix
(import ./nix).shells.default
```
And `default.nix` could just be:
```nix
import ./nix
```

jakubgs (changes requested):
Two main questions:

- Do you think we'll need `config` in status-go?
   - I'm leaning towards no, but maybe you have some ideas. In `status-mobile` we had it because we needed a way to provide things like source overrides for `status-go` or android builds. Not sure if useful here.
- Do you think we'll need different `TARGET`s?
  - I think it might be if in the future we want to provide alternative smaller shells to some targets. And this way we stray less from `status-mobile` setup, keeping contents of `nix/scripts` essentially the same.

Also, everything starting from `setup` target to `install-xtools` can be removed since we provide everything from Nix shell.

jakubgs:
Also, you're missing `gcroots` changes here I applied in:

* https://github.com/status-im/status-mobile/pull/18742

I think we should align `nix/scripts` to be exactly the same as `status-mobile` to avoid confusion.

igor-sirotin (commented):
I know nothing about nix so I'm not sure if my review is relevant at all.

I guess best I can do is test the solution.

```bash
➜  status-go git:(using-nix-shell-in-makefile) make shell
Configuring Nix shell for target ...
error: creating symlink from '/nix/var/nix/gcroots/per-user/igorsirotin/status-mobile/shell.tmp-54234-735157167' to '/nix/store/0ayykp21sa8ib79fv0k3zx064sca23c4-status-go-shell.drv': Permission denied
Configuring Nix shell for target ...
error: creating directory '/nix/var/nix/gcroots/per-user/igorsirotin/status-go': Permission denied
Configuring Nix shell for target ...
error: creating directory '/nix/var/nix/gcroots/per-user/igorsirotin/status-go': Permission denied
make: *** [shell] Error 1
➜
```

😢 

-----

UPD: it got better after running `nix-shell` after checking out the branch.

```bash
[nix-shell:~/Repositories/Status/status-go]$ make shell
Configuring Nix shell for target ...
error: creating directory '/nix/var/nix/gcroots/per-user/igorsirotin/status-go': Permission denied
Configuring Nix shell for target ...
error: creating directory '/nix/var/nix/gcroots/per-user/igorsirotin/status-go': Permission denied
make: *** [Makefile:107: shell] Error 1
```

Should I run this with sudo? 🤔

jakubgs:
@igor-sirotin Yes, this is the issue I mentioned in comment above about new way to handle `gcroots`.

Makefile
```diff
@@ -185,39 +235,44 @@ endif
...
 docker-image: BUILD_TARGET ?= statusd
 docker-image: ##@docker Build docker image (use DOCKER_IMAGE_NAME to set the image name)
 	@echo "Building docker image..."
 	docker build --file _assets/build/Dockerfile . \
 		--build-arg "build_tags=$(BUILD_TAGS)" \
-		--build-arg "build_flags=$(BUILD_FLAGS)" \
+		--build-arg 'build_flags=$(BUILD_FLAGS)' \
 		--build-arg "build_target=$(BUILD_TARGET)" \
 		--label "commit=$(GIT_COMMIT)" \
 		--label "author=$(GIT_AUTHOR)" \
```
jakubgs: Just use single quotes for all flags, not just one.

nix/scripts/version.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+# WARNING: Versions 2.14, 2.15, 2.16 have an issue creating gcroots:
+# https://github.com/NixOS/nix/issues/8564
```
jakubgs: We don't need this comment anymore.

_assets/scripts/install_deps.sh
```diff
@@ -7,7 +7,7 @@ elif [[ -x $(command -v pacman) ]]; then
 elif [[ -x $(command -v brew) ]]; then
   brew install protobuf jq
 elif [[ -x $(command -v nix-env) ]]; then
-  nix-env -iA nixos.protobuf3_17
+  nix-env -iA nixos.protobuf3_20
```
jakubgs: Completely unnecessary, this script and its references can be removed. Everything is provided by Nix.

Makefile
```diff
@@ -233,49 +288,11 @@ endif
...
-install-lint: ##@install Install Linting Tools
-	GO111MODULE=on go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.52.2
-
-install-junit-report: ##@install Install Junit Report Tool for Jenkins integration
-	GO111MODULE=on go install github.com/jstemmer/go-junit-report/v2@latest
-
-install-mock: ##@install Install Module Mocking Tools
-	GO111MODULE=on go install github.com/golang/mock/mockgen@v1.4.4
-
-install-modvendor: ##@install Install Module Vendoring Tool
```
jakubgs, line 267: We need to remove reference to this in `_assets/ci/Jenkinsfile.tests`.

jakubgs (approved):

jakubgs (approved):
We probably don't need to use `nix.shell` in Jenkinsfiles sine `make` invokes it itself.

igor-sirotin (approved):
Thank you, @apentori!

This now works for me. Below is a few minor points. We discussed them in DM, but leaving it here for history.

1. Some docs wrong links.
```bash
> make statusgo
Missing TARGET, assuming default target. See nix/README.md for more details.
```

There's nothing about `TARGET` in `nix/README.md` 😦
Also, readme refers to `nix/DETAILS.md` and `KNOWN_ISSUES.md` don't exist. 

2. When running make from JetBrains GoLand IDE, I get go version conflicts.
```bash
compile: version "go1.20.13" does not match go tool version "go1.19.9"
```
This seem to happen because of `GOROOT` env variable defined by GoLand.
Simple `GOROOT :=` in makefile fixes the issue.

3. `detected_OS` ended up empty for me>.
This fixes the issue:
```diff
diff --git a/Makefile b/Makefile
index c78e401b5..a7bb55d77 100644
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ IPFS_GATEWAY_URL ?= https://ipfs.status.im/
 ifeq ($(OS),Windows_NT)     # is Windows_NT on XP, 2000, 7, Vista, 10...
  detected_OS := Windows
 else
- detected_OS := $(strip $(SHELL=/bin/sh shell uname))
+ detected_OS := $(shell uname)
 endif
 
 ifeq ($(detected_OS),Darwin)
```

4. There're lots of `Configuring Nix shell for target 'default'...` in the build log.
As you said, this should be fixed by running all commands in a target with a single nix-shell execution. That would be to separate them with `:`. We can do this later, it's not urgent.

5. Do we actually need this message?
```bash
Missing TARGET, assuming default target. See nix/README.md for more details.
```
I suppose most people will use `default` target, maybe we shouldn't consider this as warning?

-----

Actually, while writing this, I thought of one thing. Should the makefile actually be tight that much to nix? Maybe the tools specifying should rely on users? We could provide `make shell` command, but not force all targets to be tun with nix.

jakubgs:
No, we should force devs and users to use Nix. That way we have full control over build tools and dependencies.

As for the warning, i agree, it's not necessary here as we only have the default shell.

```
compile: version "go1.20.13" does not match go tool version "go1.19.9"
```
Where does that come from? We do have bad version references in `go.mod` for sure:
https://github.com/status-im/status-go/blob/0aed93ff04b7d8194c650921ceb659441dc3d602/go.mod#L3

jakubgs:
Using `$(shell uname)` will invoke Nix shell to run `uname`, it shouldn't be necessary and will make things slower.

igor-sirotin:
> No, we should force devs and users to use Nix. That way we have full control over build tools and dependencies.

I personally disagree with this, but won't complain as I am going to use Nix 😄 
But I can simply imagine someone who don't want to use Nix, but he will be forced to use it or rewrite the Makefile. Or if we decide to switch the package manager, it shouldn't affect the Makefile, the build instructions should remain the same in theory.

As fro the "full control over build tools and dependencies": 
If we remove generated files from the repo, then it shouldn't be such a problem.

Again, I'm just posting my thoughts here. It's not a blocker or requirement from me.

igor-sirotin:
> ```
> compile: version "go1.20.13" does not match go tool version "go1.19.9"
> ```
> Where does that come from? We do have bad version references in `go.mod` for sure:

As I mentioned:

> This seem to happen because of `GOROOT` env variable defined by GoLand.
> Simple `GOROOT :=` in makefile fixes the issue.

jakubgs:
Why do you use a custom `GOROOT`? Is it common among devs?

igor-sirotin:
> Why do you use a custom `GOROOT`? Is it common among devs?

I don't do this explicitly myself. This is done my IDE (JetBrains GoLand).
And it could easily be that other devs have this env variable defined as well.

Settings `GOROOT` was previously required by Go. Not it's not. But I can't modify my IDE behaviour. And I would like not to switch from it 😄

jakubgs:
That's fine, just trying to understand the behavior.

osmaczko (approved):
Thanks for this :pray:

_assets/ci/Jenkinsfile.android
```diff
@@ -31,13 +31,13 @@ pipeline {
...
 
   environment {
-    TARGET   = 'android'
-    TMPDIR   = "${WORKSPACE_TMP}"
-    GOPATH   = "${WORKSPACE_TMP}/go"
-    GOCACHE  = "${WORKSPACE_TMP}/gocache"
-    PATH     = "${PATH}:${GOPATH}/bin"
-    REPO_SRC = "${GOPATH}/src/github.com/status-im/status-go"
-    ARTIFACT = utils.pkgFilename(name: "status-go", type: "android", ext: "aar", version: null)
+    PLATEFORM = 'android'
```
yakimant:
```suggestion
    PLATFORM = 'android'
```

nix/pkgs.nix
```diff
@@ -0,0 +1,28 @@
...
+# as well as which versions of package we use, including their overrides.
+{ config ? { } }:
+
+let 
+  # For testing local version of nixpkgs
+  #nixpkgsSrc = (import <nixpkgs> { }).lib.cleanSource "/home/jakubgs/work/nixpkgs";
+
+  # We follow the master branch of official nixpkgs.
+  nixpkgsSrc = builtins.fetchTarball {
+    url = "https://github.com/NixOS/nixpkgs/archive/e7603eba51f2c7820c0a182c6bbb351181caa8e7.tar.gz";
```
yakimant: Different form status-mobile one?

yakimant (approved):
Have few comments and there are couple unresolved from Jakub.
Great job!

nix/pkgs.nix
jakubgs: Good point, we might want to sync together with Sid's upgrade:

* https://github.com/status-im/status-mobile/pull/19369

siddarthkay:
Hmm `make statusgo-android` is failing on this branch with : 

```
unable to import bind/java: [-: # golang.org/x/mobile/bind/java
In file included from context_android.c:6:
../../../golang.org/x/mobile/bind/java/seq_android.h:9:10: fatal error: 'android/log.h' file not found]
/nix/store/0gm0yz51dh0j4afz3q2pmzkazd2n7y1l-gomobile-unstable-2022-05-18/bin/gomobile:
 /nix/store/0gm0yz51dh0j4afz3q2pmzkazd2n7y1l-gomobile-unstable-2022-05-18/bin/gobind 
 -lang=go,java -outdir=/tmp/gomobile-work -tags=gowaku_no_rln 
 github.com/status-im/status-go/mobile failed: exit status 1

make: *** [statusgo-android] Error 1
```
perhaps with newer `nixpkgs` we also will have to upgrade `gomobile` overlay?

I guess for this PR we could stick with older `nixpkgs` if that solves the issue and we can upgrade `nixpkgs` for `status-go` separately to address these issues.
cc @jakubgs

jakubgs:
I think we should use the same commit as `status-mobile`.

siddarthkay (approved):
Looks good to me and works well locally as well.
@apentori : when you get a chance do push the modified `go.mod`

Once this is merged It will make my life easier for the upcoming go upgrade.
Thank you!

jakubgs:
For future reference we had to apply some extra fixes for this:

- [`status-go#ba042379`](https://github.com/status-im/status-go/commit/ba042379) - fix(nix)_: avoid warning about default shell fallback
- [`status-go#c343d529`](https://github.com/status-im/status-go/commit/c343d529) - fix(nix)_: add USER var when sourcing profile
- [`status-go#9509fddd`](https://github.com/status-im/status-go/commit/9509fddd) - nix: fix detecting install type in Docker
- [`status-go#c57c8aab`](https://github.com/status-im/status-go/commit/c57c8aab) - ci: fix build

PR #4389, opened 2023-11-27
feat: refresh balances when Eth or ERC20 transfer is detected

### LINKED ISSUE #12138: When receiving tx, refresh wallet balance
When sending/receiving an erc721 tx, refresh the collectible
When sending/receiving an erc20, refresh the asset balance
When sending/receiving native token, refresh the native balance

dlipicar:
Fixes https://github.com/status-im/status-desktop/issues/12138

Whenever an ETH or ERC20 transfer is detected, we trigger a waller refresh (re-fetch balances).

We trigger them if the Tx happened later than 30 seconds before the last update (account for delays in the balances calculation).

We delay the actual refresh for about 30 seconds, to prevent triggering multiple refreshes if a lot of Txs are detected at the same time.

We don't have the granularity needed in our API to update only balances for the token involved in the transfer. This can be improved in the future.

### DIFF
--- a/services/wallet/reader.go
+++ b/services/wallet/reader.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"math"
 	"math/big"
+	"sync/atomic"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
@@ -15,6 +16,7 @@ import (
 	"github.com/status-im/status-go/services/wallet/async"
 	"github.com/status-im/status-go/services/wallet/market"
 	"github.com/status-im/status-go/services/wallet/thirdparty"
+	"github.com/status-im/status-go/services/wallet/transfer"
 
 	"github.com/status-im/status-go/services/wallet/token"
 	"github.com/status-im/status-go/services/wallet/walletevent"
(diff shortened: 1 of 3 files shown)


services/wallet/reader.go
```diff
@@ -159,6 +171,64 @@ func (r *Reader) Stop() {
...
+	}
+
+	// Respond to ETH/Token transfers
+	walletEventCb := func(event walletevent.Event) {
+		if event.Type != transfer.EventInternalETHTransferDetected &&
+			event.Type != transfer.EventInternalERC20TransferDetected {
+			return
+		}
+
+		timecheck := r.lastWalletTokenUpdateTimestamp.Load() - activityReloadMarginSeconds
```
Cuteivist, line 216: If I understand that correctly, does it handle case when
1. I get balances
2. 20 seconds later I get 1 tx <-- trigger is not run because it's less than 30 sec 

Next refresh will happen after 10 minutes.

dlipicar: I've just woken up so I might be screwing up, but it's supposed to be the other way around.

I fetch balances
- lastWalletTokenUpdateTimestamp = 1000

20 seconds later I get 1 tx
- Tx Timestamp = 1020 
- TimeCheck = 1000 - 30 = 970

- Tx Timestamp > TimeCheck -> Trigger refresh

I'm assuming the data we get whenever we fetch balances might have some slight delay, so we refetch balances *even when* the detected transfer is a bit older than the last fetch.

stefandunca (approved):
Great optimization

Cuteivist (approved):

PR #4391, opened 2023-11-28
feat: token order API

Cuteivist:
Task https://github.com/status-im/status-desktop/issues/12725

corresponding status-desktop change: https://github.com/status-im/status-desktop/pull/12929

* Added API for re-ordering.
* Added settings sync

Current plan is to always update full list. The reason for it is fact that (unlike wallet accounts) positions can be changed multiple times and the user is asked to save whole list.
In future if single row update will be viable it will be added.


multiaccounts/settings_wallet/database.go
```diff
@@ -0,0 +1,153 @@
...
+
+	defer func() {
+		if mainError == nil {
+			err = tx.Commit()
+			return
+		}
+		_ = tx.Rollback()
+	}()
+
+	_, err = tx.Exec("DELETE FROM token_preferences WHERE testnet = ?", testNetworksEnabled)
```
stefandunca: Why not use the `mainError` directly? There is no overwrite here. Same question also for other places.

multiaccounts/settings_wallet/database_test.go
```diff
@@ -0,0 +1,286 @@
...
+)
+
+func setupTestDB(t *testing.T) (*WalletSettings, func()) {
+	db, stop, err := helpers.SetupTestSQLDB(appdatabase.DbInitializer{}, "wallet-settings-tests-")
+	require.NoError(t, err)
+	d := NewWalletSettings(db)
+	return d, func() { require.NoError(t, stop()) }
+}
+
+func TestTokenPrefrences(t *testing.T) {
```
stefandunca: Splitting this into multiple tests should help with debugging in case something fails. What do you think?

stefandunca (approved):
Great job. Please see my questions inline.

multiaccounts/settings_wallet/database.go
```diff
@@ -0,0 +1,153 @@
...
+func NewWalletSettings(db *sql.DB) *WalletSettings {
+	return &WalletSettings{
+		db: db,
+	}
+}
+
+// This function should not be used directly, it is called from the functions which update token preferences.
+func (db *WalletSettings) setClockOfLastTokenPreferencesChange(tx *sql.Tx, clock uint64) error {
+	if tx == nil {
+		return nil
```
saledjenic: @Cuteivist why do we return nil here? Don't we need to return something like `return errors.New("database transaction is nil")`?

multiaccounts/settings_wallet/database.go
```diff
@@ -0,0 +1,153 @@
...
+	query := "SELECT wallet_token_preferences_change_clock FROM settings WHERE synthetic_id = 'id'"
+	err = db.db.QueryRow(query).Scan(&result)
+	if err != nil {
+		return 0, err
+	}
+	return result, err
+}
+
+func (db *WalletSettings) setTokenGroupByCommunity(tx *sql.Tx, value bool) error {
+	if tx == nil {
```
saledjenic: @Cuteivist the same here

multiaccounts/settings_wallet/database.go
```diff
@@ -0,0 +1,153 @@
...
+	for rows.Next() {
+		token := TokenPreferences{}
+		err := rows.Scan(&token.Key, &token.Position, &token.GroupPosition, &token.Visible, &token.CommunityID)
+		if err != nil {
+			return nil, err
+		}
+
+		result = append(result, token)
+	}
+
```
saledjenic, line 120: @Cuteivist please add this check:
```suggestion
	if err := rows.Err(); err != nil {
		return nil, err
	}
```

protocol/messenger_handlers.go
```diff
@@ -1655,3 +1658,26 @@ func (m *Messenger) handleSyncActivityCenterCommunityRequestDecisionProtobuf(mes
...
 
 
+func (m *Messenger) handleSyncTokenPreferencesProtobuf(messageState *ReceivedMessageState, protoBytes []byte, msg *v1protocol.StatusMessage, filter transport.Filter) error {
+	m.logger.Info("handling SyncTokenPreferences")
+	
+	if !common.IsPubKeyEqual(messageState.CurrentMessageState.PublicKey, &m.identity.PublicKey) {
+		m.logger.Warn("not coming from us, ignoring")
+		return nil
+	}
+
```
saledjenic, line 1668: Unnecessary new lines here.

protocol/messenger_handlers.go
```diff
@@ -1655,3 +1658,26 @@ func (m *Messenger) handleSyncActivityCenterCommunityRequestDecisionProtobuf(mes
...
+	p := &protobuf.SyncTokenPreferences{}
+	err := proto.Unmarshal(protoBytes, p)
+	if err != nil {
+		return err
+	}
+
+	m.outputToCSV(msg.TransportLayer.Message.Timestamp, msg.ApplicationLayer.ID, messageState.CurrentMessageState.Contact.ID, filter.ContentTopic, filter.ChatID, msg.ApplicationLayer.Type, p)
+
+	return m.HandleSyncTokenPreferences(messageState, p, msg)
+
```
saledjenic, line 1680: Here as well.

protocol/messenger_response.go
```diff
@@ -49,6 +50,7 @@ type MessengerResponse struct {
 	WatchOnlyAccounts             []*accounts.Account
 	Keypairs                      []*accounts.Keypair
 	AccountsPositions             []*accounts.Account
+	TokenPreferences              []walletsettings.TokenPreferences
```
saledjenic, line 53: Any reason why didn't you include `TokenPreferences` in isEmpty, Mrshal and Merge functions?

saledjenic (commented):
@Cuteivist please don't forget to update the version file.

Just interested interested do see this in play, do we already have a desktop app PR for this?

multiaccounts/settings_wallet/database.go
Cuteivist: Fixed, thanks!

multiaccounts/settings_wallet/database.go
Cuteivist: Fixed, thanks!

multiaccounts/settings_wallet/database.go
Cuteivist: Thanks! Fixed.

protocol/messenger_handlers.go
Cuteivist: It's auto generated file.

protocol/messenger_handlers.go
Cuteivist: It's auto generated file.

protocol/messenger_response.go
Cuteivist: Simply I forgotten to update them. Thanks for pointing that out.

During that process I found out the `AccountPositions` were incorrectly merged

`	r.WatchOnlyAccounts = append(r.AccountsPositions, response.AccountsPositions...)`

I fixed that line too.

multiaccounts/settings_wallet/database_test.go
Cuteivist: Done.

saledjenic (approved):

PR #4393, opened 2023-11-28
Integrating keycard signing in case of buying stickers

saledjenic:
We should let the client decide where to sign a transaction and once the tx is signed let him submit the tx providing the signature. This PR allows exactly that.


flexsurfer:
hey @saledjenic thanks for the PR, could you please give an example of steps (methods calls) how it worked before and how it works in this PR ? thanks

saledjenic:
> hey @saledjenic thanks for the PR, could you please give an example of steps (methods calls) how it worked before and how it works in this PR ? thanks

@flexsurfer yes, ofc. 

Regarding the mobile app there are no changes right now. Once the mobile app implements a new wallet API we can remove related endpoints.

These are the steps done by the desktop app:
1. `stickers_buyPrepareTx`, providing chain id, pack id...
2. the response from step 1. is updated with necessary values (gas, gasPrice, maxPriorityFeePerGas, maxFeePerGas) and sent to `wallet_buildTransaction`
3. Response from `buildTransaction` contains all necessary data (tx, tx hash, signing details) informing the client whether to:
3.1. sign the tx on the keycard - from this action signature gets created (r,s,v params)
3.2. sign the tx on the statusgo side (using private key stored in local keystore file) calling `wallet_signMessage` -- from this action signature gets created
4. Now the tx + chainId + signature is passed to `wallet_sendTransactionWithSignature` which sends the tx and `TrackPendingTransaction`
5. And as the last step we call `stickers_addPending`

services/wallet/api.go
```diff
@@ -524,6 +525,37 @@ func (api *API) GetAddressDetails(ctx context.Context, chainID uint64, address s
 	return result, nil
 }
 
+func (api *API) SignMessage(ctx context.Context, message types.HexBytes, address common.Address, password string) (string, error) {
+	log.Debug("[WalletAPI::SignMessage] message: ", message, " address: ", address, " password: ", password)
+	return api.s.transactionManager.SignMessage(message, address, password)
+}
+
+func (api *API) BuildTransaction(ctx context.Context, chainID uint64, sendTxArgsJSON string) (response *transfer.TxResponse, err error) {
+	log.Debug("[WalletAPI::BuildTransaction] chainID: ", chainID, " sendTxArgsJSON: ", sendTxArgsJSON)
```
stefandunca: Log API requires the use of a key-value approach to parameters. Otherwise, the entire log line will be an error. Same in the other log calls.
```suggestion
	log.Debug("[WalletAPI::BuildTransaction]", "chainID: ", chainID, " sendTxArgsJSON: ", sendTxArgsJSON)
```

services/wallet/api.go
```diff
@@ -524,6 +525,37 @@ func (api *API) GetAddressDetails(ctx context.Context, chainID uint64, address s
 	return result, nil
 }
 
+func (api *API) SignMessage(ctx context.Context, message types.HexBytes, address common.Address, password string) (string, error) {
+	log.Debug("[WalletAPI::SignMessage] message: ", message, " address: ", address, " password: ", password)
```
qfrank: log `password` is not a good idea 🙂 
we don't need such prefix `[WalletAPI::SignMessage]` anymore, because we have [origin information](https://github.com/status-im/status-go/pull/4277) in log now

stefandunca (approved):
Please see my comment inline.

services/wallet/api.go
saledjenic: > log password is not a good idea

@qfrank it's ok for the debug, isn't it?

> we don't need such prefix [WalletAPI::SignMessage] anymore, because we have https://github.com/status-im/status-go/pull/4277 in log now

Just followed the approach used in the rest of the file.

qfrank: not good even for debug level IMO

qfrank: except we disable log level setting for `debug` for production build

saledjenic: @qfrank I removed the pass from the log. Now I have to add it manually when I'm debugging. ;)

services/wallet/api.go
saledjenic: Updated, thanks.

qfrank (approved):

PR #4396, opened 2023-11-29
feat(wallet): add Wallet Connect state change API

stefandunca:

### DIFF
--- a/services/wallet/walletconnect/database.go
+++ b/services/wallet/walletconnect/database.go
@@ -2,6 +2,7 @@ package walletconnect
 
 import (
 	"database/sql"
+	"errors"
 )
 
 type Pairing struct {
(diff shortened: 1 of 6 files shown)


services/wallet/walletconnect/database.go
```diff
@@ -22,6 +23,25 @@ func InsertPairing(db *sql.DB, pairing Pairing) error {
...
+	if err != nil {
+		return err
+	}
+
+	res, err := stmt.Exec(active, topic)
+	if err != nil {
+		return err
+	}
+
+	rowsAffected, err := res.RowsAffected()
```
saledjenic, line 37:
```suggestion
	rowsAffected, err := res.RowsAffected()
	if err != nil {
		return err
	}
```

saledjenic (approved):

PR #4397, opened 2023-11-29
preparing release transaction data

saledjenic:
The same approach applied as we use for `ens_registerPrepareTx` and `ens_setPubKeyPrepareTx`.


flexsurfer (approved):

stefandunca (approved):
👍🏼

PR #4406, opened 2023-11-30
refactor(communities)_: delegate Community creation in Persistence

osmaczko:
In persistence.go, the lack of sufficient knowledge for constructing fully initialized Community objects required clients to manually call `initializeCommunity`. This commit addresses the issue by delegating Community creation to Manager. It also removes queries and logic duplication.

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -256,11 +256,11 @@ func NewManager(identity *ecdsa.PrivateKey, installationID string, db *sql.DB, e
 		torrentConfig:               torrentConfig,
 		torrentTasks:                make(map[string]metainfo.Hash),
 		historyArchiveDownloadTasks: make(map[string]*HistoryArchiveDownloadTask),
-		persistence: &Persistence{
-			logger:     logger,
-			db:         db,
-			timesource: timesource,
-		},
+	}
+
+	manager.persistence = &Persistence{
+		db:                      db,
+		recordBundleToCommunity: manager.dbRecordBundleToCommunity,
 	}
 
 	if managerConfig.accountsManager != nil {
@@ -622,19 +622,7 @@ func (m *Manager) publish(subscription *Subscription) {
 }
 
 func (m *Manager) All() ([]*Community, error) {
-	communities, err := m.persistence.AllCommunities(&m.identity.PublicKey, m.installationID)
-	if err != nil {
-		return nil, err
-	}
-
-	for _, c := range communities {
-		err = m.initializeCommunity(c)
-		if err != nil {
-			return nil, err
-		}
-	}
-
-	return communities, nil
+	return m.persistence.AllCommunities(&m.identity.PublicKey)
 }
 
 type CommunityShard struct {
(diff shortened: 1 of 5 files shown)


protocol/communities/persistence_mapping.go
```diff
@@ -0,0 +1,159 @@
...
+		spectated:    community.config.Spectated,
+		muted:        community.config.Muted,
+		mutedTill:    community.config.MuteTill,
+		shardCluster: shardCluster,
+		shardIndex:   shardIndex,
+	}, nil
+}
+
+func communityToEventsRecord(community *Community) (*EventsRecord, error) {
+	if community.config.EventsData == nil {
```
MishkaRogachev, line 43: suggestion: `|| len(community.config.EventsData.Events) < 1`

protocol/communities/manager.go
```diff
@@ -3220,43 +3156,31 @@ func (m *Manager) BanUserFromCommunity(request *requests.BanUserFromCommunity) (
...
-	// Workaround for https://github.com/status-im/status-desktop/issues/12188
-	HydrateChannelsMembers(community.IDString(), community.config.CommunityDescription)
+		if m.transport != nil && m.transport.WakuVersion() == 2 {
+			topic := community.PubsubTopic()
+			privKey, err := m.transport.RetrievePubsubTopicKey(topic)
+			if err != nil {
+				return err
+			}
+			community.config.PubsubTopicPrivateKey = privKey
+		}
```
MishkaRogachev, line 3173: maybe return an error if waku version is different?

protocol/communities/persistence.go
```diff
@@ -12,79 +12,201 @@ import (
...
 	LEFT JOIN communities_requests_to_join r ON c.id = r.community_id AND r.public_key = ?
 	LEFT JOIN communities_events ae ON c.id = ae.id
 	LEFT JOIN communities_control_node ccn ON c.id = ccn.community_id`
 
-func (p *Persistence) SaveCommunity(community *Community) error {
-	id := community.ID()
-	privateKey := community.PrivateKey()
-	controlNode := community.ControlNode()
-	wrappedCommunity, err := community.ToProtocolMessageBytes()
+func scanCommunity(scanner func(dest ...any) error) (*CommunityRecordBundle, error) {
```
MishkaRogachev, line 82: Cool! We need to do this for all other repeated queries

protocol/communities/persistence_mapping.go
```diff
@@ -0,0 +1,159 @@
...
+
+	return &EventsRecord{
+		id:             community.ID(),
+		rawEvents:      rawEvents,
+		rawDescription: community.config.EventsData.EventsBaseCommunityDescription,
+	}, nil
+}
+
+func recordToRequestToJoin(r *RequestToJoinRecord) *RequestToJoin {
+	// FIXME: fill revealed addresses
```
MishkaRogachev, line 60: Is it a real fixme or we add addresses on the manager/messenger level?

MishkaRogachev (approved):
Great job! Just curious that not many community tests were affected

protocol/communities/manager.go
osmaczko: Wanted to preserve the same behavior as was before. Once we get rid off wakuv1 we should make it an error I believe.

protocol/communities/persistence_mapping.go
osmaczko: This is a real fixme, although I am unsure about the impact of its absence, since the behavior hasn't changed with my refactor.

protocol/communities/manager.go
```diff
@@ -256,11 +256,11 @@ func NewManager(identity *ecdsa.PrivateKey, installationID string, db *sql.DB, e
...
-		persistence: &Persistence{
-			logger:     logger,
-			db:         db,
-			timesource: timesource,
-		},
+	}
+
+	manager.persistence = &Persistence{
+		db:                      db,
+		recordBundleToCommunity: manager.dbRecordBundleToCommunity,
```
mprakhov, line 263: it's too late, but this recordBundleToCommunity callback is not a good idea. What gonna happen if someone decides to override this callback? It can lead to unexpected behavior as it will be used on all other queries which use this callback

protocol/communities/persistence.go
```diff
@@ -146,59 +268,38 @@ func (p *Persistence) queryCommunities(memberIdentity *ecdsa.PublicKey, installa
...
-		var publicKeyBytes, privateKeyBytes, controlNodeBytes, descriptionBytes []byte
-		var joined, spectated, verified, muted bool
-		var muteTill sql.NullTime
-		var cluster, index, requestedToJoinAt sql.NullInt64
-		var installationIDStr sql.NullString
-
-		// Community events specific fields
-		var eventsBytes, eventsDescriptionBytes []byte
-		err := rows.Scan(&publicKeyBytes, &privateKeyBytes, &controlNodeBytes, &descriptionBytes, &joined, &spectated, &verified, &muted, &muteTill, &requestedToJoinAt, &eventsBytes, &eventsDescriptionBytes, &cluster, &index, &installationIDStr)
+		r, err := scanCommunity(rows.Scan)
```
mprakhov, line 271: this fn contains duplicate logic from the `rowsToCommunities`, can be easily replaced

protocol/communities/manager.go
osmaczko: Not sure I am following. It is encapsulated into the structure like everything else. The same question I can ask about the `db`, what gonna happen if someone decides to override `db` field with something else?

Can you please elaborate more of what you meant, what would be the motivation to override this callback?

PR #4409, opened 2023-12-01
chore: parse all non confirmed requests to join of the community during getting all communities

mprakhov:
- during requesting all communities, add all non-approved requests to join the Community
- send MessengerSignal after promoting self as a control node

Closes #https://github.com/status-im/status-desktop/issues/12577

status-desktop PR: https://github.com/status-im/status-desktop/pull/12928

### DIFF
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -265,6 +265,7 @@ func (o *Community) MarshalJSON() ([]byte, error) {
 		PubsubTopic                 string                               `json:"pubsubTopic"`
 		PubsubTopicKey              string                               `json:"pubsubTopicKey"`
 		Shard                       *shard.Shard                         `json:"shard"`
+		RequestsToJoinCommunity     []*RequestToJoin                     `json:"requestsToJoinCommunity"`
 	}{
 		ID:                          o.ID(),
 		MemberRole:                  o.MemberRole(o.MemberIdentity()),
@@ -287,6 +288,7 @@ func (o *Community) MarshalJSON() ([]byte, error) {
 		PubsubTopic:                 o.PubsubTopic(),
 		PubsubTopicKey:              o.PubsubTopicKey(),
 		Shard:                       o.Shard(),
+		RequestsToJoinCommunity:     []*RequestToJoin{},
 	}
 	if o.config.CommunityDescription != nil {
 		for id, c := range o.config.CommunityDescription.Categories {
@@ -325,6 +327,7 @@ func (o *Community) MarshalJSON() ([]byte, error) {
 		communityItem.OutroMessage = o.config.CommunityDescription.OutroMessage
 		communityItem.CommunityTokensMetadata = o.config.CommunityDescription.CommunityTokensMetadata
 		communityItem.ActiveMembersCount = o.config.CommunityDescription.ActiveMembersCount
+		communityItem.RequestsToJoinCommunity = o.config.RequestsToJoin
 
 		if o.config.CommunityDescription.Identity != nil {
 			communityItem.Name = o.Name()
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -622,7 +622,21 @@ func (m *Manager) publish(subscription *Subscription) {
 }
 
 func (m *Manager) All() ([]*Community, error) {
-	return m.persistence.AllCommunities(&m.identity.PublicKey)
+	communities, err := m.persistence.AllCommunities(&m.identity.PublicKey)
+	if err != nil {
+		return nil, err
+	}
+
+	nonApprovedRequestsToJoin, err := m.persistence.AllNonApprovedCommunitiesRequestsToJoin()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, community := range communities {
+		community.config.RequestsToJoin = nonApprovedRequestsToJoin[community.IDString()]
+	}
+
+	return communities, nil
 }
 
 type CommunityShard struct {
@@ -1475,7 +1489,7 @@ func (m *Manager) DeleteCategory(request *requests.DeleteCommunityCategory) (*Co
 	return changes.Community, changes, nil
 }
 
-func (m *Manager) GenerateRequestsToJoinForAutoApprovalOnNewOwnership(communityID types.HexBytes, kickedMembers map[string]*protobuf.CommunityMember) error {
+func (m *Manager) GenerateRequestsToJoinForAutoApprovalOnNewOwnership(communityID types.HexBytes, kickedMembers map[string]*protobuf.CommunityMember) ([]*RequestToJoin, error) {
 	var requestsToJoin []*RequestToJoin
 	clock := uint64(time.Now().Unix())
 	for pubKeyStr := range kickedMembers {
@@ -1493,7 +1507,7 @@ func (m *Manager) GenerateRequestsToJoinForAutoApprovalOnNewOwnership(communityI
 		requestsToJoin = append(requestsToJoin, requestToJoin)
 	}
 
-	return m.persistence.SaveRequestsToJoin(requestsToJoin)
+	return requestsToJoin, m.persistence.SaveRequestsToJoin(requestsToJoin)
 }
 
 func (m *Manager) Queue(signer *ecdsa.PublicKey, community *Community, clock uint64, payload []byte) error {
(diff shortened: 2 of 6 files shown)


protocol/communities/community.go
```diff
@@ -266,6 +266,7 @@ func (o *Community) MarshalJSON() ([]byte, error) {
 		PubsubTopic                 string                               `json:"pubsubTopic"`
 		PubsubTopicKey              string                               `json:"pubsubTopicKey"`
 		Shard                       *shard.Shard                         `json:"shard"`
+		RequestsToJoinCommunity     []*RequestToJoin                     `json:"requestsToJoinCommunity"`
```
mprakhov, line 268: RequestsToJoinCommunity will contain non-approved requests to join.
It will be filled during `func (api *PublicAPI) Communities(parent context.Context)`
Providing these changes allows status-desktop to get rid of next API's:

```
func (api *PublicAPI) AllPendingRequestsToJoinForCommunity(id types.HexBytes)
func (api *PublicAPI) DeclinedRequestsToJoinForCommunity
func (api *PublicAPI) CanceledRequestsToJoinForCommunity
func (api *PublicAPI) MyPendingRequestsToJoin()
func (api *PublicAPI) MyAwaitingAddressesRequestsToJoin()
```

@Samyoul, @cammellos could you please assign someone from the mobile team to:
1. check, if these APIs are used on the mobile side
2. if these APIs can be removed

borismelnik (approved):

protocol/communities/manager.go
```diff
@@ -627,6 +627,15 @@ func (m *Manager) All() ([]*Community, error) {
...
 	}
 
+	nonApprovedRequestsToJoin, err := m.persistence.AllNonApprovedCommunitiesRequestsToJoin()
+	if err != nil {
+		return nil, err
+	}
+
+	for index := 0; index < len(communities); index++ {
+		communities[index].config.RequestsToJoin = nonApprovedRequestsToJoin[communities[index].IDString()]
+	}
```
osmaczko: Shouldn't it be in `initializeCommunity`? Otherwise, the `RequestsToJoin` for all other manager's API that retrieves communities will be empty :thinking:

protocol/communities/manager.go
```diff
@@ -627,6 +627,15 @@ func (m *Manager) All() ([]*Community, error) {
 		return nil, err
 	}
 
+	nonApprovedRequestsToJoin, err := m.persistence.AllNonApprovedCommunitiesRequestsToJoin()
+	if err != nil {
+		return nil, err
+	}
+
+	for index := 0; index < len(communities); index++ {
```
MishkaRogachev: suggestion: `for _, community := range(communities)`

MishkaRogachev (approved):

protocol/communities/community.go
cammellos: we use `MyPendingRequestsToJoin`

protocol/communities/manager.go
mprakhov: That's what I did in the beginning, but later, working with the NIM code I understood, that this is senseless.
On the NIM side, we request All communities on the app start. Here we need all community RequestsToJoin.
If there will be any requests to join updates - they will come as a MessengerResponse::RequestsToJoin and will be parsed on the NIM side separately.
On the status-go side - if there are any operations, which require requestsToJoin - we operate with the database directly

P.S. But yes, we can get rid of `communities[index].config.RequestsToJoin` and request requests to join for all communities from the separate API call, but I decided to get rid of the API and return all necessary data for communities when we request All communities (and we do it only on startup)

osmaczko: > That's what I did in the beginning, but later, working with the NIM code I understood, that this is senseless. On the NIM side, we request All communities on the app start. Here we need all community RequestsToJoin. If there will be any requests to join updates - they will come as a MessengerResponse::RequestsToJoin and will be parsed on the NIM side separately. On the status-go side - if there are any operations, which require requestsToJoin - we operate with the database directly

What if mobile uses other endpoints to get communities? We should not assume how the API is used, client logic doesn't matter, using other endpoints should still return complete Community objects.

mprakhov: If they use another endpoint  - that means, that they do not expect to get community with the community.RequestsTojoin. This will impact their current logic. The same we do, we do not expect to get community.RequestsToJoin.
I will remove community.RequestsToJoin and from status-desktop will request all non-approved requests to join through the separate API. This will be required only on app startup, in all other cases, RequestsToJoin will come through MessengerResponse to the status-desktop
This will decrease the risks of the regression

osmaczko: > If they use another endpoint - that means, that they do not expect to get community with the community.RequestsTojoin.

By another endpoint I meant for instance: `func (m *Manager) Controlled() ([]*Community, error)`. This would return Communities without `community.RequestsTojoin` filled.

mprakhov: 1. `initializeCommunity` - does not exist, so this shouldn't be the case anymore, am I right? We will always get _**our**_ request to join
2. Does `Community` contains `RequestsToJoin` or not does not depend on my implementation in the AllCommunities request, it depends on the SQL query of the specific API
3. `func (m *Manager) Controlled()` will return `Community` with _**our**_ request to join, which is expected according to the query
4. If you want to return all requests to join during `GetByID`, we can create a separate ticket for this as this is a separate task, which is not related to the task I'm working on (and it must be tested on the regression). As you can see, I rewrote `AllCommunities` API only and it has optimized logic for getting All communities with All non-approved requests to join this community

osmaczko: > 1. `initializeCommunity` - does not exist, so this shouldn't be the case anymore, am I right? We will always get _**our**_ request to join

It does, is a part of the [callback](https://github.com/status-im/status-go/blob/f08e6f7137c543e2867a6dcd6a6224f9ddbc5221/protocol/communities/manager.go#L3160) now, so it can't be avoided. Since you decided to go with a separate API, then yes, it is not a concern anymore.

> 2. Does `Community` contains `RequestsToJoin` or not does not depend on my implementation in the AllCommunities request, it depends on the SQL query of the specific API

You added the logic in `manager.go` not in SQL query itself.

> 3. `func (m *Manager) Controlled()` will return `Community` with _**our**_ request to join, which is expected according to the query

Please correct me if I am wrong, but with your initial changes I reviewed,
- `All` endpoint would return Communities with not empty `RequestsToJoinCommunity` and not empty `MyPendingRequestsToJoin`
- `Controlled` endpoint would return Communities with empty `RequestsToJoinCommunity` and not empty `MyPendingRequestsToJoin`

Isn't the principle of least surprise violated here?

> 4. If you want to return all requests to join during `GetByID`, we can create a separate ticket for this as this is a separate task, which is not related to the task I'm working on (and it must be tested on the regression). As you can see, I rewrote `AllCommunities` API only and it has optimized logic for getting All communities with All non-approved requests to join this community

It was enough to move this part of the code to `itinializeCommunity`, not extra work required.

mprakhov: 2. Yes, because the SQL query will return `only` my request to join. So I can easily drop it as I will be receiving the request of all non-approved requests to join the community in the next step (I can modify SQL query to drop getting my request to join, I thought I changed it before).

3. `All` will return all communities and fill these communities with non-approved requests to join if they exist. Please, pay attention, that `All` will be called `only` on an application startup, we are not interested in approved requests to join on a startup

> Isn't the principle of least surprise violated here?
This question is not to me, I don't know where `Controlled` is used and what kind of requests to join it is expecting. From the query I see - that it expects only my own request to join. If you see, that it contains not full data, please, create a ticket/bug with the description. It would be good if you could reproduce it in the application

`It was enough to move this part of the code to itinializeCommunity, not extra work required.`
1. Can you guarantee that these changes do not impact anything on the status-go, desktop, and mobile side? 
2. Could you please prove that this is a bug and not expected behavior?
3. Could you please prove that this will not downgrade the performance for the case, where we have 100 communities with 1000 requests to join?

Again, if you want to put all requests to join to the GetByID - please, create a separate ticket as it is a change request and it broke existing logic in the current application. Current logic does not expect, that we are returning all requests to join in GetByID and if we do not do it, I bet there was a reason not to do it

mprakhov: and we can move discussion into this ticket 😄

PR #4411, opened 2023-12-01
fix: use community controlNode pubkey instead of community ID for verifying the received community data signer

mprakhov:
If the community has an owner token, we must verify the signature using ControlNode pubkey.
If the community does not have an owner token, calling ControlNode() will return community ID

Closes #https://github.com/status-im/status-desktop/issues/12902


igor-sirotin (approved):

MishkaRogachev (approved):
LGTM!

osmaczko (approved):

endulab (approved):

PR #4414, opened 2023-12-01
chore(walletconnect)_: various improvements applied

saledjenic:
These changes are the second step in cleaning the wallet API and moving the things to the place where they logically belong.

- `WalletConnectTransfer` identified as a new transfer type
- Wallet-related endpoints that logically belong to the wallet moved from the wallet connect service
- Wallet connect service now receives `transfer.TransactionManager` instead of `transactions.Transactor`
- Deadlock issue when trying to send the tx with the wrong nonce fixed


stefandunca (approved):
Nice simplification

alaibe (approved):

PR #4416, opened 2023-12-03
fix: merge chat response with response to joining a community

### LINKED ISSUE #17801: Sometimes channel is not fetched if opened after joining community (E2E blocker)
According to @Parveshdhull [comment](https://github.com/status-im/status-mobile/pull/17704#issuecomment-1778973906) this issue might require fix on status-go side.

This issues is often caught by our e2e causing the whole group of tests to fail. So it can be considered as test blocker. 

Sometimes this bug is also reproducible manually.

**Steps:** 

1. User A creates community which does not require admin manual approval. Use PR build to create community from mobile. 
2. User A sends community invite to User B
3. User B sends Request to join the community
4. Wait until community status is changed to Joined for User B
5. User B opens community channel
6. See if channel is fetched

**Actual result:** sometimes channel is not fetched

[logs (11).zip](https://github.com/status-im/status-mobile/files/13068070/logs.11.zip)

https://github.com/status-im/status-mobile/assets/97245802/162146e5-2e12-4589-ae82-7f0e1dc9a318

**Expected result:** channel is fetched

### Additional Information

- Status version: nightly 

siddarthkay:
### Summary: 

When an open community was created by Device A and shared with Device B and when Device B would request to join such a community, the general channel would be forever in loading state.

This happened because as part of messenger response the chatId of general channel was not sent and mobile client would not fetch that chat data.

This PR fixes that issue by sending chatId as part of messenger response right after the request to join community succeeds.

fixes https://github.com/status-im/status-mobile/issues/17801

### DIFF
--- a/protocol/messenger.go
+++ b/protocol/messenger.go
@@ -4033,6 +4033,7 @@ func (m *Messenger) MessageByChatID(chatID, cursor string, limit int) ([]*common
 		}
 
 	}
+
 	if m.httpServer != nil {
 		for idx := range msgs {
 			m.prepareMessage(msgs[idx], m.httpServer)
--- a/protocol/messenger_handler.go
+++ b/protocol/messenger_handler.go
@@ -1620,11 +1620,16 @@ func (m *Messenger) HandleCommunityRequestToJoinResponse(state *ReceivedMessageS
 		if err != nil {
 			return err
 		}
+
+		// we merge to include chats in response signal to joining a community
+		err = state.Response.Merge(response)
+		if err != nil {
+			return err
+		}
+
 		if len(response.Communities()) > 0 {
 			communitySettings := response.CommunitiesSettings()[0]
 			community := response.Communities()[0]
-			state.Response.AddCommunity(community)
-			state.Response.AddCommunitySettings(communitySettings)
 
 			magnetlink := requestToJoinResponseProto.MagnetUri
 			if m.torrentClientReady() && communitySettings != nil && communitySettings.HistoryArchiveSupportEnabled && magnetlink != "" {
--- a/protocol/communities_messenger_test.go
+++ b/protocol/communities_messenger_test.go
@@ -182,6 +182,96 @@ func (s *MessengerCommunitiesSuite) TestRetrieveCommunity() {
 	s.Require().Equal(community.IDString(), response.Messages()[0].CommunityID)
 }
 
+func (s *MessengerCommunitiesSuite) TestJoiningOpenCommunityReturnsChatsResponse() {
+	ctx := context.Background()
+
+	openCommunityDescription := &requests.CreateCommunity{
+		Name:                         "open community",
+		Description:                  "open community to join with no requests",
+		Color:                        "#26a69a",
+		HistoryArchiveSupportEnabled: true,
+		Membership:                   protobuf.CommunityPermissions_AUTO_ACCEPT,
+		PinMessageAllMembersEnabled:  false,
+	}
+
+	response, err := s.bob.CreateCommunity(openCommunityDescription, true)
+	generalChannelChatID := response.Chats()[0].ID
+	s.Require().NoError(err)
+	s.Require().NotNil(response)
+	s.Require().Len(response.Communities(), 1)
+	s.Require().Len(response.CommunitiesSettings(), 1)
+	s.Require().Len(response.Chats(), 1)
+
+	community := response.Communities()[0]
+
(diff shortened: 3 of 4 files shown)


protocol/messenger.go
```diff
@@ -4027,10 +4029,13 @@ func (m *Messenger) MessageByChatID(chatID, cursor string, limit int) ([]*common
 		}
 
 	}
-	if m.httpServer != nil {
-		for idx := range msgs {
-			m.prepareMessage(msgs[idx], m.httpServer)
-		}
+
+	if m.httpServer == nil {
```
MishkaRogachev: i'm not sure, but do we always have `httpServer`?

siddarthkay: oh I should probably revert back all of the changes to this file since these were made for debugging purposes only.

protocol/messenger_handler.go
```diff
@@ -1622,11 +1622,20 @@ func (m *Messenger) HandleCommunityRequestToJoinResponse(state *ReceivedMessageS
...
+			chatIDs, err := m.getChatIdsForCommunity(community.ID())
+			if err != nil {
+				return err
+			}
+
+			if len(chatIDs) > 0 {
+				generalChannelChatID := m.Chat(chatIDs[0])
+				state.Response.AddChat(generalChannelChatID)
+			}
+
```
osmaczko: Don't we need it as well when we spectate the community?

protocol/messenger_handler.go
```diff
@@ -1622,11 +1622,20 @@ func (m *Messenger) HandleCommunityRequestToJoinResponse(state *ReceivedMessageS
 			communitySettings := response.CommunitiesSettings()[0]
 			community := response.Communities()[0]
 			state.Response.AddCommunity(community)
+			chatIDs, err := m.getChatIdsForCommunity(community.ID())
+			if err != nil {
+				return err
+			}
+
+			if len(chatIDs) > 0 {
+				generalChannelChatID := m.Chat(chatIDs[0])
```
ilmotta: The admin can remove the `#general` channel after creation.

ilmotta (approved):

protocol/messenger_handler.go
cammellos: @siddarthkay maybe put the pr back to draft if it's not ready, we dicussed a different way to handle this, we should merge messenger responses

siddarthkay: @cammellos: I've adjusted the current PR to merge the responses, It is indeed a neater approach.
Also added a test to verify.

protocol/messenger.go
siddarthkay: @MishkaRogachev : this file was reverted.

protocol/messenger_handler.go
siddarthkay: I'm not sure what you mean @osmaczko : would you mind elaborating?

protocol/messenger_handler.go
```diff
@@ -1620,6 +1620,13 @@ func (m *Messenger) HandleCommunityRequestToJoinResponse(state *ReceivedMessageS
...
 		}
+
+		// we merge to include chats in response signal to joining a community
+		err = state.Response.Merge(response)
+		if err != nil {
+			return err
+		}
+
 		if len(response.Communities()) > 0 {
 			communitySettings := response.CommunitiesSettings()[0]
```
cammellos, line 1631: can we extend `MessengerResponse` to also merge `communitySettings` so we can get rid of these lines?

siddarthkay: ah sorry forgot it, we had discussed this as well, I will take care of this :)

protocol/messenger_handler.go
osmaczko: Sure. We can [spectate](https://github.com/status-im/status-go/blob/058120fad95056627badd8854b6655d70258db4c/protocol/messenger_communities.go#L736) community before we join it. For open communities we should see general channel in that case as well.

siddarthkay: Please correct me If I'm wrong, but I think that already happens because of this line here : https://github.com/status-im/status-go/blob/058120fad95056627badd8854b6655d70258db4c/protocol/messenger_communities.go#L751 

On the mobile client we do see the general channel when we request to join a community (after it has been shared).

The updated approach now in this PR is to merge responses (messenger response + response to joining community) so that chat information (i.e channel information) was not missed in that particular response signal. 
In the case of open community it is the general channel.

protocol/messenger_handler.go
siddarthkay: Done

cammellos (approved):

osmaczko (approved):

MishkaRogachev (approved):

Issue #4355, opened 2023-11-21
Block sending messages that exceed the waku limit and return an error


jrainville:
# Problem

A couple of times, we've been bit by the fact that waku has a size limit and we don't enforce it correctly in status-go, or at least with waku v2.

What should happen when we try to send a message that exceeds the limit is we block sending it and return an error. That way, the client can show a proper error.

What happens currently is the action is sent by the user and no error is ever shown, not even in the console, so the message will never be received by anyone.

This happened before with the community description. The community never updated anymore, but there was no info received by the clients.

# Acceptance Criteria

When a message is too big to be sent, an error is returned and gets propagated back to the client.

jrainville:
cc @richard-ramos

kaichaosun:
Hi @jrainville I wonder how to reproduce this issue, could you confirm if this issue still exist or not?

I tried to create a community with long description and big image, also send out long messages. I noticed that status-desktop has a length limit regarding to message length and description length (140 characters). Test against the build binary https://github.com/status-im/status-desktop/pull/12765.
Also test the big image which is actually compressed during our pair programming between @igor-sirotin and me.

When reading the code, I learn that there are some latest changes which split big message to segments (https://github.com/status-im/status-go/pull/4285).

Along with the max msg length check added by @chaitanyaprem.

jrainville:
The problem has mostly been circumvented by @osmaczko by splitting the message in chunks, but it's an app layer mechanism and I think fixing this problem on a lower level would still help since it would make sure we have info if this happens again (for example the limit changes and we forget to update the app level mechanism).

@osmaczko can help you deactivate the mechanism and test sending a big message.

osmaczko:
> @osmaczko can help you deactivate the mechanism and test sending a big message.

The simplest way to deactivate the message segmentation for testing purposes would be:
```diff
diff --git a/protocol/common/message_sender.go b/protocol/common/message_sender.go
index f7013b8e6..d6eb91d3d 100644
--- a/protocol/common/message_sender.go
+++ b/protocol/common/message_sender.go
@@ -1286,7 +1286,7 @@ func segmentMessage(newMessage *types.NewMessage, maxSegmentSize int) ([]*types.
 func (s *MessageSender) segmentMessage(newMessage *types.NewMessage) ([]*types.NewMessage, error) {
     // We set the max message size to 3/4 of the allowed message size, to leave
     // room for segment message metadata.
[...]

PR #4421, opened 2023-12-05
test: ensure max message length is not excessed when send

kaichaosun:
The check both happens in api here and [waku side](https://github.com/waku-org/go-waku/pull/939). 
It seems necessary since error raised in go-waku is not propagated to api.

Relates to https://github.com/status-im/status-go/issues/4355

### DIFF
--- a/wakuv2/api.go
+++ b/wakuv2/api.go
@@ -255,6 +255,14 @@ func (api *PublicWakuAPI) Post(ctx context.Context, req NewMessage) (hexutil.Byt
 		Ephemeral:    &req.Ephemeral,
 	}
 
+	encodedMsg, err := proto.Marshal(wakuMsg)
+	if err != nil {
+		return nil, err
+	}
+	if len(encodedMsg) > int(api.w.MaxMessageSize()) {
+		return nil, errors.New("message size exceeds max size allowed by waku node")
+	}
+
 	hash, err := api.w.Send(req.PubsubTopic, wakuMsg)
 
 	if err != nil {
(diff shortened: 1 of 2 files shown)


chaitanyaprem:
> The check both happens in api here and [waku side](https://github.com/waku-org/go-waku/pull/939). It seems necessary since error raised in go-waku is not propagated to api.
> 
> Relates to #4355


Isn't it better to propagate the error from go-waku rather than having a duplicate check?

kaichaosun:
Reuse the logic would be great, but when I look into the code, errors raised in go-waku is converted to an expired event without any error information [here](https://github.com/status-im/status-go/blob/develop/wakuv2/waku.go#L1050).

I don't have enough knowledge about how those events are communicated, and not sure if it's acceptable to add a new event to identify error scenarios. Please guide me if you have a plan. @chaitanyaprem

chaitanyaprem:
> Reuse the logic would be great, but when I look into the code, errors raised in go-waku is converted to an expired event without any error information [here](https://github.com/status-im/status-go/blob/develop/wakuv2/waku.go#L1050).
> 
> I don't have enough knowledge about how those events are communicated, and not sure if it's acceptable to add a new event to identify error scenarios. Please guide me if you have a plan. @chaitanyaprem



The code you have pointed out already captures the error returned and logs it as well. So, if go-waku returns the error in case the message size is too big, then this gets logged which should be sufficient right. 
Unless you want to take an action at a higher layer based on this error (which doesn't seem to be the case as per this PR), I don't see the need to have this check again in status-go code.

Issue #4419, opened 2023-12-04
Add a new API for getting all non-approved requests to join


mprakhov:
# Problem
Add a new API for getting all non-approved requests to join

PR #4422, opened 2023-12-05
chore: API for getting all non-approved requests to join for all communities

mprakhov:
API for getting all non-approved requests to join for all communities

After a discussion, we decided to reject from filling all communities with all non-approved requests to join during Communities API call

Instead, we will use a separate API for getting all non-approved communities requests to join

Closes #https://github.com/status-im/status-go/issues/4419

status-desktop PR: https://github.com/status-im/status-desktop/pull/12941


osmaczko (approved):
Thank you.

MishkaRogachev (approved):

PR #4425, opened 2023-12-05
fix: tokenMaster does not have members revealed addresses

mprakhov:
fixed:
- control node sent to TokenMaster request to join with revealed addresses
- token master saves revealed addresses from control node msg during `handleCommunityPrivilegedUserSyncMessage`
- signer msg validation in `handleCommunityPrivilegedUserSyncMessage`
- drop all requests to join data on membership role change
- if we became a privileged member and received all requests to join, send to the backend only non-approved requests to join

Closes # https://github.com/status-im/status-desktop/issues/12903

### DIFF
--- a/protocol/communities/persistence.go
+++ b/protocol/communities/persistence.go
@@ -1669,3 +1669,12 @@ func (p *Persistence) AllNonApprovedCommunitiesRequestsToJoin() ([]*RequestToJoi
 	}
 	return nonApprovedRequestsToJoin, nil
 }
+
+func (p *Persistence) RemoveAllCommunityRequestsToJoinWithRevealedAddressesExceptPublicKey(pk string, communityID []byte) error {
+	_, err := p.db.Exec(`
+	DELETE FROM communities_requests_to_join_revealed_addresses
+		WHERE request_id IN (SELECT id FROM communities_requests_to_join WHERE community_id = ? AND public_key != ?);
+	DELETE FROM communities_requests_to_join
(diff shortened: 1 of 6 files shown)


protocol/communities/persistence.go
```diff
@@ -1669,3 +1669,40 @@ func (p *Persistence) AllNonApprovedCommunitiesRequestsToJoin() ([]*RequestToJoi
 	}
 	return nonApprovedRequestsToJoin, nil
 }
+
+func (p *Persistence) RemoveAllCommunityRequestsToJoinExceptOur(pk string, communityID []byte) error {
```
MishkaRogachev: suggestion: it can be any pk, so `RemoveAllCommunityRequestsToJoinExceptPublicKey`

protocol/communities/persistence.go
```diff
@@ -1669,3 +1669,40 @@ func (p *Persistence) AllNonApprovedCommunitiesRequestsToJoin() ([]*RequestToJoi
...
+	for rows.Next() {
+		requestID := []byte{}
+		err := rows.Scan(&requestID)
+		if err != nil {
+			return err
+		}
+		requestsID = append(requestsID, requestID)
+	}
+
+	err = p.RemoveAllCommunityRequestsToJoinExceptOur(pk, communityID)
```
MishkaRogachev: suggestion: can we put all the queries into one transaction?

MishkaRogachev (approved):

PR #4427, opened 2023-12-06
fix: prevent crash when erc20 tx is contract creation

alaibe:


dlipicar (approved):

services/wallet/transfer/commands.go
```diff
@@ -399,7 +399,8 @@ func (c *transfersCommand) checkAndProcessBridgeMultiTx(ctx context.Context, tx
 
 func (c *transfersCommand) processUnknownErc20CommunityTransactions(ctx context.Context, allTransfers []Transfer) {
 	for _, tx := range allTransfers {
-		if tx.Type == w_common.Erc20Transfer {
+		// To can be nil in case of erc20 contract creation
+		if tx.Type == w_common.Erc20Transfer && tx.Transaction.To() != nil {
```
dlipicar, line 403: Ideally an erc20 contract deploy shouldn't be marked with type `w_common.Erc20Transfer`... Perhaps create an issue to look into this later

PR #4428, opened 2023-12-06
fix(wallet): use mnemonic displayed check for 'findNewBlocksCommand'

IvanBelyakoff:
Since `findNewBlocksCommand` now has its own `Run` method, avoid searching blocks with transfers for idle state as well.
Fixed some tests to use wrong sql DB for account DB

### DIFF
--- a/services/wallet/transfer/commands_sequential.go
+++ b/services/wallet/transfer/commands_sequential.go
@@ -54,7 +54,25 @@ func (c *findNewBlocksCommand) Run(parent context.Context) (err error) {
 
 func (c *findNewBlocksCommand) findAndSaveEthBlocks(parent context.Context, fromNum, headNum *big.Int) {
 	// Check ETH transfers for each account independently
+	mnemonicWasNotShown, err := c.accountsDB.GetMnemonicWasNotShown()
+	if err != nil {
+		c.error = err
+		return
+	}
+
 	for _, account := range c.accounts {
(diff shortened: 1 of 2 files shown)


Cuteivist (approved):

services/wallet/transfer/commands_sequential.go
```diff
@@ -54,7 +54,25 @@ func (c *findNewBlocksCommand) Run(parent context.Context) (err error) {
...
 	for _, account := range c.accounts {
+		if mnemonicCheckEnabled && mnemonicWasNotShown {
+			acc, err := c.accountsDB.GetAccountByAddress(nodetypes.Address(account))
+			if err != nil {
+				c.error = err
+				return
+			}
+			if acc.AddressWasNotShown {
+				log.Info("skip findBlocksCommand, mnemonic has not been shown and the address has not been shared yet", "address", account)
+				return
```
dlipicar: double-checking here, we want to skip the command for all remaining accounts if the address of only one of them has not been shared?
Otherwise this would have to be a `continue`

rasom (approved):

services/wallet/transfer/commands_sequential.go
IvanBelyakoff: Sharp eye! I overlooked it, thanx. Fixed

dlipicar (approved):

PR #4436, opened 2023-12-07
fix(wallet): quick fix for eating RPC requests. Proper fix TBD

IvanBelyakoff:
1. An error in getting head block causes early return of the `Run` function but transfers loop is already started, which will be started once again after command will call Run again on timeout.
Avoid starting transfers loop for now before getting head block.

TBD: Proper fix is WIP

2. Fixed a bug that new ETH blocks in idle state were searched only for the first account on the list. Thanx @rasom for discovering it and pointing to the actual problem.

### DIFF
--- a/services/wallet/transfer/commands_sequential.go
+++ b/services/wallet/transfer/commands_sequential.go
@@ -196,7 +196,7 @@ func (c *findNewBlocksCommand) findBlocksWithEthTransfers(parent context.Context
 
 		var newFromBlock *Block
 		var ethHeaders []*DBHeader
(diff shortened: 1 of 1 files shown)


rasom (approved):

stefandunca (approved):
👍

rasom:
the issue is not fixed by this commit

rasom:
to clarify: my issue originally was not exactly limit, but that `eth_getBalanceAt` requests did not happen for some accounts. I thought it was related to limits, because there were errors related to limits, but apparently that's just another bug in the code related to this line https://github.com/status-im/status-go/blob/develop/services/wallet/transfer/commands_sequential.go#L599

PR #4439, opened 2023-12-08
fix: a few adjustments for matterbridge

### LINKED ISSUE #12710: [Bridge] Bridge status communities to Discord
### Description

Continue the work started in the spike https://github.com/status-im/status-desktop/issues/12610

### Acceptance Criteria

- Messages sent to a Status community channel are bridged to a Discord community channel and vice-versa

endulab:
A few adjustments for matterbridge
- exposing Online info
- adding ability to disable backup loop and auto message loop (they are not needed in the bridge)

Issue #12710

### DIFF
--- a/protocol/messenger.go
+++ b/protocol/messenger.go
@@ -796,10 +796,14 @@ func (m *Messenger) Start() (*MessengerResponse, error) {
 	m.watchPendingCommunityRequestToJoin()
 	m.broadcastLatestUserStatus()
 	m.timeoutAutomaticStatusUpdates()
-	m.startBackupLoop()
-	err = m.startAutoMessageLoop()
-	if err != nil {
-		return nil, err
+	if !m.config.featureFlags.DisableCheckingForBackup {
+		m.startBackupLoop()
+	}
+	if !m.config.featureFlags.DisableAutoMessageLoop {
+		err = m.startAutoMessageLoop()
+		if err != nil {
+			return nil, err
+		}
 	}
 	m.startSyncSettingsLoop()
 	m.startSettingsChangesLoop()
@@ -930,7 +934,7 @@ func (m *Messenger) handleConnectionChange(online bool) {
 	m.ensVerifier.SetOnline(online)
 }
 
-func (m *Messenger) online() bool {
+func (m *Messenger) Online() bool {
 	switch m.transport.WakuVersion() {
 	case 2:
 		return m.transport.PeerCount() > 0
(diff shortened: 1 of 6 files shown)


MishkaRogachev (approved):
Great work!

protocol/messenger.go
```diff
@@ -801,10 +801,14 @@ func (m *Messenger) Start() (*MessengerResponse, error) {
 	m.watchPendingCommunityRequestToJoin()
 	m.broadcastLatestUserStatus()
 	m.timeoutAutomaticStatusUpdates()
-	m.startBackupLoop()
-	err = m.startAutoMessageLoop()
-	if err != nil {
-		return nil, err
+	if !m.config.featureFlags.BackupDisabled {
```
saledjenic: There is already a `settings.BackupEnabled` field that can be used for this purpose?

protocol/messenger_backup.go
```diff
@@ -32,13 +32,17 @@ func (m *Messenger) lastBackup() (uint64, error) {
 	return m.settings.LastBackup()
 }
 
+func (m *Messenger) EnableBackups(enable bool) error {
```
saledjenic: I don't see where is this used?

protocol/messenger_config.go
```diff
@@ -275,6 +275,20 @@ func WithPushNotifications() func(c *config) error {
 	}
 }
 
+func WithBackupDisabled() func(c *config) error {
```
saledjenic: The same for this and the one below, where do we use them?

protocol/messenger_backup.go
endulab: Aaa, right. I do not use this one.

protocol/messenger_config.go
endulab: It is used in the bridge repo: https://github.com/status-im/matterbridge/pull/1/files#diff-a04d7ca2cd2cc8163c50e09baf54a0c66a04242f0ced84601144cb2a7abd87c1R313

I know that we have settings switch, but bridge's settings table is empty. And default setting (for NoRows error) is true. Maybe we can change it to false? Then I can remove this one.
Also, what about checking BackupEnabled flag before running a goroutine in startBackupLoop() ?

saledjenic: > what about checking BackupEnabled flag before running a goroutine in startBackupLoop()?

I am pretty sure, that you can do that, actually that's how it should be.

> Maybe we can change it to false?

I don't think we need it at all.

When you call `backend.StartNode` you can pass `false` for `ProcessBackedupMessages` via `NodeConfig` and you can also check how to set the initial setting via the passed node config which will set `BackupEnabled` in the settings table (check the usage of `CreateSettings` function).

endulab: @saledjenic I fixed naming according to our discussion

saledjenic (approved):

PR #4441, opened 2023-12-10
feat(wallet): better lifecycle handling of blocks and transfers search commands

IvanBelyakoff:
- Infinite commands started only once and never restarted, stopped on context cancellation
- Finite commands are joined into AtomicGroup to stop the rest in the group in case one command fails. Otherwise other commands in the group will continue running and the failed command is not retried to restart until other commands finish.
- Fixed goroutines leakage in case of failure of some commands

TBD: error handling in findBlocksCommand

Updates [#11074](https://github.com/status-im/status-desktop/issues/11074)

### DIFF
--- a/services/wallet/transfer/commands_sequential.go
+++ b/services/wallet/transfer/commands_sequential.go
@@ -3,6 +3,7 @@ package transfer
 import (
 	"context"
 	"math/big"
+	"sync"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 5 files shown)


services/wallet/transfer/commands_sequential.go
```diff
@@ -730,17 +775,53 @@ type loadBlocksAndTransfersCommand struct {
...
+	errorCounter
+	mu sync.Mutex
 }
 
-func (c *loadBlocksAndTransfersCommand) Run(parent context.Context) error {
-	log.Debug("start load all transfers command", "chain", c.chainClient.NetworkID(), "accounts", c.accounts)
+func (c *loadBlocksAndTransfersCommand) incStarted() {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.loops++
```
dlipicar, line 786: matter of preference, if it's just to have a simple thread-safe counter or boolean flag I like using atomics instead, which normally have less overhead

dlipicar (approved):
In general it looks fine. I feel there's a complex mixture of commands/goroutines being spawned from several different places, there's possibly some way to organize the code differently to more easily keep track of what happens where, what's blocking and whats async. Perhaps it can be looked into later when you have the time.

IvanBelyakoff:
> In general it looks fine. I feel there's a complex mixture of commands/goroutines being spawned from several different places, there's possibly some way to organize the code differently to more easily keep track of what happens where, what's blocking and whats async. Perhaps it can be looked into later when you have the time.

Agree with you. The implementation and tracking of what commands and where are spawn is somewhat complex and confusing. I might think of some way to simplify that indeed. At least now I add more tests and hope that will help us to better understand the logic and introduce changes

PR #4446, opened 2023-12-11
chore: store node requests manager

### LINKED ISSUE #4192: refactor and fix `RequestCommunityInfoFromMailserver`

igor-sirotin:
Closes https://github.com/status-im/status-go/issues/4192

# Decsription

I implemented a new `StoreNodeRequestsManager`. 
This fixes all of the bugs we had with `requestCommunityInfoFromMailserver` and gives more abilities over the process.

## Benefits

1. Automatically wait for an available store node
2. If there's an active request for a community, subscribe to the results channel and don't start a new request.
3. Able to make multiple simultaneous calls to `FetchCommunity` and/or `FetchCommunities`
4. Inplace process the received envelopes and remove "200ms period loop"
5. Instead of processing full month of requested data, process each fetched envelopes page and stop the process if community found. 
6. Make a first "request of faith" with a single envelope page limit.
8. The code is encapsulated in a single class.
9. Common implementation for fetching a single and multiple communities.

##  Tests! Tests! Tests!
I started with implementing `MessengerStoreNodeRequestSuite`. 
All of the tests were initially failing and, but no with the new code.


protocol/messenger_mailserver_cycle.go
```diff
@@ -709,3 +709,44 @@ func (m *Messenger) disconnectStorenodeIfRequired() error {
...
+
+	go func() {
+		defer func() {
+			close(finish)
+		}()
+		wg.Wait()
+	}()
+
+	select {
+	case <-finish:
```
richard-ramos, line 778: Probably it's a good idea to add `case <-m.ctx.Done():`

protocol/messenger_store_node_request_manager.go
```diff
@@ -0,0 +1,327 @@
...
+	stats     FetchCommunityStats
+}
+
+type communitySubscriptionChannel = chan fetchCommunityResult
+
+func (r *storeNodeRequest) subscribe() *communitySubscriptionChannel {
+	channel := make(communitySubscriptionChannel, 100)
+	r.subscriptions = append(r.subscriptions, &channel)
+	return &channel
+}
```
richard-ramos: You can return the channel directly. No need to return a reference to it as channels already behave like references. (This will require changing `r.subscriptions` to not store channel references)

```suggestion
func (r *storeNodeRequest) subscribe() communitySubscriptionChannel {
	channel := make(communitySubscriptionChannel, 100)
	r.subscriptions = append(r.subscriptions, channel)
	return channel
}
```

protocol/messenger_store_node_request_manager.go
```diff
@@ -0,0 +1,327 @@
...
+
+	r.manager.logger.Debug("request finished",
+		zap.String("communityID", r.communityID),
+		zap.Bool("communityFound", r.result.community != nil),
+		zap.Error(r.result.err))
+
+	// Send the result to subscribers
+	// It's important that this is done with `activeRequestsLock` locked.
+	for _, s := range r.subscriptions {
+		*s <- r.result
```
richard-ramos: Should we close this channel right after pushing the result?

richard-ramos (approved):
:rocket: 

This PR should probably be tested by mobile and desktop QA!

protocol/messenger_store_node_request_manager.go
```diff
@@ -0,0 +1,327 @@
...
+)
+
+// FetchCommunityStats is used in tests
+type FetchCommunityStats struct {
+	FetchedEnvelopesCount int
+	FetchedPagesCount     int
+}
+
+type StoreNodeRequestManager struct {
+	messenger *Messenger
```
osmaczko, line 27: Would it make sense to use interface instead to break bidirectional dependency?

osmaczko (approved):
Nice, a lot of issues solved :muscle:

protocol/messenger_store_node_request_manager.go
igor-sirotin: So I checked this. Here's a list of Messenger functions used here:
- `.transport.FilterByChatID`
- `.transport.InitPublicFilters`
- `transport.RemoveFilters`
- `.passStoredCommunityInfoToSignalHandler`
- `.ProcessAllMessages`
- `.communitiesManager.GetByIDString`
- `.waitForAvailableStoreNode`
- `.GetCurrentTimeInMillis`
- `.performMailserverRequest`
- `.processMailserverBatchWithOptions`

For me it looks like it's better to keep bidirectional dependency than having such an interface :smile:
Though I'm sure this can be simplified to less methods.

I'll leave it as is for now and see where what we have after implementing contact fetch in the same way. It will makes more sense to refactor after that.

igor-sirotin:
Prepared PRs for QA:
- https://github.com/status-im/status-desktop/pull/13012
- https://github.com/status-im/status-mobile/pull/18170

PR #4449, opened 2023-12-12
fix(history-archive): Skip dowloading data from torrent if free space  is not enought

borismelnik:
Full description of bug is [here](https://github.com/status-im/status-desktop/issues/9404)

This PR adds the check for ability to download the archives

Important changes:

Closes #


MishkaRogachev (approved):
Seems to be good! But I don't have confidence in math here :)

mprakhov (approved):

PR #4450, opened 2023-12-12
Feat/wallet community manager

Cuteivist:
First part of https://github.com/status-im/status-desktop/issues/12985

* Added CommunityManager to have common community API for collectibles and tokens
* Added get from cached community data

Currently tokens doesn't fetch for community. It will be updated after @dlipicar changes to fetching implementation.

### DIFF
--- a/services/wallet/community/manager.go
+++ b/services/wallet/community/manager.go
@@ -1,0 +1,87 @@
+package community
+
+import (
+	"database/sql"
+	"fmt"
+	"time"
+
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/status-im/status-go/services/wallet/thirdparty"
+)
+
+const failedCommunityFetchRetryDelay = 1 * time.Hour
+
+type Manager struct {
+	db                    *DataDB
+	communityInfoProvider thirdparty.CommunityInfoProvider
+}
+
+func NewManager(db *sql.DB) *Manager {
+	return &Manager{
+		db: NewDataDB(db),
+	}
+}
+
+// Used to break circular dependency, call once as soon as possible after initialization
+func (cm *Manager) SetCommunityInfoProvider(communityInfoProvider thirdparty.CommunityInfoProvider) {
+	cm.communityInfoProvider = communityInfoProvider
+}
+
+func (cm *Manager) GetCommunityInfo(id string) (*thirdparty.CommunityInfo, *InfoState, error) {
+	return cm.db.GetCommunityInfo(id)
+}
+
+func (cm *Manager) GetCommunityID(tokenURI string) string {
+	return cm.communityInfoProvider.GetCommunityID(tokenURI)
+}
+
+func (cm *Manager) FillCollectibleMetadata(c *thirdparty.FullCollectibleData) error {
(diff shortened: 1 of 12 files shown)


services/wallet/community/manager.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/status-im/status-go/services/wallet/thirdparty"
+)
+
+const failedCommunityFetchRetryDelay = 1 * time.Hour
+
+type Manager struct {
+	db *DataDB
+}
+
+func NewManager(db *sql.DB) *Manager {
```
dlipicar, line 19: makes more sense to me to have the `Manager` hold the instance to the `CommunityInfoProvider`, I don't think that's going to change depending on who calls `FetchCommunityInfo`

services/wallet/community/manager.go
```diff
@@ -0,0 +1,73 @@
...
+	return &Manager{
+		db: NewDataDB(db),
+	}
+}
+
+func (cm *Manager) SetCommunityInfo(id string, c *thirdparty.CommunityInfo) (err error) {
```
dlipicar: this probably shouldn't be called from outside this class, better make it private

Cuteivist: Good point. Fixed.

services/wallet/community/manager.go
Cuteivist: @dlipicar 
I moved CollectiblesCommunityInfoProvider and created WalletCommunityInfoProvider. Removed instance from Collectibles Manager and moved it to Community Manager, added required calls. 

What do you think about this approach ? I like that community provider is only in Community Manager, it makes a lot of sense. Wallet provider itself doesn't introduce additional dependencies as it uses thirdparty structs, so it can also have collectibles (and probably token in future) methods.

IvanBelyakoff (approved):
LGTM

services/wallet/thirdparty/wallet_types.go
```diff
@@ -0,0 +1,6 @@
+package thirdparty
+
+type WalletCommunityInfoProvider interface {
+	CommunityInfoProvider
+	FillCollectibleMetadata(collectible *FullCollectibleData) error
```
dlipicar: Hmm if we're not gonna have separate Collectible-related and Token-related interfaces, we can simplify things and make `FillCollectibleMetadata` part of `CommunityInfoProvider` (`CommunityInfoProvider` is an interface to get wallet-relevant data from the community code, so this additional layer on top of it feels kind of redundant).
Anyway, not that important, leave it as is if you feel it's better.

services/wallet/community/manager.go
dlipicar: yeap, I'm OK with this. Just a small comment below

dlipicar (approved):
great job!

services/wallet/thirdparty/wallet_types.go
Cuteivist: Awesome. I though that `CommunityInfoProvider` interface would be used elsewhere.

That makes full sense. Updated the code.

PR #4454, opened 2023-12-13
feat: lazy load collectibles metadata

dlipicar:
Part of https://github.com/status-im/status-desktop/issues/12494
status-desktop part: https://github.com/status-im/status-desktop/pull/13010

This PR causes Community Collectible metadata to be fetched asynchronously.
Since ownership updates retrieve fresh metadata from our providers, we can use that to update our local cache for non-community tokens. However fetching collectible metadata from Community Descriptions might take long, or even timeout. For this reason, we now do this part of the process asynchronously.

Any time the metadata for some collectibles is updated, event `"wallet-collectibles-data-updated"` is emitted containing the new metadata in the `CollectibleDataTypeDetails` form. We do this in chunks so we don't transmit a huge payload in a single event. The client must react to this event and update its model with the new data.

We could use this new mechanism for async fetching of collectibles metadata for Activity entries, but that is left for a separate task.

### DIFF
--- a/services/wallet/collectibles/types.go
+++ b/services/wallet/collectibles/types.go
@@ -47,6 +47,17 @@ func idToCollectible(id thirdparty.CollectibleUniqueID) Collectible {
 	return ret
 }
 
+func idsToCollectibles(ids []thirdparty.CollectibleUniqueID) []Collectible {
+	res := make([]Collectible, 0, len(ids))
+
+	for _, id := range ids {
+		c := idToCollectible(id)
+		res = append(res, c)
+	}
+
+	return res
+}
+
 func fullCollectibleDataToHeader(c thirdparty.FullCollectibleData) Collectible {
 	ret := Collectible{
 		DataType: CollectibleDataTypeHeader,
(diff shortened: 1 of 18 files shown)


Cuteivist:
> We could use this new mechanism for async fetching of collectibles metadata for Activity entries, but that is left for a separate task.

I think it would be minor upgrade because right now activity skips collectibles metadata when filtering. After sending filter response we call asynchronously `FetchAssetsByCollectibleUniqueID`

services/wallet/collectibles/types.go
Cuteivist, line 1: The code complexity of functions here grew a lot with that change.
I think it would be good idea to add some unit tests for it.
What do you think?

Cuteivist (approved):
Looks great. Love those optimizations coming.

saledjenic (approved):

services/wallet/collectibles/types.go
dlipicar: will do

services/wallet/community/manager.go
```diff
@@ -53,36 +49,7 @@ func (cm *Manager) setCommunityInfo(id string, c *thirdparty.CommunityInfo) (err
 	return cm.db.SetCommunityInfo(id, c)
 }
 
-func (cm *Manager) mustFetchCommunityInfo(communityID string) bool {
```
dlipicar, line 56: This logic has now been added to the messenger side `func (m *Messenger) FetchCommunity`

PR #4456, opened 2023-12-13
chore(walletconnect)_: record wallet sessions instead of pairings

saledjenic:


services/wallet/walletconnect/database.go
```diff
@@ -39,65 +67,118 @@ func ChangePairingState(db *sql.DB, topic Topic, active bool) error {
...
+			&session.DappName,
+			&session.DappURL,
+			&session.DappDescription,
+			&session.DappIcon,
+			&session.DappVerifyURL,
+			&session.DappPublicKey)
 		if err != nil {
 			return nil, err
 		}
 		if err := rows.Err(); err != nil {
```
stefandunca: This check is related to iteration errors and, therefore should be left outside. Basically if there is an error iterating `rows.Next()` will return `false` and the error can be retrieved using `rows.Err()`

services/wallet/walletconnect/database.go
```diff
@@ -39,65 +67,118 @@ func ChangePairingState(db *sql.DB, topic Topic, active bool) error {
...
 		}
-		pairings = append(pairings, pairing)
-	}
-	if err := rows.Err(); err != nil {
-		return nil, err
+		sessions = append(sessions, session)
 	}
 
-	return pairings, nil
+	return sessions, rows.Err()
```
stefandunca: The way to go is to return `nil` in case of error instead of empty or partial results. This is covered by leaving the rows.Error check after the iteration and here you can kep `nil` for error.

services/wallet/walletconnect/database.go
```diff
@@ -39,65 +67,118 @@ func ChangePairingState(db *sql.DB, topic Topic, active bool) error {
...
+			&session.DappName,
+			&session.DappURL,
+			&session.DappDescription,
+			&session.DappIcon,
+			&session.DappVerifyURL,
+			&session.DappPublicKey)
+		if err != nil {
+			return nil, err
+		}
+		if err := rows.Err(); err != nil {
```
stefandunca: Same here, there will never be a chance to arrive here if an error occurred during iteration. Have to be moved outside.

stefandunca (approved):
Excellent. Please see my comments inline.

services/wallet/walletconnect/database.go
saledjenic: If I understand correctly you're saying that this is better:
```
    return nil, err
}
return sessions, nil
```
than this:
`return sessions, rows.Err()`

But isn't that the choice depends on your preference only cause I see those two the same in the end.

services/wallet/walletconnect/database.go
saledjenic: Not sure how this ended up here, I was updating old code so I either overlooked this or appeared after rebase. Thanks.

services/wallet/walletconnect/database.go
stefandunca: > If I understand correctly you're saying that this is better:

I separated the two things and it got confusing I see. The only important thing was to move the check for `rows.Err()` outside the loop.  Two steps approach is more idiomatic so that you don't provide a partial unusable result when an error is returned. However, in practice the one liner should work very well given that the first check after the call should be for the error.

PR #4458, opened 2023-12-13
fix: mailserver batches should be grouped by pubsub topics

richard-ramos:


protocol/messenger_mailserver.go
```diff
@@ -798,6 +807,8 @@ func (m *Messenger) SyncChatFromSyncedFrom(chatID string) (uint32, error) {
 			m.config.messengerSignalsHandler.HistoryRequestStarted(1)
 		}
 
+		fmt.Println("SYNC FROM", chatID, pubsubTopic, topics)
```
cammellos: debug

protocol/messenger_mailserver.go
```diff
@@ -937,61 +948,8 @@ func (m *Messenger) ConnectionChanged(state connection.State) {
 	}
 
 	if m.connectionState.Offline && !state.Offline {
-		err := m.sender.StartDatasync()
-		if err != nil {
-			m.logger.Error("failed to start datasync", zap.Error(err))
-		}
+		m.sender.StartDatasync()
```
cammellos: why removing the error checking, had no impact on the functioning but it would report an error?

protocol/messenger_mailserver.go
```diff
@@ -464,46 +469,48 @@ func (m *Messenger) syncFiltersFrom(filters []*transport.Filter, lastRequest uin
...
+					ChatIDs:     batch.ChatIDs,
+				}
 
-				// Replace og batch with new dates
-				batch.To = from
-				batches[k] = batch
-				tmpKeysToIterate = append(tmpKeysToIterate, k)
-			} else {
-				batches24h = append(batches24h, batch)
+				from := batch.To - 86400
```
cammellos: maybe we can const this number? I take is to allow some drift, but a variable/name explanation would be good

cammellos (approved):
looks ok, left some comments, we probably should have some tests for the splitting into batches logic, since it's quite involved, maybe not now but worth doing

protocol/messenger_mailserver.go
richard-ramos: Oops

protocol/messenger_mailserver.go
richard-ramos: It was removed by mistake

PR #4461, opened 2023-12-14
feat: Return seen/unseen messages count in MessengerResponse

MishkaRogachev:
For https://github.com/status-im/status-desktop/issues/12857

Add unread count and unread count with mentions to the messenger response

Important changes:
- [x] Change `MarkActivityCenterNotificationsRead`: add count and count with mentions for each affected chat in the response
- [x] Deprecate `MarkMessagesSeen`
- [x] Introduce `MarkMessagesRead` with MessengerResponse
- [x] Change `MarkMessageAsUnread` with MessengerResponse return type (not used on mobile currently)

### DIFF
--- a/protocol/messenger.go
+++ b/protocol/messenger.go
@@ -4143,7 +4143,7 @@ func (m *Messenger) DeleteMessagesByChatID(id string) error {
 	return m.persistence.DeleteMessagesByChatID(id)
 }
 
-func (m *Messenger) MarkMessageAsUnreadImpl(chatID string, messageID string) (uint64, uint64, error) {
+func (m *Messenger) markMessageAsUnreadImpl(chatID string, messageID string) (uint64, uint64, error) {
 	count, countWithMentions, err := m.persistence.MarkMessageAsUnread(chatID, messageID)
 
 	if err != nil {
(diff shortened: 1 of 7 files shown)


MishkaRogachev:
@qfrank @ilmotta in this PR i mark `MarkMessagesSeen` as deprecated, it's better to switch to 'MarkMessagesRead'. But the difference is only the response type

mprakhov (approved):

protocol/messenger_activity_center.go
```diff
@@ -182,12 +189,16 @@ func (m *Messenger) MarkActivityCenterNotificationsRead(ctx context.Context, ids
...
-	response := &MessengerResponse{}
 	notifications, err := m.persistence.MarkActivityCenterNotificationsUnread(ids, updatedAt)
 	if err != nil {
 		return nil, err
 	}
 
+	response := &MessengerResponse{}
+	response.AddActivityCenterNotifications(notifications)
+
+	// Don't mark messages unseen in chat, that looks weird
```
kounkou, line 200: Is this a leftover ? Not sure what it refers to.

protocol/messenger_response.go
```diff
@@ -812,3 +824,25 @@ func (r *MessengerResponse) GetUpdatedProfileShowcases() []*ProfileShowcase {
 	}
 	return showcases
 }
+
+func (r *MessengerResponse) AddSeveralSeenAndUnseenMessages(messages []*SeenUnseenMessages) {
```
kounkou, line 828: I am not sure what `adding` messages mean ?
Could you explain what are you trying to achieve please?

services/ext/api.go
```diff
@@ -797,6 +797,7 @@ func (api *PublicAPI) DeleteMessagesByChatID(id string) error {
 	return api.service.messenger.DeleteMessagesByChatID(id)
 }
 
+// Deprecated: Use MarkMessagesRead instead
```
kounkou, line 800: I believe you should use the deprecated pragma instead, because a comment is not enough to help future developers

Example : 

```
func (api *PublicAPI) MarkMessagesSeen(chatID string, ids []string) (*MarkMessageSeenResponse, error) {.deprecated: "use MarkMessagesRead instead".}
```

protocol/messenger.go
```diff
@@ -4202,6 +4209,7 @@ func (m *Messenger) markMessagesSeenImpl(chatID string, ids []string) (uint64, u
 	return count, countWithMentions, nil
 }
 
+// Deprecated: Use MarkMessagesRead instead
```
kounkou, line 4212: I believe you should use the deprecated pragma instead, because a comment is not enough to help future developers

Example :

```
func (m *Messenger) MarkMessagesSeen(chatID string, ids []string) (uint64, uint64, []*ActivityCenterNotification, error) {.deprecated: "use MarkMessagesRead instead".} 
```

kounkou (changes requested):
Left a few comments

protocol/messenger_activity_center.go
MishkaRogachev: This function is sort of the opposite of `MarkActivityCenterNotificationsRead` which also affects messages. I left comment here to accent we don't want to affect messages here

protocol/messenger_response.go
MishkaRogachev: This `MessengerResponse` we are using as a response type in all messenger api functions, so it's an aggregate for all possible Messenger's outputs

services/ext/api.go
MishkaRogachev: This is part of go, and it seems to me that `.deprecated` pragma is from nim.  In status-go project `// Deprecated:` comment its a common way to do it

kounkou: oh my bad, I didn't read well here 😄

kounkou (approved):

PR #4466, opened 2023-12-14
refactor(waku2): publish messages on a goroutine

### LINKED ISSUE #4462: bug: delay when using lightpush
# Problem
Observed the below while dogfooding status-desktop using lightpush. 

If a single message is sent , message is delivered almost instantly
When multiple messages are sent in succession, then a significant delay in message delivery is observed which keeps increasing with each mesage sent.

# Implementation

On debugging further noticed that in status-go, a single queue which sends a message via lightpush and waits for response from lightpush node before next message is sent.  This is causing the significant delay in delivery.
ref code: https://github.com/status-im/status-go/blob/b52a9ce0e5f72e62086f84a6a639a6de0bfd5c7e/wakuv2/waku.go#L1028-L1037 . 

This needs to be done asynchronously in order to avoid the delay observed while using lightpush. 

Quoting @richard-ramos  `I guess that code could be put into a goroutine. The important part is to call w.SendEnvelopeEvent(event) after the message is sent indicating if it was sent succesfully or not`

# Acceptance Criteria

When using lightMode, sending messages in succession should not have delay in delivering the same.

# Future Steps

Ensure all lightpush related tests pass. 
Need to ensure to test cases where there is a lot of network connectivity, messages to be sent are either sent successfully or marked for resend.

richard-ramos:
Small refactor for publishing messages doing it in a separate go routine to not delay message sending

Closes #4462

### DIFF
--- a/wakuv2/waku.go
+++ b/wakuv2/waku.go
@@ -1026,41 +1026,47 @@ func (w *Waku) broadcast() {
 	for {
 		select {
 		case envelope := <-w.sendQueue:
-			pubsubTopic := envelope.PubsubTopic()
-			var err error
-			logger := w.logger.With(zap.String("envelopeHash", hexutil.Encode(envelope.Hash())), zap.String("pubsubTopic", pubsubTopic), zap.String("contentTopic", envelope.Message().ContentTopic), zap.Int64("timestamp", envelope.Message().GetTimestamp()))
-			// For now only used in testing to simulate going offline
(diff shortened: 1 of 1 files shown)


plopezlpz (approved):

chaitanyaprem (approved):
LGTM

kaichaosun (approved):

vitvly (approved):

PR #4471, opened 2023-12-15
enable custom community store nodes

### LINKED ISSUE #4357: Communities should be able to define the list of store nodes
Message history retrieval works either by using a hardcoded list of store nodes from the status fleet, or using store nodes that were added manually via the advanced settings. Communities should be able to define the list of store nodes, either via a DNS discovery URL or specifying the list of nodes.

#### Scenario 1: _Get new storenode info from fetching community info:_
  1. Owner creates a community
  2. Owner sets the store node for the community
  3. Owner sends a message to the "general" community chat
  4. Bob joins the community and fetches the community info including the storenode info for that community
  5. Bob can see the community message fetched from the appropriate storenode

#### Scenario 2  _Get new storenode info from control message:_
  1. Owner creates a community
  2. Bob joins the community
  3. Owner sets the storenode for the community
  4. Bob gets the new storenode info from the control message
  5. Owner sends a message to the community chat
  6. Bob can see the community message fetched from the appropriate storenode 

#### Scenario 3: _Get new storenode info from fetching community info:_
  1. Owner creates a community
  2. Owner sets the store node for the community
  3. Owner sends a message to the "general" community chat
  4. Wait 30 days for the community description 
  5. Bob joins the community and fetches the community info including the storenode info for that community
  6. Bob can see the community message fetched from the appropriate storenode

plopezlpz:
A community owner can specify a store node for the community.

If the store node of a community is specified then all messages belonging to the community's chats will be requested to the specified store node.

## Current limitations
- It is assumed the store node is always connected: this should be improved to allow multiple store nodes per community and do a cycle similar to current store node cycle
- If a store node is setup for a community, we won't use any other store node for that community, maybe we want to fallback to the default store nodes.

Closes #4357


plopezlpz:
@richard-ramos I'm unsure if the two limitations in the description are acceptable

appdatabase/migrations/sql/1702022748_community_mailservers.up.sql
```diff
@@ -0,0 +1,13 @@
+-- To allow many to many relationship between communities and mailservers
+CREATE TABLE IF NOT EXISTS community_mailservers (
+    community_id BLOB NOT NULL,
+    mailserver_id VARCHAR NOT NULL,
+    PRIMARY KEY (community_id, mailserver_id),
+    FOREIGN KEY (community_id) REFERENCES communities_communities(id) ON DELETE CASCADE,
```
plopezlpz: This table is to allow many-to-many relationships between communities and mailservers, the thing is:
- `communities_communities` is migrated in `protocol/migrations` and
- `mailservers` is migrated in here `appdatabase/migrations`

I'm not sure if this table is safe here

appdatabase/migrations/sql/1702651464_community_mailservers.up.sql
```diff
@@ -0,0 +1,13 @@
...
+    mailserver_id VARCHAR NOT NULL,
+    PRIMARY KEY (community_id, mailserver_id),
+    FOREIGN KEY (community_id) REFERENCES communities_communities(id) ON DELETE CASCADE,
+    FOREIGN KEY (mailserver_id) REFERENCES mailservers(id) ON DELETE CASCADE
+) WITHOUT ROWID;
+
+CREATE INDEX community_mailservers_mailserver_id ON community_mailservers(mailserver_id);
+
+-- To avoid querying community-only mailservers
+ALTER TABLE mailservers ADD COLUMN community_only BOOLEAN DEFAULT FALSE;
```
richard-ramos: I'd use `storenodes` instead of `mailservers`, as mailserver is a concept associated to wakuv1 (eventually we'll have to rename everywhere we use mailservers to storenodes to be consistent)

protocol/messenger_communities.go
```diff
@@ -2043,6 +2044,57 @@ func (m *Messenger) SetCommunityShard(request *requests.SetCommunityShard) (*Mes
 	return response, nil
 }
 
+func (m *Messenger) SetCommunityMailServers(request *requests.SetCommunityMailServers) (*MessengerResponse, error) {
```
richard-ramos: Probably makes sense to sync up the community mailservers by calling `func (m *Messenger) syncCommunity`. That function will need to be modified to take into account the community storenodes

appdatabase/migrations/sql/1702651464_community_mailservers.up.sql
```diff
@@ -0,0 +1,13 @@
+-- To allow many to many relationship between communities and mailservers
+CREATE TABLE IF NOT EXISTS community_mailservers (
+    community_id BLOB NOT NULL,
+    mailserver_id VARCHAR NOT NULL,
```
richard-ramos: Since these can change overtime (with new storenodes added or removed), I wonder if it would make sense to add a `clock` and `removed` column to handle these type of changes? This would be similar to how bookmarks are managed in protocol/messenger_bookmarks.go

protocol/messenger_communities.go
```diff
@@ -2043,6 +2044,57 @@ func (m *Messenger) SetCommunityShard(request *requests.SetCommunityShard) (*Mes
...
+	}
+
+	err = m.mailserversDatabase.SaveMailserversForCommunity(request.CommunityID, request.MailServers)
+	if err != nil {
+		return nil, err
+	}
+	if err := m.communityStorenodes.ReloadFromDB(); err != nil {
+		return nil, err
+	}
+	response := &MessengerResponse{
```
richard-ramos, line 2088: Since we need to get these mailservers to the community members, we probably need a new message type that will be broadcasted when we save the storenodes

protocol/messenger_communities.go
```diff
@@ -2043,6 +2044,57 @@ func (m *Messenger) SetCommunityShard(request *requests.SetCommunityShard) (*Mes
...
+	community, err := m.communitiesManager.GetByID(id)
+	if err != nil {
+		return nil, err
+	}
+	if community == nil {
+		return nil, communities.ErrOrgNotFound
+	}
+	if !community.IsControlNode() {
+		return nil, errors.New("not admin or owner")
+	}
```
richard-ramos: All members of a community should be able to retrieve the community mailservers. This condition is probably not necessary

appdatabase/migrations/sql/1702651464_community_mailservers.up.sql
plaucorp: it is an existing table so I rather don't change the table name

plaucorp: `mailservers` is an already existing table, I rather leave that name

plopezlpz: mailservers is an already existing table, I rather leave that name

Issue #4438, opened 2023-12-08
Add syncing to peers for community messages


cammellos:
Given that there's some reliability issues, we decided to implement an e2e syncing mechanism between peers.
This is something that is already partially implemented for 1-to-1, and will be extended to work with communities.

chair28980:
@cammellos can you estimate a start and target date for this issue?

PR #4472, opened 2023-12-15
Surface acks, offers, requests to messenger

cammellos:
I need that information for one of the issues I am working on https://github.com/status-im/status-go/issues/4438 , and since it might be a large change, I thought I will commit this first to avoid conflicts.

I think it should also fix a potential issue where a message would be dropped if we were receiving both datasync messages & we would also be decrypting out of order messages, but not sure, I haven't had the time to replicate to ensure it was indeed a bug.
It should in theory not change anything this change.

### DIFF
--- a/protocol/common/message_sender.go
+++ b/protocol/common/message_sender.go
@@ -713,26 +713,30 @@ func (s *MessageSender) SendPublic(
 
 // unwrapDatasyncMessage tries to unwrap message as datasync one and in case of success
 // returns cloned messages with replaced payloads
-func unwrapDatasyncMessage(m *v1protocol.StatusMessage, datasync *datasync.DataSync) ([]*v1protocol.StatusMessage, [][]byte, error) {
-	var statusMessages []*v1protocol.StatusMessage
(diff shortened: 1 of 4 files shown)


protocol/common/message_sender.go
```diff
@@ -713,119 +713,117 @@ func (s *MessageSender) SendPublic(
...
 		}
 	}
 
-	return statusMessages, acks, nil
+	return response, nil
 }
 
-type handleMessageResponse struct {
-	Message          *v1protocol.StatusMessage
-	DatasyncMessages []*v1protocol.StatusMessage
```
osmaczko: I had these two separated to make the intention of one message to multiple messages transformation a bit clearer and to avoid `len(response.StatusMessages) > 0` and `StatusMessages[0]` on the client side. `func (s *MessageSender) HandleMessages(wakuMessage *types.Message)` is more complex and less readable to me now.

I see `HandleMessageResponse` has now become public and is used by messenger, so the merge to one `StatusMessages` array is justified.

Suggestion, not a blocker: maybe we could have two types, one private and one public.

osmaczko (approved):

richard-ramos (approved):

PR #4473, opened 2023-12-15
fix: return banned list for the clients

mprakhov:
Return the banned list for the clients

Closes #https://github.com/status-im/status-desktop/issues/12832

### DIFF
--- a/protocol/communities/community.go
+++ b/protocol/communities/community.go
@@ -1555,13 +1555,12 @@ func (o *Community) tokenPermissions() map[string]*CommunityTokenPermission {
 func (o *Community) PendingAndBannedMembers() map[string]CommunityMemberState {
 	result := make(map[string]CommunityMemberState)
 
-	// Non-privileged members should not see pending and banned members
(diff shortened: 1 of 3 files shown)


protocol/communities/community.go
```diff
@@ -1548,15 +1548,14 @@ func (o *Community) tokenPermissions() map[string]*CommunityTokenPermission {
 func (o *Community) PendingAndBannedMembers() map[string]CommunityMemberState {
 	result := make(map[string]CommunityMemberState)
 
-	// Non-privileged members should not see pending and banned members
```
mprakhov, line 1558: I return all updated data. If not to do it, on the client there will be unexpected behavior:
1. Before adding `PendingAndBannedMembers` fn, we returned `BanList` and didn't filter anything on go side 
2. A client on a startup has a ban list received by `ChannelGroup` struct 
3. Client must receive updates, when he's added to the ban list and when he's moved from the ban list (now we have a bug due to `PendingAndBannedMembers` always empty for non privileged members)

MishkaRogachev (approved):

PR #4474, opened 2023-12-15
fix (@status-go/stickers) Stickers still show in the recent list after being uninstalled

kounkou:
### Description

Stickers still show in the recent list after being uninstalled

fixes #12677
linked to https://github.com/status-im/status-desktop/pull/13066

Important changes:
- Updates the  private function `recentStickers` private function to filter the stickers based on if the sticker pack is installed or not. Before, the filter was not applied and returned a list that didn't take into account if the pack was installed, therefore causing a bug.


kounkou:
I updated a private function. Checking if adding a unittest is possible as there's no current infrastructure for unit testing for the entire file.

richard-ramos (approved):

igor-sirotin:
> Checking if adding a unittest is possible

A test would be good!

kounkou:
> > Checking if adding a unittest is possible
> 
> A test would be good!

Yes, I am not going to merge until I have a decent one. I wanted to put the PR out for now as a respect to people precious time. I am adding it by ~~EOD~~ Monday 👍🏾  (on my time LOL)

kounkou:
I believed this PR required special attention for testing, so I updated the code to be able to mock the functions `GetInstalledStickerPacks` and `GetRecentStickers` insisde the `func (api *API) recentStickers() ([]Sticker, error) {` function. 
While this took me a day or 2 to make sure there's no regression, I think the entire codebase should be refactored similarly to be able to easily enable capacity to mock the function, hence to create more testable code in the future.
On a another perspective, we should strive for TDD to help produce testable code.

-----

- Fixed codeClimate issues due to high number of function mocks
- Split the mocks into contextual mocks files

multiaccounts/settings/database_settings_manager.go
```diff
@@ -0,0 +1,77 @@
+package settings
```
kounkou, line 1: This interface should have been split to make the code more easily testable.
Code smell can be seen when seeing in go struct some 

```
*settings.Database
```
As this means we are forcing the developer/tester to pass an actual implementation of the Database. This is really sub-optimal because it can take some time to setup all the parameters required to run the application.
Instead, this PR proposal is to use an interface and be able to mock the critical function that are being called. 
With this PR we do not go level to call the database, but we mock the call hence we control the result from the low level API call.
As a result, we can avoid flaky tests, and have fast tests. 

To finish, it might be necessary to apply similar pattern for other portions of the code. But this is not in the scope of this PR

multiaccounts/accounts/database.go
```diff
@@ -283,7 +283,7 @@ func (a *Keypair) Operability() AccountOperable {
 
 // Database sql wrapper for operations with browser objects.
 type Database struct {
-	*settings.Database
+	settings.DatabaseSettingsManager
```
kounkou, line 286: In a separate task, we can make the Database take different dependencies like Reader, Writer,... for the Database to be designed better.
I leave this out of the scope of this ticket.

borismelnik (approved):
LGTM :)

PR #4477, opened 2023-12-15
Feat/use media server for community collectible images

### LINKED ISSUE #13022: Use media server for collectible images

dlipicar:
We used to transfer the whole base64-encoded image, which is normally very large. We now use the media server instead and transfer only a link.

Since the CommunityImage field is now much lighter, we now always include it (both Header and Details form)

This should make collectibles fetches much faster and prevent the nasty Nim json parse crash from happening.

Closes https://github.com/status-im/status-desktop/issues/13022

### DIFF
--- a/node/get_status_node.go
+++ b/node/get_status_node.go
@@ -207,7 +207,7 @@ func (n *StatusNode) StartMediaServerWithoutDB() error {
 		}
 	}
 
-	httpServer, err := server.NewMediaServer(nil, nil, n.multiaccountsDB)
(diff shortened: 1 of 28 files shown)


Cuteivist (approved):
Looks good. 
I didn't work with media server before. I might use it to pass community token images too.

stefandunca (approved):
Great improvement in terms of performance: lighter image data, qml can fetch them async so faster status-go requests and also qml cache should work very fine with this approach.

PR #4480, opened 2023-12-17
feat() add session based activity API, first steps for activity updates

stefandunca:
- [x] on top of https://github.com/status-im/status-go/pull/4658

This commit introduces the first steps towards implementing a session-based activity API to support dynamic updates of the current visualized filter in the wallet activity service. This change is necessary to move away from static paginated filtering, which was previously done in SQL, to a more dynamic approach that can handle updates in real-time.

The change is required by the new integration https://github.com/status-im/status-desktop/pull/13127 and doesn't affect the current API integration.

The main changes include:

- Added a `TODO.md` file outlining the plan and requirements for dynamic activity updates.
- New session-related API to the `activity.Service`
- `session.go` contains the logic for session management and event processing related to activity updates.
- Add test case for incremental filter updates.
    
The commit also includes:
    
- various other minor changes and refactoring to support the new session-based approach.
- Deprecation notices added to the `api.go` file for methods that are no longer used by the status-desktop application.
- Clarification comments added to the `scheduler.go` file regarding replacement policies.

Updates: [#12120](https://github.com/status-im/status-desktop/issues/12120)

### DIFF
--- a/services/wallet/activity/service.go
+++ b/services/wallet/activity/service.go
@@ -6,6 +6,8 @@ import (
 	"encoding/json"
 	"errors"
 	"strconv"
+	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
@@ -18,15 +20,19 @@ import (
 	"github.com/status-im/status-go/services/wallet/thirdparty"
 	"github.com/status-im/status-go/services/wallet/token"
 	"github.com/status-im/status-go/services/wallet/walletevent"
+	"github.com/status-im/status-go/transactions"
 )
 
 const (
-	// FilterResponse json is sent as a message in the EventActivityFilteringDone event
+	// EventActivityFilteringDone contains a FilterResponse payload
 	EventActivityFilteringDone          walletevent.EventType = "wallet-activity-filtering-done"
 	EventActivityFilteringUpdate        walletevent.EventType = "wallet-activity-filtering-entries-updated"
 	EventActivityGetRecipientsDone      walletevent.EventType = "wallet-activity-get-recipients-result"
 	EventActivityGetOldestTimestampDone walletevent.EventType = "wallet-activity-get-oldest-timestamp-result"
 	EventActivityGetCollectibles        walletevent.EventType = "wallet-activity-get-collectibles"
+
+	// EventActivitySessionUpdated contains a SessionUpdate payload
+	EventActivitySessionUpdated walletevent.EventType = "wallet-activity-session-updated"
 )
 
 var (
@@ -56,15 +62,33 @@ type Service struct {
 	eventFeed    *event.Feed
 
 	scheduler *async.MultiClientScheduler
+
+	sessions      map[SessionID]*Session
+	lastSessionID atomic.Int32
+	subscriptions event.Subscription
+	ch            chan walletevent.Event
+	// sessionsRWMutex is used to protect all sessions related members
+	sessionsRWMutex sync.RWMutex
+
+	// TODO #12120: sort out session dependencies
+	pendingTracker *transactions.PendingTxTracker
 }
 
-func NewService(db *sql.DB, tokenManager token.ManagerInterface, collectibles collectibles.ManagerInterface, eventFeed *event.Feed) *Service {
+func (s *Service) nextSessionID() SessionID {
+	return SessionID(s.lastSessionID.Add(1))
+}
+
+func NewService(db *sql.DB, tokenManager token.ManagerInterface, collectibles collectibles.ManagerInterface, eventFeed *event.Feed, pendingTracker *transactions.PendingTxTracker) *Service {
 	return &Service{
 		db:           db,
 		tokenManager: tokenManager,
 		collectibles: collectibles,
 		eventFeed:    eventFeed,
 		scheduler:    async.NewMultiClientScheduler(),
+
+		sessions: make(map[SessionID]*Session),
+
+		pendingTracker: pendingTracker,
 	}
 }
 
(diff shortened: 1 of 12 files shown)


services/wallet/activity/service.go
```diff
@@ -56,15 +62,35 @@ type Service struct {
...
+	// sessionsRWMutex is used to protect all sessions related members
+	sessionsRWMutex sync.RWMutex
+
+	// TODO #12120: sort out session dependencies
+	pendingTracker *transactions.PendingTxTracker
 }
 
-func NewService(db *sql.DB, tokenManager token.ManagerInterface, collectibles collectibles.ManagerInterface, eventFeed *event.Feed) *Service {
+func (s *Service) nextSessionID() SessionID {
+	s.lastSessionID.Add(1)
```
IvanBelyakoff: Add() returns a new value. `Load()` below is not needed

services/wallet/activity/service_test.go
```diff
@@ -57,18 +58,25 @@ func (m *mockTokenManager) LookupToken(chainID *uint64, tokenSymbol string) (tkn
...
 	require.NoError(tb, err)
 
 	eventFeed = new(event.Feed)
 	tokenMock = &mockTokenManager{}
 	collectiblesMock = &mockCollectiblesManager{}
-	service = NewService(db, tokenMock, collectiblesMock, eventFeed)
+
+	chainClient = transactions.NewMockChainClient()
+
+	pendingCheckInterval := time.Second
```
IvanBelyakoff, line 72: Why so long?

services/wallet/activity/session.go
```diff
@@ -0,0 +1,313 @@
...
+}
+
+// SessionUpdate payload for EventActivitySessionUpdated
+type SessionUpdate struct {
+	NewEntries []Entry         `json:"newEntries,omitempty"`
+	Removed    []EntryIdentity `json:"removed,omitempty"`
+	Updated    []Entry         `json:"updated,omitempty"`
+}
+
+type fullFilter struct {
```
IvanBelyakoff: this type is exactly the same as Session type but model. Maybe make fullFilter an embedded type of Session?

services/wallet/activity/session.go
```diff
@@ -0,0 +1,313 @@
...
+	// TODO #12120: prettyfy this
+	slices.SortFunc(addresses, func(a eth.Address, b eth.Address) bool {
+		return a.Hex() < b.Hex()
+	})
+	slices.Sort(chainIDs)
+	slices.SortFunc(filter.CounterpartyAddresses, func(a eth.Address, b eth.Address) bool {
+		return a.Hex() < b.Hex()
+	})
+
+	s.sessionsRWMutex.Lock()
```
IvanBelyakoff, line 100: `func (s* Service) createSession` ?
The mutex locking will be handled better as well

services/wallet/activity/session.go
```diff
@@ -0,0 +1,313 @@
...
+			session.model = make([]EntryIdentity, 0, len(entries))
+			for _, a := range entries {
+				session.model = append(session.model, EntryIdentity{
+					payloadType: a.payloadType,
+					transaction: a.transaction,
+					id:          a.id,
+				})
+			}
+		}
+		s.sessionsRWMutex.Unlock()
```
IvanBelyakoff: deferred in the beginning of the func?

services/wallet/activity/session.go
```diff
@@ -0,0 +1,313 @@
...
+
+	return sessionID
+}
+
+// TODO #12120: extend the session based API
+//func (s *Service) GetMoreForFilterSession(count int) {}
+
+// subscribeToEvents should be called with sessionsRWMutex locked for writing
+func (s *Service) subscribeToEvents() {
+	s.ch = make(chan walletevent.Event)
```
IvanBelyakoff: they say it should be buffered to avoid blocking other subscribers

services/wallet/activity/session.go
```diff
@@ -0,0 +1,313 @@
...
+		if event.Type == transactions.EventPendingTransactionUpdate {
+			var p transactions.PendingTxUpdatePayload
+			err := json.Unmarshal([]byte(event.Message), &p)
+			if err != nil {
+				log.Error("Error unmarshalling PendingTxUpdatePayload", "error", err)
+				continue
+			}
+
+			for id := range s.sessions {
+				s.sessionsRWMutex.RLock()
```
IvanBelyakoff, line 164: I'd use as much as possible a C++ RAII-like technique of holding mutexes, for example put RLock/defer RUnlock call inside checkFilterForPending and Lock/defer Unlock inside addOnTop(). I see that it is not a member, well if it makes sense, you can create a member and move mutex locking there and call non-member from there.
That would be more clear imho and avoid duplicate Unlocking at different branches of if/else

services/wallet/activity/session.go
```diff
@@ -0,0 +1,313 @@
...
+//func (s *Service) GetMoreForFilterSession(count int) {}
+
+// subscribeToEvents should be called with sessionsRWMutex locked for writing
+func (s *Service) subscribeToEvents() {
+	s.ch = make(chan walletevent.Event)
+	s.subscriptions = s.eventFeed.Subscribe(s.ch)
+	go s.processEvents()
+}
+
+func (s *Service) processEvents() {
```
IvanBelyakoff, line 153: Is this method meant to be looping forever until quit is called?

IvanBelyakoff (commented):
Stefan, good job summing it up on what should be done and a plan for it!
I have a more technical review rather than a strategical, and there is a lot to think about as you mentioned it.
Please address the inline issues

services/wallet/activity/service_test.go
stefandunca: Should be long enough that the second run is not executed given that it is not needed

services/wallet/activity/session.go
stefandunca: That could work. However, they are logically unrelated; `Session` is used to keep the session's state and `fullFilter` to provide a cleaner way to pass multiple parameters to `internalFilter` call. I renamed `fullFilter` to `fullFilterParams` to make it more verbose.

services/wallet/activity/session.go
stefandunca: The implementation does more than just create the session. It also executes the SQL query and returns the first page.

services/wallet/activity/session.go
stefandunca: Indeed, thanks for the heads up.

services/wallet/activity/session.go
stefandunca: This will work but "hide" the locking at lower function call levels. I try not to spread the mutexes at all call levels and keep it as high as possible. I plan to go away from locking in the future PRs.

services/wallet/activity/session.go
stefandunca: Yes, it will loop until `event.Unsubscribe` is called which is done in `StopFilterSession`

IvanBelyakoff (approved):

PR #4481, opened 2023-12-18
feat: add balance to collectibles api

### LINKED ISSUE #12956: Add balances to list of collectibles returned from backend
### Description

This is for now a backend-only feature, to be used by the mobile app.

They need to show a little icon to show which accounts owns the collectibles in the gallery
image

I've raised the issue of ERC1155 tokens, where the same collectible can be owned by multiple accounts and multiple units can be owned by an account. Example:

https://opensea.io/akki1994
image

Add a list of [address, balance] to each collectibles entry returned from the backend.

dlipicar:
Closes https://github.com/status-im/status-desktop/issues/12956

Add collectible balances to ownership cache, return list of owner address+balance alongside collectible data in the relevant APIs.

This is first going to be used in status-mobile.

We currently hardcode a balance of 1 to all collectibles, which is only valid for ERC721 tokens. Proper support for ERC1155 tokens (which requires fetching the balances from the providers) is left for a separate task: https://github.com/status-im/status-desktop/issues/13025

### DIFF
--- a/services/wallet/collectibles/commands.go
+++ b/services/wallet/collectibles/commands.go
@@ -3,6 +3,7 @@ package collectibles
 import (
 	"context"
 	"errors"
+	"math/big"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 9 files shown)


Cuteivist (approved):
Looks good

stefandunca (approved):
👍

PR #4482, opened 2023-12-18
feat: Save previously owned tokens

### LINKED ISSUE #12936: Preserve data on owned tokens
1) We show DAI, ETH and SNT with 0 balance from the start - new wallet
2) If a user buys or receives a token, we show that new token
3) If the user then sends all of that token, resulting in 0 balance, we keep showing that token - but the user has the option to either manually hide it or auto-hide it via setting a low balance threshold in manage tokens

Aim of this task is to update backend to preserve info about owned assets that aren't owned anymore.

Cuteivist:
closes https://github.com/status-im/status-desktop/issues/12936

* use token_balance to indicate previous ownership of the token, data from this table is not removed when token balance changes to 0
* add token entry if it is not to token_balance table for erc20, swap (2 tokens different tokens) and bridge (2 tokens - same token for 2 chains)

Screenshot containing the 0 tokens values (confirmed with etherscan balance checker):

image

### DIFF
--- a/services/wallet/reader.go
+++ b/services/wallet/reader.go
@@ -250,6 +250,11 @@ func (r *Reader) GetWalletToken(ctx context.Context, addresses []common.Address)
 		availableNetworks = append(availableNetworks, network)
 	}
 
+	cachedTokens, err := r.GetCachedWalletTokensWithoutMarketData()
+	if err != nil {
+		return nil, err
+	}
+
 	chainIDs := make([]uint64, 0)
 	for _, network := range availableNetworks {
 		chainIDs = append(chainIDs, network.ChainID)
@@ -296,7 +301,7 @@ func (r *Reader) GetWalletToken(ctx context.Context, addresses []common.Address)
 			for symbol, tokens := range getTokenBySymbols(tokenList) {
 				balancesPerChain := make(map[uint64]ChainBalance)
 				decimals := tokens[0].Decimals
-				anyPositiveBalance := false
+				isVisible := false
 				for _, token := range tokens {
 					hexBalance := balances[token.ChainID][address][token.Address]
 					balance := big.NewFloat(0.0)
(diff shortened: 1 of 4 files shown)


services/wallet/reader.go
```diff
@@ -250,6 +250,26 @@ func (r *Reader) GetWalletToken(ctx context.Context, addresses []common.Address)
 		availableNetworks = append(availableNetworks, network)
 	}
 
+	cachedTokens, err := r.GetCachedWalletTokensWithoutMarketData()
+	if err != nil {
+		return nil, err
+	}
+
+	isCachedToken := func(address common.Address, symbol string, chainID uint64) bool {
```
dlipicar: `GetWalletToken` is chunky enough already, perhaps move this definition to a separate named function instead of doing it inline?

services/wallet/token/token.go
```diff
@@ -289,6 +290,22 @@ func (tm *Manager) FindOrCreateTokenByAddress(ctx context.Context, chainID uint6
...
 
+func (tm *Manager) MarkAsPreviouslyOwnedToken(token *Token, owner common.Address) error {
+	if token == nil {
+		return errors.New("token is nil")
+	}
+	if (owner == common.Address{}) {
+		return errors.New("owner is nil")
+	}
+	count := 0
+	err := tm.db.QueryRow(`SELECT count(*) FROM token_balances WHERE user_address = ? AND token_address = ? AND chain_id = ?`, owner.Hex(), token.Address.Hex(), token.ChainID).Scan(&count)
```
dlipicar: premature optimization incoming: if you only need to know if there's more than 0 rows, there's a more efficient query:
https://stackoverflow.com/questions/9755860/valid-query-to-check-if-row-exists-in-sqlite3

dlipicar (approved):

services/wallet/reader.go
Cuteivist: Good idea. Move it to as function

services/wallet/token/token.go
Cuteivist: You are right! I need to remember that one for future.

PR #4484, opened 2023-12-18
chore: fetch contact with `StoreNodeRequestManager`

igor-sirotin:
## Description

I introduced a better way to fetch communities from store nodes here: https://github.com/status-im/status-go/pull/4446

This PR makes the `StoreNodeRequestManager` more universal for other needs, e.g. now we fetch contacts with the same approach and code.

## QA notes

I refactored the implementation of `RequestContactInfoFromMailserver` endpoint. The API remains the same, it just should work better now.

In status-desktop this endpoint is used to:
- when typed/pasted a valid public key to ad-hoc chat input
- when received a message with a contact link and contact is not found in local database

This feature is also implicitly used by status-go when generating a link preview for a contact that is not found in local database.

### DIFF
--- a/protocol/messenger_store_node_request_manager.go
+++ b/protocol/messenger_store_node_request_manager.go
@@ -20,17 +20,15 @@ const (
 	storeNodeAvailableTimeout = 30 * time.Second
 )
 
-// FetchStats is used in tests
-type FetchStats struct {
+// StoreNodeRequestStats is used in tests
+type StoreNodeRequestStats struct {
 	FetchedEnvelopesCount int
 	FetchedPagesCount     int
 }
 
-// WARNING: Is this needed?
-// This helps to fix collisions between communityID and contactID. Is this case actually supported by design?
 type storeNodeRequestID struct {
-	requestType storeNodeRequestType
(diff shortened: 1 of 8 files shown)


protocol/messenger_store_node_request_manager.go
```diff
@@ -185,7 +252,9 @@ type storeNodeRequest struct {
 
 	// request corresponding metadata to be used in finalize
 	filterToForget *transport.Filter
-	communityID    string
+
+	requestType storeNodeRequestType
+	typedID     string // FIXME: rename
```
mprakhov: fixme - is it ok?

mprakhov (approved):
LGTM 👍

protocol/messenger_store_node_request_manager.go
igor-sirotin: no, it's not 😄 
Will fix now, thank you!

igor-sirotin:
I'm facing a flaky test, but won't hold these changes.
It's now tracked separately here: https://github.com/status-im/status-go/issues/4495

Issue #3803, opened 2023-07-25
Paired device gets random response after first device makes a mutual contact


igor-sirotin:
# Problem

First reproduced here: https://github.com/status-im/status-go/pull/3799

When running `TestBlockedContactSyncing`, there're always different response on different iterations here:
https://github.com/status-im/status-go/blob/d9413755de0a6bfc65f3fc2c0c745503d439ef52/protocol/messenger_contact_requests_test.go#L1429

Possible messages:
1. `hello-1` (original CR text). ContentType: `CONTACT_REQUEST`
3. `Please add me to your contacts`, ContentType: `CONTACT_REQUEST`
2. `@Bob accepted your contact request`, ContentType: `SYSTEM_MESSAGE_MUTUAL_EVENT_ACCEPTED`

Depending on the messages, the response also contain one of (or both) of these AC notifications:
1. Type: 5, Message: `hello-1`
2. Type: 5, Message: `Please add me to your contacts`

There's some combination of messages received on 1 iteration. On iteration 2 and further another combination is received.

# Acceptance Criteria

The response is consistent.

igor-sirotin:
We spent a few hours with @MishkaRogachev trying to figure out what's happening. 

Basically, the difference comes from the order of receiving `hello-1` and `ContactUpdate`. 
But there're more things which influence this:
- AC Syncing
- System messages
- Generating not-found CR
- Updating CR mechanism
- AC notifications store Message copies in the database

We want to receiver resulting state to be the same, no matter what was the order of receiving messages.
But unfortunately with current CR implementation this means "implement a bunch of workarounds handling each corner case". 

Consider this example.

1. Setup Alice on 2 paired devices<br>2. Setup Bob<br>3. Alice-1 adds Bob as mutual friend (Bob accepts CR)<br>4. Alice-2 starts waiting for incoming message,<br>gathering a response for client app.
| Actions | Response if CR is received first | Response if ContactUpdate is received first |
|-|-|-|
|  | 1. Contact Sync<br>2. Message "hello-1"<br>5. System message<br>6. AC "hello-1" | 1. Contact Sync <br> 2. Message "Please add me to your contacts"<br>4. AC "Please add me to your contacts"<br><-- "hello-1" arrives -->7. Message(2) with updated text "hello-1"<br>8. AC(2) with updated message "hello-1"<br>9. AC "hello-1"

igor-sirotin:
We spent a few hours with @MishkaRogachev trying to figure out what's happening. 

[...]

PR #4488, opened 2023-12-19
Fix: Don't create extra mutual state messages when accepting the CR again

### LINKED ISSUE #13019: New issues with contact requests
# Bug Report
Screenshot 2023-12-15 at 14 00 12

image

## Description

I've recently noticed 2 new contact request issues, both reproduced with moderate frequency:

1) The receiver doesn't get the contact request , it looks like it gets lost in the network. But normal messages don't get lost, at least that often.
2) The sending party sometimes receives an additional message that the contact request is accepted, although the sending party does not perform any additional manipulations. May be related to https://github.com/status-im/status-go/issues/3803

## Steps to reproduce

1. Create a new account
2. Send a contact request to a user
3. Goto 1 until one of the problems is reproduced.

#### Expected behavior

1. Receiver gets a contact request
2. Sender sees 3 messages:

Screenshot 2023-12-15 at 14 09 41

#### Actual behavior

Sometimes happens (1) or (2) problem

MishkaRogachev:
For https://github.com/status-im/status-desktop/pull/13060

Important changes:
- [x] `handleAcceptContactRequestMessage` func does not create extra system messages, when it's accepted again.

Close https://github.com/status-im/status-desktop/issues/13019

### DIFF
--- a/protocol/messenger_contact_requests_test.go
+++ b/protocol/messenger_contact_requests_test.go
@@ -487,6 +487,32 @@ func (s *MessengerContactRequestSuite) TestReceiveAcceptAndRetractContactRequest
 }
 
 // The scenario tested is as follow:
+//  1. Repeat 5 times:
+//     2.1) Alice sends a contact request to Bob
+//     2.2) Bob accepts the contact request
+//     2.3) Alice removes bob from contacts
+func (s *MessengerContactRequestSuite) TestAcceptCRRemoveAndRepeat() {
+	theirMessenger := s.newMessenger()
+	_, err := theirMessenger.Start()
+	s.Require().NoError(err)
+	defer TearDownMessenger(&s.Suite, theirMessenger)
+
+	contactID := types.EncodeHex(crypto.FromECDSAPub(&theirMessenger.identity.PublicKey))
+
+	for i := 0; i < 5; i++ {
+		messageText := fmt.Sprintf("hello %d", i)
+		request := &requests.SendContactRequest{
+			ID:      contactID,
+			Message: messageText,
+		}
+		s.sendContactRequest(request, s.m)
+		contactRequest := s.receiveContactRequest(messageText, theirMessenger)
+		s.acceptContactRequest(contactRequest, s.m, theirMessenger)
+		s.retractContactRequest(contactID, theirMessenger)
+	}
+}
+
+// The scenario tested is as follow:
 // 1) Alice sends a contact request to Bob
 // 2) Bob declines the contact request
 // 3) Alice fails to send a new contact request to Bob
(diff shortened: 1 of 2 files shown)


protocol/messenger_contact_requests_test.go
```diff
@@ -539,7 +565,69 @@ func (s *MessengerContactRequestSuite) TestAliceTriesToSpamBobWithContactRequest
...
+	s.Require().NoError(err)
+
+	// Check we don't have extra messages on Alice's side
+	resp, err := WaitOnMessengerResponse(alice,
+		func(r *MessengerResponse) bool {
+			return len(r.ActivityCenterNotifications()) == 1 && len(r.Messages()) == 1
+		},
+		"contact request acceptance not received",
+	)
+	s.logResponse(resp, "acceptContactRequest")
```
cammellos: debug?

cammellos (approved):

protocol/messenger_contact_requests_test.go
```diff
@@ -1472,7 +1560,7 @@ func (s *MessengerContactRequestSuite) blockContactAndSync(alice1 *Messenger, al
 	s.Require().Equal(ContactRequestStateReceived, respContact.ContactRequestRemoteState)
 
 	// Check chats list
-	//s.Require().Len(alice2.Chats(), 1) // FIXME: Uncomment after https://github.com/status-im/status-go/issues/3800
+	s.Require().Len(alice2.Chats(), 2)
```
MishkaRogachev: @igor-sirotin, maybe it's time to remove https://github.com/status-im/status-go/blob/develop/protocol/messenger_contacts.go#L414-L415 ?

MishkaRogachev:
@igor-sirotin unfortunately this fix does not relate to https://github.com/status-im/status-go/issues/3803 and https://github.com/status-im/status-go/issues/3801 issues. It's a problem on the syncing side, as you mentioned [here](https://github.com/status-im/status-go/issues/3803#issuecomment-1654588500) should be resolved with having only one source of truth. And imo it's a bad time to refactor this part now

protocol/messenger_contact_requests_test.go
igor-sirotin: I think not yet. Maybe in a year or smth 😄

igor-sirotin: But it's better to use `AddChatsCount` in such cases.

https://github.com/status-im/status-go/blob/d535cd95f885a30387e0333f71c9a1081fc72a3a/deprecation/protocol.go#L8-L17

igor-sirotin (approved):

PR #4489, opened 2023-12-19
chore: read and use nonce from network instead from the local cache when sending tx

saledjenic:
Reading the Nonce from the local cache may be incorrect if the tx is made out of the Status app or if Status app sends a tx prepared by the dapp (via WalletConnect). A submitted tx with a wrong Nonce results in a failing tx, that's why we need to read the Nonce from the network.

Affected areas: wallet + other parts of the app that send a transaction


alaibe (approved):
Code LGTM, Just curious if we are not missing an edge case
@cammellos ?

dlipicar (approved):

saledjenic:
@cammellos I missed the PR and accidentally merged this instead of another PR, if you have any suggestions/comments let me know and I will apply them in a new PR.

PR #4494, opened 2023-12-20
Add debug level logs to make it easier for us to investigate message reliability 

### LINKED ISSUE #18206: Add more logging to PR builds to trace lost messages
## Problem
Currently, retrieving the message-id from the logs is not a straightforward process. As a result, there are no simple methods available to trace a specific message within the message history. 

This issue was highlighted during the message reliability meeting, emphasizing the need to streamline the tracing of a particular message using its unique id.

#### Acceptance criteria

To address this problem, the following steps will be taken:

1. Implement the necessary logging on the `DEBUG` level to facilitate easier debugging.
2. This enhanced logging functionality will simplify the process of tracing and debugging specific messages from the logs, i.e. from: [https://github.com/status-im/status-mobile/issues/18031](https://github.com/status-im/status-mobile/issues/18031) **only for e2e and PR builds**

By fulfilling these acceptance criteria, we aim to improve the efficiency and effectiveness of message tracing within the application.

siddarthkay:
When debugging message reliability we often get the number of messages sent and their IDs but we do not know the content of the messages and the type of message sent.

This PR adds debug level logs so that it helps in investigations.

ref : https://github.com/status-im/status-mobile/issues/18031

Closes [#18206](https://github.com/status-im/status-mobile/issues/18206)


protocol/messenger.go
```diff
@@ -2289,6 +2290,15 @@ func (m *Messenger) SendChatMessages(ctx context.Context, messages []*common.Mes
 
 // sendChatMessage takes a minimal message and sends it based on the corresponding chat
 func (m *Messenger) sendChatMessage(ctx context.Context, message *common.Message) (*MessengerResponse, error) {
+	m.logger.Debug("inside sendChatMessage",
```
cammellos: id, displayname,clock will all be blank at this stage, you want to log this at the bottom of the function

siddarthkay: thanks will amend. 
Do we want any other field that may be necessary?
I'm currently verifying this branch with 2 builds on my local environment

cammellos: that's enough, I believe

cammellos (approved):

MishkaRogachev (approved):

richard-ramos (approved):

igor-sirotin:
Logging message text sounds not quite private and secure to me, even for DEBUG logs.
Maybe it's fine to keep around temporarily, but I suppose we shouldn't at least keep it for release. 

If this is needed for e2e tests, then we should add some special flag for this, that would only be enabled in tests.
If this is needed for manual testing with app, then there're other ways to get `messageId`, e.g. in desktop (when `Settings->Advanced->Debug` is enabled) you can right-click the message and `Copy message id`:

image

cc @siddarthkay @cammellos

siddarthkay:
Thanks @igor-sirotin for pointing this out, we can further improve this by using build flags to ensure this kind of logging does not make it into release.
I have created this issue to track the work https://github.com/status-im/status-go/issues/4624

Issue #4230, opened 2023-10-30
Implement a mechanism to retrieve shard information for a given community ID


osmaczko:
As a result, we will need to develop a mechanism to retrieve shard information for a given community ID. One idea is to use a pubsub topic where the control node can publicly announce the shard information.

_Originally posted by @osmaczko in https://github.com/status-im/status-go/issues/3961#issuecomment-1785453710_

### QA Test Scenario Description
Scenario 1
1. 
2. 
Scenario 2
1. 
2.

jrainville:
Just to be sure, the Owner token Deployer mechanism to validate that someone is the rightful owner of a community can also be used without the community description right?

I'm asking, because we need to be sure that the info about the community' shard needs to be validated, so it will be signed, either by the community key or the owner's key. And since we won't have the community description yet, we need to be sure that just having the community ID is enough to validate an Owner.

osmaczko:
> Just to be sure, the Owner token Deployer mechanism to validate that someone is the rightful owner of a community can also be used without the community description right?

Yes, all we need is community id.

Currently, the trigger for owner validation is the `CommunityDescription` signed by an unverified user. This creates a chicken-and-egg problem: we can't validate shard information because we lack the signer's details, and we can't obtain the `CommunityDescription` since we don't have knowledge of the shard. However, this can be easily resolved by using `CommunityShardInfo` as an additional trigger for owner validation.

PR #4499, opened 2023-12-20
feat: implemented a mechanism to retrieve shard information for a given community ID

mprakhov:
Implemented a mechanism to retrieve shard information for a given community ID
We store in `storeNode` shard information under the content topic: `communityID + -"shard-info".` This data stored in `DefaultNonProtectedPubsubTopic` pubsubTopic

If a user tries to fetch a community, and Shard is nil, we will try to fetch shard info from the store node and use it during community fetch

Shard info published with a 5 min interval in `handleCommunitiesSubscription`

Impact:
Fetch community

Closes: 
- https://github.com/status-im/status-go/issues/4230
- https://github.com/status-im/status-desktop/issues/12585


protocol/messenger_community_shard.go
```diff
@@ -0,0 +1,130 @@
+package protocol
+
+import (
+	"context"
+	"crypto/ecdsa"
+	"errors"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/status-im/status-go/eth-node/crypto"
```
MishkaRogachev, line 11: not go-imported

protocol/communities/manager.go
```diff
@@ -3224,6 +3224,26 @@ func (m *Manager) GetByIDString(idString string) (*Community, error) {
 	return m.GetByID(id)
 }
 
+func (m *Manager) GetCommunityShard(communityID string) (*shard.Shard, error) {
```
MishkaRogachev: Can `GetCommunityShard` and `SaveCommunityShard` accept a communityId in the same format, such as `types.HexBytes`?

protocol/communities/persistence.go
```diff
@@ -1698,3 +1700,53 @@ func (p *Persistence) RemoveAllCommunityRequestsToJoinWithRevealedAddressesExcep
 		WHERE community_id = ? AND public_key != ?;`, communityID, pk, communityID, pk)
 	return err
 }
+
+func (p *Persistence) SaveCommunityShard(communityID types.HexBytes, shard *shard.Shard, clock uint64) error {
```
MishkaRogachev, line 1705: suggestion: i think `dbClock >= clock` should not be an error. The return type can be changed to `(bool, error)`

protocol/communities/persistence.go
```diff
@@ -1698,3 +1700,53 @@ func (p *Persistence) RemoveAllCommunityRequestsToJoinWithRevealedAddressesExcep
...
+	var dbClock uint64
+	// Fetch existing Community Shard
+	err := p.db.QueryRow(`SELECT clock FROM communities_shards WHERE community_id = ? `, communityID).Scan(&dbClock)
```
MishkaRogachev: suggestion: add select and insert queries in one transaction

protocol/messenger_community_shard.go
```diff
@@ -0,0 +1,130 @@
...
+	err = m.communitiesManager.SaveCommunityShard(publicShardInfo.CommunityId, shard.FromProtobuff(publicShardInfo.Shard), publicShardInfo.Clock)
+	if err != nil {
+		logError(err)
+	}
+	return err
+}
+
+func recoverCommunityShardInfoSignature(rawShardInfo *protobuf.CommunityPublicShardInfo) (*ecdsa.PublicKey, error) {
+	if rawShardInfo.Signature == nil || len(rawShardInfo.Signature) == 0 {
+		return nil, errors.New("missing signature")
```
MishkaRogachev: suggestion: declare errors globally

MishkaRogachev (commented):
Great job and this is one of the most complex PRs I've seen in a while! I like the idea of moving the sharding to a separate table. I'll run through it again to get better understanding and in the meantime I'll leave some minor questions

protocol/communities/manager.go
mprakhov: fixed

protocol/communities/persistence.go
mprakhov: no, this must be an error. Imagine, that we have a handle function, which store the shard info and you got 2 msgs... first with clock 11 and second with clock 7. Clock 7 must not rewrite more fresh data

protocol/communities/persistence.go
mprakhov: done

protocol/messenger_community_shard.go
mprakhov: low, also modified it to tell, that this is exactly shard info signature

protocol/communities/persistence.go
MishkaRogachev: I meant that message with `dbClock >= clock` should still be ignored, but no error raised because older message can arrive after fresh one and it's kind of normal

protocol/messenger_store_node_request_manager.go
```diff
@@ -66,17 +68,45 @@ func (m *StoreNodeRequestManager) FetchCommunity(community communities.Community
...
+		if err != nil {
+			return nil, StoreNodeRequestStats{}, fmt.Errorf("failed to create a shard info request: %w", err)
+		}
+
+		if !waitForResponse {
+			go func() {
+				shardResult := <-shard
+				communityShard = shardResult.shard
+
+				_, _, _ = requestCommunity(community.CommunityID, communityShard)
```
MishkaRogachev, line 99: not an issue: maybe it's possible to early return here?

MishkaRogachev (approved):
Great work!

protocol/communities/persistence.go
mprakhov: ok, understood, in the HandleCommunityShardInfo I'm not returning an error if error is an outdated clock
P.S. Added as a separate commit

protocol/messenger_store_node_request_manager.go
mprakhov: this is executed in a separate thread, so return does not make sense

protocol/common/shard/shard.go
```diff
@@ -54,3 +54,10 @@ const UndefinedShardValue = 0
 func DefaultShardPubsubTopic() string {
 	return wakuproto.NewStaticShardingPubsubTopic(MainStatusShardCluster, DefaultShardIndex).String()
 }
+
+func DefaultShard() *Shard {
+	return &Shard{
+		Cluster: MainStatusShardCluster,
+		Index:   NonProtectedShardIndex,
```
richard-ramos, line 61: Why is NonProtectedShardIndex used here instead of the DefaultShardIndex?

mprakhov: I saw that we put everything there (and my published shard info is public data), so decided to use this shard... can be changed. I didn't see we use DefaultShardIndex somewhere in the community.
Should I use DefaultShardIndex instead?

richard-ramos: The idea of the NonProtectedShardIndex is a shard used for users that have not joined a community to be able to send messages without having the DoS protection enabled. 
- ApplicationMetadataMessage_COMMUNITY_REQUEST_TO_JOIN
- ApplicationMetadataMessage_COMMUNITY_CANCEL_REQUEST_TO_JOIN
- ApplicationMetadataMessage_COMMUNITY_REQUEST_TO_JOIN_RESPONSE

Looking at code, it should be okay to keep it there for now, so no need to move it to the default shard. 
Probably in the future this type of control messages will be published into its own separate shard anyway.

PR #4503, opened 2023-12-21
fix: `fleet-shards.test.json`

richard-ramos:
Fixes some incorrect values being used in the default config for the shards.test fleet


igor-sirotin (approved):
Thanks!

PR #4505, opened 2023-12-21
chore: bump go-waku

richard-ramos:
Use last go-waku version from https://github.com/waku-org/go-waku/pull/857


plopezlpz (approved):

wakuv2/waku.go
```diff
@@ -292,6 +292,7 @@ func New(nodeKey string, fleet string, cfg *Config, logger *zap.Logger, appDB *s
 		node.WithMaxPeerConnections(cfg.DiscoveryLimit),
 		node.WithLogger(logger),
 		node.WithClusterID(cfg.ClusterID),
+		node.WithMaxMsgSize(1024 * 1024),
```
kaichaosun, line 295: Shouldn't this be `150*1024`? comply with this https://github.com/waku-org/go-waku/pull/963

chaitanyaprem: 150KB is set for The public Waku network.
Status is using its own static-sharded waku network with clusterID 16 and they don't want to change the message size limit and rather leave it at what libp2p supports by default which is 1MB.

chaitanyaprem (approved):
LGTM

kaichaosun (approved):

vitvly (approved):

PR #4509, opened 2023-12-22
`FetchCommunity`: various fixes and improvements

### LINKED ISSUE #4496: `FetchCommunity` doesn't get a new community description after adding permissions
1. This bug is now reproducible for @dlipicar and @endulab. 
2. @mprakhov was able to reproduce this with some tests.
3. I think I'm facing something similar with in https://github.com/status-im/status-go/issues/4495, but without permissions.

Let's gather the information here.

igor-sirotin:
Closes https://github.com/status-im/status-go/issues/4496
Closes https://github.com/status-im/status-go/issues/4496

## Main changes

1. Temporal fix for https://github.com/waku-org/nwaku/issues/2317 bug.
I made the manager to always fetch all envelopes in the batch (no stop when community was found).
These lines should be removed/reverted when the original bug is fixed:
https://github.com/status-im/status-go/blob/fdcdec371f37fffd6c7706cb14b26e826bfbe6df/protocol/messenger_store_node_request_manager_config.go#L24-L26
https://github.com/status-im/status-go/blob/fdcdec371f37fffd6c7706cb14b26e826bfbe6df/protocol/messenger_storenode_request_test.go#L420

2. Check if a desired community already exists in the database and require the fetched community `Clock` to be higher.
If no newer information was fethed, `nil` is returned, even if the community exists in the database

3. Implement `StoreNodeRequestManager` options:
- `WaitForResponse`
- `StopWhenDataFound`
- `InitialPageSize`
- `FurtherPageSize`
This allows a better control over the request. And made it possible to implement the fix (1) tidy.

~~4. It seems to me that the `TestRequestMultipleCommunities ` is now less flaky, but still fails at about 1% cases.~~
4. UPD: Fixes `TestRequestMultipleCommunities` flaky test

5. Extracted `DefaultMailserversByFleet` function

6. Some other minor refactoring and logging

## Test improvements

1. Added a mechanism to wait for the test store node to receive the envelopes.
This was previously leading to bugs, the case is tested with `TestSequentialUpdates`

2. Implemented a test to check the fetched envelopes order: `TestRequestCommunityEnvelopesOrder`
This is to address https://github.com/waku-org/nwaku/issues/2317 bug, to ensure that our code behaves correctly (assuming that go-waku is also working correctly).

3. Implemented a `TestFetchRealCommunity`
it allows to fetch community from all store nodes in a given fleet and check community state on each store node.

### DIFF
--- a/multiaccounts/settings/mocks/database_settings_manager_mock.go
+++ b/multiaccounts/settings/mocks/database_settings_manager_mock.go
@@ -125,8 +125,6 @@ func (mr *MockDatabaseSettingsManagerMockRecorder) CanUseMailservers() *gomock.C
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "CanUseMailservers", reflect.TypeOf((*MockDatabaseSettingsManager)(nil).CanUseMailservers))
 }
 
-
-
 // DeviceName mocks base method.
 func (m *MockDatabaseSettingsManager) DeviceName() (string, error) {
 	m.ctrl.T.Helper()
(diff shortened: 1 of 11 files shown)


mprakhov (approved):
LGTM

PR #4511, opened 2023-12-26
Correct profile showcase tokens and collectibles

MishkaRogachev:
For https://github.com/status-im/status-desktop/pull/13103

Important changes:
- [x] Separate profile showcase assets on verified and unverified tokens
- [x] Add `CommunityId` for profile showcase collectibles.
- [x] Split collectible UID on ContractAddress, ChainID and TokenID
- [x] Refactor profile showcase related tests
- [x] Fix identification for tokens and collectibles

Used by https://github.com/status-im/status-desktop/issues/13072 and https://github.com/status-im/status-desktop/issues/13073


protocol/protobuf/profile_showcase.proto
```diff
@@ -17,20 +17,33 @@ message ProfileShowcaseAccount {
 }
 
 message ProfileShowcaseCollectible {
-  string uid = 1;
+  string uid = 1 [deprecated = true];
```
kounkou, line 20: Just curious, why is it deprecated ?

kounkou (approved):

protocol/protobuf/profile_showcase.proto
MishkaRogachev: I originally used a synthetic identifier which is `chainID+contractAddress+TokenID`, but since the wallet has an api to get specific collections on contractAddress, tokenID, chainID and wallet account address, it is better to keep these fields separate. But yes, there is no sense in depreciation here, i will just remove it

protocol/migrations/sqlite/1706028033_profile_showcase_address_and_community.up.sql
```diff
@@ -0,0 +1,59 @@
+-- Recreate tables for storing current user profile showcase collectibles & assets preferences
+DROP TABLE profile_showcase_collectibles_preferences;
```
igor-sirotin, line 2: BRUTAL  💪

protocol/messenger_profile_showcase.go
```diff
@@ -104,26 +132,76 @@ func fromProfileShowcaseCollectibleProto(messages []*protobuf.ProfileShowcaseCol
...
+	for _, collectible := range preferences.Collectibles {
+		if account, ok := accountsMap[collectible.AccountAddress]; ok {
+			if account.ShowcaseVisibility < collectible.ShowcaseVisibility {
+				return errorAccountVisibilityLowerThanCollectible
+			}
+		} else {
+			return nil
+			// NOTE: with current wallet collectible implementation we don't know account on this stage
+			// return errorNoAccountAddressForCollectible
+		}
```
igor-sirotin: Nothing important, just more readable IMHO:
```suggestion
        account, ok := accountsMap[collectible.AccountAddress]
        if !ok {
            return nil
			// NOTE: with current wallet collectible implementation we don't know account on this stage
			// return errorNoAccountAddressForCollectible
	    }
	    if account.ShowcaseVisibility < collectible.ShowcaseVisibility {
		    return errorAccountVisibilityLowerThanCollectible
	    }
```

igor-sirotin (approved):
LGTM

Issue #4479, opened 2023-12-16
Order Non-Joined or Requested to Join Communities By last "Opened"


Samyoul:
Add the `lastOpenedAt` field and add a new API endpoint to call something like `CommunityOpened(communityID)` so that the UI can call it when a particular component is rendered.

This feature is needed so that "opened" / "spectated" (communities that a user is not a member of or has an active join request) communities are ordered by those that were opened last.

See https://github.com/status-im/status-mobile/issues/17337

PR #4514, opened 2023-12-31
Add last opened at key for community

ibrkhalil:
Add the last opened at key to communities to know when a community was last opened.
Required for [#17337](https://github.com/status-im/status-go/issues/4479) on status-mobile


protocol/communities_messenger_test.go
```diff
@@ -3904,3 +3904,74 @@ func (s *MessengerCommunitiesSuite) TestRequestAndCancelCommunityAdminOffline()
...
+	s.Require().Len(response.Chats(), 1)
+	// Check lastOpenedAt is zero as community wasn't opened yet
+	s.Require().True(response.Communities()[0].LastOpenedAt() == 0)
+
+	// Mock frontend triggering communityUpdateLastOpenedAt
+	response, err = s.alice.CommunityUpdateLastOpenedAt(community.IDString())
+	s.Require().NoError(err)
+
+	// Check lastOpenedAt was updated
+	s.Require().True(response.Communities()[0].LastOpenedAt() > 0)
```
Samyoul: Really nice to see a test, one suggestion is to:

- store the `response.Communities()[0].LastOpenedAt()` to a var
- add a `time.Sleep()` for 200 ms or so
- call `s.alice.CommunityUpdateLastOpenedAt(community.IDString())` again
- check that the new `LastOpenedAt` value is greater than the previous value

This check will catch any flakiness or weird behaviour with storing and retrieving timestamps (which can be a big problem sometimes).

protocol/communities/persistence_test_helpers.go
```diff
@@ -110,7 +113,7 @@ func (p *Persistence) saveRawCommunityRow(rawCommRow RawCommunityRow) error {
 
 func (p *Persistence) saveRawCommunityRowWithoutSyncedAt(rawCommRow RawCommunityRow) error {
 	_, err := p.db.Exec(
-		`INSERT INTO communities_communities ("id", "private_key", "description", "joined", "verified", "muted") VALUES (?, ?, ?, ?, ?, ?)`,
+		`INSERT INTO communities_communities ("id", "private_key", "description", "joined", "verified", "muted", "last_opened_at") VALUES (?, ?, ?, ?, ?, ?, ?)`,
```
Samyoul: The `rawCommRow` struct should have its `LastOpenedAt` property set. Otherwise you won't actually do anything with the value you've pulled from the DB.

protocol/communities_messenger_test.go
```diff
@@ -3904,3 +3904,85 @@ func (s *MessengerCommunitiesSuite) TestRequestAndCancelCommunityAdminOffline()
...
+	// Mock frontend triggering communityUpdateLastOpenedAt
+	response, err = s.alice.CommunityUpdateLastOpenedAt(community.IDString())
+	s.Require().NoError(err)
+
+	// Check lastOpenedAt was updated
+	lastOpenedAt1 := response.Communities()[0].LastOpenedAt()
+	s.Require().True(response.Communities()[0].LastOpenedAt() > 0)
+
+	// Nap for a bit
+	time.Sleep(time.Second)
```
ibrkhalil, line 3926: @Samyoul I tried 200ms and 500ms
200ms | failure rate 26/100 test runs
500ms | 1 failure out of 100 test run
1 second 100 test runs without failure.

protocol/messenger_communities.go
```diff
@@ -574,6 +574,21 @@ func (m *Messenger) JoinedCommunities() ([]*communities.Community, error) {
...
+	}
+	community.UpdateLastOpenedAt()
+	response := &MessengerResponse{}
+	response.AddCommunity(community)
+	return response, nil
+}
+
 func (m *Messenger) SpectatedCommunities() ([]*communities.Community, error) {
 	return m.communitiesManager.Spectated()
 }
```
Samyoul: Would it be possible to change this function so that it matches `SpectatedCommunities()`? The reason is any changes to the community should be handled by the communitiesManager, and not the Messenger

ibrkhalil: Done

protocol/communities/persistence_test_helpers.go
ibrkhalil: Done

protocol/communities_messenger_test.go
ibrkhalil: Done

protocol/communities/manager.go
```diff
@@ -680,10 +680,27 @@ func (m *Manager) Joined() ([]*Community, error) {
...
+func (m *Manager) UpdateLastOpenedAt(communityID types.HexBytes) error {
+	return m.persistence.UpdateLastOpenedAt(communityID)
+}
+
 func (m *Manager) Spectated() ([]*Community, error) {
 	return m.persistence.SpectatedCommunities(&m.identity.PublicKey)
 }
 
+func (m *Manager) CommunityUpdateLastOpenedAt(communityID string) (*Community, error) {
+	community, err := m.GetByIDString(communityID)
```
osmaczko: I believe we should also check if community is nil.

protocol/communities_messenger_test.go
```diff
@@ -3904,3 +3904,85 @@ func (s *MessengerCommunitiesSuite) TestRequestAndCancelCommunityAdminOffline()
...
+
+	// We join the org
+	response, err = s.alice.JoinCommunity(ctx, community.ID(), false)
+	s.Require().NoError(err)
+	s.Require().NotNil(response)
+	s.Require().Len(response.Communities(), 1)
+	s.Require().True(response.Communities()[0].Joined())
+	s.Require().Len(response.Chats(), 1)
+	// Check lastOpenedAt is zero as community wasn't opened yet
+	s.Require().True(response.Communities()[0].LastOpenedAt() == 0)
```
osmaczko:
```suggestion
	community, _ := s.createCommunity()
	s.advertiseCommunityTo(community, s.owner, s.alice)
	s.joinCommunity(community, s.owner, s.alice)
```

protocol/communities/manager.go
```diff
@@ -680,10 +680,27 @@ func (m *Manager) Joined() ([]*Community, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	err = m.UpdateLastOpenedAt(community.ID())
```
osmaczko: I believe we don't need extra public function `UpdateLastOpenedAt`, it is used only here.

```suggestion
	err = m.persistence.UpdateLastOpenedAt(communityID)
```

protocol/communities_messenger_test.go
osmaczko: That makes sense, given `time.Now().Unix()` returns time in seconds.

protocol/communities/manager.go
```diff
@@ -680,10 +680,27 @@ func (m *Manager) Joined() ([]*Community, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	community.UpdateLastOpenedAt()
```
igor-sirotin: Looks like there will be an inconsistency between database and status-go values.

You call `time.Now().Unix()` here:
https://github.com/status-im/status-go/blob/75cac42492b250a788f481f028dc5163483ca66b/protocol/communities/community.go#L974-L976

and here:
https://github.com/status-im/status-go/blob/75cac42492b250a788f481f028dc5163483ca66b/protocol/communities/persistence.go#L277-L280

Not a big deal, but better to have one source of truth.

protocol/communities/manager.go
```diff
@@ -744,6 +761,7 @@ func (m *Manager) CreateCommunity(request *requests.CreateCommunity, publish boo
 		MemberIdentity:       &m.identity.PublicKey,
 		CommunityDescription: description,
 		Shard:                nil,
+		LastOpenedAt:         time.Now().Unix(),
```
igor-sirotin: Maybe call `UpdateLastOpenedAt`? For the sake of clarity, where do we set this value.

protocol/communities/manager.go
```diff
@@ -1229,6 +1247,7 @@ func (m *Manager) ImportCommunity(key *ecdsa.PrivateKey, clock uint64) (*Communi
 			Joined:               true,
 			MemberIdentity:       &m.identity.PublicKey,
 			CommunityDescription: description,
+			LastOpenedAt:         time.Now().Unix(),
```
igor-sirotin: Same here as for `CreateCommunity`.

protocol/communities_messenger_test.go
Samyoul: @osmaczko : *proceeds to delete half the PR*. 😆 

It would be nice if we had some kind of document that pointed out these kind of boiler plate test funcs.

protocol/communities/manager.go
igor-sirotin: Hmm. Is this actually valid? does `CreateCommunity` automatically opens it from status-go side?
I think this is done from client side, then logically `LastOpenedAt` should be updated on that stage.

protocol/protobuf/pairing.proto
```diff
@@ -137,6 +137,7 @@ message SyncInstallationCommunity {
   bool spectated = 11;
   bytes encryption_keys = 12;
   SyncCommunityControlNode control_node = 13;
+  int64 last_opened_at = 14;
```
igor-sirotin: hmm. I don't see where so we set this field in the sync message 🤔

protocol/communities_messenger_test.go
osmaczko: I don't think documentation will suffice. I'm not fond of external code docs/hints because it becomes outdated very quickly. The ideal solution would be to use these helpers everywhere. Then, contributors will naturally follow the established pattern. Besides that, code reviews are a good opportunity to remind everyone about these helpers. Hope that makes sense :sweat_smile:

protocol/communities/manager.go
ibrkhalil: Done

protocol/communities/manager.go
ibrkhalil: Done

protocol/communities/manager.go
ibrkhalil: Replaced with 0

protocol/communities/manager.go
ibrkhalil: Replaced with 0

protocol/protobuf/pairing.proto
ibrkhalil: Would [this](https://github.com/status-im/status-go/pull/4514/commits/7276e361fcbeb5605afc6b6fc869a647307ddf8a#diff-12e1567f20b885c074b2a1213664d43a77fbc87c34b5dc42cdb6d4526a86460dR3026) work?

protocol/communities/manager.go
osmaczko: No need to address it, `community==nil` check is not needed anymore: https://github.com/status-im/status-go/pull/4544.

protocol/communities_messenger_test.go
```diff
@@ -3905,3 +3905,27 @@ func (s *MessengerCommunitiesSuite) TestRequestAndCancelCommunityAdminOffline()
...
+	// Nap for a bit
+	time.Sleep(time.Second)
+
+	// Check lastOpenedAt was successfully updated twice
+	response, err = s.alice.CommunityUpdateLastOpenedAt(community.IDString())
+	lastOpenedAt2 := response.Communities()[0].LastOpenedAt()
+	s.Require().NoError(err)
+
+	s.Require().True(lastOpenedAt2 > lastOpenedAt1)
+}
```
igor-sirotin, line 3933: Can you please add a test that a sync message arrives to another instance?

protocol/messenger_communities.go
```diff
@@ -576,6 +576,16 @@ func (m *Messenger) JoinedCommunities() ([]*communities.Community, error) {
...
 
+func (m *Messenger) CommunityUpdateLastOpenedAt(communityID string) (*MessengerResponse, error) {
+	updatedCommunity, err := m.communitiesManager.CommunityUpdateLastOpenedAt(communityID)
+	if err != nil {
+		return nil, err
+	}
+	response := &MessengerResponse{}
+	response.AddCommunity(updatedCommunity)
+	return response, nil
```
igor-sirotin: I believe we should call smth like `SyncCommunity` here. Otherwise no sync message is sent.

ibrkhalil: Done

protocol/communities_messenger_test.go
ibrkhalil: Done

igor-sirotin (approved):

PR #4515, opened 2024-01-02
feat: Community token received notification

Cuteivist:
Task https://github.com/status-im/status-desktop/issues/13086

Status-desktop change: https://github.com/status-im/status-desktop/pull/13116

* Added community token received notification
* Notification is send when token is discovered with new transaction
* Updated community image to use media server url instead of payload

### DIFF
--- a/protocol/activity_center.go
+++ b/protocol/activity_center.go
@@ -35,6 +35,7 @@ const (
 	ActivityCenterNotificationTypeSetSignerFailed
 	ActivityCenterNotificationTypeSetSignerDeclined
 	ActivityCenterNotificationTypeShareAccounts
+	ActivityCenterNotificationTypeCommunityTokenReceived
 )
 
 type ActivityCenterMembershipStatus int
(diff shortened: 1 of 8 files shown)


alaibe (approved):

alexjba:
@Cuteivist Can the token received event be sent if the community info is not ready?
This event will be processed in the UI and we're assuming we'll always have all the needed community info to display the community information.

Cuteivist:
@alexjba Notification from status-go will be sent for any token that has CommunityID. Status-go will try to retrieve (asynchronously) `name`, `color` and `image`, but in case it won't succeed  it will still send that user received token (with token data, e.g. symbol, name, balance in this specific tx) for community with ID. Then UI can decide whether display that info.

We assume that we will have all the info, but still it's possible there are corner cases when we won't have those.

NOTE That's about community tokens. I'm going through collectibles right now so I'm not sure how it will go there. I will adjust to work in similar manner.

PR #4516, opened 2024-01-03
fix (status-go/reply) images album disappears on restart

### LINKED ISSUE #12821: Retrieve images from Status-go to generate thumbnail images in reply
### Description

linked to https://github.com/status-im/status-go/pull/4516

From PR : https://github.com/status-im/status-desktop/pull/12806
Issue   : #10986 
 
> I think using only the messages received here might not generate a good map, because newMessagesLoaded only contains ~20 messages. In most cases, it should be ok, but if someone replied to an old message, then the album of messages wouldn't be part of the messages seq, so we'd also not find the images.

jrainville

In the code creates a map that solely relies on the received messages. The received message having a limit of ~20 messages, reply on an image for older messages, the code will not be able to obtain the images for that older message.
A solution would be to have access to the DB from status-go to fetch the corresponding message images and populate the Qml properties accordingly.

### Steps to reproduce

1- Find very old message with album images (more than 20 message earlier)
2- Reply to the message with album images

#### Expected behavior

- The old message images are not rendered in the reply thumbnails

#### Actual behavior

- The old message images should be rendered in the reply thumbnails

### Additional information

- https://github.com/status-im/status-desktop/issues/10986
- https://github.com/status-im/status-desktop/pull/12806

kounkou:
### Description

This PR (status-go) fixes [#12821](https://github.com/status-im/status-desktop/issues/12821)

linked to https://github.com/status-im/status-desktop/issues/12821

- Add album_images column to the user_messages table
- Migrate the database

### More details/decision made on the Pull Request :

This Pull Request is a continuation on [10986](https://github.com/status-im/status-desktop/issues/10986)

In the earlier version of the Status App, we employed a map associating albums with images to establish a connection between the received images and the corresponding message album. While this approach represented an improvement over the initial state of the application, where no mechanism facilitated the retrieval of quoted images, it faced limitations due to the finite number of messages received. Consequently, a quoted image might not be displayed if the associated message is too old.

The resolution presented in this Pull Request involves adding a field on the download side of the message. This field would part of fields for the quoted message, enabling their direct display in the view. It is essential to note that we intentionally avoid transmitting the images in the upload link to prevent imposing a significant processing and performance overhead on the application. Instead, within status-go, the quoted images are set at a specific stage and subsequently included in the message for the download step, facilitating their presentation in the view.

### Testing done

[Screencast from 2024-01-03 01:34:09 AM.webm](https://github.com/status-im/status-go/assets/2589171/a0063ed8-1288-4444-b1d3-e3bf33858d46)


kounkou:
The codeclimate issue happening on generate file protocol_message.pb.go, https://codeclimate.com/github/status-im/status-go/pull/4516. Action should be to remove generated files from codeclimate. Skipping as the check is also not required.

borismelnik (approved):
LGTM :)

protocol/protobuf/contact_verification.pb.go
```diff
@@ -1,245 +1,380 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
+// versions:
+// 	protoc-gen-go v1.27.1
+// 	protoc        v3.12.4
```
mprakhov: please, set up a correct version:
1. removed all my manual go-related installations (like protobuf and etc)
2. run make setup , make setup-dev
3. It will generate ~/go folder. Add path  ~/go/bin to PATH
4. Done, version should be the correct

kounkou: Okay thanks, let me perform this steps and resend a revision after testing. I understand the expectations should be that we don't have the protobuf files. Hopefully this decreases the number of files.

mprakhov: you will have some one, but not so much and no with these changes. At least you should not have this lines with the version

protocol/migrations/sqlite/1704489636_add_album_images.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE user_messages ADD COLUMN album_images TEXT DEFAULT "";
```
mprakhov, line 1: should it be stored as BLOB?

protocol/messenger.go
```diff
@@ -4040,24 +4051,68 @@ func (m *Messenger) MessageByChatID(chatID, cursor string, limit int) ([]*common
...
+			quotedImages = append(quotedImages, s.MakeImageURL(message.ID))
 		}
 	}
+	return quotedImages
 }
 
-func (m *Messenger) prepareMessage(msg *common.Message, s *server.MediaServer) {
+func (m *Messenger) prepareMessage(msg *common.Message, s *server.MediaServer) error {
 	if msg.QuotedMessage != nil && msg.QuotedMessage.ContentType == int64(protobuf.ChatMessage_IMAGE) {
 		msg.QuotedMessage.ImageLocalURL = s.MakeImageURL(msg.QuotedMessage.ID)
```
mprakhov, line 4091: can we not pass `s *server.MediaServer` and just verify it from this fn?
```
func (m *Messenger) prepareMessage(msg *common.Message, s *server.MediaServer) error {
    if m.httpServer == nil {
        return nil
    }
...
}
```

protocol/migrations/sqlite/1704489636_add_album_images.up.sql
kounkou: BLOB are best used for storing binary data which we are not storing here.

mprakhov: oh, so it is not an image content, right?

protocol/messenger.go
kounkou:
```go
func (m *Messenger) prepareMessage(msg *common.Message, s *server.MediaServer) error {
    if m.httpServer == nil {
        return nil
    }
...
```

Not really sure I understand the logic in that code snippet you provided. You meant instead ?

```go
return errors.New("...")
...
```

Furthermore, if the intent is to check if provided parameter is nil, my take on this issue is that the caller function should already have had that parameter as nil and should have checked that the parameter is nil.
The reason why I advocate for this is because, let's say there is bug in the function and this produces a bug. As developer, I want to fix the root cause, not the function down the road that processes the parameter. You can see the check example here :

```go
...
	if m.httpServer != nil {
		for idx := range msgs {
			err = m.prepareMessage(msgs[idx], m.httpServer)

			if err != nil {
				return nil, "", err
			}
		}
	}
...
```
https://github.com/status-im/status-go/blob/405d468e0e74e78a89f0653ad6514c50af9d1d6f/protocol/messenger.go#L4054

protocol/migrations/sqlite/1704489636_add_album_images.up.sql
kounkou: > oh, so it is not an image content, right?

It's kind of an image URL served. I chose that column name because it's better for understanding as I don't want to expose the internal implementation which may change in the future?

protocol/messenger.go
mprakhov: I mean, that we don't need to pass m.httpServer all over the way as m.httpServer is a member of the Messenger class

mprakhov (approved):

protocol/messenger.go
kounkou: Oh I see. Good catch 👍🏾 . I think that can be part of a refactoring I guess. I can open a ticket for this cause it'll change a bunch of files all over the places.

PR #4492, opened 2023-12-20
chore(wallet) add pending transaction test helpers

stefandunca:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #4480
* #4523
* __->__ #4492

Refactor the test helpers for pending transactions by extracting the mock implementations and helper functions into a new file `transactions/testhelpers.go`. Update event names for clarity and consistency in `pendingtxtracker.go`.

PR #4523, opened 2024-01-04
feat(wallet): propagate status with the pending tx status changed

stefandunca:
Replace usage of `eth_getTransactionByHash` with `eth_getTransactionReceipt`
when polling for changes. `eth_getTransactionReceipt` delivers also the
status of the transaction.

Update tests to account for the new changes
Propagate status
Refactor the `pendingtxtracker.go` file to emit notifications with a new payload structure that includes transaction identity and deletion status.
Includes: #4492

Closes status-desktop [#13124](https://github.com/status-im/status-desktop/issues/13124)

### DIFF
--- a/transactions/pendingtxtracker.go
+++ b/transactions/pendingtxtracker.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	eth "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/event"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/p2p"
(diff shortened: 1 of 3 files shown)


transactions/pendingtxtracker.go
```diff
@@ -24,12 +25,14 @@ import (
 )
 
 const (
-	// EventPendingTransactionUpdate is emitted when a pending transaction is updated (added or deleted). Carries StatusChangedPayload in message
+	// EventPendingTransactionUpdate is emitted when a pending transaction is updated (added or deleted). Carries PendingTxUpdatePayload in message
```
J-Son89, line 28: curious, does this change break the current api? i.e you will have to make updates to desktop for this to work with the current feature?

transactions/pendingtxtracker.go
```diff
@@ -139,6 +151,18 @@ func (tm *PendingTxTracker) fetchAndUpdateDB(ctx context.Context) bool {
 	return res
 }
 
+type nullableReceipt struct {
+	*types.Receipt
+}
+
+func (nr *nullableReceipt) UnmarshalJSON(data []byte) error {
+	if string(data) == "null" {
```
J-Son89: not sure of the best practices etc in this codebase/Go in general but I generally prefer to use named variables and let the code speak then commenting etc. consider something like:
```
const transactionNotAvailable = string(data) == "null"
if  transactionNotAvailable {
	return nil
}
```

transactions/pendingtxtracker.go
stefandunca: It doesn't. For now, we still use the deprecated `watch` API https://github.com/status-im/status-go/blob/405d468e0e74e78a89f0653ad6514c50af9d1d6f/services/wallet/api.go#L283

transactions/pendingtxtracker.go
stefandunca: Good idea, thanks for the heads up.

transactions/testhelpers.go
dlipicar, line 1: ❤️

dlipicar (approved):

PR #4524, opened 2024-01-04
feat: Add advanced token settings

Cuteivist:
Task https://github.com/status-im/status-desktop/issues/13118

Status-desktop change: https://github.com/status-im/status-desktop/pull/13128

Added settings for advanced token settings based on https://github.com/status-im/status-desktop/issues/12611


appdatabase/migrations/sql/1704281285_tokens_advanced_settings.up.sql
saledjenic, line 1: General idea, maybe we should start thinking of having a separate wallet_settings table, since this one (app settings) getting bigger and bigger day to day.

multiaccounts/settings/columns.go
```diff
@@ -402,6 +402,21 @@ var (
 		dBColumnName:   "wallet_token_preferences_group_by_community",
 		valueHandler:   BoolHandler,
 	}
+	ShowCommunityAssetWhenSendingTokens = SettingField{
```
saledjenic, line 405: @Cuteivist I guess we should add all 3 properties to sync protobuf, to have them synced among devices and when local pairing as well.

multiaccounts/settings/database.go
```diff
@@ -369,7 +369,8 @@ func (db *Database) GetSettings() (Settings, error) {
 		waku_bloom_filter_mode, webview_allow_permission_requests, current_user_status, send_status_updates, gif_recents,
 		gif_favorites, opensea_enabled, last_backup, backup_enabled, telemetry_server_url, auto_message_enabled, gif_api_key,
 		test_networks_enabled, mutual_contact_enabled, profile_migration_needed, is_sepolia_enabled, wallet_token_preferences_group_by_community, url_unfurling_mode,
-                omit_transfers_history_scan, mnemonic_was_not_shown
+		omit_transfers_history_scan, mnemonic_was_not_shown, wallet_show_community_asset_when_sending_tokens, wallet_display_assets_below_balance,
```
saledjenic, line 380: I see that the default values are set in the SQL script for these fields, but pretty sure, it's a good thing to have them added to the `CreateSettings` function as well, since that lets the status-go's client set them based on its need when initializing the app.

multiaccounts/settings/database.go
```diff
@@ -744,10 +748,46 @@ func (db *Database) GetTokenGroupByCommunity() (result bool, err error) {
 	return result, err
 }
 
+func (db *Database) GetShowCommunityAssetWhenSendingTokens() (result bool, err error) {
```
saledjenic: @Cuteivist at this moment no need for this and other functions which added below within this PR. Do we need/use them anyhow?

appdatabase/migrations/sql/1704281285_tokens_advanced_settings.up.sql
Cuteivist: That's really good idea, but it is itself bigger task because all data passing and sync should be handled properly on wallet to. Feel free to create issue for that one!

multiaccounts/settings/columns.go
Cuteivist: That's right, Added.

multiaccounts/settings/database.go
Cuteivist: Good point. Added there and default values

multiaccounts/settings/database.go
Cuteivist: Removed

appdatabase/migrations/sql/1704281285_tokens_advanced_settings.up.sql
saledjenic: @Cuteivist  issue created:
https://github.com/status-im/status-go/issues/4525

saledjenic (approved):

multiaccounts/settings/structs.go
```diff
@@ -178,29 +178,32 @@ type Settings struct {
...
+	BackupEnabled                       bool                          `json:"backup-enabled?,omitempty"`
+	AutoMessageEnabled                  bool                          `json:"auto-message-enabled?,omitempty"`
+	GifAPIKey                           string                        `json:"gifs/api-key"`
+	TestNetworksEnabled                 bool                          `json:"test-networks-enabled?,omitempty"`
+	ProfileMigrationNeeded              bool                          `json:"profile-migration-needed,omitempty"`
+	IsSepoliaEnabled                    bool                          `json:"is-sepolia-enabled?,omitempty"`
+	TokenGroupByCommunity               bool                          `json:"token-group-by-community?,omitempty"`
+	ShowCommunityAssetWhenSendingTokens bool                          `json:"show-community-asset-when-sending-tokens?,omitempty"`
+	DisplayAssetsBelowBalance           bool                          `json:"display-assets-below-balance?,omitempty"`
+	DisplayAssetsBelowBalanceThreshold  int64                         `json:"display-assets-below-balance-threshold,omitempty"`
```
dlipicar, line 205: shouldn't this be a `uint64`?
(plus the corresponding change everywhere else)

Cuteivist: @dlipicar Actually sqlite unsigned big int is just int64, so in the end uint wouldn't help here anything, just require additional handling in syncing.

dlipicar: Alright! Not gonna argue too strongly about this, I just feel that this parameter, due to it's meaning, should be represented by an unsigned variable in as many places as possible, at least to get some compiler warning if anybody wants to do something funny like setting it to a negative number. The fact that SQLite uses a signed type to store it should be taken into account, but not necessarily propagate across the rest of the code.

dlipicar (approved):

PR #4531, opened 2024-01-05
feat: use automatic peer selection for filter.

### LINKED ISSUE #4463: feat: Improve filter peer management based on shards
# Problem

While going through filter code in status-go, i had noticed that the way filter subscriptions are done towards peers does not take sharding into consideration. 

Referring to below code where filter peers are found, few observations that can be improved are

1. When looking for filter peers, not sure why relay protocol ID is also checked for. If a peer supports filter subscribe, that should be sufficient right?
2. The peer lookup doesn't consider shards supported by a peer. Which means it is possible that the selected peer does not support the required shard and hence filter subscribe might fail. This can be addressed in 2 ways, either use Automatic peer selection provided by go-waku by passing flag WithAutomaticPeerSelection, or add logic that would also consider shards and bucket peers as per shards. 

https://github.com/status-im/status-go/blob/42cf9fa7402923a9db8781f3a14d24aa7ddd5a6e/wakuv2/filter_manager.go#L303-L308

Note that point-2 may not impact as of now, because the fleet nodes support all shards. But as communities are created and assigned to different shards where various peers only support a subset of them, this will start causing issues.

# Implementation

Explained approaches on how this can be implemented above.

# Acceptance Criteria

- [ ] Filter peers are to be selected including the shards into consideration

kaichaosun:
Use automatic peer selection from waku helps reduce duplicate logic and increase performance when static shards apply.

Important changes:
- [x] use filter option `WithAutomaticPeerSelection` when subscribe

Closes https://github.com/status-im/status-go/issues/4463

### DIFF
--- a/wakuv2/filter_manager.go
+++ b/wakuv2/filter_manager.go
@@ -2,9 +2,6 @@ package wakuv2
 
 import (
 	"context"
-	"crypto/rand"
-	"errors"
-	"math/big"
 	"sync"
 	"time"
 
@@ -200,7 +197,7 @@ func (mgr *FilterManager) processEvents(ev *FilterEvent) {
 	}
 }
 
-func (mgr *FilterManager) subscribeToFilter(filterID string, peer peer.ID, tempID string) {
+func (mgr *FilterManager) subscribeToFilter(filterID string, tempID string) {
 
 	logger := mgr.logger.With(zap.String("filterId", filterID))
 	f := mgr.getFilter(filterID)
@@ -210,16 +207,16 @@ func (mgr *FilterManager) subscribeToFilter(filterID string, peer peer.ID, tempI
 		return
 	}
 	contentFilter := mgr.buildContentFilter(f.PubsubTopic, f.ContentTopics)
-	logger.Debug("filter subscribe to filter node", zap.Stringer("peer", peer), zap.String("pubsubTopic", contentFilter.PubsubTopic), zap.Strings("contentTopics", contentFilter.ContentTopicsList()))
+	logger.Debug("filter subscribe to filter node", zap.String("pubsubTopic", contentFilter.PubsubTopic), zap.Strings("contentTopics", contentFilter.ContentTopicsList()))
 	ctx, cancel := context.WithTimeout(mgr.ctx, requestTimeout)
 	defer cancel()
 
-	subDetails, err := mgr.node.FilterLightnode().Subscribe(ctx, contentFilter, filter.WithPeer(peer))
+	subDetails, err := mgr.node.FilterLightnode().Subscribe(ctx, contentFilter, filter.WithAutomaticPeerSelection(mgr.peers...))
 	var sub *subscription.SubscriptionDetails
 	if err != nil {
-		logger.Warn("filter could not add wakuv2 filter for peer", zap.Stringer("peer", peer), zap.Error(err))
+		logger.Warn("filter could not add wakuv2 filter for peers", zap.Any("peers", mgr.peers), zap.Error(err))
 	} else {
-		logger.Debug("filter subscription success", zap.Stringer("peer", peer), zap.String("pubsubTopic", contentFilter.PubsubTopic), zap.Strings("contentTopics", contentFilter.ContentTopicsList()))
+		logger.Debug("filter subscription success", zap.Any("peers", mgr.peers), zap.String("pubsubTopic", contentFilter.PubsubTopic), zap.Strings("contentTopics", contentFilter.ContentTopicsList()))
 		sub = subDetails[0]
 	}
 
@@ -320,14 +317,6 @@ func (mgr *FilterManager) findFilterPeers() []peer.ID {
 	return peers
 }
 
-func (mgr *FilterManager) findPeerCandidate() (peer.ID, error) {
-	if len(mgr.peers) == 0 {
-		return "", errors.New("filter could not select a suitable peer")
-	}
-	n, _ := rand.Int(rand.Reader, big.NewInt(int64(len(mgr.peers))))
-	return mgr.peers[n.Int64()], nil
-}
-
 func (mgr *FilterManager) resubscribe(filterID string) {
 	subs, found := mgr.filterSubs[filterID]
 	if !found {
@@ -344,16 +333,11 @@ func (mgr *FilterManager) resubscribe(filterID string) {
 	mgr.logger.Debug("filter resubscribe subs count:", zap.String("filterId", filterID), zap.Int("len", len(subs)))
 	for i := len(subs); i < mgr.settings.MinPeersForFilter; i++ {
 		mgr.logger.Debug("filter check not passed, try subscribing to peers", zap.String("filterId", filterID))
-		peer, err := mgr.findPeerCandidate()
 
-		if err == nil {
-			// Create sub placeholder in order to avoid potentially too many subs
-			tempID := uuid.NewString()
-			subs[tempID] = nil
-			go mgr.subscribeToFilter(filterID, peer, tempID)
-		} else {
-			mgr.logger.Error("filter resubscribe findPeer error", zap.Error(err))
-		}
+		// Create sub placeholder in order to avoid potentially too many subs
+		tempID := uuid.NewString()
+		subs[tempID] = nil
+		go mgr.subscribeToFilter(filterID, tempID)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


wakuv2/filter_manager.go
```diff
@@ -210,16 +207,16 @@ func (mgr *FilterManager) subscribeToFilter(filterID string, peer peer.ID, tempI
...
 
-	subDetails, err := mgr.node.FilterLightnode().Subscribe(ctx, contentFilter, filter.WithPeer(peer))
+	subDetails, err := mgr.node.FilterLightnode().Subscribe(ctx, contentFilter, filter.WithAutomaticPeerSelection(mgr.peers...))
 	var sub *subscription.SubscriptionDetails
 	if err != nil {
-		logger.Warn("filter could not add wakuv2 filter for peer", zap.Stringer("peer", peer), zap.Error(err))
+		logger.Warn("filter could not add wakuv2 filter for peers", zap.Any("peers", mgr.peers), zap.Error(err))
 	} else {
-		logger.Debug("filter subscription success", zap.Stringer("peer", peer), zap.String("pubsubTopic", contentFilter.PubsubTopic), zap.Strings("contentTopics", contentFilter.ContentTopicsList()))
+		logger.Debug("filter subscription success", zap.Any("peers", mgr.peers), zap.String("pubsubTopic", contentFilter.PubsubTopic), zap.Strings("contentTopics", contentFilter.ContentTopicsList()))
```
chaitanyaprem: Rather than mgr.peers, better to print the actual peerID with which filter subscription was successful. You can get that from `subDetails.PeerID`

chaitanyaprem: Also where is mgr.peers getting populated? We need to see if that is being used somewhere else and if it is being directly taken from node peer-store then it maybe inconsistent with actual peers with which filter subs are maintained. 

I think we may have to do detailed analysis of how filter state is maintained in status-go and update things accordingly. I am not sure changing peer selection to automatic would just be this much.

PingPeers is done based on peerIDs that are part of subscriptions, which is fine and not impacted by this change. 

But unsubscribe and any othre places where mgr.Peers is used need to taken a look at., because mgr.filterPeers uses peerStore to find filterPeers. It is possible that there maybe filterPeers in the peerStore but not subscribed to. We need to ensure Filter operations like unsubscribe etc are not sent to peers to which subscriptions are not present.

I am not sure if there is some other peer reference that needs to be looked at.

vitvly:
@kaichaosun @chaitanyaprem been trying to recall why exactly `WithPeer` was used in `FilterManager`, instead of `WithAutomaticPeerSelection`. Originally implemented here: https://github.com/status-im/status-go/pull/4020.

I think that it was planned to improve `FilterManager.findPeerCandidate` function, so that it would take distribute distinct filter subs among different peers better. Right now it's random, similarly to what `WithAutomaticPeerSelection` does. This results in `FilterManager.resubscribe` being not too smart when trying to establish `MinPeersForFilter` connections for any particular filter, as it might select the same peer twice, i think.

So maybe another option to try is improving `findPeerCandidate`, so that:
  - it does not pick the same peer for the same filter twice
  - it sorts `FilterManager.peers` somehow by priority. Before https://github.com/status-im/status-go/pull/4020 there used to be a `Waku.filterPeerDisconnectMap`, which had put recently disconnected peers to the end of `peers` list. Definitely there can be a much better logic for peer sorting. Apologies if there is already one in `go-waku` that i'm not aware of.

Alternatively, go-waku's automatic peer selection could be modified in order to take into account the above two points.

wakuv2/filter_manager.go
vitvly: Chiming in here - unsubscribes are performed via `UnsubscribeWithSubscription`, this should ensure that we're using a particular peer that already is being subscribed to.
`FilterManager.peers` is populated on a scheduled basis, using libp2p's PeerStore. If some peer pings fail in the time interval between full peer re-population, these disconnected peers are removed from `FilterManager.peers` (possibly to be added back during full re-population)

kaichaosun:
The `mgr.peers` is maintained as the candidates for connecting to a filter service node, both the connected and disconnected peers should be removed from the candidates. It originally populated from peer store with protocol support Filter and relay.
~`mgr.subscribeToFilter` is now changed to not use goroutine to avoid subscribe to the same peer.~

For sorting FilterManager.peers by priority, this is another story, and I see it can be used not only by filter, should we design and spec it first before implement it? Go-waku already provides peer selection based on delay, I believe there could be other dimensions of `priority` besides delay. (use delay based selection probably not a good idea when we are not sure about the geolocation of filter services and application users.)

@vitvly @chaitanyaprem

wakuv2/filter_manager.go
chaitanyaprem: > Chiming in here - unsubscribes are performed via `UnsubscribeWithSubscription`, this should ensure that we're using a particular peer that already is being subscribed to.

In that case we don't have to worry about unsubscribes.

chaitanyaprem: > `FilterManager.peers` is populated on a scheduled basis, using libp2p's PeerStore. If some peer pings fail in the time interval between full peer re-population, these disconnected peers are removed from `FilterManager.peers` (possibly to be added back during full re-population)

If we are relying on peer-selection from go-waku, then rather than looking at peers from peerStore, the peers selected for each filter can be stored in mgr.peers and pings can be sent to them accordingly.

kaichaosun: Sounds good to me, how do you think about this solution @vitvly ?

vitvly: Agree with this option - with automatic peer selection we don't need to refer to peer store. In that case we don't even need `mgr.peers` i think - it ceases to be a pool of possible peer candidates and instead becomes a duplicate store of peer IDs, given that we can fetch all current peer IDs from active subscriptions anyway.

chaitanyaprem: +1 to this, but how will Filter ping be issued. AFAIU, right now it is being issued based on peers listed in mgr.peers. 
Otherwise, the subscriptions can be looped and aggregated peer-wise to issue the pings.

wakuv2/filter_manager.go
```diff
@@ -341,19 +326,20 @@ func (mgr *FilterManager) resubscribe(filterID string) {
...
-			subs[tempID] = nil
-			go mgr.subscribeToFilter(filterID, peer, tempID)
-		} else {
-			mgr.logger.Error("filter resubscribe findPeer error", zap.Error(err))
-		}
+
+		// Create sub placeholder in order to avoid potentially too many subs
+		tempID := uuid.NewString()
+		subs[tempID] = nil
+		mgr.subscribeToFilter(filterID, tempID)
```
vitvly: Can you please explain a bit more on why removing goroutine helps with avoiding subscribing to the same peer? Having all calls to go-waku's `FilterLightnode` asynchronous is important for the functionality of FilterManager, as far as i recall - some of these subscribes/unsubscribes might take a long time.

wakuv2/filter_manager.go
kaichaosun: Fixed.

chaitanyaprem:
> For sorting FilterManager.peers by priority, this is another story, and I see it can be used not only by filter, should we design and spec it first before implement it? Go-waku already provides peer selection based on delay, I believe there could be other dimensions of `priority` besides delay. (use delay based selection probably not a good idea when we are not sure about the geolocation of filter services and application users.)

You are right. This is something that will have to be done eventually(but for now it makes sense to keep it simple since we are dealing with only fleet nodes which are not of large number). 
This would require more thought as you had mentioned. There is already a similar feature planned on go-waku https://github.com/waku-org/go-waku/issues/771 which will be taken up sometime this year based on the need. 

For now, to keep it simple we can use RTT probably for this.

kaichaosun:
> For now, to keep it simple we can use RTT probably for this.

Do you mean use `SelectPeerWithLowestRTT` for filter peer selection? This will probably add too many loads to the filter service nodes if most of the users are in same region (like Europe). @chaitanyaprem

chaitanyaprem:
> > For now, to keep it simple we can use RTT probably for this.
> 
> Do you mean use `SelectPeerWithLowestRTT` for filter peer selection? This will probably add too many loads to the filter service nodes if most of the users are in same region (like Europe). @chaitanyaprem

This would work better in a more decentralized operator environment where there are more service nodes which i think would be how the status communities would function in the long run.
But you are right that if we start using this with current fleet nodes only, this won't scale well.  This can be taken up at a later stage on the RTT logic is complete and we have some data showing how much traffic it adds.

vitvly (approved):

wakuv2/filter_manager.go
kaichaosun: Added back goroutine and depend on go-waku for dedupliate subscribing to same peer.

Issue #4437, opened 2023-12-08
Deprecate `protobuf.ChatMessage_Community`


ibrkhalil:
# Problem

Deprecate `protobuf.ChatMessage_Community`

According to new request to change message invitation structure, We now don't send a signed community description from admin and we just send a URL to the community.

`protobuf.ChatMessage_Community` is no longer needed.

igor-sirotin:
Some places I wasn't sure I'm doing it right:

https://github.com/status-im/status-go/blob/1794b93c16b1f68cdb7125b4b66cc8926da9eed2/protocol/messenger_communities_import_discord.go#L294

https://github.com/status-im/status-go/blob/1794b93c16b1f68cdb7125b4b66cc8926da9eed2/protocol/messenger_communities_import_discord.go#L1335

https://github.com/status-im/status-go/blob/b321f28c9deeec620aa01aa6a009c99cf537097a/protocol/messenger_community_metrics_test.go#L107

https://github.com/status-im/status-go/blob/843bae56597ecf3811d724504bd0216b867979f7/protocol/messenger.go#L3456

PR #4533, opened 2024-01-05
chore: check and manually verify community if during the fetchCommunity, the community was added to the verification loop

mprakhov:
When we were trying to fetch the community, which has the owner token and permission, `HandleCommunityDescription` put such a community into the verificationLoop.
This leads to an error when fetchCommunity was storing to DB only community without owner token permissions.

The fix is:
1. During the fetchCommunity from the store node, check if the community was added to the verification loop
2. If yes - do validation manually
3. If no - return results from communities DB

IMPACT:
1. Verification loop (moving common logic refactor)
2. Receiving the data by FetchCommunity

Closes # https://github.com/status-im/status-desktop/issues/12876

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -435,63 +435,80 @@ func (m *Manager) runOwnerVerificationLoop() {
 				for id, communities := range communitiesToValidate {
 					m.logger.Info("validating communities", zap.String("id", id), zap.Int("count", len(communities)))
 
-					for _, communityToValidate := range communities {
-						signer, description, err := UnwrapCommunityDescriptionMessage(communityToValidate.payload)
-						if err != nil {
-							m.logger.Error("failed to unwrap community", zap.Error(err))
-							continue
-						}
-
-						chainID := CommunityDescriptionTokenOwnerChainID(description)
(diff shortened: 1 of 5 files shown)


protocol/communities/manager.go
```diff
@@ -435,63 +435,80 @@ func (m *Manager) runOwnerVerificationLoop() {
...
-						ctx, cancel := context.WithTimeout(context.Background(), time.Second*3)
-						defer cancel()
+}
 
-						owner, err := m.ownerVerifier.SafeGetSignerPubKey(ctx, chainID, id)
-						if err != nil {
-							m.logger.Error("failed to get owner", zap.Error(err))
-							continue
-						}
+func (m *Manager) validateCommunity(communityToValidateData []communityToValidate) (*CommunityResponse, error) {
```
mprakhov, line 456: common code was moved from the `runOwnerVerificationLoop` into a separate fn

osmaczko (approved):
Nice, it will most likely also fix: https://github.com/status-im/status-desktop/issues/13055

igor-sirotin:
This should also help with https://github.com/status-im/status-go/issues/4437 tests.

PR #4535, opened 2024-01-06
chore: attach wrapped CommunityDescription when sharing community

### LINKED ISSUE #13055: user cannot join community after getting airdrop dropped
# Bug Report

**Version:** 0.15 RC 9, shards.test fleet

## Steps to reproduce

1. create a community
2. create a nft or token
3. set permissions to that only users with that token that was airdroped can become member
4. airdrop to someone
5. invite that user to the community

result: user cannot join the community

could be sharding related

osmaczko:
This ensures the most recent community is retrieved by the clients alongside the communnity link. Clients no longer need to rely on store node to fetch the description in that case.

fixes: status-im/status-desktop#13055

### DIFF
--- a/protocol/communities_events_utils_test.go
+++ b/protocol/communities_events_utils_test.go
@@ -1028,6 +1028,7 @@ func advertiseCommunityToUserOldWay(s *suite.Suite, community *communities.Commu
 	inputMessage.ChatId = chat.ID
 	inputMessage.Text = "some text"
 	inputMessage.CommunityID = community.IDString()
+	inputMessage.ContentType = protobuf.ChatMessage_COMMUNITY
 
 	err := owner.SaveChat(chat)
 	s.Require().NoError(err)
--- a/protocol/communities_messenger_signers_test.go
+++ b/protocol/communities_messenger_signers_test.go
@@ -562,6 +562,7 @@ func (s *MessengerCommunitiesSignersSuite) TestNewOwnerAcceptRequestToJoin() {
 	inputMessage.ChatId = chat.ID
 	inputMessage.Text = "some text"
 	inputMessage.CommunityID = community.IDString()
+	inputMessage.ContentType = protobuf.ChatMessage_COMMUNITY
 
 	err = s.alice.SaveChat(chat)
 	s.Require().NoError(err)
--- a/protocol/communities_messenger_test.go
+++ b/protocol/communities_messenger_test.go
@@ -156,6 +156,7 @@ func (s *MessengerCommunitiesSuite) TestRetrieveCommunity() {
 	inputMessage.ChatId = chat.ID
 	inputMessage.Text = "some text"
 	inputMessage.CommunityID = community.IDString()
+	inputMessage.ContentType = protobuf.ChatMessage_COMMUNITY
 
 	err = s.bob.SaveChat(chat)
 	s.Require().NoError(err)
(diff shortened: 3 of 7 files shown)


igor-sirotin:
I think I have concerns about this.
Let me take a closer look on Monday.

igor-sirotin (commented):
Sooo.

1. The bug was initially reproduced on status-desktop `0.15.0-rc9`. 
And #4341 is not part of that release. 

2. This fix covers only one case of "sharing a community". 
But user could share a community link manually. In that case no community description is attached. And what if there's several community links in a message? I don't think we should rely on user to share a community in this particular way.

3. Yes, this removes store node dependency. But I believe store nodes should be stable enough for such behaviour. 
If we speak of reliability, I think there's no guarantee that the sent community description is the most up-to-date.

4. There's intention to drop support for this property: https://github.com/status-im/status-go/issues/4437
And this PR basically reverts the idea back.

@osmaczko, wdyt?

osmaczko:
> Sooo.
> 
> 1. The bug was initially reproduced on status-desktop `0.15.0-rc9`.
>    And [New way of sharing communities #4341](https://github.com/status-im/status-go/pull/4341) is not part of that release.

The root cause of the bug is a bit different.

Long story short, when we fetch the community from store node: 
1) There is no guarantee that the newest community will be fetched or will be fetched first.
2) If community (may be the newest) is fetched and it has owner token, it will be queued for verification. As a result, store node request manager will look further in the history and will find community description that doesn't have owner token, which is usually very old one.

Point 2) was addressed by: https://github.com/status-im/status-go/pull/4533. In most cases it fixes the bug, unless very old description comes before the new one. Here point 1) kicks in and this is what I wanted to address by this PR.

> 2. This fix covers only one case of "sharing a community".
>    But user could share a community link manually. In that case no community description is attached. And what if there's several community links in a message? I don't think we should rely on user to share a community in this particular way.

If user shares community as plain text url message, for instance, through a copy-paste, then we have store nodes fetch as usual. This PR can be thought as an optimization. If user clicks share community, then plain text url message + hidden community description will be send.

> 3. Yes, this removes store node dependency. But I believe store nodes should be stable enough for such behaviour.
>    If we speak of reliability, I think there's no guarantee that the sent community description is the most up-to-date.

If someone shares the community, he shares the most up-to-date community he knows. Client can still try to ask store nodes for more up-to-date description, all what needs to be done is to set `tryDatabase = false` when we receive a community link.

With hidden community description payload approach we have guarantee that community we want to share is at least at the version we have now.

> 4. There's intention to drop support for this property: [Deprecate `protobuf.ChatMessage_Community` #4437](https://github.com/status-im/status-go/issues/4437)
>    And this PR basically reverts the idea back.

Actually, this PR [removes](https://github.com/status-im/status-go/pull/4535/files#diff-12e1567f20b885c074b2a1213664d43a77fbc87c34b5dc42cdb6d4526a86460d) the only place of sending message of `protobuf.ChatMessage_COMMUNITY` content type, community is shared through plain text. We still need to provide [backward compatibility](https://github.com/status-im/status-go/blob/19ed538c036cba7505818ee030a3f9f4a73a9b6d/protocol/messenger_handler.go#L2351-L2354), in case older clients sends us this message. Other places where `protobuf.ChatMessage_COMMUNITY` is referenced are tests which does not use `advertiseCommunityToUserOldWay` directly.

igor-sirotin:
☎️ <- as a reminder that we discussed this offline

osmaczko:
> ☎️ <- as a reminder that we discussed this offline

Yes, thank you for the discussion. In conclusion, the current solution introduces some inconsistency between sharing a community through copy-paste and using the 'share community' button. Although it's merely an optimization, it might not be the best approach since it could obscure issues related to fetching store nodes. We might consider closing this pull request and focus on improving the reliability of store node fetching. Additionally, we should look into addressing the issue of re-sending community requests to the new owner.

osmaczko:
Closing this PR as it introduces unnecessary cognitive load with minimal benefit. Given that `fetchCommunity` is functioning properly, this is not needed.

PR #4536, opened 2024-01-08
feat: link owned collectibles with latest matching transfer

dlipicar:
Part of [#12942](https://github.com/status-im/status-desktop/issues/12942)
status-dekstop part: https://github.com/status-im/status-desktop/pull/13202

- Introduce a mechanism to link each owned collectible to the latest detected matching ERC721/ERC1155 transfer.
-- When a new owned collectible appears or the balance changes, we look in the `transfers` table for a matching Tx
-- When a new ERC721/ERC1155 transfer is detected, we look for the matching owned collectible
- Add LatestTxTimestamp to the balance struct
- Add parameters indicating changes (`added/updated/removed`) to the `EventCollectiblesOwnershipUpdatePartial` and `EventCollectiblesOwnershipUpdateFinished` events.

### DIFF
--- a/services/wallet/collectibles/commands.go
+++ b/services/wallet/collectibles/commands.go
@@ -2,6 +2,7 @@ package collectibles
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"math/big"
 	"sync/atomic"
@@ -14,6 +15,7 @@ import (
 	"github.com/status-im/status-go/services/wallet/bigint"
 	walletCommon "github.com/status-im/status-go/services/wallet/common"
 	"github.com/status-im/status-go/services/wallet/thirdparty"
+	"github.com/status-im/status-go/services/wallet/transfer"
 	"github.com/status-im/status-go/services/wallet/walletevent"
 )
 
@@ -31,7 +33,22 @@ type OwnedCollectibles struct {
 	ids     []thirdparty.CollectibleUniqueID
 }
 
-type OwnedCollectiblesCb func(OwnedCollectibles)
+type OwnedCollectiblesChangeType = int
+
+const (
+	OwnedCollectiblesChangeTypeAdded OwnedCollectiblesChangeType = iota + 1
+	OwnedCollectiblesChangeTypeUpdated
+	OwnedCollectiblesChangeTypeRemoved
+)
+
+type OwnedCollectiblesChange struct {
+	ownedCollectibles OwnedCollectibles
+	changeType        OwnedCollectiblesChangeType
+}
+
+type OwnedCollectiblesChangeCb func(OwnedCollectiblesChange)
+
+type TransferCb func(common.Address, walletCommon.ChainID, []transfer.Transfer)
 
 const (
 	OwnershipStateIdle OwnershipState = iota + 1
@@ -41,12 +58,12 @@ const (
 )
 
 type periodicRefreshOwnedCollectiblesCommand struct {
-	chainID                walletCommon.ChainID
-	account                common.Address
-	manager                *Manager
-	ownershipDB            *OwnershipDB
-	walletFeed             *event.Feed
-	receivedCollectiblesCb OwnedCollectiblesCb
+	chainID                   walletCommon.ChainID
+	account                   common.Address
+	manager                   *Manager
+	ownershipDB               *OwnershipDB
+	walletFeed                *event.Feed
+	ownedCollectiblesChangeCb OwnedCollectiblesChangeCb
 
 	group *async.Group
 	state atomic.Value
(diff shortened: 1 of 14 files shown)


services/wallet/collectibles/commands.go
```diff
@@ -241,42 +260,75 @@ func (c *loadOwnedCollectiblesCommand) Run(parent context.Context) (err error) {
...
+					break
+				}
+				c.triggerEvent(EventCollectiblesOwnershipUpdatePartial, c.chainID, c.account, string(encodedMessage))
+
+				updateMessage = OwnershipUpdateMessage{}
 			}
 		}
 	}
 
+	encodedMessage, err := json.Marshal(updateMessage)
```
IvanBelyakoff: In case of error, we overwrite initial c.err here. I think it is better to move this line and check below to the line where it is used. So in some cases we dont make this unnecessary call as we already did that in `initialFetch` branch, and in other cases we wont overwrite the initial error with this one. WDYT?

services/wallet/collectibles/service.go
```diff
@@ -399,6 +411,57 @@ func (s *Service) collectibleIDsToDataType(ctx context.Context, ids []thirdparty
...
+func (s *Service) lookupTransferForCollectibles(ownedCollectibles OwnedCollectibles) {
+	// There are some limitations to this approach:
+	// - Collectibles ownership and transfers are not in sync and might represent the state at different moments.
+	// - We have no way of knowing if the latest collectible transfer we've detected is actually the latest one, so the timestamp we
+	// use might be older than the real one.
+	// - There might be detected transfers that are temporarily not reflected in the collectibles ownership.
+	// - For ERC721 tokens we should only look for incoming transfers. For ERC1155 tokens we should look for both incoming and outgoing transfers.
+	// We need to get the contract standard for each collectible to know which approach to take.
+	for _, id := range ownedCollectibles.ids {
+		fmt.Println("lookupTransferForCollectibles", id)
```
IvanBelyakoff: Log.debug or remove it?

services/wallet/transfer/commands.go
```diff
@@ -487,20 +487,23 @@ func transferTypeToEventType(transferType w_common.Type) walletevent.EventType {
...
 			c.feed.Send(walletevent.Event{
-				Type:     transferTypeToEventType(transferType),
-				Accounts: []common.Address{c.address},
-				ChainID:  c.chainClient.NetworkID(),
-				At:       int64(latestTransferTimestamp),
+				Type:        transferTypeToEventType(transferType),
+				Accounts:    []common.Address{c.address},
+				ChainID:     c.chainClient.NetworkID(),
+				At:          int64(latestTransferTimestamp),
+				EventParams: eventTransfers,
```
IvanBelyakoff, line 506: May I ask why do we need it?

services/wallet/collectibles/ownership_db.go
```diff
@@ -28,37 +28,70 @@ func NewOwnershipDB(sqlDb *sql.DB) *OwnershipDB {
...
 
-func removeAddressOwnership(creator sqlite.StatementCreator, chainID w_common.ChainID, ownerAddress common.Address) error {
-	deleteOwnership, err := creator.Prepare("DELETE FROM collectibles_ownership_cache WHERE chain_id = ? AND owner_address = ?")
+func insertTmpOwnership(
+	db *sql.DB,
+	chainID w_common.ChainID,
+	ownerAddress common.Address,
+	balancesPerContractAdddress thirdparty.TokenBalancesPerContractAddress,
+) error {
+	fmt.Println("insertTmpOwnership")
```
IvanBelyakoff: log.Debug or remove it here and other places in this PR?

services/wallet/collectibles/ownership_db.go
```diff
@@ -28,37 +28,70 @@ func NewOwnershipDB(sqlDb *sql.DB) *OwnershipDB {
...
+	// Put ols/new ownership data into temp tables
```
IvanBelyakoff: old/new
And please fix a misprint below - "createib" - created?

IvanBelyakoff (approved):
Great work, Dario! Some minor comments from me

services/wallet/collectibles/service.go
dlipicar: oops

services/wallet/collectibles/ownership_db.go
dlipicar: oops x2

services/wallet/transfer/commands.go
dlipicar: I'm using this event as a pub-subby way of informing other parts of status-go about newly detected transfers. Concretely, the collectibles service needs this to know there's a new collectibles transfer to match with some owned collectible.

We can probably do dependency injection instead or use another lighter mechanism to implement the pub-sub pattern, it's actually one of the things being discussed at status-go meetings :D

services/wallet/collectibles/commands.go
dlipicar: I still want the error event to be emitted if encoding fails for some reason, but I added a check to prevent an unnecessary call to `Marshall` in that condition.

Also, this is not done before on the `initialFetch` branch since it breaks out of the loop before that due to the `finished` condition.

services/wallet/transfer/commands.go
IvanBelyakoff: Ok, thanx, I just failed to find the usage of it first time, now I see

alaibe (approved):

PR #4537, opened 2024-01-08
chore(savedaddresses)!: favourite property removed and primary key updated

saledjenic:
- favourite column removed from the saved_addresses table
- favourite property removed from the SavedAddress struct
- ens name removed from the primary key, the primary key now is composed of address and is_test columns
- ens parameter removed from wakuext_deleteSavedAddress
- wallet_getSavedAddresses moved to wakuext_getSavedAddresses (to keep them all in a single place)
- saved addresses related endpoints removed from the wallet service, even they logically belong there, a reason for that
is avoiding emitting sync message if one uses calls from the wallet service, while that's not the case in ext service. Once
we refactor this and introduce devices syncing mechanism in the wallet service, we should not only these but other wallet
related endpoints move there (removed: wallet_getSavedAddresses, wallet_addSavedAddress and wallet_deleteSavedAddress).

Affected area:
Saved addresses


ilmotta:
cc @J-Son89 I'm not sure about the new wallet in mobile, but these upcoming breaking changes seem to at least affect our legacy wallet.

services/ext/api.go
```diff
@@ -1066,8 +1066,12 @@ func (api *PublicAPI) UpsertSavedAddress(ctx context.Context, sa wallet.SavedAdd
 	return api.service.messenger.UpsertSavedAddress(ctx, sa)
 }
 
-func (api *PublicAPI) DeleteSavedAddress(ctx context.Context, address ethcommon.Address, ens string, isTest bool) error {
-	return api.service.messenger.DeleteSavedAddress(ctx, address, ens, isTest)
+func (api *PublicAPI) DeleteSavedAddress(ctx context.Context, address ethcommon.Address, isTest bool) error {
```
stefandunca, line 1069: It's strange to see the saved address APIs here. However, I assume this is due to our current architecture.

briansztamfater (approved):

walletdatabase/migrations/sql/1704701942_remove_favourite_and_change_primary_key_for_saved_addresses.up.sql
```diff
@@ -0,0 +1,42 @@
...
+  name VARCHAR NOT NULL DEFAULT "",
+  removed BOOLEAN NOT NULL DEFAULT FALSE,
+  update_clock INT NOT NULL DEFAULT 0,
+  chain_short_names VARCHAR NOT NULL DEFAULT "",
+  ens_name VARCHAR NOT NULL DEFAULT "",
+  is_test BOOLEAN NOT NULL DEFAULT FALSE,
+  created_at INT NOT NULL DEFAULT 0,
+  color VARCHAR NOT NULL DEFAULT "primary",
+  PRIMARY KEY (address, is_test),
+  UNIQUE (name, is_test)
```
stefandunca, line 12: I see test mode polluting more of the functionality by adding extra responsibilities, I'm thinking that we should consider abstracting it at a higher level until it is not too entangled. My first thought is that having a separate wallet DB for testing might make it easier.

alaibe (approved):

stefandunca (approved):
Good job. I left some brainstorming ideas in the comments.

services/ext/api.go
alaibe: This is to have access to waku :(

J-Son89:
> cc @J-Son89 I'm not sure about the new wallet in mobile, but these upcoming breaking changes seem to at least affect our legacy wallet.

the legacy wallet is removed at this point so it should be okay. I'm sure we will communicate if any changes should affect the current mobile wallet 👌

services/ext/api.go
saledjenic: > It's strange to see the saved address APIs here. However, I assume this is due to our current architecture.

Yes, it is, I agree, but having Add/Delete saved address endpoints directly exposed as a part of wallet services avoids sending sync messages when the client uses them and that's why I removed them from there and kept them only here, to avoid misleading usage.

Logically those calls belong to the wallet service, which means once we establish an architecture we should move them there, but till then we should have them in a single place.

walletdatabase/migrations/sql/1704701942_remove_favourite_and_change_primary_key_for_saved_addresses.up.sql
saledjenic: Not sure if you're pointing to status-go unit/integration tests or test mode, in this example `is_test` field is used to mark the addresses that are saved when the app is in the test mode. 

Not entirely sure up to which level is the test mode entangled in the app at this moment, like do we differentiate the data for all wallet-related stuff based on the test mode, but if we do, that's a valid suggestion for sure.

stefandunca: > Not sure if you're pointing to status-go unit/integration tests or test mode, in this example is_test field is used to mark the addresses that are saved when the app is in the test mode.

I was referring to the blockchain test mode (where we use the chains instead of .

> Not entirely sure up to which level is the test mode entangled in the app at this moment, like do we differentiate the data for all wallet-related stuff based on the test mode, but if we do, that's a valid suggestion.

We have to use it for all user state-related data that should not be mixed between main/test modes (usually chain-related like in this case). We should keep an eye on it though if it grows too much.

PR #4129, opened 2023-10-09
chore: reject outdated CommunityDescription

osmaczko:
Prevent the inclusion of CommunityDescription with an outdated clock in
MessengerResponse to avoid false-positives in tests and reduce redundant
data exchange between status-go and clients.

PR #4543, opened 2024-01-09
Add basic peersyncing

cammellos:
This commit adds basic syncing capabilities with peers if they are both online.

It updates the work done on MVDS, but I decided to create the code in status-go instead, since it's very tight to the application (similarly the code that was the inspiration for mvds, bramble, is all tight together at the database level).

I reused parts of the protobufs.

The flow is:

1) An OFFER message is sent periodically with a bunch of message-ids and
   group-ids.
2) Anyone can REQUEST some of those messages if not present in their
   database.

3) The peer will then send over those messages.

It's still very rough and we need to add authorization (as of now it would sync for example messages with users outside of the community), backoff, retry, handling errors, avoding extra calls etc.

It's disabled by default, but I am planning to add a way to set up the flags.


protocol/messenger.go
```diff
@@ -365,6 +369,8 @@ func NewMessenger(
 		installationID,
 		logger,
 	)
+	c.featureFlags.StoreNodesDisabled = true
```
osmaczko: Shouldn't `false` be the default?

protocol/common/message_sender.go
```diff
@@ -795,27 +810,39 @@ func (s *MessageSender) HandleMessages(wakuMessage *types.Message) (*HandleMessa
...
 		DatasyncRequests: h.DatasyncRequests,
 	}
 }
 
 type handleMessageResponse struct {
+	Hash             []byte
 	Message          *v1protocol.StatusMessage
 	DatasyncMessages []*v1protocol.StatusMessage
+	DatasyncSender   *ecdsa.PublicKey
+	DatasyncGroupID  []byte
```
osmaczko: unused

protocol/messenger_peersyncing_test.go
```diff
@@ -0,0 +1,276 @@
...
+	msg, err := s.owner.peersyncing.AvailableMessages()
+	s.Require().NoError(err)
+	s.Require().Len(msg, 1)
+
+	// Bob joins the community
+	advertiseCommunityTo(&s.Suite, community, s.owner, s.bob)
+
+	s.joinCommunity(community, s.owner, s.bob)
+
+	// Bob should now send an offer
```
osmaczko, line 150:
```suggestion
	// Bob should now receive an offer and respond with a request
```

protocol/messenger_peersyncing_test.go
```diff
@@ -0,0 +1,276 @@
...
+	_, err = WaitOnMessengerResponse(
+		s.bob,
+		func(r *MessengerResponse) bool {
+			return s.bob.peersyncingOffers[messageID[2:]] != 0
+		},
+		"offer not sent",
+	)
+	s.Require().NoError(err)
+
+	// Owner should now reply to the offer
```
osmaczko, line 160:
```suggestion
	// Owner should now receive request and respond with a message
```

protocol/messenger_peersyncing.go
```diff
@@ -0,0 +1,333 @@
...
+		var chatIDs [][]byte
+		for id, _ := range community.Chats() {
+			chatIDs = append(chatIDs, []byte(community.IDString()+id))
+		}
+
+		if len(chatIDs) == 0 {
+			continue
+		}
+
+		availableMessages, err := m.peersyncing.AvailableMessagesByGroupIDs(chatIDs, 20)
```
osmaczko: We offer the 20 most recent messages every 5 seconds. If we receive more than 20 new messages in under 5 seconds, there will be a gap in the messages offered. I assume this is acceptable?

protocol/common/feature_flags.go
```diff
@@ -17,4 +17,13 @@ type FeatureFlags struct {
...
 	DisableAutoMessageLoop bool
+
+	// ResendRawMessagesDisabled indicates whether we should be disabling sending raw messages
+	ResendRawMessagesDisabled bool
+
+	// StoreNodesDisabled indicates whether we should fetch messages from store nodes
+	StoreNodesDisabled bool
+
+	// DisableAdvertisingMessages indicates whether we should advertise and sync messages with other peers
+	DisableAdvertisingMessages bool
```
osmaczko:
```suggestion
	Peersyncing bool
```

Disable means as if the feature was there before.

protocol/messenger_peersyncing.go
```diff
@@ -0,0 +1,333 @@
...
+			case <-m.quit:
+				ticker.Stop()
+				logger.Debug("peersyncing loop stopped")
+				return
+			}
+		}
+	}()
+}
+
+func (m *Messenger) peerSyncingLoop() error {
```
osmaczko:
```suggestion
func (m *Messenger) sendDatasyncOffers() error {
```

osmaczko (commented):
Cool! Looks like a significant improvement to chat messages reliability. I don't have any major concerns :+1:

Note:
I'm not certain what was pinpointed as the main issue affecting the reliability of community messages. However, if it's related to hash ratchet key distribution, as I suspect, this approach is unlikely to resolve that.

protocol/messenger.go
cammellos: oh yes, sorry, for testing :)

protocol/messenger_peersyncing.go
cammellos: actually, I haven't played with the numbers, didn't think much about it.
It should be defintely more than 5 seconds, maybe 1 minute or so, and more than 20, maybe 40/60, but that's a number to play with I guess, let me know if you have any suggestion

cammellos:
> Cool! Looks like a significant improvement to chat messages reliability. I don't have any major concerns 👍
> 
> Note: I'm not certain what was pinpointed as the main issue affecting the reliability of community messages. However, if it's related to hash ratchet key distribution, as I suspect, this approach is unlikely to resolve that.

I think this addresses mostly general message reliability issues, rather than specifically to encrypted communities (although it should help to some extent, especially if we allow syncing those kind of messages, but that's for later).

PR is ready to be reviewed I think, I would like to break it down in steps, it's all behind a feature flag (other than saving the messages, so once enabled users can share messages). Next I want to work on allowing to enable it through setting so we can easily test it.

osmaczko (approved):

protocol/protobuf/chat_message.pb.go
```diff
@@ -1,24 +1,24 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
+// versions:
```
mprakhov, line 2: Could you please configure your env to have the same configurations as other devs have in order not to generate proto which are different due to different set-up?

1. remove all my manual go-related installations (like protobuf and etc)
2. run make setup , make setup-dev
3. It will generate ~/go folder. Add path ~/go/bin to PATH
4. Done, version should be the correct
5. Run make generate and check that `versions` removed

mprakhov (approved):

protocol/protobuf/chat_message.pb.go
cammellos: we use `shell.nix`, I think that's the best way to replicate dev environments, maybe @jakubgs can give his opinion. I can try to make it the same as what we normally use, but I would lean towards all of us using `shell.nix` rather than depending on the `os` packages, since people might have different package managers etc

osmaczko: I think resurrecting https://github.com/status-im/status-go/issues/4191 would save us from such discussions.

cammellos: I tried but the env doesn't install on my system, (shell.nix works, but the version specified is too old), is it ok if I merge as it is?

jakubgs: @apentori is already working on this in:

* https://github.com/status-im/status-go/pull/4388

Just other priorities took his focus away from it.

osmaczko: @jakubgs thanks for update :pray:

protocol/communities/community_description_encryption.go
```diff
@@ -59,15 +60,31 @@ func encryptDescription(encryptor DescriptionEncryptor, community *Community, de
...
+		return nil, nil
+	}
+
+	var failedToDecrypt []*CommunityPrivateDataFailedToDecrypt
+
 	for keyIDSeqNo, encryptedDescription := range description.PrivateData {
-		decryptedDescription, err := encryptor.decryptCommunityDescription(keyIDSeqNo, encryptedDescription)
+		decryptedDescriptionResponse, err := encryptor.decryptCommunityDescription(keyIDSeqNo, encryptedDescription)
+		if decryptedDescriptionResponse != nil && !decryptedDescriptionResponse.Decrypted {
+			failedToDecrypt = append(failedToDecrypt, &CommunityPrivateDataFailedToDecrypt{GroupID: id, KeyID: decryptedDescriptionResponse.KeyID})
```
osmaczko: This loop also iterates thorugh encrypted channels, GroupID is not alwas communityID (can be communityID+channelID).

osmaczko: There is a complication because we cannot retrieve the `GroupID` from the `keyID`, which prevents us from `SaveHashRatchetMessage`. Additionally, storing the hash ratchet message for future use implies processing the same message more than once, which could lead to further issues.

The cleanest solution to me would be to invoke `decryptDescription` on already processed community once we receive a new key for it.

cammellos: >This loop also iterates thorugh encrypted channels, GroupID is not alwas communityID (can be communityID+channelID).

Ah, this I didn't know, let me check if I can replicate a failure in a test.

> Additionally, storing the hash ratchet message for future use implies processing the same message more than once, which could lead to further issues.

This I am not sure I understand fully, could you elaborate? (we intentionally reprocess the same message i.e the community description), but I think I might misunderstand you

osmaczko: > This I am not sure I understand fully, could you elaborate? (we intentionally reprocess the same message i.e the community description), but I think I might misunderstand you

Instead processing the same message and making things like https://github.com/status-im/status-go/pull/4129 impossible/difficult to implement we could introduce `HandleHashRatchetKeys` on a messenger level that will:
```
communityID := getCommunityID(groupID)
...
community := m.manager.GetByID(communityID)
...
response := m.manager.DecryptCommunity(community)
...
return response
```

PR #4544, opened 2024-01-09
chore: remove boilerplate check of `communities.GetByID` clients

osmaczko:
motivated by:
https://github.com/status-im/status-go/pull/4514#discussion_r1445808116

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -833,9 +833,6 @@ func (m *Manager) EditCommunityTokenPermission(request *requests.EditCommunityTo
 	if err != nil {
 		return nil, nil, err
 	}
-	if community == nil {
-		return nil, nil, ErrOrgNotFound
-	}
 
 	tokenPermission := request.ToCommunityTokenPermission()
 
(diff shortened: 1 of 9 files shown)


igor-sirotin (approved):
Actually I liked the logic of returning `nil, nil` in such cases.
That means "no error occurred, no community found".

But that's my own thoughts, I don't have a any strong opinion on this.
Maybe worth implementing all such getters like this? There're probably also contact getters (I didn't check tho).

ibrkhalil (approved):

osmaczko:
> Actually I liked the logic of returning `nil, nil` in such cases. That means "no error occurred, no community found".

I also thought this approach was okay until I decided to find all the places where the check for `community == nil` is missing. There are many such places, and people tend to forget to add this check, and will likely continue to forget in the future as well. The severity of missing this check is really high, as it may lead to a crash.

> Maybe worth implementing all such getters like this? There're probably also contact getters (I didn't check tho).

If a similar pattern is used there as well, where we check both the error and contact nilness separately after each invocation, then we should probably do the same.

PR #4551, opened 2024-01-11
fix: avoid losing cached community collectibles data

dlipicar:
If the community couldn't be properly fetched, we would overwrite any previously cached collectible metadata with the one given by the providers (mostly empty since most metadata comes from Waku).
This PR changes this logic, so in that situation we now only insert new items but don't replace what's in the DB.


alaibe (approved):

stefandunca (approved):
👍

PR #4552, opened 2024-01-11
chore(wallet) rename pokt urls to grove in default networks and tests

stefandunca:
Desktop PR [#13191](https://github.com/status-im/status-desktop/pull/13191)

Updates status-desktop [#13189](https://github.com/status-im/status-desktop/issues/13189)

### DIFF
--- a/api/default_networks.go
+++ b/api/default_networks.go
@@ -27,7 +27,7 @@ var ganacheTokenAddress = common.HexToAddress("0x8571Ddc46b10d31EF963aF49b6C7799
 var mainnet = params.Network{
 	ChainID:                mainnetChainID,
 	ChainName:              "Ethereum Mainnet",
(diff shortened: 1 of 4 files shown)


jakubgs:
Please rebase.

dlipicar (approved):

api/default_networks.go
```diff
@@ -45,7 +45,7 @@ var mainnet = params.Network{
 var goerli = params.Network{
```
dlipicar, line 45: why don't we have an entry for sepolia here?

stefandunca: Good point might be that we added it to the desktop only. We inject config at onboarding and login so we overwrite this. There was a discussion to drop network client injection and rely on status-go, not sure what happened with that.

stefandunca: Here it is https://github.com/status-im/status-desktop/issues/11435

saledjenic (approved):

PR #4554, opened 2024-01-11
fix: update the list of chains supported by opensea

dlipicar:
Part of https://github.com/status-im/status-desktop/issues/13190

Opensea no longer supports Arbitrum Goerli and Optimism Goerli (they were replaced by Sepolia).

Moreover, they don't return any kind of error when trying to use the endpoints for those chains, but the results are not correct (not up to date).

This PR updated the list of supported chains for the Opensea client.


stefandunca (approved):
👍

alaibe (approved):

PR #4562, opened 2024-01-12
chore: fetch community token metadata during handling community description

mprakhov:
Fetch community token metadata by all clients while handling community description

This information is required for the wallet team to restore all token data if a user receives a community token

Not doing this inside `(s *Service) fetchCommunity)` will help us to escape race conditions (when we are trying to write twice the same data and getting SQLite unique constraint error) and escape unnecessary locks

IMPACT: community tokens receive


dlipicar (approved):
Wallet is happy with this change

igor-sirotin (approved):

PR #4566, opened 2024-01-15
chore: update default wallet account name

### LINKED ISSUE #4565: Update the default wallet name to "Account 1"
Currently the default wallet name is "Ethereum account". This issue is to request this default name to be updated to "Account 1" as shown in the wallet designs

https://www.figma.com/file/HKncH4wnDwZDAhc4AryK8Y/Wallet-for-Mobile?type=design&node-id=1-46718&mode=design&t=r7mG6i8ADZHOPvXS-4

jo-mut:
This pr updates the default wallet name to "Account 1"

Closes #4565


mohsen-ghafouri (approved):
LGTM

appdatabase/migrations/sql/1705399289_.up.sql
qfrank, line 1: why there're 2 new empty sql file added? this is not normal 🤔 if you only want to change `defaults.go`, the rest can be reverted.

jo-mut: @qfrank i ran `make migration` but since there is no any migration in this pr that was not really needed. I will revert them

jo-mut: turns out `make migration` creates an empty migration file when there are not migrations. that was my mistake. I ran this command instead of `make generate`. I have reverted the files

J-Son89 (approved):

appdatabase/migrations/sql/1705399289_.up.sql
qfrank: may i know why you want `make generate` when you only need to change `defaults.go`?

jo-mut: initially there was a failing test so I ran `make generate` i guess i actually dont need to run make generate

qfrank: interesting, which test? did `make generate` fix that test? sorry i got so many questions cause of additional changes 🙂

qfrank (approved):

appdatabase/migrations/sql/1705399289_.up.sql
jo-mut: generating the files did not actually fix the failing test. It was failing on develop. Maybe it was just a flaky test as later on it worked fine

PR #4568, opened 2024-01-15
feat: implement new endpoint to set customization color

mohsen-ghafouri:
new endpoint to set/change profile customization color

related to status-mobile issue https://github.com/status-im/status-mobile/issues/18453


protocol/messenger_settings_test.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/status-im/status-go/protocol/tt"
+	"github.com/status-im/status-go/waku"
+	"github.com/stretchr/testify/suite"
+)
+
+func TestMessengerSettings(t *testing.T) {
+	suite.Run(t, new(MessengerSettingsSuite))
+}
+
+type MessengerSettingsSuite struct {
```
cammellos, line 23: I think you can inherit from BasMessengerTestSuite or similar

protocol/messenger_settings.go
```diff
@@ -20,3 +22,19 @@ func (m *Messenger) SetLogLevel(request *requests.SetLogLevel) error {
...
+func (m *Messenger) SetCustomizationColor(ctx context.Context, request *requests.SetCustomizationColor) error {
+	updatedAt := timesource.GetCurrentTimeInMillis()
+
+	acc, err := m.multiAccounts.GetAccount(request.KeyUID)
+	if err != nil {
+		return err
+	}
+	acc.CustomizationColor = request.CustomizationColor
+	acc.CustomizationColorClock = updatedAt
+	err = m.syncAccountCustomizationColor(ctx, acc)
```
cammellos: Does this save the value as well locally? (you can check if login/logout preserve the value)

mohsen-ghafouri: it will dispatch a message to `ApplicationMetadataMessage_SYNC_ACCOUNT_CUSTOMIZATION_COLOR` and at the end it will `UPDATE accounts...` with this new customization color. i will test it in mobile

protocol/messenger_settings.go
```diff
@@ -20,3 +22,30 @@ func (m *Messenger) SetLogLevel(request *requests.SetLogLevel) error {
...
+	acc.CustomizationColor = request.CustomizationColor
+
+	tNow := timesource.GetCurrentTimeInMillis()
+	if acc.CustomizationColorClock >= tNow {
+		acc.CustomizationColorClock++
+	} else {
+		acc.CustomizationColorClock = tNow
+	}
+
+	_, err = m.multiAccounts.UpdateAccountCustomizationColor(request.KeyUID, string(acc.CustomizationColor), acc.CustomizationColorClock)
```
qfrank: we should check affected rows as well here :)
so there're multiple places need to check now, maybe we should ask `UpdateAccountCustomizationColor` to return affected row number and error directly to avoid duplicating code

protocol/requests/set_customization_color.go
```diff
@@ -0,0 +1,8 @@
+package requests
+
+import "github.com/status-im/status-go/multiaccounts/common"
+
+type SetCustomizationColor struct {
+	CustomizationColor common.CustomizationColor `json:"customizationColor"`
+	KeyUID             string                    `json:"keyUid"`
+}
```
qfrank, line 15: missing validation function like [this](https://github.com/status-im/status-go/blob/81f4c860865c5e7c22b654fb06ae183db5eb7ae4/protocol/requests/send_one_to_one_message.go#L15)

protocol/messenger_settings_test.go
qfrank: I think @cammellos meant `MessengerBaseTestSuite`, there is an example you maybe interested [MessengerSyncActivityCenterSuite](https://github.com/status-im/status-go/blob/81f4c860865c5e7c22b654fb06ae183db5eb7ae4/protocol/messenger_sync_activity_center_test.go#L25)

protocol/messenger_settings.go
```diff
@@ -20,3 +22,30 @@ func (m *Messenger) SetLogLevel(request *requests.SetLogLevel) error {
...
+func (m *Messenger) SetCustomizationColor(ctx context.Context, request *requests.SetCustomizationColor) error {
+
+	acc, err := m.multiAccounts.GetAccount(request.KeyUID)
+	if err != nil {
+		return err
+	}
+
+	acc.CustomizationColor = request.CustomizationColor
+
+	tNow := timesource.GetCurrentTimeInMillis()
```
qfrank, line 40: maybe add a comment above would be better. e.g. "use a combination of wall clock + lamport timestamp, just like `Chat#NextClockAndTimestamp`"

protocol/messenger_settings.go
mohsen-ghafouri: @qfrank could you please check again ?

protocol/messenger_settings_test.go
```diff
@@ -0,0 +1,100 @@
...
+	err = s.alice.SetCustomizationColor(context.TODO(), &requests.SetCustomizationColor{KeyUID: s.alice.account.KeyUID, CustomizationColor: common.CustomizationColorBlue})
+	s.Require().NoError(err)
+	_, err = WaitOnMessengerResponse(s.alice2, func(r *MessengerResponse) bool {
+		return len(r.CustomizationColor) > 0
+	}, "message syncAccountCustomizationColor not received")
+	s.Require().NoError(err)
+	acc, err = s.alice2.multiAccounts.GetAccount(s.alice.account.KeyUID)
+	acc2, err = s.alice2.multiAccounts.GetAccount(s.alice2.account.KeyUID)
+	s.Require().NoError(err)
+	s.Require().Equal(acc.CustomizationColor, acc2.CustomizationColor)
```
qfrank, line 104: check if they changed to blue ?

protocol/requests/set_customization_color.go
```diff
@@ -0,0 +1,26 @@
...
+
+var ErrSetCustomizationColorInvalidColor = errors.New("customizationColor: invalid color")
+var ErrSetCustomizationColorInvalidKeyUID = errors.New("keyUid: invalid id")
+
+type SetCustomizationColor struct {
+	CustomizationColor common.CustomizationColor `json:"customizationColor"`
+	KeyUID             string                    `json:"keyUid"`
+}
+
+func (a *SetCustomizationColor) Validate() error {
```
qfrank, line 17: I see nowhere invoking `Validate` :)

qfrank (approved):
LGTM except additional minor comments

protocol/requests/set_customization_color.go
mohsen-ghafouri: oh sorry, i missed that. resolved :)

protocol/messenger_settings_test.go
qfrank: emm.. i saw test failed, extend from `MessengerSyncSettingsSuite` is a bad idea since it will trigger run test `TestSyncSettings_PreferredName` , didn't know why my previous comment ignored,  [MessengerSyncActivityCenterSuite](https://github.com/status-im/status-go/blob/81f4c860865c5e7c22b654fb06ae183db5eb7ae4/protocol/messenger_sync_activity_center_test.go#L25) is a better example @mohsen-ghafouri

mohsen-ghafouri: @qfrank sorry i missed that comment, will resolve it

PR #4570, opened 2024-01-16
(fix/status-go) fix profile picture update/removal

### LINKED ISSUE #9947: 1x1 chat: removing custom picture is not reflected on contact's side
### Description

1. open app
2. add mutual contact
3. open chat with this contact
4. set custom user picture for yourself and save
5. make sure the picture is shown for your contact
6. delete the custom picture from your profile to have default one
7. check contact's chat

As result, the contact is still seeing custom picture , even it was removed

Screenshot 2023-03-20 at 16 03 51

kounkou:
### Description

This PR is linked to [13197](https://github.com/status-im/status-desktop/pull/13197) and contains the status-go necessary changes to make work 
- the changing of the profile picture and
- the first step for being able to reset the profile picture (see below for more details)

This PR also contains generated files as we are augmenting the protobuf with the flag to know if the images should be considered deleted.

This PR fixes [9947](https://github.com/status-im/status-desktop/issues/9947)

### Design decisions

In this PR I evaluated Possible solutions in order to mark images as deleted :

### **[Preferred - Implemented]** Modify the client to handle the delete operation

This design decision makes more sense after discussion with Patryk and is simpler than the 2 steps delete below as it tackle the inner lack of handling of the delete profile operation.

### Execute 2 steps delete operation

The strategy is to execute the Delete in 2 steps.
Step 1 get the information of the contactcode advertisement, and fills the payload with the images.
Step 2 add a indicator that this is a delete operation
Step 2 is to perform the clean up of the Image to be deleted in the Database

### Generate an empty identity image

Sending the contactcode requires a payload, and because during the delete we don't have images, sending the default image would help work around the lack of image. This requires generating a default image and agreeing on Frontend and backend.
This approach also requires saving the image, so database resource is needed.

### Create a new API for the delete Image

Currently the sending of concactcode is done under several constraints that include having a payload. So the idea would be to trigger a deleteIndentity image that doesn't have all the constraints of the sending of contact code.
This might take longer to execute, and also I am not sure why we should use another API here, as we are still in the context of the contactcode.

### Known Limitations

- This is the reset is not persisted yet, meaning that at reboot, the reset changes are lost


osmaczko (commented):
As discussed offline, the issue lies on the receiver's side, where we don't clean up images if they arrive empty.

multiaccounts/database.go
```diff
@@ -450,9 +451,29 @@ func (db *Database) publishOnIdentityImageSubscriptions(change *IdentityImageSub
...
+	if err != nil {
+		return err
+	}
+
+	db.markedImageForDeletion = false
+
+	return nil
+}
+
+func (db *Database) DeleteIdentityImage(keyUID string) error {
```
kounkou: I had a discussion with @osmaczko, (thank you! 🙏🏾) and here is the summary:

Instead of introducing a flag and transmitting the images, we can effectively manage the absence of images within the `ContactChatIdentity` in the `messenger_handler`. This approach allows the recipient of the contact code to be informed about the change (deletion of the profile) and update the view accordingly.

Regarding the current code:

At present, we do not address the absence of images because our handling is focused solely on image updates. Enhancing the code to accommodate missing images should effectively address the reported issue. Additionally, adopting this approach will streamline the code, eliminating the need for managing a flag as proposed in this pull request.

kounkou:
> As discussed offline, the issue lies on the receiver's side, where we don't clean up images if they arrive empty.

I have updated the PR with the new revision that fix all above mentioned issues. Here is a video showing all the expected functionalities


[profile updates.webm](https://github.com/status-im/status-go/assets/2589171/60a24da8-3ff2-4efa-8d6f-b52f845c4e05)

protocol/persistence.go
```diff
@@ -755,6 +755,22 @@ func (db sqlitePersistence) SaveContactChatIdentity(contactID string, chatIdenti
 		_ = tx.Rollback()
 	}()
 
+	if len(chatIdentity.Images) == 0 {
```
osmaczko: I am wondering if this is a suitable approach. In the future, we might want to remove only images of a specific type. In this scenario, an image will be removed only if there are no other images present at all.

The solution could be:
- collect all image types that are left
- then `DELETE FROM chat_identity_contacts WHERE contact_id = ? AND image_type NOT IN (?, ?, ...)`, `tx.Exec(contactID, imageTypes)`

kounkou: Hey, thanks for your comment @osmaczko. I'll be reviewing the code with this new perspective. Seems to me like an amazing comment on my unknown unknowns 🙏🏾

kounkou: It turns out, the keys to the chatIdentity.Images is actually a string that represents the type of images. This type could be one of :

```
SmallDimName = "thumbnail"
LargeDimName = "large"
BannerIdentityName = "banner"
```

The idea is to delete the image matching only the image type above.

-----

I have updated the code to handle the removing of images of specific type. I agree that this udpate makes the code more robust for future possible improvements like stated above, for instance when deleting 1 of 3 images.
I have also renamed the confusing `ImageType` to `ImageFormat` and retested the changes again like in the video
I believe those changes should be enough to complete this status-go PR. Sending another revision.
Thanks again!

kounkou: I have updated the code to handle the removing of images of specific type. I agree that this udpate makes the code more robust for future possible improvements like stated above, for instance when deleting 1 of 3 images.
I have also renamed the confusing `ImageType` to `ImageFormat`
I believe those changes should be enough to complete this status-go PR. Sending another revision.
Thanks!

osmaczko (approved):
:muscle:

kounkou:
Once I obtain second approval, I will regenerate the files and merge.

appdatabase/migrations/bindata.go
```diff
@@ -188,7 +188,7 @@ func _1640111208_dummyUpSql() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "1640111208_dummy.up.sql", size: 258, mode: os.FileMode(0644), modTime: time.Unix(1692726719, 0)}
```
MishkaRogachev: If there are no changes in migrations, bindatas in the corresponding directories can be excluded from the commit

MishkaRogachev (approved):
Great work!

PR #4571, opened 2024-01-16
fix: allow to update rpc urls

alaibe:
* in all cases update the original urls in case they changed with a new api key
* Update current urls only if the current one is the same as the previous original one and the original one did change

### DIFF
--- a/rpc/network/network.go
+++ b/rpc/network/network.go
@@ -130,13 +130,8 @@ func (nm *Manager) Init(networks []params.Network) error {
 	// Add new networks and update related chain id for the old ones
 	for i := range networks {
 		found := false
-		// Update original urls if they are empty
-		if networks[i].OriginalRPCURL == "" {
-			networks[i].OriginalRPCURL = networks[i].RPCURL
-		}
-		if networks[i].OriginalFallbackURL == "" {
-			networks[i].OriginalFallbackURL = networks[i].FallbackURL
-		}
+		networks[i].OriginalRPCURL = networks[i].RPCURL
+		networks[i].OriginalFallbackURL = networks[i].FallbackURL
 
 		for j := range currentNetworks {
 			if currentNetworks[j].ChainID == networks[i].ChainID {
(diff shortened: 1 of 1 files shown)


saledjenic (approved):

rpc/network/network.go
```diff
@@ -148,7 +143,22 @@ func (nm *Manager) Init(networks []params.Network) error {
 						errors += fmt.Sprintf("error updating network fallback_url for ChainID: %d, %s", currentNetworks[j].ChainID, err.Error())
 					}
 				}
-				err := nm.UpdateRPCURLs(currentNetworks[j].ChainID, networks[i].RPCURL, networks[i].OriginalRPCURL, networks[i].FallbackURL, networks[i].OriginalFallbackURL)
+
+				if networks[i].OriginalRPCURL != currentNetworks[j].OriginalRPCURL && currentNetworks[j].RPCURL == currentNetworks[j].OriginalRPCURL {
+					err := nm.UpdateRPCURL(networks[i].ChainID, networks[i].OriginalRPCURL)
```
saledjenic: @alaibe wondering if it makes sense to do these queries as transactions instead of executing each query separately? 

Cause in the current implementation if an error occurs at some point, we return the errors list but do not revert the changes.

rpc/network/network.go
```diff
@@ -193,9 +203,18 @@ func (nm *Manager) UpdateRelatedChainID(chainID uint64, relatedChainID uint64) e
 	return err
 }
 
-func (nm *Manager) UpdateRPCURLs(chainID uint64, rpcURL, originalRPCURL, fallbackURL, originalFallbackURL string) error {
-	_, err := nm.db.Exec(`UPDATE networks SET rpc_url = ?, original_rpc_url = ?, fallback_url = ?, original_fallback_url = ?  WHERE chain_id = ?`,
-		rpcURL, originalRPCURL, fallbackURL, originalFallbackURL, chainID)
+func (nm *Manager) UpdateRPCURL(chainID uint64, rpcURL string) error {
```
saledjenic: @alaibe guess that we don't need to make this and the calls below public (they all should start with a small first letter)?

alaibe: Done

rpc/network/network.go
alaibe: yes currently we only log the errors, are they really independent query or are they a group of query? good question. I am wondering if we should not return the error

saledjenic: They are independent queries that are executed successively and theoretically if all of them as an entirety introduce the change then if an error occurs among any two we should roll back. That's why I proposed to make a transaction and get back as soon as we face an error.

But in practice, I don't have anything against this approach you did, since the only error which may happen in that loop is the DB error, and in reality if that happens we don't have so much to do, cause the app won't run anymore.

So you can leave it as it's now, it's already approved. :)

PR #4573, opened 2024-01-16
Comment out all logged flaky tests

rasom:
It gets more and more difficult to merge PRs because of false tests failures.
It also makes it difficult sometimes to catch genuine failures of newly introduced code/tests.

The goal of this PR is to comment out as many of those tests as needed to get relatively reliable results. The metric for "reliable" is 5 subsequent successful runs of tests suite on Jenkins (arbitrary but should be good enough).

There are separate issues for all commented out tests which can be found [here](https://github.com/status-im/status-go/issues?q=is%3Aopen+is%3Aissue+label%3A%22E%3AFlaky+Test%22).

My suggestion would be that person who fixes a flaky test runs it at least 1000 times locally and only after that files a PR with fix. We also can have a separate Jenkins job for that purpose. 

Anyway all that stuff is open for discussion.

### DIFF
--- a/api/backend_test.go
+++ b/api/backend_test.go
@@ -1,7 +1,7 @@
 package api
 
 import (
-	"context"
+	//"context"
 	"crypto/sha256"
 	"database/sql"
 	"encoding/hex"
@@ -34,7 +34,8 @@ import (
 	"github.com/status-im/status-go/protocol/requests"
 	"github.com/status-im/status-go/rpc"
 	"github.com/status-im/status-go/services/typeddata"
-	walletservice "github.com/status-im/status-go/services/wallet"
+
+	//walletservice "github.com/status-im/status-go/services/wallet"
 	"github.com/status-im/status-go/signal"
 	"github.com/status-im/status-go/sqlite"
 	"github.com/status-im/status-go/t/helpers"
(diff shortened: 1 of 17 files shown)


rasom:
ok it is done
Screenshot 2024-01-17 at 15 34 30

IvanBelyakoff (approved):

osmaczko (commented):
I would wait for the opinion from @cammellos on that one. If I remember correctly, he had some good arguments for keeping the flaky tests.

FYI: We have a nightly build for the detection of flaky tests. Please see https://github.com/status-im/status-go/issues/4264. The problem is that the reports are not usable yet because the job crashes. Once that is resolved, we should have a better view of test flakiness.

cammellos:
> I would wait for the opinion from @cammellos on that one. If I remember correctly, he had some good arguments for keeping the flaky tests.
> 
> FYI: We have a nightly build for the detection of flaky tests. Please see #4264. The problem is that the reports are not usable yet because the job crashes. Once that is resolved, we should have a better view of test flakiness.

@rasom @osmaczko 
As agreed yesterday, we are going to merge this and then work on re-enabling the tests. I'll update you at the end of the day where we are at with it.

rasom:
@cammellos thanks

@osmaczko I don't think that visibility is a problem at the moment, we have identified 30+ of those tests already. But I personally wasted many hours on build re-runs during past months. That has to stop, and at this point it doesn't look like we will fix those 30+ tests in two hours or so and we just need to wait a bit more.

PR #4579, opened 2024-01-17
replace mention name with public key 

qfrank:
With this PR, when user send chat message contains mention e.g. `hello @john ` , `john` will be replaced with corresponding public key like this: `hello @0x123...` , this is a small change so that status-mobile don't need make an extra API call to backend to replace mention name with public key before calling [SendChatMessages](https://github.com/status-im/status-go/blob/1924828f8d565d1c4f31367069edc3846dc5071e/services/ext/api.go#L891).

relate status mobile [PR](https://github.com/status-im/status-mobile/pull/18533)

status: ready.


siddarthkay (approved):
looks good

ilmotta (approved):

PR #4581, opened 2024-01-17
Cleanup transfers scheduler and handle errors gracefully

### LINKED ISSUE #11074: Clean up new status-go scheduler and revise TODOs
- Revise error handling for Sequential Fetch Strategy 
- ~Make block check timeout configurable based on chain block mining time~ Not valid since we make less requests now and do not check each block but use a balance checker contract instead to cut down on a number of requests

IvanBelyakoff:
Handle errors in findBlocksCommand and findNewBlocksCommand gracefully.
Add ErrorCounter type for async package and a FiniteCommandWithErrorCounter to exit on errors overflow
Add tests

Closes [#11074](https://github.com/status-im/status-desktop/issues/11074)

### DIFF
--- a/services/wallet/async/async.go
+++ b/services/wallet/async/async.go
@@ -47,7 +47,6 @@ func (c SingleShotCommand) Run(ctx context.Context) error {
 type FiniteCommand struct {
 	Interval time.Duration
 	Runable  func(context.Context) error
-	OnExit   *func(context.Context, error)
 }
 
 func (c FiniteCommand) Run(ctx context.Context) error {
@@ -134,7 +133,7 @@ func NewAtomicGroup(parent context.Context) *AtomicGroup {
 	return ag
 }
 
-// AtomicGroup terminates as soon as first goroutine terminates..
+// AtomicGroup terminates as soon as first goroutine terminates with error.
 type AtomicGroup struct {
 	ctx    context.Context
 	cancel func()
@@ -319,18 +318,11 @@ func (c FiniteCommandWithErrorCounter) Run(ctx context.Context) error {
 
 		if c.ErrorCounter.SetError(err) {
 			return false, err
-		} else {
-			return true, err
 		}
+		return true, err
 	}
 
 	quit, err := f(ctx)
-	defer func() {
-		if c.OnExit != nil {
-			(*c.OnExit)(ctx, err)
-		}
-	}()
-
 	if quit {
 		return err
 	}
--- a/services/wallet/async/async.go
+++ b/services/wallet/async/async.go
@@ -4,12 +4,18 @@ import (
 	"context"
 	"sync"
 	"time"
+
+	"github.com/ethereum/go-ethereum/log"
 )
 
 type Command func(context.Context) error
 
 type Commander interface {
 	Command(inteval ...time.Duration) Command
+}
+
+type Runner interface {
+	Run(context.Context) error
 }
 
 // SingleShotCommand runs once.
@@ -41,6 +47,7 @@ func (c SingleShotCommand) Run(ctx context.Context) error {
 type FiniteCommand struct {
 	Interval time.Duration
 	Runable  func(context.Context) error
+	OnExit   *func(context.Context, error)
 }
 
 func (c FiniteCommand) Run(ctx context.Context) error {
@@ -138,6 +145,20 @@ type AtomicGroup struct {
 	error error
 }
 
+type AtomicGroupKey string
+
+func (d *AtomicGroup) SetName(name string) {
+	d.ctx = context.WithValue(d.ctx, AtomicGroupKey("name"), name)
+}
+
+func (d *AtomicGroup) Name() string {
+	val := d.ctx.Value(AtomicGroupKey("name"))
+	if val != nil {
+		return val.(string)
+	}
+	return ""
+}
+
 // Go spawns function in a goroutine and stores results or errors.
 func (d *AtomicGroup) Add(cmd Command) {
 	d.wg.Add(1)
@@ -149,6 +170,7 @@ func (d *AtomicGroup) Add(cmd Command) {
 		if err != nil {
 			// do not overwrite original error by context errors
 			if d.error != nil {
+				log.Info("async.Command failed", "error", err, "d.error", d.error, "group", d.Name())
 				return
 			}
 			d.error = err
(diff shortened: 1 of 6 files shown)


services/wallet/async/async.go
```diff
@@ -244,3 +266,85 @@ func (d *QueuedAtomicGroup) onFinish() {
...
+
+func (ec *ErrorCounter) Error() error {
+	return ec.err
+}
+
+func (ec *ErrorCounter) MaxErrors() int {
+	return ec.maxErrors
+}
+
+type FiniteCommandWithErrorCounter struct {
```
dlipicar, line 307: 💯 for reusable components!

dlipicar (approved):
LGTM!

So the cause for high RPC consumption was infinitely retrying the fetches for some transfers that always failed, and now we're stopping after a couple of retries?
If so:
1) What's causing them to fail like that?
2) What do we do about these transfers? Are they somehow re-processed at some point?

IvanBelyakoff:
> * What's causing them to fail like that?
> * What do we do about these transfers? Are they somehow re-processed at some point?

Good questions, @dlipicar 
Actually it was not the goal for the PR to fix high RPC consumption, but rather to fix error handling which was not clear and messed up, and fix the commands lifecycles so that they keep working when needed, stopped when requested.

1) From what I saw the main reason was hitting some limit - either requests per second rate (then we keep retrying and reach the requests per day limit) or total requests per day. And I've rechecked this - we still keep trying loading transfers even in case of errors forever (or if an error happens for fetching blocks and reaches maxError limit, then transfer loading will stop as well) with a little delay (5 sec), so I'm going to fix it now and stop after some retries - thanx for the heads up. A good thing would be to add an exponential backoff instead of specific retry interval, but I wont do it in scope of this PR.
2) Well, if we fail to load a transfer after some retries - I think we should stop it and report an error to the UI that we have troubles loading transfers. But then we either should allow user to retry it or retry it after some long interval under the hood (can't think of a good number here, it does not make much sense, because for the user it means they wont see the transfers, but if we already tried some exponential backoff with a delay of at least some minutes - then we are out of luck in terms of good UX). This is also a design decision.
If we fail to load history blocks - it could be because of the same reasons, or because of non-archival node issue (I think we are not handling it any more now as we access archive nodes AFAIK, but we generally should) - then again we should do the same as for transfers.

So now we are not reprocessing the blocks or transfers until restart. WDYT?

services/wallet/async/async_test.go
```diff
@@ -0,0 +1,59 @@
...
+	group.Add(func(ctx context.Context) error {
+		time.Sleep(1 * time.Millisecond)
+		return err // failure
+	})
+	group.Add(func(ctx context.Context) error {
+		for {
+			select {
+			case <-ctx.Done():
+				return nil
+			case <-time.After(1 * time.Millisecond): // wait for the first command to fail and terminate the second command
```
stefandunca: This timeout should be bigger than the expected runtime of the fail task and fail the test if triggered before `ctx.Done()`, if I understand the `AtomicGroup` specifications.

services/wallet/async/async_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestAtomicGroupTerminatesOnOneCommandFailed(t *testing.T) {
+	ctx := context.Background()
+	group := NewAtomicGroup(ctx)
+
+	err := errors.New("error")
+	group.Add(func(ctx context.Context) error {
+		time.Sleep(1 * time.Millisecond)
```
stefandunca: What do you think about removing the sleep (not to make the test take unnecessarily) but adding instead this failure command after the next command instead, ensuring the other is failing if finished after the `1ms` timeout?

services/wallet/async/async_test.go
```diff
@@ -0,0 +1,59 @@
...
+			select {
+			case <-ctx.Done():
+				return nil
+			case <-time.After(1 * time.Millisecond): // wait for the first command to fail and terminate the second command
+				_ = 1
+			}
+		}
+	})
+
+	<-group.WaitAsync()
```
stefandunca: What do you think of calling `group.Wait()`? should be enough to prove the point?

services/wallet/async/async_test.go
```diff
@@ -0,0 +1,59 @@
...
+		time.Sleep(1 * time.Millisecond)
+		return nil // success
+	})
+	group.Add(func(ctx context.Context) error {
+		for {
+			select {
+			case <-ctx.Done():
+				return nil
+			case <-time.After(5 * time.Millisecond):
+				finished = true
```
stefandunca, line 44: What do you think of having it more explicitly with a `require.Fail("timeout waiting for the other command")` instead of `finished` variable?

services/wallet/async/async_test.go
```diff
@@ -0,0 +1,59 @@
...
+	require.Equal(t, err, group.Error())
+}
+
+func TestAtomicGroupWaitsAllToFinish(t *testing.T) {
+	ctx := context.Background()
+	group := NewAtomicGroup(ctx)
+
+	finished := false
+	group.Add(func(ctx context.Context) error {
+		time.Sleep(1 * time.Millisecond)
```
stefandunca, line 35: Same as above, how about replacing this sleep with a different order of adding the commands? Because we know the other command needs 5 ms, if this is added after the other we know it has a chance to run without a sleep.

stefandunca (approved):
Great finding and fix. I'm looking forward to seeing the final solution, indeed there are some challenges as @IvanBelyakoff suggested.
I left some thoughts and questions in the inline comments.

services/wallet/async/async_test.go
IvanBelyakoff: The intention was to fail the first goroutine while the second is spinning and pretending to do something. The timeouts are the same because I think in any execution order (real situations) it should work the same.

services/wallet/async/async_test.go
IvanBelyakoff: I actually was not sure what happens if I don't have any timeout here and the function returns immediately with error before the second goroutine even started and waits on ctx.Done() - whether ctx.Done case will be executed or not since cancel() for it was already called. Now due to your comments I tested it and it seems to work fine if Done channel is closed before an attempt to read on it. So I will remove the timeout.

services/wallet/async/async_test.go
IvanBelyakoff: No, Stefan, I believe this is wrong. The order of adding commands (goroutines in fact) does not define the order of execution - they will be executed in random order. So we can not rely on that, if that is what you mean.

services/wallet/async/async_test.go
IvanBelyakoff: I think what you suggest is the opposite to the intention of the test.
The description of AtomicGroup says "AtomicGroup terminates as soon as first goroutine terminates.." and that led me to a conclusion that even in case one routine finishes first with no error the group will cancel the rest routines. But the code makes it clear that the rest goroutines will be cancelled only in case of failure. So I wrote a test to avoid the confusion and prevent possible code changes when AtomicGroup will indeed terminate independent of the return value of the first finished goroutine.
So I do want to wait until the second goroutine times out and sets the `finished` flag, otherwise the test will fail if it will finish on ctx.Done() instead.

services/wallet/async/async_test.go
IvanBelyakoff: But I will simplify the test, I think it would be better. Thank you for paying attention to it.

IvanBelyakoff (commented):
@stefandunca thank you for your feedback. I will make some amendments

services/wallet/async/async_test.go
stefandunca: Thanks for the great clarifications.

> the description of AtomicGroup says "AtomicGroup terminates as soon as first goroutine terminates.." and that led me to a conclusion that even in case one routine finishes first with no error the group will cancel the rest routines

This was my understanding also looking at the original comment, hence the 3ms timeout seemed like an error case.

services/wallet/async/async_test.go
stefandunca: The comment was based on reading the old description :). Thanks for clarifying.

PR #4594, opened 2024-01-19
chore: add _test suffix to messenger_testing_utils.go

osmaczko:
Otherwise it will be compiled and included into the protocol package when building the production binary.


osmaczko:
Oops, I see now, it's used in other packages; it's not as simple as just renaming. I will create an issue for that.

PR #4598, opened 2024-01-19
fix: closing tmp db file after creation

saledjenic:
Changes done in this commit solve the " The process cannot access the file because it is being used by another process." error which was noticeable on Windows while migrating a profile keypair to a Keycard.


alexjba (approved):
👍

Issue #4605, opened 2024-01-22
Skip nonce check for generated account when mnemonic has not been revealed


rasom:
Nonce check is not needed in this case at all as all ongoing transfers happen within the app anyway.

PR #4611, opened 2024-01-23
[#4605] Skip nonce check when possible during tx detection

rasom:
There is no need to check nonce for accounts which are generated by application when mnemonic has not been revealed as all outgoing transfer should happen within the application.

### DIFF
--- a/services/wallet/transfer/commands_sequential.go
+++ b/services/wallet/transfer/commands_sequential.go
@@ -6,6 +6,8 @@ import (
 	"sync"
 	"time"
 
+	"golang.org/x/exp/slices"
+
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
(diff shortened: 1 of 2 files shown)


saledjenic (approved):

IvanBelyakoff (commented):
Great job! Just one comment inline

services/wallet/transfer/commands_sequential.go
```diff
@@ -148,36 +148,53 @@ func (c *findNewBlocksCommand) Run(parent context.Context) error {
...
 	}
 
-	if len(accountsWithDetectedChanges) != 0 || c.iteration%nonceCheckIntervalIterations == 0 {
-		c.findAndSaveEthBlocks(parent, c.fromBlockNumber, headNum, accounts)
+	if len(accountsWithDetectedChanges) != 0 {
+		c.findAndSaveEthBlocks(parent, c.fromBlockNumber, headNum, accountsToCheck)
+	} else if c.iteration%nonceCheckIntervalIterations == 0 && len(accountsWithOutsideTransfers) > 0 {
+		c.findAndSaveEthBlocks(parent, c.fromBlockNumber, headNum, accountsWithOutsideTransfers)
+		for _, account := range accountsToCheck {
+			err := c.markEthBlockRangeChecked(account, &BlockRange{nil, c.fromBlockNumber, headNum})
```
IvanBelyakoff, line 195: Saving block ranges should be done only for those accs that are in accountsToCheck but not in accountsWithOutsideTransfers?
```
if !slices.Contains(accountsWithOutsideTransfers, account) {
     err := c.markEthBlockRangeChecked(account, &BlockRange{nil, c.fromBlockNumber, headNum})
```

rasom: I was lazy as it makes no difference here, we can exclude those from the loop but it won't harm.

rasom: added

IvanBelyakoff (approved):

PR #4613, opened 2024-01-23
Fix encryption metadata issues

cammellos:
This commit fixes a few issues with communities encryption:
1) Key distribution was disconnected from the community description, this created a case where the key would arrive after the community description and that would result in the client thinking that it was kicked. 
 To overcome this, we added a message that signals the user that is kicked. Also, we distribute the key with the community description so that there's no more issues with timing.
This is a bit expensive for large communities, and it will require some further optimizations.

2) Key distribution is now also connected to the request to join response, so there are no timing issues.

3) Fixes an issue with key distribution (race condition) where the community would be modified before being compared, resulting in a comparison of two identical communities, which would result in no key being distributed. This commit only partially address the issue.

The code is still rough, as this is a release blocker. I will refactor `common/message_sender.go` in a separate PR, too many paths, but too risky to do it in the scope of this PR.
And some code will need to be extracted from the publish community loop, since the kicking code is living there, but in the interest of time I'd like to do that separately if that's ok.


yakimant:
Please don't merge the Jenkins change, it is done for debug on the branch only.

protocol/common/message_sender.go
```diff
@@ -671,7 +661,32 @@ func (s *MessageSender) SendPublic(
 		return nil, errors.Wrap(err, "failed to wrap a public message in the encryption layer")
 	}
 
-	if !rawMessage.SkipEncryptionLayer {
+	if len(rawMessage.HashRatchetGroupID) != 0 {
+
+		var ratchet *encryption.HashRatchetKeyCompatibility
+		var err error
+		// We have just rekeyed, pull the latest
+		ratchet, err = s.protocol.GetCurrentKeyForGroup(rawMessage.HashRatchetGroupID)
```
mprakhov, line 669: minor:
`ratchet, err := s.protocol.GetCurrentKeyForGroup(rawMessage.HashRatchetGroupID)`

mprakhov (approved):
LGTM

protocol/common/message_sender.go
cammellos: (I'll address in the next PR, in order not to re-run the tests), thank you!

protocol/messenger_communities.go
```diff
@@ -338,7 +392,14 @@ func (m *Messenger) handleCommunitiesSubscription(c chan *communities.Subscripti
 					return
 				}
 				if sub.Community != nil {
-					publishOrgAndDistributeEncryptionKeys(sub.Community)
+					if sub.Community == nil {
+						continue
+					}
```
osmaczko, line 397: It will be never entered :thinking:

protocol/messenger_communities.go
```diff
@@ -338,7 +392,14 @@ func (m *Messenger) handleCommunitiesSubscription(c chan *communities.Subscripti
...
+					// NOTE: because we use a pointer here, there's a race condition where the community would be updated before it's compared to the previous one.
+					// This results in keys not being propagated as the copy would not see any changes
+					communityCopy := sub.Community.CreateDeepCopy()
```
osmaczko, line 400: Perhaps instead of having a Community pointer in the subscription, we could use a Community ID. This way, we can re-fetch the latest version. This solution, together with the recently introduced `map[communityID, mutex]`, will ensure the latest version of the Community is used and will mitigate race conditions.

osmaczko (commented):
Thanks for addressing that :+1:

PR #4043, opened 2023-09-18
chore(wallet) optimize activity filer query

stefandunca:
### Updates [#11036](https://github.com/status-im/status-desktop/issues/11036)

Activity filter optimizations

# Activity filter optimizations

| Name                   | Original   | tr_type   | join     | hex      | no-db     | db_only   |     exp |    net_j |
|:-----------------------|:-----------|:----------|:---------|:---------|:----------|:----------|---------:|---------:|
| RAM_NoFilter-10        | 49580229   | 51253242  | 51112462 | 50915133 | 121217817 | 141691008 | 50908642 | 50239712 |
| SSD_NoFilter-10        | 49963604   | 51393588  | 51213038 | 50881483 | 120785679 | 141063467 | 50462767 | 49676867 |
| SSD_MovingWindow-10    | 53695712   | 54155292  | 54161733 | 54061325 | 126966633 | 146866017 | 53479929 | 53350475 |
| SSD_AllAddr_AllTos-10  | 41382804   | 41195225  | 51684175 | 52107262 | 64348100  | 97608833  | 50523529 | 49968321 |
| SSD_OneAddress-10      | 34945275   | 35103850  | 31066429 | 31328762 | 50927300  | 54322971  | 30098529 | 30252546 |
| FilterSend_AllAddr-10  | 39546808   | 37566604  | 38389725 | 38260738 | 114820458 | 125588408 | 37127625 | 36864575 |
| FilterSend_6Addr-10    | 41221458   | 41111225  | 40848288 | 40135492 | 118629700 | 128200467 | 38942521 | 39012100 |
| FilterThreeNetworks-10 | -          | -         | -        | -        | -         | -         | 50058929 | 49854450 |

Legend

- Original: before current change
- tr_type: use tr_type instead of `IN` filter_addresses
- join: filter recipients as join. I also tried from and owner but it didn't work as expected
- hex: remove usage of HEX and insert binary. Didn't notice much improvement though
- no-db: removed using temporary DB
- db_only: only used temporary DB and join
- exp: some more experiments
- net_j: join networks instead of WHERE ... IN ...

PR #4619, opened 2024-01-24
fix(wallet): do not return gas-only ETH entries of transfers table for

IvanBelyakoff:
activities request
- remove gas-fee-only ETH transfers once token transfers with the same tx_hash, nonce, networkd_id and address are loaded
- delete such history transfers on migration
- minor refactoring of `saveAndConfirmPending` (unrelated to the task)
- removed redundant parameters from `prepareUpdatedBlockRange` method

### DIFF
--- a/services/wallet/activity/filter.sql
+++ b/services/wallet/activity/filter.sql
@@ -157,6 +157,7 @@ layer2_networks(network_id) AS (
 mint_methods(method_hash) AS (
 	%s
 )
+
 SELECT
 	transfers.hash AS transfer_hash,
 	NULL AS pending_hash,
@@ -208,7 +209,7 @@ SELECT
 		WHEN transfers.tx_from_address = zeroAddress AND transfers.type = "erc20" THEN substr(json_extract(tx, '$.input'), 1, 10)
 		ELSE NULL
 	END AS method_hash,
-	CASE 
+	CASE
 		WHEN transfers.tx_from_address = zeroAddress AND transfers.type = "erc20" THEN (SELECT 1 FROM json_each(transfers.receipt, '$.logs' ) WHERE json_extract( value, '$.topics[0]' ) = communityMintEvent)
 		ELSE NULL
 	END AS community_mint_event
@@ -254,11 +255,11 @@ WHERE
 				AND (
 					transfers.type = 'erc721'
 					OR (
-						transfers.type = 'erc20' 
+						transfers.type = 'erc20'
 						AND (
 							(method_hash IS NOT NULL AND method_hash IN mint_methods)
 							OR community_mint_event IS NOT NULL
-						) 
+						)
 					)
 				)
 			)
@@ -281,11 +282,11 @@ WHERE
 			AND (
 				transfers.type = 'erc721'
 				OR (
-					transfers.type = 'erc20' 
+					transfers.type = 'erc20'
 					AND (
 						(method_hash IS NOT NULL AND method_hash IN mint_methods)
 						OR community_mint_event IS NOT NULL
-					) 
+					)
 				)
 			)
 		)
--- a/services/wallet/activity/activity_test.go
+++ b/services/wallet/activity/activity_test.go
@@ -1378,3 +1378,74 @@ func TestGetMultiTxDetails(t *testing.T) {
 	require.Equal(t, td.multiTx1Tr2.Hash, details.ChainDetails[1].Hash)
 	require.Equal(t, td.multiTx1Tr2.Contract, *details.ChainDetails[1].Contract)
 }
+
+func TestGetActivityEntriesSkipEthGasFeeOnlyTransfers(t *testing.T) {
+	deps, close := setupTestActivityDB(t)
+	defer close()
+
+	to := eth.Address{0x1}
+	from := eth.Address{0x2}
+	hash := eth.Hash{0x3}
+	blkNum := int64(1)
+	chainID := common.ChainID(1)
+	nonce := uint64(1)
+
+	// Insert 0-value gas-only ETH transfer as a result of token transfer's gas fee
+	transfer.InsertTestTransfer(t, deps.db, to, &transfer.TestTransfer{
+		TestTransaction: transfer.TestTransaction{
+			ChainID:   chainID,
+			From:      from,
+			Hash:      hash,
+			BlkNumber: blkNum,
+			Nonce:     nonce,
+		},
+		To:    to,
+		Value: 0,
+	})
+
+	entries, err := getActivityEntries(context.Background(), deps, []eth.Address{to}, true, []common.ChainID{chainID}, Filter{}, 0, 10)
+	require.NoError(t, err)
+	require.Equal(t, 1, len(entries))
+	require.Equal(t, hash, entries[0].transaction.Hash)
+
+	// Insert token transfer
+	transfer.InsertTestTransferWithOptions(t, deps.db, to,
+		&transfer.TestTransfer{
+			TestTransaction: transfer.TestTransaction{
+				ChainID:   chainID,
+				From:      from,
+				Hash:      hash,
+				BlkNumber: blkNum,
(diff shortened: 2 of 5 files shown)


services/wallet/activity/filter.sql
```diff
@@ -157,6 +157,43 @@ layer2_networks(network_id) AS (
 mint_methods(method_hash) AS (
 	%s
 )
+-- Select all token transfers
```
alaibe: Isn't the comment misleading, that will also return erc721 and contract call?

alaibe (approved):

services/wallet/activity/filter.sql
Cuteivist: I am also curious on that.

Cuteivist (approved):

services/wallet/activity/filter.sql
IvanBelyakoff: Well, AFAIK all transfers that don't have type='eth' are token transfers, so that will return erc721 and others.
https://github.com/status-im/status-go/blob/b6ade53603b4f08bfdcf316f914bee3944537a08/services/wallet/common/log_parser.go#L24

IvanBelyakoff: But now I think the only transfer type that can be not "eth" and not token is bridging eth. Not sure how we store it in DB, will check now

alaibe: how do  we categorize mint?

Cuteivist: @alaibe In SQL mint is when `from` null, and type `erc721` or `erc20` and mint method signature checked

https://github.com/status-im/status-go/blob/cd9617cad7f8934405b87a5404221f586831e9b8/services/wallet/activity/filter.sql#L275

IvanBelyakoff: > @alaibe In SQL mint is when `from` null, and type `erc721` or `erc20` and mint method signature checked
> 
> https://github.com/status-im/status-go/blob/cd9617cad7f8934405b87a5404221f586831e9b8/services/wallet/activity/filter.sql#L275

So it is still a token transfer, with ethereum gas spent. So I guess it should be fine. But I can't test it for now, as I don't have my wallet working at all neither production nor dev due to some issues of running out of requests. So if you don't mind I will merge this. If we experience any issues for some corner cases, I'll be right there to fix it or revert the fix. 
WDYT? @alaibe @Cuteivist

Cuteivist: Sounds good.

services/wallet/activity/filter.sql
```diff
@@ -157,6 +157,43 @@ layer2_networks(network_id) AS (
...
+, eth_real_zero_value_rows AS (
+    SELECT *
+    FROM eth_zero_value_rows
+    WHERE (tx_hash, address, network_id, account_nonce) NOT IN (
+        SELECT tx_hash, address, network_id, account_nonce
+        FROM token_rows
+    )
+)
+-- Merge all transfers
+, clean_transfers AS (
```
stefandunca: What is the performance impact of adding this filter layer on top of the `transfers` table?

We already have a considerable performance penalty with the complex query, that I could not find a way to optimize further. 
The last performance analysis conclusion was that the way to go is to cache as many or all the criteria, dynamically aggregated here, into an activity table. Hence, I'm thinking that the way to do this without a high-performance impact is by adding an indexed flag of "gas transfer" and filtering that out. What do you think @IvanBelyakoff, would that be possible?

stefandunca (commented):
This change looks like dropping some transaction entries (user won't have awareness and access to the hash in any way through our app). I can't think of any other transactions that we hide this way, we only group them, hence I'm wondering if this is what we want or is temporary until we sync with design?

services/wallet/activity/filter.sql
IvanBelyakoff: Thanx Stefan for the feedback. I am not aware of the status of performance of our queries for the activities. Do we have any settled way to measure it consistently to know how each PR affects performance?
Or is it manually measured by every dev with their own queries?
That should be easy to add this flag, I just did not want to add it because we can do it without it, but since we have performance issues, I would consider it now

stefandunca: > Do we have any settled way to measure it consistently to know how each PR affects performance?

There are these benchmarks that I used in the performance improvements, I'm not sure they work still https://github.com/status-im/status-go/blob/develop/services/wallet/activity/benchmarks_test.go
I had on my list to run them in CI. However, because there is nothing yet there and have to build a foundation, that was never a priority.

You can find the last results in the last commit message: https://github.com/status-im/status-go/commit/579f7e4a523b2734a589bc452d4f391334627a0c or corresponding PR [#4043](https://github.com/status-im/status-go/pull/4043)

One way to run them is by `go test -v ./services/wallet/activity/... --benchtime=10x '--run=^$' --bench=BenchmarkGetActivityEntries --tags=gowaku_no_rln,gowaku_skip_migrations"` that I found in my command history
```

stefandunca: >  ... but since we have performance issues, I would consider it now

The performance slowdown is visible on a heavy history like the one I used to do the hands-on benchmarks using the `0x1` address which has many thousands of activity entries. A regular user won't notice it so there is no urgent concern.

stefandunca: We can go with the current change but track the effects over time. AFAIK we don't yet have defined some reference user profiles to generate some performance base line and check it from time to time or regularly, therefore all my understanding is based on a lot of assumptions.

IvanBelyakoff:
> This change looks like dropping some transaction entries (user won't have awareness and access to the hash in any way through our app). I can't think of any other transactions that we hide this way, we only group them, hence I'm wondering if this is what we want or is temporary until we sync with design?

Yes, this is intended, we don't want this transaction to show up ever.

services/wallet/activity/filter.sql
IvanBelyakoff: > > Do we have any settled way to measure it consistently to know how each PR affects performance?
> 
> There are these benchmarks that I used in the performance improvements, I'm not sure they work still https://github.com/status-im/status-go/blob/develop/services/wallet/activity/benchmarks_test.go I had on my list to run them in CI. However, because there is nothing yet there and have to build a foundation, that was never a priority.
> 
> You can find the last results in the last commit message: [579f7e4](https://github.com/status-im/status-go/commit/579f7e4a523b2734a589bc452d4f391334627a0c) or corresponding PR [#4043](https://github.com/status-im/status-go/pull/4043)
> 
> One way to run them is by `go test -v ./services/wallet/activity/... --benchtime=10x '--run=^$' --bench=BenchmarkGetActivityEntries --tags=gowaku_no_rln,gowaku_skip_migrations"` that I found in my command history



> > Do we have any settled way to measure it consistently to know how each PR affects performance?
> 
> There are these benchmarks that I used in the performance improvements, I'm not sure they work still https://github.com/status-im/status-go/blob/develop/services/wallet/activity/benchmarks_test.go I had on my list to run them in CI. However, because there is nothing yet there and have to build a foundation, that was never a priority.
> 
> You can find the last results in the last commit message: [579f7e4](https://github.com/status-im/status-go/commit/579f7e4a523b2734a589bc452d4f391334627a0c) or corresponding PR [#4043](https://github.com/status-im/status-go/pull/4043)
> 
> One way to run them is by `go test -v ./services/wallet/activity/... --benchtime=10x '--run=^$' --bench=BenchmarkGetActivityEntries --tags=gowaku_no_rln,gowaku_skip_migrations"` that I found in my command history

I measured the performance. According to this benchmark, the degradation is around 15-25%, which is huge considering that we want to ignore this type of transactions always. So I am using another approach now - removing these transactions as soon as they are detected. My previous concern why I did not want to remove them was that I thought they are needed for balance chart, but I double checked and balance chart should not be affected.

Cuteivist (approved):

PR #4621, opened 2024-01-24
fix: clear waku envelopes cache when deleting a chat

### LINKED ISSUE #11686: Missing community messages when re-joining
Community-level encryption:
1. Create token-gated community - ✓
2. Join with MemberA who satisfies the criteria - ✓
3. Send message "X" to general channel - ✓
4. MemberA should see the "X" message - ✓
5. Change token permissions so that MemberA no longer satisfies the criteria - ✓
6. MemberA should be kicked from the community - ✓ 
7. Send message "Y" to general channel - ✓
8. Change token permissions so that MemberA once again meets the criteria - ✓
9. MemberA should be ~~re-added to the community - X (user has to manually re-join)~~ able to manually re-join
**10. MemberA should see both "X" and "Y" messages - X (only sees "X")**
11. Send message "Z" to general channel - ✓
12. MemberA should see "X", "Y" and "Z" messages - X (only sees "X" and "Z")

_Originally posted by @lukaszso in https://github.com/status-im/status-desktop/issues/11657#issuecomment-1650700759_
            
 I believe it happens also on non token-gated communities.

igor-sirotin:
➡️  PRO TIP: Review by commits 🤓 

# Description

When a user leaves or gets auto-kicked from a token-gated community, all of the messages are deleted from the database. These messages should be fetched again if the user rejoins the community, but waku envelopes cache prevented this.

Because we don't have a `messageId` -> `envelopeHash` information, we clear the whole cache.

"Clearing envelopes cache" consists 2 operations:
  - clear in-memory `waku.envelopes` cache
  - clear in-database `transport_message_cache` table

# Main things done:

1. Implement and expose methods for clearing waku in-memory envelopes cache
1. Clean waku enveloeps cache when deleting a chat from database
3. Update `TestBecomeMemberPermissions` to check that messages are there after rejoining the community
4. Remove `WithResendParams(3, 3)` option from communities tests

# Minor refactoring on the way:

1. Move `newTestMessenger` `extraOptions` argument to `testMessengerConfig`
6. Implement `WithTestStoreNode` messenger option, which should be used in tests
7. Define `ErrPermissionToJoinNotSatisfied` instead of duplicating `permission to join not satisfied` string
8. Use `logger.Named` instead of `logger.With(zap.String("name", ...)`

Fixes https://github.com/status-im/status-desktop/issues/11686

### DIFF
--- a/eth-node/bridge/geth/waku.go
+++ b/eth-node/bridge/geth/waku.go
@@ -274,6 +274,10 @@ func (w *gethWakuWrapper) RequestStoreMessages(ctx context.Context, peerID []byt
 
 func (w *gethWakuWrapper) ConnectionChanged(_ connection.State) {}
 
+func (w *gethWakuWrapper) ClearEnvelopesCache() {
+	w.waku.ClearEnvelopesCache()
+}
+
 type wakuFilterWrapper struct {
 	filter *wakucommon.Filter
 	id     string
--- a/eth-node/bridge/geth/wakuv2.go
+++ b/eth-node/bridge/geth/wakuv2.go
@@ -289,6 +289,10 @@ func (w *gethWakuV2Wrapper) ConnectionChanged(state connection.State) {
 	w.waku.ConnectionChanged(state)
 }
 
+func (w *gethWakuV2Wrapper) ClearEnvelopesCache() {
+	w.waku.ClearEnvelopesCache()
+}
+
 type wakuV2FilterWrapper struct {
 	filter *wakucommon.Filter
 	id     string
--- a/eth-node/types/waku.go
+++ b/eth-node/types/waku.go
@@ -161,4 +161,7 @@ type Waku interface {
 
 	// ConnectionChanged is called whenever the client knows its connection status has changed
 	ConnectionChanged(connection.State)
+
+	// ClearEnvelopesCache clears waku envelopes cache
+	ClearEnvelopesCache()
 }
--- a/protocol/communities/errors.go
+++ b/protocol/communities/errors.go
@@ -44,3 +44,4 @@ var ErrInvalidManageTokensPermission = errors.New("no privileges to manage token
 var ErrRevealedAccountsAbsent = errors.New("revealed accounts is absent")
 var ErrNoRevealedAccountsSignature = errors.New("revealed accounts without the signature")
 var ErrNoFreeSpaceForHistoryArchives = errors.New("history archive: No free space for downloading history archives")
+var ErrPermissionToJoinNotSatisfied = errors.New("permission to join not satisfied")
--- a/protocol/communities_messenger_helpers_test.go
+++ b/protocol/communities_messenger_helpers_test.go
@@ -489,7 +489,6 @@ func joinOnRequestCommunity(s *suite.Suite, community *communities.Community, co
 	s.Require().NoError(err)
 }
 
-/*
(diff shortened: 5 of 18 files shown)


richard-ramos (approved):

osmaczko (commented):
> Because we don't have a messageId -> envelopeHash information, we clear the whole cache.

I'm curious about the decision to clear the entire cache. Isn't it too much? What about side effects, like processing the same messages many times? Could you please explain the reasoning behind this approach compared to creating mentioned mapping of `messageId -> envelopeHash` and selectively removing envelopes from a specific chat?

cammellos:
> > Because we don't have a messageId -> envelopeHash information, we clear the whole cache.
> 
> I'm curious about the decision to clear the entire cache. Isn't it too much? What about side effects, like processing the same messages many times? Could you please explain the reasoning behind this approach compared to creating mentioned mapping of `messageId -> envelopeHash` and selectively removing envelopes from a specific chat?

we had a discussion about this and I suggested the approach. 

The reasoning is that it's simpler, and as well if re-processing the same message has side effects, that's a bug and we should be handling this, since that's possible to happen in normal circumstances, as we process messages "at least once", and not "at most once", so it makes only bugs harder to spot, basically our code should take that into consideration when we write it.

That combined with the fact that clearing the cache does not mean that we will refetch those messages, I think the impact is low, but we can always go from simple -> complex without much extra effort if necessary. (i.e we can go for the simplest solution first, if we see that's problematic, we can easily add a more sophisticated way to remove messages)

osmaczko:
Thanks @cammellos, that's fair enough :+1:

osmaczko (approved):
:muscle:

igor-sirotin:
I've commented out `TestBecomeMemberPermissions`, as it was commented before.
With my changes it has become extremely flaky passes very rarely.

Here's an enviable list of errors the test falls into now:

```go
2024-01-29T19:30:36.825Z        DEBUG   bob     communities/community_description_encryption.go:68      failed to decrypt community private data        {"keyIDSeqNo": "b897ac24bf4b91d264697b074609637cac59a38d4b3be464d2639b7c792c88461", "error": "no ratchet key for given keyID", "errorVerbose": "no ratchet key for given .....

    communities_messenger_token_permissions_test.go:792: 
                Error Trace:    /Users/igorsirotin/Repositories/Status/status-desktop-2/vendor/status-go/protocol/communities_messenger_token_permissions_test.go:792
                Error:          Target error should be in err chain:
                                expected: "permission to join not satisfied"
                                in chain: "already joined"
                Test:           TestMessengerCommunitiesTokenPermissionsSuite/TestBecomeMemberPermissions

```

```go
    communities_messenger_token_permissions_test.go:761: 
                Error Trace:    /Users/igorsirotin/Repositories/Status/status-desktop-2/vendor/status-go/protocol/communities_messenger_token_permissions_test.go:761
                Error:          Received unexpected error:
                                no community that satisfies criteria
                Test:           TestMessengerCommunitiesTokenPermissionsSuite/TestBecomeMemberPermissions
```

```go
    communities_messenger_token_permissions_test.go:764: 
                Error Trace:    /Users/igorsirotin/Repositories/Status/status-desktop-2/vendor/status-go/protocol/communities_messenger_token_permissions_test.go:764
                Error:          Received unexpected error:
                                timed out when waiting for key distribution
                Test:           TestMessengerCommunitiesTokenPermissionsSuite/TestBecomeMemberPermissions
```

PR #4627, opened 2024-01-25
fix: handle community shard unassignment and update

richard-ramos:
Should fix the issue described in https://github.com/status-im/status-desktop/issues/13267

### DIFF
--- a/protocol/messenger_communities.go
+++ b/protocol/messenger_communities.go
@@ -715,6 +715,16 @@ func (m *Messenger) subscribeToCommunityShard(communityID []byte, shard *shard.S
 	return m.transport.SubscribeToPubsubTopic(pubsubTopic, pubK)
 }
 
+func (m *Messenger) unsubscribeFromShard(shard *shard.Shard) error {
+	if m.transport.WakuVersion() != 2 {
+		return nil
+	}
+
+	// TODO: this should probably be moved completely to transport once pubsub topic logic is implemented
+
+	return m.transport.UnsubscribeFromPubsubTopic(shard.PubsubTopic())
+}
+
 func (m *Messenger) joinCommunity(ctx context.Context, communityID types.HexBytes, forceJoin bool) (*MessengerResponse, error) {
 	logger := m.logger.Named("joinCommunity")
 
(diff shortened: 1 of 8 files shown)


wakuv2/waku.go
```diff
@@ -610,6 +610,24 @@ func (w *Waku) getPubsubTopic(topic string) string {
...
+		return errors.New("only available for full nodes")
+	}
+
+	topic = w.getPubsubTopic(topic)
+
+	if !w.node.Relay().IsSubscribed(topic) {
+		return nil
+	}
+
+	w.node.Relay().RemoveTopicValidator(topic)
```
chaitanyaprem: This is already done as part of w.node.Relay().Unsubscribe. 
No need to do this specifically again.

wakuv2/waku.go
```diff
@@ -610,6 +610,24 @@ func (w *Waku) getPubsubTopic(topic string) string {
 	return topic
 }
 
+func (w *Waku) unsubscribeFromPubsubTopicWithWakuRelay(topic string) error {
+	if w.settings.LightClient {
```
chaitanyaprem: This check is also done in line 1559 below already, hence is duplicate and can be removed.

chaitanyaprem (commented):
Reviewed only parts that are invoking Waku API. 

Also, wondering if it is a good idea to just allow a community to switch a shard without some sort of migration warning. But this should be done at UI level i guess. 
Also, it may be better to indicate the impact of doing this, as message history would be lost and any other such impact of changing shard.

protocol/messenger_communities.go
```diff
@@ -715,6 +715,16 @@ func (m *Messenger) subscribeToCommunityShard(communityID []byte, shard *shard.S
 	return m.transport.SubscribeToPubsubTopic(pubsubTopic, pubK)
 }
 
+func (m *Messenger) unsubscribeFromShard(shard *shard.Shard) error {
+	if m.transport.WakuVersion() != 2 {
+		return nil
+	}
```
igor-sirotin, line 721: This check is redundant, as it's already done it `transport.UnsubscribeFromPubsubTopic`

protocol/messenger_communities.go
```diff
@@ -2829,21 +2875,39 @@ func (m *Messenger) HandleCommunityShardKey(state *ReceivedMessageState, message
 	return nil
 }
 
-func (m *Messenger) handleCommunityShardAndFiltersFromProto(community *communities.Community, shard *shard.Shard, privateKeyBytes []byte) error {
+func (m *Messenger) handleCommunityShardAndFiltersFromProto(community *communities.Community, clock uint64, shard *shard.Shard, privateKeyBytes []byte) error {
```
igor-sirotin: What's the `clock` is needed here for? I don't see it being used.

wakuv2/waku.go
igor-sirotin: But this also prevents of calling `RemoveTopicValidator`. Not sure what it does, just noting.

igor-sirotin (approved):

PR #4629, opened 2024-01-26
[#4604] Skip balance requests on checking nonce

### LINKED ISSUE #4604: Remove balance check when checking nonce changes every 30 mins
Currently every 60 mins nonce is checked for addresses on which zero sum eth transfer might occur. As an old algorithm is used there, the balance diff check also happens although it is not needed. It costs 2 requests per account per hour per network, which is not a big number but still can be easily skipped.

rasom:
Currently once per hour the nonce is checked for accounts which might have outgoing txs initiated outside of the app (restored account, watch-only account, account with revealed mnemonic). The existing algorithm was utilized but it also makes balance check, which adds unnecessary requests (at this point we know that there was no balance change per account).
This change addresses the problem by adding a separate nonce check before a full balance&nonce scan of the specific range. 

fix #4604

### DIFF
--- a/services/wallet/transfer/commands_sequential.go
+++ b/services/wallet/transfer/commands_sequential.go
@@ -142,6 +142,29 @@ func (c *findNewBlocksCommand) detectTransfers(parent context.Context, accounts 
 	return blockNum, addressesToCheck, nil
 }
 
+func (c *findNewBlocksCommand) detectNonceChange(parent context.Context, from, to *big.Int, accounts []common.Address) ([]common.Address, error) {
+	addressesWithChange := []common.Address{}
+	for _, account := range accounts {
+		oldNonce, err := c.balanceCacher.NonceAt(parent, c.chainClient, account, from)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


alaibe (approved):

services/wallet/transfer/commands_sequential.go
```diff
@@ -191,9 +214,18 @@ func (c *findNewBlocksCommand) Run(parent context.Context) error {
...
 	} else if c.iteration%nonceCheckIntervalIterations == 0 && len(accountsWithOutsideTransfers) > 0 {
-		c.findAndSaveEthBlocks(parent, c.fromBlockNumber, headNum, accountsWithOutsideTransfers)
+		accountsWithNonceChanges, err := c.detectNonceChange(parent, c.fromBlockNumber, headNum, accountsWithOutsideTransfers)
+		if err != nil {
+			c.error = err
+			return err
+		}
+
+		if len(accountsWithNonceChanges) > 0 {
+			c.findAndSaveEthBlocks(parent, c.fromBlockNumber, headNum, accountsWithNonceChanges)
```
alaibe, line 224: shouldn't we do that all the time as we know those block are "done"?

PR #4631, opened 2024-01-26
feat: New endpoint to return balances associated with all token criteria 

ilmotta:
## Summary

This PR adds a new endpoint `wakuext_getCommunityPermissionedBalances` that allows the mobile client (opened PR https://github.com/status-im/status-mobile/pull/18636) to show relevant balances when the user is selecting addresses to join a community.

To understand this feature, it's recommended to have a quick look at the user story in https://github.com/status-im/status-mobile/issues/18126.

If this implementation is deemed acceptable by your reviews, then I can document the endpoint, and potentially refactor more aggressively to reuse some logic from `protocol/communities/permission_checker.go#CheckPermissions`, because they're very similar. Just make the code a bit more DRY.

**Note:** *Relevant tokens* is a terminology used in Figma to describe "balances for all assets that have at least one Token Criteria associated, and for a given account address". I think "permissioned balances" captures some of that. Naming is hard, feel free to suggest something else that's still concise if possible.

### Demo

You can check the mobile PR to see the screenshot https://github.com/status-im/status-mobile/pull/18636

### Follow-up work

On the mobile side of things, we would like to have less logic on the client and more stuff in status-go. This PR is one such step, but we are considering superseding this new endpoint (or have another one) to also return the highest permission role to join, which in terms of logic, they do share quite a bit of code.

Support for ENS permissions in this endpoint will come at a later stage.


yqrashawn (approved):
LGTM!
Don't know if this fits desktop needs.

cammellos (approved):

mprakhov (approved):

PR #4632, opened 2024-01-26
feat: new endpoint added which returns accounts of all contacts that match the passed address

saledjenic:
We need this on the desktop app side, for the case when the user is about to add a new address to the saved addresses list, we need to check if that address belongs to any of his contacts, which means we need to check the list of profile showcases accounts of all of the users' contacts.

### DIFF
--- a/protocol/persistence_profile_showcase.go
+++ b/protocol/persistence_profile_showcase.go
@@ -3,6 +3,7 @@ package protocol
 import (
 	"context"
 	"database/sql"
+	"errors"
 )
 
 type ProfileShowcaseVisibility int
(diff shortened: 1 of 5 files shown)


protocol/persistence_profile_showcase.go
```diff
@@ -34,8 +35,8 @@ const selectContactProfileShowcaseCommunityQuery = "SELECT community_id, sort_or
 const removeContactProfileShowcaseCommunityQuery = "DELETE FROM profile_showcase_communities_contacts WHERE contact_id = ?"                                              // #nosec G101
 
 const upsertContactProfileShowcaseAccountQuery = "INSERT OR REPLACE INTO profile_showcase_accounts_contacts(contact_id, address, name, color_id, emoji, sort_order) VALUES (?, ?, ?, ?, ?, ?)" // #nosec G101
-const selectContactProfileShowcaseAccountQuery = "SELECT address, name, color_id, emoji, sort_order FROM profile_showcase_accounts_contacts WHERE contact_id = ?"                              // #nosec G101
-const removeContactProfileShowcaseAccountQuery = "DELETE FROM profile_showcase_accounts_contacts WHERE contact_id = ?"
+const selectContactProfileShowcaseAccountQuery = "SELECT * FROM profile_showcase_accounts_contacts WHERE contact_id = ?"                                                                       // #nosec G101
```
MishkaRogachev, line 38: n/a: i prefer explicit row names, because order should match `rows.Scan`

protocol/persistence_profile_showcase.go
```diff
@@ -392,25 +406,42 @@ func (db sqlitePersistence) saveProfileShowcaseAccountContact(tx *sql.Tx, contac
...
+func (db sqlitePersistence) getProfileShowcaseAccountsContact(tx *sql.Tx, contactID string) ([]*ProfileShowcaseAccount, error) {
+	rows, err := tx.Query(selectContactProfileShowcaseAccountQuery, contactID)
+	if err != nil {
+		return nil, err
+	}
+
+	return db.processProfileShowcaseAccounts(rows)
+}
+
+func (db sqlitePersistence) GetProfileShowcaseAccountsByAddress(address string) ([]*ProfileShowcaseAccount, error) {
```
MishkaRogachev, line 438: can you please add a test for a new function?

protocol/persistence_profile_showcase.go
```diff
@@ -392,25 +406,42 @@ func (db sqlitePersistence) saveProfileShowcaseAccountContact(tx *sql.Tx, contac
...
+		return nil, errors.New("rows is nil")
 	}
 
-	accounts := []*ProfileShowcaseAccount{}
-
 	for rows.Next() {
 		account := &ProfileShowcaseAccount{}
 
-		err := rows.Scan(&account.Address, &account.Name, &account.ColorID, &account.Emoji, &account.Order)
+		err = rows.Scan(&account.Address, &account.Name, &account.ColorID, &account.Emoji, &account.Order, &account.ContactID)
```
MishkaRogachev, line 417: i guess `&account.ContactID` should be first, that can be a reason [this](https://ci.status.im/job/status-go/job/prs/job/tests/job/PR-4632/1/testReport/github/com_status-im_status-go_protocol/TestProfileShowcasePersistenceSuite_TestProfileShowcaseContacts/) test fails

MishkaRogachev (commented):
A few questions

protocol/persistence_profile_showcase.go
saledjenic: I just wanted to avoid typing, cause we return all columns.

protocol/persistence_profile_showcase.go
saledjenic: Of course.

protocol/persistence_profile_showcase.go
saledjenic: No, it's not, but previously, `contact_id` was not included in the select query, not it is, end when you're comparing `s.Require().Equal(*showcase1.Accounts[i], *showcase1Back.Accounts[i])` then `showcase1` doesn't expect `contact_id` to be provided, but the `showcase1Back` provides it, cause I need it.

saledjenic:
@MishkaRogachev updated, test added.

MishkaRogachev (approved):
LGTM!

PR #4638, opened 2024-01-29
Fix/outdated community shard key msgs

osmaczko:
fix: ignore outdated COMMUNITY_SHARD_KEY messages 

This mitigates issue where community shard on client's side was not in
sync with owner's.

relates to: status-im/status-desktop#13217

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -1075,9 +1075,9 @@ func (m *Manager) DeleteCommunity(id types.HexBytes) error {
 	return m.persistence.DeleteCommunitySettings(id)
 }
 
-func (m *Manager) UpdateShard(community *Community, shard *shard.Shard) error {
+func (m *Manager) UpdateShard(community *Community, shard *shard.Shard, clock uint64) error {
 	community.config.Shard = shard
-	return m.persistence.SaveCommunityShard(community.ID(), shard, community.Clock())
+	return m.persistence.SaveCommunityShard(community.ID(), shard, clock)
 }
 
 // SetShard assigns a shard to a community
@@ -1092,7 +1092,7 @@ func (m *Manager) SetShard(communityID types.HexBytes, shard *shard.Shard) (*Com
 
 	community.increaseClock()
 
-	err = m.UpdateShard(community, shard)
+	err = m.UpdateShard(community, shard, community.Clock())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


protocol/communities/manager.go
```diff
@@ -1075,9 +1075,9 @@ func (m *Manager) DeleteCommunity(id types.HexBytes) error {
 	return m.persistence.DeleteCommunitySettings(id)
 }
 
-func (m *Manager) UpdateShard(community *Community, shard *shard.Shard) error {
+func (m *Manager) UpdateShard(community *Community, shard *shard.Shard, clock uint64) error {
```
igor-sirotin, line 1078: Why is `clock` needed here as an argument? 🤔 
From what I see you always use `community.Clock()` as a value when calling `UpdateShard`.

protocol/messenger.go
```diff
@@ -178,6 +178,9 @@ type Messenger struct {
 	// used to track dispatched messages
 	dispatchMessageTestCallback func(common.RawMessage)
 
+	// used to track unhandled messages
+	unhandledMessagesTracker func(*v1protocol.StatusMessage, error)
```
igor-sirotin, line 182: 👍

protocol/messenger_builder_test.go
```diff
@@ -90,10 +94,37 @@ func newTestMessenger(waku types.Waku, config testMessengerConfig, extraOptions
...
+	if _, exists := u.messages[msgType]; !exists {
+		u.messages[msgType] = []*unhandedMessage{}
+	}
+
+	newMessage := &unhandedMessage{
+		StatusMessage: msg,
+		err:           err,
+	}
+	u.messages[msgType] = append(u.messages[msgType], newMessage)
+}
```
igor-sirotin, line 130: Not a lot of code, but maybe move this to a separate file?
It would also make sense to use `unhandedMessage` as an argument in `addMessage`

igor-sirotin (approved):

protocol/messenger_builder_test.go
osmaczko: > Not a lot of code, but maybe move this to a separate file?

Was thinking about it but wasn't able to come up with proper name :laughing: This `_test.go` suffix makes test utilities very hard to name. I will try to come up with something in follow-up maybe?

> It would also make sense to use `unhandedMessage` as an argument in `addMessage`

I can't, `addMessage` must fulfill the callback interface defined in Messenger::unhandledMessagesTracker.

protocol/communities/manager.go
osmaczko: Ooops, you are right, very good catch! I overlooked one place. Will fire the PR in a sec.

osmaczko: @igor-sirotin => https://github.com/status-im/status-go/pull/4643

PR #4642, opened 2024-01-29
fix: update display name endpoint

mohsen-ghafouri:
a fix for this issue https://github.com/status-im/status-mobile/pull/18440#issuecomment-1906004966

created new endpoint for update display name


siddarthkay (approved):

protocol/messenger_identity.go
```diff
@@ -90,6 +91,44 @@ func (m *Messenger) SetDisplayName(displayName string) error {
 	return m.publishContactCode()
 }
 
+func (m *Messenger) UpdateDisplayName(request *requests.UpdateDisplayName) error {
```
saledjenic: @mohsen-ghafouri isn't this function the same as this one we already have?

https://github.com/status-im/status-go/blob/develop/protocol/messenger_identity.go#L55-L91

There is also this API endpoint for using it https://github.com/status-im/status-go/blob/ac81c15805869cab7108c2d332dca3a9b8a0673d/services/ext/api.go#L960

saledjenic (changes requested):
@mohsen-ghafouri I would rather not duplicate the thing.

protocol/messenger_identity.go
siddarthkay: using this method also updates the image which causes a weird bug in mobile , see -> https://github.com/status-im/status-mobile/pull/18440#issuecomment-1906004966

saledjenic: Then, that only makes me think that the function that is used to store images doesn't update `m.account.Images`, then the bug should be fixed there.

saledjenic: Another approach is to use this line `err = m.multiAccounts.UpdateAccount(*m.account)` instead of this one  `err = m.multiAccounts.SaveAccount(*m.account)` in `func (m *Messenger) SetDisplayName(displayName string) error` function. So in that case that's a single-line fix.

siddarthkay: no it does, if you see the conversation in the mobile PR,
This weird edge case only happens when a user logs out and logs back in. otherwise this edge case does not happen.

mohsen-ghafouri: yeah it make sense, it's better to fix `m.account.Images` in `StoreIdentityImages`. let me check

mohsen-ghafouri: @saledjenic could you please check again? I just resolve the issue in `SetDisplayName`

saledjenic (approved):

protocol/messenger_identity.go
saledjenic: @mohsen-ghafouri thanks.

PR #4643, opened 2024-01-29
fix: use CommunityShardKey's clock rather than community's clock

osmaczko:


igor-sirotin (approved):

osmaczko:
I'll add a test covering that case later this week.

PR #4654, opened 2024-01-31
feat(wallet) use sql activity filter for incremental updates

stefandunca:
### Updates status-desktop [#12120](https://github.com/status-im/status-desktop/issues/12120)

Switch from the prototype of duplicating the SQL filter as a runtime and keeping them in sync on each event that might invalidate the current filtered entries to a simpler approach of requesting the filter again and doing the diff to detect the new changes.

Also, add a new reset API to model the new entries design requirements.

The new approach shows fewer corner cases to handle and follows one source of truth, making debugging and future maintenance easier.

Other changes

- Fix pending mocking to work with multiple calls
- Refactor tests to account for the new changes

### DIFF
--- a/services/wallet/activity/session.go
+++ b/services/wallet/activity/session.go
@@ -2,13 +2,12 @@ package activity
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
+	"strconv"
 
 	"golang.org/x/exp/slices"
 
 	eth "github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/event"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/status-im/status-go/services/wallet/async"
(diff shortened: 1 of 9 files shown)


services/wallet/activity/session.go
```diff
@@ -18,12 +17,28 @@ import (
...
 
+const nilStr = "nil"
+
 type EntryIdentity struct {
 	payloadType PayloadType
 	transaction *transfer.TransactionIdentity
 	id          transfer.MultiTransactionIDType
 }
 
+// func (e EntryIdentity) same(a EntryIdentity) bool {
```
dlipicar, line 28: can this be removed?

stefandunca: I revived and used it in one of my following PRs :)

dlipicar (approved):

services/wallet/activity/session.go
Cuteivist, line 1: There's a lot of commented functions here. Is there purpose to leave them right now ?

Cuteivist (approved):
Looks good

services/wallet/activity/session.go
Cuteivist: Nvm I see they are removed in next PR

PR #4665, opened 2024-02-01
Simplify FilterManager

### LINKED ISSUE #4659: bug: Filter subscribe and unsubscribe seem to happen out of order
# Problem

Filter unsubscribe and subscribe happened out of order causing status-mobile to think it is subscribed to filter where as Fleet node thinks subscription is unsubscribed.
Reference discussion [here](https://github.com/status-im/status-go/issues/4459#issuecomment-1916144642)

# Acceptance Criteria

Where-ever there is Filter unsubscribe and subscribe, these should happen in sequence.

vitvly:
Fixes https://github.com/status-im/status-go/issues/4659

After some looking on FilterManager, it appears that it's too complicated. Since the logic for setting specific peers has been removed in favour of automatic selection, i don't actually see any need to maintain any knowledge about particular peers on FilterManager level. Also, maintaining `MinPeersForFilter` connections for each filter is also out of status-go's purview, i think. Let go-waku handle that.

There are some changes to be made in go-waku as well (will push separately):
  - `IsSubscriptionAlive` checks have to be smart enough in order to not to issue too many requests to same peers (multiple subs can share identical peers)
  - probably subscription channels returned to status-go have to be multiplexed in case of multiple peers. There should be a 1-to-1 mapping between filters and subscriptions in status-go.
 
List of other changes to `FilterManager`:
  - dropped lots of code there
  - async is there only for multiple filters
  - all operations for a single filter occur in a synchronous fashion, in order to guarantee proper ordering.


wakuv2/filter_manager.go
```diff
@@ -53,225 +28,95 @@ type FilterEvent struct {
...
-	case FilterEventGetStats:
-		stats := make(FilterSubs)
-		for id, subs := range mgr.filterSubs {
-			stats[id] = make(subscription.SubscriptionSet)
-			for subID, sub := range subs {
-				if sub == nil {
-					// Skip temp subs
-					continue
-				}
+		if sub == nil {
```
kaichaosun: Should we merge this condition into the above `else` branch?

wakuv2/filter_manager.go
```diff
@@ -284,34 +129,34 @@ func (mgr *FilterManager) buildContentFilter(pubsubTopic string, contentTopicSet
...
-	if !found {
-		mgr.logger.Error("resubscribe filter not found", zap.String("filterId", filterID))
-		return
-	}
-	if len(subs) > mgr.settings.MinPeersForFilter {
-		mgr.logger.Error("filter resubscribe too many subs", zap.String("filterId", filterID), zap.Int("len", len(subs)))
+func (mgr *FilterManager) subscribe(ctx context.Context, logger *zap.Logger, contentFilter protocol.ContentFilter) *subscription.SubscriptionDetails {
+	reqCtx, cancel := context.WithTimeout(ctx, pingTimeout)
+	defer cancel()
+	subDetails, err := mgr.node.Subscribe(reqCtx, contentFilter, filter.WithAutomaticPeerSelection())
```
kaichaosun: I'm wondering how the minimum subscribe peers is handled, maybe relate to this PR? https://github.com/waku-org/go-waku/pull/1005

kaichaosun (approved):

chaitanyaprem:
> After some looking on FilterManager, it appears that it's too complicated. Since the logic for setting specific peers has been removed in favour of automatic selection, i don't actually see any need to maintain any knowledge about particular peers on FilterManager level. Also, maintaining `MinPeersForFilter` connections for each filter is also out of status-go's purview, i think. Let go-waku handle that.

Ideally such functionality should be part of the SDK layer, it could be a thin layer above the actual Filter implementation in go-waku. Would not recommend adding it to the Filter code in go-waku as it has become too complicated. 

Wrt MinPeersForFilter, status-go should specify the count for it as the idea is for users for Filter to be able to specify and choose based on the use-case. 
Alternatively, we could always set the default to 2(which is what was being used by status-go).

wakuv2/filter_manager.go
```diff
@@ -53,225 +28,95 @@ type FilterEvent struct {
...
-				go mgr.runFilterSubscriptionLoop(ev.sub)
+	for {
+		logger.Debug("filter lifecycle started")
+		if sub != nil {
+			logger.Debug("filter health check")
+			err := mgr.isFilterSubAlive(ctx, sub)
+			if err != nil {
+				logger.Debug("filter ping error", zap.Error(err))
+				mgr.unsubscribe(ctx, logger, sub)
+				sub = nil
```
chaitanyaprem: Wondering it this identification of subscription not being alive should trigger some mechanism to fetch messages that could have been lost until we subscribe again for this Filter either via mailserver logic. 

Or will peer-syncing take care of such scenarios?

wakuv2/filter_manager.go
```diff
@@ -53,225 +28,95 @@ type FilterEvent struct {
...
-			for subID, sub := range subs {
-				if sub == nil {
-					// Skip temp subs
-					continue
-				}
+		if sub == nil {
+			// We are here either when the filter has just been added,
+			// or when ping failed
+			logger.Debug("filter try subscribe")
+			sub = mgr.subscribe(ctx, logger, contentFilter)
```
chaitanyaprem: What if filter subscription fails? 
Shouldn't there be some sort of retry for the failed `filterID` or is that driven outside of filterManager?

wakuv2/filter_manager.go
```diff
@@ -53,225 +28,95 @@ type FilterEvent struct {
...
-				}
-				mgr.eventChan <- FilterEvent{eventType: FilterEventPingResult, peerID: sub.PeerID, success: alive}
-			}(sub)
-		}
+func (mgr *FilterManager) removeFilter(filterID string) {
+	cancel, ok := mgr.filters[filterID]
+	if ok {
+		delete(mgr.filters, filterID)
+		// close goroutine running filterLifecycle
+		cancel()
```
chaitanyaprem: Shouldn't unsubscribe be called for subscription associated withthis particular filter?

wakuv2/filter_manager.go
vitvly: If subscription fails, `sub` will be `nil`, and re-subscribe will be attempted once ticker emits an event. So we will land at the same line of code.

wakuv2/filter_manager.go
vitvly: It will be called via `cancel()`, as `filterLifecycle` listens to `ctx.Done()` channel. Once it emits an event, `unsubscribe` is called (line 96).

wakuv2/filter_manager.go
vitvly: In that case we are not going to attempt re-subscribing after failed ping. Suppose `sub != nil` when we enter the loop. We invoke `isFilterSubAlive` and it fails. In case when we merge the condition with `else`, we'll skip resubscribing until the next `ticker` event. The code will be correct, but there will be a delay.

wakuv2/filter_manager.go
vitvly: This is a good point @kaichaosun . For now i've decided to not handle it at all - to be coded separately. As i understand, this might become part of the chat-sdk layer. The problem is that with a `WithMaxPeersPerContentFilter` subscribe option, `Subscribe` will return a slice of subscriptions. I don't want to handle that in `FilterManager` any more - i'd rather have a single subscription, with a single channel, and let that intermediate layer between status-go and chat-sdk handle it.

vitvly: Please let me know what you think, and if i correctly understand the go-waku code you've referenced.

vitvly:
@chaitanyaprem regarding `MinPeersForFilter` -  i think @kaichaosun raised a similar point. Just to reiterate - if i understand correctly, using `WithMaxPeersPerContentFilter` will return a slice of subscriptions, and i'd rather have a single sub in status-go. In the issue description above item #2 plans to address that, i think. Please correct me if my understanding is wrong.

vitvly:
This, i think, implies, that the notion of `Subscription` for status-go should only encompass its channel `C`. Current `SubscriptionDetails` has extra exported fields that status-go shouldn't know about.

wakuv2/filter_manager.go
vitvly: This is something i don't have answer for. Please let me know if the below is correct:
  - Once we reconnect to a healthy peer, we're only going to get freshly delivered messages
  -  For the messages lost, we have to rely on store node

chaitanyaprem:
> @chaitanyaprem regarding `MinPeersForFilter` - i think @kaichaosun raised a similar point. Just to reiterate - if i understand correctly, using `WithMaxPeersPerContentFilter` will return a slice of subscriptions, and i'd rather have a single sub in status-go. In the issue description above item #2 plans to address that, i think. Please correct me if my understanding is wrong.

Hmm, interesting point. Haven't thought about it this way, but kind of makes sense to have SDK abstract this from status-go.  Have to think a little more if this brings any unnecessary complexities. 
cc @richard-ramos WDYT?

richard-ramos:
Indeed, having a single subscription would make sense for status-go, since it makes dealing with multiple peers something simpler.  Like @chaitanyaprem , i think this should go a layer above the actual filter protocol implementation, since that's 'raw' protocol usage, and it's code in this moment is somewhat complex, while a separate abstraction could help make its usage to be more palatable, while offering an API that would make sense for status-go.

vitvly:
When testing this PR in status-mobile a PeerExchange problem was uncovered: https://github.com/status-im/status-mobile/pull/18769#issuecomment-1936038794. Debugging/troubleshooting notes here: https://github.com/status-im/status-mobile/pull/18769#issuecomment-1936053739.

wakuv2/waku.go
```diff
@@ -218,17 +218,26 @@ func New(nodeKey string, fleet string, cfg *Config, logger *zap.Logger, appDB *s
 		onHistoricMessagesRequestFailed: onHistoricMessagesRequestFailed,
 		onPeerStats:                     onPeerStats,
 	}
+	enablePeerExchange := false
+	enableDiscv5 := false
+	if cfg.LightClient {
+		enablePeerExchange = true
```
chaitanyaprem: This enables server side of peerExchange which is not required in case of LightClient. 
But in relay-mode where discv5 is enabled, we may want to enable this. 

We need to make sure peerExchange client is running in lightMode, which i did not see any setting here. Rather the start of peerExchangeLoop should happen which queries peerExchange nodes in order to retrieve peers rather than using discv5.

chaitanyaprem: If we want to have different configs to run peerExchange client and server that is also fine. But right now it looks like same config is being used for both which doesn't make sense. 

PeerExchange server makes sense only when discv5 is running locally as it will have a cache of peers that it can respond to in case a PeerExchange request comes from the PeerExchangeClient.

Also looks like some of my changes to consider bootNodes as peerExchange peers have been over-ridden which must be causing the issue reported in mobile-CI where there are no nodes identified for lightpush/filter.

chaitanyaprem: The solution is to: 
-  add the bootnodes as peers and let identify protocol run (by inviking `identifyAndConnect` method. Post this, the peerExchange loop will start functioning properly and should be able to discover peers. 
- Segregate config for PeerExchangeClient and Server and run client in case of lightMode and server in case of relayMode

chaitanyaprem: Fixed peer-exchange client related changes in the branch [fix/peer-exchange-client](https://github.com/status-im/status-go/tree/fix/peer-exchange-client). 
I don't know how to set new config and what impact it has in other places wrt persisting config etc, hence left that part out which needs to be addressed.

chaitanyaprem: cc @vitvly

chaitanyaprem: Already addressed as part of another PR https://github.com/status-im/status-go/pull/4826 , so we dont need these changes anymore.

chaitanyaprem (approved):
LGTM

Can be merged once https://github.com/waku-org/go-waku/pull/1048 is ready and CI errors are addressed.

chaitanyaprem:
> LGTM
> 
> Can be merged once [waku-org/go-waku#1048](https://github.com/waku-org/go-waku/pull/1048) is ready and CI errors are addressed.

@vitvly the relevant go-waku PR has been merged. This also can be merged once CI errors are addressed. We can test Filter in CI for any issues.

chaitanyaprem:
@vitvly Since i started working on optimizing Filter pings, i had noticed few more issues/enhancements that are required for Filter Management to effectively function. I am making those changes as part of https://github.com/waku-org/go-waku/pull/1102

I would suggest to wait for this PR to be completed and then we can use that version to run tests.

chaitanyaprem:
Waiting for https://github.com/status-im/status-go/pull/5213 to get merged as the Filter Test seems to be failing which can be fixed by this PR. 
Once the above PR is merged, then will retest Filter test to confirm there are no issues and then initiate dogfooding for this PR.

chaitanyaprem:
@richard-ramos I had noticed you haven't taken a look at the updated PR. Please review and add any comments.

wakuv2/waku_test.go
```diff
@@ -313,89 +326,92 @@ func TestWakuV2Filter(t *testing.T) {
 	if envEnrTreeAddress != "" {
 		enrTreeAddress = envEnrTreeAddress
 	}
-
+	//enrTreeAddress = "/ip4/100.118.238.47/tcp/30304/p2p/16Uiu2HAmBUejhH3b9nqTkSyj3TJck9AoMZoVJ2WYbuakc7MhU5cm"
```
richard-ramos:
```suggestion
```

wakuv2/waku_test.go
```diff
@@ -313,89 +326,92 @@ func TestWakuV2Filter(t *testing.T) {
...
-
+	testPubsubTopic := "/waku/2/rs/16/32"
 	filter := &common.Filter{
 		Messages:      common.NewMemoryMessageStore(),
+		PubsubTopic:   testPubsubTopic,
 		ContentTopics: common.NewTopicSetFromBytes([][]byte{[]byte{1, 2, 3, 4}}),
 	}
 
-	filterID, err := w.Subscribe(filter)
+	fmt.Println("### Subscribe")
```
richard-ramos:
```suggestion
```
or use logs instead

wakuv2/waku_test.go
```diff
@@ -313,89 +326,92 @@ func TestWakuV2Filter(t *testing.T) {
...
 		ContentTopic: contentTopic.ContentTopic(),
 		Version:      proto.Uint32(0),
 		Timestamp:    &msgTimestamp,
 	})
 	require.NoError(t, err)
 
-	time.Sleep(15 * time.Second)
+	time.Sleep(5 * time.Second)
 
+	fmt.Println("### Check")
```
richard-ramos:
```suggestion
```

wakuv2/waku_test.go
```diff
@@ -313,89 +326,92 @@ func TestWakuV2Filter(t *testing.T) {
...
-	isFilterSubAliveBak := w.filterManager.isFilterSubAlive
-	w.filterManager.config.MinPeersForFilter = 0
-	w.filterManager.isFilterSubAlive = func(sub *subscription.SubscriptionDetails) error {
-		return errors.New("peer down")
-	}
+	_, err = w.node.FilterLightnode().UnsubscribeWithSubscription(w.ctx, subscriptions[0])
+	require.NoError(t, err)
 
-	time.Sleep(5 * time.Second)
+	//subscriptions[0].Close()
```
richard-ramos:
```suggestion
```

wakuv2/waku.go
```diff
@@ -1227,6 +1218,49 @@ func (w *Waku) Query(ctx context.Context, peerID peer.ID, query legacy_store.Que
 	return result.Cursor(), len(result.Messages), nil
 }
 
+func (w *Waku) lightClientConnectionStatus() {
+
+	peers := w.node.Host().Network().Peers()
+	//peersByPubSubTopic := w.node.Host().Peerstore().(wps.WakuPeerstore).PeersByPubSubTopics(w.cfg.DefaultShardedPubsubTopics)
```
richard-ramos:
```suggestion
```

wakuv2/waku.go
```diff
@@ -1227,6 +1218,49 @@ func (w *Waku) Query(ctx context.Context, peerID peer.ID, query legacy_store.Que
...
+	w.logger.Debug("peer stats",
+		zap.Int("peersCount", len(peers))) //, zap.Int("peersByPubSubTopic", len(peersByPubSubTopic)))
```
richard-ramos:
```suggestion
		zap.Int("peersCount", len(peers))) 
```

wakuv2/filter_manager.go
```diff
@@ -51,223 +25,114 @@ type FilterEvent struct {
...
-				continue
-			}
-			distinctPeers[sub.PeerID] = struct{}{}
-			logger.Debug("filter ping peer", zap.Stringer("peerId", sub.PeerID))
-			go func(sub *subscription.SubscriptionDetails) {
-				err := mgr.isFilterSubAlive(sub)
-				alive := err == nil
+	}
+	//if !newStatus && mgr.peersAvailable || newStatus && !mgr.peersAvailable { //Offline or online
+	//mgr.logger.Info("node status change, notifying all filter subscriptions", zap.Bool("new-status", newStatus))
```
richard-ramos:
```suggestion
```

wakuv2/filter_manager.go
```diff
@@ -51,223 +25,114 @@ type FilterEvent struct {
...
+	mgr.Lock()
+	for _, subDetails := range mgr.filters {
+		subDetails.sub.SetNodeState(newStatus)
+	}
+	mgr.Unlock()
+	//}
```
richard-ramos:
```suggestion
```

richard-ramos (approved):
Good work. I like the changes done on filter manager.

chaitanyaprem:
Disabled pxClient for now as that is causing issues. This is being tracked as part of https://github.com/status-im/status-go/issues/5344 so that we can enable it again in lightClients.

chaitanyaprem:
Merging this PR as it has passed mobile e2e tests and also manual test with some observations which should not be related to filter.

Ref status-mobile PR : https://github.com/status-im/status-mobile/pull/20369

PR #4666, opened 2024-02-01
[#4630] Add wallet_fetchOrGetCachedWalletBalances method

### LINKED ISSUE #4630: Request ethscan balance checks only on detected transfers
Currently we check balances using ethscan contract with some interval (10 minutes), that produces a lot of unnecessary RPC calls
```
// ~16h, desktop
 eth_CallContract_0x08A8fDBddc160A7d5b957256b903dCAb1aE512C5: 1079, //ethscan
  eth_CallContract_0x9e5076DF494FC949aBc4461F4E57592B81517D81: 1079, //ethscan optimism
  eth_CallContract_0xBB85398092B83A016935a17Fc857507b7851a071: 1079, //ethscan arbitrum
  eth_CodeAt: 887, // likely ethscan

  total eth_ calls: 7971
  ethscan calls: 4219 (~53%)
```

`eth_CodeAt` part was fixed separately here https://github.com/status-im/status-go/pull/4623

`eth_CallContract` calls should be executed only on tx detection (maybe also once a day or something, but that shouldn't be necessary).

rasom:
fixes #4630

Balances are fetched from the database, if needed are fetched via RPC. The cache is invalidated after detection of any transaction.


rasom:
just noticed we do have all that data in the database

IvanBelyakoff (commented):
Roman, I am curious if we should invalidate cache for specific address/es + networks?

rasom:
@IvanBelyakoff we can, we also can check block numbers, we can block concurrent `wallet_fetchOrGetCachedWalletBalances` requests till the one which already runs will end (to make sure we do not request same balances multiple times), etc. That would only save us like very few requests but will definitely add some complexity and potential bugs with cache invalidation. I'm not sure it is worth the effort and potential problems. 

Considering a low impact of it I would add it separately (if needed) while keeping current change small and simple.

alaibe (approved):

IvanBelyakoff (approved):

PR #4667, opened 2024-02-02
fix: add node config `RuntimeLogLevel` option

### LINKED ISSUE #13139: It's not possible to switch the log level
In the advanced settings section, if using a local build, the DEBUG flag is always active, and switching it makes no difference (i tried restarting the app). The same happens in release version, except that the log level is always INFO and it's not possible to switch the DEBUG flag

I'm not entirely sure, but I believe this code might be causing the issue: https://github.com/status-im/status-desktop/blob/a2aedb5c6a699c154382f75d9c5a7e6e641fa14c/src/app/boot/app_controller.nim#L137-L143

cc: @chaitanyaprem

kounkou:
### Description

Closes https://github.com/status-im/status-desktop/issues/13139

linked to https://github.com/status-im/status-desktop/pull/13309

Previously it was not possible to change the state of the Debug toggle. This is because the code forced the setting the default value, ignoring the database setup, hence always setting the DEBUG as LogLevel.

This PR adds :
- A RuntimeLogLevel to enable setting ephemeral loggig strategy on status-go

### DIFF
--- a/api/geth_backend.go
+++ b/api/geth_backend.go
@@ -1712,6 +1712,10 @@ func (b *GethStatusBackend) loadNodeConfig(inputNodeCfg *params.NodeConfig) erro
 
 	b.config = conf
 
+	if inputNodeCfg != nil && inputNodeCfg.RuntimeLogLevel != "" {
+		b.config.LogLevel = inputNodeCfg.RuntimeLogLevel
+	}
+
 	return nil
 }
 
--- a/params/config.go
+++ b/params/config.go
@@ -412,6 +412,9 @@ type NodeConfig struct {
 	// LogFile is filename where exposed logs get written to
 	LogFile string
 
+	// RuntimeLoglevel defines minimum log level for this session only, not affecting the db-stored node configuration
+	RuntimeLogLevel string `validate:"omitempty,eq=ERROR|eq=WARN|eq=INFO|eq=DEBUG|eq=TRACE"`
+
 	// LogLevel defines minimum log level. Valid names are "ERROR", "WARN", "INFO", "DEBUG", and "TRACE".
 	LogLevel string `validate:"eq=ERROR|eq=WARN|eq=INFO|eq=DEBUG|eq=TRACE"`
 
--- a/api/backend_test.go
+++ b/api/backend_test.go
@@ -696,6 +696,74 @@ func TestBackendGetVerifiedAccount(t *testing.T) {
 	})
 }
 
+func TestRuntimeLogLevelIsNotWrittenToDatabase(t *testing.T) {
+	utils.Init()
+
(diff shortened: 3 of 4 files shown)


api/geth_backend.go
```diff
@@ -476,6 +476,10 @@ func (b *GethStatusBackend) startNodeWithKey(acc multiaccounts.Account, password
 		return err
 	}
 
+	if inputNodeCfg.RuntimeLogLevel != "" {
+		b.config.RuntimeLogLevel = inputNodeCfg.RuntimeLogLevel
+	}
+
```
igor-sirotin: Maybe we could move this to `loadNodeConfig`? 
It would make sense to make it there, as it's used not only in this function.

params/config.go
```diff
@@ -412,6 +412,9 @@ type NodeConfig struct {
 	// LogFile is filename where exposed logs get written to
 	LogFile string
 
+	// RuntimeLoglevel enables status-go to update the logging strategy without updating the database
```
igor-sirotin:
```suggestion
	// RuntimeLoglevel defines minimum log level for this session only, not affecting the db-stored node configuration
```

igor-sirotin (commented):
Thanks for the test! 💪

api/geth_backend.go
kounkou: > Maybe we could move this to `loadNodeConfig`? It would make sense to make it there, as it's used not only in this function.

Having a look...

igor-sirotin (approved):
Please wait for @osmaczko to review

params/config.go
```diff
@@ -412,6 +412,9 @@ type NodeConfig struct {
 	// LogFile is filename where exposed logs get written to
 	LogFile string
 
+	// RuntimeLoglevel defines minimum log level for this session only, not affecting the db-stored node configuration
+	RuntimeLogLevel string `validate:"omitempty,eq=ERROR|eq=WARN|eq=INFO|eq=DEBUG|eq=TRACE"`
```
osmaczko, line 416: How does it work? I don't see it being used anywhere other than in tests :thinking:

kounkou: > How does it work? I don't see it being used anywhere other than in tests 🤔

Huh, made some mistakes, fixing this.

-----
On the go side, it's used to set the LogLevel that will be used by zap to set the log level

igor-sirotin: We set it from desktop if any runtime options are found:
https://github.com/status-im/status-desktop/pull/13309/files#diff-92c30cc52000ad1754ec5845cb656b2d8fe5a050db412429093cbdd9bb83d2e1

osmaczko (approved):
lgtm now

PR #4669, opened 2024-02-02
fix: sometimes channels do not randomly appear in test community afte…

### LINKED ISSUE #18455: Sometimes channels do not randomly appear in test community after its creation
# Bug Report

## Problem 

Sometimes channels do not randomly appear in the test community after its creation, this leads to failures in e2e tests and is also reproducible manually.

Example:
https://github.com/status-im/status-mobile/pull/18374#issuecomment-1884660314 in `1. test_community_hashtag_links_to_community_channels, id: 702948`
On the [video](https://app.eu-central-1.saucelabs.com/tests/f4c8fcc2d8b145fbbd06c0718dd7cf14?auth=cf2405c3f14cb355888676d06d06f72c#245) (14:19) you can see that the `#general` channel is not displayed on the channel list and this caused the test to fail.

### Reproduction

1. Install any PR build 
2. Open the `Communities` screen
3. Tap `+`
4. `Create closed community`
5. Open the community

You may have to repeat these steps several times until the community opens without one of the channels.
Full list of channels: #general, #cats, #dogs, #rules

#### Expected behavior
Screenshot 2024-01-10 at 18 23 08

#### Actual behavior

No `#rules` channel
Screenshot 2024-01-10 at 18 23 54

### Logs

[geth.log.zip](https://github.com/status-im/status-mobile/files/13891149/geth.log.zip)
[Status.log.zip](https://github.com/status-im/status-mobile/files/13891150/Status.log.zip)

### Additional Information

- Status version: nightly (only for x86 builds used for e2e), PR builds

qfrank:
[TestCreateClosedCommunity](https://github.com/status-im/status-go/blob/3ae9ee9c6e19e5ca8cbae8ecf20889da070ef092/protocol/messenger_community_for_mobile_testing_test.go#L31) 
can reproduce the mobile issue before this PR if running `TestCreateClosedCommunity` for 1000 times.

[Relate comment](https://github.com/status-im/status-mobile/issues/18455#issuecomment-1921294314)

Closes [mobile issue](https://github.com/status-im/status-mobile/issues/18455)

### DIFF
--- a/protocol/communities/manager.go
+++ b/protocol/communities/manager.go
@@ -102,6 +102,54 @@ type Manager struct {
 	RekeyInterval                    time.Duration
 	PermissionChecker                PermissionChecker
 	keyDistributor                   KeyDistributor
+	communityLock                    *CommunityLock
+}
+
+type CommunityLock struct {
+	logger *zap.Logger
+	locks  map[string]*sync.Mutex
+	mutex  sync.Mutex
+}
+
+func NewCommunityLock(logger *zap.Logger) *CommunityLock {
+	return &CommunityLock{
+		logger: logger.Named("CommunityLock"),
+		locks:  make(map[string]*sync.Mutex),
+	}
+}
+
+func (c *CommunityLock) Lock(communityID types.HexBytes) {
+	c.mutex.Lock()
+	communityIDStr := types.EncodeHex(communityID)
+	lock, ok := c.locks[communityIDStr]
+	if !ok {
+		lock = &sync.Mutex{}
+		c.locks[communityIDStr] = lock
+	}
+	c.mutex.Unlock()
+
+	lock.Lock()
+}
+
+func (c *CommunityLock) Unlock(communityID types.HexBytes) {
+	c.mutex.Lock()
+	communityIDStr := types.EncodeHex(communityID)
+	lock, ok := c.locks[communityIDStr]
+	c.mutex.Unlock()
+
+	if ok {
+		lock.Unlock()
+	} else {
+		c.logger.Warn("trying to unlock a non-existent lock", zap.String("communityID", communityIDStr))
+	}
+}
+
+func (c *CommunityLock) Init() {
+	c.locks = make(map[string]*sync.Mutex)
+}
+
+func (c *CommunityLock) Release() {
+	c.locks = nil
 }
 
 type HistoryArchiveDownloadTask struct {
@@ -263,6 +311,7 @@ func NewManager(identity *ecdsa.PrivateKey, installationID string, db *sql.DB, e
 		torrentTasks:                make(map[string]metainfo.Hash),
 		historyArchiveDownloadTasks: make(map[string]*HistoryArchiveDownloadTask),
 		keyDistributor:              keyDistributor,
+		communityLock:               NewCommunityLock(logger),
 	}
 
 	manager.persistence = &Persistence{
@@ -368,6 +417,7 @@ func (m *Manager) Subscribe() chan *Subscription {
 
 func (m *Manager) Start() error {
 	m.stopped = false
+	m.communityLock.Init()
 	if m.ensVerifier != nil {
 		m.runENSVerificationLoop()
 	}
@@ -514,6 +564,7 @@ func (m *Manager) Stop() error {
 		close(c)
 	}
 	m.StopTorrentClient()
+	m.communityLock.Release()
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


qfrank:
need continue work:

```
2024-02-02T11:34:12.220Z	DEBUG	transport/transport.go:611	confirming message processed	{"Transport": {"ids": ["0x4e8cdb4aee798c29bd7a9f0de8030eadcc3c080f999ac05d8ae51f4451afa8ba"], "timestamp": 1706873652220}}
    communities_messenger_helpers_test.go:401: 
        	Error Trace:	/home/jenkins/workspace/status-go_prs_tests_PR-4669/protocol/communities_messenger_helpers_test.go:401
        	            				/home/jenkins/workspace/status-go_prs_tests_PR-4669/protocol/communities_events_utils_test.go:128
        	            				/home/jenkins/workspace/status-go_prs_tests_PR-4669/protocol/communities_events_owner_without_community_key_test.go:276
        	            	community not found
        	Test:       	TestOwnerWithoutCommunityKeyCommunityEventsSuite/TestMemberReceiveOwnerEventsWhenControlNodeOffline
```

osmaczko (commented):
If I understand correctly, the root cause is a race condition. The solution looks like a hack/workaround. Why would we want to save a community without a description?

osmaczko (changes requested):
Perhaps we could prevent race conditions with synchronization primitives. `map[communityID][Mutex]` is a good candidate. What do you think?

qfrank:
> If I understand correctly, the root cause is a race condition. The solution looks like a hack/workaround. Why would we want to save a community without a description?

this is the easiest way to fix to unblock qa e2e test, but yeah, we should take care of race condition in the community manager level, i'll create another PR to take care of it in community manager level. WDYT? @osmaczko

qfrank:
> need continue work:
> 
> ```
> 2024-02-02T11:34:12.220Z	DEBUG	transport/transport.go:611	confirming message processed	{"Transport": {"ids": ["0x4e8cdb4aee798c29bd7a9f0de8030eadcc3c080f999ac05d8ae51f4451afa8ba"], "timestamp": 1706873652220}}
>     communities_messenger_helpers_test.go:401: 
>         	Error Trace:	/home/jenkins/workspace/status-go_prs_tests_PR-4669/protocol/communities_messenger_helpers_test.go:401
>         	            				/home/jenkins/workspace/status-go_prs_tests_PR-4669/protocol/communities_events_utils_test.go:128
>         	            				/home/jenkins/workspace/status-go_prs_tests_PR-4669/protocol/communities_events_owner_without_community_key_test.go:276
>         	Error:      	Received unexpected error:
>         	            	community not found
>         	Test:       	TestOwnerWithoutCommunityKeyCommunityEventsSuite/TestMemberReceiveOwnerEventsWhenControlNodeOffline
> ```

fixed.

osmaczko:
> > If I understand correctly, the root cause is a race condition. The solution looks like a hack/workaround. Why would we want to save a community without a description?
> 
> this is the easiest way to fix to unblock qa e2e test, but yeah, we should take care of race condition in the community manager level, i'll create another PR to take care of it in community manager level. WDYT? @osmaczko

I would prefer to avoid this workaround being merged. e2e test can be disabled until it is fixed properly.

qfrank:
I'll make a non-hack solution then, thank you @osmaczko 

> Perhaps we could prevent race conditions with synchronization primitives. `map[communityID][Mutex]` is a good candidate. What do you think?

osmaczko:
> I'll make a non-hack solution then, thank you @osmaczko
> 
> > Perhaps we could prevent race conditions with synchronization primitives. `map[communityID][Mutex]` is a good candidate. What do you think?

Thank you @qfrank, I really appreciate that!

qfrank:
> > I'll make a non-hack solution then, thank you @osmaczko
> > > Perhaps we could prevent race conditions with synchronization primitives. `map[communityID][Mutex]` is a good candidate. What do you think?
> 
> Thank you @qfrank, I really appreciate that!

ready to review 🙂 @osmaczko

protocol/communities/manager.go
```diff
@@ -1562,6 +1590,9 @@ func (m *Manager) HandleCommunityDescriptionMessage(signer *ecdsa.PublicKey, des
 		return nil, err
 	}
 
+	mutex := m.getMutexForCommunity(types.EncodeHex(id))
```
cammellos: I think it's a bit neater if we create a separate object, that keeps it's own state:
```
type CommunityLock struct {
    locks map[string]*sync.Mutex
    mutex sync.Mutex
}
```

and then you can instance it on messenger and provide a couple of methods:

```
  m.communityLocks.LockCommunity(communityID)
  defer m.communityLocks,UnlockCommunity(communityID)
```

In general, you don't want to let consumer access locks directly (albeit this is a bit different, since it's all messenger, but conceptually there's a function gating access to state)

qfrank: fixed @cammellos

osmaczko (approved):
Thank you!

PR #4674, opened 2024-02-05
Feat: Save and dispatch profile showcase on wallet account change

MishkaRogachev:
For https://github.com/status-im/status-desktop/issues/13080

When a user updates a wallet account using [SaveOrUpdateAccount](https://github.com/status-im/status-go/blob/develop/protocol/messenger_wallet.go#L160-L175), the profile showcase preferences are updated and profile message is dispatched

### DIFF
--- a/protocol/messenger_profile_showcase.go
+++ b/protocol/messenger_profile_showcase.go
@@ -454,7 +454,7 @@ func (m *Messenger) BuildProfileShowcaseFromIdentity(state *ReceivedMessageState
 }
 
 func (m *Messenger) UpdateProfileShowcaseWalletAccount(account *accounts.Account) error {
-	profileAccount, err := m.persistence.GetProfileShowcaseAccountPreferences(account.Address.Hex())
+	profileAccount, err := m.persistence.GetProfileShowcaseAccountPreference(account.Address.Hex())
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


igor-sirotin (approved):

protocol/messenger_profile_showcase.go
```diff
@@ -206,6 +207,10 @@ func (m *Messenger) SetProfileShowcasePreferences(preferences *ProfileShowcasePr
 		return err
 	}
 
+	return m.DispatchProfileShowcase()
+}
+
+func (m *Messenger) DispatchProfileShowcase() error {
```
kounkou, line 213: Hey, just curious, was it really necessary to replace the `m.publishContactCode()` with a function containing a single function call to `m.publishContactCode()` ?

kounkou (approved):
LGTM! Thanks, just a small comment.
Also, Should you squash your commits ? I see 2 commits with same description...

protocol/messenger_profile_showcase.go
MishkaRogachev: In the next PR i'm going to add more `DispatchProfileShowcase` calls in communities and contacts sections. And it's not obvious that `m.publishContactCode` should be invoked to dispatch profile showcase message. And in future, if we decide to separate contact code and profile showcase message it's better to have just one place to modify

PR #4675, opened 2024-02-06
feat: add support for erc1155

alaibe:


services/wallet/router.go
```diff
@@ -42,11 +43,16 @@ const (
 	StickersBuy
 	Bridge
 	ERC721Transfer
+	ERC1155Transfer
 )
 
+func (s SendType) IsNFTTransfer() bool {
```
dlipicar: being pedantic here: ERC1155s are normally called "semi-fungible tokens". "non-fungible tokens" would refer only to ERC721s. I guess "Collectibles" would be the term that includes both? Or whatever better word we can come up with.

dlipicar (approved):

PR #4676, opened 2024-02-06
fix: fix permissions checker

### LINKED ISSUE #12884: Communties: Sharing my address shows fantom permissions which are not met
### Description

I have permission free community (the only condition is that it requires approval to join). As a member of that community, when clicking Request to join, there is a Select addresses to share flow which says that I dont have needed tokens. But so far none is required

Screenshot 2023-11-28 at 12 10 09
Screenshot 2023-11-28 at 12 10 45

https://github.com/status-im/status-desktop/assets/82375995/03922f26-1196-4d3d-8452-bcf3e3a90972

upd: **EXPECTED UI in this case**

image

endulab:
The CheckPermissions returned result was not correct.

Issue [#12884](https://github.com/status-im/status-desktop/issues/12884)

### DIFF
--- a/protocol/communities/permission_checker.go
+++ b/protocol/communities/permission_checker.go
@@ -10,6 +10,9 @@ import (
 
 	"go.uber.org/zap"
 
+	maps "golang.org/x/exp/maps"
+	slices "golang.org/x/exp/slices"
+
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/status-im/status-go/protocol/ens"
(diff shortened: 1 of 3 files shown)


protocol/communities/permission_checker.go
```diff
@@ -106,7 +106,7 @@ func (p *DefaultPermissionChecker) CheckPermissionToJoin(community *Community, a
 
 	accountsAndChainIDs := combineAddressesAndChainIDs(addresses, allChainIDs)
 
-	if len(becomeMemberPermissions) == 0 || len(permissionsToJoin) == 0 {
+	if len(permissionsToJoin) == 0 {
```
endulab: **permissionsToJoin** holds all permissions (admin, member and tokenMaster).
It is sufficient to check the length of it only.
Also, in case there was a tokenMaster permission only, this function returned a default response.

mprakhov: It will not work.
1. Create a community
2. Add admin permission
3. Try to join as a simple member, which does not fit the admin permission

mprakhov (approved):
LGTM 👍

